PR #24336, opened 2025-04-08
[webgpu] fix bias-add

jchen10:
"channels" should be validated before divided by "components". 
"components" should be passed to program inputs and outputs.


jchen10:
The [js-ep](https://github.com/microsoft/onnxruntime/blob/d6df4f29d32ab0632b7b1522aa4f117c36993d7d/js/web/lib/wasm/jsep/webgpu/ops/bias-add.ts#L24) for your reference.

@fs-eire @xhcao PTAL

xhcao:
LGTM

guschmue (approved):

PR #24351, opened 2025-04-08
[QNN EP] Add support for Int64 tensors

quic-ashwshan:
1. Added 'ProcessInt64Tensors' method in BaseOpBuilder to handle common input processing to the graph.
2. Added logic in ProcessOutputs to handle common Cast addition at output.
3. Adds Cast Op at the input to convert to int32 for graph input.
4. Initializers and activation inputs are handled by casting int64_t data to int32_t for QNN compatibility by resizing and copying data.
5. Modified `TransposeOpBuilder` and `GatherOpBuilder`to handle processing outputs.
6. Added unit test for a Reshape op to run with int64 inputs

### DIFF
--- a/onnxruntime/test/providers/qnn/reshape_expand_op_test.cc
+++ b/onnxruntime/test/providers/qnn/reshape_expand_op_test.cc
@@ -238,6 +238,17 @@ TEST_F(QnnHTPBackendTests, Reshape_4D_int32) {
                                      19);  // Opset
 }
 
+// Test that int64 Reshape runs on HTP backend.
+TEST_F(QnnHTPBackendTests, Reshape_4D_int64) {
+  std::vector<int64_t> input_data = {0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11};
+  RunReshapeExpandTestOnHTP<int64_t>("Reshape",
+                                     TestInputDef<int64_t>({1, 3, 2, 2}, false, input_data),
+                                     TestInputDef<int64_t>({3}, true, {1, 1, 12}),
+                                     {},  // Attributes
+                                     ExpectedEPNodeAssignment::All,
+                                     19);  // Opset
+}
+
 // Test QDQ Reshape with a shape value of 0 (copy dimension from input)
 TEST_F(QnnHTPBackendTests, Reshape_4D_0MeansCopy) {
   RunQDQReshapeExpandTestOnHTP<uint8_t>("Reshape",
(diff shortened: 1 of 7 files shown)


onnxruntime/test/providers/qnn/reshape_expand_op_test.cc
```diff
@@ -238,6 +238,17 @@ TEST_F(QnnHTPBackendTests, Reshape_4D_int32) {
                                      19);  // Opset
 }
 
+// Test that int64 Reshape runs on HTP backend.
+TEST_F(QnnHTPBackendTests, Reshape_4D_int64) {
+  std::vector<int64_t> input_data = {0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11};
+  RunReshapeExpandTestOnHTP<int64_t>("Reshape",
+                                     TestInputDef<int64_t>({1, 3, 2, 2}, false, input_data),
```
HectorSVC, line 245: >TestInputDef<int64_t>({1, 3, 2, 2}, false, input_data), [](http://example.com/codeflow?start=37&length=55)

It's OK to cast int64 to int32 for indices input since the value range won't cause overflow. Other than that, it may cause overflow. Will QNN SDK report runtime error if the int64 data out of range for int32 cast?

quic-ashwshan: So, for initializers we are resizing int64 data to int32 before converting the dtype if that is the question.

HectorSVC: not for initializers. Here the first input of the Reshape node is int64 and is dynamic data, so a Cast node will be inserted by this PR. Is there any guarantee that the int64 dynamic input can be cast to int32 without any problem?

quic-ashwshan: I believe the logic that we have is processing all the input tensors and checking for the data type of tensor being int64 and inserting a Cast should not impact even if the input would be dynamic. 
This is happening in the function ProcessInt64Tensors.

HectorSVC: What I meant is: when casting from int64 to int32, if the int64 value is outside the range of int32, the result will be truncated, and data loss will occur. What's the action for the Qnn Cast Op with int64 to int32?

HectorSVC (approved):

HectorSVC:
Please complete the CLA process so that I can merge the PR

quic-ashwshan:

PR #24353, opened 2025-04-08
Merge 'main' into 'win-ort-main' @ 39e585ff2b

mschofie:
### Description
This PR merges the 'main' branch into 'win-ort-main' at commit `39e585ff2b` - the commit with the 1.22 payload that 'win-ort-main' is aligning to. With this commit, 'main' and 'win-ort-main' are identical at that commit.

### Motivation and Context
The process of updating a private fork is to update the win-ort-main branch. So this PR updates the win-ort-main branch to the current payload. Going forward I'd like to go with a 'fork then branch' approach, rather than a 'branch then fork' approach to avoid having to update the win-ort-main branch in the public.


snnn:
We can enable pull request pipelines by adding [win-ort-main](https://github.com/microsoft/onnxruntime/tree/win-ort-main) to the triggers in the yaml files in the .github folder. Would like me to do so?

mschofie:
> We can enable pull request pipelines by adding [win-ort-main](https://github.com/microsoft/onnxruntime/tree/win-ort-main) to the triggers in the yaml files in the .github folder. Would like me to do so?

Thanks for the offer, @snnn, but I think I'd like to (eventually) move away from this public branch (as I suggested in description). We have a private fork, and performing the merges on _that_ side seems less noisy to _this_ repo. We need to accept merges into win-ort-main through the 1.22 release, but hopefully we have an improved process to leverage after that release

PR #24355, opened 2025-04-09
[CI] upgrade x86_64 image for macOS CI

fs-eire:
### Description

Upgrade the x86_64 image for macOS CI from `macos-13` to `macos-15-large` to improve build performance.

See: https://github.com/actions/runner-images/blob/main/README.md#available-images


fs-eire:
Closed as not preferred to use -large images

PR #24356, opened 2025-04-09
[Native WebGPU] Add InstanceNormalization

satyajandhyala:
### Description
Add InstanceNormalization for WebGPU EP.


satyajandhyala:
There seems to be some error with PR taking forever to process newer commits.

PR #24358, opened 2025-04-09
[WebGPU EP] Add EINSUM implementation

feich-ms:
### Description

This PR added the native implementation of einsum operator, based and expanded on existing einsum.ts. All the test cases in einsum_test.cc have been passed.

The equation attribute value of einsum op is a string which consists of left hand side (LHS) and optionally right hand side (RHS) separated by '->'. Ex.

- "ij->ji" matrix transpose
- "ii->i" diagonal elements of a square matrix
- "ij->" sum over all elements of a matrix
- "ij,jk->ik" explicit matrix multiplication
- "ij,jk" implicit matrix multiplication
- "ij,jk->" matrix multiplication and sum over all elements
- "ij,jk,kl->il" three matrix multiplication
- "...ij,...jk->...ik" batched matmul with broadcasting
- ",...i->...i" matrix element multiplication with one scalar
- "abc,cd->abc" keep the original abc matrix shape but matmul and sum over along d

LHS consists of a sequence of terms separated by commas. Each term corresponds to an input variable.
Each symbol corresponds to a dimension in the input variable. The symbol can be either a letter, 'a' to 'z' or 'A' to
'Z' or '...' to represent arbitrary dimensions or empty to represent a scalar.

Empty RHS are handleed differently for implicit vs explicit modes.
- Implicit mode - arrow is not in the equation where the equation "ij,jk" equals to "ij,jk->ik" which is actually a matrix multiplication.
- Explicit mode - arrow is in the equation where the equation "ij,jk->" contains two steps, first step is a matrix multiplication just like the implicit mode, and the second step is to sum up the matrix produced by the first step to a scalar.

For all the test cases, pls refer to einsum_test.cc

### DIFF
--- a/onnxruntime/core/providers/webgpu/math/einsum.cc
+++ b/onnxruntime/core/providers/webgpu/math/einsum.cc
@@ -46,7 +46,7 @@ bool IsInteger(const std::string& s) {
 EinsumEquation::EinsumEquation(const std::vector<const Tensor*>& inputs,
                                const std::string& raw_equation) {
   std::string lhs, rhs, equation = RemoveAllWhitespace(raw_equation);
-  int arrow_pos = equation.find("->");
+  size_t arrow_pos = equation.find("->");
   if (arrow_pos != std::string::npos) {
     lhs = equation.substr(0, arrow_pos);
     rhs = equation.substr(arrow_pos + 2);
@@ -60,8 +60,8 @@ EinsumEquation::EinsumEquation(const std::vector<const Tensor*>& inputs,
   }
 
   // Parse LHS terms.
-  int pos = 0;
-  int find;
+  size_t pos = 0;
+  size_t find;
   int input_idx = 0;
   while ((find = lhs.find(',', pos)) != std::string::npos) {
     auto term = lhs.substr(pos, find - pos);
@@ -190,7 +190,7 @@ EinsumTerm EinsumEquation::ProcessTerm(const std::string& term,
       }
     } else {
       einsum_term.symbol_to_indices[symbol].push_back(
-          i + (has_ellipsis_ ? ellipsis_dims_.size() - 1 : 0));
+          i + (has_ellipsis_ ? static_cast<int>(ellipsis_dims_.size()) - 1 : 0));
       AddSymbol(symbol, static_cast<int>(dims[next_dim++]), index);
     }
   }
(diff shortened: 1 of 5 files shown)


onnxruntime/core/providers/webgpu/math/einsum.cc
```diff
@@ -0,0 +1,370 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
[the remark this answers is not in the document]
feich-ms: lintrunner -a

feich-ms:
@satyajandhyala @xiaofeihan1 @qjia7 @guschmue @fs-eire, pls help to reiview, thanks.

onnxruntime/core/providers/webgpu/math/einsum.cc
```diff
@@ -0,0 +1,467 @@
...
+    return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT,
+                           "Einsum requires at least one input tensor.");
+  }
+
+  std::vector<const Tensor*> input_tensors;
+  for (int i = 0; i < context.InputCount(); ++i) {
+    input_tensors.push_back(context.Input<Tensor>(i));
+  }
+
+  EinsumEquation equation(input_tensors, equation_);
```
fs-eire, line 442: The `equation_` is an attribute, which means it is constant and will not change for each calls to `Einsum::ComputeInternal` in this context. Is there a way to move some preparing steps into the kernel constructor so that we don't have to do this multiple times?

onnxruntime/core/providers/webgpu/math/einsum.cc
```diff
@@ -0,0 +1,467 @@
...
+    shader.MainFunctionBody() << "var outputIndices = " << output.OffsetToIndices("global_idx")
+                              << ";\n";
+  }
+
+  // Define input indices with appropriate types.
+  for (size_t i = 0; i < input_count_; i++) {
+    const auto& input = inputs[i].get();
+    int rank = input.Rank();
+
+    shader.MainFunctionBody() << "var input" << i << "Indices: " << GetWGSLIndicesType(rank)
```
fs-eire, line 409: should not use `GetWGSLIndicesType`. Instead, use "input{i}_indices_t" offerred by indices helper.

feich-ms: done

onnxruntime/core/providers/webgpu/math/einsum.cc
feich-ms: Thanks for the suggestion. I dived deep into the logic and looks like it's not easy to move the eqaution parsing logic into the kernel constructor, the reason is that the equation parsing heavily depends on the input shapes, especially depends on the input dimensions to handle cases with "..." in the equation, like ...ij,...jk->...ik. When parsing the equation from left to right, we need to know the input dimensions to fill up the ellipsis dimensions. Another case that heaily depends on inputs is something like "ij,jk" which need to fill up the output shape implicitly. What we can move into kernel constructor is only the term string parsing which is low cost and then we still loop the parsed term strings and process term in the ComputeInternal function, so I think it makes little sense to make these changes given it brings little improvements. What do you think?

fs-eire: this sounds reasonable for me.

could you please put a comment here mentioning that a potential optimization can be applied in future based on shape inference result?

feich-ms: @fs-eire, done, added comments as above,pls take a look.

fs-eire (approved):
lgtm

PR #24371, opened 2025-04-09
Implement experimental intermediate cross CPU EP allocation

yuslepukhin:
### Description

Make sure QNN EP returns QtpShared based OrtDevice when shared allocator is enabled.
Attempt to use downstream device allocator for allocating the output from the CPU based
upstream node. This works when all the consumers of a particular input share the same CPU based device.

This logic can be made more efficient. We can also take into account other CPU based EPs requirements and see if we could meet them.

### Motivation and Context
CPU Based providers, notable QNN accept CPU based inputs, but they have a requirement of 4K allocations, otherwise the input incurs an extra copy. This is especially noticeable with intermediate values that are produced by upstream CPU based nodes.
This change is experimental.


onnxruntime/core/framework/allocation_planner.cc
```diff
@@ -876,12 +877,50 @@ class PlannerImpl {
...
+                // Find out the index for this input
+                auto hit = std::find_if(input_defs.cbegin(), input_defs.cend(),
+                                        [node_output](const NodeArg* arg) { return node_output->Name() == arg->Name(); });
+                ORT_ENFORCE(hit != input_defs.cend(), "Output not found among consumer inputs");
+                ptrdiff_t input_index = std::distance(input_defs.cbegin(), hit);
+                OrtDevice input_device = consumer_ep->GetOrtDeviceByMemType(consumer_kernel_def->InputMemoryType(narrow<size_t>(input_index)));
+                consumer_devices.insert(input_device);
+              }
+              // All outputs are on the same CPU device
+              if (consumer_devices.size() == 1) {
```
yuslepukhin, line 913: > if (consumer_devices.size() == 1) { [](http://example.com/codeflow?start=13&length=36)

Here, instead of requiring that all consumers be on the same CPU based device, we can select the device with the highest alignment and use it to allocate for all of the downstream consumers. We can also assign priorities.

yuslepukhin:
Please, be advised of this PR: https://github.com/microsoft/onnxruntime/pull/24371

onnxruntime/core/framework/utils.cc
```diff
@@ -72,6 +72,12 @@ bool ProviderIsCpuBased(const std::string& provider_type) {
          provider_type == onnxruntime::utils::kInternalTestingExecutionProvider;
 }
 
+bool ProviderIsCompiledType(const std::string& provider_type) {
```
skottmckay, line 75: is this more a 'provider wants 4K alignment'? 

e.g. coreml is compiled but afaik doesn't require 4K alignment. 

Would it be more flexible to have a map of EP to desired alignment, and we have a more generic allocator that takes the alignment as an arg (instead of one hardcoded CPUAllocatorAligned4K)?

onnxruntime/core/framework/allocator.cc
```diff
@@ -72,9 +72,9 @@ void AllocatorDefaultFree(void* p) {
 }
 
 #else
-void* AllocatorDefaultAlloc(size_t size) {
-  const size_t alignment = MlasGetPreferredBufferAlignment();
-  if (size <= 0) return nullptr;
+
+void* AllocatorDefaultAllocAligned(size_t size, size_t alignment) {
```
skottmckay, line 76: Does there need to be a AllocatorDefaultAllocAligned in the `ifdef USE_MIMALLOC` section?

onnxruntime/core/framework/allocator.cc
```diff
@@ -126,6 +139,24 @@ void* AllocateBufferWithOptions(IAllocator& alloc, size_t size, bool use_reserve
...
+}
+
+bool IsCpuDeviceWithAllocator(const OrtDevice& ort_device) {
+  if (ort_device.Type() == OrtDevice::CPU) {
+    auto mem_type = ort_device.MemType();
+    return mem_type == OrtDevice::MemType::DEFAULT ||
+           mem_type == OrtDevice::MemType::CPU_ALIGNED_4K;
+  }
+  return false;
+}
```
skottmckay, line 158: I don't really understand what a cpu device without an allocator would be, but this doesn't appear to be used anywhere.

include/onnxruntime/core/framework/ortdevice.h
```diff
@@ -26,6 +26,7 @@ struct OrtDevice {
     static const MemoryType HIP_PINNED = 2;
     static const MemoryType CANN_PINNED = 3;
     static const MemoryType QNN_HTP_SHARED = 4;
+    static const MemoryType CPU_ALIGNED_4K = 5;
```
skottmckay, line 29: This doesn't really feel like a new memory type. Is there another way to do this? Would an optional alignment field be clearer? Allocator could honor that if set.

onnxruntime/core/framework/allocation_planner.cc
```diff
@@ -881,7 +887,27 @@ class PlannerImpl {
           if (!node_output->Exists()) continue;
           OrtValueIndex index = Index(node_output->Name());
           ProcessDef(index, node_output);
-          plan_.SetLocation(static_cast<size_t>(index), exec_provider->GetOrtDeviceByMemType(p_kernel_def->OutputMemoryType(i)));
+          OrtDevice output_device = exec_provider->GetOrtDeviceByMemType(p_kernel_def->OutputMemoryType(i));
+          // The below ifdef is due to GetConsumerNodes()
+          // Does minimal build support CPU based providers?
+#if !defined(ORT_MINIMAL_BUILD) || defined(ORT_EXTENDED_MINIMAL_BUILD)
```
skottmckay, line 893: Base minimal build is CPU EP only. There was no usage of consumer nodes in that scenario so GetConsumerNodes was excluded.

onnxruntime/core/framework/allocation_planner.cc
```diff
@@ -881,7 +887,27 @@ class PlannerImpl {
...
+          if (output_device == OrtDevice()) {
```
skottmckay, line 894: This is a confusing comparison without an explanation that `OrtDevice()` == CPU memory. I assume that's the intention of the comparison.

onnxruntime/core/framework/allocation_planner.cc
```diff
@@ -881,7 +887,27 @@ class PlannerImpl {
...
+            const auto& output_name = node_output->Name();
+            const auto consumers = graph_viewer_.GetConsumerNodes(output_name);
+            for (const auto* consumer : consumers) {
+              if (consumer != nullptr) {
+                // Check if the provider is a compiling provider. They handle entire subgraphs at once with many inputs.
+                // Request input type on CPU, if not avaiable it would return the default.
+                const auto& ep_type = consumer->GetExecutionProviderType();
+                if (utils::ProviderIsCompiledType(ep_type)) {
+                  output_device = execution_providers_.Get(ep_type)->GetOrtDeviceByMemType(OrtMemType::OrtMemTypeCPUInput);
+                  break;
```
skottmckay, line 904: Is the key aspect that the provider wants 4K alignment and not whether it compiles? 

If we had an alignment field in OrtDevice could be more generic? get the mem type for the default device, check if it's CPU memory, check if there's an alignment requirement.

onnxruntime/core/framework/utils.cc
yuslepukhin: I agree about the map.

include/onnxruntime/core/framework/ortdevice.h
yuslepukhin: It does feel this way, but it is another memory type. The device type and memory type together are really the keys to request the specific allocator. The provider gives OrtDevice to the allocation planner, it records it into the plan, then the execution frame uses OrtDevice as a key to the map of the allocators.

onnxruntime/core/framework/allocation_planner.cc
yuslepukhin: Yes, I will add a comment here.

onnxruntime/core/framework/allocation_planner.cc
yuslepukhin: Yes, and no. Qnn may return either QtpSharedAllocator device or plain CPU, others currently would return CPU 4K aligned, or NPU allocator.

The key aspect here is to avoid the implicit copy, and the consumer EP knows what kind of allocator it needs depending on the build.

The alignment field alone would not do, because as I explain earlier, we need an OrtDevice instance that would uniquely id the allocator, even if we add alignment to the allocator implementation.

This concept could be extended to CPU -> CUDA allocations and avoid MemCpy node insertions. I.e. CUDAExecutionProvider may return OrtDevice with CUDAPinned memtype. This would cause the upstream output to be allocated in CPU accessible memory, so no copy would be necessary.

include/onnxruntime/core/framework/ortdevice.h
yuslepukhin: It is not only about alignment. Qnn may return a device specific allocator that is CPU accessible.

onnxruntime/core/framework/allocator.cc
yuslepukhin: Removed

onnxruntime/core/framework/utils.cc
yuslepukhin: I will be adding more providers to the list as the details become available. I renamed the function.

include/onnxruntime/core/framework/ortdevice.h
```diff
@@ -62,6 +77,8 @@ struct OrtDevice {
     auto h = std::hash<int>()(device_type);
     onnxruntime::HashCombine(memory_type, h);
     onnxruntime::HashCombine(device_id, h);
+    if (alignment)
+      onnxruntime::HashCombine(*alignment, h);
```
skottmckay, line 81: Would this be more deterministic?
```suggestion
    onnxruntime::HashCombine(alignment ? *alignment : 0, h);
```

include/onnxruntime/core/framework/ortdevice.h
```diff
@@ -71,8 +88,13 @@ struct OrtDevice {
       return device_type < other.device_type;
     if (memory_type != other.memory_type)
       return memory_type < other.memory_type;
+    if (device_id != other.device_id)
+      return device_id < other.device_id;
+
+    if (alignment && other.alignment)
+      return *alignment < *other.alignment;
```
skottmckay, line 95: Do we need a default value if only one instance has alignment set? not perfect as you don't know the implicit default, but maybe better than assuming things are equal. 

related question would be whether we should provide the MLAS default alignment when creating the CPU allocator to make any implicit alignment value explicit so the comparison here is accurate. 

```suggestion
    auto align_to = alignment ? *alignment : 0;
    auto other_align_to = other.alignment ? *other.alignment : 0;
    if (align_to != other_align_to)
      return *alignment < *other.alignment;
```

onnxruntime/core/framework/allocation_planner.cc
```diff
@@ -856,10 +857,15 @@ class PlannerImpl {
                     // we have seen
                     plan_.SetLocation(static_cast<size_t>(index), exec_provider->GetOrtDeviceByMemType(OrtMemType::OrtMemTypeDefault));
                   } else {
-                    // Default the location to CPU
-                    plan_.SetLocation(static_cast<size_t>(index),
-                                      execution_providers_.Get(CPU)->GetOrtDeviceByMemType(OrtMemType::OrtMemTypeDefault));
-                    set_implicitly_consumed_node_arg_has_heterogenous_ep_consumers.insert(index);
+                    if (utils::DoesEpLocationRequiresAdjustment(exec_provider->Type())) {
```
skottmckay, line 860: `DoesEpLocationRequiresAdjustment` feels very allocation planner specific. Does that need to be in the general utils or could it be a local function here? 

Would 'UseEpSpecificCpuAllocator' be clearer? I _think_ that would be the high level description of what you're doing, and that setting a location is a side effect of that.

onnxruntime/core/framework/allocation_planner.cc
```diff
@@ -881,7 +887,27 @@ class PlannerImpl {
           if (!node_output->Exists()) continue;
           OrtValueIndex index = Index(node_output->Name());
           ProcessDef(index, node_output);
-          plan_.SetLocation(static_cast<size_t>(index), exec_provider->GetOrtDeviceByMemType(p_kernel_def->OutputMemoryType(i)));
+          OrtDevice output_device = exec_provider->GetOrtDeviceByMemType(p_kernel_def->OutputMemoryType(i));
+#if !defined(ORT_MINIMAL_BUILD) || defined(ORT_EXTENDED_MINIMAL_BUILD)
+          // Downstream nodes of certain providers may require a CPU accessible location override
+          // to make sure the EP does not incur an unnecessary copy.
```
skottmckay, line 893: Naively 'accessible' would be shared GPU memory (accessible from CPU) vs. dedicated (not accessible). 

Is this about it being a CPU 'accessible location' or that the memory has a specific alignment requirement?

onnxruntime/core/providers/cpu/cpu_execution_provider.cc
```diff
@@ -31,11 +31,28 @@ CPUExecutionProvider::CPUExecutionProvider(const CPUExecutionProviderInfo& info)
...
+  constexpr const bool use_arena_false = false;
+  AllocatorCreationInfo device_info_cpu_aligned_4k{[](int) {
+                                                     return std::make_unique<CPUAllocator>(
+                                                         OrtMemoryInfo(
+                                                             onnxruntime::CPU_ALIGNED_4K, OrtAllocatorType::OrtDeviceAllocator,
+                                                             OrtDevice(OrtDevice::CPU, OrtDevice::MemType::DEFAULT,
+                                                                       DEFAULT_CPU_ALLOCATOR_DEVICE_ID,
+                                                                       kAlloc4KAlignment)));
+                                                   },
+                                                   DEFAULT_CPU_ALLOCATOR_DEVICE_ID, use_arena_false};
```
skottmckay, line 50: Why do we create this in the CPU EP? Naively I would have expected an EP that requires a different alignment to provide a CPU allocator that does that. IIRC the EP registration allows EPs to provide such allocators.

Otherwise, wouldn't the CPU EP need have allocators for all possible alignment values which won't scale?

onnxruntime/core/framework/allocation_planner.cc
yuslepukhin: This could be both or one of them. The point here, is that certain EPs would like to prevent copies and for that they want their input allocated using the allocator of their choice. It could be an aligned memory, or it could something like GPU Pinned. This also depends on the EP configuration.

cmake/onnxruntime_unittests.cmake
```diff
@@ -1277,18 +1277,17 @@ if (NOT onnxruntime_ENABLE_TRAINING_TORCH_INTEROP)
         list(APPEND onnxruntime_perf_test_libs ${android_shared_libs})
       endif()
       if (CMAKE_SYSTEM_NAME MATCHES "AIX")
-        list(APPEND onnxruntime_perf_test_libs onnxruntime_graph onnxruntime_session onnxruntime_providers onnxruntime_framework onnxruntime_util onnxruntime_mlas onnxruntime_optimizer onnxruntime_flatbuffers iconv re2 gtest absl_failure_signal_handler absl_examine_stack absl_flags_parse  absl_flags_usage absl_flags_usage_internal)
-    endif()
+        list(APPEND onnxruntime_perf_test_libs onnxruntime_graph onnxruntime_session onnxruntime_providers onnxruntime_util onnxruntime_framework onnxruntime_optimizer onnxruntime_flatbuffers iconv re2 gtest absl_failure_signal_handler absl_examine_stack absl_flags_parse  absl_flags_usage absl_flags_usage_internal)
```
skottmckay, line 1280: will this work? we now add mlas early on in the dependencies so will that resolve symbols required in onnxruntime_framework? I guess it's working for onnxruntime_common so maybe cmake re-orders things based on dependencies it knows about.

include/onnxruntime/core/framework/ortdevice.h
```diff
@@ -28,31 +30,42 @@ struct OrtDevice {
...
         memory_type(memory_type_),
-        device_id(device_id_) {}
+        device_id(device_id_),
+        alignment(alignment) {}
 
-  constexpr OrtDevice() : OrtDevice(CPU, MemType::DEFAULT, 0) {}
+  // OrtDevice propagates many units and calling Mlas inline would have to change the linkage for many
+  // libs. However, we should save on calling MlasGetPreferredBufferAlignment() everytime we Alloc()
+  OrtDevice(DeviceType device_type_, MemoryType memory_type_, DeviceId device_id_) noexcept
+      : OrtDevice(device_type_, memory_type_, device_id_, ::MlasGetPreferredBufferAlignment()) {}
```
skottmckay, line 42: the MLAS alignment is only relevant to CPU though, so using it as a default for all device types might not be desirable.

but isn't that alignment mainly relevant to the CPU EP, so we could call MlasGetPreferredBufferAlignment as pass it into OrtDevice when constructing an allocator in a CPU based EP? any CPU based EP could override the allocator for default CPU memory (e.g. xnnpack) so if we're going to fallback to CPU EP we still want the alignment to apply.

onnxruntime/core/framework/allocation_planner.cc
```diff
@@ -725,6 +726,13 @@ class PlannerImpl {
       ProcessDef(index, graph_viewer_.GetNodeArg(pair.first));
     }
 
+    // Check if should prefer the destination node EP.
+    auto UseEpSpecificAllocator = [](const std::string& provider_type) {
+      return provider_type == kQnnExecutionProvider ||
+             provider_type == kVitisAIExecutionProvider ||
+             provider_type == kVSINPUExecutionProvider;
+    };
```
skottmckay, line 734: nit: this won't work with fully plugin EPs in the future. can we infer we should use an EP specific allocator by checking the alignment value of the allocator for CPU memory and using the allocator with the highest alignment value?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1320,4 +1320,12 @@ std::vector<AllocatorPtr> QNNExecutionProvider::CreatePreferredAllocators() {
   return allocators;
 }
 
+OrtDevice QNNExecutionProvider::GetOrtDeviceByMemType(OrtMemType /* em_type */) const {
```
skottmckay, line 1323: Does the QNN EP require a 4K CPU allocator? or any CPU memory is always QNN_HTP_SHARED and that will have 4K alignment by default?

onnxruntime/core/providers/shared_library/provider_api.h
```diff
@@ -436,6 +436,8 @@ inline Env& GetDefaultEnv() {
 }
 }  // namespace onnxruntime
 
+size_t MLASCALL MlasGetPreferredBufferAlignment();
```
skottmckay, line 439: What provider bridge EP requires this? I don't see any usage in the plugin ep changes.

cmake/onnxruntime_unittests.cmake
yuslepukhin: I will remove this. OrtDevice is included in many places, including binaries that link to a shared lib. Will set the alignment when the allocator is created or when OrtDevice is returned from EP.

onnxruntime/core/framework/allocation_planner.cc
yuslepukhin: This is not about alignment only. Can be for other reasons.
E.g. Qnn qtp allocator when enabled is not about alignment.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
yuslepukhin: Not any more.

onnxruntime/core/providers/shared_library/provider_api.h
yuslepukhin: It is implicit from OrtDevice ctor.
Said that, We will let EP set that at the point of construction.

onnxruntime/core/providers/qnn/qnn_allocator.cc
```diff
@@ -50,7 +50,7 @@ struct AllocationHeader {
 };
 
 size_t AllocationAlignment() {
-  constexpr size_t min_allocation_alignment = 64;  // Equal to MlasGetPreferredBufferAlignment()
+  const size_t min_allocation_alignment = 64;  // Equal to MlasGetPreferredBufferAlignment()
```
yuslepukhin, line 53: >const  [](http://example.com/codeflow?start=2&length=6)

Undo this

onnxruntime/core/providers/cpu/cpu_execution_provider.cc
```diff
@@ -31,11 +31,14 @@ CPUExecutionProvider::CPUExecutionProvider(const CPUExecutionProviderInfo& info)
     : IExecutionProvider{onnxruntime::kCpuExecutionProvider}, info_{info} {}
 
 std::vector<AllocatorPtr> CPUExecutionProvider::CreatePreferredAllocators() {
+  std::vector<AllocatorPtr> result;
```
yuslepukhin, line 34: >std::v [](http://example.com/codeflow?start=2&length=6)

Undo change

onnxruntime/core/framework/allocation_planner.cc
skottmckay: how will that work with a plugin EP where we can't hardcode the name to get this behavior? doesn't need to be addressed in this PR but we need to understand what is required for this setup to work with a plugin EP if it can't be inferred from properties of the EP's allocator.

skottmckay: still wondering if we can infer this. assuming it only applies to CPU memory

* if MemType differs, prefer the non-default one
  *  e.g. QNN uses OrtDevice::MemType::QNN_HTP_SHARED
  * if both are not default, no preference.
* prefer allocator with higher alignment requirement

onnxruntime/core/providers/qnn/qnn_allocator.cc
```diff
@@ -6,91 +6,94 @@
...
-  return offset;
+std::unique_ptr<void, void (*)(void*)> WrapSharedMemoryWithUniquePtr(void* shared_memory_raw,
+                                                                     const RpcMemApi& rpcmem_api) {
+  return {shared_memory_raw, rpcmem_api.free};
 }
 
-std::byte* GetAllocationHeaderAddress(void* allocation_address) {
-  auto* allocation_header_address = reinterpret_cast<std::byte*>(allocation_address) - sizeof(AllocationHeader);
-  return allocation_header_address;
+class AllocationTracker {
```
skottmckay, line 32: @edgchen1 could you take a look over these changes as I believe you added the original implementation?

onnxruntime/core/providers/qnn/builder/qnn_model.cc
```diff
@@ -200,12 +200,17 @@ static Status BindQnnTensorMemoryToOrtValueMemory(const logging::Logger& logger,
...
-  const bool uses_shared_memory = ort_value_memory_info == HtpSharedMemoryAllocator::AssociatedMemoryInfo();
+  const static auto htp_shared_mem_info = HtpSharedMemoryAllocator::AssociatedMemoryInfo();
+  const bool uses_shared_memory = (ort_value_memory_info.device.Type() == htp_shared_mem_info.device.Type() &&
+                                   ort_value_memory_info.device.MemType() == htp_shared_mem_info.device.MemType());
 
   if (!uses_shared_memory) {
     LOGS(logger, VERBOSE) << "Setting Qnn_Tensor_t clientBuf to ORT tensor memory.";
     SetQnnTensorMemType(qnn_tensor, QNN_TENSORMEMTYPE_RAW);
     SetQnnTensorClientBuf(qnn_tensor, ort_value_data, ort_value_data_size);
+    if ((reinterpret_cast<uintptr_t>(ort_value_data) & (4095)) != 0) {
```
edgchen1, line 211: if there's no code to map the aligned CPU buffer to NPU yet, we wouldn't need to care about alignment here yet, right?

onnxruntime/core/providers/qnn/qnn_allocator.cc
edgchen1: I also made these changes. I'll give them a look but it would be good to have more eyes on them too.

onnxruntime/core/framework/allocation_planner.cc
yuslepukhin: I will have to learn more about plugin EP. 
Off the top of my head, we could come up with a boolean EP interface that would ask us to override.

Simply having a memory allocator that allocates CPU accessible memory would not work. E.g. `CUDAPinned` is CPU accessible, but we are not asked to do this currently (we could).
It also depends on the performance results.

Another idea is to always call into EP and see if it has a CPU accessible preference, we could come up with an `enum` similar to `InputOnCPU`, and the EP returns optional override, we could use it. I think like it best.

yuslepukhin: Nice suggestions with regards to selecting from multiple inputs.
Device allocators typically have a high alignment of 4K (different memtype), we could give those priority.

onnxruntime/core/framework/allocation_planner.cc
```diff
@@ -856,9 +857,22 @@ class PlannerImpl {
...
+                    if (first_device.Type() == OrtDevice::CPU && second_device.Type() == OrtDevice::CPU) {
+                      if (first_device.MemType() == OrtDevice::MemType::DEFAULT &&
+                          second_device.MemType() == OrtDevice::MemType::DEFAULT) {
+                        result = (first_device.GetAlignment() >= second_device.GetAlignment()) ? first_device : second_device;
+                      } else {
+                        result = (first_device.MemType() != OrtDevice::MemType::DEFAULT) ? first_device : second_device;
+                      }
+                    } else {
+                      result = (first_device.Type() == OrtDevice::CPU) ? first_device : second_device;
+                    }
```
skottmckay, line 874: Can we extract the logic comparing two OrtDevice instances from here and below out and use the helpful comments from below in the common code?

skottmckay (approved):

PR #11839, opened 2022-06-13
Skip Constant Folding for ops producing an optional type output

hariharans29:
**Description**: Optional tensors that contain data are tensors at runtime and hence the check that an output OrtValue of an op be a tensor before deciding whether to constant fold the op is not correct. Instead, make the decision based on the proto type of the output NodeArg and ensure that the type is of type tensor before going ahead with constant folding. 

**Motivation and Context**
Fixes https://github.com/microsoft/onnxruntime/issues/11717

PR #24373, opened 2025-04-10
Fix MatmulTransposeFusion when input A and B are the same

### LINKED ISSUE #24341: onnxruntime cannot correctly process the argument of MatMul:  Node input '/Cast_2_output_0' is not a graph input, initializer, or output of a previous node.
### Describe the issue

For the following onnx model, it can be correctly checked by onnx.checker.check_model. In this model, MatMul takes the output of Cast as inputs. 
Image

However, onnxruntime crashes when we run this model.
```c
Traceback (most recent call last):
  File "/home/carla/Documents/test_onnxruntime/0408/onnx_output51775/test.py", line 22, in <module>
    ort_session = onnxruntime.InferenceSession(
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/carla/anaconda3/envs/nnsmith-onnxruntime/lib/python3.11/site-packages/onnxruntime/capi/onnxruntime_inference_collection.py", line 472, in __init__
    self._create_inference_session(providers, provider_options, disabled_optimizers)
  File "/home/carla/anaconda3/envs/nnsmith-onnxruntime/lib/python3.11/site-packages/onnxruntime/capi/onnxruntime_inference_collection.py", line 544, in _create_inference_session
    sess.initialize_session(providers, provider_options, disabled_optimizers)
onnxruntime.capi.onnxruntime_pybind11_state.InvalidArgument: [ONNXRuntimeError] : 2 : INVALID_ARGUMENT : Invalid model. Node input '/Cast_2_output_0' is not a graph input, initializer, or output of a previous node.
```

### To reproduce

Environment
OS: Ubuntu 20.04
onnxruntime: 1.22.0.dev20250402004

Steps to reproduce
This bug can be reproduced by the following code with the model in the attachment.
```python
import onnx
import onnxruntime


onnx_model = onnx.load("model.onnx")
onnx.checker.check_model(onnx_model)

ort_session = onnxruntime.InferenceSession(
                 onnx_model.SerializeToString(), providers=["CPUExecutionProvider"]
              )
```

[model.zip](https://github.com/user-attachments/files/19643035/model.zip)

### Urgency

_No response_

### Platform

Linux

### OS Version

Ubuntu 20.04

### ONNX Runtime Installation

Built from Source

1.22.0.dev20250402004

### ONNX Runtime API

Python

### Architecture

X64

### Execution Provider

Default CPU

_No response_

fs-eire:
### Description

MatmulTransposeFusion does not work correctly when input A and B are the same for a `MatMul` node.

image

Fixes #24341


onnxruntime/test/optimizer/graph_transform_test.cc
```diff
@@ -2946,6 +2946,24 @@ TEST_F(GraphTransformationTests, TransposeMatmulTransBatchNoFusion) {
   }
 }
 
+TEST_F(GraphTransformationTests, GH24341) {
```
hanbitmyths, line 2949: Change a test name like TransposeMatmulSameInputNoFusion to make consistent with other test cases? You can put comment for github issues.

onnxruntime/test/optimizer/graph_transform_test.cc
```diff
@@ -2946,6 +2946,24 @@ TEST_F(GraphTransformationTests, TransposeMatmulTransBatchNoFusion) {
...
+  constexpr const ORTCHAR_T* model_uri = MODEL_FOLDER "fusion/gh_issue_24341.onnx";
```
hanbitmyths, line 2950: Same here. Change the filename like transpose_matmul_same_input_no_fusion.onnx.

onnxruntime/test/optimizer/graph_transform_test.cc
fs-eire: I was following this convention like previously did in https://github.com/microsoft/onnxruntime/pull/11116/files#diff-7ebe64fe3211b862da76388b3a7a3e281a40848bdafab576a9071bc9fbb24b15R35 or https://github.com/microsoft/onnxruntime/pull/11839/files#diff-2061b610e6e8d0211ebb4f304684feb30a1875de130044898377c948e4871512. Not sure if this is the preferred way.

I am OK with either and please let me know what do you think

tianleiwu (approved):

onnxruntime/test/optimizer/graph_transform_test.cc
tianleiwu: How about `TransposeMatmulSameInput_gh_issue_24341` like

https://github.com/microsoft/onnxruntime/blob/6a25fade269d965ff01775440ed6e4d0d25f83da/onnxruntime/test/optimizer/graph_transform_test.cc#L7740

tianleiwu (approved):

PR #24270, opened 2025-04-01
[WebGPU EP] Implements ceil mode for Average Pool 

prathikr:
Ceil mode is required for rtdetr model.

The actual ceil mode calculation is already implemented in the PoolAttributes::ComputeOutputSize() method from pool_attributes.h under CPU EP.

PR #24381, opened 2025-04-10
Remove restrictions for ceil() in AveragePool

### LINKED ISSUE #21206: [Web] `Error: using ceil() in shape computation is not yet supported for AveragePool`
### Describe the issue

When attempting to run https://huggingface.co/onnx-community/rtdetr_r50vd on WebGPU, I receive the following error:
image

### To reproduce

JSFiddle: https://jsfiddle.net/j8uc3k91/

```js
const url = 'https://huggingface.co/onnx-community/rtdetr_r18vd_coco_o365/resolve/main/onnx/model.onnx?download=true';
const session = await ort.InferenceSession.create(url, {
  executionProviders: ['webgpu'],
});

const dims = [1, 3, 640, 640];
const input = new ort.Tensor(
  'float32',
  new Float32Array(dims.reduce((acc, x) => acc * x, 1)),
  dims,
)

// prepare feeds. use model input names as keys.
const feeds = { pixel_values: input };

// feed inputs and run
const results = await session.run(feeds);
console.log(results)
```

### Urgency

Blocks RT-DETR use in Transformers.js w/ WebGPU

### ONNX Runtime Installation

Released Package

1.18.0

### Execution Provider

'webgpu' (WebGPU)

Dexterp37:
### Description
This is a follow-up to #24270 , which seems to wrap up the work. Thanks @prathikr and @fs-eire !

This merely removes the remaining restrictions, allowing the implementation from #24270 to be used.

### Motivation and Context

Fixes #21206

### DIFF
--- a/js/web/lib/onnxjs/backends/webgl/ops/pool.ts
+++ b/js/web/lib/onnxjs/backends/webgl/ops/pool.ts
@@ -47,11 +47,6 @@ export const parseAveragePoolAttributes: OperatorInitialization<AveragePoolAttri
   const strides = node.attributes.getInts('strides', []);
   const pads = node.attributes.getInts('pads', []);
 
-  // TODO: support attribute 'ceil_mode'
-  if (ceilMode !== 0) {
-    throw new Error('using ceil() in shape computation is not yet supported for AveragePool');
-  }
-
   return createAttributeWithCacheKey({ autoPad, ceilMode, countIncludePad, kernelShape, strides, pads });
 };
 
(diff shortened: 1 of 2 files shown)


Dexterp37:

fs-eire:
I don't think WebGL and JSEP actually implemented `ceil_mode`. Removing the checks will only cause the kernel to generate incorrect output shape.

Dexterp37:
> I don't think WebGL and JSEP actually implemented `ceil_mode`. Removing the checks will only cause the kernel to generate incorrect output shape.

Oh, I see - nevermind then! Apologies for the noise, I'm going to close this down !

PR #24387, opened 2025-04-10
[MacOS] Add MLProgram Gather op for CoreML EP

carzh:
### Description
Add MLProgram implementation for Gather

To support this change, I also added handling for converting int64 to int32 in model builder

### DIFF
--- a/onnxruntime/core/providers/coreml/builders/impl/gather_op_builder.cc
+++ b/onnxruntime/core/providers/coreml/builders/impl/gather_op_builder.cc
@@ -32,7 +32,7 @@ int64_t GetAxisAttribute(const Node& node) {
 Status GatherOpBuilder::AddToModelBuilderImpl(ModelBuilder& model_builder, const Node& node,
                                               const logging::Logger& logger) const {
   if (model_builder.CreateMLProgram()) {
-    using namespace CoreML::Specification::MILSpec;  // NOLINT
+    using CoreML::Specification::MILSpec::Operation;
     std::unique_ptr<Operation> op = model_builder.CreateOperation(node, "gather");
 
     std::optional<int32_t> output_datatype;
--- a/onnxruntime/core/providers/coreml/builders/impl/gather_op_builder.cc
+++ b/onnxruntime/core/providers/coreml/builders/impl/gather_op_builder.cc
@@ -40,9 +40,7 @@ Status GatherOpBuilder::AddToModelBuilderImpl(ModelBuilder& model_builder, const
     int32_t input_type;
     ORT_RETURN_IF_NOT(GetType(*node.InputDefs()[0], input_type, logger), "Failed to get input type");
 
-    if (input_type == ONNX_NAMESPACE::TensorProto_DataType_INT64) {
       output_datatype = ONNX_NAMESPACE::TensorProto_DataType_INT32;
-    }
 
     NodeAttrHelper helper(node);
     size_t indices_elem_count = 1;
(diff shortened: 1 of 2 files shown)


onnxruntime/core/providers/coreml/builders/impl/gather_op_builder.cc
```diff
@@ -28,13 +30,37 @@ int64_t GetAxisAttribute(const Node& node) {
...
-  auto layer = model_builder.CreateNNLayer(node);
-  layer->mutable_gather()->set_axis(GetAxisAttribute(node));
-  *layer->mutable_input()->Add() = node.InputDefs()[0]->Name();    // data
-  *layer->mutable_input()->Add() = node.InputDefs()[1]->Name();    // indices
-  *layer->mutable_output()->Add() = node.OutputDefs()[0]->Name();  // output
-  model_builder.AddLayer(std::move(layer));
+                                              const logging::Logger& logger) const {
+  if (model_builder.CreateMLProgram()) {
+    using namespace CoreML::Specification::MILSpec;  // NOLINT
+    std::unique_ptr<Operation> op = model_builder.CreateOperation(node, "gather");
```
edgchen1, line 36: if it's just for referring to one or a few types, can use `using <qualified type name>;`. that's generally preferable to `using namespace x;` which pulls all names into scope and increases the chance for conflicts.
```suggestion
    using CoreML::Specification::MILSpec::Operation;
    std::unique_ptr<Operation> op = model_builder.CreateOperation(node, "gather");
```

onnxruntime/core/providers/coreml/builders/model_builder.cc
```diff
@@ -143,18 +181,15 @@ void CopyOnnxTensorToCoreMLTensor(const ONNX_NAMESPACE::TensorProto& tensor_prot
...
     case ONNX_NAMESPACE::TensorProto_DataType_INT64: {
-      // enable when this is proven to not be the case
-      ORT_THROW(
-          "INT64 is unexpected as CoreML uses 32-bit int for indices. "
-          "Most likely an initializer that should have been skipped was not.");
-      //// from: int64_data/raw, to: longints
-      // if (has_raw_data) {
-      //   CopyRawDataToRepeatedField<int64_t>(tensor_proto, *tensor_value.mutable_longints()->mutable_values());
+      // from: int64_data/raw, to: ints (use narrow to convert to int32)
+      // CoreML uses int32
```
edgchen1, line 185: is this something we want to do for all int64 inputs?

carzh: Ideally, yeah, every CoreML op should have their int64 inputs converted to int32 and their int64 outputs converted to int32. CoreML doesn't support int64 for their ops.

onnxruntime/core/providers/coreml/builders/model_builder.cc
```diff
@@ -71,9 +71,47 @@ void CopyRawDataToRepeatedField(const ONNX_NAMESPACE::TensorProto& tensor_proto,
...
+  // Use std::transform with proper iterators
+  std::transform(data, data + element_count,
+                 google::protobuf::RepeatedFieldBackInserter(&repeated_field),
+                 [&tensor_name](int64_t v) {
+                   try {
+                     return narrow<int32_t>(v);
+                   } catch (const std::exception& e) {
+                     ORT_THROW("Error converting int64 to int32 for tensor ", tensor_name, ": ", e.what(),
+                               ". Value (", v, ") exceeds int32 range.");
+                   }
```
edgchen1, line 94: it's nice that you're including more detail in the message.

however, generally, we try to avoid using `try` or `catch` directly in order to support builds with exceptions disabled. that's why those `ORT_THROW` and `ORT_CATCH` macros are there.

in this case, I think we can just call `narrow` and not bother with adding more info to the exception. if exceptions are enabled, it should be possible to follow a call stack back to this code.

onnxruntime/core/providers/coreml/builders/model_builder.cc
```diff
@@ -71,9 +71,47 @@ void CopyRawDataToRepeatedField(const ONNX_NAMESPACE::TensorProto& tensor_proto,
...
+      ORT_THROW("Error converting int64 to int32: ", e.what(),
+                ". Value (", tensor_proto.int64_data(i), ") exceeds int32 range.");
+    }
+  }
+}
+
 // copy T data from the TensorProto.int32_t field to TensorValue.bytes
 template <typename T>
-void CopyInt32DataToBytes(const ONNX_NAMESPACE::TensorProto& tensor_proto, MILSpec::TensorValue tensor_value) {
+void CopyInt32DataToBytes(const ONNX_NAMESPACE::TensorProto& tensor_proto, MILSpec::TensorValue& tensor_value) {
```
edgchen1, line 114: good catch

edgchen1 (approved):

PR #24389, opened 2025-04-11
[QNN EP] Add support for int64 shape input of Expand Op

quic-tirupath:
### Description
1. Transform INT64 shape of Expand Op to INT32 shape.
2. Add Unit test to check INT64 Shape conversion to INT32 by QNN EP.

### Motivation and Context
QNN doesn't support INT64 shape for Expand Op. This commit delegates the Expand Ops
with INT64 shape on QNN EP. This improves the inference time.


quic-tirupath:

PR #24399, opened 2025-04-11
[Native WebGPU] Support shared memory version of ReduceOps

satyajandhyala:
### Description
 Support shared memory version of ReduceOps

### DIFF
--- a/onnxruntime/core/providers/webgpu/reduction/reduction_ops.cc
+++ b/onnxruntime/core/providers/webgpu/reduction/reduction_ops.cc
@@ -7,6 +7,7 @@
 #include "core/providers/webgpu/data_transfer.h"
 #include "core/providers/webgpu/shader_helper.h"
 #include "core/providers/webgpu/webgpu_supported_types.h"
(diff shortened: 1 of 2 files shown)


guschmue (dismissed):

guschmue:
ReductionOpTest.ReductionVariationTest has some diff

satyajandhyala:
> ReductionOpTest.ReductionVariationTest has some diff

I fixed it and verified locally that it is working. Let me try rerunning the pipeline. Also I will verify again locally.

fs-eire (approved):
I think table driven is better than the previous virtual function approach (GetOpSpecificCode).

There are slightly overhead but it should be OK considering they are only happening when generating shader.

satyajandhyala:
> I think table driven is better than the previous virtual function approach (GetOpSpecificCode).
> 
> There are slightly overhead but it should be OK considering they are only happening when generating shader.

In some cases the table driven is not possible in the case of naive because there is some computation involved.

guschmue (approved):

guschmue:
there is a ut failing on macos that we should fix

PR #24402, opened 2025-04-12
Add ability to disable Model Editor API

skottmckay:
### Description

The OrtModel clashes with a type in the winml public API leading to issues as the linker gets confused. 
As winml doesn't need any model editing capabilities the cleanest solution is to make it easy to exclude Model Editor API.

### Motivation and Context

Resolve type confusion leading to runtime crash in winml

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -255,6 +255,9 @@ option(onnxruntime_USE_OPENVINO_INTERFACE "Build ONNXRuntime shared lib which is
 option(onnxruntime_USE_VITISAI_INTERFACE "Build ONNXRuntime shared lib which is compatible with Vitis-AI EP interface" OFF)
 option(onnxruntime_USE_QNN_INTERFACE "Build ONNXRuntime shared lib which is compatible with QNN EP interface" OFF)
 
+# model editing api. by default disabled in minimal build. also disabled by winml due to a clash between their OrtModel
+# type the the ORT OrtModel type (which is also used in the EP API).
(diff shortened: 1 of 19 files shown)


onnxruntime/core/session/inference_session.cc
```diff
@@ -1225,57 +1225,6 @@ common::Status InferenceSession::Load() {
   return LoadWithLoader(loader, "model_loading_from_saved_proto");
 }
 
-common::Status InferenceSession::Load(const OrtModel& model_editor_api_model) {
```
skottmckay, line 1228: This code was moved to make the #ifdef's a bit cleaner. No changes to the code.

cmake/CMakeLists.txt
```diff
@@ -600,6 +603,14 @@ if (onnxruntime_EXTENDED_MINIMAL_BUILD AND NOT onnxruntime_MINIMAL_BUILD)
   set(onnxruntime_MINIMAL_BUILD ON)
 endif()
 
+if (onnxruntime_MINIMAL_BUILD)
```
jywu-msft, line 606: want to do the same check for onnxruntime_USE_WINML ?
(rather than disabling via build.py)

jywu-msft (approved):

skottmckay:
I think this is a better fix and only requires removing the assert on the ORT side: 

https://github.com/microsoft/win-onnxruntime/pull/37

skottmckay:
This approach resolves the issue instead of avoiding it so this PR shouldn't be needed. #24474

PR #24404, opened 2025-04-13
[webgpu] Fix batch-norm for ort-web-tests

jchen10:
'global_idx' should be used to calculate the output indices.

### DIFF
--- a/onnxruntime/core/providers/webgpu/nn/batch_norm.cc
+++ b/onnxruntime/core/providers/webgpu/nn/batch_norm.cc
@@ -51,8 +51,7 @@ Status BatchNormalizationProgram::GenerateShaderCode(ShaderHelper& shader) const
   const ShaderVariableHelper& output = shader.AddOutput("output", ShaderUsage::UseUniform | ShaderUsage::UseIndicesTypeAlias);
 
   shader.MainFunctionBody() << shader.GuardAgainstOutOfBoundsWorkgroupSizes("uniforms.output_size")
-                            << "  let idx = global_idx * " << components_ << ";\n"
-                            << "  var outputIndices = " << output.OffsetToIndices("idx") << ";\n";
+                            << "  var outputIndices = " << output.OffsetToIndices("global_idx") << ";\n";
   if (spatial_) {
     if (input_tensor.Rank() == 1) {
       shader.MainFunctionBody() << "  let cOffset = 0u;\n";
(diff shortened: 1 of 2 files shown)


jchen10:
@xhcao @fs-eire PTAL
The fix should be right per my understanding and local verification. However, the JS operator has the same [problem](https://github.com/microsoft/onnxruntime/blob/90c263f471bbce724e77d8e62831d3a9fa838b2f/js/web/lib/wasm/jsep/webgpu/ops/batch-norm.ts#L112). I am not clear yet why it can pass the same test. @qjia7 It seems these code lines were originally from you. Please help double check this, thanks!

fs-eire (approved):

jchen10:
> I think this makes sense to me. The definition of "offset" is a little bit different between JSEP and WebGPUEP. In JSEP, "offset" is defined on the "tensor data". In WebGPU EP, "offset" is defined on the "WGSL input/output data". take the following example:
> 
> we have a tensor `x` of shape [2, 3, 4], components = 4.
> 
> tensor's element is float32. the corresponding input/output element in WGSL is vec4.
> 
> The JSEP's offset range is [0, 24), and WebGPUEP's offset range is [0, 6).

Got it, thanks for the clarification.

onnxruntime/core/providers/webgpu/nn/batch_norm.cc
```diff
@@ -51,8 +51,7 @@ Status BatchNormalizationProgram::GenerateShaderCode(ShaderHelper& shader) const
   const ShaderVariableHelper& output = shader.AddOutput("output", ShaderUsage::UseUniform | ShaderUsage::UseIndicesTypeAlias);
 
   shader.MainFunctionBody() << shader.GuardAgainstOutOfBoundsWorkgroupSizes("uniforms.output_size")
-                            << "  let idx = global_idx * " << components_ << ";\n"
-                            << "  var outputIndices = " << output.OffsetToIndices("idx") << ";\n";
+                            << "  var outputIndices = " << output.OffsetToIndices("global_idx") << ";\n";
```
xhcao, line 54: Why `components_` is still needed in Line 60?

jchen10: You are right. It's no longer needed. Thanks for the catch.

fs-eire (approved):

PR #24411, opened 2025-04-14
Do not add LD_LIBRARY_PATH to search path when not defined

clementperon:
If LD_LIBRARY_PATH is not defined a blank "-L" is added to the link command. This causes the next object to be linked to get treated as if it was a search path and causes link failure.

### DIFF
--- a/cmake/onnxruntime_providers_acl.cmake
+++ b/cmake/onnxruntime_providers_acl.cmake
@@ -13,7 +13,9 @@
     onnxruntime_common onnxruntime_framework onnx onnx_proto ${PROTOBUF_LIB} flatbuffers::flatbuffers Boost::mp11 safeint_interface
   )
 
-  target_link_libraries(onnxruntime_providers_acl -L$ENV{LD_LIBRARY_PATH})
+  if (DEFINED ENV{LD_LIBRARY_PATH})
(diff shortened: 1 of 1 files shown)


tianleiwu (approved):

cmake/onnxruntime_providers_acl.cmake
```diff
@@ -13,7 +13,9 @@
     onnxruntime_common onnxruntime_framework onnx onnx_proto ${PROTOBUF_LIB} flatbuffers::flatbuffers Boost::mp11 safeint_interface
   )
 
-  target_link_libraries(onnxruntime_providers_acl -L$ENV{LD_LIBRARY_PATH})
+  if (DEFINED ENV{LD_LIBRARY_PATH})
+    target_link_libraries(onnxruntime_providers_acl -L$ENV{LD_LIBRARY_PATH})
```
tianleiwu, line 17: target_link_libraries does not accept -L options directly. It expects library targets or paths to libraries. 

There are link_directories or set_target_properties(onnxruntime_providers_acl PROPERTIES LINK_FLAGS ...) for that purpose.

Anyway, let me merge this to unblock your build.

snnn: You should use [target_link_directories](https://cmake.org/cmake/help/latest/command/target_link_directories.html)

PR #24412, opened 2025-04-14
Fix typo in option text s/buildings/bindings

clementperon:
Fix typo in option text s/buildings/bindings


fs-eire (approved):

fs-eire:
thank you for the contribution!

PR #24416, opened 2025-04-14
Add session config to return an error if model needs to be compiled

adrianlizarraga:
### Description
Adds session config option (`"session.disable_model_compile"`) that disables model compilation during session initialization.

If this option is set to "1", inference session creation will fail with error code ORT_MODEL_REQUIRES_COMPILATION if compilation is required to run the model on any Execution Provider added to the session. Only the following kinds of models are valid when this option is set to "1":
  - Pre-compiled models that have EPContext nodes for the compiling Execution Providers in the session.
  - Non-compiled models that run only on non-compiling Execution Providers, like CPU EP.

### Example usage
The following example (taken from a unit test) tries to load a model that requires compilation with a session that disables compilation. The session creation fails with error code `ORT_MODEL_REQUIRES_COMPILATION`. Then, the example compiles the model and loads the compiled model successfully.

```C++
  // Taken from a unit test ...

  // Initialize session options with QNN EP
  ProviderOptions provider_options;
  provider_options["backend_type"] = "htp";
  provider_options["offload_graph_io_quantization"] = "0";

  session_options.AppendExecutionProvider("QNN", provider_options);
  session_options.AddConfigEntry(kOrtSessionOptionsDisableEpCompile, "1");  // Disable model compilation!

  // Create an inference session that fails with error ORT_MODEL_REQUIRES_COMPILATION
  try {
    Ort::Session session(*ort_env, input_model_file, session_options);
    FAIL() << "Expected Session creation to fail but it succeeded";  // Should not get here!
  } catch (const Ort::Exception& excpt) {
    OrtErrorCode error_code = excpt.GetOrtErrorCode();
    std::string_view error_msg = excpt.what();
    ASSERT_EQ(error_code, ORT_MODEL_REQUIRES_COMPILATION);
    ASSERT_THAT(error_msg, testing::HasSubstr(kQnnExecutionProvider));
  }

  // Session creation failed because the model was not pre-compiled.
  // Try to compile it now.

  // Create model compilation options from the session options.
  Ort::ModelCompilationOptions compile_options(*ort_env, session_options);
  compile_options.SetInputModelPath(input_model_file);
  compile_options.SetOutputModelPath(output_model_file);

  // Compile the model.
  Ort::Status status = Ort::CompileModel(*ort_env, compile_options);
  ASSERT_TRUE(status.IsOK()) << status.GetErrorMessage();

  // Should be able to create a session with the compiled model and the original session options.
  Ort::Session session(*ort_env, output_model_file, session_options);
```

### Motivation and Context
Compiling models can take a very long time. Want to have a session option that requires input models that do not need to be compiled.


include/onnxruntime/core/session/onnxruntime_session_options_config_keys.h
```diff
@@ -343,3 +343,18 @@ static const char* const kOrtSessionOptionsQDQMatMulNBitsAccuracyLevel = "sessio
...
+// EP registered with the session. Only pre-compiled models (i.e., EPContext models) or models that run on
+// non-compiling EPs are supported if this option is set to "1".
+//
+// See \href https://onnxruntime.ai/docs/execution-providers/EP-Context-Design.html for details about
+// compiled models with EPContext nodes.
+//
+// Option values:
+// - "0": EP compile is not disabled. [DEFAULT]
+// - "1": EP compile is disabled.
+static const char* const kOrtSessionOptionsDisableEpCompile = "session.disable_ep_compile";
```
adrianlizarraga, line 360: Should this be renamed to `session.disable_model_compile`?

adrianlizarraga:
Hi @adrastogi, can you please take a look when able?

onnxruntime/test/providers/qnn/qnn_ep_context_test.cc
```diff
@@ -273,6 +273,62 @@ static void CheckEpContextNodeCounts(void* model_buffer, size_t model_buffer_siz
...
+  so.AddConfigEntry(kOrtSessionOptionsDisableEpCompile, "1");  // Disable model compilation!
+
+  // Create an inference session that fails with error ORT_MODEL_REQUIRES_COMPILATION
+  try {
+    Ort::Session session(*ort_env, input_model_file, so);
+    FAIL() << "Expected Session creation to fail but it succeeded";  // Should not get here!
+  } catch (const Ort::Exception& excpt) {
+    OrtErrorCode error_code = excpt.GetOrtErrorCode();
+    std::string_view error_msg = excpt.what();
+    ASSERT_EQ(error_code, ORT_MODEL_REQUIRES_COMPILATION);
```
adrastogi, line 308: Novice question for my own understanding here: in this test, what is causing ORT to determine that the input model needs compilation? Is it that the model has been constructed in a specific way?

include/onnxruntime/core/session/onnxruntime_session_options_config_keys.h
adrastogi: `session.disable_model_compile` would match the new status code, for what that's worth.

adrastogi (approved):
:shipit:

onnxruntime/test/providers/qnn/qnn_ep_context_test.cc
adrianlizarraga: Good question. There are a few things happening.

The first thing to note is that I constructed the input model so that it is not compiled (i.e., it does **not** have ONNX nodes of type `EPContext`). Secondly, some EPs always compile nodes (e.g., QNN EP) unless the nodes are already compiled. Other EPs, like CPU EP, do not compile and instead just register kernels that correspond to each node in the model.

When the application creates a Session with the input model, ORT tries to assign (partition in ORT terms) model nodes to the EPs enabled in the session. In this unit test, the only EPs are QNN and ORT CPU EP. If any of those EPs needs to compile a node and "ep.disable_ep_compile" is on, then ORT returns an error.

Note that we're not just checking for the presence of `EPContext` nodes. If we only did that, then enabling "ep.disable_ep_compile" would also return an error if the application tried to run a non-compiled with CPU EP, which we would want to allow even if "ep.disable_ep_compile" is turned on.

include/onnxruntime/core/session/onnxruntime_session_options_config_keys.h
adrianlizarraga: Updated

jywu-msft (approved):

PR #24419, opened 2025-04-14
set fp16 abs_error for WebGPU EP in InstanceNorm test

fs-eire:
### Description
set fp16 abs_error for WebGPU EP in InstanceNorm test

Prefer to use `DefaultWebGpuExecutionProvider().get() != nullptr` as condition


fs-eire:
Close in favor of #24420

PR #24421, opened 2025-04-14
[tools] remove unused pipelines in run_CI script

fs-eire:
### Description

remove unused pipelines in run_CI script

### DIFF
--- a/tools/python/run_CIs_for_external_pr.py
+++ b/tools/python/run_CIs_for_external_pr.py
@@ -20,28 +20,11 @@ def get_pipeline_names():
         "Windows ARM64 QNN CI Pipeline",
         "Windows x64 QNN CI Pipeline",
         "Windows GPU Doc Gen CI Pipeline",
-        "ONNX Runtime Web CI Pipeline",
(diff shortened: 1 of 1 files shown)


tools/python/run_CIs_for_external_pr.py
```diff
@@ -20,28 +20,11 @@ def get_pipeline_names():
...
         "Big Models",
-        # android
-        "Linux Android Emulator QNN CI Pipeline",
-        # not currently required, but running these like internal PRs.
-        "Android CI Pipeline",
-        "iOS CI Pipeline",
-        "ONNX Runtime React Native CI Pipeline",
-        "Linux DNNL CI Pipeline",
-        "Linux MIGraphX CI Pipeline",
-        "Linux ROCm CI Pipeline",
```
tianleiwu, line 44: Could we add an option like `--all` to trigger all pipelines including not required ones?

fs-eire: My understanding is that currently the CI pipelines are being migrated to GitHub Actions and only the "required" ones are needed and the other pipelines are either out of dated (replaced with Github Actions) or have known failures.

PR #24428, opened 2025-04-15
[node.js] fix handling null value for externalData

fs-eire:
### Description

Fix a bug when `sessionOptions.externalData === undefined` for Node.js binding.

### DIFF
--- a/js/node/src/session_options_helper.cc
+++ b/js/node/src/session_options_helper.cc
@@ -321,44 +321,46 @@ void ParseSessionOptions(const Napi::Object options, Ort::SessionOptions& sessio
   // external data
   if (options.Has("externalData")) {
     auto externalDataValue = options.Get("externalData");
-    ORT_NAPI_THROW_TYPEERROR_IF(!externalDataValue.IsArray(), options.Env(),
-                                "Invalid argument: sessionOptions.externalData must be an array.");
-    auto externalData = externalDataValue.As<Napi::Array>();
-    std::vector<std::basic_string<ORTCHAR_T>> paths;
-    std::vector<char*> buffs;
-    std::vector<size_t> sizes;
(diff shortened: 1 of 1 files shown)


prathikr (approved):

js/node/src/session_options_helper.cc
```diff
@@ -321,44 +321,46 @@ void ParseSessionOptions(const Napi::Object options, Ort::SessionOptions& sessio
   // external data
   if (options.Has("externalData")) {
     auto externalDataValue = options.Get("externalData");
-    ORT_NAPI_THROW_TYPEERROR_IF(!externalDataValue.IsArray(), options.Env(),
-                                "Invalid argument: sessionOptions.externalData must be an array.");
-    auto externalData = externalDataValue.As<Napi::Array>();
-    std::vector<std::basic_string<ORTCHAR_T>> paths;
-    std::vector<char*> buffs;
-    std::vector<size_t> sizes;
+    if (!externalDataValue.IsNull() && !externalDataValue.IsUndefined()) {
```
satyajandhyala, line 324: Do we need to produce an error or warning if the externalData is specified and there is no data?

fs-eire: in N-API, it's different between "the sessionOptions object does not have 'externalData' property" and "sessionOptions.externalData is `undefined`". We usually don't make a difference for this situation.

If user specify session options like this:
```js
const sessionOptions = {
  // .. other options

  externalData: undefined
};
```

This should still work. The old Node.js binding will fail in this case.

guschmue (approved):

js/node/src/session_options_helper.cc
fs-eire: In other words, I think it is fine if externalData is null or undefined. In this case it is considered there is no external data.

PR #24430, opened 2025-04-15
Add infrastructure for auto EP selection

skottmckay:
### Description

Add infrastructure to enable auto EP selection.

Device discovery for CPU/GPU/NPU on Windows.
Supports internal (CPU/DML/WebGPU) and provider bridge (CUDA) EPs currently.
Infrastructure will be used with plugin EPs next.

Selection policy implementation will be added next, so in the interim there's a temporary function with manually specified selection so unit tests can cover the end-to-end.


include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -4921,12 +4942,244 @@ struct OrtApi {
...
+  const char*(ORT_API_CALL* HardwareDevice_Vendor)(_In_ const OrtHardwareDevice* device);
+
+  /** \brief Get the hardware device's unique identifier.
+   *
+   * \param[in] device The OrtHardwareDevice instance to query.
+   * \return The bus ID of the device.
+   *
+   * \since Version 1.22.
+   */
+  // ??? does the user care about the 'bus' part or could this just be HardwareDevice_Id?
```
skottmckay, line 5126: Does this need to be 'BusId' or would the more generic 'Id' be fine?

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -5050,7 +5303,7 @@ struct OrtCustomOp {
  *
  * See onnxruntime/test/shared_lib/test_model_editor_api.cc for example usage.
  *
- * \since Version 1.21.
+ * \since Version 1.22.
```
skottmckay, line 5306: The Model Editor API wasn't included in 1.21 so updating this value for the types/functions involved to make them correct.

onnxruntime/core/session/environment.cc
```diff
@@ -312,6 +322,11 @@ Internal copy node
     // fire off startup telemetry (this call is idempotent)
     const Env& env = Env::Default();
     env.GetTelemetryProvider().LogProcessInfo();
+
+    // register internal EPs for autoep selection
+    // TODO: ??? Is there any reason not to do this like an EP allocates a large chunk of memory when created?
+    //       If that is the case the user could register by name with no library path to do registration manually.
```
skottmckay, line 328: I think the CPU EP is fine. Any concerns about always creating an instance of the WebGPU or DML EPs if they're included in the build?

jywu-msft: for static EP's right? seems ok to me. we can check how much mem those EP's allocate upon creation

jywu-msft:
bunch of build failures. missing some dxcore header ?

onnxruntime/core/session/ort_env.h
```diff
@@ -41,6 +41,10 @@ struct OrtEnv {
     return *(value_.get());
   }
 
+  onnxruntime::Environment& GetEnvironment() {
+    return *(value_.get());
```
yuslepukhin, line 45: > return *(value_.get()); [](http://example.com/codeflow?start=3&length=24)

why not simply `return *value_;`?

onnxruntime/core/session/ep_library_provider_bridge.cc
```diff
@@ -0,0 +1,69 @@
...
+#include "core/session/abi_devices.h"
+#include "core/session/abi_session_options_impl.h"
+#include "core/session/ep_factory_internal.h"
+
+namespace onnxruntime {
+
+std::unique_ptr<EpFactoryInternal> EpLibraryProviderBridge::CreateCudaEpFactory(Provider& provider) {
+  // Using the name that SessionOptionsAppendExecutionProvider uses to identify the EP as that matches the
+  // expected name in the configuration options. must be static to be valid for the lambdas
+  static const std::string ep_name = "CUDA";
```
skottmckay, line 18: Is it worth adding the 'is supported' and 'create ep' implementation to the library to avoid hardcoding things on the ORT time?

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -4921,12 +4942,244 @@ struct OrtApi {
...
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \since Version 1.22.
+   */
+  ORT_API2_STATUS(SessionOptionsAppendExecutionProvider_V2, _In_ OrtSessionOptions* session_options,
+                  _In_ OrtEnv* env, _In_ const char* ep_name,
+                  _In_reads_(num_op_options) const char* const* ep_option_keys,
+                  _In_reads_(num_op_options) const char* const* ep_option_vals,
+                  size_t num_ep_options);
```
skottmckay, line 5090: Should this take `const OrtEpDevice** ep_devices, size_t num_devices` to support 
- selecting specific EP + device combination
  - e.g. WebGPU or DML could be used with integrated and discrete GPU
- selecting EP with multiple devices
  - e.g. EP can use GPU and NPU

User would get OrtEpDevice pointers from GetEpDevices. In a single call, all devices must have the same EP.

cmake/onnxruntime_common.cmake
```diff
@@ -157,6 +159,9 @@ if(APPLE)
   target_link_libraries(onnxruntime_common PRIVATE "-framework Foundation")
 endif()
 
+if(MSVC)
+  target_link_libraries(onnxruntime_common PRIVATE dxcore.lib)
```
fs-eire, line 163: Does this result to more binary size and runtime dependency?

skottmckay: I would have expected dxcore to be a system library and this is just a shim. dxcore.lib is ~2K.

include/onnxruntime/core/framework/provider_options.h
```diff
@@ -15,4 +15,18 @@ using ProviderOptions = std::unordered_map<std::string, std::string>;
...
+  // return the prefix that is used for provider options when they're in SessionOptions.config_options.
+  // the prefix is not used when ProviderOptions are coming from the user.
+  static std::string GetProviderOptionPrefix(const std::string& name) {
+    return GetProviderOptionPrefix(name.c_str());
+  }
+
+  // return the prefix that is used for provider options when they're in SessionOptions.config_options.
+  // the prefix is not used when ProviderOptions are coming from the user.
+  static std::string GetProviderOptionPrefix(const char* name) {
+    return std::string("ep.") + name + ".";
```
adrianlizarraga, line 28: Should this be made lowercase?

onnxruntime/core/session/abi_session_options_impl.h
```diff
@@ -14,6 +14,14 @@
 
 struct OrtSessionOptions {
   onnxruntime::SessionOptions value;
+
+  // This is only set when we provide an OrtSessionOptions to OrtEpFactory::CreateEp.
+  // In that scenario the InferenceSession has been created with the user provided OrtSessionOptions,
+  // so InferenceSession::GetSessionOptions() is the source of truth and existing_value is set to that.
+  // When set, `value` should be ignored.
+  // `const SessionOptions` as the expected usage in CreateEp does not involve modifying the value.
+  std::optional<const onnxruntime::SessionOptions*> existing_value;
```
adrianlizarraga, line 23: The compile API's `ModelCompilationOptions` makes a copy of the user's original session options. Should ModelCompilationOptions be modified to copy from `existing_value` if it is set?

include/onnxruntime/core/session/onnxruntime_c_api.h
skottmckay: Changed to DeviceId in part 2 PR (#24455). can be debated there.

onnxruntime/core/session/ep_library_provider_bridge.cc
skottmckay: I think so. See #24455 and if that approach seems fine this hardcoding will go away.

include/onnxruntime/core/session/onnxruntime_c_api.h
skottmckay: Added that approach in #24455.

include/onnxruntime/core/framework/provider_options.h
skottmckay: yeah. we want it to be lowercase everywhere for consistency right? i.e. lowercase before prefixing when adding to SessionOptions.config_options.

onnxruntime/core/session/utils.cc
```diff
@@ -5,15 +5,91 @@
...
+// temporary implementation for testing. EP to 'select' is specified in config option
+Status TestAutoSelectEPsImpl(const Environment& env, InferenceSession& sess, const std::string& ep_to_select) {
+  const auto& execution_devices = env.GetOrtEpDevices();
+
+  // Create OrtSessionOptions for the CreateEp call.
+  // Once the InferenceSession is created, its SessionOptions is the source of truth and contains all the values from
+  // the user provided OrtSessionOptions.
+  auto& session_options = sess.GetMutableSessionOptions();
+  OrtSessionOptions ort_so;
+  ort_so.existing_value = &session_options;
```
adrianlizarraga, line 30: Why not copy the `value` instead of storing a pointer to it?
```C++
OrtSessionOptions ort_so;
ort_so.value = session_options;  // copy
```

onnxruntime/core/platform/windows/device_discovery.cc
```diff
@@ -0,0 +1,123 @@
...
+  // Get information about the CPU device
+  auto vendor = CPUIDInfo::GetCPUIDInfo().GetCPUVendor();
+
+  // Create an OrtDevice for the CPU and add it to the list of found devices
+  OrtHardwareDevice cpu_device{OrtHardwareDeviceType_CPU, 0, std::string(vendor), 0};
+
+  found_devices.insert(cpu_device);
+
+  // Get all GPUs and NPUs by querying WDDM/MCDM.
+  wil::com_ptr<IDXCoreAdapterFactory> adapterFactory;
```
RyanUnderhill, line 44: The naming convention is inconsistent here, don't we use snake_case vs camelCase? (my guess is it's Windows code influcence)

onnxruntime/core/session/abi_devices.h
```diff
@@ -0,0 +1,62 @@
...
+struct equal_to<OrtHardwareDevice> {
+  bool operator()(const OrtHardwareDevice& lhs, const OrtHardwareDevice& rhs) const noexcept {
+    return lhs.type == rhs.type &&
+           lhs.vendor_id == rhs.vendor_id &&
+           lhs.vendor == rhs.vendor &&
+           lhs.bus_id == rhs.bus_id &&
+           lhs.metadata.keys == rhs.metadata.keys &&
+           lhs.metadata.values == rhs.metadata.values;
+  }
+};
```
RyanUnderhill, line 50: The compiler can't auto generate operator== here?

include/onnxruntime/core/session/environment.h
```diff
@@ -86,18 +100,76 @@ class Environment {
    * Return an error if an allocator with the same OrtMemoryInfo is already registered.
    * For provider_type please refer core/graph/constants.h
    */
-  Status CreateAndRegisterAllocatorV2(const std::string& provider_type, const OrtMemoryInfo& mem_info, const std::unordered_map<std::string, std::string>& options, const OrtArenaCfg* arena_cfg = nullptr);
+  Status CreateAndRegisterAllocatorV2(const std::string& provider_type, const OrtMemoryInfo& mem_info,
+                                      const std::unordered_map<std::string, std::string>& options,
+                                      const OrtArenaCfg* arena_cfg = nullptr);
+
+  Status RegisterExecutionProviderLibrary(const std::string& registration_name, const ORTCHAR_T* lib_path);
+  Status UnregisterExecutionProviderLibrary(const std::string& registration_name);
```
RyanUnderhill, line 108: Can these use std::string_view?

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -5630,6 +5878,164 @@ struct OrtCompileApi {
...
+ *                          i.e. usage is `factories[0] = new MyEpFactory();`
+ * \param[in] max_factories The maximum number of OrtEpFactory instances that can be added to `factories`.
+ * \param[out] num_factories The number of OrtEpFactory instances created by the factory and added to `factories`.
+ *
+ * \snippet{doc} snippets.dox OrtStatus Return Value
+ *
+ * \since Version 1.22.
+ */
+typedef OrtStatus* (*CreateEpApiFactoriesFn)(_In_ const char* registered_name, _In_ const OrtApiBase* ort_api_base,
+                                             _Inout_ OrtEpFactory** factories, _In_ size_t max_factories,
```
adrianlizarraga, line 5936: nit: Is there some sort of guarantee that we can make about the minimum value of `max_factories`? If the EP library wants to add more factories than we allocated, then they have no real recourse. This is probably not a huge issue because most libraries should only need a few factories but seems to me like an unnecessary constraint.

Potential alternatives:
- EP library can export a function that returns the number of required factories (GetNumOrtEpFactories) and we set `max_factories` to that number
- Pass an OrtAllocator that the EP library can use to allocate the array that holds the factories (max_factories would become an out param)

onnxruntime/core/session/ep_library_internal.h
```diff
@@ -0,0 +1,50 @@
...
+
+#include "core/common/common.h"
+#include "core/session/ep_library.h"
+#include "core/session/ep_factory_internal.h"
+#include "core/session/onnxruntime_c_api.h"
+#include "core/session/provider_bridge_library.h"
+
+namespace onnxruntime {
+
+struct EpLibraryInternal : EpLibrary {
```
adrianlizarraga, line 14: you're probably already planning on doing this after addressing more important things. It would be good to have doc comments on these EpLibrary classes to help reader understand differences between them (EpLibraryInternal vs EpLibraryPlugin vs  EpLibraryProviderBridge )

onnxruntime/core/session/utils.cc
skottmckay: one reason was lifetime. that copy goes out of scope much sooner, so if the EP did keep the OrtSessionOptions* around and attempted to use it later like in CreateEp it would blow up. 

between the cost of the copy (which depends on what things are set in SessionOptions), potential issues with other things in SessionOptions not expecting to be copied, and the lifetime, it seemed safer to point to the 'real' instance.

the scope where this applies is limited as well. SessionOptions getters in the public API (required for OrtEp but not yet), and our interim EpLibrary* code that gets plugged into the 'create EP' path. no other usage of OrtSessionOptions needs to care about `existing_value`. 

would love suggestions of a better name though. not overly happy with `existing_value`.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -402,6 +405,24 @@ typedef enum OrtMemoryInfoDeviceType {
   OrtMemoryInfoDeviceType_FPGA = 2
 } OrtMemoryInfoDeviceType;
 
+typedef enum OrtHardwareDeviceType {
+  OrtHardwareDeviceType_CPU,
+  OrtHardwareDeviceType_GPU,
+  OrtHardwareDeviceType_NPU
+} OrtHardwareDeviceType;
```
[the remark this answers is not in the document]
skottmckay: There's no integrated vs discrete separation for GPU. NPU is Neural Processing Unit not GPU. 

How memory works comes down to an EP that supports a GPU. e.g for CUDA that's may be fairly fixed and always be discrete. for WebGPU it depends on whether it's integrated or discrete. 

I _think_ that means any integrate/discrete distinctions are implementation specific so shouldn't be pushed up to having separate INTEGRATED_GPU and DISCRETE_GPU values here.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -402,6 +405,24 @@ typedef enum OrtMemoryInfoDeviceType {
...
+/** \brief These are the default EP selection policies used by ORT when doing automatic EP selection.
+ */
+typedef enum OrtExecutionProviderDevicePolicy {
+  OrtExecutionProviderDevicePolicy_DEFAULT,
+  OrtExecutionProviderDevicePolicy_PREFER_CPU,
+  OrtExecutionProviderDevicePolicy_PREFER_NPU,
+  OrtExecutionProviderDevicePolicy_PREFER_GPU,
+  OrtExecutionProviderDevicePolicy_MAX_PERFORMANCE,
+  OrtExecutionProviderDevicePolicy_MAX_EFFICIENCY,
+  OrtExecutionProviderDevicePolicy_MIN_OVERALL_POWER,
```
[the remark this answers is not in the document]
skottmckay: Something like that. 'max efficiency' is less clear to me. the real implementation of the selection policy will hopefully clarify. that's coming later.

onnxruntime/core/platform/windows/device_discovery.cc
skottmckay: can cleanup. this was contributed as starter code. for this PR assume this is all replaced and we will make it conform to coding standards when the production ready implementation is added

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -5630,6 +5878,164 @@ struct OrtCompileApi {
...
+  const char*(ORT_API_CALL* GetName)(const OrtEp* this_ptr);
+
+  // OrtStatus* GetCapability(OrtEp* ep, const OrtGraph* graph,
+  //                          size_t* num_supported_subgraphs,
+  //                          OrtIndexedSubgraph** supported_subgraphs, OrtAllocator* allocator);
+
+  // OrtStatus* Compile(OrtEp* ep, const OrtGraph** graphs, OrtNode** fused_graph_nodes,
+  //                    size_t count, OrtNodeComputeInfo* node_compute_infos);
+
+  //  Many other functions!
```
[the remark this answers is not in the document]
skottmckay: This will be the next step. Whole new feature rather than simple TODO, but I can add one if preferred.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -4931,6 +4952,233 @@ struct OrtApi {
...
+   *         Note: ORT owns the instance so the user must not call ReleaseKeyValuePairs with it.
+   *
+   * \since Version 1.22.
+   */
+  const OrtKeyValuePairs*(ORT_API_CALL* HardwareDevice_Metadata)(_In_ const OrtHardwareDevice* device);
+
+  /** \brief Get the execution provider name.
+   *
+   * \param[in] ep_device The OrtEpDevice instance to query.
+   * \return The execution provider name.
```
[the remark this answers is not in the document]
skottmckay: Will clarify. It does not.

include/onnxruntime/core/session/onnxruntime_c_api.h
skottmckay: wanted to keep it simple given we don't expect a mass of factories, and this was entirely about avoiding the complexity of passing in allocators. in reality, one factory per device type is what I would expect to be the max. and we have 3 device types. 

would it be ok to document that currently there are always 4 available? or we can make that 8 if we want to go nuts. it's cheap to do so. avoids extra things in the EP API and lifetime management complexity.

include/onnxruntime/core/session/onnxruntime_c_api.h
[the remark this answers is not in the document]
skottmckay: Isn't that implied by it returning a const char*?

onnxruntime/core/platform/windows/device_discovery.cc
skottmckay: will cleanup in the 'part 2' changes rather than here.

onnxruntime/core/session/abi_session_options_impl.h
skottmckay: I don't think it needs to be aware. The only diff between the original OrtSessionOptions and InferenceSession::session_options_ would be the default ep_options from OrtEpDevice for selected EPs.

onnxruntime/core/session/abi_devices.h
skottmckay: It cannot. Even if I add an `operator==` to OrtKeyValuePairs it's not happy.

include/onnxruntime/core/session/environment.h
skottmckay: they could, but it doesn't end up buying you anything as it needs to be std::string to do the internal lookups like `ep_libraries_count(registration_name)`. we need a std::string in the key of that to guarantee it remains valid. 

so if I start with a std::string there's an indirection to string_view and I have to construct a new std::string inside RegisterExecutionProviderLibrary (for example), costing me at least one extra std::string allocation.

include/onnxruntime/core/session/onnxruntime_c_api.h
adrianlizarraga: Yeah, documenting that there at least 8 (or 4) available seems reasonable to me. Thanks.

onnxruntime/core/session/utils.cc
adrianlizarraga: We're passing `&ort_so` (OrtSessionOptions) to CreateEp. `ort_so` is declared on the stack, so it will also go away after this function returns. Wouldn't the same lifetime issue apply in the current case as well?

include/onnxruntime/core/session/onnxruntime_c_api.h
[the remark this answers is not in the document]
jywu-msft: yeah, unclear. preferring NPU to me implies preferring efficiency vs max perf (GPU)

onnxruntime/core/session/utils.cc
adrianlizarraga: >potential issues with other things in SessionOptions not expecting to be copied

I was concerned about this too, but it seems to be safe to copy SessionOptions for at least two reasons:
- The `InferenceSession` class always makes a copy of the `SessionOptions` passed to its constructor: https://github.com/microsoft/onnxruntime/blob/689ec470f1d959e841178aa3a1139b89d45903cb/onnxruntime/core/session/inference_session.cc#L316
- There's a public C API function that makes a copy of OrtSessionOptions: https://github.com/microsoft/onnxruntime/blob/689ec470f1d959e841178aa3a1139b89d45903cb/include/onnxruntime/core/session/onnxruntime_c_api.h#L928

>the cost of the copy

The SessionOptions struct is only 824 bytes. CPU will chew through that :) (especially since this is a one time copy upfront).

> lifetime issues

As I mentioned above, the current implementation still has lifetime issues because the `ort_so` is on the stack. Maybe the right solution is to update InferenceSession to store a `OrtSessionOptions` instead of a `onnxruntime::SessionOptions`. This way we can just directly pass its address to CreateEp() and it will live as long as the session, which is what I think the intent is.

include/onnxruntime/core/session/onnxruntime_c_api.h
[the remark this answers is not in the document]
skottmckay: Added TODO

include/onnxruntime/core/session/onnxruntime_c_api.h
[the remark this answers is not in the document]
skottmckay: Will leave this as is and we can refine with the PR to add the selection policy implementation if that's OK.

onnxruntime/core/session/utils.cc
skottmckay: > We're passing `&ort_so` (OrtSessionOptions) to CreateEp. `ort_so` is declared on the stack, so it will also go away after this function returns. Wouldn't the same lifetime issue apply in the current case as well?

The pointer to a value in InferenceSession should be valid for the entire duration of the EP's usage as we call CreateEp (or CreateIExecutionProvider) on a per-session basis. We release the EP before the InferenceSession goes away.

onnxruntime/core/session/ep_api_utils.h
```diff
@@ -0,0 +1,42 @@
...
+// Licensed under the MIT License.
+
+#include "core/session/onnxruntime_c_api.h"
+
+namespace onnxruntime {
+// helper to forward a call from the C API to an instance of the factory implementation.
+// used by EpFactoryInternal and EpFactoryProviderBridge.
+template <typename TFactory>
+struct ForwardToFactory {
+  static const char* GetFactoryName(const OrtEpFactory* this_ptr) {
```
adrianlizarraga, line 11:
```suggestion
  static const char* ORT_API_CALL GetFactoryName(const OrtEpFactory* this_ptr) {
```

onnxruntime/core/session/ep_api_utils.h
```diff
@@ -0,0 +1,42 @@
...
+    return static_cast<const TFactory*>(this_ptr)->GetName();
+  }
+
+  static const char* GetVendor(const OrtEpFactory* this_ptr) {
```
adrianlizarraga, line 15:
```suggestion
  static const char* ORT_API_CALL GetVendor(const OrtEpFactory* this_ptr) {
```

onnxruntime/core/session/ep_api_utils.h
```diff
@@ -0,0 +1,42 @@
...
+    return static_cast<const TFactory*>(this_ptr)->GetVendor();
+  }
+
+  static bool GetDeviceInfoIfSupported(const OrtEpFactory* this_ptr,
+                                       const OrtHardwareDevice* device,
+                                       OrtKeyValuePairs** ep_device_metadata,
+                                       OrtKeyValuePairs** ep_options_for_device) {
```
adrianlizarraga, line 22:
```suggestion
  static bool ORT_API_CALL GetDeviceInfoIfSupported(const OrtEpFactory* this_ptr,
                                                    const OrtHardwareDevice* device,
                                                    OrtKeyValuePairs** ep_device_metadata,
                                                    OrtKeyValuePairs** ep_options_for_device) {
```

onnxruntime/core/session/ep_api_utils.h
```diff
@@ -0,0 +1,42 @@
...
+    return static_cast<const TFactory*>(this_ptr)->GetDeviceInfoIfSupported(device, ep_device_metadata,
+                                                                            ep_options_for_device);
+  }
+
+  static OrtStatus* CreateEp(OrtEpFactory* this_ptr,
```
adrianlizarraga, line 27:
```suggestion
  static OrtStatus* ORT_API_CALL CreateEp(OrtEpFactory* this_ptr,
```

onnxruntime/core/session/ep_api_utils.h
```diff
@@ -0,0 +1,42 @@
...
+                             const OrtHardwareDevice* const* devices,
+                             const OrtKeyValuePairs* const* ep_metadata_pairs,
+                             size_t num_devices,
+                             const OrtSessionOptions* session_options,
+                             const OrtLogger* logger,
+                             OrtEp** ep) {
```
adrianlizarraga, line 33:
```suggestion
  static OrtStatus* ORT_API_CALL CreateEp(OrtEpFactory* this_ptr,
                                          const OrtHardwareDevice* const* devices,
                                          const OrtKeyValuePairs* const* ep_metadata_pairs,
                                          size_t num_devices,
                                          const OrtSessionOptions* session_options,
                                          const OrtLogger* logger,
                                          OrtEp** ep) {
```

onnxruntime/core/session/ep_api_utils.h
```diff
@@ -0,0 +1,42 @@
...
+    return static_cast<TFactory*>(this_ptr)->CreateEp(devices, ep_metadata_pairs, num_devices,
+                                                      session_options, logger, ep);
+  }
+
+  static void ReleaseEp(OrtEpFactory* this_ptr, OrtEp* ep) {
```
adrianlizarraga, line 38:
```suggestion
  static void ORT_API_CALL ReleaseEp(OrtEpFactory* this_ptr, OrtEp* ep) {
```

onnxruntime/test/autoep/library/example_plugin_ep.cc
```diff
@@ -0,0 +1,160 @@
...
+struct ExampleEpFactory : OrtEpFactory, ApiPtrs {
+  ExampleEpFactory(const char* ep_name, ApiPtrs apis) : ApiPtrs(apis), ep_name_{ep_name} {
+    GetName = GetNameImpl;
+    GetVendor = GetVendorImpl;
+    GetDeviceInfoIfSupported = GetDeviceInfoIfSupportedImpl;
+    CreateEp = CreateEpImpl;
+    ReleaseEp = ReleaseEpImpl;
+  }
+
+  static const char* GetNameImpl(const OrtEpFactory* this_ptr) {
```
adrianlizarraga, line 52:
```suggestion
  static const char* ORT_API_CALL GetNameImpl(const OrtEpFactory* this_ptr) {
```

onnxruntime/test/autoep/library/example_plugin_ep.cc
```diff
@@ -0,0 +1,160 @@
...
+    const auto* ep = static_cast<const ExampleEpFactory*>(this_ptr);
+    return ep->ep_name_.c_str();
+  }
+
+  static const char* GetVendorImpl(const OrtEpFactory* this_ptr) {
```
adrianlizarraga, line 57:
```suggestion
  static const char* ORT_API_CALL GetVendorImpl(const OrtEpFactory* this_ptr) {
```

onnxruntime/test/autoep/library/example_plugin_ep.cc
```diff
@@ -0,0 +1,160 @@
...
+    const auto* ep = static_cast<const ExampleEpFactory*>(this_ptr);
+    return ep->vendor_.c_str();
+  }
+
+  static bool GetDeviceInfoIfSupportedImpl(const OrtEpFactory* this_ptr,
+                                           const OrtHardwareDevice* device,
+                                           _Out_opt_ OrtKeyValuePairs** ep_metadata,
+                                           _Out_opt_ OrtKeyValuePairs** ep_options) {
```
adrianlizarraga, line 65:
```suggestion
  static bool ORT_API_CALL GetDeviceInfoIfSupportedImpl(const OrtEpFactory* this_ptr,
                                                        const OrtHardwareDevice* device,
                                                        _Out_opt_ OrtKeyValuePairs** ep_metadata,
                                                        _Out_opt_ OrtKeyValuePairs** ep_options) {
```

onnxruntime/test/autoep/library/example_plugin_ep.cc
```diff
@@ -0,0 +1,160 @@
...
+    return false;
+  }
+
+  static OrtStatus* CreateEpImpl(OrtEpFactory* this_ptr,
+                                 _In_reads_(num_devices) const OrtHardwareDevice* const* /*devices*/,
+                                 _In_reads_(num_devices) const OrtKeyValuePairs* const* /*ep_metadata*/,
+                                 _In_ size_t num_devices,
+                                 _In_ const OrtSessionOptions* session_options,
+                                 _In_ const OrtLogger* logger,
+                                 _Out_ OrtEp** ep) {
```
adrianlizarraga, line 89:
```suggestion
  static OrtStatus* ORT_API_CALL CreateEpImpl(OrtEpFactory* this_ptr,
                                              _In_reads_(num_devices) const OrtHardwareDevice* const* /*devices*/,
                                              _In_reads_(num_devices) const OrtKeyValuePairs* const* /*ep_metadata*/,
                                              _In_ size_t num_devices,
                                              _In_ const OrtSessionOptions* session_options,
                                              _In_ const OrtLogger* logger,
                                              _Out_ OrtEp** ep) {
```

onnxruntime/test/autoep/library/example_plugin_ep.cc
```diff
@@ -0,0 +1,160 @@
...
+    // const OrtHardwareDevice* device = devices[0];
+    // const OrtKeyValuePairs* ep_metadata = ep_metadata[0];
+
+    auto dummy_ep = std::make_unique<ExampleEp>(*factory, factory->ep_name_, *session_options, *logger);
+
+    *ep = dummy_ep.release();
+    return nullptr;
+  }
+
+  static void ReleaseEpImpl(OrtEpFactory* /*this_ptr*/, OrtEp* ep) {
```
adrianlizarraga, line 115:
```suggestion
  static void ORT_API_CALL ReleaseEpImpl(OrtEpFactory* /*this_ptr*/, OrtEp* ep) {
```

adrianlizarraga (commented):
Adding `ORT_API_CALL` to the factory functions should hopefully fix the x86 ci errors

onnxruntime/core/session/environment.cc
skottmckay: yeah - static EPs only. can refine how they behave in a separate PR if needed.

include/onnxruntime/core/session/onnxruntime_c_api.h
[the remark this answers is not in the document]
RyanUnderhill: It is, but saying the lifetime requirements is nice to be sure about.

onnxruntime/core/session/ep_library.cc
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include "core/session/ep_library.h"
+
+#include "core/framework/provider_options.h"
+#include "core/session/abi_session_options_impl.h
```
adrianlizarraga, line 7:
```suggestion
#include "core/session/abi_session_options_impl.h"
```

onnxruntime/test/util/include/asserts.h
```diff
@@ -4,7 +4,7 @@
 #pragma once
 
 #include "core/common/status.h"
-#include "core/session/onnxruntime_c_api.h"
+#include "core/session/onnxruntime_cxx_api.h"
```
adrianlizarraga, line 7: I believe this is causing the following error: `libonnxruntime_providers_cuda_ut.so: undefined symbol: OrtGetApiBase`

The test/providers/cuda/test_cases/reduction_functions_test.cc file (part of onnxruntime_providers_cuda_ut) [includes this asserts.h file](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/test/providers/cuda/test_cases/reduction_functions_test.cc#L13). Now that onnxruntime_cxx_api.h is included transitively, it needs to link with onnxruntime_session because the C++ api header calls [OrtBaseApi()->GetApi() to initialize the global api_ variable.](https://github.com/microsoft/onnxruntime/blob/main/include/onnxruntime/core/session/onnxruntime_cxx_api.h#L118)

onnxruntime/core/session/utils.cc
skottmckay: Adrian made a great point that the OrtSessionOptions instance is also scoped to this function. Removed existing_value and updated doc for CreateEp to call out that the OrtSessionOptions instance it gets should not be saved for later use.

onnxruntime/test/util/include/asserts.h
skottmckay: Thanks! Split out the assert headers to avoid this dependency as the new usage for the OrtStatus asserts is much simpler so I think it's worth maintaining.

cmake/onnxruntime_unittests.cmake
```diff
@@ -1798,6 +1799,70 @@ if (NOT CMAKE_SYSTEM_NAME STREQUAL "Emscripten" AND (NOT onnxruntime_MINIMAL_BUI
...
+  if (onnxruntime_USE_DML)
+    list(APPEND onnxruntime_autoep_test_LIBS d3d12.lib)
+  endif()
+
+  if (CPUINFO_SUPPORTED)
+    list(APPEND onnxruntime_autoep_test_LIBS cpuinfo)
+  endif()
+
+  if (CMAKE_SYSTEM_NAME MATCHES "AIX")
+    list(APPEND onnxruntime_shared_lib_test_LIBS onnxruntime_graph onnxruntime_session onnxruntime_providers
```
adrianlizarraga, line 1853:
```suggestion
    list(APPEND onnxruntime_autoep_test_LIBS onnxruntime_graph onnxruntime_session onnxruntime_providers
```

jywu-msft (approved):

skottmckay:
Will follow up any comments in separate PR

cmake/onnxruntime_common.cmake
snnn: onnxruntime_common is a static library which does not have a link step. So how would this line of code help?

PR #24433, opened 2025-04-15
Support canonical EP names in SessionOptionsAppendExecutionProvider

adrianlizarraga:
### Description
Updates the `SessionOptionsAppendExecutionProvider` function to also support full canonical provider names (e.g., QNNExecutionProvider) in addition to the short names (e.g., QNN).

### Motivation and Context
There's an inconsistency in how ORT names EPs. The `SessionOptionsAppendExecutionProvider` C API function uses short names (e.g., "QNN"), but other ORT APIs use longer names (e.g., "QNNExecutionProvider").
- Python APIs to add EP to session uses "QNNExecutionProvider" (not "QNN")
- Python and C APIs to GetAvailableProviders use "QNNExecutionProvider"
- Internal ORT code uses "QNNExecutionProvider" to assign nodes to cuda ep.
- Only `SessionOptionsAppendExecutionProvider` uses short names like "QNN".


onnxruntime/core/providers/webgpu/webgpu_provider_options.h
```diff
@@ -9,28 +9,28 @@ namespace options {
 
 // The following are the options that can be set in the WebGPU provider options.
 
-constexpr const char* kPreferredLayout = "ep.WebGPU.preferredLayout";
-constexpr const char* kEnableGraphCapture = "ep.WebGPU.enableGraphCapture";
+constexpr const char* kPreferredLayout = "ep.WebGpuExecutionProvider.preferredLayout";
```
adrianlizarraga, line 12: Hi @satyajandhyala, want to make sure that us changing the prefix for these configs won't cause any issues.

We updated the SessionOptionsAppendExecutionProvider API function to use this new prefix: `ep.<EP_NAME>.`.

fs-eire: If user creates WebGPU EP via C-API `SessionOptionsAppendExecutionProvider()`, this change should be transparent to them, right?

adrianlizarraga: Yes, this change will be transparent to user of `SessionOptionsAppendExecutionProvider`

fs-eire: thank you for confirming. I think we didn't use `AddSessionConfigEntry` to add WebGPU EP specific options. So this change should be safe.

onnxruntime/core/session/provider_registration.cc
```diff
@@ -81,6 +84,41 @@ ORT_API_STATUS_IMPL(OrtApis::SessionOptionsAppendExecutionProvider,
                     _In_reads_(num_keys) const char* const* provider_options_values,
                     _In_ size_t num_keys) {
   API_IMPL_BEGIN
+  enum class EpID {
+    INVALID = 0,
+    DML,
+    QNN,
+    OpenVINO,
+    SNPE,
```
skottmckay, line 92: is SNPE still enabled? I thought that was replaced by QNN

onnxruntime/core/session/provider_registration.cc
```diff
@@ -108,11 +146,45 @@ ORT_API_STATUS_IMPL(OrtApis::SessionOptionsAppendExecutionProvider,
...
+
+  const EpToAppend& ep_to_append = *ep_to_append_iter;
+  ORT_ENFORCE(ep_to_append.id != EpID::INVALID);
+
   // Add provider options to the session config options.
   // Use a new key with the format: "ep.<EP_NAME>.<PROVIDER_OPTION_KEY>"
   for (const auto& [key, value] : provider_options) {
     std::ostringstream new_key_builder;
-    new_key_builder << "ep." << provider_name << "." << key;
+    new_key_builder << "ep." << ep_to_append.canonical_name << "." << key;
```
skottmckay, line 187: Could we make the prefix all lowercase so it's not case sensitive? 

Longer term these values will come from plugin EPs and a case-insensitive treatment might be slightly more user-friendly.

skottmckay (dismissed):

onnxruntime/core/session/provider_registration.cc
adrianlizarraga: QNN EP replaces SNPE. I removed mention of SNPE from the `SessionOptionsAppendExecutionProvider` C API function documentation. However, I did not remove SNPE from the implementation (left a TODO to remove SNPE entirely).

onnxruntime/core/session/provider_registration.cc
adrianlizarraga: Yeah, I made this lowercase now.

onnxruntime/core/providers/qnn/builder/qnn_utils.h
```diff
@@ -26,6 +27,22 @@ class QnnOpConfigWrapper;
 class QnnModelWrapper;
 
 namespace utils {
+/**
+ * Returns a lowercase version of the input string.
+ * /param str The string to lowercase.
+ * /return The lowercased string.
+ */
+inline std::string GetLowercaseString(std::string str) {
```
adrianlizarraga, line 35: Note: this is a duplicate of the function in core/common/string_utils.h (which is internal to ORT). I don't see a need to add this simple utility function to the provider bridge api.

skottmckay (approved):

PR #24435, opened 2025-04-15
[QNN EP] Enable QnnGpu backend in QNN EP.

johnpaultaken:
### Description
Enable the GPU backend also for the onnxruntime QNN EP.

### Motivation and Context
Why is this change required? What problem does it solve? 
It allows QNN EP to run on the GPU backend also.
With this change many models can now run fully on QNN EP GPU backend, like resnet_50, google_vit_base_fp32, squeezenet1.0-7 etc. Also the onnxruntime node tests and versioned operator tests pass numbers for the GPU is comparable to the HTP now. 
Note: Currently QNN_LOG_LEVEL_DEBUG need to be enabled to run correctly.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
@@ -8,8 +8,7 @@
 #include <string>
 #include "QnnOpDef.h"
 #include "CPU/QnnCpuCommon.h"
-// TODO: not exist for Windows yet
-// #include "GPU/QnnGpuCommon.h"
+#include "GPU/QnnGpuCommon.h"
 #include "DSP/QnnDspCommon.h"
 #include "HTP/QnnHtpCommon.h"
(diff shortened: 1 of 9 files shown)


onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -8,6 +8,7 @@
 #include <string>
 #include "QnnOpDef.h"
 #include "CPU/QnnCpuCommon.h"
+#include "GPU/QnnGpuCommon.h"
 // TODO: not exist for Windows yet
 // #include "GPU/QnnGpuCommon.h"
```
HectorSVC, line 13: Please remove these comments

johnpaultaken: will do, thanks.

HectorSVC (approved):
:shipit:

PR #24207, opened 2025-03-27
[draft] Add API to compile a model

adrianlizarraga:
### Description
- Adds C/C++ API functionality to compile a model (i.e., generate a model with EPContext nodes) using explicit APIs.
- Adds support for compiling when input or output models are in memory (not just files).
- Allows specifying the threshold for when initializers are stored in an external file.

List of C API functions:
```C++
  ORT_API2_STATUS(CreateModelCompilationOptions, _In_ const OrtEnv* env, _Outptr_ OrtModelCompilationOptions** out);
  ORT_API2_STATUS(CreateModelCompilationOptionsFromSessionOptions, _In_ const OrtEnv* env,
                  _In_ OrtSessionOptions* session_options, _Outptr_ OrtModelCompilationOptions** out);
  ORT_API2_STATUS(ModelCompilationOptions_SetInputModelPath, _In_ OrtModelCompilationOptions* model_compile_options,
                  _In_ const ORTCHAR_T* input_model_path);
  ORT_API2_STATUS(ModelCompilationOptions_SetInputModelFromBuffer, _In_ OrtModelCompilationOptions* model_compile_options,
                  _In_ const void* input_model_data, size_t input_model_data_size);
  ORT_API2_STATUS(ModelCompilationOptions_SetOutputModelPath, _In_ OrtModelCompilationOptions* model_compile_options,
                  _In_ const ORTCHAR_T* output_model_path);
  ORT_API2_STATUS(ModelCompilationOptions_SetOutputModelExternalInitializersFile,
                  _In_ OrtModelCompilationOptions* model_compile_options,
                  _In_ const ORTCHAR_T* external_initializers_file_path,
                  size_t external_initializer_size_threshold);
  ORT_API2_STATUS(ModelCompilationOptions_SetOutputModelBuffer, _In_ OrtModelCompilationOptions* model_compile_options,
                  _Inout_ OrtAllocator* allocator, void** output_model_buffer_ptr, size_t* output_model_buffer_size_ptr);
  ORT_API2_STATUS(ModelCompilationOptions_SetEpContextEmbedMode, _In_ OrtModelCompilationOptions* model_compile_options,
                  bool embed_ep_context_in_model);
```

Example (see unit tests for others):
```C++
#include "onnxruntime_cxx_api.h"

// Test using the CompileModel() API with settings:
//   - input model from buffer
//   - output model file
//   - EPContext nodes in output model use embedded binary blobs.
[...]

PR #24436, opened 2025-04-15
Clean up Compile API

adrianlizarraga:
### Description
Address additional review comments on https://github.com/microsoft/onnxruntime/pull/24207:
- Remove use of `#ifdef ORT_MINIMAL_BUILD` in public C/C++ API headers for Compile API
- Use `AllocatorPtr` internally to ensure memory is properly released if an exception is thrown while serializing the output model to the user's buffer.
- Improve C API function documentation.
- Clean up internal `ModelCompilationOptions` class

### Motivation and Context
Useful review comments were left on the original PR after merge. This addresses those comments.


onnxruntime/core/session/model_compilation_options.cc
```diff
@@ -1,26 +1,136 @@
...
+#include <string>
+#include <utility>
+
+#include "core/session/allocator_adapters.h"
+#include "core/session/onnxruntime_session_options_config_keys.h"
+#include "core/session/ort_env.h"
+
 namespace onnxruntime {
+ModelCompilationOptions::ModelCompilationOptions(const OrtEnv& env, const OrtSessionOptions& session_options)
+    : env_(env), session_options_(std::make_unique<OrtSessionOptions>(session_options)) {
```
yuslepukhin, line 17: >std::make_unique<OrtSessionOptions>(session_options) [](http://example.com/codeflow?start=34&length=52)

Since the session options do not appear to be optional, what is the purpose of allocating it on a heap?

adrianlizarraga: Ah yes. It used to be optional in a previous iteration. Now it is not, so I updated this to store the session options member by value.

yuslepukhin (approved):
:shipit:

PR #24437, opened 2025-04-16
[WebNN] Always execute decomposed *SimplifiedLayerNormalization in FP32

Honry:
Decomposed [Skip]SimplifiedLayerNormalization will lose precision in FP16, we'd like to add cast (to: fp32) ops around it in WebNN EP to ensure its precision rather than manually add cast nodes in each model file.

### DIFF
--- a/onnxruntime/core/providers/webnn/builders/impl/normalization_op_builder.cc
+++ b/onnxruntime/core/providers/webnn/builders/impl/normalization_op_builder.cc
@@ -115,17 +115,20 @@ Status NormalizationOpBuilder::AddToModelBuilderImpl(ModelBuilder& model_builder
       emscripten::val common_options = emscripten::val::object();
 
       if (input_type == ONNX_NAMESPACE::TensorProto_DataType_FLOAT16) {
-        // Decomposed *SimplifiedLayerNormalization may lose precision if its data type is float16,
-        // cast all inputs to float32 to ensure precision.
+        // Decomposed *SimplifiedLayerNormalization may lose precision if its data type is float16.
+        // So cast all inputs to float32 to ensure precision.
         common_options.set("label", node.Name() + "_cast_input_to_fp32");
-        input = model_builder.GetBuilder().call<emscripten::val>("cast", input, emscripten::val("float32"));
+        input = model_builder.GetBuilder().call<emscripten::val>("cast", input,
+                                                                 emscripten::val("float32"), common_options);
 
         common_options.set("label", node.Name() + "_cast_scale_to_fp32");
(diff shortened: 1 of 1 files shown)


Honry:

guschmue (approved):

onnxruntime/core/providers/webnn/builders/impl/normalization_op_builder.cc
```diff
@@ -114,9 +114,29 @@ Status NormalizationOpBuilder::AddToModelBuilderImpl(ModelBuilder& model_builder
       ORT_RETURN_IF_NOT(GetType(*input_defs[0], input_type, logger), "Cannot get input type");
       emscripten::val common_options = emscripten::val::object();
 
+      if (input_type == ONNX_NAMESPACE::TensorProto_DataType_FLOAT16) {
+        // Decomposed *SimplifiedLayerNormalization may lose precision if its data type is float16,
+        // cast all inputs to float32 to ensure precision.
```
fdwr, line 119:
```suggestion
        // Decomposed *SimplifiedLayerNormalization may lose precision if its data type is float16.
        // So cast all inputs to float32 to ensure precision.
```

onnxruntime/core/providers/webnn/builders/impl/normalization_op_builder.cc
```diff
@@ -114,9 +114,29 @@ Status NormalizationOpBuilder::AddToModelBuilderImpl(ModelBuilder& model_builder
...
+        common_options.set("label", node.Name() + "_cast_input_to_fp32");
+        input = model_builder.GetBuilder().call<emscripten::val>("cast", input, emscripten::val("float32"));
```
fdwr, line 121:
```suggestion
        input = model_builder.GetBuilder().call<emscripten::val>("cast", input, emscripten::val("float32"), common_options);
```
I don't see `common_options` being used (below too).

fdwr (commented):
👀

onnxruntime/core/providers/webnn/builders/impl/normalization_op_builder.cc
Honry: Oops, I forgot it. Fixed.

Honry:
@fdwr, thanks for your comments, fixed in new commit, PTAL again.

fdwr (approved):
👍

fdwr:
Merging since these 5 failures are unrelated, persistent, and pervasive.

> failed to authorize: failed to fetch anonymous token: unexpected status from GET request

PR #24447, opened 2025-04-16
Fix cuda memory access violation in GQA FlashAttention

RyanUnderhill:
### Description
zeros_ memory buffer was uninitialized, but it must be initialized to zero.

### Motivation and Context
A memory allocator change in GenAI started crashing in FlashAttention and this was eventually tracked down to be the cause. The allocator change was innocent. I'm not sure how this didn't fail previously, or if it was we weren't getting the reports about it.

### DIFF
--- a/onnxruntime/contrib_ops/cuda/bert/group_query_attention.cc
+++ b/onnxruntime/contrib_ops/cuda/bert/group_query_attention.cc
@@ -63,6 +63,7 @@ GroupQueryAttention<T>::GroupQueryAttention(const OpKernelInfo& info)
 
   if (!disable_flash_attention_) {
     zeros_ = this->GetScratchBuffer<int>(kZerosCount, nullptr);
+    CUDA_CALL_THROW(cudaMemset(zeros_.get(), 0, kZerosCount * sizeof(int)));
   }
 }
 


aciddelgado (approved):

baijumeswani (approved):

onnxruntime/contrib_ops/cuda/bert/group_query_attention.cc
```diff
@@ -63,6 +63,7 @@ GroupQueryAttention<T>::GroupQueryAttention(const OpKernelInfo& info)
 
   if (!disable_flash_attention_) {
     zeros_ = this->GetScratchBuffer<int>(kZerosCount, nullptr);
+    CUDA_CALL_THROW(cudaMemset(zeros_.get(), 0, kZerosCount * sizeof(int)));
```
baijumeswani, line 66: can this by `cudaMemsetAsync`?

RyanUnderhill: We have no stream here, it's in the constructor of the op. (same reason the GetScratchBuffer has to use nullptr vs context->GetCurrentStream())

PR #24455, opened 2025-04-17
AutoSelect EP Infrastructure Part 2. 

skottmckay:
### Description

NOTE: Moving out of Draft state so the CIs run. 

Diffs are against https://github.com/microsoft/onnxruntime/pull/24430 and will be a follow-up PR once that is checked in.

Proposed changes. Can split out the 3 main areas if it's easier/quicker to review.
Assumption: We want all these changes. Please feel free to disagree with some/all and to suggest alternatives. 

Update SessionOptionsAppendExecutionProvider_V2 to make OrtEpDevices as input to support all known scenarios.
  - supports needing to pick an EP that supports multiple devices to run on a specific device
    - e.g. WebGPU could run on discrete or integrated GPU and we want to register it for one
  - supports needing to pick an EP to run on multiple devices it supports
    - e.g. OpenVINO could be some combination of CPU/GPU/NPU. not necessarily all. not necessarily one. 
  - update EpFactoryInternal to support multiple devices

Re-work setup for provider bridge EPs being registered. 
  - removes all hardcoding of EP name to identify provider bridge EP
  - add CreateEpFactories/ReleaseEpFactory to provider bridge EP
  - add CreateIExecutionProvider to Provider
  - add implementation of the pieces that are now on the provider bridge side to the CUDA EP.
  - wire up EpLibraryProviderBridge to use new setup

Add enhance device discovery
  - try and find/add more meaningful data so there's always a vendor name and description
  - add some other arbitrary metadata. TBD what we really want/need
  - I am really just fumbling my way around here to see what's possible. Happy to have all of this replaced with a better approach.

Minor cleanups
  - use uint32_t for vendor_id and device_id as that's the real type
  - rename bus_id to device_id as that matches the fields in the hardware id we're reading from
    - the new device discovery can lookup BusNumber and add to the metadata which makes bus_id confusing
    - open to suggestions on naming. we don't have a field that represents 'unique id' (not sure we need one). 'device_id' could potentially be confused as being a unique id for the device.
    -


onnxruntime/core/platform/windows/device_discovery.cc
```diff
@@ -3,46 +3,275 @@
...
+                                            (PBYTE)buffer.data(), (DWORD)buffer.size(), &size)) {
+        entry->vendor = buffer;
+      }
+
+      if (guid != GUID_DEVCLASS_PROCESSOR) {
+        DWORD busNumber = 0;
+        size = 0;
+        if (SetupDiGetDeviceRegistryPropertyW(devInfo, &devData, SPDRP_BUSNUMBER, nullptr,
+                                              reinterpret_cast<PBYTE>(&busNumber), sizeof(busNumber), &size)) {
+          // push_back in case there are two identical devices. not sure how else to tell them apart
```
adrianlizarraga, line 201: I could be completely wrong, but would [https://learn.microsoft.com/en-us/windows/win32/api/setupapi/nf-setupapi-setupdigetdeviceinstanceidw](SetupDiGetDeviceInstanceIdW) be useful here?

>The following is an example of an instance ID ("1&08") concatenated to a device ID for a PCI device:
>
>PCI\VEN_1000&DEV_0001&SUBSYS_00000000&REV_02\1&08

skottmckay: Consider all this code to be a temporary/interim solution so there are probably lots of things that could be better.

I did start with this, but couldn't find a way to join a unique identifier from the SetupApi world with the DX LUID. Best I could find was vendor_id + device_id (which seems to be good for PCI but hit-and-miss for ACPI), so the info I'm getting from SetupApi is hardware type specific not instance specific (hence the list of BusNumber if you have say 2 identical GPUs in your system). 

and once vendo id + device id was the info joining SetupApi with DX, calling SetupDiGetDeviceRegistryPropertyW didn't seem to be adding any value (although it's quite possible I have overlooked something and it can add). 

so the comment is a bit misleading and should be updated.

onnxruntime/core/providers/cuda/cuda_provider_factory.cc
```diff
@@ -279,10 +279,108 @@ struct CUDA_Provider : Provider {
...
 
 CUDA_Provider* GetProvider() {
   return &g_provider;
 }
 
 }  // namespace onnxruntime
+
+#include "core/framework/error_code_helper.h"
+
+// OrtEpApi infrastructure for
```
RyanUnderhill, line 307: for... is this comment incomplete?

onnxruntime/core/providers/cuda/cuda_provider_factory.cc
```diff
@@ -279,10 +279,108 @@ struct CUDA_Provider : Provider {
...
+                                           _Out_opt_ OrtKeyValuePairs** ep_metadata,
+                                           _Out_opt_ OrtKeyValuePairs** ep_options) {
+    const auto* factory = static_cast<const CudaEpFactory*>(this_ptr);
+
+    if (factory->ort_api.HardwareDevice_Type(device) == OrtHardwareDeviceType::OrtHardwareDeviceType_GPU &&
+        factory->ort_api.HardwareDevice_VendorId(device) == 0x10de) {
+      return true;
+    }
+
+    return false;
```
RyanUnderhill, line 338: Why don't we just return the expression result?

skottmckay: I expect the 'real' implementation should check if the device has the required compute capability. this is essentially a stub and any additional selection logic needs to be added in the `if` branch, which may include populating metadata or provider option info.

onnxruntime/core/providers/cuda/cuda_provider_factory.cc
skottmckay:
```suggestion
// OrtEpApi infrastructure to be able to use the CUDA EP as an OrtEpFactory for auto EP selection.
```

skottmckay:
Merged into #24430

PR #24456, opened 2025-04-17
NV TensorRT RTX EP - initial commit

ankan-ban:
New EP - currently based on existing TensorRT EP but meant to be used on RTX GPUs with a lean version of TensorRT.

### Description
Adding a new EP based on TensorRT EP. This is going to use a special version of TensorRT optimized for RTX GPUs. In the future we plan to make changes to the EP to streamline it further (e.g, get rid of dependency on CUDA EP completely).

### Motivation and Context
The new TensorRT for RTX is going to have:
1. Much smaller footprint 
2. Much faster model compile/load times. 
3. Better usability in terms of use of cached models across multiple RTX GPUs.

This effort is also targeting WCR ML workflows.


include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -4931,6 +4932,138 @@ struct OrtApi {
...
+ * Appends the Nv (NVIDIA) execution provider to the session options, using settings
+ * specified in the provided OrtNvTensorRtRtxProviderOptions structure.
+ *
+ * \param[in] options A pointer to an ::OrtSessionOptions instance to which the Nv execution provider will be appended.
+ * \param[in] nv_options A pointer to an ::OrtNvTensorRtRtxProviderOptions instance containing configuration for the Nv execution provider.
+ *
+ * \snippet{doc} snippets.dox OrtStatus Return Value
+ * \since Version 1.21
+ */
+ORT_API2_STATUS(SessionOptionsAppendExecutionProvider_Nv_TensorRT_RTX,
```
RyanUnderhill, line 4975: Why does this require new provider specific APIs vs using SessionOptionsAppendExecutionProvider?
@adrianlizarraga

ishwar-raut1: NvTensorRT_RTX is created in a standalone shared dll. So we require the EP-specific API.

onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc
```diff
@@ -0,0 +1,191 @@
...
+
+
+
+} g_info;
+
+struct NvProviderFactory : IExecutionProviderFactory {
+  NvProviderFactory(const NvExecutionProviderInfo& info) : info_{info} {}
+  ~NvProviderFactory() override {}
+
+  std::unique_ptr<IExecutionProvider> CreateProvider() override;
```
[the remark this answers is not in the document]
ankan-ban: Addressed this part.

include/onnxruntime/core/session/onnxruntime_c_api.h
ankan-ban: @RyanUnderhill we have removed these APIs in the latest revision. After reviewing the SessionOptionsAppendExecutionProvider API we realized there was no need for provider specific APIs.

cmake/onnxruntime_unittests.cmake
```diff
@@ -61,21 +61,21 @@ function(AddTest)
...
+      target_link_libraries(${_UT_TARGET} PRIVATE cudnn_frontend)
     endif()
     target_link_libraries(${_UT_TARGET} PRIVATE ${_UT_LIBS} GTest::gtest GTest::gmock ${onnxruntime_EXTERNAL_LIBRARIES})
   endif()
 
   onnxruntime_add_include_to_target(${_UT_TARGET} date::date flatbuffers::flatbuffers)
   target_include_directories(${_UT_TARGET} PRIVATE ${TEST_INC_DIR})
   if (onnxruntime_USE_CUDA)
-    target_include_directories(${_UT_TARGET} PRIVATE ${CMAKE_CUDA_TOOLKIT_INCLUDE_DIRECTORIES} ${CUDNN_INCLUDE_DIR})
+    target_include_directories(${_UT_TARGET} PRIVATE ${CUDAToolkit_INCLUDE_DIRS} ${CUDNN_INCLUDE_DIR})
```
snnn, line 78: Why is this change needed? What's the difference between CUDAToolkit_INCLUDE_DIRS and CMAKE_CUDA_TOOLKIT_INCLUDE_DIRECTORIES ?

jywu-msft:
@chilo-ms please review.

cmake/onnxruntime_unittests.cmake
ankan-ban: @gedoensmax do you remember why we had to do this?

gedoensmax: With these lines: https://github.com/gedoensmax/onnxruntime/blob/fb6a73162f3b72b9887b0d5577b3e46bbf3da960/cmake/CMakeLists.txt#L1573-L1576
We are setting the CUDA path and version to be used. This is AFAIK the more modern way to find CUDA dependencies and I think it would be good to use wither one or the other across a single project. I have no strong opinion though.

chilo-ms:
> @chilo-ms We will happily take any guidance on how to run more test using NV EP to find remaining bugs or implementation gaps.

We need to add a new pipeline/CI for building and testing this new NV EP which this PR hasn't done yet.
Considering this new TRT binary is not public yet, some questions here:
- In terms of building the NV EP, just to double check, i assume those header files from public TRT, i.e. NvInferXXX.h, can be used.    
- For running NV EP, our Windows pipeline uses A10, can NV EP run on A10? From what i heard it's limited to ampere+ gpus or 40xx/50xx gpus?

chilo-ms:
To add a new NV EP pipeline in GitHub Action, please duplicate from [TRT EP's](https://github.com/microsoft/onnxruntime/blob/main/.github/workflows/windows_tensorrt.yml) first and then modify accordingly.

chilo-ms (approved):
The code in ORT that makes this new NV EP known to ORT looks good to me.
As for EP code and test/validation is not the focus for this PR, we can discuss later.

jywu-msft (approved):

snnn:

PR #24460, opened 2025-04-17
add GatherBlockQuantized to webgpu ep

guschmue:


onnxruntime/core/providers/webgpu/shader_variable.cc
```diff
@@ -55,7 +60,11 @@ constexpr static const std::string_view VALUE_TYPE_ARRAY[] = {
     "u32",         // Uint8x4 (u32 as 4 elements of uint8)
     "vec2<u32>",   // Uint8x8 (vec2<u32> as 2x4 elements of uint8)
     "vec4<u32>",   // Uint8x16 (vec4<u32> as 4x4 elements of uint8)
-    "i32",         // Int8x4
+    "u32",         // Int8x4 (u32 as 4 elements of uint8)
+    "vec2<i32>",   // Int8x8 (vec2<i32> as 2x4 elements of int8)
+    "vec4<i32>",   // Int8x16 (vec4<i32> as 4x4 elements of int8)
+    "u32",         // Uint4x8
+    "i32",         // Int4x8
```
fs-eire, line 67: should this be u32 for consistency?

also , there are other inconsistency. for example, Int8x4 uses 'u32' but Int8x8 uses 'vec2<i32>'.

Since there is no (u)int8/(u)int16 in WGSL, we always uses 4-bytes-integer as bit representation, maybe always using u32 is a good idea (RSHIFT is safer since no sign bit)

guschmue: done

onnxruntime/core/providers/webgpu/shader_variable.cc
```diff
@@ -33,6 +34,10 @@ constexpr static const std::string_view STORAGE_TYPE_ARRAY[] = {
     "vec2<u32>",  // Uint8x8
     "vec4<u32>",  // Uint8x16
     "u32",        // Int8x4
+    "vec2<u32>",  // Int8x8
+    "vec4<u32>",  // Int8x16 20
```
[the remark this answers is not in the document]
guschmue: done

onnxruntime/core/providers/webgpu/program.cc
```diff
@@ -120,6 +123,8 @@ int NumberOfComponents(ProgramVariableDataType type) {
     case ProgramVariableDataType::Int64:
     case ProgramVariableDataType::Uint64:
     case ProgramVariableDataType::Float16:
+    case ProgramVariableDataType::Uint4x8:
+    case ProgramVariableDataType::Int4x8:
```
fs-eire, line 127: Regarding the number of components, how does ORT defines the shape of a INT4 tensor?

say a simple exmaple, we have a 1-D INT4 tensor representing 1024 INT4 numbers. is the tensor shape [1024] or [512] (or other)?

guschmue: [1024] in onnx

guschmue:
ups, I enabled webgpu wrong in unit test - if I fix it, I see some failing. Fixing it.

onnxruntime/core/providers/webgpu/program.cc
fs-eire: then to make the definition of "number of components" consistent, it should be 8 instead of 1 here.

onnxruntime/core/providers/webgpu/shader_variable.cc
```diff
@@ -55,7 +60,11 @@ constexpr static const std::string_view VALUE_TYPE_ARRAY[] = {
     "u32",         // Uint8x4 (u32 as 4 elements of uint8)
     "vec2<u32>",   // Uint8x8 (vec2<u32> as 2x4 elements of uint8)
     "vec4<u32>",   // Uint8x16 (vec4<u32> as 4x4 elements of uint8)
-    "i32",         // Int8x4
+    "u32",         // Int8x4 (u32 as 4 elements of uint8)
+    "vec2<i32>",   // Int8x8 (vec2<i32> as 2x4 elements of int8)
```
fs-eire, line 64: is this expected to be u32 instead of i32? if storage type is u32 but value type is i32, there need explicit bitcast

guschmue: made it i32

onnxruntime/core/providers/webgpu/program.cc
guschmue: fixed.

onnxruntime/core/providers/webgpu/program.cc
```diff
@@ -165,6 +165,10 @@ ProgramVariableDataType ToProgramVariableDataType(int32_t element_type, int comp
         return ProgramVariableDataType::Int64;
       case ONNX_TENSOR_ELEMENT_DATA_TYPE_UINT64:
         return ProgramVariableDataType::Uint64;
+      case ONNX_TENSOR_ELEMENT_DATA_TYPE_UINT4:
+        return ProgramVariableDataType::Uint4x8;
+      case ONNX_TENSOR_ELEMENT_DATA_TYPE_INT4:
+        return ProgramVariableDataType::Int4x8;
```
fs-eire, line 171: these should be removed. they are in the "components == 8" case

onnxruntime/contrib_ops/webgpu/quantization/gather_block_quantized.cc
```diff
@@ -0,0 +1,199 @@
...
+
+  return context.RunProgram(program);
+}
+
+namespace {
+const std::vector<MLDataType>& GatherBlockQuantizedT1Constraint() {
+  static std::vector<MLDataType> types{
+      DataTypeImpl::GetTensorType<Int4x2>(),
+      DataTypeImpl::GetTensorType<UInt4x2>(),
+      // DataTypeImpl::GetTensorType<uint8_t>()
```
fs-eire, line 174: is this commented intentionally?

PR #24462, opened 2025-04-17
[CoreML] Add support for int64

carzh:
### Description
Add int64 as a supported datatype for moving nodes to the CoreML EP.

We already convert constants automatically from int64 to int32 for CoreML by calling narrow. 

Adding the conversion for outputs as well.

### Motivation and Context
- More nodes supported on CoreML

### Note on the Unsqueeze op
According to #22975 there is a bug with the Unsqueeze op with scalar inputs on x86.

I was running into a bug for unsqueezes that unsqueezed a scalar input to a tensor of shape [1] since CoreML doesn't support scalar values for MLProgram. I adapted the HandleX86ArchUnsqueeze method but alternatively, can replace with an identity operator or add some additional checks. I went with adapting the HandleX86ArchUnsqueeze method since it seemed like the fastest solution.

### DIFF
--- a/onnxruntime/core/providers/coreml/builders/impl/base_op_builder.cc
+++ b/onnxruntime/core/providers/coreml/builders/impl/base_op_builder.cc
@@ -110,14 +110,14 @@ bool BaseOpBuilder::IsInputDtypeSupport(const Node& node, size_t idx,
   }
 
   // float is supported
-  if (input_type == ONNX_NAMESPACE::TensorProto_DataType_FLOAT ||
-      input_type == ONNX_NAMESPACE::TensorProto_DataType_INT64) {
+  if (input_type == ONNX_NAMESPACE::TensorProto_DataType_FLOAT) {
     return true;
   }
 
 #if CAN_BUILD_COREML6_OR_LATER
-  // only MLProgram support FP16
-  if (input_params.create_mlprogram && input_type == ONNX_NAMESPACE::TensorProto_DataType_FLOAT16) {
+  // only MLProgram support FP16 and INT64
+  if (input_params.create_mlprogram && (input_type == ONNX_NAMESPACE::TensorProto_DataType_FLOAT16 ||
+                                        input_type == ONNX_NAMESPACE::TensorProto_DataType_INT64)) {
     return true;
   }
 #endif
--- a/onnxruntime/core/providers/coreml/builders/impl/base_op_builder.cc
+++ b/onnxruntime/core/providers/coreml/builders/impl/base_op_builder.cc
@@ -110,7 +110,8 @@ bool BaseOpBuilder::IsInputDtypeSupport(const Node& node, size_t idx,
   }
 
   // float is supported
-  if (input_type == ONNX_NAMESPACE::TensorProto_DataType_FLOAT) {
+  if (input_type == ONNX_NAMESPACE::TensorProto_DataType_FLOAT ||
+      input_type == ONNX_NAMESPACE::TensorProto_DataType_INT64) {
     return true;
   }
 
--- a/onnxruntime/core/providers/coreml/builders/impl/binary_op_builder.cc
+++ b/onnxruntime/core/providers/coreml/builders/impl/binary_op_builder.cc
@@ -145,8 +145,8 @@ Status BinaryOpBuilder::AddToModelBuilderImpl(ModelBuilder& model_builder, const
     } else if (op_type == "Sub") {
       coreml_op_type = "sub";
     } else if (op_type == "Div") {
-      // we support fp32/fp16 currently. when we add support for integers we need to check the type and use
-      // "floor_div" or "real_div" accordingly
+      // Use "floor_div" op for integer division (int32 or int64)
+      // use "real_div" for float division (fp16 or fp32)
       if (ShouldUseFloorDiv(node, logger)) {
         coreml_op_type = "floor_div";
       } else {
--- a/onnxruntime/core/providers/coreml/builders/impl/binary_op_builder.cc
+++ b/onnxruntime/core/providers/coreml/builders/impl/binary_op_builder.cc
@@ -56,17 +56,14 @@ bool CheckIfBothInputShapesMatch(const Node& node, const logging::Logger& logger
 }
 
 bool ShouldUseFloorDiv(const Node& node, const logging::Logger& logger) {
+  // since ONNX spec requires both inputs to have the same type, we only need
+  // to check the first input type
   const auto& input0 = *node.InputDefs()[0];
-  const auto& input1 = *node.InputDefs()[1];
   int32_t input_type0 = ONNX_NAMESPACE::TensorProto_DataType_UNDEFINED;
-  int32_t input_type1 = ONNX_NAMESPACE::TensorProto_DataType_UNDEFINED;
   GetType(input0, input_type0, logger);
-  GetType(input1, input_type1, logger);
 
   return input_type0 == ONNX_NAMESPACE::TensorProto_DataType_INT32 ||
-         input_type0 == ONNX_NAMESPACE::TensorProto_DataType_INT64 ||
-         input_type1 == ONNX_NAMESPACE::TensorProto_DataType_INT32 ||
-         input_type1 == ONNX_NAMESPACE::TensorProto_DataType_INT64;
+         input_type0 == ONNX_NAMESPACE::TensorProto_DataType_INT64;
 }
 }  // namespace
 
--- a/onnxruntime/core/providers/coreml/builders/impl/binary_op_builder.cc
+++ b/onnxruntime/core/providers/coreml/builders/impl/binary_op_builder.cc
@@ -54,6 +54,20 @@ bool CheckIfBothInputShapesMatch(const Node& node, const logging::Logger& logger
                     y_shape_proto->dim().begin(), y_shape_proto->dim().end(),
                     dim_eq);
 }
+
+bool ShouldUseFloorDiv(const Node& node, const logging::Logger& logger) {
+  const auto& input0 = *node.InputDefs()[0];
+  const auto& input1 = *node.InputDefs()[1];
+  int32_t input_type0 = ONNX_NAMESPACE::TensorProto_DataType_UNDEFINED;
+  int32_t input_type1 = ONNX_NAMESPACE::TensorProto_DataType_UNDEFINED;
+  GetType(input0, input_type0, logger);
+  GetType(input1, input_type1, logger);
+
+  return input_type0 == ONNX_NAMESPACE::TensorProto_DataType_INT32 ||
+         input_type0 == ONNX_NAMESPACE::TensorProto_DataType_INT64 ||
+         input_type1 == ONNX_NAMESPACE::TensorProto_DataType_INT32 ||
+         input_type1 == ONNX_NAMESPACE::TensorProto_DataType_INT64;
+}
 }  // namespace
 
 static std::vector<int64_t> InferOutputShape(const std::vector<int64_t>& a, const std::vector<int64_t>& b) {
@@ -133,7 +147,11 @@ Status BinaryOpBuilder::AddToModelBuilderImpl(ModelBuilder& model_builder, const
     } else if (op_type == "Div") {
       // we support fp32/fp16 currently. when we add support for integers we need to check the type and use
       // "floor_div" or "real_div" accordingly
-      coreml_op_type = "real_div";
+      if (ShouldUseFloorDiv(node, logger)) {
+        coreml_op_type = "floor_div";
+      } else {
+        coreml_op_type = "real_div";
+      }
     } else if (op_type == "Pow") {
       coreml_op_type = "pow";
     } else {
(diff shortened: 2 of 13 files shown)


onnxruntime/core/providers/coreml/builders/impl/builder_utils.cc
```diff
@@ -261,9 +261,8 @@ MILSpec::DataType OnnxDataTypeToMILSpec(int onnx_type) {
     case ONNX_NAMESPACE::TensorProto_DataType_INT16:
       return MILSpec::DataType::INT16;
     case ONNX_NAMESPACE::TensorProto_DataType_INT32:
-      return MILSpec::DataType::INT32;
     case ONNX_NAMESPACE::TensorProto_DataType_INT64:
-      return MILSpec::DataType::INT64;
+      return MILSpec::DataType::INT32;
```
edgchen1, line 265: I think it may be more understandable to keep this helper function as a direct, one to one mapping.

a side question is, why do they even have the `MILSpec::DataType::INT64` data type?

onnxruntime/core/providers/coreml/builders/impl/builder_utils.cc
```diff
@@ -367,19 +366,15 @@ void AddIntermediateOperationOutput(COREML_SPEC::MILSpec::Operation& op, std::st
...
   output_arg.set_name(output.Name());
 
   MILSpec::ValueType& value = *output_arg.mutable_type();
   MILSpec::TensorType& tensor_type = *value.mutable_tensortype();
 
-  auto elem_type = override_element_type ? *override_element_type
-                                         : output.TypeAsProto()->tensor_type().elem_type();
-
-  SetTensorTypeInfo(tensor_type, OnnxDataTypeToMILSpec(elem_type), output.Shape(), /*convert_scalar*/ true);
+  SetTensorTypeInfo(tensor_type, OnnxDataTypeToMILSpec(output.TypeAsProto()->tensor_type().elem_type()), output.Shape(), /*convert_scalar*/ true);
```
edgchen1, line 377: if you want to get rid of the `override_element_type` argument, maybe the conversion from int64 to int32 could happen in this function instead of within `OnnxDataTypeToMILSpec`.

onnxruntime/core/providers/coreml/builders/impl/builder_utils.h
```diff
@@ -160,8 +160,7 @@ void AddIntermediateOperationOutput(COREML_SPEC::MILSpec::Operation& op, std::st
 ///   Override the element type. Only set to handle cases where we believe the data at runtime will be int32 but
 ///   the original ONNX node has type int64.
 /// </param>
-void AddOperationOutput(COREML_SPEC::MILSpec::Operation& op, const NodeArg& output,
-                        std::optional<int32_t> override_element_type = std::nullopt);
+void AddOperationOutput(COREML_SPEC::MILSpec::Operation& op, const NodeArg& output);
```
edgchen1, line 163: also update the function documentation comment.

onnxruntime/core/providers/coreml/builders/impl/base_op_builder.cc
```diff
@@ -110,7 +110,8 @@ bool BaseOpBuilder::IsInputDtypeSupport(const Node& node, size_t idx,
   }
 
   // float is supported
-  if (input_type == ONNX_NAMESPACE::TensorProto_DataType_FLOAT) {
+  if (input_type == ONNX_NAMESPACE::TensorProto_DataType_FLOAT ||
+      input_type == ONNX_NAMESPACE::TensorProto_DataType_INT64) {
```
edgchen1, line 114: it seems like a big change to enable int64 by default now. should we start by enabling it per op builder instead? e.g., by overriding `HasSupportedInputsImpl()`

onnxruntime/core/providers/coreml/builders/impl/squeeze_op_builder.cc
```diff
@@ -86,13 +86,15 @@ Status SqueezeOpBuilder::AddToModelBuilderImpl(ModelBuilder& model_builder,
   if (model_builder.CreateMLProgram()) {
     using namespace CoreML::Specification::MILSpec;
 
-#if defined(TARGET_CPU_X86_64) && TARGET_CPU_X86_64
     // expand_dims has limited requirements for static shape, however, X86_64 has a bug that it can't handle scalar input
+    // in addition, MLProgram does not support scalar values -- we convert the scalars to 1D tensors.
```
edgchen1, line 90: do we need to mention the x86_64 bug anymore if the ifdef is removed?

carzh: Ah, I'm not sure what the x86_64 bug is and I thought to leave it in because it might refer to a separate issue from what I encountered, but I'm open to removing it

onnxruntime/core/providers/coreml/builders/impl/base_op_builder.cc
skottmckay: I think it becomes a much larger change to do operator by operator for a model where the input data is int64 (vs vision models where the input is float)

By enabling it here, the ORT unit tests should run and we should have good coverage with int64 data shouldn't we?

onnxruntime/core/providers/coreml/builders/impl/builder_utils.cc
skottmckay: > a side question is, why do they even have the `MILSpec::DataType::INT64` data type?

easy to have a 'we might want this in the future' value. but AFAIK there is no usage of it in an actual MLModel.

but having a 1:1 mapping would be nicer as we could have an explicit comment on the case for INT64 saying that always gets converted to int32 in the CoreML world so the next developer to read the code understands what's going on.

onnxruntime/core/providers/coreml/builders/impl/builder_utils.cc
skottmckay: Do we/could we use OnnxDataTypeToMILSpec for graph inputs/initializers as well so doing it there covers all potential use cases?

onnxruntime/core/providers/coreml/builders/impl/cast_op_builder.cc
```diff
@@ -44,7 +44,6 @@ Status CastOpBuilder::AddToModelBuilderImpl([[maybe_unused]] ModelBuilder& model
       // CoreML operators can only produce int32 and not int64 values.
       // Due to that there should be no actual int64 values inside the CoreML model and we can infer any
       // ONNX_NAMESPACE::TensorProto::INT64 values to be int32.
-      cast_to_type = ONNX_NAMESPACE::TensorProto::INT32;
     } else if (cast_to_type == ONNX_NAMESPACE::TensorProto::FLOAT) {
```
skottmckay, line 47: A comment like this explaining the int64 -> int32 rationale would be good in OnnxDataTypeToMILSpec

onnxruntime/core/providers/coreml/builders/impl/pad_op_builder.cc
```diff
@@ -150,6 +150,17 @@ bool PadOpBuilder::IsOpSupportedImpl(const Node& node, const OpBuilderInputParam
       LOGS(logger, VERBOSE) << "constant_value must be a constant initializer.";
       return false;
     }
+
+    int32_t constant_value_type;
+    if (!GetType(*input_defs[2], constant_value_type, logger)) {
+      LOGS(logger, VERBOSE) << "Get constant_value type failed";
+      return false;
+    }
```
skottmckay, line 158: nit: an initializer always has a type as it's constant data, so this check may be unnecessary.

onnxruntime/core/providers/coreml/builders/impl/squeeze_op_builder.cc
skottmckay: @wejoncy would know more.

wejoncy: CoreML will raise an internal error if feeding  a scalar input.
Not sure if the latest CoreML fixed it not.

onnxruntime/core/providers/coreml/builders/impl/builder_utils.cc
edgchen1: ok. when I see this function `OnnxDataTypeToMILSpec`, my initial assumption is that it just maps one enum type to another. if the destination enum also has "int64", it is unexpected (to me) that the source enum "int64" maps to something else.

perhaps a good place for describing this special behavior would be the comment that documents the function.
https://github.com/microsoft/onnxruntime/blob/f76e4de3b45b13204e69fc851201848803dc1b3a/onnxruntime/core/providers/coreml/builders/impl/builder_utils.h#L99-L101

onnxruntime/core/providers/coreml/builders/impl/base_op_builder.cc
edgchen1: ok. I meant a big change in behavior vs a big code change.

> By enabling it here, the ORT unit tests should run and we should have good coverage with int64 data shouldn't we?

hopefully this is the case.

onnxruntime/core/providers/coreml/builders/impl/base_op_builder.cc
```diff
@@ -110,7 +110,8 @@ bool BaseOpBuilder::IsInputDtypeSupport(const Node& node, size_t idx,
   }
 
   // float is supported
```
edgchen1, line 112: also update this comment

onnxruntime/core/providers/coreml/builders/impl/builder_utils.cc
carzh: Ah, that's a fair point -- I added comment to both the header file and the switch case

onnxruntime/core/providers/coreml/builders/impl/squeeze_op_builder.cc
carzh: Gotcha looks to be the same issue as what I was running into so amended the comment

onnxruntime/core/providers/coreml/builders/impl/builder_utils.cc
carzh: Yes, it seems that it's already used in model_builder.cc for constants and tensors! I removed some of the ternary statements that were used there in the last commit

onnxruntime/core/providers/coreml/builders/impl/binary_op_builder.cc
```diff
@@ -131,9 +145,13 @@ Status BinaryOpBuilder::AddToModelBuilderImpl(ModelBuilder& model_builder, const
...
     } else if (op_type == "Div") {
-      // we support fp32/fp16 currently. when we add support for integers we need to check the type and use
-      // "floor_div" or "real_div" accordingly
-      coreml_op_type = "real_div";
+      // Use "floor_div" op for integer division (int32 or int64)
+      // use "real_div" for float division (fp16 or fp32)
+      if (ShouldUseFloorDiv(node, logger)) {
+        coreml_op_type = "floor_div";
+      } else {
+        coreml_op_type = "real_div";
```
skottmckay, line 153: So we have int64 unit tests for Div

onnxruntime/core/providers/coreml/builders/impl/binary_op_builder.cc
```diff
@@ -54,6 +54,20 @@ bool CheckIfBothInputShapesMatch(const Node& node, const logging::Logger& logger
...
+  const auto& input1 = *node.InputDefs()[1];
+  int32_t input_type0 = ONNX_NAMESPACE::TensorProto_DataType_UNDEFINED;
+  int32_t input_type1 = ONNX_NAMESPACE::TensorProto_DataType_UNDEFINED;
+  GetType(input0, input_type0, logger);
+  GetType(input1, input_type1, logger);
+
+  return input_type0 == ONNX_NAMESPACE::TensorProto_DataType_INT32 ||
+         input_type0 == ONNX_NAMESPACE::TensorProto_DataType_INT64 ||
+         input_type1 == ONNX_NAMESPACE::TensorProto_DataType_INT32 ||
+         input_type1 == ONNX_NAMESPACE::TensorProto_DataType_INT64;
```
skottmckay, line 69: both inputs in ONNX have to be the same type so you don't need to check input_type0 and input_type1

onnxruntime/core/providers/coreml/builders/impl/binary_op_builder.cc
carzh: Ah to summarize offline conversation for visibility -- div int64 unit test exists [here](https://github.com/microsoft/onnxruntime/blob/15b3f086fae73702b62ba3da8a10e6ad4725efd8/onnxruntime/test/providers/cpu/math/element_wise_ops_test.cc#L886) and covers the expected rounding behavior. This unit test passed previously (before the addition of floor_div handling) -- [CoreML docs](https://apple.github.io/coremltools/_modules/coremltools/converters/mil/mil/ops/defs/iOS15/elementwise_binary.html#real_div) don't specify expected behavior for real_div when using int32, but from the unit tests, it seems to have the same behavior as ONNX div. So using floor_div is optional (behavior-wise), but left the floor_div handling in because I thought it didn't hurt

edgchen1 (approved):

PR #24176, opened 2025-03-25
Validate CreateSessionFromArray with ep.context_enable enabled

HectorSVC:
### Description
Add validation in path that user CreateSessionFromArray: if ep.context_enable is set, then ep.context_file_path is expected, otherwise report error because ORT don't know where to generate the _ctx.onnx file

PR #24463, opened 2025-04-18
Allow EpContext models with input/output models completely in buffers

adrianlizarraga:
### Description
Re-enables (and fixes) generation of compiled EpContext models with **both** input and output models stored in buffers.

### Motivation and Context
Previous PR #24176 inadvertently added a check that disabled storing both input and output models in buffers. However, we need this functionality. This was actually a fortunate scenario, as it led to the discovery of a bug.


HectorSVC (approved):
:shipit:

PR #24384, opened 2025-04-10
[CPU] Add 8bit support to matmulnbits quantizer 

fajin-corp:
### Description
Add 8bit support to matmulnbits quantizer. matmul_4bits_quantizer now can quantize a const B in a MatMul to 8bits initializer.

### Motivation and Context
MatMul4Bits has accuracy issue for phi-4 model used for foundry local.
The early prototype showed >= 6bits can fix the issue.
To mitigate the issue as soon as possible, add 8bit support to MatMulNBits.

PR #24472, opened 2025-04-19
Rename matmul_4bits_quantizer.py to matmul_nbits_quantizer.py

tianleiwu:
### Description

* Rename  filename and class name since it supports 4 and 8 bits.
* Update HQQWeightOnlyQuantizer to support 8 bits.
* Update some comments.

### Motivation and Context
https://github.com/microsoft/onnxruntime/pull/24384 added 8 bits support for the default weight only quantizer.

### DIFF
--- a/onnxruntime/python/tools/quantization/matmul_nbits_quantizer.py
+++ b/onnxruntime/python/tools/quantization/matmul_nbits_quantizer.py
@@ -50,8 +50,8 @@ class WeightOnlyQuantConfig:
             quant_axes (dict[str, int], optional):
                 op:axis, which axis to quantize for an op. Default {MatMul: 0, Gather: 1}
             customized_weight_config:
-                customized weight config for nodes if needed.
-                If both customized_weight_config and nodes_to_exclude are set, nodes_to_exclude overwrites customized_weight_config.
+                customized weight config for nodes if needed. It is dictionary with node name as key,
+                and the value is a dict of customized config.
         """
         self.algorithm = algorithm
         self.quant_format = quant_format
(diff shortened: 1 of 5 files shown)


onnxruntime/python/tools/quantization/matmul_nbits_quantizer.py
```diff
@@ -1095,14 +1104,13 @@ def quantize_awq(self, model: ModelProto | str) -> ModelProto:
...
-# TODO(fajin): change class name
-class MatMul4BitsQuantizer:
+class MatMulNBitsQuantizer:
     """
     Target node:        QOperator node:            QDQ nodes:
     MatMul              MatMulNBits                DeQuantizeLinear -> MatMul
     Gather              GatherBlockQuantized       Gather, Gather, Gather (optional) -> DequantizeLinear
 
-    Perform 4b quantization of constant weights for target nodes.
+    Perform 4/8 bits quantization of constant weights for target nodes.
```
kunal-vaishnavi, line 1113: It appears [this PR](https://github.com/microsoft/onnxruntime/pull/24414) will enable N bits for `MatMulNBits` during inference. Can we add something similar for quantizing to N bits?

kunal-vaishnavi (approved):

onnxruntime/python/tools/quantization/matmul_nbits_quantizer.py
jywu-msft: it updates the quantizer as well https://github.com/microsoft/onnxruntime/blob/377d96e0c4cc9ff1cdaa4cee9b2c20bcbd1e0790/onnxruntime/python/tools/quantization/matmul_4bits_quantizer.py
@liqunfu

jiafatom (commented):
Can we rename matmul_4bits_quantizer.py to matmul_quantizer.py? Previously we have 4 bits only, so we add 4bits in between. It seems a matmul_quantizer to me

tianleiwu:
> Can we rename matmul_4bits_quantizer.py to matmul_quantizer.py? Previously we have 4 bits only, so we add 4bits in between. It seems a matmul_quantizer to me

nbits naming follows MatMulNBits op.

PR #24478, opened 2025-04-21
[WebGPU] Optimize GEMM with vec4

xiaofeihan1:
### Description

In this PR, we use vec4 to optimize GEMM when colums of A and B can be divided by 4, or use previous shader.
I will add u32/vec2 implementation in the future, and we will only keep one shader at that time.

###  Perf comparison
I run customized model only include GEMM(M = N = K = 1024) with nodejs on M2/M3 Max. Roughly 20% increase.
 || !transA&&!transB | transA     | transB     | transA&&transB |  
|------------------|------------|------------|----------------|------------|
| M2               | 9.36->7.41 | 9.45->7.54 | 11.21->8.19    | 9.66->8.37 |
| M3 max           | 8.07->6.99 | 7.54->6.53 | 8.42->5.89     | 5.47->5.29 |


fs-eire:
Is there a way to reuse the implementation of `MatMul`? My understanding is that there are some kind of duplication between GEMM and MatMul, and it would be great if we can reuse the shared code

xiaofeihan1:
> Is there a way to reuse the implementation of `MatMul`? My understanding is that there are some kind of duplication between GEMM and MatMul, and it would be great if we can reuse the shared code

Thanks for the callout. That's what I'm gonna do next.
For current PR, I want to push forward to support vec4 for GEMM. I will take the refactor work in future PRs because it also require some effort to consider(e.g.There are also some differences between gemm and matmul, e.g. the latter supports batch size, the former supports transpose, etc).

onnxruntime/core/providers/webgpu/math/gemm_vec4.cc
```diff
@@ -0,0 +1,323 @@
...
+  //   shader.MainFunctionBody() << "  if (m < uniforms.M && n < uniforms.N4) {\n"
+  //                             << "    values = values + uniforms.beta * "
+  //                             << C.GetByOffset(C.BroadcastedIndicesToOffset("vec2(m, n)", output)) << ";\n"
+  //                             << "  }\n";
+  // }
+
+  // Write output
+  // shader.MainFunctionBody() << "  if (m < uniforms.M && n < uniforms.N4) {\n"
+  //                           << "    " << output.SetByOffset("m * uniforms.N4 + n", "values") << "\n"
+  //                           << "  }\n";
```
qjia7, line 228: Just remove above code if they are needed anymore.

onnxruntime/core/providers/webgpu/math/gemm_vec4.cc
```diff
@@ -0,0 +1,323 @@
...
+  const uint32_t TILE_SIZE = 32;
+  const uint32_t num_tile_n = (N + TILE_SIZE - 1) / TILE_SIZE;
+  const uint32_t num_tile_m = (M + TILE_SIZE - 1) / TILE_SIZE;
+
+  program.CacheHint(alpha, transA_, transB_)
+      .AddOutputs({{y, ProgramTensorMetadataDependency::TypeAndRank, components}})
+      .SetDispatchGroupSize(num_tile_n * num_tile_m)
+      .SetWorkgroupSize(256, 1, 1)
+      .AddUniformVariables({
+          {static_cast<uint32_t>(num_tile_n)},  // num_tile_n
```
qjia7, line 308: `static_cast<uint32_t>` is not needed since `num_tile_n` is type `uint32_t`. Similar for other places.

onnxruntime/core/providers/webgpu/math/gemm_vec4.cc
```diff
@@ -0,0 +1,323 @@
...
+  }
+
+  const auto& c_shape = c->Shape();
+  if (c_shape.NumDimensions() == 2) {
+    uint32_t C_cols = onnxruntime::narrow<uint32_t>(c_shape[1]);
+    return C_cols % 4 == 0;
+  }
+
+  uint32_t C_rows = onnxruntime::narrow<uint32_t>(c_shape[0]);
+  return C_rows % 4 == 0;
```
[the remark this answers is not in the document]
xiaofeihan1: Thanks for the suggestion! I have changed the code. For C's last dimension
- N.Use component = 4. Bias is `C.BroadcastedIndicesToOffset("vec2(row, col)", output)`
- 1.Use component = 1
  - shape size == 1. Bias is `vec4(C[0])` 
  - shape size == M. Bias is `vec4(C[row])`
Is this OK?

onnxruntime/core/providers/webgpu/math/gemm.h
```diff
@@ -22,6 +22,10 @@ class GemmProgram final : public Program<GemmProgram> {
 
   Status GenerateShaderCode(ShaderHelper& sh) const override;
```
xiaofeihan1, line 24: delete

onnxruntime/core/providers/webgpu/math/gemm_vec4.cc
```diff
@@ -0,0 +1,306 @@
...
+                      const Tensor* b) {
+  const auto& a_shape = a->Shape();
+  const auto& b_shape = b->Shape();
+
+  uint32_t A_rows = onnxruntime::narrow<uint32_t>(a_shape[0]);
+  uint32_t A_cols = onnxruntime::narrow<uint32_t>(a_shape[1]);
+  uint32_t B_rows = onnxruntime::narrow<uint32_t>(b_shape[0]);
+  uint32_t B_cols = onnxruntime::narrow<uint32_t>(b_shape[1]);
+
+  return A_rows % 4 == 0 && A_cols % 4 == 0 && B_rows % 4 == 0 && B_cols % 4 == 0;
```
qjia7, line 237: Please loose the limitations. It's too strict so that only few cases can go to this path. Maybe `a_shape[1] % 4 == 0` and `b_shape[1]%4 == 0`  are enough.

onnxruntime/core/providers/webgpu/math/gemm_vec4.cc
```diff
@@ -0,0 +1,306 @@
...
+      << "if(col < uniforms.N4 && row < uniforms.M) {";
+  if (need_handle_bias_) {
+    const ShaderVariableHelper& C = shader.AddInput("C", ShaderUsage::UseUniform);
+    shader.AdditionalImplementation() << "    values += uniforms.beta * ";
+    if (c_component_ == 1 && c_is_scalar_) {
+      shader.AdditionalImplementation() << "output_value_t(C[0]);\n";
+    } else if (c_component_ == 1) {
+      shader.AdditionalImplementation() << "output_value_t(C[row]);\n";
+    } else {
+      shader.AdditionalImplementation() << C.GetByOffset(C.BroadcastedIndicesToOffset("vec2(row, col)", output)) << ";\n";
```
qjia7, line 44: Can we always get C's value like below no matter `c_component_` and  `c_is_scalar_`'s value?
```
shader.AdditionalImplementation() << " values += uniforms.beta * vec4<C_element_t>(";
    << C.GetByOffset(C.BroadcastedIndicesToOffset("vec2(row, col)", output)) << ");\n";`

Comment (xiaofeihan1) [reply onnxruntime/core/providers/webgpu/math/gemm_vec4.cc:44]: The reason I wrote with this way is that`BroadcastedIndicesToOffset` requires that the component dimensions of `output` and `C` be identical. 
- When `c` components is 1 but `output` component is 4, broadcast will occur the error. e.g. A(3*4), B(4*4), C(3), Y(3*4).
Given other comment you suggested that making `output` component is 1 or 4 depends on `N % 4 == 0`. I need to check all cases for C and output. That might be more complex than now.

Comment (xiaofeihan1) [reply onnxruntime/core/providers/webgpu/math/gemm_vec4.cc:44]: We need to ensure the the components of `output` and `C` is same. So I will only set `c_and_output_components` to 4 when the last dim of C is `N` and it is divisible by 4. Or it is always is 1.

Comment (xiaofeihan1) [reply onnxruntime/core/providers/webgpu/math/gemm_vec4.cc]: I take changeable output components into account and update the code.

| C's components | output's components | Note|
|---|---|---|
| 1 | 1 | Use broadcast|
| 1 | 4 | C is either {M,1} or {1,1}, so we can use C[i]/C[0] instead of broadcast|
| 4 | 1 | Such a case is impossible.|
| 4 | 4 | Use broadcast |

ps: 
`c_components = (c_last_dim == N && N % 4 == 0) ? 4 : 1;`
`output_components = N % 4 == 0 ? 4 : 1;`

File: onnxruntime/core/providers/webgpu/math/gemm_vec4.cc (hunk 5 above)
Comment (xiaofeihan1) [reply onnxruntime/core/providers/webgpu/math/gemm_vec4.cc]: Thanks for suggestion. Done. We added some code to support this, because which needs to consider the case where output may not be vec4 when B is transpose.

File: onnxruntime/core/providers/webgpu/math/gemm_vec4.cc [hunk 7]
```diff
@@ -0,0 +1,314 @@
...
+
+  if (need_handle_bias) {
+    program.AddInput({c, ProgramTensorMetadataDependency::TypeAndRank, c_components});
+  }
+
+  const uint32_t TILE_SIZE = 32;
+  const uint32_t num_tile_n = (N + TILE_SIZE - 1) / TILE_SIZE;
+  const uint32_t num_tile_m = (M + TILE_SIZE - 1) / TILE_SIZE;
+
+  program.CacheHint(alpha, transA_, transB_)
```
qjia7, line 296: `c_is_scalar` also needs to be added into `CacheHint`?

xiaofeihan1: Thanks for catching this. Added it.

qjia7 (approved):
LGTM, thanks.

fs-eire (approved):

PR #24485, opened 2025-04-21
[webgpu] Fix multihead-attention for ort-web-tests

jchen10:
Compute 'total_sequence_length' the same way as JSEP.

### DIFF
--- a/onnxruntime/contrib_ops/webgpu/bert/attention.cc
+++ b/onnxruntime/contrib_ops/webgpu/bert/attention.cc
@@ -446,7 +446,8 @@ Status ApplyAttention(const Tensor* Q, const Tensor* K, const Tensor* V, const T
                       WebgpuAttentionParameters& parameters, onnxruntime::webgpu::ComputeContext& context, const Tensor* seqlen_k) {
   const int output_count = std::min({context.OutputCount(), 1 + (past_key != nullptr ? 1 : 0) + (past_value != nullptr ? 1 : 0)});
   const int past_sequence_length = output_count > 1 ? parameters.past_sequence_length_ : 0;
-  const int total_sequence_length = past_sequence_length + parameters.kv_sequence_length_;
+  const int total_sequence_length =
+      parameters.is_gqa_ ? parameters.total_sequence_length_ : past_sequence_length + parameters.kv_sequence_length_;
 
   const TensorShapeVector probs_dims({parameters.batch_size_, parameters.num_heads_,
                                       parameters.sequence_length_, total_sequence_length});
--- a/onnxruntime/contrib_ops/webgpu/bert/attention.cc
+++ b/onnxruntime/contrib_ops/webgpu/bert/attention.cc
@@ -446,7 +446,7 @@ Status ApplyAttention(const Tensor* Q, const Tensor* K, const Tensor* V, const T
                       WebgpuAttentionParameters& parameters, onnxruntime::webgpu::ComputeContext& context, const Tensor* seqlen_k) {
   const int output_count = std::min({context.OutputCount(), 1 + (past_key != nullptr ? 1 : 0) + (past_value != nullptr ? 1 : 0)});
   const int past_sequence_length = output_count > 1 ? parameters.past_sequence_length_ : 0;
-  const int total_sequence_length = parameters.total_sequence_length_;
+  const int total_sequence_length = past_sequence_length + parameters.kv_sequence_length_;
 
   const TensorShapeVector probs_dims({parameters.batch_size_, parameters.num_heads_,
                                       parameters.sequence_length_, total_sequence_length});


jchen10:
The corresponding JSEP  code [line](https://github.com/microsoft/onnxruntime/blob/2abab8d39e08819d0e44833f9afee0a37c9aaa75/js/web/lib/wasm/jsep/webgpu/ops/attention.ts#L813).
@xhcao @fs-eire PTAL

onnxruntime/contrib_ops/webgpu/bert/attention.cc
```diff
@@ -446,7 +446,7 @@ Status ApplyAttention(const Tensor* Q, const Tensor* K, const Tensor* V, const T
                       WebgpuAttentionParameters& parameters, onnxruntime::webgpu::ComputeContext& context, const Tensor* seqlen_k) {
   const int output_count = std::min({context.OutputCount(), 1 + (past_key != nullptr ? 1 : 0) + (past_value != nullptr ? 1 : 0)});
   const int past_sequence_length = output_count > 1 ? parameters.past_sequence_length_ : 0;
-  const int total_sequence_length = parameters.total_sequence_length_;
+  const int total_sequence_length = past_sequence_length + parameters.kv_sequence_length_;
```
qjia7, line 449: Will it fix any issues? `past_sequence_length` is equal to `max_sequence_length` when it's static kv cache. The change will make it dispatch many unnecessary workgroups and increase computations which hurt performance. @guschmue ever did it for performance #23615

jchen10: Thanks jiajia.
It fixes the test cases below:
MultiHeadAttention Basic, one head and head-size=1 with pastKey and pastValue T[0]
MultiHeadAttention Basic, one head and head-size=1 with attentionBias, pastKey and pastValue T[0]
MultiHeadAttention Basic, one head and head-size=4 with attentionBias,  and pastValue T[0]

jchen10: The cases are defined [here](https://github.com/microsoft/onnxruntime/blob/2abab8d39e08819d0e44833f9afee0a37c9aaa75/js/web/test/data/ops/multihead-attention.jsonc#L195).

qjia7: For multihead-attention, `parameters.total_sequence_length_` should be equal to `past_sequence_length + parameters.kv_sequence_length_`. Can you check that why they are not  equal? Maybe we should fix there.

guschmue: Agree with Jiajia, I'd expect them to be the same.
My change for parameters.total_sequence_length_ was for GQA when the kv_cache is used. MHA does not support this.

guschmue: But for sure if we change it in attention we'd break kv_cache for GQA.

qjia7: From the cases and implementation, it seems that if `present_key` and `present_value` are not present, the `past_sequence_length` should be zero even `past_key` and `past_value` are set. But `parameters.past_sequence_length_` is not zero due to `past_key` and `past_value` are not empty. I don't know why we implement it like this. @satyajandhyala Do you know the reason? Is [case1](https://github.com/microsoft/onnxruntime/blob/2abab8d39e08819d0e44833f9afee0a37c9aaa75/js/web/test/data/ops/multihead-attention.jsonc#L316) valid? If yes, we should correctly set `parameters.past_sequence_length_` to zero if  `present_key` and `present_value` are empty. Otherwise, modify the test.

satyajandhyala: These are valid tests, as they pass wasm. May be we could make this assignment conditional; something like
`const int total_sequence_length = parameters.is_gqa ? parameters.total_sequence_length_ : past_sequence_length + parameters.kv_sequence_length_;`

or use `parameters.past_present_share_buffer_` flag.

satyajandhyala: > From the cases and implementation, it seems that if `present_key` and `present_value` are not present, the `past_sequence_length` should be zero even `past_key` and `past_value` are set. But `parameters.past_sequence_length_` is not zero due to `past_key` and `past_value` are not empty. I don't know why we implement it like this. @satyajandhyala Do you know the reason? Is [case1](https://github.com/microsoft/onnxruntime/blob/2abab8d39e08819d0e44833f9afee0a37c9aaa75/js/web/test/data/ops/multihead-attention.jsonc#L316) valid? If yes, we should correctly set `parameters.past_sequence_length_` to zero if `present_key` and `present_value` are empty. Otherwise, modify the test.

setting past_sequence_length to zero even if past_key and past_value are present may be not correct.

jchen10: > These are valid tests, as they pass wasm. May be we could make this assignment conditional; something like `const int total_sequence_length = parameters.is_gqa ? parameters.total_sequence_length_ : past_sequence_length + parameters.kv_sequence_length_;`
> 
> or use `parameters.past_present_share_buffer_` flag.

Done!

guschmue: I think this is good for now. One of these days we should cleanup the AttentionParameters struct, becomes to hard to tell what length is used for what.

fs-eire:
some pipeline failed, maybe need to merge to latest main

jchen10:
> some pipeline failed, maybe need to merge to latest main

Thanks for the reminder. Rebased to main!

fs-eire (approved):

PR #24489, opened 2025-04-21
[WebGPU EP] Implements Depth-To-Space Operator

prathikr:
Increases operator coverage for WebGPU EP.

### DIFF
--- a/onnxruntime/core/providers/webgpu/tensor/depth_to_space.cc
+++ b/onnxruntime/core/providers/webgpu/tensor/depth_to_space.cc
@@ -135,7 +135,7 @@ Status DepthToSpace<is_nhwc>::ComputeInternal(onnxruntime::webgpu::ComputeContex
 
   DepthToSpaceProgram program{perm};
   program
-      .AddInput({input, ProgramTensorMetadataDependency::Type, input_override_shape, 1})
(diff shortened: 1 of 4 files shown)


onnxruntime/core/providers/webgpu/tensor/depth_to_space.cc
```diff
@@ -0,0 +1,145 @@
...
+  std::vector<int64_t> shape_after_permutation_vec(6);
+  for (int i = 0; i < 6; i++) {
+    shape_after_permutation_vec[i] = shape[perm[i]];
+  }
+  TensorShape output_override_shape(shape_after_permutation_vec);
+
+  DepthToSpaceProgram program{perm};
+  program
+      .AddInput({input, ProgramTensorMetadataDependency::TypeAndRank, input_override_shape, 1})
+      .AddOutput({output, ProgramTensorMetadataDependency::TypeAndRank, output_override_shape, 1})
```
fs-eire, line 137: since input and output are of the same type `T`, you need either input or output depending on type should be ok.

also, it looks like input and output MUST be rank==4, so maybe remove rank here too. (what will happen if input is not rank-4?)

fs-eire (approved):

fs-eire (approved):

snnn:

PR #24496, opened 2025-04-22
webgpu: fix conv error when using MatMulNaiveProgram

xhcao:

### DIFF
--- a/onnxruntime/core/providers/webgpu/nn/conv.cc
+++ b/onnxruntime/core/providers/webgpu/nn/conv.cc
@@ -198,7 +198,7 @@ Status Conv<is_channels_last, is_fused>::ComputeInternal(ComputeContext& context
           .AddInputs({{matmul_inputs[0], ProgramTensorMetadataDependency::TypeAndRank, ReduceShapeByComponents(matmul_input_reshapes[0], a_components), int(a_components)},
                       {matmul_inputs[1], ProgramTensorMetadataDependency::TypeAndRank, ReduceShapeByComponents(matmul_input_reshapes[1], components), int(components)}});
       if (has_bias) {
-        program.AddInput({bias, ProgramTensorMetadataDependency::Rank, bias->Shape(), components});
+        program.AddInput({bias, ProgramTensorMetadataDependency::Rank, ReduceShapeByComponents(bias->Shape(), components), components});
       }
       program
           .AddOutputs({{output, ProgramTensorMetadataDependency::None, ReduceShapeByComponents(matmul_output_shape, components), int(components)}})


xhcao:
This PR resolves below error when running sd-turbo-ort-web/vae_decoder
[E:onnxruntime:sd-turbo-vae-hf, sequential_executor.cc:572 onnxruntime::ExecuteKernel] Non-zero status code returned while running Conv node. Name:'/post_quant_conv/Conv_token_3' Status Message: shader_helper.cc:192 onnxruntime::webgpu::`anonymous-namespace'::ValidateVariableShape Tensor original shape {4} cannot reshape to {4} with component number 4

@jchen10 PTAL

onnxruntime/core/providers/webgpu/nn/conv.cc
```diff
@@ -198,7 +198,7 @@ Status Conv<is_channels_last, is_fused>::ComputeInternal(ComputeContext& context
           .AddInputs({{matmul_inputs[0], ProgramTensorMetadataDependency::TypeAndRank, ReduceShapeByComponents(matmul_input_reshapes[0], a_components), int(a_components)},
                       {matmul_inputs[1], ProgramTensorMetadataDependency::TypeAndRank, ReduceShapeByComponents(matmul_input_reshapes[1], components), int(components)}});
       if (has_bias) {
-        program.AddInput({bias, ProgramTensorMetadataDependency::Rank, bias->Shape(), components});
+        program.AddInput({bias, ProgramTensorMetadataDependency::Rank, ReduceShapeByComponents(bias->Shape(), components), components});
```
qjia7, line 201:
```suggestion
        program.AddInput({bias, ProgramTensorMetadataDependency::Rank, components});
```
If `const TensorShape& override_shape` is missing and `components` > 1, the ProgramInput::ProgramInput will do it for you to get the correct reduced shape. See https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/providers/webgpu/program.cc#L257-L259

xhcao: @qjia7, your comment implicitly gets `override_shape`, my modification explicitly gets `override_shape`, which calling another constructor function. They are the same, no one is better. And in this file, take line 198|199 for examples, they all gets the reduced shapes explicitly, I want to keep the same code style here.

satyajandhyala: The reason we used override_shape in the case of input and weight is because they might have changed from the original tensor. In the case of bias, we would not have modified the shape to override the original tensor shape. For brevity we could do as @qjia7 suggested.

fs-eire: I am planning a refactor for Program and will normalize this part. I think it is OK to keep it as it is for now.

guschmue (approved):

onnxruntime/core/providers/webgpu/nn/conv.cc
guschmue: ok, merging and refactor once Yulong's change is there.

PR #16935, opened 2023-07-31
Add WAITPKG checks, add support for TPAUSE in ThreadPool spin

kleiti:
### Description
Adding necessary runtime checks for WAITPKG:
onnxruntime/core/common/cpuid_info.cc
onnxruntime/core/common/cpuid_info.h

Added a spin function with TPAUSE, use it in WorkerLoop if HW has WAITPKG support. Defining the control bit as 0, since this will provide better power savings, but also improve performance of other SMT thread on the same core. See https://www.felixcloutier.com/x86/tpause.
include/onnxruntime/core/common/spin_pause.h
include/onnxruntime/core/platform/EigenNonBlockingThreadPool.h

### Motivation and Context
Change enables TPAUSE support on Alder Lake and newer Intel microarchitectures in the context of Eigen-based threadpool implementation. This allows the CPU to enter light-weight C-state while in a spin, saving power and in this case, improving performance by 3-4%.
See section 'Active Spins in Applications' in
https://cdrdv2-public.intel.com/685865/211112_Hybrid_WP_2_Developing_v1.2.pdf

### Result
Tested with onnxruntime_perf_test.exe on a dynamically quantized model (which I can not share - however these changes are not model specific), running 
`onnxruntime_perf_test.exe -I -m times -r 10000 -o 2 model.onnx`

As to optimal TPAUSE cycle count, testing shows that ~2000 cycles provides the best results. First 2 graphs were generated from onnxruntime_perf_test.exe output, power data collected using Intel Power Gadget on ADL i9-12700H.
image
image
image

PR #24524, opened 2025-04-23
Add WAITPKG checks, add support for TPAUSE within SpinPause

omarhass47:
### Description
This change introduces `TPAUSE` support in the `SpinPause()` function in Windows and Linux to reduce power consumption and improve efficiency during spin-wait periods. `TPAUSE` is a lightweight power/performance ISA that goes into an optimized C0 power state while waiting on a delay event, compared to `_mm_pause()` which is a NOP-like instruction that provides a small delay in the CPU Pipeline. With this change, performance of First Inference Latency across certain models can also improve. Models that were tested internally have shown up to ~2x improvement in First Inference Latency and up to ~20% lower overall power consumption.

Genuine Intel CPUID detection logic was also refactored into a shared utility (`CheckIntel()`), enabling consistent platform checks across the codebase. Here `TPAUSE` is enabled by default for architectures that support it.

[Intel Intrinsics Guide (TPAUSE)](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=tpause&techs=MMX,SSE_ALL,AVX_ALL,AVX_512,AMX,SVML,Other&ig_expand=6888,6888)

### Motivation and Context
Performance and power efficiency gains - Previous PR was created which initially introduced the TPAUSE instruction in `SpinPause()` with measured improvements in power (please see previous TPAUSE PR here: [Add WAITPKG checks, add support for TPAUSE in ThreadPool spin #16935](https://github.com/microsoft/onnxruntime/pull/16935)). Additional performance testing and measurements were done across Mobile, Desktop, and Server, influencing enhancements to the PR such as a tweak to the `spin_delay_cycles`, Linux support and the refactored Intel CPUID detection logic.

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -258,13 +258,6 @@ option(onnxruntime_USE_OPENVINO_INTERFACE "Build ONNXRuntime shared lib which is
 option(onnxruntime_USE_VITISAI_INTERFACE "Build ONNXRuntime shared lib which is compatible with Vitis-AI EP interface" OFF)
 option(onnxruntime_USE_QNN_INTERFACE "Build ONNXRuntime shared lib which is compatible with QNN EP interface" OFF)
 
-# Check if the compiler is GCC or Clang
-if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
-    if (CMAKE_SYSTEM_PROCESSOR MATCHES "x86_64" OR CMAKE_SYSTEM_PROCESSOR MATCHES "i[3-6]86")
-        add_compile_options(-mwaitpkg)
-    endif()
-endif()
-
 if("${CMAKE_C_COMPILER_ID}" STREQUAL "GNU" AND CMAKE_C_COMPILER_VERSION VERSION_LESS 11.1)
   message(FATAL_ERROR  "GCC version must be greater than or equal to 11.1")
 endif()
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -258,6 +258,12 @@ option(onnxruntime_USE_OPENVINO_INTERFACE "Build ONNXRuntime shared lib which is
 option(onnxruntime_USE_VITISAI_INTERFACE "Build ONNXRuntime shared lib which is compatible with Vitis-AI EP interface" OFF)
 option(onnxruntime_USE_QNN_INTERFACE "Build ONNXRuntime shared lib which is compatible with QNN EP interface" OFF)
 
+# Check if the compiler is GCC or Clang
+if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU" OR CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+    if (CMAKE_SYSTEM_PROCESSOR MATCHES "x86_64" OR CMAKE_SYSTEM_PROCESSOR MATCHES "i[3-6]86")
+        add_compile_options(-mwaitpkg)
+    endif()
+endif()
 
 if("${CMAKE_C_COMPILER_ID}" STREQUAL "GNU" AND CMAKE_C_COMPILER_VERSION VERSION_LESS 11.1)
   message(FATAL_ERROR  "GCC version must be greater than or equal to 11.1")
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -259,7 +259,7 @@ option(onnxruntime_USE_VITISAI_INTERFACE "Build ONNXRuntime shared lib which is 
 option(onnxruntime_USE_QNN_INTERFACE "Build ONNXRuntime shared lib which is compatible with QNN EP interface" OFF)
 
 # Check if the compiler is GCC or Clang
-if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU" OR CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
     if (CMAKE_SYSTEM_PROCESSOR MATCHES "x86_64" OR CMAKE_SYSTEM_PROCESSOR MATCHES "i[3-6]86")
         add_compile_options(-mwaitpkg)
     endif()
--- a/cmake/onnxruntime_common.cmake
+++ b/cmake/onnxruntime_common.cmake
@@ -11,6 +11,8 @@ set(onnxruntime_common_src_patterns
     "${ONNXRUNTIME_ROOT}/core/common/logging/*.cc"
     "${ONNXRUNTIME_ROOT}/core/common/logging/sinks/*.h"
     "${ONNXRUNTIME_ROOT}/core/common/logging/sinks/*.cc"
+    "${ONNXRUNTIME_ROOT}/core/platform/check_intel.h"
+    "${ONNXRUNTIME_ROOT}/core/platform/check_intel.cc"
     "${ONNXRUNTIME_ROOT}/core/platform/device_discovery.h"
     "${ONNXRUNTIME_ROOT}/core/platform/device_discovery.cc"
     "${ONNXRUNTIME_ROOT}/core/platform/env.h"
--- a/cmake/onnxruntime_common.cmake
+++ b/cmake/onnxruntime_common.cmake
@@ -102,10 +102,12 @@ if(WIN32)
     target_compile_options(onnxruntime_common PRIVATE "/Zc:char8_t-")
   endif()
 elseif(NOT APPLE)
-   set_source_files_properties(
-        ${ONNXRUNTIME_ROOT}/core/common/spin_pause.cc
-        PROPERTIES COMPILE_FLAGS "-mwaitpkg"
-   )
+  if(CMAKE_SYSTEM_PROCESSOR MATCHES "x86_64" AND NOT ANDROID)
+    set_source_files_properties(
+      ${ONNXRUNTIME_ROOT}/core/common/spin_pause.cc
+      PROPERTIES COMPILE_FLAGS "-mwaitpkg"
+    )
+  endif()
 endif()
 if (onnxruntime_USE_TELEMETRY)
   set_target_properties(onnxruntime_common PROPERTIES COMPILE_FLAGS "/FI${ONNXRUNTIME_INCLUDE_DIR}/core/platform/windows/TraceLoggingConfigPrivate.h")
--- a/cmake/onnxruntime_common.cmake
+++ b/cmake/onnxruntime_common.cmake
@@ -99,6 +99,11 @@ if(WIN32)
     set_property(TARGET onnxruntime_common PROPERTY CXX_STANDARD 23)
     target_compile_options(onnxruntime_common PRIVATE "/Zc:char8_t-")
   endif()
+else()
+   set_source_files_properties(
+        ${ONNXRUNTIME_ROOT}/core/common/spin_pause.cc
+        PROPERTIES COMPILE_FLAGS "-mwaitpkg"
+   )
 endif()
 if (onnxruntime_USE_TELEMETRY)
   set_target_properties(onnxruntime_common PROPERTIES COMPILE_FLAGS "/FI${ONNXRUNTIME_INCLUDE_DIR}/core/platform/windows/TraceLoggingConfigPrivate.h")
--- a/cmake/onnxruntime_common.cmake
+++ b/cmake/onnxruntime_common.cmake
@@ -101,7 +101,7 @@ if(WIN32)
     set_property(TARGET onnxruntime_common PROPERTY CXX_STANDARD 23)
     target_compile_options(onnxruntime_common PRIVATE "/Zc:char8_t-")
   endif()
-else()
+elseif(NOT APPLE)
    set_source_files_properties(
         ${ONNXRUNTIME_ROOT}/core/common/spin_pause.cc
         PROPERTIES COMPILE_FLAGS "-mwaitpkg"
--- a/cmake/onnxruntime_common.cmake
+++ b/cmake/onnxruntime_common.cmake
@@ -101,14 +101,15 @@ if(WIN32)
     set_property(TARGET onnxruntime_common PROPERTY CXX_STANDARD 23)
     target_compile_options(onnxruntime_common PRIVATE "/Zc:char8_t-")
   endif()
-elseif(NOT APPLE)
-  if(CMAKE_SYSTEM_PROCESSOR MATCHES "x86_64" AND NOT ANDROID)
+endif()
+
+if(NOT WIN32 AND NOT APPLE AND NOT ANDROID AND CMAKE_SYSTEM_PROCESSOR MATCHES "x86_64")
     set_source_files_properties(
       ${ONNXRUNTIME_ROOT}/core/common/spin_pause.cc
       PROPERTIES COMPILE_FLAGS "-mwaitpkg"
     )
-  endif()
 endif()
+
 if (onnxruntime_USE_TELEMETRY)
   set_target_properties(onnxruntime_common PROPERTIES COMPILE_FLAGS "/FI${ONNXRUNTIME_INCLUDE_DIR}/core/platform/windows/TraceLoggingConfigPrivate.h")
 endif()
--- a/docs/ORTModule_Training_Guidelines.md
+++ b/docs/ORTModule_Training_Guidelines.md
@@ -208,6 +208,17 @@ debugging).
 	export ORTMODULE_ENABLE_COMPUTE_OPTIMIZER=0 # Disable
 	```
 
+#### ORT_TPAUSE
+- **Feature Area**: *ORTMODULE/Optimizations*
+- **Description**: By default, this is enabled. This env var enables the use of the `TPAUSE` instruction for spin-wait loops on supported x86_64 architectures. This
+instruction allows the processor to enter a low-power state during spin-wait periods, leading to reduced power consumption
+and improved efficiency.
+
+	```bash
+	export ORT_TPAUSE=1 # Enable
+	export ORT_TPAUSE=0 # Disable
+	```
+
 #### ORTMODULE_PRINT_INPUT_DENSITY
 
 - **Feature Area**: *ORTMODULE/RuntimeInspector*
--- a/docs/ORTModule_Training_Guidelines.md
+++ b/docs/ORTModule_Training_Guidelines.md
@@ -208,17 +208,6 @@ debugging).
 	export ORTMODULE_ENABLE_COMPUTE_OPTIMIZER=0 # Disable
 	```
 
-#### ORT_TPAUSE
-- **Feature Area**: *ORTMODULE/Optimizations*
-- **Description**: By default, this is enabled. This env var enables the use of the `TPAUSE` instruction for spin-wait loops on supported x86_64 architectures. This
-instruction allows the processor to enter a low-power state during spin-wait periods, leading to reduced power consumption
-and improved efficiency.
-
-	```bash
-	export ORT_TPAUSE=1 # Enable
-	export ORT_TPAUSE=0 # Disable
-	```
-
 #### ORTMODULE_PRINT_INPUT_DENSITY
 
 - **Feature Area**: *ORTMODULE/RuntimeInspector*
--- a/include/onnxruntime/core/common/spin_pause.h
+++ b/include/onnxruntime/core/common/spin_pause.h
@@ -3,10 +3,6 @@
 
 #pragma once
 
-#if defined(_M_AMD64) || defined(__x86_64__)
-#include <cstdint>
-#endif
-
 namespace onnxruntime {
 namespace concurrency {
 
--- a/include/onnxruntime/core/common/spin_pause.h
+++ b/include/onnxruntime/core/common/spin_pause.h
@@ -11,7 +11,53 @@
 #include <xmmintrin.h>
 #endif
 
+#if defined(_M_AMD64) || defined(__x86_64__)
+#if defined(__linux__)
+#include <x86intrin.h>
+#include <immintrin.h>
+#include <cstdlib>
+#elif defined _WIN32
+#include <Windows.h>
+#endif
+#include <cstdint>
+#include "core/common/cpuid_info.h"
+#endif
+
 namespace onnxruntime {
+
+#if defined(_M_AMD64) || defined(__x86_64__)
+const bool tpause = CPUIDInfo::GetCPUIDInfo().HasTPAUSE();
+const std::uint64_t spin_delay_cycles = 1000;
+
+// Check if TPAUSE Environment Variable is set
+const bool tpause_enabled = []() -> bool {
+  bool enabled = false;
+  std::string env_value;
+
+#ifdef _WIN32
+  if (!tpause) return false;  // Skip if CPU doesn't support TPAUSE
+
+  char buffer[256];
+  if (GetEnvironmentVariableA("ORT_TPAUSE", buffer, 256) != 0) {
+    env_value = buffer;
+  }
+#else
+  const auto env = getenv("ORT_TPAUSE");
+  if (env != nullptr) {
+    env_value = env;
+  }
+#endif
+
+  if (env_value.empty()) {
+    enabled = true;
+  } else {
+    enabled = (env_value == "1");
+  }
+
+  return enabled;
+}();
+
+#endif
 
 namespace concurrency {
 
@@ -19,7 +65,21 @@ namespace concurrency {
 
 inline void SpinPause() {
 #if defined(_M_AMD64) || defined(__x86_64__)
+#if defined(_WIN32)
+  if (tpause_enabled) {
+    _tpause(0x0, __rdtsc() + spin_delay_cycles);
+  } else {
+    _mm_pause();
+  }
+#elif defined(__linux__)
+  if (tpause_enabled && __builtin_cpu_supports("waitpkg")) {
+    __builtin_ia32_tpause(0x0, __rdtsc() + spin_delay_cycles);
+  } else {
+    _mm_pause();
+  }
+#else
   _mm_pause();
+#endif
 #endif
 }
 
--- a/include/onnxruntime/core/common/spin_pause.h
+++ b/include/onnxruntime/core/common/spin_pause.h
@@ -3,86 +3,15 @@
 
 #pragma once
 
-#if defined(_M_AMD64)
-#include <intrin.h>
-#endif
-
-#if defined(__x86_64__)
-#include <xmmintrin.h>
-#endif
-
 #if defined(_M_AMD64) || defined(__x86_64__)
-#if defined(__linux__)
-#include <x86intrin.h>
-#include <immintrin.h>
-#include <cstdlib>
-#elif defined _WIN32
-#include <Windows.h>
-#endif
 #include <cstdint>
-#include "core/common/cpuid_info.h"
 #endif
 
 namespace onnxruntime {
-
-#if defined(_M_AMD64) || defined(__x86_64__)
-const bool tpause = CPUIDInfo::GetCPUIDInfo().HasTPAUSE();
-const std::uint64_t spin_delay_cycles = 1000;
-
-// Check if TPAUSE Environment Variable is set
-const bool tpause_enabled = []() -> bool {
-  bool enabled = false;
-  std::string env_value;
-
-#ifdef _WIN32
-  if (!tpause) return false;  // Skip if CPU doesn't support TPAUSE
-
-  char buffer[256];
-  if (GetEnvironmentVariableA("ORT_TPAUSE", buffer, 256) != 0) {
-    env_value = buffer;
-  }
-#else
-  const auto env = getenv("ORT_TPAUSE");
-  if (env != nullptr) {
-    env_value = env;
-  }
-#endif
-
-  if (env_value.empty()) {
-    enabled = true;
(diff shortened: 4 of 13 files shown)


include/onnxruntime/core/common/spin_pause.h
```diff
@@ -11,16 +11,75 @@
...
+
 namespace concurrency {
 
 // Intrinsic to use in spin-loops
 
 inline void SpinPause() {
 #if defined(_M_AMD64) || defined(__x86_64__)
+#if defined(_WIN32)
+  if (tpause_enabled) {
+    _tpause(0x0, __rdtsc() + 1000);
```
liqunfu, line 69: make a variable: spin_delay_cycles

liqunfu: in the picture and also from: https://github.com/microsoft/onnxruntime/pull/16935, "testing shows that ~2000 cycles provides the best results". Why is changes to 1000 hear?

liqunfu (changes requested):

include/onnxruntime/core/common/spin_pause.h
omarhass47: We tested across a variety of models across desktop, mobile and server and found that across all these different types of systems and different spin_pause values, and we found that spin_pause=1000 showed the best performance in FIL gains across the models we tested with, while also not affecting performance of any other operation in the E2E execution.

liqunfu (approved):

cmake/CMakeLists.txt
```diff
@@ -255,6 +255,12 @@ option(onnxruntime_USE_OPENVINO_INTERFACE "Build ONNXRuntime shared lib which is
 option(onnxruntime_USE_VITISAI_INTERFACE "Build ONNXRuntime shared lib which is compatible with Vitis-AI EP interface" OFF)
 option(onnxruntime_USE_QNN_INTERFACE "Build ONNXRuntime shared lib which is compatible with QNN EP interface" OFF)
 
+# Check if the compiler is GCC or Clang
+if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+    if (CMAKE_SYSTEM_PROCESSOR MATCHES "x86_64" OR CMAKE_SYSTEM_PROCESSOR MATCHES "i[3-6]86")
+        add_compile_options(-mwaitpkg)
```
edgchen1, line 261: can we limit this to just the source files that need it? it might be better to move the implementation of SpinPause() out of the header.

docs/ORTModule_Training_Guidelines.md
```diff
@@ -208,6 +208,17 @@ debugging).
 	export ORTMODULE_ENABLE_COMPUTE_OPTIMIZER=0 # Disable
 	```
 
+#### ORT_TPAUSE
+- **Feature Area**: *ORTMODULE/Optimizations*
+- **Description**: By default, this is enabled. This env var enables the use of the `TPAUSE` instruction for spin-wait loops on supported x86_64 architectures. This
```
edgchen1, line 213: this is not related to ORTModule, so this does not seem like the right place to document it.

taking a step back, is there a good reason to not enable it?

onnxruntime/core/platform/posix/hardware_core_enumerator_linux.cc
```diff
@@ -0,0 +1,51 @@
+#include "hardware_core_enumerator_linux.h"
+
+#ifdef __linux__
+#include <cpuid.h>
+#include <vector>
+
+namespace onnxruntime {
+
+IntelChecks checkIntel() {
+  IntelChecks intelCheck = {false, false};
```
edgchen1, line 10: general: there are some naming convention inconsistencies. for example, the convention is `variable_name` and `FunctionName`. see https://google.github.io/styleguide/cppguide.html#Naming

docs/ORTModule_Training_Guidelines.md
omarhass47: @edgchen1 where would you recommend to add the documentation for TPAUSE? As for reasons for not enabling there are no noticeable regressions in any other metrics, hence why we enabled TPAUSE by default.

cmake/CMakeLists.txt
omarhass47: Done! Just pushed a new commit which moves implementation of SpinPause out of the header to `spin_pause.cc` with the `-mwaitpkg` compiler flag being added now to only `spin_pause.cc`.

onnxruntime/core/platform/posix/hardware_core_enumerator_linux.cc
omarhass47: @edgchen1 Thanks! I've just updated the naming conventions used, please let me know if you still have any feedback on this :)

docs/ORTModule_Training_Guidelines.md
edgchen1: > As for reasons for not enabling there are no noticeable regressions in any other metrics, hence why we enabled TPAUSE by default.

it's good that there are no known regressions. so, if I understand correctly, there's no known reason to disable it?

if we do need to provide a way to disable it, I'd prefer not using an environment variable to control core framework behavior if possible. as an alternative example, we have APIs to configure threadpool related options, like this:
https://github.com/microsoft/onnxruntime/blob/acb8ffd5df0d6c4c6b12610807d02e75ef9a70ed/include/onnxruntime/core/session/onnxruntime_c_api.h#L2591
or it could be a compile-time option.

include/onnxruntime/core/common/spin_pause.h
```diff
@@ -3,26 +3,15 @@
...
 #pragma once
 
-#if defined(_M_AMD64)
-#include <intrin.h>
-#endif
-
-#if defined(__x86_64__)
-#include <xmmintrin.h>
+#if defined(_M_AMD64) || defined(__x86_64__)
+#include <cstdint>
```
edgchen1, line 7: we don't really need to include anything here now, right?

onnxruntime/core/common/spin_pause.cc
```diff
@@ -0,0 +1,86 @@
...
+#include <cstdlib>
+#elif defined(_WIN32)
+#include <Windows.h>
+#endif
+#include <cstdint>
+#include <string>
+#include "core/common/cpuid_info.h"
+#endif
+
+#include "core/common/spin_pause.h"
```
edgchen1, line 25: nit: this one should be first.

https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

onnxruntime/core/common/spin_pause.cc
```diff
@@ -0,0 +1,86 @@
...
+// Check if TPAUSE Environment Variable is set
+static const bool tpause_enabled = []() -> bool {
+  bool enabled = false;
+  std::string env_value;
+
+#ifdef _WIN32
+  if (!tpause) return false;  // Skip if CPU doesn't support TPAUSE
+
+  char buffer[256];
+  if (GetEnvironmentVariableA("ORT_TPAUSE", buffer, 256) != 0) {
```
edgchen1, line 42: I don't suggest using an environment variable for this, but for completeness, we do have a cross-platform helper function to read one:
https://github.com/microsoft/onnxruntime/blob/acb8ffd5df0d6c4c6b12610807d02e75ef9a70ed/onnxruntime/core/platform/env.h#L271-L275

onnxruntime/core/platform/posix/hardware_core_enumerator_linux.cc
edgchen1, line 1: add copyright header

onnxruntime/core/platform/posix/hardware_core_enumerator_linux.h
edgchen1, line 1: add copyright header

onnxruntime/core/platform/posix/hardware_core_enumerator_linux.h
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+namespace onnxruntime {
+struct IntelChecks {
+  bool is_Intel;
+  bool isIntel_SpecifiedPlatform;
```
edgchen1, line 6:
```suggestion
  bool is_intel;
  bool is_intel_specified_platform;
```

onnxruntime/core/platform/posix/hardware_core_enumerator_linux.h
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+namespace onnxruntime {
+struct IntelChecks {
```
edgchen1, line 4: would `CheckIntelResult` be a clearer name for the type?

onnxruntime/core/platform/windows/hardware_core_enumerator.cc
```diff
@@ -43,6 +43,39 @@ uint32_t CountSetBits(DWORD input) {
...
+      0xc065,  // ARL-H
+      0xb065   // ARL-U
+  };
+  const int kVendorID_Intel[3] = {0x756e6547, 0x6c65746e, 0x49656e69};  // "GenuntelineI"
+  int regs_leaf0[4];
+  int regs_leaf1[4];
+  __cpuid(regs_leaf0, 0);
+  __cpuid(regs_leaf1, 0x1);
+
+  auto is_Intel = (kVendorID_Intel[0] == regs_leaf0[1]) && (kVendorID_Intel[1] == regs_leaf0[2]) && (kVendorID_Intel[2] == regs_leaf0[3]);
```
edgchen1, line 61: nit: please keep line lengths to a maximum of 120 chars

onnxruntime/core/platform/posix/hardware_core_enumerator_linux.cc
```diff
@@ -0,0 +1,51 @@
...
+
+  __get_cpuid(1, &regs[0], &regs[1], &regs[2], &regs[3]);
+
+  unsigned int base_family = (regs[0] >> 8) & 0xF;
+  unsigned int base_model = (regs[0] >> 4) & 0xF;
+  unsigned int extended_model = (regs[0] >> 16) & 0xF;
+
+  unsigned int model = (base_family == 0x6 || base_family == 0xF) ? (base_model + (extended_model << 4)) : base_model;
+
+  const std::vector<unsigned int> kVendorID_IntelSpecifiedPlatformIDs = {
```
edgchen1, line 31: don't need to use a vector here
```suggestion
  const unsigned int kVendorID_IntelSpecifiedPlatformIDs[3] = {
```

omarhass47: Hey @edgchen1 the reason why a vector is used here rather than an array is so that `kVendorID_IntelSpecifiedPlatformIDs` can be expanded on for future platform releases, so here it makes sense to use a vector over an array.

edgchen1: a vector makes sense if the collection is dynamic at runtime. here, it seems like it could be a compile-time constant (`constexpr` might be good too).

I think that this would also work, and make it easier to update the code with additional entries later on:
```suggestion
  const unsigned int kVendorID_IntelSpecifiedPlatformIDs[] = {
```

onnxruntime/core/platform/posix/hardware_core_enumerator_linux.h
omarhass47: Agreed! Change has been made.

docs/ORTModule_Training_Guidelines.md
omarhass47: Thanks @edgchen1! Thanks for the clarification, in that case then I've for now removed the environment variable check for TPAUSE based on the performance testing we've done. Please let me know if you have any other thoughts on this.

onnxruntime/core/common/spin_pause.cc
```diff
@@ -0,0 +1,54 @@
...
+void SpinPause() {
+#if defined(_M_AMD64) || defined(__x86_64__)
+#if defined(_WIN32)
+  if (tpause) {
+    _tpause(0x0, __rdtsc() + spin_delay_cycles);
+  } else {
+    _mm_pause();
+  }
+#elif defined(__linux__)
+  if (__builtin_cpu_supports("waitpkg")) {
```
edgchen1, line 42: is it possible to move this into the logic that sets `CPUIDInfo::has_tpause_` so we can just check `tpause` here?

onnxruntime/core/common/spin_pause.cc
```diff
@@ -0,0 +1,54 @@
...
+#include <immintrin.h>
+#include <cstdlib>
+#endif
+#endif
+
+namespace onnxruntime {
+
+#if defined(_M_AMD64) || defined(__x86_64__)
+static const bool tpause = CPUIDInfo::GetCPUIDInfo().HasTPAUSE();
+static const std::uint64_t spin_delay_cycles = 1000;
```
edgchen1, line 27: maybe make these function-local static variables instead. then we won't have to worry about static initialization order issues.

edgchen1, line 27: also some naming and declaration suggestions
```suggestion
static const bool has_tpause = CPUIDInfo::GetCPUIDInfo().HasTPAUSE();
static constexpr uint64_t tpause_spin_delay_cycles = 1000;
```

onnxruntime/core/platform/posix/hardware_core_enumerator_linux.cc
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include "hardware_core_enumerator_linux.h"
+
+#ifdef __linux__
```
edgchen1, line 6: should we have a fallback implementation for `!defined(__linux__)`? maybe it can just return a default unsupported value. but it's probably better to have something, otherwise we may encounter link errors.

onnxruntime/core/platform/posix/hardware_core_enumerator_linux.h
```diff
@@ -0,0 +1,13 @@
...
+
+#pragma once
+
+namespace onnxruntime {
+struct CheckIntelResult {
+  bool is_intel;
+  bool is_intel_specified_platform;
+};
+
+CheckIntelResult CheckIntel();
```
edgchen1, line 12: I know that this was refactored out of the Windows hardware_core_enumerator.h/cc file, but now it doesn't seem directly related to a "hardware core enumerator" anymore. maybe rename the files to something more specific, like check_intel.h/cc? then, we could even consolidate the header into a single one for both Windows and POSIX.

onnxruntime/core/platform/windows/hardware_core_enumerator.cc
```diff
@@ -85,30 +121,11 @@ uint32_t HardwareCoreEnumerator::DefaultIntraOpNumThreads() {
...
-  __cpuid(regs_leaf1, 0x1);
-
-  auto isIntel = (kVendorID_Intel[0] == regs_leaf0[1]) && (kVendorID_Intel[1] == regs_leaf0[2]) && (kVendorID_Intel[2] == regs_leaf0[3]);
-
-  for (int intelSpecifiedPlatform : kVendorID_IntelSpecifiedPlatformIDs) {
-    if ((regs_leaf1[0] >> 4) == intelSpecifiedPlatform) {
-      isIntelSpecifiedPlatform = true;
-    }
-  }
+  CheckIntelResult check_Intel = CheckIntel();
```
edgchen1, line 125: nit: can we consistently name the result of `CheckIntel()`, `intel_check` or `check_intel`?

onnxruntime/core/common/spin_pause.cc
omarhass47: Absolutely, latest commit moves this logic to `CPUIDInfo`, and so now `has_tpause` can be used for Windows and Linux. Thanks!

onnxruntime/core/common/spin_pause.cc
omarhass47: Great suggestion, thanks for providing it! Moved to inside `SpinPause()` function in the latest commit.

omarhass47: @edgchen1 names are updates in latest commit, thanks! :)

onnxruntime/core/platform/posix/hardware_core_enumerator_linux.cc
omarhass47: @edgchen1 so now in the latest commit the logic is all consolidated in `check_intel.cc` such that if `!defined(_WIN32)` or `!defined(__linux__)` then it will just return false for the Intel Check. Let me know if you have any other thoughts on this :)

onnxruntime/core/platform/posix/hardware_core_enumerator_linux.h
omarhass47: @edgchen1 this is a great suggestion thanks a bunch! I've now consolidated to `check_intel.cc` and `check_intel.h` for the Genuine Intel Check, and deleted the previous files that were under POSIX. Just tested and confirmed that the logic still works on Windows and Linux. Let me know if you have anymore feedback on this, thanks!

onnxruntime/core/platform/windows/hardware_core_enumerator.cc
omarhass47: Definitely! Latest commit should reflect the updated name.

onnxruntime/core/common/cpuid_info.cc
```diff
@@ -134,6 +135,24 @@ void CPUIDInfo::X86Init() {
...
+            has_tpause_ = false;
+          }
+
+          #ifdef __linux__
+          if(__builtin_cpu_supports("waitpkg")) {
+            has_tpause_ = true;
+          } else {
+            has_tpause_ = false;
+          }
+          #endif
```
edgchen1, line 154: the `#ifdef __linux__` part overwrites the previous value of `has_tpause_`. could we simplify it to this?

```suggestion
#ifdef __linux__
          has_tpause_ = __builtin_cpu_supports("waitpkg") != 0;
#else
          has_tpause_ = (data[2] & (1 << 5)) != 0;
#endif
```

onnxruntime/core/common/spin_pause.cc
```diff
@@ -0,0 +1,48 @@
...
+#endif
+
+namespace onnxruntime {
+namespace concurrency {
+
+// Intrinsic to use in spin-loops
+void SpinPause() {
+#if defined(_M_AMD64) || defined(__x86_64__)
+  static const bool has_tpause = CPUIDInfo::GetCPUIDInfo().HasTPAUSE();
+  static constexpr std::uint64_t tpause_spin_delay_cycles = 1000;
```
edgchen1, line 29: nit
```suggestion
  static constexpr uint64_t tpause_spin_delay_cycles = 1000;
```

onnxruntime/core/common/spin_pause.cc
```diff
@@ -0,0 +1,48 @@
...
+  if (has_tpause) {
+#if defined(_WIN32)
+  _tpause(0x0, __rdtsc() + tpause_spin_delay_cycles);
+#elif defined(__linux__) && !defined(__aarch64__)
```
edgchen1, line 33: is `defined(__aarch64__)` possible here? we're already within `defined(_M_AMD64) || defined(__x86_64__)`

onnxruntime/core/common/spin_pause.cc
```diff
@@ -0,0 +1,48 @@
...
+  __builtin_ia32_tpause(0x0, __rdtsc() + tpause_spin_delay_cycles);
+#else
+    _mm_pause();
+#endif
+  } else {
+    _mm_pause();
+  }
+
+#else
+  _mm_pause();
```
edgchen1, line 43: so this is `!(defined(_M_AMD64) || defined(__x86_64__))`, right? is it possible that `_mm_pause()` might not be available then?

onnxruntime/core/platform/check_intel.cc
```diff
@@ -0,0 +1,92 @@
...
+  is_intel =
+      (kVendorID_Intel[0] == regs_leaf0[1]) &&
+      (kVendorID_Intel[1] == regs_leaf0[2]) &&
+      (kVendorID_Intel[2] == regs_leaf0[3]);
+
+  if (!is_intel) {
+    return intel_check;  // if not an Intel CPU, return early
+  }
+
+  for (int intelSpecifiedPlatform : kVendorID_IntelSpecifiedPlatformIDs) {
```
edgchen1, line 42: nit: `kVendorID_IntelSpecifiedPlatformIDs` is an `unsigned int[]`
```suggestion
  for (auto intelSpecifiedPlatform : kVendorID_IntelSpecifiedPlatformIDs) {
```

edgchen1 (commented):
thanks for addressing the comments. would it also be possible to include some measurements in the PR description that show the performance improvements from this change?

onnxruntime/core/common/spin_pause.cc
omarhass47: I noticed in the CI/CD that the Android CI failed due to a difference in how the tpause intrinsic is used, so the thought here was to just include the `!defined(__aarch64__)` flag to attempt to exclude that execution path for Android. Would you recommend a different alternative?

onnxruntime/core/platform/check_intel.cc
omarhass47: Thanks for the suggestion!! Fixed in the latest commit

onnxruntime/core/common/spin_pause.cc
omarhass47: Ah yes! Referring back to the original source as well it's wrapped around `defined(_M_AMD64) || defined(__x86_64__)`, fixed in the latest commits thanks!

onnxruntime/core/common/spin_pause.cc
edgchen1: I see. `defined(__ANDROID__)` would be a more direct way to detect Android.

onnxruntime/core/common/cpuid_info.cc
```diff
@@ -134,6 +135,15 @@ void CPUIDInfo::X86Init() {
         // avx512_skylake = avx512f | avx512vl | avx512cd | avx512bw | avx512dq
         has_avx512_skylake_ = has_avx512 && (data[1] & ((1 << 16) | (1 << 17) | (1 << 28) | (1 << 30) | (1 << 31)));
         is_hybrid_ = (data[3] & (1 << 15));
+        // Check for TPAUSE
+        CheckIntelResult check_intel = CheckIntel();
+        if (check_intel.is_intel) {
+        #ifdef __linux__
```
edgchen1, line 141: the formatting PR check is complaining. to fix the formatting, you can either run clang-format or lintrunner:
https://github.com/microsoft/onnxruntime/blob/main/docs/Coding_Conventions_and_Standards.md#linting

omarhass47: Ran lintrunner with latest commit, thanks @edgchen1!

onnxruntime/core/common/spin_pause.cc
omarhass47: Added in the latest commit, thanks for the suggestion! :)

omarhass47:
Hi @edgchen1 thanks a ton for your patience, there are many errors that were generated by the CI/CD that I did not see when building on Windows, Linux or MacOS. Hopefully the latest commit will resolve the remaining warnings/errors that the CI/CD was displaying.

edgchen1:
it looks like there are still a few failing CI builds. e.g., the Linux arm64 build:
https://github.com/microsoft/onnxruntime/actions/runs/15074705006/job/42392349724?pr=24524

```
/onnxruntime_src/onnxruntime/core/platform/check_intel.cc: In function ‘onnxruntime::CheckIntelResult onnxruntime::CheckIntel()’:
/onnxruntime_src/onnxruntime/core/platform/check_intel.cc:53:3: error: ‘__get_cpuid’ was not declared in this scope
   53 |   __get_cpuid(0, &regs[0], &regs[1], &regs[2], &regs[3]);
      |   ^~~~~~~~~~~
```

omarhass47:
> it looks like there are still a few failing CI builds. e.g., the Linux arm64 build: https://github.com/microsoft/onnxruntime/actions/runs/15074705006/job/42392349724?pr=24524
> 
> ```
> /onnxruntime_src/onnxruntime/core/platform/check_intel.cc: In function ‘onnxruntime::CheckIntelResult onnxruntime::CheckIntel()’:
> /onnxruntime_src/onnxruntime/core/platform/check_intel.cc:53:3: error: ‘__get_cpuid’ was not declared in this scope
>    53 |   __get_cpuid(0, &regs[0], &regs[1], &regs[2], &regs[3]);
>       |   ^~~~~~~~~~~
> ```

Thanks @edgchen1! I've just added checks for `#if (defined(_M_AMD64) || defined(__x86_64__))` in `check_intel.cc`, that should resolve this issue. :)

omarhass47:
Hi @edgchen1! Do you have any recommendations on how to resolve the Lint error the CI/CD is throwing? When running `lintrunner -a` in my local repo I just get the following message:
```
ok No lint issues.
Successfully applied all patches.
```

Please let me know if you have any other feedback/comments on the PR as it stands, I'll look into squashing the commits once you give the okay that it's merge ready :) thanks!

onnxruntime/core/platform/check_intel.cc
```diff
@@ -0,0 +1,96 @@
...
+      (kVendorID_Intel[2] == static_cast<unsigned int>(regs_leaf0[3]));
+
+  if (!is_intel) {
+    return intel_check;  // if not an Intel CPU, return early
+  }
+
+  for (auto intelSpecifiedPlatform : kVendorID_IntelSpecifiedPlatformIDs) {
+    if (static_cast<unsigned int>(regs_leaf1[0] >> 4) == intelSpecifiedPlatform) {
+      is_intel_specified_platform = true;
+    }
```
edgchen1, line 48: - nit: variable naming
- perhaps we should use unsigned right shift. signed right shift would fill in leading 1's if the sign bit is set.
- early return?
```suggestion
  for (auto intel_specified_platform : kVendorID_IntelSpecifiedPlatformIDs) {
    if ((static_cast<unsigned int>(regs_leaf1[0]) >> 4) == intel_specified_platform) {
      is_intel_specified_platform = true;
      break;
    }
```

onnxruntime/core/platform/check_intel.cc
```diff
@@ -0,0 +1,96 @@
...
+  };
+
+  for (auto id : kVendorID_IntelSpecifiedPlatformIDs) {
+    if (model == id) {
+      is_intel_specified_platform = true;
+      break;
+    }
+  }
+#endif  //__linux__
+#endif //(_M_AMD64) || (__x86_64__)
```
edgchen1, line 88: I think this should make the formatting check happy. if not, try running `clang-format -i` on this file.
```suggestion
#endif  // (_M_AMD64) || (__x86_64__)
```

edgchen1:
> Hi @edgchen1! Do you have any recommendations on how to resolve the Lint error the CI/CD is throwing? When running `lintrunner -a` in my local repo I just get the following message:
> 
> ```
> ok No lint issues.
> Successfully applied all patches.
> ```

Hm, you can try running clang-format directly on the file it's complaining about. I think clang-format is installed as one of lintrunner's dependencies.

cmake/onnxruntime_common.cmake
```diff
@@ -99,6 +101,13 @@ if(WIN32)
     set_property(TARGET onnxruntime_common PROPERTY CXX_STANDARD 23)
     target_compile_options(onnxruntime_common PRIVATE "/Zc:char8_t-")
   endif()
+elseif(NOT APPLE)
+  if(CMAKE_SYSTEM_PROCESSOR MATCHES "x86_64" AND NOT ANDROID)
+    set_source_files_properties(
+      ${ONNXRUNTIME_ROOT}/core/common/spin_pause.cc
+      PROPERTIES COMPILE_FLAGS "-mwaitpkg"
+    )
+  endif()
```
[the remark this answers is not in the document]
omarhass47: Thanks @edgchen1! I moved to a separate if block as recommended in the latest commit.

edgchen1 (approved):

omarhass47:
> @omarhass47 thanks for your patience. I think we're almost ready to merge this PR. would you mind merging from main? it might fix the remaining CI build issues.

Hey @edgchen1 great to hear! Sure thing, I just re-based my branch with main and pushed.

edgchen1:
@omarhass47 thanks for the contribution!

omarhass47:
> @omarhass47 thanks for the contribution!

Happy to do it @edgchen1, and thank you for all your feedback and comments!

wdconinc:
It appears that GCC and Clang implement `__builtin_ia32_tpause` on linux with a different number of arguments:
- GCC: 2-argument `__builtin_ia32_tpause`, e.g. https://github.com/gcc-mirror/gcc/blob/releases/gcc-15.2.0/gcc/config/i386/waitpkgintrin.h#L51-L56
- Clang: 3-argument `__builtin_ia32_tpause`, e.g. https://github.com/llvm/llvm-project/blob/llvmorg-21.1.8/clang/lib/Headers/waitpkgintrin.h#L33-L38

The explicit use of `__builtin_ia32_tpause` introduced in this PR uses the 2-argument version and prevents compilation with clang on linux. Is there a reason to avoid the `_tpause` wrapper (which has the same 2-argument signature in both) on linux, as it is done on WIN32?

edgchen1:
> Is there a reason to avoid the _tpause wrapper (which has the same 2-argument signature in both) on linux, as it is done on WIN32?

Not that I'm aware of. @omarhass47 do you know?

PR #20282, opened 2024-04-11
Fix shape conv fuse opt

gedoensmax:
### Description

FIx:
- Multiples Convs into an Add+Relu will fuse the op although intermediates are needed @tianleiwu for viz
   image
- Also fixes an issue with Shape Initializers Merge as input, that occurs when the input initializer is the same across multiple nodes but not all nodes are Shape nodes. @er3x3 for viz

PR #19470, opened 2024-02-08
Adding CUDNN Frontend and use for CUDA NN Convolution

JTischbein:
### Description
Added CUDNN Frontend and used it for NHWC convolutions, and optionally fuse activation.

#### Backward compatible 
- For model existed with FusedConv, model can still run. 
- If ORT is built with cuDNN 8, cuDNN frontend will not be built into binary. Old kernels (using cudnn backend APIs) are used.

#### Major Changes
- For cuDNN 9, we will enable cudnn frontend to fuse convolution and bias when a provider option `fuse_conv_bias=1`.
- Remove the fusion of FusedConv from graph transformer for CUDA provider, so there will not be FusedConv be added to graph for CUDA EP in the future.
- Update cmake files regarding to cudnn settings. The search order of CUDNN installation in build are like the following:
  * environment variable `CUDNN_PATH`
  * `onnxruntime_CUDNN_HOME` cmake extra defines. If a build starts from build.py/build.sh, user can pass it through `--cudnn_home` parameter, or by environment variable `CUDNN_HOME` if `--cudnn_home` not used.
  * cudnn python package installation directory like python3.xx/site-packages/nvidia/cudnn
  * CUDA installation path

#### Potential Issues

- If ORT is built with cuDNN 8, FusedConv fusion is no longer done automatically, so some model might have performance regression.  If user still wants FusedConv operator for performance reason, they can still have multiple ways to walkaround: like use older version of onnxruntime; or use older version of ORT to save optimized onnx, then run with latest version of ORT. We believe that majority users have moved to cudnn 9 when 1.20 release (since the default in ORT and PyTorch is cudnn 9 for 3 months when 1.20 release), so the impact is small.
- cuDNN graph uses TF32 by default, and user cannot disable TF32 through the use_tf32 cuda provider option. If user encounters accuracy issue (like in testing), user has to set environment variable `NVIDIA_TF32_OVERRIDE=0` to disable TF32. Need update the document of use_tf32 later.

#### Follow ups
This is one of PRs that target to enable NHWC convolution in CUDA EP by default if device supports it. There are other changes will follow up to make it possible.
(1) Enable `prefer_nhwc` by default for device with sm >= 70. 
(2) Change `fuse_conv_bias=1` by default after more testing.
(3) Add other NHWC operators (like Resize or UpSample).

### Motivation and Context

[...]

PR #24525, opened 2025-04-24
Remove non-constant input check from ConvActivationFusion

### LINKED ISSUE #24473: ORT aborts when loading the attached model
### Describe the issue

[resnet18.zip](https://github.com/user-attachments/files/19819823/resnet18.zip)

When loading the model, inference session aborts without an informative error message.

### To reproduce

```py
import onnxruntime as ort

ort.InferenceSession("resnet18.onnx")
```

yeilds

```
/opt/rh/gcc-toolset-14/root/usr/include/c++/14/bits/stl_vector.h:1149: std::vector<_Tp, _Alloc>::const_reference std::vector<_Tp, _Alloc>::operator[](size_type) const [with _Tp = onnxruntime::NodeArg*; _Alloc = std::allocator<onnxruntime::NodeArg*>; const_reference = onnxruntime::NodeArg* const&; size_type = long unsigned int]: Assertion '__n < this->size()' failed.
fish: Job 1, 'python' terminated by signal SIGABRT (Abort)
```

### Urgency

_No response_

### Platform

Linux

### OS Version

ubuntu

### ONNX Runtime Installation

Released Package

1.21.0

### ONNX Runtime API

Python

### Architecture

X64

### Execution Provider

Default CPU

_No response_

edgchen1:
### Description

An additional check for non-constant inputs was added to ConvActivationFusion in #20282. This was to avoid fusing an Add in a Conv+Add+Relu that has another non-constant input.

https://github.com/microsoft/onnxruntime/blob/6c8cb6a6d1993f84fcf4008f468a071c0b73aad3/onnxruntime/core/optimizer/conv_activation_fusion.cc#L26-L39

However, this check fails to account for implicit inputs and will read past the end of a node's explicit input defs if any implicit inputs are present.

Moreover, this check is no longer necessary after #19470 removed Conv+Add+Relu fusion from ConvActivationFusion.

This change removes the check and some other unused code.

### Motivation and Context

Fix #24473.

### DIFF
--- a/onnxruntime/core/optimizer/conv_activation_fusion.cc
+++ b/onnxruntime/core/optimizer/conv_activation_fusion.cc
@@ -19,24 +19,12 @@ namespace {
 
 #if !defined(ORT_MINIMAL_BUILD)
 namespace selectors {
+
 const Node* GetLoneConsumerNode(const GraphViewer& graph_viewer, const Node& node) {
   if (!optimizer_utils::CheckOutputEdges(graph_viewer.GetGraph(), node, 1)) {
     return nullptr;
   }
-  const Node* next_node = &*node.OutputNodesBegin();
-  // ensure that the target node also has only one input that is not an initializer
-  const size_t input_edges_total = next_node->GetInputEdgesCount();
-  int non_const_edges = 0;
-  for (size_t edge_idx = 0; edge_idx < input_edges_total; ++edge_idx) {
-    if (!graph_utils::NodeArgIsConstant(graph_viewer.GetGraph(), *next_node->InputDefs()[edge_idx])) {
-      ++non_const_edges;
-    }
-  }
-  if (non_const_edges > 1) {
-    return nullptr;
-  } else {
-    return next_node;
-  }
+  return &*node.OutputNodesBegin();
 }
 
 bool HasElementDataType(const NodeArg& node_arg, int32_t data_type) {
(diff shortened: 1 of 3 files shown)


onnxruntime/core/optimizer/conv_activation_fusion.cc
```diff
@@ -19,24 +19,12 @@ namespace {
 
 #if !defined(ORT_MINIMAL_BUILD)
 namespace selectors {
+
 const Node* GetLoneConsumerNode(const GraphViewer& graph_viewer, const Node& node) {
   if (!optimizer_utils::CheckOutputEdges(graph_viewer.GetGraph(), node, 1)) {
     return nullptr;
   }
```
skottmckay, line 26: Not sure this is enough but that would be an issue with CheckOutputEdges.

````
bool CheckOutputEdges(const Graph& graph, const Node& node, size_t expected_output_edges) {
  if (graph.NodeProducesGraphOutput(node)) {
    return false;
  }

  return node.GetOutputEdgesCount() == expected_output_edges;
}
```

If the output edge is to an implicit input (which it was in this case) I think all bets are off and it has to be treated as unknown how many actual edges there are. You can't fuse with something in a subgraph either.  

Most likely that combination is rare if we're looking to fuse, hence this hasn't blown up before. But I think CheckOutputEdges might need to also check that no edges go to implicit inputs and return false if any do.

skottmckay (approved):

PR #24540, opened 2025-04-24
Migrate "Linux CPU Minimal Build E2E CI Pipeline" to Github Actions

snnn:
1. Migrate "Linux CPU Minimal Build E2E CI Pipeline" and "onnxruntime-binary-size-checks-ci-pipeline" to Github Action
2. Add the support for building ONNX Runtime minimal build with vcpkg.
3. Change Github Action's triggers
4. Auto format the yaml files with ruamel.yaml 

Because this change, the pipelines' triggers are set as following:

```
on:
  push:
    branches: [ main, 'rel-*']
  pull_request:
    branches: [ main, 'rel-*']
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true
```

I set "cancel-in-progress: true" because for pipeline runs triggered by pull requests if the pull request was updated(a new commit was added there), the old pipeline runs can be cancelled. However, this setting doesn't work well for the runs triggered by "push" events for the main branch. Let's say, we merged a PR , then it triggered this pipeline. Then before the pipeline is finished, we merged another PR. Then the old pipeline run will be cancelled. But we do want it to be cancelled. Each commit in the main branch should be verified.


snnn:
Very strange. Only my PRs cannot trigger license/cla.

snnn:

PR #24542, opened 2025-04-25
Fix memleakdbg call stack output.

### LINKED ISSUE #24535: ORT Debug Build Fails to Infer for Every Model
### Describe the issue

When building ONNX Runtime in Debug configuration, inference consistently fails for all models tested (e.g., SqueezeNet, AlexNet). The same models run successfully with a Release build. This issue appears to be isolated to the Debug build configuration and is reproducible across different models.

The inference attempt in Debug mode results in the following error (attached below), indicating a possible issue in debug-specific assertions, memory checks, or internal state validations.
Image

### To reproduce

1) git clone --recursive https://github.com/microsoft/onnxruntime
2) cd onnxruntime
3) build.bat --config Debug --build_shared_lib --parallel
4) cd build/Windows/Debug/Debug
5) onnxruntime_perf_test.exe -v -m times -r 1 -I "C:\Users\Administrator\Downloads\squeezenet1.1-7.onnx"

### Urgency

_No response_

### Platform

Windows

### OS Version

Windows 

### ONNX Runtime Installation

Built from Source

main 

### ONNX Runtime API

C++

### Architecture

X64

### Execution Provider

Default CPU

_No response_

edgchen1:
### Description

Fix memleakdbg call stack output.

The call stack output was getting clobbered:
`C:\dev\onnxruntime\build\Debug\_deps\googletest-src\googletest\include\gtest\internal\gtest-port.h(1631): l\gtest-port.h(1631): eadLocal<testing::Sequence *>::GetOrCreateValue`

I think the issue is that this aliasing of `buffer` and `symbol`:
https://github.com/microsoft/onnxruntime/blob/173a11a4e7a2f7a360c9db6abbe601a06a16f004/onnxruntime/core/platform/windows/debug_alloc.cc#L97-L100

does not play nicely with a call to `_snprintf_s` like this:
https://github.com/microsoft/onnxruntime/blob/173a11a4e7a2f7a360c9db6abbe601a06a16f004/onnxruntime/core/platform/windows/debug_alloc.cc#L115

The clobbered output does not match the predefined, ignored patterns, so we see spurious mem leak check output.

This change updates the memleakdbg output generation to use C++ ostreams and instead of fixed size buffers and `_snprintf_s`.

### Motivation and Context

Fix spurious mem leak check output.
Fix #24535.

### DIFF
--- a/onnxruntime/core/platform/windows/debug_alloc.cc
+++ b/onnxruntime/core/platform/windows/debug_alloc.cc
@@ -95,10 +95,10 @@ struct SymbolHelper {
   }
 
   void Lookup(const void* address, std::ostream& message) {
-    alignas(SYMBOL_INFO) char symbol_buffer[kSymbolBufferSize] = {0};
-    SYMBOL_INFO* symbol = reinterpret_cast<SYMBOL_INFO*>(symbol_buffer);
+    SYMBOL_INFO_PACKAGE symbol_info_package{};
+    SYMBOL_INFO* symbol = &symbol_info_package.si;
     symbol->SizeOfStruct = sizeof(SYMBOL_INFO);
-    symbol->MaxNameLen = MAX_SYM_NAME;
+    symbol->MaxNameLen = std::size(symbol_info_package.name);
 
     if (!LookupSymAndInitialize(address, symbol, message)) {
       return;
--- a/onnxruntime/core/platform/windows/debug_alloc.cc
+++ b/onnxruntime/core/platform/windows/debug_alloc.cc
@@ -270,7 +270,7 @@ Memory_LeakCheck::~Memory_LeakCheck() {
     DebugPrint("-----Ending Heap Trace-----\n\n");
 
     std::cout << "\n----- MEMORY LEAKS: " << leaked_bytes << " bytes of memory leaked in "
-              << leak_count << " allocations" << "\n";
+              << leak_count << " allocations\n";
     if (!IsDebuggerPresent()) {
       exit(-1);
     }
--- a/onnxruntime/core/platform/windows/debug_alloc.cc
+++ b/onnxruntime/core/platform/windows/debug_alloc.cc
@@ -75,45 +75,43 @@ struct SymbolHelper {
 
   SymbolHelper() = default;
 
-  static constexpr size_t kInitialBufferSize = sizeof(SYMBOL_INFO) + MAX_SYM_NAME;
+  static constexpr size_t kSymbolBufferSize = sizeof(SYMBOL_INFO) + MAX_SYM_NAME;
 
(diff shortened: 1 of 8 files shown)


onnxruntime/core/platform/windows/debug_alloc.cc
```diff
@@ -271,12 +269,8 @@ Memory_LeakCheck::~Memory_LeakCheck() {
...
     DebugPrint("-----Ending Heap Trace-----\n\n");
 
-    std::string string;
-    char buffer[1024];
-    _snprintf_s(buffer, _TRUNCATE, "%d bytes of memory leaked in %d allocations", static_cast<int>(leaked_bytes), static_cast<int>(leak_count));
-    string.append(buffer);
-
-    std::cout << "\n----- MEMORY LEAKS: " << string.c_str() << "\n";
+    std::cout << "\n----- MEMORY LEAKS: " << leaked_bytes << " bytes of memory leaked in "
+              << leak_count << " allocations" << "\n";
```
edgchen1, line 273:
```suggestion
              << leak_count << " allocations\n";
```

onnxruntime/core/platform/windows/debug_alloc.cc
```diff
@@ -75,45 +75,43 @@ struct SymbolHelper {
...
       }
     }
     return true;
   }
 
-  void Lookup(std::string& string, const ULONG_PTR address) {
-    alignas(SYMBOL_INFO) char buffer[kInitialBufferSize] = {0};
-    SYMBOL_INFO* symbol = reinterpret_cast<SYMBOL_INFO*>(buffer);
+  void Lookup(const void* address, std::ostream& message) {
+    alignas(SYMBOL_INFO) char symbol_buffer[kSymbolBufferSize] = {0};
```
RyanUnderhill, line 98:
```suggestion
    alignas(SYMBOL_INFO) std::byte symbol_buffer[kSymbolBufferSize] = {0};
```
To avoid anyone accidentally writing directly to symbol_buffer as a string again?

edgchen1: I found `SYMBOL_INFO_PACKAGE` in the DbgHelp.h, which is a struct containing `SYMBOL_INFO` and a char buffer following it. using that instead.

yuslepukhin: Is that buffer owned by the `SYMBOL_INFO_PACKAGE`?

edgchen1: yes. here's the definition in DbgHelp.h:

```c++
typedef struct _SYMBOL_INFO_PACKAGE {
    SYMBOL_INFO si;
    CHAR        name[MAX_SYM_NAME + 1];
} SYMBOL_INFO_PACKAGE, *PSYMBOL_INFO_PACKAGE;
```

snnn (approved):

snnn:

PR #24547, opened 2025-04-25
Update Github Actions pipelines' triggers

snnn:
### Description
1. Update Github Actions pipelines' triggers. Make all of them to be the same.
2. Format yaml files.

Before this change, the pipelines' triggers were set as following:

```
on:
  push:
    branches: [ main, 'rel-*']
  pull_request:
    branches: [ main, 'rel-*']
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true
```

I set "cancel-in-progress: true" because for pipeline runs triggered by pull requests if the pull request was updated(a new commit was added there), the old pipeline runs can be cancelled. However, this setting doesn't work well for the runs triggered by "push" events for the main branch. Let's say, we merged a PR , then it triggered this pipeline. Then before the pipeline is finished, we merged another PR. Then the old pipeline run will be cancelled. But we do want it to be cancelled. Each commit in the main branch should be verified.

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -13,8 +13,8 @@ on:
   workflow_dispatch:
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}-${{ github.event_name == 'workflow_dispatch' }}
-  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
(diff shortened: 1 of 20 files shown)


snnn:
Somehow all my PRs stuck in the license/cla check

snnn:

jchen351 (approved):

snnn:
@fs-eire It works!

PR #24552, opened 2025-04-25
Rework --enable_generic_interface and add tests

adrianlizarraga:
### Description
- Introduces `USE_<EP>_PROVIDER_INTERFACE` pre-processor macros that indicate when an EP interface is enabled but the full EP is not being compiled.
  - Previously, the CMake configuration turned on `USE_<EP>` for both use cases. This prevented tests from determining whether the full EP or only the interface was available, which caused test failures. It also turned on all EP code paths in core ORT code at the same time, which caused compilation and logic errors.
- Adds the new NV EP to list of EPs whose interface is enabled with ORT is built with `--enable_generic_interface`
- Updates the Windows Arm64 QNN CI Pipeline to actually use the `--enable_generic_interface` flag.
  - Previously, It was not actually being passed to the build command, so no unit tests were being run with the flag enabled.
- Adds unit tests to check that adding an EP to the session options fails when only the generic interface (but not the full EP) is built.

#### CI Pipelines that use --enable_generic_interface
- Windows ARM64 QNN CI Pipeline:
  - Builds ORT with `--use_qnn --enable_generic_interface` and runs all normal QNN EP unit tests.
  - Builds ORT with `--use_qnn --enable_generic_interface` and runs new unit tests that try to add the following EPs to the session options (expect failure): OpenVINO, CUDA, NV, TensorRT, VitisAI
- Build and Test OpenVINO EP (AlamLinux8, Py3.12) / build_test_pipeline:
  - Builds ORT with `--use_openvino --enable_generic_interface` and runs all normal OpenVINO EP unit tests.
  - Builds ORT with `--use_openvino --enable_generic_interface` and runs new unit tests that try to add the following EPs to the session options (expect failure): QNN, CUDA, NV, TensorRT, VitisAI
- windows_x64_release_ep_generic_interface
  - Builds ORT with `--enable_generic_interface` and now runs CPU EP unit tests (didn't previously).

### Motivation and Context
Fix use of `--enable_generic_interface` and make sure tests actually run.


tools/ci_build/github/azure-pipelines/win-qnn-arm64-ci-pipeline.yml
```diff
@@ -50,13 +50,13 @@ jobs:
...
-        qnn_build_args: '--use_qnn'
+        ExtraQnnBuildArgs: ''
       STATIC_LIB:
         QnnLibKind: 'static_lib'
-        qnn_build_args: '--use_qnn'
+        ExtraQnnBuildArgs: ''
       SHARED_LIB_GENERIC_INTERFACE:
         QnnLibKind: 'shared_lib'
-        qnn_build_args: '--use_qnn --use_generic_interface'
+        ExtraQnnBuildArgs: '--enable_generic_interface'
```
karim-vad, line 59: since you removed --use_qnn, so how do you build EPs ?

adrianlizarraga: `--use_qnn` is always passed explicitly. Please refer to the build commands below:
https://github.com/microsoft/onnxruntime/blob/92f9d4a141d6fabe33c65a9d608e10972b3e21b9/tools/ci_build/github/azure-pipelines/win-qnn-arm64-ci-pipeline.yml#L85-L96

tools/ci_build/build.py
```diff
@@ -443,7 +443,7 @@ def generate_build_tree(
         + ("ON" if args.use_tensorrt_builtin_parser and not args.use_tensorrt_oss_parser else "OFF"),
         # interface variables are used only for building onnxruntime/onnxruntime_shared.dll but not EPs
         "-Donnxruntime_USE_TENSORRT_INTERFACE=" + ("ON" if args.enable_generic_interface else "OFF"),
```
karim-vad, line 445: why is CUDA_INTERFACE removed ?  
Since TENSORRT also requires CUDA so both TENSORRT_INTERFACE and CUDA_INTERFACE must be ON or both need to be OFF ?

onnxruntime/python/onnxruntime_pybind_state_common.cc
```diff
@@ -9,7 +9,7 @@ namespace py = pybind11;
```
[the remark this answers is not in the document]
adrianlizarraga: Yep, the onnxruntime_pybind_state_common.h file has been updated to use the interface macros

cmake/CMakeLists.txt
```diff
@@ -253,7 +253,6 @@ option(onnxruntime_FORCE_GENERIC_ALGORITHMS "Disable optimized arch-specific alg
 
 option(onnxruntime_USE_TENSORRT_INTERFACE "Build ONNXRuntime shared lib which is compatible with TensorRT EP interface" OFF)
 option(onnxruntime_USE_NV_INTERFACE "Build ONNXRuntime shared lib which is compatible with NV EP interface" OFF)
-option(onnxruntime_USE_CUDA_INTERFACE "Build ONNXRuntime shared lib which is compatible with Cuda EP interface" OFF)
```
adrianlizarraga, line 256: Note: I removed CUDA from `--enable_generic_interface`. Doing cuda properly will take a lot more work. I don't think we need CUDA for now anyway.

cmake/CMakeLists.txt
```diff
@@ -1243,7 +1238,13 @@ function(onnxruntime_set_compile_flags target_name)
       target_compile_definitions(${target_name} PRIVATE ${ORT_FLAG})
     endforeach()
 
-    if("${target_name}" STREQUAL "onnxruntime")
+    # These are the targes in which USE_<EP>_PROVIDER_INTERFACE is used.
```
adrianlizarraga, line 1241:
```suggestion
    # These are the targets in which USE_<EP>_PROVIDER_INTERFACE is used.
```

adrianlizarraga, line 1241:
```suggestion
    # These are the targets in which USE_<EP>_PROVIDER_INTERFACE is used.
```

onnxruntime/core/providers/provider_factory_creators.h
```diff
@@ -105,7 +105,3 @@
 #if defined(USE_AZURE)
 #include "core/providers/azure/azure_provider_factory_creator.h"
 #endif
-
-#if defined(USE_NV)
-#include "core/providers/nv_tensorrt_rtx/nv_provider_factory_creator.h"
```
adrianlizarraga, line 110: Note: delete duplicate include

onnxruntime/test/shared_lib/test_session_options.cc
```diff
@@ -34,3 +35,115 @@ TEST(CApiTest, session_options_oversized_affinity_string) {
...
+// Test that loading NV EP when only the interface is built (but not the full EP) fails.
+TEST(CApiTest, session_options_provider_interface_fail_add_tensorrt) {
+  const OrtApi& api = Ort::GetApi();
+  OrtSessionOptions* session_options = nullptr;
+  OrtStatus* status = nullptr;
+
+  status = api.CreateSessionOptions(&session_options);
+  ASSERT_TRUE(status == nullptr);
+
+  // Should not be able to add NV EP to session options because we only built the interface.
```
adrianlizarraga, line 93:
```suggestion
  // Should not be able to add TensorRT EP to session options because we only built the interface.
```

tools/ci_build/build.py
adrianlizarraga: We always have the option of building a base onnxruntime.dll with the options: 
- `--use_cuda --enable_generic_interface`
- Or just `--enable_generic_interface`

tools/ci_build/github/azure-pipelines/win-qnn-arm64-ci-pipeline.yml
adrianlizarraga, line 59:
```suggestion
        ExtraQnnBuildArgs: '--enable_generic_interface --build_wheel'
```

tools/ci_build/github/azure-pipelines/win-qnn-arm64-ci-pipeline.yml
```diff
@@ -93,7 +93,7 @@ jobs:
         --build_shared_lib --use_vcpkg --use_vcpkg_ms_internal_asset_cache 
         --use_qnn $(QnnLibKind)
         --qnn_home $(QnnSDKRootDir)
-        --update --build --parallel
+        --update --build --parallel --build_wheel $(ExtraQnnBuildArgs)
```
adrianlizarraga, line 96:
```suggestion
        --update --build --parallel $(ExtraQnnBuildArgs)
```

tools/ci_build/github/azure-pipelines/win-qnn-arm64-ci-pipeline.yml
```diff
@@ -103,7 +103,7 @@ jobs:
         --build_shared_lib ^
         --use_qnn $(QnnLibKind) ^
         --qnn_home $(QnnSDKRootDir) ^
-        --test --enable_onnx_tests
+        --test --enable_onnx_tests --build_wheel $(ExtraQnnBuildArgs)
```
adrianlizarraga, line 106:
```suggestion
        --test --enable_onnx_tests $(ExtraQnnBuildArgs)
```

tools/ci_build/build.py
karim-vad: In the main build script  build.py 
 if args.use_tensorrt:
        args.use_cuda = True

cmake/CMakeLists.txt
```diff
@@ -1243,7 +1243,13 @@ function(onnxruntime_set_compile_flags target_name)
       target_compile_definitions(${target_name} PRIVATE ${ORT_FLAG})
     endforeach()
 
-    if("${target_name}" STREQUAL "onnxruntime")
+    # These are the targets in which USE_<EP>_PROVIDER_INTERFACE is used.
```
karim-vad, line 1246: In my opinion we should add entire list of all static libs that are currently built into onnxruntime.dll..

Even though some of these libs dont use preprocessor macro, but from the developer perspective, if they add new code, they will not know the nuance of this build logic and will create incorrect bits.

tools/ci_build/build.py
adrianlizarraga: Re-added CUDA

cmake/CMakeLists.txt
adrianlizarraga: Good call. I added all targets

adrianlizarraga: Looks like there's actually no need to specify individual targets. We can just reuse the existing `ORT_PROVIDER_FLAGS` now that we're using different preprocessor macros. This simplifies the CMake file.

adrianlizarraga:
Hi @snnn, Could you please take a look at this PR? I believe it addresses your concerns in this other PR: https://github.com/microsoft/onnxruntime/pull/24368

jywu-msft (approved):

snnn:

PR #24568, opened 2025-04-26
Give the EP all devices so it can create OrtEpDevice instances with full knowledge

skottmckay:
### Description

GetDeviceInfoIfSupported -> GetSupportedDevices

EP sees all devices so it can make decisions with full knowledge. This is mainly applicable to GPU EPs like WebGPU.

EP has to iterate device and call CreateEpDevice for devices it supports.


onnxruntime/core/session/environment.cc
```diff
@@ -482,36 +507,25 @@ Status Environment::EpInfo::Create(std::unique_ptr<EpLibrary> library_in, std::u
   ORT_RETURN_IF_ERROR(instance.library->Load());
   const auto& factories = instance.library->GetFactories();
 
+  // OrtHardwareDevice instances to pass to GetSupportedDevices. sorted by type to be slightly more structured.
+  // the set of hardware devices is static so this can also be static.
+  const static auto sorted_devices = SortDevicesByType();
```
adrianlizarraga, line 512:
```suggestion
  const static std::vector<const OrtHardwareDevice*> sorted_devices = SortDevicesByType();
```
nit: Make it easier for future reader to determine which kind of "device" type this is (since there are so many) :)

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -5889,6 +5898,27 @@ struct OrtCompileApi {
...
+   *                        ep_device will take ownership of this instance.
+   * \param[in] ep_options  Optional OrtKeyValuePairs instance for execution provider options that will be added
+   *                        to the Session configuration options if the execution provider is selected.
+   *                        ep_device will take ownership of this instance.
+   * \param ep_device OrtExecutionDevice that is created.
+   */
+  ORT_API2_STATUS(CreateEpDevice, _In_ OrtEpFactory* ep_factory,
+                  _In_ const OrtHardwareDevice* hardware_device,
+                  _In_ OrtKeyValuePairs* ep_metadata,
+                  _In_ OrtKeyValuePairs* ep_options,
```
adrianlizarraga, line 5916: I see that these are moved (via std::move) in the implementation. AFAIK, this is not common practice in a C API (as std::move is a C++ construct) and can lead to difficult to debug errors because the object provided by the user has been gutted by `std::move`. I would lean towards passing in const pointers and allowing the implementation to make a copy.

But, if we still want to use std::move, then I think we need to do the following:
- Make these `_InOut_` instead of `_In_`.
- Expand documenetation commet to state that the OrtKeyValuePairs they passed in are now empty (due to the move) and can no longer be used by the caller.

adrianlizarraga (approved):

snnn:

PR #24585, opened 2025-04-29
Fix DML autoep select test. 

skottmckay:
### Description

Fix DML autoep select test. It should only select one device as that's all the test infrastructure is setup to handle.


adrianlizarraga (approved):

snnn:

PR #24587, opened 2025-04-29
'LoadPluginOrProviderBridge' should load the library_path as absolute

mschofie:
### Description
`LoadPluginOrProviderBridge` is called when attempting to load a Plugin. It uses the passed `library_path` to attempt to load the Plugin as a `Provider` - using `ProviderLibrary` - to see if it can be treated as a 'ProviderBridge'. `ProviderLibrary` attempts to load the Provider by prefixing the path to the onnxruntime.dll. Plugins needn't be redistributed with OnnxRuntime, so the path to the Plugin _may_ be an absolute path, and if so `ProviderLibrary` fails. At the same time - however - `LoadPluginOrProviderBridge` needs to support OnnxRuntime-relative paths: As 'Providers' are migrated to 'Plugins', existing Providers should be usable as Plugins. To accommodate both scenarios, this PR:

1. Adds support to `ProviderLibrary` to be created with an absolute path.
2. Validates the path passed to `LoadPluginOrProviderBridge`;
    1. if it is absolute, the same absolute path is passed to `ProviderLibrary` and `EpLibraryPlugin`.
    2. if the path is not absolute, it is converted to an absolute path by prefixing the OnnxRuntime location, and the same path is passed to `ProviderLibrary` and `EpLibraryPlugin`.

### Motivation and Context
This PR enables `LoadPluginOrProviderBridge` to be called with an absolute path to the Plugin, allowing it to be used as a 'ProviderBridge', or with an OnnxRuntime-relative path to the Plugin.

### DIFF
--- a/onnxruntime/core/session/ep_library_plugin.h
+++ b/onnxruntime/core/session/ep_library_plugin.h
@@ -16,9 +16,9 @@ namespace onnxruntime {
 /// </summary>
 class EpLibraryPlugin : public EpLibrary {
  public:
-  EpLibraryPlugin(const std::string& registration_name, const ORTCHAR_T* library_path)
+  EpLibraryPlugin(const std::string& registration_name, std::filesystem::path library_path)
       : registration_name_{registration_name},
-        library_path_{library_path} {
+        library_path_{std::move(library_path)} {
   }
 
   const char* RegistrationName() const override {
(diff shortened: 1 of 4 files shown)


skottmckay (approved):

snnn:

PR #24594, opened 2025-04-29
[CoreML] Update Conv and Softmax ops

carzh:
### Description
Moved the dimension limit because it seems to only apply to conv operations (texture memory is typically used for conv operations in the GPU because it has a slow write but fast read -- ChromaDB model had a slice operation with an input > 16384 -- operation worked fine after I had moved the dim check)

Also added extra checks for Softmax on MLProgram that allows more softmax nodes to be moved to CoreML


onnxruntime/core/providers/coreml/builders/impl/reshape_op_builder.cc
```diff
@@ -89,37 +118,58 @@ bool ReshapeOpBuilder::IsOpSupportedImpl(const Node& node, const OpBuilderInputP
...
-  if (!GetStaticShape(*input_defs[0], input_shape, logger))
+  if (!input_params.create_mlprogram) {
+    // if we are using NeuralNetwork, input shape must be static
+    if (!GetStaticShape(*input_defs[0], input_shape, logger)) {
+      LOGS(logger, VERBOSE) << "Unable to get shape of input -- input must have static shape for NeuralNetwork.";
+      return false;
+    }
+  }
+
+  // first input must be fixed rank OR (first input has variadic rank AND shape only contains positive integers)
```
edgchen1, line 145: "OR (first input has variadic rank AND shape only contains positive integers)"

if I understand correctly, we don't handle this, right? might be worth explicitly saying so.

onnxruntime/core/providers/coreml/builders/impl/reshape_op_builder.cc
```diff
@@ -89,37 +118,58 @@ bool ReshapeOpBuilder::IsOpSupportedImpl(const Node& node, const OpBuilderInputP
...
+  // as per docs, 0 is considered an illegal shape element if the input is variadic
+  if (!GetShape(*input_defs[0], input_shape, logger)) {
```
edgchen1, line 147: instead of calling `GetStaticShape` and then `GetShape` and populating `input_shape` twice, could call `GetShape` and then `IsStaticShape` on the result.

onnxruntime/core/providers/coreml/builders/impl/reshape_op_builder.cc
```diff
@@ -89,37 +118,58 @@ bool ReshapeOpBuilder::IsOpSupportedImpl(const Node& node, const OpBuilderInputP
...
+    LOGS(logger, VERBOSE) << "Unable to get shape of input -- input must have fixed rank for reshape.";
     return false;
+  }
 
-  if (input_shape.empty()) {
-    LOGS(logger, VERBOSE) << "Reshape does not support empty input shape";
+  if (input_shape.empty() && !AllPositiveShape(new_shape)) {
+    // unknown rank & fails the positive shape check
```
edgchen1, line 153: does an empty shape mean unknown rank, or that we are dealing with a scalar?

onnxruntime/core/providers/coreml/builders/impl/base_op_builder.h
```diff
@@ -36,6 +36,11 @@ class BaseOpBuilder : public IOpBuilder {
   static bool IsInputDtypeSupport(const Node& node, size_t idx, const OpBuilderInputParams& input_params,
                                   const logging::Logger& logger);
 
+  // See this issue, https://github.com/apple/coremltools/issues/1003
+  // https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf has maximum texture widths which may be the
+  // root cause.
+  static bool CheckShapeForLimit(onnxruntime::VectorInt64& shape);
```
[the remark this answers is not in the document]
carzh: Updated to CheckShapeForConvMemoryLimit -- but open to other more descriptive name ideas :0

onnxruntime/core/providers/coreml/builders/impl/conv_op_builder.cc
```diff
@@ -237,6 +237,16 @@ bool ConvOpBuilder::IsOpSupportedImpl(const Node& node, const OpBuilderInputPara
   const auto* weight_shape = input_defs[1]->Shape();
   int64_t num_dims = weight_shape ? weight_shape->dim_size() : -1;
 
+  std::vector<int64_t> weight_shape_vec;
+  std::vector<int64_t> x_shape_vec;
+  GetShape(*input_defs[1], weight_shape_vec, logger);
+  GetShape(*input_defs[0], x_shape_vec, logger);
+
+  if (!CheckShapeForLimit(weight_shape_vec) || !CheckShapeForLimit(x_shape_vec)) {
+    LOGS(logger, VERBOSE) << "Conv [" << name << "] has a shape with dimension > 16384. CoreML does not support conv operations with dim > 16384.";
```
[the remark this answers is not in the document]
carzh: Yep moved logging o CheckShapeForLimit

onnxruntime/core/providers/coreml/builders/impl/reshape_op_builder.cc
```diff
@@ -75,11 +75,40 @@ Status ReshapeOpBuilder::AddToModelBuilderImpl(ModelBuilder& model_builder,
   return Status::OK();
 }
 
+bool AllPositiveShape(gsl::span<const int64_t> shape) {
+  return std::all_of(shape.begin(), shape.end(), [](int64_t dim) { return dim > 0 || dim == 0; });
```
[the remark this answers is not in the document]
carzh: Oops, I think that was a mistake left in from earlier debugging -- fixed

onnxruntime/core/providers/coreml/builders/impl/reshape_op_builder.cc
```diff
@@ -75,11 +75,40 @@ Status ReshapeOpBuilder::AddToModelBuilderImpl(ModelBuilder& model_builder,
...
+
+  // Case 1: If new_shape has no -1 dimensions, it's always valid
+  if (negative_one_count == 0) {
+    return true;
+  }
+
+  // Case 2: If new_shape has exactly one -1 dimension, check if input_shape has at least one -1
+  if (negative_one_count == 1) {
+    return std::any_of(input_shape.begin(), input_shape.end(),
+                       [](int64_t dim) { return dim == -1; });
```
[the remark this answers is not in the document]
carzh: I think my previous implementation was confusing and in debugging I wrote a lot of helper methods to check for weird edge cases -- updated with version that checks for the number of new symbols introduced in new_shape and also checks the number of -1's in new_shape

onnxruntime/core/providers/coreml/builders/impl/base_op_builder.h
```diff
@@ -36,6 +36,11 @@ class BaseOpBuilder : public IOpBuilder {
   static bool IsInputDtypeSupport(const Node& node, size_t idx, const OpBuilderInputParams& input_params,
                                   const logging::Logger& logger);
 
+  // See this issue, https://github.com/apple/coremltools/issues/1003
+  // https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf has maximum texture widths which may be the
+  // root cause.
+  static bool CheckShapeForConvMemoryLimit(onnxruntime::VectorInt64& shape, const logging::Logger& logger);
```
edgchen1, line 42: nit: it feels out of place in base_op_builder.h as it is not relevant to most ops. maybe move it to a separate helper function in helper.h/cc?

onnxruntime/core/providers/coreml/builders/impl/base_op_builder.cc
```diff
@@ -126,6 +127,24 @@ bool BaseOpBuilder::IsInputDtypeSupport(const Node& node, size_t idx,
   return false;
 }
 
+/* static */
+bool BaseOpBuilder::CheckShapeForConvMemoryLimit(onnxruntime::VectorInt64& shape, const logging::Logger& logger) {
```
edgchen1, line 131: for `shape`, we don't need to pass the vector by reference as we don't modify it. consider using `gsl::span<const int64_t>` as it allows more flexibility for the input.

onnxruntime/core/providers/coreml/builders/impl/base_op_builder.h
carzh: Ah, fair point -- moved to the helper.h/cc files!

onnxruntime/core/providers/coreml/builders/impl/conv_op_builder.cc
```diff
@@ -237,6 +237,16 @@ bool ConvOpBuilder::IsOpSupportedImpl(const Node& node, const OpBuilderInputPara
   const auto* weight_shape = input_defs[1]->Shape();
   int64_t num_dims = weight_shape ? weight_shape->dim_size() : -1;
 
+  std::vector<int64_t> weight_shape_vec;
+  std::vector<int64_t> x_shape_vec;
+  GetShape(*input_defs[1], weight_shape_vec, logger);
```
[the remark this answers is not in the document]
carzh: Added check for return values for both!

onnxruntime/core/providers/coreml/builders/impl/reshape_op_builder.cc
```diff
@@ -89,39 +118,64 @@ bool ReshapeOpBuilder::IsOpSupportedImpl(const Node& node, const OpBuilderInputP
...
-  if (!GetStaticShape(*input_defs[0], input_shape, logger))
-    return false;
 
-  if (input_shape.empty()) {
-    LOGS(logger, VERBOSE) << "Reshape does not support empty input shape";
+  // first input must be fixed rank OR (first input has variadic rank AND shape only contains positive integers)
+  // as per docs, 0 is considered an illegal shape element if the input is variadic
+  if (!GetShape(*input_defs[0], input_shape, logger)) {
+    if (!AllPositiveShape(new_shape)) {
+      LOGS(logger, VERBOSE) << "Unable to get shape of input -- input must have fixed rank for reshape, unless the new shape contains all positive integers. "
```
skottmckay: nit: line length

onnxruntime/core/providers/coreml/builders/impl/reshape_op_builder.cc
```diff
@@ -89,39 +118,64 @@ bool ReshapeOpBuilder::IsOpSupportedImpl(const Node& node, const OpBuilderInputP
...
   auto new_shape = unpacked_tensor.DataAsSpan<int64_t>();
+
   if (new_shape.empty()) {
     LOGS(logger, VERBOSE) << "New shape of reshape cannot be empty";
     return false;
   }
 
+  // CoreML reshape does not support 0 as a dimension
+  // Even though MLProgram Reshape docs imply 0 is supported as a dimension, it is not supported
+  if (allow_zero) {
```
skottmckay, line 129: nit: isn't this slightly different to the supposedly supported usage of zero (copy value from input)? If allow_zero is set to 1, the ONNX behavior is to copy an explicit zero (which equates to empty output). CoreML doesn't support that at all afaik.

onnxruntime/core/providers/coreml/builders/impl/reshape_op_builder.cc
```diff
@@ -89,39 +118,64 @@ bool ReshapeOpBuilder::IsOpSupportedImpl(const Node& node, const OpBuilderInputP
...
+  // as per docs, 0 is considered an illegal shape element if the input is variadic
+  if (!GetShape(*input_defs[0], input_shape, logger)) {
+    if (!AllPositiveShape(new_shape)) {
+      LOGS(logger, VERBOSE) << "Unable to get shape of input -- input must have fixed rank for reshape, unless the new shape contains all positive integers. "
+                               "Input shape: "
+                            << Shape2String(input_shape) << ", new shape: " << Shape2String(new_shape);
+      return false;
+    } else {
+      return true;
+    }
```
skottmckay: We return true if the input shape is unknown and AllPositiveShape is true, but if the input shape is known we have additional checks like `new_shape.size() > 5`. should we be falling through to at least some of the later checks instead of returning true here?

onnxruntime/core/providers/coreml/builders/impl/reshape_op_builder.cc
```diff
@@ -89,39 +118,64 @@ bool ReshapeOpBuilder::IsOpSupportedImpl(const Node& node, const OpBuilderInputP
...
+  if (!CheckNegativeOneCollision(input_shape, new_shape, logger)) {
+    return false;
   }
 
+  // Max of one -1 allowed
+  if (input_params.create_mlprogram && std::count(new_shape.begin(), new_shape.end(), -1) > 1) {
+    LOGS(logger, VERBOSE) << "Reshape does not support -1 in new shape unless it appears in the input shape. "
+                             "Input shape: "
+                          << Shape2String(input_shape) << ", new shape: " << Shape2String(new_shape);
+    return false;
```
skottmckay, line 173: nit: the onnx type/shape inferencing should have enforced this as we know the `shape` input is a const initializer.

https://github.com/onnx/onnx/blob/a570866ba4641506d5530b9b559e169ded3edfe1/onnx/defs/tensor/defs.cc#L289-L291

onnxruntime/core/providers/coreml/builders/impl/reshape_op_builder.cc
carzh: Ah this is a good point but ended up removing support for this case because there's no good way to write unit tests for it at the moment and only affects 1 node in the ChromaDB model -- will keep this in mind for future implementation though

onnxruntime/core/providers/coreml/builders/impl/softmax_op_builder.cc
```diff
@@ -30,12 +30,15 @@ Status SoftmaxOpBuilder::AddToModelBuilderImpl(ModelBuilder& model_builder,
...
   std::vector<int64_t> data_shape;
-  ORT_RETURN_IF_NOT(GetStaticShape(*node.InputDefs()[0], data_shape, logger), "Failed to get input shape.");
 
   NodeAttrHelper helper(node);
   int32_t axis_default_value = (node.SinceVersion() < 13) ? 1 : -1;
   const auto axis = helper.Get("axis", axis_default_value);
-  auto axis_nonnegative = HandleNegativeAxis(axis, data_shape.size());
+  int64_t axis_nonnegative = axis;
+  if (node.SinceVersion() < 13) {
+    ORT_RETURN_IF_NOT(GetStaticShape(*node.InputDefs()[0], data_shape, logger), "Failed to get input shape.");
```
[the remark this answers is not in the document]
carzh: Ah, it's a little unclear to see, but data_shape is only used in the codeflows / branches where node.SinceVersion() < 13 -- the data_shape is required for the cases where some kind of reshape is needed.

onnxruntime/core/providers/coreml/builders/impl/softmax_op_builder.cc
```diff
@@ -125,13 +128,14 @@ Status SoftmaxOpBuilder::AddToModelBuilderImpl(ModelBuilder& model_builder,
...
   const auto& input_defs = node.InputDefs();
   std::vector<int64_t> input_shape;
-  if (!GetStaticShape(*input_defs[0], input_shape, logger))
+  if (!GetStaticShape(*input_defs[0], input_shape, logger) && (!input_params.create_mlprogram || node.SinceVersion() < 13)) {
+    LOGS(logger, VERBOSE) << "Softmax input must have static shape for NeuralNetwork or ONNX opset < 13";
     return false;
-
+  }
   const TensorShape shape(input_shape);
   if (shape.Size() == 0) {
```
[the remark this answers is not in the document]
carzh: Oops, yes, changed to GetShape then check for IsStaticShape after

onnxruntime/core/providers/coreml/builders/impl/conv_op_builder.cc
```diff
@@ -237,6 +237,23 @@ bool ConvOpBuilder::IsOpSupportedImpl(const Node& node, const OpBuilderInputPara
...
+    LOGS(logger, VERBOSE) << "Unable to get the shape of first input, which is necessary to check for valid convolutions.";
+    return false;
+  }
+
+  if (!GetShape(*input_defs[0], x_shape_vec, logger)) {
+    LOGS(logger, VERBOSE) << "Unable to get the shape of second input, which is necessary to check for valid convolutions.";
+    return false;
+  }
+
+  if (!CheckShapeForConvMemoryLimit(weight_shape_vec, logger) || !CheckShapeForConvMemoryLimit(x_shape_vec, logger)) {
```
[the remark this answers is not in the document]
carzh: added check for output shape!

onnxruntime/core/providers/coreml/builders/impl/softmax_op_builder.cc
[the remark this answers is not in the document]
edgchen1: it's a little unintuitive that `data_shape` is only valid for certain cases now. it seems like one more thing that people modifying this code later will need to watch out for. could we do something like this instead, so that `data_shape` will be consistently valid?
```c++
  std::vector<int64_t> data_shape;
  ORT_RETURN_IF_NOT(GetShape(*node.InputDefs()[0], data_shape, logger), "Failed to get input shape.");
  ...
  if (node.SinceVersion() < 13) {
    ORT_RETURN_IF_NOT(IsStaticShape(data_shape), "Input shape is not static.");
```

onnxruntime/core/providers/coreml/builders/impl/conv_op_builder.cc
```diff
@@ -236,6 +236,30 @@ bool ConvOpBuilder::IsOpSupportedImpl(const Node& node, const OpBuilderInputPara
...
+    LOGS(logger, VERBOSE) << "Unable to get the shape of 'X' input, which is necessary to check for valid convolutions.";
+    return false;
+  }
+
+  if (!GetShape(output, output_shape_vec, logger)) {
+    LOGS(logger, VERBOSE) << "Unable to get the shape of the output, which is necessary to check for valid convolutions.";
+    return false;
+  }
+
+  if (!CheckShapeForConvMemoryLimit(weight_shape_vec, logger) || !CheckShapeForConvMemoryLimit(x_shape_vec, logger) || !CheckShapeForConvMemoryLimit(output_shape_vec, logger)) {
```
edgchen1: nit: line length

onnxruntime/core/providers/coreml/builders/impl/softmax_op_builder.cc
```diff
@@ -125,16 +128,18 @@ Status SoftmaxOpBuilder::AddToModelBuilderImpl(ModelBuilder& model_builder,
...
+
+  if (!GetShape(*input_defs[0], input_shape, logger)) {
+    LOGS(logger, VERBOSE) << "Softmax input [" << input_defs[0]->Name() << "] must have shape information.";
     return false;
+  }
 
-  const TensorShape shape(input_shape);
-  if (shape.Size() == 0) {
-    LOGS(logger, VERBOSE) << "Empty input data is not supported.";
+  if (!IsStaticShape(input_shape) && (!input_params.create_mlprogram || node.SinceVersion() < 13)) {
```
edgchen1: do we need `!input_params.create_mlprogram` here? `SoftmaxOpBuilder::AddToModelBuilderImpl()` only checks `node.SinceVersion() < 13` before calling `GetStaticShape()`.

onnxruntime/core/providers/coreml/builders/impl/reshape_op_builder.cc
```diff
@@ -75,11 +75,43 @@ Status ReshapeOpBuilder::AddToModelBuilderImpl(ModelBuilder& model_builder,
   return Status::OK();
 }
 
+bool AllPositiveShape(gsl::span<const int64_t> shape) {
```
edgchen1: is this no longer used?

onnxruntime/core/providers/coreml/builders/impl/reshape_op_builder.cc
```diff
@@ -40,7 +40,7 @@ Status ReshapeOpBuilder::AddToModelBuilderImpl(ModelBuilder& model_builder,
                                                const logging::Logger& logger) const {
   const auto& input_defs = node.InputDefs();
   std::vector<int64_t> input_shape;
-  ORT_RETURN_IF_NOT(GetStaticShape(*input_defs[0], input_shape, logger), "Cannot get shape of data");
+  ORT_RETURN_IF_NOT(GetShape(*input_defs[0], input_shape, logger), "Cannot get shape of data");
```
edgchen1: hm, it looks like ReshapeHelper does not like an input shape with dynamic dimensions:
https://github.com/microsoft/onnxruntime/blob/30415e6a3660a4bcbcc9d39ba317686faa628ed4/onnxruntime/core/providers/cpu/tensor/reshape_helper.h#L15-L16

so even if GetShape were to succeed here, if input_shape has a -1 dimension, ReshapeHelper will complain.

onnxruntime/core/providers/coreml/builders/impl/softmax_op_builder.cc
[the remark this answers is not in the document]
carzh: Oh, yes, that's a good idea -- I added GetShape for the else branch

onnxruntime/core/providers/coreml/builders/impl/reshape_op_builder.cc
```diff
@@ -75,11 +75,39 @@ Status ReshapeOpBuilder::AddToModelBuilderImpl(ModelBuilder& model_builder,
   return Status::OK();
 }
 
+// CoreML does not handle 0's in the new_shape, even though its documentation claims 0's in the new_shape will
+// match the corresponding dimension in x.shape
+// https://apple.github.io/coremltools/source/coremltools.converters.mil.mil.ops.defs.html#coremltools.converters.mil.mil.ops.defs.iOS15.tensor_transformation.reshape
+//
+// So instead, we use ReshapeHelper to populate the values from input_shape. However, input_shape may have -1's in it
```
edgchen1: "input_shape may have -1's in it"
it won't now if it must be a static shape, right?

onnxruntime/test/providers/cpu/tensor/tensor_op_test.cc
```diff
@@ -91,6 +91,39 @@ TEST(TensorOpTest, Reshape_WithOutAllowZero) {
...
 }
 
+TEST(TensorOpTest, Reshape_WithOutAllowZeroToDiffRank) {
+  OpTester test("Reshape", 14);
+
+  test.AddInput<float>("data", {2, 3, 12}, std::vector<float>(72, 1.0f));
+  test.AddInput<int64_t>("shape", {4}, {2, 3, 3, 4}, true);
+  test.AddAttribute<int64_t>("allowzero", 0);
+  test.AddOutput<float>("reshaped", {2, 3, 3, 4}, std::vector<float>(72, 1.0f));
+  // TensorRT doesn't support dynamic shape tensor for now
```
edgchen1: what does "dynamic shape tensor" mean here? the shape input is an initializer.

carzh: Oops -- the test was copied / templated from previous tests

onnxruntime/core/providers/coreml/builders/impl/softmax_op_builder.cc
```diff
@@ -125,16 +132,18 @@ Status SoftmaxOpBuilder::AddToModelBuilderImpl(ModelBuilder& model_builder,
   return Status::OK();
 }
 
-bool SoftmaxOpBuilder::IsOpSupportedImpl(const Node& node, const OpBuilderInputParams& /*input_params*/,
+bool SoftmaxOpBuilder::IsOpSupportedImpl(const Node& node, const OpBuilderInputParams& input_params,
```
tianleiwu, line 135: input_params is not used. There is build warnings.

PR #24595, opened 2025-04-29
Add Paged Attention Op for CUDA SM80 support

aciddelgado:
### Description
Adds Paged Attention Op which enables of Paged KV Cache. Inputs to this op are unpadded (packed / varlen) so Cumulative Sequence Lengths are a required input.

### Motivation and Context
Adding this op to ONNXRuntime is necessary to allow the GenAI team to enable a continuous batching server API.


onnxruntime/contrib_ops/cpu/bert/attention_parameters.h
```diff
@@ -102,6 +102,21 @@ struct GroupQueryAttentionParameters : AttentionParameters {
...
 
+// Parameters deduced from node attributes and inputs/outputs.
+struct PagedAttentionParameters : AttentionParameters {
+  int token_count;              // number of tokens in packed query
+  int kv_num_heads;
+  int kv_hidden_size;
+  int block_size;               // block size for kv cache
+  int max_num_blocks_per_seq;   // max number of blocks per sequence for kv cache
+  int num_blocks;               // number of blocks in kv cache
+  int rotary_dim;               // rotary embedding dimension
```
kunal-vaishnavi, line 113: Is this the same as `rotary_embedding` in `AttentionParameters`?

onnxruntime/contrib_ops/cpu/bert/attention_parameters.h
```diff
@@ -102,6 +102,21 @@ struct GroupQueryAttentionParameters : AttentionParameters {
   int* zero_ptr;
 };
 
+// Parameters deduced from node attributes and inputs/outputs.
+struct PagedAttentionParameters : AttentionParameters {
+  int token_count;              // number of tokens in packed query
+  int kv_num_heads;
```
kunal-vaishnavi, line 108: Given that some of these have become standard attention parameters now (`kv_num_heads`, `kv_hidden_size`, `is_packed_qkv`), can they be moved into `AttentionParameters`?

onnxruntime/contrib_ops/cuda/bert/paged_attention.cc
```diff
@@ -0,0 +1,212 @@
...
+  size_t workspace_buffer_bytes = 0;
+  if (parameters.is_packed_qkv) { // unpacking and rotary can be done with the same buffer in the same operation
+    workspace_buffer_bytes = parameters.token_count * (parameters.num_heads + 2 * parameters.kv_num_heads) * parameters.head_size * sizeof(T);
+  } else if (do_rotary_) {
+    workspace_buffer_bytes = 2 * sizeof(T) * parameters.token_count * parameters.num_heads * parameters.head_size;
+  }
+  auto workspace_buffer = GetScratchBuffer<void>(workspace_buffer_bytes, context->GetComputeStream());
+
+  // Print debug info
+  if (kernel_options_->AllowDebugInfo()) {
```
kunal-vaishnavi, line 170: For when debug mode is enabled, the parameters can also be dumped as follows.

https://github.com/microsoft/onnxruntime/blob/838b97e73289cd11caf969f9f5c01ce153d6069f/onnxruntime/contrib_ops/cuda/bert/multihead_attention.cc#L124-L135

Because of how the `DUMP_STRING` and `DUMP_STRING_INIT` methods are defined below, they don't require any checks for when debug mode is on or off.

https://github.com/microsoft/onnxruntime/blob/838b97e73289cd11caf969f9f5c01ce153d6069f/onnxruntime/contrib_ops/cpu/utils/debug_macros.h#L33-L44

onnxruntime/contrib_ops/cuda/bert/paged_attention_helper.h
```diff
@@ -0,0 +1,302 @@
...
+                   float softcap,
+                   int max_threads_per_block) {
+  if (max_threads_per_block > 0 && num_heads > max_threads_per_block) {
+    return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT, "num_heads should be no larger than ", max_threads_per_block);
+  }
+
+  return CheckInputs(query, key, value, key_cache, value_cache, cumulative_sequence_length, seqlens, max_query_len, max_seq_len, block_table, slot_mappings, cos_cache, sin_cache, parameters, num_heads, kv_num_heads, scale, softcap);
+}
+
+}  // namespace group_query_attention_helper
```
kunal-vaishnavi, line 300:
```suggestion
}  // namespace paged_attention_helper
```

onnxruntime/contrib_ops/cuda/bert/paged_attention_helper.h
```diff
@@ -0,0 +1,302 @@
...
+#include "core/providers/common.h"
+#include "contrib_ops/cpu/bert/attention_common.h"
+#include "contrib_ops/cpu/bert/attention_parameters.h"
+
+namespace onnxruntime {
+namespace contrib {
+namespace paged_attention_helper {
+
+template <typename T = Tensor>
+Status CheckInputs(const T* query,
```
kunal-vaishnavi, line 16: Can we refactor some of the helper checks done in [`group_query_attention_helper.h`](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/contrib_ops/cpu/bert/group_query_attention_helper.h) to be in separate methods similar to how the helper checks in [`multihead_attention_helper.h`](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/contrib_ops/cpu/bert/multihead_attention_helper.h) are done? Then we can re-use some of the methods in `group_query_attention_helper.h` here (e.g. the rotary embedding cache checks) while defining any new checks in this file as needed (e.g. the new checks for the KV caches).

onnxruntime/contrib_ops/cuda/bert/paged_attention_impl.cu
```diff
@@ -0,0 +1,385 @@
...
+template <typename T>
+__global__ void ReshapeAndCache(const T* __restrict__ key, const T* __restrict__ value, T* __restrict__ key_cache,
+                                T* __restrict__ value_cache, const int* __restrict__ slot_mappings,
+                                const int token_count, const int kv_hidden_size, const int block_size) {
+  const int token_id = blockIdx.x;
+  const int slot_id = slot_mappings[token_id];
+  const int block_id = slot_id / block_size;
+  const int block_offset = slot_id % block_size;
+
+  for (int i = threadIdx.x; i < kv_hidden_size; i += blockDim.x) {
```
kunal-vaishnavi, line 231: Can this be rewritten to avoid needing the for loop?

kunal-vaishnavi, line 231: Can this be rewritten to avoid needing the for-loop?

onnxruntime/core/graph/contrib_ops/bert_defs.cc
```diff
@@ -1206,6 +1206,215 @@ ONNX_MS_OPERATOR_SET_SCHEMA(
...
+      fail_shape_inference("The block-based KV cache inputs shall be 4 dimensions");
+    }
+    // KV cache in and out share the same buffer, thus they have the same shape
+    ONNX_NAMESPACE::propagateShapeFromInputToOutput(ctx, 3, 1);
+    ONNX_NAMESPACE::propagateShapeFromInputToOutput(ctx, 4, 2);
+  }
+}
+
+ONNX_MS_OPERATOR_SET_SCHEMA(
+    PagedAttention, 1,
```
kunal-vaishnavi, line 1307: Should we name the input KV caches as `past_key` and `past_value` + the output KV caches as `present_key` and `present_value` to match the names used in the other attention ops?

aciddelgado: Since the kv cache inputs are always updated in place in this op, and the kv cache outputs are optional, I think it would be innacurate to name them like that

onnxruntime/contrib_ops/cpu/bert/attention_parameters.h
aciddelgado: I can only find [one instance](https://github.com/microsoft/onnxruntime/blob/838b97e73289cd11caf969f9f5c01ce153d6069f/onnxruntime/contrib_ops/cuda/bert/attention_prepare_qkv.cu#L153) of `rotary_embedding` being used and I didn't know it existed haha. All other attention op parameters use rotary_dim or rotary_embedding_dim so perhaps it's best to collapse these into base attention params.

onnxruntime/contrib_ops/cuda/bert/paged_attention_helper.h
aciddelgado: The rotary embedding cache check would be the only reusable one but i'll refactor it.

onnxruntime/contrib_ops/cuda/bert/attention_data.h
```diff
@@ -180,6 +180,35 @@ struct GroupQueryAttentionData {
...
 
+template <typename T>
+struct PagedAttentionData {
+  // Input Tensors
+  const T* query = nullptr;
+  const T* key = nullptr;
+  const T* value = nullptr;
+  T* key_cache = nullptr;
+  T* value_cache = nullptr;
+  const int* cumulative_sequence_length = nullptr;
```
tianleiwu, line 191: what is different of cumulative_sequence_length and cumulative_seqlens_kv? It is better to distinguish them from naming like cumulative_seqlens_q and cumulative_seqlens_kv.

onnxruntime/contrib_ops/cuda/bert/flash_attention/flash_api.cc
```diff
@@ -149,6 +149,11 @@ size_t get_softmax_lse_size(size_t seqlen, size_t batch_size, size_t num_heads)
   return bytes;
 }
 
+size_t get_softmax_lse_size(size_t token_count, size_t num_heads) {
+  size_t bytes = sizeof(float) * token_count * num_heads;
```
tianleiwu, line 153: Does it need num_splits like the below function?

onnxruntime/contrib_ops/cuda/bert/paged_attention_helper.h
```diff
@@ -0,0 +1,310 @@
...
+  }
+  if (value_cache_dims[3] != head_size) {
+    return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT,
+                           "Input 'past_value' dimension 3 should be same as head_size, got ",
+                           value_cache_dims[3]);
+  }
+  return Status::OK();
+}
+
+template <typename T>
```
tianleiwu, line 159: nit: no need template T here, use int directly?

onnxruntime/core/graph/contrib_ops/bert_defs.cc
```diff
@@ -1206,6 +1206,215 @@ ONNX_MS_OPERATOR_SET_SCHEMA(
...
+        .Input(7,
+               "max_query_len",
+               "Scalar tensor equivalent to the maximum length of all sequences in the query tensor (max new seqlen). "
+               "This input is allocated on CPU.",
+               "M")
+        .Input(8,
+               "max_seq_len",
+               "Scalar tensor equivalent to the maximum length of all sequences in the KV Cache after appending new "
+               "tokens (max total seqlen). This input is allocated on CPU.",
+               "M")
```
tianleiwu, line 1375: The interface is designed for the convenience of current implementation, that might not be good for other EP (like WebGPU, CPU etc).

A generic design is to have q_seq_lens and kv_seq_lens, then you can deduce these four inputs. For example, for CPU, it is very easy to compute max seq length, so there is no need to have max_query_len and max_seq_len inputs.

I guess that the extra latency caused by such interface change might be tiny.

onnxruntime/core/graph/contrib_ops/bert_defs.cc
```diff
@@ -1206,6 +1206,215 @@ ONNX_MS_OPERATOR_SET_SCHEMA(
...
+        .Input(9,
+               "block_table",
+               "2D tensor with shape (batch_size, max_blocks_per_sequence) that maps each sequence in the batch to its"
+               "corresponding blocks in the KV cache.",
+               "M")
+        .Input(10,
```
tianleiwu, line 1381: Since we have sequence lengths, we shall be able to map each token in KV cahce dynamically.
Why do we need slot_mappings?

onnxruntime/core/graph/contrib_ops/bert_defs.cc
```diff
@@ -1206,6 +1206,215 @@ ONNX_MS_OPERATOR_SET_SCHEMA(
...
+        .Input(4,
+               "value_cache",
+               "Block-based value cache with shape (num_blocks, block_size, kv_num_heads, head_size). This is updated "
+               "in place within the op. This should be the same shape as key_cache.",
+               "T")
+        .Input(5,
+               "cumulative_sequence_length",
+               "A tensor with shape (batch_size + 1). It specifies the cumulative sequence lengths between the packed "
+               "entries in Q/K/V.",
+               "M")
```
tianleiwu, line 1361: The data type need not be same like mask's. I think we can use another type symbol like "S" for sequence length data type.

onnxruntime/core/graph/contrib_ops/bert_defs.cc
```diff
@@ -1206,6 +1206,215 @@ ONNX_MS_OPERATOR_SET_SCHEMA(
...
+                "T",
+                OpSchema::Optional)
+        .Output(2,
+                "value_cache_out",
+                "Block-based value cache with shape (num_blocks, block_size, kv_num_heads, head_size). This is always "
+                "the same tensor as value_cache.",
+                "T",
+                OpSchema::Optional)
+        .TypeConstraint("T", {"tensor(float16)", "tensor(bfloat16)"}, "Constrain input and output to float tensors.")
+        .TypeConstraint("M", {"tensor(int32)"}, "Constrain mask to int tensor.")
```
tianleiwu, line 1413: There is no mask input.

onnxruntime/core/graph/contrib_ops/bert_defs.cc
```diff
@@ -1206,6 +1206,215 @@ ONNX_MS_OPERATOR_SET_SCHEMA(
           GroupQueryAttentionTypeAndShapeInference(ctx, 3);
         }));
 
+constexpr const char* PagedAttention_ver1_doc = R"DOC(
+Paged Attention.
+
+This op leverages a block-based KV cache to enable continuous batching for LLMs.Currently, it is designed to work with
```
tianleiwu, line 1212: nit: add a space after LLMs.

onnxruntime/core/graph/contrib_ops/bert_defs.cc
```diff
@@ -1206,6 +1206,215 @@ ONNX_MS_OPERATOR_SET_SCHEMA(
...
+  // Shape inference for output tensor
+  if (hasInputShape(ctx, 0)) {
+    auto& query_shape = getInputShape(ctx, 0);
+    auto& query_dims = query_shape.dim();
+
+    if (query_dims.size() != 2) {
+      fail_shape_inference("Input 0 (query) shall be 2 dimensions");
+    }
+
+    if (hasInputShape(ctx, 2)) {
```
tianleiwu, line 1269: check whether input 2 exists instead of whether it has shape.

onnxruntime/core/graph/contrib_ops/bert_defs.cc
aciddelgado: We don't need it. I can talk with the Genai team to see if we can remove it.

onnxruntime/contrib_ops/cuda/bert/paged_attention_impl.cu
```diff
@@ -0,0 +1,391 @@
+/*
```
tianleiwu, line 1: I think the code is completely different from TRT bert demo. There is no need to add this header.

onnxruntime/contrib_ops/cuda/bert/paged_attention_helper.h
aciddelgado: T here is Tensor and since we are checking shape we need Tensors

onnxruntime/core/graph/contrib_ops/bert_defs.cc
tianleiwu: Since padding has been removed in the input, the input is treated as a long sequence. My understanding is that there is no much use of max query length as long as you know the total length of the long sequence.

For max kv length, it is only used to check the optional input cos/sin cache of rotary. But, there is no guarantee that the max kv length itself is correct. To verify max length is correct, we will have to reduce from length per sequence.

aciddelgado: Primarily, max seqlen q and k were included as input because these values are passed to flash attention. However, I believe we can remove these dependencies at a kernel level.

onnxruntime/contrib_ops/cpu/bert/attention_parameters.h
```diff
@@ -83,15 +82,13 @@ struct DecoderMaskedMultiHeadAttentionParameters : AttentionParameters {
...
+  int kv_num_heads;             // number of heads of key or value
+  int kv_hidden_size;           // hidden size of key or value
   int seqlen_past_kv_cache;     // sequence length of past kv tensor
   int seqlen_present_kv_cache;  // sequence length of present kv tensor
-  int kv_hidden_size;
-  int kv_num_heads;
-  int num_splits;         // number of splits for splitkv
-  int rotary_dim;         // rotary embedding dimension
-  int local_window_size;  // The window size excludes current token. It only includes tokens on the left side.
+  int num_splits;               // number of splits for splitkv
```
tianleiwu: num_splits exists in AttentionParameters

onnxruntime/core/graph/contrib_ops/bert_defs.cc
```diff
@@ -1206,6 +1206,195 @@ ONNX_MS_OPERATOR_SET_SCHEMA(
...
+               "Block-based value cache with shape (num_blocks, block_size, kv_num_heads, head_size). This is updated "
+               "in place within the op. This should be the same shape as key_cache.",
+               "T")
+        .Input(5,
+               "cumulative_sequence_length",
+               "A tensor with shape (batch_size + 1). It specifies the cumulative sequence lengths between the packed "
+               "entries in Q/K/V.",
+               "S")
+        .Input(6,
+               "seqlens",
```
tianleiwu: rename as past_seq_lens to avoid confusion.

onnxruntime/core/graph/contrib_ops/bert_defs.cc
```diff
@@ -1206,6 +1206,195 @@ ONNX_MS_OPERATOR_SET_SCHEMA(
...
+    ONNX_NAMESPACE::propagateElemTypeFromInputToOutput(ctx, 0, 1);
+    ONNX_NAMESPACE::propagateElemTypeFromInputToOutput(ctx, 0, 2);
+    // shapes
+    auto& key_cache_shape = getInputShape(ctx, 3);
+    auto& key_cache_dims = key_cache_shape.dim();
+    if (key_cache_dims.size() != 4) {
+      fail_shape_inference("The block-based KV cache inputs shall be 4 dimensions");
+    }
+    // KV cache in and out share the same buffer, thus they have the same shape
+    ONNX_NAMESPACE::propagateShapeFromInputToOutput(ctx, 3, 1);
```
tianleiwu, line 1301: also need propagate type

onnxruntime/core/graph/contrib_ops/bert_defs.cc
```diff
@@ -1206,6 +1206,195 @@ ONNX_MS_OPERATOR_SET_SCHEMA(
...
+    if (ctx.hasInput(2)) {
+      ONNX_NAMESPACE::TensorShapeProto output_shape;
+      propagateShapeFromInputToOutput(ctx, 0, 0);
+    } else {  // packed QKV
+      ONNX_NAMESPACE::TensorShapeProto output_shape;
+      *output_shape.add_dim() = query_dims[0];
+      int64_t num_heads = getAttribute(ctx, "num_heads", 0);
+      int64_t kv_num_heads = getAttribute(ctx, "kv_num_heads", 0);
+      int64_t hidden_size = query_dims[1].dim_value();
+      int64_t head_size = hidden_size / (num_heads + 2 * kv_num_heads);
```
tianleiwu, line 1280: do some simple check here like num_heads > 0, kv_num_heads > 0 and hidden_size % (num_heads + 2 * kv_num_heads) ==0.

tianleiwu (commented):
What design change needed if we want to support FP8 or FP4 paged attention in the future?

onnxruntime/contrib_ops/cuda/bert/paged_attention_helper.h
```diff
@@ -0,0 +1,278 @@
...
+                           "Input 'past_value' dimension 3 should be same as head_size, got ",
+                           value_cache_dims[3]);
+  }
+  return Status::OK();
+}
+
+template <typename T = Tensor>
+Status CheckSequenceLengthTensors(const T* cumulative_sequence_length, const T* seqlens, int& batch_size) {
+  const auto& cumulative_seqlen_dim = cumulative_sequence_length->Shape().GetDims();
+  if (cumulative_seqlen_dim.size() != 1) {
```
tianleiwu: also check the length is batch_size + 1

aciddelgado:
> What design change needed if we want to support FP8 or FP4 paged attention in the future?

New kernel necessary

tianleiwu (approved):

Issue #23614, opened 2025-02-07
[Performance]  ORT-WebGPU Average Pooling is working too long in edge case


grazder:
### Describe the issue

I'm using ORT with WebGPU for inference. I did profiling and found that 4 AveragePoolings takes almost half of model inference speed.

![Image](https://github.com/user-attachments/assets/9868d904-afac-4b35-b89d-68a45263df30)

Here is example of operation:

![Image](https://github.com/user-attachments/assets/973e7a05-422c-41f2-91a0-5401c9bd5b45)

AvgPool inference time has linear dependence on kernel size

Shapes // Time
* (20, 32) = 640 // 0.2ms;
* (40, 64) = 2560 (x4) // 0.8ms (x4)
* (80, 128) = 10240 (x16) // 3.2ms (x16)

Problem here is that it is an edge case, because it is literally ReduceMean on dims=(2,3). I changed the operation and got giant speed boost.

![Image](https://github.com/user-attachments/assets/96bae90a-46c7-4da4-800e-25580cfff44d)

I think that it is an edge case (that AvgPool == ReduceMean) and you can create a special handler for it.

### To reproduce

Just create AvgPool with big kernel_size

### Urgency

Not urgent, workaround was found

### Platform

Mac

### OS Version

15.2

### ONNX Runtime Installation

Released Package

### ONNX Runtime Version or Commit ID

1.20.1

### ONNX Runtime API

JavaScript

### Architecture

ARM64

### Execution Provider

Other / Unknown

### Is this a quantized model?

No

grazder:
I also noticed this behavior for `GlobalAveragePooling`

vadimkantorov:
cc @justinchuby

grazder:
@xhcao Thanks! I'm going to test this fix next week on my models.

PR #24598, opened 2025-04-30
[webgpu]: optimize pool operators

xhcao:
The patch optimizes pool operators when output size is small and kernel size is big

### DIFF
--- a/onnxruntime/core/providers/webgpu/nn/pool.cc
+++ b/onnxruntime/core/providers/webgpu/nn/pool.cc
@@ -95,6 +95,9 @@ Status PoolProgram::GenerateShaderCode(ShaderHelper& shader) const {
     var_decl_code = SS_GET(var_decl_ss);
 
     sampling_code = "      value = max(value, x_val);\n";
+    if (are_small_output_big_kernel_) {
+      downsampling_code = "  sum_or_max_shared[local_idx] = value;\n";
+    }
   } else {
     SS(var_decl_ss, kStringInitialSize);
     var_decl_ss << "  var value = " << (is_float16_ ? "f16(0)" : "f32(0)") << ";\n";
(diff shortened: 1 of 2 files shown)


xhcao:
The issue is from an user's model. 
If the input data shape is [1, 64, 128, 128] (NCHW), the original code needs one work group for 64 output elements. The kernel size is 128*128, so there is a 128*128 loop.
If the output size is small, one work group is responsible for one output element, there is only 128 loop.
@jchen10 PTAL

xhcao:
https://github.com/microsoft/onnxruntime/issues/23614

guschmue:
this ci error is unrelated to this PR - fix incoming

guschmue (approved):

vadimkantorov:
does this PR introduce a shortcut when the input size is equal to kernel size (equivalent to simply taking a mean)?

xhcao:
> does this PR also introduce a shortcut when the input size is equal to kernel size - same code path as simply taking a mean?

Hi, @vadimkantorov The path is valid when https://github.com/microsoft/onnxruntime/blob/7d719758f34ef99123cc26aa18bbfda508108c57/onnxruntime/core/providers/webgpu/nn/pool.cc#L287, when the kernel size is large and the output size is small.
`same code path as simply taking a mean?`, Do you mean `AveragePool` and `GlobalAveragePool`? The path is also valid for these operators.

vadimkantorov:
Yes, I mean global pooling, where output_size is strictly 1x1, while input_size (matching kernel_size as in https://github.com/microsoft/onnxruntime/issues/23614) can be 100 or anything else.

I guess output_size = 1, kernel_size = 100 would not pass this check, right?

xhcao:
> ut_size = kernel_size = 100 w

In https://github.com/microsoft/onnxruntime/issues/23614, the input shape is [1, 56, 80, 128], the kernel shape is [80, 128], so the output shape size is 56 (< 128), the kernel size is 80 * 128 (> 128), which will pass this check.

Issue #24500, opened 2025-04-22
[Build] The version info of onnxruntime dlls is not set correctly


BengtGustafsson:
### Describe the issue

When building your library and its provider dlls they don't get a good version number. onnxruntime.dll and onnxruntime_provider_shared.dll  get ORT_VERSION verbatim while the cuda and trt provider dlls don't get any version set at all.

The reason for ORT_VERSION is probably that you forgot the ${} that needs to surround variable names in cmake somewhere.

### Target platform

windows

### Build script

nothing special. Using your build.py with dml, cuda and trt.

### Error / output

versions on dlls wrong or missing.

### Visual Studio Version

VS2022

PR #24606, opened 2025-04-30
Fix version info for onnxruntime.dll and QNN, TensorRT, and CUDA EP dlls

adrianlizarraga:
### Description
Fixes https://github.com/microsoft/onnxruntime/issues/24500

- Fixes local build of onnxruntime.dll to have a valid version, such as "1.23.0", instead of the literal string "ORT_VERSION"
- Adds version info to onnxruntime_providers_qnn.dll, onnxruntime_providers_cuda.dll, and onnxruntime_providers_tensorrt.dll. It was missing completely. This was done by adding `onnxruntime_providers_*.rc` files to define each EP's [DLL version info](https://learn.microsoft.com/en-us/windows/win32/menurc/versioninfo-resource).

Fixed onnxruntime.dll version info (local non-ADO build):
image

Fixed onnxruntime_providers_qnn.dll version info (adds QNN SDK version too):
image

### Motivation and Context
We create dlls with invalid or missing version info.


onnxruntime/core/providers/qnn/onnxruntime_providers_qnn.rc
```diff
@@ -0,0 +1,47 @@
...
+FILETYPE        VFT_DLL
+FILESUBTYPE     VFT2_UNKNOWN
+
+BEGIN
+    BLOCK "StringFileInfo"
+    BEGIN
+        BLOCK "040904E4"
+        BEGIN
+            VALUE "CompanyName",      "Microsoft Corporation"
+            VALUE "FileDescription",  "ONNX Runtime QNN Execution Provider"
```
adrianlizarraga, line 33: Note: I copied core/dll/onnxruntime.rc and edited the FileDescription. There's an opportunity to add the QNN SDK version as well, which would be very useful.

cmake/onnxruntime_providers_cuda.cmake
```diff
@@ -122,17 +122,28 @@
   if (onnxruntime_REDUCED_OPS_BUILD)
     substitute_op_reduction_srcs(onnxruntime_providers_cuda_src)
   endif()
+
```
jywu-msft, line 125: do these changes need to be done under if(WIN32) ?

jywu-msft:
do we want to do this for only windows in the cmake file?

cmake/onnxruntime_providers_cuda.cmake
adrianlizarraga: It wasn't explicitly required as openvino and cpu didn't check and all CI passed too. But I agree that your suggestion is just cleaner. I don't feel right relying on this being ignored on non-Windows.

Addressed.

jywu-msft (approved):

snnn:

PR #24609, opened 2025-04-30
Fix unused variable warning in onnxruntime/core/providers/coreml/builders/impl/gather_op_builder.cc.

edgchen1:
### Description

Fix unused variable warning in onnxruntime/core/providers/coreml/builders/impl/gather_op_builder.cc.

### Motivation and Context

Fix build.


carzh (approved):
Thank you for making the PR fix!

PR #24610, opened 2025-05-01
[webgpu] allows atomic type for output

fs-eire:
### Description

This PR adds support for atomic types for program output. Applying atomic type on program output can be done in the following way:
```c++
program.AddOutput({output_tensor, ProgramTensorMetadataDependency::TypeAndRank, ProgramOutput::Atomic});
```
The last
```

The support for atomic type is minimal. According to [spec](https://www.w3.org/TR/WGSL/#atomic-types), the only valid operations on atomic objects are the [atomic builtin functions](https://www.w3.org/TR/WGSL/#atomic-builtin-functions). This means atomic types cannot be accessed (get/set) using the normal way. Get* and Set* functions will not be working on atomic types for indices helper. Use the WGSL builtin functions directly. OffsetToIndices and IndicesToOffset functions still work.

### DIFF
--- a/onnxruntime/core/providers/webgpu/program.h
+++ b/onnxruntime/core/providers/webgpu/program.h
@@ -233,13 +233,21 @@ struct ProgramInput {
 };
 
 struct ProgramOutput {
+ private:
+  struct AtomicTag {};
+
+ public:
+  constexpr static const AtomicTag Atomic{};
+
   ProgramOutput(Tensor* tensor);
   ProgramOutput(Tensor* tensor, ProgramTensorMetadataDependency dependency, int component = 1);
+  ProgramOutput(Tensor* tensor, ProgramTensorMetadataDependency dependency, AtomicTag);
   ProgramOutput(Tensor* tensor, ProgramTensorMetadataDependency dependency, const TensorShape& override_shape, int component);
 
   Tensor* tensor;
   ProgramTensorMetadataDependency dependency;
   ProgramVariableDataType var_type;
+  bool is_atomic;
   bool use_override_shape;
   TensorShape override_shape;
 };
--- a/onnxruntime/core/providers/webgpu/shader_helper.cc
+++ b/onnxruntime/core/providers/webgpu/shader_helper.cc
@@ -123,7 +123,12 @@ const ShaderIndicesHelper& ShaderHelper::AddIndices(const std::string& name, Sha
 #ifndef NDEBUG  // if debug build
 namespace {
 // Validate if the tensor element type matches the program variable data type
-Status ValidateVariableDataType(int32_t element_type, ProgramVariableDataType var_type) {
+Status ValidateVariableDataType(int32_t element_type, ProgramVariableDataType var_type, bool is_atomic = false) {
+  if (is_atomic) {
+    ORT_RETURN_IF_NOT(var_type == ProgramVariableDataType::Int32 || var_type == ProgramVariableDataType::Uint32,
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: onnxruntime/core/providers/webgpu/program.h [hunk 1]
```diff
@@ -233,13 +233,21 @@ struct ProgramInput {
...
 struct ProgramOutput {
+ private:
+  struct AtomicTag {};
+
+ public:
+  constexpr static const AtomicTag Atomic{};
+
   ProgramOutput(Tensor* tensor);
   ProgramOutput(Tensor* tensor, ProgramTensorMetadataDependency dependency, int component = 1);
+  ProgramOutput(Tensor* tensor, ProgramTensorMetadataDependency dependency, AtomicTag);
```
satyajandhyala, line 244: Can we  use `is_atomic` flag as argument with default value false instead of using operator overloading here, using AtomicTag,

onnxruntime/core/providers/webgpu/shader_helper.cc
```diff
@@ -408,12 +413,22 @@ Status ShaderHelper::GenerateSourceCode(std::string& code, std::vector<int>& sha
   //
   // Input/output variables
   //
-  size_t variable_count = 0;
-  for (const auto& input : input_vars_) {
-    ss << "@group(0) @binding(" << variable_count++ << ") var<storage, read> " << input->name_ << ": array<" << input->StorageType() << ">;\n";
+  for (size_t i = 0; i < input_vars_.size(); ++i) {
+    const auto& input = input_vars_[i];
+    ss << "@group(0) @binding(" << i << ") var<storage, read> " << input->name_ << ": array<" << input->StorageType() << ">;\n";
```
satyajandhyala, line 418: Do we need to use `input_vars_.size() + i` instead of `i` to make it unique?

onnxruntime/core/providers/webgpu/shader_helper.cc
```diff
@@ -408,12 +413,22 @@ Status ShaderHelper::GenerateSourceCode(std::string& code, std::vector<int>& sha
...
-  for (const auto& output : output_vars_) {
-    ss << "@group(0) @binding(" << variable_count++ << ") var<storage, read_write> " << output->name_ << ": array<" << output->StorageType() << ">;\n";
+  for (size_t i = 0; i < output_vars_.size(); ++i) {
+    const auto& output = output_vars_[i];
+    bool is_atomic = program_.Outputs()[i].is_atomic;
+    ss << "@group(0) @binding(" << i << ") var<storage, read_write> " << output->name_ << ": array<";
+    if (is_atomic) {
+      ss << "atomic<";
+    }
+    ss << output->StorageType();
```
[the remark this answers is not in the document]
fs-eire: it's checked in the updated implementation of `ValidateVariableDataType()`. the tensor data type must be int32 or uint32 otherwise it will fail.

onnxruntime/core/providers/webgpu/program.h
fs-eire: Using tag type helps to give better overload flexibility and better sementic. As described in the PR description, you can use the following way to specify:
```
program.AddOutput({output_tensor, ProgramTensorMetadataDependency::TypeAndRank, ProgramOutput::Atomic});
```

onnxruntime/core/providers/webgpu/shader_helper.cc
fs-eire: thanks for catching. updated with fix.

satyajandhyala (approved):

PR #24623, opened 2025-05-02
[Native WebGPU] Added ScatterND

satyajandhyala:
### Description
Added ScatterND operator to Native WebGPU EP.

### Motivation and Context

Required to increase coverage.

### DIFF
--- a/onnxruntime/core/providers/webgpu/tensor/scatter_nd.cc
+++ b/onnxruntime/core/providers/webgpu/tensor/scatter_nd.cc
@@ -82,12 +82,12 @@ Status ScatterNDProgram::GenerateShaderCode(ShaderHelper& shader) const {
     std::ostringstream ss;
     // The stride is size is one less than the shape size.
     if (output_rank == 1) {
-      ss << "    let element_count_dim = 1;\n";
+      ss << "    let element_count_dim = 1u;\n";
     } else if (output_rank == 2) {
-      ss << "    let element_count_dim = select(uniforms.output_stride, 1, " << (parallel ? "i - indices_start" : "i") << " == 1 " << ");\n";
+      ss << "    let element_count_dim = select(uniforms.output_stride, 1u, " << (parallel ? "i - indices_start" : "i") << " == 1 " << ");\n";
     } else {
       ss << "    let output_stride_index = " << (parallel ? "i - indices_start" : "i") << ";\n"
-         << "    let element_count_dim = select(uniforms.output_stride[output_stride_index], 1, output_stride_index == " << (output_rank - 1) << ");\n";
+         << "    let element_count_dim = select(uniforms.output_stride[output_stride_index], 1u, output_stride_index == " << (output_rank - 1) << ");\n";
     }
     if (output_rank == 1) {
       ss << "    let dim_value = uniforms.output_shape;\n";
--- a/onnxruntime/core/providers/webgpu/tensor/scatter_nd.cc
+++ b/onnxruntime/core/providers/webgpu/tensor/scatter_nd.cc
@@ -86,8 +86,8 @@ Status ScatterNDProgram::GenerateShaderCode(ShaderHelper& shader) const {
     } else {
       ss << "    let element_count_dim = select(" << GetElementAt("uniforms.output_stride", "i - indices_start", output_rank - 1) << ", 1, i - indices_start == " << (output_rank - 1) << ");\n";
     }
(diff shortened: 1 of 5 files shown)


onnxruntime/test/providers/cpu/tensor/scatter_nd_op_test.cc
```diff
@@ -210,7 +210,7 @@ TEST(ScatterNDOpTest, ScatterND_18_mul_long_shape) {
   // The linter complains if the line is split into multiple lines.
   test1.AddInput<float>("updates", {3, 2, 3, 1, 1, 1, 1}, {2.0f, 4.0f, 8.0f, 16.0f, 32.0f, 64.0f, 128.0f, 256.0f, 512.0f, 1024.0f, 2048.0f, 4096.0f, 8192.0f, 16384.0f, 32768.0f, 65536.0f, 131072.0f, 262144.0f});
   test1.AddOutput<float>("output", {2, 2, 3, 1, 1, 1, 1}, {1638.4f, 6553.6f, 26214.4f, 104857.6f, 419430.4f, 1677721.625f, 12.8f, 25.6f, 51.2f, 102.4f, 204.8f, 409.6f});
-  test1.Run(OpTester::ExpectResult::kExpectSuccess, "", {kTensorrtExecutionProvider, kOpenVINOExecutionProvider});
+  test1.Run(OpTester::ExpectResult::kExpectSuccess, "", {kTensorrtExecutionProvider, kOpenVINOExecutionProvider, kWebGpuExecutionProvider});
```
prathikr, line 213: Can you add a test for the multiple updates to a single index? You can add the SAM_float32 test case from jsep test suite.

satyajandhyala: This test case results in shape such as array<vect4<u32>> which is not handled in the current code. The fix could be from setting the shape/stride.

prathikr: Since webgpu ep has a pin to cpu feature, should we remove the duplicate indices check and instead tell customers to pin scatternd to cpu when they run into the duplicate indices edge case? Then we can investigate why their model exports scatternd in undefined behavior case?

@fs-eire @guschmue thoughts?

satyajandhyala: We are not covering the duplicate indices in this PR.

prathikr: If we are not covering duplicate indices, then we need to remove the hasDuplicates check as it will cause performance bottleneck for models with reduction=none. I'd suggest basing the webgpu kernel off the jsep scatternd kernel before I adjusted it for this edge case.

guschmue: for the duplicate indices - I think the spec says that those are not allowed but we know about 1 model that does this and we know cpu can handle it.
Is it just 1 model?

onnxruntime/core/providers/webgpu/tensor/scatter_nd.cc
```diff
@@ -0,0 +1,263 @@
...
+    data_type_str = "f32";
+  } else {
+    // Default value.
+    data_type_str = "output_element_t";
+  }
+  if (reduction_ != ScatterNDReduction::None && !reducible) {
+    ORT_THROW("ScatterND: Reduction is not supported for data type ", data_type_str);
+  }
+  shader.MainFunctionBody() << shader.GuardAgainstOutOfBoundsWorkgroupSizes("uniforms.output_size")
+                            << "  var hasDuplicates = false;\n";
```
guschmue, line 140: maybe this should be shared memory so we don't need to always do this?

guschmue, line 140: if we are not sure that we handle duplicated indicies 100% and the spec doesn't allow it, would be also also an to throw error. Just don't know a way of throwing from wgsl.

satyajandhyala: I will remove the code computing ` hasDuplicates` as suggested.

onnxruntime/test/providers/cpu/tensor/scatter_nd_op_test.cc
prathikr: SAM and Dexter's test model have hit this edge case, others may too

guschmue (approved):

PR #24629, opened 2025-05-03
[QNN EP] Enable automatic selection of QNN EP for PREFER_NPU policy

adrianlizarraga:
### Description
- Enables automatic selection of QNN EP for PREFER_NPU policy
- Fixes cpuid vendor id for Qualcomm to be `'Q' | ('C' << 8) | ('O' << 16) | ('M' << 24);`

Sample code from unit test:
```c++
// Tests autoEP feature to automatically select an EP that supports the NPU.
// Currently only works on Windows.
TEST_F(QnnHTPBackendTests, AutoEp_PreferNpu) {
  ASSERT_ORTSTATUS_OK(Ort::GetApi().RegisterExecutionProviderLibrary(*ort_env, kQnnExecutionProvider,
                                                                     ORT_TSTR("onnxruntime_providers_qnn.dll")));

  Ort::SessionOptions so;
  so.SetEpSelectionPolicy(OrtExecutionProviderDevicePolicy_PREFER_NPU);

  const ORTCHAR_T* ort_model_path = ORT_MODEL_FOLDER "nhwc_resize_sizes_opset18.quant.onnx";
  Ort::Session session(*ort_env, ort_model_path, so);
  EXPECT_TRUE(SessionHasEp(session, kQnnExecutionProvider));

  ASSERT_ORTSTATUS_OK(Ort::GetApi().UnregisterExecutionProviderLibrary(*ort_env, kQnnExecutionProvider));
}
```

### Motivation and Context
A recent feature allows ORT to automatically select an EP according to policies set by the user (e.g., prefer npu or prefer gpu). This PR allows QNN EP to be potentially selected when the user sets the `PREFER_NPU` policy.

### DIFF
--- a/onnxruntime/core/providers/qnn/qnn_provider_factory.cc
+++ b/onnxruntime/core/providers/qnn/qnn_provider_factory.cc
@@ -79,6 +79,27 @@ struct QNN_Provider : Provider {
     return std::make_shared<onnxruntime::QNNProviderFactory>(*provider_options, config_options);
   }
 
+  Status CreateIExecutionProvider(const OrtHardwareDevice* const* /*devices*/,
+                                  const OrtKeyValuePairs* const* /*ep_metadata*/,
+                                  size_t num_devices,
+                                  ProviderOptions& provider_options,
+                                  const OrtSessionOptions& session_options,
+                                  const OrtLogger& logger,
+                                  std::unique_ptr<IExecutionProvider>& ep) override {
+    if (num_devices != 1) {
+      return Status(common::ONNXRUNTIME, ORT_EP_FAIL, "QNN EP only supports one device.");
+    }
+
+    const ConfigOptions* config_options = &session_options.GetConfigOptions();
+
+    std::array<const void*, 2> configs_array = {&provider_options, config_options};
+    const void* arg = reinterpret_cast<const void*>(&configs_array);
+    auto ep_factory = CreateExecutionProviderFactory(arg);
+    ep = ep_factory->CreateProvider(session_options, logger);
+
+    return Status::OK();
+  }
+
   void Initialize() override {}
   void Shutdown() override {}
 } g_provider;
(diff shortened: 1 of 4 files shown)


onnxruntime/core/providers/qnn/qnn_provider_factory.cc
```diff
@@ -93,4 +114,119 @@ ORT_API(onnxruntime::Provider*, GetProvider) {
...
+  const OrtApi* ort_api = ort_api_base->GetApi(ORT_API_VERSION);
+
+  // Factory could use registration_name or define its own EP name.
+  auto factory_npu = std::make_unique<QnnEpFactory>(*ort_api,
+                                                    onnxruntime::kQnnExecutionProvider,
+                                                    OrtHardwareDeviceType_NPU, "htp");
+
+  // If want to support GPU, create a new factory instance because QNN EP is not currently setup to partition a single model
+  // among heterogeneous devices.
+  // std::unique_ptr<OrtEpFactory> factory_gpu = std::make_unique<QnnEpFactory>(*ort_api, "QNNExecutionProvider_GPU", OrtHardwareDeviceType_GPU, "gpu");
```
skottmckay, line 214: Could this support GPU or we don't know how well it works?

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
```diff
@@ -93,4 +114,119 @@ ORT_API(onnxruntime::Provider*, GetProvider) {
...
+  }
+
+  static void ReleaseEpImpl(OrtEpFactory* /*this_ptr*/, OrtEp* /*ep*/) {
+    // no-op as we never create an EP here.
+  }
+
+  const OrtApi& ort_api;
+  const std::string ep_name;                       // EP name
+  const std::string vendor{"Microsoft"};           // EP vendor name
+  const uint32_t vendor_id{0x4d4f4351};            // Qualcomm vendor ID
```
skottmckay, line 194: should maybe point out that this isn't an official vendor id. it's an arbitrary value created by bitshifting QCOM. a link to the cpuid_info code would be good. 

https://github.com/microsoft/onnxruntime/blob/6fa8ba107f461d72073ee4842830a9d0a96d2546/onnxruntime/core/common/cpuid_info.cc#L158

but maybe I've screwed that up. I shifted the Q by 24 bits (original code had it as the first byte) thinking that would make the values 0xQCOM but if the vendor_id is 0x4d4f4351 I've not taken into account it's little endian and I've made it 0xMOCQ which is the complete opposite of my intention.

skottmckay (approved):

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
skottmckay: But godbolt.org suggests the vendor id with the latest main should be 0x51434f4d which was what I intended.

image

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
adrianlizarraga: We don't know how well it works so didn't want to enable it yet.

jywu-msft (dismissed):

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
adrianlizarraga: Interesting. It's not getting the vendor_id from that line in cpuid_info.cc. The vendor is `""`.
image

Let me trace where it is getting the vendor_id from.

adrianlizarraga: It's getting the vendor id from dxcore. It is 'M', 'O', 'C', 'Q' = 0x4d4f4351.
image

I think I should update the cpuid_info.cc `GetVendorId` to return:
```c++
if (vendor.find("Qualcomm") == 0) return 'M' << 24 | 'O' << 16 | 'C' << 8 | 'Q';
```

adrianlizarraga: Yeah, looking around the internet, the correct value is 0x4d4f4351. Also, it makes since in a little-endian machine, since the 'Q' should be the first byte, which means it is **not** shifted.

I'm updating cpuid_info.cc to return `'M' << 24 | 'O' << 16 | 'C' << 8 | 'Q';`

skottmckay: Thanks! That needs to match. I didn't realise this was an 'official' id.

skottmckay (approved):

snnn:
This PR has been cherry-picked into the `rel-1.22.0` branch in PR #24630. Removing the `release:1.22.0` label.

PR #24614, opened 2025-05-01
Add Python bindings for "compile" and "auto EP" APIs

adrianlizarraga:
### Description
Python API updates for auto ep selection and the compilation API.
- Adds Python API `SessionOptions.add_provider()` (equivalent to C API's `SessionOptionsAppendExecutionProvider`)
- Adds Python API `SessionOptions.add_provider_for_devices()` (equivalent to C API's `SessionOptionsAppendExecutionProvider_V2`)
- Adds Python API `SessionOptions.set_provider_selection_policy()` (equivalent to C API's `SessionOptionsSetEpSelectionPolicy`)
- Adds Python API class `ModelCompiler` to compile models (wraps C API's `OrtModelCompilationOptions` and `CompileModel()`)
- TODO: Finish delegate callback. Need to add a `void*` parameter to delegate function.

### Sample program that uses autoep APIs
Adapted from a unit test.
```python
    def test_cuda_prefer_gpu_and_inference(self):
        """
        Test selecting CUDA EP via the PREFER_GPU policy and running inference.
        """
        ep_lib_path = "onnxruntime_providers_cuda.dll"
        ep_registration_name = "CUDAExecutionProvider"

        if sys.platform != "win32":
            self.skipTest("Skipping test because device discovery is only supported on Windows")

        if not os.path.exists(ep_lib_path):
            self.skipTest(f"Skipping test because EP library '{ep_lib_path}' cannot be found")

        onnxrt.register_execution_provider_library(ep_registration_name, os.path.realpath(ep_lib_path))

        # Set a policy to prefer GPU. Cuda should be selected.
        sess_options = onnxrt.SessionOptions()
        sess_options.set_provider_selection_policy(onnxrt.OrtExecutionProviderDevicePolicy.PREFER_GPU)
        self.assertTrue(sess_options.has_providers())

        # Run sample model and check output
        sess = onnxrt.InferenceSession(get_name("mul_1.onnx"), sess_options=sess_options)

        x = np.array([[1.0, 2.0], [3.0, 4.0], [5.0, 6.0]], dtype=np.float32)
        input_name = sess.get_inputs()[0].name
        res = sess.run([], {input_name: x})
        output_expected = np.array([[1.0, 4.0], [9.0, 16.0], [25.0, 36.0]], dtype=np.float32)
        np.testing.assert_allclose(output_expected, res[0], rtol=1e-05, atol=1e-08)
```
### Sample program that uses compile APIs
Adapted from a unit test that compiles using EP selection policy.
```python
[...]

PR #24634, opened 2025-05-05
Add support for EP selection delegate callback to Python bindings

adrianlizarraga:
### Description
Follow up to https://github.com/microsoft/onnxruntime/pull/24614

Example Python program (adapted from unit tests) that specifies a custom EP selection function to select a OrtEpDevice(s) for compiling:
```python
    def test_compile_with_ep_selection_delegate(self):
        # ...
        # User's custom EP selection function.
        def my_delegate(
            ep_devices: Sequence[onnxrt.OrtEpDevice],
            model_metadata: dict[str, str],
            runtime_metadata: dict[str, str],
            max_selections: int,
        ) -> Sequence[onnxrt.OrtEpDevice]:
            self.assertTrue(len(model_metadata) > 0)
            self.assertTrue(ep_devices and max_selections > 0)

            # Select the first and last devices (if there are more than one)
            selected_devices = [ep_devices[0]]
            if max_selections > 2 and len(ep_devices) > 1:
                selected_devices.append(ep_devices[-1])  # ORT CPU EP is always last

            return selected_devices

        session_options = onnxrt.SessionOptions()
        session_options.set_provider_selection_policy_delegate(my_delegate)

        model_compiler = onnxrt.ModelCompiler(
            session_options,
            input_model_path,
            embed_compiled_data_into_model=True,
            external_initializers_file_path=None,
        )
        model_compiler.compile_to_file(output_model_path)
```

How to raise an exception from the Python EP selection function:
```python
        # User's custom EP selection function.
        custom_error_message = "MY ERROR"

        def my_delegate_that_fails(
            ep_devices: Sequence[onnxrt.OrtEpDevice],
            model_metadata: dict[str, str],
            runtime_metadata: dict[str, str],
            max_selections: int,
        ) -> Sequence[onnxrt.OrtEpDevice]:
            self.assertTrue(len(ep_devices) >= 1)
            raise ValueError(custom_error_message)

        sess_options = onnxrt.SessionOptions()
        sess_options.set_provider_selection_policy_delegate(my_delegate_that_fails)

        # Create session and expect ORT to raise a Fail exception that contains our message.
        with self.assertRaises(Fail) as context:
            onnxrt.InferenceSession(get_name("mul_1.onnx"), sess_options=sess_options)
        self.assertIn(custom_error_message, str(context.exception))
```


include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -443,13 +443,14 @@ typedef enum OrtExecutionProviderDevicePolicy {
...
  *                    ORT will release the OrtStatus* if not null.
  */
-typedef OrtStatus* (*EpSelectionDelegate)(_In_ const OrtEpDevice** ep_devices,
-                                          _In_ size_t num_devices,
-                                          _In_ const OrtKeyValuePairs* model_metadata,
-                                          _In_opt_ const OrtKeyValuePairs* runtime_metadata,
-                                          _Inout_ const OrtEpDevice** selected,
-                                          _In_ size_t max_selected,
-                                          _Out_ size_t* num_selected);
+typedef OrtStatus*(ORT_API_CALL* EpSelectionDelegate)(_In_ void* user_param,
```
skottmckay, line 446: I put the `void*` last in my version as it's optional. Doesn't need to be passed for native code.

onnxruntime/python/onnxruntime_pybind_state.cc
```diff
@@ -1798,38 +1799,69 @@ void addGlobalMethods(py::module& m) {
...
-//   *num_selected = py_selected.size();
-//   for (size_t i = 0; i < py_selected.size(); ++i) {
-//     selected[i] = py_selected[i];
-//   }
-//
-//   return nullptr;
-// };
+#if !defined(ORT_MINIMAL_BUILD)
+/**
+ * Calls the user's Python EP selection function and coverts the results to a format that can be used
```
skottmckay, line 1804:
```suggestion
 * Calls the user's Python EP selection function and converts the results to a format that can be used
```

onnxruntime/test/python/onnxruntime_test_python_autoep.py
```diff
@@ -100,8 +101,83 @@ def test_cuda_prefer_gpu_and_inference(self):
...
-        # self.unregister_execution_provider_library(ep_registration_name)
+        del sess  # Delete session before unregistering library
+        self.unregister_execution_provider_library(ep_registration_name)
+
+    def test_cuda_ep_selection_delegate_and_inference(self):
+        """
+        Test selecting CUDA EP via the custom EP selection delegate function and then run inference.
+        """
+        ep_lib_path = "onnxruntime_providers_cuda.dll"
+        ep_registration_name = "CUDAExecutionProvider"
```
skottmckay, line 112: nit: the registration name and EP names don't need to match. could we call this 'ep_name' to avoid potentially creating an assumption that they always do?

skottmckay (approved):

onnxruntime/test/python/onnxruntime_test_python_autoep.py
adrianlizarraga: Good call. Will address in a separate PR as I actually used `ep_registration_name` for all tests. I hope it's ok to merge now to get it into main.

PR #24637, opened 2025-05-05
fix checks for metal when running under wasm

guschmue:
When under wasm we can't check for metal by looking at backend because it will always be WEBGPU.
Because of this we'll take the DP4A path on metal that results in sub-optimal performance.
Use vendor to check for metal instead.

### DIFF
--- a/onnxruntime/contrib_ops/webgpu/quantization/subgroup_matrix_matmul_nbits.cc
+++ b/onnxruntime/contrib_ops/webgpu/quantization/subgroup_matrix_matmul_nbits.cc
@@ -248,8 +248,8 @@ bool CanApplySubgroupMatrixMatMulNBits(onnxruntime::webgpu::ComputeContext& cont
   // some precision issues with subgroupMatrixMultiplyAccumulate. It is possible to support higher accuracy
   // by setting compute_precision to Fp32, but that will be slower. For 1K token prefill FP16 Phi 3.5 is around 5s,
   // FP322 is around 7s.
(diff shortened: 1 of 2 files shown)


onnxruntime/contrib_ops/webgpu/quantization/subgroup_matrix_matmul_nbits.cc
```diff
@@ -248,8 +248,8 @@ bool CanApplySubgroupMatrixMatMulNBits(onnxruntime::webgpu::ComputeContext& cont
   // some precision issues with subgroupMatrixMultiplyAccumulate. It is possible to support higher accuracy
   // by setting compute_precision to Fp32, but that will be slower. For 1K token prefill FP16 Phi 3.5 is around 5s,
   // FP322 is around 7s.
-  return context.AdapterInfo().backendType == wgpu::BackendType::Metal &&
-         has_subgroup_matrix &&
+  return has_subgroup_matrix &&
```
prathikr, line 251: should these be swapped to maintain short circuit behavior?

fs-eire (approved):
LGTM

onnxruntime/contrib_ops/webgpu/quantization/subgroup_matrix_matmul_nbits.cc
fs-eire: Ideally, a part of the condition can be calculated only once in initialization phrase. But I am OK with it as everything looks cheap enough.

PR #24642, opened 2025-05-05
Add RMSNormalization from ONNX opset 23

shubhambhokare1:
Add RMSNormalization op that was added in ONNX opset 23.


titaiwangms:
Replaced by #24875

Issue #24070, opened 2025-03-17
[Web] Different model output with WebGPU vs WASM (or Python with the CUDA EP)


Dexterp37:
### Describe the issue

Using the model referenced below (see STR) produces a substantially different output when using the WebGPU provider (**wrong**) vs the WASM provider (**correct**). The difference is big enough in the output to indicate some bug in some WebGPU operator.

To further confirm this, I ran inference with the same model using onnx in Python: the correct results are consistent with both the CUDA and the CPU providers.

I'm lost at how to debug this further unfortunately.

### To reproduce

1. Host the following web page with a minimal example. Note that it's using 1.21.0 from a CDN, but the result doesn't change even with the latest available dev version. The needed files can be downloaded: [model](https://drive.google.com/file/d/1Zyv1iDGFEX3DqgJ5MtuIPZDf6YFaLVC3/view?usp=sharing) and [input image](https://drive.google.com/file/d/1RYjdtl2cCUFqzwsmJWKcNWKkNaN9HH0b/view?usp=sharing).

```html
<html>

<head>
  <script src="https://cdn.jsdelivr.net/npm/onnxruntime-web@1.21.0/dist/ort.min.js"></script>
</head>

<body>
    <h1>Example</h1>
    <div>
        <img id="image" src="turtle3.png" alt="Image preview..." width="1333" height="750" hidden>
        <canvas id="mycanvas" width="1333" height="750"></canvas>
    </div>

</body>
<script>
    async function startInferenceSession(modelPath) {
      const session = await ort
        .InferenceSession
        .create(
          modelPath,
          {
            executionProviders: [
              // 'wasm',
              'webgpu'
            ],
            graphOptimizationLevel: 'all',
          }
        );
      console.log('Inference session created');
      return session;
    }


    async function runInference(session, input) {
      const feeds = {};
      feeds[session.inputNames[0]] = input;
      const outputData = await session.run(feeds);
      return outputData;
    }
    
    async function imageToTensor(
      imageData,
    ) {
      const mean = [0.485, 0.456, 0.406]
      const std = [0.229, 0.224, 0.225]

      const rawTensor = await ort.Tensor.fromImage(imageData)

      // Normalize the tensor as needed.
      const rawData = rawTensor.data
      const [_, C, H, W] = rawTensor.dims;

      const normalizedData = new Float32Array(rawData.length)

      const dataPerChannel = H * W;
      for (let c = 0; c < C; c++) {
[...]

PR #24650, opened 2025-05-06
[JSEP] Fix outputSize calculation causing duplicate indices.

satyajandhyala:
### Description
Fix the outputSize computation causing duplicate indices. The outputSize should be the size of indices tensor without counting the last dimension.

### Motivation and Context

Fix the issue https://github.com/microsoft/onnxruntime/issues/24070


satyajandhyala:
> It seems that the unit test is failing

Fix the err.

fs-eire:
Close & reopen to trigger the pipelines

Issue #24636, opened 2025-05-05
Regression in TreeEnsembleRegressor if the provided graph is a DAG


cbourjau:
### Describe the issue

Up until the introduction of the new `TreeEnsemble` operator, the ONNX standard did not support splitting on set memberships. This was problematic since such splits are commonly associated with categorical features. To support these split types, one was forced to compare each value in the set with an explicit `BRANCH_EQ` node. When applied naively, this would mean that the subtree behind the true-branch of each split had to be copied - a quite untenable situation in many scenarios.

The solution to this problem was to build a DAG rather than a Tree such that each true branch points to the same node (see ASCII diagram in the test case below). This used to work with `onnxruntime<=1.20`. However, it appears that a regression to these kinds of "trees" was introduced [here](https://github.com/microsoft/onnxruntime/pull/22333). Thus, the following test case worked as intended on `onnxruntime<=1.20` but fails on the latest 1.21.1 version.

### To reproduce

```python
import numpy as np
import onnxruntime as ort
import spox.opset.ai.onnx.ml.v3 as ml
from spox import Tensor, argument, build

# DAG structure:
#  __0
#  |  \
#  |___2
#  |    \
#  1     3
_NA = -1
kwargs = {
    "n_targets": 1,
    "nodes_truenodeids": [1, _NA, 1, _NA],
    "nodes_falsenodeids": [2, _NA, 3, _NA],
    "nodes_featureids": [0, _NA, 0, _NA],
    "nodes_missing_value_tracks_true": [1, _NA, 1, _NA],
    "nodes_modes": [
        "BRANCH_EQ",
        "LEAF",
        "BRANCH_EQ",
        "LEAF",
    ],
    "nodes_nodeids": [0, 1, 2, 3],
    "nodes_treeids": [0, 0, 0, 0],
    "nodes_values_as_tensor": np.array([0.0, _NA, 1.0, _NA], np.float64),
    "target_ids": [0, 1],
    "target_nodeids": [1, 3],
    "target_treeids": [0, 0],
    "target_weights_as_tensor": np.array([10.0, 20.0]),
}
inputs = np.array([[0.0], [1.0], [-1.0]])
expected = np.array([[10.0], [10.0], [20.0]], np.float32)

x = argument(Tensor(np.float64, ("N", 1)))
y = ml.tree_ensemble_regressor(x, **kwargs)

model = build({"x": x}, {"y": y})
with open("test.onnx", "wb") as f:
[...]

PR #24654, opened 2025-05-06
Fix initialization of same_node_ in TreeEnsemble

xadupre:
### Description

For TreeEnsemble, onnxruntime tries to fuse multiple nodes BRANCH_EQ into one node BRANCH_MEMBER. When a tree only contains BRANCH_EQ nodes, the final tree could be a mix between BRANCH_EQ and BRANCH_MEMBER. To be more efficient, onnxruntime detects that all the nodes use the same rule and avoids checking that value for every node while getting the final leaf. This detection happened before the fusion into BRANCH_MEMBER. This PR detects that this check must be done again. This extra cost only happens when a tree only contains nodes BRANCH_EQ and should not be much. It only happens during the initialization.

### Motivation and Context
Fixes issue #24636.


onnxruntime/test/providers/cpu/ml/treeregressor_test.cc
```diff
@@ -844,5 +844,46 @@ TEST(MLOpTest, TreeRegressorTrueNodeBeforeNode) {
   test.Run();
 }
 
+TEST(MLOpTest, TreeRegressorBranchEq) {
```
cbourjau, line 847: It is great to have this test here. Should we also add a test that checks that the folding of the chain of `BRANCH_EQ` nodes actually happened?

cbourjau (commented):
Excellent! Thanks for looking into this so quickly!

yuslepukhin:
>   T operator*(double val) { return has_score ? score * static_cast<T>(val) : 0; }

Potentially important nit: these two operators must be const.

---
Refers to: onnxruntime/core/providers/cpu/ml/tree_ensemble_aggregator.h:53 in 9ef3585. [](commit_id = 9ef3585bc1fdc67dd3ab211c67b360382d6f9363, deletion_comment = False)

yuslepukhin:
>   size_t limit;

Suggest both i and fpos to be size_t, that would reduce the number of casts

---
Refers to: onnxruntime/core/providers/cpu/ml/tree_ensemble_common.h:149 in 9ef3585. [](commit_id = 9ef3585bc1fdc67dd3ab211c67b360382d6f9363, deletion_comment = False)

yuslepukhin (dismissed):
:shipit:

yuslepukhin (dismissed):
:shipit:

cbourjau:
Is anything missing on this PR?

xadupre:
> Is anything missing on this PR?

No, I just forgot to merge which I just did.

PR #24657, opened 2025-05-06
[QNN EP] Fix Resize Op support translation

quic-tirupath:
- Use ResizeNearestNeighbor Op for Resize with interpolation_mode=Nearest and rank-4 inputs.
 - Add a Unit test to verify the modified translation.

### Description
ResizeNearestNeighbor Op is faster for Resize with interpolation_mode=Nearest and rank-4 inputs.

### Motivation and Context
This commit matches Resize Op behavior in QNN-EP with QNN Offline converter path. This fix also improves inference time.


HectorSVC (approved):
:shipit:

PR #24660, opened 2025-05-06
Use GUID_DEVCLASS_COMPUTEACCELERATOR for NPU discovery

ashrit-ms:
### Description
Updated device discovery to use `GUID_DEVCLASS_COMPUTEACCELERATOR` instead of `GUID_DEVCLASS_SYSTEM` when querying SetupAPI for potential NPU devices. This provides a more specific and accurate class for identifying compute accelerators like NPUs.

This change also saves us an average of 5 milliseconds by not looping through unnecessary system devices.

### Motivation and Context
When looking for NPUs, the previous code used `GUID_DEVCLASS_SYSTEM` as the class to query for potential devices and didn't return the Qualcomm NPU.


onnxruntime/core/platform/windows/device_discovery.cc
```diff
@@ -79,7 +79,7 @@ std::unordered_map<uint64_t, DeviceInfo> GetDeviceInfoSetupApi(const std::unorde
   std::array<GUID, 3> guids = {
       GUID_DEVCLASS_DISPLAY,
       GUID_DEVCLASS_PROCESSOR,
-      GUID_DEVCLASS_SYSTEM,
+      GUID_DEVCLASS_COMPUTEACCELERATOR,
```
skottmckay, line 82: Is this GUID available on older versions of windows or do we need to do something similar to this to workaround that?

https://github.com/microsoft/onnxruntime/blob/0186596691ce072289abbcf6ff2cf67a164d136e/onnxruntime/core/platform/windows/device_discovery.cc#L76-L77

ashrit-ms: Fixed.

skottmckay (approved):

PR #24664, opened 2025-05-06
Fix improper vector iterator handling and a couple of warnings

yuslepukhin:
### Description

When erasing elements from a vector while iterating over it, the iterator must be updated to the new position after the erase operation as a return value, otherwise results in decrement of end operator which is undefined behavior.
Fix a couple of warnings as well.

### Motivation and Context
Causes test assert.


HectorSVC (dismissed):
:shipit:

chilo-ms (approved):

PR #24672, opened 2025-05-07
[onnxruntimeperftest] Add option to enable IO bindings on CUDA before session run

keshavv27:
### Description
Add option to enable tensor input and output bindings on CUDA before perftest inference session run.
Output binding is handled by changing the memory allocator type to CUDA.
Input binding is handled by creating default ORT tensor on CPU, initializing it with data, then cudaMemcpy the data from CPU to CUDA allocated GPU tensor using the raw pointers.

### Motivation and Context
By this change, the end-to-end inference time reported is more accurate as the CPU<->GPU overhead is moved out of the inference run


keshavv27:

keshavv27:
@chilo-ms @jywu-msft Pinging again for review, thankyou!

keshavv27:
Hi @tianleiwu, is there anything pending for this change?

PR #24675, opened 2025-05-07
[webgpu] Add zero points support for dp4 path

qjia7:
This PR adds the zero pints support for the DP4 path.


qjia7:
@sushraja-msft @guschmue  Please take a look, thanks.

onnxruntime/contrib_ops/webgpu/quantization/matmul_nbits_common.h
```diff
@@ -0,0 +1,68 @@
...
+ * Generates WebGPU shader code for reading zero points in quantized matrix multiplication
+ *
+ * @param nbits Number of bits for quantization (4 or 8)
+ * @param has_zero_points Whether zero points are provided as an input
+ * @param output_type Type name to use for zero point values in the generated code (default: "output_element_t")
+ * @return String containing the generated WebGPU shader code
+ */
+inline std::string GenerateZeroPointReadingCode(uint32_t nbits, bool has_zero_points,
+                                                const std::string& output_type = "output_element_t") {
+  ORT_ENFORCE(nbits == 8 || nbits == 4, "Only 4/8 bits are supported for webgpu matmulnbits");
```
sushraja-msft: nit: move the implementation to matmul_nbits_common.cc

onnxruntime/contrib_ops/webgpu/quantization/matmul_nbits.cc
```diff
@@ -416,25 +381,24 @@ Status MatMulNBits::ComputeInternal(onnxruntime::webgpu::ComputeContext& context
   const uint32_t components_a = GetMaxComponents(K);
   const uint32_t components_b = GetMaxComponents(blob_size_in_words);
   uint32_t components = GetMaxComponents(N);
+  // zero_points has shape[N * CeilDiv(n_blocks_per_col * bits, 8)]. So here we need to check whether n_blocks_per_col is divisible by 8/nbits.
+  // For bits==4, this is counted by elements of uint4. Need add 1 if not divisible by 2.
+  uint32_t zero_blocks_per_col = n_blocks_per_col % (8 / nbits) == 0 ? n_blocks_per_col : n_blocks_per_col + 1;
```
sushraja-msft, line 391: I think we only support zero_points that are uint8 type - because I see code like
mul_program.AddInput({zero_points, ProgramTensorMetadataDependency::None, {(zero_points->Shape().Size() + 3) / 4}, 4});
which I think is forcing the zero_points to be u32.

https://github.com/microsoft/onnxruntime/blob/main/docs/ContribOperators.md#inputs-3---6
indicates zero points can be tensor(float16), tensor(float), tensor(bfloat16)
Can we ORT_ENFORCE zero_points is uint8 ?

onnxruntime/contrib_ops/webgpu/quantization/dp4a_matmul_nbits.cc
```diff
@@ -248,6 +299,64 @@ Status DP4AMatMulNBitsProgram::GenerateShaderCode(ShaderHelper& shader) const {
             var own_a0: vec4<u32> = tile_A[0][base_A + a_idx];
             var own_a1: vec4<u32> = tile_A[1][base_A + a_idx];
             var own_scale_a: output_element_t = scale_A[base_A + a_idx];
+    )MAIN_FN";
+  if (nbits_ == 8 && has_zero_points_) {
```
sushraja-msft, line 303: I think you need to modify the nbits_ != 8 path below as well to support zero_points

onnxruntime/contrib_ops/webgpu/quantization/matmul_nbits.cc
qjia7: https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/contrib_ops/webgpu/quantization/matmul_nbits.cc#L70 ensures that we only support uint8 type for zero_points.
>Can we ORT_ENFORCE zero_points is uint8 ?

Done

onnxruntime/contrib_ops/webgpu/quantization/dp4a_matmul_nbits.cc
qjia7: `nbits_ == 8 && has_zero_points_` is the only case we need to do special process to avoid data overflow. For other situations,  directly subtracting zero is enough when de-quantize B data, which won't cause data overflow.

guschmue (approved):
lgtm.

PR #24678, opened 2025-05-07
[VitisAI] fix graph_save dump unsorted Model

danyue333:
### Description
When calling graph_save, a model with unsorted nodes will be saved, which will cause errors when using the model later.

### Motivation and Context
The resolve function is called in the graph_save process, which causes the nodes to be sorted.


danyue333:
@HectorSVC  can you take a look for this pr?

HectorSVC (approved):

PR #24682, opened 2025-05-07
Align hardware ID parsing with DXCore representation

ashrit-ms:
### Description
Modified the `get_id` lambda to parse 4-character hardware ID components (e.g., from VEN_xxxx or DEV_yyyy) by converting the ASCII string representation directly to a uint32_t using `WStringToUint32Id`.

This replaces the previous hexadecimal string-to-integer conversion and aligns with how DXCore reports these vendor and device IDs, ensuring consistent ID interpretation.

### Motivation and Context
Before this change, we were assuming the hardware ID components were hexadecimal strings and converting them to integers. This assumption is incorrect and was leading to incorrect interpretations of the vendor and device IDs.

Down the line when we compare the vendor and device id with the ids we receive from DXCORE, there are no matches, and we fail to copy the information collected in `GetDeviceInfoSetupApi`.

### Testing
Tested with all the sample apps and stepped through the code to verify the Vendor and Device IDs match the IDs from DXCORE on both Qualcomm and AMD.

### DIFF
--- a/onnxruntime/core/platform/windows/device_discovery.cc
+++ b/onnxruntime/core/platform/windows/device_discovery.cc
@@ -120,7 +120,7 @@ std::unordered_map<uint64_t, DeviceInfo> GetDeviceInfoSetupApi(const std::unorde
                                             (PBYTE)buffer, sizeof(buffer), &size)) {
         uint32_t vendor_id = 0;
         uint32_t device_id = 0;
-                                      
+
         // PCI\VEN_xxxx&DEV_yyyy&...
         // ACPI\VEN_xxxx&DEV_yyyy&... if we're lucky.
         // ACPI values seem to be very inconsistent, so we check fairly carefully and always require a device id.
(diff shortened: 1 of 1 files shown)


onnxruntime/core/platform/windows/device_discovery.cc
```diff
@@ -110,8 +124,9 @@ std::unordered_map<uint64_t, DeviceInfo> GetDeviceInfoSetupApi(const std::unorde
         const auto get_id = [](const std::wstring& hardware_id, const std::wstring& prefix) -> uint32_t {
           if (auto idx = hardware_id.find(prefix); idx != std::wstring::npos) {
             auto id = hardware_id.substr(idx + prefix.size(), 4);
-            if (std::all_of(id.begin(), id.end(), iswxdigit)) {
-              return std::stoul(id, nullptr, 16);
+            if (id.size() == 4) {
+              // DXCore reports vendor and device IDs as 32-bit integer representations of the ASCII string.
+              return WStringToUint32Id(id);
```
skottmckay, line 129: Should this be done for all vendor ids where id size is 4 or limited to when we're checking GUID_DEVCLASS_DISPLAY?

ashrit-ms: Good catch, fixed it only use the new method for DISPLAY and COMPUTE ACCELERATOR class.

skottmckay (approved):

PR #24689, opened 2025-05-08
[NvTensorRTRTX EP] Enable automatic selection of NvTensorRTRTX EP for PREFER_GPU policy

nieubank:
### Description
Enables automatic selection of NvTensorRTRTX EP for PREFER_GPU policy similar to #24629

Sample code from unit test:
```c++
  auto env = Ort::Env();

  env.RegisterExecutionProviderLibrary(kNvTensorRTRTXExecutionProvider, ORT_TSTR("onnxruntime_providers_nv_tensorrt_rtx.dll"));

  Ort::SessionOptions so;
  so.SetEpSelectionPolicy(OrtExecutionProviderDevicePolicy_PREFER_GPU);
  Ort::Session session_object(env, model_name_ctx.c_str(), so);
  EXPECT_TRUE(SessionHasEp(session_object, kNvTensorRTRTXExecutionProvider));

  env.UnregisterExecutionProviderLibrary(kNvTensorRTRTXExecutionProvider);
```

### Motivation and Context
A recent feature allows ORT to automatically select an EP according to policies set by the user (e.g., prefer npu or prefer gpu). This PR allows NvTensorRTRTX EP to be potentially selected when the user sets the PREFER_GPU policy.

### DIFF
--- a/onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc
+++ b/onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc
@@ -114,6 +114,27 @@ struct Nv_Provider : Provider {
     return std::make_shared<NvProviderFactory>(info);
   }
 
+  Status CreateIExecutionProvider(const OrtHardwareDevice* const* /*devices*/,
+                                  const OrtKeyValuePairs* const* /*ep_metadata*/,
+                                  size_t num_devices,
+                                  ProviderOptions& provider_options,
+                                  const OrtSessionOptions& session_options,
+                                  const OrtLogger& logger,
+                                  std::unique_ptr<IExecutionProvider>& ep) override {
+    if (num_devices != 1) {
+      return Status(common::ONNXRUNTIME, ORT_EP_FAIL, "[NvTensorRTRTX EP]  only supports one device.");
+    }
+
+    const ConfigOptions* config_options = &session_options.GetConfigOptions();
+
+    std::array<const void*, 2> configs_array = {&provider_options, config_options};
+    const void* arg = reinterpret_cast<const void*>(&configs_array);
+    auto ep_factory = CreateExecutionProviderFactory(arg);
+    ep = ep_factory->CreateProvider(session_options, logger);
+
+    return Status::OK();
+  }
+
   void Initialize() override {
     InitializeRegistry();
   }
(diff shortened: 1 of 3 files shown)


onnxruntime/test/providers/nv_tensorrt_rtx/nv_basic_test.cc
```diff
@@ -331,5 +332,48 @@ TEST(NvExecutionProviderTest, ContextEmbedAndReloadDataDynamic) {
...
+
+  {
+    env.RegisterExecutionProviderLibrary(kNvTensorRTRTXExecutionProvider, ORT_TSTR("onnxruntime_providers_nv_tensorrt_rtx.dll"));
+
+    Ort::SessionOptions so;
+    so.SetEpSelectionPolicy(OrtExecutionProviderDevicePolicy_PREFER_GPU);
+    Ort::Session session_object(env, model_name_ctx.c_str(), so);
+    EXPECT_TRUE(SessionHasEp(session_object, kNvTensorRTRTXExecutionProvider));
+
+    env.UnregisterExecutionProviderLibrary(kNvTensorRTRTXExecutionProvider);
```
adrianlizarraga, line 374: Hi, we shouldn't unregister an EP library while the Session is still alive. Refer to: https://github.com/microsoft/onnxruntime/blob/main/include/onnxruntime/core/session/onnxruntime_c_api.h#L5077-L5087

Can you please move the Unregister call outside of the scope?

onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc
```diff
@@ -132,3 +153,113 @@ ORT_API(onnxruntime::Provider*, GetProvider) {
...
+
+  static void ReleaseEpImpl(OrtEpFactory* /*this_ptr*/, OrtEp* /*ep*/) {
+    // no-op as we never create an EP here.
+  }
+
+  const OrtApi& ort_api;
+  const std::string ep_name;
+  const std::string vendor{"NVIDIA"};
+
+  // NVIDIA vendor ID. Refer to the ACPI ID registry (search NVIDIA): https://uefi.org/ACPI_ID_List
```
adrianlizarraga, line 232: Does this vendor ID come from the ACPI ID registry?

nieubank: > Does this vendor ID come from the ACPI ID registry?

https://uefi.org/ACPI_ID_List?acpi_search=nvidia
Nvidia | NVDA | 01/01/2010

The value is well known so I used it directly.

adrianlizarraga: The reason I ask is because `0x10de` doesn't map to the bytes `'NVIDIA'`. For QNN, the characters `'QCOM'` from ACPI ID registry mapped directly to the vendor id value.

adrianlizarraga (approved):

nieubank:
@adrianlizarraga - I've updated the PR on the latest main with the change we took for Windows ML to fix multi-device path. When you get a chance, please review and let me know if there is anything else that needs to be done to unblock merging. Thanks!

PR #24695, opened 2025-05-08
[Compile] Add API to set compile flags; Fix behavior when no nodes are compiled

adrianlizarraga:
#### Original compile approach where an EPContext model is generated as a side-effect of creating a session:
- **Restore** previous behavior where:
  - compiling a model that generates no EPContext nodes is silently ignored (nothing is generated and no error is reported) 
  - compiling a previously compiled model is silently ignored (nothing is generated and no error is reported)

#### Explicit compile API:
- **Retains** current behavior where compiling a model that does not generate EPContext nodes still generates a model by default.
- Adds C/C++/C#/Python API called `setFlags` that allows the user to specify what is considered an error.
  - `OrtCompileApiFlags_ERROR_IF_NO_NODES_COMPILED`: CompileModel() returns `ORT_FAIL` if no nodes were compiled.
  - `OrtCompileApiFlags_ERROR_IF_OUTPUT_FILE_EXISTS`: CompileModel() returns `ORT_FAIL` if a file with the same filename as the output model exists.
- Adds logic to detect when the user is trying to compile a previously compiled model and returns an `ORT_INVALID_GRAPH` error with a relevant error message.

### Motivation and Context
A previous [PR changed the default behavior](https://github.com/microsoft/onnxruntime/commit/b4f7a905b0d636b71bd486c0ef702eb5a44eadf2#diff-e2d3910ae7593ee7ba4fd74e53f738fa973ae2fc32c069f1088ba458b91f8d4bL809) of the original "implicit" compilation approach. This PR was motivated by restoring the original behavior that users currently depend on. At the same time, we want to allow users of the new explicit compile API to determine what is considered an error.


include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -5782,6 +5782,21 @@ struct OrtModelEditorApi {
...
 
+/** \brief Flags representing options to enable when compiling a model.
+ */
+typedef enum OrtCompileApiFlags {
+  // Default. Do not enable any additional compilation options.
+  OrtCompileApiFlags_NONE = 0,
+
+  // Force compilation to return an error (ORT_FAIL) if no nodes were compiled.
+  // Otherwise, a model with basic optimizations (ORT_ENABLE_BASIC) is still generated by default.
+  OrtCompileApiFlags_ERROR_IF_NO_NODES_COMPILED = 1 << 0,
```
skottmckay, line 5793: Is there a use case that requires an error?

skottmckay (approved):

PR #24699, opened 2025-05-08
enable using vcpkg for Dawn

fs-eire:
### Description

Enable vcpkg for webgpu


tools/python/util/vcpkg_helpers.py
```diff
@@ -78,6 +78,13 @@ def add_port_configs(f, has_exception: bool, is_emscripten: bool, enable_minimal
 endif() # End ONNX-specific block
 """)
 
+    # Add Dawn specific flags
+    f.write(r"""
+if(PORT STREQUAL "dawn")
```
fs-eire, line 83: this should be `STREQUAL` instead of `MATCH` for similar lines above?

tools/ci_build/build.py
```diff
@@ -1228,7 +1228,7 @@ def generate_build_tree(
             ]
         env = {}
         if args.use_vcpkg:
-            env["VCPKG_KEEP_ENV_VARS"] = "TRT_UPLOAD_AUTH_TOKEN;EMSDK;EMSDK_NODE;EMSDK_PYTHON"
+            env["VCPKG_KEEP_ENV_VARS"] = "TRT_UPLOAD_AUTH_TOKEN;EMSDK;EMSDK_NODE;EMSDK_PYTHON;PATH"
```
fs-eire, line 1231: if not adding `PATH` here, it seems vcpkg failed to find python 3 on my machine. not sure why.

cmake/vcpkg-ports/dawn/portfile.cmake
```diff
@@ -0,0 +1,132 @@
...
+
+    # We are currently always using the D3D12 backend.
+    list(APPEND onnxruntime_vcpkg_DAWN_OPTIONS
+        -DDAWN_ENABLE_D3D11=OFF
+    )
+endif()
+
+vcpkg_from_github(
+    OUT_SOURCE_PATH SOURCE_PATH
+    REPO fs-eire/dawn
```
fs-eire, line 115: will update to official dawn before merging.

currently using my fork for easier iteration.

cmake/vcpkg-ports/dawn/portfile.cmake
```diff
@@ -0,0 +1,132 @@
...
+    -DTINT_BUILD_IR_BINARY=OFF
+    -DTINT_BUILD_SPV_READER=OFF
+    -DTINT_BUILD_WGSL_WRITER=ON
+
+    -DDAWN_ENABLE_SPIRV_VALIDATION=OFF
+
+    # explicitly set the jinja2 and markupsafe directories to empty strings
+    # when they are empty, the python script will import them from the system
+    #
+    # pip install jinja2 markupsafe
```
fs-eire, line 63: how to manage these 2 dependencies?

we can use the `pip install` on the CI machine, but it's easy to fail on a local build.

fs-eire:
The Java binding is failing on macOS. @Craigacp could you please help to take a look at it?

I need to add build flag `--cmake_extra_defines onnxruntime_BUILD_DAWN_MONOLITHIC_LIBRARY=ON` to the build, which adds a new dynamic library `libwebgpu_dawn.dylib`. While the Node.js binding tests passed, the Java binding tests failed because of cannot find "@rpath/libwebgpu_dawn.dylib". The error is still not fixed after applying commit https://github.com/microsoft/onnxruntime/pull/24699/commits/9125736d9291f2b7f73c64d61cbdc57629094359 and https://github.com/microsoft/onnxruntime/pull/24699/commits/ad12b0e3d3c28c235e95cbb82f6c574b701e756a

To reproduce the issue locally, use the following build command line:
```
./build.sh --parallel --use_binskim_compliant_compile_flags --build_shared_lib --build_nodejs --build_java --use_webgpu --cmake_extra_defines onnxruntime_BUILD_DAWN_MONOLITHIC_LIBRARY=ON --use_vcpkg --config Release
```

Craigacp:
I'm a bit confused about what the context is here. Why would the Java binding be available in a context where webgpu was being used?

fs-eire:
> I'm a bit confused about what the context is here. Why would the Java binding be available in a context where webgpu was being used?

The name of "WebGPU" is confusing. Let me try to explain it a little bit.

WebGPU is a web standard initially, which is designed for web (browser environment). However, Dawn, the native implementation of WebGPU can be integrated to native so that we can leverage the GPU capabilities in non-browser environment too. In this context, "WebGPU" means the WebGPU EP, which uses the native port of webgpu (webgpu.h and the native implementation). So, since it is native, it should work with language bindings including Java binding.

In the build, dawn will be compiled into a dynamic library (libwebgpu_dawn.dylib on macOS) and WebGPU EP will load it at runtime.

Craigacp:
Oh ok. I'll take a look.

Craigacp:
Ok, if libwebgpu_dawn.dylib is supposed to be packaged inside the jar file (which it currently is doing at `/ai/onnxruntime/native/osx-aarch64/libwebgpu_dawn.dylib`) then it'll need to be unpacked the way the shared provider lib is unpacked (https://github.com/microsoft/onnxruntime/blob/main/java/src/main/java/ai/onnxruntime/OnnxRuntime.java#L163). Otherwise it's not in the same folder as the location we load `libonnxruntime.dylib` and `libonnxruntime4j_jni.dylib` from so the linker can't find it. Presumably this would want to be conditional in the way the unpacking of the shared provider lib is, so another call to `extractProviderLibrary` with the right path should work.

Craigacp:
This diff was sufficient to make the Java tests pass:

```
diff --git a/java/src/main/java/ai/onnxruntime/OnnxRuntime.java b/java/src/main/java/ai/onnxruntime/OnnxRuntime.java
index c28c79f1e7..4de3ee6338 100644
--- a/java/src/main/java/ai/onnxruntime/OnnxRuntime.java
+++ b/java/src/main/java/ai/onnxruntime/OnnxRuntime.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019, 2023, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2019, 2025, Oracle and/or its affiliates. All rights reserved.
  * Licensed under the MIT License.
  */
 package ai.onnxruntime;
@@ -79,6 +79,9 @@ final class OnnxRuntime {
   /** The short name of the ONNX runtime QNN provider library */
   static final String ONNXRUNTIME_LIBRARY_QNN_NAME = "onnxruntime_providers_qnn";
 
+  /** The short name of the ONNX runtime DAWN library */
+  static final String ONNXRUNTIME_LIBRARY_DAWN = "webgpu_dawn";
+
   /** The OS & CPU architecture string */
   private static final String OS_ARCH_STR = initOsArch();
 
@@ -161,6 +164,8 @@ final class OnnxRuntime {
       // Extract and prepare the shared provider library but don't try to load it,
       // the ONNX Runtime native library will load it
       extractProviderLibrary(ONNXRUNTIME_LIBRARY_SHARED_NAME);
+      // Extract the DAWN WebGPU implementation library and load it if it exists.
+      load(ONNXRUNTIME_LIBRARY_DAWN);
 
       if (!isAndroid()) {
         load(ONNXRUNTIME_LIBRARY_NAME);
```

I've not checked to see if that causes problems when Dawn isn't present. We typically have a little more conditionality in the loading to avoid causing trouble (e.g. turning it off when on Android), but I'm not really sure when it's going to be packaged in, or if we want to let users override the location. If you can tell me a little more about it then I can try to figure out what guards we need.

fs-eire:
@Craigacp thank you for the explanation! The CI failure shows that `load(ONNXRUNTIME_LIBRARY_DAWN)` will fail if cannot find the file.

Is there a way to create a boolean value depending on build flags in Java? Specifically, the following C++ code does it:
```c++
#if defined(BUILD_DAWN_MONOLITHIC_LIBRARY)
constexpr bool SHOULD_LOAD_DAWN_DYNAMIC_LIB = true;
#else
constexpr bool SHOULD_LOAD_DAWN_DYNAMIC_LIB = false;
#endif
```

Craigacp:
Unfortunately Java doesn't have conditional compilation, so we'll need another signalling mechanism. Modifying the load method to make it allow optional values so we can load it if it's present is probably the best way.

snnn:
The "webgpu_build_x64_RelWithDebInfo" pipeline did not run.

fs-eire:
> webgpu_build_x64_RelWithDebInfo

"webgpu_build_x64_RelWithDebInfo" is not becoming 2 runs:
- webgpu_build_x64_RelWithDebInfo (vcpkg)
- webgpu_build_x64_RelWithDebInfo (novcpkg)

snnn (approved):
Really thanks!

PR #24700, opened 2025-05-09
[web CI] allow shader key validation failure

fs-eire:
### Description

To unblock the pipeline. 

This should be reverted once the root cause is identified and fixed.


jywu-msft (approved):

fs-eire:
close in favor of #24701

Issue #24680, opened 2025-05-07
[Feature Request] Expose tensor data size through the public C API


huningxin:
### Describe the feature request

Allows user code to get the data size of a tensor, e.g. expose [`Tensor::SizeInBytes()`](https://github.com/microsoft/onnxruntime/blob/main/include/onnxruntime/core/framework/tensor.h#L301), through the public C API.

e.g. Introduce a dedicated API
```
ORT_API2_STATUS(GetTensorSizeInBytes, _In_ OrtValue* value, _Out_ size_t* out)
```

Or extend `GetTensorMutableData` that can return its size in bytes
```
ORT_API2_STATUS(GetTensorMutableData, _In_ OrtValue* value, _Outptr_ void** buffer, _Out_ size_t* size_in_bytes);
```

### Describe scenario use case

Applications, such as [WebNN in Chromium](https://chromium-review.googlesource.com/c/chromium/src/+/6469898/comment/9136abc5_b77062f8/), may need to get the buffer of an initializer by `GetTensorMutableData()` and copies the data into the buffer. It would be important to ensure the data length is equal to the tensor size before coping that avoids the out-of-bounds write. Without the API of getting bytes size of a tensor, the user code needs to calculate it by itself which may not be the same as the implementation of ONNX Runtime.

/cc @skottmckay @guschmue @fdwr

cbourjau:
> Without the API of getting bytes size of a tensor, the user code needs to calculate it by itself which may not be the same as the implementation of ONNX Runtime.

Could you name a type where the size wouldn't be known based on the element type? Could this issue also be fixed by onnxruntime explicitly documenting the size of the types rather than extending the C-API?

fdwr:
Huh, that's very surprising that ORT has an API to get the raw data...

```c++
  /** \brief Get a pointer to the raw data inside a tensor
   *
   * Used to read/write/modify the internal tensor data directly.
   * \note The returned pointer is valid until the \p value is destroyed.
   *
   * \param[in] value A tensor type (string tensors are not supported)
   * \param[out] out Filled in with a pointer to the internal storage
   *
   * \snippet{doc} snippets.dox OrtStatus Return Value
   */
  ORT_API2_STATUS(GetTensorMutableData, _In_ OrtValue* value, _Outptr_ void** out);
```

...but evidently not a way to get the byte size of that raw data buffer. 😶 Am I just missing something?

[...]

PR #24728, opened 2025-05-12
Expose and test GetTensorSizeInBytes in C# and Python

yuslepukhin:
### Description
Expose and test GetTensorSizeInBytes in C# and Python

### Motivation and Context

https://github.com/microsoft/onnxruntime/issues/24680

### DIFF
--- a/onnxruntime/python/onnxruntime_inference_collection.py
+++ b/onnxruntime/python/onnxruntime_inference_collection.py
@@ -1006,6 +1006,13 @@ class OrtValue:
         """
         return self._ortvalue.element_type()
 
+    def tensor_size_in_bytes(self) -> int:
+        """
+        Returns the size of the data in the OrtValue in bytes
+        if the OrtValue is a tensor.
+        """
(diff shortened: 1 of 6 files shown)


tianleiwu (approved):

onnxruntime/python/onnxruntime_inference_collection.py
```diff
@@ -1006,6 +1006,13 @@ def element_type(self) -> int:
         """
         return self._ortvalue.element_type()
 
+    def tensor_size_in_bytes(self) -> int:
+        """
+        Returns the size of the data in the OrtValue in bytes
+        if the OrtValue is a tensor.
```
tianleiwu, line 1012: what's the return value if it is not a tensor?

yuslepukhin:
> Can GetTensorSizeInBytes also be exposed through C/C++ API? @yuslepukhin , WebNN uses ORT C API and needs this API. Thanks!
> 
> /cc @fdwr @skottmckay

Do you think it is possible to have in C# and not to have in C? Yes, it is both in C And C++.

PR #24729, opened 2025-05-12
[QNN EP] Fix 16x16 Conv translation

quic-tirupath:
- QNN's 16x16 Conv doesn't support asymmetric int16 weight
 - Insert Convert Op to convert from asymmetric uint16 weight to symmetric int16 weight

### Description
- QNN' Conv op doesn't support asymmetric INT16 weights.
- 16x16 Conv operators in ONNX models fallback to CPU execution provider and reporting higher inference times.
-  Insert a Convert Op to convert asymmetric uint16 weight to symmetric int16 weight to schedule 16x16 Conv's on QNN EP provider.

### Motivation and Context
- This fixes Graph execution failures for models contain 16x16 Conv op on QNN Execution provider
- This also improves Inference times of model contain 16x16 Conv op


HectorSVC (approved):
:shipit:

PR #24732, opened 2025-05-13
[x86] matmul8bit memory loading perf tuning

fajin-corp:
### Description
Use aligned load and preloading. There is ~10% token generation speed up.

### Motivation and Context
Optimize perf


liqunfu (approved):

hanbitmyths (approved):

yufenglee:
@snnn already reviewed it.

PR #24735, opened 2025-05-13
[QNN EP] Add Support for MaxPool rank-3 input QNN EP

quic-ashwshan:
- Currently QNN EP only supports MaxPool for rank 4.
- This change adds support for rank 3 input by adding Reshapes before and after the Op to ensure that the MaxPool gets input rank 4.
- Updated all attributes if converting rank 3 input to rank 4 by updating stride, pads, dilations and kernel size.
- Added unit tests which takes input rank 3 to validate MaxPool on NPU.

### Description
This change extends the support of QNN EP's MaxPool operation to handle input tensors of rank 3. To achieve this, Reshape Ops are added before and after the MaxPool Op to ensure that the input to MaxPool is always of rank 4, as required. Additionally, the attributes such as stride, pads, dilations, and kernel size are updated accordingly to accommodate the conversion of rank 3 inputs to rank 4. Unit tests have been added to validate the functionality of MaxPool on NPU with rank 3 inputs.

### Motivation and Context
This change is required to enhance the flexibility and usability of QNN EP's MaxPool operation by supporting a broader range of input tensor ranks. Previously, the operation was limited to only supporting rank 4 inputs, which restricted the support in certain scenarios. By adding support for rank 3 inputs, this change solves the problem of limited compatibility and enhances the overall functionality and makes sure that the MaxPool op offloads to the NPU (QNN HTP Backend)

### DIFF
--- a/onnxruntime/test/providers/qnn/pool_op_test.cpp
+++ b/onnxruntime/test/providers/qnn/pool_op_test.cpp
@@ -188,6 +188,38 @@ TEST_F(QnnHTPBackendTests, MaxPool_Large_Input_HTP_u8) {
                             QDQTolerance(0.00417f));
 }
 
+// 1-D MaxPool HTP test for rank-3 without ceil
+TEST_F(QnnHTPBackendTests, MaxPool_Rank3_HTP_u8) {
+  RunQDQPoolOpTest<uint8_t>(
+      "MaxPool",
+      TestInputDef<float>({1, 3, 3}, false, -10.0f, 10.0f),
+      // A single 1-D kernel of length 3
+      {utils::MakeAttribute("kernel_shape", std::vector<int64_t>{3}),
+       utils::MakeAttribute("strides", std::vector<int64_t>{3}),
+       // 1-D pad: only two values
+       utils::MakeAttribute("pads", std::vector<int64_t>{0, 0}),
+       utils::MakeAttribute("dilations", std::vector<int64_t>{1}),
+       utils::MakeAttribute("ceil_mode", static_cast<int64_t>(0)),
+       utils::MakeAttribute("storage_order", static_cast<int64_t>(0)),
+       utils::MakeAttribute("auto_pad", "NOTSET")},
+      ExpectedEPNodeAssignment::All);
+}
+
+// 1-D MaxPool HTP test for rank-3 with ceil_mode=1
+TEST_F(QnnHTPBackendTests, MaxPool_Rank3_Ceil_HTP_u8) {
+  RunQDQPoolOpTest<uint8_t>(
+      "MaxPool",
+      TestInputDef<float>({1, 3, 3}, false, -10.0f, 10.0f),
+      {utils::MakeAttribute("kernel_shape", std::vector<int64_t>{3}),
+       utils::MakeAttribute("strides", std::vector<int64_t>{3}),
+       utils::MakeAttribute("pads", std::vector<int64_t>{0, 0}),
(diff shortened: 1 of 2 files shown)


onnxruntime/test/providers/qnn/pool_op_test.cpp
```diff
@@ -188,6 +188,38 @@ TEST_F(QnnHTPBackendTests, MaxPool_Large_Input_HTP_u8) {
...
+      "MaxPool",
+      TestInputDef<float>({1, 3, 3}, false, -10.0f, 10.0f),
+      {utils::MakeAttribute("kernel_shape", std::vector<int64_t>{3}),
+       utils::MakeAttribute("strides", std::vector<int64_t>{3}),
+       utils::MakeAttribute("pads", std::vector<int64_t>{0, 0}),
+       utils::MakeAttribute("dilations", std::vector<int64_t>{1}),
+       utils::MakeAttribute("ceil_mode", static_cast<int64_t>(1)),
+       utils::MakeAttribute("storage_order", static_cast<int64_t>(0)),
+       utils::MakeAttribute("auto_pad", "NOTSET")},
+      ExpectedEPNodeAssignment::All);
```
yuslepukhin, line 220: >ExpectedEPNodeAssignment::All); [](http://example.com/codeflow?start=6&length=31)

We can test that the added Reshape nodes are present. InferenceSessionWrapper class enables us to obtain the graph and there is a utility to count the nodes.

quic-ashwshan: To obtain the graph from the InferenceSessionWrapper directly we would need to create a model and then initialize it and then get the graph from the wrapper? Since the current test does not have any helper functions to grab the graph directly using the InferenceSessionWrapper.

quic-ashwshan: Tried using the InferenceSessionWrapper to obtain the graph and the utility - NumberOfNodes(). But this only gives out the number of nodes to be 1 since at the end after initialize() the QNN model just remains with one fused node with the name QNN_(numbers).

yuslepukhin: Here is a good [example ](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/test/contrib_ops/dynamic_quantize_matmul_test.cc#L258)of doing it.

quic-ashwshan: Hi, even if we use CountOpsInGraph() it would only output 1 as the count since the final graph that we get after calling 
session.Initialize() will only have one node which is essentially fused and with the name QNN_128.. 

`auto op_to_count = CountOpsInGraph(session.GetGraph());
for (const auto& kv : op_to_count) {
  LOGS_DEFAULT(WARNING) << "Key : " << kv.first << "Value: " << kv.second; }`
output - 
Key : QNN.QNN_12862626549331715565_0Value: 1

The reason why this works in the above example is because the test is still in the ONNX world after optimizations.

yuslepukhin: See if disabling optimizations would help. If not, add commentary as above and test for one node.

quic-ashwshan: Disabling optimizations did not work. Added test to verify final output node to be a fused QNN node.

yuslepukhin (approved):
:shipit:

HectorSVC (approved):
:shipit:

PR #24740, opened 2025-05-13
Compile API: support for OrtModel input and write output to stream

adrianlizarraga:
### Description
- Adds API function to compile an `OrtModel` instance that was created with the model editor API.
  - C and C++ support.
  - No Python bindings yet until we first add the model editor API to the Python bindings.
- Adds API function to write the compiled model to a user-provided output stream.
  - C, C++, and Python support

### Examples
More examples can be found in the unit tests.

#### C++
Compiles an `OrtModel` to a stream that itself writes to a file:
```c++
static OrtStatus* ORT_API_CALL MyWrite(void* stream_state, const void* buffer, size_t buffer_num_bytes) {
  std::ofstream* outfile = reinterpret_cast<std::ofstream*>(stream_state);
  outfile->write(reinterpret_cast<const char*>(buffer), buffer_num_bytes);
  return nullptr;  // No error.
}

int main(int argc, char** argv) {
  // Not shown: build OrtModel using editor API ...
  Ort::Model model(opsets);
  model.AddGraph(graph);

  Ort::SessionOptions so;
  so.AppendExecutionProvider("QNN", QnnHTPOptionsWithoutQDQOffloading());

  const ORTCHAR_T* output_model_file = ORT_TSTR("compileapi_ortmodel_ctx.onnx");
  std::ofstream outfile(output_model_file, std::ios::binary);

  // Create model compilation options that load an OrtModel and write the compiled model bytes to MyWrite function
  Ort::ModelCompilationOptions compile_options(*ort_env, so);
  compile_options.SetInputModel(model.GetConst());
  compile_options.SetOutputModelOutStream(MyWrite, reinterpret_cast<void*>(&outfile));  // Set output stream
  compile_options.SetEpContextEmbedMode(true);

  // Compile the model.
  Ort::Status status = Ort::CompileModel(*ort_env, compile_options);
  assert(status.IsOk());
  outfile.close();

  assert(std::filesystem::exists(output_model_file));  // assert model was created.
}
```

### Motivation and Context
Provide flexibility to users of the Compile API that, like those compiling for webnn, that want to open/write files outside of ORT.

Next PR: allow loading input model from an existing file handle.


onnxruntime/core/session/compile_api.cc
```diff
@@ -106,6 +107,27 @@ ORT_API_STATUS_IMPL(OrtCompileAPI::ModelCompilationOptions_SetInputModelFromBuff
...
+
+  if (input_model == nullptr) {
+    return OrtApis::CreateStatus(ORT_INVALID_ARGUMENT, "Invalid input model: OrtModel pointer is null");
+  }
+
+  model_compile_options->SetInputOrtModel(*input_model);
+  return nullptr;
+#else
+  ORT_UNUSED_PARAMETER(ort_model_compile_options);
+  ORT_UNUSED_PARAMETER(input_model_path);
```
adrianlizarraga, line 125:
```suggestion
  ORT_UNUSED_PARAMETER(input_model);
```

onnxruntime/core/framework/tensorprotoutils.cc
```diff
@@ -179,11 +179,17 @@ Status ReadExternalDataForTensor(const ONNX_NAMESPACE::TensorProto& tensor_proto
...
+    // the value in location is the memory address of the data
+    const void* ext_data_buf = reinterpret_cast<const void*>(file_offset);
+    std::memcpy(unpacked_tensor.data(), ext_data_buf, tensor_byte_size);
+  } else {
+    ORT_RETURN_IF_ERROR(onnxruntime::Env::Default().ReadFileIntoBuffer(
+        external_file_path.c_str(),
+        file_offset,
+        tensor_byte_size,
+        gsl::make_span(reinterpret_cast<char*>(unpacked_tensor.data()), tensor_byte_size)));
+  }
```
adrianlizarraga, line 192: Note: fixes incomplete handling of external data that is stored in memory with the tag `kTensorProtoMemoryAddressTag`. I ran into this when serializing an OrtModel (created via editor api) to a model proto. There's a test in this PR that triggers this scenario.

yuslepukhin: This is re-worked in my OrtValue initializers PR

onnxruntime/core/framework/graph_partitioner.cc
```diff
@@ -894,25 +896,41 @@ static Status CreateEpContextModel(const ExecutionProviders& execution_providers
...
+    *output_buffer_holder->buffer_ptr = buffer.release();
+  } else if (output_stream_holder != nullptr) {
+    // Write output model to user's output stream.
+    ORT_RETURN_IF_ERROR(ep_context_model.MainGraph().Resolve());
+    ONNX_NAMESPACE::ModelProto model_proto = ep_context_model.ToGraphProtoWithExternalInitializers(external_ini_path,
+                                                                                                   valid_output_model_path,
+                                                                                                   model_saving_options);
+    size_t buffer_size = model_proto.ByteSizeLong();
+    ORT_RETURN_IF(buffer_size > static_cast<size_t>(std::numeric_limits<int>::max()),
+                  "Cannot serialize ONNX ModelProto larger than 2GB");
```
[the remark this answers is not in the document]
adrianlizarraga: I'm not sure. As far as I'm aware, onnx (protobuf) does not support files larger than 2gb, so weights would have to be external for large models.

gedoensmax:
@adrianlizarraga am I right in the assumption that this will fail for a model that has more than 2GB ? I would assume that the resulting photo is serialized before entering this custom function. It might be crucial to handle the proto and any external initializers separately.

onnxruntime/core/framework/graph_partitioner.cc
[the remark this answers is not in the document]
huningxin: Can the user also pass an output stream for external weights file?

PR #24743, opened 2025-05-13
[NvTensorRT RTX] Add Bfloat16

gedoensmax:
### Description

TRT supports Bfloat 16 and ORT does as well. 
In addition the `setup.py` was missing a copy for NVTRT EP and TRT EP can only be built against the packaged parser with TRT RTX.


gedoensmax:
I noticed some other things on the path for CUDA device bindings when ORT is compiled without CUDA EP and just with the CUDA EP interface enabled. I will convert this to a draft and finish up tomorrow.

gedoensmax:
We will resort to relying on CUDA EP for device bindings for the time being.

gedoensmax:
@chilo-ms can you help review this ?

chilo-ms:
Please also help add bf16 in python binding.
similar to https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/python/onnxruntime_pybind_state.cc#L619

gedoensmax:
@chilo-ms do you mind if I just delete the entire thing and replace it with https://github.com/microsoft/onnxruntime/blob/2a09f271815b15e414d6be5a13557927ce828394/onnxruntime/test/perftest/ort_test_session.cc#L151-L183 ? 
Is the type checking and explicit warning really necessary at the pybind level ?

onnxruntime/test/providers/nv_tensorrt_rtx/nv_basic_test.cc
```diff
@@ -1,4 +1,5 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
yuslepukhin, line 2: Need to make sure that there test coverage for BFloat16 type.

chilo-ms:
> @chilo-ms do you mind if I just delete the entire thing and replace it with
> 
> https://github.com/microsoft/onnxruntime/blob/2a09f271815b15e414d6be5a13557927ce828394/onnxruntime/test/perftest/ort_test_session.cc#L151-L183
> 
> ?
> Is the type checking and explicit warning really necessary at the pybind level ?

Agree that type checking and explicit warning are not necessary.
Your suggestion needs to add the ORT C API dependencies that is not added in this pybind file and need some efforts to test.
Another option is to refactor the code with what CUDA EP [did](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/python/onnxruntime_pybind_state.cc#L1063), but that requires TRT EP to expose some functions like `TensorRTExecutionProviderInfo__FromProviderOptions`.

Either way, i think we can do this in another PR?

gedoensmax:
@chilo-ms changes are done. Let me know if there is something else for this API.

chilo-ms (approved):

onnxruntime/test/providers/nv_tensorrt_rtx/nv_basic_test.cc
```diff
@@ -24,6 +25,35 @@ namespace onnxruntime {
...
+  std::string getTypeAsName() {
+    std::string dtype_name = "";
+    if constexpr (std::is_same<T, double>::value) {
+      dtype_name = "fp64";
+    } else if constexpr (std::is_same<T, float>::value) {
+      dtype_name = "fp32";
+    } else if constexpr (std::is_same<T, BFloat16>::value) {
+      dtype_name = "fp16";
+    } else if constexpr (std::is_same<T, MLFloat16>::value) {
+      dtype_name = "bf16";
```
joeyearsley, line 40: Was this missed in the original review, it seems we are setting the BFloat16 type to fp16 rather than bf16 and vice-versa?

gedoensmax: Yes it seems so. If you want to use bfloat16 I would recommend using a strongly typed ONNX rather than using this global flag.

PR #24757, opened 2025-05-14
Skipping newly added MHA test as this isn't supported by ROCm EP like…

TedThemistokleous:
… other tests (#118)

### Description

Skip this test for now for ROCm EP as we're having spotty support for MHA due to older CK repo.

### Motivation and Context

Breaks in testing for now. Add skip. This test was added after we added the skip


TedThemistokleous:
ping @tianleiwu , this is just a one line commit

tianleiwu (approved):

PR #24772, opened 2025-05-15
SkipSimplifiedLayerNorm + QuickGelu bfloat16 CUDA implementation

nenad1002:
### Description

SkipSimplifiedLayerNorm + QuickGelu bfloat16 CUDA implementation #24772


onnxruntime/contrib_ops/cuda/bert/skip_layer_norm.cc
```diff
@@ -106,19 +107,35 @@ Status SkipLayerNorm<T, Simplified>::ComputeInternal(OpKernelContext* ctx) const
...
-        reinterpret_cast<const CudaT*>(skip->Data<T>()),
-        (bias != nullptr) ? reinterpret_cast<const CudaT*>(bias->Data<T>()) : nullptr,
-        reinterpret_cast<const CudaT*>(gamma->Data<T>()),
-        (beta != nullptr) ? reinterpret_cast<const CudaT*>(beta->Data<T>()) : nullptr,
-        epsilon_,
-        hidden_size,
-        row_count,
-        skip_size);
+    if (std::is_same<T, BFloat16>::value) {
+      LaunchSkipLayerNormKernel<nv_bfloat16, Simplified>(
```
tianleiwu, line 111: DO not need this? 
Is CudaT same as nv_bfloat16 for BF16?

onnxruntime/contrib_ops/cuda/activation/activations.cc
```diff
@@ -54,6 +55,7 @@ UNARY_ACTIVATION_OP_HFD(QuickGelu, 1, kMSDomain);
 REGISTER_ACTIVATION_KERNEL(ThresholdedRelu, 1, kOnnxDomain, MLFloat16)
 REGISTER_ACTIVATION_KERNEL(ThresholdedRelu, 1, kOnnxDomain, float)
 REGISTER_ACTIVATION_KERNEL(ThresholdedRelu, 1, kOnnxDomain, double)
+REGISTER_ACTIVATION_KERNEL(ThresholdedRelu, 1, kOnnxDomain, BFloat16)
```
tianleiwu, line 58: PR title is QuickGelu. The change here5 is ThresholdedRelu. Is it expcted?

onnxruntime/contrib_ops/cuda/bert/skip_layer_norm.cc
nenad1002: It comes in as BFloat16 (ONNX type), and on CUDA the type has to change to nv_bfloat16 (CUDA type). It can be written differently maybe, but some form of this is needed.

tianleiwu: If you fix this line (change the mapped type to nv_bfloat16):
https://github.com/microsoft/onnxruntime/blob/77afa69c472659f8ae87a975faaef74dc391e589/onnxruntime/core/providers/cuda/cuda_common.h#L56

I think we do not need this.

tianleiwu (dismissed):

onnxruntime/contrib_ops/cuda/bert/skip_layer_norm.cc
nenad1002: It's not as simple, causes a bunch of issues upstream, this is used somehow, I'll try to fix it

nenad1002: This is still probably the right way to do it, but there are too many places where we manually convert to nv_bfloat16, I am not sure why is it written that way, it looks hard to correct it everywhere without breaking something. And in some places I think I would have to convert it back to BFloat16 (wrong design perhaps?) to convert it to a float or different type (e.g. cublasGemmBatchedHelper).

I think we should just get it in this way, and try to see later can we consolidate the logic or continue this way.

onnxruntime/contrib_ops/cuda/bert/skip_layer_norm.cc
```diff
@@ -106,19 +107,35 @@ Status SkipLayerNorm<T, Simplified>::ComputeInternal(OpKernelContext* ctx) const
...
-        reinterpret_cast<const CudaT*>(input->Data<T>()),
-        reinterpret_cast<const CudaT*>(skip->Data<T>()),
-        (bias != nullptr) ? reinterpret_cast<const CudaT*>(bias->Data<T>()) : nullptr,
-        reinterpret_cast<const CudaT*>(gamma->Data<T>()),
-        (beta != nullptr) ? reinterpret_cast<const CudaT*>(beta->Data<T>()) : nullptr,
-        epsilon_,
-        hidden_size,
-        row_count,
-        skip_size);
+    if (std::is_same<T, BFloat16>::value) {
```
yuslepukhin, line 110: > ( [](http://example.com/codeflow?start=6&length=2)

if constexpr

nenad1002: See the response to Tianlei above. It's not trivial (a bunch of failures happen where when I fix one, I get another, and I think some of my fixes could even slow down inference then). Since I have/had an approval on this one, I would get this in since we are stuck on bf16 models - we cannot build any without this change. I am happy to jump on a call to debug this later.

yuslepukhin (approved):

PR #24776, opened 2025-05-15
Use Memtype as MemHandle only for Graph IO Tensors

gnedanur:
### Description
Memtype Memhandle is applicable only for Graph IO tensors. For other tensors we can leave it as RAW

### Motivation and Context
Compose failed for some models as Memtype is set as MemHandle for static tensors.


HectorSVC (approved):
:shipit:

PR #24781, opened 2025-05-15
[QNN EP] Always write to profiling file if `profiling_file_path` is given.

edgchen1:
### Description

Always write to profiling file if `profiling_file_path` is given.

### Motivation and Context

Previously, on Windows, if the ETW path is enabled, the profiling data will not be written to the file even if `profiling_file_path` is given. I thought that this behavior was confusing.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
@@ -1280,7 +1280,7 @@ Status QnnBackendManager::ExtractBackendProfilingInfo() {
   }
 
   ORT_RETURN_IF(!tracelogging_provider_ep_enabled && profiling_file_path_.empty(),
-                "Need to specify a cvs file via provider option profiling_file_path if ETW not enabled.");
+                "Need to specify a CSV file via provider option profiling_file_path if ETW not enabled.");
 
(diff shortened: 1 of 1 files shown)


onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1399,11 +1400,7 @@ Status QnnBackendManager::ExtractProfilingEventBasic(
   std::string message = GetEventTypeString(event_data.type);
   std::string unit = GetUnitString(event_data.unit);
 
-#ifndef _WIN32
-  tracelogging_provider_ep_enabled = false;
-#endif
```
HectorSVC, line 1404: should we keep this in case it cause problem for non-Windows platform?
#ifndef _WIN32
  tracelogging_provider_ep_enabled = false;
#endif

HectorSVC: maybe not required? since the code is inside #ifdef _WIN32 anyway.

HectorSVC (approved):
:shipit:

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
edgchen1: yeah, I think we can just remove this ifdef.

PR #15833, opened 2023-05-06
ExecutionProvider API refactor - move allocator from EP level to SessionState level and indexed by OrtDevice

jslhcl:
### Description
This PR is to refactor ExecutionProvider API for memory management, which is to move allocators from EP level to SessionState level and indexed by OrtDevice

### Motivation and Context

This PR is to refactor ExecutionProvider API for memory management, which is to move allocators from EP level to SessionState level and indexed by OrtDevice. By this change, EP level will shift the burden of maintaining allocators, which will be user friendly for EP developers

PR #24784, opened 2025-05-16
[TRT EP] Fix trt context memory sharing

toothache:
### Description

Fixed a TRT context memory sharing bug where the context memory was assigned to a unique_ptr that was immediately destructed upon leaving scope.

### Motivation and Context

The bug seems to be introduced by a refactor work: #15833 :

image

### DIFF
--- a/include/onnxruntime/core/framework/allocator.h
+++ b/include/onnxruntime/core/framework/allocator.h
@@ -203,7 +203,8 @@ class IAllocator {
      @returns std::unique_ptr with allocated memory and deleter. Throws if it cannot allocate memory.
   */
   template <typename T>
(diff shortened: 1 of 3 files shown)


toothache:
@yf711 @chilo-ms , may I get a review for this change? Thanks!

chilo-ms (approved):

jywu-msft:
@chilo-ms can you help double check if this same issue is there on NV RTX EP?

chilo-ms:
> @chilo-ms can you help double check if this same issue is there on NV RTX EP?

@ankan-ban @ishwar-raut1 @gedoensmax @gaugarg-nv
Please see this TRT context memory sharing bug and fix in TRT EP.
Even though NV RTX EP for now doesn't expose TRT context memory sharing option, you might need the fix later.

PR #24797, opened 2025-05-16
Validate ep.context_file_path option

HectorSVC:
### Description
Validate ep.context_file_path option, make sure it failed if it's not valid file path


MayureshV1:
Looks Good to Me !

adrianlizarraga (approved):

PR #24445, opened 2025-04-16
[VitisAI EP] Implement new overload of CreateProvider() called with session options

adrianlizarraga:
### Description
A new overload of CreateProvider() was added to the to handle the extraction of EP options from the session option configurations.

### Motivation and Context
Allows use of new Compile API.
Refer to https://github.com/microsoft/onnxruntime/pull/24207

PR #24799, opened 2025-05-17
[VITISAI] pass all session configs to vitisai ep for `Ort::CompileModel` flow

wcy123:
### Description

convert all session configs, i.e. key-value pairs into provider options, the key prefixed with `ort_session_config.`

### Motivation and Context

#24445 has a bug when `Ort::CompileModel`  is used, not all session config are passed to VITISAI EP backend.
It is because that the `session_option` which holds a reference to `VitisiAIExectuionProviderFactory` is not as same as the `session_option` used for `Ort::CompileModel`. `Ort::CompileModel` create another `session_option` behind scene.

The symptom of this bug is that only the session configs in the first `SessionOptions` object is passed to  `VitisiAIExectuionProviderFactory` and session configs in the second `SessionOptions` are not, so that  VITISAI EP backend sometimes assumes that ep.cache_context is not enabled, and then ep context cache model is not created properly.

### DIFF
--- a/onnxruntime/core/providers/vitisai/vitisai_provider_factory.cc
+++ b/onnxruntime/core/providers/vitisai/vitisai_provider_factory.cc
@@ -84,7 +84,7 @@ struct VitisAI_Provider : Provider {
     }
   };
   // Get provider specific custom op domain list. Provider has the resposibility to release OrtCustomOpDomain instances it creates.
-  void GetCustomOpDomainList(IExecutionProviderFactory*, std::vector<OrtCustomOpDomain*>&) override{};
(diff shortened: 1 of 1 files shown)


tianleiwu:
@wcy123, please follow https://github.com/microsoft/onnxruntime/blob/main/docs/Coding_Conventions_and_Standards.md#linting to format the code.

wcy123:
> @wcy123, please follow https://github.com/microsoft/onnxruntime/blob/main/docs/Coding_Conventions_and_Standards.md#linting to format the code.

thank for your comment. linter is done.

HectorSVC (approved):

PR #24803, opened 2025-05-19
[WebNN] Support ConvInteger op

Honry:
WebNN doesn't provide a dedicated op for `ConvInteger`, this PR supports `ConvInteger` op by decomposing it into `DequantizeLinear x, w -> Conv -> Cast (to int32)`.

BTW, adds `ConvInteger` to layout sensitive op list for layout transformation when the preferred layout is NHWC.


fdwr (approved):
👍

guschmue (approved):

PR #24816, opened 2025-05-20
[NV TensorRt RTX EP] : Fix Domain check. 

ishwar-raut1:
### Description

Small change to remove the MS Domain check on onnx model nodes

### Motivation and Context

The check returns unsupported for some nodes having an MS Domain. Trt RTX supports some MS domain ops. if return unsupported these ops falls back to CPU EP

@ankan-ban @chilo-ms @gedoensmax @jywu-msft

### DIFF
--- a/onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
+++ b/onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
@@ -1989,10 +1989,6 @@ NvExecutionProvider::GetCapability(const GraphViewer& graph,
     if (exclude_ops_set.find(node->OpType()) != exclude_ops_set.end()) {
       supported_node = false;
     }
-    // Exclude contrib ops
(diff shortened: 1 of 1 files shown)


ankan-ban:
looks good (we plan to support some of the common contrib ops in the EP - especially for LLM support)

gedoensmax:
How "silent" does our parser do the plitting ? We should make sure that during `supportsModel` no error log is omitted. This is why the change was introduced since a support check should not log errors during checking in my opinion.

chilo-ms:
> How "silent" does our parser do the plitting ? We should make sure that during supportsModel no error log is omitted. This is why the change was introduced since a support check should not log errors during checking in my opinion.

I don't remember in the past i see error being omitted and forces program to exit when TRT parser encounters unsupported ops, it did log/print some info though. 
Anyway, i think it's still worth checking.

chilo-ms (approved):

jywu-msft (approved):

PR #24828, opened 2025-05-20
Update MatMulBNits spec and Add Input Checks

tianleiwu:
### Description

Major changes of spec:
* 2D scale shape: [N * n_blocks_per_col] => [N, n_blocks_per_col]
* 2D zero shape:  [N * CeilDiv(n_blocks_per_col * bits, 8)] => [N, CeilDiv(n_blocks_per_col * bits, 8)]
* For B, drop int32 type and only allow uint8.
* allow bfloat16 as input/output type.
* Mark input g_idx as deprecated (since it has no benefit on model size and performance in inference).

Add a function CheckInputs to verify the input shape.

The reason of the shape change is to make scale and zero compatible with other operators like DequantizeLinear and GatherBlockQuantized. That will make it easy for graph fusion and model builder.

Note that ORT can still handle the legacy 1D format for scale and zero points, and CUDA/CPU could still handle g_idx. However, they are deprecated, and our tools shall generate 2D scale and zeros, and avoid using g_idx going forward.

This change is backward compatible. Model from old spec can run in latest ORT (CheckInputs handles 1D scale and zero points), and model from latest spec can still run in older ORT (since older ORT does not check dimension of scale and zero points)

### Motivation and Context

CUDA and CPU provider does not check inputs for MatMulNBits. It could cause out of boundary access.

We are going to share the lm_head weights of MatMulNBits to GatherBlockQuantized. 2D shape can be used in Gather directly, and we can avoid Reshape nodes.

Our latest models published for foundry use 2D scale and zero points. So I update the spec to reflect that.


onnxruntime/contrib_ops/cpu/quantization/matmul_nbits_helper.h
```diff
@@ -0,0 +1,77 @@
...
+                   const T* group_index,
+                   const T* bias,
+                   int64_t n,
+                   int64_t k,
+                   int64_t block_size,
+                   int64_t bits) {
+  // activation (A)
+  // quantized_weight (B) : (N, k_blocks, blob_size) or null
+  //                        k_blocks = (K + block_size - 1) / block_size
+  //                        blob_size = (block_size * bits + 7) / 8
```
jiafatom, line 28: Currently it seems I only see the case that `block_size * bits` is a multiple of 8. I wonder whether we have some test to cover those cases when it is not a multiple of 8. Where do you get this equation, can you add the description link?

jiafatom:
Did we test this PR to the current big model we are working?

onnxruntime/contrib_ops/cpu/quantization/matmul_nbits_helper.h
tianleiwu: block_size is 16, 32, ..., 128, so block_size * bits will always be multiple of 8.
I will update it to block_size * bits / 8 for simplicity.

tianleiwu:
> Did we test this PR to the current big model we are working?

We have test case to cover 1D and 2D format of scale/zp in CUDA/CPU/WebGPU EPs. I will verify with foundry local later.

js/web/lib/wasm/jsep/webgpu/ops/matmulnbits.ts
```diff
@@ -48,8 +48,11 @@ const validateInputs = (inputs: readonly TensorView[], attributes: MatMulNBitsAt
   if (inputs.length === 4) {
     const zeroPoints = inputs[3];
     const zeroPointsShape = zeroPoints.dims;
+
+    // This assumes zero points are packed.
+    // Unpack format (zero point has same data type and shape as scale) is not supported by webgpu.
     const expectedZeroPointsSize =
-      attributes.bits > 4 ? attributes.n * nBlocksPerCol : attributes.n * Math.floor((nBlocksPerCol + 1) / 2);
+      attributes.n * (attributes.bits === 8 ? nBlocksPerCol : Math.floor((nBlocksPerCol * attributes.bits + 7) / 8));
```
nenad1002, line 55: why this change? Shouldnt for bits == 4 size be ~2x larger, and not (nBlocksPerCol * attributes.bits + 7) / 8) larger?

onnxruntime/contrib_ops/cpu/quantization/matmul_nbits.cc
```diff
@@ -677,11 +678,17 @@ template <typename T1>
 Status MatMulNBits<T1>::Compute(OpKernelContext* ctx) const {
   concurrency::ThreadPool* thread_pool = ctx->GetOperatorThreadPool();
   const Tensor* a = ctx->Input<Tensor>(InputIndex::A);
+  // If B is prepacked, B would have been removed from the context
+  const bool is_b_prepacked = packed_b_size_ > 0;
+  const Tensor* b = is_b_prepacked ? nullptr : ctx->Input<Tensor>(InputIndex::B);
```
nenad1002, line 683: what happens later if this is nullptr?

nenad1002 (approved):

onnxruntime/core/graph/contrib_ops/contrib_defs.cc
```diff
@@ -3428,56 +3428,61 @@ This op functions in much the same was as Dropout-11 and Dropout-13 do, execpt t
...
 
   ONNX_CONTRIB_OPERATOR_SCHEMA(MatMulNBits)
       .SetDomain(kMSDomain)
       .SinceVersion(1)
       .SetDoc(MatMulNBits_ver1_doc)
-      .Attr("K", "size of each input feature", AttributeProto::INT)
-      .Attr("N", "size of each output feature", AttributeProto::INT)
-      .Attr("bits", "number of bits used for weight quantization (default 4)", AttributeProto::INT)
-      .Attr("block_size", "number of groupsize used for weight quantization,(default 128). It needs to be a power of 2 and not smaller than 16.", AttributeProto::INT)
+      .Attr("K", "Input feature dimension of the weight matrix.", AttributeProto::INT)
```
kunal-vaishnavi, line 3451: Can we add an optional attribute that says whether the op is symmetric or asymmetric? AMD had requested this because they did not want to parse the op's input types to see if `uint4` vs `int4` is used.

kunal-vaishnavi, line 3451: Can we add an optional attribute that says whether the op is symmetric or asymmetric? AMD has requested this because they do not want to parse the op's input types to see if `uint4` vs `int4` is used. It is easier to read the attribute instead.

js/web/lib/wasm/jsep/webgpu/ops/matmulnbits.ts
tianleiwu: The original formula only works for bits=4 and 8. The new one works for bits = 2~8.

For example, nBlocksPerCol = 3, bits=4:
* Math.floor((nBlocksPerCol + 1) / 2) = 2
* Math.floor((nBlocksPerCol * attributes.bits + 7) / 8)) = (3 * 4 + 7) / 8 = 19 / 8 = 2
So they are the same for 4 bits.

onnxruntime/core/graph/contrib_ops/contrib_defs.cc
tianleiwu: Current kernel only support uint4 quantized_weight, not int4. 

The default zero point is 8 for 4 bits, so (quantized_weight - zero_point) becomes int4 if they do not pass zero point.

PR #24837, opened 2025-05-22
[WebGPU EP] fix NaN bug in softmax operator

prathikr:
Handle case of row_sum_shared==0.0 for softmax operation

```wgsl
let value = exp(getValue(row, col, row_stride) - row_max_shared) / row_sum_shared;\n"
```


tianleiwu:
In what situation will row_sum_shared be 0? As long as there is any element,  exp(value-max_value) = exp(0) = 1 when value = max_value.

prathikr:
Continued in https://github.com/microsoft/onnxruntime/pull/24855

PR #17953, opened 2023-10-15
Fix cast removal bug

adityagoel4512:
### Description
The `RemoveDuplicateCastTransformer` fairly naively removed Cast nodes from the graph without considering precision loss when using the same `TypeGroup`. For instance, F64 -> F32 -> F64 would be optimised out of the graph. 

I also noticed that signedness was not accounted for, which is not covered by any existing issue but is a problem. For example doing int -> unsigned int -> int produces very different values for negative inputs and so should not be optimised out

One could argue that we shouldn't be performing such cast elimination at all (at least not in this transformer). The original scope might be well restricted to only eliminating unnecessary casts from the `InsertCastTransformer` and no others.

### Motivation and Context
This should fix https://github.com/microsoft/onnxruntime/issues/17565, https://github.com/microsoft/onnxruntime/issues/9915 and https://github.com/microsoft/onnxruntime/issues/8787.

PR #24842, opened 2025-05-22
Cast Nodes Fusion

nenad1002:
### Description

We might have a case where multiple Cast nodes in the chain cast back to the original type. This fusion will remove extra nodes.
E.g.
`A ('float32') -> Cast (to='float16') ->  Cast (to='int4') ->  Cast (to='float32') -> Cast (to='float16') -> B
`
will reduce to
` A ('float32') -> Cast (to='float16') -> B
`
All the Cast nodes throughout the path need to have one input and one output to be considered for the fusion.

### Motivation and Context

Gemma3 ONNX models used to have double casting, and many new models created by the model builder might have as well. Extra Casts might reduce accuracy and increase inference time.

### DIFF
--- a/onnxruntime/core/graph/graph_utils.cc
+++ b/onnxruntime/core/graph/graph_utils.cc
@@ -610,6 +610,11 @@ bool IsGraphInput(const Graph& graph, const NodeArg* input) {
   return std::find(graph_inputs.begin(), graph_inputs.end(), input) != graph_inputs.end();
 }
 
+bool IsGraphOutput(const Graph& graph, const NodeArg* output) {
+  const std::vector<const NodeArg*>& graph_outputs = graph.GetOutputs();
+  return std::find(graph_outputs.begin(), graph_outputs.end(), output) != graph_outputs.end();
+}
+
 bool IsInitializer(const Graph& graph, const std::string& name, bool check_outer_scope) {
   bool is_initializer = false;
   const ONNX_NAMESPACE::TensorProto* initializer = nullptr;
--- a/onnxruntime/core/graph/graph_utils.cc
+++ b/onnxruntime/core/graph/graph_utils.cc
@@ -611,7 +611,7 @@ bool IsGraphInput(const Graph& graph, const NodeArg* input) {
 }
 
 bool IsGraphOutput(const Graph& graph, const NodeArg* output) {
-  const std::vector<const NodeArg*>& graph_outputs = graph.GetOutputs();
+  const auto& graph_outputs = graph.GetOutputs();
   return std::find(graph_outputs.begin(), graph_outputs.end(), output) != graph_outputs.end();
 }
 
--- a/onnxruntime/core/optimizer/cast_chain_elimination.cc
+++ b/onnxruntime/core/optimizer/cast_chain_elimination.cc
@@ -9,6 +9,29 @@
 
 namespace onnxruntime {
 
+static Node* FollowCastChain(Node* current) {
+  while (true) {
+    // A rare case when the Cast node output is branching out.
+    // We don't really want to deal with this complexity, hence we will skip it.
+    if (current->GetOutputEdgesCount() > 1) {
+      return nullptr;
+    }
+
+    auto it = current->OutputNodesBegin();
+    if (it == current->OutputNodesEnd()) {
+      break;
+    }
+
+    Node* next = const_cast<Node*>(&(*it));
+    if (next->OpType() != "Cast") {
+      break;
+    }
+    current = next;
+  }
+
+  return current;
+}
+
 Status CastChainElimination::ApplyImpl(Graph& graph, bool& modified, int graph_level, const logging::Logger& logger) const {
   GraphViewer graph_viewer(graph);
   const auto& node_topology_list = graph_viewer.GetNodesInTopologicalOrder();
@@ -20,13 +43,13 @@ Status CastChainElimination::ApplyImpl(Graph& graph, bool& modified, int graph_l
 
     Node& node = *node_ptr;
 
-    ORT_RETURN_IF_ERROR(Recurse(node, modified, graph_level, logger));
-
-    if (!graph_utils::CanRemoveNode(graph, node, logger)) {
+    if (node.OpType() != "Cast") {
       continue;
     }
 
-    if (node.OpType() != "Cast") {
+    ORT_RETURN_IF_ERROR(Recurse(node, modified, graph_level, logger));
+
+    if (graph_utils::IsGraphOutput(graph, node.OutputDefs()[0])) {
       continue;
     }
 
@@ -35,33 +58,19 @@ Status CastChainElimination::ApplyImpl(Graph& graph, bool& modified, int graph_l
       continue;
     }
 
-    // If not, find the longest chain that casts to the input type, if it exists.
-    Node* current = &node;
-    Node* final_non_cast_node = &node;
-    int matching_elem_type = input_type->tensor_type().elem_type();
-
-    while (current->OpType() == "Cast") {
-      const auto& to_attr = current->GetAttributes().at("to");
-
-      // A rare case when the Cast node output is branching out.
-      // We don't really want to deal with this complexity, hence we will skip it.
-      if (current->GetOutputEdgesCount() > 1) {
-        return Status::OK();
-      }
-
-      auto it = current->OutputNodesBegin();
-      if (it == current->OutputNodesEnd()) {
-        break;
-      }
-      current = const_cast<Node*>(&(*it));
-
-      // We found the repeating pattern.
-      if (to_attr.i() == matching_elem_type) {
-        final_non_cast_node = current;
-      }
+    if (!graph_utils::CanRemoveNode(graph, node, logger)) {
+      continue;
     }
 
-    // No repeating pattern was found.
+    // If not, find the longest chain that casts to the input type, if it exists.
+    Node* current = &node;
+    Node* final_non_cast_node = FollowCastChain(current);
+
+    if (!final_non_cast_node) {
+      continue;
+    }
+
+    // No extra casts were found.
     if (node.Index() == final_non_cast_node->Index()) {
       continue;
     }
--- a/onnxruntime/core/optimizer/cast_chain_elimination.cc
+++ b/onnxruntime/core/optimizer/cast_chain_elimination.cc
@@ -1,0 +1,113 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include "core/common/logging/logging.h"
+#include "core/optimizer/cast_chain_elimination.h"
+#include "core/optimizer/utils.h"
+#include "core/graph/graph.h"
+#include "core/graph/graph_utils.h"
+
+namespace onnxruntime {
+
+Status CastChainElimination::ApplyImpl(Graph& graph, bool& modified, int graph_level, const logging::Logger& logger) const {
+  GraphViewer graph_viewer(graph);
+  const auto& node_topology_list = graph_viewer.GetNodesInTopologicalOrder();
+
+  for (auto node_index : node_topology_list) {
+    auto* node_ptr = graph.GetNode(node_index);
+    if (node_ptr == nullptr)
+      continue;  // we removed the node as part of an earlier fusion
+
+    Node& node = *node_ptr;
+
+    ORT_RETURN_IF_ERROR(Recurse(node, modified, graph_level, logger));
+
+    if (!graph_utils::CanRemoveNode(graph, node, logger)) {
+      continue;
+    }
+
+    if (node.OpType() != "Cast") {
+      continue;
+    }
+
+    const auto* input_type = node.InputDefs()[0]->TypeAsProto();
+    if (input_type == nullptr || !input_type->tensor_type().has_elem_type()) {
+      continue;
+    }
+
+    // If not, find the longest chain that casts to the input type, if it exists.
+    Node* current = &node;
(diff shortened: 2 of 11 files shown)


tianleiwu:
We used to have similar Cast removing logic. That was reversed in https://github.com/microsoft/onnxruntime/pull/17953 to be more conservative. 

It is easy to process model offline like https://github.com/microsoft/onnxruntime/blob/ad7b0e368d453e4e6db646e141ef33e38ff83a7a/onnxruntime/python/tools/transformers/onnx_model.py#L667.

I suggest to add an option in onnxruntime_session_options_config_keys.h. The default is off, and user can turn it on if needed.

nenad1002:
> We used to have similar Cast removing logic but it causes a lot of accuracy issues. That was reversed in #17953 to be more conservative. I suggest to add an option in onnxruntime_session_options_config_keys.h. The default is off, and user can turn it on if needed.
> 
> It is easy to process model offline like
> 
> https://github.com/microsoft/onnxruntime/blob/ad7b0e368d453e4e6db646e141ef33e38ff83a7a/onnxruntime/python/tools/transformers/onnx_model.py#L667
> 
> .
> So another way to avoid the issue to use such post-processing in model builder.

Ok, interesting, did not know this could cause accuracy drop. I can go either way, add a feature filter option or just rely on offline model processing. It is nice to have onnxruntime remove it for us automatically though since we often end up with multiple casts when experimenting.

onnxruntime/core/optimizer/cast_elimination.h
```diff
@@ -11,12 +11,22 @@ namespace onnxruntime {
 @Class CastElimination
 
 Rewrite rule that eliminates Cast nodes if 'to' attribute has same data type as input tensor data type.
+Additionally, it will try to find the longest chain where the 'to' attribute has the same data type as the input of the first Cast node in the chain.
```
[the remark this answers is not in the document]
tianleiwu: It is mixed with two different rules:
* remove useless cast where input type is same as output type
* remove chained cast node.

It is better to have just one rule to keep the logic consistent.

onnxruntime/core/optimizer/cast_chain_elimination.cc
```diff
@@ -0,0 +1,113 @@
...
+  for (auto node_index : node_topology_list) {
+    auto* node_ptr = graph.GetNode(node_index);
+    if (node_ptr == nullptr)
+      continue;  // we removed the node as part of an earlier fusion
+
+    Node& node = *node_ptr;
+
+    ORT_RETURN_IF_ERROR(Recurse(node, modified, graph_level, logger));
+
+    if (!graph_utils::CanRemoveNode(graph, node, logger)) {
```
tianleiwu, line 25: This call is expensive. Please move it below other simple checks.

onnxruntime/core/optimizer/cast_chain_elimination.cc
```diff
@@ -0,0 +1,113 @@
...
+
+    // If not, find the longest chain that casts to the input type, if it exists.
+    Node* current = &node;
+    Node* final_non_cast_node = &node;
+    int matching_elem_type = input_type->tensor_type().elem_type();
+
+    while (current->OpType() == "Cast") {
+      const auto& to_attr = current->GetAttributes().at("to");
+
+      // A rare case when the Cast node output is branching out.
```
tianleiwu, line 46: Check that the Cast output is not a graph output.

onnxruntime/core/optimizer/cast_chain_elimination.cc
nenad1002: This is the first call for all Cast ops in the other fusion.

onnxruntime/core/optimizer/cast_chain_elimination.cc
```diff
@@ -0,0 +1,113 @@
...
+      continue;
+    }
+
+    // If not, find the longest chain that casts to the input type, if it exists.
+    Node* current = &node;
+    Node* final_non_cast_node = &node;
+    int matching_elem_type = input_type->tensor_type().elem_type();
+
+    while (current->OpType() == "Cast") {
+      const auto& to_attr = current->GetAttributes().at("to");
```
tianleiwu, line 44: move this line after checks (like line 55).

It is because, if code is returned or break loop, there is no need to read the to attribute.

onnxruntime/core/optimizer/cast_chain_elimination.cc
```diff
@@ -0,0 +1,113 @@
...
+      }
+
+      auto it = current->OutputNodesBegin();
+      if (it == current->OutputNodesEnd()) {
+        break;
+      }
+      current = const_cast<Node*>(&(*it));
+
+      // We found the repeating pattern.
+      if (to_attr.i() == matching_elem_type) {
```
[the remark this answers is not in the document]
nenad1002: Can we? What if the specific pair of 'from/to' casting is not implemented?

onnxruntime/core/optimizer/cast_chain_elimination.cc
```diff
@@ -0,0 +1,113 @@
...
+    Node* current = &node;
+    Node* final_non_cast_node = &node;
+    int matching_elem_type = input_type->tensor_type().elem_type();
+
+    while (current->OpType() == "Cast") {
+      const auto& to_attr = current->GetAttributes().at("to");
+
+      // A rare case when the Cast node output is branching out.
+      // We don't really want to deal with this complexity, hence we will skip it.
+      if (current->GetOutputEdgesCount() > 1) {
```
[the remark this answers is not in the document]
nenad1002: The case above will work with current code

onnxruntime/core/optimizer/cast_chain_elimination.cc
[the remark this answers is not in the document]
tianleiwu: It is elementwise static cast. A generic implementation shall implement pair-wise of all numeric types.

Also, this fusion is off by default, so user do not need to turn it on if it is not working for some EP.

onnxruntime/core/optimizer/cast_chain_elimination.cc
```diff
@@ -0,0 +1,122 @@
...
+namespace onnxruntime {
+
+static Node* FollowCastChain(Node* current) {
+  while (true) {
+    // A rare case when the Cast node output is branching out.
+    // We don't really want to deal with this complexity, hence we will skip it.
+    if (current->GetOutputEdgesCount() > 1) {
+      return nullptr;
+    }
+
```
tianleiwu, line 19: Add a check:
```
if (graph.NodeProducesGraphOutput(node)) {
    return nullptr;
}
```

onnxruntime/core/optimizer/cast_chain_elimination.cc
```diff
@@ -0,0 +1,122 @@
...
+      continue;
+    }
+
+    if (!graph_utils::CanRemoveNode(graph, node, logger)) {
+      continue;
+    }
+
+    // If not, find the longest chain that casts to the input type, if it exists.
+    Node* current = &node;
+    Node* final_non_cast_node = FollowCastChain(current);
```
tianleiwu, line 67: FollowCastChain could return nullptr. Need check return value.

onnxruntime/core/optimizer/cast_chain_elimination.cc
```diff
@@ -0,0 +1,122 @@
...
+
+    if (!final_non_cast_node) {
```
tianleiwu, line 69: final_non_cast_node == nullptr

onnxruntime/core/graph/graph_utils.cc
```diff
@@ -610,6 +610,11 @@ bool IsGraphInput(const Graph& graph, const NodeArg* input) {
   return std::find(graph_inputs.begin(), graph_inputs.end(), input) != graph_inputs.end();
 }
 
+bool IsGraphOutput(const Graph& graph, const NodeArg* output) {
+  const std::vector<const NodeArg*>& graph_outputs = graph.GetOutputs();
```
yuslepukhin: >std::vector<const NodeArg*> [](http://example.com/codeflow?start=8&length=27)

nit: consider const auto&

onnxruntime/core/optimizer/graph_transformer_utils.cc
```diff
@@ -198,6 +199,8 @@ InlinedVector<std::unique_ptr<GraphTransformer>> GenerateTransformers(
   InlinedVector<std::unique_ptr<GraphTransformer>> transformers;
   const bool disable_quant_qdq =
       session_options.config_options.GetConfigOrDefault(kOrtSessionOptionsDisableQuantQDQ, "0") == "1";
+  const bool enable_cast_chain_elimination =
+      session_options.config_options.GetConfigOrDefault(kOrtSessionOptionsEnableCastChainElimination, "0") == "1";
```
yuslepukhin, line 209: Any reason why this should not be enabled just like anything else?

onnxruntime/core/optimizer/cast_chain_elimination.cc
```diff
@@ -0,0 +1,122 @@
...
+    if (current->GetOutputEdgesCount() > 1) {
+      return nullptr;
+    }
+
+    auto it = current->OutputNodesBegin();
+    if (it == current->OutputNodesEnd()) {
+      break;
+    }
+
+    Node* next = const_cast<Node*>(&(*it));
```
yuslepukhin: >= const_cast<Node*>(&(*it)); [](http://example.com/codeflow?start=15&length=28)

Is there any way we can get rid of const_cast?

onnxruntime/core/optimizer/graph_transformer_utils.cc
nenad1002: Accuracy drops were observed in the past in some models for this type of the fusion

onnxruntime/core/optimizer/cast_chain_elimination.cc
nenad1002: I am doing refactor, it will go away

onnxruntime/core/optimizer/cast_chain_elimination.cc
```diff
@@ -0,0 +1,71 @@
...
+  return Status::OK();
+}
+
+bool CastChainElimination::SatisfyCondition(const Graph& graph, const Node& node, const logging::Logger& logger) const {
+  if (graph_utils::IsGraphOutput(graph, node.OutputDefs()[0])) {
+    return false;
+  }
+
+  const auto* input_type = node.InputDefs()[0]->TypeAsProto();
+  if (input_type == nullptr || !input_type->tensor_type().has_elem_type()) {
```
tianleiwu: No need to check input type?

onnxruntime/core/optimizer/cast_chain_elimination.cc
```diff
@@ -0,0 +1,71 @@
...
+
+  graph.RemoveNode(node.Index());
+
+  rule_effect = RewriteRuleEffect::kRemovedCurrentNode;
+
+  return Status::OK();
+}
+
+bool CastChainElimination::SatisfyCondition(const Graph& graph, const Node& node, const logging::Logger& logger) const {
+  if (graph_utils::IsGraphOutput(graph, node.OutputDefs()[0])) {
```
tianleiwu: CanRemoveNode will also check graph output.

onnxruntime/core/optimizer/cast_chain_elimination.cc
```diff
@@ -0,0 +1,71 @@
...
+#include "core/common/logging/logging.h"
+#include "core/optimizer/rewrite_rule.h"
+#include "core/optimizer/cast_chain_elimination.h"
+#include "core/optimizer/utils.h"
+#include "core/graph/graph.h"
+#include "core/graph/graph_utils.h"
+
+namespace onnxruntime {
+
+Status CastChainElimination::Apply(Graph& graph, Node& node, RewriteRuleEffect& rule_effect, const logging::Logger&) const {
```
tianleiwu, line 13: I think this function can be same like CastElimination::Apply?
https://github.com/microsoft/onnxruntime/blob/3ca8a49f235fde90318ee3055b0dd9ab23b53602/onnxruntime/core/optimizer/cast_elimination.cc#L14-L18

onnxruntime/core/optimizer/cast_chain_elimination.cc
```diff
@@ -0,0 +1,71 @@
...
+  auto nextNodeIt = node.OutputNodesBegin();
+
+  Node* next = graph.GetNode(nextNodeIt->Index());
+
+  // Skip if the next node is not of type Cast.
+  if (next->OpType() != "Cast") {
```
tianleiwu: move this check to SatisfyCondition function.

onnxruntime/core/optimizer/cast_chain_elimination.cc
nenad1002: It cannot, RemoveNode method will fail

```C++ exception with description "/home/nebanfic/onnxruntime/onnxruntime/core/graph/graph.cc:1613 void onnxruntime::Graph::AddEdge(onnxruntime::NodeIndex, onnxruntime::NodeIndex, int, int) Argument type mismatch when adding edge.```

nenad1002: It's very possible the method has a bug, but it's not obvious. We should prob avoid using this method, it was never working for me properly, it's trying to do too many things at once, I think it needs to be revisited in future and split into many smaller depending on fuctionality.

tianleiwu (approved):

Issue #24771, opened 2025-05-15
[Build] v1.22 requires dxcore.dll


pfriesch:
### Describe the issue

The latest version does not load on Windows Server as dxcore.dll is missing. DirectX can't be installed either. Is that an intended requirement or was it added by mistake?

```
C:\Users\Curiosity\Downloads\Dependencies_x64_Release>Dependencies.exe -depth 1 -chain C:\Curiosity\onnxruntime.dll
├ onnxruntime.dll (ROOT) : C:\Curiosity\onnxruntime.dll
|  ├ KERNEL32.dll (WellKnownDlls) : C:\Windows\system32\kernel32.dll
|  ├ MSVCP140.dll (WindowsFolder) : C:\Windows\system32\MSVCP140.dll
|  ├ VCRUNTIME140_1.dll (WindowsFolder) : C:\Windows\system32\VCRUNTIME140_1.dll
|  ├ VCRUNTIME140.dll (WindowsFolder) : C:\Windows\system32\VCRUNTIME140.dll
|  ├ api-ms-win-crt-runtime-l1-1-0.dll (ApiSetSchema) : C:\Windows\system32\ucrtbase.dll
|  ├ api-ms-win-crt-stdio-l1-1-0.dll (ApiSetSchema) : C:\Windows\system32\ucrtbase.dll
|  ├ api-ms-win-crt-string-l1-1-0.dll (ApiSetSchema) : C:\Windows\system32\ucrtbase.dll
|  ├ api-ms-win-crt-convert-l1-1-0.dll (ApiSetSchema) : C:\Windows\system32\ucrtbase.dll
|  ├ api-ms-win-crt-filesystem-l1-1-0.dll (ApiSetSchema) : C:\Windows\system32\ucrtbase.dll
|  ├ api-ms-win-crt-heap-l1-1-0.dll (ApiSetSchema) : C:\Windows\system32\ucrtbase.dll
|  ├ dxcore.dll (NOT_FOUND) :
|  ├ ADVAPI32.dll (WellKnownDlls) : C:\Windows\system32\advapi32.dll
|  ├ SETUPAPI.dll (WellKnownDlls) : C:\Windows\system32\Setupapi.dll
|  ├ dxgi.dll (WindowsFolder) : C:\Windows\system32\dxgi.dll
|  ├ api-ms-win-core-path-l1-1-0.dll (ApiSetSchema) : C:\Windows\system32\kernelbase.dll
[...]

PR #24845, opened 2025-05-23
Weaken dxcore dependency

skottmckay:
### Description

Do a manual load of dxcore.dll so that old Windows versions are still supported.

### Motivation and Context

#24771


onnxruntime/core/platform/windows/device_discovery.cc
```diff
@@ -362,90 +362,109 @@ std::unordered_map<uint64_t, DeviceInfo> GetDeviceInfoD3D12() {
   return device_info;
 }
 
+typedef HRESULT(WINAPI* PFN_DXCoreCreateAdapterFactory)(REFIID riid, void** ppvFactory);
```
skottmckay, line 365: Hide the whitespace diffs as there are very few actual lines that are changed. 

image

ashrit-ms (approved):

onnxruntime/core/platform/windows/device_discovery.cc
```diff
@@ -362,90 +362,111 @@ std::unordered_map<uint64_t, DeviceInfo> GetDeviceInfoD3D12() {
...
+
 // returns LUID to DeviceInfo
 std::unordered_map<uint64_t, DeviceInfo> GetDeviceInfoDxcore() {
   std::unordered_map<uint64_t, DeviceInfo> device_info;
 
-  // Get all GPUs and NPUs by querying WDDM/MCDM.
-  wil::com_ptr<IDXCoreAdapterFactory> adapterFactory;
-  if (FAILED(DXCoreCreateAdapterFactory(IID_PPV_ARGS(&adapterFactory)))) {
+  // Load dxcore.dll. We do this manually so there's not a hard dependency on dxcore which is newer.
```
snnn, line 371: There is another way to solve this problem: use apiset https://learn.microsoft.com/en-us/windows/win32/apiindex/detect-api-set-availability
But I am not quite understand how it works. Because at link time the binary still need to link to something.

skottmckay: As it doesn't matter if `DXCoreCreateAdapterFactory` is implemented via a stub (a stub returning nothing is the same to us in this usage as no stub being available), the value of using IsApiSetImplemented seems minimal here.

snnn (approved):

onnxruntime/core/platform/windows/device_discovery.cc
```diff
@@ -362,90 +362,111 @@ std::unordered_map<uint64_t, DeviceInfo> GetDeviceInfoD3D12() {
...
+  HMODULE dxcore_lib = LoadLibraryExW(L"dxcore.dll", nullptr, LOAD_LIBRARY_SEARCH_SYSTEM32);
```
edgchen1, line 372: nit: it would be safer to use an RAII type like unique_ptr to handle the FreeLibrary call. it's not obvious to me that the rest of the code before the existing FreeLibrary call will not throw or return early.

snnn: WIL has such helpers

onnxruntime/core/platform/windows/device_discovery.cc
```diff
@@ -362,89 +362,109 @@ std::unordered_map<uint64_t, DeviceInfo> GetDeviceInfoD3D12() {
...
-    if (FAILED(adapterFactory->CreateAdapterList(1, hwAttribute, IID_PPV_ARGS(&adapterList)))) {
-      continue;
+  do {
+    auto pfnDXCoreCreateAdapterFactory = reinterpret_cast<PFN_DXCoreCreateAdapterFactory>(
+        GetProcAddress(dxcore_lib.get(), "DXCoreCreateAdapterFactory"));
+
+    if (!pfnDXCoreCreateAdapterFactory) {
+      // this isn't expected to fail so ERROR not WARNING
+      LOGS_DEFAULT(ERROR) << "Failed to get DXCoreCreateAdapterFactory function address.";
+      break;
```
edgchen1, line 385: maybe simply `return device_info`, now that we don't need to explicitly call FreeLibrary at the end? and the do/while(false) doesn't seem to add much now.

edgchen1 (approved):

skottmckay:
> Does this mean that after the merge the minimum build would go from 10.0.22621 (22H2) back to 10.0.19041 (20H1)?

IIUC dxcore was added in 10.0.19041 so this should support versions prior to that by making it an optional runtime dependency. I believe Windows Server 2019 uses 10.0.17763 and the intention of this change is for device discovery to not fail on that.

> Also what about this other dxcore.lib dependency?
> 
https://github.com/microsoft/onnxruntime/blob/70de20b94fdc77a0e603ef868a2c53ef38da557c/cmake/onnxruntime_providers_dml.cmake#L62

That's for the DML execution provider. If you're building with that you have a hard dependency on dxcore being available.

JulienMaille:
> That's for the DML execution provider. If you're building with that you have a hard dependency on dxcore being available.

Can you elaborate on this? I thought ONNXRuntime used to be working fine on 10.0.17763 several versions ago. Why is dxcore mandatory for DML?

snnn:
ONNX Runtime has multiple execution providers and multiple packages. For example, if you want to use onnxruntime-gpu python package from pypi.org, you should have CUDA.  If you want to use onnxruntime-directml python package from pypi.org, you should the directml DLLs. If you just want to use the default CPU provider, you don't need to have these dependencies.

JulienMaille:
@snnn Thanks, I get that. But it seems you can have DML on old SDKs as long as you distribute `directML.dll`. `dxcore.dll` is only used to enumerate devices, it's not mandatory to run inference.
The doc states https://onnxruntime.ai/docs/get-started/with-cpp.html#builds
|  Artifact | Description  |   Supported Platforms  |
|---|---|---|
| Microsoft.ML.OnnxRuntime.DirectML  | GPU - DirectML (Release)  |  Windows 10 1709+  |

And Win10 1709 doesn't come with `dxcore.dll` AFAIK
I'm currently trying to compile 1.22.0 with an old SDK to support this theory, I can submit my patches as PR when confirmed.

snnn:
So, your goal is to run the DirectML EP down to Windows 10 1709?

JulienMaille:
@snnn Actually down to Windows 10 1809 would be enough. I will try to find a computer and install this Win10 build on it, tight now I only tested in a VM
image

aseovic:
Are there plans to release Java ONNX Runtime with this fix to Maven Central?

The latest version I see there is 1.22.0, which fails during initialization on Windows Server 2019:
```
03:04:23 Caused by: java.lang.ExceptionInInitializerError: Exception java.lang.UnsatisfiedLinkError: C:\Users\ci\AppData\Local\Temp\onnxruntime-java8448688778529667369\onnxruntime.dll: Can't find dependent libraries [in thread "main"]
03:04:23 	at java.base/jdk.internal.loader.NativeLibraries.load(Native Method)
03:04:23 	at java.base/jdk.internal.loader.NativeLibraries$NativeLibraryImpl.open(NativeLibraries.java:321)
03:04:23 	at java.base/jdk.internal.loader.NativeLibraries.loadLibrary(NativeLibraries.java:187)
03:04:23 	at java.base/jdk.internal.loader.NativeLibraries.loadLibrary(NativeLibraries.java:129)
03:04:23 	at java.base/java.lang.ClassLoader.loadLibrary(ClassLoader.java:2245)
03:04:23 	at java.base/java.lang.Runtime.load0(Runtime.java:767)
03:04:23 	at java.base/java.lang.System.load(System.java:1624)
03:04:23 	at ai.onnxruntime.OnnxRuntime.load(OnnxRuntime.java:387)
03:04:23 	at ai.onnxruntime.OnnxRuntime.init(OnnxRuntime.java:166)
03:04:23 	at ai.onnxruntime.OrtEnvironment.<clinit>(OrtEnvironment.java:34)
```

fs-eire:
This will be included in the incoming 1.23.0 release

PR #24847, opened 2025-05-23
Download protobuf dependency on ARM64 build host

qc-tbhardwa:
Windows on ARM support AMD64 emulation, so we can use win64 version of protoc.

Description
Compilation on ARM64 machine fails due to missing protoc dependency.

Motivation and Context
With this change we can compile onnxruntime on Windows on Arm devices without setting protobuf manually. CMake will download and setup protoc dependency.


HectorSVC:
We have built on ARM64 machine for a long time, never encounter any issue with protoc. Don't know what's the issue you you encountered.

qc-tbhardwa:
If we try to compile arm64ec on arm64 device then it fails with below error.

Doing crosscompiling
Numpy version: 2.2.5
-- Loading Dependencies URLs ...
-- Loading Dependencies ...
-- Abseil source dir:C:/repos/onnxruntime_gqa/build/arm64_build_err_ec/Release/_deps/abseil_cpp-src
-- CMAKE_HOST_SYSTEM_NAME: Windows
CMake Error at external/onnxruntime_external_deps.cmake:162 (message):
  ONNX_CUSTOM_PROTOC_EXECUTABLE must be set to cross-compile.
Call Stack (most recent call first):
  CMakeLists.txt:699 (include)
  
  
  For cross-compiling, we are not fetching protoc if processor is AM64.

cmake/external/onnxruntime_external_deps.cmake
```diff
@@ -132,6 +132,9 @@ if(NOT ONNX_CUSTOM_PROTOC_EXECUTABLE AND NOT onnxruntime_USE_VCPKG)
       elseif(CMAKE_HOST_SYSTEM_PROCESSOR STREQUAL "x86")
         onnxruntime_fetchcontent_declare(protoc_binary URL ${DEP_URL_protoc_win32} URL_HASH SHA1=${DEP_SHA1_protoc_win32} EXCLUDE_FROM_ALL)
         FetchContent_Populate(protoc_binary)
+      elseif(CMAKE_HOST_SYSTEM_PROCESSOR STREQUAL "ARM64")
```
HectorSVC, line 135: >CMAKE_HOST_SYSTEM_PROCESSOR STREQUAL "ARM64" [](http://example.com/codeflow?start=13&length=44)

My concern is it always uses win64 protoc for arm64 device, it may cause some perf issue.

According to your comment it's to resolve the arm64ec build on arm64 device. So we should add:
AND onnxruntime_target_platform STREQUAL "ARM64EC"

qc-tbhardwa: This error is seen whenever we are cross compiling on ARM64 processor.
On AMD64 python, both arm64 and arm64ec are cross-compilation.
On ARM64 python only arm64ec is cross-compilation.
If we ever try to compile for x86_64that will also be cross-compilation.

Also, here we are downloading protoc_binary, which will be used to compile proto files into source code, then source code will be compiled with cross compiler. If this understanding is correct, then we should not have perf issue as we are compiling compiled code in native architecture. Only proto -> source file conversion will be using emulated x64 code.

We can use native protoc_binary when we update protoc dependency to newer version, where ARM64 release is available.

HectorSVC: good point

PR #24848, opened 2025-05-23
[QNN-EP] Define SpaceToDepth fusion for YOLOv2.

minfhong-quic:
### Description

- Add SpaceToDepth fusion for QNN preprocess.
  - The pattern in YOLOv2 is uncommon while the common seen one is left as future work.
- Add entry point/API for non-quantization user to preprocess models for QNN execution.
- Revise cmake to package newly introduced directory into Python wheel.

### Motivation and Context

- While executing YOLOv2 model on QNN-EP, a sequence of Reshape and Transpose having 6D shapes are falling back to CPU due to HTP limitation. Add fusion to fuse this sequence of ops into a single SpaceToDepth which can be directly executed on QNN-EP.
- Since current QNN preprocess is provided in `onnxruntime/python/tools/quantization/execution_providers/qnn/preprocess.py` which is under quantization directory, the path may be confusing for non-quantization users. In order to allow non-quantization users to preprocess models for QNN, introduce `onnxruntime/python/tools/qnn/preprocess.py` to serve as the entry point and provide API to preprocess models.

### DIFF
--- a/cmake/onnxruntime_python.cmake
+++ b/cmake/onnxruntime_python.cmake
@@ -453,6 +453,9 @@ endif()
 file(GLOB onnxruntime_python_tools_srcs CONFIGURE_DEPENDS
     "${ONNXRUNTIME_ROOT}/python/tools/*.py"
 )
+file(GLOB onnxruntime_python_tools_qnn_src CONFIGURE_DEPENDS
+    "${ONNXRUNTIME_ROOT}/python/tools/qnn/*.py"
+)
 file(GLOB onnxruntime_python_quantization_src CONFIGURE_DEPENDS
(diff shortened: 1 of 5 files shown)


minfhong-quic:
Hi @HectorSVC, appreciate for triggering the CI pipelines.
However, one of the pipelines failed at [GradientCheckerTest.ReduceMaxGrad](https://github.com/microsoft/onnxruntime/actions/runs/15204416520/job/42811967767?pr=24848) under `orttraining\orttraining\test\gradient\gradient_ops_test.cc` which should not be relevant with my PR.
Any suggestion for this situation? Thanks!

HectorSVC (approved):
:shipit:

PR #24854, opened 2025-05-23
[CUDA] fp16 intB gemm

tianleiwu:
### Description
* Add fpA intB gemm kernel from WeightOnlyGroupwiseQuantGemmPlugin of TensorRT-LLM.
* Add prepacking to convert weight/scales/zero_points to adapt MatMulNBits to use the kernel.

Limitations:
* Only enable fp16 kernel. BF16 support will be added later.
* Requires zero points. The support of scales only might be added later.
* Bias is not enabled  since previous MatMulNBits kernel does not support bias.

### Motivation and Context

To improve performance of LLM. 

Initial result shows 2.2x throughput on prompt processing and 1.25X throughput on token generation using onnxruntime-genai benchmark_e2e.py on phi-4-mini-instruct  on A100.


onnxruntime/contrib_ops/cuda/llm/cutlass_extensions/compute_occupancy.h
```diff
@@ -0,0 +1,74 @@
...
+#include "core/providers/cuda/cuda_common.h"
+
+namespace onnxruntime::llm {
+namespace cutlass_extensions {
+
+template <typename GemmKernel, bool enable_cutlass_3x = false>
+inline int compute_occupancy_for_kernel() {
+  int smem_size = static_cast<int>(sizeof(typename GemmKernel::SharedStorage));
+
+  if (smem_size > (48 << 10)) {
```
nenad1002, line 31: is this for a specific GPU? What if we want higher in the future? Is it possible to parametrize this?

onnxruntime/contrib_ops/cuda/llm/common/logger.h
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
nenad1002, line 1: Why not use the existing logger (include/onnxruntime/core/common/logging/logging.h)?

onnxruntime/contrib_ops/cuda/llm/cutlass_preprocessors.cc
```diff
@@ -0,0 +1,687 @@
...
+//  0 1 8 9 16 17 24 25 2 3 10 11 18 19 26 27 4 5 12 13 20 21 28 29 6 7 14 15 22 23 30 31
+// For W4_A8, see the map in the code. The idea is similar to above.
+// The goal of this permutation is to ensure data ends up in the correct threads after
+// we execute LDSM. It counteracts the effect of the data being of different widths.
+// For more information about the expected layouts, see the MMA section in the PTX docs.
+std::vector<int> get_permutation_map(QuantType quant_type) {
+  if (quant_type == QuantType::W8_A16) {
+    return {0, 1, 8, 9, 2, 3, 10, 11, 4, 5, 12, 13, 6, 7, 14, 15};
+  } else if (quant_type == QuantType::W4_A16) {
+    return {0, 1, 8, 9, 16, 17, 24, 25, 2, 3, 10, 11, 18, 19, 26, 27, 4, 5, 12, 13, 20, 21, 28, 29, 6, 7, 14, 15,
```
nenad1002, line 149: Where are these numbers coming from? Any asserts needed?

onnxruntime/contrib_ops/cuda/llm/common/logger.h
tianleiwu: It's using existing logger. Just some macros.

onnxruntime/contrib_ops/cuda/llm/fpA_intB_gemm_adpator.cu
```diff
@@ -0,0 +1,275 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
kunal-vaishnavi, line 1: This file name has a spelling error. It should end with `adaptor`.

onnxruntime/contrib_ops/cuda/llm/cutlass_extensions/epilogue/thread/fused_activations.h
```diff
@@ -0,0 +1,61 @@
...
+/////////////////////////////////////////////////////////////////////////////////////////////////
+
+__forceinline__ __device__ float copysignf_pos(float a, float b) {
+  float r;
+  r = __int_as_float(__float_as_int(a) | (__float_as_int(b) & 0x80000000));
+  return r;
+}
+
+__forceinline__ __device__ float tanh_opt(float x) {
+#if (__CUDACC_VER_MAJOR__ < 11) || (__CUDA_ARCH__ < 750)
```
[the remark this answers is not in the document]
tianleiwu: I think the performance is same for sm >= 7.5. 
For sm < 7.5, may need some perf test to see whether it is faster.

onnxruntime/contrib_ops/cuda/llm/cutlass_extensions/compute_occupancy.h
tianleiwu: The 48 KB  threshold is a practical default for shared memory allocation per block in CUDA kernels. Some newer GPU architectures support more than 48 KB of shared memory per block. To utilize this, must allocate shared memory dynamically and explicitly set the desired size using cudaFuncSetAttribute just like the code here.

onnxruntime/contrib_ops/cuda/llm/cutlass_preprocessors.cc
tianleiwu: This is row permutation map for 16 (for 8 bits) or 32 (for 4 bits).

PR #24855, opened 2025-05-24
[WebGPU EP] Fix NaN bug in softmax operator

prathikr:
Handle NaN in softmax operator for WebGPU EP and JSEP.


onnxruntime/core/providers/webgpu/math/softmax.cc
```diff
@@ -141,7 +141,12 @@ Status SoftmaxProgram::GenerateShaderCode(ShaderHelper& shader) const {
 
       // Calculate the final value for each element in the row
       << "  for (var col = lindex; col < cols; col += wg) {\n"
-      << "    let value = exp(getValue(row, col, row_stride) - row_max_shared) / row_sum_shared;\n"
+      << "    var value : x_value_t;\n"
+      << "    if (all(row_sum_shared == x_value_t(0.0))) {\n"
```
tianleiwu, line 145: It's better to put this check before the loop like:
```
    if (sum_is_zero) {
          for ...
     } else {
          for ...
     }
 ```
 
  Could you add a test case that could trigger this bug?

tianleiwu (approved):

guschmue (dismissed):

tianleiwu (approved):

PR #24869, opened 2025-05-27
[NV TensorRT RTX EP] enable weight stripped engines with EP Context

thevishalagarwal:
Enable NV TRT RTX EP engines to be weight stripped always when using EP Context

We want to always use weight-stripped engines for EP Context to reduce disk footprint on end-user system. With this, there are two ways to load weights
1. provide weights via bytestream (recommended)
2. original `model.onnx` present in the same folder as the `model_ctx.onnx`

```cpp
std::vector<char> model_bytes = ReadFileFromDisk("model.onnx");

// weight refitting using bytesteam
std::unordered_map<std::string, std::string> rtx_ep_options;
rtx_ep_options[onnxruntime::nv::provider_option_names::kONNXBytestream] = std::to_string(reinterpret_cast<size_t>(model_bytes.data()));
rtx_ep_options[onnxruntime::nv::provider_option_names::kONNXBytestreamSize] = std::to_string(model_bytes.size());
```

### DIFF
--- a/onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider_info.cc
+++ b/onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider_info.cc
@@ -81,8 +81,8 @@ NvExecutionProviderInfo NvExecutionProviderInfo::FromProviderOptions(const Provi
 
   // If embed mode is not specified, default to 1 if dump_ep_context_model is true, otherwise 0
   const auto embed_mode = std::stoi(session_options.GetConfigOrDefault(kOrtSessionOptionEpContextEmbedMode, "-1"));
-  if(embed_mode == -1) {
-    if(info.dump_ep_context_model)
+  if (embed_mode == -1) {
+    if (info.dump_ep_context_model)
       embed_mode = 1;
     else
       embed_mode = 0;


thevishalagarwal:
cc @ankan-ban @gedoensmax @ishwar-raut1

ankan-ban:
@chilo-ms to review.

thevishalagarwal:
@chilo-ms ping for review

onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider_info.cc
```diff
@@ -64,23 +64,27 @@ NvExecutionProviderInfo NvExecutionProviderInfo::FromProviderOptions(const Provi
...
+  const auto ep_context_enable = session_options.GetConfigOrDefault(kOrtSessionOptionEpContextEnable, "1");
+  if (ep_context_enable == "0") {
+    info.dump_ep_context_model = false;
+  } else if (ep_context_enable == "1") {
+    info.dump_ep_context_model = true;
+    info.ep_context_embed_mode = 1;
+    info.weight_stripped_engine_enable = true;
+  } else {
+    ORT_THROW("Invalid ", kOrtSessionOptionEpContextEnable, " must 0 or 1");
+  }
```
gedoensmax: This looks a lot like we only use weight stripped engines for embed mode `1` ? Does this align with other defaults ?

gedoensmax (commented):
@adrianlizarraga how does QNN does this weight sharing concept / weight less EP context nodes ? Are the weights to a compiled node handled as initializer or are weight always baked into the EP context binary ? 

Ideally we would not require this hacky solution over `kONNXBytestream` but could use `AddExternalInitializers` or similar IMHO.

onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider_info.cc
thevishalagarwal: We use weight stripped engine only when `kOrtSessionOptionEpContextEnable` is `true`. 

If `kOrtSessionOptionEpContextEmbedMode` is not specified, then it defaults to `1` when EP context is enabled. Otherwise, it defaults to 0. I have again refactored the code just to make it simpler

gedoensmax: Ok I see yeah that makes sense since we do not enable TRT RTX to dump an engine in another way than an EP Context node.

Issue #24773, opened 2025-05-15
Intermittent crash in ETW logging


daveorourke:
### Describe the issue

We're shipping onnxruntime.dll with our product, Camtasia.  We originally shipped v1.18.0, which was stable.  Later, we updated to v1.20.0, and we started getting crash reports from our telemetry.

To troubleshoot this issue, we built from the [v1.20.0](https://github.com/microsoft/onnxruntime/releases/tag/v1.20.0) tag so we could get .pdb debug symbols for onnxruntime.dll.  We then reproduced the crash, which gave us the crash call stack shown here: 
![Image](https://github.com/user-attachments/assets/8b1948b6-a49d-49c9-bd80-26dcb5f48b58)

It's unclear (to us) what ETW event is triggering this crash, and we don't have reliable repro steps.  But we do know the following:
1. The crash was introduced between v1.18.0 and v1.20.0 with the introduction of ETW logging.
2. The crash seems to occur at random (apparently from some ETW event), while our application is open, but can occur without any user interaction, and more often on resume from sleep.
3. The crash .dmp always has the same call stack as shown above.

We did note that the crash mentioned in this [commit](https://github.com/microsoft/onnxruntime/commit/55f7f9d7a9b88c4e7f0eb7cf4d7f31004761f5cb) has a very similar call stack.  Unfortunately, that commit was prior to the v1.20.0 tag and the crash persists.

#### Crashes by OSVersion
![Image](https://github.com/user-attachments/assets/e72791c9-5773-42d8-a85f-4d16cc37e9c0)

### To reproduce

1. Launch Camtasia Windows 2025.1.1 (current version, as of 5/15/2025), which uses onnxruntime v1.20.0
2. Wait. 😐 

As shown above, the crash occurs in onnxruntime.dll stemming from an ETW callback.  But we don't have reliable repro steps to trigger it. ☹

### Urgency

As of today (5/15/2025), we have over 2,000 crash reports from users who have voluntarily submitted them (users get a crash dialog with Send / Don't Send buttons).  We think most users click "Don't Send", so the actual number is much higher.

[...]

PR #24877, opened 2025-05-27
Improve Windows ETW callback registration and fix issues

yuslepukhin:
### Description

- `EtwRegistrationManager`. Make sure all fields initialized by a constructor
- Register `ML_Ort_Provider_Etw_Callback` once for all the sessions. The first session registers, the last one to go aware removes the callback to Log all sessions.
- Register a callback object instead of a pointer to it. Store it in the map with a session unique key.
- Create a registration guard to remove callbacks in case `InferenceSession` constructor does not finish.

### Motivation and Context

This PR is inspired by https://github.com/microsoft/onnxruntime/issues/24773?reload=1.

Current code exhibits multiple issues.
- `EtwRegistrationManager` constructor does not initialize all of the fields including the `InitializationStatus`.
- Global callback object is registered and re-created by every session. Customers sometimes run thousands of models in the same sessions which results in a quadratic ETW costs. The callback object is destroyed and recreated every time a session is created.
- There is a chance that InferenceSession constructor does not finish, and the callback would remain registered. This may result in intermittent hard to diagnose bugs.

### DIFF
--- a/onnxruntime/core/framework/execution_provider.cc
+++ b/onnxruntime/core/framework/execution_provider.cc
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 #include "core/framework/execution_provider.h"
+#include "core/framework/execution_providers.h"
 
 #include "core/graph/graph_viewer.h"
 #include "core/framework/compute_capability.h"
@@ -8,6 +9,8 @@
 #include "core/framework/kernel_registry_manager.h"
 #include "core/framework/murmurhash3.h"
 #include "core/framework/op_kernel.h"
+
+#include <stdint.h>
 
 namespace onnxruntime {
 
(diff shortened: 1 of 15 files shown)


yuslepukhin:
Testing was performed using a usual build test for CPU and QNN providers.
I also used a special drop in package provided by @ivberg to run several concurrent sessions along with profiling. The .ETL file collected has the following events:


image

onnxruntime/core/session/inference_session.cc
```diff
@@ -508,87 +507,45 @@ void InferenceSession::ConstructorCommon(const SessionOptions& session_options,
...
-            LOGS(*session_logger_, VERBOSE) << "Done Removing ETW Sink from logger";
-          }
-        }
-      });
 
   // Register callback for ETW capture state (rundown)
-  etwRegistrationManager.RegisterInternalCallback(callback_ETWSink_provider_);
-
+  etwRegistrationManager.RegisterInternalCallback(callback_ETWSink_key_,
+                                                  [&etwRegistrationManager, this](
```
snnn, line 532: >this [](http://example.com/codeflow?start=76&length=4)

You should not need to register `this` pointer, because when the callback executes, the original session could have been already gone. 
The callback function should not bind to  any session.
The loggers can be the default logger.

onnxruntime/core/platform/windows/telemetry.h
```diff
@@ -68,17 +70,17 @@
...
+  static void UnregisterInternalCallback(const std::string& callback_key);
 
  private:
   static std::mutex mutex_;
   static uint32_t global_register_count_;
   static bool enabled_;
   static uint32_t projection_;
 
-  static std::vector<const EtwInternalCallback*> callbacks_;
+  static std::unordered_map<std::string, EtwInternalCallback> callbacks_;
```
snnn, line 83: >callbacks_ [](http://example.com/codeflow?start=62&length=10)

Because this is a global object, nobody can create any inference session before all such global objects are initialized. Nobody can destroy any session after "main" function is returned.  This is not a well known restriction, and it is a major source of crashes I see.

snnn:
>   callbacks_.clear();

The WindowsTelemetry obj usually is a global var.
This callbacks_ var is also a global var(it is not a member var of the WindowsTelemetry class)
Then, how do you know which one will get destroyed first?


---
Refers to: onnxruntime/core/platform/windows/telemetry.cc:93 in 6ab0185. [](commit_id = 6ab018548403b100ea2ff259d9ae2379bd0a6f2d, deletion_comment = False)

onnxruntime/core/platform/windows/logging/etw_sink.cc
```diff
@@ -182,10 +177,9 @@
...
 
   std::lock_guard<std::mutex> lock(callbacks_mutex_);
-  for (const auto& callback : callbacks_) {
-    if (callback != nullptr) {
-      (*callback)(SourceId, IsEnabled, Level, MatchAnyKeyword, MatchAllKeyword, FilterData, CallbackContext);
-    }
+  for (const auto& entry : callbacks_) {
+    const auto& cb = entry.second;
+    cb(SourceId, IsEnabled, Level, MatchAnyKeyword, MatchAllKeyword, FilterData, CallbackContext);
   }
```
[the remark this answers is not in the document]
yuslepukhin: Will discuss offline. There is not a deadlock here.

onnxruntime/core/session/inference_session.cc
snnn: Then what if:
1. create session 1, which will register a callback
2. create session 2, which will not register a callback as sizeof(active_sessions) > 1
3. destroy session 1, here you have to unregister the callback
Then there is a session alive but no callback is  working.

onnxruntime/core/platform/windows/telemetry.h
snnn: I was investigating another crash.

snnn (approved):

onnxruntime/core/platform/windows/logging/etw_sink.cc
```diff
@@ -129,30 +126,21 @@ void NTAPI EtwRegistrationManager::ORT_TL_EtwEnableCallback(
     _In_opt_ PEVENT_FILTER_DESCRIPTOR FilterData,
     _In_opt_ PVOID CallbackContext) {
   auto& manager = EtwRegistrationManager::Instance();
-  {
-    std::lock_guard<std::mutex> lock(manager.provider_change_mutex_);
-    manager.is_enabled_ = (IsEnabled != 0);
-    manager.level_ = Level;
-    manager.keyword_ = MatchAnyKeyword;
-  }
+  std::lock_guard<std::mutex> lock(manager.provider_change_mutex_);
```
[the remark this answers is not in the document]
yuslepukhin: I do not think this matters. The code is littered with mutexes that are not always applied properly. Supposedly, that mutex would allow atomic change of all the 3 properties, but getters are piece meals.

onnxruntime/core/platform/windows/logging/etw_sink.cc
```diff
@@ -129,30 +126,21 @@ void NTAPI EtwRegistrationManager::ORT_TL_EtwEnableCallback(
...
+  // Clarify: It is not clear at all why we need to assign the below 3 to the members
+  // had we passed them along on the stack, no mutex would be necessary.
```
skottmckay, line 131: Doesn't the mutex ensure all 3 values are updated in sync (vs. 2 concurrent requests and some get updated by the first and some the second)?

skottmckay:
Are there any tests we can add to prevent regression from the new behavior?

onnxruntime/core/framework/execution_provider.cc
```diff
@@ -37,4 +40,99 @@ common::Status IExecutionProvider::Compile(const std::vector<FusedNodeAndGraph>&
 }
 
 #endif
+
+ExecutionProviders::ExecutionProviders() {
+#ifdef _WIN32
+  // Register callback for ETW capture state (rundown)
+  etw_callback_key_ = "ExecutionProviders_rundown_";
```
snnn, line 47: Why do you move the code here?
ExecutionProviders is not a singleton. It can have multiple instances.

onnxruntime/core/framework/execution_provider.cc
```diff
@@ -37,4 +40,99 @@ common::Status IExecutionProvider::Compile(const std::vector<FusedNodeAndGraph>&
...
+  etw_callback_key_.append(std::to_string(reinterpret_cast<uintptr_t>(this)));
+  WindowsTelemetry::RegisterInternalCallback(
```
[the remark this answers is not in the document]
yuslepukhin: Each instance of ExecutionProviders would register a callback, this is currently not changed.
It would log all EPs that were added to the instance.
In theory, each instance can have different set of EPs.
Questions is, what should this be? @ivberg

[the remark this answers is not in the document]
ivberg: It should be that each EP logs it's options. Taking a trace with captureState rundown is an unusual diagnostic event so happens quite rarely

ivberg (approved):

snnn:
> TRACELOGGING_DEFINE_PROVIDER(telemetry_provider_handle, "Microsoft.ML.ONNXRuntime",

Perhaps it should use a different provider? One provider can only be registered once per process. 

---
Refers to: onnxruntime/core/providers/qnn/qnn_telemetry.cc:47 in e62225b. [](commit_id = e62225bbd774d8367d06561f5858e46f57fb57eb, deletion_comment = False)

snnn (approved):

onnxruntime/core/framework/execution_provider.cc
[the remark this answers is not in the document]
yuslepukhin: The issue here is that each session may have its own set of EPs, and those may have different options set for them. It is not clear, how to avoid it at this point.

snnn:
I just reviewed the telemetry code again last weekend. I found that the WindowsTelemetry object is a function local static(because it is part of the onnxruntime::Env object). So, it gets destructed earlier than any global object. And, users still need to be sure that they will not have any Inference Session alive before OrtEnv is dead(which happens pretty early).

yuslepukhin:
> I just reviewed the telemetry code again last weekend. I found that the WindowsTelemetry object is a function local static(because it is part of the onnxruntime::Env object). So, it gets destructed earlier than any global object. And, users still need to be sure that they will not have any Inference Session alive before OrtEnv is dead(which happens pretty early).

Yes, I am aware of WindowsTelemetry being a member of OrtEnv. This was factored into this PR.

PR #24891, opened 2025-05-28
faster shut down

snnn:
This PR is to reduce the chance of having crashes when a process is shutting down. The main idea is: if we know the process is shutting down(or if we know the ORT DLL won't be reloaded), we do not need to run C++ object destructors. The approach is recommended by Windows SDK's official document and many Windows developers.  For example,  18 years ago Raymond Chen wrote a blog [The old-fashioned theory on how processes exit](https://devblogs.microsoft.com/oldnewthing/20070502-00/?p=27023). This is ORT's current behavior. Raymond Chen also wrote a blog [what a better approach is](https://devblogs.microsoft.com/oldnewthing/20120105-00/?p=8683)

In our case, when onnxruntime is built as a python package, the DLL(onnxruntime_pybind11_state.pyd ) will never be manually unloaded.  Same on Linux.  Python does not unload the DLLs on exit.  Therefore,  we do not need to worry about the potential memory leaks caused by any global variables.  Therefore we do not need to call OrtEnv's destructors, and we do not need to unload any EP dlls. 

In most cases, people do not unload DLLs on Windows. And, on Linux it is even more complicated because GCC needs to maintain a unique table to avoid  odr violations, and this feature makes most C++ shared library cannot be unloaded. 

So, this change detects if the os is Windows and if the process is shutdown when calling destructors. If yes, the destructor will do nothing. 

After this change on Windows in most cases OrtEnv will not be destroyed. The only exception is: if someone manually load the DLL and manually unload the DLL, and also do not have a global threadpool. Then I think the user is an advanced user and they should know that they need to destroy all inference session objects and the OrtEnv singleton object before unloading the DLL. Besides, if they have enabled global thread pool, the DLL won't be unloaded if they haven't shutdown the thread pool and delete the OrtEnv object. And, even if the user has manually loaded/unloaded the DLL, there would still be memory leaks(that are not related to this change).  It's hard to get 100% clean.


tianleiwu (approved):
LGTM

adrianlizarraga (approved):
"The building is being demolished. Don't bother sweeping the floor" 👍

onnxruntime/core/session/environment.cc
```diff
@@ -287,23 +287,23 @@ Status Environment::Initialize(std::unique_ptr<logging::LoggingManager> logging_
     // Register MemCpy schema;
 
     // These ops are internal-only, so register outside of onnx
-    static std::vector<std::string> all_fixed_size_types = []() {
-      std::vector<std::string> all_types;
+    {
```
skottmckay: Given these are just strings, what changes by moving them to being a class member?

onnxruntime/core/session/ort_env.cc
```diff
@@ -77,23 +79,43 @@ OrtEnv* OrtEnv::GetInstance(const OrtEnv::LoggingManagerConstructionInfo& lm_inf
...
+    --ref_count_;
+    if (ref_count_ == 0) {
+      if (!g_is_shutting_down.load(std::memory_order_acquire)) {
+        instance_to_delete = p_instance_;  // Point to the instance to be deleted.
+        p_instance_ = nullptr;             // Set the static instance pointer to nullptr under the lock.
+      } else {
+        // Process is shutting down, let it leak.
+        // p_instance_ remains as is (though ref_count_ is 0), future CreateEnv calls
+        // would increment ref_count_ on this "leaked" instance.
+        // This behavior matches the requirement to "just let the memory leak out".
```
skottmckay, line 111: Could we have used the original code with unique_ptr and called `release()` here instead?

onnxruntime/test/shared_lib/dlopen_main.cc
```diff
@@ -0,0 +1,183 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
```
skottmckay, line 3: Would be good to provide some context here as the reader possibly knows nothing about there being two different files in `test/shared_lib` that have `main()`. or maybe this file should be in a different directory.

e.g. explain that it's used in a separate test app called `onnxruntime_shared_lib_dlopen_test` which only includes this file.

orttraining/orttraining/python/orttraining_python_module.cc
```diff
@@ -123,13 +123,17 @@ bool GetProviderInstanceHash(const std::string& type,
   return false;
 }
 
-ORTTrainingPythonEnv::ORTTrainingPythonEnv() : ort_env_(GetEnv()) {
+ORTTrainingPythonEnv::ORTTrainingPythonEnv(OrtEnv* ort_env) : ort_env_(ort_env) {
```
skottmckay: nit: this should take a reference to advertise `ort_env` is required. It can convert to a pointer for the member as needed (e.g. class needs to be copyable).

```suggestion
ORTTrainingPythonEnv::ORTTrainingPythonEnv(OrtEnv& ort_env) : ort_env_(&ort_env) {
```

onnxruntime/core/session/environment.cc
snnn: They will not be destructed. The strings will not be freed due to this change.

onnxruntime/core/session/ort_env.cc
snnn: The original code code makes sure that the pointer is always freed , even if a user forgot to call ReleaseEnv. 
The new code does not guarantee that, and intentionally make the memory leak even if a user has called ReleaseEnv. 
While std::unique_ptr provides strong guarantee on preventing memory leak, it will make things more complicated in this case.

snnn: For example, after this change the python extension allocates OrtEnv in such a way:
```
OrtEnv* ort_env = new OrtEnv();
```
Then the OrtEnv::~OrtEnv() function will not be called.

orttraining/orttraining/python/orttraining_python_module.cc
snnn: Updated the code.

onnxruntime/test/shared_lib/dlopen_main.cc
snnn: Updated the code.

skottmckay (approved):

PR #24894, opened 2025-05-29
Support reading in-memory external data for tensor

shiyi9801:
### Description

TensorProto may have external data in existing memory buffer. If the 'location' field of the external data info is set to \*/\_ORT_MEM_ADDR\_/\*, the 'offset' field should contain the address of the memory containing the data.

This PR fixes the issue that DML EP fails to export the EP context model when it calls `ReadExternalDataForTensor()`.


shiyi9801:
@fdwr @skottmckay PTAL, thanks!
/cc @huningxin

skottmckay:
What's the use-case where you're hitting this issue? The special tag has limited internal usage so I'm wondering what the new scenario is where we need this change.

skottmckay:
Ah ok. I think this PR hit the same issue. https://github.com/microsoft/onnxruntime/pull/24740/

shiyi9801:
> Ah ok. I think this PR hit the same issue. #24740

I see, I'll abandon my PR and let that PR fix this issue then, thanks!

PR #24898, opened 2025-05-29
[webgpu] support intel subgroup matrix on matmul_nbits

xhcao:
The patch enables intel subgroup matrix on matmul_bits operator, and temporarily supports it on vulkan backend and xe-2lpg arch, we will extend the functions on more subgroup matrix configs and platforms.


xhcao:
1. The subgroup matrix feature is very relevant with hardware vendors and their architectures, they have different subgroup matrix configs, and different vendors and architectures have the different best subgroup config,
optimizing the algorithm on one hardware will easily hurt others hardwares, so we generate code separately for different vendors in the early stages of development.
2. The PR currently only supports intel xe-2lpg architecture on vulkan, and the subgroup matrix config is f16(8*16) x f16(16*16) = f32(8*16), we will extend the features when the dawn enables more configs.
3. The current performance on intel xe-2lpg architecture is ~20% slower than dp4a, and ~10% faster than non-dp4a.

@jchen10 @daijh PTAL, thanks.

onnxruntime/core/providers/webgpu/webgpu_context.cc
```diff
@@ -150,6 +148,11 @@ void WebGpuContext::Initialize(const WebGpuBufferCacheConfig& buffer_cache_confi
     for (size_t i = 0; i < supported_features.featureCount; i++) {
       device_features_.insert(supported_features.features[i]);
     }
+    // cache adapter info
+    if (DeviceHasFeature(wgpu::FeatureName::ChromiumExperimentalSubgroupMatrix)) {
```
fs-eire, line 152: is this feature always available in all platforms? (win/linux/mac/wasm)

fs-eire:
Need to merge to latest main branch to fix the CI pipeline issue.

onnxruntime/core/providers/webgpu/webgpu_context.cc
xhcao: Subgroup matrix (named `cooperative_matrix` in graphics drivers) feature exposes powerful hardware acceleration for matrix operations, the feature could be enabled on new GPU devices, which should have XMX hardware components.
The function has been standardized on vulkan, https://registry.khronos.org/vulkan/specs/latest/man/html/VK_KHR_cooperative_matrix.html, I think gpu vendors should already support the feature on vulkan.
D3D is developing this feature and is not ready. But a similar feature named `cooperative_vector` was said it's ready yesterday, https://devblogs.microsoft.com/directx/cooperative-vector/, which benefits phi decoding stage.
Before this PR, there is already a configuration ([8, 8] * [8, 8] = [8, 8]) float16 implementation based on metal.
The feature is based on GPU hardware, I think wasm doesn't support it.

xhcao:
Please do NOT merge it to upstream today, I have some optimization to merge it tomorrow, thanks.

sushraja-msft:
xhcao, does this work on windows. Would it be possible to share instructions on how to try vk backend with ort webgpu ?

xhcao:
> xhcao, does this work on windows. Would it be possible to share instructions on how to try vk backend with ort webgpu ?

@sushraja-msft The PR enables the feature on intel+"xe-2lpg" platform, on windows, you could build onnxruntime with below command,
`build.bat --config RelWithDebInfo --build_dir build/Rel --parallel --skip_submodule_sync --skip_tests --parallel --use_webgpu --build_shared_lib  --enable_pybind --build_wheel --cmake_extra_defines onnxruntime_BUILD_UNIT_TESTS=OFF --cmake_extra_defines onnxruntime_ENABLE_DAWN_BACKEND_VULKAN=ON --cmake_extra_defines onnxruntime_ENABLE_DAWN_BACKEND_D3D12=OFF --cmake_generator "Visual Studio 17 2022"`

But after upgrading dawn yesterday, the compilation of shader would fail.

If I output `subgroup_matrix_result` to global memory directly, or output it through shared memory, it all fails. I am not sure whether your implementation on metal could work correctly now. The error logs are shown as below,
`
Exception: WebGPU validation failed. Error while parsing WGSL: :131:38 error: 'subgroupMatrixStore' requires argument 1 to be uniform
        subgroupMatrixStore(&output, matrix_c_offset + subtile_id * m_dim * uniforms.N + 3 * n_dim, matC03, false, uniforms.N);
                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
:89:31 note: builtin 'local_idx' of 'main' may be non-uniform
        let subtile_id =  u32(local_idx / sg_size);
                              ^^^^^^^^^
 - While calling [Device].CreateShaderModule([ShaderModuleDescriptor]).
`
and
`
Exception: WebGPU validation failed. Error while parsing WGSL: :163:29 error: 'subgroupMatrixStore' requires argument 0 to be uniform
        subgroupMatrixStore(&scratch[subtile_id], 0, matC03, false, n_dim);
                            ^^^^^^^^^^^^^^^^^^^^
:101:31 note: builtin 'local_idx' of 'main' may be non-uniform
        let subtile_id =  u32(local_idx / sg_size);
                              ^^^^^^^^^
 - While calling [Device].CreateShaderModule([ShaderModuleDescriptor]).
`

onnxruntime/core/providers/webgpu/webgpu_context.cc
fs-eire: > AdapterPropertiesSubgroupMatrixConfigs

looks like we need to use `#if !defined(__wasm__)` to exclude related code to make the build working for wasm.

xhcao: Done

xhcao:
After tuning the work group size (128 -> 256), tile size of A and removing the tile shared memory of C, the performance is better (~20%) than dp4a code path on LNL on win+vulkan.
It is ready to review now, PTAL

onnxruntime/core/providers/webgpu/webgpu_context.h
```diff
@@ -81,6 +81,7 @@ class WebGpuContext final {
   const wgpu::AdapterInfo& AdapterInfo() const { return adapter_info_; }
   const wgpu::Limits& DeviceLimits() const { return device_limits_; }
   bool DeviceHasFeature(wgpu::FeatureName feature) const { return device_features_.find(feature) != device_features_.end(); }
+  const wgpu::AdapterPropertiesSubgroupMatrixConfigs& SubgroupMatrixConfigs() const { return subgroup_matrix_configs_; }
```
xhcao, line 85: when building wasm target, bots report below error,
webgpu_context.h:84:15: error: no type named 'AdapterPropertiesSubgroupMatrixConfigs' in namespace 'wgpu'; did you mean 'AdapterPropertiesSubgroups'?
@fs-eire Does building wasm use old dawn files? Do you know the root-cause?

xhcao:
@sushraja-msft Could help review this PR, thanks?

onnxruntime/core/providers/webgpu/webgpu_context.h
fs-eire: I don't think `AdapterPropertiesSubgroupMatrixConfigs` (or the equivalent) exists in web, so Dawn does not generate the corresponding types and it makes sense to me. should remove them in WebAssembly build

onnxruntime/core/providers/webgpu/webgpu_context.cc
fs-eire: this should also apply to webgpu_context.h

> > AdapterPropertiesSubgroupMatrixConfigs
> 
> looks like we need to use `#if !defined(__wasm__)` to exclude related code to make the build working for wasm.

onnxruntime/contrib_ops/webgpu/quantization/subgroup_matrix_matmul_nbits.cc
```diff
@@ -1,18 +1,187 @@
...
 namespace onnxruntime {
 namespace contrib {
 namespace webgpu {
 
-Status SubgroupMatrixMatMulNBitsProgram::GenerateShaderCode(ShaderHelper& shader) const {
-  shader.AddInput("input_a", ShaderUsage::UseUniform | ShaderUsage::UseIndicesTypeAlias | ShaderUsage::UseValueTypeAlias);
-  shader.AddInput("input_b", ShaderUsage::UseUniform);
-  shader.AddInput("scales_b", ShaderUsage::UseUniform);
-  shader.AddOutput("output", ShaderUsage::UseUniform | ShaderUsage::UseElementTypeAlias);
+// The elements' sequence must match the enum class SubgroupMatrixComponentType in webgpu_cpp.h
```
fs-eire: can use `static_assert` to ensure.

onnxruntime/contrib_ops/webgpu/quantization/subgroup_matrix_matmul_nbits.cc
```diff
@@ -1,18 +1,187 @@
...
+constexpr std::string_view ComponentTypeName[] = {"unknown", "f32", "f16", "u32", "i32"};
+
+// std::tuple<architecture, backendType, componentType, resultComponentType, M, N, K, subgroupMinSize, subgroupMaxSize>
+static const std::tuple<std::string_view, wgpu::BackendType, wgpu::SubgroupMatrixComponentType, wgpu::SubgroupMatrixComponentType,
+                        uint32_t, uint32_t, uint32_t, uint32_t, uint32_t>
+    intel_supported_subgroup_matrix_configs[] = {
+        {"xe-2lpg", wgpu::BackendType::Vulkan, wgpu::SubgroupMatrixComponentType::F16, wgpu::SubgroupMatrixComponentType::F16, 8, 16, 16, 16, 32},
+        {"xe-2lpg", wgpu::BackendType::Vulkan, wgpu::SubgroupMatrixComponentType::F16, wgpu::SubgroupMatrixComponentType::F32, 8, 16, 16, 16, 32}};
 
+bool IsSubgroupMatrixConfigSupportedOnIntel(onnxruntime::webgpu::ComputeContext& context, int32_t& config_index) {
```
fs-eire, line 54: it looks like this check can be done only once in initialization?

fs-eire: It may be just fine for this PR - I am thinking of a refactoring plan for feature detection, including a subgroup matrix helper and dp4a matrix helper. but it's not necessary to be a part of the current PR.

onnxruntime/core/providers/webgpu/webgpu_context.cc
xhcao: Done

onnxruntime/core/providers/webgpu/webgpu_context.h
xhcao: Done

onnxruntime/contrib_ops/webgpu/quantization/subgroup_matrix_matmul_nbits.cc
```diff
@@ -382,9 +418,13 @@ Status SubgroupMatrixMatMulNBitsProgram::GenerateShaderCode(ShaderHelper& shader
 
   if (!vendor_.compare("apple")) {
     return GenerateShaderCodeOnApple(shader, nbits_);
```
fs-eire, line 416: does `GenerateShaderCodeOnApple()` actually work in chrome on Apple?

xhcao: Maybe. Because the implementation on Apple is hard code, if `context.HasFeature(wgpu::FeatureName::ChromiumExperimentalSubgroupMatrix)` is true, it may avoid the checking in chromium.

onnxruntime/core/providers/webgpu/webgpu_context.cc
```diff
@@ -452,6 +457,9 @@ std::vector<const char*> WebGpuContext::GetEnabledAdapterToggles() const {
   constexpr const char* toggles[] = {
       "use_dxc",
       "allow_unsafe_apis",
+#if defined(DAWN_ENABLE_VULKAN)
+      "use_vulkan_memory_model",
+#endif
```
fs-eire, line 462: what will this toggle do if the backend is D3D12 or Metal?

fs-eire:
It looks like all code path will not run subgroup matrix in wasm. (correct me if I'm wrong)

maybe simple exclude subgroup_matrix_matmul_nbits.cc and subgroup_matrix_matmul_nbits.h from web assembly build is easier to resolve the wasm build?

onnxruntime/core/providers/webgpu/webgpu_context.cc
xhcao: The toggle is only useful on Vulkan. https://source.chromium.org/chromium/chromium/src/+/main:third_party/dawn/src/dawn/native/Toggles.cpp;l=643
https://issues.chromium.org/issues/392606604

xhcao:
> It looks like all code path will not run subgroup matrix in wasm. (correct me if I'm wrong)
> 
> maybe simple exclude subgroup_matrix_matmul_nbits.cc and subgroup_matrix_matmul_nbits.h from web assembly build is easier to resolve the wasm build?

You are right. Done. Thanks.

xhcao:
Sorry. It was my fault to exclude one more code for wasm target, I had built wasm target on local machine, but not found the failure.
Is the CUDA Bot failure related with my PR?

PR #18989, opened 2024-01-03
[js/webgpu] Support capture and replay for jsep

qjia7:
### Description
This PR expands the graph capture capability to JS EP, which is similar to #16081. But for JS EP, we don't use the CUDA Graph, instead, we records all gpu commands and replay them,  which removes most of the cpu overhead to avoid the the situation that gpu waiting for cpu.

mobilenetv2-12 becomes 3.7ms from 6ms on NV 3090 and becomes 3.38ms from 4.58ms on Intel A770. 

All limitations are similar with CUDA EP:
1. Models with control-flow ops (i.e. If, Loop and Scan ops) are not supported.
2. Usage of graph capture is limited to models where-in all ops in the model can be partitioned to the JS EP or CPU EP and no memory copy between them.
3. Shapes of inputs/outputs cannot change across inference calls.
4. IObinding is required.

The usage is like below:
Method 1: specify outputs buffers explicitly.
```
    const sessionOptions = {
        executionProviders: [
          {
            name: "webgpu",
          },
        ],
        enableGraphCapture: true,
      };
    const session = await ort.InferenceSession.create('./models/mobilenetv2-12.onnx', sessionOptions);
   
    // prepare the inputBuffer/outputBuffer
    ... ...

   const feeds = {
       'input': ort.Tensor.fromGpuBuffer(inputBuffer, { dataType: 'float32', dims })
   };

   const fetches = {
       'output': ort.Tensor.fromGpuBuffer(outputBuffer, { dataType: 'float32', dims: [1, 1000] })
   };

   let results = await session.run(feeds, fetches);  // The first run will begin to capture the graph.

   // update inputBuffer content
  ... ...
   results = = await session.run(feeds, fetches);  // The 2ed run and after will directly call replay to execute the graph.

  ... ...
   session.release();
```
Method 2: Don't specify outputs buffers explicitly. Internally, when graph capture is enabled, it will set all outputs location to 'gpu-buffer'.
```
    const sessionOptions = {
        executionProviders: [
          {
            name: "webgpu",
          },
        ],
        enableGraphCapture: true,
      };
    const session = await ort.InferenceSession.create('./models/mobilenetv2-12.onnx', sessionOptions);

    // prepare the inputBuffer
    ... ...

   const feeds = {
       'input': ort.Tensor.fromGpuBuffer(inputBuffer, { dataType: 'float32', dims })
   };

   let results = await session.run(feeds);  // The first run will begin to capture the graph.
   
   // update inputBuffer content
  ... ...
[...]

PR #24900, opened 2025-05-29
[webgpu] Enable graph capture

qjia7:
This PR enables graph capture capabilities in the WebGPU provider, which is similar with jsep one #18989. 

All limitations are similar with JS/CUDA EP:
1. Models with control-flow ops (i.e. If, Loop and Scan ops) are not supported.
2. Usage of graph capture is limited to models where-in all ops in the model can be partitioned to the WebGPU EP or CPU EP and no memory copy between them.
3. Shapes of inputs/outputs cannot change across inference calls.
4. IOBinding is required. And all inputs/outputs are pre-allocated gpu buffers.

### DIFF
--- a/onnxruntime/core/providers/webgpu/buffer_manager.cc
+++ b/onnxruntime/core/providers/webgpu/buffer_manager.cc
@@ -250,6 +250,7 @@ class BucketCacheManager : public IBufferCacheManager {
           wgpuBufferRelease(buffer);
         }
       } else {
+        // TODO: Reuse the captured buffers for storage buffer to reduce peak memory.
         auto it = captured_buffers_.find(session_id);
         if (it == captured_buffers_.end()) {
           captured_buffers_.emplace(session_id, std::vector<WGPUBuffer>());
--- a/onnxruntime/core/providers/webgpu/buffer_manager.cc
+++ b/onnxruntime/core/providers/webgpu/buffer_manager.cc
@@ -37,7 +37,11 @@ class DisabledCacheManager : public IBufferCacheManager {
     wgpuBufferRelease(buffer);
   }
 
-  void OnRefresh() override {
+  void ReleaseCapturedBuffers(uint32_t /*session_id*/) override {
+    // no-op
+  }
+
+  void OnRefresh(SStatus /*session_status*/, uint32_t /*session_id*/) override {
     // no-op
   }
 };
@@ -59,7 +63,11 @@ class LazyReleaseCacheManager : public IBufferCacheManager {
     pending_buffers_.emplace_back(buffer);
   }
 
-  void OnRefresh() override {
+  void ReleaseCapturedBuffers(uint32_t /*session_id*/) override {
+    // no-op
+  }
+
+  void OnRefresh(SStatus /*session_status*/, uint32_t /*session_id*/) override {
     Release();
     pending_buffers_.clear();
   }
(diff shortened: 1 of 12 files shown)


guschmue:
wow, that was fast :)

onnxruntime/core/providers/webgpu/buffer_manager.cc
```diff
@@ -196,17 +229,33 @@ class BucketCacheManager : public IBufferCacheManager {
...
+      if (session_status == SStatus::Default) {
+        auto it = buckets_.find(buffer_size);
+        if (it != buckets_.end() && it->second.size() < buckets_limit_[buffer_size]) {
+          it->second.emplace_back(buffer);
+        } else {
+          wgpuBufferRelease(buffer);
+        }
       } else {
-        wgpuBufferRelease(buffer);
+        // TODO: Reuse the captured buffers for storage buffer to reduce peak memory.
```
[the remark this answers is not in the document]
qjia7: Added the `Graph` and `GraphSimple` mode for storage and uniform buffers when `enableGraphCapture` is enabled.

onnxruntime/core/providers/webgpu/buffer_manager.h
```diff
@@ -14,24 +14,34 @@ namespace webgpu {
 
 class WebGpuContext;
 
+// For command capture and replay
+enum class SStatus {
```
fs-eire: "SStatus" looks strange. maybe `SessionCapturingState` with `NotStarted`, `Capturing`, `Captured`?

onnxruntime/core/session/inference_session.cc
```diff
@@ -1906,6 +1906,10 @@ common::Status InferenceSession::Initialize() {
     }
 
     for (auto& ep : execution_providers_) {
+      auto end_status = ep->OnSessionInitializationStart(session_id_);
+      if (status.IsOK()) {
+        status = end_status;
+      }
```
fs-eire: seems the status is not used. change to use ORT_RETURN_IF_ERROR_SESSIONID_ or make the function return void

onnxruntime/core/providers/webgpu/buffer_manager.h
qjia7: Rename it to `SessionStatus` and keep the original enum value. The `session_status_` in webgpu_context is the current active session's status. Only if the current session is capturing or replaying, the `session_status_` is `Capturing` or `Replaying`. Otherwise, `session_status_` should be always in `Default` mode.

guschmue:
tried it out, seems to work for models like mobilenet-v4. I see ~20% gain in my setup.
Kind of unfortunate in ort that the capture hangs on the session and not a partition.
Looking if we can get a llm running with capture.

guschmue (approved):

guschmue:
there are a few ops that are partition to cpu. All because of int64 - can try to change the model to use int32.

qjia7:
@guschmue @fs-eire Rebased the code to latest. Please take another look, thanks.

onnxruntime/core/providers/webgpu/buffer_manager.cc
```diff
@@ -37,7 +37,7 @@ class DisabledCacheManager : public IBufferCacheManager {
     wgpuBufferRelease(buffer);
   }
 
-  void OnRefresh() override {
+  void OnRefresh(SessionState /*session_status*/) override {
```
fs-eire: should use const reference

qjia7: Done

PR #23294, opened 2025-01-08
[TensorRT] Support Multiple EP Context 

jingyanwangms:
### Description

- Use CreateEpContextModel from graph_partitioner.cc to save model with context ep. Now multi ep context in a model is supported
- Updated merging ep context related options from session option and tensorrt option
- Updated and adding unit test

Supported scenarios:
- Save/run **static single** ep context node using engine cache
- Save/run **static single** ep context node with embedded ep context info
- Save/run **static multiple** ep context node using engine cache
- Save/run **static multiple** ep context node with embedded ep context info
- Save/run **dynamic multiple** ep context node using engine cache
- Save/run **dynamic multiple** ep context node with embedded ep context info

Unsupported scenarios:
- Subsequent runs with embedded dynamic input ep context node where dynamic input dimension changed
This does not work because tensorrt engine might be updated during run time because of input size change but ORT does not have a call back mechanism to call CreateEpContextModel to update embedded ep context. Supporting this will require significant changes in the existing infrastructure.

PR #24901, opened 2025-05-29
[TRT RTX EP] Implement GetEPContextNodes()

thevishalagarwal:
Implements GetEPContextNodes()

### DIFF
--- a/onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.h
+++ b/onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.h
@@ -231,6 +231,8 @@ class NvExecutionProvider : public IExecutionProvider {
                                     bool serialize_refitted_engine,
                                     bool detailed_build_log);
 
+  const InlinedVector<const Node*> GetEpContextNodes() const override;
+
  private:
   mutable NvExecutionProviderInfo info_;
   bool external_stream_ = false;
@@ -266,6 +268,7 @@ class NvExecutionProvider : public IExecutionProvider {
   std::string cache_prefix_;
   std::string op_types_to_exclude_;
   int nv_profile_index_ = 0;
+  std::vector<std::unique_ptr<onnxruntime::Model>> ep_context_nodes_;
 
   // The format is as for TENSORRT_VERSION: (MAJOR * 100 + MINOR) * 100 + PATCH
   int32_t trt_version_;
--- a/onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.h
+++ b/onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.h
@@ -268,7 +268,7 @@ class NvExecutionProvider : public IExecutionProvider {
   std::string cache_prefix_;
   std::string op_types_to_exclude_;
   int nv_profile_index_ = 0;
-  std::vector<std::unique_ptr<onnxruntime::Model>> ep_context_nodes_;
+  std::vector<std::unique_ptr<onnxruntime::Model>> ep_context_models_;
 
   // The format is as for TENSORRT_VERSION: (MAJOR * 100 + MINOR) * 100 + PATCH
   int32_t trt_version_;
--- a/onnxruntime/core/providers/nv_tensorrt_rtx/onnx_ctx_model_helper.cc
+++ b/onnxruntime/core/providers/nv_tensorrt_rtx/onnx_ctx_model_helper.cc
@@ -73,7 +73,7 @@ ONNX_NAMESPACE::ModelProto* CreateCtxModel(const GraphViewer& graph_viewer,
                                            const std::string compute_capability,
                                            const std::string onnx_model_path,
                                            const logging::Logger* logger,
-                                           std::vector<std::unique_ptr<onnxruntime::Model>>& ep_context_nodes, const std::string& node_name) {
+                                           std::vector<std::unique_ptr<onnxruntime::Model>>& ep_context_models, const std::string& ep_context_node_name) {
   auto model_build = graph_viewer.CreateModel(*logger);
   auto& graph_build = model_build->MainGraph();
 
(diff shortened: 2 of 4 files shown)


onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.h
```diff
@@ -266,6 +268,7 @@ class NvExecutionProvider : public IExecutionProvider {
   std::string cache_prefix_;
   std::string op_types_to_exclude_;
   int nv_profile_index_ = 0;
+  std::vector<std::unique_ptr<onnxruntime::Model>> ep_context_models_;
```
HectorSVC, line 271: >ep_context_models_ [](http://example.com/codeflow?start=51&length=18)

The model instance here is just help to hold the EPContext nodes. Keep one instance is enough, why need a list of models?

onnxruntime/core/providers/nv_tensorrt_rtx/onnx_ctx_model_helper.cc
```diff
@@ -72,7 +72,8 @@ ONNX_NAMESPACE::ModelProto* CreateCtxModel(const GraphViewer& graph_viewer,
                                            const int64_t embed_mode,
                                            const std::string compute_capability,
                                            const std::string onnx_model_path,
-                                           const logging::Logger* logger) {
+                                           const logging::Logger* logger,
+                                           std::vector<std::unique_ptr<onnxruntime::Model>>& ep_context_models, const std::string& ep_context_node_name) {
   auto model_build = graph_viewer.CreateModel(*logger);
```
HectorSVC: >auto model_build = graph_viewer.CreateModel(*logger); [](http://example.com/codeflow?start=2&length=53)

Seems wired to me. Why create a model from a existing graph_viewer? how the lifecycle is controlled with the existing graph_viewer? ep_context_model instance need to be valid as long as the EP instance is alive.
All EP need to do with GetEpContextNodes() is, the EP creates the model instance to hold all EPContext nodes for the graph partitioner to query. EP only need to add all EPContext nodes into that model instance.
[https://github.com/microsoft/onnxruntime/blob/9705b17ed15d48b6b02c9578209e80182a90ebc3/onnxruntime/core/providers/qnn/qnn_execution_provider.cc#L1179-L1191](https://github.com/microsoft/onnxruntime/blob/9705b17ed15d48b6b02c9578209e80182a90ebc3/onnxruntime/core/providers/qnn/qnn_execution_provider.cc#L1179-L1191)

HectorSVC:
There was a fix for the Web CI pipeline, please merge the code from latest main branch.

onnxruntime/core/providers/nv_tensorrt_rtx/onnx_ctx_model_helper.cc
chilo-ms: The name of the `CreateCtxModel()` might be a bit confusing. 
It has some historic reasons

- At the beginning of TRT EP implementing EP Context feature, it didn't follow the rule to implement `GetEpContextNodes()`
- it only supported a model only containing a single EP Context node, meaning the whole graph can be run by TRT. There is no partitioning.  
- It didn't implement EP API's `GetEpContextNodes()`, instead it directly creates the ONNX model, so that's why this function has this name to really create a onnx model and dump to file.

chilo-ms: Right now, since RTX EP and TRT EP are going to follow EP Context implementation rule to implement `GetEpContextNodes()`, we could rename it to avoid confusion.

chilo-ms:
When running with TRT RTX EP, will it also handle the case where model contains contrib ops that will be fallback to run on CUDA EP or CPU? I assume it will.

If that's the case, following change need to be added to this PR as well.

- In GetCapability, it should be able the correctly parse the EP Context model that contains some EP Context nodes and some onnx nodes.

onnxruntime/core/providers/nv_tensorrt_rtx/onnx_ctx_model_helper.cc
gedoensmax: Sharing some source between TRT and TRT RTX seems like a good idea but will be kind of awkward. Any opinion on this ? Otherwise let's duplicate changes and source for now.

gedoensmax:
If we will be filling the `source` section on an EP Context node will the partitioner automatically force a node onto an EP ? I think if there are EP Context nodes there is no longer a real partitioning needed 🤔

onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.h
thevishalagarwal: if there are multiple EPContext nodes, each supported subgraph is compiled separately to a EPContext node from the `::Compile()` method unlike QNN where the entire graph is compiled at once

thevishalagarwal:
> In GetCapability, it should be able the correctly parse the EP Context model that contains some EP Context nodes and some onnx nodes.

@chilo-ms Can you please explain how to handle this in GetCapability? Should it just return Compute Capability of EPContext nodes or merge results from EPContext and regular onnx nodes?

onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.h
HectorSVC: No, QNN also supports partitioned graphs which will be compiled into multiple EPContext nodes. One model instance to hold all EPContext nodes is enough.

chilo-ms:
> > In GetCapability, it should be able the correctly parse the EP Context model that contains some EP Context nodes and some onnx nodes.
> 
> @chilo-ms Can you please explain how to handle this in GetCapability? Should it just return Compute Capability of EPContext nodes or merge results from EPContext and regular onnx nodes?

We had a unfinished PR which adds support for multiple EP Context nodes as well as implements GetEPContextNodes().
You can see the code [here](https://github.com/microsoft/onnxruntime/pull/23294/files#r1914037913).

chilo-ms:
> If we will be filling the `source` section on an EP Context node will the partitioner automatically force a node onto an EP ? I think if there are EP Context nodes there is no longer a real partitioning needed 🤔

yeah, it's not a 'real' partitioning. 
But, in ORT's workflow, even with EP Context node, ORT still needs to call EP's GetCapability to decide whether to assign that EP Context node to EP.

onnxruntime/core/providers/nv_tensorrt_rtx/onnx_ctx_model_helper.h
```diff
@@ -20,21 +20,20 @@ static const std::string COMPUTE_CAPABILITY = "hardware_architecture";
 static const std::string ONNX_MODEL_FILENAME = "onnx_model_filename";
 static const std::string EPCONTEXT_OP_DOMAIN = "com.microsoft";
 static const std::string EPCONTEXT_WARNING =
-    "It's suggested to set the ORT graph optimization level to 0 and  \
-                                              make \"embed_mode\" to 0 (\"ep_cache_context\" is the cache path)\
-                                              for the best model loading time";
+    "It's suggested to set the ORT graph optimization level to 0 for the best performance";
```
gedoensmax, line 23: Not just for best performance. We want to circumvent that initializers get constant folded in the optimizer and therefore change their name.

PR #24904, opened 2025-05-29
[QNN-EP] Fuse ChannelShuffle pattern

qti-yuduo:
### Description
Fuse transposed channel shuffle pattern into QNN op -- ONNX does not have native ChannelShuffle op. 

### Motivation and Context
Improves performance on QNN EP.


HectorSVC (approved):
:shipit:

PR #24905, opened 2025-05-29
Delete ORT module code that was for pytorch training

snnn:
Delete ORT module code that was for pytorch training. The module does not have active development since last summer.  We will put our focus on on-device training instead, which typically do not use python or pytorch.


baijumeswani:
This PR also deletesall training relted python bindings (including for non ortmodule scenarios). After this change, users can no longer build python orttraining. Is this the intention behind this change?

snnn:
Now our main usages for on-device training is for mobile devices.

baijumeswani:
There are two parts to on-device training:

1. Build the training graph offline (ahead of time). This uses [python `onnxblock`](https://onnxruntime.ai/docs/get-started/training-on-device.html#the-offline-phase). This uses the shared training pybind file as ortmodule uses.
2. Run the training on the device. We have [Python](https://onnxruntime.ai/docs/api/python/on_device_training/training_api.html), [C/C++](https://onnxruntime.ai/docs/api/c/training_c_cpp_api.html), [C#](https://onnxruntime.ai/docs/api/csharp/api/Microsoft.ML.OnnxRuntime.TrainingSession.html), [Java ](https://onnxruntime.ai/docs/api/java/ai/onnxruntime/OrtTrainingSession.html), js APIs for this.

While Python APIs for on device training are not needed on most devices, it is essential for development/prototyping so developers can build confidence on running training on the device (after validating with Python).

snnn:
I think your concerns are valid. I will talk to you offline.

PR #24906, opened 2025-05-29
[QNN EP] Add Support for If Op using Graph Transformation

quic-ashwshan:
- Currently Onnx graph transform handles only static If Op cases.
  1. When If op input is static and both branches are also static
  2. When if Op input is statifc but branches are dynamic.
- Extend If Op handling to support dynamic condition and dynamic branches.
- Lower both subgraphs to the main graph and select output using ONNX Where op.
- Added unit tests in graph transformer to test dynamic if conditions and ensure that a Where Op is added

### Description
This change adds support for lowering ONNX If operations where both the condition and the branches are dynamic. Previously, the ORT transformation logic handled only two cases:

1. Fully static If nodes were folded to constants.
2. Semi-dynamic If nodes (static condition, dynamic branches) were partially lowered by selecting the active branch at compile time.

This change now includes the case -
When both the condition and branches are dynamic, both subgraphs are lowered into the main graph, and a Where op is used to select the correct output at runtime.

### Motivation and Context
1. Enables support for dynamic control flow in ONNX models where the condition and both branches of an If node are not statically known.
2. This also helps offload the op to the NPU for QNN backend.


HectorSVC:
There was a fix for the Web CI pipeline, please merge the code from latest main branch.

onnxruntime/core/optimizer/if_to_where_transformer.cc
```diff
@@ -0,0 +1,226 @@
...
+      }
+    }
+    if (graph_utils::IsConstantInitializer(graph, cond_arg->Name(), true))
+      continue;
+
+    // Extract the two subgraph branches
+    Graph* then_sub = if_node.GetMutableGraphAttribute("then_branch");
+    Graph* else_sub = if_node.GetMutableGraphAttribute("else_branch");
+    if (!then_sub || !else_sub) {
+      LOGS(logger, ERROR) << "If node missing subgraphs!";
```
HectorSVC: >ERROR [](http://example.com/codeflow?start=19&length=5)

We should not log error message here. Info log should be sufficient.
Optimization not applied if pattern not match, it does not mean something wrong with user model. Error log is used to cause user attention to take some action.
Info log should be sufficient to help developer to understand what happened, why the optimization rules not applied. #Closed

onnxruntime/core/optimizer/if_to_where_transformer.cc
```diff
@@ -0,0 +1,226 @@
...
+    std::vector<NodeArg*> then_outputs, else_outputs;
+    ORT_RETURN_IF_ERROR(InlineSubgraph(graph, *then_sub, then_map, then_outputs, logger));
+    ORT_RETURN_IF_ERROR(InlineSubgraph(graph, *else_sub, else_map, else_outputs, logger));
+
+    // Build one Where node per original If output
+    const auto& if_outputs = if_node.MutableOutputDefs();
+    size_t num_outputs = if_outputs.size();
+
+    if (then_outputs.size() != else_outputs.size() || then_outputs.size() != if_outputs.size()) {
+      LOGS(logger, ERROR) << "Mismatch in output sizes between then/else branches and If node.";
```
HectorSVC: >ERROR [](http://example.com/codeflow?start=19&length=5)

same, please change to Info #Closed

HectorSVC (dismissed):
:shipit:

onnxruntime/core/optimizer/if_to_where_transformer.cc
```diff
@@ -0,0 +1,226 @@
...
+using namespace ONNX_NAMESPACE;
+using namespace onnxruntime::common;
+namespace onnxruntime {
+
+static Status InlineSubgraph(
+    Graph& main_graph,
+    const Graph& subgraph,
+    std::unordered_map<std::string, NodeArg*>& name_to_nodearg,
+    std::vector<NodeArg*>& subgraph_outputs,
+    const logging::Logger& logger) {
```
skottmckay, line 17: Values in a subgraph can shadow outer scope values. Due to that you have to either check if that's the case or create unique names when inlining.

https://github.com/microsoft/onnxruntime/blob/03b22ffc42e740d08ea7873ebbe0f135190469bc/onnxruntime/core/graph/graph.cc#L5317 is used to inline function based nodes (which is essentially a node with a subgraph) and uses `function_utils::Specialize` to do that.

onnxruntime/core/optimizer/if_to_where_transformer.cc
```diff
@@ -0,0 +1,226 @@
...
+    subgraph_outputs.push_back(mapped_output);
+  }
+
+  return Status::OK();
+}
+
+static bool CanTransformIfToWhere(const Node& if_node) {
+  const auto& outputs = if_node.OutputDefs();
+  for (const auto* output : outputs) {
+    const auto* type_proto = output->TypeAsProto();
```
skottmckay, line 112: The rules for `If` outputs are pretty flexible: https://github.com/onnx/onnx/blob/main/docs/Operators.md#outputs-1---

Specifically:
>The `then_branch` and `else_branch` may produce tensors with the same element type and different shapes.

Does `Where` allow for these or do the output shapes from each branch need to match exactly for it to be used?

onnxruntime/core/optimizer/if_to_where_transformer.cc
```diff
@@ -0,0 +1,226 @@
...
+    Node& if_node = *it;
+    if (if_node.OpType() != "If")
+      continue;
+
+    if (!graph_utils::IsSupportedProvider(if_node, GetCompatibleExecutionProviders())) {
+      // If not Qnn Execution Provider
+      continue;
+    }
+
+    if (!CanTransformIfToWhere(if_node)) {
```
skottmckay, line 144: IIUC this means both branches must be executed, so there's a cost to the inlining. Should there be some logic to not inline if one or both of the branches is expensive to run?

onnxruntime/core/optimizer/if_to_where_transformer.cc
```diff
@@ -0,0 +1,226 @@
...
+  for (const auto* output : outputs) {
+    const auto* type_proto = output->TypeAsProto();
+    if (!type_proto) return false;
+
+    // If any output is optional, we skip the transformation
+    if (type_proto->value_case() == onnx::TypeProto::kOptionalType) {
+      return false;
+    }
+
+    // Optionally, ensure it's a tensor
```
skottmckay, line 120: `Where` doesn't support Sequence output either.

onnxruntime/core/optimizer/if_to_where_transformer.cc
```diff
@@ -0,0 +1,226 @@
...
+
+    if (!CanTransformIfToWhere(if_node)) {
+      LOGS(logger, VERBOSE) << "Skipping IfToWhere transformation for node " << if_node.Name()
+                            << " due to optional or unsupported output types.";
+      continue;
+    }
+
+    const NodeArg* cond_arg = if_node.InputDefs()[0];
+    if (!cond_arg)
+      continue;
```
skottmckay: `cond` is a required input so model load should have failed if it wasn't present.

It's required to be a `bool` so that check should also be unnecessary.

onnxruntime/core/optimizer/if_to_where_transformer.cc
```diff
@@ -0,0 +1,226 @@
...
+    if (graph_utils::IsConstantInitializer(graph, cond_arg->Name(), true))
+      continue;
+
+    // Extract the two subgraph branches
+    Graph* then_sub = if_node.GetMutableGraphAttribute("then_branch");
+    Graph* else_sub = if_node.GetMutableGraphAttribute("else_branch");
+    if (!then_sub || !else_sub) {
+      LOGS(logger, INFO) << "If node missing subgraphs!";
+      continue;
+    }
```
skottmckay: Both subgraphs are required inputs so shouldn't need this check.

onnxruntime/core/optimizer/if_to_where_transformer.cc
```diff
@@ -0,0 +1,226 @@
...
+    ORT_RETURN_IF_ERROR(InlineSubgraph(graph, *else_sub, else_map, else_outputs, logger));
+
+    // Build one Where node per original If output
+    const auto& if_outputs = if_node.MutableOutputDefs();
+    size_t num_outputs = if_outputs.size();
+
+    if (then_outputs.size() != else_outputs.size() || then_outputs.size() != if_outputs.size()) {
+      LOGS(logger, INFO) << "Mismatch in output sizes between then/else branches and If node.";
+      continue;
+    }
```
skottmckay: Should this be prior to any graph modifications from InlineSubgraph?

onnxruntime/core/optimizer/if_to_where_transformer.cc
quic-ashwshan: Will that be a case of inferring shapes at runtime? Since if we do not know the condition up until runtime and if the output of the branches has different rank, then the if op output shape is not determined till that point. And this is not supported by Qnn

onnxruntime/core/optimizer/if_to_where_transformer.cc
quic-ashwshan: We can possibly give an option to the user itself to enable the transformation if needed. Offloading to CPU vs running on NPU with all branches inlined.

onnxruntime/core/optimizer/if_to_where_transformer.cc
skottmckay: it's not just rank. individual dimensions could differ I believe. 

type/shape inferencing runs across the subgraph, so if possible the complete output shape should be known which would like you explicitly check.

quic-ashwshan:
@skottmckay, Is there a way to run the transformations in Level 1 itself only for the Qnn Execution Provider ?

onnxruntime/core/optimizer/if_to_where_transformer.cc
yuslepukhin: This seem to be a lot of duplicate code. There is already logic to inline if branch. Check ConstantFolding and graph.cc

quic-ashwshan: Yes, we have a function to inline one of the branches at a time. But here we are inlining both branches at the same time right. This is only taking care of the case where the input condition itself is dynamic.

onnxruntime/core/optimizer/if_to_where_transformer.cc
quic-ashwshan: Yes, although this is not supported by our Qnn kernels. We would need to have a hard check that the inputs going to the WhereOp/ElementWiseSelect are the same.

quic-ashwshan:
Will close this PR for now - There is a public PR which will enable Graph Transforms inside qnn provider. Will plug it back in there.

Issue #24107, opened 2025-03-19
[Feature Request] Support for MatMul with broadcasting (eg. 3Dx2D matmul) for Xnnpack execution provider


pl121lp:
### Describe the feature request

Currently the MatMul op can be mapped to XnnPack fully connected kernels. However, only 2Dx2D matrix multiply is supported.
Now, many models have Matmul layers where the input is a 3D tensor, which requires broadcasting to apply matmul 2D and generate the resulting 3D tensor.
Since this is not supported by the Xnnpack execution provider execution falls back on MLAS implementation which can handle this case. (   sgemm.cpp:MlasGemmPackB()  )
From what I gather it would be possible to add broadcasting functionality in the Xnnpack execution provider layer?

### Describe scenario use case

Support for models that have non 2Dx2D matmul to comply with ONNX spec for the MatMul operation.
See image of layer cutout from a model

![Image](https://github.com/user-attachments/assets/486267c2-125d-4b1b-8d95-bc97e94ec325)

PR #24908, opened 2025-05-30
Support activation broadcasting in XNNPACK Matmul

hariharans29:
### Description
1. Support activation broadcasting in XNNPACK Matmul
2. Fix a subtle bug when activations is 1-D
            Per the existing gating logic, 1-D activations were allowed but the batch being passed through did not account for it. The batch size passed in was always `a->Shape()[0]` which is actually passing in the reduction dimension (K). This is incorrect as for a 1-D activation input, a `1` is to be prepended to the shape which meant that we should have actually passed in `1` for the batch. This passed the relevant test but I think it would have written outside the bounds of the output buffer because of the non-unary batch being passed through.

### Motivation and Context
Resolve https://github.com/microsoft/onnxruntime/issues/24107


onnxruntime/core/providers/xnnpack/math/matmul.cc
```diff
@@ -41,9 +41,9 @@ bool MatMul::IsOnnxNodeSupported(const NodeUnit& node_unit, const GraphViewer& g
       break;
     }
 
-    if (A_shape == nullptr || A_shape->dim_size() > 2 ||
-        (A_shape->dim_size() == 2 && A_shape->dim(1).dim_value() == 0) ||
```
hariharans29, line 45: All the gating logic around a dim alue being 0 should be taken care of in Compute() because we bail out if the output buffer's size is going to be 0

onnxruntime/core/providers/xnnpack/math/matmul.cc
```diff
@@ -162,10 +162,30 @@ Status MatMul::Compute(OpKernelContext* ctx) const {
...
+    // (excluding the actual inner reduction dim)
+
+   for (int64_t i = 0; i < rank - 1; ++i)
+   {
+      batch *= a_dims[i];
+   }
+  }
+
   if (op_type_ == OpComputeType::op_compute_type_fp32) {
-    status = xnn_reshape_fully_connected_nc_f32(op0_.get(), a->Shape()[0], threadpool);
```
hariharans29, line 166: a->Shape()[0] is incorrect for 1-D activations

onnxruntime/core/providers/xnnpack/math/matmul.cc
```diff
@@ -162,10 +161,28 @@ Status MatMul::Compute(OpKernelContext* ctx) const {
...
+  const auto& a_dims = a->Shape();
+  int64_t rank = a_dims.NumDimensions();
+
+  if (rank == 2) {
+    batch = a_dims[0];
+  } else if (rank > 2) {
+    // Input 'A' is N-dimensional, the batch is made up of the product of the outermost dims
+    // (excluding the actual inner reduction dim)
+
+    for (int64_t i = 0; i < rank - 1; ++i) {
```
jslhcl, line 177: nit: This for loop is also valid for rank == 2 case

onnxruntime/core/providers/xnnpack/math/matmul.cc
```diff
@@ -41,9 +41,8 @@ bool MatMul::IsOnnxNodeSupported(const NodeUnit& node_unit, const GraphViewer& g
       break;
     }
 
-    if (A_shape == nullptr || A_shape->dim_size() > 2 ||
-        (A_shape->dim_size() == 2 && A_shape->dim(1).dim_value() == 0) ||
-        A_shape->dim(0).dim_value() == 0) {
+    // A must at-least be 1-D
+    if (A_shape == nullptr || A_shape->dim_size() < 1) {
```
jslhcl, line 45: The original logic would also break for A_shape->dim_size() > 2 case. Is that condition missed?

hariharans29: The original logic didn't support A_shape -> dim_size > 2. This PR enables it and fixes the bug for A_shape-> dim_size == 1.

onnxruntime/core/providers/xnnpack/math/matmul.cc
hariharans29: Sure. Just wanted to avoid the one-iteration loop for rank == 2 case (the most common use-case). But I do understand that the loop overhead is probably negligible. :)

PR #24909, opened 2025-05-30
[NV TensorRT RTX EP] misc fixes towards graphics interop

ankan-ban:
### Description
- don't use cuda runtime API to set the device when a stream is already provided.
- expose option to set limit on max shared memory TensorRT can use.

### DIFF
--- a/onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
+++ b/onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
@@ -912,6 +912,7 @@ Status BindKernelOutput(Ort::KernelContext& ctx,
 }
 
 NvExecutionProvider::PerThreadContext::PerThreadContext(OrtDevice::DeviceId device_id, bool has_user_compute_stream, cudaStream_t stream) {
+  // TODO: figure out if PerThreadContext is used at all. If not, just clean it up.
   if (has_user_compute_stream) {
     CUDA_CALL_THROW(cudaSetDevice(device_id));
     (void)(stream);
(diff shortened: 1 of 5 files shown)


ankan-ban:
@chilo-ms @ishwar-raut1 @gedoensmax for review.

ishwar-raut1 (approved):

onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
```diff
@@ -1046,8 +1047,16 @@ NvExecutionProvider::NvExecutionProvider(const NvExecutionProviderInfo& info)
...
+
   // TODO(maximlianm) remove this since we should be able to compile an AOT context file without GPU
-  CUDA_CALL_THROW(cudaSetDevice(device_id_));
+
+  if (!info.has_user_compute_stream) {
+    // If the app is passing in a compute stream, it already has initialized cuda and created a context.
+    // Calling cudaSetDevice() will set the default context in the current thread
+    // which may not be compatible with the stream created by the app.
+    CUDA_CALL_THROW(cudaSetDevice(device_id_));
+  }
```
gedoensmax, line 1058: I think this is needed for loading an engine to a specific GPU using TRT. Would it be allowed to retain the context of a selected device id ? 
https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__PRIMARY__CTX.html#group__CUDA__PRIMARY__CTX_1g9051f2d5c31501997a6cb0530290a300

ankan-ban: When the user has passed a compute stream, we expect a few things from them:
1. They already initialized CUDA on the right device (they would need to do this anyway before they create the stream).
2. They pass a stream created on the correct device
I think we need to clarify these expectations in the documentation.

chilo-ms:
Please help resolve the conflict

ankan-ban:
closing as the changes have been submitted as part of https://github.com/microsoft/onnxruntime/pull/25015

PR #24621, opened 2025-05-02
Add GenAI Chat Template Function Docs

sayanshaw24:

PR #24912, opened 2025-05-30
Update genai apidocs

natke:


docs/genai/api/cpp.md
```diff
@@ -0,0 +1,757 @@
...
+
+```cpp
+auto str = tokenizer->Decode(tokens, token_count);
+```
+
+---
+
+### ApplyChatTemplate
+
+Applies a chat template to messages and tools.
```
sayanshaw24, line 190: do we want to add more info on params like in this PR: https://github.com/microsoft/onnxruntime/pull/24621? or keep it as is (more descriptive in the former of course but easier for us to maintain the latter)

if we do, at least for certain core functions, like Encode/Decode or ApplyChatTemplate, we can use that PR as reference.

natke: Just thinking if get this one merged, then you can apply your changes on top, What do you think @sayanshaw24 ?

sayanshaw24: sounds good!

PR #24913, opened 2025-05-30
Update config spec

natke:


ajindal1 (dismissed):

fs-eire:
A few links are not available. should we fix them?

natke:
> A few links are not available. should we fix them?

Are you able to find the correct link for this 404? in build/web.html?

[Windows WebAssembly CI Pipeline](https://dev.azure.com/onnxruntime/onnxruntime/_build?definitionId=161&_a=summary

fs-eire:
> > A few links are not available. should we fix them?
> 
> Are you able to find the correct link for this 404? in build/web.html?
> 
> [Windows WebAssembly CI Pipeline](https://dev.azure.com/onnxruntime/onnxruntime/_build?definitionId=161&_a=summary

We now use https://github.com/microsoft/onnxruntime/actions/workflows/web.yml

docs/genai/reference/config.md
```diff
@@ -49,128 +46,490 @@ A configuration file called genai_config.json is generated automatically if the
...
+---
 
-* **_pad_token_**: The id of the padding token.
+### Config::Model
 
-* **_context_length_**: The maximum length of sequence that the model can process.
+Describes the model architecture, files, and tokenization.
 
+- **type**: *(string)*  
+  The type of model (e.g., `"phi3"`, `"llama"`, `"gpt"`, etc.).
```
kunal-vaishnavi: Can we add a note here that `decoder` is a reserved keyword that can be used for decoder-only LLMs whose model type names may not be registered with GenAI? It allows us to let future models run with GenAI that can be created with the model builder but whose model types aren't known yet inside GenAI (e.g. `llama*` or `phi*` where `*` is a new number).

docs/genai/reference/config.md
```diff
@@ -49,128 +46,490 @@ A configuration file called genai_config.json is generated automatically if the
...
+Describes the generation/search parameters.
+
+- **do_sample**: *(bool)*  
+  Whether to use sampling (top-k/top-p) or deterministic search.
+
+- **min_length**: *(int)*  
+  Minimum length of the generated sequence.
+
+- **max_length**: *(int)*  
+  Maximum length of the generated sequence.
```
kunal-vaishnavi: Can we add a note in this doc that clarifies the difference between `context_length` and `max_length`? We should also add a note about how `max_length` and `past_present_share_buffer` work together.

docs/genai/reference/config.md
```diff
@@ -7,40 +7,37 @@ grand_parent: Generate API (Preview)
 nav_order: 1
 ---
```
kunal-vaishnavi, line 8: Can we also update the ["building models" doc](https://github.com/microsoft/onnxruntime/blob/gh-pages/docs/genai/howto/build-model.md) to point directly to the [GitHub repo's README](https://github.com/microsoft/onnxruntime-genai/blob/main/src/python/py/models/README.md) instead of having two copies to maintain?

docs/genai/reference/config.md
natke: So use "decoder" as the type?

kunal-vaishnavi: Yes, `decoder` is the name of the special type. [Here](https://github.com/microsoft/onnxruntime-genai/blob/592b7cacd365ecdb140f9e9d5fe31b4c875f621c/src/models/model.cpp#L811) is the full list of types.

natke: Thanks! How does this type interact with the other types? Like, could I replace all of the other types with decoder and everything would still work correctly?

natke: Updated. Have a look at my changes and see if that looks right

kunal-vaishnavi: Most of the other types can be replaced by `decoder`. There are some exceptions, however, such as the Phi ones because they are used to identify if the model's KV cache needs to be reset at a certain threshold. Therefore, all of the types are listed there for completeness.

docs/genai/reference/config.md
```diff
@@ -106,7 +106,33 @@ Top-level configuration object.
...
+  - phi
+  - phimoe
+  - phi3
+  - phi3small
+  - qwen2
+  - qwen3
+
+  For decoder only LLMS that are split into a pipeline of models, use "decoder-pipeline".
+  
+  Other model types:
```
kunal-vaishnavi: Should we split this section up further by category (e.g. encoder-decoder for Whisper and Marian, multi-modal for Phi vision, Phi-4 multimodal, and Gemma-3 vision)?

docs/genai/reference/config.md
```diff
@@ -49,128 +46,516 @@ A configuration file called genai_config.json is generated automatically if the
...
+  For advanced models, a pipeline of sub-models can be specified, each with its own ONNX file, session options, and input/output mappings.
+
+---
+
+## Search combinations
 
-2. Greedy search
+1. **Beam search**
+   - `num_beams > 1`
+   - `do_sample = false`
```
baijumeswani, line 554: not sure if we should mention the relationship between beam search and past present buffer sharing.

```suggestion
   - `do_sample = false`
   - `past_present_share_buffer = false`
```

docs/genai/reference/config.md
natke: done

docs/genai/reference/config.md
natke: And the past present share buffer cannot be shared because the past values do not map neatly to the present ones?

baijumeswani: Because beam search needs to swap the order of the kv cache when num beams > 1 to keep the more relevant beams up front.
And swapping the order of the kv cache cannot be done when past present buffer sharing is enabled. So we implicitly will disable past present buffer sharing when num beams > 1.

docs/genai/howto/past-present-share-buffer.md
```diff
@@ -0,0 +1,40 @@
...
+This is represented in the following diagram
+
+![alt text](../../../images/past-present-share-buffer.png)
+
+The size of the KV cache is different depending on the value of the past present share buffer.
+
+
+## When past_present_share_buffer is true
+
+Past KV caches = Present KV caches = batch_size *  num_key_value_heads + max_length + head_size
```
kunal-vaishnavi:
```suggestion
Past KV caches = Present KV caches = batch_size * num_key_value_heads * max_length * head_size
```

docs/genai/howto/past-present-share-buffer.md
```diff
@@ -0,0 +1,40 @@
...
+
+Note that the size of the cache is largely determined the value of the max_length parameter.
+
+
+## When past_present_share_buffer is false
+
+Past KV caches = batch_size * num_key_value_heads * past_sequence_length * head_size
+
+Present KV caches = batch_size *  num_key_value_heads * past_sequence_length + 1 *  head_size
```
kunal-vaishnavi:
```suggestion
Present KV caches = batch_size *  num_key_value_heads * (past_sequence_length + 1) *  head_size
```

docs/genai/howto/past-present-share-buffer.md
natke: fixed

docs/genai/howto/past-present-share-buffer.md
natke: fixed

docs/genai/howto/past-present-share-buffer.md
```diff
@@ -0,0 +1,50 @@
...
+parent: How to
+grand_parent: Generate API (Preview)
+nav_order: 6
+---
+
+# How to configure the past present share buffer
+
+The past present share buffer is an optimization that can be used to save memory and processing time.
+
+When the share buffer is used, the past and present KV cache buffers point to the same memory block.
```
kunal-vaishnavi:
```suggestion
When buffer sharing is used, the past and present KV cache buffers point to the same memory block.
```

docs/genai/howto/past-present-share-buffer.md
```diff
@@ -0,0 +1,50 @@
...
+
+When the share buffer is not used, the present KV cache buffers are re-allocated before every forward pass of the model, and copied to the past KV cache buffers.
```
kunal-vaishnavi:
```suggestion
When buffer sharing is not used, the present KV cache buffers are re-allocated before every forward pass of the model and copied to the past KV cache buffers.
```

docs/genai/howto/past-present-share-buffer.md
```diff
@@ -0,0 +1,50 @@
...
+
+This is represented in the following diagram
+
+![alt text](../../../images/past-present-share-buffer.png)
+
+The size of the KV cache is different depending on the value of the past present share buffer.
```
kunal-vaishnavi:
```suggestion
The size of the KV cache depends on whether buffer sharing is enabled or disabled.
```

PR #24942, opened 2025-06-03
Skip `_tpause` call for `_M_ARM64EC` in spin_pause.cc.

edgchen1:
### Description

Skip `_tpause` call for `_M_ARM64EC` in spin_pause.cc.

### Motivation and Context

Fix link error on ARM64EC for `_tpause` unresolved external symbol.


edgchen1:
@omarhass47 Just want to check - is it fine to not use `_tpause` when building for ARM64EC?

PR #24951, opened 2025-06-04
[WebNN] Add validation for node input tensor ranks 

NingW101:
### Description
Add function of checking if all node input tensor ranks are supported by WebNN. Add check for binary op types, other special ops check will be added in later prs

### DIFF
--- a/onnxruntime/core/providers/webnn/builders/helper.cc
+++ b/onnxruntime/core/providers/webnn/builders/helper.cc
@@ -109,15 +109,16 @@ bool IsInputRankSupportedByOp(const Node& node, const emscripten::val& wnn_limit
     return false;
   }
 
-  for (const auto& input : it->second.inputs) {
-    const std::string_view webnn_op_type = it->second.opType;
+  const std::string_view webnn_op_type = it->second.opType;
+  const std::string webnn_op_type_str(webnn_op_type);
 
+  for (const auto& input : it->second.inputs) {
     if (static_cast<size_t>(input.index) >= node.InputDefs().size() || node.InputDefs()[input.index] == nullptr) {
-      LOGS(logger, VERBOSE) << "Operator type: [" << op_type
-                            << "], input index: [" << input.index
-                            << "], corresponding WebNN op type: " << webnn_op_type
-                            << ", WebNN input name " << input.name
-                            << "] is not found.";
+      LOGS(logger, VERBOSE) << "Input index [" << input.index
+                            << "] for operator type [" << op_type
+                            << "], corresponding WebNN op type [" << webnn_op_type
+                            << "], WebNN input name [" << input.name
+                            << "] is invalid.";
       return false;
     }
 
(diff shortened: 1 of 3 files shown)


NingW101:
@honry @fdwr @guschmue PTAL. Thanks so much!

onnxruntime/core/providers/webnn/builders/helper.cc
```diff
@@ -99,6 +99,70 @@ bool IsTensorShapeSupported(const NodeArg& node_arg, const std::string& parent_n
...
+  const std::string_view op_type = node.OpType();
+
+  auto it = op_inputs_map.find(op_type);
+  if (it == op_inputs_map.end()) {
+    LOGS(logger, VERBOSE) << "Operator type: [" << op_type << "] is not found in the op inputs map.";
+    return false;
+  }
+
+  for (const auto& input : it->second.inputs) {
+    const std::string_view webnn_op_type = it->second.opType;
```
fdwr: Appears this line doesn't change. Could move it outside the loop to line 111.

onnxruntime/core/providers/webnn/builders/helper.cc
```diff
@@ -99,6 +99,70 @@ bool IsTensorShapeSupported(const NodeArg& node_arg, const std::string& parent_n
...
+        wnn_limits[std::string(webnn_op_type)][std::string(input.name)].isUndefined()) {
+      LOGS(logger, VERBOSE) << "Operator type: [" << op_type
+                            << "], input index: [" << input.index
+                            << "], corresponding WebNN op type: " << webnn_op_type
+                            << ", WebNN input name " << input.name
+                            << " is not defined in wnn_limits.";
+      return false;
+    }
+
+    const auto& input_limits = wnn_limits[std::string(webnn_op_type)][std::string(input.name)];
```
fdwr: If we really need to wrap the `string_view` with a `string` just to perform a lookup, then better to do it exactly once outside the loop rather than multiple times (which could allocate memory). Ideally I'd like `wnn_limits` to take advantage of C++14's [heterogeneous lookups](https://www.cppstories.com/2021/heterogeneous-access-cpp20/), but the object comes from emscripten, and it seems to not support that sadly.

onnxruntime/core/providers/webnn/builders/helper.cc
```diff
@@ -99,6 +99,70 @@ bool IsTensorShapeSupported(const NodeArg& node_arg, const std::string& parent_n
...
+
+  for (const auto& input : it->second.inputs) {
+    const std::string_view webnn_op_type = it->second.opType;
+
+    if (static_cast<size_t>(input.index) >= node.InputDefs().size() || node.InputDefs()[input.index] == nullptr) {
+      LOGS(logger, VERBOSE) << "Operator type: [" << op_type
+                            << "], input index: [" << input.index
+                            << "], corresponding WebNN op type: " << webnn_op_type
+                            << ", WebNN input name " << input.name
+                            << "] is not found.";
```
fdwr: Error message copy pasta. It should say than the input index is invalid (or unsupported) rather than the operator is not found.

fdwr (commented):
Thanks Wang Ning.

NingW101:
@fdwr Thanks so much for your quick and helpful comments, I addressed them in updated commit, please take another look, thanks!

fdwr (approved):
👍

fdwr:
Awaiting @Honry's review, else LGTM.

Honry (approved):
LGTM, thanks!

guschmue (approved):

PR #24953, opened 2025-06-04
Allow FP16 math in flash attention

sushraja-msft:
### Description
This change restores back fp16 math based FlashAttention.

### Motivation and Context
Earlier we noticed quality issues with deepseek-r1 attributed to overflow of qk computation when performing math in fp16 precision. https://github.com/microsoft/onnxruntime/pull/24723, addressed it by promoting math to fp32 to avoid the precision issue. 

However the topic remained that, these models are trained with FP8 precision how is it that inferencing runs into precision issues with FP16 math? using FP32 math also resulted in slight performance degradation.

In this follow up investigation, one issue identified is that we multiply scale for gqa quite late in the computation. Scale is 0.088 for deepseek-r1. Multiplying scale upfront seems to prevent the overflow issues.

For now only the prefill shaders are updated to use this approach. Pending feedback on impact across models, the generation shader can also be restored to FP16 math.

### DIFF
--- a/onnxruntime/contrib_ops/webgpu/bert/flash_attention.cc
+++ b/onnxruntime/contrib_ops/webgpu/bert/flash_attention.cc
@@ -137,13 +137,18 @@ Status FlashAttentionProgram::GenerateShaderCode(ShaderHelper& shader) const {
   shader.AdditionalImplementation() << "const qkv_head_size: u32 = " << qkv_head_size_ << ";\n"
                                     << "const num_heads: u32 =" << qkv_num_heads_ << ";\n";
 
+  if (is_fp16_)  {
+      shader.AdditionalImplementation() << "const min_value : q_element_t = q_element_t(-65504.0);\n";
+  } else {
+      shader.AdditionalImplementation() << "const min_value = f32(-3.402823e+38f);\n";
+  }
+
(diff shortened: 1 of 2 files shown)


onnxruntime/contrib_ops/webgpu/bert/flash_attention.cc
```diff
@@ -143,7 +143,7 @@ Status FlashAttentionProgram::GenerateShaderCode(ShaderHelper& shader) const {
   const max_k_step: u32 = 16u;
   const vec_factor: u32 = 4u;
   const qkv_head_size_vec: u32 = qkv_head_size / vec_factor;
-  const min_value = f32(-3.402823e+38f);;
+  const min_value : q_element_t = q_element_t(-65504.0);
```
fs-eire: is q_element_t possible to be f32? if so, should we use -3.402823e+38 as min_value in this case

tianleiwu: Is it possible to make min value to be type dependent (different thresholds for fp32 and fp16)?

qjia7 (approved):

fs-eire (approved):

guschmue:
tested with deepseek-r1 - results are ok.

guschmue (approved):

PR #24956, opened 2025-06-04
[CoreML] Add boolean values to CoreML

carzh:
This pull request improves the CoreML Execution Provider in ONNX Runtime by enhancing support for the `Equal` operator, refining partitioning logic to filter out nodes with unsupported input/output types, and adding new tests for partitioning scenarios. The most important changes are grouped below.

**Operator support and registration:**

* Added support for the `Equal` operator in the CoreML Execution Provider, including its registration and mapping to the CoreML op type. This allows models using the `Equal` operator to be offloaded to CoreML when compatible. [[1]](diffhunk://#diff-a0661d7e419206887db4877b5486565c45428dc34376223f7f9e2c5e1c776304R157-R158) [[2]](diffhunk://#diff-c228c51566bf3b858b05120f54bd07418f760579a20c22beee1223c9a49cc9d6R48)
* Implemented `IsOpSupportedImpl` in `BinaryOpBuilder` to ensure that both `Max` and `Equal` operators are only supported when using ML Program, with appropriate logging and checks. [[1]](diffhunk://#diff-a0661d7e419206887db4877b5486565c45428dc34376223f7f9e2c5e1c776304R22-R24) [[2]](diffhunk://#diff-a0661d7e419206887db4877b5486565c45428dc34376223f7f9e2c5e1c776304R216-R229) [[3]](diffhunk://#diff-a0661d7e419206887db4877b5486565c45428dc34376223f7f9e2c5e1c776304L234-L237)

**Partitioning and type compatibility:**

* Introduced logic in `CoreMLExecutionProvider` to filter out partition edge nodes whose input/output types are not supported by CoreML (e.g., only float32, float16, int32, int64 are supported for ML Program). This includes new helper functions for type checking and partition filtering, ensuring only compatible subgraphs are assigned to CoreML. [[1]](diffhunk://#diff-aa1280a87a6f57da05bd4f3e4707e1faeecf9eafd8c7351701fadfe77e91af5dR22) [[2]](diffhunk://#diff-aa1280a87a6f57da05bd4f3e4707e1faeecf9eafd8c7351701fadfe77e91af5dR41-R197) [[3]](diffhunk://#diff-aa1280a87a6f57da05bd4f3e4707e1faeecf9eafd8c7351701fadfe77e91af5dL120-R283) [[4]](diffhunk://#diff-aa1280a87a6f57da05bd4f3e4707e1faeecf9eafd8c7351701fadfe77e91af5dL142-R303) [[5]](diffhunk://#diff-f33bed88bda0e069a918e90f452faa9e746b7445d869cfa9102b556122ed6f24R9) [[6]](diffhunk://#diff-f33bed88bda0e069a918e90f452faa9e746b7445d869cfa9102b556122ed6f24R41-R58)

**Testing and validation:**

* Added new tests (`PartitioningTwoNegsThenEqual`, `PartitioningWhereAddEqual`) to verify correct partitioning and node assignment when using models with the `Equal` operator and mixed-type edges. These tests ensure that only supported nodes are assigned to the CoreML EP and that the outputs are validated.
* Introduced new ONNX model generators (`two_negs_then_equal.py`, `where_add_equal.py`) for use in the above tests, covering cases with chained `Neg` and `Equal` ops and mixed control/data flow. [[1]](diffhunk://#diff-7db9dc9dbc62b82881b2452a6f51f0f900b865b311f250801d0a27d74780631dR1-R34) [[2]](diffhunk://#diff-2a519b990b95a286f5ef906c1331ed82d8575b47c489ba1ebdff2092a7a1e731R1-R63)


onnxruntime/core/providers/coreml/builders/impl/cast_op_builder.cc
```diff
@@ -88,32 +88,13 @@ bool CastOpBuilder::IsOpSupportedImpl(const Node& node, const OpBuilderInputPara
 
   const auto& prec_node = node.InputEdgesBegin()->GetNode();
```
edgchen1, line 100: why are we removing the argmax -> cast special case handling?

onnxruntime/core/providers/coreml/builders/impl/base_op_builder.cc
```diff
@@ -114,13 +114,11 @@ bool BaseOpBuilder::IsInputDtypeSupport(const Node& node, size_t idx,
     return true;
   }
```
edgchen1, line 85: why was `#if CAN_BUILD_COREML6_OR_LATER` removed?

if it's no longer needed, `input_params` does not need to be declared with `[[maybe_unused]]` anymore.

onnxruntime/core/providers/coreml/builders/impl/binary_op_builder.cc
```diff
@@ -235,6 +237,10 @@ bool BinaryOpBuilder::HasSupportedInputsImpl(const Node& node, const OpBuilderIn
     return false;
   }
 
+  if (node.OpType() == "Equal" && !input_params.create_mlprogram) {
```
edgchen1: for this, and the `node.OpType() == "Max"` check above, maybe it would be better to put them in `IsOpSupportedImpl` instead of `HasSupportedInputsImpl`. these checks aren't checking the node inputs.

onnxruntime/core/providers/coreml/coreml_execution_provider.cc
```diff
@@ -93,9 +242,12 @@ CoreMLExecutionProvider::GetCapability(const onnxruntime::GraphViewer& graph_vie
                                             nullptr,
                                             /*drop_constant_initializers*/ true);
 
-  const auto num_of_partitions = result.size();
+  std::vector<std::unique_ptr<ComputeCapability>> filtered_result =
+      FilterIncompatibleEdgeNodesFromPartitions(std::move(result), graph_viewer, logger);
```
edgchen1, line 279: could you please add some tests for this new functionality?

onnxruntime/core/providers/coreml/builders/impl/cast_op_builder.cc
carzh: Ah sorry -- I didn't see that it was for NeuralNetwork only -- I added it back in

onnxruntime/core/providers/coreml/builders/impl/base_op_builder.cc
carzh: Removed the maybe_unused annotation

For some reason, CAN_BUILD_COREML6_OR_LATER doesn't work properly on my machine so I usually remove it for the perf tests. I accidentally added it into this PR and forgot to clean it up

onnxruntime/core/providers/coreml/builders/impl/binary_op_builder.cc
carzh: Moved checks!

onnxruntime/core/providers/coreml/coreml_execution_provider.cc
carzh: added two tests but am open to adding more for different cases

onnxruntime/core/providers/coreml/coreml_execution_provider.cc
```diff
@@ -36,6 +37,154 @@ CoreMLExecutionProvider::CoreMLExecutionProvider(const CoreMLOptions& options)
...
 
+// Create a helper function for type checking
+bool CoreMLExecutionProvider::IsSupportedType(int32_t type) const {
+  return supported_input_output_types_.find(type) != supported_input_output_types_.end();
+}
+
+// After an input node is removed from the partition, we need to update the inputs of the partition.
+// We evaluate the outputs of the input nodes
+// For each output, if its consumer node is part of the partition, we add this output to the inputs of the partition
+void UpdateInputsSetAfterNodeRemoval(std::unordered_set<NodeIndex>& partition_nodes_set,
```
edgchen1:
```suggestion
void UpdateInputsSetAfterNodeRemoval(const std::unordered_set<NodeIndex>& partition_nodes_set,
```

onnxruntime/core/providers/coreml/coreml_execution_provider.cc
```diff
@@ -36,6 +37,154 @@ CoreMLExecutionProvider::CoreMLExecutionProvider(const CoreMLOptions& options)
...
+        newInputs.insert(output->Name());
+      }
+    }
+  }
+}
+
+// we want to add the inputs of the removed node to the outputs of the partition IF:
+// 1. the input is an output of a node that is part of the partition
+// 2. the input is not already in the outputs of the partition
+void UpdateOutputsSetAfterNodeRemoval(std::unordered_set<NodeIndex>& partition_nodes_set,
```
edgchen1:
```suggestion
void UpdateOutputsSetAfterNodeRemoval(const std::unordered_set<NodeIndex>& partition_nodes_set,
```

onnxruntime/test/providers/coreml/coreml_basic_test.cc
```diff
@@ -430,5 +430,113 @@ TEST(CoreMLExecutionProviderTest, TestModelCache) {
...
+  OrtValue X = CreateInputOrtValueOnCPU<float>(X_shape, X_data);
+
+  std::vector<int64_t> cond_shape = {2, 3};
+  std::vector<uint8_t> uint8_data = gen.Uniform<uint8_t>(cond_shape, 0, 1);
+  std::vector<char> bool_data(uint8_data.begin(), uint8_data.end());
+  OrtValue cond = CreateInputOrtValueOnCPU<bool>(cond_shape, gsl::make_span(reinterpret_cast<const bool*>(bool_data.data()), bool_data.size()));
+
+  // GTest suite complains if lambda that potentially throws assert failure is passed to another function
+  // also resulting error message is not very useful -- so we use a function to store the count of CoreML nodes
+  // and assert the value after the model is ran.
```
edgchen1: what is the error? I would think that an issue like this would make `EPVerificationParams::graph_verifier` not very usable.

alternatively, can we use `EXPECT` tests in the verifier?

onnxruntime/test/testdata/coreml_partition_tests/two_negs_then_equal.onnx
edgchen1, line 1: please also add a script that can be used to generate the test model.

onnxruntime/core/providers/coreml/coreml_execution_provider.cc
```diff
@@ -36,6 +37,154 @@ CoreMLExecutionProvider::CoreMLExecutionProvider(const CoreMLOptions& options)
...
+  bool modified;
+
+  do {
+    bool inputs_modified = ProcessIncompatibleNodes(graph_viewer, partition_nodes, meta_def, true, logger);
+    bool outputs_modified = ProcessIncompatibleNodes(graph_viewer, partition_nodes, meta_def, false, logger);
+
+    modified = inputs_modified || outputs_modified;
+    if (modified) {
+      partition.nodes = std::vector<NodeIndex>(partition_nodes.begin(), partition_nodes.end());
+    }
```
edgchen1: could we just update `partition.nodes` once after the loop?

onnxruntime/core/providers/coreml/coreml_execution_provider.h
```diff
@@ -37,5 +38,31 @@ class CoreMLExecutionProvider : public IExecutionProvider {
 
   // map of fused_node_name to compiled_coreml_model
   InlinedHashMap<std::string, std::unique_ptr<onnxruntime::coreml::Model>> coreml_models_;
+
+  // Supported input and output types for CoreML MLProgram EP
+  // CoreML EP supports a more limited set of input and output types for models
+  // within the model, more data types are supported.
+  const std::unordered_set<int64_t> supported_input_output_types_ = {ONNX_NAMESPACE::TensorProto_DataType_FLOAT,
```
edgchen1: this is only used in `IsSupportedType`, maybe just define it within that function? it can be a function-local static variable.

onnxruntime/core/providers/coreml/coreml_execution_provider.h
```diff
@@ -37,5 +38,31 @@ class CoreMLExecutionProvider : public IExecutionProvider {
...
+                                                                     ONNX_NAMESPACE::TensorProto_DataType_FLOAT16,
+                                                                     ONNX_NAMESPACE::TensorProto_DataType_INT32,
+                                                                     ONNX_NAMESPACE::TensorProto_DataType_INT64};
+
+  // Helper function to check if a type is supported by CoreML
+  bool IsSupportedType(int32_t type) const;
```
edgchen1: how about `IsSupportedInputOrOutputType`?

onnxruntime/core/providers/coreml/coreml_execution_provider.cc
```diff
@@ -36,6 +37,154 @@ CoreMLExecutionProvider::CoreMLExecutionProvider(const CoreMLOptions& options)
...
+bool CoreMLExecutionProvider::IsSupportedType(int32_t type) const {
+  return supported_input_output_types_.find(type) != supported_input_output_types_.end();
+}
+
+// After an input node is removed from the partition, we need to update the inputs of the partition.
+// We evaluate the outputs of the input nodes
+// For each output, if its consumer node is part of the partition, we add this output to the inputs of the partition
+void UpdateInputsSetAfterNodeRemoval(std::unordered_set<NodeIndex>& partition_nodes_set,
+                                     const Node* node,
+                                     std::unordered_set<std::string>& newInputs,
```
edgchen1: nit: naming convention - `new_inputs` vs `newInputs`. same thing for `newOutputs` below.

onnxruntime/core/providers/coreml/coreml_execution_provider.cc
```diff
@@ -36,6 +37,154 @@ CoreMLExecutionProvider::CoreMLExecutionProvider(const CoreMLOptions& options)
...
+                                                       std::unordered_set<NodeIndex>& partition_nodes,
+                                                       IndexedSubGraph::MetaDef* meta_def,
+                                                       const bool is_input,
+                                                       const logging::Logger& logger) const {
+  std::unordered_set<std::string> new_names;
+  bool update_meta_def = false;
+  const size_t original_partition_size = partition_nodes.size();
+
+  // Get the appropriate vector (inputs or outputs)
+  auto& names = is_input ? meta_def->inputs : meta_def->outputs;
```
edgchen1: "name" is kind of general. could it be more specific, like "io_name"?

onnxruntime/core/providers/coreml/coreml_execution_provider.cc
edgchen1: make the helper functions `static` or put them in `namespace {}` so they are only scoped to this file.

onnxruntime/test/providers/coreml/coreml_basic_test.cc
carzh: Ah, for some reason, I remembered the failing test giving a very confusing error message that was unrelated to coreml or the partitions -- I implemented it again to check what the error message was + gave false values so that the test would fail, but the error this time was very clear 😅😅 so I switched it back to assert statements inside the graph verifier function

onnxruntime/core/providers/coreml/coreml_execution_provider.cc
```diff
@@ -36,6 +37,163 @@ CoreMLExecutionProvider::CoreMLExecutionProvider(const CoreMLOptions& options)
...
+void UpdateInputsSetAfterNodeRemoval(const std::unordered_set<NodeIndex>& partition_nodes_set,
+                                     const Node* node,
+                                     std::unordered_set<std::string>& new_inputs,
+                                     const onnxruntime::GraphViewer& graph_viewer) {
+  for (const auto& output : node->OutputDefs()) {
+    for (const auto& consumer_node : graph_viewer.GetConsumerNodes(output->Name())) {
+      NodeIndex consumer_node_index = consumer_node->Index();
+      if (partition_nodes_set.find(consumer_node_index) != partition_nodes_set.end() &&
+          new_inputs.find(output->Name()) == new_inputs.end()) {
+        new_inputs.insert(output->Name());
```
edgchen1, line 68: nit: we don't need to check if the set contains an element before inserting it. can save a lookup by just inserting it.

onnxruntime/core/providers/coreml/coreml_execution_provider.cc
```diff
@@ -36,6 +37,163 @@ CoreMLExecutionProvider::CoreMLExecutionProvider(const CoreMLOptions& options)
...
+// we want to add the inputs of the removed node to the outputs of the partition IF:
+// 1. the input is an output of a node that is part of the partition
+// 2. the input is not already in the outputs of the partition
+void UpdateOutputsSetAfterNodeRemoval(const std::unordered_set<NodeIndex>& partition_nodes_set,
+                                      const Node* node,
+                                      std::unordered_set<std::string>& new_outputs,
+                                      const onnxruntime::GraphViewer& graph_viewer) {
+  for (const auto& input : node->InputDefs()) {
+    const Node* producer_node_for_input = graph_viewer.GetProducerNode(input->Name());
+    NodeIndex producer_index = producer_node_for_input ? producer_node_for_input->Index() : NodeIndex(-1);
```
edgchen1, line 83: can we avoid looking up the index -1? maybe just check `producer_node_for_input != nullptr` first.

onnxruntime/core/providers/coreml/coreml_execution_provider.cc
```diff
@@ -36,6 +37,163 @@ CoreMLExecutionProvider::CoreMLExecutionProvider(const CoreMLOptions& options)
...
+    if (partition_nodes_set.find(producer_index) != partition_nodes_set.end() &&
+        new_outputs.find(input->Name()) == new_outputs.end()) {
+      new_outputs.insert(input->Name());
+    }
+  }
+}
+}  // namespace
+
+bool CoreMLExecutionProvider::ProcessIncompatibleNodes(const onnxruntime::GraphViewer& graph_viewer,
```
edgchen1, line 92: maybe a more descriptive name like "RemoveIncompatibleInputsOrOutputs"?

onnxruntime/core/providers/coreml/coreml_execution_provider.cc
```diff
@@ -36,6 +37,163 @@ CoreMLExecutionProvider::CoreMLExecutionProvider(const CoreMLOptions& options)
...
+                                << " from CoreML partition due to unsupported type: "
+                                << ONNX_NAMESPACE::TensorProto_DataType_Name(
+                                       (ONNX_NAMESPACE::TensorProto_DataType)type);
+        }
+      }
+    } else {
+      // Type is supported, check if nodes are in partition
+      for (const Node* node : nodes) {
+        if (node && partition_nodes.find(node->Index()) != partition_nodes.end()) {
+          new_names.insert(io_name);
```
edgchen1, line 146: what if a node in the partition has one supported input and one unsupported input? would the result be different depending on whether the supported input was handled first?

onnxruntime/core/providers/coreml/coreml_execution_provider.cc
```diff
@@ -36,6 +37,163 @@ CoreMLExecutionProvider::CoreMLExecutionProvider(const CoreMLOptions& options)
...
+  }
+}
+}  // namespace
+
+bool CoreMLExecutionProvider::ProcessIncompatibleNodes(const onnxruntime::GraphViewer& graph_viewer,
+                                                       std::unordered_set<NodeIndex>& partition_nodes,
+                                                       IndexedSubGraph::MetaDef* meta_def,
+                                                       const bool is_input,
+                                                       const logging::Logger& logger) const {
+  std::unordered_set<std::string> new_names;
```
edgchen1, line 97: how about `new_io_names`?

onnxruntime/core/providers/coreml/coreml_execution_provider.cc
```diff
@@ -36,6 +37,163 @@ CoreMLExecutionProvider::CoreMLExecutionProvider(const CoreMLOptions& options)
...
+
+  std::unordered_set<NodeIndex> partition_nodes(partition.nodes.begin(), partition.nodes.end());
+  bool modified;
+
+  do {
+    bool inputs_modified = ProcessIncompatibleNodes(graph_viewer, partition_nodes, meta_def, true, logger);
+    bool outputs_modified = ProcessIncompatibleNodes(graph_viewer, partition_nodes, meta_def, false, logger);
+
+    modified = inputs_modified || outputs_modified;
+  } while (modified);
```
edgchen1, line 176: it seems like this could be inefficient. e.g., if input processing and output processing don't modify the partition (i.e., return true) the same number of times, then some iterations will be unnecessary for the one that finishes earlier.

onnxruntime/core/providers/coreml/coreml_execution_provider.cc
```diff
@@ -36,6 +37,163 @@ CoreMLExecutionProvider::CoreMLExecutionProvider(const CoreMLOptions& options)
...
+  partition.nodes = std::vector<NodeIndex>(partition_nodes.begin(), partition_nodes.end());
+}
+
+// CoreML only supports a limited set of inputs and outputs (int32 and float32), so we remove all edge nodes with incompatible types
+std::vector<std::unique_ptr<ComputeCapability>> CoreMLExecutionProvider::FilterIncompatibleEdgeNodesFromPartitions(
+    std::vector<std::unique_ptr<ComputeCapability>>&& capabilities,
```
edgchen1, line 182: it may be more straightforward to take the `capabilities` by non-const reference and modify it. it looks unusual to take an rvalue reference and also return a modified copy.

you can filter out empty partitions after processing them. e.g.:
```c++
capabilities.erase(
    std::remove_if(capabilities.begin(), capabilities.end(),
                   [](const auto& capability) {
                      return capability->sub_graph->nodes.empty();
                   }),
    capabilities.end());
```

https://en.wikipedia.org/wiki/Erase%E2%80%93remove_idiom

onnxruntime/core/providers/coreml/coreml_execution_provider.cc
```diff
@@ -36,6 +37,163 @@ CoreMLExecutionProvider::CoreMLExecutionProvider(const CoreMLOptions& options)
...
+  do {
+    bool inputs_modified = ProcessIncompatibleNodes(graph_viewer, partition_nodes, meta_def, true, logger);
+    bool outputs_modified = ProcessIncompatibleNodes(graph_viewer, partition_nodes, meta_def, false, logger);
+
+    modified = inputs_modified || outputs_modified;
+  } while (modified);
+  partition.nodes = std::vector<NodeIndex>(partition_nodes.begin(), partition_nodes.end());
+}
+
+// CoreML only supports a limited set of inputs and outputs (int32 and float32), so we remove all edge nodes with incompatible types
```
edgchen1, line 180: nit: suggest not explicitly mentioning the supported types here. generally, the less we need to keep comments consistent the better.

onnxruntime/core/providers/coreml/coreml_execution_provider.cc
```diff
@@ -36,6 +37,163 @@ CoreMLExecutionProvider::CoreMLExecutionProvider(const CoreMLOptions& options)
 
 CoreMLExecutionProvider::~CoreMLExecutionProvider() {}
 
+// Create a helper function for type checking
+bool CoreMLExecutionProvider::IsSupportedInputOrOutputType(int32_t type) const {
+  // Supported input and output types for CoreML MLProgram EP
```
edgchen1, line 43: does this limitation only apply for MLProgram? if so, should we limit the partition filtering to that case?

onnxruntime/core/providers/coreml/coreml_execution_provider.h
```diff
@@ -37,5 +38,23 @@ class CoreMLExecutionProvider : public IExecutionProvider {
 
   // map of fused_node_name to compiled_coreml_model
   InlinedHashMap<std::string, std::unique_ptr<onnxruntime::coreml::Model>> coreml_models_;
+
+  // Helper function to check if a type is supported by CoreML
+  bool IsSupportedInputOrOutputType(int32_t type) const;
```
edgchen1, line 43: nit: can we put the private functions before the data members?

yuslepukhin:
@copilot resolve the merge conflicts in this pull request

yuslepukhin:
1. ONNX Spec vs CoreML Spec Alignment
- Where: ONNX requires condition to be tensor(bool), and X/Y can be many types. CoreML's select also requires a bool condition. The PR correctly maps ONNX's Where to CoreML's select. However, the PR only validates inputs 1 and 2 (X, Y) via IsInputDtypeSupport but does not explicitly validate input 0 (condition) is bool. Since the ONNX spec already constrains this and the model should have been validated, this is acceptable but could be defensive.

- Equal: ONNX Equal accepts many numeric types and outputs tensor(bool). CoreML's equal supports the same semantics. The output being bool is an internal edge — handled correctly by the partition filtering.

- Cast with BOOL: CoreML MLProgram supports cast to/from bool. The PR adds bool to both input and output type checks — matches CoreML capabilities.

2. Test coverage Gaps:

- No standalone Where operator test that verifies the Where op runs correctly on CoreML with valid inputs (i.e., a model where Where can stay in the partition and produce correct output).
- No standalone Equal operator test verifying correct output on CoreML EP.
- No test for Cast with bool input/output.
- No negative test for Where with NeuralNetwork (non-MLProgram) mode being rejected.
- No test for the edge case edgchen1 raised (mixed supported/unsupported inputs on the same node).
- No test for broadcast behavior of Where (ONNX Where supports multidirectional broadcasting).

3. Additional Issues
- Lint warnings: The PR has several unresolved cpplint warnings:

- Missing #include <memory>, #include <utility>, #include <string>, #include <unordered_set>, #include <vector>
- using namespace directive in where_op_builder.cc line 32 (should use specific using declarations)
- where_op_builder.cc does not check broadcasting compatibility: The ONNX Where spec supports multidirectional broadcasting. The builder should either verify shapes are broadcast-compatible or document that CoreML handles this.

- NodeIndex(-1) sentinel value in UpdateOutputsSetAfterNodeRemoval (line 82): This is fragile. If std::unordered_set happens to contain NodeIndex(-1) (unlikely but possible with wrapping), this could produce incorrect results. edgchen1's suggestion to null-check first is the right fix.

PR #24957, opened 2025-06-04
Cherry pick telemetry changes from win-onnxruntime

ashrit-ms:
### Description
This change cherry-picks telemetry changes from win-onnxruntime to improve telemetry data collection for  ONNX Runtime on Windows.

### Motivation and Context
These changes are already present in win-onnxruntime, so cherry-picking these changes here for Windows use cases that rely on public ONNX Runtime.

### DIFF
--- a/onnxruntime/core/session/inference_session.cc
+++ b/onnxruntime/core/session/inference_session.cc
@@ -1673,6 +1673,7 @@ static bool ModelHasFP16Inputs(const Graph& graph) {
   return false;
 }
 
+#ifdef _WIN32
 [[maybe_unused]] static std::string ModelWeightDataType(const Graph& graph) {
   std::string data_type_list;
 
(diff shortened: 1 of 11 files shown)


onnxruntime/core/session/inference_session.cc
```diff
@@ -1673,6 +1673,131 @@ static bool ModelHasFP16Inputs(const Graph& graph) {
...
+  std::ostringstream hash_stream;
+  hash_stream << std::hex << final_hash;
+  return hash_stream.str();
+}
+#endif
+
+#ifdef _WIN32
+[[maybe_unused]] static std::string ComputeModelWeightHash(const InitializedTensorSet& initializers) {
+  std::size_t final_hash = 0;
+  const std::size_t node_hash_count = TelemetrySampleCount;
```
snnn, line 1767: Abseil has helpers that can make this code much simpler https://abseil.io/docs/cpp/guides/hash

ashrit-ms: Thanks for the suggestion, I will look into updating the implementation once we sync the private and public repos.

PR #24959, opened 2025-06-05
Support 8 bit weights unpacked mode compute in MatmulNBits kernel

hariharans29:
### Description

MLAS doesn't have optimzed kernel implementations for 8-bit weights to support all user requested configurations (as allowed per the `MatMulNBits` op kernel spec) For example, only `accuracy_level` of 4 is supported when weights are 8 bits. In this case, provide a "fallback" execution path like the one available for 4 bit weights.

This change: 
 
1. Adds an execution path in the "unpacked compute" mode for 8-bit weights in `MatmulNBits` kernel
 as MlasDequantize already supports 8 bit weights
          1.1 Hitting the non-MLAS de-quantization path which doesn't support 8 bits should be very rare and an enforce is added to guard against hitting that

2. Added logging to alert users that the non-optimal "unpacked compute" mode is being used in the `MatmulNBits` kernel 

### Motivation and Context
Prevent users from hitting errors if MLAS doesn't host optimized kernels for 8 bit weights

### DIFF
--- a/onnxruntime/contrib_ops/cpu/quantization/matmul_nbits.cc
+++ b/onnxruntime/contrib_ops/cpu/quantization/matmul_nbits.cc
@@ -752,9 +752,11 @@ Status MatMulNBits<T1>::Compute(OpKernelContext* ctx) const {
   }
 
   // TODO(hasesh): Should this logging level be warning ?
-  LOGS(ctx->Logger(), INFO) << MakeString("Using unpacked compute mode for the Matmul operation ",
+  LOGS(ctx->Logger(), INFO) << MakeString("Falling back to using unpacked compute mode for the Matmul operation ",
                                           "(i.e.) the weights will be de-quantized to fp32 before invoking "
-                                          "the fp32 Matmul operation");
+                                          "the fp32 Matmul kernel."
+                                          "This is because MLAS doesn't have an optimized quantized kernel "
+                                          "for the requested compute configuration.");
 
   return ComputeBUnpacked(a, b, scales, zero_points, reorder_idx, bias, y, allocator, thread_pool, helper);
 }
--- a/onnxruntime/contrib_ops/cpu/quantization/matmul_nbits.cc
+++ b/onnxruntime/contrib_ops/cpu/quantization/matmul_nbits.cc
@@ -7,6 +7,7 @@
 #include <type_traits>
 
 #include "core/common/common.h"
+#include "core/common/make_string.h"
 #include "core/common/narrow.h"
 #include "core/common/safeint.h"
 #include "core/framework/op_kernel.h"
@@ -440,8 +441,6 @@ Status MatMulNBits<float>::ComputeBUnpacked(const Tensor* a,
                                             AllocatorPtr& allocator,
                                             concurrency::ThreadPool* thread_pool,
                                             const MatMulComputeHelper& helper) const {
-  ORT_ENFORCE(nbits_ == 4, "Only 4b quantization is supported for unpacked compute.");
-
   const auto* a_data = a->Data<float>();
   const uint8_t* b_data = b->Data<uint8_t>();
   const auto* scales_data = scales->Data<float>();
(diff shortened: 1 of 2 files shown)


onnxruntime/contrib_ops/cpu/quantization/matmul_nbits.cc
```diff
@@ -460,19 +459,36 @@ Status MatMulNBits<float>::ComputeBUnpacked(const Tensor* a,
...
+          tmp_b_data_ptr.get(),                           // dequantized output
+          b_data,                                         // quantized input
+          scales_data,                                    // quantization scales
+          static_cast<const uint8_t*>(zero_points_data),  // quantization zero points
+          static_cast<int32_t>(block_size_),              // quantization block size
+          column_wise_quant_,                             // columnwise quantization or row-wise
+          static_cast<int32_t>(K_),                       // number of rows in quantized input
+          static_cast<int32_t>(N_),                       // number of columns in quantized input
+          thread_pool);
+    } else {  // If it isn't 4bit, it has to be 8-bit quantization
```
edgchen1, line 472: should we add `ORT_ENFORCE(nbits_ == 8)` to this branch?

onnxruntime/contrib_ops/cpu/quantization/matmul_nbits.cc
```diff
@@ -720,6 +752,13 @@ Status MatMulNBits<T1>::Compute(OpKernelContext* ctx) const {
     }
   }
 
+  // TODO(hasesh): Should this logging level be warning ?
+  LOGS(ctx->Logger(), INFO) << MakeString("Falling back to using unpacked compute mode for the Matmul operation ",
```
edgchen1: looks like we don't need to use MakeString() here.

I wonder if this will be logged too often if the model is run multiple times. with the frequency, I think WARNING is maybe too high. on the other hand, if we can just log it the first time, WARNING might be appropriate.

maybe it would also be good to log the node name.

hariharans29: Re. Node Name - Great idea, thanks!

You mean keep a mutable int to track the first call into `Compute()` ? Is there a better way of tracking the first call ?

edgchen1: hm, yeah having to track that state that doesn't seem great. IMO, it's maybe too verbose for WARNING level, but INFO's probably fine.

hariharans29: Yeah I think INFO is good for now. Although, there is a case to be made for WARNING as well. The users want to be using a quantized Matmul implementation which is why the op is `MatmulNBits` to begin with. The current fallback behavior is basically `fp32 Matmul + cost of de-quantization of the weights` which is a worthy enough degradation to warrant a WARNING given the expected poor performance with this approach. But, I agree that it will then get too noisy on every `Run()`. So, for now we can keep it as INFO and encourage users to turn on more verbose logging if they don't get the expected speed-up with `MatmulNBits` to see if it is using the fallback approach.

onnxruntime/contrib_ops/cpu/quantization/matmul_nbits.cc
```diff
@@ -720,6 +753,13 @@ Status MatMulNBits<T1>::Compute(OpKernelContext* ctx) const {
     }
   }
 
+  // TODO(hasesh): Should this logging level be warning ?
+  LOGS(ctx->Logger(), INFO) << "Falling back to using unpacked compute mode for the Matmul operation "
+                            << "(i.e.) the weights will be de-quantized to fp32 before invoking "
+                            << "the fp32 Matmul kernel."
+                            << "This is because MLAS doesn't have an optimized quantized kernel "
+                            << "for the requested compute configuration.";
```
edgchen1: nit: it can be a single string
```suggestion
  LOGS(ctx->Logger(), INFO) << "Falling back to using unpacked compute mode for the Matmul operation "
                               "(i.e.) the weights will be de-quantized to fp32 before invoking "
                               "the fp32 Matmul kernel."
                               "This is because MLAS doesn't have an optimized quantized kernel "
                               "for the requested compute configuration.";
```

edgchen1 (dismissed):

PR #24961, opened 2025-06-05
[ROCm EP]Add ROCm execution provider to excluded EP for test with Cuda EP

TedThemistokleous:
Need this since ROCm EP just hipifies the CUDA kernel's used for this. Will give false failures when in fact CUDA EP is doing the same thing

### Description

Adds ROCmExecutionprovider as additional exclude on test that CudaExecutionProvider is failing on. 

### Motivation and Context

Since ROCm EP performs a HIPIFY on the CUDA EP kernels to use the equivalent functionality in ROCm to achieve inference, any errors or gaps we find in the CUDA ep will translate accordingly to the ROCm EP. Need to add the exclude here to this newly added test


TedThemistokleous:
@Tianleiwu, this is a simple one line change. Just adds ROCm EP to the additional Exclude for this test case

tianleiwu (approved):

PR #24970, opened 2025-06-06
[webgpu] add cases to cover the subgroup matrix path

xhcao:


xhcao:
@fs-eire @guschmue @jchen10 PTAL

fs-eire (approved):

PR #24973, opened 2025-06-06
Add Int4 and UInt4 support for Cast

amarin16:
Add Int4 and UInt4 support for `Cast`.

There were a few QNN pipelines [failing](https://aiinfra.visualstudio.com/PublicPackages/_build/results?buildId=841810&view=logs&j=9d976e38-31ec-50dd-b1f8-279fbf889fca&t=85ed5ad3-b72a-52c3-abe0-a87b66004fd0&l=1773) for this PR, which are fixed by this onnx PR [Update input and output tensors in pb files to match the model](https://github.com/onnx/onnx/pull/7074). The problem is that onnxruntime, which uses onnx as a submodule in `cmake/external/onnx`, points to the latest release of onnx (1.18.0), but in order to have the pipeline run with my onnx fix, we would need to point to a newer version.

Since we can't update the onnx submodule to point to a non-release commit, waiting for a new onnx release might take a long time, and creating a patch under [onnxruntime/cmake/patches/onnx](https://github.com/microsoft/onnxruntime/tree/main/cmake/patches/onnx) with the changes in my onnx PR is tricky because that fix changes some binary files, this PR skips the tests which are currently failing in the QNN pipelines.


onnxruntime/core/providers/cpu/tensor/cast_op.cc
```diff
@@ -262,6 +262,58 @@ struct TensorCaster<MLFloat16, float> {
...
+                "The Int4x2 tensor size is invalid for casting to float.");
+
+    for (size_t i = 0; i < in_shape_size; ++i) {
+      const uint8_t packed = static_cast<uint8_t>(in_data[i].bits_);
+
+      // Extract signed high and low nibble
+      int8_t high_nibble = static_cast<int8_t>(packed) >> 4;
+      int8_t low_nibble = static_cast<int8_t>(packed << 4) >> 4;
+
+      out_data[2 * i] = static_cast<float>(high_nibble);
```
tianleiwu: shall the storage order be low_nibble first?

yuslepukhin: Yes, we also have users on BE platforms so the code should account for that.

onnxruntime/test/providers/cpu/tensor/cast_op_test.cc
```diff
@@ -57,7 +57,7 @@ void TestCastOp(gsl::span<const SrcType> input,
                 const BaseTester::DimsVariant& dimensions,
                 OpTester::ExpectResult expect_result = OpTester::ExpectResult::kExpectSuccess,
                 const std::string& expected_failure_string = "",
-                int opset = 13,
+                int opset = 21,
```
hariharans29, line 60: Best to keep the opset bump local to the in4 test(s) ? (Not sure though, might be fine either way)

amarin16: All the existing `CastOpTest` tests seem to still pass

amarin16:
This PR currently needs https://github.com/onnx/onnx/pull/7074 in order for the QNN pipelines to pass. That means that the `cmake/external/onnx` submodule needs to be pointed to a newer commit, which includes the change above. However, the onnx submodule already [points](https://github.com/microsoft/onnxruntime/tree/986b66af96252488bcf885741623ba877964baca/cmake/external) to the latest release [v1.18.0](https://github.com/onnx/onnx/releases/tag/v1.18.0), so we will most likely need a new onnx release in order to point to it from onnxruntime.
Unless we can point to an unreleased commit of onnx, until we get an onnx release our only alternative is to make the required QNN pipelines that are failing not required for this PR.

onnxruntime/core/providers/cpu/tensor/cast_op.cc
```diff
@@ -60,6 +60,26 @@ template <typename T>
...
+      std::is_same_v<T, int16_t> ||
+      std::is_same_v<T, uint16_t> ||
+      std::is_same_v<T, int32_t> ||
+      std::is_same_v<T, uint32_t> ||
+      std::is_same_v<T, int64_t> ||
+      std::is_same_v<T, uint64_t>;
+};
+
+template <typename T>
+struct IsStandardFloatType {
```
edgchen1: is this different from `std::is_floating_point`?

onnxruntime/core/providers/cpu/tensor/cast_op.cc
```diff
@@ -219,37 +320,98 @@ struct TensorCaster<std::string, DstType> {
...
+    auto* out_data = out.MutableData<Int4x2>();
+
+    // Every 2 strings combine into 1 Int4x2
+    const size_t shape_size = narrow<size_t>(shape.Size());
+    size_t i = 0;
+    for (; i < shape_size - 1; i += 2) {
+      // Parse each string and clamp to int4 range (-8 to 7)
+      int v0 = std::stoi(in_data[i]);
+      int v1 = std::stoi(in_data[i + 1]);
+      int8_t val0 = static_cast<int8_t>(std::clamp(v0, -8, 7));
```
edgchen1: can we use constants for (u)int4 min/max?

onnxruntime/core/providers/cpu/tensor/cast_op.cc
```diff
@@ -262,6 +424,258 @@ struct TensorCaster<MLFloat16, float> {
...
+                    std::enable_if_t<IsStandardIntegerType<DstType>::value || IsOrtFloat16Type<DstType>::value
+#if !defined(DISABLE_FLOAT8_TYPES)
+                                     || IsOrtFloat8Type<DstType>::value
+#endif
+                                     >> {
+  void Cast(const OpKernelContext&, const TensorShape& shape, const Tensor& in, Tensor& out) const {
+    const auto* in_data = in.Data<Int4x2>();
+    auto* out_data = out.MutableData<DstType>();
+
+    for (size_t i = 0; i < narrow<size_t>(shape.Size()); ++i) {
```
edgchen1: general nit: can we avoid calling `narrow` in the loop condition?

onnxruntime/core/providers/cpu/tensor/cast_op.cc
```diff
@@ -262,6 +424,258 @@ struct TensorCaster<MLFloat16, float> {
...
+      // elem 0 is the low nibble, 1 the high nibble
+      auto val = in_data[i >> 1].GetElem(i & 0x1);
+
+      out_data[i] = Int4ElementConverter<DstType>::Convert(val);
+    }
+  }
+};
+
+template <typename DstType>
+struct TensorCaster<UInt4x2, DstType, std::enable_if_t<IsStandardFloatType<DstType>::value>> {
```
edgchen1: general: can some of the code duplication and SFINAE be reduced with `if constexpr()`?

onnxruntime/core/providers/cpu/tensor/cast_op.cc
```diff
@@ -299,6 +714,53 @@ struct TensorCaster<MLFloat16, std::string> {
...
+      out_data[i] = DstType(static_cast<float>(in_data[i]), false);
+    }
+  }
+};
+
+// TensorCasterNoSat should never be instantiated for Int4x2/UInt4x2
+template <typename DstType>
+struct TensorCasterNoSat<Int4x2, DstType, void> {
+  void Cast(const OpKernelContext&, const TensorShape&, const Tensor&, Tensor&) const {
+    ORT_THROW("Int4x2 should never use TensorCasterNoSat");
```
edgchen1: if this should not happen, can we detect at compile time? e.g. with static_assert?

onnxruntime/test/onnx/TestCase.cc
```diff
@@ -948,6 +948,16 @@ std::unique_ptr<std::set<BrokenTest>> GetBrokenTests(const std::string& provider
       {"slice_neg_steps",
        "Type parameter (Tind) bound to different types (tensor(int64) and tensor(int32) in node ()."},
       {"cast_BFLOAT16_to_FLOAT", "Unexpected input data type"},
+      {"cast_FLOAT16_to_INT4", "Skipped until onnxruntime/cmake/external/onnx points to a version of onnx which includes @onnx/onnx/pull/7074"},
```
edgchen1: can we explicitly name the next release that is expected to include it? easier than having to find the onnx PR and track down when it got included.

onnxruntime/core/providers/cpu/tensor/cast_op.cc
amarin16: do you have an example in mind?

edgchen1: for example, there are two partial specializations for `TensorCaster<UInt4x2, DstType>`. I think they could be consolidated.

onnxruntime/core/providers/cpu/tensor/cast_op.cc
amarin16: It doesn't seem to compile when `static_assert` is used. It also doesn't compile without this specialization, which is why I had to add it in the first place.

onnxruntime/core/providers/cpu/tensor/cast_op.cc
amarin16: not really, I removed it.

onnxruntime/core/providers/cpu/tensor/cast_op.cc
amarin16: addressed

onnxruntime/test/onnx/TestCase.cc
amarin16: addressed

onnxruntime/core/providers/cpu/tensor/cast_op.cc
amarin16: addressed

onnxruntime/core/providers/cpu/tensor/cast_op.cc
amarin16: updated

onnxruntime/core/providers/cpu/tensor/cast_op.cc
```diff
@@ -177,6 +190,93 @@ template <>
 struct EigenCastType<BFloat16> {
   using type = Eigen::bfloat16;
 };
+
+namespace {
```
edgchen1: I think this code is already within the anonymous namespace

onnxruntime/core/providers/cpu/tensor/cast_op.cc
```diff
@@ -177,6 +190,93 @@ template <>
...
+
+// Helper struct for converting from Int4x2/UInt4x2 elements to any destination type
+template <typename DstType>
+struct Int4ElementConverter {
+  static DstType Convert(int8_t val) {
+    if constexpr (IsOrtFloat16Type<DstType>::value) {
+      return DstType(static_cast<float>(val));
+    }
+#if !defined(DISABLE_FLOAT8_TYPES)
+    else if constexpr (IsOrtFloat8Type<DstType>::value) {
```
edgchen1: maybe we can reduce the ifdefs by having `IsOrtFloat8Type` just be false if `DISABLE_FLOAT8_TYPES` is defined

onnxruntime/core/providers/cpu/tensor/cast_op.cc
edgchen1: I see, the source types would contain int4 and uint4. is it just not implemented yet or actually invalid?

onnxruntime/core/providers/cpu/tensor/cast_op.cc
```diff
@@ -262,6 +408,225 @@ struct TensorCaster<MLFloat16, float> {
...
+    const ptrdiff_t shape_size = narrow<ptrdiff_t>(shape.Size() + 1) >> 1;
+    const auto* in_data = in.Data<Int4x2>();
+    auto* out_data = out.MutableData<UInt4x2>();
+
+    for (ptrdiff_t i = 0; i < shape_size; ++i) {
+      auto low_nibble = in_data[i].GetElem(0);
+      auto high_nibble = in_data[i].GetElem(1);
+
+      // Convert to unsigned by clamping at 0
+      uint8_t low_unsigned = static_cast<uint8_t>(std::max(0, static_cast<int>(low_nibble)) & 0x0F);
```
adrianlizarraga: Is clamping to 0 correct here? I think [onnx casting](https://github.com/onnx/onnx/blob/09dcfe787e85070f9057d566c0cba4646e435cc0/onnx/reference/ops/op_cast.py#L29) is consistent with `np.astype` and normal "C-casting" rules.

If, for example, we wanted to cast a `int8_t` value of -1 to `uint8_t`. The expected result should be 255 (0xFF). 

Similarly, for `int4` to `uint4`, a value of -1 should be casted to 15 (0xF). I think the clamp to 0 makes this evaluate to 0 instead.

amarin16: addressed this

onnxruntime/test/providers/cpu/tensor/cast_op_test.cc
```diff
@@ -207,6 +207,1036 @@ TEST(CastOpTest, ToString) {
...
+  // Negative values will be cast to their unsigned representation
+  const std::vector<uint8_t> expected_uint8_output = {248, 7, 0, UINT8_MAX, 3, 251, 6, 2};
+
+  // WHEN, THEN
+  TestCastOp(gsl::make_span(int4x2_input), gsl::make_span(expected_uint8_output), shape);
+}
+
+TEST(CastOpTest, Int4x2ToInt16) {
+  // GIVEN
+  const std::vector<int64_t> shape{2, 2, 2};
```
tianleiwu, line 245: Could you add a test of shape like {3, 1, 2} and {1, 2, 3}? I think uint4x2 and int4x2 has some odd implementation for packing.

onnxruntime/test/providers/cpu/tensor/cast_op_test.cc
```diff
@@ -1376,6 +1376,48 @@ TEST(CastOpTest, UInt4x2ToFloat8E5M2) {
                                   OpTester::ExpectResult::kExpectSuccess, "", 21, Saturate::False);
 }
 
+TEST(CastOpTest, Float8E4M3FNToInt4x2) {
+  // GIVEN
+  const std::vector<int64_t> shape{2, 2, 2};
```
tianleiwu, line 1400: Please add a test with shape like {1, 2, 3} and see how the int4x2 is padded.

onnxruntime/core/providers/cpu/tensor/cast_op.cc
```diff
@@ -58,8 +58,43 @@ using IsOrtFloat16Type = boost::mp11::mp_contains<TypeList<BFloat16, MLFloat16>,
 #if !defined(DISABLE_FLOAT8_TYPES)
 template <typename T>
 using IsOrtFloat8Type = boost::mp11::mp_contains<element_type_lists::AllFloat8, T>;
+#else
+template <typename T>
+struct IsOrtFloat8Type : std::false_type {};
```
tianleiwu, line 63: Even when float8 types is disabled, float8 type is still float8. This naming is confusing. But it is probably fine since when float8 is disabled, the Cast operator does not support float8 input.

PR #24977, opened 2025-06-06
[CUDA] upgrade cudnn front end to 1.12

tianleiwu:
### Description

Upgrade cudnn front end to 1.12

### Motivation and Context

https://github.com/NVIDIA/cudnn-frontend/releases/tag/v1.12.0


yuslepukhin:
Do you need to update the machine image?

tianleiwu:
> Do you need to update the machine image?

It is front-end compatible with cudnn 9.x, which is available in CI pipelines. So it is not required to update image in CI.

snnn (approved):

PR #23979, opened 2025-03-10
Convert graph initializers into OrtValue Phase I

yuslepukhin:
### Description
This PR converts TensorProto graph initializers to TensorProto/OrtValue pairs.
Currently, we only split the output for some optimizers to the above pairs.
Eventually, we should be able to convert all initializers to OrtValues on load.
Small weights will continue to be an exception as the are sometimes required by ONNX inference functions.
Some graph API leaks to EPs so we are not able to remove it at present, and this constrains our ability to convert everything at once.

### Motivation and Context
Lay Gound for proper layers separation. Eventually eliminate weights copies in the EPs.

PR #24978, opened 2025-06-06
Fix in-memory initializer handling for non-CPU device

### LINKED ISSUE #24768: Native WebGPU EP fails to run model with in-memory external data
### Description
For models that are built by ModelEditorAPI, they may have initializers that are TensorProto with external data in existing memory buffer. For those TensorProto, the 'location' field of the external data info is set to a special marker `*/_ORT_MEM_ADDR_/*`, and the 'offset' field contains the address of the memory buffer.

When running those models with native WebGPU EP, it reports error, for example the error log of running a "Conv" node with initializer using in-memory external data:

"ORT status error code: 6 error message: Non-zero status code returned while running Conv node. Name:'_0_token_1' Status Message: C:\workspace\onnxruntime\onnxruntime\core\providers\webgpu\webgpu_context.cc:210 onnxruntime::webgpu::WebGpuContext::Run std::all_of(inputs.begin(), inputs.end(), [](const ProgramInput& input) { const auto* tensor = input.tensor; return tensor != nullptr && tensor->Location().mem_type == OrtMemType::OrtMemTypeDefault && tensor->Location().device.Type() == OrtDevice::GPU && !strcmp(tensor->Location().name, WEBGPU_BUFFER); }) was false. All inputs must be tensors on WebGPU buffers."

It seems WebGPU EP doesn't upload the in-memory external weights to GPU buffer before running it.

/cc @fdwr @skottmckay @guschmue

fs-eire:
### Description

In-memory initializers are currently not handled correctly for non-CPU devices. If the memory planner assigned the initializer to a non-CPU device, the initializer is not copied to the device during the initialization steps.

This PR fixes this problem by adding an extra step to copy the initializer to the device if needed.

### Motivation and Context

Fixes #24768

### DIFF
--- a/onnxruntime/core/framework/session_state_utils.cc
+++ b/onnxruntime/core/framework/session_state_utils.cc
@@ -398,11 +398,7 @@ common::Status SaveInitializedTensors(
       ort_value = *(session_options.initializers_to_share_map.at(name));
       LOGS(logger, INFO) << "Using user supplied initializer with name (" << name << ").";
 
-    } else if (graph.GetOrtValueInitializer(name, ort_value)) {
-      // populated OrtValue from the Graph instance
     } else {
-      const ONNX_NAMESPACE::TensorProto& tensor_proto = *(entry.second);
-
       std::optional<MemBuffer> m;
       AllocatorPtr alloc;
       // TODO: if the tensor need be copied, does it have enough room?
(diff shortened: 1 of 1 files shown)


fs-eire:
@huningxin

snnn:
Add a test case?

fs-eire:
The original issue can be fixed by #23979. As suggested by @skottmckay we can wait for it to be merged first.

fs-eire:
Close this PR. The problem is fixed by #23979.

PR #23507, opened 2025-01-28
[DRAFT/DO NOT REVIEW] Add Rotary Embedding from ONNX Opset 23 

shubhambhokare1:
### Description
Add kernels for Rotary Embedding op that is planned to be added in ONNX opset 23.

### Motivation and Context
This PR is currently experimental, and the kernels are added to a temporary location in contrib_ops. Once the following happens:
1) https://github.com/onnx/onnx/pull/6461/ is merged
2) This op is released in the latest ONNX release
These kernels will be migrated from contrib_ops/onnx_std_exp to the correct location.

PR #24980, opened 2025-06-06
Add RotaryEmbeddings(23) - CPU

### LINKED ISSUE #24556: [Feature Request] Implement RotaryEmbedding-23

RotaryEmbedding-23 will be introduced in ONNX==1.18.0: https://onnx.ai/onnx/operators/onnx__RotaryEmbedding.html#rotaryembedding-23

N/A

titaiwangms:
### Description

Add ONNX RotaryEmbedding(23) following https://github.com/onnx/onnx/blob/main/docs/Operators.md#RotaryEmbedding. The PR uses contrib op RotaryEmbedding implementation under the hood.

The main difference between this op and the contrib op is that the `position_ids` in ONNX RotaryEmbedding is optional. When it's not provided, `cos_cache` and `sin_cache` should be 3d.

### Motivation and Context

Fix #24556
Reference #23507


onnxruntime/core/providers/cpu/llm/rotary_embedding.cc
```diff
@@ -0,0 +1,137 @@
...
+using namespace onnxruntime::rotary_embedding_helper;
+
+namespace onnxruntime {
+
+// These ops are internal-only, so register outside of onnx
+#define REGISTER_KERNEL_TYPED(T)                                        \
+  ONNX_OPERATOR_TYPED_KERNEL_EX(                                        \
+      RotaryEmbedding,                                                  \
+      kMSDomain,                                                        \
+      23,                                                               \
```
gramalingam, line 18: Is this opset version? Shouldn't it be 1 for MSDomain?

titaiwangms:
@tianleiwu Could you help me reason the CI failing? It seems the code only fails on Windows Asan

tianleiwu:
@titaiwangms, the code crashed (maybe there is bug in either test code or the kernel):

[ RUN      ] RotaryEmbeddingTest.RotaryEmbedding_Interleaved_NoPosIds_SmallData_LlamaMSFT
1: =================================================================
1: ==12148==ERROR: AddressSanitizer: unknown-crash on address 0x119c4a2fe0c0 at pc 0x7ff7dd41f007 bp 0x0010901c09f0 sp 0x0010901c0a00
1: READ of size 32 at 0x119c4a2fe0c0 thread T0
1:     #0 0x7ff7dd41f006 in rope_avx2::`anonymous namespace'::RopeKernel_Avx2_fp32_Impl<1> D:\a\_work\onnxruntime\onnxruntime\onnxruntime\core\mlas\lib\rotary_embedding_kernel_avx2.cpp:238
1:     #1 0x7ff7dd41f2d4 in rope_avx2::RopeKernel_Avx2_fp32(float const *, float const *, float const *, unsigned __int64, bool, float *)

Could you review the test case setup and the kernel code?

onnxruntime/core/providers/cpu/llm/rotary_embedding_helper.h
```diff
@@ -0,0 +1,196 @@
...
+
+template <typename T>
+Status CheckInputs(const T* input,
+                   const T* position_ids,
+                   const T* cos_cache,
+                   const T* sin_cache,
+                   int num_heads,
+                   int rotary_embedding_dim,
+                   void* parameters) {
+  //    input        : (batch_size, sequence_length, hidden_size)
```
tianleiwu: update comment here: input shape can also be (batch, num_heads, seq_len, head_size)

onnxruntime/core/providers/cpu/llm/rotary_embedding_helper.h
```diff
@@ -0,0 +1,196 @@
...
+  }
+
+  // Check num_heads and rotary_embedding_dim
+  if (rotary_embedding_dim > 0 && num_heads == 0) {
+    return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT, "num_heads must be provided if rotary_embedding_dim is ",
+                           "specified");
+  }
+  // Get attributes from inputs
+  int batch_size = static_cast<int>(input_dims[0]);
+  int sequence_length = static_cast<int>(input_dims[1]);
```
tianleiwu: when input is (batch, num_heads, seq_len, hidden), the sequence length is input_dims[2]

onnxruntime/core/providers/cpu/llm/rotary_embedding_helper.h
```diff
@@ -0,0 +1,196 @@
...
+
+  bool transposed = false;
+  if (input_dims.size() == 4) {
+    // input is [batch, seq, num_heads, head_size]
+    hidden_size = static_cast<int>(input_dims[2]) * static_cast<int>(input_dims[3]);
+    transposed = true;
+  }
+
+  int position_ids_format = -1;
+  int max_sequence_length;
```
tianleiwu, line 65: initialize max_sequence_length and head_size to -1.

tianleiwu: merge lines 65/66 with 85/86.

onnxruntime/core/providers/cpu/llm/rotary_embedding_helper.h
```diff
@@ -0,0 +1,196 @@
...
+                             sin_cache_dims.size());
+    }
+    if (cos_cache_dims[0] != sin_cache_dims[0] || cos_cache_dims[1] != sin_cache_dims[1]) {
+      return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT, "Inputs 'cos_cache' and 'sin_cache' are expected to have ",
+                             "the same shape");
+    }
+    // Check position_ids
+    const auto& position_ids_dims = position_ids->Shape().GetDims();
+    if (!onnxruntime::IsScalarOr1ElementVector(position_ids) && position_ids_dims.size() != 2) {
+      return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT, "Input 'position_ids' is expected to have 0, 1, or 2 ",
```
tianleiwu: Do we allow scalar (0 dimension) for position_ids? If so, update line 38

onnxruntime/core/providers/cpu/llm/rotary_embedding_helper.h
```diff
@@ -0,0 +1,196 @@
...
+                             "dimensions, got ", position_ids_dims.size());
+    }
+
+    max_sequence_length = static_cast<int>(cos_cache_dims[0]);
+    head_size = rotary_embedding_dim == 0 ? static_cast<int>(cos_cache_dims[1]) * 2
+                                          : static_cast<int>(hidden_size / num_heads);
+    if (rotary_embedding_dim > 0 && rotary_embedding_dim > head_size) {
+      return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT, "rotary_embedding_dim must be less than or equal to ",
```
tianleiwu, line 128: duplicated with line 89.

onnxruntime/core/providers/cpu/llm/rotary_embedding_helper.h
```diff
@@ -0,0 +1,196 @@
...
+                   const T* position_ids,
+                   const T* cos_cache,
+                   const T* sin_cache,
+                   int num_heads,
+                   int rotary_embedding_dim,
+                   void* parameters) {
+  //    input        : (batch_size, sequence_length, hidden_size)
+  //    cos cache    : (max_sequence_length, rotary_embedding_dim / 2)
+  //    sin cache    : (max_sequence_length, rotary_embedding_dim / 2)
+  //    position ids : (1) or (batch_size, sequence_length)
```
tianleiwu: onnx spec says that position ids is 2D tensor

onnxruntime/core/providers/cpu/llm/rotary_embedding_helper.h
```diff
@@ -0,0 +1,196 @@
...
+template <typename T>
+Status CheckInputs(const T* input,
+                   const T* position_ids,
+                   const T* cos_cache,
+                   const T* sin_cache,
+                   int num_heads,
+                   int rotary_embedding_dim,
+                   void* parameters) {
+  //    input        : (batch_size, sequence_length, hidden_size)
+  //    cos cache    : (max_sequence_length, rotary_embedding_dim / 2)
```
tianleiwu: This comment is not correct.

When position id is provided and rotary_embedding_dim is 0 (not provided):
* (max_position_id_plus_1, head_size / 2)                              
    
When position id is provided and rotary_embedding_dim > 0:
* (max_position_id_plus_1, rotary_embedding_dim / 2)

When position id is not provided and rotary_embedding_dim is 0 (not provided):
* (batch_size, sequence_length, head_size / 2)

When position id is not provided and rotary_embedding_dim > 0:
* (batch_size, sequence_length, rotary_embedding_dim / 2)

onnxruntime/core/providers/cpu/llm/rotary_embedding_helper.h
```diff
@@ -0,0 +1,196 @@
...
+
+  // Check input
+  const auto& input_dims = input->Shape().GetDims();
+  if (input_dims.size() != 3 && input_dims.size() != 4) {
+    return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT, "Input 'x' is expected to have 3 or 4 dimensions, got ",
+                           input_dims.size());
+  }
+
+  // Check num_heads and rotary_embedding_dim
+  if (rotary_embedding_dim > 0 && num_heads == 0) {
```
tianleiwu: No need this since we enforce same logic in constructor:
https://github.com/microsoft/onnxruntime/blob/4d59816bc941180afeef93def16ccfc8be62c184/onnxruntime/core/providers/cpu/llm/rotary_embedding.cc#L34-L36

onnxruntime/core/providers/cpu/llm/rotary_embedding_helper.h
titaiwangms: I wonder if I should ban the usage of 1D or I should add a comment to allow it when position_ids is 1D. The kernel supports it anyway.

tianleiwu: Let's follow the spec and only allows 2D.

onnxruntime/core/providers/cpu/llm/rotary_embedding_helper.h
```diff
@@ -0,0 +1,196 @@
...
+    }
+    // Check position_ids
+    const auto& position_ids_dims = position_ids->Shape().GetDims();
+    if (!onnxruntime::IsScalarOr1ElementVector(position_ids) && position_ids_dims.size() != 2) {
+      return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT, "Input 'position_ids' is expected to have 0, 1, or 2 ",
+                             "dimensions, got ", position_ids_dims.size());
+    }
+
+    max_sequence_length = static_cast<int>(cos_cache_dims[0]);
+    head_size = rotary_embedding_dim == 0 ? static_cast<int>(cos_cache_dims[1]) * 2
```
tianleiwu: One way to get head_size:
* When x is 4D, head_size is the last dimension
* When x is 3D and num_heads is provided, we can get from hidden_size / num_heads.

The second way:
 * When rotary_embedding_dim is 0, we can get from last dimension of cos/sin cache.

I suggest using the first way to get head_size, then check that the last dim of cos/sin cache is head_size / 2.

onnxruntime/core/providers/cpu/llm/rotary_embedding_helper.h
```diff
@@ -0,0 +1,196 @@
...
+
+    max_sequence_length = static_cast<int>(cos_cache_dims[1]);
+    head_size = rotary_embedding_dim == 0 ? static_cast<int>(cos_cache_dims[2]) * 2
+                                          : static_cast<int>(hidden_size / num_heads);
+    if (rotary_embedding_dim > 0 && rotary_embedding_dim > head_size) {
+      return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT, "rotary_embedding_dim must be less than or equal to ",
+                             "head_size");
+    }
+    // Check cos_cache input shapes
+    if (max_sequence_length != static_cast<int>(cos_cache_dims[1])) {
```
tianleiwu: See line 85. I think this check will always be true here.

onnxruntime/core/providers/cpu/llm/rotary_embedding_helper.h
```diff
@@ -0,0 +1,196 @@
...
+      return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT, "Input 'cos_cache' dimension 1 should be same as ",
+                             "max_sequence_length, got ", cos_cache_dims[0]);
+    }
+    if ((head_size / 2) != static_cast<int>(cos_cache_dims[2]) && (rotary_embedding_dim > 0 && (rotary_embedding_dim / 2) != static_cast<int>(cos_cache_dims[2]))) {
```
tianleiwu: The conditions cannot be joined by && here. The correct one is something like 
`if (cos_cache_dims[2] != (rotary_embedding_dim > 0 ? rotary_embedding_dim : head_size) / 2)`

Also, I suggest to add the following logic after hidden_size is avaialbe:
```
if (rotary_embedding_dim == 0) {
   head_size = hidden_size / num_heads;
   rotary_embedding_dim = head_size / 2;
}
```
So that a simple logic like `if (cos_cache_dims[2] != rotary_embedding_dim/2)` can be used here.

onnxruntime/core/providers/cpu/llm/rotary_embedding_helper.h
```diff
@@ -0,0 +1,196 @@
...
+    if (cos_cache_dims.size() != 3) {
+      return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT, "Input 'cos_cache' is expected to have 3 dimensions, got ",
+                             cos_cache_dims.size());
+    }
+    const auto& sin_cache_dims = sin_cache->Shape().GetDims();
+    if (sin_cache_dims.size() != 3) {
+      return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT, "Input 'sin_cache' is expected to have 3 dimensions, got ",
+                             sin_cache_dims.size());
+    }
+    if (cos_cache_dims[0] != sin_cache_dims[0] || cos_cache_dims[1] != sin_cache_dims[1] || cos_cache_dims[2] != sin_cache_dims[2]) {
```
tianleiwu, line 94: add a check that first two dimension of cache is batch_size and sequence_length.

onnxruntime/core/providers/cpu/llm/rotary_embedding.cc
```diff
@@ -0,0 +1,141 @@
...
+                                                                   position_ids,
+                                                                   cos_cache,
+                                                                   sin_cache,
+                                                                   num_heads,
+                                                                   rotary_embedding_dim,
+                                                                   &parameters));
+
+  Tensor* output = context->Output(0, X->Shape());
+
+  if (parameters.sequence_length > parameters.max_sequence_length) {
```
tianleiwu: This check can be moved to CheckInputs

onnxruntime/core/providers/cpu/llm/rotary_embedding_helper.h
titaiwangms: Done

onnxruntime/core/providers/cpu/llm/rotary_embedding_helper.h
titaiwangms: Done

onnxruntime/core/providers/cpu/llm/rotary_embedding_helper.h
titaiwangms: This needs to be declared outside of if/else block to be in the scope. I have done some changes.

onnxruntime/core/providers/cpu/llm/rotary_embedding_helper.h
titaiwangms: Done

onnxruntime/core/providers/cpu/llm/rotary_embedding_helper.h
titaiwangms: Banned 1D in check_inputs and in the tests.

onnxruntime/core/providers/cpu/llm/rotary_embedding_helper.h
titaiwangms: Updated to the comments! Thanks!

onnxruntime/core/providers/cpu/llm/rotary_embedding_helper.h
titaiwangms: Removed

onnxruntime/core/providers/cpu/llm/rotary_embedding_helper.h
titaiwangms: Now I get head size in the very begining.

onnxruntime/core/providers/cpu/llm/rotary_embedding_helper.h
titaiwangms: Good catch!

onnxruntime/core/providers/cpu/llm/rotary_embedding_helper.h
titaiwangms: I choose the first one (`if (cos_cache_dims[2] != (rotary_embedding_dim > 0 ? rotary_embedding_dim : head_size) / 2)`). Semantically, it is more readable in terms of the checking input logic.

onnxruntime/core/providers/cpu/llm/rotary_embedding_helper.h
titaiwangms: Done

onnxruntime/core/providers/cpu/llm/rotary_embedding.cc
titaiwangms: Done

onnxruntime/core/providers/cpu/llm/rotary_embedding.cc
```diff
@@ -0,0 +1,135 @@
...
+  const int head_size = parameters.head_size;
+  const int head_stride = parameters.head_stride;
+  const int seq_stride = parameters.seq_stride;
+  const int batch_stride = parameters.batch_stride;
+  const int position_ids_format = parameters.position_ids_format;
+  const int rotary_emb_dim = parameters.rotary_embedding_dim;
+  const int half_rotary_emb_dim = rotary_emb_dim / 2;
+
+  const int loop_len = batch_size * sequence_length * n_heads;
+  const double cost = static_cast<double>(rotary_emb_dim);
```
tianleiwu: cost is not accurate, which will impact parallel (smaller cost will use less threads)

* read: head_size * sizeof(T)
* write: head_size * sizeof(T)
* compute: rotary_emb_dim * c.  It seems that the logic is a little complex, I guess it will need dozens of clock circles. Maybe use 32 as c.

titaiwangms: Done

titaiwangms: Thanks for this great suggestion!

onnxruntime/core/providers/cpu/llm/rotary_embedding.cc
```diff
@@ -0,0 +1,139 @@
...
+                    input_data + rotary_emb_dim,
+                    (head_size - rotary_emb_dim) * sizeof(T));
+      }
+    }
+  });
+
+  return Status::OK();
+}
+
+template Status RunRotaryEmbedding<float>(concurrency::ThreadPool* tp, RotaryParameters parameters, const float* input,
```
xadupre, line 98: Do you need this?

onnxruntime/core/providers/cpu/llm/rotary_embedding_helper.h
```diff
@@ -0,0 +1,207 @@
...
+    sequence_length = static_cast<int>(input_dims[2]);
+    num_heads = static_cast<int>(input_dims[1]);
+    head_size = static_cast<int>(input_dims[3]);
+    hidden_size = num_heads * head_size;
+    transposed = true;
+  } else if (input_dims.size() == 3) {
+    // If it's 3d, it is expected to have shape [batch, seq_len, hidden_size].
+    sequence_length = static_cast<int>(input_dims[1]);
+    hidden_size = static_cast<int>(input_dims[2]);
+    head_size = hidden_size / num_heads;
```
xadupre, line 72: Do you need to check it returns an integer?

onnxruntime/core/providers/cpu/llm/rotary_embedding.cc
```diff
@@ -0,0 +1,139 @@
...
+      T* output_data = output + block_offset;
+
+      const T* cos_data;
+      const T* sin_data;
+      int cache_offset;
+      if (position_ids_format == 0) {
+        cache_offset = (b * sequence_length + s) * half_rotary_emb_dim;
+      } else {
+        // Cache is (M, H/2) or (M, rotary_embedding_dim/2)
+        const int position_id = static_cast<int>(position_ids[b * sequence_length + s]);
```
xadupre, line 79: Should we enforce ``b * sequence_length + s`` is not outside boundary? This is not the common case but the cost does not seem much compare to some potential crash.

gramalingam: Is it possible to check outside the two enclosing loops, i.e., outside the TryParallelFor? We should be able to check the shapes of tensors to guarantee something like this, I suspect.

onnxruntime/core/providers/cpu/llm/rotary_embedding.cc
titaiwangms: Yes, If I am not mistaken, we need to initialize the kernel with the kinds of data type we need to enable the op usage, so here we have both float and MLFloat16 initialized.

onnxruntime/core/providers/cpu/llm/rotary_embedding_helper.h
titaiwangms: `num_heads` is required to be passed in as int attribute when it's 3d input. We should be fine.

onnxruntime/core/providers/cpu/llm/rotary_embedding.cc
titaiwangms: We don't need to check the boundary again, as in `rotary_embedding_helper.h`, we already guaranteed that position_ids is [batch, sequence_length] when position_ids_format==1. Here, the loop is bounded to 

```cpp
const int loop_len = batch_size * sequence_length * n_heads;
```

where we know it does not go beyond position_ids shape.

titaiwangms:
@tianleiwu @gramalingam @xadupre PTAL
I need this one merged to start RotaryEmbedding with cuda, I can fix the minor issues in the next PR if there is any.

tianleiwu (approved):

PR #24985, opened 2025-06-08
Make OrtDevice generic to support plugin EP usage.

skottmckay:
### Description

Make OrtDevice generic by removing vendor specific hardcoded memory types and replacing with generic DEFAULT/HOST_ACCESSIBLE values and an explicit vendor id to differentiate.

Remove unnecessary device id from OrtMemoryInfo. The OrtDevice in OrtMemoryInfo provides this.

### Motivation and Context

Support plugin EPs


include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -5328,6 +5343,30 @@ struct OrtApi {
    * \since Version 1.23.
    */
   ORT_API2_STATUS(AllocatorGetStats, _In_ const OrtAllocator* ort_allocator, _Outptr_ OrtKeyValuePairs** out);
+
+  /** \brief Create an ::OrtMemoryInfo
+   *
+   * \param[in] name Arbitrary name.
```
yuslepukhin, line 5349: The name is not entirely arbitrary. I think it should be one of those handled in allocator.cc.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -5328,6 +5343,30 @@ struct OrtApi {
...
+   * \param[out] out Newly created ::OrtMemoryInfo. Must be freed with OrtAPi::ReleaseMemoryInfo
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \since Version 1.23
+   */
+  ORT_API2_STATUS(CreateMemoryInfo_V2, _In_ const char* name, _In_ enum OrtMemoryInfoDeviceType device_type,
+                  _In_ uint32_t vendor_id, _In_ int16_t device_id, _In_ enum OrtDeviceMemoryType mem_type,
+                  _In_ size_t alignment, enum OrtAllocatorType allocator_type,
+                  _Outptr_ OrtMemoryInfo** out);
```
skottmckay, line 5369: This differs from CreateMemoryInfo in 2 ways. One is the obvious vendor_id. The other is that it takes OrtDeviceMemoryType and not OrtMemType for clarity. 

OrtMemType is more about a node input/output (e.g. node assigned to GPU uses the mem type to indicate an input should be on CPU) and not a general description of the memory info. OrtMemType is used internally by EPs and kernel definitions. No usage in custom ops via OrtMemoryInfo that I can find. There's `GetInputMemoryType` to return the OrtMemType, but that is used to map to a device or cpu/host accessible allocator and doesn't require an OrtMemoryInfo instance.

yuslepukhin (approved):
:shipit:

onnxruntime/core/providers/dml/DmlExecutionProvider/src/ExecutionProvider.h
```diff
@@ -249,8 +249,8 @@ namespace Dml
 
         bool CanCopy(const OrtDevice& srcDevice, const OrtDevice& dstDevice) const final
         {
-          return (srcDevice.Type() == OrtDevice::DML) ||
-                 (dstDevice.Type() == OrtDevice::DML);
+          return ((srcDevice.Type() == OrtDevice::GPU && srcDevice.Vendor() == OrtDevice::VendorIds::MICROSOFT) ||
+                  (dstDevice.Type() == OrtDevice::GPU && dstDevice.Vendor() == OrtDevice::VendorIds::MICROSOFT));
```
edgchen1, line 253: why is DML's DataTransfer::CanCopy() the only one that checks vendor id?

include/onnxruntime/core/session/onnxruntime_c_api.h
skottmckay: As in we should support the old things like CUDA_PINNED via the name? I wouldn't have thought that was necessary, and if it isn't doesn't that make the name arbitrary?

onnxruntime/core/providers/dml/DmlExecutionProvider/src/ExecutionProvider.h
skottmckay: As discussed. Preserved existing behavior with other EPs (only check the device), but this changed to device+vendor as it's no longer using the special-cased 'DML' device type.

Follow up PR will add vendor check to all EPs that implement IDataTransfer CanCopy.

edgchen1 (approved):

include/onnxruntime/core/session/onnxruntime_c_api.h
skottmckay: As it's just a comment can address/update in a follow up PR if it needs changing. With the new setup you shouldn't need to know about random magic internal strings as the PCI vendor id can be used.

PR #24989, opened 2025-06-09
Add version check for IAlloactor::GetStats

toothache:
### Description

Add version check for IAlloactor::GetStats.

### DIFF
--- a/onnxruntime/core/session/allocator_adapters.cc
+++ b/onnxruntime/core/session/allocator_adapters.cc
@@ -12,6 +12,7 @@ namespace onnxruntime {
 
 namespace {
 constexpr uint32_t kOrtAllocatorReserveMinVersion = 18;
+constexpr uint32_t kOrtAllocatorStatsMinVersion = 23;
 }  // namespace
 
 OrtAllocatorImplWrappingIAllocator::OrtAllocatorImplWrappingIAllocator(onnxruntime::AllocatorPtr&& i_allocator)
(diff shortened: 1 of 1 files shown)


onnxruntime/core/session/allocator_adapters.cc
```diff
@@ -27,16 +28,18 @@ OrtAllocatorImplWrappingIAllocator::OrtAllocatorImplWrappingIAllocator(onnxrunti
...
-      [](const OrtAllocator* this_, OrtKeyValuePairs** stats) noexcept -> OrtStatusPtr {
-    API_IMPL_BEGIN
-    auto kvp = std::make_unique<OrtKeyValuePairs>();
-    auto stats_map = static_cast<const OrtAllocatorImplWrappingIAllocator*>(this_)->Stats();
-    kvp->Copy(stats_map);
-    *stats = reinterpret_cast<OrtKeyValuePairs*>(kvp.release());
-    return nullptr;
-    API_IMPL_END
-  };
+  if (OrtAllocator::version >= kOrtAllocatorStatsMinVersion) {
```
toothache, line 37: Also check for `IAllocatorImplWrappingOrtAllocator::GetStats` and add test cases

skottmckay (approved):

skottmckay (dismissed):

onnxruntime/core/session/allocator_adapters.cc
```diff
@@ -12,6 +12,7 @@ namespace onnxruntime {
 
 namespace {
 constexpr uint32_t kOrtAllocatorReserveMinVersion = 18;
+constexpr uint32_t kOrtAllocatorStatsMinVersion = 23;
```
yuslepukhin, line 21: It would be great if you placed a detailed explanation here and in the PR as to how this works.
E.g. let' say one has a C program that is compiled against the older the binary.
Now we are calling the newer binary. What happens?

This file is a part of the onnxruntime compiled binary. As far as I can see both kOrtAllocatorStatsMinVersion and ORT_API_VERSON come from this build.

Where do we get the version of the older binary?

toothache: Added explanation based on my current understanding. @skottmckay , correct me if anything wrong.

toothache: But I'm not very sure if it's really necessary to check in `OrtAllocatorImplWrappingIAllocator` as `IAllocator` is internal and user will not have chance to call new function pointers if they build against an older library.

skottmckay: The `version` field is set on the client side when they compile against the ORT API. As the API is backwards compatible, they're free to update the onnxruntime library they use at runtime without having to recompile the client code. In this case they will be calling an ORT binary that knows about the GetStats function with an OrtAllocator that does not, and has garbage in that offset.

yuslepukhin (approved):
:shipit:

PR #24994, opened 2025-06-09
[webgpu] Restore FP16 math in flash attention generation

qjia7:
This PR restores FP16 math in flash attention generation shader. It follows the changes in #24953 to use scale to multiply Q first instead of calculating it after QK to avoid data overflow in FP16.


guschmue:
I can take it for a test with deepseek-r1 to make sure we don't overflow.

guschmue:
results for deepseek-r1 are ok.

PR #25007, opened 2025-06-10
Remove deprecated literal operator check from CMake configuration

fs-eire:
This pull request removes support for the `-Wdeprecated-literal-operator` compiler flag in the build configuration. It was previously added in #23347, but it is no longer needed since ONNX version upgraded.


fs-eire:
still need this as there are other build breaks :(

PR #25008, opened 2025-06-10
[webgpu] Fix test case 'Asin' on Vulkan backend

fs-eire:
### Description

Set the atol to 0.00022f for Vulkan(linux)


fs-eire:
Close this for now. Seems this only fails on an emulated environment (devbox) and passing on all other places.

PR #25016, opened 2025-06-10
[QNN EP] Enable boolean support for expand op

qti-yuduo:
### Description
Enable boolean support for expand op

### Motivation and Context
Additional op data type coverage.


HectorSVC (approved):
:shipit:

PR #25017, opened 2025-06-10
[QNN EP] MatMul+Add->Gemm fusion when AttentionFusion isn't enabled

qti-yuduo:
### Description
MatMul+Add->Gemm fusion when AttentionFusion isn't enabled.

### Motivation and Context

Graph transformation [MatMulAddFusion](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/optimizer/matmul_add_fusion.cc) fold `ONNX::MatMul` followed by `ONNX::Add` into `ONNX::GEMM`, however, it [intentionally skipping the portion belongs to "Attention Pattern"](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/optimizer/matmul_add_fusion.cc#L21). This result in poor performance on QNN EP (and other EPs who does not run *AttentionFusion transformers) due to unfused MatMul + Add pairs. 

image

With this change, additional GEMM would be fused *post* AttentionFusions.

### DIFF
--- a/onnxruntime/core/optimizer/graph_transformer_mgr.cc
+++ b/onnxruntime/core/optimizer/graph_transformer_mgr.cc
@@ -4,6 +4,9 @@
 #include "core/optimizer/graph_transformer_mgr.h"
 #include "core/optimizer/rule_based_graph_transformer.h"
 
+#include <memory>
+#include <utility>
+
 using namespace onnxruntime;
 using namespace ::onnxruntime::common;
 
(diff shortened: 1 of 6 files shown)


qti-yuduo:
Rebased (to resolve merge conflicts).

HectorSVC:
We got agreement last week that we want to move the fusion inside the QNN EP considering future EP ABI work. I'm not sure Ashish told you that.

qti-yuduo:
> We got agreement last week that we want to move the fusion inside the QNN EP considering future EP ABI work. I'm not sure Ashish told you that.

Thanks for lemme know, I can follow up with Ashish to get more details. In the meantime, this change is actually beyond just QNN EP. It more broader issue for all EPs that does not implement fused Attention. I feel this is the right place for the change, would you say even this should go limit to QNN only?

HectorSVC (approved):
:shipit:

Issue #25013, opened 2025-06-10
[Build] Tests fail for debug builds with recent VS versions


samwebster:
### Describe the issue

Build x64 debug version from main with tests enabled and they fail due to segfaults. Tested with VS 17.14.4 and 17.14.3. This is a regression from up to 4 months ago.

### Urgency

Blocking us from updating to the latest ORT

### Target platform

Windows x64

### Build script

```
onnxruntime/tools/ci_build/build.py --config Debug --cmake_path "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\Common7\IDE\CommonExtensions\Microsoft\CMake\CMake\bin\cmake.exe" --ctest_path "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\Common7\IDE\CommonExtensions\Microsoft\CMake\CMake\bin\ctest.exe" --cmake_generator "Visual Studio 17 2022" --build_shared_lib --parallel 32 --build_dir C:\work\onnxruntime\build_x64-cpu --cmake_extra_defines "onnxruntime_USE_NEURAL_SPEED=OFF" --enable_msvc_static_runtime --use_binskim_compliant_compile_flags
```

### Error / output

```
The following tests FAILED:
          1 - onnxruntime_test_all (SEGFAULT)
          2 - onnx_test_pytorch_converted (SEGFAULT)
          3 - onnx_test_pytorch_operator (SEGFAULT)
          4 - onnxruntime_shared_lib_dlopen_test (SEGFAULT)
          5 - onnxruntime_shared_lib_test (SEGFAULT)
          6 - onnxruntime_global_thread_pools_test (SEGFAULT)
          7 - onnxruntime_customopregistration_test (SEGFAULT)
          8 - onnxruntime_autoep_test (SEGFAULT)
          9 - onnxruntime_logging_apis_test (SEGFAULT)
```

This seems to be due to https://github.com/microsoft/onnxruntime/pull/24542

Call stack at point of read access violation:
```
>	std::ctype<char>::widen(char _Byte) Line 2755	C++
 	std::basic_ios<char,std::char_traits<char>>::widen(char _Byte) Line 113	C++
 	std::basic_ios<char,std::char_traits<char>>::init(std::basic_streambuf<char,std::char_traits<char>> * _Strbuf, bool _Isstd) Line 149	C++
[...]

PR #25019, opened 2025-06-10
Remove the builtin memory leak checker

snnn:
A temp solution for #25013 . Unfortunately we do not have people to maintain this code now.


yuslepukhin (approved):
:shipit:

skottmckay:
I found the info to be valuable during recent development work, and I haven't seen segfaults from its usage (different build params though). 

Can we not instead improve the flag to explicitly exclude from a build if needed? If you specify `--disable_memleak_checker` why would debug_alloc.h/cc be compiled?

baijumeswani:
I've found the mem leak checker valuable as well when I was working on the training c api. 

It would be nice if we can keep this functionality.

PR #24659, opened 2025-05-06
[QNN-EP] Einsum QDQ tests followup: node selector

1duo:
[QNN-EP] Einsum QDQ tests followup: node selector. 

Verified now Einsum QDQ tests are running with `QNN_DATATYPE_UFIXED_POINT_8`: 

```
2025-05-06 11:22:09.738247549 [V:onnxruntime:qdq_model_logger, qnn_model_wrapper.cc:305 ComposeQnnGraph] Qnn_OpConfig node name: node_token_15 package_name: qti.aisw QNN_op_type: MatMul num_of_inputs: 2 num_of_outputs: 1 num_of_params: 2
 node_inputs:
 name=node id=2 version=1 type=QNN_TENSOR_TYPE_NATIVE dataFormat=0 dataType=QNN_DATATYPE_UFIXED_POINT_8 rank=2 dimensions=(2 3 ) memType=QNN_TENSORMEMTYPE_RAW quantizeParams: encodingDefinition=QNN_DEFINITION_DEFINED quantizationEncoding=QNN_QUANTIZATION_ENCODING_SCALE_OFFSET scale=0.000980392 offset=-102
 name=node_token_6 id=4 version=1 type=QNN_TENSOR_TYPE_NATIVE dataFormat=0 dataType=QNN_DATATYPE_UFIXED_POINT_8 rank=2 dimensions=(3 4 ) memType=QNN_TENSORMEMTYPE_RAW quantizeParams: encodingDefinition=QNN_DEFINITION_DEFINED quantizationEncoding=QNN_QUANTIZATION_ENCODING_SCALE_OFFSET scale=0.00215686 offset=-46
 node_outputs:
 name=node_token_16 id=5 version=1 type=QNN_TENSOR_TYPE_NATIVE dataFormat=0 dataType=QNN_DATATYPE_UFIXED_POINT_8 rank=2 dimensions=(2 4 ) memType=QNN_TENSORMEMTYPE_RAW quantizeParams: encodingDefinition=QNN_DEFINITION_DEFINED quantizationEncoding=QNN_QUANTIZATION_ENCODING_SCALE_OFFSET scale=0.000441176 offset=-40
 node_params:
 type=QNN_PARAMTYPE_SCALAR name=transpose_in0 value=0
 type=QNN_PARAMTYPE_SCALAR name=transpose_in1 value=0
```

PR #25021, opened 2025-06-11
[QNN EP] Add support for Mean Op in QNN EP

quic-ashwshan:
- Implemented MeanOpBuilder to support ONNX Mean operator in QNN EP.
- Decomposed Mean into a sequence of element-wise Add operations followed by a Div Op.
- Added unit tests for Mean op running on HTP

### Description
Adds support for the ONNX Mean operator in QNN EP via Add + Div decomposition.

### Motivation and Context
Enables execution of models using Mean on QNN backend, improving Op support.


onnxruntime/test/providers/qnn/simple_op_htp_test.cc
```diff
@@ -991,6 +991,60 @@ TEST_F(QnnHTPBackendTests, BinaryOp_And4D) {
...
+}
+
+TEST_F(QnnHTPBackendTests, Mean_TwoInputs_QU8) {
+  RunQDQOpTest<uint8_t>("Mean",
+                        {TestInputDef<float>({1, 2, 2}, false, GetFloatDataInRange(0.0f, 10.0f, 4)),
+                         TestInputDef<float>({1, 2, 2}, false, GetFloatDataInRange(10.0f, 20.0f, 4))},
+                        {},  // No attributes for Mean
+                        13,  // Opset version
+                        ExpectedEPNodeAssignment::All);
+}
```
HectorSVC, line 1052: Please check the doc to make sure there are no steps missing: [https://onnxruntime.ai/docs/execution-providers/QNN-ExecutionProvider.html#add-new-operator-support-in-qnn-ep](https://onnxruntime.ai/docs/execution-providers/QNN-ExecutionProvider.html#add-new-operator-support-in-qnn-ep)

Please make sure the DQ -> Mean -> Q are grouped as node unit? so that this test is running with quantized uint8 instead of fp32 precision with fp16 mode?

Another recent example is: https://github.com/microsoft/onnxruntime/pull/24659

quic-ashwshan: Yes, the Mean Op is also running on Float 32 precision. Let me know if follow up PRs are required. Or I can append to this one as well.

HectorSVC: would be great to be in one PR.

quic-ashwshan: Hi Hector, I think the issue with running the Mean op in QDQ mode is that we are converting this to Add + Div. 
If we have multiple inputs to the Mean op or even 2 inputs, we are essentially creating these intermediate Add output tensors whose encodings are not known to us. It is not recommended to calculate these encodings manually on the fly. This was not an issue with Reciprocal we it is only replacing to Div so the encodings remain the same. Hence we will only be able to run the op in Float precision for QDQ models

quic-ashwshan:
Resolved merge conflict.

quic-ashwshan:
Looks like there is another merge conflict. Can this be resolved during merge? or do I need to pull latest change again?

quic-ashwshan:
> Looks like there is another merge conflict. Can this be resolved during merge? or do I need to pull latest change again?

resolved

HectorSVC (approved):

PR #25022, opened 2025-06-11
[WebNN] Expose Module.webnnRegisterMLContext to EM_ASM block

Honry:
Use WebNN EP from C++ code needs to call the Module.webnnRegisterMLContext from EM_ASM block.

### DIFF
--- a/js/web/lib/wasm/jsep/init.ts
+++ b/js/web/lib/wasm/jsep/init.ts
@@ -278,11 +278,11 @@ export const init = async (
     const backend = new WebNNBackend(env);
     jsepInit('webnn', [
       backend,
-      // jsepReserveTensorId
+      // webnnReserveTensorId
       () => backend.reserveTensorId(),
-      // jsepReleaseTensorId,
+      // webnnReleaseTensorId
       (tensorId: number) => backend.releaseTensorId(tensorId),
-      // jsepEnsureTensor
+      // webnnEnsureTensor
       async (
         sessionId: number | undefined,
         tensorId: number,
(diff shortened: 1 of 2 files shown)


Honry:
@guschmue, @fs-eire, PTAL, thanks!

onnxruntime/wasm/pre-jsep.js
```diff
@@ -104,21 +104,20 @@ Module["jsepInit"] = (name, params) => {
       Module["webnnEnsureTensor"],
       Module.webnnUploadTensor,
       Module["webnnDownloadTensor"],
+      Module.webnnRegisterMLContext,
```
fs-eire, line 107: use `Module["webnnRegisterMLContext"]` directly here.

js/web/lib/wasm/jsep/init.ts
```diff
@@ -283,14 +283,21 @@ export const init = async (
...
+        shape: number[],
+        copyOld: boolean,
+      ) => backend.ensureTensor(sessionId, tensorId, onnxDataType, shape, copyOld),
       // jsepUploadTensor
       (tensorId: number, data: Uint8Array) => {
         backend.uploadTensor(tensorId, data);
       },
       // jsepDownloadTensor
       async (tensorId: number, dstBuffer: ArrayBufferView | ArrayBuffer) => backend.downloadTensor(tensorId, dstBuffer),
+      // jsepRegisterMLContext
```
fs-eire: could you please help to revise the names of those "jsep*" to "webnn*"? I think they are already "webnn*" just the comments are not up-to-date

Honry: Good point, done and PTAL again. Thanks!

PR #25024, opened 2025-06-11
[QNN EP] Add fixing dynamic input shapes in qnn.preprocess

huaychou:
### Description

- Add argument "--dynamic_input_shapes" to fix dynamic input shapes in qnn.preprocess

### Motivation and Context

- Because QNN-EP doesn't support dynamic shape, add this argument to support model with dynamic input dims but static input shape after qnn.preprocess.


huaychou:

HectorSVC (approved):
:shipit:

PR #25026, opened 2025-06-11
[WebGPU EP] adjusts pow op to use sqrt when exponent is 0.5

prathikr:
Resolves the below error for florence2 model by using WebGPU built-in sqrt(x) function instead of pow(x, y) when the exponent is 0.5. The sqrt(x) built-in is both faster and more stable than using pow(x, 0.5).

```
Non-zero status code returned while running Reshape node. Name:'/Reshape_1' Status 
Message: /Users/runner/work/1/s/onnxruntime/core/providers/cpu/tensor/reshape_helper.h:47 
  onnxruntime::ReshapeHelper::ReshapeHelper(const TensorShape &, TensorShapeVector &, bool) input_shape_size == size was false. 
  The input tensor cannot be reshaped to the requested shape. Input shape:{1,576,32}, requested shape:{1,23,23,32}
```


onnxruntime/core/providers/webgpu/math/binary_elementwise_ops.cc
```diff
@@ -293,6 +293,8 @@ std::string GetPowImpl(int lhs_element_type, int /* rhs_element_type */) {
   s << "fn pow_custom(a : input_a_element_t, b : f32) -> input_a_element_t {\n"
        "  if (b == 0.0) {\n"
        "    return input_a_element_t(1.0);\n"
+       "  } else if (a >= input_a_element_t(0.0) && b == 0.5) {\n"
+       "    return input_a_element_t(sqrt(f32(a)));\n"
```
guschmue: why do we need the f32(a) ?

guschmue (changes requested):

onnxruntime/core/providers/webgpu/math/binary_elementwise_ops.cc
```diff
@@ -293,6 +293,8 @@ std::string GetPowImpl(int lhs_element_type, int /* rhs_element_type */) {
   s << "fn pow_custom(a : input_a_element_t, b : f32) -> input_a_element_t {\n"
        "  if (b == 0.0) {\n"
        "    return input_a_element_t(1.0);\n"
+       "  } else if (a >= input_a_element_t(0.0) && b == 0.5) {\n"
+       "    return input_a_element_t(sqrt(" << (lhs_element_type == ONNX_TENSOR_ELEMENT_DATA_TYPE_FLOAT16 ? "f16" : "f32") << "(a)));\n"
```
guschmue: input_a_element_t(a) ?

prathikr: good catch, need to use the round_str here

fs-eire:
please add a new unittest for the fix

onnxruntime/core/providers/webgpu/math/binary_elementwise_ops.cc
```diff
@@ -293,6 +293,8 @@ std::string GetPowImpl(int lhs_element_type, int /* rhs_element_type */) {
   s << "fn pow_custom(a : input_a_element_t, b : f32) -> input_a_element_t {\n"
        "  if (b == 0.0) {\n"
        "    return input_a_element_t(1.0);\n"
+       "  } else if ((input_a_element_t == f32 || input_a_element_t == f16) && (a >= input_a_element_t(0.0) && b == 0.5)) {\n"
```
fs-eire: looks like `input_a_element_t == f32` is not a valid expression in WGSL. Need to use `if` in C++.

onnxruntime/core/providers/webgpu/math/binary_elementwise_ops.cc
```diff
@@ -289,11 +289,18 @@ std::string GetPowImpl(int lhs_element_type, int /* rhs_element_type */) {
...
+  } else {
+    is_float = "false";
+  }
 
   s << "fn pow_custom(a : input_a_element_t, b : f32) -> input_a_element_t {\n"
        "  if (b == 0.0) {\n"
        "    return input_a_element_t(1.0);\n"
-       "  } else if ((input_a_element_t == f32 || input_a_element_t == f16) && (a >= input_a_element_t(0.0) && b == 0.5)) {\n"
+       "  } else if (" << is_float << " && (a >= input_a_element_t(0.0) && b == 0.5)) {\n"
        "    return sqrt(a);\n"
```
fs-eire: you can use 
```c++
if (lhs_element_type == ONNX_TENSOR_ELEMENT_DATA_TYPE_FLOAT ||
    lhs_element_type == ONNX_TENSOR_ELEMENT_DATA_TYPE_FLOAT16) {
  ss << "} else if (a >= input_a_element_t(0.0) && b == 0.5) {\n"
        "  return sqrt(a);\n";
}
```

onnxruntime/core/providers/webgpu/math/binary_elementwise_ops.cc
```diff
@@ -289,22 +289,21 @@ std::string GetPowImpl(int lhs_element_type, int /* rhs_element_type */) {
...
-  std::string is_float;
-  if (lhs_element_type == ONNX_TENSOR_ELEMENT_DATA_TYPE_FLOAT ||
-      lhs_element_type == ONNX_TENSOR_ELEMENT_DATA_TYPE_FLOAT16) {
-    is_float = "true";
-  } else {
-    is_float = "false";
+
+  SS(float_cond, 1024);
+  if (lhs_element_type == ONNX_TENSOR_ELEMENT_DATA_TYPE_FLOAT || lhs_element_type == ONNX_TENSOR_ELEMENT_DATA_TYPE_FLOAT16) {
+    float_cond << " else if (a >= input_a_element_t(0.0) && b == 0.5) {\n"
```
fs-eire: you can just use the existing stream `s`. no need to create a new one

PR #25011, opened 2025-06-10
[webgpu] Bump version of Dawn to 991810463a

fs-eire:
### Description

This PR is a redo of the reverted PR https://github.com/microsoft/onnxruntime/pull/24935.

  - Upgrade version of Dawn to 991810463a.
    - Comparing to previous commit c3999d7e3, it includes the fix to GCC13, so that now Linux build passes.
    - Includes previous merged patches:
      - https://dawn-review.googlesource.com/c/dawn/+/242335
      - https://dawn-review.googlesource.com/c/dawn/+/242634
      - https://dawn-review.googlesource.com/c/dawn/+/243614

  - Remaining patches:
    - `dawn_destroy_buffer_on_destructor.patch`
      **Allow WGPUBufferImpl class to destroy the buffer in the destructor**
      In native implementation, wgpuBufferRelease will trigger the buffer destroy (if refcount decreased to 0). But in emwgpu implementation, the buffer destroy won't happen. This change adds a destructor to the buffer class to destroy the buffer when the refcount is 0 for non-external buffers.
      
    - `dawn_force_enable_f16_nvidia_vulkan.patch`
      **Force enable f16 support for NVIDIA Vulkan**
      Dawn disabled f16 support for NVIDIA Vulkan by default because of crashes in f16 CTS tests (crbug.com/tint/2164).
      Since the crashes are limited to specific GPU models, we patched Dawn to remove the restriction.
      
    - `dawn_allow_non_uniform_for_subgroup_matrix_args.patch` (new)
      **Revert change "[tint] Check uniformity for subgroup matrix builtin arguments"**
      The following change was made in upstream to force the subgroup matrix builtin arguments to be uniform:
      https://dawn-review.googlesource.com/c/dawn/+/236054  (https://issues.chromium.org/issues/403611487)
      Since we use `subgroup_id` as the subgroup matrix builtin argument, we have to revert this change to allow the subgroup matrix builtin arguments to be non-uniform.

PR #25028, opened 2025-06-11
[WebGPU-EP] Disable SubgroupMatrix uniformity checks

jchen10:
This disables the uniformity checks for SubgroupMatrix in the WebGPU EP.


jchen10:
@fs-eire @xhcao PTAL
This may help us roll Dawn as usual without manually reverting the patch.

fs-eire:
Thank you for this change! It's a cleaner solution compared to adding a patch (https://github.com/microsoft/onnxruntime/pull/25011)

PR #25036, opened 2025-06-12
Fix some `OrtStatus*` leaks

edgchen1:
### Description

`OrtStatus*` values should be released with `ReleaseStatus()`. In some places this was not being done.

This change adds a helper function to convert to the internal `Status` type and also release the `OrtStatus*`. Existing leaky usages of the helper function `ToStatus()` were updated.

### Motivation and Context

Fix `OrtStatus*` leaks.


onnxruntime/core/framework/error_code_helper.h
```diff
@@ -8,11 +8,16 @@
...
+
 // Convert onnxruntime::common::Status to OrtStatus*.
 _Ret_notnull_ OrtStatus* ToOrtStatus(const onnxruntime::common::Status& st);
 
 // Convert OrtStatus* to onnxruntime::common::Status.
+// Note: ToStatus() will not release the OrtStatus*. MoveToStatus() will convert and release the OrtStatus*.
 Status ToStatus(const OrtStatus* ort_status, common::StatusCategory category = common::StatusCategory::ONNXRUNTIME);
+
+// Convert OrtStatus* to onnxruntime::common::Status and release the OrtStatus*.
+Status MoveToStatus(OrtStatus* ort_status, common::StatusCategory category = common::StatusCategory::ONNXRUNTIME);
```
edgchen1: I'm open to suggestions for a better name. "move" here is in the sense of C++ move semantics.

snnn (approved):

onnxruntime/core/framework/error_code.cc
```diff
@@ -43,6 +54,21 @@ _Check_return_ _Ret_notnull_ OrtStatus* ORT_API_CALL OrtApis::CreateStatus(OrtEr
 }
 
 namespace onnxruntime {
+
+namespace {
+
+struct OrtStatusDeleter {
```
yuslepukhin, line 64: >truct OrtStatusDeleter { [](http://example.com/codeflow?start=1&length=24)

Ort::Status is also a good way to do RAAI

edgchen1: it's a good idea. I tried it, but I encountered a link error about an undefined reference to OrtGetApiBase. I would like to not add additional build dependencies if possible, so I'll just use the unique_ptr with custom deleter for now.

adrianlizarraga (approved):

onnxruntime/core/framework/error_code_helper.h
skottmckay: Can we instead update ToStatus to take a non-const OrtStatus and do the release? 

From what I've seen there's no usage where you want to preserve the OrtStatus if you're converting to onnxruntime::Status. Similar for Status -> OrtStatus. We're calling this at an internal/external API boundary when returning from a function.

edgchen1: IMO, the name `ToStatus` implies a conversion (and not more) and I'd find it unexpected if the source `OrtStatus*` also got released. we're a bit limited in the expressiveness of the signature since we only have a pointer and not some other type where we can use an rvalue reference argument.

in addition to `MoveToStatus`, I also considered `ConvertToStatusAndRelease`. is that any better?

> From what I've seen there's no usage where you want to preserve the OrtStatus if you're converting to onnxruntime::Status.

this is what I've seen too. I can just remove the version that doesn't release it entirely.

onnxruntime/core/framework/error_code.cc
```diff
@@ -26,6 +30,13 @@ inline OrtStatus* NewStatus(size_t clen) {
   if (buf == nullptr) return nullptr;  // OOM. What we can do here? abort()?
   return new (buf) OrtStatus;
 }
+
+inline void DeleteStatus(OrtStatus* ort_status) {
+#if defined(_MSC_VER) && !defined(__clang__)
+#pragma warning(suppress : 26409)
+#endif
```
skottmckay: Should this push/pop the warning?  #Closed

onnxruntime/core/framework/error_code_helper.h
skottmckay: I find the 'move' slightly misleading as there's a copy of the error message so it's not quite equivalent to a C++ move.

I'd definitely want there to be only one function as I don't think there's a need for the non-releasing version. 

Personally I'd vote for `ToStatusAndRelease`. The `ToStatus` pairs with `ToOrtStatus` so `Convert` seems unnecessary unless you add it to both.

onnxruntime/core/framework/error_code.cc
edgchen1: apparently suppress really does only suppress it for the next line, which is `#endif`. oops.

onnxruntime/core/framework/error_code_helper.h
edgchen1: ok, renamed it to `ToStatusAndRelease`.

skottmckay (approved):
:shipit:

PR #25043, opened 2025-06-13
Use non-CPU device type and id for host accessible memory

skottmckay:
### Description

Use the non-CPU device type and id for host accessible memory to make the link between CPU and the non-CPU device explicit.

Update the data transfer implementations to check vendor id.


onnxruntime/core/providers/cann/npu_data_transfer.cc
```diff
@@ -67,16 +78,11 @@ common::Status NPUDataTransfer::CopyTensorAsync(const Tensor& src, Tensor& dst,
...
-  if (dst_device.Type() == OrtDevice::NPU) {
-    if (src_device.Type() == OrtDevice::CPU) {
-      // copy from pinned memory to NPU, this is non-blocking
-      CANN_RETURN_IF_ERROR(aclrtMemcpyAsync(dst_data,
-                                            bytes,
-                                            src_data,
-                                            bytes,
-                                            ACL_MEMCPY_HOST_TO_DEVICE,
-                                            static_cast<aclrtStream>(stream.GetHandle())));
-    } else if (src_device.Type() == OrtDevice::NPU) {
```
skottmckay, line 79: Swapped the order to match the non-async method above. i.e. check NPU -> NPU first then NPU -> CPU/host accessible. 
The `else if` in the original code is misleading as there are only 3 src options: NPU, host accessible and CPU, so it was really an `else`.

This change applied in a couple of other places as well given implementations for data transfer have been copied from existing EPs when being created.

onnxruntime/core/providers/cann/npu_data_transfer.cc
```diff
@@ -86,17 +92,23 @@ common::Status NPUDataTransfer::CopyTensorAsync(const Tensor& src, Tensor& dst,
...
+      // copy from pinned or CPU memory to NPU, this is non-blocking
+      CANN_RETURN_IF_ERROR(aclrtMemcpyAsync(dst_data,
+                                            bytes,
+                                            src_data,
+                                            bytes,
+                                            ACL_MEMCPY_HOST_TO_DEVICE,
+                                            static_cast<aclrtStream>(stream.GetHandle())));
     }
-  } else if (src_device.Type() == OrtDevice::NPU) {
-    if (dst_device.Type() == OrtDevice::CPU) {
```
skottmckay, line 91: this `if` was unnecessary as CPU in the old code covered CPU and host accessible, so there are no other destination options as we did npu to npu earlier. 

if it was necessary data transfer would have been broken as there's no `else` that would handle other cases.

onnxruntime/core/providers/cuda/cuda_allocator.h
```diff
@@ -53,11 +53,11 @@ class CUDAExternalAllocator : public CUDAAllocator {
 // TODO: add a default constructor
 class CUDAPinnedAllocator : public IAllocator {
  public:
-  CUDAPinnedAllocator(const char* name)
+  CUDAPinnedAllocator(OrtDevice::DeviceId device_id, const char* name)
```
edgchen1, line 56: do we need to associate the pinned memory allocator with a specific device? I didn't see where this device_id is used by the allocator implementation.

skottmckay: I'm not an expert in CUDA but I believe the device is selected outside of cudaMallocHost via cudaSetDevice. I would have expected the device to have been set in general for the EP given an instance is only using one device. So they get the device id now, but there's probably nothing that needs to be done in the current setup.

edgchen1: as far as I can see, cudaSetDevice is called for CUDAAllocator::Alloc/Free but not CUDAPinnedAllocator::Alloc/Free. is a CUDAPinnedAllocator instance intended to only be used with one device?

onnxruntime/core/framework/allocation_planner.cc
```diff
@@ -733,19 +733,22 @@ class PlannerImpl {
...
     // we check which one has higher alignment and use that one if it is so.
     // If the suggested device is CPU, but not the default mem type, then
-    // it is a CPU accessible memory device allocator. They typically have a page aligment
+    // it is a CPU accessible memory device allocator. They typically have a page alignment
     // so that would satisfy the alignment requirement of any other CPU consumers.
     // If one device is not on CPU, we default on the one that is CPU.
     auto determine_device = [](const OrtDevice& output_device, const OrtDevice& suggested_device) -> OrtDevice {
-      if (output_device.Type() == OrtDevice::CPU && suggested_device.Type() == OrtDevice::CPU) {
-        if (output_device.MemType() == OrtDevice::MemType::DEFAULT &&
-            suggested_device.MemType() == OrtDevice::MemType::DEFAULT) {
```
adrianlizarraga, line 742: The above comment mentions "default mem type". Is that still relevant or should it be updated?

onnxruntime/core/framework/debug_node_inputs_outputs_utils.cc
```diff
@@ -399,8 +399,7 @@ void DumpTensor(
   // check tensor is on CPU before dumping it
   auto& tensor_location = tensor.Location();
   if (tensor_location.device.Type() == OrtDevice::CPU ||
```
HectorSVC: tensor_location.device.UsesCpuMemory()

onnxruntime/core/providers/cuda/cuda_allocator.h
```diff
@@ -53,11 +53,11 @@ class CUDAExternalAllocator : public CUDAAllocator {
...
       : IAllocator(
             OrtMemoryInfo(name, OrtAllocatorType::OrtDeviceAllocator,
-                          OrtDevice(OrtDevice::CPU, OrtDevice::MemType::HOST_ACCESSIBLE, OrtDevice::VendorIds::NVIDIA,
-                                    0 /*CPU device always with id 0*/),
+                          OrtDevice(OrtDevice::GPU, OrtDevice::MemType::HOST_ACCESSIBLE, OrtDevice::VendorIds::NVIDIA,
```
HectorSVC, line 59: >GPU [](http://example.com/codeflow?start=47&length=3)

should still be CPU since it's host memory.

onnxruntime/core/providers/cuda/cuda_allocator.h
HectorSVC: CUDA Pinned memory is host memory.

onnxruntime/core/providers/cuda/cuda_allocator.h
```diff
@@ -53,11 +53,11 @@ class CUDAExternalAllocator : public CUDAAllocator {
...
+                                    device_id),
```
HectorSVC, line 60: >device_id [](http://example.com/codeflow?start=36&length=9)

should set device_id to 0 for host memory.

onnxruntime/core/providers/cuda/cuda_execution_provider.cc
```diff
@@ -2846,22 +2846,17 @@
   if (mem_type == OrtMemTypeCPUInput)
     return OrtDevice();
   if (mem_type == OrtMemTypeCPUOutput)
-    return OrtDevice(OrtDevice::CPU, OrtDevice::MemType::HOST_ACCESSIBLE, OrtDevice::VendorIds::NVIDIA,
-                     0 /*CPU device id always be 0*/);
+    return OrtDevice(OrtDevice::GPU, OrtDevice::MemType::HOST_ACCESSIBLE, OrtDevice::VendorIds::NVIDIA,
+                     default_device_.Id());
```
HectorSVC, line 2850: need to revert the change

skottmckay: Can you explain? There was a TODO on line 2859 below saying it's more correct to use the GPU device id. Do we need to share the pinned memory allocator across devices?

HectorSVC: I see. Keep the change for device_id, but need to change to device type back to CPU.

onnxruntime/core/providers/cuda/cuda_allocator.h
skottmckay: I believe the CUDA EP only uses one device per inference session, and the doc for CUDA says the pinned memory lifetime is associated with the cuda device.

https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DEVICE.html#group__CUDART__DEVICE_1g159587909ffa0791bbe4b40187a4c6bb
> Any host memory allocated from this host thread using [cudaMallocHost()](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html#group__CUDART__MEMORY_1gab84100ae1fa1b12eaca660207ef585b) or [cudaHostAlloc()](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html#group__CUDART__MEMORY_1gb65da58f444e7230d3322b6126bb4902) or [cudaHostRegister()](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html#group__CUDART__MEMORY_1ge8d5c17670f16ac4fc8fcb4181cb490c) will have its lifetime associated with device

AFAICS the device shouldn't change once the EP is created. There is an ORT API `SetCurrentGpuDeviceId` function to set it, but that only updates the provider options so implicitly must be called before the EP and its allocators are created as there's nothing that appears to recreate them. If you called that after the EP was created, you'd end up with the CUDA EP having an arena in front of device memory with a mix of allocations from different devices as `CUDAAllocator::Alloc` sets the device each time. 

We call `cudaSetDevice` in the EP ctor, in the per-thread context ctor, when creating a `CudaStream`, and in `CUDAAllocator`. I can understand setting it in the EP ctor and the per-thread context as `cudaSetDevice` is thread-local. Not sure it's necessary elsewhere or alternatively that we're doing it in the right places. 

Theoretically an app could have multiple inference sessions that use difference devices, and the thread used to call `InferenceSession::Run` on could change. If that happened, I think it would imply needing to call `setCudaDevice` at the start of the usage of a thread that is new or could have been used with a different inference session. so maybe instead of setting it in `CUDAAllocator::Alloc` and when creating a `CudaStream` we might only need to do it at the start of `InferenceSession::Run` (which would equate to inside of `CUDAExecutionProvider::OnRunStart`).

onnxruntime/core/providers/cuda/cuda_execution_provider.cc
skottmckay: Why does the device type need to be CPU? The main premise of the PR is to make a clearer association between the actual device when the memory is host accessible. That requires setting both the device to the non-CPU value and the device id. 

Is something missing an update that would not preserve the current behavior or treatment of host accessible memory?

HectorSVC: It's confusing while saying a device memory (GPU memory) is host (CPU) accessible. Because physically it's CPU memory. May makes more sense in another way like the CPU memory is device (GPU, NPU) accessible.

skottmckay: Isn't the typical terminology (at least for GPU) to call it 'host accessible' memory, so using 'device accessible' could also be confusing? And if we avoid 'device accessible' having a device id with type of CPU is going to be confusing.

FWIW according to AI it isn't necessarily always physically CPU memory.

Platform / Architecture | Memory Physically on | Host Accessible? | Use Case
-- | -- | -- | --
Integrated GPU | Host (RAM) | ✅ | Shared graphics memory
PCIe BAR mapping | GPU (VRAM) | ✅ | GPUDirect, CUDA P2P
Unified Memory (UMA) | Shared fabric | ✅ | Apple M1, AMD APU
Vulkan w/ HOST_VISIBLE | GPU (VRAM) | ✅ | Explicit GPU buffer sharing

jywu-mysoft: agree. host accessible memory is memory cpu can access but the memory does not need to reside on cpu.

HectorSVC:
>   if (strcmp(name1, onnxruntime::CPU) == 0) {

A question not relate to this PR. How would we expect user to create IOBinding start from OrtMemoryInfo but without this string compare (user need to specify string like "Cuda" for example), for the case that EP is created from auto EP path?

---
Refers to: onnxruntime/core/framework/allocator.cc:190 in f05df1d. [](commit_id = f05df1d2e6c4b76ee760fe4200375ad89ff549e5, deletion_comment = False)

skottmckay:
>   if (strcmp(name1, onnxruntime::CPU) == 0) {

I believe we need a way for the user to retrieve the OrtMemoryInfo for model inputs/outputs after partitioning is done.

With that, they should also be able to call GetSharedAllocator to allocate memory on that device if needed, and create an OrtValue for the device memory.

Other missing piece is an API function to copy between a src and dst OrtValue using the data transfer implementation from the EP.

---
In reply to: [3025030899](https://github.com/microsoft/onnxruntime/pull/25043#issuecomment-3025030899) [](http://example.com/codeflow?ancestors=3025030899)

---
Refers to: onnxruntime/core/framework/allocator.cc:190 in f05df1d. [](commit_id = f05df1d2e6c4b76ee760fe4200375ad89ff549e5, deletion_comment = False)

PR #25055, opened 2025-06-13
Revert "Improve Windows ETW callback registration and fix issues"

tianleiwu:
### Description

Revert [Improve Windows ETW callback registration and fix issues](https://github.com/microsoft/onnxruntime/pull/24877) to unblock python packaging pipeline.

### Motivation and Context

Python packaging pipeline is failing due to the changes in the PR.


snnn:
We should also revert 74028e4bdcfa3069113b657bc1e2f32389372dfa , because PR #24877 was for fixing crashes caused by that change. However, it is not an easy job.

PR #25056, opened 2025-06-13
Disable RMSNorm in onnx_backend_test.py and Topk(uint_64)

titaiwangms:
Missing test failures from ONNX 1.18 integration. 

1. The RMSNorm test models is invalid in terms of ONNX spec (e.g. Model rms_normalization_2d_axis1_expanded failed to load:Node () Op (Range) [ShapeInferenceError] Input to 'Range' op should be scalars (Tensor with only one element and shape empty)).
Screenshot 2025-06-13 120308

2. TopK with uint64 is currently not supported.


titaiwangms:
@snnn @tianleiwu I think this should be able to skip the failing onnx backend tests that are added since ONNX==1.18.0 (I have not tested it on CI machine though.).

tianleiwu (approved):

PR #25070, opened 2025-06-14
Allocator and data transfer support for plugin EP API

skottmckay:
### Description

Add allocator and data transfer infrastructure for plugin EP API

Allocators are created via the OrtEpFactory using OrtMemoryInfo that as added to the OrtEpDevice instances the factory returns. This allows allocators to be created outside of an inference session and shared.

When a library is loaded a default instance of each allocator is added to the shared allocators if there is no existing allocator (e.g. user provided custom allocator).
CreateSharedAllocator can be used to replace this default instance with a user configured one. e.g. add an arena or provide other configuration options that are passed through to the OrtEpFactory's CreateAllocator function.
 
Similarly IDataTransfer is supported by the factory implementing OrtDataTransferImpl, which will also enable data transfer outside of a session. That will be added in a future PR as the synchronization requirements need to be figured out and will affect the public API.

### DIFF
--- a/include/onnxruntime/core/framework/allocator.h
+++ b/include/onnxruntime/core/framework/allocator.h
@@ -7,10 +7,17 @@
 
 #include "core/common/common.h"
 #include "core/framework/allocator_stats.h"
+#include "core/session/abi_key_value_pairs.h"
 // some enums are defined in session/onnxruntime_c_api.h but used in ortdevice.h/ortmemory.h
 #include "core/session/onnxruntime_c_api.h"
 #include "core/framework/ortdevice.h"
 #include "core/framework/ortmemoryinfo.h"
+
+namespace onnxruntime {
+namespace common {
+class Status;
+}
+}  // namespace onnxruntime
 
 // This configures the arena based allocator used by ORT
 // See docs/C_API.md for details on what these mean and how to choose these values
@@ -37,6 +44,26 @@ struct OrtArenaCfg {
   int max_dead_bytes_per_chunk;           // use -1 to allow ORT to choose the default
   int initial_growth_chunk_size_bytes;    // use -1 to allow ORT to choose the default
   int64_t max_power_of_two_extend_bytes;  // use -1 to allow ORT to choose the default
+
+  bool IsValid() {
+    return arena_extend_strategy >= -1 && arena_extend_strategy <= 1 &&
+           initial_chunk_size_bytes >= -1 &&
+           max_dead_bytes_per_chunk >= -1 &&
+           initial_growth_chunk_size_bytes >= -1 &&
+           max_power_of_two_extend_bytes >= -1;
+  }
+
+  // config key names that we parse in FromKeyValuePairs
+  struct ConfigKeyNames {
+    static constexpr const char* ArenaExtendStrategy = "arena.extend_strategy";
+    static constexpr const char* InitialChunkSizeBytes = "arena.initial_chunk_size_bytes";
+    static constexpr const char* MaxDeadBytesPerChunk = "arena.max_dead_bytes_per_chunk";
+    static constexpr const char* InitialGrowthChunkSizeBytes = "arena.initial_growth_chunk_size_bytes";
+    static constexpr const char* MaxPowerOfTwoExtendBytes = "arena.max_power_of_two_extend_bytes";
+    static constexpr const char* MaxMem = "arena.max_mem";
+  };
+
+  static onnxruntime::common::Status FromKeyValuePairs(const OrtKeyValuePairs& kvps, OrtArenaCfg& cfg);
 };
 
 namespace onnxruntime {
--- a/include/onnxruntime/core/session/environment.h
+++ b/include/onnxruntime/core/session/environment.h
@@ -28,6 +28,9 @@ class EpFactoryInternal;
 class InferenceSession;
 struct IExecutionProviderFactory;
 struct SessionOptions;
+namespace plugin_ep {
+class DataTransfer;
+}  // namespace plugin_ep
 
 /**
    Provides the runtime environment for onnxruntime.
@@ -75,7 +78,7 @@ class Environment {
    * Registers an allocator for sharing between multiple sessions.
    * Return an error if an allocator with the same OrtMemoryInfo is already registered.
    */
-  Status RegisterAllocator(AllocatorPtr allocator);
+  Status RegisterAllocator(OrtAllocator* allocator);
 
   /**
    * Creates and registers an allocator for sharing between multiple sessions.
@@ -120,6 +123,15 @@ class Environment {
   const std::vector<const OrtEpDevice*>& GetOrtEpDevices() const {
     return execution_devices_;
   }
+
+  // return a shared allocator from a plugin EP or custom allocator added with RegisterAllocator
+  OrtAllocator* GetSharedAllocator(const OrtMemoryInfo& mem_info);
+
+  Status CreateSharedAllocator(const OrtEpDevice& ep_device,
+                               OrtDeviceMemoryType mem_type, OrtAllocatorType allocator_type,
+                               const OrtKeyValuePairs* allocator_options, OrtAllocator** allocator);
+  Status ReleaseSharedAllocator(const OrtEpDevice& ep_device, OrtDeviceMemoryType mem_type);
+
 #endif  // !defined(ORT_MINIMAL_BUILD)
   ~Environment();
 
@@ -130,11 +142,32 @@ class Environment {
                     const OrtThreadingOptions* tp_options = nullptr,
                     bool create_global_thread_pools = false);
 
+  Status RegisterAllocatorImpl(AllocatorPtr allocator, bool replace_existing);
+  Status UnregisterAllocatorImpl(const OrtMemoryInfo& mem_info, bool error_if_not_found = true);
+  Status CreateSharedAllocatorImpl(const OrtEpDevice& ep_device,
+                                   const OrtMemoryInfo& memory_info, OrtAllocatorType allocator_type,
+                                   const OrtKeyValuePairs* allocator_options, OrtAllocator** allocator,
+                                   bool replace_existing);
+
   std::unique_ptr<logging::LoggingManager> logging_manager_;
   std::unique_ptr<onnxruntime::concurrency::ThreadPool> intra_op_thread_pool_;
   std::unique_ptr<onnxruntime::concurrency::ThreadPool> inter_op_thread_pool_;
   bool create_global_thread_pools_{false};
+
+  std::mutex mutex_;
+
+  // shared allocators from various sources.
+  // CreateAndRegisterAllocator[V2]: IAllocator allocators created by ORT
+  // RegisterAllocator: IAllocatorImplWrappingOrtAllocator custom allocators registered by the user.
+  //                    TODO: How can we detect registration of an allocator from an InferenceSession?
+  // OrtEpDevice: We create a default shared IAllocatorImplWrappingOrtAllocator for each OrtEpDevice memory info.
   std::vector<AllocatorPtr> shared_allocators_;
+
+  // RegisterAllocator and CreateSharedAllocator pointers. Used for GetAllocator.
+  // Every instance here is also in shared_allocators_.
+  std::unordered_set<OrtAllocator*> shared_ort_allocators_;
+
+  using OrtAllocatorUniquePtr = std::unique_ptr<OrtAllocator, std::function<void(OrtAllocator*)>>;
 
 #if !defined(ORT_MINIMAL_BUILD)
   // register EPs that are built into the ORT binary so they can take part in AutoEP selection
--- a/include/onnxruntime/core/session/onnxruntime_c_api.h
+++ b/include/onnxruntime/core/session/onnxruntime_c_api.h
@@ -142,6 +142,9 @@ extern "C" {
 // __VA_ARGS__ on Windows and Linux are different
 #define ORT_API(RETURN_TYPE, NAME, ...) RETURN_TYPE ORT_API_CALL NAME(__VA_ARGS__) NO_EXCEPTION
 
+#define ORT_API_T(RETURN_TYPE, NAME, ...) \
+  RETURN_TYPE(ORT_API_CALL* NAME)(__VA_ARGS__) NO_EXCEPTION
+
 #define ORT_API_STATUS(NAME, ...)                                                                   \
   _Success_(return == 0) _Check_return_ _Ret_maybenull_ OrtStatusPtr ORT_API_CALL NAME(__VA_ARGS__) \
   NO_EXCEPTION ORT_MUST_USE_RESULT
@@ -238,7 +241,7 @@ typedef enum OrtLoggingLevel {
   ORT_LOGGING_LEVEL_VERBOSE,  ///< Verbose informational messages (least severe).
   ORT_LOGGING_LEVEL_INFO,     ///< Informational messages.
   ORT_LOGGING_LEVEL_WARNING,  ///< Warning messages.
-  ORT_LOGGING_LEVEL_ERROR,    ///< Error messages.
+  ORT_LOGGING_LEVEL_ERROR,    ///< Error messages.M
   ORT_LOGGING_LEVEL_FATAL,    ///< Fatal error messages (most severe).
 } OrtLoggingLevel;
 
@@ -4561,7 +4564,8 @@ struct OrtApi {
    *  \param[in] provider_options_values value of the provider options map
    *  \param[in] num_keys Length of the provider options map
    */
-  ORT_API2_STATUS(CreateAndRegisterAllocatorV2, _Inout_ OrtEnv* env, _In_ const char* provider_type, _In_ const OrtMemoryInfo* mem_info, _In_ const OrtArenaCfg* arena_cfg,
+  ORT_API2_STATUS(CreateAndRegisterAllocatorV2, _Inout_ OrtEnv* env, _In_ const char* provider_type,
+                  _In_ const OrtMemoryInfo* mem_info, _In_ const OrtArenaCfg* arena_cfg,
                   _In_reads_(num_keys) const char* const* provider_options_keys, _In_reads_(num_keys) const char* const* provider_options_values, _In_ size_t num_keys);
 
   /** \brief Run the model asynchronously in a thread owned by intra op thread pool
@@ -5367,6 +5371,91 @@ struct OrtApi {
                   _In_ uint32_t vendor_id, _In_ int16_t device_id, _In_ enum OrtDeviceMemoryType mem_type,
                   _In_ size_t alignment, enum OrtAllocatorType allocator_type,
                   _Outptr_ OrtMemoryInfo** out);
+
+  /** \brief Get the OrtMemoryInfo for the device.
+   *
+   * \param[in] ep_device The OrtEpDevice instance to query.
+   * \return A pointer to the OrtMemoryInfo for the device.
+   *
+   * \since Version 1.23
+   */
+  ORT_API_T(const OrtMemoryInfo*, EpDevice_MemoryInfo, _In_ const OrtEpDevice* ep_device);
+
+  /** \brief Create/replace a shared allocator for the OrtEpDevice in the OrtEnv.
+   *
+   * OrtEpDevice maps to the EP factory, and the factory provides the allocator implementation.
+   *
+   * OrtDeviceMemoryType_DEFAULT is always supported for non-CPU based devices.
+   * OrtDeviceMemoryType_HOST_ACCESSIBLE is optional and dependent on the device as to whether it's required/supported.
+   *
+   * If a shared allocator already exists and is a custom allocator added with RegisterAllocator, an error is returned
+   * as it must be remove with UnregisterAllocator first.
+   *
+   * If a shared allocator already exists for the OrtEpDevice and OrtDeviceMemoryType, it is replaced. This allows
+   * changing the shared allocator configuration from the default. e.g. adding an arena.
+   *
+   * \param[in] env The OrtEnv instance to create the shared allocator in.
+   * \param[in] ep_device The OrtEpDevice instance to create the shared allocator for.
+   * \param[in] mem_type The memory type to use for the shared allocator.
+   * \param[in] allocator_type The type of allocator to create (e.g. OrtAllocatorType::OrtArenaAllocator).
+   * \param[in] allocator_options Optional key-value pairs to configure the allocator.
+   * \param[out] allocator A pointer to the created shared allocator. Owned by the OrtEnv.
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \since Version 1.23
+   */
+  ORT_API2_STATUS(CreateSharedAllocator, _In_ OrtEnv* env, _In_ const OrtEpDevice* ep_device,
+                  _In_ OrtDeviceMemoryType mem_type, _In_ OrtAllocatorType allocator_type,
+                  _In_opt_ const OrtKeyValuePairs* allocator_options,
+                  _Outptr_opt_ OrtAllocator** allocator);
+
+  /** \brief Get a shared allocator from the OrtEnv.
+   *
+   * By default there is a shared allocator created for all OrtEpDevice instances, so if you get the OrtMemoryInfo
+   * from the OrtEpDevice using EpDevice_MemoryInfo a shared allocator is guaranteed to exist.
+   *
+   * This will also match and return custom allocators created with RegisterAllocator.
+   *
+   * \param[in] env The OrtEnv instance to get the shared allocator from.
+   * \param[in] mem_info The OrtMemoryInfo instance to get the shared allocator for.
+   * \return A pointer to the shared allocator, or nullptr if no shared allocator exists for the given memory info.
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \since Version 1.23
+   */
+  ORT_API_T(OrtAllocator*, GetSharedAllocator, _In_ OrtEnv* env, _In_ const OrtMemoryInfo* mem_info);
+
+  /** \brief Release a shared allocator from the OrtEnv for the OrtEpDevice and memory type.
+   *
+   * This will release the shared allocator for the given OrtEpDevice and memory type.
+   * If no shared allocator exists, this is a no-op.
+   *
+   * \param[in] env The OrtEnv instance to release the shared allocator from.
+   * \param[in] ep_device The OrtEpDevice instance to release the shared allocator for.
+   * \param[in] mem_type The memory type of the shared allocator to release.
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \since Version 1.23
+   */
+  ORT_API2_STATUS(ReleaseSharedAllocator, _In_ OrtEnv* env, _In_ const OrtEpDevice* ep_device,
+                  _In_ OrtDeviceMemoryType mem_type);
+
+  /** \brief Get a const pointer to the raw data inside a tensor
+   *
+   * Used to read the internal tensor data directly.
+   * \note The returned pointer is valid until the \p value is destroyed.
+   *
+   * \param[in] value A tensor type (string tensors are not supported)
+   * \param[out] out Filled in with a pointer to the internal storage
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \since Version 1.23
+   */
+  ORT_API2_STATUS(GetTensorData, _In_ const OrtValue* value, _Outptr_ const void** out);
 };
 
 /*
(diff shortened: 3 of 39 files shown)


include/onnxruntime/core/framework/allocator.h
```diff
@@ -37,6 +44,26 @@ struct OrtArenaCfg {
   int max_dead_bytes_per_chunk;           // use -1 to allow ORT to choose the default
   int initial_growth_chunk_size_bytes;    // use -1 to allow ORT to choose the default
   int64_t max_power_of_two_extend_bytes;  // use -1 to allow ORT to choose the default
+
+  bool IsValid() {
```
skottmckay, line 42: Moving validity check that was scattered around to be here.

include/onnxruntime/core/session/environment.h
```diff
@@ -75,7 +78,7 @@ class Environment {
    * Registers an allocator for sharing between multiple sessions.
    * Return an error if an allocator with the same OrtMemoryInfo is already registered.
    */
-  Status RegisterAllocator(AllocatorPtr allocator);
+  Status RegisterAllocator(OrtAllocator* allocator);
```
skottmckay, line 93: Separating external from internal usage by making the public method take OrtAllocator*. This enables GetSharedAllocator to return any of the OrtAllocator based instances.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -6096,7 +6185,109 @@ struct OrtCompileApi {
...
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \since Version 1.23.
+   */
+  ORT_API2_STATUS(CopyTensors, _In_ void* this_ptr,
+                  _In_reads_(num_tensors) const OrtValue** src_tensors,
+                  _In_reads_(num_tensors) OrtValue** dst_tensors,
+                  _In_reads_(num_tensors) OrtSyncStream** streams,
+                  _In_ size_t num_tensors);
```
skottmckay: Design choice: Minimize the EP API by having a single function that can do asynchronous batched copies. The library is free to implement that as synchronous copies. This avoids having additional functions for copying a single tensor.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -6096,7 +6185,109 @@ struct OrtCompileApi {
 
 ORT_RUNTIME_CLASS(Ep);
 ORT_RUNTIME_CLASS(EpFactory);
+ORT_RUNTIME_CLASS(MemoryDevice);  // opaque class to wrap onnxruntime::OrtDevice
```
skottmckay: I couldn't avoid adding OrtDevice to the API as IDataTransfer uses it. To try and disambiguate/clarify I called it OrtMemoryDevice as we have OrtHardwareDevice, and calling it `OrtDevice` in the API felt too vague.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -142,6 +142,9 @@ extern "C" {
 // __VA_ARGS__ on Windows and Linux are different
 #define ORT_API(RETURN_TYPE, NAME, ...) RETURN_TYPE ORT_API_CALL NAME(__VA_ARGS__) NO_EXCEPTION
 
+#define ORT_API_T(RETURN_TYPE, NAME, ...) \
+  RETURN_TYPE(ORT_API_CALL* NAME)(__VA_ARGS__) NO_EXCEPTION
+
```
skottmckay: We were missing the NO_EXCEPTION on a number of functions in the initial plugin EP setup. This macro should help keep things consistent by adding NO_EXCEPTION like ORT_API2_STATUS does.

onnxruntime/core/framework/allocator.cc
```diff
@@ -15,6 +17,61 @@
 
 #include "core/framework/bfc_arena.h"
 
+using Status = onnxruntime::common::Status;
+
+Status OrtArenaCfg::FromKeyValuePairs(const OrtKeyValuePairs& kvps, OrtArenaCfg& cfg) {
```
skottmckay, line 23: Support creation from OrtKeyValuePairs to keep the public API minimal. We use OrtKeyValuePairs in general for configuration/options/metadata and this avoids having to add OrtArenaCfg as an additional parameter in some functions like CreateSharedAllocator.

onnxruntime/core/framework/data_transfer_manager.cc
```diff
@@ -16,6 +16,20 @@ Status DataTransferManager::RegisterDataTransfer(std::unique_ptr<IDataTransfer>
   return Status::OK();
 }
 
+Status DataTransferManager::UnregisterDataTransfer(IDataTransfer* data_transfer) {
```
skottmckay, line 19: Need to unregister if the plugin EP library is unloaded.

onnxruntime/core/session/allocator_adapters.cc
```diff
@@ -115,30 +125,31 @@ ORT_API_STATUS_IMPL(OrtApis::CreateAllocator, const OrtSession* sess,
...
+  if (!ort_env) {
     return OrtApis::CreateStatus(ORT_INVALID_ARGUMENT, "Env is null");
   }
 
   if (!mem_info) {
     return OrtApis::CreateStatus(ORT_INVALID_ARGUMENT, "OrtMemoryInfo is null");
   }
 
-  auto st = env->CreateAndRegisterAllocator(*mem_info, arena_cfg);
+  auto& env = ort_env->GetEnvironment();
```
skottmckay, line 186: Removed unnecessary call from here -> OrtEnv.CreateAndRegisterAllocator -> Environment.CreateAndRegisterAllocator.

As the Environment instance is available from OrtEnv there's no need to it to have a CreateAndRegisterAllocator method that forwards.

include/onnxruntime/core/framework/allocator.h
```diff
@@ -7,11 +7,18 @@
...
 #include "core/framework/allocator_stats.h"
+#include "core/session/abi_key_value_pairs.h"
 // some enums are defined in session/onnxruntime_c_api.h but used in ortdevice.h/ortmemory.h
 #include "core/session/onnxruntime_c_api.h"
 #include "core/framework/ortdevice.h"
 #include "core/framework/ortmemoryinfo.h"
 
+namespace onnxruntime {
+namespace common {
+class Status;
```
edgchen1: nit: common.h already includes status.h, so we don't need a forward declaration. we could explicitly include status.h too for clarity.

a user of this file would need a complete Status definition anyway to call FromKeyValuePairs().

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -5367,6 +5372,90 @@ struct OrtApi {
...
+   *
+   * \since Version 1.23
+   */
+  ORT_API_T(const OrtMemoryInfo*, EpDevice_MemoryInfo, _In_ const OrtEpDevice* ep_device);
+
+  /** \brief Create/replace a shared allocator for the OrtEpDevice in the OrtEnv.
+   *
+   * OrtEpDevice maps to the EP factory, and the factory provides the allocator implementation.
+   *
+   * OrtDeviceMemoryType_DEFAULT is always supported for non-CPU based devices.
```
edgchen1: is the device memory allocator always required? does this also apply to the QNN EP, which only has the HtpSharedMemoryAllocator?

onnxruntime/core/framework/allocator.cc
```diff
@@ -15,6 +17,61 @@
...
+                             "Failed to parse value for ", key, " from ", str);
+    }
+
+    return Status::OK();
+  };
+
+  int64_t value = 0;
+  if (auto it = kvps.entries.find(ConfigKeyNames::ArenaExtendStrategy); it != kvps.entries.end()) {
+    ORT_RETURN_IF_ERROR(from_string(it->first, it->second, value));
+    cfg.arena_extend_strategy = onnxruntime::narrow<int32_t>(value);
```
edgchen1: consider using `ParseStringWithClassicLocale()` in parse_string.h. it handles conversions to specific int types directly.

include/onnxruntime/core/session/onnxruntime_c_api.h
skottmckay: Discussed offline. It's EP dependent as to what is available.

An NVIDIA EP would have a GPU device allocator as well as a host accessible allocator (CUDA pinned memory) and nodes assigned to the EP their input on GPU.
c.f. nodes assigned to QNN take their input on CPU, but can also use HTP shared memory (host accessible), so would only provide the host accessible allocator and not a device allocator, even though the OrtEpDevice would be for a GPU or NPU.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -5930,6 +5932,90 @@ struct OrtApi {
...
+                  _In_ OrtDeviceMemoryType mem_type, _In_ OrtAllocatorType allocator_type,
+                  _In_opt_ const OrtKeyValuePairs* allocator_options,
+                  _Outptr_opt_ OrtAllocator** allocator);
+
+  /** \brief Get a shared allocator from the OrtEnv.
+   *
+   * By default there is a shared allocator created for all OrtEpDevice instances, so if you get the OrtMemoryInfo
+   * from the OrtEpDevice using EpDevice_MemoryInfo a shared allocator is guaranteed to exist.
+   *
+   * This will also match and return custom allocators added with RegisterAllocator.
```
adrianlizarraga, line 5978: Does this also match allocators created with `CreateAndRegisterAllocator` and `CreateAndRegisterAllocatorV2`?

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -169,6 +225,75 @@ struct OrtEpApi {
...
+   *
+   * This is required for OrtDataTransferImpl (which implements onnxruntime::IDataTransfer) where the OrtMemoryDevice
+   * is used in the CanCopy and CopyTensors functions.
+   *
+   * \param[in] memory_info The OrtMemoryInfo instance to get the memory device from.
+   * \return The OrtMemoryDevice associated with the OrtMemoryInfo instance.
+   *
+   * \since Version 1.23.
+   */
+  ORT_API_T(const OrtMemoryDevice*, OrtMemoryInfo_GetMemoryDevice, _In_ const OrtMemoryInfo* memory_info);
```
adrianlizarraga: Should this be named `MemoryInfo_GetMemoryDevice` (without the Ort prefix)? Not sure what the convention should be.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -475,6 +600,45 @@ struct OrtEpFactory {
    * \since Version 1.22.
    */
   void(ORT_API_CALL* ReleaseEp)(OrtEpFactory* this_ptr, struct OrtEp* ep);
+
+  /** \brief Create an OrtAllocator for the given OrtMemoryInfo.
+   *
+   * This is used to create an allocator that an execution provider created by the factory requires.
```
adrianlizarraga: Is this supposed to say `created but the factory requires`?

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -169,6 +225,75 @@ struct OrtEpApi {
    * \since Version 1.23.
    */
   ORT_API_T(const char*, NodeComputeContext_NodeName, _In_ const OrtNodeComputeContext* context);
+
+  /** \brief Register an allocator with the OrtEpDevice.
+   *
+   * This allows an EP to provide OrtMemoryInfo for DEFAULT and HOST_ACCESSIBLE memory types as needed.
```
adrianlizarraga: nit: could we clarify that this only allows the EP to provide one DEFAULT memory type and one HOST_ACCESSIBLE memory type. The implementation would keep replacing memory infos if this function is called multiple times with HOST_ACCESSIBLE (for example).

onnxruntime/test/autoep/library/ep.cc
```diff
@@ -0,0 +1,500 @@
...
+                                                 void** compute_state);
+  static OrtStatus* ORT_API_CALL ComputeImpl(OrtNodeComputeInfo* this_ptr, void* compute_state,
+                                             OrtKernelContext* kernel_context);
+  static void ORT_API_CALL ReleaseStateImpl(OrtNodeComputeInfo* this_ptr, void* compute_state);
+
+  ExampleEp& ep;
+};
+
+ExampleEp::ExampleEp(ExampleEpFactory& factory, const std::string& name, const Config& config, const OrtLogger& logger)
+    : ApiPtrs(static_cast<const ApiPtrs&>(factory)),
```
adrianlizarraga: I think this needs a call `OrtEp()` to ensure that `OrtEp` function pointers are initialized to `nullptr`. There are some optional `OrtEp` functions, like `OnRunStart`, that should be set to nullptr.

onnxruntime/test/autoep/library/ep.cc
```diff
@@ -0,0 +1,500 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
adrianlizarraga, line 1: Thanks for separating out the example ep files. Much easier to understand.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -5930,6 +5932,90 @@ struct OrtApi {
...
+   *
+   * \param[in] env The OrtEnv instance to get the shared allocator from.
+   * \param[in] mem_info The OrtMemoryInfo instance to get the shared allocator for.
+   * \return A pointer to the shared allocator, or nullptr if no shared allocator exists for the given memory info.
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
```
edgchen1, line 5987: the docs say it returns OrtStatus but it doesn't. IMO, most C APIs should return OrtStatus so we have a way of reporting errors.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -5930,6 +5932,90 @@ struct OrtApi {
...
+   * \param[in] mem_type The memory type of the shared allocator to release.
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \since Version 1.23
+   */
+  ORT_API2_STATUS(ReleaseSharedAllocator, _In_ OrtEnv* env, _In_ const OrtEpDevice* ep_device,
+                  _In_ OrtDeviceMemoryType mem_type);
+
+  /** \brief Get a const pointer to the raw data inside a tensor
```
edgchen1, line 6010: nice to finally add this :)

onnxruntime/core/framework/plugin_data_transfer.cc
```diff
@@ -0,0 +1,68 @@
...
+  std::vector<OrtValue> values;
+  values.resize(src_dst_pairs.size() * 2);
+
+  for (size_t i = 0; i < src_dst_pairs.size(); ++i) {
+    const auto& pair = src_dst_pairs[i];
+
+    // we need to remove the const from the src to wrap it in an OrtValue.
+    // it's passed to the impl as a const OrtValue, and the deleter is a no-op so this should be safe.
+    Tensor* src_tensor = const_cast<Tensor*>(&(pair.src.get()));
+    values[i * 2].Init(reinterpret_cast<void*>(src_tensor), ml_tensor_type, no_op_deleter);
```
adrianlizarraga: please feel free to ignore this as this const_cast seems safe as long as the lifetime of the Tensor exceeds that of the OrtValue.

If storing unowned data into OrtValue becomes common, we could create a new overload of `OrtValue::Init()` that does this internally and knows to use a no-op deleter.

```c++
  void Init(const void* pUnownedData, onnxruntime::MLDataType type) {
    const std::function<void(void*)> no_op_deleter = [](void*) {};  // could be a static variable
    data_.reset(const_cast<void*>(pUnownedData), no_op_deleter);
    type_ = type;
  }
```

onnxruntime/core/framework/plugin_data_transfer.cc
```diff
@@ -0,0 +1,68 @@
...
+  for (size_t i = 0; i < src_dst_pairs.size(); ++i) {
+    src_values.push_back(&values[i * 2]);
+    dst_values.push_back(&values[i * 2 + 1]);
+    streams.push_back(nullptr);  // static_cast<OrtSyncStream*>(src_dst_pairs[i].src_stream));
+  }
+
+  auto* status = impl_.CopyTensors(&impl_, src_values.data(), dst_values.data(), streams.data(),
+                                   src_dst_pairs.size());
+
+  return status == nullptr ? Status::OK() : ToStatusAndRelease(status);
```
adrianlizarraga: nit: I think `return ToStatusAndRelease(status)` would work for both cases.

onnxruntime/core/platform/windows/debug_alloc.cc
```diff
@@ -253,7 +253,8 @@ Memory_LeakCheck::~Memory_LeakCheck() {
         string.find("testing::internal::Mutex::ThreadSafeLazyInit") == std::string::npos &&
         string.find("testing::internal::ThreadLocalRegistryImpl::GetThreadLocalsMapLocked") == std::string::npos &&
         string.find("testing::internal::ThreadLocalRegistryImpl::GetValueOnCurrentThread") == std::string::npos &&
-        string.find("PyInit_onnxruntime_pybind11_state") == std::string::npos) {
+        string.find("PyInit_onnxruntime_pybind11_state") == std::string::npos &&
+        string.find("google::protobuf::internal::InitProtobufDefaultsSlow") == std::string::npos) {
```
adrianlizarraga, line 257: Thanks! I'm assuming this gets rid of the protobuf related memleaks that are reported when running tests in a debug build. It makes it hard to parse out relevant memleaks.

onnxruntime/core/framework/allocator.cc
```diff
@@ -15,6 +18,58 @@
...
+  }
```
edgchen1: `narrow` failure has a pretty generic error message. how about something like this:

```suggestion
  const auto from_string = [](const std::string& key, const std::string& str, auto& value) -> Status {
    if (!onnxruntime::TryParseStringWithClassicLocale(str, value)) {
      return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT, "Failed to parse value for ", key, " from ", str);
    }

    return Status::OK();
  };

  if (auto it = kvps.entries.find(ConfigKeyNames::ArenaExtendStrategy); it != kvps.entries.end()) {
    ORT_RETURN_IF_ERROR(from_string(it->first, it->second, cfg.arena_extend_strategy));
  }
```

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -5930,6 +5932,90 @@ struct OrtApi {
...
+   *
+   * \param[in] env The OrtEnv instance to release the shared allocator from.
+   * \param[in] ep_device The OrtEpDevice instance to release the shared allocator for.
+   * \param[in] mem_type The memory type of the shared allocator to release.
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \since Version 1.23
+   */
+  ORT_API2_STATUS(ReleaseSharedAllocator, _In_ OrtEnv* env, _In_ const OrtEpDevice* ep_device,
```
adrianlizarraga, line 6007: Is this only called for allocators created with `CreateSharedAllocator` (but not `GetSharedAllocator`)?

onnxruntime/core/framework/plugin_data_transfer.cc
edgchen1: nit:
```suggestion
  return ToStatusAndRelease(impl_.CopyTensors(&impl_, src_values.data(), dst_values.data(), streams.data(),
                                              src_dst_pairs.size()));
```

include/onnxruntime/core/session/environment.h
```diff
@@ -140,12 +152,33 @@ class Environment {
...
+  std::mutex mutex_;
+
+  // shared allocators from various sources.
+  // CreateAndRegisterAllocator[V2]: IAllocator allocators created by ORT
+  // RegisterAllocator: IAllocatorImplWrappingOrtAllocator custom allocators registered by the user.
+  //                    TODO: How can we detect registration of an allocator from an InferenceSession?
+  // OrtEpDevice: We create a default shared IAllocatorImplWrappingOrtAllocator for each OrtEpDevice memory info.
   std::vector<AllocatorPtr> shared_allocators_;
 
+  // RegisterAllocator and CreateSharedAllocator pointers. Used for GetAllocator.
```
adrianlizarraga: Is this supposed to be `Used for GetSharedAllocator`?

onnxruntime/core/framework/plugin_data_transfer.cc
```diff
@@ -0,0 +1,68 @@
...
+  OrtValue src, dst;
+  Tensor* src_tensor_ptr = const_cast<Tensor*>(&src_tensor);
+  src.Init(reinterpret_cast<void*>(src_tensor_ptr), ml_tensor_type, no_op_deleter);
+  dst.Init(reinterpret_cast<void*>(&dst_tensor), ml_tensor_type, no_op_deleter);
+  const OrtValue* src_ptr = &src;
+  OrtValue* dst_ptr = &dst;
+  OrtSyncStream* stream_ptr = nullptr;  // static_cast<OrtSyncStream*>(stream);
+  auto* status = impl_.CopyTensors(&impl_, &src_ptr, &dst_ptr, &stream_ptr, 1);
+
+  return status == nullptr ? Status::OK() : ToStatusAndRelease(status);
```
edgchen1:
```suggestion
  return ToStatusAndRelease(impl_.CopyTensors(&impl_, &src_ptr, &dst_ptr, &stream_ptr, 1));
```

onnxruntime/core/session/environment.cc
```diff
@@ -111,22 +95,105 @@ Status Environment::RegisterAllocator(AllocatorPtr allocator) {
...
+      shared_ort_allocators_.erase(it);
+    }
+
+    shared_ort_allocators_.insert(allocator);
+  }
+
+  return status;
+}
+
+Status Environment::RegisterAllocatorImpl(AllocatorPtr allocator, bool replace_existing) {
```
adrianlizarraga: In what situation would `replace_existing` be set to `false`? Just curious, maybe I missed it.

onnxruntime/core/session/environment.cc
```diff
@@ -381,6 +420,12 @@ Status Environment::CreateAndRegisterAllocatorV2(const std::string& provider_typ
 
 Environment::~Environment() = default;
 
+OrtAllocator* Environment::GetSharedAllocator(const OrtMemoryInfo& mem_info) {
+  // doesn't matter whether we match a custom allocator or an EP allocator so match_name is false
```
adrianlizarraga: Should this lock the mutex too since `shared_ort_allocators_` could be read while it is modified (via Register) in a different thread?

onnxruntime/core/session/ep_api.cc
```diff
@@ -86,6 +90,63 @@ ORT_API(const char*, NodeComputeContext_NodeName, _In_ const OrtNodeComputeConte
   return compute_context->node_name;
 }
 
+ORT_API_STATUS_IMPL(EpDevice_AddAllocatorInfo, _In_ OrtEpDevice* ep_device,
+                    _In_ const OrtMemoryInfo* allocator_memory_info) {
+  const OrtDevice& info = allocator_memory_info->device;
+  if (info.MemType() == OrtDevice::MemType::DEFAULT) {
+    ep_device->device_memory_info = allocator_memory_info;
+  } else {
+    ep_device->host_accessible_memory_info = allocator_memory_info;
```
adrianlizarraga: nit: should we explicitly check for `OrtDevice::MemType::HOST_ACCESSIBLE` to make it more explicit and return an error status otherwise (in case a new enum value is added later).

onnxruntime/core/session/ep_plugin_provider_interfaces.cc
```diff
@@ -442,5 +449,42 @@ Status PluginExecutionProvider::SetEpDynamicOptions(gsl::span<const char* const>
...
+  }
+
+  return std::make_unique<plugin_ep::DataTransfer>(*data_transfer_impl);
+}
+
+std::vector<AllocatorPtr> PluginExecutionProvider::CreatePreferredAllocators() {
+  std::vector<AllocatorPtr> allocators;
+  allocators.reserve(allocator_mem_infos_.size());
+
+  for (const auto* memory_info : allocator_mem_infos_) {
```
adrianlizarraga, line 480: Is `allocator_mem_infos_` initialized with elements? Maybe I missed it.

onnxruntime/core/session/onnxruntime_c_api.cc
```diff
@@ -1090,6 +1090,13 @@ ORT_API_STATUS_IMPL(OrtApis::GetTensorMutableData, _Inout_ OrtValue* value, _Out
   API_IMPL_END
 }
 
+ORT_API_STATUS_IMPL(OrtApis::GetTensorData, _Inout_ const OrtValue* value, _Outptr_ const void** output) {
```
adrianlizarraga, line 1093: Nice. now we have a const-correct version of this API.

include/onnxruntime/core/session/onnxruntime_c_api.h
skottmckay: It won't currently. Those require a session to be created to acquire via the API, so the allocator has a lifetime bound to the session.

But it should cover all devices if we update the OrtEpFactory implementations for internal EPs to add their allocators to the shared ones.

onnxruntime/test/autoep/library/ep.cc
```diff
@@ -0,0 +1,500 @@
...
+    RETURN_IF_ERROR(ort_api.GetTensorShapeElementCount(type_shape, &num_elems));
+
+    size_t num_dims = 0;
+    RETURN_IF_ERROR(ort_api.GetDimensionsCount(type_shape, &num_dims));
+
+    shape.resize(num_dims, 0);
+    RETURN_IF_ERROR(ort_api.GetDimensions(type_shape, shape.data(), shape.size()));
+
+    const float* raw_data = nullptr;
+    RETURN_IF_ERROR(ort_api.GetTensorMutableData(const_cast<OrtValue*>(input), (void**)&raw_data));  // No const-correct API?
```
adrianlizarraga:
```suggestion
    RETURN_IF_ERROR(ort_api.GetTensorData(input, reinterpret_cast<const void**>(&raw_data)));
```

include/onnxruntime/core/session/onnxruntime_c_api.h
skottmckay: Yes. Create/Release are based on OrtEpDevice. GetSharedAllocator is OrtMemoryInfo based.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
skottmckay: I think dropping the prefix in function names is good. You're calling through the ORT API pointer which provides the ORT context.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
skottmckay: No. But I can try and reword it. The factory provides allocator creation, so needs to provide all the allocators that any EP it creates requires.

onnxruntime/core/framework/plugin_data_transfer.cc
skottmckay: The wrapper OrtValue instances for src/dst tensors only exist for the duration of this function. As the Tensor being wrapped is an input to the function that should guarantee the lifetime of the Tensor exceeds that of the OrtValue.

As this might be the only place we need this wrapping, and the object lifetime semantics are very clear here but might not be in a general "wrap existing Tensor in OrtValue" usage, I'll hold off on pushing changes into OrtValue for now.

onnxruntime/core/platform/windows/debug_alloc.cc
skottmckay: at least the latest ones I saw. I thought we had a protobuf entry in here previously, so I was surprised I needed to add it.

onnxruntime/core/session/environment.cc
skottmckay: User registers custom allocator. That should override any default shared allocator from an EP

onnxruntime/core/session/ep_plugin_provider_interfaces.cc
skottmckay: Great catch. Got lost in a merge. Added back.

onnxruntime/core/session/onnxruntime_c_api.cc
skottmckay: Kinda surprised I needed to add it. People must love const_cast more than I thought. :P

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -10,8 +10,64 @@ extern "C" {
...
+   */
+  ORT_API_T(void, Release, _In_ void* this_ptr);
+
+  /** \brief Check if the implementation can copy between the source and destination memory devices.
+   *
+   * \param[in] this_ptr Pointer to the OrtDataTransferImpl instance.
+   * \param[in] src_memory_device Source OrtMemoryDevice to copy from.
+   * \param[in] dst_memory_device Destination OrtMemoryDevice to copy to.
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
```
adrianlizarraga: this function doesn't return an `OrtStatus`

onnxruntime/core/session/environment.cc
adrianlizarraga: Isn't that scenario setting `replace_existing` to `true`? I'm wondering if it is ever set to `false` (i'm not seeing it).

skottmckay: ah sorry. previously the EP registration would create the allocator and call RegisterAllocatorImpl with it set to false. I changed the library load to use CreateSharedAllocatorImpl after adding shared_ort_allocators_. will remove.

PR #25082, opened 2025-06-16
vitisai/imp/graph.cc: fix graph_save() for models with external data

schuermans-slx:
### Description

This fixes the export of a model to disk (e.g. for debugging) in case the original model uses an external data file.
Before, the external data file was not found if the model is exporter to a different directory than the original model.

### Motivation and Context

Reading initializer data from an external data file happens on demand and uses the name of the model file to determine the directory containing the data file.
Thus, when creating a new Model object in memory for the export to disk, it is important to use the path to the original model file in that object instead of the path to the model file to be exported to disk. This ensures that the original data file can still be found for reading the initializer data during the export.


HectorSVC (approved):
:shipit:

PR #25090, opened 2025-06-17
Fix the input order in gen_qnn_ctx_onn_model.py

chuteng-quic:
### Description
In the original script, the order of inputs in the generated ONNX model follows the order in the JSON file instead of the id.
Therefore, the script will first sort the inputs by ID before wrapping the context into the new model.

### Motivation and Context
If we use gen_qnn_ctx_onnx_model.py, it will wrap the context based on the order in qnn_model_net.json instead of the id in qnn_model_net.json, which may result in the onnx model having an incorrect order of inputs.

### DIFF
--- a/onnxruntime/python/tools/qnn/gen_qnn_ctx_onnx_model.py
+++ b/onnxruntime/python/tools/qnn/gen_qnn_ctx_onnx_model.py
@@ -11,13 +11,16 @@ from onnx import TensorProto, helper
 
 
 class QnnTensorStruct:
-    def __init__(self):
-        self.name = ""
-        self.onnx_data_type = TensorProto.FLOAT
-        self.is_quantized = False
-        self.scale = 0.0
-        self.offset = 0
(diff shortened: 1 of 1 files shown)


chuteng-quic:
Hi @HectorSVC , the check failure doesn't seem to be related to my change.

D:\a\_work\onnxruntime\onnxruntime\onnxruntime\core\providers\webgpu\webgpu_context.cc:135 onnxruntime::webgpu::WebGpuContext::Initialize::<lambda_80b5f183808855c70c5c2feabd4839b3>::()::<lambda_4e6515c14c43e84ea6bc837c5ef2449d>::<lambda_invoker_cdecl> status == wgpu::RequestDeviceStatus::Success was false. Failed to get a WebGPU device: D3D12 create command queue failed with DXGI_ERROR_DEVICE_REMOVED (0x887A0005)

Could you help re-trigger the CI checks?

PR #25110, opened 2025-06-18
[MLAS] Add 8-bit weights ARM64 Gemm implementation

hariharans29:
### Description
Enable 8-bit weights Gemm on ARM64 via MLAS

1. Supports 2 flavors of the 8-bit Gemm kernel - one uses `vdotq` (U8U8) and the other uses `vusdotq` (U8S8) on platforms where I8MM is supported.

2. Provides access to these new MLAS Gemm kernels via the `MatmulNBits` contrib operator

3. Tests:
    **MLAS**
  3 new sets of tests:
    - `SQ8BitQuantA` : Tests the dynamic activation quantization MLAS kernel (`fp32 -> uint8_t` or `fp32 -> int8_t` on I8MM platforms)
    - `SQ8BitPrepack`: Tests the prepacking of the weights for the 8-bit Gemm kernels
    - `SQ8BitGemm`: Tests the 8-bit Gemm kernels

    **MatmulNBits contrib tests**
    - Enables the 8-bit Gemm tests on ARM64 (previously only enabled on x86)  

### Motivation and Context
Enable 8-bit weights Gemm on ARM64 via MLAS

Based on work and contribution by @fajin-corp 

Phi-4-mini-instruct perf numbers (before and after this change):

image


onnxruntime/test/contrib_ops/matmul_8bits_test.cc
```diff
@@ -24,7 +24,7 @@
 #include "core/session/ort_env.h"
 #include "core/util/qmath.h"
 
-#if (defined(MLAS_TARGET_AMD64_IX86) && !defined(USE_DML) && !defined(USE_WEBGPU) && !defined(USE_COREML)) || defined(USE_CUDA) || defined(USE_WEBGPU)
+#if ((defined(MLAS_TARGET_AMD64_IX86) || defined(MLAS_TARGET_ARM64)) && !defined(USE_DML) && !defined(USE_WEBGPU) && !defined(USE_COREML)) || defined(USE_CUDA) || defined(USE_WEBGPU)
```
hariharans29: Enables tests on ARM64

onnxruntime/core/mlas/lib/qnbitgemm_kernel_neon.cpp
```diff
@@ -278,6 +452,77 @@ UseKleidiAI(size_t K, size_t BlkLen, bool HasZp)
...
+
+            c_blk += ldc * RowsHandled;
+            a_blksum_row += BlockCountK * RowsHandled;
+            RowsRemaining -= RowsHandled;
+        }
+    }
+
+    if constexpr (QuantAUnsigned) {
+        {
+            assert(QuantBBlkSum2 != nullptr);
```
vraspar: The assertion fails on android `[QuantAUnsigned = true]: assertion "QuantBBlkSum2 != nullptr" failed` when testing with  `phi-4-mini-instruct: cpu-int4-kquant-block-128-mixed-acc-level-4/v3`

hariharans29: This should be fixed now.

onnxruntime/contrib_ops/cpu/quantization/matmul_nbits.cc
```diff
@@ -193,10 +193,39 @@ Status MatMulNBits<T1>::PrePack(const Tensor& tensor, int input_idx, /*out*/ All
     auto qptr = tensor.DataRaw();
     auto scale_ptr = scales ? scales->DataRaw() : nullptr;
     packed_b_ = IAllocator::MakeUniquePtr<void>(alloc, packed_b_size_, true);
+#if defined(MLAS_TARGET_ARM64)
+    if (nbits_ == 8) {
+      // On ARM64, the scales should NOT be provided while packing the weights for nbits_ == 8.
+      MlasQNBitGemmPackQuantBData(N_, K_, nbits_, block_size_, compute_type_, qptr, packed_b_.get(), nullptr,
```
edgchen1: is it possible for the ARM64 8b implementation to ignore the scales? then maybe we can simplify this call site

onnxruntime/contrib_ops/cpu/quantization/matmul_nbits.cc
```diff
@@ -193,10 +193,39 @@ Status MatMulNBits<T1>::PrePack(const Tensor& tensor, int input_idx, /*out*/ All
...
+        auto sptr = tensor.Data<float>();
+        MlasQNBitGemmPackQuantBData(N_, K_, nbits_, block_size_, compute_type_, nullptr, packed_b_.get(), sptr,
+                                    has_zp_input_, nullptr, nullptr);
+        is_packed = false;
+      } else if (input_idx == InputIndex::zero_points && packed_b_ != nullptr) {
+        auto zptr = tensor.Data<uint8_t>();
+        MlasQNBitGemmPackQuantBData(N_, K_, nbits_, block_size_, compute_type_, nullptr, packed_b_.get(), nullptr,
+                                    has_zp_input_, zptr, nullptr);
+        is_packed = false;
+      }
```
edgchen1, line 223: this code looks like a duplicate of the `#ifdef MLAS_TARGET_AMD64_IX86` block below.

could we consolidate them?

e.g.,
```c++
const bool should_pack_scale_and_zp_inputs = [&]() {
#if defined(MLAS_TARGET_AMD64_IX86)
  return true;
#else
  return nbits_ == 8;
#endif
}();

if (should_pack_scale_and_zp_inputs) {
    // ...
}
```

onnxruntime/core/mlas/inc/mlas_qnbit.h
```diff
@@ -48,6 +48,7 @@ struct MLAS_QNBIT_GEMM_DATA_PARAMS {
     const T* QuantBScale = nullptr;             ///< address of scale values of quantized B, one per block
     const void* QuantBZeroPoint = nullptr;          ///< optional address of zero point values of quantized B, one per block
     const T* QuantBBlkSum = nullptr;            ///< optional address of scale * zp, one per block
+    const T* QuantBBlkSum2 = nullptr;           ///< optional address of scale * accumulate(quant - zp), one per block. Used when QuantA is uint8.
```
edgchen1: can we come up with a better name than "BlkSum2"?

onnxruntime/core/mlas/lib/mlasi.h
```diff
@@ -1297,6 +1298,8 @@ struct MLAS_PLATFORM {
     // TODO: move to cpuinfo
     bool Avx2Supported_ = false;
     bool Avx512Supported_ = false;
+    bool ArmNeonQuantAUnsigned = false;
```
edgchen1: can this name be more specific? without context, "A" is not very meaningful.

onnxruntime/core/mlas/lib/sqnbitgemm_kernel_neon_int8.cpp
```diff
@@ -187,6 +187,233 @@ QuantizeARow_CompInt8(
...
+        return v;
+    } else {
+        assert(count == 1);
+        float32x4_t v = vdupq_n_f32(0.0f);
+        v = vld1q_lane_f32(src, v, 0);
+        return v;
+    }
+}
+
+template <bool QuantAUnsigned>
```
edgchen1: nit: maybe the name `IsQuantAUnsigned` would make it more obvious that it is a boolean value.

onnxruntime/core/mlas/lib/sqnbitgemm_kernel_neon_int8.cpp
```diff
@@ -187,6 +187,233 @@ QuantizeARow_CompInt8(
...
+                const uint16x8_t i_8_u16_0 = vmovl_u8(vget_low_u8(v_16_u8));
+                const uint16x8_t i_8_u16_1 = vmovl_high_u8(v_16_u8);
+                sum_8_i16_0 = vaddq_u16(sum_8_i16_0, i_8_u16_0);
+                sum_8_i16_1 = vaddq_u16(sum_8_i16_1, i_8_u16_1);
+            } else {
+                const int8x16_t v_16_i8 = vcombine_s8(vqmovn_s16(v_8_i16_0), vqmovn_s16(v_8_i16_1));
+                vst1q_s8(blob + k + kk, v_16_i8);
+
+                // accumulate Sum(a_i)
+                const int16x8_t i_8_i16_0 = vmovl_s8(vget_low_s8(v_16_i8));
```
edgchen1: why do we need to widen from s8 to s16 again? don't we already have s16 values in `v_8_i16_0` and `v_8_i16_1`?

onnxruntime/core/mlas/lib/sqnbitgemm_kernel_neon_int8.cpp
```diff
@@ -1439,6 +1666,723 @@ SQ4BitGemmKernel_CompInt8(
...
+    size_t CountM,
+    size_t CountN,
+    size_t BlockCountK,
+    const float* Bias,
+    size_t ldc
+)
+{
+    constexpr size_t NCols4 = 4;
+    constexpr size_t NCols8 = 8;
+    constexpr size_t NRows2 = 2;
```
edgchen1: general - would a better name be "MRows2" since it seems to be rows of M?

onnxruntime/test/contrib_ops/matmul_8bits_test.cc
```diff
@@ -25,7 +25,7 @@
...
```
edgchen1, line 31: nit: line length

onnxruntime/core/mlas/lib/qnbitgemm_kernel_neon.cpp
```diff
@@ -199,6 +224,152 @@ SQ4BitGemmPackQuantBDataAndBlkSum(
...
+            *(QuantBScaleBegin + n4 * BlockCountK + k_blk * 4 + n4_res) = QuantBScale;
+        } else { // full 8 cols
+            *(QuantBScaleBegin + n8 * BlockCountK + k_blk * 8 + n8_res) = QuantBScale;
+        }
+    });
+}
+
+/**
+ * 4 rows x 8 cols pack together, along all K. Then 4 rows x 4 cols, along all K.
+ * When rol < 4, keep original layout.
```
edgchen1: is "rol" meant to be "row"?

onnxruntime/core/mlas/lib/mlasi.h
hariharans29: Generally in the context of Matmul development, I think "A" is reasonably well-known to be the first input - but I can go ahead and name it `ArmNeonQuantActivationsToUnsigned ` to re-inforce that it is indicating activations (first input). Would that be reasonable ?

EDIT: Saw the next comment and `ArmNeonIsQuantActivationsUnsigned` seems like a better name. Please let me know if that is okay.

onnxruntime/contrib_ops/cpu/quantization/matmul_nbits.cc
hariharans29: Fixed in latest refactor

onnxruntime/contrib_ops/cpu/quantization/matmul_nbits.cc
hariharans29: Fixed in latest refactor

onnxruntime/core/mlas/lib/sqnbitgemm_kernel_neon_int8.cpp
hariharans29: Done

onnxruntime/core/mlas/lib/sqnbitgemm_kernel_neon_int8.cpp
hariharans29: Good idea, done

onnxruntime/test/contrib_ops/matmul_8bits_test.cc
hariharans29: Fixed

onnxruntime/core/mlas/lib/qnbitgemm_kernel_neon.cpp
hariharans29: Yeah, I think it was (or it was meant to be row length shortened to rol). Either way, I have changed it to be more clear.

onnxruntime/core/mlas/inc/mlas_qnbit.h
hariharans29: Done

onnxruntime/core/mlas/lib/sqnbitgemm_kernel_neon_int8.cpp
hariharans29: Great catch - thanks! I have removed these redundant instructions and re-used those existing values in `v_8_i16_0` and `v_8_i16_1`

onnxruntime/core/mlas/lib/qnbitgemm.cpp
```diff
@@ -774,7 +779,8 @@ SQ8BitGemm_CompInt8(
                 bias,
                 ldc,
                 ABlockSum,
-                b_blk_sum
+                b_blk_sum,
+                b_blk_sum2
```
edgchen1: nit: can we rename the variable to something like `a_blk_zp_correction`?

onnxruntime/core/mlas/inc/mlas_qnbit.h
```diff
@@ -48,6 +48,15 @@ struct MLAS_QNBIT_GEMM_DATA_PARAMS {
     const T* QuantBScale = nullptr;             ///< address of scale values of quantized B, one per block
     const void* QuantBZeroPoint = nullptr;          ///< optional address of zero point values of quantized B, one per block
     const T* QuantBBlkSum = nullptr;            ///< optional address of scale * zp, one per block
+
+    /// <summary>
+    /// Optional address of scale * accumulate(quant - zp), one per block. Used when A is quantized is uint8.
+    /// When converting the activation input (A) to uint8, we first convert the values first convert to
+    /// int8 and then add a "bias" of +128 to convert the range of values from [-128, +127] to [0, +255].
+    /// This optional input helps to "de-bias" the output of the +128 bias added to the activation input.
```
edgchen1: thanks for adding the description. it's helpful.

```suggestion
    /// Optional address of scale * accumulate(quant - zp), one per block. Used when A is quantized to uint8.
    /// When converting the activation input (A) to uint8, we first convert the values to
    /// int8 and then add a "bias" of +128 to convert the range of values from [-128, +127] to [0, +255].
    /// This optional input helps to "de-bias" the output of the +128 bias added to the activation input.
```

a few other points:
- this value is not so much "optional" as it is "only used when A is quantized to uint8", right?
- in the formula `scale * accumulate(quant - zp)`, it might be good to explain that scale/zp/quant are for B (if I understand correctly). it wasn't obvious to me, since we are quantizing both B and A.

jywu-msft (dismissed):

hariharans29:
Pending final perf validation and accuracy verification post PR comments addressing

onnxruntime/core/mlas/lib/platform.cpp
```diff
@@ -593,16 +592,22 @@ Return Value:
...
+#if defined(__linux__)
+
         this->GemmU8U8Dispatch = &MlasGemmU8X8DispatchUmmla;
         this->GemmU8S8Dispatch = &MlasGemmU8X8DispatchUmmla;
         this->GemmS8S8Dispatch = &MlasGemmS8S8DispatchSmmla;
-    }
 #endif
+    }
+
+    this->ArmNeonIsQuantActivationsUnsigned = HasI8MMInstructions ? false : true;
```
edgchen1, line 609: if `ArmNeonIsQuantActivationsUnsigned` is just the negation of `MLAS_CPUIDINFO::GetCPUIDInfo().HasArmNeon_I8MM()`, can we just keep track of the latter? it seems that whether quantized A is unsigned or not (based off of whether i8mm is supported) is an MLAS implementation detail.

perhaps within the implementation, we can have a helper function like `bool IsQuantActivationsUnsigned() { return !MLAS_CPUIDINFO::GetCPUIDInfo().HasArmNeon_I8MM(); }`.

onnxruntime/core/mlas/lib/sqnbitgemm_kernel_avx2.cpp
```diff
@@ -602,7 +602,8 @@ SQ8BitGemmKernel_BlkSum_CompInt8_avx2(
     const float* Bias,
     size_t ldc,
     const float* ABlockSum,
-    const float* QuantBBlkSum
+    const float* QuantBBlkSum,
+    const float* /*QuantBBlkSum2*/
```
edgchen1, line 606: general nit: there are still some occurrences of the old name "BlkSum2"

edgchen1 (approved):

onnxruntime/core/mlas/lib/qnbitgemm.cpp
hariharans29: Fixed.

onnxruntime/core/mlas/inc/mlas_qnbit.h
hariharans29: Fixed.

onnxruntime/core/mlas/lib/platform.cpp
hariharans29: Will address in a future PR

onnxruntime/core/mlas/lib/sqnbitgemm_kernel_avx2.cpp
hariharans29: Will address in a future PR

hariharans29:
Merging as the React Native pipeline failure is unrelated to this change

PR #25117, opened 2025-06-19
Fix: delay CUDADriverWrapper instantiation to avoid uncaught exception

krahenbuhl:
Fix: delay CUDADriverWrapper instantiation to avoid uncaught exceptions when CUDA is unavailable

### Description

This PR moves the static instantiation of CUDADriverWrapper from a class-level static field to a function-local static inside CUDADriverWrapper::GetInstance(). This change ensures that the CUDA driver is only loaded when the instance is actually needed, rather than at static initialization time. It preserves the singleton behavior while deferring instantiation to runtime.

### Motivation and Context

When libcuda.so.1 is not available on the system, the constructor of CUDADriverWrapper throws an exception. Previously, this exception was triggered during static initialization, leading to an uncatchable std::terminate() and process termination. By moving the instance into GetInstance() as a function-local static, the exception can now be caught by client code (e.g., in try/catch), allowing graceful fallback when CUDA is unavailable.

### DIFF
--- a/onnxruntime/contrib_ops/cuda/bert/tensorrt_fused_multihead_attention/cudaDriverWrapper.cc
+++ b/onnxruntime/contrib_ops/cuda/bert/tensorrt_fused_multihead_attention/cudaDriverWrapper.cc
@@ -136,10 +136,8 @@ CUresult CUDADriverWrapper::cuLaunchKernel(
       f, gridDimX, gridDimY, gridDimZ, blockDimX, blockDimY, blockDimZ, sharedMemBytes, hStream, kernelParams, extra);
 }
 
-// Initialize the singleton instance
-CUDADriverWrapper CUDADriverWrapper::instance;
-
 const CUDADriverWrapper* CUDADriverWrapper::GetInstance() {
+  static CUDADriverWrapper instance;
   return &instance;
 }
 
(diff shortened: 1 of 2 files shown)


krahenbuhl:
@microsoft-github-policy-service agree company="Pix4D"

tianleiwu (approved):

onnxruntime/contrib_ops/cuda/bert/tensorrt_fused_multihead_attention/cudaDriverWrapper.cc
```diff
@@ -136,10 +136,8 @@ CUresult CUDADriverWrapper::cuLaunchKernel(
       f, gridDimX, gridDimY, gridDimZ, blockDimX, blockDimY, blockDimZ, sharedMemBytes, hStream, kernelParams, extra);
 }
 
-// Initialize the singleton instance
-CUDADriverWrapper CUDADriverWrapper::instance;
-
 const CUDADriverWrapper* CUDADriverWrapper::GetInstance() {
+  static CUDADriverWrapper instance;
```
snnn, line 140: Please be very careful that this object will be deleted before all global variables.  Please check who depends on this.

tianleiwu: The usage looks fine: https://github.com/search?q=repo%3Amicrosoft%2Fonnxruntime%20%20CUDADriverWrapper%3A%3AGetInstance&type=code

PR #25119, opened 2025-06-19
Update index.md

ThomasXilinx:
### Description
Changed Xilinx to AMD
Removed preview from VitisAI EP in the Summary of supported Execution Providers.

### Motivation and Context
Xilinx is now AMD
This EP is no longer in preview.


ThomasXilinx:
There are no details on what is causing the "CheckLinks / Check website links" failure. How can this be resolved? We would like to this doc update to be published rapidly. Thanks.

XIAY0009:
Would be great to get preview removed, change Xilinx to AMD as well as to add Vitis-AI under IoT/Edge/Mobile column as well, thank you for your help.

PR #25120, opened 2025-06-20
Add dynamic bucket cache mode to improve peak and avg gpu buffer memory usage

feich-ms:
### Description
Add a dynamic bucket cache mode to improve the GPU buffer memory usage, including reducing the peak memory and average memory usage without any other performance regressions.

### Motivation and Context

Default buckets solution uses default buckets with coarse-grained bucket keys to cache buffers and don’t update the bucket keys over the runs while dynamic buckets solution uses dynamic buckets with fine-grained bucket keys to cache buffers and update the bucket keys over the runs. Dynamic buckets solution will overall request less buffers than default buckets solution but keeps the similar cache missed buffer bytes. We summarized the optimization solution details in this doc https://microsoftapc-my.sharepoint.com/:w:/g/personal/feich_microsoft_com/Eb1nPP2LyDZNq9bpG-jXp64BbEVcbbMzYeOPJHc3ITUu-w?e=tQLJVo.

A experimental result shows that dynamic buckets solution can reduce the **peak memory** usage by **16.9%** aginst Edge build-in AI Phi3 model with GenAI tool command: python benchmark/python/benchmark_e2e.py -i /Users/feich/models/phi3 -l 1000 -o perf_result.csv

**Open questions**: if this solution works well, is it possible to replace current default bucket solution?

### Test Device 
- Windows with NVIDIA GeForce RTX 5080 chip 
- Mac Mini with Apple M2 Pro chip 

### Model Memory and Perf Analysis
The peak/avg memory calculation is based on this commit https://github.com/microsoft/onnxruntime/commit/61cdfb9bfdf5baca7013f1e4f8c39c848b5f8359

The token gen latency and tokens/sec are from the output file of benchmark_e2e.py script
Token Gen Latency -> prompt_processing_latency_ms
Tokens/sec -> token_generation_throughput_tps

**Phi3-Windows**
Optimization Strategy | Peak Memory (MB) | Avg Memory (MB) | Token Gen Latency (ms) | Tokens/sec
-- | -- | -- | -- | --
Default Bucket | 3603.83 | 3127.05 | 7.17 | 139.50
Dynamic Bucket | **2994.94 (+16.90%)** | 2510.18 (+19.73%) | 7.15 (+0.24%) | 139.84 (+0.24%)

**DeepSeek-R1-Windows**
Optimization Strategy | Peak Memory (MB) | Avg Memory (MB) | Token Gen Latency (ms) | Tokens/sec
-- | -- | -- | -- | --
Default Bucket | 2089.03 | 1716.15 | 6.07 | 164.67
Dynamic Bucket | **1752.53 (+16.11%)** | 1345.29 (+21.61%) | 6.10 (-0.43%) | 163.97 (-0.42%)

**LLAMA3.2-1B-Windows**
Optimization Strategy | Peak Memory (MB) | Avg Memory (MB) | Token Gen Latency (ms) | Tokens/sec
-- | -- | -- | -- | --
Default Bucket | 1736.03 | 1424.64 | 3.37 | 296.53
Dynamic Bucket | **1550.81 (+10.67%)** | 1222.38 (+14.20%) | 3.37 (+0.06%) | 296.71 (+0.06%)

**Phi3-Mac**
Optimization Strategy | Peak Memory (MB) | Avg Memory (MB) | Token Gen Latency (ms) | Tokens/sec
-- | -- | -- | -- | --
Default Bucket | 3603.83 | 3121.60 | 7.17 | 139.50
Dynamic Bucket | **2994.94 (+16.90%)** | 2508.87 (+19.63%) | 7.15 (+0.24%) | 139.84 (+0.24%)

**DeepSeek-R1-Mac**
Optimization Strategy | Peak Memory (MB) | Avg Memory (MB) | Token Gen Latency (ms) | Tokens/sec
-- | -- | -- | -- | --
Default Bucket | 2065.88 | 1697.92 | 11.23 | 89.01
Dynamic Bucket | **1739.98 (+15.78%)** | 1344.08 (+20.84%) | 11.22 (+0.11%) | 89.10 (+0.11%)

**LLAMA3.2-1B-Mac**
Optimization Strategy | Peak Memory (MB) | Avg Memory (MB) | Token Gen Latency (ms) | Tokens/sec
-- | -- | -- | -- | --
Default Bucket | 1652.87 | 1361.16 | 8.52 | 117.32
Dynamic Bucket | **1535.11 (+7.12%)** | 1220.86 (+10.31%) | 8.53 (-0.02%) | 117.30 (-0.02%)

### DIFF
--- a/onnxruntime/core/providers/webgpu/buffer_manager.cc
+++ b/onnxruntime/core/providers/webgpu/buffer_manager.cc
@@ -278,20 +278,20 @@ class DynamicBucketCacheManager : public IBufferCacheManager {
   }
 
   size_t CalculateBufferSize(size_t request_size) override {
-    request_size = NormalizeBufferSize(request_size);
+    size_t normalized_request_size = NormalizeBufferSize(request_size);
 
     // Track usage for the current run
-    current_run_usage_[request_size]++;
+    current_run_usage_[normalized_request_size]++;
 
     // Check if we already have a bucket for this size. If not, create a new bucket so that it can cache buffers of
     // this size in the current session run if the buffer is quickly released in the same session.
-    if (buckets_.find(request_size) == buckets_.end()) {
-      buckets_.emplace(request_size, std::vector<WGPUBuffer>());
-      buckets_keys_.push_back(request_size);
+    if (buckets_.find(normalized_request_size) == buckets_.end()) {
+      buckets_.emplace(normalized_request_size, std::vector<WGPUBuffer>());
+      buckets_keys_.push_back(normalized_request_size);
       std::sort(buckets_keys_.begin(), buckets_keys_.end());
     }
 
-    return request_size;
+    return normalized_request_size;
   }
 
   WGPUBuffer TryAcquireCachedBuffer(size_t buffer_size) override {
--- a/onnxruntime/core/providers/webgpu/buffer_manager.cc
+++ b/onnxruntime/core/providers/webgpu/buffer_manager.cc
@@ -276,9 +276,11 @@ class DynamicBucketCacheManager : public IBufferCacheManager {
     // Adjust buckets based on the collected memory patterns every 2 runs.
     // The reason for this is to allow the cache to adapt to the memory usage patterns
     // of previous runs of last completed token generation session.
-    if ((session_id_ + 1) % 2 == 0) {
+    static size_t run_id = 0;
+    if ((run_id + 1) % 2 == 0) {
       AdjustBuckets();
     }
+    ++run_id;
   }
 
   size_t CalculateBufferSize(size_t request_size) override {
--- a/onnxruntime/core/providers/webgpu/buffer_manager.cc
+++ b/onnxruntime/core/providers/webgpu/buffer_manager.cc
@@ -304,7 +304,7 @@ class DynamicBucketCacheManager : public IBufferCacheManager {
     return nullptr;
   }
 
-  void RegisterBuffer(WGPUBuffer buffer, size_t request_size) override {
+  void RegisterBuffer(WGPUBuffer /*buffer*/, size_t /*request_size*/) override {
     // no-op
   }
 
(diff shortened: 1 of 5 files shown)


feich-ms:
Hi @fs-eire @guschmue, I'm working on the BucketCacheManager in buffer_manager.cc to improve peak/average GPU buffer memory usage currently, this is an initial solution I tried to propose, using dynamic buckets to replace default buckets, with dynamic bucket keys and dynamic updating buckets, can you help to review the code and doc and provide feedback if this is a reasonable optimization. CC @qjia7. Thanks.

fs-eire:
The cache algorithm looks good to me. But I have concerns about the `OnRunStart()` and `OnRunEnd()`. Let me explain it.

### Background

When ORT is running, there are all different kinds of states. Those states are managed by different components. For example, a `SessionState` manages the graph and all initializers, a `ExecutionFrame` manages all intermediate values, and a `WebGpuContext` manages the WebGPU Device and all Dawn objects created from it including the buffers.

Each state has its lifecycle. For example, an initializer is created during the session initialization and will live until the session is released. An intermediate value, however, will have a shorter lifecycle that created at the start of a session run and destroyed no more later than the end of a session run.

From the lifecycle perspective, there are 3 different layers of lifecycles:
- global or global-like: for example, `OrtEnv`. it can be simply assumed to be always available.
- per-session: the life cycle is bound to the session. for example, a `SessionState`.
- per-run: the life cycle is bound to an inference run call. for example, a `ExecutionFrame`.

For WebGPU, there are a new different lifecycle layer:
- per-logical-device: the `WebGpuContext`. It's bound to the lifecycle of the `WGPUDevice` object.

### Problem

In the main branch, the buffer manager and the cache manager are all per-logical-device. Since one WebGpuContext instance can be used by multiple sessions, the buffer manager and the buffer cache manager are also used by multiple sessions.

In the current PR, the new cache algorithm is exposing a new requirement:
1. It needs to know when an inference run is started and ended.
2. It requires between the start and end, no buffer operation from a different inference run happen.

While (1) is quite straightforward, (2) is actually missing from the current implementation.

Per the current design, WebGpuContext is not bound to any session state. This assumption is very important because it can safely allow multiple inference runs of different sessions at the same time. This is the problem: (2) is never guaranteed.

### Solution

There are a few things need to change for the framework to support the new cache algorithm. Considering the latest requirement for graph capture, this need to be carefully considered to ensure no design flaw or regression.

I will think about it and come up with a design to work with both the buffer manager optimization and the graph capture feature. I will update it here when it's ready.

PR #25124, opened 2025-06-20
[EP ABI] Add support for creating EP Context models.

adrianlizarraga:
### Description
- Updates `OrtEp::Compile()` to allow a plugin EP to create and return EPContext nodes.
- Updates the example EP plugin to generate an example EPContext model:
image

### Motivation and Context
Adds more of the functionality missing from the EP ABI used for plugin EPs.


onnxruntime/test/autoep/library/example_plugin_ep_utils.h
```diff
@@ -0,0 +1,58 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
adrianlizarraga, line 1: Note: Moved random utility functions from `example_plugin_ep.cc` into this new utils file as the example EP code was getting a bit long.

HectorSVC (approved):

PR #25127, opened 2025-06-20
[ORT EP API] Add some additional ORT EP APIs

edgchen1:
### Description

Adding the following ORT EP APIs:
- `GetPreferredDataLayout()`
- `SetDynamicOptions()`
- `OnRunStart()`
- `OnRunEnd()`

### Motivation and Context

Expose additional EP APIs.


include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -217,6 +227,71 @@ struct OrtEp {
...
+   * \param[in] run_options The run options for this run.
+   *
+   * \note Implementation of this function is optional.
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \since Version 1.23.
+   */
+  OrtStatus*(ORT_API_CALL* OnRunStart)(_In_ OrtEp* this_ptr,
+                                       _In_ const OrtRunOptions* run_options);
```
adrianlizarraga, line 293: I could be wrong, but I suspect that the C APIs for getting string configs from `OrtRunOptions` don't yet exist. There may be other run option APIs that are used by EPs that do not have public C API counterparts. If so, should those APIs be added in this PR or would you prefer a separate PR?

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -217,6 +227,71 @@ struct OrtEp {
   void(ORT_API_CALL* ReleaseNodeComputeInfos)(_In_ OrtEp* this_ptr,
                                               OrtNodeComputeInfo** node_compute_infos,
                                               _In_ size_t num_node_compute_infos);
+
+  /** \brief Get the EP's preferred data layout.
+   *
+   * \note Implementation of this function is optional.
+   *       If not implemented, ORT will assume that this EP prefers the data layout `OrtEpDataLayout::NCHW`.
```
adrianlizarraga, line 248: Thanks for doing this. I forgot to do this for other APIs.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
edgchen1: good point, I'll look into it.

edgchen1: added `GetRunConfigEntry()` to the C API. hopefully that covers it.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -217,6 +227,71 @@ struct OrtEp {
   void(ORT_API_CALL* ReleaseNodeComputeInfos)(_In_ OrtEp* this_ptr,
                                               OrtNodeComputeInfo** node_compute_infos,
                                               _In_ size_t num_node_compute_infos);
+
+  /** \brief Get the EP's preferred data layout.
```
adrianlizarraga, line 245: Should we describe how this function interacts with `OrtEp::GetCapability`? Or is that too much implementation detail?

 i.e., If not set to the default `NCHW`, ONNX Runtime calls `OrtEp::GetCapability` twice:
  - The first call to `OrtEp::GetCapability` provides nodes with the default `NCHW` layout. The execution provider should indicate which nodes would be supported if they had the preferred (non-default) data layout.
  - The second call `OrtEp::GetCapability` provides nodes with the execution provider's preferred data layout and a domain of `"?"`. In this case, additional Transpose nodes may have been inserted into the graph to ensure the correct data layout.

For compiling EPs, `OrtEp::Compile` receives nodes with the preferred data layout and a domain of `"?"`.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -217,6 +227,71 @@ struct OrtEp {
...
+   * \param[out] preferred_data_layout The EP's preferred data layout.
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \since Version 1.23.
+   */
+  OrtStatus*(ORT_API_CALL* GetPreferredDataLayout)(_In_ OrtEp* this_ptr,
+                                                   _Out_ OrtEpDataLayout* preferred_data_layout);
+
+  /** \brief Set dynamic options on this EP.
```
adrianlizarraga, line 260: Should we state that this may be called after the OrtSession has been created via `OrtApi::SetEpDynamicOptions` but before run is called? And maybe link to the common dynamic options:
https://github.com/microsoft/onnxruntime/blob/4e45b295ef5f58abbf68875382765ca8701f501d/include/onnxruntime/core/session/onnxruntime_c_api.h#L4949-L4952

adrianlizarraga (dismissed):

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -135,6 +135,16 @@ struct OrtEpApi {
   ORT_API_T(const char*, NodeComputeContext_NodeName, _In_ const OrtNodeComputeContext* context);
 };
 
+/**
+ * \brief The data layout type that is preferred by an EP.
+ * \since Version 1.23.
+ */
+typedef enum OrtEpDataLayout {
+  NCHW = 0,
```
adrianlizarraga: There have been different conventions for naming enum values. I think most recently we started appending the enum type as a prefix to avoid name conflicts. Ex:

```c++
typedef enum OrtEpDataLayout {
  OrtEpDataLayout_NCHW = 0,
  OrtEpDataLayout_NHWC,
} OrtEpDataLayout
```

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -135,6 +135,16 @@ struct OrtEpApi {
...
+  NHWC,
+  NCHWC,
```
adrianlizarraga: I could be wrong, but I think `NCHWC` is not used by EPs (or maybe only used for L2 opts?). If so, should we remove this for now?

edgchen1: yeah, I didn't find any usages of `DataLayout::NCHWC`. I'll try removing it.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
edgchen1: added a reference to `OrtApi::SetEpDynamicOptions`. I think they can refer to the documentation for that API for more details.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
edgchen1: it seems good to document this interaction between ORT and the EPs somewhere. I'm also not that familiar with the details - e.g., I didn't know about the domain `?`. maybe we can add more API usage documentation in a separate PR.

adrianlizarraga: there's a domain for nhwc. I used `?` because I don't remember what it is exactly :)

onnxruntime/test/autoep/library/example_plugin_ep.cc
```diff
@@ -122,7 +122,7 @@ struct ExampleEp : OrtEp, ApiPtrs {
   };
 
   ExampleEp(ApiPtrs apis, const std::string& name, const Config& config, const OrtLogger& logger)
-      : ApiPtrs(apis), name_{name}, config_{config}, logger_{logger} {
+      : OrtEp(), ApiPtrs(apis), name_{name}, config_{config}, logger_{logger} {
```
edgchen1, line 125: note: now that some `OrtEp` APIs are optional to implement, we should initialize the unimplemented ones to nullptr.

adrianlizarraga (approved):

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
edgchen1: let's update the docs in a separate PR

PR #25131, opened 2025-06-22
[java] Auto EP and compile model support

Craigacp:
### Description
Java API for compile model and EP discovery APIs. Roughly equivalent to the C# version in #24604.

cc: @skottmckay.

I haven't quite got the CMake configured so the Java tests for the ep registration only run when the ONNX Runtime shared provider support is built, but everything else works. I expect that to be a quick fix, but I'm not sure in what conditions it should be built and how we should handle it so I don't know where/when to plumb it through.

### Motivation and Context
API parity for Java.


cmake/onnxruntime_unittests.cmake
```diff
@@ -1657,6 +1657,10 @@ if (NOT CMAKE_SYSTEM_NAME STREQUAL "Emscripten")
           add_custom_command(TARGET onnxruntime_providers_qnn POST_BUILD
               COMMAND ${CMAKE_COMMAND} -E copy ${QNN_LIB_FILES} ${JAVA_NATIVE_TEST_DIR})
         endif()
+	if (WIN32)
+          set(EXAMPLE_PLUGIN_EP_DST_FILE_NAME $<IF:$<BOOL:${WIN32}>,$<TARGET_FILE_NAME:example_plugin_ep>,$<TARGET_LINKER_FILE_NAME:example_plugin_ep>>)
+          add_custom_command(TARGET custom_op_library POST_BUILD COMMAND ${CMAKE_COMMAND} -E copy_if_different $<TARGET_FILE:example_plugin_ep> ${JAVA_NATIVE_TEST_DIR}/${EXAMPLE_PLUGIN_EP_DST_FILE_NAME})
+	endif()
```
skottmckay, line 1663: nit: the `if` and `endif` lines have tabs

java/src/main/java/ai/onnxruntime/OrtSession.java
```diff
@@ -1273,6 +1272,37 @@ public void addCoreML(EnumSet<CoreMLFlags> flags) throws OrtException {
       addCoreML(OnnxRuntime.ortApiHandle, nativeHandle, OrtFlags.aggregateToInt(flags));
     }
 
+    /**
+     * Adds the specified execution provider and device tuples as an execution backend.
```
skottmckay, line 1280: Would it be worthwhile to mention that the order EPs are added in is the priority order for node assignment? First one is the highest priority.

This is the same as for the existing AddExecutionProvider so maybe not required as we've gotten this far without that info. :)

java/src/main/java/ai/onnxruntime/OrtModelCompilationOptions.java
skottmckay, line 1: @adrianlizarraga Could you please take a look at the model compilation side of things?

skottmckay:
> I haven't quite got the CMake configured so the Java tests for the ep registration only run when the ONNX Runtime shared provider support is built, but everything else works. I expect that to be a quick fix, but I'm not sure in what conditions it should be built and how we should handle it so I don't know where/when to plumb it through

The C++ unit tests only run in a Windows shared lib build, and also exclude minimal builds

https://github.com/microsoft/onnxruntime/blob/7a6cef6fe3672c5d59f1925e5f7c0a03dcfbc90e/cmake/onnxruntime_unittests.cmake#L1832-L1837

cmake/onnxruntime_unittests.cmake
Craigacp: Ooops, I'll fix that.

java/src/main/java/ai/onnxruntime/OrtSession.java
Craigacp: I think I have that note somewhere in the documentation, but it's good to add on those methods.

Craigacp:
> > I haven't quite got the CMake configured so the Java tests for the ep registration only run when the ONNX Runtime shared provider support is built, but everything else works. I expect that to be a quick fix, but I'm not sure in what conditions it should be built and how we should handle it so I don't know where/when to plumb it through
> 
> The C++ unit tests only run in a Windows shared lib build, and also exclude minimal builds
> 
> https://github.com/microsoft/onnxruntime/blob/7a6cef6fe3672c5d59f1925e5f7c0a03dcfbc90e/cmake/onnxruntime_unittests.cmake#L1832-L1837

Ok, I can lock them off. Should I use conditional compilation in the C binding to make the EP/device listing methods throw OrtException when used on unsupported platforms? It's a bit weird for them to return an empty list.

java/src/main/java/ai/onnxruntime/OrtModelCompilationOptions.java
adrianlizarraga: Yes, just took a look. It looks good. Thank you @Craigacp.

Like the C# bindings, the `ModelCompilationOptions_SetOutputModelBuffer` function is not included in these bindings, but perhaps the future/potential streaming API ([different PR](https://github.com/microsoft/onnxruntime/pull/24740)) would be a better fit. That can be done at a later time.

skottmckay:
> Should I use conditional compilation in the C binding to make the EP/device listing methods throw OrtException when used on unsupported platforms? It's a bit weird for them to return an empty list.

We'll be adding device discovery on linux and macos platforms very soon (should be in next release) so probably not necessary.

Craigacp:
> > Should I use conditional compilation in the C binding to make the EP/device listing methods throw OrtException when used on unsupported platforms? It's a bit weird for them to return an empty list.
> 
> We'll be adding device discovery on linux and macos platforms very soon (should be in next release) so probably not necessary.

Ok, can I get cc'd on those PRs so I can enable the tests on those platforms?

skottmckay:
> > > Should I use conditional compilation in the C binding to make the EP/device listing methods throw OrtException when used on unsupported platforms? It's a bit weird for them to return an empty list.
> > 
> > 
> > We'll be adding device discovery on linux and macos platforms very soon (should be in next release) so probably not necessary.
> 
> Ok, can I get cc'd on those PRs so I can enable the tests on those platforms?

@edgchen1 Who will be adding them

java/src/main/java/ai/onnxruntime/OrtSession.java
Craigacp: I've added a note to the top of this class explaining EP assignment, and updated this method's javadoc.

Craigacp:
I've improved the javadoc a little and made the `onnxruntime_providers_shared` library an unconditional part of the Java build. Previously we only copied that in if an EP required it, but now with the dynamic loading of EPs it needs to always be there.

Craigacp:
The EP tests are still only enabled on Windows, let me know if support has been added for other platforms, it's a quick fix.

Craigacp:
Enabling the shared library copy caused macOS and Android to fail. Not sure why it's not supported on macOS, but I can easily lock off the Android one. I had assumed if macOS was getting EP discovery support that the shared EP lib would be built automatically?

cmake/onnxruntime_java.cmake
```diff
@@ -157,9 +157,7 @@ if (WIN32)
   if(NOT onnxruntime_ENABLE_STATIC_ANALYSIS)
     add_custom_command(TARGET onnxruntime4j_jni POST_BUILD COMMAND ${CMAKE_COMMAND} -E copy_if_different $<TARGET_FILE:onnxruntime> ${JAVA_PACKAGE_LIB_DIR}/$<TARGET_FILE_NAME:onnxruntime>)
     add_custom_command(TARGET onnxruntime4j_jni POST_BUILD COMMAND ${CMAKE_COMMAND} -E copy_if_different $<TARGET_FILE:onnxruntime4j_jni> ${JAVA_PACKAGE_JNI_DIR}/$<TARGET_FILE_NAME:onnxruntime4j_jni>)
-    if (onnxruntime_USE_CUDA OR onnxruntime_USE_DNNL OR onnxruntime_USE_OPENVINO OR onnxruntime_USE_TENSORRT OR (onnxruntime_USE_QNN AND NOT onnxruntime_BUILD_QNN_EP_STATIC_LIB))
-      add_custom_command(TARGET onnxruntime4j_jni POST_BUILD COMMAND ${CMAKE_COMMAND} -E copy_if_different $<TARGET_FILE:onnxruntime_providers_shared> ${JAVA_PACKAGE_LIB_DIR}/$<TARGET_FILE_NAME:onnxruntime_providers_shared>)
-    endif()
+    add_custom_command(TARGET onnxruntime4j_jni POST_BUILD COMMAND ${CMAKE_COMMAND} -E copy_if_different $<TARGET_FILE:onnxruntime_providers_shared> ${JAVA_PACKAGE_LIB_DIR}/$<TARGET_FILE_NAME:onnxruntime_providers_shared>)
```
edgchen1: hm, for some reason, the `onnxruntime_providers_shared` target is not enabled sometimes.

https://github.com/microsoft/onnxruntime/blob/b0d4f005e6f94929f3d8a2a735768ae7d0977627/cmake/onnxruntime_providers_cpu.cmake#L210-L213

should we add a check like `if (TARGET onnxruntime_providers_shared)` around this?

Craigacp: Sure, I can put that in.

Craigacp: I've added a guard around the copy in `cmake/onnxruntime_java.cmake`.

Craigacp:
It doesn't look like the 2 failing checks are due to this PR, they seem to be happening well before anything I've changed is touched.

Craigacp:
Those failures were a javadoc error which I've fixed. Not sure how the other tests didn't catch it.

Craigacp:
@edgchen1 please can I get this merged in? I'd prefer to make a fresh small PR for the Java bits of #25878 rather than change this one again given we've got all the tests green now. The Java bits of #25878 will need this PR to land to add the OrtEpDevice Java implementation.

java/src/main/java/ai/onnxruntime/OrtEpDevice.java
```diff
@@ -0,0 +1,117 @@
...
+  long getNativeHandle() {
+    return nativeHandle;
+  }
+
+  /**
+   * Gets the EP name.
+   *
+   * @return The EP name.
+   */
+  public String getName() {
```
edgchen1, line 50: would it be more consistent with the C API (`EpDevice_EpName`) to call it `getEpName`? similarly comment for the other methods here.

e.g., both OrtEpDevice and OrtHardwareDevice have a `getVendor`. I wonder if calling it `getEpVendor` here would be clearer.

including the "Ep" in these method names would also be more consistent with the `toString` output.

edgchen1 (approved):

java/src/main/java/ai/onnxruntime/OrtEpDevice.java
edgchen1: can consider this for a later PR

Craigacp: Will do. They are already in an `OrtEpDevice` object so the Ep seemed a bit redundant, but I'm happy to change it for uniformity.

Issue #25038, opened 2025-06-12
[Build] libc++abi: terminating due to uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument


phial3:
### Describe the issue

env: macOS 15.5 M2 Pro
version: onnxruntime-osx-arm64-1.22.0
error: runtime error, this is debug with lldb
```text
libc++abi: terminating due to uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument
Process 67904 stopped
* thread #1, name = 'main', queue = 'com.apple.main-thread', stop reason = signal SIGABRT
    frame #0: 0x0000000199bd9388 libsystem_kernel.dylib`__pthread_kill + 8
libsystem_kernel.dylib`__pthread_kill:
->  0x199bd9388 <+8>:  b.lo   0x199bd93a8    ; <+40>
    0x199bd938c <+12>: pacibsp 
    0x199bd9390 <+16>: stp    x29, x30, [sp, #-0x10]!
    0x199bd9394 <+20>: mov    x29, sp
Target 0: (yolo) stopped.
(lldb) bt
* thread #1, name = 'main', queue = 'com.apple.main-thread', stop reason = signal SIGABRT
  * frame #0: 0x0000000199bd9388 libsystem_kernel.dylib`__pthread_kill + 8
    frame #1: 0x0000000199c1288c libsystem_pthread.dylib`pthread_kill + 296
    frame #2: 0x0000000199b1bc60 libsystem_c.dylib`abort + 124
    frame #3: 0x0000000199bc839c libc++abi.dylib`abort_message + 132
    frame #4: 0x0000000199bb6cf0 libc++abi.dylib`demangling_terminate_handler() + 316
    frame #5: 0x000000019983cde4 libobjc.A.dylib`_objc_terminate() + 172
    frame #6: 0x0000000199bc76b0 libc++abi.dylib`std::__terminate(void (*)()) + 16
    frame #7: 0x0000000199bc7654 libc++abi.dylib`std::terminate() + 108
    frame #8: 0x000000010f302f34 libonnxruntime.1.22.0.dylib`__clang_call_terminate + 12
    frame #9: 0x000000010e68b054 libonnxruntime.1.22.0.dylib`std::__1::unique_ptr<OrtEnv, std::__1::default_delete<OrtEnv>>::~unique_ptr[abi:ue170006]() + 84
[...]

PR #25134, opened 2025-06-23
Leak logger mutex

### LINKED ISSUE #24579: mutex issue at process exit on MacOS since v1.21.*
### Describe the issue

Hello,

I am trying to use the latest onnxruntime release (1.21.X) (tried both 0 and 1), they work fine for unix and windows but for mac I get the following issue at the end of the inference when everything seems to be closing down:

`libc++abi: terminating with uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument
`
No code changes from my side, the issue is only present on the latest onnxruntime release and only for Mac.
On previous releases all are working just fine, the only thing changed is the onnx runtime.

Googling says it has something todo with static mutex and the order in which they get destroyed/accessed - see more [here](https://stackoverflow.com/questions/66773247/libcabi-dylib-terminating-with-uncaught-exception-of-type-std-1system-er).

This happens both on the cpu case and on the gpu (coreML) case.

I have got in my hands only macOS 12 and macOS13 and both fail. Also both arm64 and x86 fail on all scenarios.

Important to note is that I can compile and link against onnxruntime just fine, its during running that the issue appears. I believe more specifically when the inference has finished and things are getting destructed/closing down

### To reproduce

For the current moment can't share a model to repro, and not sure If I will ever get approval for this.
Sharing to see if anyone has seen this or if it brings something in mind.

### Urgency

_No response_

### Platform

Mac

### OS Version

MacOs 12 and MacOs13

### ONNX Runtime Installation

Built from Source

1.21.0 and 1.21.1

### ONNX Runtime API

C++

### Architecture

X64

### Execution Provider

Default CPU

_No response_

decahedron1:
### Description
Allowing the destructor for `DefaultLoggerMutex` to run may cause a crash on macOS in certain usages of ONNX Runtime (most notably, with the [`ort`](https://ort.pyke.io/) Rust bindings). Since the mutex is a static local variable, it's fine to just leak the memory.

### Motivation and Context
Resolves #24579, #25038.

This change was applied to the ONNX Runtime binaries that ship with `ort` and no further issues have been reported.

### DIFF
--- a/onnxruntime/core/common/logging/logging.cc
+++ b/onnxruntime/core/common/logging/logging.cc
@@ -64,8 +64,8 @@ LoggingManager* LoggingManager::GetDefaultInstance() {
 #pragma warning(disable : 26426)
 #endif
 
-static std::mutex& DefaultLoggerMutex() noexcept {
-  static std::mutex mutex;
+static std::mutex* DefaultLoggerMutex() noexcept {
+  static std::mutex* mutex = new std::mutex();
   return mutex;
 }
 
@@ -107,7 +107,7 @@ LoggingManager::LoggingManager(std::unique_ptr<ISink> sink, Severity default_min
 
     // lock mutex to create instance, and enable logging
     // this matches the mutex usage in Shutdown
-    std::lock_guard<std::mutex> guard(DefaultLoggerMutex());
+    std::lock_guard<std::mutex> guard(*DefaultLoggerMutex());
 
     if (DefaultLoggerManagerInstance().load() != nullptr) {
       ORT_THROW("Only one instance of LoggingManager created with InstanceType::Default can exist at any point in time.");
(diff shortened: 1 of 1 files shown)


onnxruntime/core/common/logging/logging.cc
```diff
@@ -64,8 +64,8 @@ LoggingManager* LoggingManager::GetDefaultInstance() {
 #pragma warning(disable : 26426)
 #endif
 
-static std::mutex& DefaultLoggerMutex() noexcept {
-  static std::mutex mutex;
+static std::mutex* DefaultLoggerMutex() noexcept {
+  static std::mutex* mutex = new std::mutex();
```
snnn, line 68: Define it as a global var would be fine. 
Then there is no memory leak and no ordering issue.

snnn: Oh ..  I was wrong

snnn: I just asked Gemini. Unfortunately, the destructor of std::mutex on macOS platform is not trivial. This is really sad!!!!!

snnn: Could someone please help me confirm that? If we define std::mutex as a global var with default parameters:
```
std::mutext m;
```
Will its destructor call pthread_mutex_destroy ?

snnn: It seems that Android has the same problem.

snnn:
But, this PR will introduce memory leaks.  We have a memory leak checker in the windows_x64_debug pipeline, and it  failed because of that.  Do we have to make this change?

PR #25137, opened 2025-06-23
[EP ABI] Add ability to drop constant initializers for fused nodes specified in GetCapability

adrianlizarraga:
### Description
- Add ability to drop constant initializers for fused nodes specified in GetCapability.
- Rework how an EP specifies nodes that should be fused into one node within GetCapability.
  - Instead of passing the set of nodes as arguments to `GraphSupportInfo_AddNodesToFuse()`, the EP creates an `OrtNodeFusionOptions` object to specify the nodes and other relevant options. This makes it easier to extend the API in the future since we can't add more parameters to an existing function, but we can add more functions that modify an options object.

### Motivation and Context
Add more functionality missing from GetCapability() in the EP ABI.


include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -10,6 +10,7 @@ extern "C" {
 ORT_RUNTIME_CLASS(Ep);
 ORT_RUNTIME_CLASS(EpFactory);
 ORT_RUNTIME_CLASS(EpGraphSupportInfo);
+ORT_RUNTIME_CLASS(NodeFusionOptions);
```
adrianlizarraga: I'm not sure this is the best name. Suggestions are very much appreciated.

chilo-ms:
Another request that EP might want is to create the fused node's id hashed from not only graph_viewer but also other info, ex: TRT version. 
The case for TRT is to distinguish engines built from same OrtGraph with different version of TRT libraries. The workflow is TRT EP provides the id to ORT and gets the fused node's name and assign that name for the engine cache file, and next time when running the new inference, it can search for the cache with the name.

`OrtNodeFusionOptions` can address this issue by letting EP provide its own hashed id.

onnxruntime/core/session/abi_ep_types.h
```diff
@@ -17,30 +18,32 @@ struct EpGraph;
...
-    kInvalidGrouping = 0,
-    kSingleAssignedNode,
-    kFusedNode,
-  };
-
   // A grouping of supported nodes that should be handled in a single ComputeCapability.
   struct NodeGrouping {
-    NodeGrouping(NodeGroupingKind kind, std::vector<const onnxruntime::EpNode*>&& nodes)
-        : kind(kind), nodes(std::move(nodes)) {}
+    NodeGrouping(const onnxruntime::EpNode* single_node) : variant_(single_node) {}
```
adrianlizarraga:
```suggestion
    explicit NodeGrouping(const onnxruntime::EpNode* single_node) : variant_(single_node) {}
```

onnxruntime/core/session/abi_ep_types.h
```diff
@@ -17,30 +18,32 @@ struct EpGraph;
...
+    NodeGrouping(const OrtNodeFusionOptions& node_fusion_options) : variant_(node_fusion_options) {}
```
adrianlizarraga:
```suggestion
    explicit NodeGrouping(const OrtNodeFusionOptions& node_fusion_options) : variant_(node_fusion_options) {}
```

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -86,28 +87,64 @@ struct OrtEpApi {
...
    * a path between two of the provided nodes. Otherwise, the graph will become invalid.
+   *
+   * \param[in] nodes Array of nodes supported by the EP that should be fused.
+   * \param[in] num_nodes The number of nodes.
+   * \param[out] out Output parameter set to the OrtNodeFusionOptions instance that is created.
+   *
+   * \since Version 1.23.
+   */
+  ORT_API2_STATUS(CreateNodeFusionOptions, _In_reads_(num_nodes) const OrtNode* const* nodes, _In_ size_t num_nodes,
+                  _Outptr_ OrtNodeFusionOptions** out);
```
skottmckay: Do we need to put the nodes inside the options? I think the usage would be slightly more natural to keep them separate, which also allows the node fusion options argument to EpGraphSupportInfo_AddNodesToFuse to be optional. 

Do we need an opaque type for this or would a struct with a version number be simpler for this usage? Opaque types are more flexible, but I'm mindful about how large the API surface could get if we make everything an opaque type. In this case the usage seems to be one-way (EP author populates the options, passes it in to ORT, and they're done with it), so I'm not sure we would need to stuff anything extra into it in the background.

onnxruntime/test/autoep/library/example_plugin_ep.cc
```diff
@@ -194,6 +241,53 @@ struct ExampleEp : OrtEp, ApiPtrs {
     // Clean up the execution provider
   }
 
+  OrtStatus* SaveConstantInitializers(const OrtGraph* graph) {
+    OrtArrayOfConstObjects* initializers = nullptr;
+    DeferOrtRelease<OrtArrayOfConstObjects> release_initializers(&initializers, ort_api.ReleaseArrayOfConstObjects);
+    size_t num_initializers = 0;
+
+    RETURN_IF_ERROR(ort_api.Graph_GetInitializers(graph, &initializers));
```
skottmckay, line 192: If getting the constant initializers is going to be a common operation (I assume it will be given that determines nodes you can fuse), should we a) cache the initializers in the OrtGraph instance so we're not creating a new list on every call; and b) have a Graph_GetConstantInitialiizers?

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
adrianlizarraga: Updated to make the options struct not opaque. Nodes are also provided to AddNodesToFuse(). It is simpler. Thanks.

I think as we port TRT EP, we'll find out soon enough if it needs to change.

onnxruntime/test/autoep/library/example_plugin_ep.cc
adrianlizarraga: > have a Graph_GetConstantInitialiizers?

I don't know how much value that provides given that the user can get all initializer value infos and check if the value info is a constant initializer directly by calling `ValueInfo_IsConstantInitializer`. Do you think that warrants an extra API? I suppose it doesn't hurt.

skottmckay: technically the EP author could do this as a one-off. fetch all initializers, create list of constant initializers, to minimize the number of times you need to process the overall list. 

but if you can get that info via the value info now I think that's fine

skottmckay (approved):
:shipit:

PR #24887, opened 2025-05-28
Add GetCapability/Compile infrastructure for EP ABI

adrianlizarraga:
### Description
This PRs sets the foundation for the EP ABI, which allows plugin-EPs to interface with ORT using a binary stable interface. A plugin-EP can be built separately from ORT and is not tied to a specific commit of ORT.

Currently, this PR adds basic APIs necessary to allow an example plugin-EP to compile and run a simple model with a single `Mul` node.

- Example plugin-EP implementation: https://github.com/microsoft/onnxruntime/blob/adrianl/ep-abi/onnxruntime/test/autoep/library/example_plugin_ep.cc
- APIs: 
  - Graph IR: https://github.com/microsoft/onnxruntime/blob/adrianl/ep-abi/include/onnxruntime/core/session/onnxruntime_c_api.h#L5290-L5439
  - Plugin EP: https://github.com/microsoft/onnxruntime/blob/adrianl/ep-abi/include/onnxruntime/core/session/onnxruntime_c_api.h#L6177-L6481
- Example app code (from unit tests): https://github.com/microsoft/onnxruntime/blob/adrianl/ep-abi/onnxruntime/test/autoep/test_autoep_selection.cc#L614

### Motivation and Context
Based on https://github.com/microsoft/onnxruntime/pull/21450

PR #25143, opened 2025-06-23
Add Node_GetAttributes C API for EP ABI

chilo-ms:
This PRs adds additional Node_GetAttributes C API for EP ABI use.
It's based on https://github.com/microsoft/onnxruntime/pull/24887

### DIFF
--- a/include/onnxruntime/core/session/onnxruntime_c_api.h
+++ b/include/onnxruntime/core/session/onnxruntime_c_api.h
@@ -5878,14 +5878,14 @@ struct OrtApi {
   /** \brief Returns a node's attributes as OrtOpAttr instances.
    *
    * \param[in] node The OrtNode instance.
-   * \param[out] inputs Output parameter set to the OrtArrayOfConstObjects instance containing the node's attributes
+   * \param[out] attributes Output parameter set to the OrtArrayOfConstObjects instance containing the node's attributes
    *                    as OrtOpAttr instances. Must be released by calling ReleaseArrayOfConstObjects.
    *
    * \snippet{doc} snippets.dox OrtStatus Return Value
    *
    * \since Version 1.23.
    */
-  ORT_API2_STATUS(Node_GetAttributes, _In_ const OrtNode* node, _Outptr_ OrtArrayOfConstObjects** inputs);
+  ORT_API2_STATUS(Node_GetAttributes, _In_ const OrtNode* node, _Outptr_ OrtArrayOfConstObjects** attributes);
 
   /** \brief Get the subgraphs, as OrtGraph instances, contained by the given node.
    *
--- a/include/onnxruntime/core/session/onnxruntime_c_api.h
+++ b/include/onnxruntime/core/session/onnxruntime_c_api.h
@@ -5887,6 +5887,18 @@ struct OrtApi {
    */
   ORT_API2_STATUS(Node_GetAttributes, _In_ const OrtNode* node, _Outptr_ OrtArrayOfConstObjects** attributes);
 
+  /** \brief Gets the OrtNode's attribute as OrtOpAttr by name.
+   *
+   * \param[in] node The OrtNode instance.
+   * \param[in] attribute_name The name of the attribute
+   * \param[out] attribute Output the attribute if its name matches 'attribute_name', otherwise output nullptr.
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \since Version 1.23.
+   */
+  ORT_API2_STATUS(Node_GetAttributeByName, _In_ const OrtNode* node, _In_ const char* attribute_name, _Outptr_ const OrtOpAttr** attribute);
+
   /** \brief Get the attribute type as OrtOpAttrType from an OrtOpAttr.
    *
    * \param[in] attribute The OrtOpAttr instance.
--- a/include/onnxruntime/core/session/onnxruntime_c_api.h
+++ b/include/onnxruntime/core/session/onnxruntime_c_api.h
@@ -497,6 +497,7 @@ typedef OrtStatus*(ORT_API_CALL* EpSelectionDelegate)(_In_ const OrtEpDevice** e
 typedef enum OrtTypeTag {
   ORT_TYPE_TAG_Void,
   ORT_TYPE_TAG_OrtValueInfo,
+  ORT_TYPE_TAG_OrtOpAttr,
   ORT_TYPE_TAG_OrtNode,
   ORT_TYPE_TAG_OrtGraph,
 } OrtTypeTag;
@@ -5873,6 +5874,18 @@ struct OrtApi {
    * \since Version 1.23.
    */
   ORT_API2_STATUS(Node_GetImplicitInputs, _In_ const OrtNode* node, _Outptr_ OrtArrayOfConstObjects** implicit_inputs);
+
+  /** \brief Returns a node's attributes as OrtOpAttr instances.
+   *
+   * \param[in] node The OrtNode instance.
+   * \param[out] inputs Output parameter set to the OrtArrayOfConstObjects instance containing the node's attributes
+   *                    as OrtOpAttr instances. Must be released by calling ReleaseArrayOfConstObjects.
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \since Version 1.23.
+   */
+  ORT_API2_STATUS(Node_GetAttributes, _In_ const OrtNode* node, _Outptr_ OrtArrayOfConstObjects** inputs);
 
   /** \brief Get the subgraphs, as OrtGraph instances, contained by the given node.
    *
--- a/include/onnxruntime/core/session/onnxruntime_c_api.h
+++ b/include/onnxruntime/core/session/onnxruntime_c_api.h
@@ -270,6 +270,8 @@ typedef enum OrtOpAttrType {
   ORT_OP_ATTR_FLOATS,
   ORT_OP_ATTR_STRING,
   ORT_OP_ATTR_STRINGS,
+  ORT_OP_ATTR_GRAPH,
+  ORT_OP_ATTR_GRAPHS,
 } OrtOpAttrType;
 
 //! @}
@@ -5886,6 +5888,17 @@ struct OrtApi {
    * \since Version 1.23.
    */
   ORT_API2_STATUS(Node_GetAttributes, _In_ const OrtNode* node, _Outptr_ OrtArrayOfConstObjects** attributes);
+
+  /** \brief Get the attribute type as OrtOpAttrType from an OrtOpAttr.
+   *
+   * \param[in] attribute The OrtOpAttr instance.
+   * \param[out] type Output the attribute type as OrtOpAttrType.
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \since Version 1.23.
+   */
+  ORT_API2_STATUS(OpAttr_GetType, _In_ const OrtOpAttr* attribute, _Out_ OrtOpAttrType* type);
 
   /** \brief Get the subgraphs, as OrtGraph instances, contained by the given node.
    *
--- a/include/onnxruntime/core/session/onnxruntime_c_api.h
+++ b/include/onnxruntime/core/session/onnxruntime_c_api.h
@@ -270,8 +270,6 @@ typedef enum OrtOpAttrType {
   ORT_OP_ATTR_FLOATS,
   ORT_OP_ATTR_STRING,
   ORT_OP_ATTR_STRINGS,
-  ORT_OP_ATTR_GRAPH,
-  ORT_OP_ATTR_GRAPHS,
 } OrtOpAttrType;
 
 //! @}
--- a/onnxruntime/core/graph/ep_api_types.cc
+++ b/onnxruntime/core/graph/ep_api_types.cc
@@ -101,7 +101,7 @@ Status EpNode::Create(const Node& node, const EpGraph* ep_graph,
     for (const auto& item : node_attrs) {
       auto attr = std::make_unique<ONNX_NAMESPACE::AttributeProto>(item.second);  // Copy AttributeProto and owned by this EpNode object.
       ep_node_attributes.push_back(reinterpret_cast<OrtOpAttr*>(attr.get()));
-      ep_node_attributes_map[item.first] = std::move(attr);
+      ep_node_attributes_map.emplace(item.first, std::move(attr));
     }
   }
 
--- a/onnxruntime/core/graph/ep_api_types.cc
+++ b/onnxruntime/core/graph/ep_api_types.cc
@@ -224,6 +224,15 @@ gsl::span<const EpValueInfo* const> EpNode::GetOutputsSpan() const {
   return outputs_;
 }
 
+const OrtOpAttr* EpNode::GetAttribute(const std::string& name) const {
+  auto iter = attributes_map_.find(name);
+  if (iter == attributes_map_.end()) {
+    return nullptr;
+  } else {
+    return reinterpret_cast<const OrtOpAttr*>(iter->second.get());
+  }
+}
+
 //
 // EpValueInfo
 //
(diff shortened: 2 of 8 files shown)


include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -5874,6 +5875,18 @@ struct OrtApi {
    */
   ORT_API2_STATUS(Node_GetImplicitInputs, _In_ const OrtNode* node, _Outptr_ OrtArrayOfConstObjects** implicit_inputs);
 
+  /** \brief Returns a node's attributes as OrtOpAttr instances.
+   *
+   * \param[in] node The OrtNode instance.
+   * \param[out] inputs Output parameter set to the OrtArrayOfConstObjects instance containing the node's attributes
```
adrianlizarraga: `inputs` -> `attributes`

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -5874,6 +5875,18 @@ struct OrtApi {
...
+   *                    as OrtOpAttr instances. Must be released by calling ReleaseArrayOfConstObjects.
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \since Version 1.23.
+   */
+  ORT_API2_STATUS(Node_GetAttributes, _In_ const OrtNode* node, _Outptr_ OrtArrayOfConstObjects** inputs);
```
adrianlizarraga: Same for this one: `inputs` -> `attributes`

onnxruntime/core/session/onnxruntime_c_api.cc
```diff
@@ -2842,6 +2842,20 @@ ORT_API_STATUS_IMPL(OrtApis::Node_GetImplicitInputs, _In_ const OrtNode* node,
   API_IMPL_END
 }
 
+ORT_API_STATUS_IMPL(OrtApis::Node_GetAttributes, _In_ const OrtNode* node, _Outptr_ OrtArrayOfConstObjects** attrs) {
+  API_IMPL_BEGIN
+  if (attrs == nullptr) {
+    return OrtApis::CreateStatus(ORT_INVALID_ARGUMENT, "'inputs' argument is NULL");
```
adrianlizarraga: `inputs` -> `attributes`

onnxruntime/core/graph/ep_api_types.cc
```diff
@@ -91,6 +91,22 @@ Status EpNode::Create(const Node& node, const EpGraph* ep_graph,
   ConvertNodeArgsToValueInfos(ep_graph, value_infos_map, node_inputs, ep_node_inputs);
   ConvertNodeArgsToValueInfos(ep_graph, value_infos_map, node_outputs, ep_node_outputs);
 
+  const auto& node_attrs = node.GetAttributes();
+  std::unordered_map<std::string, std::unique_ptr<ONNX_NAMESPACE::AttributeProto>> ep_node_attributes_map;
+  std::vector<OrtOpAttr*> ep_node_attributes;
+
+  if (node_attrs.size() > 0) {
+    ep_node_attributes.resize(node_attrs.size(), nullptr);
```
adrianlizarraga: nit: could use `ep_node_attributes.reserve(node_attrs.size())` combined with `ep_node_attributes.push_back()`. This removes need for `attr_idx` and gets rid of an unnecessary initialization to `nullptr`.

onnxruntime/test/ep_graph/test_ep_graph.cc
```diff
@@ -449,6 +449,24 @@ static void CheckGraphCApi(const GraphViewer& graph_viewer, const OrtGraph& api_
...
+      DeferOrtRelease<OrtArrayOfConstObjects> release_node_attributes(&api_node_attributes,
+                                                                      ort_api.ReleaseArrayOfConstObjects);
+      ASSERT_ORTSTATUS_OK(ort_api.Node_GetAttributes(api_node, &api_node_attributes));
+      CheckArrayObjectType(api_node_attributes, ORT_TYPE_TAG_OrtOpAttr);
+
+      for (size_t attr_idx = 0; attr_idx < node_attrs.size(); attr_idx++) {
+        const OrtOpAttr* api_node_attr = nullptr;
+        ASSERT_ORTSTATUS_OK(ort_api.ArrayOfConstObjects_GetElementAt(api_node_attributes, attr_idx,
+                                                                     reinterpret_cast<const void**>(&api_node_attr)));
+        ASSERT_NE(api_node_attr, nullptr);
```
adrianlizarraga, line 467: I was about to suggest checking that the attribute retrieved from the API has the same type as the attribute from the internal graph (no need to check contents maybe). However, I don't think we have a C API to get the `OrtOpAttrType` from an `OrtOpAttr`. I think that API would be necessary for an EP author, right? Do you think that should be added in this PR?

chilo-ms: That's a good point and user might need to know the attribute type.
Just wondering if we really need to add a C API to get the type of OrtOpAttr? Since OrtOpAttr is interchangeable with ONNX::AttributeProto, and user can cast it to AttributeProto and call type() to get the ONNX::AttributeProto_AttributeType
Or we shouldn't assume users will add ONNX as a dependency?

adrianlizarraga: I think for normal users of the API we shouldn't assume or require them to cast to onnx::AttributeProto.

chilo-ms: OpAttr_GetType C API is added

onnxruntime/core/session/onnxruntime_c_api.cc
```diff
@@ -2856,6 +2856,54 @@ ORT_API_STATUS_IMPL(OrtApis::Node_GetAttributes, _In_ const OrtNode* node, _Outp
...
+    case ONNX_NAMESPACE::AttributeProto_AttributeType::AttributeProto_AttributeType_GRAPH: {
+      *type = OrtOpAttrType::ORT_OP_ATTR_GRAPH;
+      break;
+    }
+    case ONNX_NAMESPACE::AttributeProto_AttributeType::AttributeProto_AttributeType_GRAPHS: {
+      *type = OrtOpAttrType::ORT_OP_ATTR_GRAPHS;
+      break;
+    }
+    default:
+      return OrtApis::CreateStatus(OrtErrorCode::ORT_INVALID_ARGUMENT, "Unexpected attribute type.");
```
chilo-ms, line 2916: OrtOpAttrType is not 1:1 to ONNX::AttributeProto_AttributeType
How do we better return if type is in AttributeProto_AttributeType but not in OrtOpAttrType?
How about returning OrtOpAttrType::ORT_OP_ATTR_UNDEFINED in that case?

adrianlizarraga: Hmm good question. I think it is fine to return an error status if there's an onnx attribute type that ORT does not yet support. My intuition is that we wouldn't want to overload `UNDEFINED` with multiple possible meanings: it is actually an onnx undefined type or it is a type ORT doesn't support.

But maybe others have a different opinion.

adrianlizarraga (approved):

onnxruntime/test/ep_graph/test_ep_graph.cc
```diff
@@ -449,6 +449,69 @@ static void CheckGraphCApi(const GraphViewer& graph_viewer, const OrtGraph& api_
...
+
+        OrtOpAttrType api_node_attr_type = OrtOpAttrType::ORT_OP_ATTR_UNDEFINED;
+        ASSERT_ORTSTATUS_OK(ort_api.OpAttr_GetType(api_node_attr, &api_node_attr_type));
+
+        ONNX_NAMESPACE::AttributeProto_AttributeType node_attr_type = node_attr.second.type();
+        switch (node_attr_type) {
```
adrianlizarraga, line 484: Looks like this needs a `default` case to make CI happy (handle all possible types).

chilo-ms: Some tests failed in CIs due to i didn't handle correctly in the case where attribute type is not defined in OrtOpAttrType.
Add default to switch to make compiler happy.

adrianlizarraga (dismissed):

onnxruntime/core/graph/ep_api_types.cc
```diff
@@ -91,6 +91,20 @@ Status EpNode::Create(const Node& node, const EpGraph* ep_graph,
...
+  std::unordered_map<std::string, std::unique_ptr<ONNX_NAMESPACE::AttributeProto>> ep_node_attributes_map;
+  std::vector<OrtOpAttr*> ep_node_attributes;
+
+  if (node_attrs.size() > 0) {
+    ep_node_attributes.reserve(node_attrs.size());
+
+    for (const auto& item : node_attrs) {
+      auto attr = std::make_unique<ONNX_NAMESPACE::AttributeProto>(item.second);  // Copy AttributeProto and owned by this EpNode object.
+      ep_node_attributes.push_back(reinterpret_cast<OrtOpAttr*>(attr.get()));
+      ep_node_attributes_map[item.first] = std::move(attr);
```
edgchen1: nit: `operator[]` will first construct an element if one doesn't already exist and then return a reference to it. the extra initialization can be avoided with `emplace()`
```suggestion
      ep_node_attributes_map.emplace(item.first, std::move(attr));
```

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -270,6 +270,8 @@ typedef enum OrtOpAttrType {
   ORT_OP_ATTR_FLOATS,
   ORT_OP_ATTR_STRING,
   ORT_OP_ATTR_STRINGS,
+  ORT_OP_ATTR_GRAPH,
+  ORT_OP_ATTR_GRAPHS,
```
edgchen1: how will users read graph attributes? I see there's an existing `ReadOpAttr` API, but that fills out a byte buffer. the interpretation of the byte buffer seems more straightforward for primitive types and null-terminated strings, but what about graphs?

onnxruntime/core/session/onnxruntime_c_api.cc
edgchen1: I also think an error would be good in that case.

onnxruntime/test/ep_graph/test_ep_graph.cc
```diff
@@ -449,6 +449,78 @@ static void CheckGraphCApi(const GraphViewer& graph_viewer, const OrtGraph& api_
...
+          case ONNX_NAMESPACE::AttributeProto_AttributeType::AttributeProto_AttributeType_GRAPH: {
+            ASSERT_EQ(api_node_attr_type, OrtOpAttrType::ORT_OP_ATTR_GRAPH);
+            break;
+          }
+          case ONNX_NAMESPACE::AttributeProto_AttributeType::AttributeProto_AttributeType_GRAPHS: {
+            ASSERT_EQ(api_node_attr_type, OrtOpAttrType::ORT_OP_ATTR_GRAPHS);
+            break;
+          }
+          default:
+            break;
```
edgchen1: should the test fail if the default case is hit?

include/onnxruntime/core/session/onnxruntime_c_api.h
chilo-ms: For graphs, the existing `ReadOpAttr` would have to be able to read the `GraphProto` returning from `AttributeProto::g()`

````
inline const ::onnx::GraphProto& AttributeProto::g() const {
  // @@protoc_insertion_point(field_get:onnx.AttributeProto.g)
  return _internal_g();
}
````
I think since GraphProto is not a primitive type and we can't return it in our C API, so we didn't add graph/graphs in 
OrtOpAttrType in the first place? 
I will remove them in OrtOpAttrType to be aligned with `ReadOpAttr`

File: onnxruntime/test/ep_graph/test_ep_graph.cc (hunk 10 above)
Comment (chilo-ms) [reply onnxruntime/test/ep_graph/test_ep_graph.cc]: The models, e.g. bart_tiny.onnx, that the tests use contain node attribute types that are in OrtOpAttrType. 
We don't check and make it fail for the unsupported OrtOpAttrType.

File: include/onnxruntime/core/session/onnxruntime_c_api.h [hunk 11]
```diff
@@ -5874,6 +5875,29 @@ struct OrtApi {
...
+  /** \brief Get the attribute type as OrtOpAttrType from an OrtOpAttr.
+   *
+   * \param[in] attribute The OrtOpAttr instance.
+   * \param[out] type Output the attribute type as OrtOpAttrType.
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \since Version 1.23.
+   */
+  ORT_API2_STATUS(OpAttr_GetType, _In_ const OrtOpAttr* attribute, _Out_ OrtOpAttrType* type);
```
edgchen1, line 5911: should we also provide a way to get the attribute name?

or have a way to get an attribute by name? e.g., like this API:
`ORT_API2_STATUS(ShapeInferContext_GetAttribute, _In_ const OrtShapeInferContext* context, _In_ const char* attr_name, _Outptr_ const OrtOpAttr** attr);`

onnxruntime/test/ep_graph/test_ep_graph.cc
edgchen1: we should already `continue` above if `ort_api.OpAttr_GetType(api_node_attr, &api_node_attr_type)` doesn't support the type, right? so wouldn't it be unexpected to hit this default case?

include/onnxruntime/core/session/onnxruntime_c_api.h
edgchen1: stepping back a bit, it would be good to understand what an EP author needs to be able to do with the attributes.

my assumption was that they would want to read a particular attribute's value. and for that we'd probably need to be able to identify a particular attribute by name.

onnxruntime/test/ep_graph/test_ep_graph.cc
chilo-ms: oh okay i got your meaning, let me make test fail for the default case.

include/onnxruntime/core/session/onnxruntime_c_api.h
chilo-ms: Node_GetAttribute API added

onnxruntime/core/graph/ep_api_types.h
```diff
@@ -196,6 +199,9 @@ struct EpNode : public OrtNode {
   InlinedVector<EpValueInfo*> inputs_;
   InlinedVector<EpValueInfo*> outputs_;
 
+  std::unordered_map<std::string, std::unique_ptr<ONNX_NAMESPACE::AttributeProto>> attributes_map_;
```
adrianlizarraga, line 205: nit: is this `attribute_map_` used anywhere? Perhaps it could be used in the API that gets an attribute by name instead of searching through the attributes array. (though I admit that searching through a small array is probably faster than using a map).

onnxruntime/core/session/onnxruntime_c_api.cc
```diff
@@ -2862,17 +2863,8 @@ ORT_API_STATUS_IMPL(OrtApis::Node_GetAttributeByName, _In_ const OrtNode* node,
...
-
-  *attribute = nullptr;
-  for (auto attr : array->storage) {
-    auto ort_op_attr = reinterpret_cast<const OrtOpAttr*>(attr);
-    if (ort_op_attr->attr_proto.has_name() && (ort_op_attr->attr_proto.name() == attribute_name)) {
-      *attribute = ort_op_attr;
-      break;
-    }
-  }
+  const EpNode* ep_node = EpNode::ToInternal(node);
```
adrianlizarraga, line 2866: One other thing here. `EpNode::ToInternal` returns `nullptr` if this is not an `EpNode` (it could be a `ModelEditorNode`). In that case, we should return a status error.

chilo-ms: Ah, right, just addressed this.

adrianlizarraga (approved):

PR #25147, opened 2025-06-24
Add `ShouldConvertDataLayoutForOp()` API to allow EPs to customize layout sensitive ops

edgchen1:
### Description

Add `IExecutionProvider::ShouldConvertDataLayoutForOp()` to allow EPs to customize layout sensitive ops. Move existing hardcoded EP-specific logic out of layout transformer code.

Add `OrtEp::ShouldConvertDataLayoutForOp` to ABI EP API to allow similar customization by plugin EPs.

### Motivation and Context

Enable layout sensitive op customization through internal EP interface and the ABI EP API.

### DIFF
--- a/include/onnxruntime/core/session/onnxruntime_ep_c_api.h
+++ b/include/onnxruntime/core/session/onnxruntime_ep_c_api.h
@@ -257,6 +257,31 @@ struct OrtEp {
   OrtStatus*(ORT_API_CALL* GetPreferredDataLayout)(_In_ OrtEp* this_ptr,
                                                    _Out_ OrtEpDataLayout* preferred_data_layout);
 
+  /** \brief Determine whether a node with `domain` and `op_type` requires its data layout to be converted to NHWC.
+   *         If the EP prefers NHWC data layout (see `GetPreferredDataLayout()`), this function will be called during
+   *         layout transformation.
+   *
+   * \note Implementation of this function is optional.
+   *       If an EP prefers NHWC data layout, it may implement this to customize the specific NHWC op preferences at a
+   *       finer granularity.
+   *
+   * \param[in] this_ptr The OrtEp instance.
+   * \param[in] node_domain The node's op domain. An empty string means the ONNX domain.
+   * \param[in] node_op_type The node's op type.
+   * \param[out] should_convert Indicates whether the node's layout should be converted to NHWC.
+   *                            If greater than 0, convert.
+   *                            If 0, don't convert.
+   *                            Otherwise, if less than 0, leave the decision to ORT.
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \since Version 1.23.
+   */
+  OrtStatus*(ORT_API_CALL* ShouldConvertNodeLayoutToNhwc)(_In_ OrtEp* this_ptr,
+                                                          _In_z_ const char* node_domain,
+                                                          _In_z_ const char* node_op_type,
+                                                          _Outptr_ int* should_convert);
+
   /** \brief Set dynamic options on this EP.
    *
    * Dynamic options can be set by the user at any time after session creation with `OrtApi::SetEpDynamicOptions()`.
--- a/include/onnxruntime/core/session/onnxruntime_ep_c_api.h
+++ b/include/onnxruntime/core/session/onnxruntime_ep_c_api.h
@@ -136,12 +136,18 @@ struct OrtEpApi {
 };
 
 /**
- * \brief The data layout type that is preferred by an EP.
+ * \brief The data layout type.
+ *
+ * EPs may specify a preferred data layout type. ORT's default layout type is OrtEpDataLayout_NCHW, or
+ * OrtEpDataLayout_Default.
+ *
  * \since Version 1.23.
  */
 typedef enum OrtEpDataLayout {
   OrtEpDataLayout_NCHW = 0,
   OrtEpDataLayout_NHWC,
+
+  OrtEpDataLayout_Default = OrtEpDataLayout_NCHW,
 } OrtEpDataLayout;
 
 /**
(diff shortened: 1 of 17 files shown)


fs-eire:
What is the difference between the newly added interface and the existing `GetPreferredLayout()`?

edgchen1:
> What is the difference between the newly added interface and the existing `GetPreferredLayout()`?

GetPreferredLayout returns a single preferred layout value for the whole EP. some EPs have further customization of what ops should be considered layout sensitive. this is hardcoded here currently:
https://github.com/microsoft/onnxruntime/blob/3a47bd2821370df1315c0eead9189bda4e0ea186/onnxruntime/core/optimizer/layout_transformation/layout_transformation.cc#L72-L117

this new interface provides a way for EPs to specify this originally hardcoded logic.

onnxruntime/core/session/ep_plugin_provider_interfaces.cc
```diff
@@ -412,6 +412,26 @@ DataLayout PluginExecutionProvider::GetPreferredLayout() const {
...
+std::optional<bool> PluginExecutionProvider::ShouldConvertNodeLayoutToNhwc(std::string_view node_domain,
+                                                                           std::string_view node_op_type) const {
+  if (ort_ep_->ShouldConvertNodeLayoutToNhwc == nullptr) {
+    return Base::ShouldConvertNodeLayoutToNhwc(node_domain, node_op_type);
+  }
+
+  int should_convert = -1;
+
+  ORT_THROW_IF_ERROR(ToStatusAndRelease(
+    ort_ep_->ShouldConvertNodeLayoutToNhwc(ort_ep_.get(), node_domain.data(), node_op_type.data(), &should_convert)));
```
edgchen1: ensure c strings are null terminated? could convert to std::string to be extra sure

onnxruntime/core/providers/cuda/cuda_nhwc_kernels.cc
```diff
@@ -24,7 +24,7 @@
 
 namespace onnxruntime::cuda {
 
-// When adding new supported NHWC operations make sure to also integrate them into: ConvertNodeLayout
+// When adding new supported NHWC operations make sure to also integrate them into: ShouldConvertNodeLayoutToNhwc
 // in onnxruntime/core/optimizer/layout_transformation/layout_transformation.cc
```
edgchen1:
```suggestion
// When adding new supported NHWC operations make sure to also integrate them into
// CUDAExecutionProvider::ShouldConvertNodeLayoutToNhwc()
```

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -257,6 +257,31 @@ struct OrtEp {
...
+   * \param[out] should_convert Indicates whether the node's layout should be converted to NHWC.
+   *                            If greater than 0, convert.
+   *                            If 0, don't convert.
+   *                            Otherwise, if less than 0, leave the decision to ORT.
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \since Version 1.23.
+   */
+  OrtStatus*(ORT_API_CALL* ShouldConvertNodeLayoutToNhwc)(_In_ OrtEp* this_ptr,
```
adrianlizarraga: We have the `GetPreferredDataLayout` interface which could return any layout (including any we may add later). This function, however, is specific to the NHWC layout only. To future proof, should we rename this to something like `ShouldConvertNodeLayout` and add the layout as a parameter? This gives us flexibility to add new layouts in the future and reuse the same interface functions.

```c++
OrtStatus*(ORT_API_CALL* ShouldConvertNodeLayout)(_In_ OrtEp* this_ptr,
                                                  _In_ OrtEpDataLayout data_layout, /*other params*/);
```

skottmckay: Good idea.

edgchen1: updated the name and added a `target_data_layout` parameter.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -257,6 +263,34 @@ struct OrtEp {
...
 
+  /** \brief Given an op with domain `domain` and type `op_type`, determine whether an associated node's data layout
+   *         should be converted to `target_data_layout`.
+   *         If the EP prefers a non-default data layout (see `GetPreferredDataLayout()`), this function will be called
+   *         during layout transformation with `target_data_layout` set to the EP's preferred data layout.
+   *
+   * \note Implementation of this function is optional.
+   *       If an EP prefers a non-default data layout, it may implement this to customize the specific op data layout
+   *       preferences at a finer granularity.
+   *
```
adrianlizarraga, line 274: I'm not sure if we have official documentation regarding the ORT layout transformer, but maybe we could link to these dev notes that explain the relationship between `GetPreferredDataLayout`, `GetCapability`, and "layout-sensitive ops": https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/optimizer/layout_transformation/layout_transformation_dev_notes.md

adrianlizarraga (approved):

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
skottmckay: AFAIK there's no existing documentation as it was always an internal feature (hence not part of the previous EP API).

skottmckay (approved):
:shipit:

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
edgchen1: I think it would be good to document this feature for EP implementors. Let me see if I can come up with something in another PR.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1066,6 +1066,16 @@ DataLayout QNNExecutionProvider::GetPreferredLayout() const {
   return DataLayout::NHWC;
 }
 
+std::optional<bool> QNNExecutionProvider::ShouldConvertNodeLayoutToNhwc(std::string_view node_domain,
+                                                                        std::string_view node_op_type) const {
+  if (node_domain == kOnnxDomain && node_op_type == "Upsample") {
+    // Upsample is translated to QNN's Resize, which requires the NHWC layout for processing.
+    return true;
+  }
```
HectorSVC, line 1079: The Op list for QNN EP is not correct. should include Conv, ConvTranspose, Pooling ops, etc. Please double check.

HectorSVC: Please check the code path for
[https://github.com/microsoft/onnxruntime/blob/a17e3462ffb44df06299bcbc9af7de23f5eba367/onnxruntime/core/optimizer/transpose_optimization/onnx_transpose_optimization.cc#L3304](https://github.com/microsoft/onnxruntime/blob/a17e3462ffb44df06299bcbc9af7de23f5eba367/onnxruntime/core/optimizer/transpose_optimization/onnx_transpose_optimization.cc#L3304)

PR #25156, opened 2025-06-24
Attention Operator (CPU)

xadupre:
### Description
Implementation Attention(23) for CPU.

The backend tests from onnx were wrong for Attention (see https://github.com/onnx/onnx/pull/7142). The onnx version needs to be updated to make all tests pass. The implementation matches the reference implementation after onnx was fixed.


onnxruntime/core/providers/cpu/llm/attention.cc
```diff
@@ -0,0 +1,749 @@
...
+  int k_dims = onnxruntime::narrow<int>(K->Shape().NumDimensions());
+  int v_dims = onnxruntime::narrow<int>(V->Shape().NumDimensions());
+  ORT_ENFORCE(q_dims == 3 || q_dims == 4, "Q must be a 3D or 4D tensor");
+  ORT_ENFORCE(q_dims == k_dims, "Q and K must have the same rank.");
+  ORT_ENFORCE(q_dims == v_dims, "Q and V must have the same rank.");
+
+  ORT_ENFORCE(Q->Shape()[0] == K->Shape()[0], "inconsistent batch_size");
+  ORT_ENFORCE(Q->Shape()[0] == V->Shape()[0], "inconsistent batch_size");
+  ORT_ENFORCE(past_key == nullptr || Q->Shape()[0] == past_key->Shape()[0], "inconsistent batch_size");
+  ORT_ENFORCE(past_value == nullptr || Q->Shape()[0] == past_value->Shape()[0], "inconsistent batch_size");
```
gramalingam: Should we check that `(past_key == nullptr) == (past_value == nullptr)` ?

onnxruntime/core/providers/cpu/llm/attention.cc
```diff
@@ -0,0 +1,749 @@
...
+    // From shapes
+    parameters.transpose_output = false;                                      // whether to transpose the output from BxNxSxH to BxSxNxH
+    parameters.q_sequence_length = onnxruntime::narrow<int>(Q->Shape()[2]);   // Q.shape[2] (4D)
+    parameters.head_size = onnxruntime::narrow<int>(Q->Shape()[3]);           // Q.shape[3] (4D)
+    parameters.kv_sequence_length = onnxruntime::narrow<int>(K->Shape()[2]);  // K.shape[2] or V.shape[2] (4D)
+    parameters.v_head_size = onnxruntime::narrow<int>(V->Shape()[3]);         // V.shape[3] (4D)
+    parameters.past_sequence_length = past_key == nullptr                     // past_key.shape[2] or past_value.shape[2] (4D) or given by the mask
+                                          ? (attn_mask == nullptr
+                                                 ? 0
+                                                 : onnxruntime::narrow<int>(attn_mask->Shape()[attn_mask->Shape().NumDimensions() - 1]) - parameters.kv_sequence_length)
```
gramalingam: Does this make sense? How can we have a non-zero past_sequence_length without a past_key being present?

onnxruntime/core/providers/cpu/llm/attention.cc
```diff
@@ -0,0 +1,749 @@
...
+    ORT_ENFORCE(past_key == nullptr || K->Shape()[3] == past_key->Shape()[3], "inconsistent head_size");
+    ORT_ENFORCE(past_value == nullptr || V->Shape()[3] == past_value->Shape()[3], "inconsistent head_size");
+
+    // From shapes
+    parameters.transpose_output = false;                                      // whether to transpose the output from BxNxSxH to BxSxNxH
+    parameters.q_sequence_length = onnxruntime::narrow<int>(Q->Shape()[2]);   // Q.shape[2] (4D)
+    parameters.head_size = onnxruntime::narrow<int>(Q->Shape()[3]);           // Q.shape[3] (4D)
+    parameters.kv_sequence_length = onnxruntime::narrow<int>(K->Shape()[2]);  // K.shape[2] or V.shape[2] (4D)
+    parameters.v_head_size = onnxruntime::narrow<int>(V->Shape()[3]);         // V.shape[3] (4D)
+    parameters.past_sequence_length = past_key == nullptr                     // past_key.shape[2] or past_value.shape[2] (4D) or given by the mask
```
gramalingam: When this is set from past_key, should we check that past_value.shape[2] == past_key.shape[2]?

onnxruntime/core/providers/cpu/llm/attention.cc
```diff
@@ -0,0 +1,749 @@
...
+                                              static_cast<int64_t>(parameters.v_head_size)};
+  std::vector<int64_t> output_qk_shape = {static_cast<int64_t>(parameters.batch_size),
+                                          static_cast<int64_t>(parameters.q_num_heads),
+                                          static_cast<int64_t>(parameters.q_sequence_length),
+                                          static_cast<int64_t>(parameters.past_sequence_length + parameters.kv_sequence_length)};
+
+  Tensor* Y = context->Output(0, y_shape);
+  Tensor* present_key = context->Output(1, present_key_shape);
+  Tensor* present_value = context->Output(2, present_value_shape);
+  Tensor* output_qk = context->Output(3, output_qk_shape);
```
gramalingam: Should we have any sanity checks for nullptrs here, or does `ApplyAttention` do those checks?

onnxruntime/core/providers/cpu/llm/attention.cc
```diff
@@ -0,0 +1,749 @@
...
+                              present_value,  // present V output tensor (if separating present KV)
+                              output_qk,      // Q*K output tensor (if returning Q*K value)
+                              parameters,     // attention parameters
+                              nullptr,        // const Tensor* attn_bias,  // additive bias applied on scaled QK.
+                              false           // bool past_present_share_buffer
+  );
+}
+
+template <typename T>
+void AttentionBase<T>::ComputeAttentionProbs(T* attention_probs,                       // output buffer with size BxNxSxT
```
gramalingam: Just wondering: so, we can't reuse any shared/common parts of the implementation with ORT contrib op implementation?

onnxruntime/core/providers/cpu/llm/attention.cc
gramalingam: It's been a while since I looked at the `context->Output` API: I guess it will return nullptr if the optional output is not specified in ONNX model? Is it safe to invoke when the output is not specified? I assume so, but just double checking

onnxruntime/core/providers/cpu/llm/attention_helper.h
```diff
@@ -0,0 +1,66 @@
...
+namespace onnxruntime {
+namespace attention_helper {
+
+enum AttentionMaskType {
+  MASK_NONE,  // No mask
+  MASK_BOOL,  // Boolean mask with shape [batch_size, sequence_length, total_sequence_length] or [batch_size, q_num_heads, q_sequence_length, total_sequence_length
+  MASK_ADD,   // Additive mask with shape [batch_size, sequence_length, total_sequence_length] or [batch_size, q_num_heads, q_sequence_length, total_sequence_length
+};
+
+enum AttentionType {
```
gramalingam: Is this used anywhere? I didn't see it in the attention implementation, maybe I missed it.

onnxruntime/core/providers/cpu/llm/attention.cc
```diff
@@ -0,0 +1,749 @@
...
+      Attention,                                                      \
+      23,                                                             \
+      T,                                                              \
+      KernelDefBuilder()                                              \
+          .TypeConstraint("T1", DataTypeImpl::GetTensorType<T>())     \
+          .TypeConstraint("T2", DataTypeImpl::GetTensorType<T>())     \
+          .TypeConstraint("U", BuildKernelDefConstraints<bool, T>()), \
+      Attention<T>);
+
+REGISTER_ONNX_KERNEL_TYPED(float)
```
titaiwangms, line 31: Is there a way we can support fp16? https://github.com/microsoft/onnxruntime/blob/b6ce761b420c7ad2c3bbc702bd9a19298cb0ca8b/onnxruntime/contrib_ops/cpu/cpu_contrib_kernels.cc#L25

onnxruntime/core/providers/cpu/llm/attention.cc
```diff
@@ -0,0 +1,749 @@
...
+  parameters.softmax_precision = softmax_precision_;                          // precision for softmax, 0 for float16, 1 for float32
+  parameters.qk_matmul_output_mode = qk_matmul_output_mode_;                  // output mode for Q*K matmul
+  parameters.mask_type = attn_mask == nullptr ? AttentionMaskType::MASK_NONE  // mask type
+                                              : (attn_mask->IsDataType<bool>() ? AttentionMaskType::MASK_BOOL : AttentionMaskType::MASK_ADD);
+  parameters.batch_size = onnxruntime::narrow<int>(Q->Shape()[0]);  // Q.shape[0], K.shape[0], V.shape[0] (4D)
+
+  ORT_ENFORCE(parameters.batch_size > 0, "Batch size must be greater than 0");
+  ORT_ENFORCE(parameters.mask_type == AttentionMaskType::MASK_NONE ||
+                  parameters.mask_type == AttentionMaskType::MASK_BOOL ||
+                  attn_mask->IsDataType<float>(),
```
tianleiwu: nit: `IsDataType<T>`

onnxruntime/core/providers/cpu/llm/attention.cc
```diff
@@ -0,0 +1,749 @@
...
+      });
+}
+
+template <typename T>
+Status AttentionBase<T>::ApplyAttention(OpKernelContext* context,
+                                        const T* Q,                             // Q data with shape BxNxSxH
+                                        const T* K,                             // K data with shape BxNxLxH
+                                        const T* V,                             // V value with size BxNxLxH_v
+                                        const Tensor* mask_index,               // mask index. nullptr if no mask or its size is B
+                                        const Tensor* past,                     // past state
```
gramalingam: What is this input "past"? It seems nullptr in the only call in this implementation. What is it for?

onnxruntime/core/providers/cpu/llm/attention.cc
```diff
@@ -0,0 +1,749 @@
...
+
+  // mask_data has been filled with 0, and its shape is BxSxT
+  T* p_mask = mask_data;
+
+  // 4D mask in Megatron GPT2 is currently not support in CPU kernel
+  if (nullptr != mask_index && mask_index_dims.size() == 4) {
+    ORT_NOT_IMPLEMENTED("4D mask in attention cpu kernel is not supported");
+  }
+
+  // For 3D mask, convert values 0 to mask_filter_value, and 1 to 0.0, then apply unidirectional mask if any.
```
tianleiwu: This code is for the contrib op. 
I think the Attention in onnx domain only support 4D mask.

onnxruntime/core/providers/cpu/llm/attention.cc
```diff
@@ -0,0 +1,749 @@
...
+  ORT_RETURN_IF_ERROR(context->GetTempSpaceAllocator(&allocator));
+
+  auto* tp = context->GetOperatorThreadPool();
+
+  Tensor* present = nullptr;
+  int past_sequence_length;
+  if (parameters.past_sequence_length == 0) {
+    if (present_key == nullptr && present_value == nullptr) {
+      past_sequence_length = (nullptr != past) ? static_cast<int>(past->Shape().GetDims()[3]) : 0;
+      present = this->GetPresent(context,
```
gramalingam: I am a bit confused about what this is doing. It is invoked if present-key and present-value are nullptr, but the implementation seems to be accessing Output(1), which should be null anyway?

onnxruntime/core/providers/cpu/llm/attention_helper.h
```diff
@@ -0,0 +1,66 @@
...
+#include "core/common/common.h"
+#include "core/providers/common.h"
+
+namespace onnxruntime {
+namespace attention_helper {
+
+enum AttentionMaskType {
+  MASK_NONE,  // No mask
+  MASK_BOOL,  // Boolean mask with shape [batch_size, sequence_length, total_sequence_length] or [batch_size, q_num_heads, q_sequence_length, total_sequence_length
+  MASK_ADD,   // Additive mask with shape [batch_size, sequence_length, total_sequence_length] or [batch_size, q_num_heads, q_sequence_length, total_sequence_length
```
tianleiwu: Why do we need this enum? We can just use whether the mask input exists and its data type in the code if needed. Like `mask == nullptr`, or `std::is_same_v<U, bool>` etc.

The additive mask is similar to attention_bias in the contrib op.

onnxruntime/core/providers/cpu/llm/attention.cc
```diff
@@ -0,0 +1,749 @@
...
+template <typename T>
+inline void ComputeAttentionSoftcapInplace(T* scores, int sequence_length, T softcap) {
+  MlasComputeSoftcap(scores, scores, sequence_length, softcap);
+}
+
+template <typename T>
+Attention<T>::Attention(const OpKernelInfo& info) : AttentionBase<T>(info) {
+  is_causal_ = static_cast<int>(info.GetAttrOrDefault<int64_t>("is_causal", 0)) == 1;
+  kv_num_heads_ = static_cast<int>(info.GetAttrOrDefault<int64_t>("kv_num_heads", 0));
+  q_num_heads_ = static_cast<int>(info.GetAttrOrDefault<int64_t>("q_num_heads", 0));
```
titaiwangms, line 100: Better add a reason why we set default here (3d/4d discussion).

onnxruntime/core/providers/cpu/llm/attention.cc
```diff
@@ -0,0 +1,749 @@
...
+
+template <typename T>
+inline void ComputeAttentionSoftcapInplace(T* scores, int sequence_length, T softcap) {
+  MlasComputeSoftcap(scores, scores, sequence_length, softcap);
+}
+
+template <typename T>
+Attention<T>::Attention(const OpKernelInfo& info) : AttentionBase<T>(info) {
+  is_causal_ = static_cast<int>(info.GetAttrOrDefault<int64_t>("is_causal", 0)) == 1;
+  kv_num_heads_ = static_cast<int>(info.GetAttrOrDefault<int64_t>("kv_num_heads", 0));
```
titaiwangms, line 99: Should we add ORT_ENFORCE about the relationship between kn_num_heads and q_num_heads?
We only support the following three types:

* Multi-headed Attention (MHA):  q_num_heads = kv_num_heads.
* Group-query Attention (GQA): q_num_heads > kv_num_heads, q_num_heads % kv_num_heads == 0.
* Multi-query Attention (MQA):  q_num_heads > kv_num_heads, kv_num_heads=1.

onnxruntime/core/providers/cpu/llm/attention.cc
```diff
@@ -0,0 +1,749 @@
...
+  q_num_heads_ = static_cast<int>(info.GetAttrOrDefault<int64_t>("q_num_heads", 0));
+  int mode = static_cast<int>(info.GetAttrOrDefault<int64_t>("qk_matmul_output_mode", 0));
+  qk_matmul_output_mode_ = static_cast<QKMatMulOutputMode>(mode);
+  ORT_ENFORCE(qk_matmul_output_mode_ == QKMatMulOutputMode::kQK ||
+                  qk_matmul_output_mode_ == QKMatMulOutputMode::kQKMask ||
+                  qk_matmul_output_mode_ == QKMatMulOutputMode::kQKSoftCap ||
+                  qk_matmul_output_mode_ == QKMatMulOutputMode::kQKSoftMax,
+              "qk_matmul_output_mode must be 0, 1, 2, or 3.");
+  scale_ = info.GetAttrOrDefault<float>("scale", std::numeric_limits<T>::quiet_NaN());
```
titaiwangms, line 112: We can put a comment saying the default will be applied later in the implementation.

onnxruntime/core/providers/cpu/llm/attention.cc
```diff
@@ -0,0 +1,749 @@
...
+  ORT_ENFORCE(Q != nullptr && K != nullptr && V != nullptr,
+              "Q, K, and V inputs must not be null");
+  int q_dims = onnxruntime::narrow<int>(Q->Shape().NumDimensions());
+  int k_dims = onnxruntime::narrow<int>(K->Shape().NumDimensions());
+  int v_dims = onnxruntime::narrow<int>(V->Shape().NumDimensions());
+  ORT_ENFORCE(q_dims == 3 || q_dims == 4, "Q must be a 3D or 4D tensor");
+  ORT_ENFORCE(q_dims == k_dims, "Q and K must have the same rank.");
+  ORT_ENFORCE(q_dims == v_dims, "Q and V must have the same rank.");
+
+  ORT_ENFORCE(Q->Shape()[0] == K->Shape()[0], "inconsistent batch_size");
```
titaiwangms:
```suggestion
  ORT_ENFORCE(Q->Shape()[0] == K->Shape()[0], "inconsistent batch_size of Q and K");
```

onnxruntime/core/providers/cpu/llm/attention.cc
```diff
@@ -0,0 +1,749 @@
...
+  ORT_ENFORCE(Q->Shape()[0] == V->Shape()[0], "inconsistent batch_size");
```
titaiwangms:
```suggestion
  ORT_ENFORCE(Q->Shape()[0] == V->Shape()[0], "inconsistent batch_size of Qand V");
```

onnxruntime/core/providers/cpu/llm/attention.cc
```diff
@@ -0,0 +1,749 @@
...
+  ORT_ENFORCE(past_key == nullptr || Q->Shape()[0] == past_key->Shape()[0], "inconsistent batch_size");
```
titaiwangms: The message can be more specific to who that are having inconsistent batch_size.

onnxruntime/core/providers/cpu/llm/attention.cc
titaiwangms: This one as well.

onnxruntime/core/providers/cpu/llm/attention.cc
```diff
@@ -0,0 +1,749 @@
...
+  std::vector<int64_t> y_shape;
+  if (q_dims == 4) {
+    // 4D
+    parameters.kv_num_heads = kv_num_heads_ > 0 ? kv_num_heads_ : onnxruntime::narrow<int>(K->Shape()[1]);  // K.shape[1] or V.shape[1] (4D)
+    parameters.q_num_heads = q_num_heads_ > 0 ? q_num_heads_ : onnxruntime::narrow<int>(Q->Shape()[1]);     // Q.shape[1] (4D)
+
+    ORT_ENFORCE(parameters.kv_num_heads == onnxruntime::narrow<int>(K->Shape()[1]), "kv_num_heads different from K.shape[1]");
+    ORT_ENFORCE(parameters.kv_num_heads == onnxruntime::narrow<int>(V->Shape()[1]), "kv_num_heads different from V.shape[1]");
+    ORT_ENFORCE(parameters.q_num_heads == onnxruntime::narrow<int>(Q->Shape()[1]), "q_num_heads different from Q.shape[1]");
+    ORT_ENFORCE(Q->Shape()[3] == K->Shape()[3], "inconsistent head_size");
```
titaiwangms:
```suggestion
    ORT_ENFORCE(Q->Shape()[3] == K->Shape()[3], "inconsistent head_size of Q and K");
```

onnxruntime/core/providers/cpu/llm/attention_helper.h
```diff
@@ -0,0 +1,66 @@
...
+  int kv_num_heads;  // K.shape[1] or V.shape[1] (4D)
+  int q_num_heads;   // Q.shape[1] (4D)
+  float scale;
+  float softcap;
+  int softmax_precision;
+  QKMatMulOutputMode qk_matmul_output_mode;
+  AttentionMaskType mask_type;
+
+  // From shapes
+  int batch_size;            // Q.shape[0], K.shape[0], V.shape[0] (4D)
```
titaiwangms: Do we need to provide 3D note also? Otherwise, it seems a bit confusing (looks like we only allow 4D here).

onnxruntime/core/providers/cpu/llm/attention.cc
```diff
@@ -0,0 +1,749 @@
...
+                                          static_cast<int64_t>(parameters.q_num_heads),
+                                          static_cast<int64_t>(parameters.q_sequence_length),
+                                          static_cast<int64_t>(parameters.past_sequence_length + parameters.kv_sequence_length)};
+
+  Tensor* Y = context->Output(0, y_shape);
+  Tensor* present_key = context->Output(1, present_key_shape);
+  Tensor* present_value = context->Output(2, present_value_shape);
+  Tensor* output_qk = context->Output(3, output_qk_shape);
+
+  return this->ApplyAttention(context,
```
titaiwangms, line 161: In general, I think we should wrap the input check above into CheckInputs function (following ORT convention), because we can really reuse those lines when we implement CUDA Attention.

onnxruntime/core/providers/cpu/llm/attention.cc
```diff
@@ -0,0 +1,749 @@
...
+                                             bool past_present_share_buffer,
+                                             int max_sequence_length,
+                                             attention_helper::QKMatMulOutputMode qk_matmul_output_mode) const {
+  const int total_sequence_length = past_sequence_length + kv_sequence_length;              // T = P + L
+  const size_t past_chunk_length = static_cast<size_t>(past_sequence_length) * head_size;   // P x H
+  const size_t q_input_chunk_length = static_cast<size_t>(sequence_length) * head_size;     // S x H
+  const size_t k_input_chunk_length = static_cast<size_t>(kv_sequence_length) * head_size;  // L x H
+  const size_t present_chunk_length = past_chunk_length + k_input_chunk_length;             // T x H
+  const size_t cache_chunk_length = static_cast<size_t>(max_sequence_length) * head_size;   // M x H
+
```
titaiwangms: Why are the following omitted?
```cpp
    DUMP_CPU_TENSOR_INIT();
    DUMP_CPU_TENSOR("Q", Q, batch_size, num_heads_, sequence_length, head_size);
    DUMP_CPU_TENSOR("K", K, batch_size, num_heads_, total_sequence_length, head_size);
    DUMP_CPU_TENSOR("Attn_Bias", attn_bias_data, attn_bias_dims);
```

onnxruntime/core/providers/cpu/llm/attention.cc
```diff
@@ -0,0 +1,749 @@
...
+                                   int kv_sequence_length,
+                                   int past_sequence_length) const {
+  const int all_sequence_length = past_sequence_length + kv_sequence_length;
+
+  // mask_data has been filled with 0, and its shape is BxSxT
+  T* p_mask = mask_data;
+
+  // 4D mask in Megatron GPT2 is currently not support in CPU kernel
+  if (nullptr != mask_index && mask_index_dims.size() == 4) {
+    ORT_NOT_IMPLEMENTED("4D mask in attention cpu kernel is not supported");
```
titaiwangms: Does this also apply to ONNX Attention?

onnxruntime/test/providers/cpu/llm/attention_op_test.cc
```diff
@@ -0,0 +1,893 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
titaiwangms, line 1: Why do we only keep a subset of Attention op tests from `contrib_ops/attention_op_test.cc`? There are more than half of them not covered.

onnxruntime/core/providers/cpu/llm/attention_helper.h
```diff
@@ -0,0 +1,66 @@
...
+
+// Parameters deduced from node attributes and inputs/outputs.
+struct AttentionParameters {
+  // Attribute from ONNX definition
+  bool is_causal;
+  int kv_num_heads;  // K.shape[1] or V.shape[1] (4D)
+  int q_num_heads;   // Q.shape[1] (4D)
+  float scale;
+  float softcap;
+  int softmax_precision;
```
titaiwangms, line 33: Is this used?

onnxruntime/core/providers/cpu/llm/attention.cc
xadupre: I tested that in debug mode and it returns null when the onnx does not specify any name in the output.

onnxruntime/core/providers/cpu/llm/attention.cc
xadupre: True, I forgot to remove, I used the code from the contribops as a baseline. I kept it because I thaught it would be needed for one the cases supported by the onnx ops. I forgot to remove.

onnxruntime/core/providers/cpu/llm/attention.cc
xadupre: According to the documentation, the shape of the mask must be broadcastable so 4D, 3D, 2D are allowed. Most of the examples uses 2D.

onnxruntime/core/providers/cpu/llm/attention_helper.h
xadupre: I forgot to remove.

onnxruntime/core/providers/cpu/llm/attention.cc
xadupre: There is but I need to use mlas in a couple of places (mask addition for example).

xadupre: I prefer to do that in a second PR when this one is shipped.

onnxruntime/core/providers/cpu/llm/attention.cc
xadupre: I assumed it was used to debug the implementation and I did not need them. So I removed them.

onnxruntime/test/providers/cpu/llm/attention_op_test.cc
xadupre: The test I implemented here are coming from the backend test not the contrib ops. I implemented a couple of cases in 4D to cover all the cases. Then I ran the backend tests in python to make sure it was working on all the cases. CI is running those tests as well.

onnxruntime/core/providers/cpu/llm/attention.cc
xadupre: done

onnxruntime/core/providers/cpu/llm/attention.cc
xadupre: done

onnxruntime/core/providers/cpu/llm/attention.cc
xadupre: I was not sure I could take a dependency on contrib_ops which meant I would have to move some code and then include it from the contrib ops. But Attention is one op when contrib ops have three. It also did not support all the options for qk_matmul_ouput_mode. It seems easier to do the reverse, implement something supporting Attention and refactor contrib ops to use.

onnxruntime/core/providers/cpu/llm/attention_helper.h
xadupre: No. Removed.

onnxruntime/core/providers/cpu/llm/attention.cc
xadupre: fixed

onnxruntime/core/providers/cpu/llm/attention.cc
xadupre: I removed the unused parameters specific to the contrib ops.

onnxruntime/core/providers/cpu/llm/attention_helper.h
xadupre: It is used.

onnxruntime/core/providers/cpu/llm/attention_helper.h
xadupre: 3D is allowed, but mathematically it is 4D as well. I'll do another pass to remove the misleading information.

onnxruntime/core/providers/cpu/llm/attention.cc
xadupre: done

onnxruntime/core/providers/cpu/llm/attention.cc
xadupre: There is no test in the backend test with 4D mask but technically, it should work.

onnxruntime/core/providers/cpu/llm/attention_helper.h
xadupre: In fact having a 4D mask would mean having a batch dimension for the mask which would does not make sense for this op.

onnxruntime/core/providers/cpu/llm/attention_helper.h
xadupre: Not yet, it will be after enabling float16.

onnxruntime/core/providers/cpu/llm/attention.cc
xadupre: done

onnxruntime/core/providers/cpu/llm/attention.cc
xadupre: The documentation says:

**Attention mask**. Shape must be broadcastable to 4D tensor with shape (batch_size, q_num_heads, q_sequence_length, total_sequence_length) where total_sequence_length = past_sequence_length + kv_sequence_length. Two types of masks are supported. A boolean mask where a value of True indicates that the element should take part in attention. Also supports a float mask of the same type as query, key, value that is added to the attention score.

onnxruntime/test/providers/cpu/llm/attention_op_test.cc
titaiwangms: Sounds like ModelBuilder will need to change if it wants to use opset23.Attention.

onnxruntime/core/providers/cpu/llm/attention.cc
xadupre: The onnx backend tests only have 2D example. This PR only supports 2D for the time being. I'll extend that.

onnxruntime/core/providers/cpu/llm/attention.cc
xadupre: Some contrib ops seems to allow it but it is not allowed here. Another enforce would catch it. I'll simplify.

onnxruntime/core/providers/cpu/llm/attention.cc
xadupre: added one comment.

onnxruntime/core/providers/cpu/llm/attention.cc
xadupre: I assume kv_num_heads>1 for GQA otherwise MQA is included in GQA. All cases are already covered by ORT_ENFORCE.

kunal-vaishnavi:
This doesn't have to be done for this PR but what is the plan for differentiating between the many attention kernel implementations? There are now some implementations for the ORT contrib ops and some implementations for the new ONNX ops. We should find a way to consolidate them (similar to how the consolidation was done for LayerNorm when it was added to ONNX) because it seems code is being repeated everywhere across CPU and CUDA.

onnxruntime/core/providers/cpu/llm/attention.cc
```diff
@@ -0,0 +1,497 @@
...
+      // Convert bool mask to 0/1
+      make_copy(mask_data, mask_index->Data<bool>(), mask_index->Shape().Size());
+    } else if (mask_index != nullptr) {
+      // We make a copy because causal is True.
+      make_copy(mask_data, mask_index->Data<T>(), mask_index->Shape().Size());
+    }
+    if (causal) {
+      // This loop could be parallelized.
+      int n_iter = mask_batch_size * mask_num_heads;
+      for (int i = 0; i < n_iter; ++i) {
```
kunal-vaishnavi, line 262: Can the [existing parallelized implementation](https://github.com/microsoft/onnxruntime/blob/4ac95cee1d94e58e4f7ab569f4dbafe8ba1d26bd/onnxruntime/contrib_ops/cpu/bert/gqa_attention_base.h#L192) be used?

onnxruntime/core/providers/cpu/llm/attention_helper.h
```diff
@@ -0,0 +1,86 @@
...
+
+enum AttentionMaskType {
+  MASK_NONE,  // No mask
+  MASK_BOOL,  // Boolean mask with shape [batch_size, sequence_length, total_sequence_length] or [batch_size, q_num_heads, q_sequence_length, total_sequence_length
+  MASK_ADD,   // Additive mask with shape [batch_size, sequence_length, total_sequence_length] or [batch_size, q_num_heads, q_sequence_length, total_sequence_length
+};
+
+enum AttentionType {
+  kInvalid = 0,               // Invalid attention type
+  kMultiHeadedAttention = 1,  // Multi-headed attention (MHA)
```
kunal-vaishnavi:
```suggestion
  kMultiHeadAttention = 1,  // Multi-headed attention (MHA)
```

onnxruntime/core/providers/cpu/llm/attention.cc
xadupre: This implementation is about GQA. In the existing contrib ops, there are different implementation for GQA, MQA, ... The implementation is roughly the same except the parallelization settings. With GQA, Gemm is call with no bias, it is applied after as well as the mask. So the causal mask is computed more than in the current implementation if the mask has not the same shape as the inputs. We can move things around. I suggest we keep this implementation and try to improve it in a second PR.

onnxruntime/core/providers/cpu/llm/attention_helper.cc
```diff
@@ -0,0 +1,135 @@
...
+    parameters.transpose_output = false;                                      // whether to transpose the output from BxNxSxH to BxSxNxH
+    parameters.q_sequence_length = onnxruntime::narrow<int>(Q->Shape()[2]);   // Q.shape[2] (4D)
+    parameters.head_size = onnxruntime::narrow<int>(Q->Shape()[3]);           // Q.shape[3] (4D)
+    parameters.kv_sequence_length = onnxruntime::narrow<int>(K->Shape()[2]);  // K.shape[2] or V.shape[2] (4D)
+    parameters.v_head_size = onnxruntime::narrow<int>(V->Shape()[3]);         // V.shape[3] (4D)
+    parameters.past_sequence_length = past_key == nullptr                     // past_key.shape[2] or past_value.shape[2] (4D) or given by the mask
+                                          ? 0
+                                          : onnxruntime::narrow<int>(past_key->Shape()[2]);
+
+    ORT_ENFORCE(attn_mask == nullptr || attn_mask->Shape()[attn_mask->Shape().NumDimensions() - 1] == parameters.past_sequence_length + parameters.kv_sequence_length,
```
tianleiwu: Set `parameters.total_sequence_length=parameters.past_sequence_length + parameters.kv_sequence_length` then use it in this function.

Please also check whether there are other parameters not set.

Normally, we have a function CheckInputs in helper to check inputs and set parameters. That could be re-used in both CPU and CUDA.

onnxruntime/core/providers/cpu/llm/attention_helper.h
tianleiwu: Not removed yet?

onnxruntime/core/providers/cpu/llm/attention_helper.h
```diff
@@ -0,0 +1,86 @@
...
+};
+
+enum AttentionType {
+  kInvalid = 0,              // Invalid attention type
+  kMultiHeadAttention = 1,   // Multi-headed attention (MHA)
+  kGroupQueryAttention = 2,  // Group query attention (GQA)
+  kMultiQueryAttention = 3,  // Multi-query attention (MQA)
+};
+
+enum QKMatMulOutputMode {
```
tianleiwu, line 12: Shall we have a value for not output?

onnxruntime/core/providers/cpu/llm/attention_helper.h
xadupre: It is used to guess the AttentionType. It is not really used so I can remove it. I let it as it can be used to improve the error messages.

onnxruntime/core/providers/cpu/llm/attention_helper.h
```diff
@@ -0,0 +1,86 @@
...
+  // From shapes
+  int batch_size;            // Q.shape[0], K.shape[0], V.shape[0] (4D)
+  int q_sequence_length;     // Q.shape[2] (4D)
+  int head_size;             // Q.shape[3] or K.shape[3 (4D)
+  int kv_sequence_length;    // K.shape[2] or V.shape[2] (4D)
+  int v_head_size;           // V.shape[4] (4D)
+  int past_sequence_length;  // pask_key.shape[2] or past_value.shape[2] (4D)
+  bool transpose_output;     // Whether to transpose the output from BxNxSxH to BxSxNxH
+
+  AttentionType getAttentionType() const {
```
tianleiwu: This function is not needed.

onnxruntime/core/providers/cpu/llm/attention_helper.h
xadupre: In that case, the onnx models only specifies 3 outputs at most. The pointer for this output is null.

onnxruntime/core/providers/cpu/llm/attention_helper.cc
xadupre: I do this in ComputeOutputShapeForAttention as well as computing the output shape. It is part of attention_helper.h.

xadupre: I'll add ``parameters.total_sequence_length=parameters.past_sequence_length + parameters.kv_sequence_length``.

onnxruntime/core/providers/cpu/llm/attention_helper.h
xadupre: I'll remove.

onnxruntime/core/providers/cpu/llm/attention_helper.cc
```diff
@@ -0,0 +1,135 @@
...
+    // 4D
+    parameters.kv_num_heads = kv_num_heads > 0 ? kv_num_heads : onnxruntime::narrow<int>(K->Shape()[1]);  // K.shape[1] or V.shape[1] (4D)
+    parameters.q_num_heads = q_num_heads > 0 ? q_num_heads : onnxruntime::narrow<int>(Q->Shape()[1]);     // Q.shape[1] (4D)
+
+    ORT_ENFORCE(parameters.kv_num_heads == onnxruntime::narrow<int>(K->Shape()[1]), "kv_num_heads different from K.shape[1]");
+    ORT_ENFORCE(parameters.kv_num_heads == onnxruntime::narrow<int>(V->Shape()[1]), "kv_num_heads different from V.shape[1]");
+    ORT_ENFORCE(parameters.q_num_heads == onnxruntime::narrow<int>(Q->Shape()[1]), "q_num_heads different from Q.shape[1]");
+    ORT_ENFORCE(Q->Shape()[3] == K->Shape()[3], "inconsistent head_size");
+    ORT_ENFORCE(K->Shape()[2] == V->Shape()[2], "inconsistent kv_sequence_length");
+    ORT_ENFORCE(attn_mask == nullptr || attn_mask->Shape()[attn_mask->Shape().NumDimensions() - 2] == Q->Shape()[2], "inconsistent q_sequence_length (between attn_mask and Q)");
```
tianleiwu, line 78: what if the attn_mask shape dimension is less than 2? I think we need validate the dimension ==4 first.

May use some macro similar to: https://github.com/microsoft/onnxruntime/blob/9de58ac7a3d18d6ae7f7ae502b3f91361067f1b5/onnxruntime/core/util/shape_checker.h#L18

That macro has return. It is easy to add one with ENFORCE. It will be easy to read like:

`ENFORCE_TENSOR_DIMS(attn_mask, batch_size, kv_num_heads, q_sequence_length, total_sequence_length)`

onnxruntime/core/providers/cpu/llm/attention_helper.cc
```diff
@@ -0,0 +1,135 @@
...
+  parameters.is_causal = is_causal;
+  parameters.softcap = softcap;
+  parameters.softmax_precision = softmax_precision;
+  parameters.qk_matmul_output_mode = qk_matmul_output_mode;                   // output mode for Q*K matmul
+  parameters.mask_type = attn_mask == nullptr ? AttentionMaskType::MASK_NONE  // mask type
+                                              : (attn_mask->IsDataType<bool>() ? AttentionMaskType::MASK_BOOL : AttentionMaskType::MASK_ADD);
+  parameters.batch_size = onnxruntime::narrow<int>(Q->Shape()[0]);  // Q.shape[0], K.shape[0], V.shape[0] (4D)
+
+  ORT_ENFORCE(parameters.batch_size > 0, "Batch size must be greater than 0");
+  ORT_ENFORCE(parameters.mask_type == AttentionMaskType::MASK_NONE ||
```
tianleiwu: This is not needed. It is in operator spec, which means mask must be either None, bool or T. Invalid case will never reach this line.

onnxruntime/core/providers/cpu/llm/attention_helper.h
```diff
@@ -0,0 +1,86 @@
...
+#pragma once
+#include "core/common/common.h"
+#include "core/providers/common.h"
+
+namespace onnxruntime {
+namespace attention_helper {
+
+enum AttentionMaskType {
+  MASK_NONE,  // No mask
+  MASK_BOOL,  // Boolean mask with shape [batch_size, sequence_length, total_sequence_length] or [batch_size, q_num_heads, q_sequence_length, total_sequence_length
```
tianleiwu: I think onnx spec only mention 4D mask for the op. 

If we really want support 2D like or 3D like mask, we may start with [B, 1, S, T] or [1, 1, S, T] shape. Since such shape allows broadcasting. That's the shape we supported for attention_bias (or MASK_ADD here).

onnxruntime/core/providers/cpu/llm/attention_helper.cc
gramalingam: The ONNX spec says the mask should be broadcastable to this 4D shape (it doesn't have to be this shape).

onnxruntime/core/providers/cpu/llm/attention_helper.h
gramalingam: The ONNX spec says `Attention mask. Shape must be broadcastable to 4D tensor with shape (batch_size, q_num_heads, q_sequence_length, total_sequence_length)`

onnxruntime/core/providers/cpu/llm/attention_helper.h
gramalingam: While it is not necessary for this opset 23 version, I think it would be useful to have a value for "no output", may be -1. We should fix this in ONNX 24 spec.

tianleiwu: I understand that we can check the output to see whether it need output. Adding a enum value to use internally could avoid confusion.

onnxruntime/core/providers/cpu/llm/attention_helper.cc
tianleiwu: You are right. Then we shall have a function to check broadcast able. 

At least, we shall support the following shapes:
[batch_size, 1, q_sequence_length, total_sequence_length]
[batch_size, kv_num_heads, q_sequence_length, total_sequence_length]

If possible, we can also support another shape for padding mask:
[batch_size, 1, 1, total_sequence_length]

onnxruntime/core/providers/cpu/llm/attention_helper.h
tianleiwu: Right. I think the comment need be updated to mention the broadcast able.

BTW, [batch_size, sequence_length, total_sequence_length] cannot be broadcasted to the target shape based on broadcasting rules.

onnxruntime/core/providers/cpu/llm/attention_helper.h
xadupre: removed

onnxruntime/core/providers/cpu/llm/attention_helper.h
xadupre: removed

onnxruntime/core/providers/cpu/llm/attention_helper.cc
xadupre: done

onnxruntime/core/providers/cpu/llm/attention_helper.h
xadupre: done

onnxruntime/core/providers/cpu/llm/attention_helper.cc
xadupre: removed mask_type

onnxruntime/core/providers/cpu/llm/attention_helper.cc
xadupre: done

onnxruntime/core/providers/cpu/llm/attention_helper.h
xadupre: done

onnxruntime/core/providers/cpu/llm/attention.cc
```diff
@@ -178,11 +182,14 @@ void AttentionBase<T>::ComputeAttentionProbs(T* attention_probs,
   const float alpha = parameters.scale;
 
   TensorOpCost unit_cost;
-  const ptrdiff_t probs_matrix_size = SafeInt<ptrdiff_t>(parameters.q_sequence_length) * total_sequence_length;
+  const ptrdiff_t probs_matrix_size = SafeInt<ptrdiff_t>(parameters.q_sequence_length) *
```
tianleiwu, line 200: nit: it is better to define loop_len and unit_test near the main loop (before line 273).

onnxruntime/core/providers/cpu/llm/attention.cc
```diff
@@ -316,17 +329,19 @@ void AttentionBase<T>::ComputeAttentionProbs(T* attention_probs,
...
-                         MLFloat16(beta),                                                                                                                                    // 8
-                         output,                                                                                                                                             // 9
-                         &c_shape,                                                                                                                                           // 10
-                         output,                                                                                                                                             // 11
-                         nullptr);                                                                                                                                           // 12
+          TensorShape c_shape({parameters.q_sequence_length, parameters.total_sequence_length});
+          Gemm_MLFloat16(CblasNoTrans, CblasTrans,                                  // 0, 1
```
tianleiwu: nit: make the comment more helpful.

onnxruntime/core/providers/cpu/llm/attention.cc
```diff
@@ -473,19 +490,22 @@ void AttentionBase<T>::ComputeVxAttentionScore(T* output,                 // buf
...
-                               v,                                                                                                                            // 7
-                               MLFloat16(0.f),                                                                                                               // 8
-                               nullptr,                                                                                                                      // 9
-                               nullptr,                                                                                                                      // 10
-                               current_tmp_data,                                                                                                             // 11
-                               nullptr);                                                                                                                     // 12
+                Gemm_MLFloat16(CblasNoTrans, CblasNoTrans,                     // 0, 1
```
tianleiwu: nit: update/remove the comment. For example
```
// trans_a
// trans_b,
// M
// N
// K,
// alpha,
// a_data
// b_data,
// beta,
// c_data
// c_shape,
// y_data,
// thread_pool
```

onnxruntime/core/providers/cpu/llm/attention.cc
```diff
@@ -522,19 +542,22 @@ void AttentionBase<T>::ComputeVxAttentionScore(T* output,                 // buf
...
-                               v,                                                                                                                            // 7
-                               MLFloat16(0.f),                                                                                                               // 8
-                               nullptr,                                                                                                                      // 9
-                               nullptr,                                                                                                                      // 10
-                               dest,                                                                                                                         // 11
-                               nullptr);                                                                                                                     // 12
+                Gemm_MLFloat16(CblasNoTrans, CblasNoTrans,                     // 0, 1
```
tianleiwu: nit: update/remove the comment.

onnxruntime/core/providers/cpu/llm/attention_helper.cc
```diff
@@ -126,10 +141,15 @@ void ComputeOutputShapeForAttention(
...
-                     static_cast<int64_t>(parameters.q_num_heads),
-                     static_cast<int64_t>(parameters.q_sequence_length),
-                     static_cast<int64_t>(parameters.past_sequence_length + parameters.kv_sequence_length)};
+  if (qk_matmul_output_mode == QKMatMulOutputMode::kNone) {
+    output_qk_shape.clear();
+  } else {
+    output_qk_shape = {static_cast<int64_t>(parameters.batch_size),
+                       static_cast<int64_t>(parameters.q_num_heads),
+                       static_cast<int64_t>(parameters.q_sequence_length),
+                       static_cast<int64_t>(parameters.past_sequence_length + parameters.kv_sequence_length)};
```
tianleiwu: nit: use parameters.total_sequence_length here. Same above

onnxruntime/core/providers/cpu/llm/attention_helper.cc
```diff
@@ -2,11 +2,26 @@
...
+  ORT_ENFORCE(q_sequence_length > 0, "Q sequence length must be greater than 0");
+  ORT_ENFORCE(kv_sequence_length > 0, "KV sequence length must be greater than 0");
+  ORT_ENFORCE(head_size > 0, "Head size must be greater than 0");
+  ORT_ENFORCE(v_head_size > 0, "V head size must be greater than 0");
+  ORT_ENFORCE(past_sequence_length >= 0, "Past sequence length must be non-negative");
+  ORT_ENFORCE(total_sequence_length > 0, "Total sequence length must be greater than 0");
+  ORT_ENFORCE(kv_num_heads > 0, "KV number of heads must be greater than 0");
+  ORT_ENFORCE(q_num_heads > 0, "Q number of heads must be greater than 0");
+  ORT_ENFORCE(total_sequence_length == past_sequence_length + kv_sequence_length,
+              "Total sequence length must be equal to past sequence length plus KV sequence length");
```
tianleiwu, line 21: Also check q_num_heads % kv_num_heads == 0

onnxruntime/core/providers/cpu/llm/attention.cc
xadupre: done

onnxruntime/core/providers/cpu/llm/attention.cc
xadupre: I removed. I used position because the error message from VS use positions.

onnxruntime/core/providers/cpu/llm/attention_helper.cc
xadupre: done

onnxruntime/core/providers/cpu/llm/attention_helper.cc
xadupre: done

onnxruntime/core/providers/cpu/llm/attention.cc
xadupre: done

onnxruntime/core/providers/cpu/llm/attention.cc
xadupre: removed

justinchuby (approved):

onnxruntime/core/providers/cpu/llm/attention.cc
```diff
@@ -0,0 +1,736 @@
...
+                               ? static_cast<QKMatMulOutputMode>(mode)
+                               : QKMatMulOutputMode::kNone;
+  ORT_ENFORCE(qk_matmul_output_mode_ == QKMatMulOutputMode::kNone ||
+                  qk_matmul_output_mode_ == QKMatMulOutputMode::kQK ||
+                  qk_matmul_output_mode_ == QKMatMulOutputMode::kQKMask ||
+                  qk_matmul_output_mode_ == QKMatMulOutputMode::kQKSoftCap ||
+                  qk_matmul_output_mode_ == QKMatMulOutputMode::kQKSoftMax,
+              "qk_matmul_output_mode must be 0, 1, 2, or 3.");
+  // The default scale depends on the input dimensions. It is set to nan to indicate that it should be computed.
+  scale_ = info.GetAttrOrDefault<float>("scale", std::numeric_limits<T>::quiet_NaN());
```
kunal-vaishnavi, line 112: Shouldn't the [default value](https://onnx.ai/onnx/operators/onnx__Attention.html#attributes) be $\frac{1}{\sqrt{H}}$ where $H$ = head size?

xadupre: (scale A) @ (scale B) = scale^2 (A @ B)

onnxruntime/core/providers/cpu/llm/attention.cc
```diff
@@ -0,0 +1,736 @@
...
+                                             const T* Q,                             // Q data. Its size is BxNxSxH
+                                             const T* K,                             // k data. Its size is BxNxLxH
+                                             const Tensor* mask_index,               // mask
+                                             const AttentionParameters& parameters,  // attention parameters
+                                             const T* past_key,                      // past key only (if not using past state)
+                                             T* present_key,                         // present key only (if not using present state)
+                                             T* output_qk,                           // Q*K output
+                                             ThreadPool* tp,
+                                             AllocatorPtr allocator) const {
+  // The case pask_key != nullptr and present_key == nullptr is not supported.
```
kunal-vaishnavi:
```suggestion
  // The case past_key != nullptr and present_key == nullptr is not supported.
```

justinchuby (approved):

onnxruntime/test/testdata/onnx_backend_test_series_filters.jsonc
```diff
@@ -31,6 +31,7 @@
     "current_failing_tests": [
         "^test_adagrad",
         "^test_adagrad_multiple",
+        "^test_attention_3d.*",  // wrong expected values in onnx==1.18.0, fixed in 1.19.0
```
titaiwangms, line 34: Better add it as TODO(someone)

onnxruntime/core/providers/cpu/math/gemm.cc
```diff
@@ -237,6 +236,18 @@ void Gemm<MLFloat16>::ComputeGemm(CBLAS_TRANSPOSE trans_a, CBLAS_TRANSPOSE trans
...
+template <>
+void Gemm<MLFloat16>::ComputeGemm(CBLAS_TRANSPOSE trans_a, CBLAS_TRANSPOSE trans_b,
+                                  ptrdiff_t M, ptrdiff_t N, ptrdiff_t K,
+                                  MLFloat16 alpha,
+                                  const MLFloat16* a_data, const MLFloat16* b_data,
+                                  MLFloat16 beta,
+                                  const MLFloat16* c_data, const TensorShape* c_shape,
+                                  MLFloat16* y_data,
+                                  concurrency::ThreadPool* thread_pool) {
+  Gemm_MLFloat16(trans_a, trans_b, M, N, K, alpha, a_data, b_data, beta, c_data, c_shape, y_data, thread_pool);
```
titaiwangms, line 248: Do we have tests for Gemma fp16?

titaiwangms: Looks like this is a new support.

xadupre: In that file, I just moved the code of gemm in a standalone function to be able to call it from another kernel. So the code is already tested by the test covered Gemm.

onnxruntime/test/testdata/onnx_backend_test_series_filters.jsonc
titaiwangms: But you can merge it to unblock yourself from CUDA work first

tianleiwu (approved):

snnn:

PR #25158, opened 2025-06-24
Qnn license file update

HectorSVC:
### Description
Use the license file from QNN SDK to make sure it's up to date.


jywu-mysoft:
do we need a way of detecting if the license file in QNN SDK gets renamed or is missing.  
should we fail

adrianlizarraga:
> do we need a way of detecting if the license file in QNN SDK gets renamed or is missing. should we fail

I could be wrong but I think the tools/nuget/generate_nuspec_for_native_nuget.py errors out if the license file is not there?
image

snnn:

PR #25159, opened 2025-06-24
[EP ABI] Update OrtGraph to use new OrtValues stored in internal Graph

adrianlizarraga:
### Description
Updates the `OrtGraph` implementation to take advantage of the work done in PR https://github.com/microsoft/onnxruntime/pull/23979, which sets the infrastructure to store initializers as `OrtValue` instances in the `onnxruntime::Graph`.

There still needs to be second part to the [aforementioned PR](https://github.com/microsoft/onnxruntime/pull/23979) to ensure that all initializers are stored as `OrtValue`s in the Graph.

### DIFF
--- a/onnxruntime/core/graph/ep_api_types.cc
+++ b/onnxruntime/core/graph/ep_api_types.cc
@@ -544,9 +544,9 @@ Status EpGraph::Create(const GraphViewer& graph_viewer, /*out*/ std::unique_ptr<
 
       EpValueInfo* outer_value_info = value_info_iter->second.get();
       bool is_constant = false;
-      auto initializer_value = std::make_unique<OrtValue>();
+      auto outer_initializer_value = std::make_unique<OrtValue>();
       const ONNX_NAMESPACE::TensorProto* outer_initializer = parent_graph->GetInitializer(implicit_name,
-                                                                                          *initializer_value,
+                                                                                          *outer_initializer_value,
                                                                                           is_constant,
                                                                                           /*check_outer_scope*/ true);
       outer_value_info->SetFlag(EpValueInfo::kIsOuterScope);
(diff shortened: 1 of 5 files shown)


onnxruntime/core/graph/ep_api_types.cc
```diff
@@ -485,12 +485,18 @@ Status EpGraph::Create(const GraphViewer& graph_viewer, /*out*/ std::unique_ptr<
...
     initializer_value_infos.push_back(value_info);
 
-    // Temporary: Copy onnx::TensorProto into OrtValue objects owned by this EpGraph.
-    // TODO: Remove this logic once a separate PR that updates onnxruntime::Graph to store initializers as
-    // OrtValue instances is merged.
+    // Initialize OrtValue for the initializer.
     auto initializer_value = std::make_unique<OrtValue>();
-    ORT_RETURN_IF_ERROR(utils::TensorProtoToOrtValue(Env::Default(), graph_viewer.ModelPath(), *tensor_proto,
-                                                     initializer_allocator, *initializer_value));
+    bool graph_has_ortvalue = graph_viewer.GetGraph().GetOrtValueInitializer(initializer_name, *initializer_value,
```
adrianlizarraga, line 490: Currently `graph_has_ortvalue` will almost always be `false`. There needs to a follow-up to https://github.com/microsoft/onnxruntime/pull/23979 to convert all initializers to `OrtValue`.

onnxruntime/core/graph/ep_api_types.cc
```diff
@@ -538,23 +544,26 @@ Status EpGraph::Create(const GraphViewer& graph_viewer, /*out*/ std::unique_ptr<
 
       EpValueInfo* outer_value_info = value_info_iter->second.get();
       bool is_constant = false;
+      auto initializer_value = std::make_unique<OrtValue>();
       const ONNX_NAMESPACE::TensorProto* outer_initializer = parent_graph->GetInitializer(implicit_name,
```
skottmckay, line 548: Do we need to heap allocate the OrtValue instance and use unique_ptr, or could we just use OrtValue directly given it contains a shared_ptr for the data?

adrianlizarraga: We could store them by value. The C API returns a pointer to an OrtValue. We could store them by value in the map (e.g., `std::unordered_map<std::string_view, OrtValue> outer_values`) and then return their address (e.g., `&outer_values.find(name)->second`). I didn't default to this because a std::unordered_map can be reallocated as it grows, and if we ever add modification capabilities to `OrtGraph`, any pointers we handed to the user would be silently invalidated.

This is not a problem now since `OrtGraph` is read-only (and we have the model editor api), but I supposed I'm being overly conservative. 

I'll update to store OrtValue by value and add maybe a comment that mentions this possibility.

adrianlizarraga: Updated

skottmckay: Good point about us handing out pointers. 

There's also a slightly different cost to using OrtValue vs unique_ptr<OrtValue> when it resizes, as moving the shared_ptr in OrtValue is a bit more expensive than moving a unique_ptr.

Maybe between making it robust against resizes, and any resize being cheaper, the heap allocation is better.

adrianlizarraga: Updated.

PR #25161, opened 2025-06-25
[CUDA] bfloat16 MatMulNBits

tianleiwu:
### Description

Support bfloat16 for MatMulNBits in CUDA.

### Motivation and Context

For LLM model with bfloat16 data type.


onnxruntime/contrib_ops/cuda/quantization/matmul_8bits.cu
```diff
@@ -125,22 +135,54 @@ __device__ __forceinline__ void AccumulateEightElements8b(
...
+  sums_f[5] = fmaf(v[5], a_vec_1.y, sums_f[5]);
+  sums_f[6] = fmaf(v[6], a_vec_1.z, sums_f[6]);
+  sums_f[7] = fmaf(v[7], a_vec_1.w, sums_f[7]);
+}
+
+
+// --- Device Function: Accumulate 8 Elements (bfloat16 precision) ---
+// Dequantizes 8 uint8_t values and accumulates the result with 8 nv_bfloat16 values from A.
+// sums_f += A_bf16 * dequant(B_quant)
+__device__ __forceinline__ void AccumulateEightElements8b(
```
nenad1002, line 152: Whats the purpose of this? To process more bfloat16 elements at the same time? Why only 8, warp specific?

onnxruntime/contrib_ops/cuda/quantization/matmul_4bits.cu
```diff
@@ -167,6 +168,59 @@ __device__ __forceinline__ void AccumulateEightElements4b(uint32_t values_quant,
...
+
+__device__ __forceinline__ void AccumulateEightElements4b(uint32_t values_quant, nv_bfloat16 scale, uint8_t zp, const nv_bfloat16* a, nv_bfloat16* sums) {
+#if defined(__CUDA_ARCH__) && __CUDA_ARCH__ >= 800
+    __nv_bfloat162 scale_bf162 = __bfloat162bfloat162(scale);
+    nv_bfloat16 zp_adjust = -scale * __uint2bfloat16_rn(zp);
+    __nv_bfloat162 zp_adjust2 = __bfloat162bfloat162(zp_adjust);
+
+    const uint4 vec_a = *(reinterpret_cast<const uint4*>(a));
+
+    constexpr uint32_t kLowHalf2 = 0x5410;
```
kunal-vaishnavi, line 199: Can we declare kernel constants at the top of the file?

https://github.com/microsoft/onnxruntime/blob/c95b016f7964643abe843c37ad5e713d67c76805/onnxruntime/contrib_ops/cuda/quantization/matmul_8bits.cu#L20-L24

onnxruntime/contrib_ops/cuda/quantization/matmul_4bits.cu
```diff
@@ -167,6 +168,59 @@ __device__ __forceinline__ void AccumulateEightElements4b(uint32_t values_quant,
   sums[7] += v7 * a_vec_1.w;
 }
 
+// Convert 8 4-bit integers stored in one uint32_t to 8 bfloat16s.
+// The output order is [0,4], [1,5], [2,6], [3,7]
+__device__ __forceinline__ void Convert8xInt4To8xBF16s(uint32_t value, __nv_bfloat162* bf16_2x4) {
+#if defined(__CUDA_ARCH__) && __CUDA_ARCH__ >= 800
+  const int i0 = (value >> 0) & 0xF;
```
kunal-vaishnavi, line 175:
```suggestion
    const int i0 = (value >> 0) & 0xF;
```

onnxruntime/core/providers/cuda/cuda_type_conversion.h
```diff
@@ -0,0 +1,92 @@
...
+#include <type_traits>
+#include <cstdint>
+#include "core/framework/int4.h"
+#include "core/framework/float8.h"
+#include "core/framework/float16.h"
+
+namespace onnxruntime {
+namespace cuda {
+
+// Type mapping for ORT Type to CUDA Type
```
kunal-vaishnavi, line 20: This mapping will be very helpful. Thanks for adding this!

onnxruntime/contrib_ops/cuda/quantization/matmul_8bits.cu
tianleiwu: It is because 4 bits x 8 could fit a register of 32 bits. That means you can load 8 elements from quantized weight into a register, then process them in a batch.

kElementsPerThreadPerIteration=8, which means each thread need accumulate 8 elements per iteration.

onnxruntime/test/contrib_ops/matmul_4bits_test.cc
```diff
@@ -473,27 +499,32 @@ namespace {
...
+template <typename T>
+void RunTest(int64_t M, int64_t N, int64_t K, int64_t block_size, bool has_zeropoint, bool zp_is_4bit = true,
+             float abs_error = 0.f, bool has_g_idx = false, bool has_bias = false) {
   TestOptions opts{};
   opts.M = M;
   opts.N = N;
   opts.K = K;
   opts.block_size = block_size;
-  opts.accuracy_level = accuracy_level;
+  opts.accuracy_level = 0;
```
kunal-vaishnavi, line 510: Why are we only testing accuracy level = 0?

onnxruntime/contrib_ops/cuda/quantization/matmul_4bits.cu
tianleiwu: This is specified for the order used in this function. I think it's fine to define it here.

The constants at the top of file might be shared by multiple functions.

onnxruntime/test/contrib_ops/matmul_4bits_test.cc
tianleiwu: accuracy level has been tested in `Float32_Accuracy0/1/2/3`.

For cuda kernel, accuracy level is ignored, so we only test level 0 for some test cases designed for GPU.

PR #25163, opened 2025-06-25
Re-enable Android CI tests

edgchen1:
### Description

Re-enable unit tests in Android CI build.

### Motivation and Context

The CI build is not running the unit tests. It should run them.


.github/workflows/android.yml
```diff
@@ -168,6 +168,7 @@ jobs:
           --build_shared_lib
           --cmake_generator=Ninja
           --build_java
+          --update --build --test
```
edgchen1, line 171: note that for cross-compiled builds (such as Android), the default behavior does not include `--test`.

https://github.com/microsoft/onnxruntime/blob/7d22c09d6488a77ee9d4503a7529774921e94f72/tools/ci_build/build_args.py#L892-L903

snnn: Good to know that. Thanks!

PR #25166, opened 2025-06-25
Make DMMHA kernel inside MHA optional for Whisper

kunal-vaishnavi:
### Description
This PR sets adding support for the `DecoderMaskedMultiHeadAttention` (DMMHA) kernel inside `MultiHeadAttention` (MHA) to false by default.

### Motivation and Context
The models containing the extra inputs for DMMHA (i.e. `past_sequence_length` and `cache_indirection`) have some runtime issues. Additionally, not all execution providers implement the DMMHA kernel inside MHA and will therefore not support these extra inputs.

### DIFF
--- a/onnxruntime/python/tools/transformers/models/whisper/convert_to_onnx.py
+++ b/onnxruntime/python/tools/transformers/models/whisper/convert_to_onnx.py
@@ -158,6 +158,14 @@ def parse_arguments(argv=None):
     )
     conversion_args.set_defaults(no_beam_search_op=False)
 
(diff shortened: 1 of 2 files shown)


baijumeswani (approved):
Would this impact perf on whisper cuda models?

kunal-vaishnavi:
> Would this impact perf on whisper cuda models?

Yes, performance will be temporarily reduced for now. Once the DMMHA models are working with ONNX Runtime GenAI for word-level timestamps, the flag can be set to true by default.

Issue #25108, opened 2025-06-18
Initializers use wrong allocator


AndreyOrb:
### Describe the issue

I have shared env. allocators in my program:
```
// Create shared GPU allocator (registered in env.)
try
{
	ASSERT_CUDA_STATUS(cudaSetDevice(gpuNum));

	static Ort::MemoryInfo memoryInfo("Cuda", OrtArenaAllocator, gpuNum, OrtMemTypeDefault);
	static Ort::ArenaCfg arenaConfig(0, ArenaExtendStrategy::kSameAsRequested, -1, -1);

	// This creates an ORT-internal allocator instance and registers it in the environment for sharing
	vector<const char*> keys, values;
	ASSERT_ORT_STATUS(api.CreateAndRegisterAllocatorV2(*_env, "CUDAExecutionProvider", memoryInfo, arenaConfig, keys.data(), values.data(), 0));

	NN_LOG(NN_INFO) << "Created ORT-internal shared GPU memory allocator for GPU device " << to_string(gpuNum + 1);
}
catch (const exception& e)
{
	THROW_EXCEPTION_WITH_EX("Could not create an ORT-internal shared GPU memory allocator", e);
}
```

The session is defined like this:
```
Ort::SessionOptions sessionOptions;
sessionOptions.AddConfigEntry(kOrtSessionOptionsUseDeviceAllocatorForInitializers, "0"); // Allocate initializers via Arena allocator
sessionOptions.AddConfigEntry(kOrtSessionOptionsConfigUseEnvAllocators, "1"); // Use allocators registered in the env instead of per session
```

As it turned out, initializers are loaded using session-local allocators, and only then replaced with env. allocators (https://github.com/microsoft/onnxruntime/blob/13c063186d572baa18cdb58f9395ba873efdfad2/onnxruntime/core/session/inference_session.cc#L2088):

```
    // now that we have all the execution providers, create the session state
    session_state_ = std::make_unique<SessionState>(
        model_->MainGraph(),
        execution_providers_,
        GetIntraOpThreadPoolToUse(),
        GetInterOpThreadPoolToUse(),
        data_transfer_mgr_,
        external_data_loader_mgr_,
        *session_logger_,
        session_profiler_,
        session_options_,
        prepacked_weights_container_);

    bool use_env_allocators =
        session_options_.config_options.GetConfigOrDefault(kOrtSessionOptionsConfigUseEnvAllocators, "0") == "1";
    if (use_env_allocators) {
[...]

PR #25175, opened 2025-06-25
Pass environment allocators into the session state, if the "session.use_env_allocators" flag was activated.

AndreyOrb:
### Description
By default, session state creates allocators for every activated EP.
It uses these session-local allocators for initializers.

Later on, if environment allocators are activates, session state replaces session-local allocators with global ones.
The right way is to use environment allocators for initializers, if "session.use_env_allocators" flag was activated.

This PR fixes the issue by providing env. allocators as a parameter to SessionState constructor.
This parameter is already present, but was not in use in this flow up till now.

### Motivation and Context
https://github.com/microsoft/onnxruntime/issues/25108

### DIFF
--- a/onnxruntime/core/framework/session_state.cc
+++ b/onnxruntime/core/framework/session_state.cc
@@ -103,19 +103,24 @@ SessionState::SessionState(Graph& graph,
 {
   enable_mem_pattern_ = sess_options_.enable_mem_pattern &&
                         sess_options_.execution_mode == ExecutionMode::ORT_SEQUENTIAL;
-  if (parent_allocators) {
-    allocators_ = parent_allocators;
-  } else {
-    allocators_unique_ptr_ = std::make_unique<AllocatorMap>();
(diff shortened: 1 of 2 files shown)


AndreyOrb:
The proposed PR has bug in it. 
If environment shared allocator for CudaPinned memory is created using CreateAndRegisterAllocatorV2, it will use CPUAllocator as a device_allocator backend. Instead, it must create allocator with CUDAPinnedAllocator backend.

So, in the proposed PR, when the shared env. allocators are activated, the CudaPinned allocator will use CPUAllocator. It will then pass this "buggy" allocator into the SessionState. SessionState will then disregard CUDA EP's CudaPinned pointing the correct CUDAPinnedAllocator.

I'm not sure if the same issue exists with other not regular allocators (OpenVINO, DML, etc), but the PR can only be merged after solving the problem above.

Closing the PR for now,..

PR #25177, opened 2025-06-25
[WebGPU EP] extend concat to handle large number of inputs

prathikr:
Resolves the following issue affecting the patchtst model:

```
Non-zero status code returned while running Concat node. Name:'/patching/Concat' 
Status Message: The number of storage buffer (input=32, output=1) exceeds the limit 
(8) of the device.
```


onnxruntime/core/providers/webgpu/tensor/concat.cc
```diff
@@ -125,12 +125,84 @@ Status Concat::ComputeInternal(ComputeContext& context) const {
...
+      // Calculate output shape for this pass
+      auto pass_shape = current_inputs[0]->Shape();
+      int64_t pass_concat_axis_size = 0;
+      for (size_t i = 0; i < max_inputs_per_pass; ++i) {
+        pass_concat_axis_size += current_inputs[i]->Shape()[axis];
+      }
+      pass_shape[axis] = pass_concat_axis_size;
+
+      // Create intermediate tensor to store result of this pass
+      auto intermediate = context.CreateGPUTensor(current_inputs[0]->DataType(), std::move(pass_shape));
```
fs-eire: Is it possible to always use `prepare.output_tensor` as output so that creating intermediate buffers can be avoided?

prathikr:
Continued in https://github.com/microsoft/onnxruntime/pull/25390

PR #25178, opened 2025-06-25
Add RotaryEmbeddings(23) - CUDA

titaiwangms:
Follow up #24980
Fix https://github.com/microsoft/onnxruntime/issues/24556

Add ONNX RotaryEmbedding(23) following https://github.com/onnx/onnx/blob/main/docs/Operators.md#RotaryEmbedding. The PR uses contrib op RotaryEmbedding implementation under the hood.

The main difference between this op and the contrib op is that the position_ids in ONNX RotaryEmbedding is optional. When it's not provided, cos_cache and sin_cache should be 3d.


onnxruntime/core/providers/cuda/llm/rotary_embedding_impl.cu
```diff
@@ -0,0 +1,165 @@
...
+#include "core/providers/cuda/cu_inc/common.cuh"
+#include <cuda_fp16.h>
+
+using namespace onnxruntime::cuda;
+
+namespace onnxruntime {
+namespace cuda {
+
+template <typename T>
+__global__ void RotaryEmbeddingBSNH(T* output,                    // BxSxNxH
```
kunal-vaishnavi, line 20: We should avoid copying this and re-use the same existing kernel

titaiwangms: I wanted to separate them because `position_ids` is optional to standard ONNX op: https://github.com/onnx/onnx/blob/main/docs/Operators.md#RotaryEmbedding.

https://github.com/microsoft/onnxruntime/blob/ce1029fe83eef0a008124d0004d5381241eff4c0/onnxruntime/core/providers/cuda/llm/rotary_embedding_impl.cu#L51-L65

titaiwangms: Just for my education, is the benefit of having fewer cuda kernels that it would be more efficient to onnxruntime? Or it's for cleaner coding? If it's latter, I think coupling contrib op and onnx op piles up complexity because the inputs (shape/value) of the function are different. I also give `position_ids_format` different definition where 0 means nullptr, and 1 means 2D now.

kunal-vaishnavi: Can we update how `position_ids_format` is defined below to have an option for when `position_ids = nullptr`?

https://github.com/microsoft/onnxruntime/blob/47ddaaa09d72f67299ccb9d70bd87523fdc696c5/onnxruntime/contrib_ops/cpu/bert/rotary_embedding_helper.h#L13-L26

titaiwangms: I have gone another direction to create the new CheckInputs though:
https://github.com/microsoft/onnxruntime/blob/47ddaaa09d72f67299ccb9d70bd87523fdc696c5/onnxruntime/core/providers/cpu/llm/rotary_embedding_helper.h#L23

These are the lines we need to add into helper if we want to merge them.
https://github.com/microsoft/onnxruntime/blob/47ddaaa09d72f67299ccb9d70bd87523fdc696c5/onnxruntime/core/providers/cpu/llm/rotary_embedding_helper.h#L82-L115

titaiwangms: Size 1 is banned in onnx rotary embedding to respect the spec: https://github.com/microsoft/onnxruntime/blob/47ddaaa09d72f67299ccb9d70bd87523fdc696c5/onnxruntime/core/providers/cpu/llm/rotary_embedding_helper.h#L134-L137

titaiwangms:
@tianleiwu PTAL

onnxruntime/core/providers/cuda/llm/rotary_embedding.cc
```diff
@@ -0,0 +1,85 @@
...
+
+template <typename T>
+Status RotaryEmbedding<T>::ComputeInternal(OpKernelContext* context) const {
+  const Tensor* input = context->Input<Tensor>(0);
+  const Tensor* cos_cache = context->Input<Tensor>(1);
+  const Tensor* sin_cache = context->Input<Tensor>(2);
+  const Tensor* position_ids = context->Input<Tensor>(3);  // Optional, can be nullptr
+
+  RotaryParameters parameters = {};
+  ORT_RETURN_IF_ERROR(rotary_embedding_helper::CheckInputs<Tensor>(input,
```
gramalingam, line 48: Is this the contrib op helper? Does it handle the case where position_ids is null?

onnxruntime/core/providers/cuda/llm/rotary_embedding_impl.cu
```diff
@@ -0,0 +1,165 @@
...
+  int4 out_strides;
+  if (is_input_bnsh_format) {
+    int in_head_stride = sequence_length * head_size;
+    int out_head_stride = sequence_length * head_size;
+    in_strides = int4{num_heads * in_head_stride, in_head_stride, in_head_stride / sequence_length, 1};
+    out_strides = int4{num_heads * out_head_stride, out_head_stride, out_head_stride / sequence_length, 1};
+  } else {
+    int in_head_stride = head_size;
+    int out_head_stride = head_size;
+    in_strides = int4{sequence_length * num_heads * in_head_stride, in_head_stride, num_heads * in_head_stride, 1};
```
tianleiwu: Maybe add a comment that the strides are for B, N, S, H here even though input format is BSNH.

onnxruntime/core/providers/cuda/llm/rotary_embedding_impl.cu
```diff
@@ -0,0 +1,165 @@
...
+                                   const int sequence_length, const int num_heads, const int head_size,
+                                   const int rotary_embedding_dim, const int max_sequence_length,
+                                   const int position_ids_format, const bool interleaved,
+                                   const int max_threads_per_block, const bool is_input_bnsh_format) {
+  int4 in_strides;
+  int4 out_strides;
+  if (is_input_bnsh_format) {
+    int in_head_stride = sequence_length * head_size;
+    int out_head_stride = sequence_length * head_size;
+    in_strides = int4{num_heads * in_head_stride, in_head_stride, in_head_stride / sequence_length, 1};
```
tianleiwu: nit: `in_head_stride / sequence_length` can be replaced by head_size.

tianleiwu (approved):

onnxruntime/core/providers/cuda/llm/rotary_embedding_impl.cu
```diff
@@ -0,0 +1,165 @@
...
+    return;
+  }
+
+  // Cache is (M, H/2)
+  const int half_rotary_embedding_dim = rotary_embedding_dim / 2;
+  int cache_offset;
+
+  // position_ids_format == 0 means position_ids is nullptr
+  // position_ids_format == 1 means position_ids is a 2D array of size (batch_size, sequence_length)
+  if (position_ids_format == 0) {
```
gramalingam: minor nit: may be rewriting as below may be a bit clearer?
```cpp
   b_s_index = b * sequence_length + s;
   if (position_ids_format != 0) {
      b_s_index = static_cast<int>(position_ids[b_s_index]);
   }
   cache_offset = b_s_index * half_rotary_embedding_dim;
```

onnxruntime/core/providers/cuda/llm/rotary_embedding_impl.cu
```diff
@@ -0,0 +1,165 @@
...
+    cache_offset = (b * sequence_length + s) * half_rotary_embedding_dim;
+  } else {
+    // Cache is (M, H/2) or (M, rotary_embedding_dim/2)
+    const int position_id = static_cast<int>(position_ids[b * sequence_length + s]);
```
gramalingam: Is the multiplication using sequence_length or should it be H/2 or rotary_embedding_dim/2?

onnxruntime/core/providers/cuda/llm/rotary_embedding.cc
titaiwangms: No, it's the helper I wrote for RotaryEmbedding-CPU (line 5).

onnxruntime/core/providers/cuda/llm/rotary_embedding_impl.cu
titaiwangms: Done

onnxruntime/core/providers/cuda/llm/rotary_embedding_impl.cu
titaiwangms: Done

onnxruntime/core/providers/cuda/llm/rotary_embedding_impl.cu
titaiwangms: This should be correct:

position_ids: (batch, seq_len)
caches: (batch, seq_len, H/2 or R/2)

We use b and s to represent the current index of batch and sequence length.

onnxruntime/core/providers/cuda/llm/rotary_embedding_impl.cu
titaiwangms: I flatten them into using sequence_length, num_heads, and head_size now: they are different. Could you elaborate on this comment?

titaiwangms: Added a comment saying the input is BSNH

gramalingam (approved):

titaiwangms:
Merged to catch feature complete deadline. Let me know if there ae follow-ups needed.

onnxruntime/core/providers/cuda/llm/rotary_embedding_impl.cu
gramalingam: Ok. I got confused by the comment `Cache is (M, H/2)` ... may be worth updating the comment, I guess there are two cases, the 3D and 2D case.

PR #25188, opened 2025-06-26
Avoid empty temp external initializer file creation inside the call

HectorSVC:
### Description
While EPContext model generation is enabled and some Nodes fallback on CPU. If the CPU nodes depend on external data. ORT force all external data to be embedded into new generated EPContext model by default. Ort used to create a dummy externa initializer file with maximum size threshold to force all initializer data dump into generated Onnx model file. Internally, a "./model_ext_ini.bin" file is created and got removed at the end of the call. It causes problem if multiple session doing the same thing.
This fix is to avoid creating the temp empty external initializer file by adding a flag to force all external data to be embedded into new generated EPContext model.

### DIFF
--- a/onnxruntime/core/framework/graph_partitioner.cc
+++ b/onnxruntime/core/framework/graph_partitioner.cc
@@ -920,6 +920,7 @@ static Status CreateEpContextModel(const ExecutionProviders& execution_providers
   }
 
   ModelSavingOptions model_saving_options{ini_size_threshold};
+  model_saving_options.force_embed_external_ini = force_embed_external_ini;
 
   if (saving_to_buffer) {
     ORT_RETURN_IF_ERROR(ep_context_model.MainGraph().Resolve());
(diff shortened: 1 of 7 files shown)


onnxruntime/core/framework/graph_partitioner.cc
```diff
@@ -911,10 +911,12 @@ static Status CreateEpContextModel(const ExecutionProviders& execution_providers
...
   size_t ini_size_threshold = ep_context_gen_options.output_external_initializer_size_threshold;
   std::filesystem::path external_ini_path = ep_context_gen_options.output_external_initializers_file_path;
+  bool force_embed_external_ini = false;
   if (external_ini_path.empty()) {
-    // Set the threshold to the max so all initializers are forced into the Onnx file
+    // if no external ini file specified, set force_embed_external_ini to true to avoid intermedia file creation
+    // and force all initializers embed into the Onnx file
     ini_size_threshold = SIZE_MAX;
-    external_ini_path = "./model_ext_ini.bin";
+    force_embed_external_ini = true;
```
adrianlizarraga, line 919: Would it make sense to move this flag inside of the `ModelSavingOptions` struct? The `ModelSavingOptions` struct should probably have all settings related to saving models with initializers:

https://github.com/microsoft/onnxruntime/blob/9534ab41a04bca790001e21bd05ff942f52fe098/include/onnxruntime/core/graph/model_saving_options.h#L10-L42

adrianlizarraga (approved):

PR #25191, opened 2025-06-27
[EP ABI] Add Graph_GetGraphView API to get a OrtGraph from a subset of nodes

chilo-ms:
Added an API that creates a sub-graph from a set of nodes in an OrtGraph.
This API is needed in the GetCapability EP ABI porting when EP wants to check whether a 'sub-graph' of the graph is supported by the hardware backend.


include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -5776,6 +5776,26 @@ struct OrtApi {
    */
   ORT_API2_STATUS(Graph_GetParentNode, _In_ const OrtGraph* graph, _Outptr_result_maybenull_ const OrtNode** node);
 
+  /** \brief Create a sub-graph from a subset of nodes in an OrtGraph.
+   *
+   * NOTE: A 'sub-graph' is a graph formed by a subset of nodes within the current OrtGraph.
```
jywu-mysoft: is there a way to make more clear the difference between this and the Node_GetSubgraphs api?

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -5776,6 +5776,26 @@ struct OrtApi {
...
+   * \param[in] graph The source OrtGraph instance.
+   * \param[in] nodes A subset of the nodes/OrtNodes in 'graph'.
+   * \param[in] copy_in_memory_initializer When constructing the graph, do copy the initializers from source graph to dst graph.
+   * \param[out] sub_graph An OrtGraph created from a given set of nodes. Must be released by calling ReleaseGraph.
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \since Version 1.23.
+   */
+  ORT_API2_STATUS(Graph_GetSubGraph, _In_ const OrtGraph* graph, _In_ const OrtArrayOfConstObjects* nodes, _In_ bool copy_in_memory_initializer, _Outptr_ OrtGraph** sub_graph);
```
adrianlizarraga: nit: line length exceeds 120

include/onnxruntime/core/session/onnxruntime_c_api.h
adrianlizarraga: Yeah, this can be confusing. In ONNX, a subgraph refers to a "nested subgraph" contained by control flow nodes. Maybe say this a non-nested subgraph, or alternatively, maybe we can refer to this as a "graph view".

onnxruntime/core/graph/ep_api_types.cc
```diff
@@ -426,10 +426,10 @@ void EpGraph::IndexToEpNodeMap::SetEpNode(NodeIndex node_index, EpNode* ep_node)
 EpGraph::EpGraph(const GraphViewer& graph_viewer, PrivateTag)
     : OrtGraph(OrtGraphIrApi::kEpApi), graph_viewer_(graph_viewer) {}
 
-// Static class function to create a std::unique_ptr<EpGraph>.
-Status EpGraph::Create(const GraphViewer& graph_viewer, /*out*/ std::unique_ptr<EpGraph>& result) {
-  auto ep_graph = std::make_unique<EpGraph>(graph_viewer, PrivateTag{});
+EpGraph::EpGraph(std::unique_ptr<GraphViewer> graph_viewer, std::unique_ptr<Model> model, PrivateTag)
+    : OrtGraph(OrtGraphIrApi::kEpApi), graph_viewer_(*graph_viewer.get()), model_(std::move(model)), graph_viewer_from_graph_in_model_(std::move(graph_viewer)) {}
 
+Status EpGraph::GreateImpl(std::unique_ptr<EpGraph> ep_graph, const GraphViewer& graph_viewer, /*out*/ std::unique_ptr<EpGraph>& result) {
```
adrianlizarraga: `GreatImpl` -> `CreateImpl`

onnxruntime/core/graph/ep_api_types.h
```diff
@@ -226,15 +227,38 @@ struct EpGraph : public OrtGraph {
...
+
+  /// <summary>
+  /// The real implementation of creating an EpGraph instance.
+  /// Please use one of the above 'Create' functions that internally call this function, and avoid calling this function directly.
+  /// </summary>
+  /// <param name="ep_graph"></param>
+  /// <param name="graph_viewer"></param>
+  /// <param name="result"></param>
+  /// <returns></returns>
+  static Status GreateImpl(std::unique_ptr<EpGraph> ep_graph, const GraphViewer& graph_viewer, /*out*/ std::unique_ptr<EpGraph>& result);
```
adrianlizarraga: `GreatImpl` -> `CreateImpl`. Also, should this be private?

onnxruntime/core/session/onnxruntime_c_api.cc
```diff
@@ -2737,6 +2740,98 @@ ORT_API_STATUS_IMPL(OrtApis::Graph_GetParentNode, _In_ const OrtGraph* graph, _O
...
 }
 
+ORT_API_STATUS_IMPL(OrtApis::Graph_GetSubGraph, _In_ const OrtGraph* src_graph,
+                    _In_ const OrtArrayOfConstObjects* ort_nodes_container,
+                    _In_ bool copy_in_memory_initializer,
+                    _Outptr_ OrtGraph** dst_graph) {
+  API_IMPL_BEGIN
+  const GraphViewer& graph_viewer = EpGraph::ToInternal(src_graph)->GetGraphViewer();
+
+  // This API constructs an onnxruntime::Graph from scratch using a given set of nodes,
```
adrianlizarraga: Do we have to construct a onnxruntime::Graph from scratch? Could we instead reuse the same Graph and construct a new GraphVIewer with a new filter: https://github.com/microsoft/onnxruntime/blob/7fc6235861d0b01fe6bba3e6b2f233f5797a92cf/onnxruntime/core/graph/graph_viewer.cc#L52

chilo-ms: I think a question needs to be asked here is do we want to let this API returns a subgraph that is a standalone OrtGraph instance? 

If we want it to be standalone OrtGraph, then reusing the same Graph to construct a new GraphViewer is not achieving the goal, because we need to make sure the same Graph is always around in order to access the nodes, meaning the lifetime of the source OrtGraph instance should be longer than the lifetime of the subgraph OrtGraph instance.

However, if we construct a onnxruntime::Graph from scratch and keep it inside the OrtGraph, we can make the OrtGraph instance standalone.

include/onnxruntime/core/session/onnxruntime_c_api.h
chilo-ms: Do we need to change the API's name? currently it's called `Graph_GetSubGraph`.
Can't think of better name for this. 

Updated API documentation comment, i explained the difference between 'sub-graph' and subgraph.
'sub-graph': a graph formed by a subset of nodes.
subgraph: a nested subgraph contained by control flow nodes.
Hopefully it can clear tell the difference between the two.

onnxruntime/core/graph/ep_api_types.h
chilo-ms: good catch, moved to private.

onnxruntime/core/session/onnxruntime_c_api.cc
chilo-ms: (updated)
I added a param 'to_create_standalone_sub_graph' in the API.
If it's false, the implementation creates a GraphViewer with filtered info and reuses the same Graph.
If it's true, the implementation constructs a onnxruntime::Graph from scratch.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -5706,6 +5706,38 @@ struct OrtApi {
...
+   * \param[in] copy_in_memory_initializer Only be used when 'to_create_standalone_sub_graph' is true.
+                When constructing the graph, do copy the initializers from source graph to dst graph.
+   * \param[out] dst_sub_graph An OrtGraph created from a given set of nodes. Must be released by calling ReleaseGraph.
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \since Version 1.23.
+   */
+  ORT_API2_STATUS(Graph_GetSubGraph, _In_ const OrtGraph* src_graph, _In_ const OrtNode** nodes, _In_ size_t num_nodes,
+                  _In_ bool to_create_standalone_sub_graph, _In_ bool copy_in_memory_initializer,
```
adrianlizarraga: nit: maybe remove the `to_` prefix

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -5706,6 +5706,38 @@ struct OrtApi {
...
+   * In that case, set copy_in_memory_initializer to true.
+   *
+   * \param[in] src_graph The source OrtGraph instance.
+   * \param[in] nodes A subset of the nodes/OrtNodes in 'graph'.
+   * \param[in] num_nodes Number of nodes.
+   * \param[in] to_create_standalone_sub_graph If it's false, this API constructs a sub-graph as a OrtGraph instance
+   *            that requires the 'src_graph' to be around in order to access the nodes. If it's true, this API
+   *            constructs a standalone sub-graph instead.
+   * \param[in] copy_in_memory_initializer Only be used when 'to_create_standalone_sub_graph' is true.
+                When constructing the graph, do copy the initializers from source graph to dst graph.
```
```suggestion
  *             When constructing the graph, do copy the initializers from source graph to dst graph.
```

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -5706,6 +5706,38 @@ struct OrtApi {
    */
   ORT_API2_STATUS(Graph_GetParentNode, _In_ const OrtGraph* graph, _Outptr_result_maybenull_ const OrtNode** node);
 
+  /** \brief Create a "sub-graph" from a subset of nodes in an OrtGraph.
+   *
+   * A "sub-graph" refers to a graph formed from a subset of nodes within the current OrtGraph and is not a
```
adrianlizarraga: nit: I'm not sure if differentiating `sub-graph` (with a hyphen) from `subgraph` is easy to remember. I'd be interested to hear what others think, but I would prefer one of the following:
- Call it `Graph_GetGraphView`. Still returns `OrtGraph`, but we explain it contains a subset of nodes. This is similar to the distinction between `std::string` and `std::string_view`, where the `std::string_view` can refer to a subset of characters in the original `std::string`.
- Call it `Graph_GetSubgraph` (not "SubGraph" or "sub-graph") and explain that this is not a nested control flow subgraph.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -5706,6 +5706,38 @@ struct OrtApi {
...
+   * 'to_create_standalone_sub_graph' needs to be true to use 'copy_in_memory_initializer'.
+   * Regarding how initializers should be handled when constructing a new graph, in some cases,
+   * initializers that refer to a memory location in OrtValue can not be handled by some hardware backends
+   * (unlike those that are on disk). This prevents us from sharing the data and we have to make a copy here.
+   * In that case, set copy_in_memory_initializer to true.
+   *
+   * \param[in] src_graph The source OrtGraph instance.
+   * \param[in] nodes A subset of the nodes/OrtNodes in 'graph'.
+   * \param[in] num_nodes Number of nodes.
+   * \param[in] to_create_standalone_sub_graph If it's false, this API constructs a sub-graph as a OrtGraph instance
```
adrianlizarraga: Just curious, is there currently a need in TRT EP to create a completely standalone `OrtGraph` that lives outside of the calls to `GetCapability` or `Compile`?

onnxruntime/core/graph/ep_graph_utils.cc
```diff
@@ -0,0 +1,115 @@
...
+      new_graph.AddNode(node.Name(), node.OpType(), node.Description(), inputs, outputs, &node.GetAttributes(), node.Domain());
+    }
+  }
+
+  // TODO: There is an edge case where the OrtValueInfo is outer scope and it's upper-level graph's input (not the initializer).
+
+  ORT_API_RETURN_IF_STATUS_NOT_OK(new_graph.Resolve());
+  out_model = std::move(model);
+#else
+  fprintf(stderr, "The GetSubGraphAsModelFromGraph is not supported in a minimal build.\n");
```
adrianlizarraga: Should this return an error `Status` instead of printing?

onnxruntime/core/session/onnxruntime_c_api.cc
```diff
@@ -2691,6 +2694,109 @@ ORT_API_STATUS_IMPL(OrtApis::Graph_GetParentNode, _In_ const OrtGraph* graph, _O
...
+  API_IMPL_BEGIN
+
+  if (num_nodes == 0) {
+    return OrtApis::CreateStatus(ORT_INVALID_ARGUMENT, "'num_nodes' argument should be > 0");
+  }
+
+  if (to_create_standalone_sub_graph) {
+#if !defined(ORT_MINIMAL_BUILD)
+    std::unique_ptr<Model> model;
+    ep_graph_utils::GetSubGraphAsModelFromGraph(src_graph, nodes, num_nodes, copy_in_memory_initializer, model);
```
adrianlizarraga: Should this handle the potential error status returned by `GetSubGraphAsModelFromGraph`?

onnxruntime/core/graph/ep_graph_utils.cc
```diff
@@ -0,0 +1,115 @@
...
+#include "core/graph/graph.h"
+#include "core/graph/ep_api_types.h"
+#include "core/graph/graph_utils.h"
+#include "core/framework/error_code_helper.h"
+
+#include <memory>
+
+namespace onnxruntime {
+namespace ep_graph_utils {
+OrtStatusPtr GetSubGraphAsModelFromGraph(const OrtGraph* src_graph,
```
adrianlizarraga: nit: perhaps this function should return an `onnxruntime::Status` and the caller can convert to `OrtStatus*` if necessary.

onnxruntime/core/session/onnxruntime_c_api.cc
adrianlizarraga: Good point about lifetimes. I didn't expect TRT EP to need to store a standalone `OrtGraph` outside of the call to `GetCapability` or `Compile`.

include/onnxruntime/core/session/onnxruntime_c_api.h
chilo-ms: No, TRT EP only needs the `OrtGraph` that lives within the calls to `GetCapability `and `Compile`.
Not sure about other EPs, but I assume they don't need the OrtGraph to be completely standalone either.

Even though standalone OrtGraph is not needed in current EPs, i'm thinking do we want to keep the "create_standalone_ortgraph" in the API signature for future need? or we can have a new API then?

include/onnxruntime/core/session/onnxruntime_c_api.h
chilo-ms: `Graph_GetGraphView` is better than the old one! Good idea of barrowing the concept of `std::string_view`. 
`std::string_view` is a non-owning reference to sequence of characters, so if we later decide not to provide a standalone OrtGraph, then the naming is perfect i think. If not, we can still explain the standalone usage in API comment.
Updated to use this API name.

FYI, the use of "SubGraph" in first naming "Graph_GetSubGraph" is borrowed from the existing "IndexedSubGraph".

include/onnxruntime/core/session/onnxruntime_c_api.h
adrianlizarraga: I would personally prefer to not add that feature if we don't need it yet. I think we could add a new api function if that is needed, or perhaps the model editor api would be a better fit for constructing a new model based on an existing input model. We could make that determination in the future I think.

adrianlizarraga (dismissed):

chilo-ms:
closing and reopening for license/cla CI to run.

onnxruntime/test/ep_graph/test_ep_graph.cc
```diff
@@ -474,6 +488,48 @@ static void CheckValueInfosCApi(const GraphViewer& graph_viewer, gsl::span<const
...
+  GraphViewerToProto(sub_graph_viewer, *model_proto->mutable_graph(), true, true, static_cast<ExecutionOrder>(1));
+  model_proto->set_ir_version(ONNX_NAMESPACE::Version::IR_VERSION);
+
+  const char* graph_name = nullptr;
+  ASSERT_ORTSTATUS_OK(ort_api.Graph_GetName(&api_graph, &graph_name));
+  std::string name = graph_name;
+  name += "_half.onnx";
+
+  // Dump subgraph for debugging
+  std::fstream dump(name, std::ios::out | std::ios::trunc | std::ios::binary);
```
adrianlizarraga: nit: can we comment this out or delete the generated file? Don't want to generate a ton of files in our test directory.

adrianlizarraga (approved):

PR #25192, opened 2025-06-27
[webgpu] Optimize DP4AMatMulNBitsSmallMProgram for intel

jing-bao:
### Description
This PR optimizes the Intel GPU path for the `DP4AMatMulNBitsSmallMProgram` by tuning  `tile_size` and `tile_size_k_vec`. 

### Motivation and Context
With this change, we achieved >8% performance boost on Intel iGPUs (Xe-LP and Xe2-LPG) for phi-4-mini-accuracy4 model.

### DIFF
--- a/onnxruntime/contrib_ops/webgpu/quantization/dp4a_matmul_nbits.cc
+++ b/onnxruntime/contrib_ops/webgpu/quantization/dp4a_matmul_nbits.cc
@@ -596,6 +596,11 @@ Status ApplyDP4AMatrixMatMulNBits(const Tensor* a, const Tensor* b, const Tensor
     uint32_t tile_size_k_vec = 16;
     uint32_t tile_size = 32;
 
+    if (context.AdapterInfo().vendor == std::string_view{"intel"}) {
+      tile_size_k_vec = 32;
+      tile_size = 4;
+    }
+
     DP4AMatMulNBitsSmallMProgram mul_program{tile_size_k_vec, tile_size, nbits, has_zero_points};
     uint32_t num_N_tile = (N + tile_size - 1) / tile_size;
     mul_program.SetWorkgroupSize(128);


jing-bao:
@qjia7 could you help to take a look? Thanks!

qjia7:
> @qjia7 could you help to take a look? Thanks!

How is the perf compared with the non-dp4 path for generation? Can you find a configuration that we can enable the dp4 generation path by default?

jing-bao:
> 
> How is the perf compared with the non-dp4 path for generation? Can you find a configuration that we can enable the dp4 generation path by default?

We are collecting more solid data to compare with non-dp4 path, will update here ASAP.

jing-bao:
> How is the perf compared with the non-dp4 path for generation? Can you find a configuration that we can enable the dp4 generation path by default?

Compared with the non-dp4 path, the optimized configuration has ~3% performance advantage on Xe-LP (thanks to its ~22.7% improvement over the base dp4 path), and ~1% gap remained on Xe2-LPG.

onnxruntime/contrib_ops/webgpu/quantization/dp4a_matmul_nbits.cc
```diff
@@ -596,6 +596,11 @@ Status ApplyDP4AMatrixMatMulNBits(const Tensor* a, const Tensor* b, const Tensor
     uint32_t tile_size_k_vec = 16;
     uint32_t tile_size = 32;
 
+    if (context.AdapterInfo().vendor == std::string_view{"intel"}) {
+      tile_size_k_vec = 32;
+      tile_size = 4;
```
qjia7, line 601: To use a small tile_size, it means that there will be more dispatched workgroups. More workgroups mean more parallelism. But in another side, when the dispatched workgroups exceed a threshold, the perf will start to degrade due to the limited hardware EUs. I have a little concern for the last matmulnbis with output shape [1,1,200064], which will dispatch 200064 / 4 = 50016 workgroups with this changes. Maybe we also need to consider the shape to choose the suitable tile_size.
For xe-lp, do you also test the desktop gpu? I remember desktop only has 32 EUs but the laptop has 96 or 128 or higher.  If both desktop and laptop work well with this change, it looks good for me. Thanks.
BTW, your test model is f32 or f16?

qjia7: Offline talked about this. All collected data is good on various devices. LGTM, thanks.

qjia7 (approved):

daijh:
@fs-eire
The CI failure appears to be unrelated to this PR. 
Could you please take a look?

fs-eire:
rerun completed successfully

PR #22646, opened 2024-10-29
Refactor the cmake code that is related to delay loading

snnn:
### Description
Refactor the cmake code that is related to delay loading. Provide a cmake option to control if delay loading should be enabled or not.
Disabling the option when python is enabled, due to a bug we are investigating. 

### Motivation and Context
ONNX Runtime's python package depends on DirectML.dll, but supposedly the DLL should be delay loaded.
This PR only refactor the code. It doesn't change the behavior.

PR #25200, opened 2025-06-27
add back linker flags "-z noexecstack" that were lost by accident

arnej27959:
### Description

Add back the linker option to make stack non-executable, which was accidentally lost here:
https://github.com/microsoft/onnxruntime/pull/22646

This just adds back the option in the same place where it was.

### Motivation and Context
After upgrading to 1.22.0 we saw this warning:

OpenJDK 64-Bit Server VM warning: You have loaded library /opt/vespa-deps/lib64/libonnxruntime.so.1.22.0 which might have disabled stack guard. The VM will try to fix the stack guard now.
It's highly recommended that you fix the library with 'execstack -c <libfile>', or link it with '-z noexecstack'.


Craigacp:
Ah, I'd wondered why that warning had reappeared but didn't have time to run it down yet.

PR #25202, opened 2025-06-27
Fix TRT-EP build for EP graph tests

kevinch-nv:
### Description
New binary `onnxruntime_ep_graph_test` uses `test_main.cc`, which contains deprecated declarations from `NvInfer.h`. Ignore these warnings when building TRT EP.

### Motivation and Context
Fixes build for TRT EP.


kevinch-nv:
@chilo-ms for viz.

chilo-ms (approved):

PR #25214, opened 2025-06-29
GatherBlockQuantized supports zero points and 8 bits for uint8 dtype

tianleiwu:
Add support for unit8 GatherBlockQuantized for the following two areas:
* Allow zero points.
* Add bits attribute and support bits=8.

Major change is to update shape inference; and update unit tests to cover these.

Note that only CPU implementation, and CUDA implementation will be added later in another PR.

### Motivation and Context

Previously, zero points are not supported when dtype is uint8. Only 4 bit quantization without zero points were supported.
This change is to  share weights of lm_head with 8 bit quantization between GatherBlockQuantized and MatMulNBits.
 
For example, when K is multiple of `block_size`, typical input and output shapes are like the following:
 * data has shape (N, K) for 8 bits, or (N, K / 2) for 4 bits.
 * scales has shape (N, k_blocks), where k_blocks = (K / block_size).
 * zero_points has shape (N, k_blocks) for 8 bits, (N, (k_blocks + 1) / 2) for 4 bits.
 * output will have shape (..., K), where ... is the shape of `indices`.

### DIFF
--- a/onnxruntime/contrib_ops/cpu/quantization/gather_block_quantized.cc
+++ b/onnxruntime/contrib_ops/cpu/quantization/gather_block_quantized.cc
@@ -18,17 +18,18 @@ namespace contrib {
 
 namespace {
 template <typename T1>
-int32_t GetDataElement(const T1* data_ptr, int64_t data_idx) {
+int32_t Get4BitElement(const T1* data_ptr, int64_t data_idx) {
   return static_cast<int32_t>(data_ptr[data_idx >> 1].GetElem(narrow<size_t>(data_idx & 1)));
 }
 
 template <>
-int32_t GetDataElement<uint8_t>(const uint8_t* data_ptr, int64_t data_idx) {
+int32_t Get4BitElement<uint8_t>(const uint8_t* data_ptr, int64_t data_idx) {
   const uint8_t data_val_u8 = data_ptr[data_idx >> 1];
   // Weights are stored as (nibble2)(nibble1) in uint8_t.
   auto data_val = static_cast<int32_t>((data_idx & 1) ? ((data_val_u8 >> 4) & 0x0F) : (data_val_u8 & 0x0F));
   return data_val;
 }
+
 }  // namespace
 
 template <typename T1, typename Tind>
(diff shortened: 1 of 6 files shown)


onnxruntime/contrib_ops/cpu/quantization/gather_block_quantized.cc
```diff
@@ -186,21 +218,44 @@ Status GatherBlockQuantized<T1, Tind>::CopyDataAndDequantize(const T1* data_ptr,
     int64_t output_idx = output_idx_base;
     int64_t data_idx = data_idx_base;
     for (int64_t i = 0; i < gather_block; ++i, ++output_idx, ++data_idx) {
-      auto data_val = GetDataElement(data_ptr, data_idx);
+      int32_t data_val;
```
jiafatom, line 221: Shall we initialize this variable?

onnxruntime/contrib_ops/cpu/quantization/gather_block_quantized.cc
```diff
@@ -232,7 +287,7 @@ template <typename T1, typename Tind>
 Status GatherBlockQuantized<T1, Tind>::Compute(OpKernelContext* context) const {
   Prepare p;
   ORT_RETURN_IF_ERROR(PrepareForCompute(context, p));
-  auto components = (std::is_same_v<T1, uint8_t>) ? 2 : 1;
+  int64_t components = std::is_same_v<T1, uint8_t> ? (8 / static_cast<int>(bits_)) : 1;
```
jiafatom, line 290: Do we need `static_cast<int>` since `bits_` is already `int64_t`

tianleiwu: It fixes a compiler warning.  8 is int, bits is int64_t, the warning is that need to match type for /

onnxruntime/contrib_ops/cpu/quantization/gather_block_quantized.cc
tianleiwu: It is fine. Every path in lines 222-230 below will set a value.

jiafatom (approved):

kunal-vaishnavi (approved):

PR #25225, opened 2025-06-30
Add telemetry error logging to InferenceSession

ashrit-ms:
### Description
Add telemetry error logging to InferenceSession::Run() methods to track runtime errors that occur during inference execution.

### Motivation and Context
Currently, we do not have any telemetry error logging in the InferenceSession class to track runtime errors that occur during inference execution. This data would allow us to prioritize and identify the most frequent errors.

### DIFF
--- a/onnxruntime/core/session/inference_session.cc
+++ b/onnxruntime/core/session/inference_session.cc
@@ -2522,6 +2522,12 @@ common::Status InferenceSession::Initialize() {
         status = end_status;
       }
     }
+  } else {
+    // Log runtime error telemetry if the return value is not OK
+    // NOTE: The error line number here is not the line number of the error in the code,
+    // but the line number of this function call in the code. The purpose of this telemetry is to log the runtime error
+    // code that occurred during the inference session initialize, not the specific line of code that caused the error.
+    env.GetTelemetryProvider().LogRuntimeError(session_id_, status, __FILE__, __FUNCTION__, __LINE__);
   }
 
   return status;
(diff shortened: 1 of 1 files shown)


onnxruntime/core/session/inference_session.cc
```diff
@@ -2522,6 +2522,12 @@ common::Status InferenceSession::Initialize() {
         status = end_status;
       }
     }
+  } else {
+    // Log runtime error telemetry if the return value is not OK
+    // NOTE: The error line number here is not the line number of the error in the code,
+    // but the line number of this function call in the code. The purpose of this telemetry is to log the runtime error
+    // code that occurred during the inference session initialize, not the specific line of code that caused the error.
+    env.GetTelemetryProvider().LogRuntimeError(session_id_, status, __FILE__, __FUNCTION__, __LINE__);
```
snnn: It should have already been logged. Check common.h

ashrit-ms: Thanks, fixed.

onnxruntime/core/session/inference_session.cc
```diff
@@ -3127,6 +3127,15 @@ Status InferenceSession::Run(const RunOptions& run_options,
     LOGS(*session_logger_, INFO) << "Start another run for necessary memory allocation or graph capture.";
     ORT_RETURN_IF_ERROR(Run(run_options, feed_names, feeds, output_names, p_fetches, p_fetches_device_info));
   }
+
+  // Log runtime error telemetry if the return value is not OK
```
snnn, line 3131: Can we use the ORT_RETURN_IF_ERROR_SESSIONID macro that is defined in [https://github.com/microsoft/onnxruntime/blob/main/include/onnxruntime/core/common/common.h](https://github.com/microsoft/onnxruntime/blob/main/include/onnxruntime/core/common/common.h) ?

ashrit-ms: Fixed.

snnn (approved):

Issue #25145, opened 2025-06-24
Fallback provider logic bug can lose GPU acceleration when creating inference session


exitflynn:
### Describe the issue

I seem to have stumbled upon a bug in the `_create_inference_session` method that causes execution provider fallback to be incorrectly overwritten, defaulting to CPU even when GPU acceleration is available. Additionally there also seems to be a slight typo in the code which can affect using the `NvTensorRTRTXExecutionProvider` provider when passed as a tuple.

https://github.com/microsoft/onnxruntime/blob/986b66af96252488bcf885741623ba877964baca/onnxruntime/python/onnxruntime_inference_collection.py#L488-L537

here I'm referring to this structure:
```python
if "TensorrtExecutionProvider" in available_providers:
    # Sets fallback to ["CUDAExecutionProvider", "CPUExecutionProvider"] when both TensorRT and CUDA requested
    
if "NvTensorRTRTXExecutionProvider" in available_providers:  # Should be elif!
    ...
    
elif "MIGraphXExecutionProvider" in available_providers: 
    ...
else:
    self._fallback_providers = ["CPUExecutionProvider"]
```
where any run of the if-elif-else statements can overwrite the fallback providers set where TensorRT is available. 

Also what appears to be the typo I mentioned earlier:
https://github.com/microsoft/onnxruntime/blob/986b66af96252488bcf885741623ba877964baca/onnxruntime/python/onnxruntime_inference_collection.py#L517-L520

I'll be happy to contribute a fix if this is confirmed!

### To reproduce

```python
from onnxruntime import InferenceSession, get_available_providers

print("Available providers:", get_available_providers())
providers = ["TensorrtExecutionProvider", "CUDAExecutionProvider"]
session = InferenceSession("model.onnx", providers=providers)
print(f"Actual fallback: {session._fallback_providers}")
```

(where available providers: `['TensorrtExecutionProvider', 'CUDAExecutionProvider', 'CPUExecutionProvider']`)

When trying to reproduce locally on my laptop I hit the scenario where the TensorRT files were not installed and I was wondering if this scenario could be a valuable one in itself since even now I should be able to fall back to the available CUDA instead of losing GPU acceleration entirely? So I've attached outputs from that run itself.

- Expected behavior: When TensorRT fails but CUDA is explicitly requested and available, should fall back to: `["CUDAExecutionProvider", "CPUExecutionProvider"]`

- Actual behavior: Falls back to: `["CPUExecutionProvider"]`

Output:
[...]

PR #25226, opened 2025-06-30
Inference fallback provider bugfix

exitflynn:
### Description

Bugfix(es) in the Inference Session creation flow which otherwise wouldn't set CUDA execution providers for users having TRT or NvRTX.

### Motivation and Context

the `_create_inference_session` method is unable to set CUDA execution providers even when they're available due to a minor bug in the control flow (`if` statement instead of `elif`) (for TRT) and a typo (for NvRTX). Further context in #25145

### DIFF
--- a/onnxruntime/python/onnxruntime_inference_collection.py
+++ b/onnxruntime/python/onnxruntime_inference_collection.py
@@ -516,7 +516,7 @@ class InferenceSession(Session):
                 )
                 and any(
                     provider == "NvTensorRTRTXExecutionProvider"
-                    or (isinstance(provider, tuple) and provider[0] == "NvExecutionProvider")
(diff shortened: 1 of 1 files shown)


xadupre (approved):

exitflynn:
hey @xadupre are the failing tests an issue i should look into or expected at this point?

robertsonwang:
FYI I ran into the same issue with `TensorrtExecutionProvider`. Specifically, if `TensorrtExecutionProvider` fails to initialize then onnxruntime will fall back to CPU support and ignore `CUDAExecutionProvider`. Unfortunately, that fall back happens during session initialization (in C++) and not during session run time. Therefore, this sort of fix will not work in the updated code base.

PR #25235, opened 2025-07-01
[webgpu] extend cast version to 23

xhcao:


xhcao:
webnn needs Cast-21 operator.
The changes of `Cast` operator from version 19 to version 23 only add three types, which are tensor(float4e2m1) tensor(int4) tensor(uint4), so could simply extend to version 23.

guschmue (approved):

PR #25238, opened 2025-07-01
[ARM CPU] SVE support for Elementwise kernels

sanketkaleoss:
### Description

Ports the `MlasErfKernel`, `MlasLogisticKernel` and `MlasComputeSoftmax` kernels to the ARM SVE backend. Specifically, the following functions have been ported.
- `MlasErfKernel` (lib/erf.cpp)
- `MlasLogisticKernel` (lib/logistic.cpp)
- `MlasComputeSumExpF32Kernel` (lib/compute.cpp)
- `MlasReduceMaximumF32Kernel` (lib/compute.cpp)
- `MlasComputeSoftmaxOutputF32Kernel` (lib/compute.cpp)
- `MlasComputeSoftmaxThreaded` (lib/compute.cpp)

This PR uses the following design structure: adds new wrapper implementations of SVE functions in `lib/mlasi_sve.h` similar to `mlasi.h` and calls these wrapper functions in each kernel's implementation.

### Motivation and Context

This work is a step toward making ONNX Runtime more performant and architecture-aware on ARM platforms.

### Performance Analysis
image
- Observed upto 1.4x speedup at the operator level
- Performance is tested on AWS Graviton3E

This PR is a joint contribution by:
- @NishantPrabhuFujitsu
- @sanketkaleoss


sanketkaleoss:
> @sanketkaleoss please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> ```
> 
> > Options:
> > 
> > 
> > ```
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > 
> > ```
> > ```
> 

@microsoft-github-policy-service agree company=“Fujitsu Research of India Private Ltd”

sanketkaleoss:
Hi @edgchen1 , @hariharans29 , @snnn Kindly enable the CI pipeline.

sanketkaleoss:
Hi @yufenglee Please trigger the CI pipeline. Thanks

cmake/onnxruntime_mlas.cmake
```diff
@@ -5,13 +5,29 @@ set(MLAS_ROOT ${ONNXRUNTIME_ROOT}/core/mlas)
...
+# Detect SVE support in AArch64 (ARM 64-bit)
+if(CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64")
+    message(STATUS "Detecting SVE support for AArch64...")
+    # Include module to check compiler flags
+    include(CheckCXXCompilerFlag)
+    # Test if the compiler supports SVE
+    check_cxx_compiler_flag("-march=armv8.2-a+sve+fp16" COMPILER_SUPPORTS_SVE)
+    if (COMPILER_SUPPORTS_SVE)
+        message(STATUS "Compiler supports SVE!")
+        add_compile_definitions(COMPILER_SUPPORTS_SVE=1) # Define COMPILER_SUPPORTS_SVE for source files
```
hariharans29: `target_compile_defintions` instead of `add_compile_definitions` ? 
https://github.com/microsoft/onnxruntime/blob/main/docs/cmake_guideline.md

onnxruntime/core/common/cpuid_info.cc
```diff
@@ -188,6 +188,7 @@ void CPUIDInfo::ArmLinuxInit() {
     has_arm_neon_dot_ = cpuinfo_has_arm_neon_dot();
     has_fp16_ = cpuinfo_has_arm_neon_fp16_arith();
     has_arm_neon_i8mm_ = cpuinfo_has_arm_i8mm();
+    has_arm_sve_ = cpuinfo_has_arm_sve();
```
hariharans29, line 197: I don't see this in `ArmAppleInit()` -> is it intentional to limit these to just Linux/Windows ?

onnxruntime/core/mlas/lib/mlasi.h
```diff
@@ -50,6 +50,10 @@ Module Name:
 #if defined(__arm__) || defined(__aarch64__)
 #include <arm_neon.h>
 #endif
+#ifdef COMPILER_SUPPORTS_SVE
```
hariharans29: We don't need these changes since they are in mlasi_sve.h right ?

onnxruntime/core/mlas/lib/logistic.cpp
```diff
@@ -58,6 +59,92 @@ MLAS_INTERNAL_DATA const struct {
     0.5f,
 };
 
+#ifdef __ARM_FEATURE_SVE
+void 
+MLASCALL
+MlasSveLogisticKernel(
```
hariharans29: Can we create a top level folder in `mlas/lib` called `sve` and place the SVE specific implementations there and reference them here through `mlasi_sve.h` ?

onnxruntime/core/mlas/lib/logistic.cpp
```diff
@@ -184,6 +271,10 @@ Return Value:
 #if defined(MLAS_TARGET_AMD64)
     GetMlasPlatform().LogisticKernelRoutine(Input, Output, N);
 #else
-    MlasLogisticKernel(Input, Output, N);
+    #ifdef __ARM_FEATURE_SVE
```
hariharans29: Newbie question - So this is set in the arm_sve header if the compiler supports SVE ? Don't we need to check the running processor supports SVE through `HasArmSVE()` before invoking the SVE routine ?

onnxruntime/core/mlas/lib/mlasi.h
```diff
@@ -50,6 +50,10 @@ Module Name:
...
+#include <arm_sve.h>  // SVE intrinsic header
+#define MLAS_SVE_INTRINSICS 1
```
hariharans29, line 55: Is this used somewhere ?

onnxruntime/core/mlas/lib/mlasi_sve.h
```diff
@@ -0,0 +1,549 @@
...
+    This module contains the procedure prototypes for the SVE intrinsics.
+
+--*/
+
+#pragma once
+
+#include "mlasi.h"
+
+
+#ifdef COMPILER_SUPPORTS_SVE
```
hariharans29: Nit: Suggest renaming `COMPILER_SUPPORTS_SVE` to `USE_SVE` to be consistent with other options in the project

onnxruntime/core/mlas/lib/mlasi_sve.h
```diff
@@ -0,0 +1,549 @@
...
+#include <arm_sve.h>  // SVE intrinsic header
+#define MLAS_SVE_INTRINSICS 1
```
hariharans29: Do we need this macro - `MLAS_SVE_INTRINSICS ` ?

cmake/onnxruntime_mlas.cmake
sanketkaleoss: Will update this, thanks

onnxruntime/core/common/cpuid_info.cc
sanketkaleoss: This PR is intended to support ArmLinux mainly. Development and testing is done on ArmLinux platform.

onnxruntime/core/mlas/lib/mlasi.h
sanketkaleoss: Thanks for pointing this out, will remove this part.

onnxruntime/core/mlas/lib/logistic.cpp
sanketkaleoss: Yes, that sounds like a more appropriate approach. I will update the PR to follow this structure.

onnxruntime/core/mlas/lib/logistic.cpp
sanketkaleoss: The runtime check like HasArmSVE() causes CI failures on systems like macOS (or x86-based runners), because the code still needs to compile even if it won't run. That's why it's essential to guard SVE-specific code at compile time using `#ifdef __ARM_FEATURE_SVE`, so it’s excluded entirely on unsupported platforms. This prevents build errors in CI environments.

onnxruntime/core/mlas/lib/mlasi_sve.h
sanketkaleoss: Will update, thanks

onnxruntime/core/mlas/lib/mlasi_sve.h
sanketkaleoss: Since it is not used elsewhere, it can be safely removed.

onnxruntime/core/mlas/lib/logistic.cpp
hariharans29: Can you please elaborate on the exception you see while running `HasArmSVE()` ? I am not very familiar with the code but it should just check if the processor supports SVE or not and that in itself should not error out (unless there is a bug there that needs to be ironed out).

The issue with the pure compile time check (and not runtime check) is if the MLAS library is built on an SVE supporting platform and run on a platform where it isn't supported. Without a runtime check, wouldn't that cause an issue as well ?

hariharans29:
The ARM Linux CI is failing on some tests

onnxruntime/core/mlas/lib/logistic.cpp
sanketkaleoss: I think the HasArmSVE() function calls svcntw() kind of instructions, which perform a runtime check for the presence of SVE. Although this check may never actually execute on non-SVE platforms, the compiler still emits these instructions, potentially causing illegal instruction exceptions at load time or runtime.

sanketkaleoss:
> The ARM Linux CI is failing on some tests

I’m currently working on fixing the errors. One of them is caused by the accuracy issue in the SVE implementation of the exp() function, and the other originates from the sigmoid function.

The issue in the sigmoid function was also present in the NEON implementation prior to the last commit, but it was resolved by applying a Clamp. I’m replicating the same fix in the SVE version.

cmake/onnxruntime_mlas.cmake
```diff
@@ -50,6 +51,21 @@ onnxruntime_add_static_library(onnxruntime_mlas
   ${MLAS_SRC_DIR}/saturation_check.cpp
 )
 
+# Detect SVE support in AArch64 (ARM 64-bit)
+if(CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64")
```
hariharans29: Could we have a build option to disable building with SVE support if needed ? (You can follow the KleidiAI template - https://github.com/microsoft/onnxruntime/blob/2678b9ca3fa29b1a418f5e7d22647d27896fb5e5/tools/ci_build/build.py#L894C17-L894C28)

onnxruntime/core/mlas/lib/mlasi.h
```diff
@@ -202,6 +207,7 @@ class MLASCPUIDInfo
     bool has_arm_neon_dot_{false};
     bool has_fp16_{false};
     bool has_arm_neon_i8mm_{false};
+     bool has_arm_sve_{false};
```
hariharans29: Nit: Formatting

onnxruntime/core/mlas/lib/logistic.cpp
hariharans29: `That's why it's essential to guard SVE-specific code at compile time using #ifdef __ARM_FEATURE_SVE, so it’s excluded entirely on unsupported platforms` -> But what happens if we build on platforms that support SVE and deploy on non-SVE supported platforms ? We will run into similar issues right ? 

Is there anyway we could improve `HasArmSVE()` so that it doesn't crash on non-SVE supported platforms ? CC: @edgchen1

onnxruntime/core/mlas/lib/sve/elementwise_sve.cpp
```diff
@@ -0,0 +1,671 @@
+#ifdef __ARM_FEATURE_SVE
```
hariharans29: Please add copyright ?

sanketkaleoss: Thanks, will add

cmake/onnxruntime_mlas.cmake
sanketkaleoss: Thanks for the suggestion, will update the code

onnxruntime/core/common/cpuid_info.cc
```diff
@@ -319,6 +321,7 @@ void CPUIDInfo::ArmWindowsInit() {
     // has_arm_neon_i8mm_ = cpuinfo_has_arm_i8mm();
     has_arm_sve_i8mm_ = cpuinfo_has_arm_sve() && has_arm_neon_i8mm_;
     has_arm_neon_bf16_ = cpuinfo_has_arm_neon_bf16();
+    has_arm_sve_ = cpuinfo_has_arm_sve();
```
edgchen1: has `cpuinfo_has_arm_sve()` been tested on Windows? I don't see `cpuinfo_isa.sve` being set here.

https://github.com/pytorch/cpuinfo/blob/8a9210069b5a37dd89ed118a783945502a30a4ae/src/arm/windows/init.c#L192-L222

cmake/CMakeLists.txt
```diff
@@ -83,6 +83,24 @@ option(onnxruntime_USE_SNPE "Build with SNPE support" OFF)
...
 option(onnxruntime_USE_DNNL "Build with DNNL support" OFF)
 option(onnxruntime_USE_JSEP "Build with JavaScript implemented kernels support" OFF)
+option(onnxruntime_USE_SVE "Build with SVE support in MLAS" OFF)
+# disable explicitly on unsupported Apple x86_64 targets
+if(APPLE AND CMAKE_OSX_ARCHITECTURES MATCHES "x86_64")
+    message(WARNING "Disabling SVE: not supported on Apple x86_64 platforms")
+    set(onnxruntime_USE_SVE OFF CACHE BOOL "" FORCE)
+endif()
+
+# Detect SVE support in AArch64 (ARM 64-bit)
```
edgchen1: this is between other `option` declarations. please move it further down.

e.g., we have other similar checks here:
https://github.com/sanketkaleoss/onnxruntime/blob/3dbe7b48c8e413a76813b53d453751b43dbde563/cmake/CMakeLists.txt#L618-L625

cmake/CMakeLists.txt
```diff
@@ -83,6 +83,24 @@ option(onnxruntime_USE_SNPE "Build with SNPE support" OFF)
...
+if(onnxruntime_USE_SVE)
+  if(CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64")
+      message(STATUS "Compiler supports SVE!")
```
edgchen1: it seems that this message is printed a little early, before the check actually happens

cmake/CMakeLists.txt
```diff
@@ -83,6 +83,24 @@ option(onnxruntime_USE_SNPE "Build with SNPE support" OFF)
...
+      # Include module to check compiler flags
+      include(CheckCXXCompilerFlag)
+      # Test if the compiler supports SVE
+      check_cxx_compiler_flag("-march=armv8.2-a+sve+fp16" USE_SVE)
```
edgchen1: nit: can we rename `USE_SVE` to `HAS_ARM64_SVE` for consistency with the other compiler flag check variables?

also, is this line checking for both sve and fp16 features? could we reuse the fp16 flag here?
https://github.com/sanketkaleoss/onnxruntime/blob/3dbe7b48c8e413a76813b53d453751b43dbde563/cmake/CMakeLists.txt#L622

cmake/CMakeLists.txt
```diff
@@ -1027,6 +1045,10 @@ function(onnxruntime_set_compile_flags target_name)
       target_compile_definitions(${target_name} PRIVATE USE_KLEIDIAI)
     endif()
 
+    if (onnxruntime_USE_SVE)
```
edgchen1: should it also check the variable that indicates whether the compiler supports SVE?

onnxruntime/core/common/cpuid_info.cc
edgchen1: I think we should either:
- only enable it on Linux, where it has been tested, and add a comment about why it is not enabled on other platforms yet.
- also enable it in ArmWindowsInit() and ArmAppleInit().

onnxruntime/core/mlas/lib/erf.cpp
```diff
@@ -22,7 +22,10 @@ Module Name:
 --*/
 
 #include "mlasi.h"
-
+#include <iostream>
```
edgchen1: do we need `<iostream>`?

onnxruntime/core/mlas/lib/compute.cpp
```diff
@@ -20,7 +20,11 @@ Module Name:
 --*/
 
 #include "mlasi.h"
+#ifdef USE_SVE
+#include "sve/mlasi_sve.h"
```
edgchen1: can we make use of the function pointers in `MLAS_PLATFORM` to limit the platform-specific code to fewer files?

onnxruntime/core/mlas/lib/logistic.cpp
edgchen1: `HasArmSVE()` should work on any platform. if it doesn't, we should fix it. but it is just returning a boolean that was set earlier in the CPUIDInfo ctor.

if we compile code conditionally here, we should ensure that it compiles code that can be run. possibly a simpler alternative is to dispatch with the existing function pointer mechanism in MLAS_PLATFORM, where we would only assign the SVE implementation to the function pointer if we know that it can run, instead of directly calling the SVE implementation here.

onnxruntime/core/mlas/lib/compute.cpp
sanketkaleoss: @hariharans29 @edgchen1 Updated the code to incorporate the suggested changes

onnxruntime/core/mlas/lib/erf.cpp
sanketkaleoss: removed

cmake/CMakeLists.txt
sanketkaleoss: Updated

cmake/CMakeLists.txt
sanketkaleoss: Thanks, updated the flag

onnxruntime/core/common/cpuid_info.cc
sanketkaleoss: Removed it entirely as it's not required for this PR

cmake/CMakeLists.txt
```diff
@@ -612,6 +619,23 @@ else()
...
+  if(onnxruntime_USE_SVE)
+    if(CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64")
+      execute_process(
+          COMMAND lscpu
+          OUTPUT_VARIABLE LSCPU_OUTPUT
+          OUTPUT_STRIP_TRAILING_WHITESPACE
+      )
+      if(LSCPU_OUTPUT MATCHES "sve")
+          message(STATUS "Host CPU supports SVE!")
+          set(HAS_ARM64_SVE "ON")
```
edgchen1: I think we want to determine whether the host can compile SVE code. This is different from whether the host CPU actually supports SVE.

How about this?
`check_cxx_compiler_flag(-march=armv8.2-a+sve HAS_ARM64_SVE)`

cmake/onnxruntime_mlas.cmake
```diff
@@ -860,4 +868,4 @@ if (NOT onnxruntime_ORT_MINIMAL_BUILD)
     endif()
   endif()
 
-endif()
+endif()
```
edgchen1: general nit: please keep the newlines at the ends of the files

onnxruntime/core/mlas/lib/platform.cpp
```diff
@@ -592,6 +594,15 @@ Return Value:
     }
 #endif
 
+#if defined(USE_SVE)
+    this->ErfKernelRoutine = MlasSveErfKernel;
+    this->LogisticKernelRoutine = MlasLogisticKernel;
```
edgchen1: should this be set to `MlasSveLogisticKernel`?

tools/ci_build/build_args.py
```diff
@@ -731,6 +731,9 @@ def add_execution_provider_args(parser: argparse.ArgumentParser) -> None:
     acl_group.add_argument(
         "--no_kleidiai", action="store_true", help="Disable KleidiAI integration (used with ACL/ArmNN)."
     )
+    acl_group.add_argument(
```
edgchen1: hm, I am not so sure about this inclusion in `acl_group`. it is not related to the ACL EP, right? I wonder if it makes sense to add a group for the CPU EP?

tools/ci_build/build.py
```diff
@@ -898,6 +898,10 @@ def generate_build_tree(
         ):
             cmake_args += ["-Donnxruntime_USE_KLEIDIAI=ON"]
 
+    if not args.no_sve:
+        if "aarch64" in platform.machine().lower() and not args.build_wasm:
```
edgchen1: can we consolidate the checks in one place? now they are both here and in the CMake code.

https://github.com/sanketkaleoss/onnxruntime/blob/e73a676b0566a36a5224ad009b487f569adb9981/cmake/CMakeLists.txt#L92-L95

sanketkaleoss: Updated and put the conditions in `build.py` and removed unnecessary conditional checks from `CMakeLists.txt`

tools/ci_build/build_args.py
sanketkaleoss: Thanks, added the suggested fix

onnxruntime/core/mlas/lib/platform.cpp
sanketkaleoss: thanks, fixed

cmake/CMakeLists.txt
sanketkaleoss: Thanks @edgchen1 for the suggestion. Initially, I had applied the fix you’re suggesting with cxx, but it resulted in CI failures on non-SVE ARM machines (e.g., Graviton2) since the code is compiled for `armv8.2` without `+sve`, leading to illegal instruction errors at runtime.

However, by using `lscpu`, we can identify whether the system supports SVE and then enable the flag accordingly.

edgchen1: The runtime check is separate from the compilation check. The binary might be run on another machine than the one it was compiled on, where SVE support may be different. Running `lscpu` here only tells us about the host machine.

I think that we need both checks. The runtime check would be appropriate for determining whether to assign the SVE kernels to the function pointers when initializing `MLAS_PLATFORM`.

sanketkaleoss: @edgchen1 I have updated the code by adding both runtime and compile time checks. Please start the CI pipeline if the code fix is okay.

cmake/CMakeLists.txt
```diff
@@ -612,6 +614,19 @@ else()
      message(FATAL_ERROR  "The compiler doesn't support FLOAT16!!!")
    endif()
   endif()
+
+  if(onnxruntime_USE_SVE)
+    if(CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64")
+      include(CheckCXXCompilerFlag)
```
edgchen1: nit: this is already included here:
https://github.com/microsoft/onnxruntime/blob/ec3bf7f03d9363ebf5c6c952a7f017fc42d7417f/cmake/CMakeLists.txt#L29

tools/ci_build/build.py
edgchen1: thanks for putting them in one place.

I think that the CMake code may be a more suitable location for these checks. e.g., we can also determine whether the compiler supports it at that point.

sanketkaleoss: @edgchen1 Updated the code, let me know if the fix is okay

onnxruntime/core/mlas/lib/sve/mlasi_sve.h
```diff
@@ -0,0 +1,655 @@
...
+{
+    return svzip2_f32(Vector1, Vector2);
+}
+
+MLAS_SVE_TARGET
+MLAS_FORCEINLINE
+MLAS_SVFLOAT32
+MlasSveAddFloat32(MLAS_SVBOOL Pred, MLAS_SVFLOAT32 Vector1, MLAS_SVFLOAT32 Vector2)
+{
+    return svadd_f32_m(Pred, Vector1, Vector2); //_m
```
edgchen1: what is the comment for?
`//_m`

onnxruntime/core/mlas/lib/sve/mlasi_sve.h
```diff
@@ -0,0 +1,655 @@
...
+{
+   return svrintm_f32_z(Pred, Vector);
+}
+
+MLAS_SVE_TARGET
+MLAS_FORCEINLINE
+MLAS_SVFLOAT32
+MlasSveMultiplyAddFloat32(MLAS_SVBOOL Pred, MLAS_SVFLOAT32 Vector1, MLAS_SVFLOAT32 Vector2, MLAS_SVFLOAT32 Vector3)
+{
+    // return svmla_f32_m(Pred, Vector1, Vector2, Vector3);
```
edgchen1: why keep this commented out code?

cmake/CMakeLists.txt
```diff
@@ -1048,6 +1062,10 @@ function(onnxruntime_set_compile_flags target_name)
       target_compile_definitions(${target_name} PRIVATE USE_KLEIDIAI)
     endif()
 
+    if (HAS_ARM64_SVE)
+      target_compile_definitions(${target_name} PRIVATE USE_SVE)
```
edgchen1: is it possible to limit this to the onnxruntime_mlas target?

onnxruntime/core/mlas/lib/sve/mlasi_sve.h
sanketkaleoss: Not necessary, will remove it

onnxruntime/core/mlas/lib/sve/mlasi_sve.h
sanketkaleoss: Not necessary, will remove

cmake/CMakeLists.txt
sanketkaleoss: Incorporated into latest commit

onnxruntime/core/mlas/lib/sve/elementwise_sve.cpp
```diff
@@ -0,0 +1,685 @@
...
+Module Name:
+
+    elementwise_sve.cpp
+
+Abstract:
+
+    This module contains the implementation of SVE-based elementwise operations
+--*/
+
+#ifdef __ARM_FEATURE_SVE
```
hariharans29: Is this guard still needed with all the cmake changes that guard the compilation of the SVE files ?

onnxruntime/core/common/cpuid_info.cc
hariharans29: Should we leave behind the comment about why this is enabled on Linux only ? @sanketkaleoss ?

onnxruntime/core/mlas/lib/compute.cpp
```diff
@@ -834,7 +833,7 @@ Return Value:
 --*/
 {
     const auto* WorkBlock = (MLAS_SOFTMAX_WORK_BLOCK<float>*)Context;
-
+
```
hariharans29, line 835: Nit: Formating ?

onnxruntime/core/mlas/lib/mlasi.h
```diff
@@ -50,6 +50,9 @@ Module Name:
 #if defined(__arm__) || defined(__aarch64__)
 #include <arm_neon.h>
 #endif
+#ifdef USE_SVE
+#include <arm_sve.h>  // SVE intrinsic header
```
hariharans29: Is this header inclusion in this fie necessary given that we have an mlasi_sve.h now ?

onnxruntime/core/mlas/lib/compute.cpp
```diff
@@ -21,6 +21,7 @@ Module Name:
 
 #include "mlasi.h"
 #include "softmax.h"
+#include <cmath>
```
hariharans29: Nit: Is this inclusion necessary ?

sanketkaleoss: Not needed, removed

onnxruntime/core/mlas/lib/mlasi.h
sanketkaleoss: Not needed now, removed

onnxruntime/core/mlas/lib/sve/elementwise_sve.cpp
sanketkaleoss: Not needed, works even without the guard now

onnxruntime/core/common/cpuid_info.cc
hariharans29: Ping on this

onnxruntime/core/mlas/lib/compute.cpp
hariharans29: Ping on this. 

There are some other minor formatting nits: some whitespaces in a couple of files. If these can be addressed, that will be great. Thanks.

onnxruntime/core/mlas/lib/erf.cpp
```diff
@@ -261,9 +260,9 @@ Return Value:
 
 --*/
 {
-#if defined(MLAS_TARGET_AMD64)
+#if defined(MLAS_TARGET_AMD64) || defined(USE_SVE)
     GetMlasPlatform().ErfKernelRoutine(Input, Output, N);
-#else
+#else
```
hariharans29: Nit: Whitespace

cmake/CMakeLists.txt
```diff
@@ -1048,6 +1062,12 @@ function(onnxruntime_set_compile_flags target_name)
       target_compile_definitions(${target_name} PRIVATE USE_KLEIDIAI)
     endif()
 
+    if(HAS_ARM64_SVE)
+      if("${target_name}" STREQUAL "onnxruntime_mlas")
```
edgchen1: I think the preferred place to call `target_compile_definitions(onnxruntime_mlas ...)` would be in onnxruntime_mlas.cmake instead of this general helper function.

onnxruntime/core/mlas/lib/compute.cpp
```diff
@@ -876,11 +874,12 @@ Return Value:
         //
         // Find the maximum value for the row.
         //
+        float Maximum;
 
-#if defined(MLAS_TARGET_AMD64) || defined(MLAS_TARGET_LARCH64)
-        float Maximum = GetMlasPlatform().ReduceMaximumF32Kernel(Input, D);
-#else
-        float Maximum = MlasReduceMaximumF32Kernel(Input, D);
+#if defined(MLAS_TARGET_AMD64) || defined(MLAS_TARGET_LARCH64) || defined(USE_SVE)
```
edgchen1: nit: what about a more specific macro name like `MLAS_USE_SVE`?

cmake/CMakeLists.txt
```diff
@@ -612,6 +614,18 @@ else()
...
+  if(onnxruntime_USE_SVE)
+    if(CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64" AND CMAKE_SYSTEM_NAME STREQUAL "Linux")
+      check_cxx_compiler_flag("-march=armv8.2-a+sve" HAS_ARM64_SVE)
+      if(HAS_ARM64_SVE)
+        message(STATUS "Compiler supports SVE!")
+      else()
+        message(STATUS "Compiler does not support SVE!")
+      endif()
+    endif()
+  endif()
```
edgchen1: how about updating `onnxruntime_USE_SVE` to be `OFF` if it is not supported? then we can simply check for `onnxruntime_USE_SVE` later on.
```suggestion
  if(onnxruntime_USE_SVE)
    if(CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64" AND CMAKE_SYSTEM_NAME STREQUAL "Linux")
      check_cxx_compiler_flag("-march=armv8.2-a+sve" HAS_ARM64_SVE)
      if(HAS_ARM64_SVE)
        message(STATUS "Compiler supports SVE!")
      else()
        message(WARNING "onnxruntime_USE_SVE was set but compiler does not support SVE. It will be disabled.")
        set(onnxruntime_USE_SVE OFF)
      endif()
    else()
      message(WARNING "onnxruntime_USE_SVE was set but it is not supported on this platform. It will be disabled.")
      set(onnxruntime_USE_SVE OFF)
    endif()
  endif()
```

edgchen1: also, I noticed that this code is within an `else()`. can it be moved outside so that it always runs?

```
if(WIN32)
  ...
else()
  <this check>
endif()
```

sanketkaleoss: Thanks, updated!

onnxruntime/core/mlas/lib/compute.cpp
sanketkaleoss: Sure, modified

cmake/CMakeLists.txt
sanketkaleoss: Okay, incorporated into the latest commit

hariharans29:
There maybe some pending comments to be resolved - they have "pings" on them. Could you please address them ? Also, can you physically resolve (hit Resolve) on the comments that have been resolved - it is needed to merge? Thanks.

sanketkaleoss:
> There maybe some pending comments to be resolved - they have "pings" on them. Could you please address them ? Also, can you physically resolve (hit Resolve) on the comments that have been resolved - it is needed to merge? Thanks.

@hariharans29  Updated and resolved the comments

cmake/onnxruntime_mlas.cmake
```diff
@@ -432,6 +432,15 @@ else()
           ${MLAS_SRC_DIR}/eltwise_kernel_neon.cpp
           ${MLAS_SRC_DIR}/sqnbitgemm_kernel_neon_int8_i8mm.cpp
         )
+        
+        # Conditionally add the SVE implementation if compiler supports it
+        if (onnxruntime_USE_SVE AND HAS_ARM64_SVE)
```
edgchen1: nit: we should only need to check `onnxruntime_USE_SVE` now
```suggestion
        if (onnxruntime_USE_SVE)
```

hariharans29:
Can you please fix the lint issues : https://github.com/microsoft/onnxruntime/blob/main/docs/Coding_Conventions_and_Standards.md#linting

cmake/onnxruntime_mlas.cmake
sanketkaleoss: @edgchen1 Sure, incorporated in latest commit

sanketkaleoss:
> Can you please fix the lint issues : https://github.com/microsoft/onnxruntime/blob/main/docs/Coding_Conventions_and_Standards.md#linting

@hariharans29 Sure, resolved in the latest commit

hariharans29:
Hi @sanketkaleoss  : This PR crashes a test related to the 8 bit Gemms. See here- https://aiinfra.visualstudio.com/Lotus/_build/results?buildId=952427&view=logs&j=d3350775-409a-5282-627a-f3b59b82cd3f&t=a42124b0-66aa-50d6-7780-383fc566e097&l=32556. 

Any chance you know why ?

sanketkaleoss:
> Hi @sanketkaleoss : This PR crashes a test related to the 8 bit Gemms. See here- https://aiinfra.visualstudio.com/Lotus/_build/results?buildId=952427&view=logs&j=d3350775-409a-5282-627a-f3b59b82cd3f&t=a42124b0-66aa-50d6-7780-383fc566e097&l=32556.
> 
> Any chance you know why ?

@hariharans29 I'm unable to open this link, says can't access the application.
Anyhow, this PR doesn't modify any 8-bit ops, it is only for FP32 elementwise_ops. Not sure why it would cause the crash for 8 bit gemm test. Can you provide more info on this?

hariharans29:
> > Hi @sanketkaleoss : This PR crashes a test related to the 8 bit Gemms. See here- https://aiinfra.visualstudio.com/Lotus/_build/results?buildId=952427&view=logs&j=d3350775-409a-5282-627a-f3b59b82cd3f&t=a42124b0-66aa-50d6-7780-383fc566e097&l=32556.
> > Any chance you know why ?
> 
> @hariharans29 I'm unable to open this link, says can't access the application. Anyhow, this PR doesn't modify any 8-bit ops, it is only for FP32 elementwise_ops. Not sure why it would cause the crash for 8 bit gemm test. Can you provide more info on this?

Hi @sanketkaleoss - The relationship is unclear, but it definitely crashes this test - https://github.com/microsoft/onnxruntime/blob/210954779264e3a195dcee3a7ba24c866f7b1457/onnxruntime/test/mlas/unittest/test_sq8bitgemm.cpp#L865.

The test ARM machine SKU is Standard_D16pds_v5. More details here: https://learn.microsoft.com/en-us/azure/virtual-machines/sizes/general-purpose/dpdsv5-series?tabs=sizebasic. I think the Ampere Altra processor does not support SVE ?

Would you be able to try running the test on a Linux machine (since SVE PR support is only added for that platform) and with the processor not having SVE support if you don't have access to an Ampere Altra ?

Just build onnxruntime, and run `./onnxruntime_mlas_test.exe --gtest_filter=SQ8BitQuantA.ShortExecute`

sanketkaleoss:
> > > Hi @sanketkaleoss : This PR crashes a test related to the 8 bit Gemms. See here- https://aiinfra.visualstudio.com/Lotus/_build/results?buildId=952427&view=logs&j=d3350775-409a-5282-627a-f3b59b82cd3f&t=a42124b0-66aa-50d6-7780-383fc566e097&l=32556.
> > > Any chance you know why ?
> > 
> > 
> > @hariharans29 I'm unable to open this link, says can't access the application. Anyhow, this PR doesn't modify any 8-bit ops, it is only for FP32 elementwise_ops. Not sure why it would cause the crash for 8 bit gemm test. Can you provide more info on this?
> 
> Hi @sanketkaleoss - The relationship is unclear, but it definitely crashes this test -
> 
> https://github.com/microsoft/onnxruntime/blob/210954779264e3a195dcee3a7ba24c866f7b1457/onnxruntime/test/mlas/unittest/test_sq8bitgemm.cpp#L865
> 
> .
> The test ARM machine SKU is Standard_D16pds_v5. More details here: https://learn.microsoft.com/en-us/azure/virtual-machines/sizes/general-purpose/dpdsv5-series?tabs=sizebasic. I think the Ampere Altra processor does not support SVE ?
> 
> Would you be able to try running the test on a Linux machine (since SVE PR support is only added for that platform) and with the processor not having SVE support if you don't have access to an Ampere Altra ?
> 
> Just build onnxruntime, and run `./onnxruntime_mlas_test.exe --gtest_filter=SQ8BitQuantA.ShortExecute`

@hariharans29 Getting "Bus error (core dumped)" 
while running that test on both AWS Graviton3E(SVE machine) and Graviton2(Non-sve machine)

hariharans29:
> > > > Hi @sanketkaleoss : This PR crashes a test related to the 8 bit Gemms. See here- https://aiinfra.visualstudio.com/Lotus/_build/results?buildId=952427&view=logs&j=d3350775-409a-5282-627a-f3b59b82cd3f&t=a42124b0-66aa-50d6-7780-383fc566e097&l=32556.
> > > > Any chance you know why ?
> > > 
> > > 
> > > @hariharans29 I'm unable to open this link, says can't access the application. Anyhow, this PR doesn't modify any 8-bit ops, it is only for FP32 elementwise_ops. Not sure why it would cause the crash for 8 bit gemm test. Can you provide more info on this?
> > 
> > 
> > Hi @sanketkaleoss - The relationship is unclear, but it definitely crashes this test -
> > https://github.com/microsoft/onnxruntime/blob/210954779264e3a195dcee3a7ba24c866f7b1457/onnxruntime/test/mlas/unittest/test_sq8bitgemm.cpp#L865
> > 
> > .
> > The test ARM machine SKU is Standard_D16pds_v5. More details here: https://learn.microsoft.com/en-us/azure/virtual-machines/sizes/general-purpose/dpdsv5-series?tabs=sizebasic. I think the Ampere Altra processor does not support SVE ?
> > Would you be able to try running the test on a Linux machine (since SVE PR support is only added for that platform) and with the processor not having SVE support if you don't have access to an Ampere Altra ?
> > Just build onnxruntime, and run `./onnxruntime_mlas_test.exe --gtest_filter=SQ8BitQuantA.ShortExecute`
> 
> @hariharans29 Getting "Bus error (core dumped)" while running that test on both AWS Graviton3E(SVE machine) and Graviton2(Non-sve machine)

Can you please debug it ? I don't see it in the pipeline with  commit before the SVE PR. I got access to a Graviton4 machine and I see it the core dump there too.

hariharans29:
Also when I build with `--no_sve `, the test passes

sanketkaleoss:
> > > > > Hi @sanketkaleoss : This PR crashes a test related to the 8 bit Gemms. See here- https://aiinfra.visualstudio.com/Lotus/_build/results?buildId=952427&view=logs&j=d3350775-409a-5282-627a-f3b59b82cd3f&t=a42124b0-66aa-50d6-7780-383fc566e097&l=32556.
> > > > > Any chance you know why ?
> > > > 
> > > > 
> > > > @hariharans29 I'm unable to open this link, says can't access the application. Anyhow, this PR doesn't modify any 8-bit ops, it is only for FP32 elementwise_ops. Not sure why it would cause the crash for 8 bit gemm test. Can you provide more info on this?
> > > 
> > > 
> > > Hi @sanketkaleoss - The relationship is unclear, but it definitely crashes this test -
> > > https://github.com/microsoft/onnxruntime/blob/210954779264e3a195dcee3a7ba24c866f7b1457/onnxruntime/test/mlas/unittest/test_sq8bitgemm.cpp#L865
> > > 
> > > .
> > > The test ARM machine SKU is Standard_D16pds_v5. More details here: https://learn.microsoft.com/en-us/azure/virtual-machines/sizes/general-purpose/dpdsv5-series?tabs=sizebasic. I think the Ampere Altra processor does not support SVE ?
> > > Would you be able to try running the test on a Linux machine (since SVE PR support is only added for that platform) and with the processor not having SVE support if you don't have access to an Ampere Altra ?
> > > Just build onnxruntime, and run `./onnxruntime_mlas_test.exe --gtest_filter=SQ8BitQuantA.ShortExecute`
> > 
> > 
> > @hariharans29 Getting "Bus error (core dumped)" while running that test on both AWS Graviton3E(SVE machine) and Graviton2(Non-sve machine)
> 
> Can you please debug it ? I don't see it in the pipeline with commit before the SVE PR. I got access to a Graviton4 machine and I see it the core dump there too.

Sure, thanks for the insight, working on fixing it

Issue #24774, opened 2025-05-15
[Build] CCCL API migration issue.


mc-nv:
### Describe the issue

We would like to admit that following deprecation required to be completed. 
https://nvidia.github.io/cccl/cccl/3.0_migration_guide.html

We eventually start seeing issue during the compilation process against product that are in development for NVIDIA. 

### Urgency

It need to be patched before June.

### Target platform

Linux

### Build script

```
build [DEBUG] - Command line arguments:
  --build_dir /workspace/onnxruntime/build/Linux --config Release --skip_submodule_sync --parallel --build_shared_lib --compile_no_warning_as_error --build_dir /workspace/build --cmake_extra_defines 'CMAKE_CUDA_ARCHITECTURES='"'"'75;80;86;89;90;100;120'"'"'' --cmake_extra_defines CMAKE_POLICY_VERSION_MINIMUM=3.5 --update --build --use_cuda --cuda_home /usr/local/cuda --cudnn_home /usr --use_tensorrt --use_tensorrt_builtin_parser --tensorrt_home /usr/src/tensorrt --allow_running_as_root --use_openvino CPU
```

### Error / output

```
[ 80%] Building CUDA object CMakeFiles/onnxruntime_providers_cuda.dir/workspace/onnxruntime/onnxruntime/contrib_ops/cuda/bert/cudnn_fmha/cudnn_flash_attention.cu.o
/workspace/onnxruntime/onnxruntime/contrib_ops/cuda/bert/attention_softmax.cu(98): error: namespace "cub" has no member "Max"
    const auto max = BlockReduce(tmp_storage).Reduce(thread_data_max, cub::Max());
                                                                           ^

/workspace/onnxruntime/onnxruntime/contrib_ops/cuda/bert/attention_softmax.cu(117): error: namespace "cub" has no member "Sum"
    const auto sum = BlockReduce(tmp_storage).Reduce(thread_data_sum, cub::Sum());
                                                                           ^

/workspace/onnxruntime/onnxruntime/contrib_ops/cuda/bert/attention_softmax.cu(174): error: namespace "cub" has no member "Max"
    const auto max = BlockReduce(tmp_storage).Reduce(input_data, cub::Max(), end);
                                                                      ^

/workspace/onnxruntime/onnxruntime/contrib_ops/cuda/bert/attention_softmax.cu(187): error: namespace "cub" has no member "Sum"
[...]

Issue #25182, opened 2025-06-26
[Build] Build CUDA 11.8 Visual Studio limitation


AndreyOrb:
### Describe the issue

Well, this commit together with some other issue (that no one could predict) turned out to be a problem.

Here's the problem:
There's still a general support for CUDA 11.8 in ORT, but because of [this](https://stackoverflow.com/questions/78515942/cuda-compatibility-with-visual-studio-2022-version-17-10) limitation, only Visual Studio 2019 or Visual Studio 2022 v17.9.2 (or less) can still compile the CUDA 11.8

In my case, I must be able to build ORT with both CUDA 11.8 and CUDA 12.x (as separate ORT builds, of course).

From different resources ([microsoft](https://learn.microsoft.com/en-us/visualstudio/install/install-visual-studio-versions-side-by-side?view=vs-2022), [microsoft](https://learn.microsoft.com/en-us/answers/questions/1623842/is-it-possible-to-install-different-versions-or-ed)) I learned that several VS versions with the same major version can be installed. 
Then, I will probably have to run build.bat from the correct VS CMD (unverified).

For now, the easiest way for me to solve the problem was to add back "Visual Studio 16 2019" in build.py.
This way, I can build ORT with CUDA 11.8 using --cmake_generator "Visual Studio 16 2019" --compile_no_warning_as_error

### Urgency

Not urgent

### Target platform

Windows

### Build script

.\build.bat --config Debug --build_shared_lib --parallel --use_cuda --cuda_home "C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v11.8" --cudnn_home "C:\Program Files\NVIDI
A GPU Computing Toolkit\cuDNN\cudnn-windows-x86_64-8.9.0.131_cuda11-archive" --cuda_version 11.8 --use_tensorrt --tensorrt_home "C:\Program Files\NVIDIA GPU Computing Toolkit\TensorRT\TensorRT-10.9.0.34.Windows.win10.cuda-11.8\TensorRT-10.9.0.34" --use_tensorrt_oss_parser --cmake_generator "Visual Studio 16 2019" --compile_no_warning_as_error --cmake_path E:\3rdParties\cmake-4.0.3\build\bin\Release\cmake.exe --skip_tests --enable_cuda_line_info --use_mimalloc --cmake_extra_defines onnxruntime_BUILD_UNIT_TESTS=OFF --skip_tests

### Error / output

[...]

PR #25246, opened 2025-07-01
Update deprecated CCCL API

mc-nv:
### Description
Update API 

### Motivation and Context
Address issues for: https://github.com/microsoft/onnxruntime/issues/24774


mc-nv:

snnn (dismissed):

snnn:
@mc-nv , it has some build failures. Please fix them.

tianleiwu:
CI pipeline is using CUDA 12.2, which does not have ::cuda::maximum(). Need code like the following to fix build error:
https://github.com/microsoft/onnxruntime/blob/b05b252de225d02be3da72be2c029065b0f42089/onnxruntime/contrib_ops/cuda/bert/embed_layer_norm_impl.cu#L42-L47

mc-nv:
@snnn  where I can find build environment configuration, or may be you can provide the version of CCCL installed. 
At glance build uses CUDA 12.2 which `18 month old` for NVIDIA. 
Will it be possible to validate changes within 12.9 build environment? 

Why we asking for it because CCCL API deprecation was introduced in version 2.8.0 
https://github.com/NVIDIA/cccl/releases/tag/v2.8.0 
And that version is `4 month old`.

snnn:
People are even asking for CUDA 11.x support.  #25182 

I think we still need to make the source code be compatible with CUDA 12.2 before the next release is done. Then we can discuss upgrading.

mc-nv:
got it I will address those changes and get back to this PR

tianleiwu (approved):

snnn (approved):

mc-nv:
I'm afraid I need to push one more commit towards to this change...

snnn:
That's fine. Welcome to submit a new PR.

tianleiwu:
I tested Windows build using cuda 12.9 with/without this PR. Both are good.

CUDA_VERSION is defined in cuda.h;  CUDART_VERSION is defined in cuda_runtime_api.h. It is better to include cuda.h or  cuda_runtime_api.h explicitly.

PR #25247, opened 2025-07-01
[EP ABI] Use pre-allocated input buffers for APIs that return arrays.

adrianlizarraga:
### Description
- Remove `OrtArrayOfConstObjects` from C API
- Rework graph APIs that return an array of objects to take pre-allocated buffers as input.
- Rename `Node_GetParentGraph` to `Node_GetGraph`
- Fixes C/C++ API documentation generation: https://github.com/microsoft/onnxruntime/actions/runs/16029991022

### Motivation and Context
Make the graph C APIs easier to use. The `OrtArrayOfConstObjects` approach was too verbose to use and made the API harder to understand because the function signatures did not show the element data types.

Example usage with `OrtArrayOfConstObjects`:

```c++
const OrtGraph* graph;  // Assumed is initialized

OrtArrayOfConstObjects* nodes = nullptr;
RETURN_IF_ERROR(ort_api.Graph_GetNodes(graph, &nodes));  // Get array

size_t num_nodes = 0;
RETURN_IF_ERROR(ort_api.ArrayOfConstObjects_GetSize(nodes, &num_nodes)); // Get size

// Use the nodes.
for (size_t i = 0; i < num_nodes; i++) {
  const OrtNode* node = nullptr;
  RETURN_IF_ERROR(ort_api.ArrayOfConstObjects_GetElementAt(nodes, i,
                                                           reinterpret_cast<const void**>(&node)));

   // Inspect OrtNode properties ...
}

// Have to manually release the OrtArrayOfConstObjects
// A C++ ORT wrapper class would help via RAII, but the same C api calls are made under the hood.
ort_api.ReleaseArrayOfConstObjects(nodes);
```

Example usage with "pre-allocated" buffers style:

```c++
const OrtGraph* graph;  // Assumed is initialized

// Get number of nodes.
size_t num_nodes = 0;
RETURN_IF_ERROR(ort_api.Graph_GetNumNodes(graph, &num_nodes));

// Pre-allocate buffer of OrtNode* and get nodes.
std::vector<const OrtNode*> nodes(num_nodes);
RETURN_IF_ERROR(ort_api.Graph_GetNodes(graph, nodes.data(), nodes.size()));

// Use the nodes.
for (size_t i = 0; i < num_nodes; i++) {
  const OrtNode* node = nodes[i];
  // Inspect OrtNode properties.
}

// std::vector destructor cleans up for us.
```


include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -3864,11 +3852,11 @@ struct OrtApi {
    *      assigned to QNN EP is dumped to a separate file.
    *   "json_qnn_graph_dir": Directory in which to dump QNN JSON graphs. If not specified, QNN graphs are dumped in the
    *      program's current working directory. Ignored if "dump_json_qnn_graph" is not set.
-   * "op_packages": QNN UDO op_package for QNN EP, allowed format:
-   *   <op_type>:<op_package_path>:<interface>[:<target>],<op_type2>:<op_package_path2>:<interface2>[:<target>],
-   *   where op_type is the name of the operation, op_package_path is the path to the op package shared library,
-   *   interface is the symbol name to register the op life cycle functions, and target is the backend type. For more
-   *   details, refer to: https://docs.qualcomm.com/bundle/publicresource/topics/80-63442-50/op_packages.html
+   *   "op_packages": QNN UDO op_package for QNN EP, allowed format:
+   *     "<op_type>:<op_package_path>:<interface>[:<target>],<op_type2>:<op_package_path2>:<interface2>[:<target>]",
```
adrianlizarraga, line 3856: note: `<op_type>` caused C/C++ api documentation generation to fail. We should probably consider enabling this github action for every PR that touches the C/C++ header files

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -5567,21 +5427,22 @@ struct OrtApi {
...
-   * \param[out] input_indices Pre-allocated array of `max_num_consumers` elements that will be filled
+   * \param[out] nodes Pre-allocated array of size `num_consumers` that is filled with OrtNode instances.
+   * \param[out] input_indices Pre-allocated array of `num_consumers` elements that is filled
    *                           with input indices. Index is set to -1 for an "implicit" input to a consumer node
    *                           that contains a subgraph (e.g., If, Loop) with nodes that use the value internally.
-   * \param[in] max_num_consumers The maximum size of the `consumer_nodes` and `consumer_input_indices` arrays.
-   *                              Typical usage sets this to the value of ValueInfo_GetValueNumConsumers().
+   * \param[in] num_consumers The size of the `consumer_nodes` and `consumer_input_indices` arrays.
+   *                          Typical usage sets this to the value of ValueInfo_GetValueNumConsumers().
+   *                          An error status returned if `num_consumers` is less than the number of actual consumers.
```
```suggestion
   *                          An error status is returned if `num_consumers` is less than the number of actual
   *                          consumers.
```

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -5702,34 +5564,78 @@ struct OrtApi {
    */
   ORT_API2_STATUS(Graph_GetOnnxIRVersion, _In_ const OrtGraph* graph, _Out_ int64_t* onnx_ir_version);
 
+  /** \brief Returns the number of graphs inputs.
```
```suggestion
  /** \brief Returns the number of graph inputs.
```

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -5702,34 +5564,78 @@ struct OrtApi {
...
+   *
+   * Counts initializers that are included in the list of graph inputs.
```
edgchen1: nit: it could be interpreted as counting initializers. but it actually counts initializers and non-initializer inputs.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -5702,34 +5564,78 @@ struct OrtApi {
...
    *
    * \snippet{doc} snippets.dox OrtStatus Return Value
    *
    * \since Version 1.23.
    */
-  ORT_API2_STATUS(Graph_GetInputs, _In_ const OrtGraph* graph, _Outptr_ OrtArrayOfConstObjects** inputs);
+  ORT_API2_STATUS(Graph_GetInputs, _In_ const OrtGraph* graph,
+                  _Out_writes_(num_inputs) const OrtValueInfo** inputs, _In_ size_t num_inputs);
+
+  /** \brief Returns the number of graphs outputs.
```
```suggestion
  /** \brief Returns the number of graph outputs.
```

onnxruntime/core/graph/ep_api_types.cc
```diff
@@ -22,6 +22,21 @@
 
 namespace onnxruntime {
 
+template <typename DstElem>
+static Status CheckCopyDestination(std::string_view error_array_label, size_t src_size, gsl::span<DstElem const> dst) {
+  if (dst.size() < src_size) {
+    return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT,
+                           "Not enough space for ", error_array_label, ": expected buffer with size >= ",
+                           src_size, ", but got buffer of size ", dst.size());
```
edgchen1: should we clarify that it is not byte size? e.g.,
```suggestion
                           "Not enough space for ", error_array_label, ": expected buffer with room for at least ",
                           src_size, " elements, but got buffer with room for only ", dst.size(), " elements.");
```

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -5702,34 +5564,78 @@ struct OrtApi {
...
    *
    * Includes initializers that are included in the list of graph inputs.
    *
    * \param[in] graph The OrtGraph instance.
-   * \param[out] inputs Output parameter set to a new OrtArrayOfConstObjects instance containing the graph inputs
-   *                    as OrtValueInfo instances. Must be released by calling ReleaseArrayOfConstObjects.
+   * \param[out] inputs Pre-allocated array of `num_inputs` elements that is filled with the graph's inputs.
+   * \param[in] num_inputs The size of the `inputs` array.
+   *                       Typical usage sets this to the result of Graph_GetNumInputs(). An error status is
+   *                       returned if `num_inputs` is less than the number of graph inputs.
```
edgchen1, line 5589: would it be worth having another error code to indicate "not enough space in buffer"?

include/onnxruntime/core/session/onnxruntime_c_api.h
adrianlizarraga: What do you think of something like: 
>Counts both required and optional graph inputs. An optional graph input (introduced in ONNX IR version 4) has a matching initializer that serves as the default value.

edgchen1: that's good. should we also explain what it means for IR version < 4, or perhaps it's not practically relevant now?
I think something simply like "The count includes initializers that are included in the list of graph inputs." would also work.

include/onnxruntime/core/session/onnxruntime_c_api.h
adrianlizarraga: Existing APIs that use pre-allocated buffer currently return `ORT_INVALID_ARGUMENT`. Given that we need to merge this very soon, I'd rather remain consistent with older APIs if that's ok.

adrianlizarraga: actually, let me double check that claim.

adrianlizarraga: Yeah, so APIs that use pre-allocated buffers (like [KernelInfo_GetInputName](https://github.com/microsoft/onnxruntime/blob/d81905edfdd845d05640f409c2a5596bfafe8c68/include/onnxruntime/core/session/onnxruntime_c_api.h#L4125)) return `ORT_INVALID_ARGUMENT` if the buffer is too small: https://github.com/microsoft/onnxruntime/blob/6707dd4ec5a4b03d33852539ed16cb90ff6d751b/onnxruntime/core/session/utils.cc#L119

I would prefer not changing that now and it does make sense that this would be an invalid argument.

include/onnxruntime/core/session/onnxruntime_c_api.h
adrianlizarraga: Thanks. Updated with your suggestion.

include/onnxruntime/core/session/onnxruntime_c_api.h
edgchen1: I think it is a more specific type of "invalid argument". there are multiple ways for an argument to be invalid.

I suppose that with a separate API to get the size it doesn't matter as much. it'd be more useful if we had a single API that could first query the required size.

onnxruntime/test/autoep/library/ep.cc
```diff
@@ -233,53 +229,39 @@ OrtStatus* ORT_API_CALL ExampleEp::GetCapabilityImpl(OrtEp* this_ptr, const OrtG
...
+      RETURN_IF_ERROR(ep->ort_api.Node_GetNumInputs(node, &num_inputs));
+      RETURN_IF_ERROR(ep->ort_api.Node_GetNumOutputs(node, &num_outputs));
       RETURN_IF(num_inputs != 2 || num_outputs != 1, ep->ort_api, "Mul should have 2 inputs and 1 output");
 
-      const void* const* inputs_data = nullptr;
-      const void* const* outputs_data = nullptr;
-      RETURN_IF_ERROR(ep->ort_api.ArrayOfConstObjects_GetData(inputs_array, &inputs_data));
-      RETURN_IF_ERROR(ep->ort_api.ArrayOfConstObjects_GetData(outputs_array, &outputs_data));
+      std::vector<const OrtValueInfo*> inputs(num_inputs);
+      std::vector<const OrtValueInfo*> outputs(num_inputs);
```
edgchen1:
```suggestion
      std::vector<const OrtValueInfo*> outputs(num_outputs);
```

edgchen1 (approved):

PR #25250, opened 2025-07-02
Refactor LogProviderOptions

ashrit-ms:
### Description
This pull request refactors the logging of provider options in the ONNX Runtime framework to improve telemetry functionality. The changes include consolidating logging logic, introducing platform-independent methods, and enhancing the telemetry interface for better extensibility.

### DIFF
--- a/onnxruntime/core/platform/windows/telemetry.cc
+++ b/onnxruntime/core/platform/windows/telemetry.cc
@@ -478,4 +478,36 @@ void WindowsTelemetry::LogAutoEpSelection(uint32_t session_id, const std::string
                     TraceLoggingString(available_execution_provider_string.c_str(), "availableExecutionProviderIds"));
 }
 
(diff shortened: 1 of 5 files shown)


onnxruntime/core/platform/windows/telemetry.cc
```diff
@@ -478,4 +478,36 @@ void WindowsTelemetry::LogAutoEpSelection(uint32_t session_id, const std::string
                     TraceLoggingString(available_execution_provider_string.c_str(), "availableExecutionProviderIds"));
 }
 
+void WindowsTelemetry::LogProviderOptions(const std::string& provider_id, const std::string& provider_options_string, bool captureState) const {
```
snnn, line 481: For newer code we should consider using std::string_view instead of std::string&

snnn (approved):
Thanks.

onnxruntime/core/platform/windows/telemetry.cc
ashrit-ms: Thanks, let me create a follow-up to update all the events to use string_view

PR #25253, opened 2025-07-02
check nchwc_inputs before accessing

s0md3v:
### Description
Fixes a build error by checking that `nchwc_inputs` is not empty before accessing its elements in `TransformBinary`.

### Motivation and Context
Compiler flagged a possible uninitialized access to `nchwc_inputs[0]` (when empty), causing the build to fail. This check ensures safe access and avoids undefined behavior. No impact on valid graphs.


tianleiwu:
It is false alert since the function is applied to **Add**, **Sum**, **Mul** which shall have at least two inputs.

https://github.com/microsoft/onnxruntime/blob/c122f0d8b80dc54548469d29e79b9bb3dcd58d7d/onnxruntime/core/optimizer/nchwc_transformer.cc#L1164-L1168

Anyway, an extra check is fine to me even though not necessary.

PR #25254, opened 2025-07-02
Plugin EP data transfer and Stream support.

skottmckay:
### Description

Plugin EP data transfer and Stream support.

Add the ability for a plugin EP to provide an IDataTransfer implementation and an OrtSyncStream implementation to do async data copy outside of an inference session.

Example usage added for CUDA EP.

Caveat: Support for providing the OrtSyncStream from the data copy to Session.Run will be a follow up PR. For the CUDA EP we can pass in the native cudaStream_t from the OrtSyncStream used for the data copy to the Run via CUDA EP provider options.


skottmckay:
@gedoensmax 
@gaugarg-nv

bil-ash:
Will this PR allow streaming the models from disk, thereby reducing device memory usage?
Or is it between CPU and GPU only?
If it is between CPU/GPU and disk, please add support for WebGPU or WASM also.

onnxruntime/core/framework/plugin_ep_stream.h
```diff
@@ -0,0 +1,105 @@
...
+struct OrtSyncNotification : onnxruntime::synchronize::Notification {};
+
+using onnxruntime::logging::Logger;
+
+namespace onnxruntime {
+namespace plugin_ep {
+
+class Notification : public synchronize::Notification {
+ public:
+  Notification(Stream& stream, OrtSyncNotificationImpl& impl)
```
gaugarg-nv: Any reason why this class takes `Stream` as a constructor argument?
At least for CUDA, events can be created independent of a stream.

onnxruntime/core/framework/plugin_ep_stream.h
```diff
@@ -0,0 +1,105 @@
...
+    this_ptr->impl_.WaitOnDevice(&this_ptr->impl_, static_cast<OrtSyncStream*>(stream));
+  }
+
+  static void WaitNotificationOnHost(onnxruntime::Stream* /*stream*/, synchronize::Notification& notification) {
+    auto* this_ptr = static_cast<Notification*>(&notification);
+    this_ptr->impl_.WaitOnHost(&this_ptr->impl_);
+  }
+
+  void Activate() override {
+    impl_.Activate(&impl_);
```
gaugarg-nv: `Activate` or maybe we can call it `Signal` can take a stream as input, on which event can be recorded.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -5378,10 +5384,18 @@ struct OrtApi {
    * \since Version 1.23
    */
   ORT_API2_STATUS(CreateMemoryInfo_V2, _In_ const char* name, _In_ enum OrtMemoryInfoDeviceType device_type,
-                  _In_ uint32_t vendor_id, _In_ int16_t device_id, _In_ enum OrtDeviceMemoryType mem_type,
+                  _In_ uint32_t vendor_id, _In_ int32_t device_id, _In_ enum OrtDeviceMemoryType mem_type,
```
skottmckay, line 5392: This is for consistency. The rest of the C API uses int32_t for device_id even though it's stored in an int16_t internally.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -6072,6 +6089,147 @@ struct OrtApi {
...
+                  _In_reads_(num_tensors) OrtValue** dst_tensors,
+                  _In_reads_opt_(num_tensors) OrtSyncStream** streams,
+                  _In_ size_t num_tensors);
+
+  // create a Notification for input being available.
+  // user should call Activate.
+  // ORT will wait on the notification if the stream is provided to Run
+  ORT_API2_STATUS(CreateInputSyncNotification, _In_ OrtSyncStream* stream);
+  ORT_API2_STATUS(ActivateInputSyncNotification, _In_ OrtSyncStream* stream);
+  ORT_API2_STATUS(ReleaseInputSyncNotification, _In_ OrtSyncStream* stream);
```
skottmckay: Temporary addition for discussion

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -6086,6 +6103,140 @@ struct OrtApi {
    * \since Version 1.23.
    */
   ORT_API2_STATUS(GetSessionOptionsConfigEntries, _In_ const OrtSessionOptions* options, _Outptr_ OrtKeyValuePairs** out);
+
+  /** \brief Get the OrtMemoryInfo for each input of the session.
+   *
+   * The memory info can be used to determine where the input tensors are required.
```
skottmckay, line 6200: To allocate inputs/outputs on the correct device the user needs OrtMemoryInfo. They can call GetSharedAllocator with the memory info, create OrtValue instances, and use CopyTensors to copy to/from device.

Considerations:
  - we need the OrtMemoryInfo from the input to check if it's DEFAULT or HOST_ACCESSIBLE memory
    - this depends on the spec for the node consuming the input
  - a custom allocator may be registered for the OrtMemoryInfo, and to honor that we need to lookup the allocator using GetSharedAllocator

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -6086,6 +6103,140 @@ struct OrtApi {
...
+   * \param[out] inputs_ep_devices Pre-allocated array of size `num_inputs` that will be filled with
+   *                               OrtEpDevice* values for each input.
+   *                               The order is the same as returned by SessionGetInputName.
+   * \param[in] num_inputs The number of inputs in the session. Must match SessionGetInputCount.
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \since Version 1.23
+   */
+  ORT_API2_STATUS(SessionGetEpDeviceForInputs, _In_ const OrtSession* session,
```
skottmckay, line 6266: Creating an OrtSyncStream is via the OrtEpDevice (as that contains a pointer to the factory internally), so we also need the ability to get the OrtEpDevice for the inputs.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -6086,6 +6103,140 @@ struct OrtApi {
...
+   * \param[in] num_tensors The number of tensors to copy. The size of `src_tensors` and `dst_tensors` must match.
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \since Version 1.23
+   */
+  ORT_API2_STATUS(CopyTensors, _In_ const OrtEnv* env,
+                  _In_reads_(num_tensors) const OrtValue** src_tensors,
+                  _In_reads_(num_tensors) OrtValue** dst_tensors,
+                  _In_opt_ OrtSyncStream* stream,
```
skottmckay, line 6330: Using a single stream instance for the copy as that seems to be sufficient given the implementation requires the copy to be between two devices only. If for some reason multiple streams were required the user can call CopyTensors once for each stream.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -67,6 +66,115 @@ typedef struct OrtDataTransferImpl {
...
+  /** \brief Create an OrtSyncNotificationImpl for the OrtSyncStream
+   *
+   * \param[in] this_ptr Pointer to the OrtSyncStreamImpl instance
+   * \param[in] stream The OrtSyncStream instance to create the notification for.
+   * \param[in] num_consumers The number of consumers that will use this notification. 0 if unknown/irrelevant.
+   * \param[out] notification Pointer to the OrtSyncNotificationImpl instance that will be created.
+   *
+   * \since Version 1.23.
+   */
+  ORT_API2_STATUS(CreateNotification, _In_ void* this_ptr,
```
skottmckay: Should we call this `CreateNotification` or `CreateNotificationImpl`?

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -709,13 +817,46 @@ struct OrtEpFactory {
...
    *
    * \param[in] this_ptr The OrtEpFactory instance.
-   * \param[out] data_transfer The created OrtDataTransferImpl instance.
+   * \param[out] data_transfer The created OrtDataTransferImpl instance. Set to nullptr if not required.
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \since Version 1.23.
+   */
+  ORT_API2_STATUS(CreateDataTransfer, _In_ OrtEpFactory* this_ptr,
```
skottmckay, line 869: Should we call this `CreateDataTransfer` or `CreateDataTransferImpl`?

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -709,13 +817,46 @@ struct OrtEpFactory {
...
+   * \param[out] stream The created OrtSyncStreamImpl instance. nullptr if the execution provider is not stream aware.
    *
    * \snippet{doc} snippets.dox OrtStatus Return Value
    *
    * \since Version 1.23.
    */
-  ORT_API2_STATUS(CreateDataTransfer, _In_ OrtEpFactory* this_ptr, _Outptr_ OrtDataTransferImpl** data_transfer);
+  ORT_API2_STATUS(CreateSyncStreamForDevice, _In_ OrtEpFactory* this_ptr,
+                  _In_ const OrtMemoryDevice* memory_device,
+                  _In_opt_ const OrtEp* ep,  // ??? Is it better to have this optional arg, or to add a CreateSyncStream to OrtEp that can be optionally implemented
```
skottmckay: There are scenarios where session/EP options affect how the Stream and Allocator is configured. 

CUDA: Some stream settings depend on whether graph capture is enabled, which is a per-session value.
WebGPU: Allocator settings for usage in an inference session differ from settings for external data copies.

2 options: 

1) Add the optional `OrtEp*` parameter to CreateSyncStreamForDevice and CreateAllocator in the OrtEpFactory. Our wrapper EP implementation would provide the ep instance when used internally to an inference session, and this differentiates the two usages.

2) Add CreateSyncStream and CreateAllocator functions to OrtEp as well. Optional to implement. We can fallback to the OrtEpFactory version if they are set to nullptr.

Any preferences? I can't think of any significant differences between what the two options enabled, but maybe I'm overlooking something.

onnxruntime/core/providers/cuda/cuda_provider_factory.cc
```diff
@@ -342,15 +705,84 @@ struct CudaEpFactory : OrtEpFactory {
...
                                             size_t* p_num_ep_devices) noexcept {
     size_t& num_ep_devices = *p_num_ep_devices;
-    auto* factory = static_cast<CudaEpFactory*>(this_ptr);
+    auto& factory = *static_cast<CudaEpFactory*>(this_ptr);
+
+    int num_cuda_devices = 0;
+    cudaGetDeviceCount(&num_cuda_devices);
+    RETURN_IF_ERROR(factory.CreateMemoryInfoForDevices(num_cuda_devices));
+
+    /* in theory we can match on the LUID in the OrtHardwareDevice metadata, but that requires the CUDA Driver API
```
skottmckay, line 721: If there's a different way to match the device to the Windows hardware info from SetupApi/Dxgi/Dxcore let me know.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
skottmckay: One aspect that would be clearer if we have additional functions on OrtEp is that logger to use.

In general we always want to prefer the session logger. CreateEp receives the session logger. OrtEpFactory will (soon) have the default logger). An implementation having to know if should use the logger from OrtEp when the `ep` argument is provided is not overly intuitive so I'm leaning towards adding extra functions to OrtEp for creating per-session allocators and streams.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -67,6 +66,115 @@ typedef struct OrtDataTransferImpl {
...
+   * \return The handle of the stream.
+   *
+   * \since Version 1.23.
+   */
+  ORT_API_T(void*, GetHandle, _In_ void* this_ptr);
+
+  /** \brief Create an OrtSyncNotificationImpl for the OrtSyncStream
+   *
+   * \param[in] this_ptr Pointer to the OrtSyncStreamImpl instance
+   * \param[in] stream The OrtSyncStream instance to create the notification for.
```
jywu-mysoft: doc doesn't match the signature?  there's no stream and num_consumers params?

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
jywu-mysoft: my preference is without the Impl for public c api

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
jywu-mysoft: without Impl is my preference.

onnxruntime/core/providers/cuda/cuda_provider_factory.cc
```diff
@@ -342,15 +705,84 @@ struct CudaEpFactory : OrtEpFactory {
                                             size_t max_ep_devices,
                                             size_t* p_num_ep_devices) noexcept {
     size_t& num_ep_devices = *p_num_ep_devices;
-    auto* factory = static_cast<CudaEpFactory*>(this_ptr);
+    auto& factory = *static_cast<CudaEpFactory*>(this_ptr);
+
+    int num_cuda_devices = 0;
+    cudaGetDeviceCount(&num_cuda_devices);
+    RETURN_IF_ERROR(factory.CreateMemoryInfoForDevices(num_cuda_devices));
```
pgonegandla, line 719: Should this be the min of num_cuda_devices and num_devices instead?
The loop below where CreateEpDevice() is called seems to do `const OrtMemoryInfo* gpu_mem_info = factory.gpu_memory_infos[device_id].get();` for all GPU devices.
This might go out of bounds if the memory is created only for the count of CUDA GPU devices, in a case like we have non-zero CUDA GPU devices and non-zero non-CUDA GPU devices.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -5383,10 +5389,18 @@ struct OrtApi {
    * \since Version 1.23
    */
   ORT_API2_STATUS(CreateMemoryInfo_V2, _In_ const char* name, _In_ enum OrtMemoryInfoDeviceType device_type,
-                  _In_ uint32_t vendor_id, _In_ int16_t device_id, _In_ enum OrtDeviceMemoryType mem_type,
+                  _In_ uint32_t vendor_id, _In_ int32_t device_id, _In_ enum OrtDeviceMemoryType mem_type,
                   _In_ size_t alignment, enum OrtAllocatorType allocator_type,
                   _Outptr_ OrtMemoryInfo** out);
 
+  /** \brief Get the device memory type from ::OrtMemoryInfo
```
adrianlizarraga, line 5396: nit: remainder of documentation

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -5383,10 +5389,18 @@ struct OrtApi {
...
+   */
+  ORT_API2_STATUS(MemoryInfoGetDeviceMemType, _In_ const OrtMemoryInfo* ptr, _Out_ OrtDeviceMemoryType* out);
+
+  /** \brief Get the vendor id from ::OrtMemoryInfo
```
adrianlizarraga, line 5407: nit: remainder of documentation

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -6163,6 +6180,140 @@ struct OrtApi {
...
+                  _Out_writes_(num_inputs) const OrtMemoryInfo** inputs_memory_info,
+                  _In_ size_t num_inputs);
+
+  /** \brief Get the OrtMemoryInfo for each output of the session.
+   *
+   * The memory info can be used to determine the device the output tensors are produced on.
+   * The user can pre-allocate an OrtValue using this information or use IOBinding to keep the data on the device.
+   * ORT will copy the output to CPU otherwise.
+   *
+   * The session must be fully initialized before calling this function as the input locations are not known until
```
```suggestion
   * The session must be fully initialized before calling this function as the output locations are not known until
```

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -6163,6 +6180,140 @@ struct OrtApi {
...
+   * \since Version 1.23
+   */
+  ORT_API_T(void*, SyncStream_GetHandle, _In_ OrtSyncStream* stream);
+
+  ORT_CLASS_RELEASE(SyncStream);
+
+  /** \brief Copy OrtValue instances containing Tensors between devices.
+   *
+   * The overall copy must be between two devices.
+   * i.e. all src_tensors and dst_tensors must have the same OrtMemoryInfo.
```
adrianlizarraga: nit: perhaps something like `all src_tensors must have the same source OrtMemorInfo, all dst_tensors must have the same destination OrtMemoryInfo.`

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -67,6 +66,115 @@ typedef struct OrtDataTransferImpl {
...
+  /** \brief Release the OrtSyncNotificationImpl instance.
+   *
+   * This is called by ORT when the OrtSyncNotificationImpl instance is no longer needed.
+   * The implementation should release any resources held by the instance.
+   *
+   * \param[in] this_ptr Pointer to the OrtSyncNotificationImpl instance.
+   *
+   * \since Version 1.23.
+   */
+  ORT_API_T(void, Release, _In_ void* this_ptr);
```
adrianlizarraga: Just curious, is there a reason to prefer `void* this_ptr` versus `OrtSyncNotificationImpl* this_ptr`?  Perhaps it is more flexible as it doesn't require use of C++ inheritance to store state.

We've used the non-void version of "this" pointers for other structs. Should some of those be updated to use `void*` since it could be more flexible?

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -709,13 +819,46 @@ struct OrtEpFactory {
...
    * \param[in] this_ptr The OrtEpFactory instance.
-   * \param[out] data_transfer The created OrtDataTransferImpl instance.
+   * \param[out] data_transfer The created OrtDataTransferImpl instance. Set to nullptr if not required.
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \since Version 1.23.
+   */
+  ORT_API2_STATUS(CreateDataTransfer, _In_ OrtEpFactory* this_ptr,
+                  _Outptr_result_maybenull_ OrtDataTransferImpl** data_transfer);
```
adrianlizarraga, line 870: Since the caller can set this to null, should it have `_opt_`? i.e., `_Outptr_result_maybenull_opt_`? (not sure if that SAL2 def exists).

onnxruntime/core/framework/plugin_ep_stream.h
```diff
@@ -0,0 +1,112 @@
...
+    std::unique_ptr<Notification> plugin_notification;
+
+    auto* ort_status = CreateNotificationImpl(num_consumers, plugin_notification);
+    if (ort_status != nullptr) {
+      ORT_THROW("Failed to create Notification: [", OrtApis::GetErrorCode(ort_status), "] ",
+                OrtApis::GetErrorMessage(ort_status));
+    }
+
+    std::unique_ptr<synchronize::Notification> result(plugin_notification.release());
+    return result;
```
adrianlizarraga: I think should be able to just `return plugin_notification` since std::unique_ptr can auto cast to base class.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -6163,6 +6180,140 @@ struct OrtApi {
...
+   * \param[in] stream Optional OrtSyncStream that can be used to perform the copy asynchronously. May be nullptr.
+   * \param[in] num_tensors The number of tensors to copy. The size of `src_tensors` and `dst_tensors` must match.
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \since Version 1.23
+   */
+  ORT_API2_STATUS(CopyTensors, _In_ const OrtEnv* env,
+                  _In_reads_(num_tensors) const OrtValue** src_tensors,
+                  _In_reads_(num_tensors) OrtValue** dst_tensors,
```
adrianlizarraga: This function doesn't replace the pointers in the src and dst arrays, so perhaps those could be marked const? i.e.,

```suggestion
                  _In_reads_(num_tensors) const OrtValue* const* src_tensors,
                  _In_reads_(num_tensors) OrtValue* const* dst_tensors,
```

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
skottmckay: I think I started with what you're suggesting but came to the conclusion that the 'result maybenull' meant the caller could set it to nullptr, and that an 'opt' would indicate the input arg may be nullptr (optional param) and that would indicate the implementer would need to check `data_transfer != nullptr` before modifying it. 

I'm not a SAL expert though so maybe my interpretation is incorrect.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
skottmckay: I think the non-void is clearer as Release should always be given the address of the OrtSyncNotificationImpl returned by CreateNotification.

onnxruntime/core/providers/cuda/cuda_provider_factory.cc
skottmckay: The loop filters on the vendor id being nvidia (0x10d3). Is it possible to have an nvidia GPU that does not show up in cudaGetDeviceCount?

pgonegandla: I think yes. I am running a sample app where I see that the num_devices is three (two NVIDIA GPU and a CPU), but the number of CUDA devices obtained from cudaGetDeviceCount(&num_cuda_devices); is 1.
This is causing a crash with the expression 'vector subscript out of range'.

onnxruntime/core/framework/plugin_ep_stream.h
skottmckay: This class is on the ORT side and is implementing ORT internals. The EP side provides the OrtSyncNotificationImpl.

onnxruntime/core/framework/plugin_ep_stream.h
skottmckay: This is the existing internal ORT class. A public type to allow sync outside of ORT is TBD and would have more generic naming.

adrianlizarraga (approved):

jywu-msft (approved):

onnxruntime/core/providers/cuda/cuda_provider_factory.cc
skottmckay: Will address in smaller PR

snnn:

PR #25261, opened 2025-07-02
Make TRT plugins optional

gedoensmax:
### Description

The parser does no longer link agains the plugin library but also loads it dynamic. Due to that I think we should also make the library optional in ORT. @chilo-ms

### DIFF
--- a/onnxruntime/core/providers/nv_tensorrt_rtx/nv_allocator.h
+++ b/onnxruntime/core/providers/nv_tensorrt_rtx/nv_allocator.h
@@ -6,7 +6,6 @@
 
 #include "core/common/inlined_containers.h"
 #include "core/framework/allocator.h"
-#include "core/providers/shared_library/provider_api.h"
 #include <mutex>
 
 namespace onnxruntime {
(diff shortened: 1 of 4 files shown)


onnxruntime/core/providers/nv_tensorrt_rtx/nv_allocator.h
```diff
@@ -56,9 +57,13 @@ class CUDAPinnedAllocator : public IAllocator {
...
-                          OrtDevice(OrtDevice::GPU, OrtDevice::MemType::HOST_ACCESSIBLE, OrtDevice::VendorIds::NVIDIA,
-                                    device_id),
-                          OrtMemTypeCPUOutput)) {}
+                          OrtDevice(OrtDevice::CPU, OrtDevice::MemType::HOST_ACCESSIBLE, OrtDevice::VendorIds::NVIDIA,
+                                    0 /*CPU device always with id 0*/),
+                          OrtMemTypeCPUOutput)) {
+    if (device_id != 0) {
+      LOGS_DEFAULT(WARNING) << "[NvTensorRTRTX EP] Device ID was non 0 for a CudaPinned allocator. The ID will be ignored.";
+    }
+  }
```
gedoensmax: @skottmckay is there a reason the pinned allocator was changed to be a GPU allocator ? If so I can change it back, but since it is a normal CPU allocation that does not need any changes to be used by any other EP I did set it back to CPU.

skottmckay: Yes. The HOST_ACCESSIBLE is now what says the memory is CPU accessible. As the memory may be tied to a specific device (I believe that is the case for CUDA_PINNED memory) we were trying to make that association clearer.

skottmckay: We were going off documentation like [this](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DEVICE.html#group__CUDART__DEVICE_1g159587909ffa0791bbe4b40187a4c6bb) where it says "Any host memory allocated from this host thread using [cudaMallocHost()](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html#group__CUDART__MEMORY_1gab84100ae1fa1b12eaca660207ef585b) or [cudaHostAlloc()](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html#group__CUDART__MEMORY_1gb65da58f444e7230d3322b6126bb4902) or [cudaHostRegister()](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html#group__CUDART__MEMORY_1ge8d5c17670f16ac4fc8fcb4181cb490c) will have its lifetime associated with device."

gedoensmax: Ok that makes sense let me revert my change. I was not aware that this is the case to be honest.

chilo-ms (approved):

Issue #25264, opened 2025-07-02
Incorrect cubic resizing with antialias on CUDA


amancini-N:
### Describe the issue

I spotted a bug in Resize operation when using cubic filter and antialiasing on CUDA EP. Essentially, 2 problems happen:
- the CUDA kernel setup for level 1 resizing is not correct. If it's upsampling, we have too many blocks per grid. If it's downsampling, we get an uncomplete output
- Antialiasing is ignoring the passed cubic coefficient and using an hardcoded constant instead.

CPU EP output is correct

### To reproduce

Use following snippet:
```python
import io

import numpy as np
import onnx
import onnxruntime as ort
from PIL import Image, ImageDraw

def create_dummy_image():

    # Create a 64x64 grayscale image
    size = (64, 64)
    img = Image.new('L', size, 0)  # Start with black background
    draw = ImageDraw.Draw(img)

    # Define two triangles that divide the image
    # Triangle 1: Upper triangle (lighter gray)
    triangle1 = [(0, 0), (64, 0), (32, 32)]
    # Triangle 2: Lower triangle (darker gray)
    triangle2 = [(0, 64), (64, 64), (32, 32)]

    # Draw the triangles with different gray values
    draw.polygon(triangle1, fill=180)  # Light gray
    draw.polygon(triangle2, fill=80)   # Dark gray

    # Convert to numpy array
    arr = np.array(img).reshape((1, 1, 64, 64)).astype(np.uint8)

    return arr

def create_model(bytes_io):
    node = onnx.helper.make_node(
        'Resize',
        inputs=['X', '', '', 'sizes'],
        outputs=['Y'],
        mode='cubic',
        axes=[2, 3],
        antialias=1,
        coordinate_transformation_mode="pytorch_half_pixel",
        cubic_coeff_a=-0.5,
    )
    graph = onnx.helper.make_graph(
        [node],
        'test',
        inputs=[
            onnx.helper.make_tensor_value_info('X', onnx.TensorProto.UINT8, ['batch_size', 'n_channels', 'height', 'width']),
            onnx.helper.make_tensor_value_info('sizes', onnx.TensorProto.INT64, [2])
        ],
        outputs=[
            onnx.helper.make_tensor_value_info('Y', onnx.TensorProto.UINT8, ['batch_size', 'n_channels', 'r_height', 'r_width'])
        ]
    )
[...]

PR #25265, opened 2025-07-02
Fix antialias downsample on CUDA EP

amancini-N:
### Description

This PR addresses 3 issues:
- Compilation errors when DISABLE_CONTRIB_OPS flag is on
- Solve a CUDA compute kernel setup issue on Resize op with cubic filter and antialiasing
- Solve cubic_coeff_a parameter being ignored in CUDA kernel of Resize op

### Motivation and Context

#25264

### DIFF
--- a/onnxruntime/core/providers/cuda/tensor/resize_impl.h
+++ b/onnxruntime/core/providers/cuda/tensor/resize_impl.h
@@ -98,7 +98,7 @@ void ResizeAntiAliasImpl(
     int rank,
     const UpsampleMode upsample_mode,
     ResizeCoordinateTransformationMode coordinate_transform_mode,
-    const float cubic_coeff_a,
+    float cubic_coeff_a,
     gsl::span<const int64_t> input_shape,
     gsl::span<const int64_t> output_shape,
     int64_t batch_size, int64_t num_channels,


onnxruntime/core/providers/cuda/tensor/resize_impl.h
```diff
@@ -98,6 +98,7 @@ void ResizeAntiAliasImpl(
     int rank,
     const UpsampleMode upsample_mode,
     ResizeCoordinateTransformationMode coordinate_transform_mode,
+    const float cubic_coeff_a,
```
yuslepukhin: > cons [](http://example.com/codeflow?start=3&length=5)

const is unnecessary

yuslepukhin:
I think this comment needs to be addressed: https://github.com/microsoft/onnxruntime/issues/25264#issuecomment-3032329374

amancini-N:
> I think this comment needs to be addressed: [#25264 (comment)](https://github.com/microsoft/onnxruntime/issues/25264#issuecomment-3032329374)

As I replied in the discussion, I don't think the proposed example is correct

onnxruntime/core/providers/cuda/tensor/resize_impl.h
amancini-N: removed

PR #21867, opened 2024-08-27
Support Smooth Softmax in GroupQueryAttention

tianleiwu:
### Description

Softmax (formula 1) is like the following:
```math
y_{i} = \frac{exp(x_{i})}{\sum_{i} exp(x_{i})}
```

**Smooth Softmax** (formula 2) is a modified version that  introduces a smooth factor like the following:
```math
s_{i} = \frac{exp(x_{i})}{1+ \sum_{i} exp(x_{i})}
```
Since exponential is prone to overflow or underflow, to get stable result, formula 3 can be used:
```math
s_{i} = \frac{exp(x_{i} + c)}{exp(c)+ \sum_{i} exp(x_{i} +c)}
```
where choice of constant c can like the following formula 4:
```math
c=-\max_{i} \{ x_i \}
```
or  apply a constraint that c <=0 like the following formula 5:

```math
c=-\max(0, \max_{i} \{ x_i \})
```
The latter one (formula 5) ensures that $s_{i}$ will fallback to formula 2 of when all elements are negative.

Here we implement the smooth softmax in flash attention for CUDA (requires Ampere or newer GPU) and MLAS (for CPU).

Note that the flash attention uses formula 4, and CPU implementation uses formula 5. The result are compatible as long as not all elements are much smaller than 0.

PR #25269, opened 2025-07-03
[CPU] GQA supports head_sink input for smooth softmax

tianleiwu:
### Description
It is an extension of [Smooth Softmax](https://github.com/microsoft/onnxruntime/pull/21867) feature. The difference is that each head has a learnable smooth factor that adding to the denominator of softmax. The smooth factor is like an extra element that joins the softmax.

The usage of the smooth factor in softmax is like the following:
```math
softmax_{i} = \frac{exp(x_{i})}{exp(s)+ \sum_{j} exp(x_{j})}
```

The head_sink is a float tensor with length of number of attention heads. For h-th head, `head_sink[h]` is used as smooth factor s. When head_sink is not provided, constant 0 is used as smooth factor s.

Changes:
- [x] Update operator spec to add an optional new input `head_sink`
- [x] Implement CPU (MLAS) kernel.
- [x] Update test_gqa_cpu.py to test it.

CUDA kernel will be updated later in a separate PR.


onnxruntime/core/mlas/lib/compute.cpp
```diff
@@ -880,11 +882,12 @@ Return Value:
 #else
         float Maximum = MlasReduceMaximumF32Kernel(Input, D);
 #endif
-        float NegativeMaximum = -Maximum;
-        if (SmoothSoftmax && NegativeMaximum > 0.0f) {
-            NegativeMaximum = 0.0f;
+        if (SmoothSoftmax && Sink > Maximum) {
+            Maximum = Sink;
         }
```
fs-eire, line 887: What the value of parameter `SmoothSoftmax` should be, when per head smooth softmax is used?

nenad1002 (approved):

onnxruntime/core/mlas/lib/compute.cpp
tianleiwu: It is true when either the attirbute is set or head_sink is provided. The sink value is loaded from head_sink for current head, or 0 if head_sink is not provided:

https://github.com/microsoft/onnxruntime/blob/60de045cdea5dc3822f5e680d9559c712e183d5f/onnxruntime/contrib_ops/cpu/bert/gqa_attention_base.h#L315-L317

PR #25272, opened 2025-07-03
Add OrtEpFactory::GetVersion and store EP version in EP metadata.

edgchen1:
### Description

Add OrtEpFactory::GetVersion and store EP version in EP metadata.

### Motivation and Context

Enforce plugin EP version specification and make it accessible from EP metadata.


onnxruntime/core/session/abi_key_value_pairs.h
```diff
@@ -18,28 +18,31 @@ struct OrtKeyValuePairs {
...
-  void Add(const std::string& key, const std::string& value) {
+  bool Add(const std::string& key, const std::string& value) {
     if (key.empty()) {  // ignore empty keys
-      return;
+      return false;
     }
 
-    auto iter_inserted = entries.insert({key, value});
-    bool inserted = iter_inserted.second;
+    auto [it, inserted] = entries.insert({key, value});
```
edgchen1: `insert` won't overwrite. maybe it should be `insert_or_assign`.

edgchen1: addressing this in #25284

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -659,6 +663,17 @@ struct OrtEpFactory {
    */
   void(ORT_API_CALL* ReleaseEp)(OrtEpFactory* this_ptr, struct OrtEp* ep);
 
+  /** \brief Get the version of the execution provider that the factory creates.
+   *
+   * The returned string should be a null-terminated, UTF-8 encoded string. ORT will copy it.
```
skottmckay, line 671: Should we say that the string should be semver compliant?

onnxruntime/test/autoep/library/ep_factory.cc
```diff
@@ -107,7 +114,6 @@ OrtStatus* ORT_API_CALL ExampleEpFactory::GetSupportedDevicesImpl(OrtEpFactory*
       factory->ort_api.CreateKeyValuePairs(&ep_options);
 
       // random example using made up values
-      factory->ort_api.AddKeyValuePair(ep_metadata, "version", "0.1");
```
skottmckay, line 110: can we have a different random example of adding to metadata?

skottmckay (approved):

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
edgchen1: hm, I'd be inclined to leave the versioning scheme up the EP author. unless we have a reason to expect semver somewhere else?

skottmckay: makes it easier for users to parse the version if we ask for semver. e.g. in webnn they want to look for specific versions of an EP, and we'll want to do the same in foundry local. if we have to process completely different formats for different EPs that translates to a lot of EP specific code.

is there a reason why it would be hard or limiting for the EP to conform to semver?

edgchen1: updated to require and enforce semver

onnxruntime/core/session/ep_api.cc
```diff
@@ -34,6 +37,18 @@ ORT_API_STATUS_IMPL(CreateEpDevice, _In_ OrtEpFactory* ep_factory,
...
   }
 
+  if (ep_device->ep_metadata.Entries().find(kOrtEpDevice_EpMetadataKey_Version) !=
+      ep_device->ep_metadata.Entries().end()) {
+    return OrtApis::CreateStatus(ORT_INVALID_ARGUMENT,
+                                 "The provided EP metadata should not explicitly specify the EP version.");
+  }
+
+  {
+    std::string ep_version = ep_factory->GetVersion(ep_factory);
```
skottmckay: Should this check the ORT version the EP was compiled against before calling the new function (ort_version_supported)?

skottmckay (dismissed):

jywu-msft (approved):

onnxruntime/core/providers/cuda/cuda_provider_factory.cc
```diff
@@ -329,6 +330,10 @@ struct CudaEpFactory : OrtEpFactory {
     return factory->vendor.c_str();
   }
 
+  static const char* ORT_API_CALL GetVersionImpl(const OrtEpFactory* this_ptr) noexcept {
```
jywu-mysoft: build warning, unused param this_ptr

PR #25277, opened 2025-07-03
Add a new ORT API `GetSessionOptionConfigEntries`

wcy123:
### Description

Add a new ORT API `GetSessionOptionConfigEntries`.

### Motivation and Context

#24887 allows plugin-EPs to interface with ORT using a binary stable interface. #24445 allows an EP to handle the extraction of EP options from the session option configurations.  For an EP like VitisAI EP to comply with the requirements,

it is necessary for a plugin-EPs to access all config entries in a session option.

```c++
    OrtKeyValuePairs * kvps = nullptr;
    auto status = GetSessionOptionConfigEntries(session_option, &kvps);
    if(status) {
        throw status;
    }
    std::unique_ptr<OrtKeyValuePairs, void (*)(OrtKeyValuePairs*)>
        config_entries(kvps,
                       ort_api.ReleaseKeyValuePairs);
    const char* const* keys = nullptr;
    const char* const* values = nullptr;
    size_t num_keys = 0;
    // Get keys and values from the config entries
    Ort::GetApi().GetKeyValuePairs(config_entries.get(), &keys, &values, &num_keys);

    for (size_t i = 0; i < num_keys; ++i) {
        // process keys[i] and values[i]
    }
```

### DIFF
--- a/include/onnxruntime/core/session/onnxruntime_c_api.h
+++ b/include/onnxruntime/core/session/onnxruntime_c_api.h
@@ -6074,6 +6074,16 @@ struct OrtApi {
    * \since Version 1.23
    */
   ORT_API2_STATUS(GetTensorData, _In_ const OrtValue* value, _Outptr_ const void** out);
+
+  /** \brief Get Session configuration entries.
+   *
+   * \param[in] options The session options.
+   * \return An OrtKeyValuePairs instance containing the all session config enties.
+   *         Note: the user should call OrtApi::ReleaseKeyValuePairs.
+   *
+   * \since Version 1.22.
+   */
+  OrtKeyValuePairs*(ORT_API_CALL* GetSessionOptionConfigEntries)(_In_ const OrtSessionOptions* options);
 };
 
 /*
--- a/include/onnxruntime/core/session/onnxruntime_c_api.h
+++ b/include/onnxruntime/core/session/onnxruntime_c_api.h
@@ -6085,7 +6085,7 @@ struct OrtApi {
    *
    * \since Version 1.23.
    */
-  ORT_API2_STATUS(GetSessionOptionConfigEntries, _In_ const OrtSessionOptions* options, _Outptr_ OrtKeyValuePairs** out);
+  ORT_API2_STATUS(GetSessionOptionsConfigEntries, _In_ const OrtSessionOptions* options, _Outptr_ OrtKeyValuePairs** out);
 };
 
 /*
--- a/include/onnxruntime/core/session/onnxruntime_c_api.h
+++ b/include/onnxruntime/core/session/onnxruntime_c_api.h
@@ -6078,12 +6078,14 @@ struct OrtApi {
   /** \brief Get Session configuration entries.
    *
    * \param[in] options The session options.
-   * \return An OrtKeyValuePairs instance containing all session configuration entries.
-   *         Note: the user should call OrtApi::ReleaseKeyValuePairs.
+   * \param[out] out A pointer to a newly created OrtKeyValuePairs instance.
+   *
+   *  An OrtKeyValuePairs instance containing all session configuration entries.
+   *  Note: the user should call OrtApi::ReleaseKeyValuePairs.
    *
    * \since Version 1.23.
    */
-  OrtKeyValuePairs*(ORT_API_CALL* GetSessionOptionConfigEntries)(_In_ const OrtSessionOptions* options);
+  ORT_API2_STATUS(GetSessionOptionConfigEntries, _In_ const OrtSessionOptions* options, _Outptr_ OrtKeyValuePairs** out);
 };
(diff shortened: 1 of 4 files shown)


include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -6068,6 +6068,16 @@ struct OrtApi {
...
    */
   ORT_API2_STATUS(GetTensorData, _In_ const OrtValue* value, _Outptr_ const void** out);
+
+  /** \brief Get Session configuration entries.
+   *
+   * \param[in] options The session options.
+   * \return An OrtKeyValuePairs instance containing all session configuration entries.
+   *         Note: the user should call OrtApi::ReleaseKeyValuePairs.
+   *
+   * \since Version 1.22.
```
```suggestion
   * \since Version 1.23.
``` #Closed

File: include/onnxruntime/core/session/onnxruntime_c_api.h [hunk 2]
```diff
@@ -6068,6 +6068,16 @@ struct OrtApi {
...
+   */
+  OrtKeyValuePairs*(ORT_API_CALL* GetSessionOptionConfigEntries)(_In_ const OrtSessionOptions* options);
```
edgchen1: can it return an OrtStatus to provide a way to indicate an error?

onnxruntime/core/session/abi_session_options.cc
```diff
@@ -278,6 +278,18 @@ ORT_API_STATUS_IMPL(OrtApis::GetSessionConfigEntry, _In_ const OrtSessionOptions
...
+ORT_API(OrtKeyValuePairs*, OrtApis::GetSessionOptionConfigEntries, _In_ const OrtSessionOptions* options) {
+  if (!options) {
+    return nullptr; // Return nullptr to indicate invalid input.
+  }
+  auto& config_options = options->value.config_options.GetConfigOptionsMap();
+  auto kvps = std::make_unique<OrtKeyValuePairs>();
+  for (auto& kv : config_options) {
+    kvps->Add(kv.first.c_str(), kv.second.c_str());
+  }
+  return reinterpret_cast<OrtKeyValuePairs*>(kvps.release());
```
edgchen1: need to ensure that exceptions will not be thrown from this function. `API_IMPL_BEGIN` and `API_IMPL_END` are helper macros to ensure this that work with OrtStatus returning API functions.

chilo-ms:
Do we want to make it more flexible by adding an additional key's prefix as a parameter, so that EP can get more useful session configs back, ex: "ep.vitis_ep.XXX", instead of all the session configs?
If key's prefix is provided as nullptr, return all the session configs.

include/onnxruntime/core/session/onnxruntime_c_api.h
wcy123: Yes, I update the source code according to your comment.

onnxruntime/core/session/abi_session_options.cc
wcy123: `API_IMPL_BEGIN` and `API_IMPL_END` are added.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -6068,6 +6068,18 @@ struct OrtApi {
...
+   *
+   * \param[in] options The session options.
+   * \param[out] out A pointer to a newly created OrtKeyValuePairs instance.
+   *
+   *  An OrtKeyValuePairs instance containing all session configuration entries.
+   *  Note: the user should call OrtApi::ReleaseKeyValuePairs.
+   *
+   * \since Version 1.23.
+   */
+  ORT_API2_STATUS(GetSessionOptionConfigEntries, _In_ const OrtSessionOptions* options, _Outptr_ OrtKeyValuePairs** out);
```
skottmckay:
```suggestion
  ORT_API2_STATUS(GetSessionOptionsConfigEntries, _In_ const OrtSessionOptions* options, _Outptr_ OrtKeyValuePairs** out);
```

wcy123:
> Do we want to make it more flexible by adding an additional key's prefix as a parameter, so that EP can get more useful session configs back, ex: "ep.vitis_ep.XXX", instead of all the session configs? If key's prefix is provided as nullptr, return all the session configs.

VitisAI EP would handle some other configuration entries, i.e. "ep.shared_ep_context", "ep.context_file_path", "ep.stop_share_ep_contexts" etc. Additionally it would be easy for an EP plugin to filter out the prefix, soreturning all session configuration could be fine.

include/onnxruntime/core/session/onnxruntime_c_api.h
wcy123: renaming is done.

HectorSVC (approved):
:shipit:

PR #25278, opened 2025-07-03
Fix INT32 bias overflow in QOperator INT8 symmetric quantization by adjusting weight scale and requantizing

Bonoy0328:
### Overview

This PR introduces a critical fix for **QOperator INT8 symmetric quantization** in ONNX Runtime. It addresses a situation where the computed **bias scale** (`input_scale * weight_scale`) becomes too small, leading to **int32 overflow** or **precision clipping** during bias quantization.

### Problem

In symmetric quantization (i.e., zero_point = 0), the bias tensor is quantized using a fixed-point scale:
**bias_scale = input_scale * weight_scale**

When this value is too small, the quantized int32 bias may exceed the range of `int32`, causing saturation or significant quantization error. This was observed to cause **>51% accuracy loss** in some models.

### Solution

This PR adds two new functions to mitigate this:

---

#### 🔧 `_adjust_weight_scale_for_int32_bias(...)`

Located in `onnx_quantizer.py`, this function:

- **Inspects the float bias range** to compute the smallest valid bias scale (based on int32 dynamic range)
- **Compares** this threshold against `input_scale * weight_scale`
- If too small, **scales up the weight scale** accordingly, to prevent overflow
- Supports both per-tensor and per-channel weight quantization cases

This logic is **only triggered when**:
- The weight's zero point is exactly zero (i.e. symmetric)
- The weight data type is `INT8` or `INT16`

---

#### 🔄 `_requantize_weight(...)`

After weight scale adjustment, this function:
- **Finds the original quantized weight** (`q_weight`), scale, and zero point from the initializer list
- **Removes** the outdated quantized weight and scale
- **Re-quantizes** the original float weights using the new scale and the same zero point
- **Re-inserts** them into the model to maintain consistency

---

### Summary of Benefits

- ✅ Prevents int32 overflow or saturation during symmetric bias quantization
- ✅ Ensures weight and bias quantization remain consistent
- ✅ Reduced quantization error from >51.4% to ~3% in test models
- ✅ Fix is limited in scope to QOperator + symmetric INT8/INT16 flow (safe for other modes)
- ✅ Improves robustness of static quantization for hardware that performs integer-only inference

---

### Code Location

- `onnxruntime/quantization/onnx_quantizer.py`
  - `def _adjust_weight_scale_for_int32_bias(...)`
  - `def _requantize_weight(...)`
  - Integrated in `quantize_bias_static(...)`

---

Please let me know if you'd like additional test coverage or integration points. Thanks!

### DIFF
--- a/onnxruntime/python/tools/quantization/onnx_quantizer.py
+++ b/onnxruntime/python/tools/quantization/onnx_quantizer.py
@@ -785,12 +785,12 @@ class ONNXQuantizer(BaseQuantizer):
         weight_zero_point = (
             onnx.numpy_helper.to_array(weight_zp_init) if weight_zp_init is not None else None
         )
-        is_per_channel = self.quantized_value_map[weight_name].axis is not None
+        is_per_channel = self.per_channel
         if (
             weight_zero_point is not None
             and weight_zero_point.size
             and not weight_zero_point.any()
-            and self.weight_qType in (onnx_proto.TensorProto.INT8)
+            and self.weight_qType in (onnx_proto.TensorProto.INT8,)
         ):
             bias_initializer = find_by_name(bias_name, self.model.initializer())
             did_update, new_weight_scale = self._adjust_weight_scale_for_int32_bias(
--- a/onnxruntime/python/tools/quantization/onnx_quantizer.py
+++ b/onnxruntime/python/tools/quantization/onnx_quantizer.py
@@ -30,6 +30,7 @@ from .quant_utils import (
     ms_domain,
     save_and_reload_model_with_shape_infer,
     tensor_proto_to_array,
+    quantize_onnx_initializer,
 )
 from .registry import CreateOpQuantizer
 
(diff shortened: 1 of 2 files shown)


Bonoy0328:

onnxruntime/python/tools/quantization/onnx_quantizer.py
```diff
@@ -660,6 +762,31 @@ def quantize_bias_static(self, bias_name, input_name, weight_name, beta=1.0):
...
+        weight_zp_init = find_by_name(weight_zp_name, self.model.initializer())
+        weight_zero_point = (
+            onnx.numpy_helper.to_array(weight_zp_init) if weight_zp_init is not None else None
+        )
+        is_per_channel = self.quantized_value_map[weight_name].axis is not None
+        if (
+            weight_zero_point is not None
+            and weight_zero_point.size
+            and not weight_zero_point.any()
+            and self.weight_qType in (onnx_proto.TensorProto.INT8, onnx_proto.TensorProto.INT16)
```
jiafatom: I thought `QLinearConv` and `QGemm` only support `int8` `uint8`
(1) do we need `int16` here?
(2) can we support `uint8` here?

onnxruntime/python/tools/quantization/onnx_quantizer.py
```diff
@@ -635,6 +636,107 @@ def find_quantized_value(self, input_name):
...
+        """Checks if the bias scale is too small and increases the weight scale if needed."""
+
+        if not weight_scale.size:
+            return False, None
+
+        bias_float_data = tensor_proto_to_array(bias_tp)
+
+        int32_info = np.iinfo(np.int32)
+        multiplicative_epsilon = 1.0001
+        qrange = np.array(int32_info.max, dtype=np.float64) - np.array(int32_info.min + 1, dtype=np.float64)
```
jiafatom, line 692: For `qrange`, is it related to the type of `weight_scale`? Why this is `int32` here?

onnxruntime/python/tools/quantization/onnx_quantizer.py
```diff
@@ -635,6 +636,107 @@ def find_quantized_value(self, input_name):
...
+    ) -> tuple[bool, np.ndarray | None]:
+        """Checks if the bias scale is too small and increases the weight scale if needed."""
+
+        if not weight_scale.size:
+            return False, None
+
+        bias_float_data = tensor_proto_to_array(bias_tp)
+
+        int32_info = np.iinfo(np.int32)
+        multiplicative_epsilon = 1.0001
```
jiafatom, line 691: Can you explain more about 1.0001 here? Could you write the equation here to help me understand why smallest valid scale is `multiplicative_epsilon * (2.0 * absmax) / qrange` ?

onnxruntime/python/tools/quantization/onnx_quantizer.py
```diff
@@ -635,6 +636,107 @@ def find_quantized_value(self, input_name):
...
+
+            if (bias_candidate_scale < bias_smallest_valid_scale) and (bias_candidate_scale > 0.0):
+                ratio = bias_smallest_valid_scale / bias_candidate_scale
+                logging.info(
+                    f"Increasing scale for weight `{weight_name}` by the ratio {ratio} to "
+                    f"ensure bias `{bias_tp.name}` has a valid scale."
+                )
+                new_scale = weight_scale_fp64 * ratio
+                weight_scale = np.array(new_scale, dtype=weight_scale_dtype)
+                updated = True
```
jiafatom, line 712: Can we write a function for this code?
from `bias_smallest_valid_scale = multiplicative_epsilon * (2.0 * absmax) / qrange`
to `updated=True`
we just need something like
`weight_scale = func()`
```
def func():
            bias_smallest_valid_scale = multiplicative_epsilon * (2.0 * absmax) / qrange

            input_scale_fp64 = np.array(input_scale.item(), dtype=np.float64)
            weight_scale_fp64 = np.array(weight_scale.item(), dtype=np.float64)
            bias_candidate_scale = input_scale_fp64 * weight_scale_fp64

            if (bias_candidate_scale < bias_smallest_valid_scale) and (bias_candidate_scale > 0.0):
                ratio = bias_smallest_valid_scale / bias_candidate_scale
                logging.info(
                    f"Increasing scale for weight `{weight_name}` by the ratio {ratio} to "
                    f"ensure bias `{bias_tp.name}` has a valid scale."
                )
                new_scale = weight_scale_fp64 * ratio
                weight_scale = np.array(new_scale, dtype=weight_scale_dtype)
                updated = True
return updated, weight_scale
```

jiafatom:
Thank you for this important fix! Can we have some unit test for this code change?

onnxruntime/python/tools/quantization/onnx_quantizer.py
Bonoy0328: (1) You're right — we don't actually need INT16 here. The current implementation doesn't support QLinearConv or QGemm with INT16 weights, so this can be safely removed.
(2) As for UINT8, it's conditionally supported — specifically when the weight zero point is all zeros. This corresponds to symmetric quantization, where the bias scale adjustment is valid. However, typical UINT8 weight quantization uses an asymmetric zero point (usually 127), in which case the bias scale adjustment is no longer applicable. Supporting generic UINT8 weights would require additional handling due to the non-zero zero point.

onnxruntime/python/tools/quantization/onnx_quantizer.py
Bonoy0328: As shown in lines 261–266 of onnxruntime/python/tools/quantization/base_quantizer.py :

```python
# Clip quantized data to the range of a int32
int32_min = np.float64(np.iinfo(np.int32).min)
int32_max = np.float64(np.iinfo(np.int32).max)
if np.any(quantized_data < int32_min) or np.any(quantized_data > int32_max):
    logging.warning(
        f"Quantized bias `{bias_name}` exceeds the range of a int32. The bias scale is too small."
    )
```
So qrange is computed based on the representable range of int32. This constant is independent of the data type of weight_scale (which is typically float). The purpose of the helper function is to check whether input_scale * weight_scale might lead to int32 bias overflow; if so, it increases the weight scale.

Therefore, qrange does not depend on weight_scale’s type, but rather reflects how many values an int32 bias tensor can represent. The code uses it to ensure the computed bias scale prevents quantized bias values from exceeding int32 limits.

onnxruntime/python/tools/quantization/onnx_quantizer.py
Bonoy0328: Bias is quantized using:

$$
\mathrm{bias\_quantized} = \frac{\mathrm{bias\_fp32}}{\mathrm{input\_scale} \cdot \mathrm{weight\_scale}}
$$

This value is cast to int32. To avoid overflow:

$$
\left| \frac{\mathrm{bias}}{\mathrm{input\_scale} \cdot \mathrm{weight\_scale}} \right| < \mathrm{int32\_max}
$$

Rearranged:

$$
\mathrm{input\_scale} \cdot \mathrm{weight\_scale} > \frac{|\mathrm{bias}|}{\mathrm{int32\_max}}
$$

More conservatively:

$$
\mathrm{bias\_scale} > \frac{2 \cdot |\mathrm{bias}|}{\mathrm{qrange}}
$$

To avoid floating-point precision issues, we scale it slightly up:

$$
\mathrm{bias\_scale} = 1.0001 \cdot \frac{2 \cdot |\mathrm{bias}|}{\mathrm{qrange}}
$$

1.0001 is an empirical value used as the lower limit of the "safe amplification" bias scale in floating-point quantization

It is used to prevent the floating-point division calculation result from overflowing int32 after being amplified by too small a scale. 

For example In TensorFlow quantization, this value is 1.00001, and in PyTorch quantizer scale check, this value is 1.0001

onnxruntime/python/tools/quantization/onnx_quantizer.py
Bonoy0328: Got it, I'll update it.

Bonoy0328:
@jiafatom Thank you for the review. I’ve made the updates according to your comments. Please take another look when you have time.

jiafatom:
Looks good to me! Thank you very much for the contribution! It can merge if the pipeline passes.

jiafatom (approved):

PR #25280, opened 2025-07-03
Fix Sign and Clip operation on int64 tensors

cbourjau:
### Description
The Clip and sign operators exhibit buggy behavior for some int64 values on Linux and Windows (see added test cases). Using [Eigen's Array-API](https://eigen.tuxfamily.org/dox/group__TutorialArrayClass.html#title0) appears to mitigate the issue. A failing test case on CI can be found [here](https://github.com/cbourjau/onnxruntime/actions/runs/16053021996/job/45300270472) (uses the conda-forge toolchain).

### DIFF
--- a/onnxruntime/core/providers/cpu/math/clip.cc
+++ b/onnxruntime/core/providers/cpu/math/clip.cc
@@ -131,7 +131,7 @@ struct Clip::ComputeImpl {
               ConstEigenVectorMap<T>(X->Data<T>() + start, count)
                   .array()
                   .unaryExpr([=](auto x) {
-                    return std::clamp(x, min_val, max_val);
+		    return std::min(max_val, std::max(x, min_val));
                   });
         },
         0);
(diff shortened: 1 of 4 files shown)


onnxruntime/core/providers/cpu/math/clip.cc
```diff
@@ -129,8 +129,10 @@ struct Clip::ComputeImpl {
 
           EigenVectorMap<T>(Y->MutableData<T>() + start, count) =
               ConstEigenVectorMap<T>(X->Data<T>() + start, count)
-                  .cwiseMax(min_val)
-                  .cwiseMin(max_val);
+                  .array()
+                  .unaryExpr([=](auto x) {
+                    return std::clamp(x, min_val, max_val);
```
tianleiwu: It will not get expected result when min_val > max_val. std::clamp will return x in this case, while onnx definition is `Min(max_value, Max(x, min_value))` will return max_value.

cbourjau:
Sorry, I lost sight of this PR. However, I was now able to fine the correct cause for this issue. The currenly-used Eigen version is buggy on SSE2 code paths. There is a fix upstream, but it is not part of any release, yet: https://gitlab.com/libeigen/eigen/-/merge_requests/2831

cbourjau:
I'm closing this in favor of https://github.com/microsoft/onnxruntime/pull/32475

PR #25282, opened 2025-07-03
Fix Windows build of cuda 12.8

tianleiwu:
### Description

Fix Windows cuda build that using cuda 12.8, where nvcc raises error that C:/DNDEBUG not found.

The root cause is current build script directly pass flags starting with /D to nvcc. It does not cause issue in CI build since it uses cuda 12.2, and we have a flag to let nvcc pass unknown flags to host compiler. That is not ideal since different version of nvcc may have different behavior on this.

Here we distinguish nvcc flags (starts with -) and MSVC flags (starts with /), and pass MSVC flags to host compiler explicitly.

### DIFF
--- a/tools/ci_build/build.py
+++ b/tools/ci_build/build.py
@@ -1155,7 +1155,8 @@ def generate_build_tree(
                         else:
                             if not compile_flag.startswith("/"):
                                 log.warning(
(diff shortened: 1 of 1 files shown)


snnn:
Probably it is related to #25246

snnn (approved):

PR #25284, opened 2025-07-03
`OrtKeyValuePairs` updates

edgchen1:
### Description

- Change the entries container type from `std::unordered_map` to `std::map`. This enables deterministic iteration order.

- Enforce internal container state consistency. `OrtKeyValuePairs` has several internal containers that must stay consistent. Previously, the internal containers were public. This change makes them private and also fixes the copy/move behavior.

- Add unit tests.

### Motivation and Context

Some fixes and improvements.

### DIFF
--- a/onnxruntime/core/session/abi_key_value_pairs.h
+++ b/onnxruntime/core/session/abi_key_value_pairs.h
@@ -39,10 +39,6 @@ struct OrtKeyValuePairs {
     Sync();
   }
 
-  void CopyFromMap(const std::unordered_map<std::string, std::string>& src) {
-    CopyFromMap(std::map<std::string, std::string>(src.begin(), src.end()));
-  }
-
   void Add(const char* key, const char* value) {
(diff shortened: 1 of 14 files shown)


onnxruntime/core/session/abi_key_value_pairs.h
```diff
@@ -4,38 +4,58 @@
...
 #include <algorithm>
+#include <map>
 #include <string>
-#include <vector>
 #include <unordered_map>
+#include <vector>
+
+#include "gsl/gsl"
 
 struct OrtKeyValuePairs {
```
skottmckay, line 14: Given the members are now private this is really a `class` isn't it?

edgchen1: doesn't really matter in C++. :) it's forward declared as a struct in `onnxruntime_c_api.h` so I think we probably shouldn't define it as a class here.

PR #25285, opened 2025-07-04
[webgpu] support smooth softmax for non-FA GQA implementation

fs-eire:
### Description

support smooth softmax for non-FA GQA implementation

This change depends on:
- #25269

Work items:

- [x] support smooth softmax
- [x] support bias
- [x] support head sink (per-head smooth softmax)

The following will not be included in this PR:
- support for FlashAttention
- support sliding window


onnxruntime/contrib_ops/cpu/bert/group_query_attention_helper.h
```diff
@@ -377,6 +378,18 @@ Status CheckCustomAttentionInputs(const T* position_ids,
     }
   }
 
+  if (head_sink != nullptr) {
```
guschmue, line 381: code later uses if (head_sink) ... else if (use_smooth_softmax_) ...
we should check that only one of them is used.

onnxruntime/contrib_ops/webgpu/bert/group_query_attention.cc
```diff
@@ -184,8 +194,8 @@ Status GroupQueryAttention::ComputeInternal(onnxruntime::webgpu::ComputeContext&
   Tensor* present_value = context.Output(2, present_kv_shape);
   parameters.past_present_share_buffer_ = present_key != nullptr && present_value != nullptr && past_key != nullptr && past_value != nullptr && past_key->DataRaw() == present_key->DataRaw() && past_value->DataRaw() == present_value->DataRaw();
 
-  if (!do_rotary_ && CanApplyFlashAttention(nullptr /* bias */, present_key, present_value, parameters, context)) {
-    return ApplyFlashAttention(query, key, value, nullptr /* attention_bias */, output, past_key, present_key, past_value,
+  if (!do_rotary_ && CanApplyFlashAttention(attention_bias, present_key, present_value, parameters, context)) {
```
guschmue: don't use FA if use_smooth_softmax_ is set

guschmue (approved):

PR #25286, opened 2025-07-04
[WebNN] Always create a new constant for zero_points

Honry:
MatMulNBits is a decomposed op in WebNN EP. Previously, we share the WebNN constant of zero_points if they have the same value and data type. However, this brings a lot of complexity for developers to fuse it back to MatMulNBits in the underlying WebNN implementation in Chromium.

In this PR, we will always create a new constant for zero_points.


Honry:

fdwr (approved):
👍

PR #25292, opened 2025-07-05
[EP ABI] Utility to serialize OrtGraph to GraphProto

adrianlizarraga:
### Description
- Provides utility functions that serialize an `OrtGraph` to a `GraphProto` or `ModelProto`.
  - Header-only file that can be copied to a project that builds with ORT and ONNX.
  - Available in [include/onnxruntime/core/providers/utils/ort_graph_to_proto.h](https://github.com/microsoft/onnxruntime/blob/adrianl/ep-abi-ort-graph-to-onnx-protobuf/include/onnxruntime/core/providers/utils/ort_graph_to_proto.h)
- Updates the `Node_GetSubgraphs` API function to also return the attribute names associated with each subgraph. This is required to determine which subgraph corresponds to a given attribute.
- Adds `Graph_GetNumOperatorSets` and `Graph_GetOperatorSets` API functions to get the opset version for each domain.

### Motivation and Context
Provide a utility to facilitate porting of existing execution providers to the new EP ABI. The utilities introduced by this PR convert an `OrtGraph` into an ONNX protobuf representation, which some existing EPs currently convert to their internal representation. Ideally, we would prefer a more direct conversion from a `OrtGraph` to the EP's internal representation, but this is a large effort. These utilities enable an incremental transition.


include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -5568,6 +5570,45 @@ struct OrtApi {
...
+   * an empty string.
+   *
+   * \param[in] graph The OrtGraph instance.
+   * \param[out] num_operator_sets Output parameter set to the number of operator sets that the graph's model uses.
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \since Version 1.23.
+   */
+  ORT_API2_STATUS(Graph_GetNumOperatorSets, _In_ const OrtGraph* graph, _Out_ size_t* num_operator_sets);
```
adrianlizarraga, line 5587: Note: "Operator set" is terminology from ONNX: https://github.com/onnx/onnx/blob/3236bf77da50c6bd365e7a3d0e79ef47acace640/onnx/onnx.proto#L888-L901

ORT also uses "domain opset versions". Not sure which is clearer.

jywu-mysoft:
Thanks @adrianlizarraga! we should discuss with @chilo-ms where the utility functions should live. there are several options.

chilo-ms (approved):

adrianlizarraga:
@chilo-ms @jywu-msft , I made this a header-only file that can be copied into a project that builds with ORT + ONNX. I moved it into `include/onnxruntime/core/providers/utils/ort_graph_to_proto.h` and added documentation comments describing how to include the file.

Please let me know if you have any feedback. Thanks.

include/onnxruntime/core/providers/utils/ort_graph_to_proto.h
```diff
@@ -0,0 +1,718 @@
...
+       offset = out_file.tellp();
+       location = external_file_path;
+       out_file.write(static_cast<const char*>(data), bytes);
+       out_file.flush();
+       is_external = true;  // True if is external initializer
+       return Ort::Status{nullptr};
+     }
+
+     ONNX_NAMESPACE::GraphProto graph_proto;
+     OrtEpUtils::OrtGraphToProto(test_graph->GetOrtGraph(), graph_proto, handle_initializer_data);
```
adrianlizarraga:
```suggestion
     OrtEpUtils::OrtGraphToProto(*ort_graph, graph_proto, handle_initializer_data);
```

adrianlizarraga:
Closing and re-opening because some checks are not running.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -5933,20 +5974,24 @@ struct OrtApi {
...
+   *                             attribute names that correspond to the subgraphs.
    *
    * \snippet{doc} snippets.dox OrtStatus Return Value
    *
    * \since Version 1.23.
    */
   ORT_API2_STATUS(Node_GetSubgraphs, _In_ const OrtNode* node,
-                  _Out_writes_(num_subgraphs) const OrtGraph** subgraphs, _In_ size_t num_subgraphs);
+                  _Out_writes_(num_subgraphs) const OrtGraph** subgraphs, _In_ size_t num_subgraphs,
+                  _Out_writes_(num_subgraphs) const char** attribute_names);
```
chilo-ms: nit: In some cases, user only needs the subgraphs and not cares about the subgraph's name. For example, in TRT EP, it needs to check whether all nodes in the subgraph are assigned to TRT EP. In order to use this API, user still needs to have `std::vector<const char*> attr_names(num_subgraphs)` but it's not used. User can't pass in `nullptr` and it will fail due to gsl::span in the implementation.

adrianlizarraga: Good point. I can make it optional so that the user can pass in nullptr.

chilo-ms (approved):

PR #2654, opened 2019-12-13
define private version of dxcore enum that is added in 19H1 SDK.

orilevari:
Use a privately defined enum value for the dxcore feature level so that we can build on a pool without 19H1 SDK.

In future iterations, onnxruntime build definitions will be reworked to allow for building with vs2019 which will allow us to update the release agent pool VM bootstrap scripts to pull down 19H1. For now this PR will allow us to run release pipelines in the meantime.

PR #25297, opened 2025-07-06
FIX: dxcore include when compiling with older Windows SDK

JulienMaille:
### Description
Confirmed to work with `Windows 10 SDK, version 1803 (10.0.17134.0)`

### DIFF
--- a/onnxruntime/core/platform/windows/device_discovery.cc
+++ b/onnxruntime/core/platform/windows/device_discovery.cc
@@ -35,8 +35,16 @@ using Microsoft::WRL::ComPtr;
 
 //// For DXCore info.
 #include <initguid.h>
+#if __has_include(<directx/dxcore.h>)
 #include <directx/dxcore.h>
+#else
+#include <dxcore.h>
+#endif
+#if __has_include(<directx/dxcore_interface.h>)
 #include <directx/dxcore_interface.h>
+#else
+#include <dxcore_interface.h>
+#endif
 #include <wil/com.h>
 
 #include "core/common/cpuid_info.h"


fs-eire:
```
D:\a\_work\onnxruntime\onnxruntime\onnxruntime\core\platform\windows\device_discovery.cc(38,10): fatal error C1083: Cannot open include file: 'directx/dxcore.h': No such file or directory
```

the change is failing for the existing build.

Could you share more information how this change fixed your build? Is there a header file path change for a specific range of SDK?

JulienMaille:
@fs-eire That's weird.
So 
- with `#include <directx/dxcore.h>` it will include `_deps\directx_headers-src\include\directx\dxcore.h`
- with `#include <dxcore.h>` I get `D:\Dev\onnxruntime\ort-1.23.0\onnxruntime\core\platform\windows\device_discovery.cc(38,10): error C1083: Cannot open include file: 'dxcore.h': No such file or directory`

Note that with `Windows SDK v17134` there's no dxcore.h while when compiling for `SDK v26100` it is present in`C:\Program Files (x86)\Windows Kits\10\Include\10.0.26100.0\um\dxcore.h`

I suppose your pipeline doesn't populate `_deps\directx_headers-src\`?

fs-eire:
Seems currently `_deps\directx_headers-src\` only exists when building with `--use_dml`.

@skottmckay @snnn do we want to always include "directx_headers" for Windows build so that ORT can build against older Windows SDK

JulienMaille:
@fs-eire Do you have a pipeline that builds with `--use_dml`? Because I'm currently testing and it seems I need the same fix with latest Windows SDK.

fs-eire:
I don't think we have an active DML CI pipeline.

JulienMaille:
For the record, These are my build parameters (when testing without forcing older sdk): `--config Release --build_dir 'D:\dev\onnxruntime\bld-x64-1.23.0-mimalloc-ov2025.2.0' --skip_submodule_sync --build_shared_lib --parallel --use_dml --use_mimalloc --compile_no_warning_as_error --skip_submodule_sync --skip_tests --skip_onnx_tests --cmake_extra_defines 'CMAKE_POLICY_VERSION_MINIMUM=3.5 onnxruntime_ENABLE_MEMLEAK_CHECKER=OFF' onnxruntime_BUILD_UNIT_TESTS=OFF CMAKE_INTERPROCEDURAL_OPTIMIZATION=ON --use_openvino MULTI:GPU,CPU`

JulienMaille:
I see other includes in dml
https://github.com/microsoft/onnxruntime/blob/0ccecf71f4347abac08660a984058da4bb07ae9c/onnxruntime/core/providers/dml/dml_provider_factory_creator.h#L15
https://github.com/microsoft/onnxruntime/blob/0ccecf71f4347abac08660a984058da4bb07ae9c/onnxruntime/core/providers/dml/dml_provider_factory.cc#L8
but if really needed we could have
```c++
#if __has_include(<directx/dxcore.h>)
#include <directx/dxcore.h>
#else
#include <dxcore.h>
#endif
#if __has_include(<directx/dxcore_interface.h>)
#include <directx/dxcore_interface.h>
#else
#include <dxcore_interface.h>
#endif
```
Also as @fs-eire said, `_deps\directx_headers-src\` are not only required for dml but also for device discovery.

JulienMaille:
@fs-eire Any chance to re-run the CI? Would be nice to have a pipeline that builds the dml plugin to show you what issue I'm trying to fix.

snnn:
If I remember correctly, DirectML EP did not really work good with older Windows SDKs. Like this: https://github.com/microsoft/onnxruntime/pull/2654

Newer version of Windows SDKs not only introduce new APIs, but also have bug fixes and provide new macros.  Usually people should always use the latest Windows SDK. And the compatibility for older Windows versions is controlled by the macros listed in https://learn.microsoft.com/en-us/cpp/porting/modifying-winver-and-win32-winnt?view=msvc-170 . In theory the macros should provide enough guarantee. I know it is not the case.  Anyway, I think we still should use the latest SDK, and find ways to do the compatibility testing.  Find a Windows 10 ver 1803 machine and run the tests, which should be enough. 

So, in my opinion, this PR is optional.

PR #25298, opened 2025-07-06
FIX printDebugInfo in ov_interface.cc

JulienMaille:
error C2181: illegal else without matching if

Fix compilation in debug due to {} mismatch

### DIFF
--- a/onnxruntime/core/providers/openvino/ov_interface.cc
+++ b/onnxruntime/core/providers/openvino/ov_interface.cc
@@ -57,452 +57,452 @@ void printDebugInfo(const ov::CompiledModel& obj) {
 }
 #endif
 
-  // Function to check if a given OV property is enabled
-  std::optional<bool> queryOVProperty(const std::string& property, const std::string& device_type) {
-    try {
-      // Get the property value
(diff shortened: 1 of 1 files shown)


fs-eire:
Please use clang-format to format the code

JulienMaille:
@fs-eire the whole file is weirdly formatted, regarding the bit I fixed I have followed the same formatting.
BTW do you have a pipeline that build with OpenVino in debug? This bug with {} should never have been committed.

fs-eire:
There is one. But I think the build pipeline only build release so the code is never included

JulienMaille:
I’m not sure how to format or indent my fix without conflicting with the formatting of the surrounding code.

fs-eire:
> I’m not sure how to format or indent my fix without conflicting with the formatting of the surrounding code.

I checked it locally. you are right, the file is not initially well formated. The Lint job only checks modified files. It's OK to commit the file after formatting a lot of lines - it should have been formatted before.

JulienMaille:
@fs-eire Done

fs-eire (approved):

PR #25305, opened 2025-07-07
[webgpu] a few optimizations to graph capture implementation

fs-eire:
### Description

1. rename `SessionState` to `GraphCaptureState`, since there is already one SessionState type in ORT.
2. optimize implementation of `ComputeContext::BufferManager()`


qjia7 (approved):
LGTM, thanks.

PR #25308, opened 2025-07-07
[EP ABI] Infer OrtDevice for plugin EP from registered OrtMemoryInfo

adrianlizarraga:
### Description
- Infer `OrtDevice` for a plugin EP from the registered `OrtMemoryInfo` for device memory.
- Fix potential `nullptr` dereference when a `PluginExecutionProvider` tries to log a message without a valid logger. Now, constructing a `PluginExecutionProvider` requires passing a valid logger.

### Motivation and Context
Address a `TODO` to properly set the `OrtDevice` for a `PluginExecutionProvider` instance.

### DIFF
--- a/onnxruntime/core/session/ep_plugin_provider_interfaces.cc
+++ b/onnxruntime/core/session/ep_plugin_provider_interfaces.cc
@@ -99,8 +99,19 @@ static OrtDevice GetOrtDeviceForPluginEp(gsl::span<const OrtEpDevice* const> ep_
 
   // Check assertion that all OrtEpDevice instances must have the same device_memory_info
   bool all_match = std::all_of(ep_devices.begin() + 1, ep_devices.end(),
-                               [&device_memory_info](const OrtEpDevice* ep_device) {
-                                 return ep_device->device_memory_info == device_memory_info;
+                               [mem_a = device_memory_info](const OrtEpDevice* ep_device) {
+                                 const OrtMemoryInfo* mem_b = ep_device->device_memory_info;
+
+                                 if (mem_a == mem_b) {
+                                   return true;  // Point to the same OrtMemoryInfo instance.
+                                 }
+
+                                 if (mem_a == nullptr || mem_b == nullptr) {
+                                   return false;  // One is nullptr and the other is not.
+                                 }
+
+                                 // Both non-null but point to different instances. Use operator==.
+                                 return *mem_a == *mem_b;
                                });
   if (!all_match) {
     ORT_THROW("Error creating execution provider '", ep_devices[0]->ep_name,
--- a/onnxruntime/core/session/ep_plugin_provider_interfaces.cc
+++ b/onnxruntime/core/session/ep_plugin_provider_interfaces.cc
@@ -84,17 +84,11 @@ struct PluginEpMetaDefNameFunctor {
 // PluginExecutionProvider
 //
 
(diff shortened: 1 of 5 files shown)


skottmckay (approved):

onnxruntime/core/session/ep_plugin_provider_interfaces.cc
```diff
@@ -86,10 +86,34 @@ struct PluginEpMetaDefNameFunctor {
 // PluginExecutionProvider
 //
 
+static OrtDevice GetOrtDeviceForPluginEp(gsl::span<const OrtEpDevice* const> ep_devices) {
+  // Get the OrtDevice from OrtEpDevice.device_memory_info if it is set. Otherwise, we set it to CPU.
+  // If there are multiple OrtEpDevice instances,the device_memory_info must be consistent for all.
```
adrianlizarraga:
```suggestion
  // If there are multiple OrtEpDevice instances, the device_memory_info must be consistent for all.
```

onnxruntime/core/session/ep_plugin_provider_interfaces.cc
```diff
@@ -86,10 +86,34 @@ struct PluginEpMetaDefNameFunctor {
 // PluginExecutionProvider
 //
 
+static OrtDevice GetOrtDeviceForPluginEp(gsl::span<const OrtEpDevice* const> ep_devices) {
+  // Get the OrtDevice from OrtEpDevice.device_memory_info if it is set. Otherwise, we set it to CPU.
+  // If there are multiple OrtEpDevice instances, the device_memory_info must be consistent for all.
+
+  if (ep_devices.empty()) {
+    // Should never be the case that we have no OrtEpDevices, but just in case.
+    return OrtDevice();
```
yuslepukhin: Would logging a warning be helpful here?

skottmckay: I don't think this is possible. If there's no OrtEpDevice how would we be creating an EP instance?

adrianlizarraga: I'll update this to `ORT_ENFORCE`

onnxruntime/core/session/provider_policy_context.cc
```diff
@@ -355,7 +356,7 @@ Status ProviderPolicyContext::CreateExecutionProvider(const Environment& env, Or
         info.ep_factory->CreateEp(info.ep_factory, info.hardware_devices.data(), info.ep_metadata.data(),
                                   info.hardware_devices.size(), &options, &logger, &api_ep)));
     ep = std::make_unique<PluginExecutionProvider>(UniqueOrtEp(api_ep, OrtEpDeleter(*info.ep_factory)), options,
-                                                   *info.ep_factory, info.devices);
+                                                   *info.ep_factory, info.devices, *logger.ToInternal());
```
adrianlizarraga, line 359: Note: explicitly passing the logger to `PluginExecutionProvider()`, which currently assumes it always has a valid logger. This would have caused a segfault if the EP tried to log a message in a session that uses auto EP.

onnxruntime/core/session/ep_plugin_provider_interfaces.cc
adrianlizarraga: updated.

adrianlizarraga:
Will close and reopen to get the `license/cla` check to run.

skottmckay (approved):

PR #25319, opened 2025-07-07
[CPU] GQA supports attention scores output

derdeljan-msft:
### Description
1. Add optional output to CPU impl of GQA op for storing attention scores (QK). Buffer is of shape (B, N, S, T) and can either be fp16 or fp32, depending on the type of other inputs
2. Add `qk_output` attribute to GQA, which controls if attention scores should be saved before or after softmax is applied
3. Add unit tests to cover this use case
4. Added asserts on other EPs if this feature is used


onnxruntime/core/graph/contrib_ops/bert_defs.cc
```diff
@@ -1127,6 +1127,10 @@ ONNX_MS_OPERATOR_SET_SCHEMA(
               "Use a smooth factor in softmax.",
               AttributeProto::INT,
               static_cast<int64_t>(-1))
+        .Attr("output_post_softmax_qk",
```
kunal-vaishnavi: There are [many cases](https://github.com/onnx/onnx/pull/6501#discussion_r1975912197) where the QK output is needed. Can we generalize this attribute name to be similar to the ONNX attention op?

xadupre:
I just finished the implementation of Attention(23) on CPU (float only): https://github.com/microsoft/onnxruntime/pull/25156. This operator supports GQA and all the options related to qk matmul output. Then I'll add support float16 and CUDA.

onnxruntime/core/graph/contrib_ops/bert_defs.cc
```diff
@@ -1200,6 +1204,11 @@ ONNX_MS_OPERATOR_SET_SCHEMA(
                 "(k-v buffer), it is of length max_sequence_length... otherwise of length past_sequence_length +"
                 "kv_sequence_length.",
                 "T")
+        .Output(3,
```
gramalingam, line 1247: The type-and-shape-inference method should be extended to propagate the shape for this optional output. It would also be useful to check in that method that the output is specified if-and-only-if the corresponding attribute value is 0 or 1.

aciddelgado (dismissed):
Suggest adding an error check for when this feature is attempted to be used on other EP's.

onnxruntime/core/graph/contrib_ops/bert_defs.cc
derdeljan-msft: Sure, but should we still have -1, 0 and 1 as the options (for now)?

fs-eire: do we need this attribute at all? My understanding is that ORT should be able to figure out whether to output or not by looking at whether output[3] exists.

derdeljan-msft: The idea behind this attribute was to allow the user to specify which version of QK should be written to the output, right now it supports before softmax (0) or after softmax (1). Finer granulation can be implemented, if necessary, I don't need it for my use case though.

derdeljan-msft: Renamed the attribute and introduced enum class with valid values for better readability.

onnxruntime/contrib_ops/cpu/bert/attention_common.h
```diff
@@ -53,6 +53,12 @@ enum AttentionKernelType {
   AttentionKernel_Default
 };
 
+enum class QKOutputType : int {
```
kunal-vaishnavi, line 56: Can you update the MHA op implementation to use this class as well?

onnxruntime/contrib_ops/cpu/bert/group_query_attention_helper.cc
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include "contrib_ops/cpu/bert/group_query_attention_helper.h"
+
+namespace onnxruntime {
+namespace contrib {
+namespace group_query_attention_helper {
+
+Status CheckCPUSpecificFeatures(const std::string& epName, int num_outputs, int qk_output) {
```
kunal-vaishnavi: I think we can rename this method to `CheckQKOutput`. We can also remove having the EP name and just say that `output_qk` is not supported.

onnxruntime/contrib_ops/cpu/bert/attention_common.h
derdeljan-msft: As far I as I see, MHA always outputs QK before the softmax and after the masking, based on whether the output buffer was provided, it doesn't have the corresponding attribute to control this.

onnxruntime/contrib_ops/cpu/bert/group_query_attention_helper.cc
derdeljan-msft: Done

onnxruntime/contrib_ops/cuda/bert/group_query_attention.cc
```diff
@@ -109,6 +109,10 @@ Status GroupQueryAttention<T>::ComputeInternal(OpKernelContext* context) const {
   parameters.do_rotary = do_rotary_;
   parameters.rotary_interleaved = rotary_interleaved_;
 
+  ORT_RETURN_IF_ERROR(group_query_attention_helper::CheckQKOutput(
```
kunal-vaishnavi: The current GQA CUDA implementation will never be able to have a QK output. GQA CUDA uses either flash attention or memory efficient attention. Neither kernel supports returning the QK output.

Side note: This is why GQA is not used in models like Whisper that require the QK output.

onnxruntime/core/graph/contrib_ops/bert_defs.cc
```diff
@@ -1205,10 +1240,15 @@ ONNX_MS_OPERATOR_SET_SCHEMA(
...
+                "output_qk",
+                "Values of QK matrix multiplication, either before or after SoftMax noramlization",
```
kunal-vaishnavi:
```suggestion
                "Values of QK matrix multiplication, either before or after softmax normalization",
```

onnxruntime/contrib_ops/cuda/bert/group_query_attention.cc
derdeljan-msft: I see, should I remove the assert then or is it fine to stay "just-in-case", as @aciddelgado suggested to add asserts for other EPs?

kunal-vaishnavi: The assert will be useful if an unfused implementation is added in the future that can have a QK output. The assert can stay but adding a comment summarizing the above info may be useful.

derdeljan-msft: Done

onnxruntime/core/graph/contrib_ops/bert_defs.cc
```diff
@@ -1127,6 +1158,10 @@ ONNX_MS_OPERATOR_SET_SCHEMA(
               "Use a smooth factor in softmax.",
               AttributeProto::INT,
               static_cast<int64_t>(-1))
+        .Attr("qk_output",
+              "Output values of QK matrix multiplication before (0) or after (1) softmax normalization. Default value is -1 (don't output).",
```
kunal-vaishnavi, line 1162: Can you define the list of valid values here? For example, what values can be used and what they represent?

kunal-vaishnavi: Can you define the list of valid values here? What values can be used and what do they represent?

onnxruntime/contrib_ops/cpu/bert/gqa_attention_base.h
```diff
@@ -151,8 +156,9 @@ class GQAAttentionBase {
                              const bool past_present_share_buffer,                 // whether present key and value share the same buffer
                              const bool packed_qkv,                                // whether Q, K, V are packed
                              const bool is_prompt,                                 // whether it is prompt
-                             ThreadPool* tp,                                       // thread pool
-                             AllocatorPtr allocator) const {                       // allocator for temporary buffer
+                             T* output_qk,
```
kunal-vaishnavi: Let's insert this parameter after `present_key` to keep all of the `T*` output parameters together and add a comment to say what it is for.

onnxruntime/contrib_ops/cpu/bert/gqa_attention_base.h
```diff
@@ -312,13 +323,6 @@ class GQAAttentionBase {
             }
           }
 
-          if (use_smooth_softmax_ || head_sink != nullptr) {
-            float sink = (head_sink != nullptr) ? static_cast<float>(head_sink[head_index]) : 0.0f;
-            ComputeSmoothSoftmaxInplace(output_softmax + start_offset, static_cast<int>(window_size), sink, nullptr);
```
kunal-vaishnavi, line 317: I don't think the computation for softmax need to be moved below. The computation should be able to stay here and the `QKOutputType::BEFORE_SOFTMAX` can move locations as needed.

kunal-vaishnavi, line 317: I don't think the computation for softmax need to be moved below. The computation should be able to stay here and the computation for `QKOutputType::BEFORE_SOFTMAX` can move locations as needed.

derdeljan-msft: It needs to be moved bellow because we want to first apply the attention mask and then output QK before softmax.

onnxruntime/core/graph/contrib_ops/bert_defs.cc
derdeljan-msft: But it is already listed here: before(0), after(1) or don't output (-1)

onnxruntime/contrib_ops/cpu/bert/gqa_attention_base.h
derdeljan-msft: Done

onnxruntime/contrib_ops/cpu/bert/gqa_attention_base.h
```diff
@@ -80,8 +84,7 @@ class GQAAttentionBase {
     // Compute the attention score.
     bool gqa_mlas_supported = MlasGQASupported<T>(CblasNoTrans, CblasTrans) &&
                               MlasGQASupported<T>(CblasNoTrans, CblasNoTrans);
-    size_t bytes = SafeInt<size_t>(batch_size) * num_heads_ * sequence_length * seqlen_present_kv_cache *
-                   (gqa_mlas_supported ? sizeof(T) : sizeof(float));
+    size_t bytes = SafeInt<size_t>(batch_size) * num_heads_ * sequence_length * seqlen_present_kv_cache * (gqa_mlas_supported ? sizeof(T) : sizeof(float));
```
tianleiwu, line 88: I think output data type is fixed in onnx graph. Not decided by whether MLAS support it or not.

onnxruntime/contrib_ops/cpu/bert/attention_common.h
```diff
@@ -53,6 +53,12 @@ enum AttentionKernelType {
...
+  NO_OUTPUT = -1,
```
tianleiwu: nit: suggest to use 0, 1, 2 instead of -1, 0, 1 since 0 usually means no, and -1 means invalid.

onnxruntime/contrib_ops/cpu/bert/gqa_attention_base.h
tianleiwu, line 88: why seqlen_present_kv_cache instead of total_sequence_length? 

Normally, the qk or softmax output has shape like (batch_size, q_num_heads, q_sequence_length, total_sequence_length) where total_sequence_length = past_sequence_length + kv_sequence_length.

onnxruntime/test/python/transformers/test_gqa_cpu.py
```diff
@@ -1591,12 +1702,26 @@ def parity_check_gqa_prompt_no_buff(
 
     head_sink = get_custom_head_sink(config.num_heads, torch_type=torch_type) if config.has_head_sink else None
 
+    output_qk = (
+        torch.zeros(
+            config.batch_size,
+            config.num_heads,
+            config.kv_sequence_length,
```
tianleiwu, line 1709: This shape does not looks right. It shall be like (batch_size, num_head, q_sequence_length, total_sequence_length).

onnxruntime/test/python/transformers/test_gqa_cpu.py
```diff
@@ -1873,10 +2018,24 @@ def parity_check_gqa_past(
         else None
     )
 
+    output_qk = (
```
tianleiwu, line 2021: Do not need this. You can see other outputs like out has no zeros tensor passed to gqa_past_func

derdeljan-msft: It is required since we expect the user to provide the actual buffer with allocated memory, in which the op will write the output.

onnxruntime/contrib_ops/cpu/bert/attention_common.h
derdeljan-msft: Done

onnxruntime/contrib_ops/cpu/bert/gqa_attention_base.h
derdeljan-msft: This logic is unchanged, here we actually allocate the temp buffer for performing this computation. Later, if the user provided the actual buffer in which to store output QK, this temp buffer will be copied to, and potentially casted to the appropriate data type.

onnxruntime/test/python/transformers/test_gqa_cpu.py
derdeljan-msft: For prompt case, q_sequence_length is the same as kv_sequence_length from what I've seen

onnxruntime/contrib_ops/cpu/bert/gqa_attention_base.h
derdeljan-msft: Update output shape to be total_sequence_length. The allocation logic here remains unchanged (I am not sure why seqlen_present_cache instead of total_sequence_length is being allocated here, I am not really familiar with MLAS kernel for this one)

onnxruntime/core/graph/contrib_ops/bert_defs.cc
```diff
@@ -336,19 +339,47 @@ void BaseGroupQueryAttentionTypeAndShapeInference(ONNX_NAMESPACE::InferenceConte
           updateOutputShape(ctx, 2, present_shape);
         }
       }
+
+      if (output_qk_index >= 0) {
+        const bool did_supply_qk_buffer = ctx.hasOutput(output_qk_index);
+        const int64_t qk_output_type = getAttribute(ctx, "qk_output", 0);
```
gramalingam: nit: safer to use `QKOutputType::NO_OUTPUT` instead of 0 in this line and lines below. (For some reason, I thought it was -1 for no-output.)

derdeljan-msft: It used to be, I changed it to be 0, 1, 2 instead of -1, 0, 1. Makes sense to use the enum here for the checks as well, updated.

onnxruntime/core/graph/contrib_ops/bert_defs.cc
```diff
@@ -1205,10 +1244,15 @@ ONNX_MS_OPERATOR_SET_SCHEMA(
                 "(k-v buffer), it is of length max_sequence_length... otherwise of length past_sequence_length +"
                 "kv_sequence_length.",
                 "T")
+        .Output(3,
+                "output_qk",
+                "Values of QK matrix multiplication, either before or after softmax noramlization",
```
kunal-vaishnavi:
```suggestion
                "Values of QK matrix multiplication, either before or after softmax normalization",
```

onnxruntime/contrib_ops/cpu/bert/group_query_attention_helper.h
```diff
@@ -398,6 +398,37 @@ Status CheckCustomAttentionInputs(const T* position_ids,
...
+
+  if (qk_output != static_cast<int>(QKOutputType::NO_OUTPUT) && output_qk == nullptr) {
+    return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT,
+                           "qk_output attribute was configured but output buffer was not provided");
+  }
+
+  return Status::OK();
+}
+
+inline Status CheckQKOutput(int num_outputs, int qk_output) {
```
tianleiwu: nit: Maybe rename this function to `CheckNoQkOutput`

onnxruntime/contrib_ops/cpu/bert/group_query_attention_helper.h
```diff
@@ -398,6 +398,37 @@ Status CheckCustomAttentionInputs(const T* position_ids,
...
+  if (num_outputs > 3) {
+    return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT,
+                           "output_qk optional output is not supported on");
```
tianleiwu: nit: remove on. Same below

tianleiwu:
A related work is to update symbolic shape inference: https://github.com/microsoft/onnxruntime/blob/9de58ac7a3d18d6ae7f7ae502b3f91361067f1b5/onnxruntime/python/tools/symbolic_shape_infer.py#L2548
I think it could be in another PR if needed in the future.

It would be nice to have same qk output modes like the onnx Attention op. We may update it later to make them consistent.

Overall, it looks good to me.

tianleiwu (approved):

derdeljan-msft:
> A related work is to update symbolic shape inference:
> 
> https://github.com/microsoft/onnxruntime/blob/9de58ac7a3d18d6ae7f7ae502b3f91361067f1b5/onnxruntime/python/tools/symbolic_shape_infer.py#L2548
> 
> 
> I think it could be in another PR if needed in the future.
> It would be nice to have same qk output modes like the onnx Attention op. We may update it later to make them consistent.
> 
> Overall, it looks good to me.

Makes sense, opened an issue to track this item https://github.com/microsoft/onnxruntime/issues/25412

kunal-vaishnavi (approved):

tianleiwu (approved):

PR #25320, opened 2025-07-08
Convert Initializers to OrtValues  Phase 2

yuslepukhin:
### Description

Make protobuf weights refer to OrtValues on load.
Create OrtValues for initializers that are loaded from ORT format for uniformity.
Create OrtValues for ORT format initializers.
Adjust exporting Graph::ToGraphProto() so it does not export in memory references in external data.
Make CoreML process external data including in memory references so it can copy it.

### Motivation and Context
Follow up for https://github.com/microsoft/onnxruntime/pull/23979


onnxruntime/core/graph/graph.cc
```diff
@@ -4213,67 +4250,153 @@ const ONNX_NAMESPACE::GraphProto& Graph::ToGraphProto() {
...
+    }
+  }
 
-    // copy any in-memory external data into raw data
-    if (utils::HasExternalDataInMemory(initializer)) {
-      const std::filesystem::path ignored;
-      std::basic_string<ORTCHAR_T> location;
-      onnxruntime::FileOffsetType file_offset;
-      SafeInt<size_t> tensor_byte_size;
+  // When graph_proto is copied from graph_proto, initializers already present in the main graph
```
adrianlizarraga, line 4315: is one of these `graph_proto_` (member variable)? #Resolved

onnxruntime/core/graph/graph.cc
```diff
@@ -1320,7 +1341,22 @@ Graph::Graph(const Model& owning_model,
...
+    // or during session state finalization.
+    // If data is already in memory, do nothing.
+    if (!utils::HasExternalData(tensor)) {
+      const bool is_sparse = sparse_tensor_names_.count(tensor.name());
+      if (is_sparse) {
+        sparse_tensor_names_.erase(tensor.name());
+      }
+      put_data_maybe_in_memory(tensor);
+      if (is_sparse) {
+        sparse_tensor_names_.emplace(tensor.name());
```
adrianlizarraga, line 1358: Do we need to remove and re-add the tensor name to the set because it is a set over references to strings? I wonder if updating `sparse_tensor_names_` to be a set over string_view would simplify things... probably better left for another PR though #Resolved

onnxruntime/core/graph/graph.cc
```diff
@@ -1320,7 +1341,22 @@ Graph::Graph(const Model& owning_model,
   }
 
   // Copy initial tensors to a map.
-  for (auto& tensor : graph_proto_->initializer()) {
+  for (int i = 0, lim = graph_proto_->initializer_size(); i < lim; ++i) {
+    auto& tensor = *graph_proto_->mutable_initializer(i);
+    // If data is on disk, it will be loaded either by optimizers
```
adrianlizarraga, line 1348: when external data is loaded by optimizers and EPs, it the file memory mapped multiple times? #Resolved

onnxruntime/core/graph/graph.cc
yuslepukhin: You are correct. The thing appeared before string_view, changing it would be the same story though.

onnxruntime/core/graph/graph.cc
yuslepukhin: Optimizers only briefly map the external initializers for its purposes, and then promptly unmap. Often times, they create other initializers in memory often as replacements for the original ones.

We do not map external initializers (and, therefore, convert it to OrtValues) until SessionState finalization. By that time, all unused initializers are removed and will not be mapped. Among unused are the ones that have been consumed by EPs and no longer referenced by any node in the graph.

This PR converts in memory initializers to OrtValues when an instance of a Graph is created, however, it does not memap external initializers. The reason for this is as above: most of the EPs currently copy it into memory, either by briefly memapping them and copying or somehow loading it internally (I do not know how exactly).

Let me know, if this needs to be changed.

include/onnxruntime/core/graph/graph.h
```diff
@@ -1561,6 +1569,16 @@ class Graph {  // NOLINT(clang-analyzer-optin.performance.Padding): preserve exi
...
 
+  /// <summary>
+  /// This is a helper function to ::ToGraphProto()
+  /// It examines the subgraph initializers and converts data in memory references to inline.
+  /// Subgraph initializers are already copied using Node::ToPtoro()
+  /// </summary>
+  /// <param name="output_graph_proto"></param>
+  /// <param name="process_main">process main graph if true</param>
+  /// <returns></returns>
+  Status ProcessSubgraphsInmemoryData(ONNX_NAMESPACE::GraphProto& output_graph_proto, bool process_main) const;
```
skottmckay:
```suggestion
  /// Subgraph initializers are already copied using Node::ToProto()
  /// </summary>
  /// <param name="output_graph_proto"></param>
  /// <param name="process_main">process main graph if true</param>
  /// <returns></returns>
  Status ProcessSubgraphsInMemoryData(ONNX_NAMESPACE::GraphProto& output_graph_proto, bool process_main) const;
``` #Pending

File: onnxruntime/core/graph/graph.cc [hunk 5]
```diff
@@ -4213,67 +4250,153 @@ const ONNX_NAMESPACE::GraphProto& Graph::ToGraphProto() {
...
-    output = initializer;
+Status Graph::ProcessSubgraphsInmemoryData(ONNX_NAMESPACE::GraphProto& output_graph_proto,
+                                           bool process_main) const {
+  for (const auto& node : Nodes()) {
+    if (node.ContainsSubgraph()) {
+      // Let's find this node in the output_graph_proto
+      auto hit = std::find_if(output_graph_proto.mutable_node()->begin(),
+                              output_graph_proto.mutable_node()->end(),
+                              [&node](const ONNX_NAMESPACE::NodeProto& proto) {
+                                return proto.name() == node.Name();
```
skottmckay: A name is an optional value in onnx::Node. Most models set it, and we have tools that can, but I'm not sure we can rely on that always being the case. May need to use an output value name for the check given a value can only be assigned once in an individual graph.

onnxruntime/core/graph/graph.cc
```diff
@@ -5762,6 +5919,21 @@ common::Status Graph::LoadFromOrtFormat(const onnxruntime::fbs::Graph& fbs_graph
...
+        std::unique_ptr<ExternalDataInfo> external_data_info;
+        ORT_RETURN_IF_ERROR(ExternalDataInfo::Create(initializer->external_data(), external_data_info));
+        auto data_offset = external_data_info->GetOffset();
+        OrtValue ort_value;
+        // const_cast due to Tensor constructor requiring non-const void* for data.
+        Tensor::InitOrtValue(DataTypeImpl::TensorTypeFromONNXEnum(
+                                 initializer->data_type())
+                                 ->GetElementType(),
+                             onnxruntime::utils::GetTensorShapeFromTensorProto(*initializer),
+                             reinterpret_cast<void*>(data_offset), CPUAllocator::DefaultInstance()->Info(),
```
skottmckay, line 5947: Is this taking a copy or assuming the flatbuffer will remain in memory? Not sure it's guaranteed to stay in memory. 

There are a couple of session config values that allow that when loading from bytes (kOrtSessionOptionsConfigUseORTModelBytesDirectly + kOrtSessionOptionsConfigUseORTModelBytesForInitializers). Not sure what happens when loading from memory. Don't we close the file after we create the onnxruntime::Model instance?

yuslepukhin: I checked both settings, and if the proto has an external in memory reference, we should be fine. This is simply creating an OrtValue around the buffer that is supposed stay valid anyway.

onnxruntime/core/graph/graph.cc
yuslepukhin: Yes

onnxruntime/core/framework/tensorprotoutils.cc
```diff
@@ -264,12 +265,32 @@ Status TensorProtoWithExternalDataToTensorProto(
...
+  // Strings can only be in memory
+  if (utils::HasString(ten_proto)) {
+    ORT_RETURN_IF_NOT(HasExternalDataInMemory(ten_proto),
+                      "TensorProto with string data can only be in memory");
+
+    std::unique_ptr<onnxruntime::ExternalDataInfo> external_data_info;
+    ORT_RETURN_IF_ERROR(onnxruntime::ExternalDataInfo::Create(ten_proto.external_data(), external_data_info));
+
+    // file_offset is address
+    if (utils::HasString(ten_proto)) {
```
edgchen1: redundant? given the check above:

https://github.com/microsoft/onnxruntime/blob/64de3151513384a3f32ee14017e880e904b1831b/onnxruntime/core/framework/tensorprotoutils.cc#L268-L269

onnxruntime/core/graph/graph.cc
```diff
@@ -1225,6 +1226,24 @@ Graph::Graph(const Model& owning_model,
...
+    ORT_THROW_IF_ERROR(utils::GetSizeInBytesFromTensorProto<0>(tensor_proto, &size_in_bytes));
+    if (size_in_bytes > utils::kSmallTensorExternalDataThreshold) {
+      OrtValue ort_value;
+      ORT_THROW_IF_ERROR(utils::TensorProtoToOrtValue(Env::Default(), model_path, tensor_proto,
+                                                      CPUAllocator::DefaultInstance(), ort_value));
+      constexpr const bool use_tensor_buffer_true = true;
+      auto tensor_proto_to_add = utils::TensorToTensorProto(ort_value.Get<Tensor>(), tensor_proto.name(),
+                                                            use_tensor_buffer_true);
+      assert(ort_value.IsAllocated());
+      ortvalue_initializers_.insert_or_assign(tensor_proto_to_add.name(), std::move(ort_value));
```
edgchen1: is `insert_or_assign`'s overwrite behavior desired here, or should we check that we are not overwriting an existing entry? #Resolved

onnxruntime/core/graph/graph_utils.h
```diff
@@ -189,6 +189,15 @@ bool MatchesOpSinceVersion(const Node& node, gsl::span<const ONNX_NAMESPACE::Ope
 /** Checks if the node has the same op set domain as the given one. */
 bool MatchesOpSetDomain(const Node& node, std::string_view domain);
 
+/// <summary>
+/// The function checks that the external data offset points to the same data as the tensor.
+/// </summary>
+/// <param name="tensor_proto"></param>
+/// <param name="tensor"></param>
+/// <returns>true if successfull</returns>
```
edgchen1:
```suggestion
/// <returns>true if successful</returns>
```

onnxruntime/core/graph/graph.cc
```diff
@@ -4213,67 +4250,159 @@ const ONNX_NAMESPACE::GraphProto& Graph::ToGraphProto() {
...
+  if (utils::HasExternalDataInMemory(initializer)) {
+    OrtValue ort_value;
+    ORT_RETURN_IF_NOT(src_graph.GetOrtValueInitializer(initializer.name(), ort_value, /*check_outer_scope=*/false),
+                      "Unable to find OrtValue for initializer with data in memory");
+    ORT_RETURN_IF_NOT(graph_utils::CheckInMemoryDataMatch(initializer, ort_value.Get<Tensor>()),
+                      "Unable to match OrtValue for initializer with data in memory");
+    output = utils::TensorToTensorProto(ort_value.Get<Tensor>(),
+                                        initializer.name(), /* use_tensor_buffer */ false);
+  } else {
+    output = initializer;
```
edgchen1, line 4271: is copying necessary? or can it be avoided if this function takes `initializer` as a non-const reference and optionally updates it?

onnxruntime/core/graph/graph.cc
```diff
@@ -4213,67 +4250,159 @@ const ONNX_NAMESPACE::GraphProto& Graph::ToGraphProto() {
...
+        }
+      }
     }
-  }
 #else
-  for (const auto& initializer : graph_proto_->initializer()) {
-    add_initializer(*mutable_initializers, initializer);
-  }
+    // Add initializers to parent graph by copy converting them from graph_proto_
+    // ToGraphProtoInternal() does not copy initializers for the main graph
```
edgchen1: this comment seems redundant since there is a copy of it outside of the ifdef as well.

onnxruntime/core/graph/graph.cc
edgchen1: I see that it only needs to be copied at some call sites. in that case, maybe we could have an in-place version and a copying version that calls the in-place one?

also, regarding the name, the "ToGraphProto" part doesn't seem relevant. it is going from a src tensorproto to a dst tensorproto.

onnxruntime/core/optimizer/attention_fusion_helper.h
```diff
@@ -272,9 +273,9 @@ struct MatchUnidirMaskResult {
   std::vector<NodeIndex> node_indices;  // id of all nodes in the subgraph for removing later.
 };
 
-// Return true when mask is unidirectionl (lower trigular) or all elements are 1.
+// Return true when mask is unidirectional (lower trigular) or all elements are 1.
```
edgchen1:
```suggestion
// Return true when mask is unidirectional (lower triangular) or all elements are 1.
```

onnxruntime/core/optimizer/attention_fusion_helper.h
```diff
@@ -3,6 +3,7 @@
 #include <limits>
 #include "onnx/defs/shape_inference.h"
 #include "onnx/defs/tensor_proto_util.h"
+#include "core/common/span_utils.h"
```
edgchen1: why is this one required?

onnxruntime/core/graph/graph.cc
yuslepukhin: I think this is the correct behavior since protos and ort values must be tightly coupled. However, I can add a check that the entry should not exist prior.

onnxruntime/core/graph/graph.cc
yuslepukhin: Good for the name comment.

There is not an in-place version since it is different graphs. I am not sure we need a function for a simple `operator=` invocation.

include/onnxruntime/core/graph/graph.h
```diff
@@ -1561,6 +1569,14 @@ class Graph {  // NOLINT(clang-analyzer-optin.performance.Padding): preserve exi
   Status AddConstantProtoAsInitializer(const ONNX_NAMESPACE::NodeProto& constant_node_proto,
                                        std::optional<std::string_view> new_name);
 
+  /// <summary>
+  /// Subgraph initializers are already copied using Node::ToProto()
```
edgchen1: nit: this doesn't describe what the function does.

include/onnxruntime/core/graph/graph.h
```diff
@@ -1198,8 +1198,16 @@ class Graph {  // NOLINT(clang-analyzer-optin.performance.Padding): preserve exi
 #endif
 
 #if !defined(ORT_MINIMAL_BUILD)
-  /** Gets the GraphProto representation of this Graph. */
+  /** Gets the GraphProto representation of this Graph only. */
   const ONNX_NAMESPACE::GraphProto& ToGraphProto();
+
+  /// <summary>
+  // This function recurses subgraphs and examines each initializer
```
edgchen1, line 1205: perhaps also mention the main purpose of this function, which I assume is to create a GraphProto.

onnxruntime/core/graph/graph.cc
```diff
@@ -1225,6 +1226,26 @@ Graph::Graph(const Model& owning_model,
...
+      OrtValue ort_value;
+      ORT_THROW_IF_ERROR(utils::TensorProtoToOrtValue(Env::Default(), model_path, tensor_proto,
+                                                      CPUAllocator::DefaultInstance(), ort_value));
+      constexpr const bool use_tensor_buffer_true = true;
+      auto tensor_proto_to_add = utils::TensorToTensorProto(ort_value.Get<Tensor>(), tensor_proto.name(),
+                                                            use_tensor_buffer_true);
+      assert(ort_value.IsAllocated());
+      auto ins_result = ortvalue_initializers_.insert_or_assign(tensor_proto_to_add.name(), std::move(ort_value));
+      ORT_ENFORCE(ins_result.second, "Unexpected duplicate insert or assign OrtValue for tensor: ", tensor_proto_to_add.name(),
+                  " in the initializer list.");
```
edgchen1, line 1244: how about `"OrtValue for initializer <name> already exists."`

onnxruntime/core/graph/graph.cc
edgchen1: I was referring to call sites like this:
```c++
      if (utils::HasExternalDataInMemory(initializer)) {
        // If the initializer has external data in memory, we need to inline it.
        ORT_RETURN_IF_ERROR(InlineOrCopyInitializer(*this, initializer, initializer));
      }
```

I think we could avoid the copy there. maybe it doesn't matter.

yuslepukhin: There is no copy here since the function would replace inline.

yuslepukhin: There is not a copy here since the original initializer is replaced by another version of it.

onnxruntime/core/graph/graph.cc
yuslepukhin: I think what is there is more accurate.

onnxruntime/core/graph/graph.cc
```diff
@@ -5052,24 +5193,38 @@ Node& Graph::FuseSubGraph(const IndexedSubGraph& sub_graph,
...
-  auto insert_result = name_to_initial_tensor_.emplace(tensor->name(), tensor);
-  ORT_ENFORCE(insert_result.second, "Constant node name: ", tensor->name(),
-              " conflicts with graph initializer. Check that the node names have been made unique.");
-  if (GetNodeArg(tensor->name()) == nullptr) {
-    TypeProto t{utils::TypeProtoFromTensorProto(*tensor)};
-    ORT_IGNORE_RETURN_VALUE(GetOrCreateNodeArg(tensor->name(), &t));
+  // In the constant node, we won't have symbolic dims.
+  const auto tensor_shape = utils::GetTensorShapeFromTensorProto(tensor_proto);
+  auto ml_data = DataTypeImpl::TensorTypeFromONNXEnum(tensor_proto.data_type())->GetElementType();
+  const size_t size_in_bytes = SafeInt<size_t>(ml_data->Size()) * tensor_shape.Size();
```
adrianlizarraga, line 5205: Sorry i should have caught this earlier.

This will over compute the number of bytes for Int4. Should use the static function `Tensor::CalculateTensorStorageSize()` to correctly get the byte size for any tensor type.

https://github.com/microsoft/onnxruntime/blob/d29a985adb33f1bc8c1eb7a07b60cdf8c44cb3ce/include/onnxruntime/core/framework/tensor.h#L146C17-L146C43

adrianlizarraga: I can update this in my follow up PR.

adrianlizarraga (approved):

snnn:

PR #25323, opened 2025-07-08
[NvTensorRTRTX EP]Disable Fast GELU operator in base model used for NV EP Unit Tests

keshavv27:
### Description
Remove fast_gelu operator from the base model created in NV TRT RTX EP unit tests.

### Motivation and Context
The operator was added in the model to partition the model into subgraphs which can be assigned to NV TRT RTX EP and CUDA EP, which supports fast_gelu. But CUDA EP is not built when building ORT with NV TRT RTX EP hence the unit tests fail with unsupported op error.  

@ishwar-raut1 @ankan-ban


keshavv27:
This change does not affect QNN EP, not sure why the check is failing.

jywu-mysoft:
please merge latest main

snnn:

PR #25327, opened 2025-07-08
[OVEP] Fix debug build

jnagi-intel:
### Description
Fix build issues in debug code under #ifndef NDEBUG check.
Switch constexpr std::string to static const

### Motivation and Context
MSVC fails to compile constexpr std::string under _DEBUG


jnagi-intel:
This is already addressed in existing PRs from @JulienMaille Thanks!  

https://github.com/microsoft/onnxruntime/pull/25299
https://github.com/microsoft/onnxruntime/pull/25298

@fs-eire FYI

PR #25332, opened 2025-07-08
[QNN EP] Fix pool with reshape name conflicts

qti-yuduo:
Naming conflicts when expand-pool2d-squeeze (implemented as reshape) logic is invoked during ONNX -> QNN op lowering. Model with multiple pool 1D ops would hit this issue.


HectorSVC (approved):
:shipit:

PR #25338, opened 2025-07-09
[MIGraphx EP] Sync AMD changes upstream 

TedThemistokleous:
Set of changes that were added to generate ROCm 7.0 branch pertaining to the MIGraphX Execution provider. Upstreams a bunch of changes

Adds the following features

Contrib Operator support:
groupnorm
nhwcconv
attention
skip_layernorm

Changes to EP flags
save/load has been removed for model_cache_dir
bf16 quantizer support + appropriate EP flags
hashing of saved .mxr file names based on gpu target, model name, and IO

Additional fixes involving renaming, tidy and other checks

Motivation and Context
Fixes, changes and updates to MIGraphX EP thats been done for ROCm development. Pushing this back upstream to ensure mainline onnxruntime is using the latest changes. Moving forward MIGraphX EP will be cut from the latest official release tag as a base point while also adding additional features that will be contributed back.


TedThemistokleous:
ping @tianleiwu  changes that AMD wants to upstream that are going into ROCm 7.0 release. Likely more incoming in another pR

tianleiwu:
@TedThemistokleous, please merge main since there is a recent change in lintrunner settting. Need `lintrunner init` and `lintrunner -a --all-files` to format the code after merge.

TedThemistokleous:
@tianleiwu done. Please let me know if there are anymore issues.

TedThemistokleous:
@tianleiwu done. Please let me know if there are anymore issues.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -711,15 +711,13 @@ typedef struct OrtTensorRTProviderOptions {
 typedef struct OrtMIGraphXProviderOptions {
```
snnn, line 711: This is a breaking change.  ONNX Runtime maintains ABI backwards compatibility. Certainly, this change breaks that.  As it will only impact the MIGraphX EP, I could accept this change. So, ORT's ABI backwards compatibility promise becomes if ORT provides ABI backwards compatibility if you don't use the functions that are not backwards compatible. (A little bit funny)
 
The correct way of doing this is to move the whole data structure out of this header file and make it opaque(invisible), and use getters/setters to access the fields. 

https://community.kde.org/Policies/Binary_Compatibility_Issues_With_C%2B%2B

onnxruntime/core/providers/migraphx/migraphx_allocator.cc
```diff
@@ -23,11 +23,11 @@ void MIGraphXAllocator::CheckDevice() const {
 #endif
 }
 
-void* MIGraphXAllocator::Alloc(size_t size) {
+void* MIGraphXAllocator::Alloc(const size_t size) {
```
snnn, line 26: Why do you need to add const here?

onnxruntime/core/providers/migraphx/migraphx_allocator.h
```diff
@@ -11,27 +11,27 @@ namespace onnxruntime {
 
 class MIGraphXAllocator : public IAllocator {
  public:
-  MIGraphXAllocator(int device_id, const char* name)
+  MIGraphXAllocator(const OrtDevice::DeviceId device_id, const char* name)
```
snnn, line 14: You do not need to use const here.  See https://abseil.io/tips/109 and https://google.github.io/styleguide/cppguide.html#Use_of_const

onnxruntime/core/providers/dml/dml_provider_factory.cc
```diff
@@ -562,7 +562,10 @@ static D3D12_COMMAND_LIST_TYPE CalculateCommandListType(ID3D12Device* d3d12_devi
       sizeof(feature_levels)
       ));
 
-  auto use_compute_command_list = (feature_levels.MaxSupportedFeatureLevel <= D3D_FEATURE_LEVEL_1_0_CORE);
+  // Use compute queue whenever possible on supported hardware to avoid TDR and maintain UI QoS
+  // Core and generic devices only have compute queues, DX11 has "immediate" submission, DX12 has both
+  auto use_compute_command_list = (feature_levels.MaxSupportedFeatureLevel <= D3D_FEATURE_LEVEL_1_0_CORE) ||
```
snnn, line 567: How does it relate to MIGRAPHX?

onnxruntime/python/onnxruntime_pybind_ortvalue.cc
```diff
@@ -390,10 +390,10 @@ void addOrtValueMethods(pybind11::module& m) {
         py::object obj = GetPyObjFromTensor(*ml_value, nullptr, GetRocmToHostMemCpyFunction());
 #elif USE_CANN
         py::object obj = GetPyObjFromTensor(*ml_value, nullptr, GetCannToHostMemCpyFunction());
-#elif USE_DML
-        py::object obj = GetPyObjFromTensor(*ml_value, nullptr, GetDmlToHostMemCpyFunction());
 #elif USE_MIGRAPHX
         py::object obj = GetPyObjFromTensor(*ml_value, nullptr, GetMIGraphXToHostMemCpyFunction());
+#elif USE_DML
```
snnn, line 395: So this change is to support the case that both USE_DML and USE_MIGRAPHX are defined?

cmake/onnxruntime_providers_migraphx.cmake
```diff
@@ -33,26 +23,24 @@
...
-
   file(GLOB_RECURSE onnxruntime_providers_migraphx_cc_srcs CONFIGURE_DEPENDS
     "${ONNXRUNTIME_ROOT}/core/providers/migraphx/*.h"
     "${ONNXRUNTIME_ROOT}/core/providers/migraphx/*.cc"
     "${ONNXRUNTIME_ROOT}/core/providers/shared_library/*.h"
     "${ONNXRUNTIME_ROOT}/core/providers/shared_library/*.cc"
   )
   source_group(TREE ${ONNXRUNTIME_ROOT}/core FILES ${onnxruntime_providers_migraphx_cc_srcs})
-  onnxruntime_add_shared_library_module(onnxruntime_providers_migraphx ${onnxruntime_providers_migraphx_cc_srcs})
+  onnxruntime_add_shared_library(onnxruntime_providers_migraphx ${onnxruntime_providers_migraphx_cc_srcs})
```
snnn, line 33: Why did you change it from onnxruntime_add_shared_library_module to onnxruntime_add_shared_library ? A Module Library is a plugin that may not be linked by other targets, but may be dynamically loaded at runtime using dlopen-like functionality.

onnxruntime/core/providers/migraphx/migraphx_execution_provider.cc
```diff
@@ -1470,19 +1528,24 @@ Status MIGraphXExecutionProvider::Compile(const std::vector<FusedNodeAndGraph>&
...
-      // input shapes are different, needs to re-parse onnx and
-      // re-compile the program
+      // input shapes are different, needs to reparse onnx and recompile the program
       if (!input_shape_match) {
-        if (!load_precompiled_model(prog, load_compiled_model_, std::string{load_compiled_path_})) {
-          LOGS_DEFAULT(VERBOSE) << "Input shape mismatch detected. Recompiling" << std::endl;
+        std::filesystem::path model_cache_file;
+        // empty cache path means the MXR caching is disabled - always compile
+        if (!model_cache_path_.empty()) {
+          model_cache_file = mgx_state->model_cache_dir / (mxr_filename_prefix + make_hash(input_shapes) + ".mxr");
```
snnn, line 1542: This part won't work good on Windows if the model_cache_dir contains non-ASCII chars. 
First, in your code you need to put a comment to clarify the encoding of mgx_state->model_cache_dir.  Is it ANSI encoding(that maps to a Windows codepage) or UTF-8?
Usually, most multibyte strings in ORT are UTF-8 strings. Then please convert the string to a wide string and do all the concatenation/modification in UTF-16.
On Linux all these strings should be UTF-8.

onnxruntime/core/providers/migraphx/migraphx_execution_provider.h
```diff
@@ -16,20 +17,17 @@
...
-static const char kCachePath[] = "ORT_MIGRAPHX_CACHE_PATH";
-static const char kINT8UseNativeMIGraphXCalibrationTable[] = "ORT_MIGRAPHX_INT8_USE_NATIVE_CALIBRATION_TABLE";
-static const char kSaveCompiledModel[] = "ORT_MIGRAPHX_SAVE_COMPILED_MODEL";
-static const char kSavedModelPath[] = "ORT_MIGRAPHX_SAVE_COMPILED_PATH";
-static const char kLoadCompiledModel[] = "ORT_MIGRAPHX_LOAD_COMPILED_MODEL";
-static const char kLoadModelPath[] = "ORT_MIGRAPHX_LOAD_COMPILED_PATH";
-static const char kExhaustiveTune[] = "ORT_MIGRAPHX_EXHAUSTIVE_TUNE";
-
-};  // namespace migraphx_env_vars
+constexpr auto kFP16Enable = "ORT_MIGRAPHX_FP16_ENABLE";
```
snnn, line 20: better to follow this guideline: https://abseil.io/tips/168 You may directly use std::string_view instead of absl::string_view since we have C++17. 
Or, I think you need to declare them as inline static constexpr const char*.

onnxruntime/core/providers/migraphx/migraphx_execution_provider.cc
```diff
@@ -158,10 +163,7 @@ void MIGraphXExecutionProvider::get_flags_from_session_info(const MIGraphXExecut
   }
 
   // Save/load migraphx compiled models
-  save_compiled_model_ = info.save_compiled_model;
-  save_compiled_path_ = info.save_model_file;
-  load_compiled_model_ = info.load_compiled_model;
-  load_compiled_path_ = info.load_model_file;
+  model_cache_path_ = info.model_cache_dir;
```
snnn, line 166: While you are welcomed to use std::filesystem::path, every time when you assign a value to this type you need carefully check the source string's encoding. 
On Windows we expect the source string is in native code page encoding. Otherwise if the string is encoded as UTF-8 you must manually convert it to std::wstring before assign it to a path.  
Or you may change type of model_cache_dir to const ORTCHAR_T*.

onnxruntime/core/providers/migraphx/migraphx_execution_provider.cc
```diff
@@ -239,69 +255,50 @@ void MIGraphXExecutionProvider::get_flags_from_env() {
...
-  if (!load_comp_model_env.empty()) {
-    load_compiled_model_ = (std::stoi(load_comp_model_env) == 0 ? false : true);
-    LOGS_DEFAULT(WARNING) << "\nORT_MIGRAPHX_LOAD_COMPILED_MODEL: " << load_compiled_model_;
-  }
-
-  const std::string load_model_path_env = onnxruntime::GetEnvironmentVar(migraphx_env_vars::kLoadModelPath);
-  if (load_compiled_model_ && !load_model_path_env.empty()) {
-    load_compiled_path_ = load_model_path_env;
-    LOGS_DEFAULT(WARNING) << "\nORT_MIGRAPHX_LOAD_COMPILED_PATH: " << load_compiled_path_;
+  const auto model_cache_path_env = GetEnvironmentVar(migraphx_env_vars::kModelCachePath);
```
snnn, line 258: Make the type explicit, and check the encoding.

onnxruntime/python/onnxruntime_pybind_ortvalue.cc
TedThemistokleous: Yes. I'm integrating and up-streaming changes that were in a separate branch related to windows side development.

onnxruntime/core/providers/migraphx/migraphx_execution_provider.cc
```diff
@@ -379,13 +375,16 @@ static bool IsTypeSupported(const NodeArg* node_arg) {
   }
 }
 
-static bool getMIGraphXType(ONNXTensorElementDataType type,
+static bool getMIGraphXType(const ONNXTensorElementDataType type,
```
snnn, line 378: Remove the const

onnxruntime/core/providers/dml/dml_provider_factory.cc
TedThemistokleous: Its a bunch of changes that are being integrated on the ROCm side for our EPs we maintain.

onnxruntime/core/providers/migraphx/migraphx_allocator.h
TedThemistokleous: Your linter and tidy said otherwise. These were ran through clang-tidy

onnxruntime/core/providers/migraphx/migraphx_allocator.cc
TedThemistokleous: We ran this through clang-tidy

onnxruntime/core/providers/migraphx/migraphx_execution_provider_info.h
```diff
@@ -14,6 +14,23 @@
 
 namespace onnxruntime {
 
+namespace migraphx_provider_option {
+constexpr auto kDeviceId = "device_id";
```
snnn, line 18: Same as the above. Consider using std::string_view instead.

onnxruntime/core/providers/migraphx/migraphx_execution_provider.cc
TedThemistokleous: Why?

onnxruntime/core/providers/migraphx/migraphx_provider_factory.cc
```diff
@@ -61,15 +68,15 @@ struct ProviderInfo_MIGraphX_Impl final : ProviderInfo_MIGraphX {
...
+  std::shared_ptr<IAllocator> CreateMIGraphXAllocator(const OrtDevice::DeviceId device_id, size_t migx_mem_limit, onnxruntime::ArenaExtendStrategy arena_extend_strategy, onnxruntime::MIGraphXExecutionProviderExternalAllocatorInfo& external_allocator_info, const OrtArenaCfg* default_memory_arena_cfg) override {
     return MIGraphXExecutionProvider::CreateMIGraphXAllocator(device_id, migx_mem_limit, arena_extend_strategy, external_allocator_info, default_memory_arena_cfg);
   }
 } g_info;
 
 struct MIGraphX_Provider : Provider {
   void* GetInfo() override { return &g_info; }
 
-  std::shared_ptr<IExecutionProviderFactory> CreateExecutionProviderFactory(int device_id) override {
+  std::shared_ptr<IExecutionProviderFactory> CreateExecutionProviderFactory(const int device_id) override {
```
snnn, line 81: Remove the const.

onnxruntime/core/providers/migraphx/migraphx_execution_provider.cc
TedThemistokleous: Odd, these came from windows side changes and we haven't seen this effecting our runs.

onnxruntime/core/providers/migraphx/migraphx_provider_factory.cc
```diff
@@ -131,10 +134,7 @@ struct MIGraphX_Provider : Provider {
 
     migx_options.migraphx_use_native_calibration_table = internal_options.int8_use_native_calibration_table;
 
-    migx_options.migraphx_save_compiled_model = internal_options.save_compiled_model;
-    migx_options.migraphx_save_model_path = internal_options.save_model_file.c_str();
-    migx_options.migraphx_load_compiled_model = internal_options.load_compiled_model;
-    migx_options.migraphx_load_model_path = internal_options.load_model_file.c_str();
+    migx_options.migraphx_cache_dir = internal_options.model_cache_dir.string().c_str();
```
snnn, line 139: .string().c_str() returns a temporary pointer I believe.  You need to store the result of internal_options.model_cache_dir.string() at somewhere.

onnxruntime/core/providers/migraphx/migraphx_execution_provider.h
TedThemistokleous: wouldn't that be covered by the auto here though? We're enforcing style checks that aren't in the liner? 

Just want some clarity on this.

onnxruntime/core/providers/migraphx/migraphx_execution_provider.cc
TedThemistokleous: Should all be UTF-8 I assume. We shouldn't be using non ascii characters here.

onnxruntime/core/providers/migraphx/migraphx_provider_factory.cc
```diff
@@ -145,6 +145,19 @@ struct MIGraphX_Provider : Provider {
...
   void Initialize() override {
+#ifdef _WIN32
+    HMODULE module = nullptr;
+    if(GetModuleHandleEx(GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS |
+                              GET_MODULE_HANDLE_EX_FLAG_UNCHANGED_REFCOUNT,
+                          static_cast<LPCSTR>(static_cast<void*>(InitializeRegistry)),
+                          &module) != 0) {
+      char buffer[MAX_PATH];
+      if(GetModuleFileName(module, buffer, sizeof(buffer)) != 0) {
+        PathRemoveFileSpec(buffer);
```
snnn, line 158: Try to see if std::filesystem has a function for doing so. Then you will not need to use the low level Windows APIs.

onnxruntime/core/providers/migraphx/migraphx_provider_factory.cc
```diff
@@ -145,6 +145,19 @@ struct MIGraphX_Provider : Provider {
   }
 
   void Initialize() override {
+#ifdef _WIN32
+    HMODULE module = nullptr;
+    if(GetModuleHandleEx(GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS |
+                              GET_MODULE_HANDLE_EX_FLAG_UNCHANGED_REFCOUNT,
+                          static_cast<LPCSTR>(static_cast<void*>(InitializeRegistry)),
+                          &module) != 0) {
+      char buffer[MAX_PATH];
```
snnn, line 156: Always use wide Windows APIs when possible.

cmake/onnxruntime_providers_migraphx.cmake
TedThemistokleous: Changed due to development for our windows side run. Is there one way or the other preferred here?  let me know so I can flow this back to the ROCm internal branch

include/onnxruntime/core/session/onnxruntime_c_api.h
TedThemistokleous: Was added in an earlier commit and I assume moving forward anyone using OnnxRT will be using the latest release supported with their ROCm install since we have a mapping in the docs.

https://onnxruntime.ai/docs/execution-providers/MIGraphX-ExecutionProvider.html#requirements

onnxruntime/core/providers/migraphx/migraphx_provider_factory.cc
```diff
@@ -145,6 +145,19 @@ struct MIGraphX_Provider : Provider {
   }
 
   void Initialize() override {
+#ifdef _WIN32
+    HMODULE module = nullptr;
+    if(GetModuleHandleEx(GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS |
```
snnn: This change usually is not needed.   Usually it prefers the folder  of the executable module that LoadLibraryEx is loading

See: https://learn.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-search-order#alternate-search-order-for-unpackaged-apps

onnxruntime/core/providers/migraphx/migraphx_provider_factory.cc
TedThemistokleous: Should this be split between windows/linux builds then around #ifdefs?

include/onnxruntime/core/session/onnxruntime_c_api.h
snnn: It is part of ORT's public API that everyone uses, no matter if MIGRAPHX EP is used or not.

onnxruntime/core/providers/migraphx/migraphx_allocator.cc
snnn: Our clang-tidy configs need to be improve.  Our coding style is inherited from Google's coding style  See https://github.com/microsoft/onnxruntime/blob/main/docs/Coding_Conventions_and_Standards.md. The document is the golden standard.

onnxruntime/core/providers/dml/dml_provider_factory.cc
snnn: Who made this change? Was someone from AMD or not?

cmake/onnxruntime_providers_migraphx.cmake
snnn: This change should not be made, since nobody would link to this library.

onnxruntime/core/providers/migraphx/migraphx_execution_provider.cc
snnn: Not that simple. You need to carefully examine the code.  Add a comment to the definition of the variable.

onnxruntime/core/providers/migraphx/migraphx_execution_provider.h
snnn: Actually, auto is discouraged to use except in the cases where the type is very obvious to see. https://google.github.io/styleguide/cppguide.html#Type_deduction 
"Use type deduction only if it makes the code clearer to readers who aren't familiar with the project, or if it makes the code safer. Do not use it merely to avoid the inconvenience of writing an explicit type."

I think there it is not that obvious. The keyword inline is needed.

onnxruntime/core/providers/migraphx/migraphx_execution_provider.cc
snnn: So, if you have a std::wstring or ORT_CHART* string, you have no problem. Otherwise try to explicitly state the encoding. Otherwise people will assume the string is a UTF-8 string.

snnn: On Windows You should never assign a UTF-8 string to a std::filesystem::path.

onnxruntime/core/providers/dml/dml_provider_factory.cc
TedThemistokleous: From AMD

onnxruntime/core/providers/migraphx/migraphx_execution_provider.cc
snnn: I just checked, onnxruntime::GetEnvironmentVar returns an ANSI string, not a UTF-8 string. As you are define a new var, it's better to state it there this string is not a UTF-8 string(while most std::string variables are).

onnxruntime/core/providers/migraphx/migraphx_provider_factory.cc
snnn: I mean when a Windows API has an ANSI version and Wide version, you need to use the Wide version.
The buffer here should be std::wstring.

cmake/onnxruntime_providers_migraphx.cmake
TedThemistokleous: We have a use case which does link to this. Hence why this was added and integrated back on the ROCm side. Let me know whats the optimal way to handle this. Looks like I'll need to handle the integration work on this. Appreciate your comments sand insights btw @snnn

onnxruntime/core/providers/migraphx/migraphx_allocator.h
snnn: Then I think we do not have good clang-tidy config here.

onnxruntime/core/providers/dml/dml_provider_factory.cc
snnn: Would it be possible to split this change to a separate PR? I need to find someone from the directml team to review this change.

cmake/onnxruntime_providers_migraphx.cmake
snnn: Then where are the header files of this library?  I need to know more background.

snnn: And, different link type results different construction/destruction order for global vars, which is common source of crashes. It's not just a build type thing.

include/onnxruntime/core/session/onnxruntime_c_api.h
TedThemistokleous: Can we pad this instead if this is a concern then? We're already using this internally for our builds and release, adding two padding vars should keep things consistent since we're missing a `char*` and `int` no?

TedThemistokleous: Or is it better to have these items but tie them to nothing in the EP then and have a `/*depricated */` comment for sunset flags.?

onnxruntime/core/providers/migraphx/migraphx_execution_provider.h
TedThemistokleous: Sure but again, why is this style comment not part of your linter profile checks then? This isn't a technical comment but style and if you want to enforce that, make it part of your lint runner checks/lint stage.

include/onnxruntime/core/session/onnxruntime_c_api.h
snnn: Padding is also a breaking change. Because clients built with older ORT versions will end-up with garbage data for the new fields. Please try to minimize the number of changes of this struct. Going forward you should move it to an opaque type like what CUDA EP did.

TedThemistokleous:
@snnn rolled back latest commit since that was the cause of a few issues. Let me know if you have any more review comments.

snnn:
Thanks. I don't have other comments.

snnn:
Sorry, one more thing: onnxruntime/tools/ci_build/github/linux/docker/migraphx-ci-pipeline-env.Dockerfile needs to be updated:
1. The docker file uses Ubuntu, which I could not get an approval from our internal. We need to use almalinux8 or AzureLinux3. This restriction only applies to the pipelines that runs on Microsoft's first-party machines(which do not include the public build machines that AzureDevOps or Github offers). 
2. The docker file uses conda, which is not free. Please consider using system python, or at least you need to disable the default channels and only use conda-forge.

onnxruntime/core/providers/dml/dml_provider_factory.cc
TedThemistokleous: Sure, I can pull this one out. I'm getting back to the other devs responsible for these to make sure these are up to standard to merge back upstream

cmake/onnxruntime_providers_migraphx.cmake
TedThemistokleous: I don't disagree with you. Unfortunately this was pushed from windows side changes since the original interface was changed. Seems like a better solution is to keep the old flags then (save/load) and add the newer one (model cache) so that we don't break abi compatibility then and just extend it?

onnxruntime/core/providers/migraphx/migraphx_provider_factory.cc
apwojcik: The change is required for AMD MIGraphX, not for the MIGraphX Execution Provider. The MIGraphX DLL loads the target DLLs (i.e., MIGraphX GPU target) at runtime. It does not use an absolute path and expects the target DLLs to be in the same directory as the MIGraphX DLL.

For example, we bundle MIGraphX binaries with MIGraphX EP Python WHL. Without this change, the MIGraphX does not know where to look for the GPU target. For Python, the executable working path is usually the location of a Python script or the Python executable.

cmake/onnxruntime_providers_migraphx.cmake
apwojcik: We changed to `onnxruntime_add_shared_library` because `onnxruntime_add_shared_library_module` puts the provider DLL in `lib` directory when installing with `cmake --install ...`, and we need the DLL in `bin` directory, and `onnxruntime_add_shared_library` does precisely that. The side effects are the header and interface library installed alongside.

TedThemistokleous:
Closing this out as our windows side team will be upstreaming their changes instead of me bundling everything together. @snnn  I'll be adding my own set of changes from Linux side in a separate PR, ensure the DML stuff is split out and make sure anything going in ROCm 7.0 linux side is up to parity. 

@apwojcik  is the point of contact for Windows changes and will be opening up integration off this branch with fixes.

onnxruntime/core/providers/migraphx/migraphx_provider_factory.cc
snnn: So there is something wrong with the LoadLibrary calls 

> For Python, the executable working path is usually the location of a Python script or the Python executable.

It's not about working dir. It's more about DLL search order. We should never search that directory.

snnn:

PR #25341, opened 2025-07-09
Remove arm 32 references

ispysoftware:
### Description
The library no longer supports arm 32 bit. This removes mentions of the non-extant libraries.

### Motivation and Context
Existing files are confusing as they specify arm 32 is supported when it isn't.


ispysoftware:

PR #25346, opened 2025-07-09
Update python bindings to be able to use a shared allocator and/or IDataTransfer registered by a plugin EP in the Environment

skottmckay:
### Description

Add ability to get shared allocator from env.
Add ability to create a MemCpyFunc using the IDataTransfer from the environment.

### DIFF
--- a/onnxruntime/python/onnxruntime_inference_collection.py
+++ b/onnxruntime/python/onnxruntime_inference_collection.py
@@ -899,7 +899,7 @@ class OrtValue:
         :param numpy_obj: The Numpy object to construct the OrtValue from
         :param device_type: e.g. cpu, cuda, cann, cpu by default
         :param device_id: device id, e.g. 0
-        :param vendor_id: If provided the device type should be "gpu" or "npu".
+        :param vendor_id: The device's PCI vendor id. If provided, the device_type should be "gpu" or "npu".
         """
         # Hold a reference to the numpy object (if device_type is 'cpu') as the OrtValue
         # is backed directly by the data buffer of the numpy object and so the numpy object
--- a/onnxruntime/python/onnxruntime_pybind_state.cc
+++ b/onnxruntime/python/onnxruntime_pybind_state.cc
@@ -1912,6 +1912,10 @@ void addObjectMethods(py::module& m, ExecutionProviderRegistrationFn ep_registra
 #elif USE_ROCM || USE_MIGRAPHX
                vendor = OrtDevice::VendorIds::AMD;
 #endif
+             } else if (type == OrtDevice::NPU) {
+#if USE_CANN
+               vendor = OrtDevice::VendorIds::HUAWEI;
+#endif
              }
 
              return OrtDevice(type, mem_type, vendor, device_id);


skottmckay:
@gedoensmax will this provide what you need? not sure if there are any other places needing updates.

onnxruntime/python/onnxruntime_inference_collection.py
```diff
@@ -32,8 +32,10 @@ def get_ort_device_type(device_type: str, device_index) -> C.OrtDevice:
         return C.OrtDevice.dml()
     elif device_type == "webgpu":
         return C.OrtDevice.webgpu()
-    elif device_type == "ort":
-        return C.get_ort_device(device_index).device_type()
```
skottmckay, line 36: the 'ort' type doesn't exist in the C++ bindings so this seems to be out of date/invalid.

https://github.com/microsoft/onnxruntime/blob/1975fba6ac383fa962dec2db92b1836ae1d0804d/onnxruntime/python/onnxruntime_pybind_state.cc#L1897-L1903

onnxruntime/test/python/onnxruntime_test_python_autoep.py
```diff
@@ -244,6 +244,41 @@
...
+        gpu_value = onnxrt.OrtValue.ortvalue_from_numpy(data, "gpu", 0, 0xBE57)
+        # copy back to CPU
+        cpu_data = gpu_value.numpy()
+        np.testing.assert_equal(data, cpu_data)
+
+        gpu_value.update_inplace(data2)  # update the fake GPU data
+        cpu_data_2 = gpu_value.numpy()  # copy back to CPU
+        np.testing.assert_equal(data2, cpu_data_2)
+
+        gpu_value = None  # Delete OrtValue before unregistering library as the allocator will be destroyed.
```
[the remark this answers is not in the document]
skottmckay: Bad suggestion. Assigning to None allows the underlying OrtValue to be freed, and that value is attached to an allocator which is coming from the EP library. If we do not do this the library can get unloaded prior to the OrtValue being freed, which causes a crash as the allocator is no longer valid.

onnxruntime/python/onnxruntime_pybind_ortvalue.cc
```diff
@@ -23,42 +23,59 @@ std::unique_ptr<OrtValue> OrtValueFromShapeAndType(const std::vector<int64_t>& s
...
+
   if (strcmp(GetDeviceName(device), CPU) == 0) {
     allocator = GetAllocator();
-  } else if (strcmp(GetDeviceName(device), CUDA) == 0) {
+  } else {
+#if !defined(ORT_MINIMAL_BUILD)
+    // if a plugin EP has been registered we can get a shared allocator from the environment.
+    // we use this as the fallback option.
+    allocator = GetSharedAllocator(device);
+#endif
```
ishwar-raut1, line 34: The allocator obtained by GetSharedAllocator will be replaced by the EP allocators. 
for example,  
`allocator = GetSharedAllocator(device);` 
will replaced by 
`allocator = GetCudaAllocator(device.Id());`

skottmckay: The intention was for the one from GetSharedAllocator to be the fallback option to preserve the current behavior.

Happy to make it the preferred allocator in a build with USE_CUDA if you want that.

ishwar-raut1: We aim to leverage the shared environment allocator and data transfer capabilities in Python for the NV TRT-RTX Execution Provider. The current behavior prioritizes the EP-specific allocator over the shared allocator. The desired behavior is to use the shared allocator by default. If no shared allocator is available, users should be warned and guided to register the EP explicitly, after which the EP-specific allocator can be used.
Another issue observed in the current implementation is that when the NV TRT-RTX EP is registered, the allocator from the CUDA EP will be queried. This will occur because the Python C API is built to support all providers. As a result, it attempts to load the CUDA EP DLL—even if the CUDA EP is not registered—which can lead to unintended behavior.

skottmckay: Preferring the shared allocator for a CUDA device is fine. Not sure what other vendors prefer so I'll leave those as is as it is closer to existing behavior.

> If no shared allocator is available, users should be warned and guided to register the EP explicitly

I expect a gradual transition to using ABI EPs but that will take time, and registration is specific to that type of usage. Given that, warnings about registering the EP explicitly may be premature. Is there a clear place to get EPs from and information on how to register an EP that we would point a user to so they would understand what they need to do?

>the allocator from the CUDA EP will be queried

Can you clarify this? Where is the CUDA EP coming from and what is resulting in it being queried?

>This will occur because the Python C API is built to support all providers

My understanding is that the code _enables_ support for all providers, but in general only one (or maybe two in a case like CUDA + TRT) would be enabled in any specific build of the python package.

ishwar-raut1: >Can you clarify this? Where is the CUDA EP coming from, and what is resulting in it being queried?

`allocator = GetCudaAllocator(device.Id());`

This code will load the Cuda  EP dll. The allocator is implemented inside the CUDA EP DLL.  This will add dependency on CUDA EP. 

>My understanding is that the code enables support for all providers, but in general only one (or maybe two in a case like CUDA + TRT) would be enabled in any specific build of the python package.

I think the onnxruntime-WinML package is built to support all the EPs.

skottmckay: This only applies if the CUDA EP is internally included and USE_CUDA is defined.

onnxruntime/core/providers/cuda/cuda_provider_factory.cc
```diff
@@ -734,6 +734,10 @@ struct CudaEpFactory : OrtEpFactory {
     }
     */
 
+    // guard against bad device discovery. max devices we expect to add is num_cuda_devices. if we're attempting
+    // to add more than that we have duplicates in the `devices` array.
+    max_ep_devices = std::min(max_ep_devices, static_cast<size_t>(num_cuda_devices));
```
adrianlizarraga, line 739: Is this supposed to modify `num_devices` instead of `max_ep_devices`? I thought `max_ep_devices` is currently always set to some maximum number (8) that is unrelated to the number of devices obtained in device discovery.

onnxruntime/python/onnxruntime_inference_collection.py
```diff
@@ -889,26 +891,23 @@ def _get_c_value(self) -> C.OrtValue:
...
-    def ortvalue_from_numpy(cls, numpy_obj: np.ndarray, /, device_type="cpu", device_id=0) -> OrtValue:
+    def ortvalue_from_numpy(cls, numpy_obj: np.ndarray, /, device_type="cpu", device_id=0, vendor_id=-1) -> OrtValue:
         """
         Factory method to construct an OrtValue (which holds a Tensor) from a given Numpy object
         A copy of the data in the Numpy object is held by the OrtValue only if the device is NOT cpu
 
         :param numpy_obj: The Numpy object to construct the OrtValue from
         :param device_type: e.g. cpu, cuda, cann, cpu by default
         :param device_id: device id, e.g. 0
+        :param vendor_id: If provided the device type should be "gpu" or "npu".
```
adrianlizarraga: Is this comment accurate? The `vendor_id` parameter is an integer, but this comment could make it seem like a string is accepted.

adrianlizarraga: Oh maybe: `The device's vendor ID. If provided, the device type should be "gpu" or "npu".`

onnxruntime/core/providers/cuda/cuda_provider_factory.cc
skottmckay: max_ep_devices is the number of slots that can be written to. It's using std::min so it should never exceed the number of slots whilst also limiting to the number of actual CUDA devices detected (which conceptually is the other max we want to apply given we shouldn't have multiple OrtEpDevice instances for the same device).

onnxruntime/python/onnxruntime_inference_collection.py
skottmckay: Updated to be more explicit and also use 'device_type' so the relationship to that parameter is clearer.

TedThemistokleous:
@snnn  this broke Mainline for AMD, not sure why this got in without anyone from out end even looking at this or CI checks right now.

skottmckay:
> @snnn this broke Mainline for AMD, not sure why this got in without anyone from out end even looking at this or CI checks right now.

I'm not sure why the conflict detection didn't work. Typically if `main` has changed a PR that has conflicts with that change should be blocked from being checked in. The signature of GetCudaToHostMemCpyFunction for example was changed in different ways (return value in one PR, arguments in the other) by the two PRs and that should be trivial to detect as a conflict. 

@snnn is something not setup or working correctly in the ORT repo?

TedThemistokleous:
@skottmckay - Relevant read https://github.com/microsoft/onnxruntime/issues/25532

You guys are flying blind on all MIGraphX/ROCm/AMD stuff. Feel free to reach out if you need further support and resources from AMD.

skottmckay:
> @skottmckay - Relevant read #25532
> 
> You guys are flying blind on all MIGraphX/ROCm/AMD stuff. Feel free to reach out if you need further support and resources from AMD.

I'm slightly confused as to what the breakage was though. Was the breakage due to the checkin of this PR with the pending MIGraphX PR or did it break checked in code? If the latter, the conflict detection should have caught it, but based on the sequence of changes it's not clear to me that that was the case.

Original `main` from June 13
`const std::unordered_map<OrtDevice::DeviceType, MemCpyFunc>* GetCudaToHostMemCpyFunction() {`

Change from this PR updated the return value but the function still took no args.
`const std::unordered_map<OrtDevice, MemCpyFunc>* GetCudaToHostMemCpyFunction() {`

Change in new PR adds a new argument which matches what's in #25583 but as that's not checked in yet wouldn't have blocked any other changes going into main.
`const std::unordered_map<OrtDevice, MemCpyFunc>* GetCudaToHostMemCpyFunction(const OrtDevice& device) {`

What is 'mainline for AMD' building against?

PR #25348, opened 2025-07-09
Support read-only allocator for use with initializers

skottmckay:
### Description

Add new allocator type of OrtReadOnlyAllocator to enable providing a separate allocator that is only used for initializers.

Update the SessionState logic to support this allocator type being provided, and use it when doing device allocations for initializers.

### Motivation and Context

Performance.

### DIFF
--- a/onnxruntime/core/framework/session_state_utils.cc
+++ b/onnxruntime/core/framework/session_state_utils.cc
@@ -37,8 +37,10 @@ namespace session_state_utils {
 
 // The following method will allocate memory directly using the device allocator.
 // It can handle arena-based allocators and non-arena based allocators.
-static common::Status AllocateBufferUsingDeviceAllocatorFromShapeAndType(const TensorShape& tensor_shape, const DataTypeImpl* type,
-                                                                         const AllocatorPtr& alloc, /*out*/ void*& p_data) {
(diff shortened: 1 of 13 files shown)


onnxruntime/core/framework/session_state_utils.cc
```diff
@@ -370,6 +374,9 @@ common::Status SaveInitializedTensors(
       AllocatorPtr alloc;
       // TODO: if the tensor need be copied, does it have enough room?
       ORT_RETURN_IF_ERROR(planner.GetPreallocatedBuffer(ort_value_index, name, memory_buffer, alloc));
+
+      // ??? Should we ignore this session option if the EP is explictly providing the read only allocator?
```
fs-eire, line 378: i think it's safe to do so

fs-eire:
does it also support `onnxruntime::IExecutionProvider`?

skottmckay:
> does it also support `onnxruntime::IExecutionProvider`?

Yes. The ORT implementation deals in IExecutionProvider. If it's a plugin EP the OrtEp gets wrapped by an IExecutionProvider based class. If the IExecutionProvider's CreatePreferredAllocators returns an allocator with type or OrtReadOnlyAllocator in the SessionState ctor it will be used for initializers.

PR #25349, opened 2025-07-09
Add patch for WebGPU on Android to handle fp16 in uniforms

vraspar:
### Motivation and Context
Android devices (like S24) doesn't seem to allow fp16 in uniforms so the WebGPU EP has to manually handle passing an fp32 in the uniform and converting to fp16 before using.

### DIFF
--- a/cmake/external/onnxruntime_external_deps.cmake
+++ b/cmake/external/onnxruntime_external_deps.cmake
@@ -746,6 +746,16 @@ if (onnxruntime_USE_WEBGPU)
           #   Some build warnings are not allowed to be disabled in project level.
           ${Patch_EXECUTABLE} --binary --ignore-whitespace -p1 < ${PROJECT_SOURCE_DIR}/patches/dawn/dawn_binskim.patch)
 
+      if(Android)
+        # Android devices doesn't seem to allow fp16 in uniforms so the WebGPU EP has to manually handle passing an fp32
+        # in the uniform and converting to fp16 before using.
+        set(ONNXRUNTIME_Dawn_PATCH_COMMAND
+            "${ONNXRUNTIME_Dawn_PATCH_COMMAND} &&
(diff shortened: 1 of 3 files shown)


fs-eire:
Is there any existing f16 access to uniform buffer?

fs-eire:
I think the patch should just work for all builds, including Linux, right? If it's already working with f16 in the old code, it should not be affected with the patched code.

vraspar:
> I think the patch should just work for all builds, including Linux, right? If it's already working with f16 in the old code, it should not be affected with the patched code.

Sorry if I am misunderstanding the question, but the patch should only impact Android. I believe it should work on Linux theoretically, but wanted to minimize the impact so I limited to android

cmake/external/onnxruntime_external_deps.cmake
```diff
@@ -744,7 +744,13 @@ if (onnxruntime_USE_WEBGPU)
...
           # - (private) Fulfill the BinSkim requirements
           #   Some build warnings are not allowed to be disabled in project level.
-          ${Patch_EXECUTABLE} --binary --ignore-whitespace -p1 < ${PROJECT_SOURCE_DIR}/patches/dawn/dawn_binskim.patch)
+          ${Patch_EXECUTABLE} --binary --ignore-whitespace -p1 < ${PROJECT_SOURCE_DIR}/patches/dawn/dawn_binskim.patch &&
+
+          # Android devices doesn't seem to allow fp16 in uniforms so the WebGPU EP has to manually handle passing an fp32
+          # in the uniform and converting to fp16 before using.
+          ${Patch_EXECUTABLE} --binary --ignore-whitespace -p1 < ${PROJECT_SOURCE_DIR}/patches/dawn/uniform_and_storage_buffer_16_bit_access.patch)
+
+      add_compile_definitions(ORT_WEBGPU_NO_FP16_IN_UNIFORMS)
```
fs-eire: I don't see where it's used. should we remove this?

jywu-mysoft:
@fs-eire, @vraspar since https://github.com/microsoft/onnxruntime/pull/25391 is now marked for 1.23.1 , should this one be as well?

PR #25354, opened 2025-07-10
[WebNN] Fix bug in Float16Array availability check

Honry:
The `from` is not a property of `Float16Array` but an inherited function, we can use `Float16Array['from']` to check if it is available.


Honry:
@fdwr, PTAL, thanks!

PR #25360, opened 2025-07-10
Fix SigLIP casual mask bug

nenad1002:
### Description

SigLIP architecture inside the vision encoder should not use a causal mask on the attention. This change will fix Phi 4 MM accuracy issues we have seen.


onnxruntime/python/tools/transformers/fusion_attention_clip.py
```diff
@@ -320,7 +320,7 @@ def fuse(self, normalize_node, input_name_to_nodes, output_name_to_node):
             output=attention_last_node.output[0],
             add_qk_str=add_qk,
             scale=None,
-            causal=(add_mask is not None),
+            causal=False,
```
tianleiwu: Need to set based on the causal mask matching. 

For example, lines 291 to 304 match causal mask. If it is matched, this shall be set as True.

nenad1002: It's a pretty hacky way to check, but can work as well

onnxruntime/python/tools/transformers/fusion_attention_clip.py
```diff
@@ -320,7 +323,7 @@ def fuse(self, normalize_node, input_name_to_nodes, output_name_to_node):
             output=attention_last_node.output[0],
             add_qk_str=add_qk,
             scale=None,
-            causal=(add_mask is not None),
+            causal=(causal_mask_nodes_1 is not None) or (causal_mask_nodes_2 is not None),
```
tianleiwu, line 329: set causal to True only when the causal mask has not been added to the attention bias. If the bias has included the causal mask, there is no need to set the causal flag.

tianleiwu: You might manually test the clip optimization according to https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/python/tools/transformers/models/stable_diffusion/README.md.

nenad1002: I am still testing this in DRAFT, but we'll most probably have to pull this logic out from this file altogether since some tests rely on the add_mask being there which I cannot, I will be creating a new optimizer.

nenad1002: I've tested this, and it's also passing automated tests now, this should work. I am not following your first comment here, if we see no casual mask nodes, we should have no casual mask. For example, our Phi model has no casual mask added nor should we set it to be True here

tianleiwu: Could you add a test case for this?

nenad1002: It will require some changes, this model is not really testable, it does a bunch of weird stuff that's not trivial to replicate with torch

tianleiwu (approved):

PR #25363, opened 2025-07-11
[EP ABI] Update to use Node_GetEpName

chilo-ms:
Change to use `Node_GetEpName` API name to avoid confusion.
For plugin EPs, the EP factory can use whatever name that registered with ORT, so make the API name `Node_GetEpName` to align with `OrtEpFactory.GetName.`

### DIFF
--- a/include/onnxruntime/core/session/onnxruntime_c_api.h
+++ b/include/onnxruntime/core/session/onnxruntime_c_api.h
@@ -6028,7 +6028,7 @@ struct OrtApi {
 
   /** \brief Returns the execution provider name that this node is assigned to run on.
    *         Returns NULL if the node has not been assigned to any execution provider yet.
-   *         For plugin execution providers, the name is the one registered via RegisterExecutionProviderLibrary.
+   *         For plugin execution providers, the name is the one returned by OrtEp::GetName.
    *
    * \param[in] node The OrtNode instance.
    * \param[out] out Output execution provider type and can be NULL if node has not been assigned.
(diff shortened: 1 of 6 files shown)


include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -6026,8 +6026,9 @@ struct OrtApi {
    */
   ORT_API2_STATUS(Node_GetGraph, _In_ const OrtNode* node, _Outptr_result_maybenull_ const OrtGraph** graph);
 
-  /** \brief Returns the execution provider type (name) that this node is assigned to run on.
+  /** \brief Returns the execution provider name that this node is assigned to run on.
    *         Returns NULL if the node has not been assigned to any execution provider yet.
+   *         For plugin execution providers, the name is the one registered via RegisterExecutionProviderLibrary.
```
skottmckay: RegisterExecutionProviderLibrary uses a registration name but that isn't necessarily the EP name as a library can provide multiple EP factories. It is in the example EP though as they're 1:1.

The EP name primarily comes from OrtEpFactory::GetName (the OrtEpFactory populates the name in the OrtEpDevice), but it's also available from OrtEp::GetName. 

As this code is relevant to the EP instance, I think the most accurate to say here is that it's the value returned by OrtEp::GetName.

```suggestion
   *         For plugin execution providers, the name is the one returned by OrtEp::GetName.
```

skottmckay (approved):

PR #25364, opened 2025-07-11
[JSEP] Fix inputShape index OOB in slice.ts

jchen10:
Use `inputShape.length - 1` instead of `inputShape.length` to avoid out-of-bounds access.


jchen10:
Thanks to @Jiawei-Shao.  He happened to catch this bug. I just fix it for him in passing.

@fs-eire @guschmue PTAL

PR #25365, opened 2025-07-11
Add vendor id to OrtEpFactory and default ORT logger to CreateEpFactories

skottmckay:
### Description

Add vendor id to OrtEpFactory. It's easier to get the vendor id than name on other platforms.
Update the selection policy to prefer match on vendor id with fallback to vendor name.

Add default ORT logger to CreateEpFactories. 
The OrtEpFactory currently has no way to log informational messages or issues.
CreateEp is given the session logger for use by the OrtEp instance so that part of things is good.

Misc cleanups. Make usage of ORT_API2_STATUS and ORT_API_T consistent on onnxruntime_ep_c_api.h.
See ort_version_supported in some EP factories where it was missed.

### Motivation and Context

Vendor id is easier to match against OrtHardwareDevice when doing auto EP selection.
OrtEpFactory should have a logger. 
Last chance to cleanup APIs before 1.23 release

### DIFF
--- a/include/onnxruntime/core/session/onnxruntime_ep_c_api.h
+++ b/include/onnxruntime/core/session/onnxruntime_ep_c_api.h
@@ -358,7 +358,7 @@ struct OrtEp {
    *
    * \since Version 1.22.
    */
-  const char*(ORT_API_CALL* GetName)(_In_ const OrtEp* this_ptr);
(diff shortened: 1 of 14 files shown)


jywu-msft (approved):

snnn:

PR #25367, opened 2025-07-11
[webgpu] Enable per-run control for graph capture

qjia7:
This PR uses the existed RunOption `gpu_graph_id` to control whether to skip the graph capture. When the webgpu ep option `enableGraphCapture` is enabled, in RunOption, gpu_graph_id = -1 means skipping graph capture. Otherwise, go to the graph capture path for each session.run. If gpu_graph_id is not specified in RunOption, it will respect `enableGraphCapture `'s value to see whether to go to graph capture path.


fs-eire (approved):
LGTM. In future we can probably support multiple records but for now the current implementation is OK.

PR #25368, opened 2025-07-11
Enable CUDA Graph in nv_tensorrt_rtx EP

umangb-09:
Added support for CUDA graph enabling for NV TensorRT RTX EP. This enables significant performance improvements for repeated inference runs by capturing and replaying CUDA operations instead of re-executing them each time.

### Motivation and Context
- **Performance Optimization**: CUDA graph capture eliminates kernel launch overhead for repeated inference runs
- **Nv TensorRT RTX Integration**: Enables CUDA graph support specifically for NV TensorRT RTX operations in ONNX Runtime

### DIFF
--- a/onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
+++ b/onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
@@ -892,10 +892,10 @@ Status BindKernelOutput(Ort::KernelContext& ctx,
 }
 
 NvExecutionProvider::PerThreadContext::PerThreadContext(OrtDevice::DeviceId device_id, bool has_user_compute_stream, cudaStream_t stream) {
-  // TODO: figure out if PerThreadContext is used at all. If not, just clean it up.
-  if (has_user_compute_stream) {
+  // Only set device if user hasn't provided a compute stream
+  if (!has_user_compute_stream) {
     CUDA_CALL_THROW(cudaSetDevice(device_id));
-    (void)(stream);
+    (void)stream;
   }
 }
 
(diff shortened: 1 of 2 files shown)


onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
```diff
@@ -1044,6 +1104,12 @@ NvExecutionProvider::NvExecutionProvider(const NvExecutionProviderInfo& info)
     external_stream_ = true;
     stream_ = static_cast<cudaStream_t>(info.user_compute_stream);
   }
+  else if(cuda_graph_enable_) {
+    // Create our own stream for NV TensorRT operations if CUDA graph is enabled
```
ankan-ban, line 1108: how does this interact with user provided stream ?
Capturing the graph on a separate stream and then executing it on the stream provided by the app is fine, but aren't you just overwriting the user provided "stream_" here ?

onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
```diff
@@ -1260,41 +1326,80 @@ NvExecutionProvider::~NvExecutionProvider() {
...
 }
 
+// Umang as we have these in poer thread context we dont need these
 bool NvExecutionProvider::IsGraphCaptureEnabled() const {
   return cuda_graph_enable_;
 }
 
-bool NvExecutionProvider::IsGraphCaptureAllowed() const {
-  return regular_run_count_before_graph_capture_ >= min_num_runs_before_cuda_graph_capture_;
+// bool NvExecutionProvider::IsGraphCaptureAllowed(CudaGraphAnnotation_t cuda_graph_annotation_id) const {
```
ankan-ban, line 1334: please delete any commented/unused code.

onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
```diff
@@ -892,10 +892,10 @@ Status BindKernelOutput(Ort::KernelContext& ctx,
 }
 
 NvExecutionProvider::PerThreadContext::PerThreadContext(OrtDevice::DeviceId device_id, bool has_user_compute_stream, cudaStream_t stream) {
```
ankan-ban, line 894: What is the purpose of this PerThreadContext ? (Sorry, not super famalier with the codebase)

onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
umangb-09: No, we have 3 way flow.
If user has provided, we will use his stream
If he has not provided and we have cuda graph enabled, we will create here
else, we will create in compute function

We can however avoid the check of cuda graph here and simply do
if user has given we will use that else we will create one here and that will only be used 

so we can remove these lines 

void* cuda_stream;
    Ort::ThrowOnError(api->KernelContext_GetGPUComputeStream(context, &cuda_stream));
    cudaStream_t stream = static_cast<cudaStream_t>(cuda_stream);

onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
umangb-09: I am too not fully aware. Based on what I can see from CUDA EP is that Cuda graph capture needs to be managed per-thread for proper capture and reply. So we are using this perThreadContext mainly for that. I have just seen what CUDA EP does and tied to replicate that. 

For Cuda Graph this is what i have as info
CUDA graphs must be captured and replayed on the same thread
Each thread needs its own graph instance to avoid conflicts
The graph state (captured vs not captured) must be maintained per-thread

anujj:
Instead of using nv_cuda_graph_enable in the NV EP, could you use the enable_cuda_graph flag instead? This would give us the same flag everywhere—model_builder.py, GenAI, and the NV EP. GenAI already uses enable_cuda_graph to enable CUDA graphs for the other EPs.

PR #25374, opened 2025-07-12
ORT perf test support for plugin EP

chilo-ms:
Add support for onnxruntime_perf_test to run plugin EP dll.

It can run the ORT's example_plugin_ep.dll built from test/autoep/library/*.cc
````sh
-L "example_ep|C:\Users\lochi\repos\ort_ep_abi_7\build\Windows\Debug\Debug\example_plugin_ep.dll" -e example_ep -r 1 C:\Users\lochi\repos\mul_op_test\mul_1.onnx
````

### DIFF
--- a/onnxruntime/test/perftest/command_args_parser.cc
+++ b/onnxruntime/test/perftest/command_args_parser.cc
@@ -39,8 +39,8 @@ namespace perftest {
       "\t-A: Disable memory arena\n"
       "\t-I: Generate tensor input binding. Free dimensions are treated as 1 unless overridden using -f.\n"
       "\t-c [parallel runs]: Specifies the (max) number of runs to invoke simultaneously. Default:1.\n"
-      "\t-e [cpu|cuda|dnnl|tensorrt|openvino|dml|acl|nnapi|coreml|qnn|snpe|rocm|migraphx|xnnpack|vitisai|webgpu]: Specifies the provider 'cpu','cuda','dnnl','tensorrt', "
-      "'nvtensorrtrtx', 'openvino', 'dml', 'acl', 'nnapi', 'coreml', 'qnn', 'snpe', 'rocm', 'migraphx', 'xnnpack', 'vitisai' or 'webgpu'. "
+      "\t-e [cpu|cuda|dnnl|tensorrt|openvino|dml|acl|nnapi|coreml|qnn|snpe|rocm|migraphx|xnnpack|vitisai|webgpu|plugin_ep]: Specifies the provider 'cpu','cuda','dnnl','tensorrt', "
+      "'nvtensorrtrtx', 'openvino', 'dml', 'acl', 'nnapi', 'coreml', 'qnn', 'snpe', 'rocm', 'migraphx', 'xnnpack', 'vitisai', 'webgpu' or plugin execution provider that provided via ep library. "
       "Default:'cpu'.\n"
       "\t-b [tf|ort]: backend to use. Default:ort\n"
       "\t-r [repeated_times]: Specifies the repeated times if running in 'times' test mode.Default:1000.\n"
@@ -195,7 +195,7 @@ static bool ParseDimensionOverride(std::basic_string<ORTCHAR_T>& dim_identifier,
 
 /*static*/ bool CommandLineParser::ParseArguments(PerformanceTestConfig& test_config, int argc, ORTCHAR_T* argv[]) {
   int ch;
-  while ((ch = getopt(argc, argv, ORT_TSTR("m:e:r:t:p:x:y:c:d:o:u:i:f:F:S:T:C:AMPIDZvhsqznlgR:X"))) != -1) {
+  while ((ch = getopt(argc, argv, ORT_TSTR("m:e:r:t:p:x:y:c:d:o:u:i:f:F:S:T:C:AMPIDZvhsqznlgR:XL:"))) != -1) {
     switch (ch) {
       case 'f': {
         std::basic_string<ORTCHAR_T> dim_name;
@@ -273,7 +273,8 @@ static bool ParseDimensionOverride(std::basic_string<ORTCHAR_T>& dim_identifier,
         } else if (!CompareCString(optarg, ORT_TSTR("nvtensorrtrtx"))) {
           test_config.machine_config.provider_type_name = onnxruntime::kNvTensorRTRTXExecutionProvider;
         } else {
-          return false;
+          // Could be plugin EP, save it first and handle later.
+          test_config.machine_config.provider_type_name = ToUTF8String(optarg);
         }
         break;
       case 'r':
@@ -403,6 +404,9 @@ static bool ParseDimensionOverride(std::basic_string<ORTCHAR_T>& dim_identifier,
         break;
       case 'X':
         test_config.run_config.use_extensions = true;
+        break;
+      case 'L':
+        test_config.plugin_ep_names_and_libs = optarg;
         break;
       case '?':
       case 'h':
--- a/onnxruntime/test/perftest/main.cc
+++ b/onnxruntime/test/perftest/main.cc
@@ -42,6 +42,9 @@ int real_main(int argc, char* argv[]) {
     if (failed)
       return -1;
   }
+
+  auto status = Status::OK();
+
   {
     std::random_device rd;
     perftest::PerformanceRunner perf_runner(env, test_config, rd);
@@ -52,16 +55,22 @@ int real_main(int argc, char* argv[]) {
       return 0;
     }
 
-    auto status = perf_runner.Run();
+    status = perf_runner.Run();
+
     if (!status.IsOK()) {
       printf("Run failed:%s\n", status.ErrorMessage().c_str());
-      return -1;
+    } else {
+      perf_runner.SerializeResult();
     }
-
-    perf_runner.SerializeResult();
   }
 
-  // unregister any plugin ep lib if it's registered
+  // Unregister all registered plugin EP libraries before program exits.
+  //
+  // This is necessary because unregistering the plugin EP also unregisters any associated shared allocators.
+  // If we don't do this first and program returns, the factories stored inside the environment will be destroyed when the environment goes out of scope.
+  // Later, when the shared allocator's deleter runs, it may cause a segmentation fault because it attempts to use the already-destroyed factory to call ReleaseAllocator.
+  //
+  // See "ep_device.ep_factory->ReleaseAllocator" in Environment::CreateSharedAllocatorImpl.
   std::unordered_map<std::string, std::string> ep_names_to_libs;
 #ifdef _MSC_VER
   std::string ep_names_and_libs_string = ToUTF8String(test_config.plugin_ep_names_and_libs);
@@ -72,6 +81,11 @@ int real_main(int argc, char* argv[]) {
   for (auto& pair : ep_names_to_libs) {
     env.UnregisterExecutionProviderLibrary(pair.first.c_str());
   }
+
+  if (!status.IsOK()) {
+    return -1;
+  }
+
   return 0;
 }
 
--- a/onnxruntime/test/perftest/main.cc
+++ b/onnxruntime/test/perftest/main.cc
@@ -6,6 +6,7 @@
 #include <random>
 #include "command_args_parser.h"
 #include "performance_runner.h"
+#include "strings_helper.h"
 #include <google/protobuf/stubs/common.h>
 
 using namespace onnxruntime;
@@ -41,23 +42,36 @@ int real_main(int argc, char* argv[]) {
     if (failed)
       return -1;
   }
-  std::random_device rd;
-  perftest::PerformanceRunner perf_runner(env, test_config, rd);
+  {
+    std::random_device rd;
+    perftest::PerformanceRunner perf_runner(env, test_config, rd);
 
-  // Exit if user enabled -n option so that user can measure session creation time
-  if (test_config.run_config.exit_after_session_creation) {
-    perf_runner.LogSessionCreationTime();
-    return 0;
+    // Exit if user enabled -n option so that user can measure session creation time
+    if (test_config.run_config.exit_after_session_creation) {
+      perf_runner.LogSessionCreationTime();
+      return 0;
+    }
+
+    auto status = perf_runner.Run();
+    if (!status.IsOK()) {
+      printf("Run failed:%s\n", status.ErrorMessage().c_str());
+      return -1;
+    }
+
+    perf_runner.SerializeResult();
   }
 
-  auto status = perf_runner.Run();
-  if (!status.IsOK()) {
-    printf("Run failed:%s\n", status.ErrorMessage().c_str());
-    return -1;
+  // unregister any plugin ep lib if it's registered
+  std::unordered_map<std::string, std::string> ep_names_to_libs;
+#ifdef _MSC_VER
+  std::string ep_names_and_libs_string = ToUTF8String(test_config.plugin_ep_names_and_libs);
+#else
+  std::string ep_names_and_libs_string = performance_test_config.plugin_ep_names_and_libs;
+#endif
+  onnxruntime::perftest::ParseSessionConfigs(ep_names_and_libs_string, ep_names_to_libs);
+  for (auto& pair : ep_names_to_libs) {
+    env.UnregisterExecutionProviderLibrary(pair.first.c_str());
   }
-
-  perf_runner.SerializeResult();
-
   return 0;
 }
 
--- a/onnxruntime/test/perftest/ort_test_session.cc
+++ b/onnxruntime/test/perftest/ort_test_session.cc
@@ -62,6 +62,56 @@ OnnxRuntimeTestSession::OnnxRuntimeTestSession(Ort::Env& env, std::random_device
     : rand_engine_(rd()), input_names_(m.GetInputCount()), input_names_str_(m.GetInputCount()), input_length_(m.GetInputCount()) {
   Ort::SessionOptions session_options;
 
+#ifdef _MSC_VER
+  std::string ep_names_and_libs_string = ToUTF8String(performance_test_config.plugin_ep_names_and_libs);
+#else
+  std::string ep_names_and_libs_string = performance_test_config.plugin_ep_names_and_libs;
+#endif
+  std::unordered_map<std::string, std::string> ep_names_to_libs;
+  ParseSessionConfigs(ep_names_and_libs_string, ep_names_to_libs);
+  bool is_plugin_ep_avaiable = false;
+
+  if (ep_names_to_libs.size() > 0) {
+    // Register plugin EP libraries if provided via "-L" argument.
+    for (auto& pair : ep_names_to_libs) {
+      const std::filesystem::path library_path = pair.second;
+      const std::string registration_name = pair.first;
+      env.RegisterExecutionProviderLibrary(registration_name.c_str(), library_path.c_str());
+      registered_plugin_ep_names_.push_back(registration_name);
+    }
+
+    std::vector<Ort::ConstEpDevice> ep_devices = env.GetEpDevices();
+    std::vector<Ort::ConstEpDevice> added_ep_devices;
+
+    // All OrtEpDevice instances must be from the same execution provider.
+    // Find the OrtEpDevice associated with the execution provider provided via "-e" argument.
+    Ort::ConstEpDevice plugin_ep_device;
+    for (Ort::ConstEpDevice& device : ep_devices) {
+      if (std::string(device.EpName()) == performance_test_config.machine_config.provider_type_name) {
+        plugin_ep_device = device;
+        added_ep_devices.push_back(plugin_ep_device);
+      }
+    }
+
+    if (added_ep_devices.empty()) {
+      for (auto ep_name : registered_plugin_ep_names_) {
+        env.UnregisterExecutionProviderLibrary(ep_name.c_str());
+      }
+      ORT_THROW(
+          "[ERROR] [plugin EP] No matching execution provider name found in EP library's factory.");
+    }
+
+#if defined(_MSC_VER)
+    std::string provider_option_string = ToUTF8String(performance_test_config.run_config.ep_runtime_config_string);
+#else
+    std::string provider_option_string = performance_test_config.run_config.ep_runtime_config_string;
+#endif
+    std::unordered_map<std::string, std::string> provider_options;
+    ParseSessionConfigs(provider_option_string, provider_options);
+    session_options.AppendExecutionProvider_V2(env, added_ep_devices, provider_options);
+    is_plugin_ep_avaiable = true;
+  }
+
   provider_name_ = performance_test_config.machine_config.provider_type_name;
   std::unordered_map<std::string, std::string> provider_options;
   if (provider_name_ == onnxruntime::kDnnlExecutionProvider) {
@@ -574,7 +624,8 @@ select from 'TF8', 'TF16', 'UINT8', 'FLOAT', 'ITENSOR'. \n)");
 #endif
   } else if (!provider_name_.empty() &&
              provider_name_ != onnxruntime::kCpuExecutionProvider &&
-             provider_name_ != onnxruntime::kOpenVINOExecutionProvider) {
+             provider_name_ != onnxruntime::kOpenVINOExecutionProvider &&
+             !is_plugin_ep_avaiable) {
     ORT_THROW("This backend is not included in perf test runner.\n");
   }
 
--- a/onnxruntime/test/perftest/ort_test_session.h
+++ b/onnxruntime/test/perftest/ort_test_session.h
@@ -51,6 +51,7 @@ class OnnxRuntimeTestSession : public TestSession {
   const int input_length_;
   std::string provider_name_;
   std::string device_memory_name_;  // Device memory type name to use from the list in allocator.h
+  std::vector<std::string> registered_plugin_ep_names_;
 };
 
 }  // namespace perftest
--- a/onnxruntime/test/perftest/test_configuration.h
+++ b/onnxruntime/test/perftest/test_configuration.h
@@ -74,6 +74,7 @@ struct PerformanceTestConfig {
   ModelInfo model_info;
   MachineConfig machine_config;
   RunConfig run_config;
+  std::basic_string<ORTCHAR_T> plugin_ep_names_and_libs;
 };
 
 }  // namespace perftest
--- a/onnxruntime/test/perftest/performance_runner.h
+++ b/onnxruntime/test/perftest/performance_runner.h
@@ -117,6 +117,7 @@ class PerformanceRunner {
   std::unique_ptr<ITestCase> test_case_;
 
   std::mutex results_mutex_;
+  bool unregister_ep_lib = false;
 };
 }  // namespace perftest
 }  // namespace onnxruntime

### CODE REVIEW

File: onnxruntime/test/perftest/main.cc [hunk 1]
```diff
@@ -41,23 +42,50 @@ int real_main(int argc, char* argv[]) {
...
+  // See "ep_device.ep_factory->ReleaseAllocator" in Environment::CreateSharedAllocatorImpl.
+  std::unordered_map<std::string, std::string> ep_names_to_libs;
+#ifdef _MSC_VER
+  std::string ep_names_and_libs_string = ToUTF8String(test_config.plugin_ep_names_and_libs);
+#else
+  std::string ep_names_and_libs_string = test_config.plugin_ep_names_and_libs;
+#endif
+  onnxruntime::perftest::ParseSessionConfigs(ep_names_and_libs_string, ep_names_to_libs);
+  for (auto& pair : ep_names_to_libs) {
+    env.UnregisterExecutionProviderLibrary(pair.first.c_str());
```
yuslepukhin: What happens if these lines are not reached due to error exception? Should this be exception safe and automatically dregister on program exit? #Resolved

onnxruntime/test/perftest/ort_test_session.cc
```diff
@@ -62,6 +62,56 @@
     : rand_engine_(rd()), input_names_(m.GetInputCount()), input_names_str_(m.GetInputCount()), input_length_(m.GetInputCount()) {
   Ort::SessionOptions session_options;
 
+#ifdef _MSC_VER
```
yuslepukhin: Here and other places #ifdef is not needed. since ToUTF8 string has a non windows version.
Also, it is not about the compiler, but the OS #Resolved

onnxruntime/test/perftest/main.cc
chilo-ms: You are right, deregistering should also be called upon exception and before program exit, i will add this.

jywu-mysoft:
@javier-intel please review and provide requirements for OVEP here. thanks.

javier-intel:
> @javier-intel please review and provide requirements for OVEP here. thanks.

Hi @jywu-msft, to support OVEP ABI changes in perf_test I suggest introducing new options to facilitate device selection:

- `-list_devices` prints all the index and properties (including metadata) of every entry in env.GetEpDevices()

- `-select_devices` takes a (semicolon separated?) list of device indices, and use that to further list the devices in [this code](https://github.com/microsoft/onnxruntime/pull/25374/files#diff-f3908cf4d2982fd2b008beb8c951d9cab67b28cb07077493b8d1dd8b448d6e18R88-R94) before calling `session_options.AppendExecutionProvider_V2(env, added_ep_devices, provider_options);`

jywu-mysoft:
> > @javier-intel please review and provide requirements for OVEP here. thanks.
> 
> Hi @jywu-msft, to support OVEP ABI changes in perf_test I suggest introducing new options to facilitate device selection:
> 
> * `-list_devices` prints all the index and properties (including metadata) of every entry in env.GetEpDevices()
> * `-select_devices` takes a (semicolon separated?) list of device indices, and use that to further list the devices in [this code](https://github.com/microsoft/onnxruntime/pull/25374/files#diff-f3908cf4d2982fd2b008beb8c951d9cab67b28cb07077493b8d1dd8b448d6e18R88-R94) before calling `session_options.AppendExecutionProvider_V2(env, added_ep_devices, provider_options);`

@chilo-ms

onnxruntime/test/perftest/windows/utils.cc
```diff
@@ -75,6 +78,43 @@ std::unique_ptr<ICPUUsage> CreateICPUUsage() {
...
+    utf8_argv.push_back(&str[0]);  // safe since std::string is mutable
+  }
+  return utf8_argv;
+}
+#endif
+
+std::basic_string<ORTCHAR_T> Utf8ToOrtString(const std::string& utf8_str) {
+  // ORTCHAR_T == char -> just convert to std::basic_string<char>
+  if constexpr (std::is_same_v<ORTCHAR_T, char>) {
+    return std::basic_string<ORTCHAR_T>(utf8_str.begin(), utf8_str.end());
```
edgchen1: this code won't ever be compiled, right?

https://github.com/microsoft/onnxruntime/blob/23910fdc3b8d8a27625e098d7ad9e4d6736639ea/include/onnxruntime/core/session/onnxruntime_c_api.h#L92

for a sanity check, you could add `static_assert(std::is_same_v<ORTCHAR_T, wchar_t>);`.

onnxruntime/test/perftest/windows/utils.cc
```diff
@@ -75,6 +78,43 @@ std::unique_ptr<ICPUUsage> CreateICPUUsage() {
   return std::make_unique<CPUUsage>();
 }
 
+#ifdef _WIN32
```
edgchen1: we don't need `#ifdef _WIN32` in this Windows-only file.

onnxruntime/test/perftest/utils.h
```diff
@@ -22,6 +23,20 @@ class ICPUUsage {
...
+std::vector<const char*> ConvertArgvToUtf8CharPtrs(std::vector<std::string>& utf8_args);
+#endif
+
+std::basic_string<ORTCHAR_T> Utf8ToOrtString(const std::string& utf8_str);
+
+bool RegisterExecutionProviderLibrary(Ort::Env& env, PerformanceTestConfig& test_config);
+
+bool UnregisterExecutionProviderLibrary(Ort::Env& env, PerformanceTestConfig& test_config);
+
+void list_devices(Ort::Env& env);
```
edgchen1: nit: naming convention - `ListDevices`

could the Ort::Env be const?

cmake/onnxruntime_unittests.cmake
```diff
@@ -1247,6 +1255,14 @@ if (NOT onnxruntime_ENABLE_TRAINING_TORCH_INTEROP)
       endif()
     endif()
 
+    if(onnxruntime_MINIMAL_BUILD)
+      add_definitions(-DCXXOPTS_NO_RTTI)
+    endif()
+
+    if(onnxruntime_DISABLE_EXCEPTIONS)
+     add_definitions(-DDISABLE_EXCEPTIONS=1)
```
edgchen1: can we reuse the existing `ORT_NO_EXCEPTIONS` definition?

https://github.com/microsoft/onnxruntime/blob/23910fdc3b8d8a27625e098d7ad9e4d6736639ea/cmake/adjust_global_compile_flags.cmake#L157

onnxruntime/test/perftest/command_args_parser.cc
```diff
@@ -19,6 +22,10 @@
 #include <unistd.h>
 #endif
 
+#else
+#include <cxxopts.hpp>
```
edgchen1: does cxxopts support disabling exceptions? would it be simpler to do that?

https://github.com/jarro2783/cxxopts/blob/5e6d1e29f7546d9d94fe3d193b788b280bf1f37d/include/cxxopts.hpp#L578-L587

perhaps we can just have a single version of `ParseArguments()`

onnxruntime/test/perftest/windows/utils.cc
```diff
@@ -75,6 +78,43 @@ std::unique_ptr<ICPUUsage> CreateICPUUsage() {
...
+  std::vector<const char*> utf8_argv;
+  utf8_argv.reserve(utf8_args.size());
+  for (auto& str : utf8_args) {
+    utf8_argv.push_back(&str[0]);  // safe since std::string is mutable
+  }
+  return utf8_argv;
+}
+#endif
+
+std::basic_string<ORTCHAR_T> Utf8ToOrtString(const std::string& utf8_str) {
```
edgchen1: could probably just use the (perhaps poorly named) helper `ToPathString()`.

onnxruntime/test/perftest/windows/utils.cc
```diff
@@ -75,6 +78,43 @@ std::unique_ptr<ICPUUsage> CreateICPUUsage() {
...
+    utf8_args.push_back(ToUTF8String(argv[i]));
+  }
+  return utf8_args;
+}
+
+std::vector<const char*> ConvertArgvToUtf8CharPtrs(std::vector<std::string>& utf8_args) {
+  std::vector<const char*> utf8_argv;
+  utf8_argv.reserve(utf8_args.size());
+  for (auto& str : utf8_args) {
+    utf8_argv.push_back(&str[0]);  // safe since std::string is mutable
```
edgchen1: `utf8_args` can be const.

onnxruntime/test/perftest/common_utils.cc
```diff
@@ -0,0 +1,64 @@
...
+bool RegisterExecutionProviderLibrary(Ort::Env& env, PerformanceTestConfig& test_config) {
+  if (!test_config.plugin_ep_names_and_libs.empty()) {
+    std::unordered_map<std::string, std::string> ep_names_to_libs;
+    ParseSessionConfigs(ToUTF8String(test_config.plugin_ep_names_and_libs), ep_names_to_libs);
+    if (ep_names_to_libs.size() > 0) {
+      for (auto& pair : ep_names_to_libs) {
+        const std::filesystem::path library_path = pair.second;
+        const std::string registration_name = pair.first;
+        env.RegisterExecutionProviderLibrary(registration_name.c_str(), Utf8ToOrtString(library_path.string()));
+        test_config.registered_plugin_eps.push_back(registration_name);
```
edgchen1: instead of modifying `test_config`, can we just return the list of registered EP names?

onnxruntime/test/perftest/main.cc
```diff
@@ -41,23 +47,73 @@ int real_main(int argc, char* argv[]) {
...
+  // Unregister all registered plugin EP libraries before program exits.
+  //
+  // This is necessary because unregistering the plugin EP also unregisters any associated shared allocators.
+  // If we don't do this first and program returns, the factories stored inside the environment will be destroyed when the environment goes out of scope.
+  // Later, when the shared allocator's deleter runs, it may cause a segmentation fault because it attempts to use the already-destroyed factory to call ReleaseAllocator.
+  //
+  // See "ep_device.ep_factory->ReleaseAllocator" in Environment::CreateSharedAllocatorImpl.
+  if (!test_config.registered_plugin_eps.empty()) {
+    perftest::utils::UnregisterExecutionProviderLibrary(env, test_config);
+  }
```
edgchen1: instead of using a try/catch, consider using an RAII type to manage the cleanup.

for example, with GSL `finally`, we can do something like this:

```c++

if (!test_config.plugin_ep_names_and_libs.empty()) {
  perftest::utils::RegisterExecutionProviderLibrary(env, test_config);
}

auto unregister_plugin_eps_at_scope_exit = gsl::finally([&]() {
  if (!test_config.registered_plugin_eps.empty()) {
    perftest::utils::UnregisterExecutionProviderLibrary(env, test_config);  // TODO ensure that this won't throw since it is called from the gsl::final_action destructor.
  }
});

```

https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#e19-use-a-final_action-object-to-express-cleanup-if-no-suitable-resource-handle-is-available

onnxruntime/test/perftest/ort_test_session.cc
```diff
@@ -62,6 +62,65 @@ OnnxRuntimeTestSession::OnnxRuntimeTestSession(Ort::Env& env, std::random_device
...
+    std::unordered_set<int> added_ep_device_index_set;
+
+    // Select devices by provided device index
+    if (!performance_test_config.selected_devices.empty()) {
+      std::vector<int> device_list;
+      device_list.reserve(performance_test_config.selected_devices.size());
+      ParseDeviceList(performance_test_config.selected_devices, device_list);
+      for (auto index : device_list) {
+        if (static_cast<size_t>(index) > (ep_devices.size() - 1)) {
+          fprintf(stderr, "%s", "The device index provided is not correct. Will skip this device id.");
```
edgchen1, line 82: should it skip the rest of the iteration? otherwise, there will be an out of bounds access at `Ort::ConstEpDevice& device = ep_devices[index];`
```suggestion
          fprintf(stderr, "%s", "The device index provided is not correct. Will skip this device id.");
          continue;
```

onnxruntime/test/perftest/windows/utils.cc
chilo-ms: good catch, i forgot to remove this code after i added the `Utf8ToOrtString` function in posix/utils.cc as well.
Will put a sanity check instead.

onnxruntime/test/perftest/common_utils.cc
chilo-ms: In the constructor of `OnnxRuntimeTestSession`, it checks whether there is ep plugin library registered and then get the ep devices from there. So, i think updating the `test_config` with the registered EP names might be the clear way to save the info. If we return the list of registered EP names, then we might need to pass the list to the constructor of `OnnxRuntimeTestSession`
How do you think?

edgchen1: I see. I guess that's ok then.

onnxruntime/test/perftest/command_args_parser.cc
```diff
@@ -166,15 +161,21 @@ namespace perftest {
       "\t-X [Enable onnxruntime-extensions custom ops]: Registers custom ops from onnxruntime-extensions. "
       "onnxruntime-extensions must have been built in to onnxruntime. This can be done with the build.py "
       "'--use_extensions' option.\n"
+#ifndef DISABLE_EXCEPTIONS
```
edgchen1: is this ifdef needed?

onnxruntime/test/perftest/command_args_parser.h
```diff
@@ -13,6 +13,7 @@ class CommandLineParser {
  public:
   static void ShowUsage();
   static bool ParseArguments(PerformanceTestConfig& test_config, int argc, ORTCHAR_T* argv[]);
+  static bool ParseArgumentsV2(PerformanceTestConfig& test_config, int argc, ORTCHAR_T* argv[]);
```
edgchen1: do we need a separate v2 version or can we just update the old one?

onnxruntime/test/perftest/ort_test_session.h
```diff
@@ -51,6 +51,7 @@ class OnnxRuntimeTestSession : public TestSession {
   const int input_length_;
   std::string provider_name_;
   std::string device_memory_name_;  // Device memory type name to use from the list in allocator.h
+  std::vector<std::string> registered_plugin_ep_names_;
```
edgchen1: is this used?

onnxruntime/test/perftest/command_args_parser.cc
```diff
@@ -193,240 +194,292 @@ static bool ParseDimensionOverride(std::basic_string<ORTCHAR_T>& dim_identifier,
...
+#else
+    auto result = options.parse(argc, argv);
+#endif
+
+    /*
+    if (result.count("help")) {
+      std::cout << options.help() << std::endl;
+      return false;
+    }
+    */
```
edgchen1: remove commented code

onnxruntime/test/perftest/utils.h
```diff
@@ -22,6 +23,18 @@ class ICPUUsage {
 
 std::unique_ptr<ICPUUsage> CreateICPUUsage();
 
+#ifdef _WIN32
+std::vector<std::string> ConvertArgvToUtf8Strings(int argc, wchar_t* argv[]);
+
+std::vector<const char*> ConvertArgvToUtf8CharPtrs(std::vector<std::string>& utf8_args);
```
edgchen1: nit: maybe name it something like `CStringsFromStrings`. it's not doing any utf8 conversion. also the input can be const.

onnxruntime/test/perftest/main.cc
```diff
@@ -41,6 +43,30 @@ int real_main(int argc, char* argv[]) {
...
+      perftest::utils::UnregisterExecutionProviderLibrary(env, test_config);  // this won't throw
+    }
+  });
+
+  if (test_config.list_available_devices) {
+    perftest::utils::ListDevices(env);
+    if (test_config.registered_plugin_eps.empty()) {
+      fprintf(stdout, "No plugin execution provider libraries are registered. Please specify them using \"--plugin_ep_libs\"; otherwise, only CPU may be available.\n");
+    }
+    return 0;
```
edgchen1, line 67: probably worth documenting that this option will make the program exit early. could also name it list_devices_and_exit or something.

onnxruntime/test/perftest/command_args_parser.cc
```diff
@@ -166,15 +161,21 @@ namespace perftest {
       "\t-X [Enable onnxruntime-extensions custom ops]: Registers custom ops from onnxruntime-extensions. "
       "onnxruntime-extensions must have been built in to onnxruntime. This can be done with the build.py "
       "'--use_extensions' option.\n"
+#ifndef DISABLE_EXCEPTIONS
+      "\t--plugin_ep_libs [registration names and libraries] Specifies a list of plugin execution provider(EP) registration names and their corresponding shared libraries to register.\n"
+      "\t    [Usage]: --plugin_ep_libs 'plugin_ep_1|plugin_ep_2.dll plugin_ep_2|plugin_ep_2.dll'\n"
+      "\t--list_devices Prints all available device indices and their properties (including metadata).\n"
```
edgchen1: general - should we call them "EP devices" instead of "devices"?

onnxruntime/test/perftest/command_args_parser.cc
```diff
@@ -193,240 +194,292 @@ static bool ParseDimensionOverride(std::basic_string<ORTCHAR_T>& dim_identifier,
...
+    options.add_options()("Z", "Disable spinning between runs", cxxopts::value<bool>()->default_value("false")->implicit_value("true"));
+    options.add_options()("n", "Exit after session creation", cxxopts::value<bool>()->default_value("false")->implicit_value("true"));
+    options.add_options()("l", "Load model via path", cxxopts::value<bool>()->default_value("false")->implicit_value("true"));
+    options.add_options()("g", "Enable CUDA IO binding", cxxopts::value<bool>()->default_value("false")->implicit_value("true"));
+    options.add_options()("X", "Use extensions", cxxopts::value<bool>()->default_value("false")->implicit_value("true"));
+    options.add_options()("plugin_ep_libs", "Plugin EP names and libs", cxxopts::value<std::string>());
+    options.add_options()("list_devices", "Prints all available device indices and their properties (including metadata)");
+    options.add_options()("select_devices", "A semicolon-separated list of device indices to add to the session and run with", cxxopts::value<std::string>());
+    options.add_options()("h,help", "Print usage");
```
edgchen1: can we just specify the option description in one place? now it is both here and in ShowUsage()

onnxruntime/test/perftest/strings_helper.cc
```diff
@@ -5,6 +5,7 @@
 
 #include <iostream>
 #include <sstream>
+#include <vector>
```
edgchen1: nit: it's already included in the header file

onnxruntime/test/perftest/strings_helper.cc
```diff
@@ -53,5 +54,16 @@ void ParseSessionConfigs(const std::string& configs_string,
...
   }
 }
+
+void ParseDeviceList(const std::string& input, std::vector<int>& result) {
+  std::stringstream ss(input);
+  std::string item;
+
+  while (std::getline(ss, item, ';')) {
+    if (!item.empty()) {
+      result.push_back(std::stoi(item));
```
edgchen1: nit: stoi may not parse the entire string if there is a non-numeric part at the end. it will also allow leading whitespace. consider using std::from_chars or one of the parsing helpers:

https://github.com/microsoft/onnxruntime/blob/223c4a08fad0dd9dee5aae53bf142c386ddf853e/include/onnxruntime/core/common/parse_string.h#L34-L50

https://github.com/microsoft/onnxruntime/blob/223c4a08fad0dd9dee5aae53bf142c386ddf853e/include/onnxruntime/core/common/parse_string.h#L103-L111

cmake/onnxruntime_unittests.cmake
```diff
@@ -1210,6 +1210,14 @@ endif()
 
 if (NOT onnxruntime_ENABLE_TRAINING_TORCH_INTEROP)
   if(NOT IOS)
+    onnxruntime_fetchcontent_declare(
+      cxxopts
```
yuslepukhin: >  cxxopts [](http://example.com/codeflow?start=4&length=9)

Does this require new component registration?
Is it available via vcpkg?

What do we use for IOS? #Resolved

onnxruntime/test/perftest/command_args_parser.cc
```diff
@@ -39,8 +34,8 @@ namespace perftest {
       "\t-A: Disable memory arena\n"
       "\t-I: Generate tensor input binding. Free dimensions are treated as 1 unless overridden using -f.\n"
       "\t-c [parallel runs]: Specifies the (max) number of runs to invoke simultaneously. Default:1.\n"
-      "\t-e [cpu|cuda|dnnl|tensorrt|openvino|dml|acl|nnapi|coreml|qnn|snpe|rocm|migraphx|xnnpack|vitisai|webgpu]: Specifies the provider 'cpu','cuda','dnnl','tensorrt', "
-      "'nvtensorrtrtx', 'openvino', 'dml', 'acl', 'nnapi', 'coreml', 'qnn', 'snpe', 'rocm', 'migraphx', 'xnnpack', 'vitisai' or 'webgpu'. "
+      "\t-e or --ep [cpu|cuda|dnnl|tensorrt|openvino|dml|acl|nnapi|coreml|qnn|snpe|rocm|migraphx|xnnpack|vitisai|webgpu|plugin_ep]: Specifies the provider 'cpu','cuda','dnnl','tensorrt', "
+      "'nvtensorrtrtx', 'openvino', 'dml', 'acl', 'nnapi', 'coreml', 'qnn', 'snpe', 'rocm', 'migraphx', 'xnnpack', 'vitisai', 'webgpu' or plugin execution provider that provided via ep library. "
```
yuslepukhin: >or plugin execution provider that provided via ep library. [](http://example.com/codeflow?start=135&length=58)

It is not clear here that an additional option may be required to specify the library. #Resolved

onnxruntime/test/perftest/command_args_parser.cc
```diff
@@ -166,15 +161,19 @@ namespace perftest {
       "\t-X [Enable onnxruntime-extensions custom ops]: Registers custom ops from onnxruntime-extensions. "
       "onnxruntime-extensions must have been built in to onnxruntime. This can be done with the build.py "
       "'--use_extensions' option.\n"
+      "\t--plugin_ep_libs [registration names and libraries] Specifies a list of plugin execution provider(EP) registration names and their corresponding shared libraries to register.\n"
```
yuslepukhin: >\t--plugin_ep_libs [ [](http://example.com/codeflow?start=7&length=20)

The -e option seems to imply that we can only test with one ep at a time. Here it is a collection. How does this match one with another? #Resolved

onnxruntime/test/perftest/command_args_parser.cc
```diff
@@ -193,240 +192,286 @@ static bool ParseDimensionOverride(std::basic_string<ORTCHAR_T>& dim_identifier,
...
+    auto utf8_argv = utils::CStringsFromStrings(utf8_strings);
+    auto result = options.parse(static_cast<int>(utf8_argv.size()), utf8_argv.data());
+#else
+    auto result = options.parse(argc, argv);
+#endif
+
+    if (result.count("f")) {
+      std::basic_string<ORTCHAR_T> dim_name;
+      int64_t override_val;
+      std::basic_string<ORTCHAR_T> opt_str = ToPathString(result["f"].as<std::string>());
```
yuslepukhin: >std::basic_string<ORTCHAR_T> [](http://example.com/codeflow?start=6&length=28)

Should we use PathString typedef? #Resolved

onnxruntime/test/perftest/ort_test_session.cc
```diff
@@ -62,6 +62,62 @@ OnnxRuntimeTestSession::OnnxRuntimeTestSession(Ort::Env& env, std::random_device
...
+        } else {
+          std::string err_msg = "[Plugin EP] [WARNING] : The EP device index and its corresponding OrtEpDevice is not created from " +
+                                performance_test_config.machine_config.provider_type_name + ". Will skip adding this device.\n";
+          fprintf(stderr, "%s", err_msg.c_str());
+        }
+      }
+    } else {
+      // All OrtEpDevice instances must be from the same execution provider.
+      // Find and select the OrtEpDevice associated with the execution provider provided via "-e" argument.
+      for (int index = 0; static_cast<size_t>(index) < ep_devices.size(); ++index) {
```
yuslepukhin: >int index = 0; [](http://example.com/codeflow?start=11&length=14)

Why not declare index as size_t and avoid cast? #Resolved

cmake/onnxruntime_unittests.cmake
chilo-ms: `cxxopts` is available via vcpkg, but i think ORT relies on `onnxruntime_fetchcontent_declare`/`FetchContent_Declare` and `onnxruntime_fetchcontent_makeavailable`/`FetchContent_MakeAvailable`  to get the 3rd party dependency.

Hmm, from the cmake, we don't build onnxruntime_perf_test for IOS, am i right?

edgchen1: > Hmm, from the cmake, we don't build onnxruntime_perf_test for IOS, am i right?

right.

https://github.com/microsoft/onnxruntime/blob/e0ad8050d04c55318d1336874cc6d640509ddbc0/cmake/onnxruntime_unittests.cmake#L1211-L1232

cmake/onnxruntime_unittests.cmake
```diff
@@ -1245,6 +1253,14 @@ if (NOT onnxruntime_ENABLE_TRAINING_TORCH_INTEROP)
       endif()
     endif()
 
+    if(onnxruntime_MINIMAL_BUILD)
+      add_definitions(-DCXXOPTS_NO_RTTI)
```
edgchen1: nit: use `target_compile_definitions` instead to scope the definitions to the target.
```suggestion
      target_compile_definitions(onnxruntime_perf_test PRIVATE CXXOPTS_NO_RTTI)
```

same thing for CXXOPTS_NO_EXCEPTIONS

onnxruntime/test/perftest/command_args_parser.cc
edgchen1: how are the EP options specified with `-i` interpreted if there is more than one EP?

onnxruntime/test/perftest/command_args_parser.cc
```diff
@@ -193,240 +192,286 @@ static bool ParseDimensionOverride(std::basic_string<ORTCHAR_T>& dim_identifier,
...
+#ifdef _WIN32
+    auto utf8_strings = utils::ConvertArgvToUtf8Strings(argc, argv);
+    auto utf8_argv = utils::CStringsFromStrings(utf8_strings);
+    auto result = options.parse(static_cast<int>(utf8_argv.size()), utf8_argv.data());
+#else
+    auto result = options.parse(argc, argv);
+#endif
+
+    if (result.count("f")) {
+      std::basic_string<ORTCHAR_T> dim_name;
```
edgchen1: since we are converting to utf8 strings first on Windows, maybe we can simplify the string handling in general and just use utf8 std::strings. the exception to this would be for actual file paths, like the model path.

onnxruntime/test/perftest/command_args_parser.cc
chilo-ms: Do we get requests to register more than one plugin EP libraries and test multiple EP devices created by the EPs?
One usage i'm thinking about is the Plugin TRT EP + CUDA EP, which requires two EPs (yes, might need two corresponding `-i`)
If there is no such request, maybe i can make only one ep library to be registered and one ep to test for now.

cmake/onnxruntime_unittests.cmake
snnn: Can we use abseil instead? The cxxopts was only used by training, which is no longer maintained.

yuslepukhin: Why are we not using the existing abseil functionality and instead introducing new one and two different chunks of code?

onnxruntime/test/perftest/command_args_parser.cc
```diff
@@ -4,20 +4,15 @@
 // Licensed under the MIT License.
 
 #include "command_args_parser.h"
+#include "utils.h"
 
 #include <string.h>
 #include <iostream>
 #include <sstream>
 #include <string_view>
 #include <unordered_map>
```
snnn, line 14: Please use abseil instead of cxxopt so that we do not need to introduce a new dependency to ORT inference build.

onnxruntime/test/perftest/command_args_parser.cc
chilo-ms: i added `--plugin_eps` and `--plugin_ep_options` options to support EPs and their provider options.
Please see the PR description for details and usage.
Let me know if you have other suggestions.

The existing `-e` and `-i` are for the provider bridge EP.

onnxruntime/test/perftest/command_args_parser.cc
chilo-ms: That's a good point.
Looking at test_configuration.h, `free_dim_name_overrides `and `free_dim_denotation_overrides` can be changed to use utf8 std::string, and i just updated them.  Other string-based options all have to deal with file paths, so i will leave them as `std::basic_string<ORTCHAR_T>`

onnxruntime/test/perftest/command_args_parser.cc
chilo-ms: updated to use abseil.

cmake/onnxruntime_unittests.cmake
chilo-ms: Updated to use abseil.

onnxruntime/test/perftest/command_args_parser.cc
chilo-ms: updated to use PathString

onnxruntime/test/perftest/command_args_parser.cc
```diff
@@ -26,6 +22,45 @@
 #include "test_configuration.h"
 #include "strings_helper.h"
 
+// Declare flags. See ShowUsage() for detailed option descriptions
```
edgchen1: could we specify the help text here instead of in a separate `ShowUsage()` function? it would be easier to read (without having to switch between two places), and we could also make use of the absl flags help display.

onnxruntime/test/perftest/command_args_parser.cc
```diff
@@ -26,6 +22,45 @@
...
+ABSL_FLAG(std::string, f, "", "");
+ABSL_FLAG(std::string, F, "", "");
+ABSL_FLAG(std::string, m, "", "");
+ABSL_FLAG(std::string, e, "", "");
+ABSL_FLAG(size_t, r, 1000, "");
```
edgchen1: it would be nice if we could just specify the default value in one place. e.g., here:

https://github.com/microsoft/onnxruntime/blob/4a8a289e635f5beabd33d111fd7d31f549300682/onnxruntime/test/perftest/test_configuration.h#L43

could we do something like this? we could also possibly avoid hardcoding the default value in the description and let absl help display it.

```c++
static const PerformanceTestConfig& DefaultPerformanceTestConfig() {
  PerformanceTestConfig default_config{};
  return default_config;
}

...

ABSL_FLAG(size_t, r, DefaultPerformanceTestConfig().run_config.repeated_times, "Specifies the repeated times if running in 'times' test mode.");
```

onnxruntime/test/perftest/command_args_parser.cc
```diff
@@ -193,240 +238,303 @@ static bool ParseDimensionOverride(std::basic_string<ORTCHAR_T>& dim_identifier,
...
+      } else if (opt_str == "nvtensorrtrtx") {
+        test_config.machine_config.provider_type_name = onnxruntime::kNvTensorRTRTXExecutionProvider;
+      } else {
+        return false;
+      }
+    }
+
+    // -r
+    size_t val = absl::GetFlag(FLAGS_r);
+    if (val != static_cast<size_t>(1000) /* default value for absl flag -r */) {
```
edgchen1: can we avoid hardcoding the default value? I think it is fine to not check for the default value and just assign it.

onnxruntime/test/perftest/command_args_parser.cc
```diff
@@ -193,240 +238,303 @@ static bool ParseDimensionOverride(std::basic_string<ORTCHAR_T>& dim_identifier,
...
+          test_config.run_config.optimization_level = ORT_ENABLE_EXTENDED;
+          break;
+        case ORT_ENABLE_LAYOUT:
+          test_config.run_config.optimization_level = ORT_ENABLE_LAYOUT;
+          break;
+        case ORT_ENABLE_ALL:
+          test_config.run_config.optimization_level = ORT_ENABLE_ALL;
+          break;
+        default: {
+          if (val > ORT_ENABLE_ALL) {  // relax constraint
```
edgchen1: I think this should be checking `val_int` instead.

in general, it seems kind of error-prone to use these generic option value variables.

it might be clearer to use variables with more descriptive names and smaller scopes. e.g.:

```c++
{
  const auto optimization_level = absl::GetFlag(FLAGS_o);
  // read optimization_level flag value
}
```

onnxruntime/test/perftest/command_args_parser.cc
```diff
@@ -193,240 +238,303 @@ static bool ParseDimensionOverride(std::basic_string<ORTCHAR_T>& dim_identifier,
...
-      break;
-    default:
+    } else if (positional.size() == 3) {
+      test_config.model_info.model_file_path = ToPathString(positional[1]);
+      test_config.model_info.result_file_path = ToPathString(positional[2]);
+    } else {
       return false;
+    }
+  }
+  ORT_CATCH(const std::exception& ex) {
```
edgchen1: do we need to catch exceptions here? I believe `absl::ParseCommandLine` won't throw.

onnxruntime/test/perftest/ort_test_session.cc
```diff
@@ -62,6 +62,82 @@ OnnxRuntimeTestSession::OnnxRuntimeTestSession(Ort::Env& env, std::random_device
...
+    }
+
+    std::string ep_option_string = ToUTF8String(performance_test_config.run_config.ep_runtime_config_string);
+
+    // A list of EP's associated provider options
+    std::vector<std::unordered_map<std::string, std::string>> ep_options_list;
+    ParseEpOptions(ep_option_string, ep_options_list);
+
+    // If user only provide the EPs' provider options for the first several EPs,
+    // add empty options for the rest EPs.
```
edgchen1: what if the user provides too many EP provider option lists? should that be an error?

onnxruntime/test/perftest/strings_helper.cc
```diff
@@ -53,5 +54,67 @@ void ParseSessionConfigs(const std::string& configs_string,
...
+ * @param input The input string to split.
+ * @param delim The delimiter character to split on.
+ * @param out   The output vector to store the resulting substrings. It will be appended to, not cleared.
+ *
+ * @example
+ *   std::vector<std::string> tokens;
+ *   SplitAndHandleEmptyTokens(";a|b;;x|y;", ';', tokens);
+ *   // tokens = ["", "a|b", "", "x|y", ""]
+ */
+void SplitAndHandleEmptyTokens(const std::string& input, char delim, std::vector<std::string>& out) {
```
edgchen1: FYI, we have this helper function:

https://github.com/microsoft/onnxruntime/blob/4a8a289e635f5beabd33d111fd7d31f549300682/onnxruntime/core/common/string_utils.h#L18-L41

cmake/onnxruntime_unittests.cmake
snnn: Thanks!

onnxruntime/core/platform/windows/debug_alloc.cc
```diff
@@ -254,7 +274,13 @@ Memory_LeakCheck::~Memory_LeakCheck() {
...
         string.find("testing::internal::ThreadLocalRegistryImpl::GetValueOnCurrentThread") == std::string::npos &&
         string.find("PyInit_onnxruntime_pybind11_state") == std::string::npos &&
-        string.find("google::protobuf::internal::InitProtobufDefaultsSlow") == std::string::npos) {
+        string.find("google::protobuf::internal::InitProtobufDefaultsSlow") == std::string::npos &&
+        string.find("flags_internal::ParseCommandLineImpl") == std::string::npos &&
+        string.find("flags_internal::FlagImpl::Init") == std::string::npos &&
+        string.find("SetFlagsUsageConfig") == std::string::npos &&
+        string.find("perftest::utils::ConvertArgvToUtf8Strings") == std::string::npos &&
+        string.find("perftest::utils::CStringsFromStrings") == std::string::npos &&
+        string.find("perftest::PerformanceTestConfig::PerformanceTestConfig") == std::string::npos) {
```
edgchen1: I'm not sure if it is worth adding special cases here for running `onnxruntime_perf_test --help` with a Debug build. when exiting early from `--help`, I think leaking is fine. also, I'd expect that most users would not use a Debug build for performance reasons.

is the memleak debug output too distracting? maybe we can consider other options, like dumping it to a file. this can be a refinement for another PR.

onnxruntime/test/perftest/command_args_parser.cc
edgchen1: cool, thanks for updating this.

onnxruntime/test/perftest/utils.h
```diff
@@ -22,6 +23,18 @@ class ICPUUsage {
...
 
+std::vector<std::string> ConvertArgvToUtf8Strings(int argc, ORTCHAR_T* argv[]);
+
+std::vector<char*> CStringsFromStrings(std::vector<std::string>& utf8_args);
+
+void RegisterExecutionProviderLibrary(Ort::Env& env, PerformanceTestConfig& test_config);
+
+void UnregisterExecutionProviderLibrary(Ort::Env& env, PerformanceTestConfig& test_config);
+
+void ListDevices(const Ort::Env& env);
```
edgchen1: nit: "ListEpDevices"

onnxruntime/test/perftest/test_configuration.h
```diff
@@ -74,6 +75,10 @@ struct PerformanceTestConfig {
   ModelInfo model_info;
   MachineConfig machine_config;
   RunConfig run_config;
+  std::basic_string<ORTCHAR_T> plugin_ep_names_and_libs;
+  std::vector<std::string> registered_plugin_eps;
+  std::string selected_devices;
+  bool list_available_devices = false;
```
edgchen1: nit: call them EP devices instead. e.g.,
```suggestion
  std::string selected_ep_device_indices;
  bool list_available_ep_devices = false;
```

onnxruntime/test/perftest/utils.h
```diff
@@ -22,6 +23,18 @@ class ICPUUsage {
...
+
+std::string_view GetBasename(std::string_view filename);
```
edgchen1: can we use `std::filesystem::path::filename()`?

onnxruntime/test/perftest/strings_helper.h
```diff
@@ -5,12 +5,19 @@
...
 
 void ParseSessionConfigs(const std::string& configs_string,
                          std::unordered_map<std::string, std::string>& session_configs,
                          const std::unordered_set<std::string>& available_keys = {});
+
+void ParseEpList(const std::string& input, std::vector<std::string>& result);
+
+void ParseEpOptions(const std::string& input, std::vector<std::unordered_map<std::string, std::string>>& result);
+
+void ParseEpDeviceList(const std::string& input, std::vector<int>& result);
```
edgchen1: nit: "ParseEpDeviceIndexList" or "ParseIntList"

onnxruntime/core/platform/windows/debug_alloc.cc
chilo-ms: The memleak output is lengthy and distracting.
Agree that we can have a refinement for it in another PR.

PR #25383, opened 2025-07-14
[WebNN] Add rank range validation for rest ops

Honry:
- Add common rank range validation to base_op_builder.cc
- Handle specific rank range validation for rest ops
- Remove duplicated input_shape validation
- Fix some typos BTW

### DIFF
--- a/onnxruntime/core/providers/webnn/builders/impl/einsum_op_builder.cc
+++ b/onnxruntime/core/providers/webnn/builders/impl/einsum_op_builder.cc
@@ -28,6 +28,8 @@ class EinsumOpBuilder : public BaseOpBuilder {
                          const WebnnDeviceType /* device_type */, const logging::Logger& logger) const override;
   bool HasSupportedInputsImpl(const GraphViewer&, const Node& node,
                               const emscripten::val& wnn_limits, const logging::Logger& logger) const override;
+  bool HasSupportedOutputsImpl(const Node& /* node */, const emscripten::val& /* wnn_limits */,
+                               const logging::Logger& /* logger */) const override;
 };
 
 // Helper functions, thanks for DML EP's OperatorHelper.
@@ -40,12 +42,6 @@ enum class RecognizedOperatorType {
   Multiply,
   Pairwise,
   Total,
-};
-
-struct RecognizedOperatorInfo {
-  RecognizedOperatorType recognized_operator_type;
-  std::initializer_list<uint32_t> component_ranks;
-  std::initializer_list<uint32_t> label_indices;
 };
 
 struct Component {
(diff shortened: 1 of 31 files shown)


Honry:

onnxruntime/core/providers/webnn/builders/impl/einsum_op_builder.cc
```diff
@@ -724,23 +714,21 @@ bool EinsumOpBuilder::IsOpSupportedImpl(const GraphViewer&,
     return false;
   }
 
-  RecognizedOperatorType recognized_operator_type = DetermineRecognizedOperatorType(label_indices, components,
```
fdwr, line 727: ~~This removal seems orthogonal to rank validation? Don't we still need to check for unsupported patterns?~~

Ah, maybe it's redundant from the check below.

onnxruntime/core/providers/webnn/builders/impl/einsum_op_builder.cc
```diff
@@ -770,17 +765,54 @@ bool EinsumOpBuilder::HasSupportedInputsImpl(const GraphViewer&, const Node& nod
...
-    return IsDataTypeSupportedByWebNNOp(op_type, "matmul", input0_type, wnn_limits, "a", "inputs", logger);
+  } else if (recognized_operator_type == RecognizedOperatorType::Multiply) {
+    decomposed_op_type = "Mul";
   } else if (recognized_operator_type == RecognizedOperatorType::ReduceSum) {
-    return IsDataTypeSupportedByWebNNOp(op_type, "reduceSum", input0_type, wnn_limits, "input", "inputs", logger);
-  } else {
-    return IsDataTypeSupportedByWebNNOp(op_type, "identity", input0_type, wnn_limits, "input", "inputs", logger);
+    decomposed_op_type = "ReduceSum";
+  } else if (recognized_operator_type == RecognizedOperatorType::Diagonal) {
+    decomposed_op_type = "Trilu";
```
fdwr, line 777: This doesn't seem right. Trilu masks a given input to keep the upper or lower triangle, but an Einstein summation diagonal would be one of the following equations `One repeated character in input, ii->i / iij->ij / iijk -> ijk` from https://github.com/microsoft/onnxruntime/blob/fab30696f314d75d450b87a4755b3ab63a1520d6/onnxruntime/core/providers/webnn/builders/impl/einsum_op_builder.cc#L493 which is not the same. Variations include:

| Formula | Note |
|----------|-------|
| ('ii->i', A2) |	diag(A2)	view main diagonal of A2 |
| ('ii->', A2) |	trace(A2)	sums main diagonal of A2 |
| ("ii,ii->i", A2, B2) |	A2.diag() * B2.diag()	diagonals multiplied by each other |
| ("ii,ii->", A2, B2) |	dot(A2.diag(), B2.diag())	dot product of diagonals |

fdwr: Actually I see now how `trilu` is utilized, achieving a trace via two trilu calls for the upper and lower sides, then reduceSum. So, it's good.

fdwr (approved):
👍

onnxruntime/core/providers/webnn/builders/impl/einsum_op_builder.cc
Honry: Yes, it is.

PR #25389, opened 2025-07-14
Update the customer op API ReadOpAttr for string type to avoid adding '\0' to the end

HectorSVC:
Update the customer op API ReadOpAttr for string type to avoid adding '\0' at the end.

### DIFF
--- a/onnxruntime/core/session/custom_ops.cc
+++ b/onnxruntime/core/session/custom_ops.cc
@@ -436,23 +436,20 @@ ORT_API_STATUS_IMPL(OrtApis::ReadOpAttr, _In_ const OrtOpAttr* op_attr, _In_ Ort
       }
       case OrtOpAttrType::ORT_OP_ATTR_STRING: {
         const auto& s = attr->s();
-        if (len < s.size() + 1) {
+        if (len < s.size()) {
           ret = OrtApis::CreateStatus(OrtErrorCode::ORT_INVALID_ARGUMENT,
                                       "Size of data not large enough to hold the string.");
         } else {
           char* output_c = reinterpret_cast<char*>(data);
-          for (char c : s) {
-            *output_c++ = c;
-          }
-          *output_c = '\0';
+          memcpy(output_c, s.data(), s.size());
         }
-        *out = s.size() + 1;
+        *out = s.size();
         break;
       }
       case OrtOpAttrType::ORT_OP_ATTR_STRINGS: {
         const auto& ss = attr->strings();
         size_t num_bytes = 0;
-        for_each(ss.begin(), ss.end(), [&num_bytes](const std::string& s) { num_bytes += s.size() + 1; });
+        for_each(ss.begin(), ss.end(), [&num_bytes](const std::string& s) { num_bytes += s.size(); });
 
         if (len < num_bytes) {
           ret = OrtApis::CreateStatus(OrtErrorCode::ORT_INVALID_ARGUMENT,
@@ -460,10 +457,7 @@ ORT_API_STATUS_IMPL(OrtApis::ReadOpAttr, _In_ const OrtOpAttr* op_attr, _In_ Ort
         } else {
           char* output_c = reinterpret_cast<char*>(data);
           for (const auto& s : ss) {
-            for (char c : s) {
-              *output_c++ = c;
-            }
-            *output_c++ = '\0';
+            memcpy(output_c, s.data(), s.size());
           }
         }
         *out = num_bytes;


onnxruntime/test/shared_lib/custom_op_utils.cc
```diff
@@ -453,8 +453,9 @@ void StandaloneCustomKernel::InitGru() {
   float betas[1] = {2.f};
   Ort::OpAttr activation_beta = Ort::OpAttr("activation_beta ", betas, 1, OrtOpAttrType::ORT_OP_ATTR_FLOATS);
 
-  const char* direction_string = "bidirectional";
-  Ort::OpAttr direction = Ort::OpAttr("direction", direction_string, 1, OrtOpAttrType::ORT_OP_ATTR_STRING);
+  const std::string direction_string = "bidirectional";
+  Ort::OpAttr direction = Ort::OpAttr("direction", direction_string.c_str(), static_cast<int>(direction_string.length()),
```
adrianlizarraga, line 457: I don't believe the second argument (`len`) needs to be changed.

https://github.com/microsoft/onnxruntime/blob/58954ba7a969d7200365498f0d77b24da51e98ac/include/onnxruntime/core/session/onnxruntime_c_api.h#L3670

The implementation of `CreateOpAttr` ignores the `len` parameter for `ORT_OP_ATTR_STRING` and relies on the null-terminator to get the string length:

https://github.com/microsoft/onnxruntime/blob/58954ba7a969d7200365498f0d77b24da51e98ac/onnxruntime/core/session/standalone_op_invoker.cc#L348

adrianlizarraga: Ah It looks like `CreateOpAttr` was updated. Perhaps should update the CreateOpAttr documentation to state that the len is not ignored for string attributes.

Note: this would be a breaking change for previous users of the CreateOpAttr API.

HectorSVC: The doc was 
\param[in] len Number of bytes stored in data.
Then your PR changed it to:
\param[in] len Number of elements if data represents an array (e.g., ORT_OP_ATTR_INTS). Otherwise, set to 1.

The changes aligns with the old comments. But I really have no idea how user used it. Maybe the safest way is to add new API v2 for this.

yuslepukhin:
>       str->resize(total_attr_bytes, '\0');

I believe this just wastes cycles.

---
Refers to: include/onnxruntime/core/providers/utils/ort_graph_to_proto.h:668 in 3863990. [](commit_id = 3863990b71bb74d7d3ad08e31b1cd3849e775139, deletion_comment = False)

onnxruntime/core/session/custom_ops.cc
```diff
@@ -436,17 +436,14 @@ ORT_API_STATUS_IMPL(OrtApis::ReadOpAttr, _In_ const OrtOpAttr* op_attr, _In_ Ort
       }
       case OrtOpAttrType::ORT_OP_ATTR_STRING: {
         const auto& s = attr->s();
-        if (len < s.size() + 1) {
+        if (len < s.size()) {
```
yuslepukhin, line 439: >len < s.size()) [](http://example.com/codeflow?start=12&length=15)

Should this be an exact match to avoid trailing garbage?

onnxruntime/core/session/standalone_op_invoker.cc
```diff
@@ -345,7 +345,7 @@ onnxruntime::Status CreateOpAttr(const char* name, const void* data, int len, Or
       attr->set_type(ONNX_NAMESPACE::AttributeProto_AttributeType::AttributeProto_AttributeType_FLOATS);
       break;
     case OrtOpAttrType::ORT_OP_ATTR_STRING:
-      attr->set_s(std::string{str});
+      attr->set_s(std::string{str, str + len});
```
yuslepukhin: >str + len [](http://example.com/codeflow?start=35&length=9)

This should be (str, len) ? Or should this be an assign() call?

onnxruntime/core/session/custom_ops.cc
HectorSVC: from the existing API, it seems like it allows user provide a buffer with more room to carry the data, and the API return the real length via out parameter.

onnxruntime/core/session/standalone_op_invoker.cc
HectorSVC: good catch. will update

HectorSVC:
>       str->resize(total_attr_bytes, '\0');

will remove it.

---
In reply to: [3080645591](https://github.com/microsoft/onnxruntime/pull/25389#issuecomment-3080645591) [](http://example.com/codeflow?ancestors=3080645591)

---
Refers to: include/onnxruntime/core/providers/utils/ort_graph_to_proto.h:668 in 3863990. [](commit_id = 3863990b71bb74d7d3ad08e31b1cd3849e775139, deletion_comment = False)

yuslepukhin:
> > ```
> >   str->resize(total_attr_bytes, '\0');
> > ```
> 
> will remove it.
> 
`resize()` may still be necessary, but filling it with zeros, not really.

onnxruntime/test/shared_lib/custom_op_utils.cc
HectorSVC: updated the API doc

jywu-msft (approved):

snnn:

PR #25395, opened 2025-07-15
[webgpu] Optimize FlashAttention for prefill

daijh:
### Description
This PR enhances unidirectional `FlashAttention` by applying causal masking inside the main loop. This optimization eliminates unnecessary memory loads by avoiding future entries in the KV cache.

Testing on Lunar Lake shows up to a 20% performance improvement for `phi-4-mini-accuracy4` (with a prompt of 4096). Similar performance gains were also observed for other models, including `Qwen3-0.6B-accuracy4`.

This PR now uses the more readable `unidirectional` attribute instead of `is_gqa`, to control causal masking.

### Motivation and Context
See above.

### DIFF
--- a/onnxruntime/contrib_ops/webgpu/bert/attention_common.h
+++ b/onnxruntime/contrib_ops/webgpu/bert/attention_common.h
@@ -84,7 +84,7 @@ struct WebgpuAttentionParameters {
   int v_head_size_ = 0;            // hidden size per head of V
   int num_heads_ = 0;
   int rotary_embedding_ = 0;
-  bool is_unidirectional_ = false;
+  bool is_unidirectional_ = true;
   bool past_present_share_buffer_ = false;
   bool do_rotary_ = false;
   bool broadcast_attn_bias_dim_0_ = false;
--- a/onnxruntime/contrib_ops/webgpu/bert/attention_common.h
+++ b/onnxruntime/contrib_ops/webgpu/bert/attention_common.h
@@ -50,6 +50,7 @@ struct WebgpuAttentionParameters {
                                                                                                        v_hidden_size_(parameters.kv_hidden_size),
                                                                                                        v_head_size_(parameters.kv_hidden_size / parameters.kv_num_heads),
                                                                                                        num_heads_(parameters.num_heads),
+                                                                                                       is_unidirectional_(true),
                                                                                                        do_rotary_(parameters.do_rotary),
                                                                                                        scale_(parameters.scale),
                                                                                                        seqlen_past_kv_cache_(parameters.seqlen_past_kv_cache),
@@ -84,7 +85,7 @@ struct WebgpuAttentionParameters {
   int v_head_size_ = 0;            // hidden size per head of V
   int num_heads_ = 0;
   int rotary_embedding_ = 0;
-  bool is_unidirectional_ = true;
+  bool is_unidirectional_ = false;
   bool past_present_share_buffer_ = false;
   bool do_rotary_ = false;
   bool broadcast_attn_bias_dim_0_ = false;
--- a/onnxruntime/contrib_ops/webgpu/bert/flash_attention.cc
+++ b/onnxruntime/contrib_ops/webgpu/bert/flash_attention.cc
@@ -275,8 +275,23 @@ Status FlashAttentionProgram::GenerateShaderCode(ShaderHelper& shader) const {
 
   var previous_max : q_element_t = min_value;
   var previous_denom : q_element_t = 0;
+)MAIN_FN";
 
-  for(var k_start = 0u; k_start < uniforms.total_sequence_length; k_start+=capped_sg_size)
+  if (is_unidirectional_) {
+    // If attention is unidirectional, set the loop bound to enforce causal masking.
+    shader.MainFunctionBody() << R"MAIN_FN(
+  let max_causal_len_for_workgroup = uniforms.past_sequence_length +
+                                     (workgroup_idx % uniforms.num_seq_tile + 1) * workgroup_size_x;
+  let loop_bound = min(uniforms.total_sequence_length, max_causal_len_for_workgroup);
+)MAIN_FN";
+  } else {
+    shader.MainFunctionBody() << R"MAIN_FN(
+  let loop_bound = uniforms.total_sequence_length;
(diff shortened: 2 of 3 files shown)


daijh:
Lunar Lake, Phi-4-mini-accuracy4:

| Prompt          | Default Prefill Speed (tps) | Opt Prefill Speed (tps) | Improvement |
| :-------------- | :-------------------------: | :---------------------: | :---------: |
| **Prompt-1024** |           561.40            |         593.76          |    5.76%    |
| **Prompt-2048** |           498.60            |         549.59          |   10.23%    |
| **Prompt-3072** |           465.98            |         537.64          |   15.38%    |
| **Prompt-4096** |           430.61            |         513.40          |   19.23%    |

daijh:
@sushraja-msft @qjia7 pls take a look.

cc @jchen10 @xhcao

onnxruntime/contrib_ops/webgpu/bert/flash_attention.cc
```diff
@@ -275,8 +275,23 @@ Status FlashAttentionProgram::GenerateShaderCode(ShaderHelper& shader) const {
...
   var previous_max : q_element_t = min_value;
   var previous_denom : q_element_t = 0;
+)MAIN_FN";
 
-  for(var k_start = 0u; k_start < uniforms.total_sequence_length; k_start+=capped_sg_size)
+  if (is_unidirectional_) {
+    // If attention is unidirectional, set the loop bound to enforce causal masking.
+    shader.MainFunctionBody() << R"MAIN_FN(
+  let max_causal_len_for_workgroup = uniforms.past_sequence_length +
+                                     (workgroup_idx % uniforms.num_seq_tile + 1) * workgroup_size_x;
```
qjia7, line 284: `(workgroup_idx % uniforms.num_seq_tile + 1) * workgroup_size_x` is equal to sequence_length, right? But total_sequnce_length = past_sequence_length + sequence_length. I suppose max_causal_len_for_workgroup is equal to total_sequence_length. Why you can see that max_causal_len_for_workgroup is less than total_sequence_length?

qjia7: Sorry, my misunderstanding. `(workgroup_idx % uniforms.num_seq_tile + 1) * workgroup_size_x` is the currently `seq_tile` which is less than sequence_length.

onnxruntime/contrib_ops/webgpu/bert/flash_attention.cc
```diff
@@ -275,8 +275,23 @@ Status FlashAttentionProgram::GenerateShaderCode(ShaderHelper& shader) const {
 
   var previous_max : q_element_t = min_value;
   var previous_denom : q_element_t = 0;
+)MAIN_FN";
 
-  for(var k_start = 0u; k_start < uniforms.total_sequence_length; k_start+=capped_sg_size)
+  if (is_unidirectional_) {
```
qjia7, line 280: Why you can't keep to use `uniforms.is_gqa` but use `is_unidirectional_`?

onnxruntime/contrib_ops/webgpu/bert/attention_common.h
```diff
@@ -84,7 +84,7 @@ struct WebgpuAttentionParameters {
   int v_head_size_ = 0;            // hidden size per head of V
   int num_heads_ = 0;
   int rotary_embedding_ = 0;
-  bool is_unidirectional_ = false;
+  bool is_unidirectional_ = true;
```
qjia7: For MHA, this value should be get from attribute. You can't just set it to true.

onnxruntime/contrib_ops/webgpu/bert/flash_attention.cc
daijh: The only purpose of `is_gqa` in the shader is to check if the attention mechanism is unidirectional. 
Renaming it to `is_unidirectional` would be more accurate and self-documenting.

The current name also implies that:
- Grouped-Query Attention (GQA) is always unidirectional 
- and Multi-Head Attention (MHA) is always bidirectional. 

The assumption would change, that we may want to apply a unidirectional to MHA in the future.

onnxruntime/contrib_ops/webgpu/bert/attention_common.h
daijh: MHA always set this attribute [1], while GQA will use the default value [2].

[1] https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/contrib_ops/webgpu/bert/attention_common.h#L31

[2] https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/contrib_ops/webgpu/bert/attention_common.h#L42

qjia7: Thanks for your info. I prefer we keep the default value as false since for MHA, default value is 0. And explicitly set it to true for GQA.

onnxruntime/contrib_ops/webgpu/bert/flash_attention.cc
```diff
@@ -337,7 +352,7 @@ Status FlashAttentionProgram::GenerateShaderCode(ShaderHelper& shader) const {
       qk_4 = qk_4 + loadAttentionBias(q_idx_global, k_start+12, head_idx);
     }
 
-    let seq_causal_length = select(uniforms.total_sequence_length, uniforms.past_sequence_length + q_idx_global + 1, uniforms.is_gqa > 0);
+    let seq_causal_length = select(uniforms.total_sequence_length, uniforms.past_sequence_length + q_idx_global + 1, uniforms.is_unidirectional > 0);
```
qjia7, line 355: nit: Is it better if we move this assignment out of the for loop?

qjia7 (dismissed):
LGTM with nits.

onnxruntime/contrib_ops/webgpu/bert/flash_attention.cc
daijh: Sure. Let me address it in follow-up PRs to keep this one focused.

daijh:
@guschmue @fs-eire 
Please take a look.

fs-eire (approved):

PR #25397, opened 2025-07-15
Fix 2 device discovery issues.

skottmckay:
### Description

Fix vendor and device id conversion from SetupApi info.
Detect Remote Display Adapter and skip. This results in a bogus device appearing when you're connected to a machine using remote desktop.

### DIFF
--- a/onnxruntime/core/platform/windows/device_discovery.cc
+++ b/onnxruntime/core/platform/windows/device_discovery.cc
@@ -91,7 +91,8 @@ uint64_t GetLuidKey(LUID luid) {
 
 // returns info for display and processor entries. key is (vendor_id << 32 | device_id)
 // npus: (vendor_id << 32 | device_id) for devices we think are NPUs from DXCORE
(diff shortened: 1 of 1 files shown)


onnxruntime/core/platform/windows/device_discovery.cc
```diff
@@ -327,6 +312,12 @@ std::unordered_map<uint64_t, DeviceInfo> GetDeviceInfoD3D12() {
       continue;
     }
 
+    ComPtr<IDXGIOutput> output;
+    if (adapter->EnumOutputs(0, &output) == DXGI_ERROR_NOT_FOUND) {
+      // D3D_DRIVER_TYPE_WARP. Software based. Most likely Microsoft Remote Display Adapter so we skip
+      continue;
+    }
```
skottmckay: https://learn.microsoft.com/en-us/windows/win32/api/dxgi/nf-dxgi-idxgiadapter-enumoutputs

> If the adapter came from a device created using D3D_DRIVER_TYPE_WARP, then the adapter has no outputs, so DXGI_ERROR_NOT_FOUND is returned.

jywu-msft (approved):

snnn:

PR #25398, opened 2025-07-15
fix shape inference error for ep context nodes

wcy123:
### Description

To support writing an Execution Provider (EP) using the new EP ABI introduced in #24887, this PR adds value info for EP Context nodes to prevent shape inference errors during `Graph::Resolve`.

### Motivation and Context

When creating a new EP Context node whose input is the output of another EP Context node, Graph::Resolve fails to set the type for the new node's arguments. This is because EP Context nodes do not have a TypeAndShapeInferenceFunction defined, as shown here:

https://github.com/microsoft/onnxruntime/blob/5fdd4e4f2a2b6705a9a49a378a3b3496805067ee/onnxruntime/core/graph/contrib_ops/contrib_defs.cc#L3289-L3337

As a result, an exception is thrown during shape inference:

https://github.com/microsoft/onnxruntime/blob/5fdd4e4f2a2b6705a9a49a378a3b3496805067ee/onnxruntime/core/graph/graph.cc#L2964

Specifically:

EP Context nodes lack TypeAndShapeInferenceFunction, so onnx_inferred_type is unavailable.
existing_type is nullptr due to the logic in:
https://github.com/microsoft/onnxruntime/blob/9de58ac7a3d18d6ae7f7ae502b3f91361067f1b5/onnxruntime/core/session/ep_plugin_provider_interfaces.cc#L279
https://github.com/microsoft/onnxruntime/blob/9de58ac7a3d18d6ae7f7ae502b3f91361067f1b5/onnxruntime/core/session/ep_plugin_provider_interfaces.cc#L285

### Implementation
This PR attempts to add type information to EP Context nodes with best effort, ensuring that Graph::Resolve can proceed without errors even when type inference is not explicitly defined.

### DIFF
--- a/onnxruntime/core/session/ep_plugin_provider_interfaces.cc
+++ b/onnxruntime/core/session/ep_plugin_provider_interfaces.cc
@@ -244,11 +244,12 @@ Status PluginExecutionProvider::FusedNodeState::AddFusedNode(const Node& fused_n
 /// Note that the EP plugin uses the model editor API to create the OrtNode instances.
 /// </summary>
 /// <param name="ep_name">Name of the plugin EP.</param>
+/// <param name="fused_nodes">fused nodes provided by ORT.</param>
 /// <param name="plugin_ep_context_nodes">EPContext nodes provided by the plugin EP.</param>
 /// <param name="result_nodes">Output parameter set to the resulting array of EPContext nodes.</param>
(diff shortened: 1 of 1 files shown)


HectorSVC (approved):
:shipit:

snnn:

PR #25404, opened 2025-07-15
Update fusion_attention to properly convert bfloat16 values

justinchuby:


onnxruntime/python/tools/transformers/fusion_attention.py
```diff
@@ -362,15 +367,15 @@ def create_combined_qkv_bias(
         name_prefix: str,
     ) -> NodeProto | None:
         q_bias = self.model.get_initializer(q_add.input[1]) or self.model.get_initializer(q_add.input[0])
-        qb = NumpyHelper.to_array(q_bias)
+        qb = to_array(q_bias)
```
kunal-vaishnavi: Instead of replacing all of the `NumpyHelper` references, can we instead update the APIs inside `NumpyHelper` to use the ONNX IR? Otherwise, there may be downstream effects where some places use `ir.from_proto(tensor).numpy()` and other places use `NumpyHelper.to_array`.

https://github.com/microsoft/onnxruntime/blob/1e5fdd1224a445afaff7741da5841464d51c7d83/onnxruntime/python/tools/transformers/fusion_utils.py#L306-L317

justinchuby: Done

onnxruntime/python/tools/transformers/fusion_utils.py
```diff
@@ -5,9 +5,10 @@
 from logging import getLogger
 
 import numpy
+import onnx
+import onnx_ir as ir
```
tianleiwu: Could we import it conditionally (like when the data type is bf16, fp8, fp4, int4x2, uint4x2 etc) in NumpyHelper class? In this way, user might not need to install it when they optimize models of float/fp16 data types.

I think CI pipeline need install the package. Need add it to https://github.com/microsoft/onnxruntime/blob/main/tools/ci_build/requirements/transformers-test/requirements.txt.

Also, we can add it to onnxruntime extra dependency, like add a section of "transformers" here:
https://github.com/microsoft/onnxruntime/blob/2911e70ec9af62d2036477241f5f94db75d63f28/setup.py#L784

justinchuby: Would it make sense to always import it? We are promoting onnx_ir as a standard component.

tianleiwu: It's better to have smooth transition, since it will break user's existing script if they upgrade ort.

justinchuby: Done

onnxruntime/python/tools/transformers/fusion_utils.py
```diff
@@ -19,7 +19,7 @@ def __init__(self, model: OnnxModel):
 
     def cast_graph_input_to_int32(self, input_name: str) -> tuple[bool, str]:
         graph_input = self.model.find_graph_input(input_name)
-        if graph_input is not None and graph_input.type.tensor_type.elem_type != TensorProto.INT32:
+        if graph_input is not None and graph_input.type.tensor_type.elem_type != onnx.TensorProto.INT32:
```
kunal-vaishnavi: Why do we need to change all of the `TensorProto` calls to be `onnx.TensorProto`? Is `from onnx import TensorProto` and then `TensorProto.dtype` not sufficient?

justinchuby: I was cleaning up the imports. Reverted

onnxruntime/python/tools/transformers/fusion_base.py
```diff
@@ -95,7 +95,7 @@ def add_initializer(self, name: str, data_type: int, dims: Sequence[int], vals:
             if not isinstance(vals, np.ndarray):
                 bytes = np.array(vals, dtype=np_type).tobytes()
             else:
-                bytes = vals.astype(np_type).tobytes()
```
kunal-vaishnavi, line 98: Why is type casting no longer needed?

justinchuby: vals are already in the desired dtype (due to how we handle the arrays in the updated logic). So there is no need to cast again. The dtype should really only be used when the input is a list. I updated the logic slightly to reflect that. I think this is a compromise we need to make in the impl for this api to keep it simple.

PR #25405, opened 2025-07-15
increase timeout for nuget-test-android pipeline

prathikr:
Pipeline has been timing out after 1hr for the past several runs on [zip-nuget-pipeline](https://aiinfra.visualstudio.com/Lotus/_build?definitionId=940&_a=summary). Increasing from 1hr to 2hr in accordance with linux-test-pipeline

### DIFF
--- a/tools/ci_build/github/azure-pipelines/nuget/templates/test_android.yml
+++ b/tools/ci_build/github/azure-pipelines/nuget/templates/test_android.yml
@@ -10,6 +10,7 @@ stages:
   - job:  NuGet_Test_Android
     workspace:
       clean: all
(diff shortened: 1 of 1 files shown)


vraspar (approved):
I have also noticed test timing out a lot. It seems reasonable to increase timing unless there are any downside @snnn

carzh:
Stage is hanging indefinitely. I don't think a timeout would fix the issue since it takes 2 minutes to run locally and also previously ran in 2 minutes when it was succeeding. Extending the timeout would just make the pipeline run for longer and then fail.

prathikr:
Actually @carzh is investigating root cause for why the stage times out

PR #25410, opened 2025-07-15
Fix the is_leaf check in TreeEnsemble

### LINKED ISSUE #24679: TreeEnsemble can incorrectly decide a root branch is a leaf.
### Describe the issue

In the albeit unlikely (but not impossible) situation when the root of a tree has one leaf and one branch child, which happen to both have the same index in `leaf_*` and `node_*` respectively, the check to make sure that this isn't just a tree with just one node in line 312 of [tree_ensemble_attribute.h](https://github.com/microsoft/onnxruntime/blob/1f3ae1fb13e3fff209d2171d21ae7d7e8ac3e727/onnxruntime/core/providers/cpu/ml/tree_ensemble_attribute.h#L312) will incorrectly decide that the root is a leaf rather than a branch.

I think it would be enough to change 
```
nodes_falsenodeids[tree_root_size_t] == nodes_truenodeids[tree_root_size_t],
```
 to something like
```
(nodes_falsenodeids[tree_root_size_t] == nodes_truenodeids[tree_root_size_t]) &&
      nodes_trueleafs[tree_root_size_t] == true && nodes_falseleafs[tree_root_size_t] == true,
```
or even
```
(nodes_falsenodeids[tree_root_size_t] == nodes_truenodeids[tree_root_size_t]) &&
      (nodes_trueleafs[tree_root_size_t] == nodes_falseleafs[tree_root_size_t]),
```
to ensure that in addition to the indices being the same, they are in fact indices to the same vector.

I have not checked providers other than `cpu`, similar issues may be present in all of them.

### To reproduce

Attached in human readable form below is a very simple forest I made by hand to try and illustrate this issue. Considering e.g. the input [7,7,4] and going through manually, it's clear the result should be 16 (25 from the first tree and -9 from the second), but at the moment onnxruntime (both in C++ and python) is returning 50.

It does this because the code looks at the root of the second tree, (index 2 in `node_*`), "sees" that both its true and false children have index 3, and assumes this means this is functionally a leaf. However, one of the children is index 3 in `node_*`, and the other is index 3 in `leaf_*`, so this is not a leaf. This would be fixed by looking at `nodes_falseleafs` and `nodes_trueleafs` as well.

```
<
   ir_version: 10,
   opset_import: ["" : 22]
>
BDT_Forest (float[?,3] X) => (float[?,1] Y) {
   Y = ai.onnx.ml.TreeEnsemble <aggregate_function: int = 1, leaf_targetids: ints = [0, 0, 0, 0, 0, 0, 0], leaf_weights: tensor = float[7] leaf_weights {100,0,25,0.5,-0.5,-5,-9}, n_targets: int = 1, nodes_falseleafs: ints = [1, 1, 0, 0, 1], nodes_falsenodeids: ints = [2, 1, 3, 4, 6], nodes_featureids: ints = [0, 1, 0, 1, 2], nodes_missing_value_tracks_true: ints = [0, 0, 0, 0, 0], nodes_modes: tensor = uint8[5] nodes_modes {0,0,0,0,0}, nodes_splits: tensor = float[5] nodes_splits {2,2,3,2,1}, nodes_trueleafs: ints = [0, 1, 1, 1, 1], nodes_truenodeids: ints = [1, 0, 3, 4, 5], tree_roots: ints = [0, 2]> (X)
}
```

### Urgency

This will be rare, but I've now found at least one "real" tree where this caused an issue, in addition to the test example above.

### Platform

Linux

### OS Version

Ubuntu 24

### ONNX Runtime Installation

Built from Source

1f4156c6146d04999e5e419df4ae6628e928aaad

### ONNX Runtime API

C++

### Architecture

X64

### Execution Provider

Default CPU

_No response_

meakbiyik:
### Description
Fixes #24679.

### Motivation and Context
The original check for a leaf node was insufficient because a branch child and a leaf child could have the same index. The bug described in issue #24679 is not a rare occasion; in fact, it is something likely to be faced in estimators with small and balanced trees. I encountered it myself in a unit test.

The corrected check ensures that for a node to be considered a leaf, both of its children must be leaves and share the same index.

### DIFF
--- a/onnxruntime/core/providers/cpu/ml/tree_ensemble_attribute.h
+++ b/onnxruntime/core/providers/cpu/ml/tree_ensemble_attribute.h
@@ -308,9 +308,10 @@ struct TreeEnsembleAttributesV5 {
     int64_t curr_treeid = 0;
     for (const int64_t& tree_root : tree_roots) {
       size_t tree_root_size_t = onnxruntime::narrow<size_t>(tree_root);
+      bool is_leaf = (nodes_falsenodeids[tree_root_size_t] == nodes_truenodeids[tree_root_size_t] &&
+                      nodes_falseleafs[tree_root_size_t] && nodes_trueleafs[tree_root_size_t]);
       transformInputOneTree(tree_root_size_t, curr_treeid, 0, 0U,
-                            nodes_falsenodeids[tree_root_size_t] == nodes_truenodeids[tree_root_size_t],
-                            membership_values_by_id, output);
+                            is_leaf, membership_values_by_id, output);
       curr_treeid++;
     }
   }


xadupre:
Thanks for doing it. Is it possible to add a unit test? It should be quick by copy/pasting an existing unit test and using the tree from the issue.

meakbiyik:
@microsoft-github-policy-service agree company="TripleLift"

meakbiyik:
@xadupre added a unit test based on the original issue's example, it runs on my side. I think it's waiting for your approval here.

When do you think this one and your PR #25423 might be released? I believe onnxruntime does not have regular release windows, right?

meakbiyik:
Ready for review ⬆️

xadupre (approved):

meakbiyik:
@xadupre I think this PR can be merged, not sure why the webGPU tests are failing.

meakbiyik:
They now passed, though the why some pipelines are pending is beyond me @xadupre 😅

meakbiyik:
Ready to merge!

meakbiyik:
Hello @snnn, can we include this fix to the 1.23.0 release as well? https://github.com/microsoft/onnxruntime/pull/25423 also got in, it would make sense to have them both together. cc @xadupre

PR #25411, opened 2025-07-15
[EP ABI] Add documentation for OrtValue and ort_graph_to_proto util

adrianlizarraga:
### Description
- Adds documentation to state that the data pointer for an `OrtValue` owned by an `OrtGraph` is stable during the lifetime of the `OrtSession` that owns the `OrtGraph`.
- Adds documentation to the ort_graph_to_proto.h utils to show how to create a `onnx::GraphProto` with external initializers that actually point to in-memory data (same approach used internally within ORT).

### Motivation and Context
Clarification of usage of new graph apis.

### DIFF
--- a/include/onnxruntime/core/session/onnxruntime_ep_c_api.h
+++ b/include/onnxruntime/core/session/onnxruntime_ep_c_api.h
@@ -413,8 +413,6 @@ struct OrtEp {
    * \note Do NOT cache the provided OrtGraph instances in any of the OrtNodeComputeInfo functions because the
    *       graphs are only valid for the duration of the call to Compile. Any graph/node/input/output
    *       names that are needed by the OrtNodeComputeInfo functions must be copied and stored by the OrtEp.
-   *       The OrtValue data for initializers (obtained via GetTensorData), however, is stable for the lifetime of
-   *       the owning OrtSession instance.
    *
    * \since Version 1.23.
    */


include/onnxruntime/core/providers/utils/ort_graph_to_proto.h
```diff
@@ -75,6 +75,44 @@
...
+                                             OrtEpGraphSupportInfo* graph_support_info) {
+     auto handle_initializer_data = [](const OrtValueInfo* value_info,
+                                       const void* data, size_t bytes,
+                                       bool& is_external, std::string& location,
+                                       int64_t& offset) -> Ort::Status {
+       (void)value_info;
+       (void)bytes;
+
+       offset = reinterpret_cast<int64_t>(data);
+       location = "_MEM_ADDR_";  // Some special location tag that indicates the offset is a pointer.
```
yuslepukhin, line 104: >// Some special location tag that indicates the offset is a pointer. [](http://example.com/codeflow?start=33&length=68)

There is a constant in utils

onnxruntime/test/ep_graph/test_ep_graph.cc
```diff
@@ -178,6 +178,77 @@ TEST(EpGraphTest, SerializeToProto_Mnist) {
...
+
+  auto handle_initializer_data = [](const OrtValueInfo* value_info,
+                                    const void* data, size_t bytes,
+                                    bool& is_external, std::string& location,
+                                    int64_t& offset) -> Ort::Status {
+    (void)value_info;
+    (void)bytes;
+
+    offset = reinterpret_cast<int64_t>(data);
+    location = "_MEM_ADDR_";
```
yuslepukhin, line 198: >= "_MEM_ADDR_"; [](http://example.com/codeflow?start=13&length=15)

Try using a constant that is actually a little different from this string

include/onnxruntime/core/providers/utils/ort_graph_to_proto.h
```diff
@@ -75,6 +75,44 @@
...
+       is_external = true;  // True if is external initializer
+       return Ort::Status{nullptr};
+     }
+
+     ONNX_NAMESPACE::GraphProto graph_proto;
```
yuslepukhin, line 109: >ONNX_NAMESPACE::GraphProto graph_proto; [](http://example.com/codeflow?start=5&length=39)

In utls:: I have introduced a validation function that ensures that TensorProto offset is the actual data ptr in the given OrtValue

include/onnxruntime/core/providers/utils/ort_graph_to_proto.h
adrianlizarraga: This is supposed to be a standalone header-only utility file that EP implementers can copy into their own source code. It is also a reference implementation of how an EP could convert an OrtGraph to onnx proto. We don't want to include any unnecessary code from ORT to make it easy to copy and modify.

onnxruntime/test/ep_graph/test_ep_graph.cc
adrianlizarraga: Refer to https://github.com/microsoft/onnxruntime/pull/25411#discussion_r2211714611

include/onnxruntime/core/providers/utils/ort_graph_to_proto.h
adrianlizarraga: Refer to https://github.com/microsoft/onnxruntime/pull/25411#discussion_r2211714611

onnxruntime/test/ep_graph/test_ep_graph.cc
adrianlizarraga: I'm not sure I understand why this is necessary. This unit test is an example of how an EP could use these utilities. An EP implementer would select their own constant anyway that is different than what ORT uses internally.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -5459,9 +5459,12 @@ struct OrtApi {
    * Supports initializers defined in an outer scope (i.e., a parent graph).
    *
    * \param[in] value_info The OrtValueInfo instance.
-   * \param[out] initializer_value Output parameter set to the initializer value or NULL.
+   * \param[out] initializer_value Output parameter set to the initializer value or NULL. The OrtValue data pointer
+   *                               (obtained via GetTensorData) is stable during the lifetime of the OrtSession
+   *                               that owns the OrtGraph.
```
javier-intel, line 5464: Perhaps add that in case `ep.share_ep_contexts` is used the lifetime is extended until all the sessions that share the context are destroyed.

adrianlizarraga: I'm not sure I understand as each OrtValue is owned by its session. Perhaps we can discuss separately.

snnn:

PR #25424, opened 2025-07-16
[EP ABI] Add Graph_GetModelPath API function

adrianlizarraga:
### Description
Adds `Graph_GetModelPath` API function.

### Motivation and Context
An EP that loads an EPContext model (i.e., compiled model) needs to know the base path of the model to determine the path the EPContext node binary (if stored in a file).


onnxruntime/core/session/onnxruntime_c_api.cc
```diff
@@ -2579,6 +2579,10 @@ ORT_API_STATUS_IMPL(OrtApis::Graph_GetName, _In_ const OrtGraph* graph, _Outptr_
   API_IMPL_END
 }
 
+ORT_API(const ORTCHAR_T*, OrtApis::Graph_GetModelPath, _In_ const OrtGraph* graph) {
```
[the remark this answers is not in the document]
adrianlizarraga: The `onnxruntime_c_api.h` header documents that parameters marked with `_In_` should not be NULL. API functions typically do not check this for `_In_` parameters (otherwise we could almost never have API functions that do not return a status).

https://github.com/microsoft/onnxruntime/blob/d22a7819fb8da16c89262743d06a365f664fbf42/include/onnxruntime/core/session/onnxruntime_c_api.h#L127-L128

onnxruntime/core/graph/model_editor_api_types.h
```diff
@@ -173,6 +173,8 @@ struct ModelEditorGraph : public OrtGraph {
 
   const std::string& GetName() const override { return name; }
 
+  const ORTCHAR_T* GetModelPath() const override { return ORT_TSTR(""); }
```
yuslepukhin: >return ORT_TSTR(""); [](http://example.com/codeflow?start=51&length=20)

Can the model path be part of the base class so this would be a meaningful impl?

onnxruntime/core/session/onnxruntime_c_api.cc
```diff
@@ -2579,6 +2579,10 @@ ORT_API_STATUS_IMPL(OrtApis::Graph_GetName, _In_ const OrtGraph* graph, _Outptr_
...
+  return graph->GetModelPath();
```
yuslepukhin: >return graph->GetModelPath() [](http://example.com/codeflow?start=2&length=28)

I would still prefer exception guards here. And with that it would raise a question of how to report potential error translated from the exception.

Even standard types such as std::path can throw.

yuslepukhin (changes requested):
:clock1:

onnxruntime/core/graph/model_editor_api_types.h
adrianlizarraga: Would a model from model editor every have a valid model path?

onnxruntime/core/session/onnxruntime_c_api.cc
adrianlizarraga: Good point. I'll make the api just return a status.

adrianlizarraga: Updated

onnxruntime/core/graph/model_editor_api_types.h
adrianlizarraga: Updated

jywu-msft (dismissed):

jywu-mysoft:
@yuslepukhin can you review again?

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -5560,6 +5560,20 @@ struct OrtApi {
    */
   ORT_API2_STATUS(Graph_GetName, _In_ const OrtGraph* graph, _Outptr_ const char** graph_name);
 
+  /** \brief Get the filepath to the ONNX model from which an OrtGraph is constructed.
+   *
+   * \note The model's filepath is empty if the filepath is unknown, such as when the model is loaded from bytes
+   * via CreateSessionFromArray.
+   *
+   * \param[in] graph The OrtGraph instance.
+   * \param[out] model_path Output parameter set to the model's filepath, which is an empty path string if unknown.
```
yuslepukhin: Should we specify the string is zero terminated?

```suggestion
   * \param[out] model_path Output parameter set to the model's null-terminated filepath.
   *                        Set to an empty path string if unknown.
```

adrianlizarraga: @yuslepukhin updated. thanks

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -5560,6 +5560,21 @@ struct OrtApi {
    */
   ORT_API2_STATUS(Graph_GetName, _In_ const OrtGraph* graph, _Outptr_ const char** graph_name);
 
+  /** \brief Get the filepath to the ONNX model from which an OrtGraph is constructed.
```
edgchen1: > Get the filepath to the ONNX model

should we limit this to ONNX models or do we also want to support ORT format models? and on that note, I'm not sure if Model::model_path_ is getting set correctly for ORT format models.

```suggestion
  /** \brief Get the filepath to the model from which an OrtGraph is constructed.
```

adrianlizarraga: Updated doc to remove explicit mention of "ONNX". Will look into the question about model_path not being set for ORT format models separately.

edgchen1 (approved):

adrianlizarraga:
closing and reopening to make CI rerun

adrianlizarraga:
hi @yuslepukhin, can you please review again?

yuslepukhin (approved):

snnn:

PR #25425, opened 2025-07-16
Enable free dimension override for graph optimization level 0

chilo-ms:
Enable free dimension override even when the graph optimization level is 0.
If the optimization level is above 0, the override will be applied during level 1 optimization.

### DIFF
--- a/onnxruntime/core/optimizer/graph_transformer_utils.cc
+++ b/onnxruntime/core/optimizer/graph_transformer_utils.cc
@@ -221,8 +221,10 @@ InlinedVector<std::unique_ptr<GraphTransformer>> GenerateTransformers(
 
   switch (level) {
     case TransformerLevel::Default: {
-      transformers.emplace_back(std::make_unique<FreeDimensionOverrideTransformer>(
-          session_options.free_dimension_overrides));
+      if (!session_options.free_dimension_overrides.empty()) {
+        transformers.emplace_back(std::make_unique<FreeDimensionOverrideTransformer>(
+            session_options.free_dimension_overrides));
+      }
     } break;
     case TransformerLevel::Level1: {
       // RewriteRule optimizations are the simplest (they generally remove unnecessary nodes and are cheap to run)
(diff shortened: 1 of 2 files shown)


skottmckay:
Can we add a unit test to validate? There should already be free dimension override tests that we can leverage.

chilo-ms:
> Can we add a unit test to validate? There should already be free dimension override tests that we can leverage.

unit tests added

onnxruntime/core/optimizer/graph_transformer_utils.cc
```diff
@@ -220,6 +220,10 @@ InlinedVector<std::unique_ptr<GraphTransformer>> GenerateTransformers(
   AllocatorPtr cpu_allocator = CPUAllocator::DefaultInstance();
 
   switch (level) {
+    case TransformerLevel::Default: {
+      transformers.emplace_back(std::make_unique<FreeDimensionOverrideTransformer>(
+          session_options.free_dimension_overrides));
+    } break;
     case TransformerLevel::Level1: {
```
skottmckay, line 229: Can we skip adding the transformer if session_options.free_dimension_overrides is empty?

chilo-ms: yes, added the check to only add the transformer if ession_options.free_dimension_overrides is not empty

onnxruntime/core/session/inference_session.cc
```diff
@@ -3645,34 +3646,50 @@ common::Status InferenceSession::AddPredefinedTransformers(
...
-      auto transformers_to_register = [&]() {
-        const bool use_full_build_optimizations =
-            level == TransformerLevel::Level1 ||
-            minimal_build_optimization_handling == MinimalBuildOptimizationHandling::ApplyFullBuildOptimizations;
-
-        if (use_full_build_optimizations) {
+    std::function<onnxruntime::InlinedVector<std::unique_ptr<GraphTransformer>>()> transformers_to_register;
+
+    // Enable free dimension override even when the graph optimization level is 0.
+    // If the optimization level is above 0, the override will be applied during level 1 optimization.
```
skottmckay, line 3654: Do we need the complexity of having the free dimension override potentially being registered in level 0 or level 1 and needing to figure out which is used, or would it be fine to only have it in level 0. I think the check on there being overrides before adding the transformer is good regardless of graph_optimization_level.

skottmckay (approved):
:shipit:

snnn:

PR #25428, opened 2025-07-16
upgrade emsdk to 4.0.9

fs-eire:
### Description

upgrade emsdk to 4.0.9

### Motivation and Context

A bug in emscan-deps blocks the build when using Ninja. The bug is fixed in upstream at https://github.com/emscripten-core/emscripten/pull/24506. The fix is included in emsdk v4.0.9

Why not upgrade to latest (4.0.11 at the moment):

- later version introduced a breaking change in https://github.com/emscripten-core/emscripten/pull/24384. ORT-web need to make some changes to work with it, and it takes more time. For now, upgrading to 4.0.9 should be enough to unblock the build.

Why the CI is not affected:

- The build issue seems to occur when building using Ninja. CI is using Linux with default CMake generator. For WebAssembly build, the default CMake generator is:
  - Ninja (Windows)
  - Makefile (UNIX)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,4 +7,4 @@
 [submodule "cmake/external/emsdk"]
 	path = cmake/external/emsdk
 	url = https://github.com/emscripten-core/emsdk.git
-	branch = 4.0.8
+	branch = 4.0.9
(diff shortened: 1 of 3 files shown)


guschmue (approved):

fs-eire:
after double check , the fix is in 4.0.10. drop this PR.

PR #25432, opened 2025-07-17
[CUDA] Support head_sink in flash attention for GQA

tianleiwu:
### Description

Update Flash Attention to support softmax sink in GQA.

Changes:
- [x] Update flash attention to support head_sink
- [x] Add test_gqa.py to test cuda, and remove test_gqa_cuda.py.

Note that the sink is treated as scaled, while the elements in QK GEMMs is not scaled. The sink value does not need scaling or softcap, and it joins softmax with those scaled or soft-capped values. There are two ways to add sink to softmax:
* One way is to [patch normalize_softmax_lse](https://github.com/microsoft/onnxruntime/blob/1cf1aa786f6e7f7e6abd6fba8b8aea2e7a43092c/onnxruntime/contrib_ops/cuda/bert/flash_attention/softmax.h#L143-L178) to use sink to update max and sum. Pros is major change in one function; Cons is the logic is a little complex since row_max is unscaled, while row_sum is scaled.
* Another way is to change softmax_rescale_o to handle the sink directly in the first block of online softmax by using an unscaled sink value. It is a robust way to keep core algorithm consistent. Cons is need change in multiple places, and it is little hard to work with softcap.

This PR use the the first approach for easy integration.

Note: Memory efficient attention change will be in separated PR.

### Motivation and Context
https://github.com/microsoft/onnxruntime/pull/25269


onnxruntime/test/python/transformers/test_gqa.py
```diff
@@ -0,0 +1,1167 @@
...
+                                        yield name, config
+
+
+def gqa_cuda_past_test_cases(allow_head_sink: bool = True):
+    batches = [5] if pipeline_mode else [1, 3, 5]
+    # s: new sequence length, s2: past sequence length
+    seqs = [(1, 1024)] if pipeline_mode else [(1, 128), (1, 1024), (1, 2048), (1, 5000)]
+    num_h = [(32, 8)] if pipeline_mode else [(6, 3), (9, 9), (32, 8)]
+    h_sizes = [256] if pipeline_mode else [64, 128, 256]
+    smmoth_softmax__head_sink = get_softmax_options(allow_head_sink)
```
kunal-vaishnavi, line 1043:
```suggestion
    smooth_softmax_head_sink = get_softmax_options(allow_head_sink)
```

onnxruntime/contrib_ops/cuda/bert/flash_attention/softmax.h
```diff
@@ -85,38 +84,6 @@ __forceinline__ __device__ void scale_apply_exp2(Tensor<Engine0, Layout0>& tenso
   }
 }
 
-// Apply the exp to all the elements.
-template <bool zero_init = true, typename Engine0, typename Layout0, typename Engine1, typename Layout1>
-__forceinline__ __device__ void max_scale_exp2_sum(Tensor<Engine0, Layout0>& tensor, Tensor<Engine1, Layout1>& max, Tensor<Engine1, Layout1>& sum, const float scale) {
```
kunal-vaishnavi, line 90: Why are we removing this kernel?

tianleiwu: not used

onnxruntime/test/python/transformers/test_gqa.py
tianleiwu: Good catch. I could fix it in next PR for efficient attention.

PR #25443, opened 2025-07-18
[webgpu] support float16 type for Einsum operator

xhcao:


xhcao:
@feich-ms Please take a look, thanks. Can I simply support float16 type for Einsum operator as this PR?
@jchen10

onnxruntime/core/providers/webgpu/math/einsum.cc
```diff
@@ -34,13 +34,13 @@ bool IsInteger(const std::string& s) {
...
 }  // namespace
 
-#define WEBGPU_EINSUM_TYPED_KERNEL_DECL(version)                                               \
-  ONNX_OPERATOR_TYPED_KERNEL_EX(                                                               \
-      Einsum, kOnnxDomain, version, float, kWebGpuExecutionProvider,                           \
-      (*KernelDefBuilder::Create()).TypeConstraint("T", DataTypeImpl::GetTensorType<float>()), \
+#define WEBGPU_EINSUM_KERNEL_DECL(version)                                               \
+  ONNX_OPERATOR_KERNEL_EX(                                                               \
+      Einsum, kOnnxDomain, version, kWebGpuExecutionProvider,                            \
+      (*KernelDefBuilder::Create()).TypeConstraint("T", WebGpuSupportedFloatTypes()),    \
```
qjia7: Can you relax it to WebGpuSupportedNumberTypes if all related tests can pass?

onnxruntime/test/providers/cpu/math/einsum_test.cc
```diff
@@ -595,9 +595,11 @@ TEST(Einsum, ImplicitEinsumAsTensorContraction) {
 // Theme: Half support
 
 TEST(Einsum, ExplicitEinsumAsIdentity_1D_input_Half) {
+#if !defined(USE_WEBGPU)
```
qjia7, line 598: Is it better if we move `#if !defined(USE_WEBGPU)` check into `HasCudaEnvironment`? It's noisy to scatter so many places.

onnxruntime/core/providers/webgpu/math/einsum.cc
xhcao: Jiajia, there is not uint32 case and only one int32 case, if relaxing the types, I am not sure whether there will be some uncovered errors.

onnxruntime/test/providers/cpu/math/einsum_test.cc
xhcao: I also think it is not good that there are many `#if !defined(USE_WEBGPU)`. But I think it is also unreasonable move it into a CUDA special file. Is there Any other good ideas from anyone?

xhcao (commented):
Jiajia, thanks for reviewing.

xhcao:
@guschmue @fs-eire Please help trigger the bots, thanks.

xhcao:
It seems that the failures of bots are not related with my PR.

xhcao:
1: [  FAILED  ] 1 test, listed below:
1: [  FAILED  ] ModelTests/ModelTest.Run/cpu__models_opset10_faster_rcnn_faster_rcnn_R_50_FPN_1x, where GetParam() = L"cpu_..\\models\\opset10\\faster_rcnn\\faster_rcnn_R_50_FPN_1x.onnx"

It seems that the failed case is not related with the PR

PR #25444, opened 2025-07-18
[QNN EP] Enable Conv Op with "auto_pad" param set as VALID

aparmp-quic:
Enable Conv Op and ConvTranspose Op with "auto_pad" param set as VALID

### Description
QNN_EP reject the Conv Op and ConvTranspose on HTP if "auto_pad" is "VALID". This configuration is supported on HTP.

### Motivation and Context
To enable Conv and ConvTranspose op with auto_pad as "VALID" running on NPU and prevent them from falling back to CPU.


aparmp-quic:

jywu-msft (approved):

PR #25446, opened 2025-07-18
[NV RTX EP] Set Compute Capability only on Turing architecture

keshavv27:
### Description

Set compute capability only on Turing arch

### Motivation and Context

Setting the native compute capability was causing a regression in performance. 

@gaugarg-nv @ishwar-raut1 @ankan-ban


ankan-ban:
Did we ask TRT-RTX guys why setting "current" profile causes perf regression ?

keshavv27:
> Did we ask TRT-RTX guys why setting "current" profile causes perf regression ?

Did not get a response yet on the root cause for this. For current state of TRT RTX this was the change suggested.

ishwar-raut1:
> > Did we ask TRT-RTX guys why setting "current" profile causes perf regression ?
> 
> Did not get a response yet on the root cause for this. For current state of TRT RTX this was the change suggested.

Yes, change should be in the TRT RTX. But this WAR is fine for now.

ankan-ban:
ok. the WAR makes sense then. Thanks.

PR #25455, opened 2025-07-19
Compile API: output model and initializer stream write functions

adrianlizarraga:
### Description
- Adds `ModelCompilationOptions_SetOutputModelWriteFunc` to the compile API to allow writing the output model ONNX bytes to a user-provided write function (i.e., for streaming).
- Adds `ModelCompilationOptions_SetOutputModelHandleInitializerFunc` to the compile API to allow the user to write individual initializers to some destination. Also allows specifying if an initializer should be embedded within the ONNX model or written to a custom file.
- Adds C++, Python, and C# bindings for the new APIs.

A follow-up PR adds a write function for EPContext node binary data: https://github.com/microsoft/onnxruntime/pull/25471

### Example
`ModelCompilationOptions_SetOutputModelWriteFunc`: https://github.com/microsoft/onnxruntime/blob/c62ed23c328cbbfefd3083c1f7a6ced604772c19/onnxruntime/test/providers/qnn/qnn_ep_context_test.cc#L2075-L2131

`ModelCompilationOptions_SetOutputModelHandleInitializerFunc`:
https://github.com/microsoft/onnxruntime/blob/c62ed23c328cbbfefd3083c1f7a6ced604772c19/onnxruntime/test/providers/qnn/qnn_ep_context_test.cc#L2160-L2292

### Motivation and Context
Add output streaming capabilities when saving compiled models.


include/onnxruntime/core/session/onnxruntime_cxx_api.h
```diff
@@ -3060,5 +3064,41 @@ struct Model : detail::ModelImpl<OrtModel> {
 
   ConstModel GetConst() const { return ConstModel{this->p_}; }
 };
+
+namespace detail {
+template <typename T>
+struct ExternalInitializerInfoImpl : Base<T> {
```
yuslepukhin: The naming should start with `Const` #Resolved

include/onnxruntime/core/session/onnxruntime_cxx_api.h
```diff
@@ -3060,5 +3064,41 @@ struct Model : detail::ModelImpl<OrtModel> {
...
+  using B::B;
+
+  std::basic_string<ORTCHAR_T> GetFilePath() const;  ///< Wraps ExternalInitializerInfo_GetFilePath
+  int64_t GetFileOffset() const;                     ///< Wraps ExternalInitializerInfo_GetFileOffset
+  size_t GetByteSize() const;                        ///< Wraps ExternalInitializerInfo_GetByteSize
+};
+}  // namespace detail
+
+// Wrapper around an unowned ::OrtExternalInitializerInfo.
+using UnownedExternalInitializerInfo = detail::ExternalInitializerInfoImpl<detail::Unowned<OrtExternalInitializerInfo>>;
```
yuslepukhin: You do not need `Unowned` since there is not a non `Const` version of it. #Closed

yuslepukhin: Said that, it is in my PR as well.

onnxruntime/core/framework/ep_context_options.h
```diff
@@ -0,0 +1,128 @@
...
+    // Generate an output model even if it doesn't have compiled nodes.
+    // The explicit Compile API defaults to this value.
+    kGenerateModel,
+
+    // Return an error if the model does not have compiled nodes.
+    // The explicit Compile API can be configured to this value.
+    kReturnError,
+  };
+
+  ModelGenOptions() = default;
```
yuslepukhin: Did you know this increases the size of the binary much more, than if you declared `ModelGenOptions::ModelGenOptions() = default` in a .cc file. #Closed

onnxruntime/core/framework/graph_partitioner.cc
```diff
@@ -794,9 +797,51 @@ static Status GetValidatedEpContextPath(const std::filesystem::path& ep_context_
   return Status::OK();
 }
 
+// Serialize an EPContext model into a onnx::ModelProto.
+static Status EpContextModelToProto(const onnxruntime::Model& ep_context_model,
```
yuslepukhin: Should this belong to either tensorprotoutils or EpContenxt. Paritioninaer sees out of place. #Closed

include/onnxruntime/core/graph/graph.h
```diff
@@ -1247,6 +1247,18 @@ class Graph {  // NOLINT(clang-analyzer-optin.performance.Padding): preserve exi
                                                                   const std::filesystem::path& model_file_path,
                                                                   const ModelSavingOptions& model_saving_options) const;
 
+  /// <summary>
```
yuslepukhin, line 1250: Is it possible to incorporate the new function in the existing code? #Closed

yuslepukhin (changes requested):

csharp/src/Microsoft.ML.OnnxRuntime/OrtExternalInitializerInfo.shared.cs
```diff
@@ -0,0 +1,139 @@
...
+        /// <param name="constHandle">Native OrtExternalInitializerInfo handle.</param>
+        /// <param name="ownsHandle">True if the OrtExternalInitializerInfo instance owns the native handle.
+        /// Defaults to false.</param>
+        /// <exception cref="InvalidOperationException"></exception>
+        internal OrtExternalInitializerInfo(IntPtr constHandle, bool ownsHandle = false)
+            : base(IntPtr.Zero, ownsHandle)
+        {
+            if (constHandle == IntPtr.Zero)
+            {
+                throw new InvalidOperationException($"{nameof(OrtExternalInitializerInfo)}: Invalid instance.");
```
yuslepukhin: It looks like an invalid argument, but may be an invalid instance received from another API call, in this case the error codeo of that call should have been checked. #Closed

onnxruntime/core/graph/graph.cc
```diff
@@ -4643,6 +4644,137 @@ ONNX_NAMESPACE::GraphProto Graph::ToGraphProtoWithExternalInitializers(
...
+Status Graph::ToGraphProtoWithInitializerHandlerImpl(OrtHandleInitializerDataFunc handle_initializer_func,
+                                                     void* state,
+                                                     /*out*/ ONNX_NAMESPACE::GraphProto& output_graph_proto) const {
+  // This loop processes subgraphs bottom up.
+  for (const auto& node : Nodes()) {
+    if (node.ContainsSubgraph()) {
+      // Let's find this node in the output_graph_proto
+      // The node name is optional, so we may need to check by the output value name
+      // given that they can only assigned once.
+      auto hit = std::find_if(output_graph_proto.mutable_node()->begin(),
```
yuslepukhin: This looks like a code that is used at other places.
1) Can we modify that code so it can take a function?
2) Can be abstract the subgraph iteration/recursion so it is re-used? We already have a function ForEachSubgraph() but that is driven by a map, this is driven by protobuf side of subgraphs, but the principle is the same. #Closed

onnxruntime/core/graph/model.cc
```diff
@@ -415,6 +415,25 @@ ModelProto Model::ToGraphProtoWithExternalInitializers(const std::filesystem::pa
...
+
+  // Sync current model_metadata_ back to protobuf metadata_props
+  model_proto.clear_metadata_props();
+  for (const auto& metadata : model_metadata_) {
+    const gsl::not_null<StringStringEntryProto*> prop{model_proto.add_metadata_props()};
+    prop->set_key(metadata.first);
+    prop->set_value(metadata.second);
+  }
+
+  const auto& graph = *graph_;
```
yuslepukhin, line 431: Do we need to check if the graph was synchronized? #Resolved

onnxruntime/core/graph/graph.cc
```diff
@@ -4643,6 +4644,137 @@ ONNX_NAMESPACE::GraphProto Graph::ToGraphProtoWithExternalInitializers(
...
+                          node.Name(), " while attempting to recurse into it.");
+        auto& result_subgraph = *sub_hit->mutable_g();
+        ORT_RETURN_IF_ERROR(subgraph->ToGraphProtoWithInitializerHandlerImpl(handle_initializer_func,
+                                                                             state, result_subgraph));
+      }
+    }
+  }
+
+  // Create a sorted std::vector of initializers so that we always process them in a deterministic order.
+  std::vector<const ONNX_NAMESPACE::TensorProto*> initializers;
```
yuslepukhin: InlinedVector #Resolved

onnxruntime/core/graph/graph.cc
```diff
@@ -4643,6 +4644,137 @@ ONNX_NAMESPACE::GraphProto Graph::ToGraphProtoWithExternalInitializers(
...
+
+  // Call user's handler function for each initializer. We store the initializer externally
+  // or within the model depending on the result returned by the handler function.
+  for (gsl::not_null<const ONNX_NAMESPACE::TensorProto*> initializer : initializers) {
+#if !defined(DISABLE_SPARSE_TENSORS)
+    if (IsSparseInitializer(initializer->name())) {
+      // Sparse tensors are added to the ONNX file directly.
+      auto& sparse_initializer = *output_graph_proto.add_sparse_initializer();
+      auto status = utils::DenseTensorToSparseTensorProto(*initializer, ModelPath(), sparse_initializer);
+      ORT_RETURN_IF_NOT(status.IsOK(), "Failed to convert dense initializer to sparse");
```
yuslepukhin: It would help to diagnose if the original message from status is propagated. #Resolved

onnxruntime/python/onnxruntime_pybind_model_compiler.h
```diff
@@ -14,11 +13,23 @@ namespace onnxruntime {
 class Environment;
 
 namespace python {
+// Type of the function provided by Python code that is called by ORT to write out the compiled model.
+// Returns the number of bytes written to the underlying stream.
+using PyOutStreamWriteFunc = std::function<void(const pybind11::bytes& buffer)>;
+
+using PyHandleInitializerFunc = std::function<std::shared_ptr<const OrtExternalInitializerInfo>(const std::string& initializer_name,
```
yuslepukhin: Truncate to 120 #Resolved

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -533,6 +533,56 @@ typedef OrtStatus*(ORT_API_CALL* EpSelectionDelegate)(_In_ const OrtEpDevice** e
...
+ *                               If the function implementation sets `new_external_info` to NULL,
+ *                               ORT stores the initializers within the model.
+ *
+ * \note ORT takes ownership of the `new_external_info` output parameter.
+ *
+ * \return OrtStatus* Write status. Return nullptr on success.
+ *                    Use CreateStatus to provide error info. Use ORT_FAIL as the error code.
+ *                    ORT will release the OrtStatus* if not null.
+ */
+typedef OrtStatus*(ORT_API_CALL* OrtHandleInitializerDataFunc)(_In_ void* state,
```
yuslepukhin: 120 chars #Resolved

csharp/src/Microsoft.ML.OnnxRuntime/NativeMethods.shared.cs
```diff
@@ -422,6 +422,7 @@ public struct OrtApi
         public IntPtr Node_GetGraph;
         public IntPtr Node_GetEpName;
         public IntPtr ReleaseExternalInitializerInfo;
+        public IntPtr CreateExternalInitializerInfo;
```
yuslepukhin: The order must match the order in the C++ OrtApi declaration and added at the end. #Resolved

include/onnxruntime/core/session/onnxruntime_cxx_api.h
adrianlizarraga: This PR adds a non-const version of ExternalInitializerInfo, so I believe we need this impl class to not be const.

include/onnxruntime/core/session/onnxruntime_cxx_api.h
adrianlizarraga: This PR adds a non-const version of ExternalInitializerInfo,

onnxruntime/core/framework/ep_context_options.h
adrianlizarraga: Moved to .cc file. Thanks.

onnxruntime/core/framework/graph_partitioner.cc
adrianlizarraga: Yeah, I agree. Unfortunately, all of the EPContext generation code is currently in the partitioner at the moment. I did the following:
- moved this specific function to a separate file (ep_context_utils.h/cc)
- Added TODO comments on other existing EPContext-related functions to note that they should also be moved out of the partitioner. I didn't do that move now because that would obfuscate the actual changes to those functions.

include/onnxruntime/core/graph/graph.h
adrianlizarraga: I strongly considered moving this into the existing `ToGraphProtoWithExternalInitializers` but decided against it due to the mismatch in what both functions are trying to do and the complexity in shoehorning this new functionality in there.

To be more specific, the existing `ToGraphProtoWithExternalInitializers` function:
- Takes in a path to an external file that ORT creates and writes into. This is not what we want with the new API since we want to allow the user to write out the initializer bytes into whatever destination they choose.
- Has very heavy use of conditionals and branching logic to accommodate various "saving modes", which do not apply to this new functionality. Trying to put this new logic in there would complicate the code, make it hard to understand, and would be error prone IMO.

Lastly, the new `ToGraphProtoWithInitializerHandler` function is a relatively small function that does only exactly what we need it to do. It is decoupled from existing uses of `ToGraphProtoWithExternalInitializers` in production.

csharp/src/Microsoft.ML.OnnxRuntime/OrtValue.shared.cs
```diff
@@ -1464,11 +1473,155 @@ protected virtual void Dispose(bool disposing)
...
             _disposed = true;
         }
 
         #endregion
     }
+
+    /// <summary>
+    /// Interface for all readonly methods implemented by OrtValue.
+    /// </summary>
+    public interface IReadOnlyOrtValue
```
adrianlizarraga: Note:
The C# bindings only implements an `OrtValue` owned by the application. However, we now need to be able to represent an `OrtValue` that is owned by ORT (so that we can pass it to the user's callback function when compiling). 

So, I created a "readonly" interface that contains all OrtValue getters. This makes it possible to pass a `IReadOnlyOrtValue` ortvalue to the user's callback. #Resolved

csharp/src/Microsoft.ML.OnnxRuntime/NativeMethods.shared.cs
```diff
@@ -2382,6 +2408,70 @@ out IntPtr lora_adapter
...
         public static DOrtGetCompileApi OrtGetCompileApi;
+
+        /// <summary>
+        /// Delegate called by ORT to write a buffer (ONNX model bytes) to a custom destination (e.g., file or stream).
+        /// </summary>
+        /// <param name="state">State that was provided in when the delegate was registered.</param>
+        /// <param name="buffer">The buffer to write.</param>
+        /// <param name="bufferNumBytes">The size of the buffer in bytes.</param>
+        /// <returns>OrtStatus*</returns>
+        [UnmanagedFunctionPointer(CallingConvention.Winapi)]
```
yuslepukhin, line 2420: If this and below function used by a client code to supply their own delegates, these two must be declared in some more public space. #Resolved

csharp/src/Microsoft.ML.OnnxRuntime/SessionOptions.shared.cs
```diff
@@ -449,7 +449,7 @@ public void AppendExecutionProvider(string providerName, Dictionary<string, stri
         /// <param name="epOptions">Optional options to configure the execution provider. May be null.</param>
         /// <exception cref="ArgumentException">epDevices was empty.</exception>
         /// <see cref="OrtEnv.GetEpDevices" />
-        public void AppendExecutionProvider(OrtEnv env, IReadOnlyList<OrtEpDevice> epDevices, 
+        public void AppendExecutionProvider(OrtEnv env, IReadOnlyList<OrtEpDevice> epDevices,
```
yuslepukhin: Looks like this file only has whitespace differences. #Closed

csharp/src/Microsoft.ML.OnnxRuntime/OrtValue.shared.cs
yuslepukhin: Well, not really. The values that are returned from inference are owned by ORT and this class represents them, and they also need to be released. So I am not sure which purpose this serves and what happens when the underlying native object is destroyed. Every `OrtValue` should be released.
If you want to create an alias to that, then it should be ref-counted internally via some new API, like Clone.
Then the existing class will own it and will properly release it.

include/onnxruntime/core/session/onnxruntime_cxx_api.h
yuslepukhin: See my offline comments. The impl should be Const since it groups const methods.

csharp/src/Microsoft.ML.OnnxRuntime/SessionOptions.shared.cs
adrianlizarraga: The auto linter in my editor did this (it uses our linter settings in the ORT repo).

csharp/src/Microsoft.ML.OnnxRuntime/NativeMethods.shared.cs
adrianlizarraga: This is the actual order in the C OrtApi struct. It is added in between functions because all of the surrounding functions are not yet released.

csharp/src/Microsoft.ML.OnnxRuntime/OrtValue.shared.cs
adrianlizarraga: ORT passes a `const OrtValue*` to the user's function. Don't want the user's function to modify it or release it (though release is less of an issue since OrtValue wraps a shared_ptr)

csharp/src/Microsoft.ML.OnnxRuntime/OrtExternalInitializerInfo.shared.cs
adrianlizarraga: Would you be ok with changing this to a Debug.Assert so that it helps catch misuse?

csharp/src/Microsoft.ML.OnnxRuntime/NativeMethods.shared.cs
adrianlizarraga: This is not used directly by client code. This is only used by internal code.

The client uses the delegate definition in https://github.com/microsoft/onnxruntime/blob/c62ed23c328cbbfefd3083c1f7a6ced604772c19/csharp/src/Microsoft.ML.OnnxRuntime/CompileModel.shared.cs#L153

and
https://github.com/microsoft/onnxruntime/blob/c62ed23c328cbbfefd3083c1f7a6ced604772c19/csharp/src/Microsoft.ML.OnnxRuntime/CompileModel.shared.cs#L192

csharp/src/Microsoft.ML.OnnxRuntime/SessionOptions.shared.cs
adrianlizarraga: i'll undo it to reduce noise

csharp/src/Microsoft.ML.OnnxRuntime/OrtExternalInitializerInfo.shared.cs
yuslepukhin: Absolutely!

onnxruntime/core/graph/graph.cc
adrianlizarraga: Refactored this common code.

onnxruntime/core/graph/model.cc
adrianlizarraga: I don't think we need to check if the graph was synchronized. As I understand, if the graph is synchronized, we can copy the internal `graph_proto_` member. However, this would unnecessarily copy initializer tensor protos, which we don't to do because the user's function will decide where to store the initializers.

I believe this is the same reason that the existing `ToGraphProtoWithExternalInitializers` also does not check if the graph is synchronized.

onnxruntime/core/graph/graph.cc
adrianlizarraga: updated.

onnxruntime/core/graph/graph.cc
adrianlizarraga: updated.

csharp/src/Microsoft.ML.OnnxRuntime/NativeMethods.shared.cs
adrianlizarraga: I updated to order just to be safe.

csharp/src/Microsoft.ML.OnnxRuntime/CompileModel.shared.cs
```diff
@@ -159,6 +159,271 @@ public void SetGraphOptimizationLevel(GraphOptimizationLevel graphOptimizationLe
                     handle, graphOptimizationLevel));
         }
 
+        /// <summary>
+        /// Delegate to write/save a buffer containing ONNX model bytes to a custom destination.
```
yuslepukhin: Need more description about what the user function is supposed to do. It would not be convinient to reach out and read C docs. #Resolved

csharp/src/Microsoft.ML.OnnxRuntime/CompileModel.shared.cs
```diff
@@ -159,6 +159,271 @@ public void SetGraphOptimizationLevel(GraphOptimizationLevel graphOptimizationLe
...
+        /// <summary>
+        /// Delegate called by ORT for every initializer when generating the compiled model.
+        /// The delegate allows the user to determine whether the initializer should be stored within the compiled
+        /// model or externally in a file. If the delegate chooses to store an initializer externally, the delegate
+        /// implementation is responsible for writing the initializer data to a file.
+        /// </summary>
+        /// <param name="initializerName">The initializer's name.</param>
+        /// <param name="initializerValue">The readonly OrtValue instance containing the data, type, and
+        /// shape of the initializer.</param>
+        /// <param name="optionalExternalInfo">May be null. If the initializer is originally stored externally,
```
yuslepukhin: Suggest `originalInitializerLocation` #Resolved

csharp/src/Microsoft.ML.OnnxRuntime/CompileModel.shared.cs
```diff
@@ -159,6 +159,271 @@ public void SetGraphOptimizationLevel(GraphOptimizationLevel graphOptimizationLe
                     handle, graphOptimizationLevel));
         }
 
+        /// <summary>
+        /// Delegate to write/save a buffer containing ONNX model bytes to a custom destination.
+        /// </summary>
+        /// <param name="buffer">The buffer to write out.</param>
+        public delegate void WriteBufferDelegate(ReadOnlySpan<byte> buffer);
```
yuslepukhin: Suggest `WriteBufferToUserDesinationDelegate` or some other alternative. #Resolved

csharp/src/Microsoft.ML.OnnxRuntime/CompileModel.shared.cs
```diff
@@ -159,6 +159,271 @@ public void SetGraphOptimizationLevel(GraphOptimizationLevel graphOptimizationLe
...
+        /// </summary>
+        /// <param name="initializerName">The initializer's name.</param>
+        /// <param name="initializerValue">The readonly OrtValue instance containing the data, type, and
+        /// shape of the initializer.</param>
+        /// <param name="optionalExternalInfo">May be null. If the initializer is originally stored externally,
+        /// this contains the file path, file offset, and data size. Otherwise, this is null.</param>
+        /// <returns>A new OrtExternalInitializerInfo indicating the new location of the initializer.
+        /// Returns null if the initializer should be stored within the generated compiled model.</returns>
+        /// <remarks>The return value may be null.</remarks>
+        public delegate OrtExternalInitializerInfo HandleInitializerDelegate(
```
yuslepukhin: Suggest `GetInitializerDestinationDelegate` as a name #Resolved

csharp/src/Microsoft.ML.OnnxRuntime/CompileModel.shared.cs
```diff
@@ -159,6 +159,271 @@ public void SetGraphOptimizationLevel(GraphOptimizationLevel graphOptimizationLe
...
+                    IReadOnlyOrtValue readOnlyInitializerValue = new OrtValue(initializerValue, owned: false);
+                    IReadOnlyExternalInitializerInfo readOnlyExternalInfo = null;
+
+                    if (externalInfo != IntPtr.Zero)
+                    {
+                        readOnlyExternalInfo = new OrtExternalInitializerInfo(externalInfo, ownsHandle: false);
+                    }
+
+                    // Call user's delegate, which may return the new location of the initializer.
+                    OrtExternalInitializerInfo optionalNewExternalInfo = connector._csharpDelegate(
```
yuslepukhin: Suggest instead of `optionalNewExternalInfo` `newInitializerLocation` #Resolved

yuslepukhin: Looks like C# external info and native are disjoint. If it was otherwise, then it would be clearler and the instance would just be passed on to the native code.

csharp/src/Microsoft.ML.OnnxRuntime/CompileModel.shared.cs
```diff
@@ -159,6 +159,271 @@ public void SetGraphOptimizationLevel(GraphOptimizationLevel graphOptimizationLe
...
+            }
+
+            #region IDispose implementation
+            public void Dispose()
+            {
+                DisposeImpl();
+                GC.SuppressFinalize(this);
+            }
+
+            protected virtual void DisposeImpl()
```
yuslepukhin: I think this should override `Dispose(bool)`. There is a long explanation to this, but it is worth doing by the book. #Resolved

csharp/src/Microsoft.ML.OnnxRuntime/CompileModel.shared.cs
```diff
@@ -175,7 +440,22 @@ protected override bool ReleaseHandle()
         {
             NativeMethods.CompileApi.OrtReleaseModelCompilationOptions(handle);
             handle = IntPtr.Zero;
+
+            _writeBufferDelegateState?.Dispose();
```
yuslepukhin: It looks like the class is both `SafeHandle` and custom disposable. 
You can not have both. Well, you can, but it would make things more compicated.
You can not dispose things in ReleaseHandle().
You need to remobe SafeHandle inhertiance, Handle and ReleaseHandle and properly implement IDisposable. This would be simpler and robust. #Resolved

csharp/src/Microsoft.ML.OnnxRuntime/OrtValue.shared.cs
```diff
@@ -1464,11 +1473,155 @@ protected virtual void Dispose(bool disposing)
...
+                NativeMethods.OrtReleaseValue(_handle);
+            }
             _handle = IntPtr.Zero;
             _disposed = true;
         }
 
         #endregion
     }
+
+    /// <summary>
```
yuslepukhin: Suggest moving this interface definition before OrtValue class. #Resolved

csharp/test/Microsoft.ML.OnnxRuntime.Tests.Common/CompileApiTests.cs
```diff
@@ -122,6 +122,141 @@ public void BasicUsage()
             }
         }
     }
+
+    [Fact]
+    public void WriteOutModelWithDelegate()
+    {
+        var sess_options = new SessionOptions();
```
yuslepukhin: sess_options mustbe disposed.
You can use the same line using statement without making a block #Resolved

csharp/test/Microsoft.ML.OnnxRuntime.Tests.Common/CompileApiTests.cs
```diff
@@ -122,6 +122,141 @@ public void BasicUsage()
...
+            {
+                File.Delete(outputModelFilePath);
+            }
+        }
+    }
+
+    [Fact]
+    public void HandleInitializersDelegate()
+    {
+        var sess_options = new SessionOptions();
```
yuslepukhin: Ditto #Resolved

csharp/test/Microsoft.ML.OnnxRuntime.Tests.Common/CompileApiTests.cs
```diff
@@ -122,6 +122,141 @@ public void BasicUsage()
...
+    public void WriteOutModelWithDelegate()
+    {
+        var sess_options = new SessionOptions();
+
+        using (var compileOptions = new OrtModelCompilationOptions(sess_options))
+        {
+            var model = TestDataLoader.LoadModelFromEmbeddedResource("squeezenet.onnx");
+            var outputModelFilePath = "squeezenet_write_delegate_ctx.onnx";
+
+            using (FileStream fs = new FileStream(outputModelFilePath, FileMode.Create, FileAccess.Write))
```
yuslepukhin: Can we add an option `DeleteOnClose` so it is garanteed to be removed? #Resolved

csharp/test/Microsoft.ML.OnnxRuntime.Tests.Common/CompileApiTests.cs
```diff
@@ -122,6 +122,141 @@ public void BasicUsage()
...
+            {
+                File.Delete(outputModelFilePath);
+            }
+        }
+    }
+
+    [Fact]
+    public void HandleInitializersDelegateWithReuse()
+    {
+        var sess_options = new SessionOptions();
```
yuslepukhin: Ditto #Resolved

csharp/src/Microsoft.ML.OnnxRuntime/CompileModel.shared.cs
```diff
@@ -156,26 +156,354 @@ public void SetGraphOptimizationLevel(GraphOptimizationLevel graphOptimizationLe
...
+
+            public IntPtr GetFunctionPointerForDelegate()
+            {
+                return Marshal.GetFunctionPointerForDelegate(_delegate);
+            }
+
+            public IntPtr GetConnectorHandleAsPointer()
+            {
+                return GCHandle.ToIntPtr(_connectorHandle);
+            }
```
yuslepukhin, line 394: Suggest these two being private or internal

csharp/src/Microsoft.ML.OnnxRuntime/CompileModel.shared.cs
```diff
@@ -156,26 +156,354 @@ public void SetGraphOptimizationLevel(GraphOptimizationLevel graphOptimizationLe
...
+
+            private Connector _connector = null;
+            private Delegate _delegate = null;
+            private GCHandle _connectorHandle = default;
+            private GCHandle _delegateHandle = default;
+            private bool _disposed = false;
+        }
+        #endregion
+
+        internal IntPtr Handle => _handle;
```
yuslepukhin: >internal IntPtr Handle => _handle [](http://example.com/codeflow?start=8&length=33)

I suggest to make this readonly

csharp/src/Microsoft.ML.OnnxRuntime/CompileModel.shared.cs
```diff
@@ -156,26 +156,354 @@ public void SetGraphOptimizationLevel(GraphOptimizationLevel graphOptimizationLe
...
+
+            if (disposing)
+            {
+                _writeBufferToDestinationDelegateState?.Dispose();
+                _getInitializerLocationDelegateState?.Dispose();
+            }
+
+            Debug.Assert(_handle != IntPtr.Zero);
+            NativeMethods.CompileApi.OrtReleaseModelCompilationOptions(_handle);
+            _handle = IntPtr.Zero;
```
yuslepukhin, line 472: Dispose() can be called more than once. We want guard against it by checking for Zero.

adrianlizarraga: There's a check above to see if dispose has already been called:
```c#
if (_disposed)
{
    return;
}
```

That will catch the case when it is called a second time.

csharp/src/Microsoft.ML.OnnxRuntime/CompileModel.shared.cs
adrianlizarraga: Made them internal

csharp/src/Microsoft.ML.OnnxRuntime/CompileModel.shared.cs
adrianlizarraga: I removed this getter as it was not used.

yuslepukhin (approved):

PR #25459, opened 2025-07-19
Update some new API functions to directly return the value if they're simple getters.

skottmckay:
### Description

Update some new API functions to directly return the value if they're simple getters.

### Motivation and Context

Simplify API usage.


snnn (dismissed):

snnn:

PR #25463, opened 2025-07-21
[build] fix NPM packaging pipeline

fs-eire:

### DIFF
--- a/tools/ci_build/github/azure-pipelines/templates/linux-wasm-ci.yml
+++ b/tools/ci_build/github/azure-pipelines/templates/linux-wasm-ci.yml
@@ -83,6 +83,20 @@ jobs:
       versionSpec: '3.12'
       addToPath: true
       architecture: $(buildArch)
+  - task: NodeTool@0
+    inputs:
+      # requires Node.js v22 for float16 testing (the V8 flag "--js-float16array")
+      versionSpec: '22.x'
+  # The image used in this workflow has an old version of Node.js installed at /bin/node and npm at /bin/npm.
+  # We need to remove the system Node.js and npm, otherwise CMake will not be able to find the correct Node.js and npm executables.
+  - script: |
+      if [ -f /bin/node ]; then
+        rm /bin/node
+      fi
+      if [ -f /bin/npm ]; then
+        rm /bin/npm
+      fi
+    displayName: Remove the system Node.js
 
   - ${{if eq(parameters.WithCache, true)}}:
     - script: |


tools/ci_build/github/azure-pipelines/templates/linux-wasm-ci.yml
```diff
@@ -83,6 +83,20 @@ jobs:
...
       addToPath: true
       architecture: $(buildArch)
+  - task: NodeTool@0
+    inputs:
+      # requires Node.js v22 for float16 testing (the V8 flag "--js-float16array")
+      versionSpec: '22.x'
+  # The image used in this workflow has an old version of Node.js installed at /bin/node and npm at /bin/npm.
+  # We need to remove the system Node.js and npm, otherwise CMake will not be able to find the correct Node.js and npm executables.
+  - script: |
+      if [ -f /bin/node ]; then
```
snnn, line 93: But, both Azure DevOps agent and Github agent are based on nodejs. if you delete this, will it break things?

snnn: Oh, the image has 
```
jammy-updates/universe amd64 nodejs amd64 12.22.9~dfsg-1ubuntu3.6 
```

snnn: Please hold on. Let me see if I can avoid installing nodejs when building the image.

snnn: I know which artifact pulled it in.

fs-eire: The previous line installed Node.js, and it will be searched correctly in shell.

The problem is about CMake, it seems to search `/bin/` folder first when using `find_program()`

snnn: Every system image has one. 

Also I need to address the following issue:
https://ubuntu.com/security/notices/USN-7469-3

I mean, when I built the image, there was some code pulled in a very old version of nodejs from apt-get. 
And because of the above security advisor, I need to upgrade the operating system to 24.04.

snnn: Please try to solve this problem from vcpkg side, instead of modifying the system image. If the system needs nodejs, you cannot delete it. If it doesn't need nodejs, I will not install it.

fs-eire:
replaced by #25466

tools/ci_build/github/azure-pipelines/templates/linux-wasm-ci.yml
fs-eire: change continues in a new PR. This is not a VCPKG issue, it is a Cmake issue.

PR #25465, opened 2025-07-21
Add arena that uses EP API so that an EP library can be self-sufficient.

skottmckay:
### Description

Add arena that uses EP API so that an EP library can be self-sufficient.
Remove cross stream sharing from BFCArena. Nothing is using it and it creates a dependency on synchronizing streams inside the arena implementation.
Tried to simplify the Stream/Notification usage. 

Current setup adds an AllocOnStream to OrtAllocator. There's no stream aware Free at this point as ORT does not attach the Stream to the memory usage so can't pass it in to the Free call. 

### Motivation and Context

If ORT adds BFCArena to an OrtAllocator from the EP we have OrtAllocator -> IAllocator wrapper -> BFCArena IAllocator [-> OrtAllocator wrapper for external usage].

The EP managing its own arena is much simpler.

### DIFF
--- a/include/onnxruntime/core/framework/allocator.h
+++ b/include/onnxruntime/core/framework/allocator.h
@@ -86,8 +86,11 @@ class Stream;
 namespace synchronize {
 class Notification;
 }
+
 using WaitNotificationFn = std::function<void(Stream*, synchronize::Notification&)>;
-void* AllocateBufferWithOptions(IAllocator& allocator, size_t size, bool use_reserve, Stream* stream, WaitNotificationFn wait_fn);
+void* AllocateBufferWithOptions(IAllocator& allocator, size_t size, bool use_reserve, Stream* stream,
+                                // wait fn is for backwards compat with provider bridge
+                                WaitNotificationFn ignored = nullptr);
 
 template <typename T>
 using IAllocatorUniquePtr = std::unique_ptr<T, std::function<void(T*)>>;
@@ -104,6 +107,9 @@ class IAllocator {
    * @remarks Use SafeInt when calculating the size of memory to allocate using Alloc.
    */
   virtual void* Alloc(size_t size) = 0;
+
+  virtual bool IsStreamAware() const { return false; }
+  virtual void* AllocOnStream(size_t size, Stream* /*stream*/) { return Alloc(size); }
 
   /**
    * Free memory at p.
@@ -192,7 +198,7 @@ class IAllocator {
   template <typename T>
   static IAllocatorUniquePtr<T> MakeUniquePtr(std::shared_ptr<IAllocator> allocator, size_t count_or_bytes,
                                               bool use_reserve = false,
-                                              Stream* stream = nullptr, WaitNotificationFn wait_fn = nullptr) {
+                                              Stream* stream = nullptr) {
     ValidateAllocator(allocator);
 
     // for now limit to fundamental types. we could support others, but to do so either we or the caller
@@ -210,7 +216,7 @@ class IAllocator {
     }
 
     // allocate
-    T* p = static_cast<T*>(AllocateBufferWithOptions(*allocator, alloc_size, use_reserve, stream, std::move(wait_fn)));
+    T* p = static_cast<T*>(AllocateBufferWithOptions(*allocator, alloc_size, use_reserve, stream, nullptr));
     ValidateAllocation(p, alloc_size);
 
     return IAllocatorUniquePtr<T>{p,
--- a/include/onnxruntime/core/framework/stream_handles.h
+++ b/include/onnxruntime/core/framework/stream_handles.h
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 #pragma once
 
+#include <atomic>
 #include <functional>
 #include <optional>
 #include <unordered_map>
@@ -21,9 +22,9 @@ namespace synchronize {
 class Notification;
 }
 
-// a stream abstraction which hold an opaque handle, and a reference to which OrtDevice instance this stream belong to.
-// it need to be OrtDevice instance as we might have different stream on different OrtDevice with same type.
-// i.e. different cuda stream on different GPU.
+/// <summary>
+/// Class to represent a stream on the OrtDevice.
+/// </summary>
 class Stream {
  public:
   Stream(StreamHandle h, const OrtDevice& d)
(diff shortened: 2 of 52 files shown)


onnxruntime/test/autoep/library/ep_arena.cc
```diff
@@ -0,0 +1,778 @@
...
+
+  auto safe_alloc = [this](size_t alloc_bytes) {
+    void* new_mem = nullptr;
+    try {
+      new_mem = device_allocator_->Alloc(device_allocator_.get(), alloc_bytes);
+    } catch (const std::bad_alloc&) {
+      // attempted allocation can throw std::bad_alloc. we want to treat this the same as if it returned nullptr
+      // so swallow the exception
+    }
+    // catch (const MyException& exception) {
```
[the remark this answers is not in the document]
skottmckay: Intentional.

yuslepukhin:
LGTM

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -372,8 +377,26 @@ typedef struct OrtAllocator {
...
+   *
+   * \param[in] this_ OrtAllocator instance
+   * \param[in] size Size of the allocation in bytes. nullptr if size was 0 or allocation failed.
+   * \param[in] stream The stream to allocate on.
+   *
+   * \return pointer to an allocated block of `size` bytes
+   *
+   * \since 1.23
+   */
+  void*(ORT_API_CALL* AllocOnStream)(struct OrtAllocator* this_, size_t size, OrtSyncStream* stream);
```
skottmckay, line 405: Not sure what the best option is here. This is the minimal required for something based on BFCArena to be used by the EP. requires something on the EP side (arena or otherwise) to track the stream that was used for an allocation.

Alternatively there's async alloc/free (in CUDA at least) which uses a stream, but that's not what we're using currently. We have a stream aware arena implementation not async alloc/free. currently nothing on the ORT side tracks the stream that was used in an allocation in order to provide it to an async free call, but this could be done on the ORT or EP side.

Also worth considering if we should split out stream support from OrtAllocator into a separate struct so we can add new stream related functions (if needed) in that struct instead of the base OrtAllocator. But a little weird with it only having one function in it, and potentially that may never change.

onnxruntime/core/framework/execution_frame.cc
```diff
@@ -441,13 +430,23 @@ ExecutionFrame::ExecutionFrame(gsl::span<const int> feed_mlvalue_idxs, gsl::span
...
+                // given that, it's unclear why we need to update the sync id in all other streams to allow them
+                // to take this buffer if it was free.
+                //
+                // device_stream_collection calls ReleaseStreamBuffers for all streams including the root stream in
+                // CleanUp, so the chunk will become available to other streams at that point.
+                //
+                // Commenting out to verify.
+                // for (size_t j = 0; j < device_streams_->NumStreams(); j++) {
+                //  stream_aware_arena->WaitOnChunk(mem_pattern_stream, device_streams_->GetStream(j));
+                //}
```
skottmckay, line 449: This doesn't seem to be required and adds complexity/more dependencies to the arena implementation to support.

onnxruntime/core/framework/tensor_allocator_with_mem_pattern.h
```diff
@@ -39,14 +39,8 @@ class TensorAllocatorWithMemPattern : public ITensorAllocator {
...
-      void* buffer;
-      if (alloc->Info().alloc_type == OrtArenaAllocator) {
-        // Arena has a specific way to store static memory.
-        // Arena does not reuse static memory allocated by Reserve.
-        buffer = static_cast<BFCArena*>(alloc.get())->Reserve(peak_size);
-      } else {
-        buffer = alloc->Alloc(peak_size);
-      }
+      // use Reserve for initializers so they don't affect arena growth patterns if an arena is involved.
+      void* buffer = alloc->Reserve(peak_size);
```
skottmckay, line 43: Most allocators simply forward `Reserve` to `Alloc` so I don't see a need to limit the usage of Reserve to only the internal BFCArena.

onnxruntime/core/session/allocator_adapters.cc
```diff
@@ -297,7 +329,6 @@ ORT_API_STATUS_IMPL(OrtApis::CreateSharedAllocator,
                     [[maybe_unused]] _In_ OrtEnv* ort_env,
                     [[maybe_unused]] _In_ const OrtEpDevice* ep_device,
                     [[maybe_unused]] _In_ OrtDeviceMemoryType mem_type,
-                    [[maybe_unused]] _In_ OrtAllocatorType allocator_type,
```
skottmckay: This is to go back to the currently ORT API limitation that OrtArenaAllocator is only used for the ORT BFCArena. 

EPs being in control of whether they use an arena or not means ORT doesn't need to be in the business of wrapping the EP allocator, and doing so was adding a lot of messy complexity to environment.cc/h.

onnxruntime/core/session/ep_api.cc
```diff
@@ -179,6 +179,31 @@ ORT_API(uint32_t, MemoryDevice_GetDeviceId, _In_ const OrtMemoryDevice* memory_d
...
+}
+
+ORT_API(uint64_t, GetSyncIdForLastWaitOnSyncStream, _In_ const OrtSyncStream* producer_stream,
+        _In_ const OrtSyncStream* consumer_stream) {
+  uint64_t id{0};
+  if (producer_stream && consumer_stream) {
+    const auto& producer = *static_cast<const Stream*>(producer_stream);
+    const auto& consumer = *static_cast<const Stream*>(consumer_stream);
+
+    // If both streams are valid, we can return the sync id for the last wait on the consumer stream.
```
skottmckay:
```suggestion
    // If both streams are valid, we can return the sync id for the last wait on the producer stream.
```

onnxruntime/core/session/inference_session.cc
```diff
@@ -3571,7 +3571,7 @@ common::Status InferenceSession::ValidateAndParseShrinkArenaString(const std::st
       ++iter;
     }
 
-    // Shrink if it is an arena based allocator
+    // Shrink if it is a BFCArena allocator
```
skottmckay, line 3574: Currently the EP arena does not have any shrink support. EP author is free to add it. Would we need/want anything in the ORT API to trigger that, or can they handle it internally (e.g. based on events like EP OnRunEnd)?

onnxruntime/test/autoep/library/ep_arena.h
```diff
@@ -0,0 +1,629 @@
...
+  }
+};
+
+// A memory allocator that implements a 'best-fit with coalescing' algorithm.
+// This is essentially a very simple version of Doug Lea's malloc (dlmalloc).
+//
+// The goal of this allocator is to support defragmentation via coalescing.
+// One assumption we make is that the process using this allocator owns pretty much all of the memory, and that nearly
+// all requests to allocate memory go through this interface.
+class ArenaImpl {
```
skottmckay, line 126: Copied from BFCArena
- updated to use EP API
- removed Shrink to simplify
- arena is stream aware by default
  - passing in a stream is optional
- tracking stream to chunk usage directly via stream_to_chunks_
  - BFCArena iterates all allocations to find chunks using a stream when marking that stream as free which felt inefficient given we're going to call that logic once per stream used in the session (i.e. num streams * num allocations worth of checks)
- added tracking of OrtSyncStreamImpl to OrtSyncStream
  - the OrtEpFactory/OrtEp creates an OrtSyncStreamImpl. ORT creates a Stream from this with the wrapper class. OrtAllocator uses the opaque OrtSyncStream. need to connect things in OrtSyncStreamImpl::OnSessionRunEnd as that is the point where we know it's safe to remove the assignment to the stream for a chunk.
- ArenaImp uses ArenaConfig directly instead of having individual members for all the values in that
- default settings etc. are unchanged from BFCArena

onnxruntime/test/autoep/library/ep_factory.cc
```diff
@@ -206,17 +209,62 @@ OrtStatus* ORT_API_CALL ExampleEpFactory::CreateAllocatorImpl(OrtEpFactory* this
...
+    return factory.ort_api.CreateStatus(ORT_INVALID_ARGUMENT, "Unexpected OrtDeviceMemoryType.");
+  }
+
   // NOTE: The factory implementation is free to return a shared OrtAllocator* instance instead of creating a new
   //       allocator on each call. To do this have an allocator instance as an OrtEpFactory class member and make
   //       ReleaseAllocatorImpl a no-op.
-  auto cpu_allocator = std::make_unique<CustomAllocator>(memory_info, factory);
-  *allocator = cpu_allocator.release();
+  //
+  // NOTE: EP should implement it's own arena logic. ep_arena.cc/h is provided as a reference and we use it here for
```
skottmckay:
```suggestion
  // NOTE: EP should implement its own arena logic. ep_arena.cc/h is provided as a reference and we use it here for
```

include/onnxruntime/core/framework/allocator.h
```diff
@@ -105,6 +108,9 @@ class IAllocator {
    */
   virtual void* Alloc(size_t size) = 0;
 
+  virtual bool IsStreamAware() const { return false; }
+  virtual void* AllocOnStream(size_t size, Stream* /*stream*/) { return Alloc(size); }
```
edgchen1: documentation would be helpful. e.g., explaining the intended usage.

onnxruntime/core/framework/execution_steps.cc
```diff
@@ -23,25 +23,25 @@ std::string BarrierStep::ToString() const {
...
+      wait_fn_(handle),
+      notification_idx_(idx) {}
 
 Status WaitOnEPStep::Execute(StreamExecutionContext& ctx,
                              size_t stream_idx,
                              SessionScope& /*session_scope*/,
                              const bool& /*terminate_flag*/,
                              bool& continue_flag) {
-  ORT_ENFORCE(wait_handle_, "WaitOnEPStep.wait_handle is null");
+  ORT_ENFORCE(wait_fn_, "WaitOnEPStep.wait_handle is null");
```
edgchen1:
```suggestion
  ORT_ENFORCE(wait_fn_, "WaitOnEPStep::wait_fn_ is null");
```

edgchen1: I wonder if this check should be done in the ctor instead.

onnxruntime/core/framework/utils.cc
```diff
@@ -74,28 +74,21 @@ static common::Status AllocateHelper(const AllocatorPtr& allocator,
...
-                             allocator, target_mlvalue);
-      } else {
-        Tensor::InitOrtValue(source_tensor.DataType(),
-                             source_tensor.Shape(),
-                             p_data,
-                             allocator, target_mlvalue);
+    void* p_data = nullptr;
+    if (target_stream && allocator->IsStreamAware()) {
+      size_t len = Tensor::CalculateTensorStorageSize(source_tensor.DataType(), source_tensor.Shape());
+      p_data = allocator->AllocOnStream(len, target_stream);
```
edgchen1, line 80: is the stream aware allocation stuff only relevant for tensors? I realize the existing code only handles tensors too.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -372,8 +377,26 @@ typedef struct OrtAllocator {
...
+   *
+   * \since 1.23
    */
   ORT_API2_STATUS(GetStats, _In_ const struct OrtAllocator* this_, _Outptr_ OrtKeyValuePairs** out);
+
+  /** \brief Allocate using a stream.
+   *
+   * If the allocator is stream aware this performs allocation using a stream.
+   *
+   * Alloc will be used if this is nullptr.
```
edgchen1, line 394: in general, would be good to have a consistent way to indicate which functions are optional to implement.

perhaps something like this:

```
  \note Implementation of this function is optional.
```

in this case, perhaps it is worth mentioning that this function is implemented if and only if the allocator is stream aware.

onnxruntime/core/framework/utils.cc
skottmckay: I assume it could be used for any allocation (the arena is allocating a size not a data type), but was keeping it consistent with the current code.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -6132,8 +6154,7 @@ struct OrtApi {
    * \since Version 1.23
    */
   ORT_API2_STATUS(CreateSharedAllocator, _In_ OrtEnv* env, _In_ const OrtEpDevice* ep_device,
-                  _In_ OrtDeviceMemoryType mem_type, _In_ OrtAllocatorType allocator_type,
-                  _In_opt_ const OrtKeyValuePairs* allocator_options,
+                  _In_ OrtDeviceMemoryType mem_type, _In_opt_ const OrtKeyValuePairs* allocator_options,
```
skottmckay: This change was to keep the OrtArenaAllocator as an internal only type.

But there's also the new read-only allocator from #25348. Is there a valid use case where a user would need/want to use the read-only allocator to create on-device shared weights outside of an inference session. If so, leaving the OrtAllocatorType in here would be required.

include/onnxruntime/core/framework/stream_handles.h
```diff
@@ -21,133 +23,127 @@ namespace synchronize {
...
+  // current sync id. equivalent to a counter for the number of events we have recorded in the underlying stream.
+  // 0 == no events recorded. sync_id_ is updated prior to recording a new event.
+  std::atomic<uint64_t> sync_id_{0};
+
+  // This is a map to track synchronization points between streams. When we wait on another stream (the producer)
+  // we add an entry to the map for that stream.
+  // The entry has the sync id from the producer stream for the event we waited on.
+  //
   // TODO: use inline container.
   // currently this class is header only, but abseil doesn't compile with nvcc
```
edgchen1, line 101: > currently this class is header only

it isn't anymore

onnxruntime/core/session/ep_api.cc
```diff
@@ -176,6 +176,31 @@ ORT_API(uint32_t, MemoryDevice_GetDeviceId, _In_ const OrtMemoryDevice* memory_d
   return memory_device->Id();
 }
 
+ORT_API(const OrtSyncStreamImpl*, SyncStream_GetImpl, _In_ const OrtSyncStream* ort_stream) {
```
edgchen1, line 183: we should ensure that no exceptions will be thrown from API functions.

onnxruntime/core/session/ort_apis.h
```diff
@@ -729,4 +728,7 @@ ORT_API_STATUS_IMPL(CopyTensors, _In_ const OrtEnv* env,
                     _In_reads_(num_tensors) OrtValue* const* dst_tensors,
                     _In_opt_ OrtSyncStream* stream,
                     _In_ size_t num_tensors);
+
+ORT_API_STATUS_IMPL(GetSessionOptionsConfigEntries, _In_ const OrtSessionOptions* options,
```
edgchen1: this is already declared above

include/onnxruntime/core/framework/stream_handles.h
skottmckay: We could inline the implementation of UpdateWithAwaitedNotification but I don't quite see why this requires that given it shouldn't be performance critical. That seems separate to concerns about usage across the provider bridge though, and abseil compatibility with nvcc seems separate again.

I'll remove the comment for now. If someone has the time and wants to try and use an inlined container here and test nvcc compatibility they're free to do so in the future.

include/onnxruntime/core/session/onnxruntime_c_api.h
skottmckay: Going to remove this change. The allocator type is potentially useful in the future for read-only memory, and we could also additional OrtAllocatorType values

onnxruntime/core/session/ep_api.cc
skottmckay: What do you propose? This would apply to all existing API functions that don't return a status which afaics are simple getters or 'release' functions. Exception would be from the user providing an invalid pointer to the object a value is being retrieved from or for the object being freed. 

Best I can think of is a try/catch that uses the default logger, but that feels pretty heavy for these scenarios. Have we had any issues before with this?

onnxruntime/core/session/allocator_adapters.cc
skottmckay: Reverted to allow specifying a read-only allocator in the future. Will be added in #25348

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -5243,7 +5272,7 @@ struct OrtApi {
   /** \brief Get the hardware device type.
    *
    * \param[in] device The OrtHardwareDevice instance to query.
-   * \return The hardware device type.
+   * \return The hardware deGetTevice type.
```
jslhcl: typo?

onnxruntime/core/session/ep_api.cc
edgchen1: explicit try/catch at the level of the API function would work. this is probably the more maintainable approach. do you mean "heavy" in the sense of code to write or performance overhead? we already have macros which do this for OrtStatus APIs. otherwise, we need to ensure that code and other function calls in the API functions will not throw exceptions.

skottmckay: I was thinking 'heavy' in the sense that the implementation cost of this handling is relatively high compared to the expected failure rate. 

if the function returns a status it's simple. catch exception, return error message. that also has overhead on the client side (we have to check every single status for error and handle it), but that's reasonable if there's any potential reasons the function may fail even if the input to it was valid.

for functions that do not return a status, my base assumption would be that unless the input is completely invalid, there is zero chance the function would fail. e.g. if you were implementing the C++ getter for a validly initialized class I don't think you would ever consider adding error handling to that getter or the caller.

so having wrapper try/catch and a macro to get the default logger and format up a log message and define a valid return value feels like a lot of cost if we never expect it to fail.

if we want to go to that effort and pay that cost, does that also imply that for every single API function implementation the first lines of code should be validating that no required input pointer args are a nullptr because if we deference a nullptr it's a segfault not an exception isn't it and the try/catch isn't going to help us? 

we are certainly free to make that choice to ensure all the api functions are bullet proof, but I'm not sure there's a requirement driving that currently.

skottmckay: Resolving purely to checkin the PR as any change along these lines will be a separate PR.

snnn:

PR #25469, opened 2025-07-21
POWER : MlasGemmQuantKernel implemented with VSX when M = 1

BODAPATIMAHESH:
### Description
POWER : Implement MlasGemmQuantKernel using VSX builtins for M = 1
Added a VSX-based implementation of MlasGemmQuantKernel optimized for the case when M = 1.
Verified correctness using ONNX Runtime's built-in tests and onnxruntime_mlas_tests; no regressions observed.
Evaluated performance using a Granite 8-bit quantized model and observed approximately 5% improvement in token generation speed.

### Motivation and Context
To improve token generation performance for models with a batch size of 1


BODAPATIMAHESH:
Created a new PR .closing this PR.

PR #25473, opened 2025-07-21
Fix EP context name for cache file name with two dots

GoranDubajic:
As `name` is already derived from stem (of cache file name), ff cache file name contains two dots then blob file name is incorrect. This enables cache file name with two dots.


jywu-mysoft:
@sfatimar , @javier-intel fyi

javier-intel (approved):

jywu-msft (approved):

snnn:

PR #25474, opened 2025-07-21
Compile API: disable optimizations by default

adrianlizarraga:
### Description
- Disables graph optimizations by default when using the explicit compiling API.
- Adds `ModelCompilationOptions_SetGraphOptimizationLevel` to allow the user to set an optimization level.
- Adds C++, Python, and C# bindings for the new API function.
- Updates `ModelCompilationOptions_SetFlags` to take in a `uint32_t flags` parameter instead of `size_t flags` to ensure the same size across platforms. This API is not yet in a public ORT release, so safe to modify.

### Motivation and Context
When compiling, prefer allowing the EP to do the optimizations instead of ORT.


onnxruntime/core/session/compile_api.h
```diff
@@ -32,5 +32,8 @@ ORT_API_STATUS_IMPL(ModelCompilationOptions_SetFlags, _In_ OrtModelCompilationOp
                     size_t flags);
 ORT_API_STATUS_IMPL(ModelCompilationOptions_SetEpContextBinaryInformation, _In_ OrtModelCompilationOptions* model_compile_options,
                     _In_ const ORTCHAR_T* output_dir, _In_ const ORTCHAR_T* model_name);
+ORT_API_STATUS_IMPL(ModelCompilationOptions_SetGraphOptimizationLevel,
```
yuslepukhin, line 35: I am a little confused about `ModelCompilationOptions_SetOutputModelBuffer`. Should this be Get* since it obtains the buffer?

onnxruntime/core/session/compile_api.h
```diff
@@ -32,5 +32,8 @@ ORT_API_STATUS_IMPL(ModelCompilationOptions_SetFlags, _In_ OrtModelCompilationOp
...
+                    _In_ OrtModelCompilationOptions* model_compile_options,
+                    _In_ GraphOptimizationLevel graph_optimization_level);
```
yuslepukhin, line 38: Another comment about `ModelCompilationOptions_SetFlags`. When we deal with flags, the number of bits must be fixed, but `size_t` is platform and arch dependent.

onnxruntime/test/providers/qnn/qnn_ep_context_test.cc
```diff
@@ -234,15 +234,20 @@ static void CreateTestModel(test::GetTestModelFn graph_builder,
 // Helper that checks that a compiled model has the expected number of EPContext nodes.
 static void CheckEpContextNodeCounts(const onnxruntime::Model& ep_ctx_model,
                                      int expected_ep_context_node_count,
-                                     int expected_other_node_count) {
+                                     int expected_other_node_count,
+                                     bool optimizations_disabled = false) {
```
yuslepukhin: Good comment from Co-pilot about that disabled does not mean default.

onnxruntime/core/session/model_compilation_options.cc
```diff
@@ -170,6 +172,31 @@ void ModelCompilationOptions::ResetInputModelSettings() {
   input_model_data_size_ = 0;
 }
 
+Status ModelCompilationOptions::SetGraphOptimizationLevel(GraphOptimizationLevel graph_optimization_level) {
+  switch (graph_optimization_level) {
+    case ORT_DISABLE_ALL:
+      session_options_.value.graph_optimization_level = onnxruntime::TransformerLevel::Default;
```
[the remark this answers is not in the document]
adrianlizarraga: ORT currently maps `ORT_DISABLE_ALL` to `TransformerLevel::Default`. That decision was not made in this PR.

csharp/src/Microsoft.ML.OnnxRuntime/CompileModel.shared.cs
```diff
@@ -130,6 +132,33 @@ public void SetFlags(OrtCompileApiFlags flags)
...
         }
 
+        /// <summary>
+        /// Sets information related to EP context binary file. The Ep uses this information to decide the
+        /// location and context binary file name when compiling with both the input and output models
+        /// stored in buffers.
+        /// </summary>
+        /// <param name="outputDirectory">Path to the model directory.</param>
+        /// <param name="modelName">The name of the model.</param>
+        public void SetEpContextBinaryInformation(string outputDirectory, string modelName)
```
adrianlizarraga, line 142: Note: this API was added in a previous PR and is missing a corresponding C# binding. Adding it here.

onnxruntime/core/session/model_compilation_options.cc
```diff
@@ -170,6 +172,31 @@ void ModelCompilationOptions::ResetInputModelSettings() {
...
+      break;
+    case ORT_ENABLE_LAYOUT:
+      session_options_.value.graph_optimization_level = onnxruntime::TransformerLevel::Level3;
+      break;
+    case ORT_ENABLE_ALL:
+      session_options_.value.graph_optimization_level = onnxruntime::TransformerLevel::MaxLevel;
+      break;
+    default:
+      return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT, "graph_optimization_level with value ",
+                             static_cast<int>(graph_optimization_level), " is invalid");
```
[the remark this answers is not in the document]
adrianlizarraga: Updated

onnxruntime/core/session/compile_api.h
adrianlizarraga: That API function was released in ORT 1.22 so we can't update the name.

But to your point, yeah this may be a bit confusing. It is called Set* because this function sets options for compilation and does not return anything (other than a status).

To be more specific, the user calls `ModelCompilationOptions_SetOutputModelBuffer` to set a pointer and an allocator on an options object. Later, when `Compile()` is called, ORT uses the allocator to allocate a buffer with the model bytes. The user's pointer is updated to point to the buffer.

onnxruntime/core/session/compile_api.h
adrianlizarraga: Good call. I updated `ModelCompilationOptions_SetFlags` to use `uint32_t` instead of `size_t`.

onnxruntime/test/providers/qnn/qnn_ep_context_test.cc
adrianlizarraga: Added comments in code. Hopefully this is clearer.

In summary, `ORT_DISABLE_ALL` does currently map to `onnxruntime::TransformerLevel::Default` everywhere in ORT. I did not change that mapping in this PR. Perhaps `TransformerLevel::Default` is not the best name but in terms of functionality, it disables all transformers except for required ones.

onnxruntime/core/session/compile_api.h
yuslepukhin: Ok, I figured it out.

yuslepukhin (approved):

PR #25475, opened 2025-07-21
[VitisAI] Update CreateEpFactories with default logger 

nieubank:
Update to accomodate changes from #25456


snnn:

PR #25480, opened 2025-07-21
fix warning as error for QNN EP

HectorSVC:
### Description
Fix an issue from warning as error for QNN EP
Always enable warning as error for QNN EP including CI

### DIFF
--- a/tools/ci_build/build.py
+++ b/tools/ci_build/build.py
@@ -186,8 +186,6 @@ def use_dev_mode(args):
         return False
     if args.use_armnn:
         return False
-    if args.use_qnn:
-        return True
     if is_macOS() and (args.ios or args.visionos or args.tvos):
         return False
     SYSTEM_COLLECTIONURI = os.getenv("SYSTEM_COLLECTIONURI")  # noqa: N806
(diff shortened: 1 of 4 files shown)


jywu-mysoft:
lint error

HectorSVC:
> lint error

that's CI issue

tools/ci_build/build.py
```diff
@@ -188,6 +188,8 @@ def use_dev_mode(args):
         return False
     if is_macOS() and (args.ios or args.visionos or args.tvos):
         return False
+    if args.use_qnn:
```
jywu-mysoft, line 191: did we confirm this is needed because of the check (lines 194-195) below?
why is it needed currently? would we want to get rid of it? 
are there other pipelines that would fail if we got rid of it

snnn (approved):

tools/ci_build/build.py
HectorSVC: The CIs which are not Github Actions still build with compile_no_warning_as_error. Need more effort to resolve all the warning as errors.

PR #25482, opened 2025-07-21
[DRAFT] [EP ABI] Load external initializer OrtValues on demand

adrianlizarraga:

### DIFF
--- a/include/onnxruntime/core/graph/graph.h
+++ b/include/onnxruntime/core/graph/graph.h
@@ -25,6 +25,7 @@
 #include "core/common/inlined_containers.h"
 #endif
 #include "core/common/span_utils.h"
+#include "core/common/safeint.h"
 #include "core/common/status.h"
 #include "core/common/logging/logging.h"
 #include "core/framework/ort_value.h"
@@ -688,6 +689,15 @@ class Node {
   bool can_be_saved_;
 };
 
+// Stores information on an initializer (TensorProto) that is stored in an external file.
+// Graph tracks this information because most initializers are now TensorProtos that refer to an
+// OrtValue Tensor, which looks like an external initializer, but isn't.
+struct ExternalInitializerInfo {
+  std::basic_string<ORTCHAR_T> file_path;
+  int64_t file_offset = 0;
+  SafeInt<size_t> tensor_byte_size = 0;
+};
+
 /**
 @class Graph
 The Graph representation containing the graph inputs and outputs, the Node instances,
@@ -787,6 +797,11 @@ class Graph {  // NOLINT(clang-analyzer-optin.performance.Padding): preserve exi
   /** Populate `value` if an externally allocated OrtValue exists for an initializer with the given name.
    */
   bool GetOrtValueInitializer(const std::string& name, OrtValue& value, bool check_outer_scope = false) const;
+
+  Status LoadInitializerAsOrtValue(const std::string& name, OrtValue& value);
+
+  bool GetExternalInitializerInfo(const std::string& name, ExternalInitializerInfo& ext_info,
+                                  bool check_outer_scope = false) const;
 
   /** Gets all the initializer tensors in this Graph. */
   const InitializedTensorSet& GetAllInitializedTensors() const noexcept { return name_to_initial_tensor_; }
@@ -1800,6 +1815,8 @@ class Graph {  // NOLINT(clang-analyzer-optin.performance.Padding): preserve exi
   // As we need to convert to TensorProto for the optimizers to work and keep the deleter information we store them
   // in the Graph instance and retrieve during session state finalization.
   std::unordered_map<std::string, OrtValue> ortvalue_initializers_;
+
+  std::unordered_map<std::string, ExternalInitializerInfo> ext_initializers_;
 
   std::unordered_set<std::reference_wrapper<const std::string>,
                      std::hash<std::string>, std::equal_to<std::string>>
--- a/onnxruntime/core/graph/ep_api_types.cc
+++ b/onnxruntime/core/graph/ep_api_types.cc
@@ -452,7 +452,7 @@ Status EpValueInfo::GetInitializerValue(const OrtValue*& result) const {
 
   // This gets an initializer value defined in this graph or in a parent graph (as long as the value
   // is used in this graph).
-  result = graph_->GetInitializerValue(name_);
+  ORT_RETURN_IF_ERROR(graph_->GetInitializerValue(name_, result));
   ORT_RETURN_IF(result == nullptr, "Unable to find initializer value named '", name_, "'.");
   return Status::OK();
 }
@@ -598,10 +598,16 @@ Status EpGraph::CreateImpl(std::unique_ptr<EpGraph> ep_graph, const GraphViewer&
                                                                              /*check_outer_scope*/ false);
 
     if (!graph_has_ortvalue) {
-      // onnxruntime::Graph does not have an OrtValue for this initializer, so create one from the TensorProto.
-      // This should only happen for small initializers that are needed for ONNX shape inferencing.
-      ORT_RETURN_IF_ERROR(utils::TensorProtoToOrtValue(Env::Default(), graph_viewer.ModelPath(), *tensor_proto,
-                                                       initializer_allocator, *initializer_value));
+      ExternalInitializerInfo ext_info = {};
+
+      if (graph_viewer.GetGraph().GetExternalInitializerInfo(initializer_name, ext_info, /*check_outer_scope*/ false)) {
+        // Do nothing for external initializers. Will load/mmap into an OrtValue on demand.
+        assert(!initializer_value->IsAllocated());
+      } else {
+        // Copy to OrtValue. This should only happen for small initializers.
+        ORT_RETURN_IF_ERROR(utils::TensorProtoToOrtValue(Env::Default(), graph_viewer.ModelPath(), *tensor_proto,
+                                                         initializer_allocator, *initializer_value));
+      }
     }
 
     initializer_values.emplace(value_info->GetName(), std::move(initializer_value));
@@ -665,11 +671,16 @@ Status EpGraph::CreateImpl(std::unique_ptr<EpGraph> ep_graph, const GraphViewer&
       // Add the OrtValue if this is an initializer.
       if (outer_initializer != nullptr) {
         if (!outer_initializer_value->IsAllocated()) {
-          // onnxruntime::Graph does not have an OrtValue for this initializer, so create one from the TensorProto.
-          // This should only happen for small initializers that are needed for ONNX shape inferencing.
-          ORT_RETURN_IF_ERROR(utils::TensorProtoToOrtValue(Env::Default(), parent_graph->ModelPath(),
-                                                           *outer_initializer, initializer_allocator,
-                                                           *outer_initializer_value));
+          ExternalInitializerInfo ext_info = {};
+
+          if (parent_graph->GetExternalInitializerInfo(implicit_name, ext_info, /*check_outer_scope*/ true)) {
+            // Do nothing for external initializers. Will load/mmap into an OrtValue on demand.
+          } else {
+            // Copy to OrtValue. This should only happen for small initializers.
+            ORT_RETURN_IF_ERROR(utils::TensorProtoToOrtValue(Env::Default(), parent_graph->ModelPath(),
+                                                             *outer_initializer, initializer_allocator,
+                                                             *outer_initializer_value));
+          }
         }
         outer_scope_initializer_values.emplace(outer_value_info->GetName(), std::move(outer_initializer_value));
       }
@@ -804,20 +815,39 @@ const EpNode* EpGraph::GetNode(NodeIndex node_index) const {
   return index_to_ep_node_.GetEpNode(node_index);
 }
 
-const OrtValue* EpGraph::GetInitializerValue(std::string_view name) const {
+Status EpGraph::GetInitializerValue(std::string_view name, const OrtValue*& result) const {
+  auto ensure_ort_value_loaded = [&](const std::unique_ptr<OrtValue>& ort_value) -> Status {
+    if (!ort_value->IsAllocated()) {
+      // Lazy load the OrtValue. This happens for external initializers.
+      Graph& graph = const_cast<Graph&>(graph_viewer_.GetGraph());
+      ORT_RETURN_IF_ERROR(graph.LoadInitializerAsOrtValue(std::string(name), const_cast<OrtValue&>(*ort_value)));
+    }
+
+    return Status::OK();
+  };
+
   // Check for initializer value in the graph's scope.
   if (auto iter = initializer_values_.find(name);
       iter != initializer_values_.end()) {
-    return iter->second.get();
+    const std::unique_ptr<OrtValue>& ort_value = iter->second;
+    ORT_RETURN_IF_ERROR(ensure_ort_value_loaded(ort_value));
+
+    result = ort_value.get();
+    return Status::OK();
   }
 
   // Check for the initializer value in an outer scope.
   // Only finds a value if the outer initializer value is used within this graph.
   if (auto iter = outer_scope_initializer_values_.find(name);
       iter != outer_scope_initializer_values_.end()) {
-    return iter->second.get();
+    const std::unique_ptr<OrtValue>& ort_value = iter->second;
+    ORT_RETURN_IF_ERROR(ensure_ort_value_loaded(ort_value));
+
+    result = ort_value.get();
+    return Status::OK();
   }
 
-  return nullptr;
+  result = nullptr;
+  return Status::OK();
 }
 }  // namespace onnxruntime
--- a/onnxruntime/core/graph/ep_api_types.cc
+++ b/onnxruntime/core/graph/ep_api_types.cc
@@ -593,6 +593,7 @@ Status EpGraph::CreateImpl(std::unique_ptr<EpGraph> ep_graph, const GraphViewer&
     initializer_value_infos.push_back(value_info);
 
     // Initialize OrtValue for the initializer.
+    // Note: using std::unique_ptr<OrtValue> because we return a OrtValue* to the user and we want it to be stable.
     auto initializer_value = std::make_unique<OrtValue>();
     bool graph_has_ortvalue = graph_viewer.GetGraph().GetOrtValueInitializer(initializer_name, *initializer_value,
                                                                              /*check_outer_scope*/ false);
@@ -656,8 +657,10 @@ Status EpGraph::CreateImpl(std::unique_ptr<EpGraph> ep_graph, const GraphViewer&
       }
 
       EpValueInfo* outer_value_info = value_info_iter->second.get();
-      bool is_constant = false;
+
+      // Note: using std::unique_ptr<OrtValue> because we return a OrtValue* to the user and we want it to be stable.
       auto outer_initializer_value = std::make_unique<OrtValue>();
+      bool is_constant = false;
       const ONNX_NAMESPACE::TensorProto* outer_initializer = parent_graph->GetInitializer(implicit_name,
                                                                                           *outer_initializer_value,
                                                                                           is_constant,
--- a/onnxruntime/core/graph/ep_api_types.h
+++ b/onnxruntime/core/graph/ep_api_types.h
@@ -348,7 +348,7 @@ struct EpGraph : public OrtGraph {
   // Considers both constant and non-constant initializers.
   // Supports initializers defined in an outer scope as long as that initializer is used
   // within this graph.
-  const OrtValue* GetInitializerValue(std::string_view name) const;
+  Status GetInitializerValue(std::string_view name, const OrtValue*& value) const;
 
  private:
   /// <summary>
--- a/onnxruntime/core/graph/graph.cc
+++ b/onnxruntime/core/graph/graph.cc
@@ -1355,6 +1355,15 @@ Graph::Graph(const Model& owning_model,
       if (is_sparse) {
         sparse_tensor_names_.emplace(tensor.name());
       }
+    } else {
+      std::basic_string<ORTCHAR_T> tensor_proto_dir;
+      ORT_THROW_IF_ERROR(GetDirNameFromFilePath(model_path.c_str(), tensor_proto_dir));
+
+      ExternalInitializerInfo ext_info = {};
+      ORT_THROW_IF_ERROR(utils::GetExternalDataInfo(tensor, tensor_proto_dir, ext_info.file_path, ext_info.file_offset,
+                                                    ext_info.tensor_byte_size));
+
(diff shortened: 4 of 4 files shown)


onnxruntime/core/graph/graph.cc
```diff
@@ -3813,6 +3822,69 @@ bool Graph::GetOrtValueInitializer(const std::string& name, OrtValue& value, boo
...
+    if (size_in_bytes > utils::kSmallTensorExternalDataThreshold) {
+      ORT_RETURN_IF_ERROR(utils::TensorProtoToOrtValue(Env::Default(), ModelPath(), tensor_proto,
+                                                       CPUAllocator::DefaultInstance(), value));
+      constexpr const bool use_tensor_buffer_true = true;
+      auto tensor_proto_to_add = utils::TensorToTensorProto(value.Get<Tensor>(), tensor_proto.name(),
+                                                            use_tensor_buffer_true);
+      assert(value.IsAllocated());
+      ORT_RETURN_IF_ERROR(ReplaceInitializedTensor(tensor_proto_to_add, value));
+    }
+  }
```
yuslepukhin: So EP would not gain access to small weights? #Resolved

adrianlizarraga: Haven't done it yet. Will do after adding new APIs and testing. I'll add a todo in the meantime.

adrianlizarraga: Modified to only load OrtValue for external initializers

include/onnxruntime/core/providers/utils/ort_graph_to_proto.h
```diff
@@ -456,11 +456,14 @@ Ort::Status OrtGraphToProto(const OrtGraph& ort_graph,
...
       onnx::StringStringEntryProto* location_entry = ext_data_entries->Add();
       onnx::StringStringEntryProto* offset_entry = ext_data_entries->Add();
+      onnx::StringStringEntryProto* length_entry = ext_data_entries->Add();
 
       location_entry->set_key("location");
       location_entry->set_value(ext_location);
       offset_entry->set_key("offset");
       offset_entry->set_value(std::to_string(ext_offset));
+      length_entry->set_key("length");
+      length_entry->set_value(std::to_string(data_bytes));
```
yuslepukhin, line 507: Would it be possible to use ExternalData utility function to do just this to avoid code repetition and potential bugs? #Resolved

onnxruntime/core/graph/ep_api_types.h
```diff
@@ -348,7 +348,7 @@ struct EpGraph : public OrtGraph {
   // Considers both constant and non-constant initializers.
   // Supports initializers defined in an outer scope as long as that initializer is used
```
yuslepukhin, line 353: >// Supports initializers defined  [](http://example.com/codeflow?start=2&length=33)

I am seeing a potential issue here.
If the initializer's OrtValue is not added in the local scope for some reason (too small, bug, etc), it does not mean the outer scope would be a correct value. #Resolved

onnxruntime/core/graph/ep_api_types.cc
```diff
@@ -593,15 +593,22 @@ Status EpGraph::CreateImpl(std::unique_ptr<EpGraph> ep_graph, const GraphViewer&
...
     auto initializer_value = std::make_unique<OrtValue>();
     bool graph_has_ortvalue = graph_viewer.GetGraph().GetOrtValueInitializer(initializer_name, *initializer_value,
                                                                              /*check_outer_scope*/ false);
 
     if (!graph_has_ortvalue) {
-      // onnxruntime::Graph does not have an OrtValue for this initializer, so create one from the TensorProto.
-      // This should only happen for small initializers that are needed for ONNX shape inferencing.
-      ORT_RETURN_IF_ERROR(utils::TensorProtoToOrtValue(Env::Default(), graph_viewer.ModelPath(), *tensor_proto,
-                                                       initializer_allocator, *initializer_value));
+      ExternalInitializerInfo ext_info = {};
```
yuslepukhin: >ExternalInitializerInfo ext_info = {}; [](http://example.com/codeflow?start=6&length=38)

Can we employ our utils::HasExternalData() here for quick checks? #Resolved

onnxruntime/core/graph/graph.cc
```diff
@@ -3813,6 +3823,63 @@ bool Graph::GetOrtValueInitializer(const std::string& name, OrtValue& value, boo
   return false;
 }
 
+Status Graph::LoadExternalInitializerAsOrtValue(const std::string& name, OrtValue& value) {
+  auto tensor_proto_iter = name_to_initial_tensor_.find(name);
+  ORT_RETURN_IF(tensor_proto_iter == name_to_initial_tensor_.end(), "Cannot load unknown initializer named '",
+                name, "'.");
+  const ONNX_NAMESPACE::TensorProto& tensor_proto = *tensor_proto_iter->second;
+
+  auto ext_info_iter = ext_initializers_.find(name);
```
yuslepukhin: >ext_initializers_ [](http://example.com/codeflow?start=23&length=17)

TODO: We have lots of structures to keep in sync. Need to cut down on it.

Perhaps, we can simply use utils::HasExternalData()
Also should we assert that this is not an external data in memory here to double check since we have some many places to keep structures in sync.

Also, do we want to check ReplaceInitialzier() functionality?

onnxruntime/core/graph/graph.cc
```diff
@@ -3813,6 +3823,63 @@ bool Graph::GetOrtValueInitializer(const std::string& name, OrtValue& value, boo
...
+
+  // Replace old initializer's tensor_proto and OrtValue.
+  auto& mutable_initializers = *(graph_proto_->mutable_initializer());
+  auto existing_entry = std::find(mutable_initializers.pointer_begin(), mutable_initializers.pointer_end(),
+                                  &tensor_proto);
+  ORT_RETURN_IF_NOT(existing_entry != mutable_initializers.pointer_end(),
+                    "graph_proto_ is not in sync with name_to_initial_tensor_");
+
+  **existing_entry = std::move(tensor_proto_to_add);
+  ORT_IGNORE_RETURN_VALUE(ortvalue_initializers_.emplace(name, value));
```
yuslepukhin: >value [](http://example.com/codeflow?start=63&length=5)

std::move() #Resolved

include/onnxruntime/core/graph/graph.h
```diff
@@ -1801,6 +1831,8 @@ class Graph {  // NOLINT(clang-analyzer-optin.performance.Padding): preserve exi
   // in the Graph instance and retrieve during session state finalization.
   std::unordered_map<std::string, OrtValue> ortvalue_initializers_;
 
+  std::unordered_map<std::string, ExternalInitializerInfo> ext_initializers_;
```
yuslepukhin: Just a thought. Is this going to be used all the time?
It is populated all the time.
We could do the checks on the fly when the external data is requested and generate that information just before returning it one by one instead of just collecting it at load time (which BTW is a concern now) and consuming memory which may not be needed at the end.

We could cut down on one structure.

include/onnxruntime/core/providers/utils/ort_graph_to_proto.h
adrianlizarraga: Not for this file. This file is meant to be a header-only reference implementation that can be copied into an EPs code repository. This file is not used by ORT core code (except for some unit tests).

onnxruntime/core/graph/ep_api_types.h
adrianlizarraga: I don't believe this should be a problem here. For small initializers (or any initializers that don't have an OrtValue), we are always explicitly creating new OrtValues for them here (in ep_api_types.cc):

https://github.com/microsoft/onnxruntime/blob/024ae9e54734eb0ad9ca092c70229470c8e2e4c7/onnxruntime/core/graph/ep_api_types.cc#L654-L675

onnxruntime/core/graph/ep_api_types.cc
adrianlizarraga: Yep, good call.

onnxruntime/core/graph/graph.cc
adrianlizarraga: can't `std::move` because `value` is a function parameter (we return the OrtValue to caller if already exists).

include/onnxruntime/core/graph/graph.h
adrianlizarraga: I'll give it a shot.

onnxruntime/core/graph/ep_api_types.cc
adrianlizarraga: Updated to use the new `utils::HasExternalDataInFile()`

onnxruntime/core/graph/graph.cc
adrianlizarraga: >Perhaps, we can simply use utils::HasExternalData()

Yeah, updated to use the new `utils::HasExternalDataInFile()` instead of checking the map.

>Also should we assert that this is not an external data in memory here to double check since we have some many places to keep structures in sync.

Use of `utils::HasExternalDataInFile()` takes care of this now.

>Also, do we want to check ReplaceInitialzier() functionality?

Thanks for catching this. I added code in `ReplaceInitializedTensorImpl()` to remove an entry from the map if it is being replaced.

include/onnxruntime/core/graph/graph.h
adrianlizarraga: I updated this branch to now only add elements to `external_data_infos_` (renamed) when needed. This happens in two places:
- `Graph::GetExternalInitializerInfo`: adds an entry if it doesn't exist.
- `Graph::LoadExternalInitializerAsOrtValue`: adds an entry for the initializer before loading it into an OrtValue, which overwrites an TensorProto.external_data() information.

Note that this change requires violating const-correctness because `GetExternalInitializerInfo` is called with a const GraphViewer/Graph. If this is not acceptable, then I'll have to revert to computing all external data info on construction of Graph.

include/onnxruntime/core/graph/graph.h
```diff
@@ -25,6 +25,7 @@
 #include "core/common/inlined_containers.h"
 #endif
 #include "core/common/span_utils.h"
+#include "core/common/safeint.h"
```
skottmckay: nit: do we need to add this here? may add a new dependency for consumers of this header and I didn't see any usage of SafeInt in the changes to this file.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -6069,6 +6089,42 @@ struct OrtApi {
 
   /// @}
 
+  /// \name OrtExternalInitializerInfo
```
skottmckay, line 6094: Given this is an opaque type can we do something to handle the special 'in-memory' external data more nicely? External user shouldn't need to be parsing for magic tags to indicate that.

onnxruntime/core/graph/graph.cc
```diff
@@ -3652,8 +3653,10 @@ Status Graph::ReplaceInitializedTensorImpl(ONNX_NAMESPACE::TensorProto new_initi
 
   // New initializers data generally are within OrtValues
   // Small initializers are still stored inside TensorProto
-  ORT_RETURN_IF_NOT(utils::HasExternalDataInMemory(new_initializer) || !ort_value.IsAllocated(),
-                    "All TensorProtos are expected to point to an OrtValue");
+  ORT_RETURN_IF_NOT(utils::HasExternalDataInMemory(new_initializer) ||
+                        (!ort_value.IsAllocated() && !utils::HasExternalData(new_initializer)),
```
skottmckay: This explanation isn't quite clear to me but maybe it's all the negations in the condition between the IF_NOT and the `!`s. #Closed

javier-intel (approved):

onnxruntime/core/graph/graph.cc
```diff
@@ -3815,6 +3823,77 @@ bool Graph::GetOrtValueInitializer(const std::string& name, OrtValue& value, boo
   return false;
 }
 
+Status Graph::LoadExternalInitializerAsOrtValue(const std::string& name, OrtValue& value) {
+  auto tensor_proto_iter = name_to_initial_tensor_.find(name);
+  ORT_RETURN_IF(tensor_proto_iter == name_to_initial_tensor_.end(), "Cannot load unknown initializer named '",
+                name, "'.");
+  const ONNX_NAMESPACE::TensorProto& tensor_proto = *tensor_proto_iter->second;
+
+  ORT_RETURN_IF(!utils::HasExternalDataInFile(tensor_proto), "Initializer '", name, "' does not have external data.");
```
skottmckay: If someone calls this and there's in-memory external data should that be an error? 

Should we do this after checking if there's already a value in `ortvalue_initializers_`? #Closed

onnxruntime/core/graph/graph.cc
```diff
@@ -3815,6 +3823,77 @@ bool Graph::GetOrtValueInitializer(const std::string& name, OrtValue& value, boo
...
+
+  // Create OrtValue that memory maps external initializer from file.
+  ORT_RETURN_IF_ERROR(utils::GetExtDataFromTensorProto(Env::Default(), ModelPath(), tensor_proto, value));
+  ONNX_NAMESPACE::TensorProto tensor_proto_to_add = utils::TensorToTensorProto(value.Get<Tensor>(),
+                                                                               tensor_proto.name(),
+                                                                               /*use_tensor_buffer*/ true);
+  assert(value.IsAllocated());
+
+  // Since we're about overwrite the TensorProto's external_data information (file path, offset, etc.),
+  // need to store it so that is available for callers of Graph::GetExternalInitializerInfo.
```
skottmckay:
```suggestion
  // Since we're about to overwrite the TensorProto's external_data information (file path, offset, etc.),
  // need to store it so that it is available for callers of Graph::GetExternalInitializerInfo.
```

onnxruntime/core/session/onnxruntime_c_api.cc
```diff
@@ -2514,6 +2514,29 @@ ORT_API_STATUS_IMPL(OrtApis::ValueInfo_GetInitializerValue, _In_ const OrtValueI
   API_IMPL_END
 }
 
+ORT_API_STATUS_IMPL(OrtApis::ValueInfo_GetExternalInitializerInfo, _In_ const OrtValueInfo* value_info,
+                    _Outptr_result_maybenull_ const OrtExternalInitializerInfo** info) {
+  API_IMPL_BEGIN
+  const onnxruntime::ExternalDataInfo* ext_data_info = nullptr;
+  ORT_API_RETURN_IF_STATUS_NOT_OK(value_info->GetExternalInitializerInfo(ext_data_info));
+
+  *info = ext_data_info != nullptr ? reinterpret_cast<const OrtExternalInitializerInfo*>(ext_data_info) : nullptr;
```
skottmckay: nit: if you add a declaration in the .cc for the opaque type to connect it to the internal type we can do a static_cast instead.

`struct OrtExternalInitializerInfo : onnxruntime::ExternalDataInfo {};`

include/onnxruntime/core/session/onnxruntime_c_api.h
adrianlizarraga: The public API isn't exposing the 'in-memory' external data trick we use internally within ORT. plugin EPs don't need to know about that. Is there some code that does?

include/onnxruntime/core/graph/graph.h
adrianlizarraga: ah this was a leftover of a previous commit. I'll remove.

adrianlizarraga:
```suggestion
```

onnxruntime/core/graph/graph.cc
adrianlizarraga: thanks, good catch.

include/onnxruntime/core/graph/graph.h
adrianlizarraga: removed

onnxruntime/core/graph/graph.cc
adrianlizarraga: updated. hopefully it's a bit clearer.

onnxruntime/core/session/onnxruntime_c_api.cc
adrianlizarraga: updated.

include/onnxruntime/core/session/onnxruntime_c_api.h
adrianlizarraga: To clarify a bit more: this public API always returns real external file information. Even if ORT has loaded the external data (mmap), we'll return the original external file information.

onnxruntime/core/graph/graph.cc
```diff
@@ -3815,6 +3824,78 @@ bool Graph::GetOrtValueInitializer(const std::string& name, OrtValue& value, boo
...
+  // Create OrtValue that memory maps external initializer from file.
+  ORT_RETURN_IF_ERROR(utils::GetExtDataFromTensorProto(Env::Default(), ModelPath(), tensor_proto, value));
+  ONNX_NAMESPACE::TensorProto tensor_proto_to_add = utils::TensorToTensorProto(value.Get<Tensor>(),
+                                                                               tensor_proto.name(),
+                                                                               /*use_tensor_buffer*/ true);
+  assert(value.IsAllocated());
+
+  // Since we're about to overwrite the TensorProto's external_data information (file path, offset, etc.),
+  // need to store it so that it is available for callers of Graph::GetExternalInitializerInfo.
+  if (external_data_infos_.count(name) == 0) {
```
yuslepukhin: The thought I had was to get rid of the additional structure within the Graph, but return each ExternalDataInfo instance separately without storing it inside the Graph because it is only needed for the provider.

adrianlizarraga: Loading an external initializer into an OrtValue overwrites the information about which file the data comes from. Because EPs need to be able to query this file information at any time (before or after loading the OrtValue), I don't see a way to track this without this additional data structure in Graph.

adrianlizarraga: To give a more concrete example of use within EP:

Suppose we have an external initializer with data in the file "data.bin". If EP makes the following two API calls in order:
1. EP calls `ValueInfo_GetInitializerValue`: ORT returns an OrtValue with the external data memory-mapped. This normally overwrites TensorProto's original location info.
2. Ep calls `ValueInfo_GetExternalInitializerInfo`: ORT _should_ return that the initializer is stored in "data.bin" even though ORT has already memory-mapped it into an OrtValue. Thus, we need to save the information (filepath, offset) in step 1.

yuslepukhin: I am not convinced we need to save it.
As far as I can see it the following are the cases:
1) The OrtValue already exists: return it
2) No OrtValue:
  a) The value is was too small to convert it on Graph load. Create it now and return, but do not save to ortvalue_initializers and does not modify TensorProto.
  b) Data is on disk two cases:
     - the customer does not need custom logic and does not need external data info. It relies on this function or  and additional entry point to do the loading
     - the customer needs external data info. Use a separate method to create ExternalDataInfo and return to the customer. This is called lazy loading.

adrianlizarraga: I agree on cases 1 and 2a, which we already do in graph/ep_graph_api.cc while creating OrtGraph.

2b is the problem:
>b) Data is on disk two cases:
the customer does not need custom logic and does not need external data info. It relies on this function or and additional entry point to do the loading
the customer needs external data info. Use a separate method to create ExternalDataInfo and return to the customer. This is called lazy loading.

If we do this, then the result of calling `ValueInfo_GetInitializerValue` and `ValueInfo_GetExternalInitializerInfo` will _change_ depending on what _order_ they are called in. That is, if you call `ValueInfo_GetExternalInitializerInfo` second, it will not return the original file path. If you call it first, it will.

This is a violation of the const contract of these API functions. Both API functions operate on a const Graph. The result of the function calls should not change depending on the order they are called.

adrianlizarraga: we spoke offline and agreed to not cache OrtValues for external initializers in the Graph. This solves the above const/order issue but the downside is that every call to `Graph::LoadExternalInitiazerAsOrtValue` memory maps the initializer data. We would need to document this.

adrianlizarraga: Made the change.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -5483,10 +5484,13 @@ struct OrtApi {
...
+   * the initializer data on the first call to this function. If caller needs custom memory mapping,
+   * use ValueInfo_GetExternalInitializerInfo to get the location of the initializer data.
+   *
    * \param[in] value_info The OrtValueInfo instance.
-   * \param[out] initializer_value Output parameter set to the initializer value or NULL. The OrtValue data pointer
-   *                               (obtained via GetTensorData) is stable during the lifetime of the OrtSession
-   *                               that owns the OrtGraph.
+   * \param[out] initializer_value Output parameter set to the initializer value or NULL. Do not cache the OrtValue
+   *                               as it is released when the owning OrtGraph is released.
    *
```
yuslepukhin, line 5494: I thought we wanted EPs to cache it. OrtValues are shared Ptrs so why not to cache them if they needed?
We would need a way to clone it though, but I can see it come useful at some point.

adrianlizarraga: Because of this recent change to not cache the `OrtValue` (for external) in `Graph`, the memory mapped data will be released when the `OrtGraph` is released (after GetCapability and after Compile). That's another downside of not caching it in Graph.

onnxruntime/test/ep_graph/test_ep_graph_utils.h
```diff
@@ -42,6 +42,30 @@ struct NodeArgConsumer {
   int64_t input_index = -1;
 };
 
+// Helper to release Ort one or more objects obtained from the public C API at the end of their scope.
+template <typename T>
```
yuslepukhin, line 46: Looks like the absent C++ API just showed up :)

yuslepukhin (approved):
:shipit:

include/onnxruntime/core/session/onnxruntime_c_api.h
skottmckay: Was thinking about the model editor usage where we get pre-allocated OrtValue instances from WebNN, create the model on the fly and run it. Potentially it never goes to disk. Is that hidden from the EP via the OrtValue they get or would it still show up as having external data?

adrianlizarraga: In that case:
- calling `ValueInfo_GetInitializerValue` returns an OrtValue, which is agnostic of origin. There are no API functions off OrtValue that allow querying whether this is an "external initializer".
- calling `ValueInfo_GetExternalInitializerInfo` returns a NULL `OrtExternalInitializerInfo` if the tensor proto did not have file location info, which is probably the case here.

snnn:

PR #25484, opened 2025-07-22
[WebNN] Respect to the shape of zero_point for ConvInteger

Honry:
WebNN requires the shapes of zeroPoint and scale for a qdq op to be same. However the ONNX allows [1] as scalar shape and some models may use [1] as the shape for x_zero_point. We should explicitly set the shape of scale to x_zero_point.


Honry:

onnxruntime/core/providers/webnn/builders/impl/conv_op_builder.cc
```diff
@@ -317,32 +317,34 @@ Status ConvOpBuilder::AddToModelBuilderImpl(ModelBuilder& model_builder, const N
...
+      ORT_RETURN_IF_NOT(GetShape(*input_defs[2], x_zero_point_shape, logger), "Cannot get shape of x_zero_point");
     } else {
       x_zero_point = model_builder.CreateOrGetConstant<uint8_t>(x_type, 0);
     }
-
     // Scale is not used by ConvInteger but required by DequantizeLinear. So set it to default value 1.0f.
     // The x_zero_point must be a scalar and the scale input should have the same shape as the zero point input.
     // So the x_scale must be a scalar too.
-    x_scale = model_builder.CreateOrGetConstant<float>(ONNX_NAMESPACE::TensorProto_DataType_FLOAT, 1.0f);
+    // ONNX allows [1] as scalar, explicitly set the shape of x_scale to x_zero_point_shape.
```
fdwr: What do you mean by `[1]`? Do you mean an array index, or data list like `[1.0]`?

```suggestion
ONNX allows [1.0] as scalar. So explicitly set the shape of x_scale to x_zero_point_shape.
```

fdwr (commented):
One comment else LGTM.

onnxruntime/core/providers/webnn/builders/impl/conv_op_builder.cc
Honry: I mean the scalar can be 1D shape as `[1]`. 

How about: 
`ONNX allows 1D tensor of size 1 as scalar. So explicitly set the shape of x_scale to x_zero_point_shape.`

See https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/providers/common.h#L37C1-L47C2. 
```
/**
Returns true if given tensor is a scalar or 1D tensor of size 1
**/
inline bool IsScalarOr1ElementVector(const Tensor* input) {
  if (input->Shape().NumDimensions() == 0 ||
      (input->Shape().NumDimensions() == 1 && input->Shape().Size() == 1)) {
    return true;
  } else {
    return false;
  }
}
```

Issue #25486, opened 2025-07-22
[Build] Linux Build failure: "absl/container/inlined_vector.h: No such file or directory"


VishalX:
### Describe the issue

Trying to build with **VitisAI** on Linux, errors out with `absl/container/inlined_vector.h: No such file or directory`

### Target platform

Linux

### Build script

Command to build:

> /usr/bin/python tools/ci_build/build.py --use_vitisai --build_shared_lib --skip_submodule_sync --enable_pybind --build_wheel --config Release --parallel 8 --build_dir /workspace/vai-rt-build/build/onnxruntime --skip_tests --use_binskim_compliant_compile_flags --compile_no_warning_as_error --cmake_extra_defines onnxruntime_USE_NEURAL_SPEED=OFF CMAKE_INSTALL_PREFIX=/workspace/vai-rt-build/install CMAKE_EXPORT_COMPILE_COMMANDS=ON CMAKE_TLS_VERIFY=FALSE BUILD_ONNX_PYTHON=ON onnxruntime_DONT_VECTORIZE=0 CMAKE_PREFIX_PATH=/workspace/vai-rt-build/install

### Error / output

Error:

Generating symbol file for ['cpu', 'vitisai']
VERSION:1.23.0
[  4%] Building C object _deps/pytorch_cpuinfo-build/CMakeFiles/cpuinfo.dir/src/log.c.o
[  4%] Building CXX object _deps/protobuf-build/CMakeFiles/libprotobuf.dir/src/google/protobuf/arenastring.cc.o
[  4%] Built target onnxruntime_generate_def
[  4%] Building CXX object CMakeFiles/onnxruntime_mocked_allocator.dir/workspace/vai-rt-build/src/onnxruntime/onnxruntime/test/util/test_allocator.cc.o
[  4%] Building C object _deps/pytorch_cpuinfo-build/CMakeFiles/cpuinfo.dir/src/x86/init.c.o
In file included from /workspace/vai-rt-build/src/onnxruntime/onnxruntime/core/providers/shared_library/provider_api.h:26,
                 from /workspace/vai-rt-build/src/onnxruntime/onnxruntime/core/providers/shared_library/provider_bridge_provider.cc:6:
/workspace/vai-rt-build/src/onnxruntime/include/onnxruntime/core/common/inlined_containers_fwd.h:25:10: fatal error: absl/container/inlined_vector.h: No such file or directory
   25 | #include <absl/container/inlined_vector.h>
[...]

PR #25487, opened 2025-07-22
[VitisAI] Fix Linux build can't find abseil header

mingyueliuh:
### Description
The build error started after ONNXRuntime update abseil (#25033). 
The EP needs to explicitly add the abseil target to its link dependencies.

### Motivation and Context
For fix the issue #25486


mingyueliuh:
@HectorSVC  please take a look when you are available. Thanks.

snnn:
So, every provider EP dll needs to have its own copy of abseil? It really surprised me.

PR #25187, opened 2025-06-26
KleidiAI SGEMM/IGEMM/Quantized MatMul - Modular MLAS API Changes for KleidiAI

damdoo01-arm:
This PR introduces the initial integration of KleidiAI-optimized microkernels into ONNX Runtime's MLAS backend, focusing on support for:

- SGEMM
- IGEMM
- Dynamic Quantized MatMuls

Key changes:
Implements overrides for MlasGemmBatch, MlasGemmPackBSize, and MlasGemmPackB using KleidiAI where applicable.
Applies dispatch logic based on TransA == CblasNoTrans and SME2 availability.
Supports float32 and int8 GEMM workloads with conditionally invoked SME2 paths.
Maintains fallback paths to default MLAS implementations to ensure coverage and stability.

**Known Issues / Next Steps:**
Requesting feedback specifically on the API structure:
Does the new MLAS interface design align with long-term extensibility?
Are the dispatch points and override boundaries well-structured?

Indicative Performance figures:
The kernels added are particularly effective for Conv2D operators:
* Based on KleidiAI SME running mobilenet_v1_ssd_f32 on Mac Mini M4 on a single thread
image

PR #25488, opened 2025-07-22
Correct tvOS and visionOS build flag

hans00:
### Description

Correct build flag for tvOS and visionOS


snnn:
Would you please explain why this change is needed?

hans00:
> Would you please explain why this change is needed?

Sync build flag with iOS
https://github.com/microsoft/onnxruntime/blob/024ae9e54734eb0ad9ca092c70229470c8e2e4c7/tools/ci_build/github/apple/default_full_ios_framework_build_settings.json#L21

P.S. Currently without disable kleidiai will build fail

skottmckay:
Should be fixed by #25187

PR #25509, opened 2025-07-23
Fix race condition in ProviderLibrary

skottmckay:
### Description

Apply bugfix from internal team.


snnn:

PR #25514, opened 2025-07-23
[QNN-EP] Fix data type check to skip optional I/Os

qti-yuduo:
### Motivation and Context
Fix data type check to skip optional I/Os.

Optional inputs/outputs would have empty name, and it's valid ONNX syntax. Without this fix, any model with optional inputs/outputs would fail the check due to missing protobuf fields. 

Without this fix, we'd hit error fetching `elem_type` from protobuf.
```
2025-07-22 11:14:40.117740035 [I:onnxruntime:, qnn_execution_provider.cc:740 GetSupportedNodes] Validation FAILED for 1 nodes in NodeUnit (Pad) :
	Operator type: Pad Node name: /blocks.4/Pad Node index: 176
	REASON : The tensor doesn't have elem_type.
```

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/base_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/base_op_builder.cc
@@ -8,6 +8,13 @@
 namespace onnxruntime {
 namespace qnn {
 
+namespace {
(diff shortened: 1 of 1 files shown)


HectorSVC:
close and reopen to restart the job

PR #25517, opened 2025-07-23
Fix node_helper.cmake for wasm build on Windows

fs-eire:
### Description

When building targeting wasm, `WIN32` is `false`. Use `CMAKE_HOST_WIN32` instead.

### Motivation and Context

Fix the problem when building wasm on Windows, node_helper cannot find NPM correctly.


guschmue (approved):

snnn:

PR #25520, opened 2025-07-24
Fix QNN SDK download problem

snnn:
Previously the machine pool had a User-assigned managed identity (UMI) which was used for accessing the blob storage. Now the UMI was removed. to improve security. Therefore we baked the data into the VM image instead.

### DIFF
--- a/tools/ci_build/github/azure-pipelines/templates/jobs/init_linux_qnn_sdk_x64.yml
+++ b/tools/ci_build/github/azure-pipelines/templates/jobs/init_linux_qnn_sdk_x64.yml
@@ -4,12 +4,8 @@ parameters:
     default: '2.36.1.250708'
 
 steps:
-  - script: |
-      azcopy cp --recursive https://lotusscus.blob.core.windows.net/models/qnnsdk/qnn-v${{ parameters.QnnSDKVersion }} $(Agent.TempDirectory)
-    displayName: 'Download QNN SDK v${{ parameters.QnnSDKVersion }}'
-
   - bash: |
-      echo "##vso[task.setvariable variable=QnnSDKRootDir]$(Agent.TempDirectory)/qnn-v${{ parameters.QnnSDKVersion }}"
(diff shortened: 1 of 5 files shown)


tools/ci_build/github/azure-pipelines/templates/jobs/init_linux_qnn_sdk_x64.yml
```diff
@@ -0,0 +1,42 @@
...
+      # Compare provided version with version from sdk.yaml
+      if [[ "$downloaded_qnn_sdk_version" == "$expected_qnn_sdk_version" ]]; then
+        echo "Success: QnnSDKVersion matches sdk.yaml version ($downloaded_qnn_sdk_version)."
+      else
+        echo "Error: QnnSDKVersion ($expected_qnn_sdk_version) does not match sdk.yaml version ($downloaded_qnn_sdk_version) in the QNN SDK directory"
+        exit 1
+      fi
+    displayName: "Sanity Check: QnnSDKVersion vs sdk.yaml version"
+
+
```
adrianlizarraga, line 38: Looks like this is no longer copying the "Qualcomm AI Hub Proprietary License.pdf" into the QNN SDK folder. 

@jywu-msft I can't recall, but is that still needed?

snnn: Yeah, I removed it because the dest folder actually is readonly(owned by root). We cannot copy files there. If the file is still needed, we can put it in the blob storage folder

adrianlizarraga: Actually, we would not need to use the "Qualcomm AI Hub Proprietary License.pdf" license file if we use the license already within QNN SDK.

Would need to merge this PR though: https://github.com/microsoft/onnxruntime/pull/25158

adrianlizarraga (approved):

PR #25521, opened 2025-07-24
[TRT RTX EP] Add support for user provided external initializers with new TRT APIs

thevishalagarwal:
This PR adds support for user provided external initializers which are already in memory, instead of loading from disk. 

Adds 3 new EP options:
* `nv_use_external_data_initializer` : when enabled, EP will own the weights in memory, instead of serializing it to ModelProto. This helps to run models >2GB
* `nv_external_data_bytestream` : bytestream for external data
* `nv_external_data_bytestream_size` : bytestream size for external data

These are implemented with new `loadModelProto` and `loadIntializer` APIs available in TRT RTX 1.1 onwards

### DIFF
--- a/include/onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_options.h
+++ b/include/onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_options.h
@@ -35,6 +35,8 @@ constexpr const char* kCudaGraphEnable = "nv_cuda_graph_enable";
 constexpr const char* kONNXBytestream = "nv_onnx_bytestream";
 constexpr const char* kONNXBytestreamSize = "nv_onnx_bytestream_size";
 constexpr const char* kMultiProfileEnable = "nv_multi_profile_enable";
+constexpr const char* kExternalDataBytestream = "nv_external_data_bytestream";
+constexpr const char* kExternalDataBytestreamSize = "nv_external_data_bytestream_size";
 
 }  // namespace provider_option_names
 namespace run_option_names {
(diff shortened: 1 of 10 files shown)


include/onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_options.h
```diff
@@ -35,6 +35,9 @@ constexpr const char* kCudaGraphEnable = "nv_cuda_graph_enable";
 constexpr const char* kONNXBytestream = "nv_onnx_bytestream";
 constexpr const char* kONNXBytestreamSize = "nv_onnx_bytestream_size";
 constexpr const char* kMultiProfileEnable = "nv_multi_profile_enable";
+constexpr const char* kUseExternalDataInitializer = "nv_use_external_data_initializer";
+constexpr const char* kExternalDataBytestream = "nv_external_data_bytestream";
+constexpr const char* kExternalDataBytestreamSize = "nv_external_data_bytestream_size";
```
gedoensmax, line 40: DO we really want to expose these options on a new EP ? We want to move to constant initializers as input very soon and I do not know what the deprecation policy on such provider options is for us. 

This is not something that we want to maintain IMHO.

onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
```diff
@@ -1701,7 +1712,28 @@ SubGraphCollection_t NvExecutionProvider::GetSupportedList(SubGraphCollection_t
...
+            auto* tp = entry.second;
+            if (tp->has_raw_data()) {
+              userWeights.emplace_back(tp->name(), tp->raw_data());
+            } else if (utils::HasExternalDataInMemory(*tp)) {
+              std::unique_ptr<ONNX_NAMESPACE::TensorProto> full_init;
+              ORT_THROW_IF_ERROR(utils::GetTensorProtoWithDataIfInMemory(*tp, full_init));
+              userWeights.emplace_back(full_init->name(), full_init->raw_data());
+            }
+          }
+        }
```
gedoensmax, line 1732: How much memory duplication does this trigger ? I was wondering if there is some attribute on the protobuf tensors that allows us a better decision on if we need to copy or if we can save references to a memory area.

gedoensmax (commented):
How does this interact with #24901 ? And which tests will need to be added.

PR #25528, opened 2025-07-24
Split windows_tensorrt.yml to two parts

snnn:
It reduces the pipeline time for about 30 minutes.  The tests still take about 1 hour, which should be reduced.


.github/workflows/windows_tensorrt.yml
```diff
@@ -69,33 +64,201 @@ jobs:
...
+          azcopy.exe cp --recursive "https://lotusscus.blob.core.windows.net/models/cuda_sdk/v12.2" .
+          dir
+        shell: pwsh
+
+      - name: Download TensorRT-10.9.0.34.Windows10.x86_64.cuda-12.8
+        run: 'azcopy.exe cp --recursive "https://lotusscus.blob.core.windows.net/models/local/TensorRT-10.9.0.34.Windows10.x86_64.cuda-12.8" ${{ runner.temp }}'
+        shell: pwsh
+        env:
+          AZCOPY_AUTO_LOGIN_TYPE: MSI
+          AZCOPY_MSI_CLIENT_ID: 63b63039-6328-442f-954b-5a64d124e5b4
```
adrianlizarraga: nit: not needed?

snnn: We still download CUDA SDK from azure blob storage.

adrianlizarraga: Right, but isn't this same info in the `env:` section for the job down below on line 256

adrianlizarraga (approved):

.github/workflows/windows_tensorrt.yml
snnn: Ah! You are right! Thanks for catching this!

PR #25503, opened 2025-07-23
Update OrtEpFactory in new EPs

skottmckay:
### Description

Update OrtEpFactory in new EPs to add allocator, data transfer and stream stubs.

PR #25537, opened 2025-07-25
[VitisAI] bugfix: add empty CreateDataTransferImpl

wcy123:
### Description

bugfix segfault when when registering VitisAI EP.

```python
import onnxruntime as ort
ort.register_execution_provider_library("VitisAIExecutionProvider", "./onnxruntime_providers_vitisai.dll")
```
it crashes at here 

https://github.com/microsoft/onnxruntime/blob/7e59947dcf8f489f02e1d9dfb7ca0f3de0d677d0/onnxruntime/core/session/environment.cc#L515

It is related to #25254

### Implementation

and an empty implementation for `CreateDataTransferImpl`


wcy123:
duplicated with #25503

PR #25544, opened 2025-07-25
DynamicQuantizeMatMul - handle case where B zero point input is provided but not constant.

edgchen1:
### Description

In DynamicQuantizeMatMul KleidiAI-specific prepacking logic, handle case where B zero point input is provided but not constant. In this case, we should not prepack.

Add some unit tests that test the prepacking code path.

Add check for ARM SME instructions in DynamicQuantizeMatMul before calling `MlasDynamicQGemmBatch()` and associated functions.

### Motivation and Context

Follow up to #25187

### DIFF
--- a/.github/workflows/macos-ci-build-and-test-workflow.yml
+++ b/.github/workflows/macos-ci-build-and-test-workflow.yml
@@ -63,10 +63,6 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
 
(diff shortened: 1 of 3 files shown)


hariharans29 (approved):
LGTM

hariharans29 (approved):

damdoo01-arm:
Thanks @edgchen1 for handling this case.

PR #25558, opened 2025-07-28
Asonawane/qmoe

apsonawane:
### Description

Add CPU kernel support for QMoE

### DIFF
--- a/onnxruntime/contrib_ops/cpu/cpu_contrib_kernels.h
+++ b/onnxruntime/contrib_ops/cpu/cpu_contrib_kernels.h
@@ -6,6 +6,9 @@
 #include "core/framework/op_kernel.h"
 #include "core/framework/kernel_registry.h"
 
+// Forward declarations for QMoE
+#include "contrib_ops/cpu/quantization/moe_quantization_cpu.h"
+
 namespace onnxruntime {
 namespace contrib {
 Status RegisterCpuContribKernels(KernelRegistry& kernel_registry);
--- a/onnxruntime/contrib_ops/cpu/moe/moe_base_cpu.h
+++ b/onnxruntime/contrib_ops/cpu/moe/moe_base_cpu.h
@@ -9,7 +9,6 @@
 
 namespace onnxruntime {
 namespace contrib {
-namespace cpu {
 
 enum class MoEParallelType {
   None = 0,
@@ -51,6 +50,7 @@ class MoEBaseCPU {
                      const Tensor* fc1_experts_bias_optional, const Tensor* fc2_experts_weights,
                      const Tensor* fc2_experts_bias_optional, const Tensor* fc3_experts_weights_optional,
                      const Tensor* fc3_experts_bias_optional) const {
+    ORT_UNUSED_PARAMETER(fc3_experts_bias_optional);
     const auto& input_dims = input->Shape().GetDims();
     const auto& router_probs_dims = router_probs->Shape().GetDims();
     const auto& fc1_experts_weights_dims = fc1_experts_weights->Shape().GetDims();
@@ -224,6 +224,5 @@ class MoEBaseCPU {
   ActivationType activation_type_;
 };
 
-}  // namespace cpu
 }  // namespace contrib
 }  // namespace onnxruntime
--- a/onnxruntime/contrib_ops/cpu/moe/moe_base_cpu.h
+++ b/onnxruntime/contrib_ops/cpu/moe/moe_base_cpu.h
@@ -1,0 +1,229 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#pragma once
+
+#include "core/common/common.h"
+#include "core/framework/tensor_shape.h"
+#include "core/framework/op_kernel.h"
+
+namespace onnxruntime {
+namespace contrib {
+namespace cpu {
+
+enum class MoEParallelType {
+  None = 0,
+  EP = 1,
+  TP = 2,
+  EPAndTP = 3,
+};
+
+enum class MoEQuantType {
+  None = 0,
+  UINT4 = 1,
+  UINT8 = 2,
+};
+
+enum class ActivationType {
+  Relu = 0,
+  Gelu = 1,
+  Silu = 2,
+  Identity = 3,
+};
+
+struct MoEParameters {
+  MoEParameters() {}
+  explicit MoEParameters(int64_t tensor_shards) : tensor_shards(tensor_shards) {}
+  int64_t num_rows;
+  int64_t num_experts;
+  int64_t local_num_experts;
+  int64_t hidden_size;
+  int64_t inter_size;
+
+  MoEParallelType parallel_type;
+  int64_t tensor_shards{1};
+};
+
+class MoEBaseCPU {
+ public:
+  Status CheckInputs(MoEParameters& parameters, MoEQuantType& quant_type, const Tensor* input,
+                     const Tensor* router_probs, const Tensor* fc1_experts_weights,
+                     const Tensor* fc1_experts_bias_optional, const Tensor* fc2_experts_weights,
+                     const Tensor* fc2_experts_bias_optional, const Tensor* fc3_experts_weights_optional,
+                     const Tensor* fc3_experts_bias_optional) const {
+    const auto& input_dims = input->Shape().GetDims();
+    const auto& router_probs_dims = router_probs->Shape().GetDims();
+    const auto& fc1_experts_weights_dims = fc1_experts_weights->Shape().GetDims();
+    const auto& fc2_experts_weights_dims = fc2_experts_weights->Shape().GetDims();
+
+    int64_t num_rows = input_dims.size() == 2 ? input_dims[0] : input_dims[0] * input_dims[1];
+    int64_t hidden_size = input_dims[input_dims.size() - 1];
+    int64_t local_num_experts = fc1_experts_weights_dims[0];
+    int64_t num_experts = router_probs_dims[1];
+    int64_t inter_size = fc2_experts_weights_dims[1];
+
+    if (fc1_experts_weights_dims.size() != 3) {
+      return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT, "fc1_experts_weights_dims must be 3D, got ",
+                             fc1_experts_weights_dims.size());
+    }
+    if (fc2_experts_weights_dims.size() != 3) {
+      return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT, "fc2_experts_weights_dims must be 3D, got ",
+                             fc2_experts_weights_dims.size());
+    }
+    if (fc1_experts_weights_dims[1] != hidden_size) {
+      return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT,
+                             "fc1_experts_weights_dims[1] must be equal to hidden_size, got ",
+                             fc1_experts_weights_dims[1], " and ", hidden_size);
+    }
+    if (fc2_experts_weights_dims[1] != inter_size) {
+      return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT,
+                             "fc2_experts_weights_dims[1] must be equal to inter_size, got ",
+                             fc2_experts_weights_dims[1], " and ", inter_size);
+    }
+
+    const int64_t coe = quant_type == MoEQuantType::UINT4 ? 2 : 1;
+    if (fc1_experts_weights_dims[2] != inter_size / coe) {
+      return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT,
+                             "fc1_experts_weights_dims[2] must be equal to inter_size, got ",
+                             fc1_experts_weights_dims[2], " and ", inter_size);
+    }
+    if (fc2_experts_weights_dims[2] != hidden_size / coe) {
+      return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT,
+                             "fc2_experts_weights_dims[2] must be equal to hidden_size, got ",
+                             fc2_experts_weights_dims[2], " and ", hidden_size);
+    }
+
+    if (router_probs_dims.size() != 2) {
+      return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT, "router_probs_dims must be 2D, got ",
+                             router_probs_dims.size());
+    }
+    if (router_probs_dims[0] != num_rows) {
+      return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT, "router_probs_dims[0] must be equal to num_rows, got ",
+                             router_probs_dims[0], " and ", num_rows);
+    }
+
+    // Optional bias validation
+    if (fc1_experts_bias_optional != nullptr && fc2_experts_bias_optional != nullptr) {
+      const auto& fc1_experts_bias_dims = fc1_experts_bias_optional->Shape().GetDims();
+      const auto& fc2_experts_bias_dims = fc2_experts_bias_optional->Shape().GetDims();
+      if (fc1_experts_bias_dims.size() != 2) {
+        return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT, "fc1_experts_bias_dims must be 2D, got ",
+                               fc1_experts_bias_dims.size());
+      }
+      if (fc2_experts_bias_dims.size() != 2) {
+        return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT, "fc2_experts_bias_dims must be 2D, got ",
+                               fc2_experts_bias_dims.size());
+      }
+      if (fc1_experts_bias_dims[0] != local_num_experts) {
+        return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT, "fc1_experts_bias_dims[0] must be equal to local_num_experts, got ",
+                               fc1_experts_bias_dims[0], " and ", local_num_experts);
+      }
+      if (fc2_experts_bias_dims[0] != local_num_experts) {
+        return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT, "fc2_experts_bias_dims[0] must be equal to local_num_experts, got ",
+                               fc2_experts_bias_dims[0], " and ", local_num_experts);
+      }
+      if (fc1_experts_bias_dims[1] != inter_size) {
+        return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT, "fc1_experts_bias_dims[1] must be equal to inter_size, got ",
+                               fc1_experts_bias_dims[1], " and ", inter_size);
+      }
+      if (fc2_experts_bias_dims[1] != hidden_size) {
+        return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT, "fc2_experts_bias_dims[1] must be equal to hidden_size, got ",
+                               fc2_experts_bias_dims[1], " and ", hidden_size);
+      }
+    }
+
+    // Optional fc3 validation - CPU implementation doesn't support FC3 yet
+    if (fc3_experts_weights_optional != nullptr) {
+      return ORT_MAKE_STATUS(ONNXRUNTIME, NOT_IMPLEMENTED,
+                             "FC3 gating is not yet implemented for CPU quantized MoE. "
+                             "Please use the CUDA execution provider for gated experts or disable FC3 gating.");
+    }
+
+    // Set output parameters
+    parameters.num_rows = num_rows;
+    parameters.num_experts = num_experts;
+    parameters.local_num_experts = local_num_experts;
+    parameters.hidden_size = hidden_size;
+    parameters.inter_size = inter_size;
+    parameters.parallel_type = MoEParallelType::None;
+
+    return Status::OK();
+  }
+
+  Status CheckInputScales(const Tensor* fc1_experts_scales, const Tensor* fc2_experts_scales, const Tensor* fc3_experts_scales_optional,
+                          int64_t num_experts, int64_t hidden_size, int64_t inter_size) const {
+    if (fc1_experts_scales == nullptr || fc2_experts_scales == nullptr) {
+      return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT, "fc1_experts_scales and fc2_experts_scales cannot be null for quantized MoE");
+    }
+
+    const auto& fc1_experts_scales_dims = fc1_experts_scales->Shape().GetDims();
+    const auto& fc2_experts_scales_dims = fc2_experts_scales->Shape().GetDims();
+
+    if (fc1_experts_scales_dims.size() != 2) {
+      return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT, "fc1_experts_scales must be 2D, got ",
+                             fc1_experts_scales_dims.size());
+    }
+    if (fc2_experts_scales_dims.size() != 2) {
+      return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT, "fc2_experts_scales must be 2D, got ",
+                             fc2_experts_scales_dims.size());
+    }
+    if (fc1_experts_scales_dims[0] != num_experts) {
+      return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT, "fc1_experts_scales[0] must be equal to num_experts, got ",
+                             fc1_experts_scales_dims[0], " and ", num_experts);
+    }
+    if (fc1_experts_scales_dims[1] != inter_size) {
+      return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT, "fc1_experts_scales[1] must be equal to inter_size, got ",
+                             fc1_experts_scales_dims[1], " and ", inter_size);
+    }
+    if (fc2_experts_scales_dims[0] != num_experts) {
+      return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT, "fc2_experts_scales[0] must be equal to num_experts, got ",
+                             fc2_experts_scales_dims[0], " and ", num_experts);
+    }
+    if (fc2_experts_scales_dims[1] != hidden_size) {
+      return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT, "fc2_experts_scales[1] must be equal to hidden_size, got ",
+                             fc2_experts_scales_dims[1], " and ", hidden_size);
+    }
+    if (fc3_experts_scales_optional != nullptr && fc1_experts_scales_dims != fc3_experts_scales_optional->Shape().GetDims()) {
+      return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT,
+                             "fc3_experts_scales must be equal to fc1_experts_scales, got ",
+                             fc3_experts_scales_optional->Shape(), " and ", TensorShape(fc1_experts_scales_dims));
+    }
+
+    return Status::OK();
+  }
+
+ protected:
+  MoEBaseCPU(const OpKernelInfo& op_kernel_info) {
+    ORT_ENFORCE(op_kernel_info.GetAttr<int64_t>("k", &k_).IsOK());
+
+    std::string activation_type_str;
+    ORT_ENFORCE(op_kernel_info.GetAttr<std::string>("activation_type", &activation_type_str).IsOK());
+    if (activation_type_str == "relu") {
+      activation_type_ = ActivationType::Relu;
+    } else if (activation_type_str == "gelu") {
+      activation_type_ = ActivationType::Gelu;
+    } else if (activation_type_str == "silu") {
+      activation_type_ = ActivationType::Silu;
+    } else if (activation_type_str == "identity") {
+      activation_type_ = ActivationType::Identity;
+    } else {
+      ORT_THROW("Unsupported MoE activation type: ", activation_type_str);
+    }
+
+    normalize_routing_weights_ = op_kernel_info.GetAttrOrDefault<int64_t>("normalize_routing_weights", 0) == 1;
+
+    use_sparse_mixer_ = op_kernel_info.GetAttrOrDefault<int64_t>("use_sparse_mixer", 0) == 1;
+    if (use_sparse_mixer_) {
+      ORT_ENFORCE(k_ == 2, "Sparse mixer only supports k=2");
+    }
+  }
+
+  bool normalize_routing_weights_;
+  bool use_sparse_mixer_;
+  int64_t k_;
+  ActivationType activation_type_;
+};
+
+}  // namespace cpu
+}  // namespace contrib
+}  // namespace onnxruntime
--- a/onnxruntime/contrib_ops/cpu/quantization/moe_quantization_cpu.cc
+++ b/onnxruntime/contrib_ops/cpu/quantization/moe_quantization_cpu.cc
@@ -1,0 +1,119 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include "contrib_ops/cpu/quantization/moe_quantization_cpu.h"
+
+#include <algorithm>
+#include <vector>
+#include <cmath>
+
+#include "core/common/safeint.h"
+#include "core/framework/float16.h"
+#include "core/framework/int4.h"
+
+using namespace onnxruntime::common;
+using namespace ONNX_NAMESPACE;
+
+namespace onnxruntime {
+namespace contrib {
+namespace cpu {
+
+QMoE::QMoE(const OpKernelInfo& op_kernel_info) : OpKernel(op_kernel_info), MoEBaseCPU(op_kernel_info) {
+  ORT_ENFORCE(op_kernel_info.GetAttr<int64_t>("expert_weight_bits", &expert_weight_bits_).IsOK());
+  ORT_ENFORCE(expert_weight_bits_ == 8 || expert_weight_bits_ == 4,
+              "expert_weight_bits must be 4 or 8, but got ", expert_weight_bits_);
+}
+
+Status QMoE::QuantizedMoEImpl_UInt4x2(OpKernelContext* context,
+                                      MoEParameters& moe_params,
+                                      const Tensor* input,
+                                      const Tensor* router_probs,
+                                      const Tensor* fc1_experts_weights,
+                                      const Tensor* fc1_experts_bias_optional,
+                                      const Tensor* fc2_experts_weights,
+                                      const Tensor* fc2_experts_bias_optional,
+                                      const Tensor* fc3_experts_weights_optional,
+                                      const Tensor* fc3_experts_bias_optional,
+                                      const Tensor* fc1_scales,
+                                      const Tensor* fc2_scales,
+                                      const Tensor* fc3_scales_optional) const {
+  // FC3 (gating) check - throw error if present
+  if (fc3_experts_weights_optional != nullptr) {
+    ORT_THROW("FC3 gating is not yet implemented for CPU quantized MoE. Please use the CUDA version for gated experts or disable FC3 gating.");
+  }
+
+  // Create output tensor
+  Tensor* output = context->Output(0, input->Shape());
+  MLFloat16* output_data = output->MutableData<MLFloat16>();
+
+  // Simple implementation: just return zeros for now
+  std::fill(output_data, output_data + moe_params.num_rows * moe_params.hidden_size, MLFloat16(0.0f));
+
+  return Status::OK();
+}
+
+Status QMoE::QuantizedMoEImpl_UInt8(OpKernelContext* context,
+                                    MoEParameters& moe_params,
+                                    const Tensor* input,
+                                    const Tensor* router_probs,
+                                    const Tensor* fc1_experts_weights,
+                                    const Tensor* fc1_experts_bias_optional,
+                                    const Tensor* fc2_experts_weights,
+                                    const Tensor* fc2_experts_bias_optional,
+                                    const Tensor* fc3_experts_weights_optional,
+                                    const Tensor* fc3_experts_bias_optional,
+                                    const Tensor* fc1_scales,
+                                    const Tensor* fc2_scales,
+                                    const Tensor* fc3_scales_optional) const {
+  // FC3 (gating) check - throw error if present
+  if (fc3_experts_weights_optional != nullptr) {
+    ORT_THROW("FC3 gating is not yet implemented for CPU quantized MoE. Please use the CUDA version for gated experts or disable FC3 gating.");
+  }
+
+  // Create output tensor
+  Tensor* output = context->Output(0, input->Shape());
+  MLFloat16* output_data = output->MutableData<MLFloat16>();
+
+  // Simple implementation: just return zeros for now
+  std::fill(output_data, output_data + moe_params.num_rows * moe_params.hidden_size, MLFloat16(0.0f));
+
+  return Status::OK();
+}
+
+Status QMoE::Compute(OpKernelContext* context) const {
+  const Tensor* input = context->Input<Tensor>(0);
+  const Tensor* router_probs = context->Input<Tensor>(1);
+  const Tensor* fc1_experts_weights = context->Input<Tensor>(2);
+  const Tensor* fc1_scales = context->Input<Tensor>(3);
+  const Tensor* fc1_experts_bias_optional = context->Input<Tensor>(4);
+  const Tensor* fc2_experts_weights = context->Input<Tensor>(5);
+  const Tensor* fc2_scales = context->Input<Tensor>(6);
+  const Tensor* fc2_experts_bias_optional = context->Input<Tensor>(7);
+  const Tensor* fc3_experts_weights_optional = context->Input<Tensor>(8);
+  const Tensor* fc3_scales_optional = context->Input<Tensor>(9);
+  const Tensor* fc3_experts_bias_optional = context->Input<Tensor>(10);
+
+  MoEQuantType quant_type = expert_weight_bits_ == 4 ? MoEQuantType::UINT4 : MoEQuantType::UINT8;
+  MoEParameters moe_params;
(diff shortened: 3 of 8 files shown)


onnxruntime/test/contrib_ops/moe_test.cc
```diff
@@ -19,7 +19,7 @@ static void RunMoETest(const std::vector<float>& input, const std::vector<float>
   constexpr int min_cuda_arch = 700;
   constexpr int max_cuda_arch = 900;
 
-  bool enable_cuda = HasCudaEnvironment(min_cuda_arch) && !NeedSkipIfCudaArchGreaterEqualThan(max_cuda_arch);
+  bool enable_cuda = HasCudaEnvironment(min_cuda_arch) && NeedSkipIfCudaArchGreaterEqualThan(max_cuda_arch);
```
tianleiwu: bool enable_cuda = HasCudaEnvironment(min_cuda_arch)

Also other places

onnxruntime/contrib_ops/cpu/moe/moe_base_cpu.h
```diff
@@ -132,11 +135,14 @@ class MoEBaseCPU {
       }
     }
 
-    // Optional fc3 validation - CPU implementation doesn't support FC3 yet
-    if (fc3_experts_weights_optional != nullptr) {
+    // FC3 validation - match CUDA FasterTransformer behavior
+    if (activation_type_ == ActivationType::SwiGLU && fc3_experts_weights_optional != nullptr) {
+      return ORT_MAKE_STATUS(ONNXRUNTIME, NOT_IMPLEMENTED,
+                             "SwiGLU activation is not supported with fc3. For SwiGLU, the gate weights should be concatenated with FC1 weights.");
```
tianleiwu: Remove "For SwiGLU, the gate weights should be concatenated with FC1 weights." since it is confusing.

Just mention that SwiGLU does not support FC3 is enough. Also update the other message for other FC3 related inputs.

onnxruntime/contrib_ops/cpu/quantization/moe_quantization_cpu.cc
```diff
@@ -355,28 +394,54 @@ Status QMoE::QuantizedMoEImpl(OpKernelContext* context,
           if (routing_weight <= 1e-6f) continue;  // Skip experts with negligible routing weight
 
           // FC1: input -> intermediate using pre-dequantized weights + MLAS SGEMM
-          const float* fc1_expert_weights = dequant_fc1_weights.get() + expert_idx * moe_params.hidden_size * moe_params.inter_size;
-          const MLFloat16* fc1_expert_bias_typed = fc1_bias_data ? fc1_bias_data + expert_idx * moe_params.inter_size : nullptr;
+          int act = activation_type_ == ActivationType::SwiGLU ? 2 : 1;
+          const float* fc1_expert_weights = dequant_fc1_weights.get() + expert_idx * moe_params.hidden_size * moe_params.inter_size * act;
+          const MLFloat16* fc1_expert_bias_typed = fc1_bias_data ? fc1_bias_data + expert_idx * moe_params.inter_size * act : nullptr;
```
tianleiwu, line 398: why the weights and bias have different types (float and float16)?

onnxruntime/contrib_ops/cpu/quantization/moe_quantization_cpu.cc
```diff
@@ -0,0 +1,530 @@
...
+      const uint8_t* fc1_expert_weights = fc1_weights_data + expert_idx * moe_params.hidden_size * moe_params.inter_size * act;
+      const float* fc1_expert_scales = fc1_scales_data + expert_idx * moe_params.inter_size * act;
+      float* dequant_fc1_expert = dequant_fc1_weights.get() + expert_idx * moe_params.hidden_size * moe_params.inter_size * act;
+
+      for (int64_t out_col = 0; out_col < moe_params.inter_size * act; ++out_col) {
+        for (int64_t in_col = 0; in_col < moe_params.hidden_size; ++in_col) {
+          size_t weight_idx = static_cast<size_t>(out_col * moe_params.hidden_size + in_col);
+          uint8_t quantized_weight = fc1_expert_weights[weight_idx];
+          // Symmetric quantization with zero point = 128
+          dequant_fc1_expert[weight_idx] = (static_cast<float>(quantized_weight) - 128.0f) * fc1_expert_scales[out_col];
```
tianleiwu: Default zero points is different from 4 bits and 8 bits. This assumes 8 bits.

onnxruntime/contrib_ops/cpu/quantization/moe_quantization_cpu.cc
```diff
@@ -0,0 +1,530 @@
...
+      float* thread_fc2_output = thread_fc2_buffers.get() + thread_id * moe_params.hidden_size;
+      float* thread_local_results = thread_results.get() + thread_id * moe_params.num_rows * moe_params.hidden_size;
+
+      // Process each token in this thread's range
+      for (int64_t token_idx = start_token; token_idx < end_token; ++token_idx) {
+        const MLFloat16* token_input_typed = input_data + token_idx * moe_params.hidden_size;
+
+        // Convert input from MLFloat16 to float for MLAS computation
+        std::vector<float> token_input_float(static_cast<size_t>(moe_params.hidden_size));
+        for (int64_t i = 0; i < moe_params.hidden_size; ++i) {
```
tianleiwu: Is there MLAS API to convert fp16 to fp32?

onnxruntime/contrib_ops/cpu/moe/moe_base_cpu.h
apsonawane: Done

onnxruntime/contrib_ops/cpu/quantization/moe_quantization_cpu.cc
```diff
@@ -0,0 +1,530 @@
...
+
+    // Pre-dequantize all expert weights once (shared across all threads)
+    const int64_t fc1_output_size = is_swiglu ? 2 * moe_params.inter_size : moe_params.inter_size;
+    auto dequant_fc1_weights = IAllocator::MakeUniquePtr<float>(allocator,
+                                                                static_cast<size_t>(moe_params.num_experts * moe_params.hidden_size * fc1_output_size));
+    auto dequant_fc2_weights = IAllocator::MakeUniquePtr<float>(allocator,
+                                                                static_cast<size_t>(moe_params.num_experts * moe_params.inter_size * moe_params.hidden_size));
+
+    // Dequantize FC1 weights for all experts (Int4 unpacking)
+    for (int64_t expert_idx = 0; expert_idx < moe_params.num_experts; ++expert_idx) {
```
kunal-vaishnavi: Can we use `ThreadPool::TryParallelFor` here?

onnxruntime/contrib_ops/cpu/quantization/moe_quantization_cpu.cc
```diff
@@ -0,0 +1,530 @@
...
+          }
+
+          // Dequantize from 4-bit to float (symmetric quantization, zero point = 8)
+          dequant_fc1_expert[linear_idx] = (static_cast<float>(quantized_weight) - 8.0f) * fc1_expert_scales[out_col];
+        }
+      }
+    }
+
+    // Dequantize FC2 weights for all experts (Int4 unpacking)
+    for (int64_t expert_idx = 0; expert_idx < moe_params.num_experts; ++expert_idx) {
```
kunal-vaishnavi: Can we use `ThreadPool::TryParallelFor` here?

onnxruntime/contrib_ops/cpu/quantization/moe_quantization_cpu.cc
```diff
@@ -0,0 +1,530 @@
...
+            quantized_weight = (packed_value >> 4) & 0x0F;  // Upper 4 bits
+          }
+
+          // Dequantize from 4-bit to float (symmetric quantization, zero point = 8)
+          dequant_fc2_expert[linear_idx] = (static_cast<float>(quantized_weight) - 8.0f) * fc2_expert_scales[out_col];
+        }
+      }
+    }
+
+    auto process_token_range = [&](ptrdiff_t start_token, ptrdiff_t end_token) {
```
kunal-vaishnavi: Can we define the `TryParallelFor` here and wrap the computations inside it? Something like this:

https://github.com/microsoft/onnxruntime/blob/b9575476e94daa9c6578aba92d8f04324dd15815/onnxruntime/contrib_ops/cpu/bert/attention_cpu_base.h#L255-L263

onnxruntime/contrib_ops/cpu/quantization/moe_quantization_cpu.cc
```diff
@@ -0,0 +1,530 @@
...
+
+    // Pre-dequantize all expert weights once (shared across all threads)
+    int act = activation_type_ == ActivationType::SwiGLU ? 2 : 1;
+    auto dequant_fc1_weights = IAllocator::MakeUniquePtr<float>(allocator,
+                                                                static_cast<size_t>(moe_params.num_experts * moe_params.hidden_size * moe_params.inter_size * act));
+    auto dequant_fc2_weights = IAllocator::MakeUniquePtr<float>(allocator,
+                                                                static_cast<size_t>(moe_params.num_experts * moe_params.inter_size * moe_params.hidden_size));
+
+    // Dequantize FC1 weights for all experts (including concatenated weights for SwiGLU)
+    for (int64_t expert_idx = 0; expert_idx < moe_params.num_experts; ++expert_idx) {
```
kunal-vaishnavi: Can we use `ThreadPool::TryParallelFor` here?

onnxruntime/contrib_ops/cpu/quantization/moe_quantization_cpu.cc
```diff
@@ -0,0 +1,530 @@
...
+          size_t weight_idx = static_cast<size_t>(out_col * moe_params.hidden_size + in_col);
+          uint8_t quantized_weight = fc1_expert_weights[weight_idx];
+          // Symmetric quantization with zero point = 128
+          dequant_fc1_expert[weight_idx] = (static_cast<float>(quantized_weight) - 128.0f) * fc1_expert_scales[out_col];
+        }
+      }
+    }
+
+    // Dequantize FC2 weights for all experts
+    for (int64_t expert_idx = 0; expert_idx < moe_params.num_experts; ++expert_idx) {
```
kunal-vaishnavi: Can we use `ThreadPool::TryParallelFor` here?

onnxruntime/contrib_ops/cpu/quantization/moe_quantization_cpu.cc
```diff
@@ -0,0 +1,530 @@
...
+        for (int64_t in_col = 0; in_col < moe_params.inter_size; ++in_col) {
+          size_t weight_idx = static_cast<size_t>(out_col * moe_params.inter_size + in_col);
+          uint8_t quantized_weight = fc2_expert_weights[weight_idx];
+          // Symmetric quantization with zero point = 128
+          dequant_fc2_expert[weight_idx] = (static_cast<float>(quantized_weight) - 128.0f) * fc2_expert_scales[out_col];
+        }
+      }
+    }
+
+    auto process_token_range = [&](ptrdiff_t start_token, ptrdiff_t end_token) {
```
kunal-vaishnavi: Can we define the `TryParallelFor` here and wrap the computations inside it? Something like this:

https://github.com/microsoft/onnxruntime/blob/b9575476e94daa9c6578aba92d8f04324dd15815/onnxruntime/contrib_ops/cpu/bert/attention_cpu_base.h#L255-L263

onnxruntime/contrib_ops/cpu/quantization/moe_quantization_cpu.cc
```diff
@@ -0,0 +1,530 @@
...
+            }
+            token_result[i] += routing_weight * thread_fc2_output[i];
+          }
+        }
+      }
+    };  // Execute token processing in parallel across threads
+    concurrency::ThreadPool::TryParallelFor(thread_pool, static_cast<std::ptrdiff_t>(moe_params.num_rows),
+                                            static_cast<double>(std::max<int64_t>(1, moe_params.num_rows / num_threads)),
+                                            process_token_range);
+  } else {
```
kunal-vaishnavi, line 237: Can we generalize both the `if` and `else` branches to re-use the same code and pass parameters as needed to the generalized code to distinguish between `uint4x2` and `uint8`?

onnxruntime/contrib_ops/cpu/quantization/moe_quantization_cpu.cc
```diff
@@ -0,0 +1,530 @@
...
+      int act = activation_type_ == ActivationType::SwiGLU ? 2 : 1;
+      float* thread_fc1_output = thread_fc1_buffers.get() + thread_id * moe_params.inter_size * act;
+      float* thread_fc2_output = thread_fc2_buffers.get() + thread_id * moe_params.hidden_size;
+      float* thread_local_results = thread_results.get() + thread_id * moe_params.num_rows * moe_params.hidden_size;
+
+      // Process each token in this thread's range
+      for (int64_t token_idx = start_token; token_idx < end_token; ++token_idx) {
+        const MLFloat16* token_input_typed = input_data + token_idx * moe_params.hidden_size;
+
+        // Convert input from MLFloat16 to float for MLAS computation
```
kunal-vaishnavi: You can use `MlasConvertHalfToFloatBuffer` here.

https://github.com/microsoft/onnxruntime/blob/b9575476e94daa9c6578aba92d8f04324dd15815/onnxruntime/core/providers/cpu/llm/attention.cc#L73

onnxruntime/contrib_ops/cpu/quantization/moe_quantization_cpu.cc
```diff
@@ -0,0 +1,530 @@
...
+  // Allocate thread-local buffers
+  auto thread_fc1_buffers = IAllocator::MakeUniquePtr<float>(allocator, static_cast<size_t>(num_threads * moe_params.inter_size * (is_swiglu ? 2 : 1)));
+  auto thread_fc2_buffers = IAllocator::MakeUniquePtr<float>(allocator, static_cast<size_t>(num_threads * moe_params.hidden_size));
+  auto thread_results = IAllocator::MakeUniquePtr<float>(allocator, static_cast<size_t>(num_threads * moe_params.num_rows * moe_params.hidden_size));
+
+  // Initialize thread results to zero
+  std::fill(thread_results.get(),
+            thread_results.get() + static_cast<size_t>(num_threads * moe_params.num_rows * moe_params.hidden_size), 0.0f);
+
+  // Helper function to convert MLFloat16 to float
```
kunal-vaishnavi: You can use `MlasConvertHalfToFloatBuffer` here.

https://github.com/microsoft/onnxruntime/blob/b9575476e94daa9c6578aba92d8f04324dd15815/onnxruntime/core/providers/cpu/llm/attention.cc#L73

onnxruntime/contrib_ops/cpu/quantization/moe_quantization_cpu.cc
```diff
@@ -0,0 +1,530 @@
...
+      const int64_t thread_id = start_token / ((moe_params.num_rows + num_threads - 1) / num_threads);
+      float* thread_fc1_output = thread_fc1_buffers.get() + thread_id * moe_params.inter_size * (is_swiglu ? 2 : 1);
+      float* thread_fc2_output = thread_fc2_buffers.get() + thread_id * moe_params.hidden_size;
+      float* thread_local_results = thread_results.get() + thread_id * moe_params.num_rows * moe_params.hidden_size;
+
+      // Process each token in this thread's range
+      for (int64_t token_idx = start_token; token_idx < end_token; ++token_idx) {
+        const MLFloat16* token_input_typed = input_data + token_idx * moe_params.hidden_size;
+
+        // Convert input from MLFloat16 to float for computation
```
kunal-vaishnavi: You can use `MlasConvertHalfToFloatBuffer` here.

https://github.com/microsoft/onnxruntime/blob/b9575476e94daa9c6578aba92d8f04324dd15815/onnxruntime/core/providers/cpu/llm/attention.cc#L73

onnxruntime/contrib_ops/cpu/quantization/moe_quantization_cpu.cc
apsonawane: Yes using MlasConvertHalfToFloatBuffer

onnxruntime/contrib_ops/cpu/quantization/moe_quantization_cpu.cc
```diff
@@ -0,0 +1,530 @@
...
+      }
+    };
+
+    // Execute token processing in parallel across threads
+    concurrency::ThreadPool::TryParallelFor(thread_pool, static_cast<std::ptrdiff_t>(moe_params.num_rows),
+                                            static_cast<double>(std::max<int64_t>(1, moe_params.num_rows / num_threads)),
+                                            process_token_range);
+  }
+
+  // Main thread reduction: combine all thread-local results into final output
```
kunal-vaishnavi: If you want to use MLAS, you can convert the buffer from FP16 to FP32, get the final output, and then convert back from FP32 to FP16. Here is an example.

https://github.com/microsoft/onnxruntime/blob/b9575476e94daa9c6578aba92d8f04324dd15815/onnxruntime/core/providers/cpu/llm/attention.cc#L73-L75

onnxruntime/core/graph/contrib_ops/contrib_defs.cc
```diff
@@ -1478,6 +1478,7 @@ ONNX_MS_OPERATOR_SET_SCHEMA(
                 "T")
         .TypeConstraint("T", {"tensor(float16)"}, "Constrain input and output types to float or float16 tensors.")
```
kunal-vaishnavi:
```suggestion
        .TypeConstraint("T", {"tensor(float)", "tensor(float16)"}, "Constrain input and output types to float or float16 tensors.")
```

apsonawane: Right now, it supports float16 so we do not need to update this right now. Will add support for float after this PR

onnxruntime/contrib_ops/cpu/quantization/moe_quantization_cpu.cc
apsonawane: Done

onnxruntime/core/graph/contrib_ops/contrib_defs.cc
apsonawane: Will add it after this PR is pushed

onnxruntime/contrib_ops/cpu/quantization/moe_quantization_cpu.cc
apsonawane: Updated now it uses correct zero points

onnxruntime/contrib_ops/cpu/cpu_contrib_kernels.h
```diff
@@ -6,6 +6,9 @@
 #include "core/framework/op_kernel.h"
 #include "core/framework/kernel_registry.h"
 
+// Forward declarations for QMoE
+#include "contrib_ops/cpu/quantization/moe_quantization_cpu.h"
```
kunal-vaishnavi: Why are forward declarations needed here? Other ops don't seem to need to include their headers here

onnxruntime/contrib_ops/cpu/quantization/moe_quantization_cpu.cc
```diff
@@ -0,0 +1,438 @@
...
+                          (*KernelDefBuilder::Create())                                    \
+                              .MayInplace(0, 0)                                            \
+                              .TypeConstraint("T", BuildKernelDefConstraints<MLFloat16>()) \
+                              .TypeConstraint("T1", BuildKernelDefConstraints<uint8_t>())  \
+                              .TypeConstraint("T2", BuildKernelDefConstraints<float>()),   \
+                          QMoE);
+
+REGISTER_KERNEL();
+
+// QMoE CPU kernel registration is handled in cpu_contrib_kernels.cc
```
kunal-vaishnavi, line 31: Why not register it here?

onnxruntime/contrib_ops/cpu/moe/moe_base_cpu.h
```diff
@@ -0,0 +1,243 @@
...
+#pragma once
+
+#include "core/common/common.h"
+#include "core/framework/tensor_shape.h"
+#include "core/framework/op_kernel.h"
+
+namespace onnxruntime {
+namespace contrib {
+
+enum class MoEParallelType {
```
kunal-vaishnavi, line 13: Similar to how [`attention_common.h`](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/contrib_ops/cpu/bert/attention_common.h) has enums used in a variety of attention ops, can we make a similar file called `moe_common.h` for existing and any future MoE ops? We can then move these enums and structs there.

onnxruntime/contrib_ops/cpu/quantization/moe_quantization_cpu.cc
```diff
@@ -0,0 +1,438 @@
...
+      case ActivationType::SwiGLU:
+        // SwiGLU: This is handled specially as it requires gating, not applied here
+        return x;
+      default:
+        return x;  // Default to identity
+    }
+  };
+
+  // Helper function to apply SwiGLU activation: gate * sigmoid(1.702 * gate) * (linear + 1)
+  auto ApplySwiGLU = [](const float* fc1_output, float* result, int64_t inter_size) {
```
kunal-vaishnavi: Can we move this into a function for other ops to also leverage in the future?

onnxruntime/contrib_ops/cpu/quantization/moe_quantization_cpu.cc
```diff
@@ -0,0 +1,438 @@
...
+                                         router_probs_float.get(),
+                                         static_cast<size_t>(moe_params.num_rows * moe_params.num_experts),
+                                         thread_pool);
+
+  // Initialize thread results to zero using optimized memset
+  std::memset(thread_results.get(), 0,
+              static_cast<size_t>(num_threads * moe_params.num_rows * moe_params.hidden_size) * sizeof(float));
+
+  // Helper function to apply activation
+  auto ApplyActivation = [](float x, ActivationType activation_type) {
```
kunal-vaishnavi: Can we move this into a function for other ops to also leverage in the future?

onnxruntime/contrib_ops/cpu/quantization/moe_quantization_cpu.cc
```diff
@@ -0,0 +1,438 @@
...
+      return (static_cast<float>(weights[weight_idx]) - zero_point) * scales[scale_idx];
+    }
+  };
+
+  // Dequantize FC1 weights for all experts
+  concurrency::ThreadPool::TryParallelFor(
+      thread_pool, static_cast<std::ptrdiff_t>(moe_params.num_experts),
+      static_cast<double>(std::max<int64_t>(1, moe_params.num_experts / num_threads)),
+      [&](ptrdiff_t expert_start, ptrdiff_t expert_end) {
+        for (int64_t expert_idx = expert_start; expert_idx < expert_end; ++expert_idx) {
```
kunal-vaishnavi, line 218: Keeping the type as `std::ptrdiff_t` is safer.

```suggestion
        for (std::ptrdiff_t expert_idx = expert_start; expert_idx < expert_end; ++expert_idx) {
```

We can then use `SafeInt` to prevent integer overflow when using `expert_idx`.

kunal-vaishnavi: Keeping the type as `std::ptrdiff_t` is safer.

```suggestion
        for (std::ptrdiff_t expert_idx = expert_start; expert_idx < expert_end; ++expert_idx) {
```

We can then use [`SafeInt`](https://learn.microsoft.com/en-us/cpp/safeint/safeint-library?view=msvc-170) to prevent integer overflow when using `expert_idx`. Here is an example.

https://github.com/microsoft/onnxruntime/blob/4a8a289e635f5beabd33d111fd7d31f549300682/onnxruntime/contrib_ops/cpu/bert/attention_cpu_base.h#L260

onnxruntime/contrib_ops/cpu/quantization/moe_quantization_cpu.cc
```diff
@@ -0,0 +1,438 @@
...
+          }
+        }
+      });
+
+  // Dequantize FC2 weights for all experts
+  concurrency::ThreadPool::TryParallelFor(
+      thread_pool, static_cast<std::ptrdiff_t>(moe_params.num_experts),
+      static_cast<double>(std::max<int64_t>(1, moe_params.num_experts / num_threads)),
+      [&](ptrdiff_t expert_start, ptrdiff_t expert_end) {
+        for (int64_t expert_idx = expert_start; expert_idx < expert_end; ++expert_idx) {
```
kunal-vaishnavi: Keeping the type as `std::ptrdiff_t` is safer.

```suggestion
        for (std::ptrdiff_t expert_idx = expert_start; expert_idx < expert_end; ++expert_idx) {
```

We can then use [`SafeInt`](https://learn.microsoft.com/en-us/cpp/safeint/safeint-library?view=msvc-170) to prevent integer overflow when using `expert_idx`. Here is an example.

https://github.com/microsoft/onnxruntime/blob/4a8a289e635f5beabd33d111fd7d31f549300682/onnxruntime/contrib_ops/cpu/bert/attention_cpu_base.h#L260

onnxruntime/contrib_ops/cpu/quantization/moe_quantization_cpu.cc
```diff
@@ -0,0 +1,438 @@
...
+      [&](ptrdiff_t start_token, ptrdiff_t end_token) {
+        const int64_t thread_id = start_token / ((moe_params.num_rows + num_threads - 1) / num_threads);
+        const int64_t thread_fc1_size = is_4bit ? (moe_params.inter_size * (is_swiglu ? 2 : 1)) : (moe_params.inter_size * act_multiplier);
+        float* thread_fc1_output = thread_fc1_buffers.get() + thread_id * thread_fc1_size;
+        float* thread_fc2_output = thread_fc2_buffers.get() + thread_id * moe_params.hidden_size;
+        float* thread_local_results = thread_results.get() + thread_id * moe_params.num_rows * moe_params.hidden_size;
+        float* thread_bias_buffer = thread_bias_buffers.get() + thread_id * max_bias_size;
+
+        // Process each token in this thread's range
+        for (int64_t token_idx = start_token; token_idx < end_token; ++token_idx) {
```
kunal-vaishnavi: Keeping the type as `std::ptrdiff_t` is safer.

```suggestion
        for (std::ptrdiff_t token_idx = start_token; token_idx < end_token; ++token_idx) {
```

We can then use [`SafeInt`](https://learn.microsoft.com/en-us/cpp/safeint/safeint-library?view=msvc-170) to prevent integer overflow when using `expert_idx`. Here is an example.

https://github.com/microsoft/onnxruntime/blob/4a8a289e635f5beabd33d111fd7d31f549300682/onnxruntime/contrib_ops/cpu/bert/attention_cpu_base.h#L260

onnxruntime/contrib_ops/cpu/quantization/moe_quantization_cpu.cc
```diff
@@ -0,0 +1,438 @@
...
+  void* float_output_ptr = allocator->Alloc(static_cast<size_t>(total_output_size * sizeof(float)));
+  BufferUniquePtr float_output_buffer(float_output_ptr, BufferDeleter(allocator));
+  float* float_output = reinterpret_cast<float*>(float_output_ptr);
+
+  // Main thread reduction: combine all thread-local results into float buffer
+  concurrency::ThreadPool::TryParallelFor(
+      thread_pool, static_cast<std::ptrdiff_t>(moe_params.num_rows),
+      static_cast<double>(std::max<int64_t>(1, moe_params.num_rows / num_threads)),
+      [&](ptrdiff_t token_start, ptrdiff_t token_end) {
+        for (int64_t token_idx = token_start; token_idx < token_end; ++token_idx) {
```
kunal-vaishnavi: Keeping the type as `std::ptrdiff_t` is safer.

```suggestion
        for (std::ptrdiff_t token_idx = token_start; token_idx < token_end; ++token_idx) {
```

We can then use [`SafeInt`](https://learn.microsoft.com/en-us/cpp/safeint/safeint-library?view=msvc-170) to prevent integer overflow when using `expert_idx`. Here is an example.

https://github.com/microsoft/onnxruntime/blob/4a8a289e635f5beabd33d111fd7d31f549300682/onnxruntime/contrib_ops/cpu/bert/attention_cpu_base.h#L260

onnxruntime/contrib_ops/cpu/moe/moe_base_cpu.h
apsonawane: This is aligned with CUDA kernel: https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/contrib_ops/cuda/moe/moe_base.h

onnxruntime/contrib_ops/cpu/quantization/moe_quantization_cpu.cc
apsonawane: All the CPU contrib ops are registered in that file so added there for consistency, also we are registering the kernel in its specific file.
https://github.com/microsoft/onnxruntime/pull/25558/files/e1d970da006dcf40d09137a99aec5e7f57b94c18#diff-6ba1e6f7854ee66e091299c4b63318d1b240173d135b3dc6c9e5143dd88f4066R19

onnxruntime/contrib_ops/cpu/quantization/moe_quantization_cpu.cc
apsonawane: Done

onnxruntime/contrib_ops/cpu/quantization/moe_quantization_cpu.cc
apsonawane: Done

onnxruntime/contrib_ops/cpu/quantization/moe_quantization_cpu.cc
apsonawane: Done

onnxruntime/contrib_ops/cpu/cpu_contrib_kernels.h
apsonawane: Removed

tianleiwu (approved):

kunal-vaishnavi (dismissed):

onnxruntime/contrib_ops/cpu/moe/moe_base_cpu.h
```diff
@@ -0,0 +1,243 @@
...
+      return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT, "router_probs_dims must be 2D, got ",
+                             router_probs_dims.size());
+    }
+    if (router_probs_dims[0] != num_rows) {
+      return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT, "router_probs_dims[0] must be equal to num_rows, got ",
+                             router_probs_dims[0], " and ", num_rows);
+    }
+
+    // Optional bias validation
+    if (fc1_experts_bias_optional != nullptr && fc2_experts_bias_optional != nullptr) {
```
tianleiwu: why use && here? 
I think there is chance that fc1 has bias, but fc2 has not.

onnxruntime/contrib_ops/cpu/moe/moe_base_cpu.h
```diff
@@ -0,0 +1,243 @@
...
+      }
+      if (fc1_experts_bias_dims[0] != local_num_experts) {
+        return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT, "fc1_experts_bias_dims[0] must be equal to local_num_experts, got ",
+                               fc1_experts_bias_dims[0], " and ", local_num_experts);
+      }
+      if (fc2_experts_bias_dims[0] != local_num_experts) {
+        return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT, "fc2_experts_bias_dims[0] must be equal to local_num_experts, got ",
+                               fc2_experts_bias_dims[0], " and ", local_num_experts);
+      }
+      if (fc1_experts_bias_dims[1] != inter_size) {
```
tianleiwu: For swiglu, this shall be 2 * inter_size

onnxruntime/test/python/transformers/test_parity_moe.py
```diff
@@ -1093,6 +1093,16 @@ def test_phi3_moe_parity(self, batch_size, sequence_length, quant_bits):
...
         # phi3_moe.benchmark_ort()
 
+    @parameterized.expand([(b, s, q) for b, s, q in phi3_test_params if q in (8, 4)])
+    def test_phi3_qmoe_cpu_parity(self, batch_size, sequence_length, quant_bits):
+        if "CPUExecutionProvider" not in onnxruntime.get_available_providers():
+            self.skipTest("CPUExecutionProvider is not available.")
+        config = PhiMoEConfig(hidden_size=256, intermediate_size=1024)
+        phi3_moe = PhiMoESparseMoeBlock(config, batch_size, sequence_length, quant_bits)
+        if phi3_moe.ort_sess is not None:
+            phi3_moe.ort_sess.set_providers(["CPUExecutionProvider"])
```
tianleiwu: This will not work. We shall not update providers after the session is created. It is because, when session creates with CUDA, some memory might be allocated in GPU memory.

onnxruntime/contrib_ops/cpu/moe/moe_utils.cc
```diff
@@ -0,0 +1,43 @@
...
+    default:
+      return x;  // Default to identity
+  }
+}
+
+void ApplySwiGLU(const float* fc1_output, float* result, int64_t inter_size) {
+  constexpr float swiglu_alpha = 1.702f;
+  for (int64_t i = 0; i < inter_size; ++i) {
+    float linear_val = fc1_output[i];             // First half: linear projection
+    float gate_val = fc1_output[i + inter_size];  // Second half: gate projection
```
tianleiwu: We shall use interleave format like cuda by default. Interleave might be faster due to cache.

onnxruntime/contrib_ops/cpu/quantization/moe_quantization_cpu.cc
```diff
@@ -0,0 +1,407 @@
...
+namespace onnxruntime {
+namespace contrib {
+
+#define REGISTER_KERNEL()                                                                  \
+  ONNX_OPERATOR_KERNEL_EX(QMoE, kMSDomain, 1, kCpuExecutionProvider,                       \
+                          (*KernelDefBuilder::Create())                                    \
+                              .MayInplace(0, 0)                                            \
+                              .TypeConstraint("T", BuildKernelDefConstraints<MLFloat16>()) \
+                              .TypeConstraint("T1", BuildKernelDefConstraints<uint8_t>())  \
+                              .TypeConstraint("T2", BuildKernelDefConstraints<float>()),   \
```
tianleiwu: Suggest to use same T2 as T if perf and accuracy no significant difference. That will be consistent with cuda.

onnxruntime/contrib_ops/cpu/quantization/moe_quantization_cpu.cc
```diff
@@ -0,0 +1,407 @@
...
+      uint8_t packed_value = weights[packed_idx];
+      uint8_t quantized_weight = (linear_idx % 2 == 0) ? (packed_value & 0x0F) : ((packed_value >> 4) & 0x0F);
+      return (static_cast<float>(quantized_weight) - zero_point) * scales[scale_idx];
+    } else {
+      // For Int8, direct access
+      return (static_cast<float>(weights[weight_idx]) - zero_point) * scales[scale_idx];
+    }
+  };
+
+  // Dequantize FC1 weights for all experts
```
tianleiwu: Could we dequantize the weights only once (like in prepacking)?

onnxruntime/contrib_ops/cpu/quantization/moe_quantization_cpu.cc
```diff
@@ -0,0 +1,407 @@
...
+          // Process all experts for this token
+          for (std::ptrdiff_t expert_idx = 0; expert_idx < moe_params.num_experts; ++expert_idx) {
+            float routing_weight = router_probs_float.get()[static_cast<int64_t>(SafeInt<int64_t>(token_idx)) * moe_params.num_experts + static_cast<int64_t>(SafeInt<int64_t>(expert_idx))];
+            if (routing_weight <= 1e-6f) continue;  // Skip experts with negligible routing weight
+
+            // FC1: input -> intermediate using pre-dequantized weights + MLAS SGEMM
+            const int64_t fc1_weight_offset = is_4bit ? (static_cast<int64_t>(SafeInt<int64_t>(expert_idx)) * moe_params.hidden_size * fc1_output_size) : (static_cast<int64_t>(SafeInt<int64_t>(expert_idx)) * moe_params.hidden_size * moe_params.inter_size * act_multiplier);
+            const float* fc1_expert_weights = dequant_fc1_weights.get() + fc1_weight_offset;
+
+            const int64_t fc1_bias_size = is_4bit ? fc1_output_size : (moe_params.inter_size * act_multiplier);
```
tianleiwu: nit: fc1_bias_size shall be same as fc1_output_size no matter it is 4 or 8 bits

onnxruntime/contrib_ops/cpu/moe/moe_base_cpu.h
```diff
@@ -0,0 +1,243 @@
...
+  int64_t hidden_size;
+  int64_t inter_size;
+
+  MoEParallelType parallel_type;
+  int64_t tensor_shards{1};
+};
+
+class MoEBaseCPU {
+ public:
+  Status CheckInputs(MoEParameters& parameters, MoEQuantType& quant_type, const Tensor* input,
```
tianleiwu, line 49: It is possible to share the same helper function for CheckInputs with CUDA op.

apsonawane: Will add it in next PR

onnxruntime/test/python/transformers/test_parity_moe.py
apsonawane: Fixed

tianleiwu:
The Python code utilizes a symmetric quantization approach, while the C++ code employs asymmetric dequantization. They are not matched.

test_phi_moe_parity attempts to create a MoE graph using fc3_experts_weights. However, the C++ kernel contains an explicit check that returns a NOT_IMPLEMENTED error.

ThreadPool::TryParallelFor gives each thread a distinct range of tokens to process. Each thread then writes its results into a large, unique, thread-local buffer (thread_local_results). Why do we need main thread to accumulate results since different threads have no overlapped in tokens?

Inconsistent SwiGLU Implementation: The logic for applying the SwiGLU activation is different depending on whether the weights are 4-bit or 8-bit quantized.
For 4-bit, it uses a helper (ApplySwiGLU) that assumes an interleaved [linear, gate, linear, gate, ...] memory layout.
For 8-bit, it implements the logic inline, assuming a chunked [linear..., gate...] layout.

Repetitive Bias Conversion: When using MLFloat16 inputs, the bias tensors are converted to float inside the tightest loop (for every token and every expert). This is highly inefficient.
Recommendation: Convert the entire bias tensor from MLFloat16 to a float buffer once before the main parallel loop begins.

onnxruntime/contrib_ops/cpu/moe/moe_utils.cc
```diff
@@ -26,17 +26,43 @@ float ApplyActivation(float x, ActivationType activation_type) {
...
+      // SwiGLU: gate * sigmoid(alpha * gate) * (linear + 1)
+      float sigmoid_arg = swiglu_alpha * gate_val;
+      float sigmoid_out = 1.0f / (1.0f + std::exp(-sigmoid_arg));
+      float swish_out = gate_val * sigmoid_out;
+      result_buffer[i] = swish_out * (linear_val + 1.0f);
+    }
   }
+
+  // Copy result back to data (first inter_size elements only - rest is overwritten by GEMM)
+  std::memcpy(data, result_buffer.get(), inter_size * sizeof(float));
```
tianleiwu: Can we not use a temp buffer? Just directly write to the data instead?

onnxruntime/contrib_ops/cpu/quantization/moe_quantization_cpu.cc
```diff
@@ -152,15 +152,31 @@ Status QMoE::QuantizedMoEImpl(OpKernelContext* context,
...
+  // Pre-convert bias tensors to float (if they exist)
+  const int64_t fc1_bias_size = is_swiglu ? 2 * moe_params.inter_size : moe_params.inter_size;
+  const int64_t fc2_bias_size = moe_params.hidden_size;
+
+  // Allocate buffers for converted biases
+  std::unique_ptr<float[]> fc1_bias_float;
+  std::unique_ptr<float[]> fc2_bias_float;
+
+  if (fc1_bias_data) {
+    fc1_bias_float = std::make_unique<float[]>(static_cast<size_t>(moe_params.num_experts * fc1_bias_size));
```
tianleiwu: Please use ORT allocator. The Arena could help performance than allocating in heap.

onnxruntime/contrib_ops/cpu/quantization/moe_quantization_cpu.cc
```diff
@@ -173,17 +189,41 @@ Status QMoE::QuantizedMoEImpl(OpKernelContext* context,
...
+    }
   } else {
     // For float, copy directly
     std::memcpy(input_float.get(), input_data,
                 static_cast<size_t>(moe_params.num_rows * moe_params.hidden_size) * sizeof(float));
     std::memcpy(router_probs_float.get(), router_probs_data,
                 static_cast<size_t>(moe_params.num_rows * moe_params.num_experts) * sizeof(float));
+
+    // For float, just point to the original data
+    if (fc1_bias_data) {
```
tianleiwu, line 248: For float, can we avoid memcpy and let pointer points to the existing bias?

onnxruntime/contrib_ops/cpu/moe/moe_utils.cc
apsonawane: Fixed

onnxruntime/contrib_ops/cpu/quantization/moe_quantization_cpu.cc
apsonawane: Fixed

onnxruntime/contrib_ops/cpu/quantization/moe_quantization_cpu.cc
apsonawane: Fixed

tianleiwu (dismissed):

kunal-vaishnavi (dismissed):

PR #25561, opened 2025-07-28
Workaround fix memory allocation for MIGraphX EP

urpetkov-amd:
This PR has workaround fix for allocation GPU memory for MIGraphX EP. For other EPs it's working same as before. Currently, unnecessary amount of memory has been allocated to CPU.


urpetkov-amd:
This PR has been closed because it has been made from forked repo and wrong branch.

PR #25563, opened 2025-07-28
Cherry-pick round 1

snnn:
- **DynamicQuantizeMatMul - handle case where B zero point input is provided but not constant. (#25544)**
- **Refactor plugin EP support (#25541)**
- **Remove the python installation steps from win-qnn-arm64-ci-pipeline.yml (#25552)**


snnn:
There are some test failures in the Nuget Test pipelines. They were caused by recent moves of our pipelines and build machines. I will address them later this week.

edgchen1 (approved):

fs-eire (approved):

snnn:
I am sorry. This PR targeted the wrong branch. I should set it to rel-1.23.0.  I will create a new PR for this.

jywu-mysoft:
how should we fix this? should we revert this commit?

fs-eire:
> how should we fix this? should we revert this commit?

It’s OK. If you check the commit history, this commit contains no code change, so there is no need to revert anything. We need another PR CP to the release branch.

PR #25565, opened 2025-07-28
[EP ABI] Node_GetAttrByName returns ORT_NOT_FOUND with non-existing attr name

adrianlizarraga:
### Description
Updates `Node_GetAttributeByName` to return an error status with code `ORT_NOT_FOUND` and set the `attribute` output parameter to `NULL` when called with a non-existing attribute name.

Why? Currently, a caller has to do string comparison of the `OrtStatus` error message to determine if the attribute does not exist or if another error occurred. This can be somewhat cumbersome. With this change, the caller can just check the error code.

### Motivation and Context
Make it easier to use `Node_GetAttributeByName`.


onnxruntime/core/graph/ep_api_types.cc
```diff
@@ -87,6 +87,28 @@ static void ConvertNodeArgsToValueInfos(const EpGraph* ep_graph,
...
+    return false;  // Not an attribute for this operator type.
+  }
+
+  const ONNX_NAMESPACE::OpSchema::Attribute& attr_schema = attr_schema_iter->second;
+
+  return !attr_schema.required;
+#else
+  ORT_UNUSED_PARAMETER(node);
+  ORT_UNUSED_PARAMETER(attr_name);
+  return false;
```
edgchen1: nit: having `IsOptionalAttribute()` unconditionally return false seems inaccurate. IMO, it shouldn't be implemented in a minimal build. maybe we can put the `ORT_MINIMAL_BUILD` ifdef in `EpNode::GetAttribute()` and document that `is_unset_optional_attr` is only set properly in a non-minimal build.

onnxruntime/test/ep_graph/test_ep_graph.cc
```diff
@@ -87,6 +87,92 @@ TEST(EpGraphTest, Check3LayerNestedSubgraphV2) {
...
+  ASSERT_ORTSTATUS_OK(ort_api.Graph_GetNumNodes(&ort_graph, &num_nodes));
+  ASSERT_EQ(num_nodes, 1);
+
+  std::vector<const OrtNode*> nodes(num_nodes);
+  ASSERT_ORTSTATUS_OK(ort_api.Graph_GetNodes(&ort_graph, nodes.data(), nodes.size()));
+
+  const OrtNode* conv_node = nodes[0];
+  const char* op_type = nullptr;
+  ASSERT_ORTSTATUS_OK(ort_api.Node_GetOperatorType(conv_node, &op_type));
+  ASSERT_EQ(std::string(op_type), "Conv");
```
edgchen1: btw, there's also `ASSERT_STREQ` for comparing C strings

onnxruntime/test/ep_graph/test_ep_graph.cc
```diff
@@ -87,6 +87,92 @@ TEST(EpGraphTest, Check3LayerNestedSubgraphV2) {
   CheckGraphCApi(test_graph->GetGraphViewer(), test_graph->GetOrtGraph());
 }
 
+TEST(EpGraphTest, GetAttributeByName) {
+  // Load model with a single Conv that has no explicit attributes set.
+  auto test_graph = TestGraph::Load(ORT_TSTR("testdata/conv_default_attrs.onnx"));
```
edgchen1, line 92: please add a script that can generate this file.

csharp/src/Microsoft.ML.OnnxRuntime/Exceptions.shared.cs
```diff
@@ -23,8 +23,10 @@ internal enum ErrorCode
         ModelLoaded = 8,
         NotImplemented = 9,
         InvalidGraph = 10,
-        ShapeInferenceNotRegistered = 11,
-        RequirementNotRegistered = 12,
+        ShapeInferenceNotRegistered = 11,  // ORT_EP_FAIL
+        RequirementNotRegistered = 12,     // ORT_MODEL_LOAD_CANCELED
```
edgchen1: why do these have different names (and seemingly different meanings) than the OrtErrorCode values?

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -6031,6 +6032,10 @@ struct OrtApi {
    *                           Typical usage sets this to the result of Node_GetNumAttributes(). An error status is
    *                           returned if `num_attributes` is less than the number of node attributes.
    *
+   * \note ONNX Runtime automatically sets optional (unset) attributes to their default values if the default value
+   * is a constant expression that does not depend on other tensor/model characteristics. Conv's 'kernel_shape'
+   * attribute is an example of an optional attribute that does not have a constant default value.
```
edgchen1: good information in the note.

maybe also mention that this function will not include any optional attributes without constant default values in its output.

onnxruntime/test/ep_graph/test_ep_graph.cc
adrianlizarraga: added

csharp/src/Microsoft.ML.OnnxRuntime/Exceptions.shared.cs
adrianlizarraga: I don't know. I'm considering removing the new error code from all other language bindings so that the owners can add it properly themselves.

onnxruntime/core/graph/ep_api_types.cc
adrianlizarraga: Updated

onnxruntime/test/ep_graph/test_ep_graph.cc
adrianlizarraga: thanks, updated.

csharp/src/Microsoft.ML.OnnxRuntime/Exceptions.shared.cs
adrianlizarraga: Left TODO and removed the new `ORT_NOT_FOUND` enum. Will wait until someone more familiar with the csharp bindings takes a look.

include/onnxruntime/core/session/onnxruntime_c_api.h
adrianlizarraga: thanks, updated.

edgchen1 (approved):

PR #25566, opened 2025-07-28
[EP ABI] Support for TENSOR type attribute

chilo-ms:
### Description
Add a new `Node_GetTensorAttributeAsOrtValue` API to support attribute that is a `TENSOR` type.
This API returns a const OrtValue that represents the TensorProto in the `TENSOR `attribute.


include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -6052,6 +6053,19 @@ struct OrtApi {
...
+   *
+   * \param[in] node The OrtNode instance.
+   * \param[in] attribute The OrtOpAttr instance.
+   * \param[out] name Output parameter set to the attribute's name. The name is a null-terminated string.
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \since Version 1.23.
+   */
+  ORT_API2_STATUS(Node_GetTensorAttributeAsOrtValue, _In_ const OrtNode* node, _In_ const OrtOpAttr* attribute,
```
chilo-ms, line 6080: Thought about changing the API to `OpAttr_GetTensorAttributeAsOrtValue` and make the OrtOpAttr own the OrtValue instance. Then i need to make sure OrtOpAttr is still interchangeable with AttributeProto.

onnxruntime/core/graph/ep_api_types.cc
```diff
@@ -113,12 +113,33 @@ Status EpNode::Create(const Node& node, const EpGraph* ep_graph,
 
   if (node_attrs.size() > 0) {
     ep_node_attributes.reserve(node_attrs.size());
+    std::unordered_map<std::string_view, std::unique_ptr<OrtValue>> tensor_attribute_values;
```
chilo-ms: `Constant` op and `ConstantOfShape` op both support 'TENSOR' attribute.
For `ConstantOfShape`, it even supports if 'TENSOR' attribute is not provided, it defaults to a tensor of value 0 and datatype float32. Will add the support here as well.

No need to do anything in the OrtNode internally.

include/onnxruntime/core/providers/utils/ort_graph_to_proto.h
```diff
@@ -758,6 +758,102 @@ static Ort::Status OrtOpAttrToProto(const OrtOpAttr& ort_attr, onnx::AttributePr
...
+
+      const void* data = nullptr;
+      size_t data_bytes = 0;
+      ORT_EP_UTILS_C_RETURN_IF_ERROR(ort_api.GetTensorData(ort_value, &data));
+      ORT_EP_UTILS_C_RETURN_IF_ERROR(ort_api.GetTensorSizeInBytes(ort_value, &data_bytes));
+
+      // Copy the Ortvalue to TensorProto as raw data
+      tensor_proto->set_raw_data(data, data_bytes);
+
+      ort_api.ReleaseTensorTypeAndShapeInfo(type_shape_info);
```
edgchen1: can we use the C++ API so we don't have to explicitly call release? if we do not, we need to ensure that `type_shape_info` gets released in every code path, including early returns.

include/onnxruntime/core/providers/utils/ort_graph_to_proto.h
```diff
@@ -758,6 +758,102 @@ static Ort::Status OrtOpAttrToProto(const OrtOpAttr& ort_attr, onnx::AttributePr
...
+
+      *(attr_proto.mutable_t()) = *tensor_proto;  // Copy TensorProto into attribute
```
edgchen1: is there a move assignment?

include/onnxruntime/core/providers/utils/ort_graph_to_proto.h
```diff
@@ -758,6 +758,102 @@ static Ort::Status OrtOpAttrToProto(const OrtOpAttr& ort_attr, onnx::AttributePr
 
       break;
     }
+    case OrtOpAttrType::ORT_OP_ATTR_TENSOR: {
+      attr_proto.set_type(onnx::AttributeProto_AttributeType_TENSOR);
+
+      std::unique_ptr<onnx::TensorProto> tensor_proto = std::make_unique<onnx::TensorProto>();
```
edgchen1: do we need a `unique_ptr<onnx::TensorProto>` or can we just use a `onnx::TensorProto`?

onnxruntime/core/graph/abi_graph_types.h
```diff
@@ -251,6 +251,14 @@ struct OrtNode {
   /// <returns>A status indicating success or an error.</returns>
   virtual onnxruntime::Status GetAttributes(gsl::span<const OrtOpAttr*> attrs) const = 0;
 
+  /// <summary>
+  /// Gets the node's attributes.
+  /// </summary>
+  /// <param name="attrs">Buffer into which to copy the attributes.</param>
```
edgchen1: update documentation

onnxruntime/core/graph/ep_api_types.cc
```diff
@@ -110,12 +110,42 @@ Status EpNode::Create(const Node& node, const EpGraph* ep_graph,
   const auto& node_attrs = node.GetAttributes();
   std::unordered_map<std::string, std::unique_ptr<ONNX_NAMESPACE::AttributeProto>> ep_node_attributes_map;
   std::vector<OrtOpAttr*> ep_node_attributes;
+  std::unordered_map<std::string, std::unique_ptr<OrtValue>> tensor_attribute_values;
+
+  auto tensor_proto_to_ort_value = [&](const ONNX_NAMESPACE::TensorProto& tensor_proto,
+                                       std::string& tensor_proto_name) -> Status {
```
edgchen1:
```suggestion
                                       const std::string& tensor_proto_name) -> Status {
```

onnxruntime/core/graph/ep_api_types.cc
```diff
@@ -147,6 +177,11 @@ Status EpNode::Create(const Node& node, const EpGraph* ep_graph,
   ep_node->inputs_ = std::move(ep_node_inputs);
   ep_node->outputs_ = std::move(ep_node_outputs);
   ep_node->attributes_map_ = std::move(ep_node_attributes_map);
+
+  if (!tensor_attribute_values.empty()) {
+    ep_node->tensor_attribute_values_ = std ::move(tensor_attribute_values);
+  }
```
edgchen1: nit: it feels unnecessary and inconsistent to check for an empty value. e.g., we don't do that for `ep_node_subgraphs`.
also, extra space in `std ::move`.
```suggestion
  ep_node->tensor_attribute_values_ = std::move(tensor_attribute_values);
```

onnxruntime/core/graph/ep_api_types.cc
```diff
@@ -230,6 +265,30 @@ Status EpNode::GetAttributes(gsl::span<const OrtOpAttr*> dst) const {
   return Status::OK();
 }
 
+Status EpNode::GetTensorAttributeAsOrtValue(const OrtOpAttr* attribute, const OrtValue*& result) const {
+  const auto attr_proto = reinterpret_cast<const ONNX_NAMESPACE::AttributeProto*>(attribute);
```
edgchen1:
```suggestion
  const auto* attr_proto = reinterpret_cast<const ONNX_NAMESPACE::AttributeProto*>(attribute);
```

onnxruntime/core/graph/ep_api_types.cc
```diff
@@ -230,6 +265,30 @@ Status EpNode::GetAttributes(gsl::span<const OrtOpAttr*> dst) const {
...
+  if (attr_proto->type() != onnx::AttributeProto::TENSOR) {
+    result = nullptr;
+    return ORT_MAKE_STATUS(ONNXRUNTIME, FAIL, "This OrtOpAttr instance is not a 'TENSOR' attribute");
+  }
+
+  const auto& tensor_proto = attr_proto->t();
+  std::string tensor_proto_name = tensor_proto.name();
+  if (tensor_proto.name().empty()) {
+    tensor_proto_name = node_.Name() + "_" + attr_proto->name();
+  }
```
edgchen1: maybe put this tensor proto name generation logic into a helper function so we only have to define it once.

onnxruntime/core/graph/model_editor_api_types.h
```diff
@@ -137,6 +137,11 @@ struct ModelEditorNode : public OrtNode {
                            "OrtModelEditorApi does not support getting attribute OrtOpAttr for OrtNode");
   }
 
+  Status GetTensorAttributeAsOrtValue(const OrtOpAttr* /*attribute*/, const OrtValue*& /*attr_tensor*/) const override {
+    return ORT_MAKE_STATUS(ONNXRUNTIME, NOT_IMPLEMENTED,
+                           "OrtModelEditorApi does not support getting 'TENSOR' attribute OrtOpAttr for OrtNode");
```
edgchen1:
```suggestion
                           "OrtModelEditorApi does not support getting 'TENSOR' attribute for OrtNode");
```

onnxruntime/core/session/onnxruntime_c_api.cc
```diff
@@ -3015,6 +3015,32 @@ ORT_API_STATUS_IMPL(OrtApis::Node_GetAttributeByName, _In_ const OrtNode* node,
...
+  if (attr_tensor == nullptr) {
+    return OrtApis::CreateStatus(ORT_INVALID_ARGUMENT, "attr_tensor argument is null");
+  }
+  if (attribute == nullptr) {
+    return OrtApis::CreateStatus(ORT_INVALID_ARGUMENT, "attribute argument is null");
+  }
+
+  const EpNode* ep_node = EpNode::ToInternal(node);
+  if (ep_node == nullptr) {
+    return OrtApis::CreateStatus(OrtErrorCode::ORT_INVALID_ARGUMENT, "node is a ModelEditorNode which doesn't support Node_GetAttributeByName.");
```
edgchen1: update status message

onnxruntime/core/session/onnxruntime_c_api.cc
```diff
@@ -3015,6 +3015,32 @@ ORT_API_STATUS_IMPL(OrtApis::Node_GetAttributeByName, _In_ const OrtNode* node,
...
+  }
+
+  const auto& tensor_proto = reinterpret_cast<const ONNX_NAMESPACE::AttributeProto*>(attribute)->t();
+
+  ORT_ENFORCE(utils::HasDataType(tensor_proto));
+  ORT_ENFORCE(ONNX_NAMESPACE::TensorProto::DataType_IsValid(tensor_proto.data_type()));
```
edgchen1: nit: add error messages

onnxruntime/core/graph/ep_api_types.cc
```diff
@@ -110,12 +110,42 @@ Status EpNode::Create(const Node& node, const EpGraph* ep_graph,
...
+    return Status::OK();
+  };
 
   if (node_attrs.size() > 0) {
     ep_node_attributes.reserve(node_attrs.size());
 
     for (const auto& item : node_attrs) {
       auto attr = std::make_unique<ONNX_NAMESPACE::AttributeProto>(item.second);  // Copy AttributeProto and owned by this EpNode object.
+
+      // Create and cache an OrtValue for the 'TENSOR' attribute
```
edgchen1: why do we cache all of the tensor valued attributes at initialization vs. creating and returning an `OrtValue` for them on demand?

do we expect the caching to provide a benefit in most usages and not have excessive memory overhead?

onnxruntime/core/graph/abi_graph_types.h
```diff
@@ -251,6 +251,14 @@ struct OrtNode {
   /// <returns>A status indicating success or an error.</returns>
   virtual onnxruntime::Status GetAttributes(gsl::span<const OrtOpAttr*> attrs) const = 0;
 
+  /// <summary>
+  /// Gets the node's attributes.
```
adrianlizarraga: Looks like comment is for a different funciton.

onnxruntime/core/session/onnxruntime_c_api.cc
```diff
@@ -3015,6 +3015,32 @@ ORT_API_STATUS_IMPL(OrtApis::Node_GetAttributeByName, _In_ const OrtNode* node,
...
+    return OrtApis::CreateStatus(ORT_INVALID_ARGUMENT, "attr_tensor argument is null");
+  }
+  if (attribute == nullptr) {
+    return OrtApis::CreateStatus(ORT_INVALID_ARGUMENT, "attribute argument is null");
+  }
+
+  const EpNode* ep_node = EpNode::ToInternal(node);
+  if (ep_node == nullptr) {
+    return OrtApis::CreateStatus(OrtErrorCode::ORT_INVALID_ARGUMENT, "node is a ModelEditorNode which doesn't support Node_GetAttributeByName.");
+  }
```
adrianlizarraga: I think this cast to `EpNode` can be removed because the `GetTensorAttributeAsOrtValue` function is now an interface implemented by both `EpNode` and `ModelEditorNode`.

onnxruntime/core/graph/ep_api_types.cc
```diff
@@ -110,12 +110,42 @@ Status EpNode::Create(const Node& node, const EpGraph* ep_graph,
...
+      if (attr->type() == onnx::AttributeProto::TENSOR) {
+        const auto& tensor_proto = attr->t();
+        // Some tensor proto could have no name.
+        // Create a name for that case since we need the name as the key for lookup later.
+        std::string tensor_proto_name = tensor_proto.name();
+        if (tensor_proto.name().empty()) {
```
adrianlizarraga: if the tensor_proto_name is **not** empty, is it guaranteed to be unique in the model?

onnxruntime/core/graph/ep_api_types.cc
chilo-ms: That's a good point. I changed to use on demand approach.

include/onnxruntime/core/providers/utils/ort_graph_to_proto.h
```diff
@@ -758,6 +758,103 @@ static Ort::Status OrtOpAttrToProto(const OrtOpAttr& ort_attr, onnx::AttributePr
...
+      }
+
+      size_t element_count = type_shape_info.GetElementCount();
+      size_t data_bytes = element_count * element_size;
+      const void* data = tensor.GetTensorData<void>();
+
+      // Copy the Ortvalue to TensorProto as raw data
+      tensor_proto.set_raw_data(data, data_bytes);
+
+      *(attr_proto.mutable_t()) = tensor_proto;
```
edgchen1:
```suggestion
      *(attr_proto.mutable_t()) = std::move(tensor_proto);
```

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -6065,6 +6066,21 @@ struct OrtApi {
...
+   * \param[out] attr_tensor Output parameter set to the 'TENSOR' attribute value or nullptr
+   *                         if it's not a 'TENSOR' attribute. Do not cache the OrtValue as
+   *                         it is released when the owning OrtGraph is released.
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \since Version 1.23.
+   */
+  ORT_API2_STATUS(Node_GetTensorAttributeAsOrtValue, _In_ const OrtNode* node, _In_ const OrtOpAttr* attribute,
+                  _Outptr_ const OrtValue** attr_tensor);
```
```suggestion
                  _Outptr_result_maybenull_ const OrtValue** attr_tensor);
```

onnxruntime/core/graph/ep_api_types.cc
edgchen1: do we need to cache them at all? there's the issue of caching on demand requiring a non-const `OrtNode*`. could we just return a newly created OrtValue?

onnxruntime/core/graph/ep_api_types.h
```diff
@@ -216,6 +219,10 @@ struct EpNode : public OrtNode {
   // Helper that gets the execution provider name that this node is assigned to run on.
   const std::string& GetEpName() const;
 
+  // Helper to get the unique name for the 'TENSOR' attribute. Returns empty string if
+  // attribute is not 'TENSOR' type.
+  const std::string GetUniqueTensorAttributeName(const OrtOpAttr* attr) const;
```
edgchen1: typically, we don't want to return const by value. it can be copied into a non-const variable anyway. just return `std::string`.

can it be a non-member helper function that is only in ep_abi_types.cc?

chilo-ms:
> do we need to cache them at all? there's the issue of caching on demand requiring a non-const OrtNode*. could we just return a newly created OrtValue?

Made the API return a newly create OrtValue, and caller now owns it and have to manually clean up by ReleaseValue or use Ort::Value.

onnxruntime/core/graph/ep_api_types.h
chilo-ms: This function is not needed anymore, since we are not doing caching. I just removed it.

onnxruntime/core/graph/ep_api_types.h
```diff
@@ -228,6 +231,7 @@ struct EpNode : public OrtNode {
 
   std::unordered_map<std::string, std::unique_ptr<ONNX_NAMESPACE::AttributeProto>> attributes_map_;
   std::vector<OrtOpAttr*> attributes_;
+  std::unordered_map<std::string, std::unique_ptr<OrtValue>> tensor_attribute_values_;  // The 'TENSOR' Attribute as an OrtValue
```
edgchen1: don't need this anymore right?

onnxruntime/core/graph/ep_api_types.cc
```diff
@@ -248,6 +248,36 @@ Status EpNode::GetAttributes(gsl::span<const OrtOpAttr*> dst) const {
...
+    return ORT_MAKE_STATUS(ONNXRUNTIME, FAIL, "This OrtOpAttr instance is not a 'TENSOR' attribute");
+  }
+
+  auto tensor_proto_to_ort_value = [&](const ONNX_NAMESPACE::TensorProto& tensor_proto,
+                                       OrtValue** result) -> Status {
+    const auto& graph_viewer = ep_graph_->GetGraphViewer();
+
+    // Initialize OrtValue for tensor attribute.
+    // Note: using std::unique_ptr<OrtValue> because we return a OrtValue* to the user and we want it to be stable.
+    auto tensor_attribute_value = std::make_unique<OrtValue>();
```
edgchen1: nit: this comment about unique_ptr seems unnecessary now.

also, I'd slightly prefer keeping the unique_ptr at the function scope instead of in this lambda. we can call unique_ptr::release() at the end of the function and it is obvious that we won't leak before that. perhaps we don't need the lambda and we can just put its logic inline.

onnxruntime/test/ep_graph/test_ep_graph.cc
```diff
@@ -436,6 +469,65 @@ TEST(EpGraphTest, SerializeToProto_ExternalInitializersInMemory) {
...
+    std::string ext_ini_file_path = "constant_of_shape_serialized.bin";
+    std::filesystem::remove(ext_ini_file_path);
+    std::ofstream ext_ini_ofs(ext_ini_file_path, std::ios::binary);
+    auto handle_initializer_data = [&ext_ini_ofs, &ext_ini_file_path](const OrtValueInfo* value_info,
+                                                                      const void* data, size_t bytes,
+                                                                      bool& is_external, std::string& location,
+                                                                      int64_t& offset) -> Ort::Status {
+      // OrtValueInfo* could be used to query initializer's name, type, shape,
+      // node consumers, etc.
+      (void)value_info;
```
edgchen1: nit
```suggestion
      static_cast<void>(value_info);
```

onnxruntime/core/graph/ep_api_types.cc
```diff
@@ -248,6 +248,36 @@ Status EpNode::GetAttributes(gsl::span<const OrtOpAttr*> dst) const {
   return Status::OK();
 }
 
+Status EpNode::GetTensorAttributeAsOrtValue(const OrtOpAttr* attribute, OrtValue** result) const {
+  const auto* attr_proto = reinterpret_cast<const ONNX_NAMESPACE::AttributeProto*>(attribute);
+
+  if (attr_proto->type() != onnx::AttributeProto::TENSOR) {
+    *result = nullptr;
```
edgchen1: it seems a little inconsistent where `*result` is set to nullptr. e.g., it doesn't get set for other error cases. more ideally, we would always set it to nullptr or a valid value, or only set it to a valid value if the function is successful.

onnxruntime/core/session/onnxruntime_c_api.cc
```diff
@@ -3018,6 +3018,27 @@ ORT_API_STATUS_IMPL(OrtApis::Node_GetAttributeByName, _In_ const OrtNode* node,
...
+  if (attribute == nullptr) {
+    return OrtApis::CreateStatus(ORT_INVALID_ARGUMENT, "attribute argument is null");
+  }
+
+  const auto& tensor_proto = reinterpret_cast<const ONNX_NAMESPACE::AttributeProto*>(attribute)->t();
+
+  ORT_ENFORCE(utils::HasDataType(tensor_proto), "Tensor proto doesn't have data type.");
+  ORT_ENFORCE(ONNX_NAMESPACE::TensorProto::DataType_IsValid(tensor_proto.data_type()), "Tensor proto has invalid data type.");
+  ORT_ENFORCE(!utils::HasExternalData(tensor_proto),
+              "Tensor proto with external data for value attribute is not supported.");
```
adrianlizarraga: This is accessing `(attribute)->t()` without first checking that the `AttributeProto` is actually a tensor type.

Perhaps it would make sense to move all of these `ORT_ENFORCE` checks inside `GetTensorAttributeAsOrtValue`, which is already checking that it is a tensor attribute:

```c++
Status EpNode::GetTensorAttributeAsOrtValue(const OrtOpAttr* attribute, OrtValue** result) const {
  const auto* attr_proto = reinterpret_cast<const ONNX_NAMESPACE::AttributeProto*>(attribute);

  if (attr_proto->type() != onnx::AttributeProto::TENSOR) {
    return ORT_MAKE_STATUS(ONNXRUNTIME, FAIL, "This OrtOpAttr instance is not a 'TENSOR' attribute");
  }

  const auto& graph_viewer = ep_graph_->GetGraphViewer();
  const auto& tensor_proto = attr_proto->t();

  // Check that TensorProto is valid.
  ORT_RETURN_IF_NOT(utils::HasDataType(tensor_proto), "Tensor proto doesn't have data type.");
  ORT_RETURN_IF_NOT(ONNX_NAMESPACE::TensorProto::DataType_IsValid(tensor_proto.data_type()), "Tensor proto has invalid data type.");
  ORT_RETURN_IF(utils::HasExternalData(tensor_proto),
                "Tensor proto with external data for value attribute is not supported.");

  // Initialize OrtValue for tensor attribute.
  auto tensor_attribute_value = std::make_unique<OrtValue>();
  AllocatorPtr tensor_attribute_allocator = CPUAllocator::DefaultInstance();
  ORT_RETURN_IF_ERROR(utils::TensorProtoToOrtValue(Env::Default(), graph_viewer.ModelPath(), tensor_proto,
                                                   tensor_attribute_allocator, *tensor_attribute_value));

  *result = tensor_attribute_value.release();
  return Status::OK();
}
```

chilo-ms: good catch, okay, it makes sense to me to move the tensor attributes checks into `EpNode::GetTensorAttributeAsOrtValue`.

include/onnxruntime/core/providers/utils/ort_graph_to_proto.h
```diff
@@ -758,6 +758,103 @@ static Ort::Status OrtOpAttrToProto(const OrtOpAttr& ort_attr, onnx::AttributePr
...
+      for (auto& dim : shape) {
+        tensor_proto.add_dims(dim);
+      }
+
+      size_t element_count = type_shape_info.GetElementCount();
+      size_t data_bytes = element_count * element_size;
+      const void* data = tensor.GetTensorData<void>();
+
+      // Copy the Ortvalue to TensorProto as raw data
+      tensor_proto.set_raw_data(data, data_bytes);
```
edgchen1, line 853: should we ensure that the data is little-endian before we copy it directly? TensorProto's raw_data requires this.

https://github.com/onnx/onnx/blob/ee724f6635ed6c1e81890c9f120ac360000c5370/onnx/onnx.in.proto#L719-L720

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -6065,6 +6066,20 @@ struct OrtApi {
   ORT_API2_STATUS(Node_GetAttributeByName, _In_ const OrtNode* node, _In_ const char* attribute_name,
                   _Outptr_result_maybenull_ const OrtOpAttr** attribute);
 
+  /** \brief Get the OrtNode's 'TENSOR' attribute as an OrtValue.
+   *
+   * \param[in] node The OrtNode instance.
+   * \param[in] attribute The OrtOpAttr instance.
+   * \param[out] attr_tensor Returns the newly created OrtValue if it's a 'TENSOR' attribute.
```
edgchen1: maybe indicate that the output OrtValue is only created if the function is successful. e.g.,
```suggestion
   * \param[out] attr_tensor If successful, contains the 'TENSOR' attribute as a newly created OrtValue.
```

adrianlizarraga (approved):

onnxruntime/core/graph/abi_graph_types.h
```diff
@@ -251,6 +251,16 @@ struct OrtNode {
...
 
+  /// <summary>
+  /// Gets the node's 'TENSOR' attribute as an OrtValue.
+  /// </summary>
+  /// <param name="attr">Node's 'TENSOR' attribute.</param>
+  /// <param name="value">Output parameter set to the 'TENSOR' attribute value or nullptr
+  ///                     if it's not a 'TENSOR' attribute.</param>
+  /// <returns>A status indicating success or an error.</returns>
+  virtual onnxruntime::Status GetTensorAttributeAsOrtValue(const OrtOpAttr* attr,
+                                                           OrtValue** value) const = 0;
```
edgchen1: update "value" comment - `*value` will no longer get set to nullptr.

also, maybe consider `OrtValue*& value` so we don't have to worry about `value` being nullptr within this function.

include/onnxruntime/core/providers/utils/ort_graph_to_proto.h
edgchen1: `std::endian` would be useful, but it was introduced in c++20.

could use something like this:

https://github.com/microsoft/onnxruntime/blob/c22f70d48e012de2a480d6ce2ef9920f320f3f22/include/onnxruntime/core/framework/endian.h#L8-L25

and check that `endian::native == endian::little`

adrianlizarraga: Good point. I would have thought that `TensorProtoToOrtValue` would store data in little-endian, but I could be wrong. (edit:  the OrtValue should store data in the machine's native endianness, so we do need to convert to little endian here)

In any case, this `ort_graph_to_proto.h` file is not really tied to the release as it is a utility file that EP implementers can copy and modify. Perhaps, we (maybe me) can address this separately since I think I made the same potential mistake elsewhere in this file.

chilo-ms: The data is from OrtValue created by `utils::TensorProtoToOrtValue` and it ended up calling `TensorProtoToTensor`.
Looking at `TensorProtoToTensor`, it gets the raw data from the raw_data field of the source tensor proto if it has it, so we can ensure the raw data is little-endian for this case. 
But i'm wondering why `TensorProtoToTensor` doesn't have the code to read the value from other fields, ex: float_data, int32_data ...?

edgchen1: > I would have thought that TensorProtoToOrtValue would store data in little-endian

if I understand correctly, `OrtValue`'s data will be in native byte order. perhaps most of the time, the native byte order is little-endian. but `TensorProto raw_data` is specified to be little-endian.

>  i'm wondering why TensorProtoToTensor doesn't have the code to read the value from other fields, ex: float_data, int32_data

I think it does support reading from other fields. it's not so easy to follow, with all the macros...

https://github.com/microsoft/onnxruntime/blob/b9575476e94daa9c6578aba92d8f04324dd15815/onnxruntime/core/framework/tensorprotoutils.cc#L520-L526

edgchen1: sure, we can address this later

adrianlizarraga: Yeah, it makes sense that the OrtValue data is in the native byte order (it has to be). We do have to convert to little-endian here. I can take this on.

adrianlizarraga (approved):

PR #24282, opened 2025-04-02
[WebNN EP] Automatically use ml-tensor for outputs

egalli:
### Description
If it would improve performance, this patch moves outputs to MLTensor backed Tensors.

### Motivation and Context
We are currently performing an extra copy on output tensors located in the CPU when using the WebNN EP (MLTensor -(copy)-> wasm heap -(copy)-> JS). This patch removes this copy by moving the readback to JS instead of wasm. As an extra benefit, we can also start the readbacks and wait for them in parallel.

This change is similar to #23073

PR #25571, opened 2025-07-29
[Web] Avoid unnecessary data copy for pre-allocated tensors

Honry:
### Description
Ensure all pre-allocated tensors do not trigger unnecessary data copying. e.g. the WebNN EP always binds its tensor to 'ml-tensor'. In such cases, the tensor ID might change after binding, but copying data for these tensors should still be avoided.

### Motivation and Context
This improves efficiency and avoids redundant operations.

### DIFF
--- a/js/web/lib/wasm/wasm-core-impl.ts
+++ b/js/web/lib/wasm/wasm-core-impl.ts
@@ -725,6 +725,7 @@ export const run = async (
   const inputTensorHandles: number[] = [];
   const outputTensorHandles: number[] = [];
   const inputOutputAllocs: number[] = [];
+  const preAllocatedOutputs: number[] = [];
 
   const beforeRunStack = wasm.stackSave();
   const inputValuesOffset = wasm.stackAlloc(inputCount * ptrSize);
@@ -797,7 +798,8 @@ export const run = async (
         const location = outputTensors[i]?.[3]; // undefined means output is not pre-allocated.
 
         if (location) {
-          // output is pre-allocated. bind the tensor.
+          // output is pre-allocated, store and bind the tensor.
+          preAllocatedOutputs.push(outputTensorHandles[i]);
           const errorCode = wasm._OrtBindOutput(handle, outputNamesUTF8Encoded[index], outputTensorHandles[i], 0);
           if (errorCode !== 0) {
             checkLastError(`Can't bind pre-allocated output[${i}] for session=${sessionId}.`);
@@ -861,7 +863,10 @@ export const run = async (
     TRACE_EVENT_BEGIN('wasm ProcessOutputTensor');
     for (let i = 0; i < outputCount; i++) {
       const tensor = Number(wasm.getValue(outputValuesOffset + i * ptrSize, '*'));
-      if (tensor === outputTensorHandles[i]) {
+      // Certain pre-allocated tensors may already be bound in the IO binding. e.g. the WebNN backend
+      // always binds its tensor to 'ml-tensor'. In such cases, the tensor ID might change after binding,
+      // but copying data for these tensors should still be avoided.
+      if (tensor === outputTensorHandles[i] || preAllocatedOutputs.includes(outputTensorHandles[i])) {
         // output tensor is pre-allocated. no need to copy data.
         output.push(outputTensors[i]!);
         continue;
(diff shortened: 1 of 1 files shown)


Honry:
@fs-eire, PTAL, thanks!

fs-eire:
I may need to understand more details about why in this case, `tensor` is not equal to `outputTensorHandles[i]` but they are the same bound output. Is this a WebNN only case? In my understanding, if `tensor !== outputTensorHandles[i]`, we should release `tensor` otherwise it's a memory leak.

Honry:
> I may need to understand more details about why in this case, `tensor` is not equal to `outputTensorHandles[i]` but they are the same bound output. Is this a WebNN only case?

Currently this is WebNN only case, usually if we use pre-allocated output tensors, the `ioBindingState` will be `null`. After this [PR](https://github.com/microsoft/onnxruntime/pull/24282/), WebNN always bind its outputs to 'ml-tensor', therefore the `ioBindingState` is not `null` and then it will always call the `wasm._OrtRunWithBinding()` (which leads to the `tensor !== outputTensorHandles[i]`)

>  In my understanding, if tensor !== outputTensorHandles[i], we should release tensor otherwise it's a memory leak.

 So I need to release this `tensor` early?

fs-eire:
> > I may need to understand more details about why in this case, `tensor` is not equal to `outputTensorHandles[i]` but they are the same bound output. Is this a WebNN only case?
> 
> Currently this is WebNN only case, usually if we use pre-allocated output tensors, the `ioBindingState` will be `null`. After this [PR](https://github.com/microsoft/onnxruntime/pull/24282/), WebNN always bind its outputs to 'ml-tensor', therefore the `ioBindingState` is not `null` and then it will always call the `wasm._OrtRunWithBinding()` (which leads to the `tensor !== outputTensorHandles[i]`)
> 
> > In my understanding, if tensor !== outputTensorHandles[i], we should release tensor otherwise it's a memory leak.
> 
> So I need to release this `tensor` early?

For `OrtRunWithBinding`, the `tensor` should be the same to `outputTensorHandles[i]`:

https://github.com/microsoft/onnxruntime/blob/131cf40dc25a1611e239b782eb5239cc4921f34e/onnxruntime/wasm/api.cc#L605-L607

Honry:
> > > I may need to understand more details about why in this case, `tensor` is not equal to `outputTensorHandles[i]` but they are the same bound output. Is this a WebNN only case?
> > 
> > 
> > Currently this is WebNN only case, usually if we use pre-allocated output tensors, the `ioBindingState` will be `null`. After this [PR](https://github.com/microsoft/onnxruntime/pull/24282/), WebNN always bind its outputs to 'ml-tensor', therefore the `ioBindingState` is not `null` and then it will always call the `wasm._OrtRunWithBinding()` (which leads to the `tensor !== outputTensorHandles[i]`)
> > > In my understanding, if tensor !== outputTensorHandles[i], we should release tensor otherwise it's a memory leak.
> > 
> > 
> > So I need to release this `tensor` early?
> 
> For `OrtRunWithBinding`, the `tensor` should be the same to `outputTensorHandles[i]`:
> 
> https://github.com/microsoft/onnxruntime/blob/131cf40dc25a1611e239b782eb5239cc4921f34e/onnxruntime/wasm/api.cc#L605-L607

Oh, that's odd, as I tested `tensor !== outputTensorHandles[i]` after `OrtRunWithBinding`, I will do more debug.

Honry:
> Oh, that's odd, as I `tested tensor !== outputTensorHandles[i]` after `OrtRunWithBinding`, I will do more debug.

@fs-eire, after further debugging, I find there're two places change the address of the `tensor`.

1. `wasm._OrtBindOutput` -> `BindOutputImpl` https://github.com/microsoft/onnxruntime/blob/e57dc2a2e5fc366a6391598082155859aebdb3e7/onnxruntime/core/session/IOBinding.cc#L94
2. `wasm._OrtRunWithBinding` -> `GetBoundOutputValues`
https://github.com/microsoft/onnxruntime/blob/e57dc2a2e5fc366a6391598082155859aebdb3e7/onnxruntime/core/session/onnxruntime_c_api.cc#L1018

Honry:
@fs-eire, friendly ping.

fs-eire:
> @fs-eire, friendly ping.

please merge to latest main branch and add comments in line 866 "// TODO: revisit this part to ensure it works for WebGPU when both pre-allocated outputs and preferred location are specified"

Honry:
> > @fs-eire, friendly ping.
> 
> please merge to latest main branch and add comments in line 866 "// TODO: revisit this part to ensure it works for WebGPU when both pre-allocated outputs and preferred location are specified"

@fs-eire, done.

fs-eire (approved):

Honry:
@fs-eire, one unexpected failure CI, do I need to re-merge to latest main branch?

PR #25578, opened 2025-07-29
[webgpu] Optimize dp4 prefill shader for Qualcomm

qjia7:
This change uses subgroupShuffle for sg_size=64 to perform the matmul. It also uses a loop instead of loop unrolling to reduce the register pressure.

Phi4 prefill for 1K tokens becomes 8.8s from 11.32s on Qualcomm Adreno X1-85 GPU.


onnxruntime/contrib_ops/webgpu/quantization/dp4a_matmul_nbits.cc
```diff
@@ -266,18 +266,24 @@ Status DP4AMatMulNBitsProgram::GenerateShaderCode(ShaderHelper& shader) const {
...
+        var subtile_id = u32(local_idx / subtile_size);
+        var subtile_idx = u32(subtile_id / 4);
+        var subtile_idy = u32(subtile_id % 4);
+        var base_A = subtile_idx * 16;
+        var base_B = subtile_idy * 16;
         // For each subtile we have 16 threads assigned.
-        let a_idx = u32(local_idx % subtile_size);
+        var a_idx = u32(local_idx % subtile_size);
+
+        if (sg_size == 64) {
```
sushraja-msft: Would it be possible to pull this shader out into a template file before making this change ?

onnxruntime/contrib_ops/webgpu/quantization/dp4a_matmul_nbits.cc
```diff
@@ -266,18 +266,24 @@ Status DP4AMatMulNBitsProgram::GenerateShaderCode(ShaderHelper& shader) const {
...
+          subtile_id = u32(local_idx / sg_size);
+          subtile_idx = u32(subtile_id / 4);
+          subtile_idy = u32(subtile_id % 4);
+          base_A = subtile_idx * sg_size;
+          base_B = subtile_idy * 16;
+          a_idx = sg_id;
+        }
```
sushraja-msft: At sg_size 64 the whole algorithm can change. The full tile size is just 64x64. Every single own_a, own_b that is curretly loaded into shared memory is accessible via subgroup shuffle.

Either we can have load_A/load_B just keep the results in registry and use subgroup shuffle or we can change the tile size to be larger to actually leverage cycling through shared memory.

onnxruntime/contrib_ops/webgpu/quantization/dp4a_matmul_nbits.cc
```diff
@@ -266,18 +266,24 @@ Status DP4AMatMulNBitsProgram::GenerateShaderCode(ShaderHelper& shader) const {
...
 
-        var lane_output1: vec4<output_element_t>;
-        var lane_output2: vec4<output_element_t>;
-        var lane_output3: vec4<output_element_t>;
-        var lane_output4: vec4<output_element_t>;
+        var lane_outputs: array<output_element_t, 16>;
```
sushraja-msft: the reason I had lane_outputs split into lane_output1, lane_output2 etc was because I believe I noticed intel would not place lane_outputs in registers if it was a single array requiring consecutive 16 floats of space.

If we are changing it we would need to measure perf on Alderlake/Tigerlake with old and new drivers to ensure no regression.

qjia7:
Will refactor it based on #25724 and constraint the changes to QC.

qjia7:
Rebase is done. Please take another look, thanks.

guschmue (approved):

onnxruntime/contrib_ops/webgpu/quantization/dp4a_matmul.wgsl.template
```diff
@@ -119,18 +120,30 @@ $MAIN {
...
+    var base_A = subtile_idx * 16;
+    var base_B = subtile_idy * 16;
     // For each subtile we have 16 threads assigned.
-    let a_idx = u32(local_idx % subtile_size);
+    var a_idx = u32(local_idx % subtile_size);
 
+#if is_qualcomm
+    if (sg_size == 64) {
+        subtile_id = u32(local_idx / sg_size);
+        subtile_idx = u32(subtile_id / 4);
```
sushraja-msft, line 158: // subtile_idx is always 0
// subtile_idy is one of {0,1,2,3} 
// The subtile is now rectangular for qualcomm case, this way we don't need to increase the number of
// lane_output each thread needs to track. That is if we want to use a subtile that is 64x64 we would need 
// var lane_outputs: array<output_element_t, 64>;

onnxruntime/contrib_ops/webgpu/quantization/dp4a_matmul.wgsl.template
```diff
@@ -154,6 +167,34 @@ $MAIN {
         var own_scale_a: output_element_t = scale_A[base_A + a_idx];
 
 #if has_zero_points && n_bits == 8
+#if is_qualcomm
```
sushraja-msft: indent this preprocessor block by 1.

sushraja-msft (dismissed):

onnxruntime/contrib_ops/webgpu/quantization/dp4a_matmul.wgsl.template
```diff
@@ -4,6 +4,7 @@
 #param block_size
 #param n_bits
 #param has_zero_points
+#param is_qualcomm
```
sushraja-msft, line 7: signed off but wondering, can this be made as generic sg_size == 64 support, without is_qualcomm.

the 

else
        {
            for (var i = 0u; i < 16u; i++)
            {
                lane_outputs[i] += SDP8AI(own_a0, tile_B[0][base_B + i], own_a1, tile_B[1][base_B + i],  own_scale_a * scale_B[base_B + i], zeroes[base_B + i]);
            }
        }
        
blocks can be eliminated that way.

sushraja-msft: I think I remember now, you wanted lane_output to be an array which doesnt work for intel.

guschmue (approved):

sushraja-msft (approved):

PR #25255, opened 2025-07-02
[Mlas] optimize MlasConv using thread partition opt

zoeczy:
### Description

This PR enhances `MlasConv` in ONNX Runtime by introducing a thread partitioning strategy based on Batch Size (`bs`) and Group Count (`group`). This change improves multi-threading efficiency in convolution scenarios where scaling with core/thread count was previously limited.

The PR also includes updates to the `bench_sconv` utility to support and evaluate multi-threaded performance under the new partitioning strategy.
* Command to run multi-threaded benchmarks under core binding: `numactl -C core_num0-core_num_1 ./onnxruntime_mlas_benchmark --benchmark_filter=Teams`
* The following results demonstrate the performance improvement of the optimized MlasConv under a 4-thread configuration:
image
Compared to the current master implementation, the optimized version exhibits nearly 3× performance improvement, showing effective scaling with thread count. In contrast, the master branch shows no meaningful performance gain when increasing the number of threads, due to insufficient parallelization in the original implementation.

### Motivation and Context

- Previously, `MlasConv` exhibited minimal performance gains when increasing the number of threads or CPU cores in scenarios with small batch sizes or grouped convolutions.
- This change introduces a finer-grained workload distribution across threads by splitting the work along batch and group dimensions.
- Benchmarks using `bench_sconv` show a noticeable performance improvement in multi-threaded runs, especially on multi-core CPUs.
#### Releated Issues
https://github.com/microsoft/onnxruntime/issues/25152

PR #25580, opened 2025-07-29
NEON kernels for NCHWc Convolution and Pooling

### LINKED ISSUE #24790: [Feature Request] Optimized Convolution kernels for aarch64

Similar to how x86 has optimized convolution kernels written using AVX, for example - [SconvKernelAvx512F.S](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/mlas/lib/x86_64/SconvKernelAvx512F.S), it would be nice to have SVE/NEON convolution kernels for aarch64.
This would greatly benefit ONNX users running CV models on Mac M-series chips, NVIDIA Grace, AWS Graviton, Fujitsu-Monaka, etc.
I want to understand from Arm ISA experts if this is a planned activity or not; @MaajidKhan @milpuz01  @chenfucn  @snnn  @yufenglee  

Models containing convolutional operators will benefit from these optimized kernels if they are running on aarch64 machines.

Rohanjames1997:
### Description
This PR implements optimized Arm NEON kernels for NCHWc (channels-last with channel blocking) convolution and pooling operations in MLAS, significantly improving performance on Arm64 platforms. 

### Motivation and Context
Fixes #24790 

The new NCHWc kernels improve performance by 5-6x, depending on the configuration of threads, model, etc.
For example, here is the performance gain witnessed during mobilenet inference: Focus on the "Number of inferences per second" (93 inf/s -> 498 inf/s)

<details>
  <summary>System configuration</summary>

```
Architecture:             aarch64
  CPU op-mode(s):         64-bit
  Byte Order:             Little Endian
CPU(s):                   64
  On-line CPU(s) list:    0-63
Vendor ID:                ARM
  Model name:             Neoverse-V2
    Model:                1
    Thread(s) per core:   1
    Core(s) per socket:   64
    Socket(s):            1
    Stepping:             r0p1
    BogoMIPS:             2000.00
    Flags:                fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics fphp asimdhp cpuid asimdrdm jscvt fcma lrcpc dcpop sha3 asimddp sha512 sve asimdfhm dit uscat ilrcpc flagm ssbs sb paca pacg dcpodp 
                          sve2 sveaes svepmull svebitperm svesha3 flagm2 frint svei8mm svebf16 i8mm bf16 dgh rng bti
Caches (sum of all):      
  L1d:                    4 MiB (64 instances)
  L1i:                    4 MiB (64 instances)
  L2:                     128 MiB (64 instances)
  L3:                     36 MiB (1 instance)
NUMA:                     
  NUMA node(s):           1
  NUMA node0 CPU(s):      0-63
Vulnerabilities:          
  Gather data sampling:   Not affected
  Itlb multihit:          Not affected
  L1tf:                   Not affected
  Mds:                    Not affected
  Meltdown:               Not affected
  Mmio stale data:        Not affected
  Reg file data sampling: Not affected
  Retbleed:               Not affected
  Spec rstack overflow:   Not affected
  Spec store bypass:      Mitigation; Speculative Store Bypass disabled via prctl
  Spectre v1:             Mitigation; __user pointer sanitization
  Spectre v2:             Not affected
  Srbds:                  Not affected
  Tsx async abort:        Not affected
```
</details>
<details>
  <summary>Perf with current upstream kernels</summary>

```
./build/Linux/Release/onnxruntime_perf_test -x 32 -I -m times -r 1000 ~/scripts/mobilenet.onnx

Setting intra_op_num_threads to 32
Session creation time cost: 0.0238608 s
First inference time cost: 11 ms
Total inference time cost: 10.7458 s
Total inference requests: 1000
Average inference time cost: 10.7458 ms
Total inference run time: 10.7465 s
Number of inferences per second: 93.0534 
Avg CPU usage: 50 %
Peak working set size: 70410240 bytes
Avg CPU usage:50
Peak working set size:70410240
Runs:1000
Min Latency: 0.0106707 s
Max Latency: 0.0113617 s
P50 Latency: 0.0107453 s
P90 Latency: 0.0107695 s
P95 Latency: 0.0107785 s
P99 Latency: 0.0107965 s
P999 Latency: 0.0113617 s
```
</details>
<details>
  <summary>Perf with NCHWc kernels</summary>

```
./build/Linux/Release/onnxruntime_perf_test -x 32 -I -m times -r 1000 ~/scripts/mobilenet.onnx

Setting intra_op_num_threads to 32
Session creation time cost: 0.0358121 s
First inference time cost: 2 ms
Total inference time cost: 2.00561 s
Total inference requests: 1000
Average inference time cost: 2.00561 ms
Total inference run time: 2.00607 s
Number of inferences per second: 498.488 
Avg CPU usage: 50 %
Peak working set size: 92467200 bytes
Avg CPU usage:50
Peak working set size:92467200
Runs:1000
Min Latency: 0.00198387 s
Max Latency: 0.00204784 s
P50 Latency: 0.00200537 s
P90 Latency: 0.0020155 s
P95 Latency: 0.00201822 s
P99 Latency: 0.0020251 s
P999 Latency: 0.00204784 s
```
</details>

Happy to run further performance tests as required.


Rohanjames1997:

Rohanjames1997:
@skottmckay @snnn @yufenglee, appreciate it if the CI can be triggered. Thanks!

hariharans29:
I think there are some build failures because some fields in the MLAS struct are only made available on Linux and they are referenced without proper ifdefs elsewhere ? I get similar build breaks on Windows when I try to pull this change and try it.

As an aside, is there any special intrinsic being used that works only on Linux. Can this be enabled on all platforms supporting NEON ?

onnxruntime/core/mlas/lib/sconv.h
```diff
@@ -0,0 +1,41 @@
...
+
+#include <cstddef>
+
+//
+// Define the calling convention for MLAS functions.
+//
+
+#ifndef MLASCALL
+#if defined(_WIN32) && !defined(_WIN64)
+#define MLASCALL __stdcall
```
hariharans29: Isn't this already defined in mlas.h ?

onnxruntime/core/mlas/lib/sconv_kernel_neon.cpp
```diff
@@ -0,0 +1,567 @@
...
+
+Abstract:
+
+    This module implements the single precision convolution kernels for ARM NEON.
+
+--*/
+
+#include "sconv.h"
+
+#if defined(__aarch64__) || defined(_M_ARM64)
```
hariharans29: Use MLAS_TARGET_ARM64 ?

onnxruntime/core/mlas/lib/spool_kernel_neon.cpp
```diff
@@ -0,0 +1,298 @@
...
+
+    spool_kernel_neon.cpp
+
+Abstract:
+
+    This module implements the single precision pooling kernels for ARM NEON.
+
+--*/
+
+#if defined(__aarch64__) || defined(_M_ARM64)
```
hariharans29: Use MLAS_TARGET_ARM64 ?

onnxruntime/core/mlas/lib/sconv.h
Rohanjames1997: Ack, removed.

onnxruntime/core/mlas/lib/sconv_kernel_neon.cpp
Rohanjames1997: Done

onnxruntime/core/mlas/lib/spool_kernel_neon.cpp
Rohanjames1997: Done

Rohanjames1997:
Thanks for the review @hariharans29!


I initially tried to restrict to Linux alone (but did not add proper ifdefs).
I will now enable it for Windows and other platforms as well through `MLAS_TARGET_ARM64`

onnxruntime/core/mlas/lib/sconv_kernel_neon.cpp
```diff
@@ -0,0 +1,567 @@
...
+                        const float* input_row_end = input_row_start + InputWidthElements;
+
+                        if (input_element >= input_row_start && input_element < input_row_end) {
+                            input_values[i] = *input_element;
+                        } else {
+                            input_values[i] = 0.0f;
+                        }
+                    }
+                    InputVector0 = MlasLoadFloat32x4(input_values);
+                }
```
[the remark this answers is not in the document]
Rohanjames1997: Done

hariharans29:
> Thanks for the review @hariharans29!
> 
> I initially tried to restrict to Linux alone (but did not add proper ifdefs). I will now enable it for Windows and other platforms as well through `MLAS_TARGET_ARM64`

Thanks. Could you please go over the remaining Copilot comments and see if they are relevant and please "resolve" them. Thanks for this excellent contribution!

onnxruntime/core/mlas/lib/mlasi.h
```diff
@@ -1332,6 +1339,14 @@ struct MLAS_PLATFORM {
...
     const MLAS_GEMM_QUANT_DISPATCH* GemmU8S8Dispatch;
     const MLAS_GEMM_QUANT_DISPATCH* GemmS8S8Dispatch;
+#endif
+#if defined(__aarch64__) && defined(__linux__)
+    MLAS_CONV_FLOAT_KERNEL* ConvNchwFloatKernel;
+    MLAS_CONV_FLOAT_KERNEL* ConvNchwcFloatKernel;
+    MLAS_CONV_DEPTHWISE_FLOAT_KERNEL* ConvDepthwiseFloatKernel;
+    MLAS_CONV_POINTWISE_FLOAT_KERNEL* ConvPointwiseFloatKernel;
+    MLAS_POOL_FLOAT_KERNEL* PoolFloatKernel[MlasPoolingKindCount];
+    uint32_t NchwcBlockSize;
```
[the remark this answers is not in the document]
Rohanjames1997: Done.

onnxruntime/core/mlas/lib/mlasi.h
```diff
@@ -949,6 +949,13 @@ extern "C" {
 #if defined(__aarch64__) && defined(__linux__)
     MLAS_SBGEMM_FLOAT_KERNEL MlasSbgemmKernelZero;
     MLAS_SBGEMM_FLOAT_KERNEL MlasSbgemmKernelAdd;
+    MLAS_CONV_FLOAT_KERNEL MlasConvNchwFloatKernelNeon;
+    MLAS_CONV_FLOAT_KERNEL MlasConvNchwcFloatKernelNeon;
+    MLAS_CONV_DEPTHWISE_FLOAT_KERNEL MlasConvDepthwiseFloatKernelNeon;
+    MLAS_CONV_POINTWISE_FLOAT_KERNEL MlasConvPointwiseFloatKernelNeon;
+    MLAS_POOL_FLOAT_KERNEL MlasPoolMaximumFloatKernelNeon;
+    MLAS_POOL_FLOAT_KERNEL MlasPoolAverageExcludePadFloatKernelNeon;
+    MLAS_POOL_FLOAT_KERNEL MlasPoolAverageIncludePadFloatKernelNeon;
```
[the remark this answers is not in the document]
Rohanjames1997: Done.

cmake/onnxruntime_mlas.cmake
```diff
@@ -449,6 +449,8 @@ else()
             ${MLAS_SRC_DIR}/qgemm_kernel_smmla.cpp
             ${MLAS_SRC_DIR}/qgemm_kernel_ummla.cpp
             ${MLAS_SRC_DIR}/sbgemm_kernel_neon.cpp
+            ${MLAS_SRC_DIR}/sconv_kernel_neon.cpp
```
hariharans29: I think this should be added to the Windows/Apple sources list as well ?

Rohanjames1997: Sure, I can move it a few lines above, outside of the `if (NOT APPLE)`, and also add it to `setup_mlas_source_for_windows`. 

Fyi, the CI is the only way I have to test Windows/Apple. Thanks for running the CI so promptly!

Rohanjames1997: Done.

onnxruntime/core/mlas/lib/spool_kernel_neon.cpp
```diff
@@ -0,0 +1,298 @@
...
+                    SumVector0 = MlasAddFloat32x4(SumVector0, InputVector0);
+                    SumVector1 = MlasAddFloat32x4(SumVector1, InputVector1);
+                    SumVector2 = MlasAddFloat32x4(SumVector2, InputVector2);
+                    SumVector3 = MlasAddFloat32x4(SumVector3, InputVector3);
+                }
+            }
+        }
+
+        if (ExcludePad) {
+            std::vector<float> results(BlockSize);
```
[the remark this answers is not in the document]
Rohanjames1997: Done

onnxruntime/core/mlas/lib/spool_kernel_neon.cpp
```diff
@@ -0,0 +1,298 @@
...
+                    SumVector2 = MlasAddFloat32x4(SumVector2, InputVector2);
+                    SumVector3 = MlasAddFloat32x4(SumVector3, InputVector3);
+
+                    if (ExcludePad) {
+                        for (size_t i = 0; i < BlockSize; i++) {
+                            valid_count[i]++;
+                        }
+                    }
+                } else {
+                    std::vector<float> values(BlockSize);
```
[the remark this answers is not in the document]
Rohanjames1997: Done

onnxruntime/core/mlas/lib/spool_kernel_neon.cpp
```diff
@@ -0,0 +1,298 @@
...
+                    for (size_t i = 0; i < BlockSize; i++) {
+                        const float* element_ptr = input_ptr + i;
+                        if (element_ptr >= row_start && element_ptr < row_end) {
+                            values[i] = *element_ptr;
+                        } else {
+                            values[i] = MaxPaddingValue;
+                        }
+                    }
```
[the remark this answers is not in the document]
Rohanjames1997: Done.

cmake/onnxruntime_mlas.cmake
hariharans29: I understand your constraint. I am happy to run CI as many times as you need. Btw - there is a conflict - can you please resolve it ?

Rohanjames1997: Yup! Resolved

hariharans29:
I think there are some low hanging fruit build failures that need some minor fixes ?

Rohanjames1997:
Thanks, yes, it seems that `BlockSize` needs to be `constexpr` instead of `const`. 

Since `MlasNchwcGetBlockSize()` cannot return a `constexpr`, I think the fix for this would be to hardcode the `BlockSize` to be `16` in both the kernel files, as the implementation certainly expects it to be `16`.

Similar to how it's done for the x86 backend [here](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/mlas/lib/x86_64/SconvKernelAvx512F.S#L340-L342)

Let me know your thoughts.

hariharans29:
> Thanks, yes, it seems that `BlockSize` needs to be `constexpr` instead of `const`.
> 
> Since `MlasNchwcGetBlockSize()` cannot return a `constexpr`, I think the fix for this would be to hardcode the `BlockSize` to be `16` in both the kernel files, as the implementation certainly expects it to be `16`.
> 
> Let me know your thoughts.



> MlasNchwcGetBlockSize

Just thinking out loud - can we have a vector of size `BlockSize` and use its buffer to pass through to `MlasLoadFloat32x4()` ? But that is a dynamic memory allocation - so it may affect perf...

If we hardcode the value here - we would have to remember to update it if it ever changes ...

Rohanjames1997:
True, I see your point. Even on the x86 backend, the values is indeed [hardcoded](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/mlas/lib/x86_64/SconvKernelAvx512F.S#L340-L342) and it enables the compiler to unroll loops efficiently.

hariharans29:
Got it - then we can go with hard-coded and add a comment there about the hard-coded value being the same as NCHWC Block size ? Thanks.

Rohanjames1997:
Sure! I can do that. Thanks.

Fyi: For both pooling and conv files, I will be using the same blocksize. So, I am adding the blocksize to `sconv.h` and importing that header in both the conv and pooling kernels.

Rohanjames1997:
Hopefully the CI passes this time 🤞 
I will handle any remaining failures later tonight/tomorrow - thanks again!

hariharans29:
CI looks good so far. The binary size check fails indicating that the binary size has grown above the threshold we had set previously. I will discuss internally and get back on the best steps to remedy that (either increase the threshold or disable this feature on minimal builds).

Rohanjames1997:
Great! Thank you!

Rohanjames1997:
Looks like all required external CI passed 🎉 
Were there any failures internally?

hariharans29:
> Looks like all required external CI passed 🎉 Were there any failures internally?

Are you referring to the release pipelines ? Unfortunately, I wouldn't know. I think they run at a specific cadence like once a day (not for all PRs like CIs) after the merge to main happens. In any case, I wouldn't be able to trigger ad-hoc runs against a fork. Although not ideal, like last time, we will let you know if something goes wrong post-merge.

Rohanjames1997:
Sure thing, that works!

hariharans29:
Could you please increase the binary size limit here (in 2 lines) so that the additional binary growth can be accommodated  - https://github.com/microsoft/onnxruntime/blob/f418a44c40b86676049951b393f69594c84c0ca4/.github/workflows/android.yml#L74. I am thinking since these are ARM64 kernels they are relevant in Andorid/Mobile builds and the extra binary growth is a cost to pay for kernels which are relvant to these builds. If need be, we will re-visit later if we want to exclude these from binary growth sensitive builds.

onnxruntime/core/mlas/lib/sconv.h
```diff
@@ -0,0 +1,32 @@
...
+    operations including output accumulation, bias addition, and activations.
+
+--*/
+
+/*
+    The MLAS_NEON_BLOCK_SIZE has to be the equal to the NchwcBlockSize in platform.cpp.
+    Refer to the discussion in https://github.com/microsoft/onnxruntime/pull/25580.
+*/
+
+constexpr size_t MLAS_NEON_BLOCK_SIZE = 16;
```
hariharans29: Nit: Maybe name it `MLAS_NEON_NCHWC_BLOCK_SIZE` to indicate that the block is for NCHWC (plain "Block" is leaving things a bit ambiguous...)

onnxruntime/core/mlas/lib/sconv_kernel_neon.cpp
```diff
@@ -0,0 +1,524 @@
...
+    This module implements the single precision convolution kernels for ARM NEON.
+
+--*/
+
+#include "mlasi.h"
+#include "sconv.h"
+
+#if defined(MLAS_TARGET_ARM64)
+
+constexpr size_t BlockSize = MLAS_NEON_BLOCK_SIZE;
```
hariharans29: Nit - Since we are only adding these files to the sources if the platform is ARM64 - are these guards necessary ?

onnxruntime/core/mlas/lib/spool_kernel_neon.cpp
```diff
@@ -0,0 +1,294 @@
...
+Abstract:
+
+    This module implements the single precision pooling kernels for ARM NEON.
+
+--*/
+
+#include "mlasi.h"
+#include "sconv.h"
+
+#if defined(MLAS_TARGET_ARM64)
```
hariharans29: Same as above - 

Nit - Since we are only adding these files to the sources if the platform is ARM64 - are these guards necessary ?

hariharans29:
LGTM - Can you please address the remaining comments and I will merge it. Thanks !

onnxruntime/core/mlas/lib/sconv.h
```diff
@@ -0,0 +1,32 @@
...
+Abstract:
+
+    This module defines convolution kernel flags for configuring convolution
+    operations including output accumulation, bias addition, and activations.
+
+--*/
+
+/*
+    The MLAS_NEON_BLOCK_SIZE has to be the equal to the NchwcBlockSize in platform.cpp.
+    Refer to the discussion in https://github.com/microsoft/onnxruntime/pull/25580.
```
edgchen1: can we use MLAS_NEON_BLOCK_SIZE to initialize MLAS_PLATFORM::NchwcBlockSize?

hariharans29: That's a comment I thought of as well. But then we would have to include this header in platform.cpp - thus far it doesn't include any "implementation specific" headers - and I wasn't sure if setting such a precent is good or bad at this stage (I guess it is fine to do so...)

edgchen1: it could be moved to a different header (i.e., the one containing the platform-specific kernel declarations) if that's more acceptable. I'd prefer that to having two separate constant values and a comment saying they should be equal.

onnxruntime/core/mlas/lib/sconv.h
Rohanjames1997: Done. Will push in next commit.

onnxruntime/core/mlas/lib/sconv_kernel_neon.cpp
Rohanjames1997: Nope, they're not necessary. 

The only reason I had kept it around was so that if anyone on x86 instance opened this file on a good editor, they would see the kernel code dimmed out/disabled and know that it was an Arm64 kernel :)

I will remove it in the next commit.

onnxruntime/core/mlas/lib/spool_kernel_neon.cpp
Rohanjames1997: Nope, they're not necessary. Will remove

onnxruntime/core/mlas/lib/sconv.h
Rohanjames1997: Will try moving it to `mlasi.h`.

Rohanjames1997:
@hariharans29, a question reg the binary size limit: What should I change it to?

From the CI, it looked like my build size was`1565969` bytes. So should the new limit be 5-10% higher than that? Let me know what the convention is for ORT development

Edit: Updated it to 10% higher.

onnxruntime/core/mlas/lib/sconv.h
Rohanjames1997: Done

Rohanjames1997:
The iphone simulator arm64 test failed with 

```
  Request completely sent off

  HTTP/1.1 404 Not Found

  Date: Fri, 12 Sep 2025 18:24:54 GMT

  Content-Type: text/plain; charset=utf-8

  Content-Length: 9

  Vary: X-PJAX, X-PJAX-Container, Turbo-Visit, Turbo-Frame,
  X-Requested-With,Accept-Encoding, Accept, X-Requested-With

  Cache-Control: no-cache

  Strict-Transport-Security: max-age=31536000; includeSubdomains; preload

  X-Frame-Options: deny

  X-Content-Type-Options: nosniff

  X-XSS-Protection: 0

  Referrer-Policy: origin-when-cross-origin, strict-origin-when-cross-origin

  Content-Security-Policy: default-src 'none'; base-uri 'self'; connect-src
  'self'; form-action 'self'; img-src 'self' data:; script-src 'self';
  style-src 'unsafe-inline'

  Server: github.com

  X-GitHub-Request-Id: BB0C:12DB8F:181056:186713:68C46575

  

  The requested URL returned error: 404

  Closing connection
```


I've updated my fork to use the latest github actions (@snnn merged it an hour ago)

Rohanjames1997:
Failure: [ONNX Runtime CUDA Builds / Windows GPU CUDA CI Pipeline Test Job (pull_request)](https://github.com/microsoft/onnxruntime/actions/runs/17682883787/job/50263559433?pr=25580)

Weird. Looks like a transient issue but not sure.

hariharans29:
> Failure: [ONNX Runtime CUDA Builds / Windows GPU CUDA CI Pipeline Test Job (pull_request)](https://github.com/microsoft/onnxruntime/actions/runs/17682883787/job/50263559433?pr=25580)
> 
> Weird. Looks like a transient issue but not sure.

Restarted, let us see.

Rohanjames1997:
Thanks, could you also run the CI? Looks like many have to be re-run, not just one.

hariharans29:
The Windows CUDA build issue is unrelated to your PR but the check is still a required check for merging. I will update you once a fix is in the main branch and you can rebase !

Rohanjames1997:
The `InferenceSessionTests.CheckRunProfilerWithSessionOptions` is failing 

```
2025-09-15T20:45:57.9327872Z 1: [ RUN      ] InferenceSessionTests.CheckRunProfilerWithSessionOptions
2025-09-15T20:45:58.0219177Z 1: E:\_work\onnxruntime\onnxruntime\onnxruntime\test\framework\inference_session_test.cc(684): error: Value of: has_kernel_info
2025-09-15T20:45:58.0219933Z 1:   Actual: false
2025-09-15T20:45:58.0220200Z 1: Expected: true
```

Looks like the profiler output [doesn't contain some lines](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/test/framework/inference_session_test.cc#L654-L681) that it expects

hariharans29:
> The `InferenceSessionTests.CheckRunProfilerWithSessionOptions` is failing
> 
> ```
> 2025-09-15T20:45:57.9327872Z 1: [ RUN      ] InferenceSessionTests.CheckRunProfilerWithSessionOptions
> 2025-09-15T20:45:58.0219177Z 1: E:\_work\onnxruntime\onnxruntime\onnxruntime\test\framework\inference_session_test.cc(684): error: Value of: has_kernel_info
> 2025-09-15T20:45:58.0219933Z 1:   Actual: false
> 2025-09-15T20:45:58.0220200Z 1: Expected: true
> ```
> 
> Looks like the profiler output [doesn't contain some lines](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/test/framework/inference_session_test.cc#L654-L681) that it expects

Yes. We will investigate.

Come to think of it, (I think) the failing check is not a required check for merging. I will merge this PR once the remaining builds pass if it allows me to.

Rohanjames1997:
Awesome! 
Thank you

Rohanjames1997:
Thanks @hariharans29 & @edgchen1 for the valuable review comments!
Happy to keep contributing to ORT!

hariharans29:
Hi @Rohanjames1997: Just curious - Are you able to repro the perf numbers that you shared in the PR description after the PR comments have been resolved ?

Rohanjames1997:
Yup, I was able to reproduce them even after I set the block size through the header file

hariharans29:
Thanks. I am guessing the kernels have been tuned using the hardware SKU you mention in the PR: Neoverse-V2?
Any idea how the perf looks on other hardware SKUs ?

Rohanjames1997:
Yup, neoverse-v2.

I can benchmark on neoverse-v1 as well and post the numbers, but I don't have access to other SKUs like neoverse-n1

Rohanjames1997:
Yup, neoverse-v2.

I can benchmark on neoverse-v1 as well and post the numbers, but I don't have access to other SKUs like neoverse-n1

onnxruntime/core/mlas/lib/sconv_kernel_neon.cpp
```diff
@@ -0,0 +1,520 @@
...
+    const float32x4_t ZeroVector = MlasBroadcastFloat32x4(0.0f);
+
+    const size_t StrideWidthElements = StrideWidth / sizeof(float);
+    const size_t DilationWidthElements = DilationWidth / sizeof(float);
+    const size_t FilterStrideElements = FilterStride / sizeof(float);
+    const size_t OutputStrideElements = OutputStride / sizeof(float);
+    const size_t InputWidthElements = InputWidth / sizeof(float);
+    const size_t DilatedInputWidthElements = DilatedInputWidth / sizeof(float);
+
+    (void)InputStride;
```
hariharans29, line 61: Just curious - Is InputStride not needed in this implementation ?

onnxruntime/core/mlas/lib/sconv_kernel_neon.cpp
```diff
@@ -0,0 +1,520 @@
...
+    const bool ReluActivation = (KernelFlags & MLAS_CONV_KERNEL_FLAG_RELU_ACTIVATION) != 0;
+
+    const float32x4_t ZeroVector = MlasBroadcastFloat32x4(0.0f);
+
+    const size_t StrideWidthElements = StrideWidth / sizeof(float);
+    const size_t DilationWidthElements = DilationWidth / sizeof(float);
+    const size_t InputStrideElements = InputStride / sizeof(float);
+    const size_t DilatedInputWidthElements = DilatedInputWidth / sizeof(float);
+
+    (void)InputStrideElements;
```
hariharans29, line 346: Same question

onnxruntime/core/mlas/lib/sconv_kernel_neon.cpp
Rohanjames1997: Correct. It is not required. Afaik, even in the x86 asm implementation it is not used (although it's difficult to read what is used and not)

onnxruntime/core/mlas/lib/sconv_kernel_neon.cpp
Rohanjames1997: Answered above

onnxruntime/core/mlas/lib/sconv_kernel_neon.cpp
hariharans29: Thanks. But what if InputStride is not 1 ?

hariharans29:
> Yup, neoverse-v2.
> 
> I can benchmark on neoverse-v1 as well and post the numbers, but I don't have access to other SKUs like neoverse-n1

Thanks - please share them when you can !

Any recommendations on how to tune other related perf parameters - like thread count, etc. to maximize perf ? 
Are there any other input shape related nuances with this implementation (i.e.) are there filter size, batch size related dependencies that make these kernels faster or slower than baseline ? 

The reason I ask is: I have a Windows ARM64 machine (Qualcomm Snapdragon X Elite) and I have a Conv heavy model given to me by a customer (unfortunately not shareable right now - I would need their permission) and I see a ~20% increase in forward pass latency with this change included and I am trying to work out how to mitigate that. They reported a similar perf degradation on their environment as well (I have asked them what their SKU is) but I am able to repro the same with my Qualcomm device which I tried as a sanity check.

onnxruntime/core/mlas/lib/sconv_kernel_neon.cpp
Rohanjames1997: Good question. The unit tests cover the scenarios for different strides. I tested them with `./build/Linux/Release/onnxruntime_mlas_test --gtest_filter=Conv2dNchwc*`

Having said that, I think `InputStride` is useful only when `InputChannels` is specified - like in the Pointwise kernel.

Rohanjames1997:
On a Neoverse-V1 instance, the peak performance is 450 inf/sec; so, 10% less than on Neoverse-V2. 
All my benchmarking has been done on Ubuntu 22.04 & 24.04

To maximize perf, I have noticed that these NCHWc kernels scale well (almost linearly) with threads, until 32 threads. 

For reference, the mobilenet model I used was from [ONNX/models](https://github.com/onnx/models/blob/refs/heads/main/validated/vision/classification/mobilenet/model/mobilenetv2-7.onnx), and this model does not accept batched inputs (tracking issue: https://github.com/onnx/models/issues/680) but that is apparently normal for ONNX models in that repository.

That is interesting, I could not get a chance to benchmark on Windows. Would you be able to find/tell me a similar model in https://github.com/onnx/models? We can benchmark it using `onnxruntime_perf_test -x <threads>`.

hariharans29:
> On a Neoverse-V1 instance, the peak performance is 450 inf/sec; so, 10% less than on Neoverse-V2. All my benchmarking has been done on Ubuntu 22.04 & 24.04
> 
> To maximize perf, I have noticed that these NCHWc kernels scale well (almost linearly) with threads, until 32 threads.
> 
> For reference, the mobilenet model I used was from [ONNX/models](https://github.com/onnx/models/blob/refs/heads/main/validated/vision/classification/mobilenet/model/mobilenetv2-7.onnx), and this model does not accept batched inputs (tracking issue: [onnx/models#680](https://github.com/onnx/models/issues/680)) but that is apparently normal for ONNX models in that repository.
> 
> That is interesting, I could not get a chance to benchmark on Windows. Would you be able to find/tell me a similar model in https://github.com/onnx/models? We can benchmark it using `onnxruntime_perf_test -x <threads>`.

Sure, let me get back to you on this one - I will either get a model from the zoo or a shareable version of the customer's model ? Thanks for the support.


One last question - I am guessing the kernels are called on multiple threads to cover different filters, batched inputs - is that correct ?

Rohanjames1997:
Yes, that's right. 

This is effectively an implementation of [grouped convolution](https://medium.com/data-science/grouped-convolutions-convolutions-in-parallel-3b8cc847e851), where channels and filters are split into groups. So, (I'm pretty sure) what happens is, each thread works on a different channel-group and a filter-group and that is why this algorithm is conducive to threading.

hariharans29:
> Yes, that's right.
> 
> This is effectively an implementation of [grouped convolution](https://medium.com/data-science/grouped-convolutions-convolutions-in-parallel-3b8cc847e851), where channels and filters are split into groups. So, (I'm pretty sure) what happens is, each thread works on a different channel-group and a filter-group and that is why this algorithm is conducive to threading.

Did you happen to see what the perf looked like when you didn't use `-x 32` (i.e.) using the default thread count in the threadpool as is on Neoverse V2 ?

Rohanjames1997:
Hmm so the default value of `-x` is 0, which uses all cores in the system. I was bencharking on a Neoverse-V2 system with 64 vCPUs, so I tested multiple configs: 2^0 to 2^6.

I don't have the exact numbers now, but the NCHWc kernels outperformed the baseline when using >=4 threads. The baseline was only faster in the single-threaded case, and it scaled very poorly after 4 threads.

hariharans29:
Hi @Rohanjames1997 -

Here is a shareable version of the model: https://drive.google.com/file/d/1dpzSVMvSFVbLodAE6aQYm_qAOgkmAtkZ/view?usp=drive_link. Let me know if you are able to access this.

I have tried to time the Conv node latencies with and without the NCHWc layout transformation by making this change in the sequential_executor.cc file (attached)
[sequential_executor_timing_updated.txt](https://github.com/user-attachments/files/22476787/sequential_executor_timing_updated.txt). I do find the Conv nodes are costlier than before the change. Would you be able to try and confirm ?

Rohanjames1997:
Hi, I don't have access to this model without requesting for it. Can you make it a publicly available object?

Also, do you mind explaining to me the change you made in `sequential_executor.cc`? Also, is there a diff on one of your branches that I can look at to understand it?

Also, what machine & OS did you benchmark on?

Ty!

hariharans29:
> Hi, I don't have access to this model without requesting for it. Can you make it a publicly available object?
> 
> Also, do you mind explaining to me the change you made in `sequential_executor.cc`? Also, is there a diff on one of your branches that I can look at to understand it?
> 
> Also, what machine & OS did you benchmark on?
> 
> Ty!

Thanks ! I just made the link public.

The change in sequential executor just times the call of an OpKernel's `Compute()` method - the MLAS routines you added will be called an OpKernel's instance - specifically it should be called from this method here - https://github.com/microsoft/onnxruntime/blob/45457323a31f40022edddc778872e265702e3531/onnxruntime/contrib_ops/cpu/nchwc_ops.cc#L154.

As for the diff, it can be diffed with current main branch's version of the file. Actually, you don't really need that change if you already have a kernel profiling methodology.

I benchmarked on a Copilot+Pc Surface Laptop - Qualcomm Snapdragon X Elite + Windows OS. The customer is running this on an m8g.xlarge (Graviton4)

hariharans29:
Did you happen to check performance of any model other than mobilenet ? I see that mobilenet only invokes these 2 code paths - [Pointwise](https://github.com/microsoft/onnxruntime/blob/88f2652dee465faa485cf68cb39c4489f28945bb/onnxruntime/core/mlas/lib/snchwc.cpp#L1316) and [Depthwise](https://github.com/microsoft/onnxruntime/blob/88f2652dee465faa485cf68cb39c4489f28945bb/onnxruntime/core/mlas/lib/snchwc.cpp#L1321) convolutions, whereas the sample model I pasted above uses that and ["regular" NCHWc path](https://github.com/microsoft/onnxruntime/blob/88f2652dee465faa485cf68cb39c4489f28945bb/onnxruntime/core/mlas/lib/snchwc.cpp#L1318). I see that is the path that is more costlier. Possibly that kernel has some perf deficiencies ?

Rohanjames1997:
Hi!

No, I couldn't benchmark it yet. I'll spend some time on it today. 

Thanks for the note about the NCHWc path. 

Do you happen to know if it is deficient when allocating more threads too?

Rohanjames1997:
Here are the benchmark results using the shared model on a Graviton4 & Ubuntu22.
Inference is faster at higher thread counts and slower at lower thread counts.

All numbers are inferences/sec

| threads | 1       | 4       | 8       | 16      | 32      | 64      | 
|---------|---------|---------|---------|---------|---------|---------| 
| vanilla | 3.35551 | 9.88463 | 14.071  | 17.6474 | 18.954  | 19.1331 | 
| nchwc   | 1.89976 | 7.17679 | 13.0548 | 21.4388 | 29.3814 | 32.0701 | 
| speedup | 0.566   | 0.726   | 0.928   | 1.215   | 1.550   | 1.676   | 
|         |         |         |         |         |         |         | 

@hariharans29 I can look more into optimizing the perf at lower thread counts. But it might be tricky. 
Thinking out loud - do you know if we can have a heuristic that checks for thread pool size and reroutes inference to the more optimized path - vanilla or NCHWc - based on that?

hariharans29:
Thanks @Rohanjames1997. I am figuring my way around the MLAS library as well - but I'd say being to identify when to use regular (im2col + MlasGemm) vs when to use NCHWc might be equally hard as well - both code-wise and identifying the right cut-off points on different platforms - that would probably need some "online benchmarking" (i.e.) use the first run to run both paths and pick the fastest algo to use from subsequent runs. Even this would be a change in first Run() behavior and needs more intenral discussion.

Just a quick clarifying question - Which Graviton powered instance are you using to allow using upto 64 threads ? The customer is on a m8g.xlarge which is a 4 core instance I believe. So from the table above, given the regression with lower thread counts, I think that explains the regression ?

Rohanjames1997:
That's true. Even I shall think more about the online benchmarking method and about heuristics.
I know Pytorch has heuristics for choosing the right backend for matmuls based on the size of the input matrices. So, I'm drawing inspiration from there.

Yes, the lower thread count explains the regression. I've been running on a c8g.16xlarge, mainly to build ORT faster and to benchmark on a variety of thread counts. The number of vCPUs is (usually) `=4n` for a `.nxlarge` instance

hariharans29:
Thanks. 

Was the benchmarking setup you used same as the mobilenet one for the new model ? The new model supports batching. Were you able to batch more images ?

The reason I ask is - there was a nice contribution that improves the thread utlization for batched inputs and group convolutions (regular path) [here](https://github.com/microsoft/onnxruntime/pull/25255). Here is my "internal" copy of the [same code](https://github.com/microsoft/onnxruntime/pull/26103). I had seen this a while ago but this only just struck me last week. With the increased thread counts, this code may close the gap of the "regular" with the NCHWc variant.

Rohanjames1997:
Yes, it was the same setup.

How do I batch using `onnxruntime_perf_test`? 
I tried looking at the help section and I think the `-f` or `-F` flags may help but I would require either the `dimension_name` or the `dimension_denotation`. By default, when specifying `-I`,  _"Free dimensions are treated as 1 unless overridden using -f."_

That's an interesting PR, and you're right, it could close the gap! 
One thing I don't know yet is which CPU platform that PR targets, or if it is platform-agnostic. When I was developing, I noticed that on x86, MlasConv was not invoked by default, and the NCHWc variant was always invoked.

One thing worth mentioning is that even while running `onnxruntime_perf_test` with the regular conv (MlasConv), the specified number of cores are being utilized to 100%, i.e. there is no "underutilization".

hariharans29:
**Which platform does that PR target ?**

That PR should improve the "regular" Conv perf for all platforms. It is not platform specific. It re-works the thread utilization for the "regular" Conv (Im2Col + MlasGemm). The MlasGemm implementation would be platform specific of course.

**Why is NCHWc Conv invoked by default on x86?**

It depends on the graph optimization level the user sets - by default it is ORT_ENABLE_ALL which by default includes layout transformation optimizations. Keep in mind though if there is a Conv that doesn't qualify for layout transformations - it will drop down to using the "regular" Conv.

See [here](https://github.com/microsoft/onnxruntime/blob/42fcd71a8a236143bd5995262a89e973d5b5bbcd/onnxruntime/core/optimizer/graph_transformer_utils.cc#L429)

Prior to your PR, ARM64 was not NCHWc compliant and hence it was always using the regular Conv path

**How do I batch using perf_test?**

I think your test data located in the same file as the model should contain batched data. 

Here is some sample code to generate and serialize a batched numpy array using the onnx python package. Please make appropriate changes where needed:

import numpy as np
from onnx.numpy_helper import from_array, to_array
import onnx

shape = (2, 384, 384, 3) # Batch size 2
arr = np.zeros(shape, dtype=np.float32)
onnx_tensor = from_array(arr)
serialized_tensor = onnx_tensor.SerializeToString()

with open("input_1.pb", "wb") as f:
    f.write(serialized_tensor)


You can drop in the generated test data into the sub-folder containing the test data of the folder that hosts the model.

Rohanjames1997:
Thanks for the explanation!

Yes, I was aware that `MlasNchwcGetBlockSize` controlled whether NCHWc was invoked or not.

On x86, since `MlasNchwcGetBlockSize` was always defined, can you give me an example of a  _"Conv that doesn't qualify for layout transformations"_ ?

I see, thanks. I always generated testing data automatically. I'll give this a shot next week.

hariharans29:
**On x86, since MlasNchwcGetBlockSize was always defined, can you give me an example of a "Conv that doesn't qualify for layout transformations" ?**

I am talking about cases like [these](https://github.com/microsoft/onnxruntime/blob/42fcd71a8a236143bd5995262a89e973d5b5bbcd/onnxruntime/core/optimizer/nchwc_transformer.cc#L361) where we bail out while transform the regular Conv into an NCHWc compliant one. There are certain criteria to be met before we can use the NCHWc version - although I am not sure how often in practice, these criteria are not met.

**I see, thanks. I always generated testing data automatically. I'll give this a shot next week.**

No pressure, thanks for your contribution and your time. Since, there is atleast one known case where the regular Conv performs better than the NCHWc one - I am thinking it may be prudent to temporarily revert this change so that we can let the main branch go back to the last know stable state. What do you think ?

Rohanjames1997:
I see, thanks!

Regarding reverting the PR- it is totally up to you and the other maintainers.

NCHWc seems to underperform at lower thread counts and outperform at higher thread counts. Also, the peak achievable throughput is always higher using NCHWc kernels - this was seen on Mobilenet and the shared model.

I think adding a thread-count based heuristic would be the better approach, but if you do decide to revert the change, I respect that as well.

hariharans29:
I think there are arguments to be made for both - reverting the change for now and keeping the change as is.

1) The model transformations happen statically (i.e.) conversion from a regular Conv to NCHWc Conv happen as the model loads and we will need to think about what is the best way to use the threadpool information to trigger the optimizer or not to. This needs some design discussions. 

2) Not all client ARM devices may be able to increase their thread count to see the improvements.

3) It would be interesting to see if the x86 NCHWc kernel hits the same issue (i.e.) performs poorly with lowered thread count. If it doesn't, would it compute model be applicable to ARM64 as well ?

Regards to reverting, I will discuss internally and get back. :)

hariharans29:
Hi @Rohanjames1997 - 

As a temporary measure, could you consider making this an optional feature (turned OFF by default initially) ? Maybe following the model here - https://github.com/microsoft/onnxruntime/pull/25238 ? That way users who would like to leverage this feature get to do so by building from source. 

CC: @edgchen1

Rohanjames1997:
Sure thing, that makes sense.

I will probably get to work on this only on the week of Oct 6th. So until then, I don't mind if it's reverted/kept as is.

hariharans29:
I just went ahead and added it: https://github.com/microsoft/onnxruntime/pull/26171.

I will test that it works as expected and merge it.

PR #23534, opened 2025-01-29
[MIGraphX EP] Add migx ep fp8 support and int4 weights

TedThemistokleous:
* Add fp8 and int4 types in supported list for Onnxruntime EP

* Add support for int4 inputs

Map things to int8 right now as we don't explicitly set an int4 input type and pack/unpack int4 operands

* Add flag to allow for fp8 quantization through Onnxruntime API

* Add fp8 quantization to the compile stage of the MIGraphX EP

Mirror the same calibration code we use for int8 and just change which quantize we call through the MIGraphx API

* cleanup logging

* Cleanup and encapsulate quantization / compile functions

- Add additional flags for fp8 thats shared for int8

- Add lockout warning message when int8/fp8 used at the same time

* Run lintrunner pass

* Fix session options inputs + add better logging.

Previous runs using session options failed as we were missing pulling in inputs from the python interface. This plus additional logging allowed me to track what options were invoked via env and what were added during the start of an inference session

* Fix naming for save/load path varibles to be consistent with  enable.

* Print only env variables that are set as warnings

need this so the user knows there's any of the environment variables running in the background to ensure proper consistently between runs.

---------

### Description

Changes to cleanup the MIGraphX EP quantization code as well as adding fp8 quantization support along with int4 support.

Cleanup changes handle a few instances of issues seen with the python interface when taking in provider options

### Motivation and Context

Required as we fix ignored flags when using provider_options for the MIGraphX EP
Adding fp8 quantization through the MIGraphX API
Adding int4 weight support for packed int4 weights for MIGraphX inference

PR #25583, opened 2025-07-29
[MIGraphX EP] Syncing AMD changes upstream

apwojcik:
A set of changes required for WCR/WindowsML that were added to the MIGraphX Execution provider. The development was done in the ROCm repository, now we want to sync with the main branch with a single drop.

The PR incorporates the review comments from the previous closed PR #25338.

Motivation and Context
Fixes, changes, and updates to MIGraphX EP that have been done for ROCm development. Pushing this back upstream to ensure mainline onnxruntime is using the latest changes. Moving forward, MIGraphX EP will be cut from the latest official release tag as a base point while also adding additional features that will be contributed back.

### DIFF
--- a/onnxruntime/python/onnxruntime_pybind_ortvalue.cc
+++ b/onnxruntime/python/onnxruntime_pybind_ortvalue.cc
@@ -421,7 +421,7 @@ void addOrtValueMethods(pybind11::module& m) {
       // Converts Tensor into a numpy array
       .def("numpy", [](const OrtValue* ml_value) -> py::object {
         ORT_ENFORCE(ml_value->IsTensor(), "Only OrtValues that are Tensors are convertible to Numpy objects");
-        const auto& device = ml_value->Get<Tensor>().Location().device;
+        [[maybe_unused]] const auto& device = ml_value->Get<Tensor>().Location().device;
 #ifdef _MSC_VER
 // The switch statement may only contain the 'default' label. In such a case, the MSVC compiler
 // will warn about it, and since the warnings are treated as errors, the compilation will break.


snnn (changes requested):
Please avoid adding new functions the public C API header file. All the EPs should use SessionOptionsAppendExecutionProvider_V2 to set the options.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -773,7 +773,11 @@ typedef struct OrtMIGraphXProviderOptions {
    *   \note If a ::OrtArenaCfg has been applied, it will override this field
    */
   int migraphx_arena_extend_strategy;
-
+  int migraphx_bf16_enable;             // MIGraphX BF16 precision. Default 0 = false, nonzero = true
```
snnn: This is an ABI change. Could you please avoid making such changes?

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -6439,6 +6443,88 @@ struct OrtApi {
...
+
+  /** \brief Create an OrtMIGraphXProviderOptions
+   *
+   * \param[out] out Newly created ::OrtMIGraphXProviderOptions. Must be released with OrtApi::ReleaseMIGraphXProviderOptions
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \since Version 1.xx.
+   */
+  ORT_API2_STATUS(CreateMIGraphXProviderOptions, _Outptr_ OrtMIGraphXProviderOptions** out);
```
snnn: Could you please use SessionOptionsAppendExecutionProvider_V2 function instead? So that we don't need to add new APIs.

onnxruntime/core/providers/migraphx/migraphx_call.cc
```diff
@@ -48,8 +48,8 @@ std::conditional_t<THRW, void, Status> RocmCall(
       (void)hipGetDevice(&currentHipDevice);
       (void)hipGetLastError();  // clear last HIP error
       static char str[1024];
-      snprintf(str, 1024, "%s failure %d: %s ; GPU=%d ; hostname=%s ; file=%s ; line=%d ; expr=%s; %s",
-               libName, (int)retCode, RocmErrString(retCode), currentHipDevice,
+      snprintf(str, sizeof(str), "%s failure %d: %s ; GPU=%d ; hostname=%s ; file=%s ; line=%d ; expr=%s; %s",
```
snnn: This looks like a security bug. Please check the function's document (for Windows and Linux)

onnxruntime/core/providers/shared_library/provider_interfaces.h
```diff
@@ -1354,6 +1354,11 @@ struct ProviderHost {
   virtual std::unique_ptr<ModelMetadefIdGenerator> ModelMetadefIdGenerator__construct() = 0;
   virtual void ModelMetadefIdGenerator__operator_delete(ModelMetadefIdGenerator* p) = 0;
   virtual int ModelMetadefIdGenerator__GenerateId(const ModelMetadefIdGenerator* p, const GraphViewer& graph_viewer, HashValue& model_hash) = 0;
+
+#ifdef _WIN32
+  virtual std::string ToUTF8String(std::wstring_view s) = 0;
```
snnn: Is it an ABI change? Adding new functions to the EP interface.

onnxruntime/core/session/onnxruntime_c_api.cc
```diff
@@ -1378,9 +1378,16 @@ ORT_API_STATUS_IMPL(OrtApis::SessionGetOverridableInitializerTypeInfo, _In_ cons
...
-  memcpy(output_string, str.c_str(), str.size());
+char* onnxruntime::StrDup(std::string_view str, OrtAllocator* allocator) {
+  char* output_string = static_cast<char*>(allocator->Alloc(allocator, str.size() + 1));
+  memcpy(output_string, str.data(), str.size());
+  output_string[str.size()] = '\0';
+  return output_string;
+}
+
+wchar_t* onnxruntime::StrDup(std::wstring_view str, OrtAllocator* allocator) {
+  auto* output_string = static_cast<wchar_t*>(allocator->Alloc(allocator, str.size() + 1));
```
snnn: This is a security bug. You didn't allocate enough memory! 
Besides, should the two functions be static?

snnn:
This PR has conflicts with #25566 , though git does not show any conflict here. Because #25566 added a new C API in the middle.

include/onnxruntime/core/session/onnxruntime_c_api.h
apwojcik: We need these for the C# interface to MIGraphX EP to enable users to set provider options.

onnxruntime/core/providers/migraphx/migraphx_call.cc
apwojcik: The change here is from `1024` to `sizeof(str)` (the second argument). The call to the function remains unchanged, and the function itself has not changed since the beginning.
Do you want us to remove that function and replace it with something safer?

onnxruntime/core/session/onnxruntime_c_api.cc
apwojcik: Yeah, will fix it. Thanks!

include/onnxruntime/core/session/onnxruntime_c_api.h
apwojcik: You were accepting such changes before #23404, #23534, #21599, and #20643.

onnxruntime/core/providers/shared_library/provider_interfaces.h
apwojcik: In the previous PR, #25338, you asked us to use wide strings for Windows paths and string views for literals. The EP interface is missing methods for the wide string view type.

snnn: It's a very simple function, we don't have to share the implementation across DLL boundaries?  Since we have cut the  1.23 release,  I hope we can avoid making ABI changes.

onnxruntime/core/providers/migraphx/migraphx_call.cc
snnn: snprintf does not ensure the dest string is nul-terminated .

include/onnxruntime/core/session/onnxruntime_c_api.h
snnn: Does SessionOptionsAppendExecutionProvider_V2  work for your use cases?  All EPs should use the new SessionOptionsAppendExecutionProvider_V2  function to register themselves.

include/onnxruntime/core/session/onnxruntime_c_api.h
snnn: These changes should not be made.  ONNX Runtime maintains ABI backward compatibility. This is our commitment to our customers.

onnxruntime/core/providers/migraphx/migraphx_call.cc
snnn: You didn't introduce the security bug. But, the security bug still exists and it should be fixed. You may open a new PR for  this if you'd like.
On Windows we prefer to use _snprintf , which has a C++ version that can handle such situations well. 
Visual Studio's static analyzer usually can find such issues.

include/onnxruntime/core/session/onnxruntime_c_api.h
hisham-hchowdhu: @snnn  could we add this change in the follow up PR? we have not implemented or test this path - needs little more time post GA

apwojcik: > Does SessionOptionsAppendExecutionProvider_V2 work for your use cases? All EPs should use the new SessionOptionsAppendExecutionProvider_V2 function to register themselves.

I understand that a client application now requires calling SessionOptionsAppendExecutionProvider_V2 instead of SessionOptionsAppendExecutionProvider_MIGraphX, right?

MIGraphX EP utilizes these functions to convert the C# representation of provider options (dictionary) into C++ provider options, as well as to convert data formats. Especially, string formats are not the exact data type.
How can this be achieved with SessionOptionsAppendExecutionProvider_V2? Do we still need that conversion? Can you guide me to the documentation chapter describing how to use the function in C#?

PS. Like Hisham mentioned, we have not implemented or tested this path yet.

onnxruntime/core/providers/migraphx/migraphx_call.cc
apwojcik: I will fix that here.

setup.py
```diff
@@ -285,15 +285,43 @@ def run(self):
...
                 assert self.dist_dir is not None
                 file = glob(path.join(self.dist_dir, "*linux*.whl"))[0]
                 logger.info("repairing %s for manylinux1", file)
                 auditwheel_cmd = ["auditwheel", "-v", "repair", "-w", self.dist_dir, file]
-                for i in cuda_dependencies + rocm_dependencies + tensorrt_dependencies + cann_dependencies:
+                for i in (
+                    cuda_dependencies
+                    + hipsdk_dependencies
+                    + rocm_dependencies
+                    + migraphx_dependencies
```
skottmckay: The new dependencies are defined in the `else` branch of `if is_manylinux` but this is inside `if is_manylinux and not disable_auditwheel_repair and not is_openvino and not is_qnn:`, so won't they be undefined if this code is run?

include/onnxruntime/core/session/onnxruntime_c_api.h
skottmckay: i.e. you can add things at the end of the struct but not change existing names or data types.

include/onnxruntime/core/session/onnxruntime_c_api.h
skottmckay: EPs should be registered via SessionOptionsAppendExecutionProvider (when included as part of the ORT build) and SessionOptionsAppendExecutionProvider_V2 (when added using RegisterExecutionProviderLibrary) instead of adding an EP specific function to the ORT API. There are existing bindings in various languages for these functions. As we're moving to plugin EPs there should be no new EP specific functions in the ORT API going forward.

EP options are passed as string:string pairs. Those need to be converted by the EP implementation which you appear to already have an implementation of in `MIGraphXExecutionProviderInfo(const ProviderOptions& options)`

include/onnxruntime/core/common/common.h
```diff
@@ -294,12 +294,26 @@ inline std::string ToUTF8String(const std::string& s) { return s; }
...
+inline std::string ToUTF8String(const wchar_t* s) {
+  return ToUTF8String(std::wstring_view{s});
+}
+inline std::string ToUTF8String(const std::wstring& s) {
+  return ToUTF8String(std::wstring_view{s});
+}
+std::wstring ToWideString(std::string_view s);
+inline std::wstring ToWideString(const char* s) {
+  return ToWideString(std::string_view{s});
+}
```
skottmckay, line 307: Are all these additions required, especially the ones taking a raw pointer.

In general, we try and limit binary size where possible, so avoid adding every possible variant unless there's a specific use case that requires them.

include/onnxruntime/core/framework/provider_options_utils.h
```diff
@@ -89,6 +89,21 @@ class ProviderOptionsParser {
...
+        "Provider option \"", name, "\" already has a value parser.");
+    return *this;
+  }
+
+  template <typename ValueParserType>
+  ProviderOptionsParser& AddValueParser(
+      const char* name, ValueParserType value_parser) {
+    return AddValueParser<ValueParserType>(std::string_view{name}, value_parser);
+  }
+
```
skottmckay: Why are these needed given ProviderOptions is a map of std::string to std::string?

`using ProviderOptions = std::unordered_map<std::string, std::string>;`

onnxruntime/core/providers/migraphx/migraphx_provider_factory.cc
```diff
@@ -184,9 +223,12 @@ struct MigraphXEpFactory : OrtEpFactory {
       : ort_api{ort_api_in}, default_logger{default_logger_in}, ep_name{ep_name}, ort_hw_device_type{hw_type} {
     GetName = GetNameImpl;
     GetVendor = GetVendorImpl;
+    GetVersion = GetVersionImpl;
     GetSupportedDevices = GetSupportedDevicesImpl;
     CreateEp = CreateEpImpl;
     ReleaseEp = ReleaseEpImpl;
+    GetVendorId = GetVendorIdImpl;
+    CreateDataTransfer = CreateDataTransferImpl;
```
skottmckay: Please add the other missing functions. See #25503 for example implementations.

onnxruntime/core/providers/shared_library/provider_api.h
```diff
@@ -326,6 +326,12 @@ std::unordered_set<NodeIndex> GetCpuPreferredNodes(const onnxruntime::GraphViewe
                                                    const logging::Logger& logger);
 
 std::string GetEnvironmentVar(const std::string& var_name);
+inline std::string GetEnvironmentVar(std::string_view var_name) {
+  return GetEnvironmentVar(std::string{var_name});
+}
+inline std::string GetEnvironmentVar(const char* var_name) {
+  return GetEnvironmentVar(std::string_view{var_name});
+}
```
skottmckay: I don't think these are required so it would be better not to bloat the API by adding them.

onnxruntime/python/onnxruntime_pybind_mlvalue.cc
```diff
@@ -230,7 +231,8 @@ AllocatorPtr GetMIGraphXAllocator(OrtDevice::DeviceId id) {
 
   if (id_to_allocator_map->find(id) == id_to_allocator_map->end()) {
     // TODO: Expose knobs so that users can set fields associated with OrtArenaCfg so that we can pass it to the following method
-    id_to_allocator_map->insert({id, GetProviderInfo_MIGraphX().CreateMIGraphXAllocator(id, gpu_mem_limit, arena_extend_strategy, migx_external_allocator_info, nullptr)});
+    id_to_allocator_map->insert({id, GetProviderInfo_MIGraphX().CreateMIGraphXAllocator(id, gpu_mem_limit, arena_extend_strategy,
```
skottmckay: nit: please keep lines to 120 chars  #Closed

onnxruntime/python/onnxruntime_pybind_ortvalue.cc
```diff
@@ -383,21 +383,27 @@ void addOrtValueMethods(pybind11::module& m) {
       // Converts Tensor into a numpy array
       .def("numpy", [](const OrtValue* ml_value) -> py::object {
         ORT_ENFORCE(ml_value->IsTensor(), "Only OrtValues that are Tensors are convertible to Numpy objects");
-
+        const auto& device = ml_value->Get<Tensor>().Location().device;
+        switch (device.Vendor()) {
```
skottmckay, line 432: This potentially breaks backwards compatibility so it would be safer to leave things unchanged in this PR. Existing python code is not aware of the vendor id as it is new, so my concern would be there might be a code path where the vendor id is not correctly set and it relies upon the current implementation.

#25346 has changes to maintain the current behavior and supports getting a shared allocator for a plugin EP.

We need to go through and validate that there are no code paths in python that are not setting the vendor id correctly before we update this logic, however all the EP specific logic in the ORT code go away once all the EPs are plugin based.

tools/ci_build/build.py
```diff
@@ -2031,6 +2030,9 @@ def build_nuget_package(
     elif use_tensorrt:
         execution_provider = "/p:ExecutionProvider=tensorrt"
         package_name = "/p:OrtPackageId=Microsoft.ML.OnnxRuntime.TensorRT"
+    elif use_migraphx:
+        execution_provider = "/p:ExecutionProvider=migraphx"
+        package_name = "/p:OrtPackageId=Microsoft.ML.OnnxRuntime.MIGraphX"
```
skottmckay, line 2038: What's the reason for this change? Is there a plan to build a nuget package with ORT + MiGraphX directly included? That's unexpected given that with a plugin EP the build is independent of ORT so the EP could be a separate nuget package.

include/onnxruntime/core/common/common.h
apwojcik: We've been asked to use WideChar for paths on Windows. The functions facilitate UTF8 to WideChar conversions, making the code cleaner and more straightforward. The explicit variants of `const char*` and `const wchar_t*` are required because of the ambiguity the compiler reports - it does not know which function to use, since both `std::string` and `std::string_view` have a constructor accepting `const char*`.
Those functions do not enlarge the size of binaries. They are only hints to a compiler to resolve ambiguities correctly.

setup.py
```diff
@@ -435,6 +464,26 @@ def finalize_options(self):
     libs.extend(qnn_deps)
     if nightly_build:
         libs.extend(["onnxruntime_pywrapper.dll"])
+    migraphx_deps = [
```
snnn, line 437: Similar to Scott's question: https://github.com/microsoft/onnxruntime/pull/25583/files#r2250309763

Should the DLLs be distributed separately ? As we already have auto EP.

include/onnxruntime/core/framework/provider_options_utils.h
apwojcik: In the sync PR we submitted before, the reviewer asked us to use `std::string_view` for literals [25338](https://github.com/microsoft/onnxruntime/pull/25338#discussion_r2201921195).
The `std::string_view` has to be explicitly converted to `std::string`, otherwise the compiler reports an error.
Both method variants are needed because of the ambiguity reported by a compiler, and they help to resolve it correctly. The methods are only hints to a compiler. They should not increase the resulting binary size.

onnxruntime/core/providers/shared_library/provider_api.h
apwojcik: We've been asked to use `std::string_view` for literals. The functions help to resolve ambiguities correctly. They do not enlarge the resulting binary size. They are only hints to a compiler.

onnxruntime/core/session/provider_bridge_ort.cc
```diff
@@ -3077,3 +3088,125 @@ ORT_API_STATUS_IMPL(OrtApis::SessionOptionsAppendExecutionProvider_VitisAI, _In_
...
+  API_IMPL_END
+}
+
+ORT_API(void, OrtApis::ReleaseMIGraphXProviderOptions, _Frees_ptr_opt_ OrtMIGraphXProviderOptions* ptr) {
+#ifdef USE_MIGRAPHX
+  std::unique_ptr<OrtMIGraphXProviderOptions> p(ptr);
+  OrtAllocator* allocator;
+  GetAllocatorWithDefaultOptions(&allocator);
+  if (ptr->migraphx_cache_dir != nullptr) {
+    allocator->Free(allocator, const_cast<wchar_t*>(ptr->migraphx_cache_dir));
```
snnn: Was the `ptr` pointer allocated from this allocator?

onnxruntime/python/onnxruntime_pybind_mlvalue.cc
apwojcik: We run `lintrunner` before every PR submission. This change is made by the tool. We saw `lintrunner` reports no configuration file found. Do you have the up-to-date configuration file we can use?

onnxruntime/python/onnxruntime_pybind_ortvalue.cc
apwojcik: One of our scenarios involves using DML EP and MIGraphX EP within the same ONNXRT application. This code makes OrtValue work for both EPs. The previous method favours MIGraphX EP and is breaking DML EP.

tools/ci_build/build.py
apwojcik: Yes, we need those changes to enable MIGraohX EP for the C# ONNXRT (classic EP, not plugin EP) application that one of our customers is working on.

onnxruntime/python/onnxruntime_pybind_mlvalue.cc
skottmckay: There's no automatic enforcement of the 120 char limit as clang-format greatly decreases readability if you enable it. It tries to put as much as possible on every line and any user formatting is lost. 

clang-format can be a little cryptic as to where it keeps/removes line breaks so sometimes it takes some experimentation. this would make it happy:

```
    id_to_allocator_map->insert(
        {id, GetProviderInfo_MIGraphX().CreateMIGraphXAllocator(
                 id, gpu_mem_limit, arena_extend_strategy,
                 migraphx::external::alloc_fn, migraphx::external::free_fn, migraphx::external::empty_cache_fn,
                 nullptr)});
```

setup.py
apwojcik: Some of those DLLs are currently available with the installation of the HIP SDK on Windows. However, we do not want to force customers to do such a prestep, and they asked us for a self-contained solution. Furthermore, none of the installation packages from AMD currently distribute MIGraphX on Windows.
There is one more aspect. The DLLs are backward compatible (in most cases), but not forward compatible. For the separate distribution, we risk losing control, and customers may encounter compatibility issues if the versions do not match.

onnxruntime/core/session/provider_bridge_ort.cc
apwojcik: Yes, it does use the same allocator. This code is working and has been tested internally for some time already.

setup.py
snnn: Where are you going to publish this python package? And, what's the package name? (We might have a problem here)

onnxruntime/python/onnxruntime_pybind_ortvalue.cc
skottmckay: I _think_ it should be ok as the legacy python OrtDevice ctor sets the vendor id and supports DML + one other GPU EP. Needs an update for CANN though. Could you possibly add this [here](https://github.com/microsoft/onnxruntime/blob/14ca6df1f21cb9cef80879e6f786016273cae6e1/onnxruntime/python/onnxruntime_pybind_state.cc#L1888) for completeness?

```c++
             } else if (type == OrtDevice::NPU) {
#if USE_CANN
               vendor = OrtDevice::VendorIds::HUAWEI;
#endif
```

include/onnxruntime/core/framework/provider_options_utils.h
skottmckay: The ORT_ENFORCE adds a branch and a throw (which adds exception unwind info), and as it's templatized that occurs for every type it's used with. That can be mitigated if the `const std::string&` variant implementation calls the `string_view` implementation.

onnxruntime/core/providers/shared_library/provider_api.h
skottmckay: why does `const char*` go via string_view instead of directly to std::string?

include/onnxruntime/core/session/onnxruntime_c_api.h
apwojcik: The change is not a rename of an existing structure member. It's a new member flag that enables BFloat16 support in MIGraphX. I'd appreciate a clear recommendation. Is this change allowed? If not, could you please provide information on how to pass new flags/options to an existing classic execution provider?

onnxruntime/python/onnxruntime_pybind_ortvalue.cc
apwojcik: It's done.

onnxruntime/core/providers/shared_library/provider_interfaces.h
apwojcik: It's done. I have removed those abstract methods from the EP interface.

onnxruntime/core/session/onnxruntime_c_api.cc
apwojcik: It's done.

setup.py
apwojcik: I removed those changes. Thanks!

include/onnxruntime/core/framework/provider_options_utils.h
apwojcik: It's done.

onnxruntime/core/providers/shared_library/provider_api.h
apwojcik: It's done. I changed to `std::string` instead of `std::string_view`

onnxruntime/python/onnxruntime_pybind_mlvalue.cc
apwojcik: I fixed the formatting as you requested. Thanks!

onnxruntime/python/onnxruntime_pybind_ortvalue.cc
apwojcik: I have added the missing fragment as you requested.

setup.py
apwojcik: As I can see now, the Python wheel name (for Python version 3.12) is `onnxruntime-1.23.0-cp312-cp312-win_amd64.whl`, but it should be `onnxruntime_migraphx-1.23.0-cp312-cp312-win_amd64.whl`. Is that why you are asking about the name?

apwojcik: I fixed the name of the Python wheel to `onnxruntime_migraphx-*.whl`

onnxruntime/core/providers/migraphx/migraphx_provider_factory.cc
apwojcik: We are not maintaining those changes. The changes were added by the WCR team from Microsoft. We are unsure what additional elements are required to complete an EP as a plugin. For this PR, I prefer not to mix upstream sync changes with modifications to make an EP as a plugin. The goal of this PR is to upstream AMD local changes so that all can work on MIGraphX EP development from the same source code base.

apwojcik:
> Please avoid adding new functions to the public C API header file. All the EPs should use SessionOptionsAppendExecutionProvider_V2 to set the options.

I have removed changes made to the C API header file and the respective Cpp source files. I will submit a separate PR once we've completed the C# interface to MIGraphX's new approach.

include/onnxruntime/core/session/onnxruntime_c_api.h
skottmckay: I think the current change is fine in terms of not changing any existing values or data types and appending at the end.

Does the implementation handle a user creating the EP with the old version of the options gracefully? i.e. how do you know if the user is aware of the new fields and is providing them? example use case would be existing user code where the onnxruntime.dll is replaced with a new version. user code doesn't need to be recompiled in this scenario but would only be aware of the original definition of this struct.

skottmckay:
> I have removed changes made to the C API header file

I still see the change to add fields to the public OrtMIGraphXProviderOptions. how will you prevent an issue if someone registers the migraphx ep with the header from an older version of ORT? the new values in OrtMIGraphXProviderOptions will be random data in that case.

e.g. in OrtAllocator the user sets this to the ORT version from the ORT header they're using, and we know based on that whether recently added functions like `GetStats` can be used or not. if the user doesn't have a way to provide the ORT version that their code was compiled with, I'm not aware of a way for you to determine whether fields added after that version can be used.

https://github.com/microsoft/onnxruntime/blob/ee794afec24bd448eabf15102d1f14a110c35516/include/onnxruntime/core/session/onnxruntime_c_api.h#L344-L345

to be safe I would have expected you'd need to create something like OrtMIGraphXProviderOptions_V2 internally and populate that from OrtMIGraphXProviderOptions (support existing usage) or from ProviderOptions (support existing and new usage). the only way to set the new fields would be via the latter.

apwojcik:
@skottmckay  I removed all changes from the C API files, except comment typos. We made OrtMIGraphXProviderOptions deprecated in favour of ProviderOptions.

skottmckay (approved):

nieubank (dismissed):

onnxruntime/python/onnxruntime_pybind_ortvalue.cc
```diff
@@ -421,21 +421,39 @@ void addOrtValueMethods(pybind11::module& m) {
       // Converts Tensor into a numpy array
       .def("numpy", [](const OrtValue* ml_value) -> py::object {
         ORT_ENFORCE(ml_value->IsTensor(), "Only OrtValues that are Tensors are convertible to Numpy objects");
-
+        const auto& device = ml_value->Get<Tensor>().Location().device;
```
skottmckay: Based on the build failures in #25683 this might be needed. Do you want to merge that changes from that PR into here so everything goes in at once?

```suggestion
        [[maybe_unused]] const auto& device = ml_value->Get<Tensor>().Location().device;
```

nieubank (approved):

adrianlizarraga (approved):

onnxruntime/core/providers/migraphx/migraphx_call.cc
adrianlizarraga: snprintf was replaced with use of std::stringstream. Marking this as resolved.

PR #25584, opened 2025-07-29
[QNN EP] Bug fix: multiple consumer for cast result in name conflicts

qti-yuduo:
Bug fix for QNN EP when multiple consumer of same cast node inserted for Gather indices. Now conditionally skip adding the node when already happen.

### DIFF
--- a/onnxruntime/test/onnx/main.cc
+++ b/onnxruntime/test/onnx/main.cc
@@ -916,7 +916,10 @@ select from 'TF8', 'TF16', 'UINT8', 'FLOAT', 'ITENSOR'. \n)");
         ORT_TSTR("sce_none_weights_log_prob_expanded"),
         ORT_TSTR("sce_none_weights_expanded"),
         ORT_TSTR("convtranspose_3d"),
-        ORT_TSTR("gather_elements_negative_indices")};
+        ORT_TSTR("gather_elements_negative_indices"),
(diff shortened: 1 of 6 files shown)


onnxruntime/test/onnx/main.cc
```diff
@@ -916,7 +916,10 @@ select from 'TF8', 'TF16', 'UINT8', 'FLOAT', 'ITENSOR'. \n)");
         ORT_TSTR("sce_none_weights_log_prob_expanded"),
         ORT_TSTR("sce_none_weights_expanded"),
         ORT_TSTR("convtranspose_3d"),
-        ORT_TSTR("gather_elements_negative_indices")};
+        ORT_TSTR("gather_elements_negative_indices"),
+        ORT_TSTR("rotary_embedding_3d_input_expanded"),
+        ORT_TSTR("rotary_embedding_expanded"),
+        ORT_TSTR("rotary_embedding_interleaved_expanded")};
```
qti-yuduo, line 922: Hi @HectorSVC, assuming it's okay to temporarily disable these 3 test cases? 

Previously these are also failures, they were surfaced as `Not Implemented` due to the compile error. Now the `Gather` op compile issue fixed, it compiles but some *other* portions of the graph failed at runtime, thus, showing up as new failures. 

It requires some additional digging so I mark skip for now.

HectorSVC (approved):

PR #25589, opened 2025-07-30
Cache opSupportLimits to improve the performance and update tracing e…

qwu16:
### Description
Cached opSupportLimits in webnn backend and avoid quering it from lower layer each time to improve the performance. Update the trace event in data transfer.

### Motivation and Context
In current implementation, each time calling ensureTensor API to check input/output tensor, MLContext.opSupportLimits API will be called to query support ops capability from chromium and this function call will be the hotspot. Call this API when session is created and then cache it will avoid the frequent lower API call.

### DIFF
--- a/js/web/lib/wasm/jsep/backend-webnn.ts
+++ b/js/web/lib/wasm/jsep/backend-webnn.ts
@@ -177,7 +177,6 @@ export class WebNNBackend {
     sessionIds.add(sessionId);
 
     if (!this.mlOpSupportLimitsBySessionId.has(sessionId)) {
-      // If support limits are not cached, get them from the context.
       this.mlOpSupportLimitsBySessionId.set(sessionId, mlContext.opSupportLimits());
     }
 


qwu16:
@Honry PTAL, thanks!

Honry (approved):
LGTM, thanks!

qwu16:
@fdwr PTAL, thanks~

fdwr:
> In current implementation, each time calling ensureTensor API to check input/output tensor, MLContext.opSupportLimits API will be called to query support ops capability from chromium and this function call will be the hotspot. 

Are there any hotspots in the Chromium implementation that warrant improvement (reducing the need for this cache)? It appears the code after still goes through an `MLOpSupportLimits`, but just avoids the call to `this.mlContextBySessionId.get(sessionId).opSupportLimits()`.

fdwr (approved):
👍

qwu16:
> > In current implementation, each time calling ensureTensor API to check input/output tensor, MLContext.opSupportLimits API will be called to query support ops capability from chromium and this function call will be the hotspot.
> 
> Are there any hotspots in the Chromium implementation that warrant improvement (reducing the need for this cache)? It appears the code after still goes through an `MLOpSupportLimits`, but just avoids the call to `this.mlContextBySessionId.get(sessionId).opSupportLimits()`.

In current Chromium implementation, queried properties for different backend have been cached in MLContext, so when calling MLContext.opSupportLimits, there is no extra IPC between render and GPU process. But in https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/ml/ml_context.cc;l=241, hundreds of op limit properties will be set before return, that takes some time. This PR will cache the MLOpSUpportLimits object and avoid setting hundreds of op limit settings in calling MLContext::opSupportLimits.

PR #25590, opened 2025-07-30
add session_id_ to LogEvaluationStart/Stop, LogSessionCreationStart

xieofxie:
### Description

use session id to track them with LogSessionCreation

if we call Run in different threads, we could differentiate them with thread id given Run is not async


snnn (approved):

snnn:

PR #25372, opened 2025-07-11
add sliding window support for webgpu gqa

guschmue:

PR #25594, opened 2025-07-30
[webgpu] Apply Flash Attention if sliding window exceeds KV cache length

daijh:
### Description

#25372 adds sliding window support for Group Query Attention, disabling Flash Attention as it's not yet supported.

This PR adds a check for the sliding window and applies Flash Attention when the window size exceeds the KV cache length or total sequence length.

### Motivation and Context
See above.

### DIFF
--- a/onnxruntime/contrib_ops/webgpu/bert/flash_attention.cc
+++ b/onnxruntime/contrib_ops/webgpu/bert/flash_attention.cc
@@ -910,9 +910,7 @@ Status ComputeFlashAttentionDecodeVxReduce(onnxruntime::webgpu::ComputeContext& 
 
 Status ApplyFlashAttention(const Tensor* Q, const Tensor* K, const Tensor* V, const Tensor* attention_bias,
                            Tensor* output, const Tensor* past_key, Tensor* present_key, const Tensor* past_value, Tensor* present_value,
-                           const WebgpuAttentionParameters& parameters, onnxruntime::webgpu::ComputeContext& context, int local_window_size) {
-  ORT_ENFORCE(local_window_size == -1, "Sliding window is not supported yet in FlashAttention.");
-
+                           const WebgpuAttentionParameters& parameters, onnxruntime::webgpu::ComputeContext& context) {
   ORT_RETURN_IF_ERROR(CopyKVCache(context, parameters, K, past_key, present_key, V, past_value, present_value));
   const int present_sequence_length = parameters.is_gqa_ ? parameters.seqlen_present_kv_cache_ : parameters.total_sequence_length_;
   if (parameters.sequence_length_ > 1) {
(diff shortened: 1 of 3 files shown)


daijh:
@guschmue @fs-eire @qjia7

onnxruntime/contrib_ops/webgpu/bert/group_query_attention.cc
```diff
@@ -198,9 +198,10 @@ Status GroupQueryAttention::ComputeInternal(onnxruntime::webgpu::ComputeContext&
   Tensor* present_value = context.Output(2, present_kv_shape);
   parameters.past_present_share_buffer_ = present_key != nullptr && present_value != nullptr && past_key != nullptr && past_value != nullptr && past_key->DataRaw() == present_key->DataRaw() && past_value->DataRaw() == present_value->DataRaw();
 
+  bool use_sliding_window = (local_window_size_ != -1 && local_window_size_ < parameters.seqlen_present_kv_cache_ && local_window_size_ < parameters.total_sequence_length_);
```
qjia7: Maybe `bool use_sliding_window = (local_window_size_ != -1 && local_window_size_ < parameters.total_sequence_length_);` is enough.

daijh: It is unclear if there are conditions where `seqlen_present_kv_cache_` < `total_sequence_length_`.
If no, we can simply remove it.

qjia7: `seqlen_present_kv_cache_`'s value either is the max_sequence_length or the past_sequence_length +kv_sequence_length (total_sequence_length). So I think it's safe otherwise the current fa has problem since it's based on this assumption.

onnxruntime/contrib_ops/webgpu/bert/flash_attention.cc
```diff
@@ -910,7 +910,9 @@ Status ComputeFlashAttentionDecodeVxReduce(onnxruntime::webgpu::ComputeContext&
 
 Status ApplyFlashAttention(const Tensor* Q, const Tensor* K, const Tensor* V, const Tensor* attention_bias,
                            Tensor* output, const Tensor* past_key, Tensor* present_key, const Tensor* past_value, Tensor* present_value,
-                           const WebgpuAttentionParameters& parameters, onnxruntime::webgpu::ComputeContext& context) {
+                           const WebgpuAttentionParameters& parameters, onnxruntime::webgpu::ComputeContext& context, int local_window_size) {
+  ORT_ENFORCE(local_window_size == -1, "Sliding window is not supported yet in FlashAttention.");
```
qjia7: Even we go to this path, it doesn't mean that local_window_size is -1. It's not consistent with your check `!use_sliding_window`. Maybe it's ok to keep flash_attention unchanged.

onnxruntime/contrib_ops/webgpu/bert/group_query_attention.cc
daijh: Add `ENFORCE` to guard this assumption.

qjia7 (approved):
LGTM, thanks.

daijh:
CI infra issue by check the logs, please help to re-run.

PR #25597, opened 2025-07-30
Add reduceSum support for uint32 and uint64

konstantin-pueckler-qc:
### Description
This PR adds ``uint32`` and ``uint64`` kernels for the ``reduceSum`` operator.

### Motivation and Context
We ran into the missing kernels in [ndonnx](https://github.com/Quantco/ndonnx). Especially the support for ``uint64`` is important for us since otherwise we need a workaround consisting of a lossy cast from ``uint64`` to ``double``.

### DIFF
--- a/onnxruntime/test/providers/cpu/reduction/reduction_ops_test.cc
+++ b/onnxruntime/test/providers/cpu/reduction/reduction_ops_test.cc
@@ -2356,20 +2356,37 @@ TEST(ReductionOpTest, ReduceSum_int32) {
   test.Run();
 }
 
+TEST(ReductionOpTest, ReduceSum_uint32) {
+  OpTester test("ReduceSum");
+  test.AddAttribute("axes", std::vector<int64_t>{0, 2});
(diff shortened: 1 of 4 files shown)


konstantin-pueckler-qc:
@microsoft-github-policy-service agree company="QuantCo"

onnxruntime/test/providers/cpu/reduction/reduction_ops_test.cc
```diff
@@ -2356,6 +2356,23 @@ TEST(ReductionOpTest, ReduceSum_int32) {
   test.Run();
 }
 
+TEST(ReductionOpTest, ReduceSum_uint64) {
```
xadupre, line 2376: You should add a test for the uint32 as well.

konstantin-pueckler-qc: thank you for the remark, now it has a test for uint32 as well

konstantin-pueckler-qc:
superseded by #25734

PR #23548, opened 2025-01-31
remove log spam from cpuinfo

jywu-msft:
cpuinfo outputs error when cpu is not recognized. 
this has been a longstanding issue e.g. 
https://github.com/microsoft/onnxruntime/issues/21947
https://github.com/microsoft/onnxruntime/issues/21393

this issue has been exacerbated by https://github.com/microsoft/onnxruntime/pull/22856
this change https://github.com/microsoft/onnxruntime/blob/4fa0f1e0edb43141c68302859e410f2ec1232c3a/onnxruntime/core/mlas/lib/qnbitgemm_kernel_neon.cpp#L189 causes the messages to appear during static initialization.

this means for python, when you import onnxruntime you immediately see the errors.

```
>>> import onnxruntime
Error in cpuinfo: Unknown chip model name 'snapdragon (tm) 8cx gen 3 @ 3.40 GHz'.
Please add new Windows on Arm SoC/chip support to arm/windows/init.c!
unknown Qualcomm CPU part 0x1 ignored
unknown Qualcomm CPU part 0x1 ignored
unknown Qualcomm CPU part 0x1 ignored
unknown Qualcomm CPU part 0x1 ignored
unknown Qualcomm CPU part 0x1 ignored
unknown Qualcomm CPU part 0x1 ignored
unknown Qualcomm CPU part 0x1 ignored
unknown Qualcomm CPU part 0x1 ignored
unknown Qualcomm CPU part 0x1 ignored
unknown Qualcomm CPU part 0x1 ignored
unknown Qualcomm CPU part 0x1 ignored
unknown Qualcomm CPU part 0x1 ignored
```

Fix is to patch pytorch_cpuinfo and to comment out std::cerr lines in cpuid_uarch.cc 
the errors are not actionable by the user, so they should not be emitted.

tested that after these changes, these errors no longer show up.

PR #25601, opened 2025-07-30
[build] fix build break on arm

fs-eire:
### Description

Remove the out-of-dated patch file.


qti-kromero:
this is causing us issues with our internal builds at QCOM - is there a reason this wasn't caught by the Azure pipeline?

fs-eire:
> this is causing us issues with our internal builds at QCOM - is there a reason this wasn't cause by the Azure pipeline?

I think it may be because of unexpected behavior of the source cache of the pipeline

cmake/external/onnxruntime_external_deps.cmake
```diff
@@ -366,25 +366,13 @@ if (CPUINFO_SUPPORTED)
...
   set(CPUINFO_BUILD_MOCK_TESTS OFF CACHE INTERNAL "")
   set(CPUINFO_BUILD_BENCHMARKS OFF CACHE INTERNAL "")
-  if (onnxruntime_target_platform STREQUAL "ARM64EC" OR onnxruntime_target_platform STREQUAL "ARM64")
-      message(STATUS "Applying a patch for Windows ARM64/ARM64EC in cpuinfo")
-      onnxruntime_fetchcontent_declare(
-        pytorch_cpuinfo
-        URL ${DEP_URL_pytorch_cpuinfo}
-        URL_HASH SHA1=${DEP_SHA1_pytorch_cpuinfo}
-        EXCLUDE_FROM_ALL
-        PATCH_COMMAND ${Patch_EXECUTABLE} -p1 < ${PROJECT_SOURCE_DIR}/patches/cpuinfo/9bb12d342fd9479679d505d93a478a6f9cd50a47.patch
```
adrianlizarraga, line 376: There are two different aspects to this patch file:
1. patch `include/cpu_info.h` to account for arm64EC when defining preprocessor macros like `CPUINFO_ARCH_*`.
2. patch `src/arm/windows/init.c` to remove the log spam when importing onnxruntime in python. (fyi @jywu-msft)
    - ORT PR: https://github.com/microsoft/onnxruntime/pull/23548
    - cpuinfo PR that changed this: https://github.com/pytorch/cpuinfo/pull/283

I think it is only 2 that is out-of-date and we would need to evaluate the new version of cpuinfo to see if it still spams log messages. Perhaps this can be handled separately?

However, we probably should keep 1 for this PR. Would require remaking that patch file.

fs-eire: makes sense. let me update the PR.

adrianlizarraga: I built this branch locally on my arm64 machine. I can confirm that it builds and that the new cpuinfo version does not spam log messages.

adrianlizarraga (approved):

qti-kromero (approved):

PR #25608, opened 2025-07-31
Refactor Java Test Pipeline

snnn:
### Description
We have a big packaging pipeline that build nuget/java/nodejs packages. After that we run these. This PR split the tests to a dedicated pipeline and refactored the code that use maven to download deps instead of using direct HTTP fetch. The new approach allows us to use Azure DevOps artifacts as an internal mirror to meet network isolation requirements.
This PR also updated tools/python/run_packaging_pipelines.py a little bit to add the support for RC releases.

### Motivation and Context
Make the packaging pipelines smaller and easier to use.


tools/ci_build/github/azure-pipelines/c-api-noopenmp-test-pipelines.yml
```diff
@@ -121,109 +121,6 @@ stages:
   parameters:
     StageSuffix: 'macOS_CPU_x64'
 
-- template: templates/final-jar-testing.yml
```
snnn, line 124: This part is moved to jar_package_testing.yml as a new pipeline.

tools/ci_build/github/azure-pipelines/c-api-noopenmp-test-pipelines.yml
```diff
@@ -394,7 +294,7 @@ stages:
           displayName: 'test'
           inputs:
             scriptPath: '$(Build.SourcesDirectory)\tools\ci_build\build.py'
-            arguments: '--config RelWithDebInfo --use_binskim_compliant_compile_flags --enable_lto --disable_rtti --build_dir $(Build.BinariesDirectory) --skip_submodule_sync --build_shared_lib --test --enable_onnx_tests  $(TelemetryOption) '
+            arguments: '--config RelWithDebInfo --use_binskim_compliant_compile_flags --enable_lto --disable_rtti --build_dir $(Build.BinariesDirectory) --skip_submodule_sync --build_shared_lib --test --enable_onnx_tests'
```
carzh, line 297: Not too familiar with the telemetry steps + telemetryoption but was wondering why it's being removed

tools/ci_build/github/azure-pipelines/templates/final-jar-testing-linux.yml
```diff
@@ -0,0 +1,104 @@
+parameters:
+- name: OS
+  displayName: Opserating System
```
carzh: tiny nit: typo

tools/python/run_packaging_pipelines.py
```diff
@@ -230,8 +232,24 @@ def evaluate_single_pipeline(
...
+                if packaging_type == "release":
+                    yaml_params = data.get("parameters", [])
+                    if isinstance(yaml_params, list):
+                        param_names = {p.get("name") for p in yaml_params if isinstance(p, dict)}
+                        if (
+                            "PreReleaseVersionSuffixString" in param_names
+                            and "PreReleaseVersionSuffixNumber" in param_names
+                        ):
+                            has_pre_release_params = True
+                            print("  - OK: Found pre-release versioning parameters.")
```
carzh, line 245: nit: in the rare case where PreReleaseVersionSuffixString is set to something but PreReleaseVersionSuffixNumber is not set, then has_pre_release_params would be false -- later in line 343, then neither the suffix string nor suffix number would be appended -- not sure if this is the desired behavior though

tools/ci_build/github/azure-pipelines/c-api-noopenmp-test-pipelines.yml
snnn: The telemetryoption  variable is either empty or equals to "--use_telemetry" which is a flag in build.py.  It is only used for generating cmake files and build the code. It is not needed when running tests.

tools/ci_build/github/azure-pipelines/templates/final-jar-testing-linux.yml
snnn: Fixed.

tools/python/run_packaging_pipelines.py
snnn: Added a check there.

tools/ci_build/github/linux/java_linux_final_test.sh
```diff
@@ -20,23 +20,22 @@ EXIT_CODE=1
...
+java -version
 
 echo "Directories created"
 echo  "Library path:" "$LD_LIBRARY_PATH"
 
-pushd test
-jar xf "$BINARY_DIR/final-jar/testing.jar"
-popd
-
-curl -O -sSL https://oss.sonatype.org/service/local/repositories/releases/content/org/junit/platform/junit-platform-console-standalone/1.6.2/junit-platform-console-standalone-1.6.2.jar
```
snnn, line 34: Mainly, this change is to avoid directly downloading such files.  Use maven package manager instead.

carzh (approved):
approving since looks like test run is successful on this branch

fs-eire (approved):

PR #25610, opened 2025-07-31
[webgpu] And int64 to cast

qjia7:
This pull request extends the WebGPU execution provider to support int64 data type casting in the `Cast` operator, with conditional support based on whether graph capture is enabled. It refactors kernel registration to allow toggling int64 support and updates the shader code and kernel logic to handle int64 tensors efficiently.

It's part of the work to enable graph capture in phi4 https://github.com/microsoft/onnxruntime/pull/25868

### DIFF
--- a/onnxruntime/core/providers/webgpu/webgpu_execution_provider.cc
+++ b/onnxruntime/core/providers/webgpu/webgpu_execution_provider.cc
@@ -871,22 +871,12 @@ std::vector<std::unique_ptr<ComputeCapability>> WebGpuExecutionProvider::GetCapa
 }
 
 std::shared_ptr<KernelRegistry> WebGpuExecutionProvider::GetKernelRegistry() const {
-  // Use different registries based on graph capture configuration
-  static std::shared_ptr<KernelRegistry> registry_with_graph_capture;
-  static std::shared_ptr<KernelRegistry> registry_without_graph_capture;
-  static std::once_flag init_flag_with_graph_capture;
-  static std::once_flag init_flag_without_graph_capture;
-
   if (enable_graph_capture_) {
-    std::call_once(init_flag_with_graph_capture, []() {
-      registry_with_graph_capture = webgpu::RegisterKernels(true);
-    });
-    return registry_with_graph_capture;
+    static std::shared_ptr<KernelRegistry> registry = webgpu::RegisterKernels(true);
+    return registry;
   } else {
-    std::call_once(init_flag_without_graph_capture, []() {
-      registry_without_graph_capture = webgpu::RegisterKernels(false);
-    });
-    return registry_without_graph_capture;
+    static std::shared_ptr<KernelRegistry> registry = webgpu::RegisterKernels(false);
+    return registry;
   }
 }
 


qjia7:
Sorry, please hold on the review. I may miss some situations.

qjia7:
@guschmue @fs-eire This PR is ready for review. Please take a look, thanks.
cc @sushraja-msft @xhcao

xhcao:
Hi, @qjia7 , is it not very reasonable that bind the usage of Cast(tensor_int64) and `enable_graph_capture` together, sometimes someone only want (or only can use) Cast(tensor_int64)?
Why not add an extra string list option in `SessionOptions`,  for example add `enable_cast_int64` into the string list if want Cast(tensor_int64), and query the string when register EP kernels? It also easily extend to `enable_pad_int64`, `enable_gathernd_int64`, etc.
It is my personal point. If other reviewers think the current code is fine, it is also ok for me.

onnxruntime/core/providers/webgpu/webgpu_execution_provider.cc
```diff
@@ -869,9 +871,17 @@ std::vector<std::unique_ptr<ComputeCapability>> WebGpuExecutionProvider::GetCapa
 }
 
 std::shared_ptr<KernelRegistry> WebGpuExecutionProvider::GetKernelRegistry() const {
-  static std::shared_ptr<KernelRegistry> registry = webgpu::RegisterKernels();
+  // Use different registries based on graph capture configuration
+  static std::shared_ptr<KernelRegistry> registry_with_int64;
+  static std::shared_ptr<KernelRegistry> registry_without_int64;
```
fs-eire: change name to registry_with{out}_graph_capture

also, would be good if support lazy load for either (use 2 once_flag)

qjia7:
> Hi, @qjia7 , is it not very reasonable that bind the usage of Cast(tensor_int64) and `enable_graph_capture` together, sometimes someone only want (or only can use) Cast(tensor_int64)? Why not add an extra string list option in `SessionOptions`, for example add `enable_cast_int64` into the string list if want Cast(tensor_int64), and query the string when register EP kernels? It also easily extend to `enable_pad_int64`, `enable_gathernd_int64`, etc. It is my personal point. If other reviewers think the current code is fine, it is also ok for me.

Hi Xinghua, this change is target for graph capture and doesn't affect the existing logic. I understand other cases may also need the cast_int64 support in webgpu. But I'm unsure if a session option is the best solution since many other ops may also need such kind of changes. I also offline talked with @fs-eire. He is considering an alternative way to support int64 in GetCapability. Therefore, we need to consider more factors. I prefer to proceed with this PR to unblock progress. Even if the final solution involves adding a session option, it won't affect the main changes of this PR—only the dynamic registration conditions would need to be modified.

onnxruntime/core/providers/webgpu/webgpu_execution_provider.cc
```diff
@@ -872,16 +872,22 @@ std::vector<std::unique_ptr<ComputeCapability>> WebGpuExecutionProvider::GetCapa
...
+      registry_with_graph_capture = webgpu::RegisterKernels(true);  // with int64 support
+    });
+    return registry_with_graph_capture;
+  } else {
+    std::call_once(init_flag_without_graph_capture, []() {
+      registry_without_graph_capture = webgpu::RegisterKernels(false);  // without int64 support
+    });
+    return registry_without_graph_capture;
+  }
 }
```
fs-eire, line 881:
```suggestion
std::shared_ptr<KernelRegistry> WebGpuExecutionProvider::GetKernelRegistry() const {
  if (enable_graph_capture_) {
    static std::shared_ptr<KernelRegistry> registry = webgpu::RegisterKernels(true);
    return registry;
  } else {
    static std::shared_ptr<KernelRegistry> registry = webgpu::RegisterKernels(false);
    return registry;
  }
}
```

PR #25615, opened 2025-07-31
[build] fix macOS x86_64 cross-compile warning

fs-eire:
### Description

This pull request introduces a new compiler flag check for `-Warray-bounds` and addresses a Clang-specific warning in the `MlasQ4Int8TileGemmKernelBlkLen32Avx2` function.

The warning is generated in unreachable code, however it is still triggered the warning. Anyway add the change to make Clang happy.

Fixes #23180

PR #25461, opened 2025-07-20
upgrade dawn to 794b6fadc4171f7b853a77ffdf0948fbec431f41

fs-eire:
### Description

upgrade dawn to latest

PR #25617, opened 2025-07-31
[build] use cross-compile to build macOS x86_64 target for WebGPU

fs-eire:
### Description

use cross-compile to build x86_64 target for WebGPU. This is required for incoming Dawn upgrade (https://github.com/microsoft/onnxruntime/pull/25461). Latest dawn need at least Clang v16 to compile, but available macOS in CI pipelines are up to v15.2.

This PR depends on #25615.


.github/workflows/mac.yml
```diff
@@ -23,32 +23,42 @@ jobs:
     uses: ./.github/workflows/macos-ci-build-and-test-workflow.yml
     with:
       # Only build arm64 for CPU
-      matrix_exclude: >-
+      matrix_include: >-
```
snnn, line 26: Do we have enough machines for the new job configs?

fs-eire:
merged latest main and reverted packaging pipeline changes. now this PR only contains changes for CI pipeline

tianleiwu (approved):

PR #25618, opened 2025-07-31
Reduce CMake's CUDA_ARCHITECTURES

snnn:
Our python and CUDA packages are over size. Need to reduce the size.


tools/ci_build/github/azure-pipelines/stages/nuget-win-cuda-packaging-stage.yml
```diff
@@ -45,7 +45,7 @@ stages:
     msbuildPlatform: x64
     packageName: x64-cuda
     CudaVersion: ${{ parameters.CudaVersion }}
-    buildparameter: --use_cuda --cuda_home=${{ parameters.win_cuda_home }} --enable_onnx_tests --enable_wcos --cmake_extra_defines "CMAKE_CUDA_ARCHITECTURES=52-real;61-real;75-real;86-real;89-real;90a-virtual"
+    buildparameter: --use_cuda --cuda_home=${{ parameters.win_cuda_home }} --enable_onnx_tests --enable_wcos --cmake_extra_defines "CMAKE_CUDA_ARCHITECTURES=70-real;75-real;80-real;90a-real;90a-virtual"
```
tianleiwu, line 48: The original ones are optimized for major consumer GPUs. Like GTX 980 (52), RTX 1080 (61), RTX 2080 (75), RTX 3080 (86), RTX 4080 (89).

70/80/90 are data center GPU V100/A100/H100, which usually not run in Windows.

snnn: So we want to use different ones for Linux/Windows?

snnn: What do you suggest? Would you like to take over this task?

tianleiwu: Right, currently we use different ones for Linux/Windows.

You can just remove one by one from the left side, until the binary size is ok. I guess it only need one or two tries.

snnn:
Abandoned.

PR #25619, opened 2025-08-01
Update MoE and qMoE spec

tianleiwu:
### Weight Shape Update
Make sure the shape reflects actual memory layout. The weight is stored in column major.

### Add support for SwiGLU activation attributes
Add spec for the new activation type SwiGLU (Swish-Gated Linear Unit) by introducing a few new attributes. For reference, see the [Triton kernel implementation](https://github.com/triton-lang/triton/blob/main/python/triton_kernels/triton_kernels/swiglu.py).

#### New Attributes for SwiGLU

* **`swiglu_fusion`**:

  * `0`: Not fused — two separate GEMMs (FC1 and FC3).
  * `1`: Fused GEMMs using **interleaved** format (g and l are interleaved per row).
  * `2`: Fused GEMMs using **non-interleaved** (concatenated) format.

* **`swiglu_limit`**: Clamp threshold applied to `g` and `l`.

* **`activation_alpha`**: Scalar multiplier applied to `g` before sigmoid.

* **`activation_beta`**: Added to `l` before the final output computation.

---

### SwiGLU Activation Function

The SwiGLU function is defined as:

```
g = xW + b
l = xV + c
G = min(g, limit)
L = max(min(l, limit), -limit)
swiglu = G * sigmoid(alpha * G) * (L + beta)
```

* `x`: Input
* `W`, `V`: Weight matrices
* `b`, `c`: Bias vectors
* `alpha`, `beta`, `limit`: Float constants

---

### Fusion Behavior

* When `swiglu_fusion = 0`:

  * Two GEMMs are computed independently.
  * FC1 → computes `g`, FC3 → computes `l`.

* When `swiglu_fusion = 1`:

  * `g` and `l` are computed in a **single fused GEMM** (FC1).
  * Output is **interleaved** per row as: `gate, linear, gate, linear, ...`.

* When `swiglu_fusion = 2`:

  * `g` and `l` are computed in a single GEMM (FC1).
  * Output is **concatenated** per row: `[g | l]`.

### Implement swiglu_limit for CUDA
Update CUDA kernel to use default swiglu limit.
Update test_moe_cuda.py to have same logic in reference implementation.

### Remaining Works
The main purpose of this PR is to update spec instead of implementing them.
Note that MoE/qMoE ops and tests still use hard-coded parameters and will be changed later to read from those attributes.

Column-wise symmetric quantization is used for qMoE. We will add more quantization details when we add support of block-wise quantization soon.


onnxruntime/core/graph/contrib_ops/contrib_defs.cc
```diff
@@ -1429,6 +1445,10 @@ ONNX_MS_OPERATOR_SET_SCHEMA(
               "Number of bits used in quantized weights. Default is 4 bits",
               AttributeProto::INT,
               static_cast<int64_t>(4))
+        .Attr("swiglu_fusion", "0: not fused, 1: fused and interleaved. 2: fused and not interleaved.", AttributeProto::INT, static_cast<int64_t>(0))
```
kunal-vaishnavi, line 1448: Should we make these attributes optional?

tianleiwu: Yes, it is optional since it has default value. You can see that one test does not add any attribute:
https://github.com/microsoft/onnxruntime/blob/0635f1135399288e55212321e27a1935e0a2cac0/onnxruntime/test/python/transformers/test_moe_cuda.py#L172-L180

However, the default value is for non-fused and traditional formula.

onnxruntime/test/python/transformers/test_moe_cuda.py
```diff
@@ -57,11 +57,13 @@
 
 
 def quant_dequant(weights, is_4_bit_quantization: bool = True):
-    # use the test version `_symmetric_...` to get the non-interleaved weights
     type = torch.quint4x2 if is_4_bit_quantization else torch.int8
-    # This import is needed to use torch.ops.trtllm._symmetric_quantize_last_axis_of_batched_matrix()
-    # Comment out this line for passing the lintrunner check in the CI.
-    # import tensorrt_llm
+
+    import tensorrt_llm  # noqa: PLC0415
```
kunal-vaishnavi, line 62: Should we wrap this in a try-except if `tensorrt_llm` is not installed?

tianleiwu: This function will not run in CI pipeline, so it is fine. If you run in local, you will need it anyway to test qMoE.

onnxruntime/core/graph/contrib_ops/contrib_defs.cc
kunal-vaishnavi: To clarify, should we add `OPTIONAL_VALUE`? Otherwise, the attributes will still have to be set when creating the op.

kunal-vaishnavi (approved):

PR #25622, opened 2025-08-01
[QNN-EP] Resolve VTCM buffer sharing bugs

quic-calvnguy:
### Description
 - Pre-allocated memory for HTP context params list during context creation when VTCM backup buffer sharing is enabled. This is done to avoid memory-based issues due to vector resizing/re-allocation.
 - Handle case where new binary contexts need to be processed

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
@@ -787,10 +787,12 @@ Status QnnBackendManager::CreateContextVtcmBackupBufferSharingEnabled(std::unord
 
   std::vector<QnnContext_Params_t> context_params_list;
   std::vector<QnnContext_ParamsV1_t> context_paramsv1_list;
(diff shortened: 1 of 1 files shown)


onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1178,6 +1180,14 @@ Status QnnBackendManager::SetupBackend(const logging::Logger& logger,
 
 #if QNN_API_VERSION_MAJOR == 2 && (QNN_API_VERSION_MINOR >= 26)
     if (vtcm_backup_buffer_sharing_enabled_) {
+      // If a context bin filepath has not been processed yet,
+      // then a new context must be created for the set of context bins
+      auto first_mapping_it = ep_context_handle_map_.find(context_bin_map.begin()->first);
```
quic-calvnguy, line 1185: The assumption here is that if the model points to any EP context binary that has not been processed, the entire set contains new EP context binaries. Logically, a model shouldn't be pointing to both processed and unprocessed EP context binaries.

HectorSVC (approved):
:shipit:

PR #25625, opened 2025-08-01
[MIGraphX EP] Fix CreateExecutionProviderFactory with correct struct and change vendor_id

Zhaeong:
This fixes the CreateIExecutionProvider for MIGraphX EP when calling CreateExecutionProviderFactory, using     OrtMIGraphXProviderOptions instead of ProviderOptions.
Also changes the vendor_id so that OrderDevices in provider_policy_context.cc will default dml ep when ep_policy is set to GPU. Will update pending more changes to MIGraphX EP.


psakhamoori:
@snnn Please help review this PR.

Issue #25609, opened 2025-07-31
[Build] CUDAExecutionProvider fails to load due to missing libcudnn.so.9 in LD_LIBRARY_PATH when using onnxruntime-gpu==1.22.0


yuxuan-z19:
### Describe the issue

When using `onnxruntime-gpu==1.22.0` alone, running inference with `providers=["CUDAExecutionProvider"]` fails at runtime with the following error:

```bash
2025-07-31 15:09:16.647314988 [E:onnxruntime:Default, provider_bridge_ort.cc:2195 TryGetProviderInfo_CUDA] /onnxruntime_src/onnxruntime/core/session/provider_bridge_ort.cc:1778 onnxruntime::Provider& onnxruntime::ProviderLibrary::Get() [ONNXRuntimeError] : 1 : FAIL : Failed to load library libonnxruntime_providers_cuda.so with error: libcudnn.so.9: cannot open shared object file: No such file or directory

2025-07-31 15:09:16.647341662 [W:onnxruntime:Default, onnxruntime_pybind_state.cc:1055 CreateExecutionProviderInstance] Failed to create CUDAExecutionProvider. Require cuDNN 9.* and CUDA 12.*. Please install all dependencies as mentioned in the GPU requirements page (https://onnxruntime.ai/docs/execution-providers/CUDA-ExecutionProvider.html#requirements), make sure they're in the PATH, and that your GPU is supported.
```

This happens even when the appropriate version of cuDNN (installed via PyPI, i.e., nvidia-cudnn-cu12) is already present in the environment.

The root cause appears to be that the cuDNN shared library path (e.g., `libcudnn.so.9`) is not added to `LD_LIBRARY_PATH`. In contrast, PyTorch dynamically patches `LD_LIBRARY_PATH` to include the path of nvidia.cudnn, which prevents this error.

A minimal workaround is to manually patch LD_LIBRARY_PATH at runtime, as follows:

```python
def _patch_cudnn_ld_lib_path():
    if os.environ.get("DIFFONNX_PATCHED") == "1":
        return

    try:
        import nvidia.cudnn
    except ImportError:
        return

    cudnn_lib = pathlib.Path(nvidia.cudnn.__file__).parent / "lib"
    if cudnn_lib.exists():
        cudnn_lib = str(cudnn_lib.resolve())
        old_ld = os.environ.get("LD_LIBRARY_PATH", "")
        if cudnn_lib not in old_ld.split(":"):
[...]

PR #25628, opened 2025-08-01
fix(ort): add automatic patching for nvidia cudnn library path

MengAiDev:
### Description
- Automatically add nvidia library paths to LD_LIBRARY_PATH on Linux when using CUDAExecutionProvider
- This patch addresses potential issues with missing cudnn libraries in LD_LIBRARY_PATH
- The patch is applied only once and only on Linux systems
- It checks for the presence of the nvidia package and cudnn library path before applying the patch

### Motivation and Context
[](https://github.com/microsoft/onnxruntime/issues/25609)

### DIFF
--- a/onnxruntime/__init__.py
+++ b/onnxruntime/__init__.py
@@ -11,6 +11,67 @@ or the `Github project <https://github.com/microsoft/onnxruntime/>`_.
 __version__ = "1.23.0"
 __author__ = "Microsoft"
 
+import os
+import platform
+import sys
+
+# Function to automatically patch LD_LIBRARY_PATH for nvidia-cudnn libraries
+def _patch_nvidia_library_path():
+    """Automatically add nvidia library paths to LD_LIBRARY_PATH on Linux when using CUDAExecutionProvider."""
+    # Only apply this patch on Linux systems
+    if platform.system() != "Linux":
+        return
+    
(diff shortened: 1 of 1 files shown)


yuxuan-z19:
This `os` import **clearly breaks** the C-API abstraction — ONNX Runtime's C interface shouldn't rely on Python internals. Maybe consider a CMake-level patch instead, like [PyTorch's approach](https://github.com/pytorch/pytorch/tree/main/cmake/Modules_CUDA_fix).

By the way, I noticed that the recent code appears to be **directly copied** from my [original implementation](https://github.com/yuxuan-z19/diffonnx/blob/main/src/diffonnx/utils.py#L410) **without attribution**. Could you please explicitly credit the source or add a reference note? That would be appreciated.

MengAiDev:
> This `os` import **clearly breaks** the C-API abstraction — ONNX Runtime's C interface shouldn't rely on Python internals. Maybe consider a CMake-level patch instead, like [PyTorch's approach]
> 
> By the way, I noticed that the recent code appears to be **directly copied** from my [original implementation](https://github.com/yuxuan-z19/diffonnx/blob/main/src/diffonnx/utils.py#L410) **without attribution**. Could you please explicitly credit the source or add a reference note? That would be appreciated.

Thanks. I'll add the comment, but actually I haven't copy your code

PR #25635, opened 2025-08-01
[QNN EP] Add support for GatherNd Op in QNN EP

quic-ashwshan:
- Added new op builder for GatherNd Op
- Added unit tests for GatherNd and QDQ tests
- Disabled two tests in ORT test as QNN CPU does not support negative indices

### Description
Adding support for GatherNd op in QNN EP

### Motivation and Context
Currently GatherNd op is not supported in QNN EP and hence falls back to ORT CPU.


HectorSVC (approved):
:shipit:

PR #25626, opened 2025-08-01
Move moving weights to memory to the end of Graph::Resolve()

yuslepukhin:
### Description

Move moving weights to memory to the end of Graph::Resolve().
Modify Inject so it copies data into TensorProto according to the C API docs.

### Motivation and Context

TypeAndShape inference runs as a part of `Resolve()` and it unable to inspect and load the initializers that point to OrtValues at that time. We choose to move TensorProto to OrtValue conversion at the end of `Resolve()`.

References: https://github.com/microsoft/onnxruntime/pull/25579

PR #25579, opened 2025-07-29
Draft: [NV TRT RTX EP] Fix onnx checker for constants in subgraph

gedoensmax:
### Description

This PR is supposed to fix:
[ ] Loading large models using the `AddExternalInitializersFromFilesInMemory` API
[ ] Fix Phi4 128k instruct parsing 

The first issue is resolved by adding a check in `Graph::InjectExternalInitializersFromFilesInMemory(` from what I can tell. The other issue that parsing of the `If` node fails during `Graph::Resolve` within `NvExecutionProvider::GetSupportedList` [here](https://github.com/gedoensmax/onnxruntime/blob/223a1f7bc00adaa5ae6bd033f1a4631b0f623510/onnxruntime/core/graph/graph.cc#L3212). I tried to fix this by loading the external data in memory to raw data.
This did not resolve the error though:
```
←[1;31m2025-07-29 17:00:31.7917863 [E:onnxruntime:, graph.cc:3212 onnxruntime::Graph::VerifyNodeAndOpMatch] This is an i
nvalid model. In Node, ("/model/rotemb_caches_subgraph/If", If, "", -1) : ("/model/rotemb_caches_subgraph/Greater/output
_0": tensor(bool),) -> ("cos_cache": tensor(float16),"sin_cache": tensor(float16),) , Error Data of TensorProto ( tensor
 name: cos_cache_large) should be stored in */_ORT_MEM_ADDR_/*, but it doesn't exist or is not accessible.←[m
```

@chilo-ms @skottmckay would you be able to help out ? My guess is this has something to do with the ORT Graph wrapping.

PR #25640, opened 2025-08-03
Move moving weights to memory to the end of Graph::Resolve() (#25626)

snnn:
### Description

Move moving weights to memory to the end of Graph::Resolve().
Modify Inject so it copies data into TensorProto according to the C API
docs.

### Motivation and Context

TypeAndShape inference runs as a part of `Resolve()` and it unable to
inspect and load the initializers that point to OrtValues at that time.
We choose to move TensorProto to OrtValue conversion at the end of
`Resolve()`.

References: https://github.com/microsoft/onnxruntime/pull/25579

### DIFF
--- a/include/onnxruntime/core/graph/graph.h
+++ b/include/onnxruntime/core/graph/graph.h
@@ -740,7 +740,7 @@ class Graph {  // NOLINT(clang-analyzer-optin.performance.Padding): preserve exi
   common::Status InjectExternalInitializedTensors(const InlinedHashMap<std::string, OrtValue>& external_initializers);
 
   /** This function takes externally provided files in memory for initializers with external
-   *    data and replaces graph initializers with its content.
+   *    data and replaces main graph initializers with its content.
    */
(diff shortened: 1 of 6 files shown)


fs-eire:
Not sure if this failure on MacOS CI (webgpu/x64) is related or not: https://github.com/microsoft/onnxruntime/actions/runs/16700824576/job/47271521600

```
2: libc++abi: terminating due to uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument
 2/10 Test  #2: onnxruntime_test_all ....................Subprocess aborted***Exception: 212.93 sec
```

snnn:
Oh .... Definitely it's a bug. Destruction order problem.

yuslepukhin:
> Oh .... Definitely it's a bug. Destruction order problem.

Which mac is being used for the CI. Is it Intel or M1?

yuslepukhin:
I am approving it. I will be debugging Mac today.

yuslepukhin (approved):

PR #25641, opened 2025-08-03
Update qMoE spec to support block quantization

tianleiwu:
Update operator spec to support block quantization in qMoE.
Implementation will come later.


apsonawane:

PR #25646, opened 2025-08-04
DequantizeLinear should support non-zero zero_point when input type is int32

wangw-1991:
### Description
This PR makes DequantizeLinear support non-zero zero_point when input data type is int32.

### Motivation and Context
For WebNN use case, we have some scenarios that input data type is int32 and the zero_point is not zero for DequantizeLinear.


wangw-1991:
@fdwr please take a look, thanks.

fdwr (approved):
So the existing code in ORT already worked with this case, and this validation just needed to be relaxed? Since there's no test case, did you test this locally via a little ONNX file or via Chromium's ORT backend?

wangw-1991:
> **Wei**: So the existing code in ORT already worked with this case, and this validation just needed to be relaxed? Since there's no test case added, did you test this locally via a little ONNX file or via Chromium's ORT backend?
> 
> FYI @adrianlizarraga and @xadupre.

I tested this via Chromium's ORT backend, because we have four WPTs meet this issue in [https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/web_tests/external/wpt/webnn/conformance_tests/dequantizeLinear.https.any.js](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/web_tests/external/wpt/webnn/conformance_tests/dequantizeLinear.https.any.js), they are: 
`dequantizeLinear int32 1D tensor with float32 1D scale`, `dequantizeLinear with float32 3D scale as an intermediate node`, `dequantizeLinear int32 1D tensor with float16 1D scale` and `dequantizeLinear with float16 3D scale as an intermediate node`

xadupre:
You should add a unit test for your changes and DequantizeLinear should be aligned with this change. Since, the current behaviour is to force zero point to be 0, I doubt an existing unit test breaks. onnx specifications do not specify anything particular (+ @gramalingam). It looks ok to move forward.

fdwr:
> You should add a unit test for your changes and DequantizeLinear should be aligned with this change. Since, the current behaviour is to force zero point to be 0, I doubt an existing unit test breaks. onnx specifications do not specify anything particular (+ @gramalingam). It looks ok to move forward.

👍 See `onnxruntime\test\providers\cpu\tensor\quantize_linear_test.cc` and run with `onnxruntime\build\Windows\Debug\Debug\onnxruntime_test_all.exe --gtest_filter=*TensorOpTest.Deq*`.

wangw-1991:
> You should add a unit test for your changes and DequantizeLinear should be aligned with this change. Since, the current behaviour is to force zero point to be 0, I doubt an existing unit test breaks. onnx specifications do not specify anything particular (+ @gramalingam). It looks ok to move forward.

I am OoO last week and sorry for the slow response. I think there is no existing unit test cover this case and please take a look at the newly  added test, thanks.

wangw-1991:
The failed CI seems to be unrelated to this PR.

fdwr:
@xadupre Any comments? Otherwise I'll merge. ⌛

> You should add a unit test for your changes and DequantizeLinear should be aligned with this change. Since, the current behaviour is to force zero point to be 0, I doubt an existing unit test breaks. onnx specifications do not specify anything particular (+ @gramalingam). It looks ok to move forward.

wangw-1991:
> @xadupre Any comments? Otherwise I'll merge. ⌛
> 
> > You should add a unit test for your changes and DequantizeLinear should be aligned with this change. Since, the current behaviour is to force zero point to be 0, I doubt an existing unit test breaks. onnx specifications do not specify anything particular (+ @gramalingam). It looks ok to move forward.

Hi, any update? Can we merge this now.

fdwr:
> Hi, any update? Can we merge this now.

**Wei**: Although @xadupre didn't explicitly approve, he did say "It looks ok to move forward", and you added a unit test with all tests passing. So I will merge it. Thanks for updating.

PR #25652, opened 2025-08-05
Properly remove in-memory references

yuslepukhin:
### Description

Make Node::ToProto() const  
call Graph::ToGraphProto() const so it does process all of the
  subgraphs recursively and removes all the in-memory references.

### Motivation and Context

Node::ToProto() const does not respect constness of subgraph attributes
and calls non-const version of Graph::ToGraphProto() which does not process
subgraph initializers and does not remove in-memory references.

### DIFF
--- a/include/onnxruntime/core/graph/graph.h
+++ b/include/onnxruntime/core/graph/graph.h
@@ -1220,7 +1220,10 @@ class Graph {  // NOLINT(clang-analyzer-optin.performance.Padding): preserve exi
 #endif
 
 #if !defined(ORT_MINIMAL_BUILD)
-  /** Gets the GraphProto representation of this Graph only. */
+  /** Gets the GraphProto representation of this Graph only.
+   * Deprecated. Do not use in main code. This does not recurse the subgraphs,
+   * does not populate the initializers of the root graph.
+   */
   const ONNX_NAMESPACE::GraphProto& ToGraphProto();
 
   /// <summary>
@@ -1606,9 +1609,12 @@ class Graph {  // NOLINT(clang-analyzer-optin.performance.Padding): preserve exi
   ///   as needed.
   /// </summary>
   /// <param name="output_graph_proto">The GraphProto to process</param>
-  /// <param name="process_main">Whether to process the main graph initializers</param>
+  /// <param name="main_graph">The main graph that ToGraphProto() was invoked for</param>
+  /// <param name="process_main_graph">Whether to process the initializers for the graph
+  /// that ToGraphProto() was invoked for</param>
   /// <returns>Status indicating success or failure</returns>  ///
-  Status ProcessSubgraphsInMemoryData(ONNX_NAMESPACE::GraphProto& output_graph_proto, bool process_main) const;
+  Status ProcessSubgraphsInMemoryData(ONNX_NAMESPACE::GraphProto& output_graph_proto,
+                                      const Graph& main_graph, bool process_main_graph) const;
 
   /// <summary>
   /// This function traverses the graph bottom up and externalizes
--- a/onnxruntime/core/graph/graph.cc
+++ b/onnxruntime/core/graph/graph.cc
@@ -666,12 +666,17 @@ void Node::ToProto(NodeProto& proto, bool update_subgraphs) const {
 
   // Set attributes.
   proto.clear_attribute();
-  for (const auto& attribute : attributes_) {
+  for (const auto& [name, attribute] : attributes_) {
     const gsl::not_null<AttributeProto*> attr{proto.add_attribute()};
-    *attr = attribute.second;  // copy
-    if (update_subgraphs && attr->has_g()) {
+    *attr = attribute;  // copy
+    if (update_subgraphs && utils::HasGraph(*attr)) {
+      auto find_hit = attr_to_subgraph_map_.find(name);
+      ORT_ENFORCE(find_hit != attr_to_subgraph_map_.cend(),
+                  "Node [", name_, "] op_type [", op_type_, "] does not have the subgraph for ", name);
+      // Force ToGraphProto() const to be called
+      const Graph& subgraph = *find_hit->second;
       attr->clear_g();
-      *attr->mutable_g() = attr_to_subgraph_map_.find(attribute.first)->second->ToGraphProto();
+      *attr->mutable_g() = subgraph.ToGraphProto();
     }
   }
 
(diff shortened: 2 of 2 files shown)


include/onnxruntime/core/graph/graph.h
```diff
@@ -1220,7 +1220,10 @@ class Graph {  // NOLINT(clang-analyzer-optin.performance.Padding): preserve exi
 #endif
 
 #if !defined(ORT_MINIMAL_BUILD)
-  /** Gets the GraphProto representation of this Graph only. */
+  /** Gets the GraphProto representation of this Graph only.
+   * Deprecated. Do not use in main code. This does not recurse the subgraphs,
+   * does not populate the initializers of the root graph.
```
skottmckay: I don't think this statement is correct. The implementation of this calls ToGraphProtoInternal which iterates all nodes. When calling Node::ToProto that will descend into the subgraphs if an update is required to create the GraphProto for the subgraph attribute in the node, and that attribute contains any nested subgraphs. 

Prior to converting TensorProto instances to use the in-memory marker this returned a valid GraphProto that could be saved and used any time. In the current implementation if there are in-memory TensorProto instances created the returned GraphProto is only valid for the lifetime of the InferenceSession as there's no logic to change in-memory TensorProtos back to being inlined. Isn't that a breaking change so we can't just say it's deprecated?

onnxruntime/core/graph/graph.cc
```diff
@@ -666,12 +666,17 @@ void Node::ToProto(NodeProto& proto, bool update_subgraphs) const {
...
+  for (const auto& [name, attribute] : attributes_) {
     const gsl::not_null<AttributeProto*> attr{proto.add_attribute()};
-    *attr = attribute.second;  // copy
-    if (update_subgraphs && attr->has_g()) {
+    *attr = attribute;  // copy
+    if (update_subgraphs && utils::HasGraph(*attr)) {
+      auto find_hit = attr_to_subgraph_map_.find(name);
+      ORT_ENFORCE(find_hit != attr_to_subgraph_map_.cend(),
+                  "Node [", name_, "] op_type [", op_type_, "] does not have the subgraph for ", name);
+      // Force ToGraphProto() const to be called
```
skottmckay:
```suggestion
      // Force ToGraphProto() const to be called so that any in-memory TensorProto initializers go back to being inlined
``` #Closed

File: onnxruntime/core/graph/graph.cc [hunk 3]
```diff
@@ -666,12 +666,17 @@ void Node::ToProto(NodeProto& proto, bool update_subgraphs) const {
...
-  for (const auto& attribute : attributes_) {
+  for (const auto& [name, attribute] : attributes_) {
     const gsl::not_null<AttributeProto*> attr{proto.add_attribute()};
-    *attr = attribute.second;  // copy
-    if (update_subgraphs && attr->has_g()) {
+    *attr = attribute;  // copy
+    if (update_subgraphs && utils::HasGraph(*attr)) {
+      auto find_hit = attr_to_subgraph_map_.find(name);
+      ORT_ENFORCE(find_hit != attr_to_subgraph_map_.cend(),
+                  "Node [", name_, "] op_type [", op_type_, "] does not have the subgraph for ", name);
```
skottmckay: nit: ORT_ENFORCE adds binary size. what use case makes this check required? I would have expected validation when the Node is created to be sufficient.  #Closed

onnxruntime/core/graph/graph.cc
```diff
@@ -4312,7 +4317,7 @@ Status InlineOrCopyInitializer(const Graph& src_graph, const ONNX_NAMESPACE::Ten
 }  // namespace
 
 Status Graph::ProcessSubgraphsInMemoryData(ONNX_NAMESPACE::GraphProto& output_graph_proto,
-                                           bool process_main) const {
+                                           const Graph& main_graph, bool process_main_graph) const {
```
skottmckay: When would parent_graph_ be null in a subgraph? There are a number of places like Graph::Resolve where that is the check for it being the main graph, so I'm wondering why that isn't sufficient here. #Closed

yuslepukhin: `ToGraphProto()` is invoked for subgraphs usually. So `parent_graph_` is not nullptr most of the time.
The issue here is that initializers are not copied by `ToGraphProtoInternal()` for the graph it is being called for, so we keep track for that and call it `main_graph` for this particular invocation.

skottmckay (dismissed):
:shipit:

gedoensmax:
Ok thank you, that does the trick for NV TRT EP.

onnxruntime/core/session/inference_session.cc
```diff
@@ -1421,6 +1421,8 @@ common::Status InferenceSession::TransformGraph(onnxruntime::Graph& graph, bool
     }
   }
 
+  ORT_RETURN_IF_ERROR_SESSIONID_(graph.ConvertInitializersIntoOrtValues());
```
skottmckay, line 1445: nit: Comment explaining what we're doing and why we do it here would be very helpful to someone unfamiliar with all the implementation details.

skottmckay (approved):

gaugarg-nv:
@skottmckay Can you please tag this PR for cherry-picking into the 1.23.0 release? This is a blocker to run Phi models with TRT-RTX EP.

adrianlizarraga (approved):

onnxruntime/core/graph/graph.cc
```diff
@@ -4343,103 +4388,48 @@ Status Graph::ProcessSubgraphsInMemoryData(ONNX_NAMESPACE::GraphProto& output_gr
...
-                                  mutable_initializers->end());
-    }
-#else
-    for (auto& initializer : *output_graph_proto.mutable_initializer()) {
-      if (utils::HasExternalDataInMemory(initializer)) {
-        // If the initializer has external data in memory, we need to inline it.
-        ORT_RETURN_IF_ERROR(InlineOrCopyInitializer(*this, initializer, initializer));
-      }
+  // Filter in iterators for weights that are present in the name_to_initial_tensor_ map
+  // and preserve the order. This is needed for tests.
```
skottmckay, line 4397: Can we not change the tests?

yuslepukhin: It is not tests code, it is the model serialization dumb text comparison. I have done this once and it depends on the hash table iteration order.

PR #25656, opened 2025-08-05
[NV TRT RTX EP] Cumulative TRT RTX EP merge

gedoensmax:
This currently holds 2 major improvements:
- dynamic shape models should have much lower memory usage and in addition to that the management is move towards ORT allocators
- the overhead for shape binding and address updates is reduce per inference

### DIFF
--- a/onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
+++ b/onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
@@ -353,193 +353,6 @@ bool ApplyProfileShapesFromProviderOptions(std::vector<nvinfer1::IOptimizationPr
   return true;
 }
 
-/*
- * Apply TensorRT optimization profile shapes from input tensor value.
- *
- * This function supports single/multiple profile(s).
- * (Note: An optimization profile describes a range of dimensions for each network input)
- *
- * @param shape_tensor_values holds "shape tensor -> shape values" for the INT32 shape tensor input across this inference run
- * @param shape_tensor_values_int64 holds "shape tensor -> shape values" for the INT64 shape tensor input across this inference run
- */
-Status ApplyProfileShapesFromInputTensorValue(std::vector<nvinfer1::IOptimizationProfile*>& trt_profiles,
-                                              Ort::KernelContext ctx,
-                                              nvinfer1::ITensor* input,
-                                              ShapeRangesMap& shape_ranges,
-                                              const std::unordered_map<std::string, size_t>& input_indexes,
-                                              std::unordered_map<std::string, std::vector<int32_t>>& shape_tensor_values,
-                                              std::unordered_map<std::string, std::vector<int64_t>>& shape_tensor_values_int64,
-                                              cudaStream_t stream,
-                                              bool* engine_update) {
-  for (size_t i = 0; i < trt_profiles.size(); i++) {
-    const std::string& input_name = input->getName();
-    nvinfer1::Dims dims = input->getDimensions();
-    int nb_dims = dims.nbDims;
-
-    size_t input_index = 0;
-    const auto& iter = input_indexes.find(input_name);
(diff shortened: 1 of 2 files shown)


skottmckay (approved):
:shipit:

onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.h
```diff
@@ -110,6 +114,45 @@ class OutputAllocator : public nvinfer1::IOutputAllocator {
...
+      return true;
+
+    for (int i = 0; i < dims.nbDims; ++i) {
+      if (dims.d[i] != other.dims.d[i])
+        return true;
+    }
+    return false;
+  }
+};
+
```
gedoensmax, line 155: @gaurag reorganized the input and output tensor updates on this PR to reduce CPU overhead per inference. This does not require reassigning tensor addresses and shapes each inference which also make managing CUDA graphs in a later PR much easier.

onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
```diff
@@ -2806,44 +2610,51 @@ Status NvExecutionProvider::CreateNodeComputeInfoFromGraph(const GraphViewer& gr
...
+        mem_size = trt_context->updateDeviceMemorySizeForShapes();
+      }
+      if (trt_state->context_memory_size != mem_size) {
+        LOGS_DEFAULT(INFO) << "[NvTensorRTRTX EP] A new context memory was allocated with size " << mem_size;
+        trt_state->context_memory = IAllocator::MakeUniquePtrFromOrtAllocator<void>(alloc, mem_size, false /*use_reserve*/);
+        trt_state->context_memory_size = mem_size;
+        trt_context->setDeviceMemoryV2(trt_state->context_memory.get(), mem_size);
       }
-      trt_context->setDeviceMemory(IAllocator::MakeUniquePtrFromOrtAllocator<void>(alloc, *max_context_mem_size_ptr).get());
     }
```
gedoensmax, line 2657: I changed this to always use the ORT arena allocator for intermediate memory. This support dynamic profiles like LLM's with much less memory than the maximal required storage. 
Plan for the future is to allocated and deallocate this on a stream to be able to reuse memory on for multiple sessions. 

For any recommendations how this could be best done for CUDA graphs we are very open ! Basically it would require a single "max" allocation across sessions and I am tending to say this will just happen after a few inferences of each model. An open question is how CUDA graph and arena shrinkage should work together for e.g. a scenario like diffusion where we have  1x encode -> 30x backbone -> 1x decoder since CUDA graphs might be very desirable for the backbone.

onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
```diff
@@ -3094,20 +2891,25 @@ Status NvExecutionProvider::CreateNodeComputeInfoFromPrecompiledEngine(const Gra
...
-    for (size_t i = 0, end = input_binding_names.size(); i < end; ++i) {
-      char const* input_name = input_binding_names[i];
-
-      size_t input_index = 0;
-      const auto iter = input_indexes.find(input_name);
-      if (iter != input_indexes.end()) {
-        input_index = iter->second;
-      }
+    auto& scratch_buffers = trt_state->scratch_buffers;
+    if (require_io_binding) {
```
gedoensmax, line 2895: We will hold on to IO buffers if possible in the TRT state here.

chilo-ms (approved):

onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.h
gaugarg-nv: Max, my handle is @gaugarg-nv  :)

PR #25567, opened 2025-07-28
Update OrtEpFactory in MiGraphX EP 

psakhamoori:
### Description
Update OrtEpFactory in new EPs to add allocator, data transfer and stream stubs.

PR #25516, opened 2025-07-23
[MIGraphX EP] Fix compilation after cherry-picking from win-onnxruntime

apwojcik:
After cherry-picking from win-onnxruntime (#25481), the MIGraphX EP stopped compiling on the main branch.

PR #25659, opened 2025-08-05
Rel 123 with migraphx compilation fix's

psakhamoori:
Cherry-pick MiGraphX EP fixes from upstream for rel-1.23.0

This PR cherry-picks three critical fixes for the MiGraphX Execution Provider:

1. Fix compilation after cherry-picking from win-onnxruntime (#25516)
   - Adds ORT_UNUSED_PARAMETER(num_devices) to fix unused parameter warning
   - Corrects struct usage in CreateIExecutionProvider method
   
2. Fix CreateExecutionProviderFactory with correct struct and change vendor_id (#25625)  
   - Updates vendor_id from 0x1002 to 0x9999 to allow DML EP to be default
   - Ensures proper device ordering in provider_policy_context.cc

3. Update OrtEpFactory in MiGraphX EP (#25567)
   - Adds complete OrtEpFactory infrastructure for auto EP selection
   - Implements all required factory methods with noexcept specifiers
   - Sets ort_version_supported to ORT_API_VERSION
   - Enables MiGraphX/AMDGPU EP integration with hardware device detection

These fixes ensure MiGraphX EP builds correctly and integrates properly with 
the ORT execution provider selection framework in the 1.23.0 release.

Cherry-picked commits:
- 87f149971f6b47cc82e84e08685193f31bb0cf8e
- 14ca6df1f21cb9cef80879e6f786016273cae6e1  
- 131cf40dc25a1611e239b782eb5239cc4921f34e

### DIFF
--- a/onnxruntime/core/providers/migraphx/migraphx_provider_factory.cc
+++ b/onnxruntime/core/providers/migraphx/migraphx_provider_factory.cc
@@ -155,7 +155,10 @@ struct MIGraphX_Provider : Provider {
     const ConfigOptions* config_options = &session_options.GetConfigOptions();
 
     std::array<const void*, 2> configs_array = {&provider_options, config_options};
-    auto ep_factory = CreateExecutionProviderFactory(&provider_options);
+    OrtMIGraphXProviderOptions migraphx_options;
+    UpdateProviderOptions(&migraphx_options, provider_options);
+
+    auto ep_factory = CreateExecutionProviderFactory(&migraphx_options);
     ep = ep_factory->CreateProvider(session_options, logger);
 
     return Status::OK();
@@ -250,7 +253,8 @@ struct MigraphXEpFactory : OrtEpFactory {
   const std::string ep_name;
   const std::string vendor{"AMD"};
 
-  const uint32_t vendor_id{0x1002};
+  // Not using AMD vendor id 0x1002 so that OrderDevices in provider_policy_context.cc will default dml ep
+  const uint32_t vendor_id{0x9999};
   const OrtHardwareDeviceType ort_hw_device_type;  // Supported OrtHardwareDevice
 };
 
--- a/onnxruntime/core/providers/migraphx/migraphx_provider_factory.cc
+++ b/onnxruntime/core/providers/migraphx/migraphx_provider_factory.cc
@@ -185,11 +185,22 @@ struct MigraphXEpFactory : OrtEpFactory {
                     OrtHardwareDeviceType hw_type,
                     const OrtLogger& default_logger_in)
       : ort_api{ort_api_in}, default_logger{default_logger_in}, ep_name{ep_name}, ort_hw_device_type{hw_type} {
+    ort_version_supported = ORT_API_VERSION;  // set to the ORT version we were compiled with
     GetName = GetNameImpl;
     GetVendor = GetVendorImpl;
+    GetVendorId = GetVendorIdImpl;
+    GetVersion = GetVersionImpl;
+
     GetSupportedDevices = GetSupportedDevicesImpl;
     CreateEp = CreateEpImpl;
     ReleaseEp = ReleaseEpImpl;
+
+    CreateAllocator = CreateAllocatorImpl;
+    ReleaseAllocator = ReleaseAllocatorImpl;
+    CreateDataTransfer = CreateDataTransferImpl;
+
+    IsStreamAware = IsStreamAwareImpl;
+    CreateSyncStreamForDevice = CreateSyncStreamForDeviceImpl;
   }
 
   // Returns the name for the EP. Each unique factory configuration must have a unique name.
@@ -202,6 +213,16 @@ struct MigraphXEpFactory : OrtEpFactory {
   static const char* GetVendorImpl(const OrtEpFactory* this_ptr) noexcept {
     const auto* factory = static_cast<const MigraphXEpFactory*>(this_ptr);
     return factory->vendor.c_str();
+  }
+
+  static uint32_t GetVendorIdImpl(const OrtEpFactory* this_ptr) noexcept {
+    const auto* factory = static_cast<const MigraphXEpFactory*>(this_ptr);
+    return factory->vendor_id;
+  }
+
+  static const char* GetVersionImpl(const OrtEpFactory* this_ptr) noexcept {
+    const auto* factory = static_cast<const MigraphXEpFactory*>(this_ptr);
+    return factory->version.c_str();
   }
 
   // Creates and returns OrtEpDevice instances for all OrtHardwareDevices that this factory supports.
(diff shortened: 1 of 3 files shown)


psakhamoori:
@nieubank Please help review this commit 

This is to address MiGraphX Compilation errors with below commits cherry-picked

Cherry-picked commits:
- 87f149971f6b47cc82e84e08685193f31bb0cf8e
- 14ca6df1f21cb9cef80879e6f786016273cae6e1  
- 131cf40dc25a1611e239b782eb5239cc4921f34e

psakhamoori:
@adrianlizarraga @skottmckay 

Please help approve this PR

onnxruntime/core/providers/migraphx/migraphx_provider_factory.cc
```diff
@@ -212,7 +236,7 @@ struct MigraphXEpFactory : OrtEpFactory {
                                             size_t num_devices,
                                             OrtEpDevice** ep_devices,
                                             size_t max_ep_devices,
-                                            size_t* p_num_ep_devices) {
+                                            size_t* p_num_ep_devices) noexcept {
```
skottmckay, line 239: What's the reason for the vendor id check in the implementation being commented out? That would result in an OrtEpDevice being created for non-AMD hardware if present which would be incorrect I assume. e.g. user has an nvidia gpu. You should only ever create an OrtEpDevice for a device that the EP can use. 

Also is any check required on the minimum capability of the user's device? #Closed

onnxruntime/core/providers/migraphx/migraphx_provider_factory.cc
```diff
@@ -237,20 +261,59 @@ struct MigraphXEpFactory : OrtEpFactory {
...
+
   const OrtApi& ort_api;
   const OrtLogger& default_logger;
   const std::string ep_name;
   const std::string vendor{"AMD"};
+  const std::string version{"1.0.0"};  // MigraphX EP version
 
-  const uint32_t vendor_id{0x1002};
+  // Not using AMD vendor id 0x1002 so that OrderDevices in provider_policy_context.cc will default dml ep
+  const uint32_t vendor_id{0x9999};
```
skottmckay, line 316: Why is defaulting to the DML EP preferred?

Using a random vendor id feels like the wrong way to address a selection issue, and relies on implementation details of OrderDevices that could change (it's arbitrarily sorting on EP name as the tie breaker currently).

The user can provide a custom policy delegate if they want this level of control over selection.

onnxruntime/test/providers/migraphx/migraphx_basic_test.cc
```diff
@@ -188,6 +188,24 @@ TEST(MIGraphXExecutionProviderTest, canEvalArgument) {
   ASSERT_EQ(canEvalNodeArgument(gv, node2, {1}, input_nodes), true);
 }
 
+static bool SessionHasEp(Ort::Session& session, const char* ep_name) {
```
skottmckay, line 191: Nothing calls this in the unit test. Is that intentional?

psakhamoori: @Zhaeong to comment

onnxruntime/core/providers/migraphx/migraphx_provider_factory.cc
psakhamoori: @Zhaeong to comment

onnxruntime/core/providers/migraphx/migraphx_provider_factory.cc
psakhamoori: @Zhaeong to comment

onnxruntime/core/providers/migraphx/migraphx_provider_factory.cc
Zhaeong: This is a workaround currently based on the source given to us @karim-vad for when --ep_policy is set to GPU. We want the behavior to default using dml, and sideload using MIGraphX. If there's another way of doing that when  --ep_policy is set to GPU we can change that.

onnxruntime/test/providers/migraphx/migraphx_basic_test.cc
Zhaeong: @apwojcik to comment. This was added to prevent build failure.

onnxruntime/core/providers/migraphx/migraphx_provider_factory.cc
Zhaeong: This was commented out before GetVendorIdImpl was added I believe. @karim-vad  to comment. We can uncomment if there's another way to default to dml when ep_selection_policy is set to PREFER_GPU (--ep_policy GPU)

onnxruntime/test/providers/migraphx/migraphx_basic_test.cc
apwojcik: It is called by the test below.
https://github.com/microsoft/onnxruntime/pull/25659/files#diff-20dc7bfb58201736ee9f7d15f47555015eef91cfda22949020c710a5ca778af4L210

The code is identical to the NV Tensor RTX test for the auto EP feature.

https://github.com/microsoft/onnxruntime/blob/b1546da6f524f4995ed3273648ebf1152f9f727a/onnxruntime/test/providers/nv_tensorrt_rtx/nv_basic_test.cc#L397

onnxruntime/core/providers/migraphx/migraphx_provider_factory.cc
skottmckay: Those seem like separate concerns.

Creating the OrtEpDevice means that you're saying migraphx can execute the model using that device. If this was for an NVIDIA GPU that is not the case is it?

The EP having a different vendor id is used in the autoep selection policy. That matches against the OrtHardwareDevice vendor id to prefer using the EP from the vendor that manufactured the device.

onnxruntime/core/providers/migraphx/migraphx_provider_factory.cc
skottmckay: I'm confused as to what behavior you're after. When would the migraphx library be used? 

If you want DML to be used why register the migraphx library?

Zhaeong: It would be used when sideloaded in session via AppendExecutionProvider_V2 or AppendExecutionProvider_MIGraphX

Zhaeong: The behavior we want in this release is for DML as the default EP when policy is set to OrtExecutionProviderDevicePolicy_PREFER_GPU in a session with both EPs registered.
The function:
PreferGpuEpPolicy::SelectProvidersForDevices
Selects the first GPU EP post OrderDevices call in SelectEpsForSession function.
It matches on vendor id, which will pick MIGraphX EP over DML if both are present. We set dummy vendor_id{0x9999}; so it won't match and DML EP will be selected.
If there's a better way to do this, please advise and we can change it.

onnxruntime/core/providers/migraphx/migraphx_provider_factory.cc
Zhaeong: Because of the workaround to set vendor_id{0x9999}, it will return no devices for MIGraphX during RegisterExecutionProviderLibrary if it checks for:
factory->ort_api.HardwareDevice_VendorId(&device) == factory->vendor_id

skottmckay: You can check that the device has the AMD vendor id. You're not forced to use the current (fake) value of factory->vendor_id.

onnxruntime/core/providers/migraphx/migraphx_provider_factory.cc
skottmckay: To clarify, AppendExecutionProvider_V2 and AppendExecutionProvider_MIGraphX are explicitly selecting the EP (vs. auto selection via policy or policy delegate) rather than 'sideloading'.

There is no notion that we wouldn't want to strongly prefer the EP from the hardware vendor when doing auto selection. The expected way if you need a custom selection policy would be to provide the policy delegate and override the default implementation. 

If that isn't feasible this hack 'works' but it's not going to be reliable going forward. Would need to be validated for each release given the default policy implementation may change and the reason it's selecting DML is based on the alphabetical order of the EP names.

Zhaeong: This would be a temporary until the MIGraphX EP is in a more stable state. At which point we would want to default to using MIGraphX. This is so that users not using custom selection for now will use DML as default for this release.

onnxruntime/core/providers/migraphx/migraphx_provider_factory.cc
Zhaeong: @psakhamoori can you add to this PR:
Change:
      if (factory->ort_api.HardwareDevice_Type(&device) == factory->ort_hw_device_type) {
        // factory->ort_api.HardwareDevice_VendorId(&device) == factory->vendor_id) {
To:
      if (factory->ort_api.HardwareDevice_Type(&device) == factory->ort_hw_device_type &&
          factory->ort_api.HardwareDevice_VendorId(&device) == 0x1002) {

Zhaeong: @psakhamoori, or add this commit from here:
https://github.com/microsoft/onnxruntime/pull/25677

skottmckay (approved):

PR #25663, opened 2025-08-06
fixed matmul broadcasting

konstantin-pueckler-qc:
### Description
currently, the broadcasting of the ``MatMul`` operator does not correctly handle zero-sized batch dimensions. By the numpy standard, when one matrix has length 0 in one batch dimension and the other matrix has length 1 in that dimension, it should be broadcasted to length 0. Instead, onnxruntime currently raises an error in that case.
I modified the computation of ``output_dims`` to fix that error.

### Motivation and Context
We came across this problem when testing the onnxruntime via [onnx-tests](https://github.com/cbourjau/onnx-tests).


konstantin-pueckler-qc:
@microsoft-github-policy-service agree company="QuantCo"

konstantin-pueckler-qc:
@xadupre  might you be in a position to review this?

Issue #25504, opened 2025-07-23
[Bug] Auto EP selection rejects the combination of DML EP with other EPs like OpenVINO EP


mingmingtasd:
On my device with the Intel CPU, and both NVIDIA and Intel GPUs,  if the DML EP and OpenVINO EP are both registered, [PreferGpuEpPolicy::SelectProvidersForDevices](https://github.com/microsoft/onnxruntime/blob/3b97d79b3c12dbf93aa0d563f345714596dc8ab6/onnxruntime/core/session/provider_policy_context.cc#L430) will select three EP devices: DML GPU for NVIDIA, OpenVINO CPU for Intel and default ORT CPU during the session initialization. But the [InferenceSession::HasInvalidCombinationOfExecutionProviders](https://github.com/microsoft/onnxruntime/blob/3b97d79b3c12dbf93aa0d563f345714596dc8ab6/onnxruntime/core/session/inference_session.cc#L1986) method rejects the combination of DML EP with OpenVINO EP, and reports error: `"DML EP can be used with only CPU EP."` which causes session initialization to fail.

1. Is it true that DML EP can only be used with the default CPU EP?
2. If so, why not reject the combination in [PreferGpuEpPolicy::SelectProvidersForDevices](https://github.com/microsoft/onnxruntime/blob/3b97d79b3c12dbf93aa0d563f345714596dc8ab6/onnxruntime/core/session/provider_policy_context.cc#L430) earlier? It doesn't make sense an invalid combination can be created and then be rejected with an error.

mingmingtasd:
@huningxin @fdwr @skottmckay Any ideas? Thanks!

fdwr:
I vaguely recall this PR https://github.com/microsoft/onnxruntime/pull/22657 that Pranav added, but I forget the context of why DML was considered mutually exclusive with other EP's (one would think ORT could shuffle memory between them via system RAM). @pranavsharma Do you recall? (*edit: oops, forgot you're not at MS anymore*)

I know a few years back (before the above PR), I was able to locally build onnxruntime.dll with both the DML EP and CUDA EP (did not try OpenVINO one) and switch between them at runtime (but not run _both_ at the same time in the same session).

skottmckay:
I'm guessing based on the linked PR that the update was involving a build with DML and CUDA enabled in it. 

[...]

PR #25664, opened 2025-08-06
Allow DML EP to be used with any CPU EP

skottmckay:
### Description

Relax restriction on DML EP so other CPU based EPs can be used. 

### Motivation and Context

#25504

### DIFF
--- a/onnxruntime/core/session/inference_session.cc
+++ b/onnxruntime/core/session/inference_session.cc
@@ -1984,13 +1984,15 @@ static void ResolveMemoryPatternFlags(SessionState& session_state) {
 // For now, this function only checks for invalid combination of DML EP with other EPs.
 // TODO: extend this function to check for other invalid combinations of EPs.
 common::Status InferenceSession::HasInvalidCombinationOfExecutionProviders() const {
-  // DML EP is only allowed with CPU EP
+  // DML EP is not allowed with other GPU or NPU EPs.
+  // historical reason for this is unknown. relaxing the limit that it must only be used with the CPU EP to support
+  // scenarios where alternative EPs are CPU based (e.g. openvino).
   bool has_dml_ep = execution_providers_.Get(kDmlExecutionProvider) != nullptr;
   if (has_dml_ep) {
-    const auto& ep_list = execution_providers_.GetIds();
-    for (const auto& ep : ep_list) {
-      if (ep == kDmlExecutionProvider || ep == kCpuExecutionProvider) continue;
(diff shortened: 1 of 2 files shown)


onnxruntime/core/session/inference_session.cc
```diff
@@ -1984,13 +1984,15 @@ static void ResolveMemoryPatternFlags(SessionState& session_state) {
 // For now, this function only checks for invalid combination of DML EP with other EPs.
 // TODO: extend this function to check for other invalid combinations of EPs.
 common::Status InferenceSession::HasInvalidCombinationOfExecutionProviders() const {
-  // DML EP is only allowed with CPU EP
+  // DML EP is not allowed with other GPU or NPU EPs.
+  // historical reason for this is unknown. relaxing the limit that it must only be used with the CPU EP to support
```
fdwr, line 1988:
```suggestion
  // The historical reason for this is unknown. Relaxing the limit that it must only be used with the CPU EP to support
```
Yeah, I forget. Pranav might recall, but alas he left MS. I think part of the concern was that the combo had never been tested before (since the ORT with WinML only ever used DML), and that ORT might not do GPU-to-GPU copies between partitions nicely. I wonder how robustly ORT works with *other* heteregenous combinations, like say a graph containing nodes partitioned to OpenVINO and CUDA (or ROCM or TensorRT)?

fdwr (approved):
👍 Thanks Scott.

mingmingtasd:
Thanks a lot! @skottmckay

edgchen1 (approved):

onnxruntime/core/session/inference_session.cc
skottmckay: The memcpy transformer handles copying between a non-CPU OrtDevice and CPU. In general that works perfectly fine as there's usually (at least previously) only the CPU EP and an EP for a non-CPU device.

Compiling EPs tend to register with CPU based memory (or host accessible) so there's no device copy required by ORT there. 

In theory you could do something like enable an EP with static kernels for a discrete GPU (let's say CUDA EP for an nvidia card) and a different EP with static kernels for an integrated GPU (let's say DML). if you did that, and the model got partitioned and data needed to go from the dGPU to the iGPU for two adjacent nodes it wouldn't work as there's no special-casing to determine it needs to do that via CPU memory. 

but given that doesn't feel like a realistic scenario, adding that special-casing feels like it's a lot of cost for possibly zero benefit.

Issue #25494, opened 2025-07-22
Get OpSchemas for registered Custom Operators


JorgeMIng:
### Describe the feature request

The function "get_all_operator_schema" (onnxruntime.capi.onnxruntime_pybind11_state.get_all_operator_schema)
is not available for custom Operators. Session Option or Session inference could be enhanced by giving the option of getting the information of the available operators for the session.

Extending the functionality of these functions helps when creating systems like NVIDIA NSight DL Designer or ONNX model node editors, where only the predefined operators are available.

I have checked the code to find if there was any method to get this metadata from registered custom operators. If this functionality already exists, I would be happy to know.

Best Regards

### Describe scenario use case

My main use case now is for a project to convert ONNX models to RDF "https://github.com/JorgeMIng/ONNX2RDF" where I was using "get_all_operator_schema" but was not able to extend it to custom operators.

It could be used to create an ONNX Operator registry compiling multiple ONNX custom libraries for easily finding operators for specific tasks. Moreover, if considering systems like Nvidia NSight DL Designer, where only the predefined ONNX operators are available, this could introduce a method to add compatibility for custom libraries.

PR #25665, opened 2025-08-06
Python binding for listing custom operators. 

JorgeMIng:
Lists the OpSchemas of custom operators introduce when installing custom libraries on a Session.
This is done in response to the issue, see issue https://github.com/microsoft/onnxruntime/issues/25494. Where i asked for the possibility of listing the OpSchemas of custom operators.

### Description
I used the already existing custom registry for custom ops and add a new function for SessionOptions, given a session options it is possible to list the OpSchemas installed by custom libraries. The implementation is quite easy, when the function is created a CustomRegistry is created and the function CreateCustomRegistry is used with the custom_ops_domains variable of SessionOptions. Then the GetOpschemaRegistry is taken a listed all its Schemas. For listing the schemas, a new function was added to OnnxRuntimeOpSchemaRegistry listing the operators.

### Motivation and Context
The main reason this change is important is a lot of information about custom operators is not accessible when for predefined operators it is already an option with the functions "from onnxruntime.capi.onnxruntime_pybind11_state import get_all_operator_schema". The information given by OpSchemas helps with creating graphical user interfaces for viewing or editing ONNX models. For example, Nvidia Nsight DL has no compatibility with custom operators as the information about types, descriptions, input and outputs is not directly accessible by onnxruntime.

In my case, I am developing a program to extract metadata given Onnx models and expressing the information in RDF. I was using the function to list operators to expresses in the ontology a lot of information about the operators. It was not compatible with custom operators, this function helps solve that issue.

Best Regards

### DIFF
--- a/include/onnxruntime/core/graph/schema_registry.h
+++ b/include/onnxruntime/core/graph/schema_registry.h
@@ -92,6 +92,12 @@ class OnnxRuntimeOpSchemaRegistry : public IOnnxRuntimeOpSchemaCollection {
                            const ONNX_NAMESPACE::OpSchema** latest_schema,
                            int* earliest_opset_where_unchanged) const override;
 
+  std::vector<ONNX_NAMESPACE::OpSchema> get_all_schemas_with_history();
+
+
+
+
+
   bool empty() const {
     return map_.empty();
   }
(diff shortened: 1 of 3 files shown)


JorgeMIng:

JorgeMIng:
Any feedback???

JorgeMIng:
¿¿¿Being able to list the OpSchema of custom operators registered on SessionOptions is something that can already be done ??? I had tried to do it within the provided bindings and using the public interfaces but i have not found a working solution. The other option would be to do a case by case for each custom libary which is more expensive to maintain.

PR #25667, opened 2025-08-06
[QNN-EP] Add CastLoneQFusion to transform Cast and QNode into Convert

qti-hungjuiw:
### Description

- Introduced `CastLoneQFusion` in QNNEP to fuse `Cast` followed by `QuantizeLinear` into a single `Convert` operation.
- Added corresponding test cases for **UINT8-to-FLOAT** `Cast` combined with `QuantizeLinear`, covering various **QuantType** scenarios.

### Motivation and Context

- To optimize the model by reducing unnecessary QDQ nodes, this fusion transformation has been implemented.

### DIFF
--- a/onnxruntime/test/providers/qnn/qnn_basic_test.cc
+++ b/onnxruntime/test/providers/qnn/qnn_basic_test.cc
@@ -1042,10 +1042,11 @@ TEST_F(QnnHTPBackendTests, QnnContextPriorityHigh) {
 // cast_input -> Cast -> Q -> DQ ----
 //                                   |
 //             input2 -> Q -> DQ -> Add -> Q -> DQ -> output
-static GetTestModelFn BuildCastAddTestCase() {
-  return [](ModelTestBuilder& builder) {
+template <typename InputType, typename QuantType>
+static GetTestQDQModelFn<QuantType> BuildCastAddQDQTestCase() {
+  return [](ModelTestBuilder& builder, std::vector<QuantParams<QuantType>>& output_qparams) {
     // Creat Cast node int32 -> float32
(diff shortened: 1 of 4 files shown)


onnxruntime/test/providers/qnn/qnn_basic_test.cc
```diff
@@ -1100,10 +1119,60 @@ TEST_F(QnnHTPBackendTests, CastAddHTPAccuracyTest) {
 #endif
```
HectorSVC, line 1119: should be replace by
provider_options["backend_type"] = "htp";
seems we didn't clean up all the  places like this.

onnxruntime/test/providers/qnn/qnn_basic_test.cc
```diff
@@ -1100,10 +1119,60 @@ TEST_F(QnnHTPBackendTests, CastAddHTPAccuracyTest) {
...
+                                BuildCastAddQDQTestCase<uint8_t, uint8_t>(),
+                                provider_options,
+                                21,
+                                ExpectedEPNodeAssignment::All);
+}
+
+TEST_F(QnnHTPBackendTests, CastAddQDQU16) {
+  ProviderOptions provider_options;
+#if defined(_WIN32)
+  provider_options["backend_path"] = "QnnHtp.dll";
```
HectorSVC: should be replaced by
provider_options["backend_type"] = "htp";
seems we didn't clean up all the places like this.

qti-hungjuiw: Done. Thanks for pointing this out : )

PR #25673, opened 2025-08-06
Relax WeightBiasQuantization constraint for larger QDQ node group

qti-yuduo:
### Description
Relax WeightBiasQuantization constraint for larger QDQ node group

### Motivation and Context
The transformer `WeightBiasQuantization` quantizes float weights on  `Q -> DQ -> Conv/ConvTranspose/Gemm's Weights -> Q-> DQ` sequence; The check on `Weights -> Q` (`children_nodes.size() != 1 || children_nodes[0]->OpType() != QDQ::QOpName`) is an issue due to it would skip quantization for many common patterns such as unfused activations followed by `Conv` (`DQ - Conv -> ReLU -> Q`).

It's actually unnecessary to check ending Q here (the fold can happen anyway without changing model semantics). However, in order to minimize the current behavior change, this PR simply extend the pattern to include single path (no branch), type-preserving path lead to `Q` to enable more quantization support.

### DIFF
--- a/onnxruntime/core/optimizer/qdq_transformer/weight_bias_quantization.cc
+++ b/onnxruntime/core/optimizer/qdq_transformer/weight_bias_quantization.cc
@@ -13,6 +13,39 @@
 
 namespace onnxruntime {
 
+/**
+ * Checks whether or not the output path from a given node leads to a QuantizeLinear op, optionally, with no
+ * branching ReLU or Clip op in between. See also: NodeGroupSelector::GetQDQSelection() in qdq_selectors.cc.
+ *
+ * @param node The starting node to check the output path from.
+ * @param graph The graph containing the nodes.
+ *
+ * @return true if the path exist, false otherwise.
+ */
+static bool IsNoBranchPathToQuantizeLinear(const Node& node, const Graph& graph) {
+  const Node* current = &node;
+  while (true) {
+    // Conv / ConvTranspose / Gemm produces single output
+    if (current->OutputDefs().size() != 1) {
+      return false;
+    }
+    const std::vector<const Node*>& consumers = graph.GetConsumerNodes(current->OutputDefs()[0]->Name());
+    // Branching or no consumer: not eligible
+    if (consumers.size() != 1) {
+      return false;
+    }
+    const Node* consumer = consumers[0];
+    if (consumer->OpType() == QDQ::QOpName) {
+      return true;
+    }
+    // Allow ReLU or Clip, see also: NodeGroupSelector::GetQDQSelection() in qdq_selectors.cc.
+    if (consumer->OpType() != "Relu" && consumer->OpType() != "Clip") {
+      return false;
+    }
+    current = consumer;
+  }
+}
+
 Status WeightBiasQuantization::ApplyImpl(Graph& graph, bool& modified, int graph_level,
                                          const logging::Logger& logger) const {
   const GraphViewer graph_viewer{graph};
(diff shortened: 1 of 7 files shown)


onnxruntime/core/optimizer/qdq_transformer/weight_bias_quantization.cc
```diff
@@ -13,6 +13,76 @@
...
+/**
+ * Checks whether or not the output path from a given node leads to a `QuantizeLinear` node,
+ * with type and shape are preserved and no branching in between.
+ *
+ * @param node The starting node to check the output path from.
+ * @param graph The graph containing the nodes.
+ *
+ * @return true if the path exist, false otherwise.
+ */
+static bool IsTypePreserveNoBranchPathToQuantizeLinear(const Node& node, const Graph& graph) {
```
adrianlizarraga: Instead of making this generically handle any operator type between the Conv and the Q (e.g., `Conv -> Any* -> Q`), could we instead limit this explicitly to only `Clip` and `Relu` operator types?

The reason I originally added the check for the final Q is to ensure that this optimizer does not generate a non-valid QDQ node unit as defined by the rules here: https://github.com/microsoft/onnxruntime/blob/5d77b73c4ea102f5d4ea653e5c6461244f553def/onnxruntime/core/optimizer/qdq_transformer/selectors_actions/qdq_selectors.cc#L361. These rules only allow a Relu/Clip to appear between a Conv (or Gemm) and the output Q. Any other operator type would result in an invalid QDQ node unit.

Also, this was originally added to fix certain customer models. We should ensure that still works.

qti-yuduo: Thanks @adrianlizarraga for taking a look! 

> could we instead limit this explicitly to only Clip and Relu operator types?

Sure, I think it'll cover majority of the use cases. 

> Also, this was originally added to fix certain customer models. We should ensure that still works.

Is this covered by any existing tests? Do we see this would break the use case?

qti-yuduo: Updated.

onnxruntime/test/optimizer/qdq_transformer_test.cc
```diff
@@ -5444,8 +5444,59 @@ TEST(QDQTransformerTests, WeightBiasQuantization_Conv_Weight_Bias) {
...
+      conv_node.AddAttribute("strides", std::vector<int64_t>{1, 1});
+      conv_node.AddAttribute("group", static_cast<int64_t>(1));
+      conv_node.AddAttribute("pads", std::vector<int64_t>{0, 0, 0, 0});
+      builder.AddNode("Relu", {conv_fp32}, {relu_fp32});
+      builder.AddQuantizeLinearNode<uint8_t>(relu_fp32, 0.69f, static_cast<uint8_t>(127), relu_q, use_contrib_qdq);
+      builder.AddDequantizeLinearNode<uint8_t>(relu_q, 0.69f, static_cast<uint8_t>(127), relu_dq, use_contrib_qdq);
+    };
+
+    // Conv's weights should be quantized and folded, one additional Q/DQ pair inserted for weight
+    auto check_transformed_graph = [use_contrib_qdq](InferenceSessionWrapper& session) {
```
HectorSVC: >use_contrib_qdq [](http://example.com/codeflow?start=36&length=15)

Please fix the build error:
/Users/runner/work/onnxruntime/onnxruntime/onnxruntime/test/optimizer/qdq_transformer_test.cc:5474:37: error: lambda capture 'use_contrib_qdq' is not used [-Werror,-Wunused-lambda-capture]
 5474 |     auto check_transformed_graph = [use_contrib_qdq](InferenceSessionWrapper& session) {

qti-yuduo: Done.

snnn:

PR #25674, opened 2025-08-06
[webgpu] support bool for binary operators

fs-eire:
### Description

Currently boolean types are not supported as inputs of binary operators in WebGPU.

This change adds the support.

### Motivation and Context

In WebGPU, `bool` is not a valid type for storage. To make it work with a storage buffer, we have to store a `u32` value representing 4 bool values.

To make it work with the existing WebGPU framework, we need to ensure all modes of the binary operator program always use components == 4 for bool.


xhcao:
@fs-eire I had integrated and tested the cases on my side, the PR worked well, thanks.

guschmue (approved):

PR #25677, opened 2025-08-06
Add vendorid check to GetSupportedDevicesImpl for MIGraphx EP

Zhaeong:
This adds vendorid check for GetSupportedDevicesImpl during EP registration


snnn:
So, this change has been released out, but it is not in the main branch yet?

PR #25684, opened 2025-08-07
Skeleton for Attention(23) on CUDA

xadupre:
### Description
Draft for Attention(23) on CUDA.

Two directions for the implementations.

**contribops**

As it is, it does not seem to support SoftCap options and all the modes for the qk output. What about bfloat16?

**cublasLtMatMul**

Follows the same implementation the one made on CPU, relies on cublasLtMatMul, it should handle all types (bfloat16, float8). It misses cuda code for the cache copy and softcap.


tianleiwu:
MultiHeadAttention cuda implementation is close to onnx Attention definition.
Softcap is easy to add, and qk output need pass the buffers. You can follow GroupQueryAttention cuda kernel to add these.

titaiwangms:
In favor of #26466

PR #25689, opened 2025-08-07
Move provider tests to `onnxruntime_provider_test` and enable use of plugin EPs

edgchen1:
### Description

Move provider-specific unit tests that were formerly in `onnxruntime_test_all` to a new test program, `onnxruntime_provider_test`. Notably, this includes the op tests which test different provider implementations.

Enable some tests in `onnxruntime_provider_test` (those using `ModelTester` or `OpTester`) to use a plugin EP. The plugin EP usage is specified at runtime, so it is referred to as "dynamic". The dynamic plugin EP configuration can be specified with environment variable `ORT_UNIT_TEST_MAIN_DYNAMIC_PLUGIN_EP_CONFIG_JSON`.

This is an example value for `ORT_UNIT_TEST_MAIN_DYNAMIC_PLUGIN_EP_CONFIG_JSON`. The test infrastructure will register the plugin EP library at `path/to/example_plugin_ep.dll`. The op tests will use a plugin EP instance with the name `example_ep`.
```json
{
  "ep_library_registration_name": "example_ep",
  "ep_library_path": "path/to/example_plugin_ep.dll",
  "selected_ep_name": "example_ep"
}
```

### Motivation and Context

Enable use of plugin EPs in some provider-specific unit tests.

### Notes for Reviewers
- There are many modified files but most of the changes are just `#include` path updates.
- I suggest setting the diff view to ignore whitespace changes.


onnxruntime/test/util/test_dynamic_plugin_ep.cc
```diff
@@ -0,0 +1,185 @@
...
+  // just get the EP name from the first one.
+  std::string ep_name = Ort::ConstEpDevice{selected_c_ep_devices.front()}.EpName();
+
+  auto state = PluginEpInfrastructureState{};
+  state.config = std::move(config);
+  state.plugin_ep_library_registration_handle = std::move(ep_library_registration_handle);
+  state.ep_factory = std::move(ep_factory);
+  state.selected_c_ep_devices = std::move(selected_c_ep_devices);
+  StrMapToKeyValueCstrVectors(state.config.default_ep_options,
+                              state.default_ep_option_key_cstrs, state.default_ep_option_value_cstrs);
```
edgchen1: the map gets moved around. maybe it would be safer to construct the c string vectors later.

tools/ci_build/github/linux/test_custom_ops_pytorch_export.sh
```diff
@@ -13,4 +13,4 @@ for filename in /onnxruntime_src/onnxruntime/test/python/contrib_ops/onnx_test_*
   cd /build/Release && python3 -m pytest -v $filename || exit 1
 done
 
-cd /build/Release && ./onnxruntime_test_all --gtest_filter=ShapeInferenceTests.* || exit 1
+cd /build/Release && ./onnxruntime_provider_test --gtest_filter=ShapeInferenceTests.* || exit 1
```
edgchen1: this file is not referenced in any other repo code. can it be removed?

tools/ci_build/github/pai/pai_test_launcher.sh
```diff
@@ -13,3 +13,6 @@ echo ""
 
 echo "Running ./onnxruntime_test_all .."
 $build_dir/onnxruntime_test_all --gtest_filter=$gtest_filter
+
+echo "Running ./onnxruntime_provider_test .."
+$build_dir/onnxruntime_provider_test --gtest_filter=$gtest_filter
```
edgchen1: this file is not referenced in any other repo code. can it be removed?

onnxruntime/test/providers/provider_test_utils.h
```diff
@@ -3,47 +3,8 @@
 
 #pragma once
 
-#include "test/providers/checkers.h"
-#include "test/providers/op_tester.h"
-#include "test/providers/model_tester.h"
-
-namespace onnxruntime {
```
edgchen1, line 10: note: moved this to test/unittest_util/conversion.h

cmake/onnxruntime_unittests.cmake
```diff
@@ -775,6 +883,9 @@ file(GLOB onnxruntime_test_framework_src CONFIGURE_DEPENDS
   ${onnxruntime_test_framework_src_patterns}
   )
 
+# TODO either update the below comment or follow it...
+# it's not obvious why this limitation of not using onnxruntime internal symbols exists
```
edgchen1: @snnn could you provide more detail about why this comment says we shouldn't use ORT internal symbols?

```
#This is a small wrapper library that shouldn't use any onnxruntime internal symbols(except onnxruntime_common).
#Because it could dynamically link to onnxruntime. Otherwise you will have two copies of onnxruntime in the same
#process and you won't know which one you are testing.
```

we currently aren't following that. e.g., in `onnxruntime/test/util/compare_ortvalue.cc`, it calls a helper from `onnxruntime/core/framework/tensorprotoutils.h`.
https://github.com/microsoft/onnxruntime/blob/d6e712c5b7b6260a61e54d1fe40107cf5366ee77/onnxruntime/test/util/compare_ortvalue.cc#L597

I was hoping to add more files to `onnxruntime_test_utils` to make them reusable between `onnxruntime_test_all` and other test programs.

cmake/onnxruntime_unittests.cmake
```diff
@@ -1095,6 +1218,87 @@ if (NOT onnxruntime_ENABLE_TRAINING_TORCH_INTEROP)
...
+      set_property(TARGET onnxruntime_provider_test APPEND_STRING PROPERTY LINK_FLAGS " --pre-js \"${ONNXRUNTIME_ROOT}/wasm/pre-jsep.js\"")
+    endif()
+
+    ###
+    ### if you want to investigate or debug a test failure in onnxruntime_provider_test, replace the following line.
+    ### those flags slow down the CI test significantly, so we don't use them by default.
+    ###
+    #   set_property(TARGET onnxruntime_provider_test APPEND_STRING PROPERTY LINK_FLAGS " -s ASSERTIONS=2 -s SAFE_HEAP=1 -s STACK_OVERFLOW_CHECK=2")
+    set_property(TARGET onnxruntime_provider_test APPEND_STRING PROPERTY LINK_FLAGS " -s ASSERTIONS=0 -s SAFE_HEAP=0 -s STACK_OVERFLOW_CHECK=1")
+  endif()
```
edgchen1, line 1261: @fs-eire do you think any of these should be added to test programs more generally, e.g., in `AddTest()`?

onnxruntime/test/util/include/test_dynamic_plugin_ep.h
```diff
@@ -0,0 +1,72 @@
...
+  std::map<std::string, std::string> default_ep_options{};
+};
+
+// Parses `InitializationConfig` from JSON.
+// The configuration JSON object should have keys and values that match the `InitializationConfig` fields.
+// E.g.:
+// {
+//   "ep_library_registration_name": "example_plugin_ep",
+//   "ep_library_path": "/path/to/example_plugin_ep.dll",
+//   "selected_ep_device_indices": [1],
```
adrianlizarraga: Maybe this is a naive question, but why do we specify the indices of the OrtEpDevices instead of the EP names?

I would imagine that getting the indices would require the user to call GetEpDevices to get the indices.

edgchen1: it's a naive implementation :). open to better ways of specifying it. yes, at the moment the user should check the EP devices first. this can be done with `onnxruntime_perf_test --list_ep_devices`.

specifying the particular EP devices seemed to be the most flexible, but perhaps it is just passing more complexity to the user. if we specify an EP name instead, what would be a sensible way to handle multiple EP devices associated with one EP name?

cmake/onnxruntime_unittests.cmake
snnn: Let's assume we are building it for Linux, you have a test program that links to onnxruntime_framework.lib and a bunch of other open source libraries, and the test program also links to libonnxruntime.so which also contains onnxruntime_framework.lib and some others things.  Would the two copies of onnxruntime_framework.lib  have conflicts? I am not sure. 
Now it uses onnxruntime_common.lib which is relatively simple. onnxruntime_common.lib  didn't have much global variables that are stateful.

snnn: Previously I found in some cases that libonnxruntime.so would prefer to use the symbols from the main executable instead of the so itself if the symbol can  be found in both places. I don't remember the details.  It highly depends on which link flags were used when generating the two binaries.  This is rather complicated. Therefore I hope we can avoid it   as much as possible.

tools/ci_build/github/pai/pai_test_launcher.sh
snnn: Yes.

onnxruntime/test/framework/allocation_planner_test.cc
```diff
@@ -33,13 +33,6 @@ using json = nlohmann::json;
 #include "core/session/onnxruntime_session_options_config_keys.h"
 using namespace ONNX_NAMESPACE;
 
-// Explicitly provide a definition for the static const var 'GPU' in the OrtDevice struct,
-// GCC 4.x doesn't seem to define this and it breaks the pipelines based on CentOS as it uses
-// GCC 4.x.
-// (This static var is referenced in some tests below)
-const OrtDevice::DeviceType OrtDevice::GPU;
-const OrtDevice::DeviceType OrtDevice::CPU;
```
edgchen1, line 41: note: changed these to be constexpr. the explicit definition here was causing problems elsewhere.

onnxruntime/test/unittest_util/framework_test_utils.cc
```diff
@@ -1,6 +1,6 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-#include "test_utils.h"
+#include "framework_test_utils.h"
```
edgchen1:
```suggestion
#include "test/unittest_util/framework_test_utils.h"
```

cmake/onnxruntime_unittests.cmake
edgchen1: ok. I added a new static library `onnxruntime_unittest_utils` for utilities that can use internal onnxruntime libraries. we can avoid using it when we need to link to the ORT shared library.

onnxruntime/test/util/include/test_dynamic_plugin_ep.h
edgchen1: I added the option to use an EP name. we'll only allow one plugin EP here.

cmake/onnxruntime_unittests.cmake
```diff
@@ -1095,6 +1196,79 @@ if (NOT onnxruntime_ENABLE_TRAINING_TORCH_INTEROP)
   endif()
 endif()
 
+# onnxruntime_provider_test
+# Execution provider-related tests.
+# These also have some support for dynamically specified plugin EPs.
+if (NOT onnxruntime_MINIMAL_BUILD AND NOT onnxruntime_REDUCED_OPS_BUILD)
+block()
```
adrianlizarraga, line 1203: I tried looking at the CMake docs for block() but didn't fully understand it. Is the main purpose of block() here to prevent internal variables from leaking to the parent scope? Just curious.

onnxruntime/test/providers/cpu/tensor/quantize_linear_test.cc
```diff
@@ -973,10 +973,11 @@ void DequantizeLinearOp21BlockedTest_InvalidBlockSize_Int4(int64_t block_size,
   so.session_log_verbosity_level = 1;
   so.graph_optimization_level = TransformerLevel::Default;
 
+  using UnpackedType = typename Tin::UnpackedType;
   for (int64_t i = 0, n = zero_point_block_count; i < n; ++i) x_zero_point.push_back(Tin(0, 0));
   for (int64_t i = 0, n = 2 * scale_block_count; i < n; i++) x_scale.push_back(Tout(2.0f));
   for (int i = 0; i < 8; ++i) {
-    if (i & 1) x.push_back(Tin(i - 1, i));
+    if (i & 1) x.push_back(Tin(static_cast<UnpackedType>(i - 1), static_cast<UnpackedType>(i)));
```
adrianlizarraga, line 980: Is the lack of static casting now being caught because the new onnxruntime_provider_test executable has more strict warnings enabled? If so, great :)

onnxruntime/test/unittest_util/base_tester.cc
```diff
@@ -681,9 +681,25 @@ void BaseTester::RunWithConfig(size_t* number_of_pre_packed_weights_counter,
       }
 #endif
 
+      const auto dynamic_plugin_ep_name = dynamic_plugin_ep_infra::GetEpName();
```
adrianlizarraga, line 684: Does this need to be guarded by `#ifdef ORT_UNIT_TEST_ENABLE_DYNAMIC_PLUGIN_EP` ?

cmake/onnxruntime_unittests.cmake
edgchen1: yep, that's exactly the intent.

onnxruntime/test/providers/cpu/tensor/quantize_linear_test.cc
edgchen1: right, onnxruntime_test_all was disabling additional warnings.

onnxruntime/test/unittest_util/base_tester.cc
edgchen1: we don't need to check that here. `ORT_UNIT_TEST_ENABLE_DYNAMIC_PLUGIN_EP` mainly controls whether the dynamic plugin EP infra `Initialize()` and `Shutdown()` are called from test_main.cc. if `Initialize()` is not called, `GetEpName()` will return `std::nullopt`.

maybe I can make the macro name clearer.

cmake/onnxruntime_unittests.cmake
edgchen1: will leave it as is for now

TedThemistokleous:
Hey , just seeing this now @edgchen1 @snnn.

The MIGraphX was using the pai_test_launcher.sh script  in our CI so this has caused a build break for MIGraphX. We typically sync weekly to your Mainline and build against Onnxruntime to ensure MIGraphX EP builds off your main won't cause breaks. |

Is there something else we should be using now here instead? Should we just run the test launcher directly? From what I can tell, it seems like the Onnxruntime_test_all binary still exists I think we just need to add a filter for excluded tests still? 

Is there a larger change here for builds/testing that we need to incorporate in our infrastructure? If so please let me know

edgchen1:
> Hey , just seeing this now @edgchen1 @snnn.
> 
> The MIGraphX was using the pai_test_launcher.sh script in our CI so this has caused a build break for MIGraphX. We typically sync weekly to your Mainline and build against Onnxruntime to ensure MIGraphX EP builds off your main won't cause breaks. |
> 
> Is there something else we should be using now here instead? Should we just run the test launcher directly? From what I can tell, it seems like the Onnxruntime_test_all binary still exists I think we just need to add a filter for excluded tests still?
> 
> Is there a larger change here for builds/testing that we need to incorporate in our infrastructure? If so please let me know

Sorry. I didn't find any existing usages in the repo, so I assumed it was fine to delete. You can add it back if you want.

The main change is to move some tests from onnxruntime_test_all to a new test program, onnxruntime_provider_test. To get the same test coverage, you'd want to also run the new test program. You may need to adjust your test filters accordingly, e.g., if they match a test that got moved.

TedThemistokleous:
Good to know. Sure I can add that back in but if I can just call in the same piece, I can adjust CI directly then if this is the case.

TedThemistokleous:
I think this was used when there was MIGraphX EP/ROCm EP CI but that got removed recently and we're in the process of adding it back

Issue #25714, opened 2025-08-11
Only 1 free dimension argument will be parsed and set in latest command parser of onnxruntime_perf_test


qwu16:
After PR https://github.com/microsoft/onnxruntime/pull/25374, the latest onnxruntime_perf_test.exe will only set 1 free dimension argument even we set multiple free dimensions in command line, so the previous command line settings like "onnxruntime_perf_test.exe -I -r 50 -o 1 -e openvino -i "device_type|GPU" -f batch:2 -f sequence:77 sd-v1.5-text-encoder-demo.onnx", only "sequence" will be overridden, "batch" will be not overridden with 2.

chilo-ms:
Please see the [PR ](https://github.com/microsoft/onnxruntime/pull/25716)for fixing this bug.

If an option appears multiple times, unlike `getopt` just returns it again in the parsing loop, `Abseil` processes them in order, and the last one wins (overwrites earlier values).

Now please specify free dimensions by name/denotation to override with the specific values as key-value pairs are separated by space.
i.e. `-f  "batch:2 sequence:77 "`.

qwu16:
> Please see the [PR ](https://github.com/microsoft/onnxruntime/pull/25716)for fixing this bug.
> 
> If an option appears multiple times, unlike `getopt` just returns it again in the parsing loop, `Abseil` processes them in order, and the last one wins (overwrites earlier values).
> 
> Now please specify free dimensions by name/denotation to override with the specific values as key-value pairs are separated by space. i.e. `-f "batch:2 sequence:77 "`.

This will make the previous way to set multiple free dimensions not work.

chilo-ms:
> This will make the previous way to set multiple free dimensions not work.

I updated to keep the previous way to set multiple free dimensions

MayureshV1:
@qwu16 . Can this issue be closed since @chilo-ms has a [PR ](https://github.com/microsoft/onnxruntime/pull/25716)which is already merged?

qwu16:
Sure.

PR #25716, opened 2025-08-11
Fix bug for same option provided multiple times in perf test

chilo-ms:
### Description
If an option appears multiple times:
Unlike `getopt` just returns it again in the parsing loop, `Abseil` processes them in order, and the last one wins (overwrites earlier values).

This PR fixes the bug for `-f` free dimension override by name and `-F `free dimension override by denotation. 
see https://github.com/microsoft/onnxruntime/issues/25714

### DIFF
--- a/onnxruntime/test/perftest/command_args_parser.cc
+++ b/onnxruntime/test/perftest/command_args_parser.cc
@@ -30,8 +30,12 @@ static const onnxruntime::perftest::PerformanceTestConfig& DefaultPerformanceTes
   return default_config;
 }
 
-ABSL_FLAG(std::string, f, "", "Specifies a free dimension by name to override to a specific value for performance optimization.");
-ABSL_FLAG(std::string, F, "", "Specifies a free dimension by denotation to override to a specific value for performance optimization.");
+ABSL_FLAG(std::string, f, "",
+          "Specifies free dimensions by name to override with the specific values for performance optimization. Key-value paris are separated by space.\n"
+          "[Usage]: -f \"dimension_name_1:override_value_1 dimension_name_2:override_value_2 ... \". Override value must > 0.");
+ABSL_FLAG(std::string, F, "",
+          "Specifies free dimensions by denotation to override with the specific values for performance optimization. Key-value paris are separated by space.\n"
+          "[Usage]: -F \"denotation_name_1:override_value_1 denotation_name_2:override_value_2 ... \". Override value must > 0.");
 ABSL_FLAG(std::string, m, "duration", "Specifies the test mode. Value could be 'duration' or 'times'.");
 ABSL_FLAG(std::string, e, "cpu", "Specifies the provider 'cpu','cuda','dnnl','tensorrt', 'nvtensorrtrtx', 'openvino', 'dml', 'acl', 'nnapi', 'coreml', 'qnn', 'snpe', 'rocm', 'migraphx', 'xnnpack', 'vitisai' or 'webgpu'.");
 ABSL_FLAG(size_t, r, DefaultPerformanceTestConfig().run_config.repeated_times, "Specifies the repeated times if running in 'times' test mode.");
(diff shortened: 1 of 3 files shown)


onnxruntime/test/perftest/strings_helper.cc
```diff
@@ -56,6 +56,34 @@ void ParseSessionConfigs(const std::string& configs_string,
...
+      std::string override_val_str = free_dim_str.substr(delimiter_location + 1, std::string::npos);
+      ORT_TRY {
+        int64_t override_val = std::stoll(override_val_str.c_str());
+        if (override_val <= 0) {
+          return false;
+        }
+        free_dim_override_map[dim_identifier] = override_val;
+      }
+      ORT_CATCH(...) {
+        return false;
```
yuslepukhin, line 82: Perhaps, we could print a message from std::exception to give more information?

onnxruntime/test/perftest/command_args_parser.cc
```diff
@@ -30,8 +30,12 @@ static const onnxruntime::perftest::PerformanceTestConfig& DefaultPerformanceTes
   return default_config;
 }
 
-ABSL_FLAG(std::string, f, "", "Specifies a free dimension by name to override to a specific value for performance optimization.");
-ABSL_FLAG(std::string, F, "", "Specifies a free dimension by denotation to override to a specific value for performance optimization.");
+ABSL_FLAG(std::string, f, "",
+          "Specifies free dimensions by name to override with the specific values for performance optimization. Key-value pairs are separated by space.\n"
+          "[Usage]: -f \"dimension_name_1:override_value_1 dimension_name_2:override_value_2 ... \". Override value must > 0.");
+ABSL_FLAG(std::string, F, "",
```
yuslepukhin, line 37: This would break the existing scripts.
This is a public API.

chilo-ms: There is a constraint in `Abseil`, using it to parse the same option provided multiple times, only the last one will be set. 
It seems we can't make `-f`/`-F` work without changing the way of using them.

onnxruntime/test/perftest/strings_helper.cc
chilo-ms: added

onnxruntime/test/perftest/strings_helper.cc
```diff
@@ -56,6 +56,37 @@ void ParseSessionConfigs(const std::string& configs_string,
...
+      ORT_TRY {
+        int64_t override_val = std::stoll(override_val_str.c_str());
+        if (override_val <= 0) {
+          return false;
+        }
+        free_dim_override_map[dim_identifier] = override_val;
+      }
+      ORT_CATCH(const std::exception& ex) {
+        ORT_HANDLE_EXCEPTION([&]() {
+          fprintf(stderr, "Error parsing free dimension override value: %s, %s\n", override_val_str.c_str(), ex.what());
```
yuslepukhin: This may not flush in time if the process goes down. I suggest using std::cerr with `std::endl` at the end which includes `std::flush`.

onnxruntime/test/perftest/command_args_parser.cc
yuslepukhin: Yes, it is a library constraing, but imaging how many things it would break. This very ticket is about breaking the commnand line expectations.

chilo-ms: okay, i have a workaround to manually parse the values for '-f' and '-F' in argv to keep previous usage.

onnxruntime/test/perftest/strings_helper.cc
chilo-ms: updated

onnxruntime/test/perftest/strings_helper.h
```diff
@@ -14,6 +15,10 @@ void ParseSessionConfigs(const std::string& configs_string,
                          std::unordered_map<std::string, std::string>& session_configs,
                          const std::unordered_set<std::string>& available_keys = {});
 
+bool ParseDimensionOverride(const std::string& input, std::map<std::string, int64_t>& free_dim_override_map);
+
+bool ParseDimensionOverrideFromArgv(int argc, std::vector<std::string> argv, std::string option, std::map<std::string, int64_t>& free_dim_override_map);
```
yuslepukhin: > std::string option [](http://example.com/codeflow?start=76&length=19)

reference or string_view?

chilo-ms: changed to use reference.

qwu16:
Hi, is there any plan when this PR will be merged? It broke our daily performance process. It would be great that onnxruntime_perf_test tool can support multiple free dimension settings again.

chilo-ms:
> Hi, is there any plan when this PR will be merged? It broke our daily performance process. It would be great that onnxruntime_perf_test tool can support multiple free dimension settings again.

Once all CI pass, it will be merged to main. Should be done today.

PR #25721, opened 2025-08-12
Fix build with musl

jaeyoonjung:
### Description

Added __GLIBC__ macro for 'execinfo.h' and related functions as they are GLIBC specific and not available in other libc implementation like musl.

### Motivation and Context

This allows to compile onnxruntime with musl, an alternative libc implementation.


jaeyoonjung:
Resubmitted as https://github.com/microsoft/onnxruntime/pull/25726

PR #25741, opened 2025-08-13
[c++] Eliminate dynamic initialization of static Ort::Global<void>::api_

chwarr:
### Description

Delay the call to `OrtGetApiBase()` until the first call to `Ort::GetApi()` so that `OrtGetApiBase()` is typically called after dynamic library loading.

### Motivation and Context

When ORT_API_MANUAL_INIT is not defined (which is the default), the static `Ort::Global<void>::api_` has a dynamic initializer that calls `OrtGetApiBase()->GetApi(ORT_API_VERSION)` This dynamic initialization can cause problems when it interacts with other global/static initialization. On Windows in particular, it can also cause deadlocks when used in a dynamic library if OrtGetApiBase()->GetApi() attempts to load any other libraries.

* Replace the templated `Global<void>::api_` with an inline static initialized to nullptr.
* `Ort::GetApi()` now calls `detail::Global::GetApi()` which calls `detail::Global::DefaultInit()` if initialization is needed.
  * When `ORT_API_MANUAL_INIT` is defined, `DefaultInit()` returns nullptr, which will eventually cause the program to crash. The callers have violated the initialization contract by not calling one of the `Ort::InitApi` overloads.
  * When `ORT_API_MANUAL_INIT` is not defined, `DefaultInit()` uses a function-level static to compute the result of `OrtGetApiBase()->GetApi(ORT_API_VERSION)` once and return it.
* `Ort::Global<void>` has been replaced with a non-templated type and moved inside a `detail` namespace. Since the `Global<void>` object was documented as being used internally, it is believed that these changes here are non-breaking, as they do not impact a public API. The public APIs, `Ort::InitApi()` and `Ort::InitApi(const OrtApi*)` remain unchanged.
* Add `#pragma detect_mismatch` to surface issues with compilation units that disagree on how ORT_API_MANUAL_INIT is defined. (MSVC only.)

### DIFF
--- a/include/onnxruntime/core/session/onnxruntime_cxx_api.h
+++ b/include/onnxruntime/core/session/onnxruntime_cxx_api.h
@@ -166,7 +166,7 @@ inline const OrtApi* detail::Global::DefaultInit() noexcept {
   // This makes it safe to have a Ort::Env constructed as a static member.
   //
   // This DOES NOT make it safe to _use_ arbitrary ORT C++ APIs when
-  // initializaing static members, however.
+  // initializing static members, however.
   static const OrtApi* api = OrtGetApiBase()->GetApi(ORT_API_VERSION);
   return api;
 }
--- a/include/onnxruntime/core/session/onnxruntime_cxx_api.h
+++ b/include/onnxruntime/core/session/onnxruntime_cxx_api.h
@@ -79,22 +79,17 @@ struct Exception : std::exception {
   throw Ort::Exception(string, code)
 #endif
 
-// This is used internally by the C++ API. This class holds the global variable that points to the OrtApi,
-//  it's in a template so that we can define a global variable in a header and make
-// it transparent to the users of the API.
-template <typename T>
-struct Global {
-  static const OrtApi* api_;
-};
+// If macro ORT_API_MANUAL_INIT is defined, no static initialization will be
+// performed. Instead, users must call InitApi() before using ORT.
+//
+// InitApi() sets the global API object using the default initialization
+// logic. Users call this to initialize the ORT C++ APIs at a time that
+// makes sense in their program.
+inline void InitApi() noexcept;
 
-// If macro ORT_API_MANUAL_INIT is defined, no static initialization will be performed. Instead, user must call InitApi() before using it.
-template <typename T>
-#ifdef ORT_API_MANUAL_INIT
-const OrtApi* Global<T>::api_{};
-inline void InitApi() noexcept { Global<void>::api_ = OrtGetApiBase()->GetApi(ORT_API_VERSION); }
-
-// Used by custom operator libraries that are not linked to onnxruntime. Sets the global API object, which is
-// required by C++ APIs.
+// InitApi(const OrtApi*) is used by custom operator libraries that are not
+// linked to onnxruntime. It sets the global API object, which is required
+// by the ORT C++ APIs.
 //
 // Example mycustomop.cc:
 //
@@ -107,22 +102,78 @@ inline void InitApi() noexcept { Global<void>::api_ = OrtGetApiBase()->GetApi(OR
 //   // ...
 // }
 //
-inline void InitApi(const OrtApi* api) noexcept { Global<void>::api_ = api; }
-#else
-#if defined(_MSC_VER) && !defined(__clang__)
-#pragma warning(push)
-// "Global initializer calls a non-constexpr function." Therefore you can't use ORT APIs in the other global initializers.
-// Please define ORT_API_MANUAL_INIT if it conerns you.
-#pragma warning(disable : 26426)
-#endif
-const OrtApi* Global<T>::api_ = OrtGetApiBase()->GetApi(ORT_API_VERSION);
-#if defined(_MSC_VER) && !defined(__clang__)
-#pragma warning(pop)
-#endif
+inline void InitApi(const OrtApi* api) noexcept;
+
+namespace detail {
+// This is used internally by the C++ API. This class holds the global
+// variable that points to the OrtApi.
+struct Global {
+  static const OrtApi* GetApi() noexcept {
+    if (!api_) {
+      api_ = DefaultInit();
+    }
+
+    return api_;
+  }
+
+ private:
+  inline static const OrtApi* api_ = nullptr;
+
+  // Has different definitions based on ORT_API_MANUAL_INIT
+  static const OrtApi* DefaultInit() noexcept;
+
+  // Public APIs to set the OrtApi* to use.
+  friend void ::Ort::InitApi() noexcept;
+  friend void ::Ort::InitApi(const OrtApi*) noexcept;
+};
+}  // namespace detail
+
+#ifdef ORT_API_MANUAL_INIT
+
+inline void InitApi(const OrtApi* api) noexcept { detail::Global::api_ = api; }
+inline void InitApi() noexcept { InitApi(OrtGetApiBase()->GetApi(ORT_API_VERSION)); }
+
+#if defined(__clang__) || defined(_MSC_VER)
+// If you get a linker error about a mismatch here, you are trying to
+// link two compilation units that have different definitions for
+// ORT_API_MANUAL_INIT together. All compilation units must agree on the
+// definition of ORT_API_MANUAL_INIT.
+#pragma detect_mismatch("ORT_API_MANUAL_INIT", "enabled")
 #endif
 
+inline const OrtApi* detail::Global::DefaultInit() noexcept {
+  // When ORT_API_MANUAL_INIT is defined, there's no default init that can
+  // be done.
+  return nullptr;
+}
+
+#else  // ORT_API_MANUAL_INIT
+
+#if defined(__clang__) || defined(_MSC_VER)
+// If you get a linker error about a mismatch here, you are trying to link
+// two compilation units that have different definitions for
+// ORT_API_MANUAL_INIT together. All compilation units must agree on the
+// definition of ORT_API_MANUAL_INIT.
+#pragma detect_mismatch("ORT_API_MANUAL_INIT", "disabled")
+#endif
+
+inline const OrtApi* detail::Global::DefaultInit() noexcept {
+  // This block-level static will be initialized once when this function is
+  // first executed, delaying the call to OrtGetApiBase()->GetApi() until it
+  // is first needed. This helps avoid issues with static initialization
+  // order and dynamic libraries loading other dynamic libraries.
+  //
+  // This makes it safe to have a Ort::Env constructed as a static member.
+  //
+  // This DOES NOT make it safe to _use_ arbitrary ORT C++ APIs when
+  // initializaing static members, however.
+  static const OrtApi* api = OrtGetApiBase()->GetApi(ORT_API_VERSION);
+  return api;
+}
+#endif  // ORT_API_MANUAL_INIT
+
 /// This returns a reference to the ORT C API.
-inline const OrtApi& GetApi() noexcept { return *Global<void>::api_; }
+inline const OrtApi& GetApi() noexcept { return *detail::Global::GetApi(); }
 
 /// <summary>
 /// This function returns the onnxruntime version string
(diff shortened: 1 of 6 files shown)


jonwis (approved):

onnxruntime/test/autoep/library/ep_arena.h
```diff
@@ -21,7 +21,10 @@ limitations under the License.
 #include <mutex>
 #include <set>
 
+#define ORT_API_MANUAL_INIT
 #include "onnxruntime_cxx_api.h"
+#undef ORT_API_MANUAL_INIT
```
skottmckay, line 26: What's the reason for this change? It changes the code to require manual initialization but there's no other change to do the initialization. Should something in example_plugin_ep.cc be calling InitApi with the OrtApiBase* the library receives on load?

skottmckay:
> On Windows in particular, it can also cause deadlocks when used in a dynamic library if OrtGetApiBase()->GetApi() attempts to load any other libraries.

When does this happen? AFAIK the ORT implementation simply returns a struct from OrtGetApiBase and GetApi(version) does a version check. What is triggering other library loads as part of that process?

https://github.com/microsoft/onnxruntime/blob/3a94a670e4502b03991ad5f56ac4169c7398dcf1/onnxruntime/core/session/onnxruntime_c_api.cc#L4162-L4164

https://github.com/microsoft/onnxruntime/blob/3a94a670e4502b03991ad5f56ac4169c7398dcf1/onnxruntime/core/session/onnxruntime_c_api.cc#L4142-L4152

onnxruntime/test/autoep/library/ep_arena.h
chwarr: It fixes an ODR violation due to [example_plugin_ep_util.h already manual init](https://github.com/microsoft/onnxruntime/blob/3a94a670e4502b03991ad5f56ac4169c7398dcf1/onnxruntime/test/autoep/library/example_plugin_ep_utils.h#L9C1-L11C27) but this file not.

With the `#pragma detect_mismatch` that I added, the linker told me that there were object files in the example_plugin_ep shared library that had different definition. So, since this shared library was already using manual init, I made all its compilation units agree.

ORT_API_MANUAL_INIT changes the definition the current tree's  Global<T>::api_ initializer and my proposed DefaultInit. I couldn't come up with a way to implementing this that didn't involve some ODR risk.

include/onnxruntime/core/session/onnxruntime_cxx_api.h
```diff
@@ -79,22 +79,17 @@ struct Exception : std::exception {
...
-
-// Used by custom operator libraries that are not linked to onnxruntime. Sets the global API object, which is
-// required by C++ APIs.
+// If macro ORT_API_MANUAL_INIT is defined, no static initialization will be
+// performed. Instead, users must call InitApi() before using ORT.
+//
+// InitApi() sets the global API object using the default initialization
+// logic. Users call this to initialize the ORT C++ APIs at a time that
+// makes sense in their program.
+inline void InitApi() noexcept;
```
chwarr, line 90: With this change, `InitApi` is now always visible. Before, it was only visible when `ORT_API_MANUAL_INIT` was defined. Preserve that behavior.

include/onnxruntime/core/session/onnxruntime_cxx_api.h
```diff
@@ -107,22 +102,78 @@ inline void InitApi() noexcept { Global<void>::api_ = OrtGetApiBase()->GetApi(OR
...
+
+  // Public APIs to set the OrtApi* to use.
+  friend void ::Ort::InitApi() noexcept;
+  friend void ::Ort::InitApi(const OrtApi*) noexcept;
+};
+}  // namespace detail
+
+#ifdef ORT_API_MANUAL_INIT
+
+inline void InitApi(const OrtApi* api) noexcept { detail::Global::api_ = api; }
```
yuslepukhin: Need a comment about intended usage referring to the above #Resolved

yuslepukhin (changes requested):

include/onnxruntime/core/session/onnxruntime_cxx_api.h
```diff
@@ -107,22 +102,78 @@ inline void InitApi() noexcept { Global<void>::api_ = OrtGetApiBase()->GetApi(OR
...
+  static const OrtApi* GetApi() noexcept {
+    if (!api_) {
+      api_ = DefaultInit();
+    }
+
+    return api_;
+  }
+
+ private:
+  inline static const OrtApi* api_ = nullptr;
```
yuslepukhin: >inline static const OrtApi* api_ = nullptr; [](http://example.com/codeflow?start=2&length=43)

Two questions.
1) Have you considered this to be a function local static?
2) Is this going to be a separate instance in each compilation unit?

chwarr: 1. I played around with that, but found that the function was harder to understand. If that's the way you'd prefer I go, I can do that. I think it looked something like this with an extra param to control whether this was intentionally setting to nullptr. Maybe that isn't needed, and nullptr for the parameter can be interpreted as just getting the value.
   ```cpp
   OrtApi* GetOrSetTheApi(OrtApi* newApi, bool shouldSet) {
     static OrtApi* api = nullptr;
     if (shouldSet) {
       api = newApi;
     }

     return api;
   }
   ```
2. class static members are guaranteed to be shared across compilation units (unless the class is inside an anonymous namespace, which is not the case here). It's not the standard, but [cppreference says](https://en.cppreference.com/w/cpp/language/static.html):
   > Static data members are not associated with any object. They exist even if no objects of the class have been defined. There is only one instance of the static data member in the entire program with static [storage duration](https://en.cppreference.com/w/cpp/language/storage_duration.html), unless the keyword [thread_local](https://en.cppreference.com/w/cpp/keyword/thread_local.html) is used, in which case there is one such object per thread with thread storage duration

chwarr: 2. The existing code uses a static member as well. The `inline` keyword I'm adding here allows me to specify the initialization inside the class definition. It doesn't change other semantics. That is, this could be written as
   ```cpp
   class Global {
     static const OrtApi* api_;
   }
   Global::api_ = nullptr;
   ```

chwarr:
> > On Windows in particular, it can also cause deadlocks when used in a dynamic library if OrtGetApiBase()->GetApi() attempts to load any other libraries.
> 
> When does this happen? AFAIK the ORT implementation simply returns a struct from OrtGetApiBase and GetApi(version) does a version check. What is triggering other library loads as part of that process?

@skottmckay, yeah, I could have made that sentence clearer. :-) How's this? "On Windows in particular, it can also cause deadlocks when used in a dynamic library if calling `OrtGetApiBase()->GetApi()` causes other libraries to be loaded."

With the current implementation, if a .cpp that is part of foo.dll has `#include <onnxruntime_cxx_api.h>`, the presence of the `Global<void>::api_ = OrtGetApiBase()->GetApi(ORT_API_VERSION);` static initializer means that a call to `OrtGetApiBase()` must happen when foo.dll is being loaded and it holds the loader lock. If `OrtGetApiBase()` is provided by another DLL like `onnxruntime.dll` and that DLL isn't loaded yet (e.g., it's [delayloaded](https://learn.microsoft.com/cpp/build/reference/linker-support-for-delay-loaded-dlls?view=msvc-170)), then foo.dll ends up attempting to make a call to LoadLibrary while holding the loader lock.

The [guidance on Windows is to do as little as possible when the DLL is loaded](https://learn.microsoft.com/windows/win32/dlls/dynamic-link-library-best-practices). In C++, this means that your global and static initializers effectively to be 0/constant values.

This change does that: `Ort::Global::api_` is now initialized to 0. The interesting initialization logic will happen the first time `Ort::GetApi()` is called, hopefully outside of the loader lock.

include/onnxruntime/core/session/onnxruntime_cxx_api.h
yuslepukhin: Thanks for the additional explanation.
I have thought about it, and I would favor making it a function local static because during concurrent calls to `GetApi()`
1) it would make it easier to reason about the `api_` ptr initialization
2) alleviate some static object concerns during DLL loading.

chwarr: I've revised this to store the OrtApi* as a function-level static.
Instead of having a `bool shouldSet`, I picked using `nullptr` to skip the set. This does mean that you can no longer call InitApi(nullptr) to clear the point. If this will be a problem, I can revise this again.

include/onnxruntime/core/session/onnxruntime_cxx_api.h
chwarr: I moved the existing documentation above to the declarations. I've added a comment here point those that.

yuslepukhin (dismissed):

yuslepukhin (approved):
:shipit:

chwarr:
Five of the pending checks have been "Waiting for status to be reported" for the past two days. How can I kick them again? I don't see any GitHub UX to do this. I assume I'm lacking the right permissions.

I'm unsure how to handle [this build issue from the CUDA build](https://github.com/microsoft/onnxruntime/actions/runs/17225475677/job/48869109524?pr=25741). It looks like `vcpkg install` failed to process telemetry. Is there a way to re-run this check? I don't see any GitHub UX to do this. I assume I'm lacking the right permissions.

```
 -- Running vcpkg install
warning: feature cuda-ep was passed, but that is not a feature supported by onnxruntime supports.
Fetching registry information from https://github.com/Microsoft/vcpkg (HEAD)...
Fatal error. System.Runtime.InteropServices.SEHException (0x80004005): External component has thrown an exception.
   at System.Collections.Generic.Dictionary`2[[System.__Canon, System.Private.CoreLib, Version=8.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e],[System.Collections.Generic.KeyValuePair`2[[System.__Canon, System.Private.CoreLib, Version=8.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e],[System.Double, System.Private.CoreLib, Version=8.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]], System.Private.CoreLib, Version=8.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]]..ctor(Int32)
   at Microsoft.ApplicationInsights.Extensibility.Implementation.Property.SanitizeMeasurements(System.Collections.Generic.IDictionary`2<System.String,Double>)
   at Microsoft.ApplicationInsights.Extensibility.Implementation.JsonSerializer.SerializeToStream(System.Collections.Generic.IEnumerable`1<Microsoft.ApplicationInsights.Channel.ITelemetry>, System.IO.TextWriter)
   at Microsoft.ApplicationInsights.Extensibility.Implementation.JsonSerializer.Serialize(System.Collections.Generic.IEnumerable`1<Microsoft.ApplicationInsights.Channel.ITelemetry>, Boolean)
   at Microsoft.ApplicationInsights.Channel.InMemoryTransmitter.Send(System.Collections.Generic.IEnumerable`1<Microsoft.ApplicationInsights.Channel.ITelemetry>, System.TimeSpan)
   at Microsoft.ApplicationInsights.Channel.InMemoryTransmitter.DequeueAndSend(System.TimeSpan)
   at Microsoft.ApplicationInsights.Channel.InMemoryTransmitter.Runner()
   at System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object)
   at System.Threading.Tasks.Task.ExecuteWithThreadLocal(System.Threading.Tasks.Task ByRef, System.Threading.Thread)
```

chwarr:
Thanks for the help re-running, @yuslepukhin.

Are you able to merge this too?

The [PR guidelines](https://github.com/microsoft/onnxruntime/blob/main/docs/PR_Guidelines.md) say that I should be the one to do the merge, but I only see "Close with comment" and "Comment" buttons. No "Merge" button. I assume this is a permissions issue too.

PR #25744, opened 2025-08-14
Ort 1 23 ovep fixes

javier-intel:
### Description
This PR fixes additional load_config handling logic issues and a failure when (ep.share_ep_contexts && !ep.context_enable) are true.

### DIFF
--- a/onnxruntime/core/providers/openvino/openvino_execution_provider.cc
+++ b/onnxruntime/core/providers/openvino/openvino_execution_provider.cc
@@ -94,18 +94,23 @@ common::Status OpenVINOExecutionProvider::Compile(
   auto& logger = *GetLogger();
   Status status = Status::OK();
 
+  bool is_epctx_model = false;
   if (!fused_nodes.empty()) {
     // Assume these properties are constant for all the model subgraphs, otherwise move to SubGraphContext
     const auto& graph_body_viewer_0 = fused_nodes[0].filtered_graph.get();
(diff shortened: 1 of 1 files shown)


sfatimar (approved):

HectorSVC:
some test failure?

javier-intel:
> some test failure?

That's odd. I see OV failures but not sure where they're coming from, those logs a huge and can't find the error message. The validation team tested each commit before the PR was raised so I'm not so concerned about the OV issues. Regarding the MacOS failures the PR only touches OVEP files, this is likely a false negative.

jywu-mysoft:
@javier-intel any updates?

javier-intel:
> @javier-intel any updates?

More or less, here's what I have. We had ran `onnxruntime_test_all` and it passed so I'm working on getting the previous results to compare as well as doing a retest to see if we can reproduce the issue locally. I'll post another update with the repro results on our side.

jywu-mysoft:
> > @javier-intel any updates?
> 
> More or less, here's what I have. We had ran `onnxruntime_test_all` and it passed so I'm working on getting the previous results to compare as well as doing a retest to see if we can reproduce the issue locally. I'll post another update with the repro results on our side.

there is some config related to precision that he changed. is it resulting in a different default precision being used?  because the test failure is output mismatch but the results/expected look close. perhaps the tolerance requires adjusting?

javier-intel:
We were able to reproduce the regression. Given there are other changes for 1.23 I'm closing this PR and raise a new one once our full scope validation completes.

PR #25745, opened 2025-08-14
[QNN-EP] Fix int64 graph output issue

kuanyul-qti:
### Description

- Change the output data type of the last node from int64/uint64 to int32/uint32, then a Cast op is added to convert the output tensor from int32/uint32 to int64/uint64.

### Motivation and Context

- Currently we only add the cast op (int32->int64) when the input name contains "_cast_int32", but the input name may not have this string because it can follow the data type of the previous node. In this case, the input data type of the op is int32, and the output data type of the op is int64, causing an error.

- Unit test -https://github.com/microsoft/onnxruntime/blob/4b1838b29608f5a19c0997971fd83bee6732ee56/onnxruntime/test/providers/qnn/reshape_expand_op_test.cc#L242

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/base_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/base_op_builder.cc
@@ -286,11 +286,10 @@ Status BaseOpBuilder::ProcessOutputs(QnnModelWrapper& qnn_model_wrapper,
     // Check if we need to add a cast node for int64
     bool needs_int64_cast = false;
     if (is_graph_output) {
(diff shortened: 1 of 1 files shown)


HectorSVC (approved):

HectorSVC:
could you merge the latest code from main branch. I have no idea why the MacOS CI always fails.

kuanyul-qti:
> could you merge the latest code from main branch. I have no idea why the MacOS CI always fails.

Hi @HectorSVC ,
I've rebased the branch. Could you help trigger the CI pipeline again?
Thanks.

PR #25747, opened 2025-08-14
[TRT RTX EP] EP context changes

thevishalagarwal:
* Implements `GetEPContextNodes()`
* Enables usage of `AddExternalInitializersFromFilesInMemory` for models that have to be communicated as byte stream but are larger than 2GB
* Add EP context unit tests for file, bytestreams and both embed modes

NOTE: For large models > 2GB, `embed_mode=0` must be used. `embed_mode=1` fails due to protobuf limitations

### DIFF
--- a/onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
+++ b/onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
@@ -1773,19 +1773,19 @@ NvExecutionProvider::GetCapability(const GraphViewer& graph,
   // If there are "EPContext" contrib op nodes, it means TRT EP can fetch the precompiled engine info from the node and
   // load the engine directly without having to go through the processes of graph proto reconstruction, calling TRT
   // parser and engine compilation. So, simply return subgraphs consists of single ep context nodes here.
-  size_t node_idx = 0;
-  if (GraphHasCtxNode(graph, node_idx)) {
-    int subgraph_idx = 0;
-    for (size_t node_idx : node_index) {
-      const auto& node = graph.GetNode(node_idx);
-      const bool is_context_node = node && !node->OpType().empty() && node->OpType() == EPCONTEXT_OP;
-      if (is_context_node) {
-        SubGraph_t supported_node_vector(std::make_pair(std::vector<size_t>{node_idx}, true));
-        std::unique_ptr<IndexedSubGraph> sub_graph = GetSubGraph(supported_node_vector, graph, model_hash, subgraph_idx++);
+  int subgraph_idx = 0;
+  for (size_t node_idx : node_index) {
+    const auto& node = graph.GetNode(node_idx);
+    const bool is_context_node = node && !node->OpType().empty() && node->OpType() == EPCONTEXT_OP;
+    if (is_context_node) {
+      SubGraph_t supported_node_vector(std::make_pair(std::vector<size_t>{node_idx}, true));
+      std::unique_ptr<IndexedSubGraph> sub_graph = GetSubGraph(supported_node_vector, graph, model_hash, subgraph_idx++);
 
-        result.push_back(ComputeCapability::Create(std::move(sub_graph)));
-      }
+      result.push_back(ComputeCapability::Create(std::move(sub_graph)));
     }
+  }
+  // return early if context nodes where found
+  if (!result.empty()) {
     return result;
   }
 
@@ -2061,9 +2061,11 @@ common::Status NvExecutionProvider::RefitEngine(std::string onnx_model_filename,
 
     // Extract weight information from the Refitter.
     int required_weights = refitter->getAllWeights(0, nullptr);
-    std::vector<char const*> refit_names(required_weights);
-    refitter->getAllWeights(required_weights, refit_names.data());
+    std::vector<char const*> refit_names_prealocated(required_weights);
+    refitter->getAllWeights(required_weights, refit_names_prealocated.data());
     LOGS_DEFAULT(VERBOSE) << "[NvTensorRTRTX EP] Refitter requires " << required_weights << " weights";
+    std::unordered_set<std::string> refit_names(std::make_move_iterator(refit_names_prealocated.begin()),
+                                      std::make_move_iterator(refit_names_prealocated.end()));
 
     // Vectors to keep track of data pointers.
     std::vector<std::string> names;
@@ -2094,8 +2096,7 @@ common::Status NvExecutionProvider::RefitEngine(std::string onnx_model_filename,
     for (int initializer_idx = 0; initializer_idx < allInitializers_byte_stream->size(); ++initializer_idx) {
       auto& proto = allInitializers_byte_stream->at(initializer_idx);
       auto& proto_name = proto.name();
-      bool weight_is_refittable = std::find(refit_names.begin(), refit_names.end(), proto_name) != refit_names.end();
-      if (weight_is_refittable) {
+      if (refit_names.find(proto_name) != refit_names.end()) {
         if (proto.has_data_location()) {
           if (proto.data_location() == TensorProto_DataLocation_EXTERNAL) {
             // Default values for reading into external_data blob.
--- a/onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
+++ b/onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
@@ -866,6 +866,14 @@ NvExecutionProvider::NvExecutionProvider(const NvExecutionProviderInfo& info)
                                        "When providing either 'trt_onnx_bytestream_size' or "
                                        "'trt_onnx_bytestream' both have to be provided"));
   }
+  onnx_external_data_bytestream_ = info.external_data_bytestream;
+  onnx_external_data_bytestream_size_ = info.external_data_bytestream_size;
+  if ((onnx_external_data_bytestream_ != nullptr && onnx_external_data_bytestream_size_ == 0) ||
+      (onnx_external_data_bytestream_ == nullptr && onnx_external_data_bytestream_size_ != 0)) {
+    ORT_THROW_IF_ERROR(ORT_MAKE_STATUS(ONNXRUNTIME, EP_FAIL,
+                                        "When providing either 'trt_external_data_bytestream_size' or "
+                                        "'trt_external_data_bytestream' both have to be provided"));
+  }
   detailed_build_log_ = info.detailed_build_log;
   dump_ep_context_model_ = info.dump_ep_context_model;
   ep_context_file_path_ = info.ep_context_file_path;
@@ -1029,6 +1037,7 @@ NvExecutionProvider::NvExecutionProvider(const NvExecutionProviderInfo& info)
                         << ", nv_ep_context_embed_mode: " << ep_context_embed_mode_
                         << ", nv_cache_prefix: " << cache_prefix_
                         << ", nv_onnx_model_bytestream_size_: " << onnx_model_bytestream_size_
+                        << ", nv_onnx_external_bytestream_size_: " << onnx_external_data_bytestream_size_
                         << ", nv_op_types_to_exclude: " << op_types_to_exclude_;
 }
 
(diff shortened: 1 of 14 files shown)


gedoensmax:
@chilo-ms we will need a review of this :) 
We disabled the option to refit for now (no weightless engines), since we want to make the interaction cleaner.

thevishalagarwal:
cc @gaugarg-nv @gedoensmax @ishwar-raut1

onnxruntime/test/providers/nv_tensorrt_rtx/nv_ep_context_test.cc
```diff
@@ -0,0 +1,198 @@
...
+  const auto& test_param = GetCompileParam();
+  SmallModelTest(test_param, false);
+}
+
+TEST_P(CompileApiTest, LargeModel) {
+  const auto& test_param = GetCompileParam();
+  // with embed mode == 1 the resulting file will be over the 2GB proto limit
+  if (test_param.embed_mode == 1) {
+    GTEST_SKIP();
+  }
```
gedoensmax, line 131: @adrianlizarraga we discused the mentioned limitation of context files via mail. What do you think of the approach of saving the context attribute on an EP context node as external initializer it it is above the set threshold. 
We need some way of communicating the engine as a byte stream for sensitive customers that want to encrypt their ONNX and their resulting EP context. Reloading the EP context would be possible with this change by also using the external initializer API.

gedoensmax:
@jywu-msft We are adding more unit tests that i believe will also help test the compile API etc in ORT. Can we resurface the topic of running NV EP in the official ORT CI ?

onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
```diff
@@ -1502,7 +1529,27 @@ SubGraphCollection_t NvExecutionProvider::GetSupportedList(SubGraphCollection_t
...
+            auto* tp = entry.second;
+            if (tp->has_raw_data()) {
+              userWeights.emplace_back(tp->name(), tp->raw_data());
+            } else if (utils::HasExternalDataInMemory(*tp)) {
+              std::unique_ptr<ONNX_NAMESPACE::TensorProto> full_init;
+              ORT_THROW_IF_ERROR(utils::GetTensorProtoWithDataIfInMemory(*tp, full_init));
+              userWeights.emplace_back(full_init->name(), full_init->raw_data());
+            }
+          }
+        }
```
gedoensmax: @yuslepukhin can you help us with a better recommendation to handle memory that has external data in memory ? We are copying all raw data into these `TensorrtUserWeights` since otherwise we were seeing some uninitialized memory reads in the case we do not copy but simply store a data pointer and its size.

Especially for data that is external and in memory I suppose there is a much nicer way to do this !

yuslepukhin: As far as the external data in memory is concerned (pointing to OrtValue) one only needs one call and that is utils::GetTensorProtoWithDatIfInMemory() (please, read the doc for it). If the resulting full_init is nullptr, means the data was not in memory, use the original TensorProto (the data can still be on disk though). If full_init is not nullptr, use that TensorProto with data now inlined. No version checks and other unncessary code.

gedoensmax: Ok understood, so simply calling `GetTensorProtoWithDataIfInMemory` does not copy the data string. Our `TensorrtUserWeights` constructor copies though since we were not able to rely on using a data pointer to the returned raw data. Am I missing something ? Is there a way to reference raw data without copying it out ?  
@kevinch-nv do you remember if we only saw issues with the "delayed" compile ? Meaning compile at first inference - we do no longer have that scenario with TRT RTX due to better dynamic shape support.

onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
```diff
@@ -1449,9 +1465,20 @@ SubGraphCollection_t NvExecutionProvider::GetSupportedList(SubGraphCollection_t
           SetGraphOuterScopeValuesAndInputs(graph_build, graph.GetGraph());
           SetAllGraphInputs(graph_build);
         }
-
-        ORT_ENFORCE(graph_build.Resolve().IsOK());
-
+        // for (auto& tensor:graph.GetAllInitializedTensors()) {
+        //   if (utils::HasExternalDataInMemory(*tensor.second)) {
+        //     std::unique_ptr<ONNX_NAMESPACE::TensorProto> full_init;
+        //     ORT_THROW_IF_ERROR(utils::GetTensorProtoWithDataIfInMemory(*tensor.second, full_init));
```
yuslepukhin: Also here, please, read the doc.

onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
```diff
@@ -1502,7 +1529,27 @@ SubGraphCollection_t NvExecutionProvider::GetSupportedList(SubGraphCollection_t
...
+
+        // save user provided external data in memory instead of writing to ModelProto
+        // needed for models > 2GB
+        std::vector<TensorrtUserWeights> userWeights;
+        auto allInitializers = graph_viewer->GetAllInitializedTensors();
+
+        if (use_external_data_initializer_) {
+          for (auto& entry : allInitializers) {
+            auto* tp = entry.second;
+            if (tp->has_raw_data()) {
```
yuslepukhin: >if (tp->has_raw_data()) { [](http://example.com/codeflow?start=12&length=25)

utils::HasRawData(*tp)

onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
yuslepukhin: I reread the code, and the issue here, is that you are trying to avoid copying the data when raw data is present,
since you are relying tensor_protos being around.

But raw data is not guaranteed to be present within the TensorProto.

`utils::GetTensorProtoWithDataIfInMemory` only deals with TensorProto that has external data in memory, leaving you to deal with other forms of data representation just as you did before, addressing this narrow scenario that we introduced.

Essentially, you have to deal with the following scenarios:

1. data is inline with TensorProto
  - raw data present.  Simply retain name -> data pointer as you are doing now and rely on data to be alive within the existing TensorProto.
  - data is not in a raw format. Do as you did before.
    Alternative:
        use TensorProto utils:: available via provider bridge. This takes care of all of the external data including in memory
        and data that is not in a raw format.
        Cost - data copy, you will have to find the place where to keep the returned raw data std::vector which does not have metadata, you will need to associate it with the original TensorProto.
        
        common::Status UnpackInitializerData(const ONNX_NAMESPACE::TensorProto& initializer,
                                         const std::filesystem::path& model_path,
                                         std::vector<uint8_t>& unpacked_tensor);


2. external data is on disk. Do as you did before or use UnpackInitializerData() as described above.
   It is preferred to mmap the data which I assume you would do. We usually memmap data on disk.

3. If you are dealing with the external data yourself and prefer only to code a special case of external data in memory then you have two alternatives:

   - use `utils::GetTensorProtoWithDataIfInMemory` that creates a new TensorProto with raw data present, otherwise,
     it does not do anything. Cost: data copy, you will need to find the place where to keep new TensorProto to keep raw_data alive so you can take a reference to it as you are doing now with raw data.

   - If utils::HasExternalDataInMememory() returns true, then the data is actually in the OrtValue inside the Graph instance.
     I am seeing that GetOrtValueInitializer() is not available via  a bridge. it is an oversight. I will make a PR today.
     Cost: you will have to find a place where to store OrtValue to keep the memory alive.
     Pros: no data copy, you can gain access to tensor data and store a reference to it as you are doing it today with raw data.

gedoensmax: >  Cost: you will have to find a place where to store OrtValue to keep the memory alive.

If these ORT Values are available during GetCapability and Compile the lifetime is sufficient. Only TRT EP (@chilo-ms) which compiles deferred if shapes are unknown would have to keep data in scope until the first inference is done.
I changed the `TensorrtUserWeights` class to not copy data in. From what i understand this is a safe usage of the raw data ? 
```
const InitializedTensorSet& allInitializers = graph_viewer->GetAllInitializedTensors();
std::vector<std::pair<const void*, const int64_t>> data;
for (auto& entry : allInitializers) {
      auto* tp = entry.second;
      if (utils::HasRawData(*tp)) {
        data.push_back(tp->raw_data().data(), tp->raw_data().size());
      }
}

// raw data here is alive safe to use since the graph_viewer lives in the same scope
data[0].first[0]
data[0].first[data[0].second - 1]
```
If i would be copying the tensor set like below the usage of raw data would be dependent it being in scope i assume since raw data is also copied from what i see. 
```
InitializedTensorSet allInitializers = graph_viewer->GetAllInitializedTensors();
```
Thanks for answering the questions with so much detail ! I want to make sure that we get this right, will try to better read the header docs in the future. 
Would it be preferred to leverage `TensorProtoToOrtValue` inside the EP instead of accessing the tensor proto's raw data ? Since this copies data as well and there is no `utils::GetRawData` i would stick to directly using `tp->raw_data()`.


I switched all raw data checks to use utility functions. `UnpackInitializerData` is not a good fit since we want to leverage external data inside TRT if it is not already loaded in memory.

yuslepukhin: 1. Yes, taking a view to the raw data of the existing `TensorProto` is safe.
2. OrtValues are available, but not all of them. Currently, we do not create OrtValues on top of the external data on disk, and we do not create it on top of the small initializers due to ONNX shape and type inference constraints. Thus, it is possible that an OrtValue is not present for a given initializer.
There are interfaces available in plug-in EP API that would return you OrtValues for any initializer and create them on the fly if necessary.

onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.h
```diff
@@ -356,6 +397,7 @@ class NvExecutionProvider : public IExecutionProvider {
   std::unordered_map<std::string, ShapeRangesMap> input_shape_ranges_;  // The profile shape ranges that the engine is built with
   std::unordered_map<std::string, std::vector<nvinfer1::IOptimizationProfile*>> profiles_;
   std::unordered_map<std::string, DDSOutputAllocatorMap> dds_output_allocator_maps_;
+  std::unordered_map<std::string, std::unique_ptr<std::vector<TensorrtUserWeights>>> weights_;  // User provided weights
```
chilo-ms: is this ever being used?

onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
gedoensmax: Ok I made use of your added API and rebased this branch. Thanks.

onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.h
gedoensmax: No - removed.

gedoensmax:
@chilo-ms Since i resolve the cases where weights are unnecessarily copied based on Dimitri's comments this should be ready to merge.

onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
```diff
@@ -1502,7 +1525,37 @@ SubGraphCollection_t NvExecutionProvider::GetSupportedList(SubGraphCollection_t
...
+            if (utils::HasRawData(*tp)) {
+              userWeights.emplace_back(TensorrtUserWeights(tp->name(), tp->raw_data().data(), tp->raw_data().size()));
+            } else if (graph_viewer->GetOrtValueInitializer(tp->name(), initializer_value)) {
+              // the initializer was marked as external data by the ORT graph at load time since it was provided in memory
+              size_t size = 0;
+              const void* ptr = nullptr;
+              c_api.GetTensorSizeInBytes(&initializer_value, &size);
+              c_api.GetTensorData(&initializer_value, &ptr);
+              userWeights.emplace_back(tp->name(), ptr, size);
+            } else if (utils::HasExternalDataInMemory(*tp)) {
```
chilo-ms, line 1548: Will this `else if (utils::HasExternalDataInMemory(*tp))` ever be executed?

> If utils::HasExternalDataInMememory() returns true, then the data is actually in the OrtValue inside the Graph instance.

As per other long discussion thread here in the PR with Dmitri, it seems to me  if `HasExternalDataInMemory` is true implies `graph_viewer->GetOrtValueInitializer()` returning true.

(update) Discussed offline. the "else if" here is redundant but won't cause issue, and we can have follow-up PR to refactor this later.

PR #25749, opened 2025-08-14
Add support for generating and validating compiled model compatibility info

adrastogi:
### Description
This pull request introduces a new mechanism for validating compiled model compatibility with execution providers (EPs) in ONNX Runtime. It adds infrastructure for EPs to generate and store compatibility information in model metadata, and for the runtime to enforce compatibility checks during session initialization. 

### Motivation and Context

The APIs proposed in this PR address two requirements:

1. Apps that have an already pre-compiled model on device need a way to determine if the pre-compiled app is still valid (given the EPs / drivers / etc. on the system).
2. Apps may have many different pre-compiled versions of a model stored on a remote server, and want to figure out which of those models they should download for the device where they are running.

### Testing
Validated that the new suite of tests passes cleanly. 
Created a private build of this ORT and the AMD Vitis EP. I stepped through the core logic (the EP doesn't have this support wired up as yet so there is no compatibility info written out) and for regression purposes, confirmed I could compile and run inferences through ResNet.


onnxruntime/core/session/utils.cc
```diff
@@ -206,6 +207,110 @@ OrtStatus* CreateSessionAndLoadModel(_In_ const OrtSessionOptions* options,
...
+    if (metadata_it != custom_metadata.end()) {
+      const std::string& compatibility_info = metadata_it->second;
+
+      // Get the actual EP instance to call validation
+      const IExecutionProvider* ep = execution_providers.Get(ep_type);
+
+      if (ep != nullptr) {
+        // Call the EP's validation method (virtual method with default implementation)
+        OrtCompiledModelCompatibility compatibility_status;
+        Status validation_result = ep->ValidateCompiledModelCompatibilityInfo(
```
adrianlizarraga, line 263: Looks like `IExecutionProvider::ValidateCompiledModelCompatibilityInfo` is not defined for minimal builds according to this CI build error: https://github.com/microsoft/onnxruntime/actions/runs/16967226629/job/48094281125?pr=25749#step:7:471

```
/onnxruntime_src/onnxruntime/core/session/utils.cc: In function ‘onnxruntime::common::Status ValidateCompiledModelCompatibility(onnxruntime::InferenceSession&)’:
/onnxruntime_src/onnxruntime/core/session/utils.cc:257:40: error: ‘const class onnxruntime::IExecutionProvider’ has no member named ‘ValidateCompiledModelCompatibilityInfo’
  257 |         Status validation_result = ep->ValidateCompiledModelCompatibilityInfo(
      |                                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Do we need to wrap this function (and call site) with `#if !defined(ORT_MINIMAL_BUILD)`?

onnxruntime/core/framework/graph_partitioner.cc
```diff
@@ -865,6 +866,36 @@ static Status CreateEpContextModel(const ExecutionProviders& execution_providers
...
+  {
+    const GraphViewer graph_viewer(graph);
+    for (const auto& ep : execution_providers) {
+      try {
+        // Generate the compatibility string for this EP
+        std::string compatibility_string = ep->GetCompiledModelCompatibilityInfo(graph_viewer);
+        if (!compatibility_string.empty()) {
+          // Create a unique key for this EP's compatibility info
+          // Use format: "ep_compatibility_info.<EP_TYPE>"
+          // TODO: Will this guarantee uniqueness across all EPs?
```
adrianlizarraga: All EPs in a session must have a unique "Type()" so this will be unique for the generated model. #Pending

onnxruntime/core/framework/graph_partitioner.cc
```diff
@@ -865,6 +866,36 @@ static Status CreateEpContextModel(const ExecutionProviders& execution_providers
...
+          auto& model_metadata = ep_context_model.MetaData();
+          auto [it, was_inserted] =
+              model_metadata.insert_or_assign(metadata_key, compatibility_string);
+          if (!was_inserted) {
+            LOGS(logger, WARNING) << "Overwriting existing EP compatibility info for key: " << metadata_key << " (EP: " << ep->Type() << ")";
+          }
+          LOGS(logger, INFO) << "Added EP compatibility info for " << ep->Type() << " with key: " << metadata_key;
+
+        } else {
+          LOGS(logger, WARNING) << "EP " << ep->Type() << " returned empty compatibility string";
```
adrianlizarraga: nit: because an empty compatibility string is the default behavior, perhaps we don't need to log this? #Pending

onnxruntime/core/framework/graph_partitioner.cc
```diff
@@ -865,6 +866,36 @@ static Status CreateEpContextModel(const ExecutionProviders& execution_providers
...
+          // Use format: "ep_compatibility_info.<EP_TYPE>"
+          // TODO: Will this guarantee uniqueness across all EPs?
+          std::string metadata_key = std::string(kOrtModelMetadata_EpCompatibilityInfoPrefix) + ep->Type();
+          auto& model_metadata = ep_context_model.MetaData();
+          auto [it, was_inserted] =
+              model_metadata.insert_or_assign(metadata_key, compatibility_string);
+          if (!was_inserted) {
+            LOGS(logger, WARNING) << "Overwriting existing EP compatibility info for key: " << metadata_key << " (EP: " << ep->Type() << ")";
+          }
+          LOGS(logger, INFO) << "Added EP compatibility info for " << ep->Type() << " with key: " << metadata_key;
```
adrianlizarraga: nit: maybe `VERBOSE` would be more appropriate? #Pending

onnxruntime/core/framework/graph_partitioner.cc
```diff
@@ -865,6 +866,36 @@ static Status CreateEpContextModel(const ExecutionProviders& execution_providers
   auto& ep_graph = ep_context_model.MainGraph();
   ep_graph.SetDescription(graph.Description());
 
+  // Generate EP compatibility strings for OrtEp types and add to model metadata
+  {
+    const GraphViewer graph_viewer(graph);
+    for (const auto& ep : execution_providers) {
+      try {
+        // Generate the compatibility string for this EP
+        std::string compatibility_string = ep->GetCompiledModelCompatibilityInfo(graph_viewer);
```
adrianlizarraga: I'm not sure how easily an EP can utilize the model that it receives in this call to GetCompiledModelCompatibilityInfo(). This model is in a state between the original uncompiled model and the final EPContext model. Specifically, the nodes assigned to a compiling EP are currently fused into a generic "fused node" with a computation function (i.e., it is not an EPContext node yet).

Would it make more sense to have this logic at the end of `CreateEpContextModel`, once all generic "fused nodes" have been replaced with proper EPContext nodes?
 #Pending

onnxruntime/core/graph/model.cc
```diff
@@ -377,6 +381,15 @@ ModelProto Model::ToProto() const {
   // out dense duplicates of sparse initializers and leave the original
   // proto intact.
   ModelProto result(model_proto_);
+
+  // Sync current model_metadata_ back to protobuf metadata_props
+  result.clear_metadata_props();
```
adrianlizarraga, line 386: Perhaps we also need to do this in ` Model::ToGraphProtoWithExternalInitializers` #Pending

onnxruntime/core/session/plugin_ep/ep_plugin_provider_interfaces.cc
```diff
@@ -644,4 +644,31 @@ void PluginExecutionProvider::RegisterStreamHandlers(IStreamCommandHandleRegistr
...
+  // Call EP plugin's OrtEp::GenerateCompiledModelCompatibilityInfo() function.
+  std::string compatibility_info_string;
+  compatibility_info_string = ort_ep_->GetCompiledModelCompatibilityInfo(ort_ep_.get(), ep_graph.get());
+  return compatibility_info_string;
+}
+
+Status PluginExecutionProvider::ValidateCompiledModelCompatibilityInfo(const std::string& compatibility_info,
+                                                                       OrtCompiledModelCompatibility& model_compatibility) const {
+  // Delegate to the EP factory's validation method
+  ORT_RETURN_IF_ERROR(ToStatusAndRelease(ep_factory_.ValidateCompiledModelCompatibilityInfo(&ep_factory_,
```
adrianlizarraga, line 672: Do we need to check if `ep_factory_.ValidateCompiledModelCompatibilityInfo` is null, and if it is, call a default implementation? I wouldn't think that an `OrtEpFactory` is required to implement this function.

adrastogi: I _thought_ `PluginExecutionProvider` instances must take an `OrtEpFactory` reference as part of construction and therefore must be valid- but I might be mistaken here. Is there a bug in my reasoning?

onnxruntime/core/framework/graph_partitioner.cc
adrastogi: Great callout. I shifted this logic to right before the model is saved (either to file or buffer).

onnxruntime/core/framework/graph_partitioner.cc
adrastogi: Thanks for helping confirm this! I added a comment to the code.

onnxruntime/core/session/plugin_ep/ep_plugin_provider_interfaces.cc
adrianlizarraga: A `PluginExecutionProvider` instance must take a non-null `OrtEpFactory`. However, the `OrtEpFactory` may not have an implementation of `ValidateCompiledModelCompatibilityInfo`.

So, the code could look something like this:

```c++
  if (ep_factory->ValidateCompiledModelCompatibilityInfo == nullptr) {
    // Plugin EP factory did not provide an implementation of this function, so we call a default implementation.
    return Base::ValidateCompiledModelCompatibilityInfo(compatibility_info, model_compatibility);
  }

  // Delegate to the EP factory's validation method
  ORT_RETURN_IF_ERROR(ToStatusAndRelease(ep_factory_.ValidateCompiledModelCompatibilityInfo(&ep_factory_,
                                                                                            compatibility_info.c_str(),
                                                                                            &model_compatibility)));
  return Status::OK();
```

adrastogi: Ahh, got it, similar to the pattern you'd introduced for the Get method. Sorry, I wasn't connecting the dots here- thanks for the explanation.

thevishalagarwal:
Hi @adrastogi / @adrianlizarraga - is [GetCompiledModelCompatibility](https://github.com/microsoft/onnxruntime/blob/cb4d4af2ce20d9ff45ce04394fcf6c623a5b64a7/include/onnxruntime/core/session/onnxruntime_ep_c_api.h#L739) API going to be available for non-ABI interfaces as well? Do you have an ETA for this draft PR to merge?

adrastogi:
Hey @thevishalagarwal- yes, the implementation supports non-ABI interfaces as well. We're trying to get this in as soon as we can (pending a few loose ends that we're chasing down). 

---
In reply to: [3209754229](https://github.com/microsoft/onnxruntime/pull/25749#issuecomment-3209754229) [](http://example.com/codeflow?ancestors=3209754229)

onnxruntime/core/framework/graph_partitioner.cc
```diff
@@ -909,6 +910,33 @@ static Status CreateEpContextModel(const ExecutionProviders& execution_providers
     }
   }
 
+  // Generate EP compatibility strings for OrtEp types and add to model metadata
+  // At this point, the graph has been populated with all the EPContext nodes
+  {
+    const GraphViewer graph_viewer(graph);
```
adrianlizarraga: I believe this needs to be `ep_graph` instead of `graph`. The `ep_graph` is the one that has EPContext nodes.

Also, we should probably call `ORT_RETURN_IF_ERROR(ep_graph.Resolve());` before this line to ensure that any unused initializers are removed. Otherwise, when PluginExecutionProvider converts graph_viewer to a OrtGraph, it will copy many unused initializers.

```c++
{
  ORT_RETURN_IF_ERROR(ep_graph.Resolve());
  const GraphViewer graph_viewer(ep_graph);
  // ...
}
```

onnxruntime/core/session/utils.cc
```diff
@@ -206,6 +207,112 @@ OrtStatus* CreateSessionAndLoadModel(_In_ const OrtSessionOptions* options,
...
+}
+
+static Status ValidateCompiledModelCompatibility(InferenceSession& sess) {
+  // Get model metadata
+  auto [status, model_metadata] = sess.GetModelMetadata();
+  if (!status.IsOK() || !model_metadata) {
+    // No metadata available, skip validation
+    return Status::OK();
+  }
+
```
adrianlizarraga, line 233: nit: Should we have an early return here if the model_metadata->custom_metadata_map is empty?

PR #25759, opened 2025-08-15
[VitisAI] Add a hook to Vitis AI EP when an EP factory is created

cferry-AMD:
### Description
This PR adds a `vitisai_on_ep_factory_created` hook after a factory is created, to make the Vitis AI EP library aware of which options were given by the user.

### Motivation and Context
The Vitis AI EP library needs to process user-given options before an EP is created. This is permitted by the introduction of this hook.


cferry-AMD:

cferry-AMD:
Hi @jywu-msft and @wcy123, do you know who is best suited to review this?

cferry-AMD:
PR not needed any longer.

PR #25763, opened 2025-08-15
[webgpu] Add support for Q2 in matmulnbits

sushraja-msft:
### Description
This change adds support for Q2 quantized matmulnbits, in webgpu.

### Motivation and Context
An alternate way to support bitnets is through adding support for lower bits in matmulnbits, this reuses our shaders and is more maintainable than a separate op. The model size grows a bit however for a 2B parameter model using 1.58bpw vs 2bpw the size difference is just 100MB. The simpler dequantization also improves perf, on an Intel XE matmul looks to be 20% faster using q2 weights vs q4 weights for the same matrix dimensions.


onnxruntime/contrib_ops/webgpu/quantization/matmul_nbits.cc
```diff
@@ -127,8 +128,11 @@ Status MatMulNBits::ComputeInternal(onnxruntime::webgpu::ComputeContext& context
   const uint32_t block_size = onnxruntime::narrow<uint32_t>(block_size_);
   const uint32_t nbits = onnxruntime::narrow<uint32_t>(bits_);
 
-  const uint32_t n_blocks_per_col = (K + block_size - 1) / block_size;
-  const uint32_t blob_size = (block_size / 8) * nbits;
+  // Special case matrix used by bitnets where there is a single scale for the entire
+  const bool single_scale_weights = (block_size == K * N);
+  const uint32_t block_size_per_col = single_scale_weights ? K : block_size;
+  const uint32_t n_blocks_per_col = (K + block_size_per_col - 1) / block_size_per_col;
+  const uint32_t blob_size = (block_size_per_col / 8) * nbits;
```
qjia7, line 136: Please also update the `zero_blocks_per_col`'s value in line 142. The previous computing is not suitable for nbits = 2.

onnxruntime/contrib_ops/webgpu/quantization/dp4a_matmul_small_m.wgsl.template
```diff
@@ -96,6 +102,11 @@ $MAIN {
                 inter_results[row_offset + local_row][local_col] += SDP8AI(own_a, own_b, own_a1, own_b1, own_scale_a * own_scale_b);
 #endif
 
+#elif n_bits == 2
```
qjia7, line 114: Maybe move the assignment of zero and own_scalue_b in line 89-90 out of the loop like you did in matmul_nbits.wgsl.template when it's a single scale weights?

onnxruntime/contrib_ops/webgpu/quantization/matmul_nbits.cc
sushraja-msft: I have not had a chance to look at zero points support for q2. Let me add an enforce block to block that for now.

qjia7 (dismissed):
LGTM, thanks.

onnxruntime/test/contrib_ops/matmul_2bits_test.cc
```diff
@@ -244,22 +244,47 @@ void TestMatMul2BitsTyped(float abs_error = 0.1f, float rel_error = 0.02f) {
...
+TYPED_TEST_SUITE(MatMulNBits, TestTypes);
+
+TYPED_TEST(MatMulNBits, Float32_2Bits_Accuracy0) {
+  TestMatMul2BitsTyped<float, TestFixture::batch_size, TestFixture::M, TestFixture::N, TestFixture::K, 0>();
+}
+
+TYPED_TEST(MatMulNBits, Float32_2Bits_Accuracy4) {
+  TestMatMul2BitsTyped<float, TestFixture::batch_size, TestFixture::M, TestFixture::N, TestFixture::K, 4>();
 }
+
```
qjia7, line 290: Sorry, forgot one thing. You should add webgpu ep to the execution_providers in L198 like https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/test/contrib_ops/matmul_4bits_test.cc#L540-L543 did. Otherwise, these tests are not tested on webgpu ep. In this file, it also tests the zero points in L225 and L239. You can use them to test the zero points path or skip those tests for webgpu.

guschmue (dismissed):

onnxruntime/test/contrib_ops/matmul_2bits_test.cc
sushraja-msft: ah thanks for the tip !

guschmue (approved):

bil-ash:
@sushraja-msft The python code available at can be used for quantizing onnx models to q2-not just bitnets but also other mdels with fp16 weights. However, the files are not publicly available. I had requested for access from the huggingface profile "bh4" but it is still pending. Please grant me access so that I can quantize other fp16 models to q2 following your quantization script.

PR #20362, opened 2024-04-18
[CPU EP] Int4 support for QuantizeLinear, DequantizeLinear, and Transpose

adrianlizarraga:
### Description
- 4-bit QuantizeLinear(21). **Blocked quantization still missing (i.e., do not support the new `block_size` attribute)**
- 4-bit DequantizeLinear(21). **Blocked dequantization still missing (i.e., do not support the new `block_size` attribute)**
- 4-bit Transpose(21).
- Update quantization tool with int4 types.
- Disable QDQ fusions for 4-bit types. See: https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/optimizer/qdq_transformer/selectors_actions/qdq_selector_action_transformer.cc
- MLAS 4-bit quantization kernels for intel, neon, powerpc.

##### Notes
To calculate a tensor's storage size, we normally get the number of elements from the shape (i.e., `tensor_shape.Size()`) and multiply by the size of a single element. This does not directly work for sub-byte elements like int4 as each element in a `Tensor<Int4x2>` stores **two** packed int4 elements in a byte. The `Tensor:: CalculateTensorStorageSize` should be called to perform the correct calculation for any tensor element type. 

### Motivation and Context
ONNX 1.16 added the int4 and uint4 types. This initial PR adds the int4 type to ORT and adds int4 implementations for the Quant, Dequant, and Transpose ops on CPU EP. We still need to add int4 support for many ops and execution providers. See the ONNX 1.16 release notes: https://github.com/onnx/onnx/releases.

PR #14731, opened 2023-02-17
[WIP] First draft to introduce float 8 types

xadupre:
### Description
The PR implements FloatE4M3, FloatE5M2 as described in PR https://github.com/onnx/onnx/pull/4805. It only works with CUDA>=11.8. First draft ensures it works. The implementation is not yet optimal.

PR #25767, opened 2025-08-16
Support fp4 type in ORT 

hariharans29:
### Description
https://github.com/onnx/onnx/pull/6318 and https://github.com/onnx/onnx/pull/6283 added FP4 support to ONNX. This change introduces the FP4 type in ORT and adds type support to one relevant operator (`Cast`) as a proof-of-concept for the type integration into ORT. More op support will be added on a need-basis.

This change took inspiration from the following PRs:

https://github.com/microsoft/onnxruntime/pull/14731
https://github.com/microsoft/onnxruntime/pull/22228
https://github.com/microsoft/onnxruntime/pull/20362 

Some notes:

1) Only `tensor` type gets support for FP4 initially. Secondary types like `seq(tensor)`, `sparse_tensor`, `optional` do not get support (so as to not introduce unnecessary bloat to the framework without a solid use-case)

2) Flatbuffer related files receive no updates in this PR

### Motivation and Context
Be able to run FP4 models with ORT


cmake/CMakeLists.txt
```diff
@@ -155,6 +155,7 @@ option(onnxruntime_DISABLE_ML_OPS "Disable traditional ML ops" OFF)
 option(onnxruntime_DISABLE_SPARSE_TENSORS "Disable sparse tensors data types" OFF)
 option(onnxruntime_DISABLE_OPTIONAL_TYPE "Disable optional type" OFF)
 option(onnxruntime_DISABLE_FLOAT8_TYPES "Disable float 8 types" OFF)
+option(onnxruntime_DISABLE_FLOAT_TYPES "Disable float 4 types" OFF)
```
tianleiwu: onnxruntime_DISABLE_FLOAT4_TYPES?

onnxruntime/test/framework/float_4_test.cc
```diff
@@ -0,0 +1,155 @@
...
+// Licensed under the MIT License.
+
+#if !defined(DISABLE_FLOAT4_TYPES)
+
+#include <vector>
+#include <map>
+
+#ifdef USE_CUDA
+// Needed for CUDA_VERSION check in float4.h
+#include <cuda.h>
```
hariharans29, line 11: Including this on CUDA enabled builds will ensure that CUDA based converion routines are used in CUDA12.8+ builds as CUDA_VERSION (from cuda.h) is used as a gating check in float4.h

onnxruntime/test/framework/float_8_test.cc
```diff
@@ -5,6 +5,11 @@
 
 #include <vector>
 
+#ifdef USE_CUDA
+// Needed for CUDA_VERSION check in float8.h
+#include <cuda.h>
```
hariharans29: Same here as in float_4_test.cc - We want to include the cuda header on CUDA enabled builds to bring in CUDA_VERSION which is a gating check to use CUDA conversion routines in float_8.h

include/onnxruntime/core/framework/float4.h
```diff
@@ -0,0 +1,297 @@
...
+    if (((float_bits & exponent_bitmask) == exponent_bitmask) && ((float_bits & mantissa_bitmask) == zero)) {
+      return ((res | 0x07) << shift);
+    }
+
+    float f_abs = std::abs(f);
+    if (f_abs > 0.25 && f_abs < 0.75) {
+      res |= 0x01;
+    } else if (f_abs >= 0.75 && f_abs < 1.25) {
+      res |= 0x02;
+    } else if (f_abs > 1.25 && f_abs < 1.75) {
```
tianleiwu, line 118: what if the f_abs is middle point 1.25 here. Shall we use f_abs <= 1.25 instead of f_abs < 1.25 in the above.

include/onnxruntime/core/framework/float4.h
```diff
@@ -0,0 +1,297 @@
...
+
+    float sign = 1.f;
+    if (bits_shifted & sign_bitmask) {
+      sign = -1.f;
+    }
+
+    float exponent = static_cast<float>((bits_shifted & exponent_bitmask) >> 1);
+    float mantissa = static_cast<float>(bits_shifted & mantissa_bitmask);
+
+    return (exponent == 0) ? (sign * (mantissa / 2.f)) : (sign * (1.f + mantissa / 2.f) * std::pow(2.f, (exponent - 1)));
```
tianleiwu: For perfromance, use ldexpf or using bit shift instead of pow here.

onnxruntime/test/framework/float_4_test.cc
```diff
@@ -0,0 +1,155 @@
...
+  EXPECT_EQ(pair.first, 6.f);
+  EXPECT_EQ(pair.second, -0.f);
+
+  pair = Float4E2M1x2(0x7F, Float4E2M1x2::FromBits()).ToFloat2();
+  EXPECT_EQ(pair.first, -6.f);
+  EXPECT_EQ(pair.second, 6.f);
+
+  // Bit representation test
+  uint8_t bits = Float4E2M1x2(-6.f, 6.f).ToBits();
+  // First nibble is the second value and the second nibble is the first value
```
tianleiwu: The comment is slightly confusing. 

From this example, the high nibble stores the first value -6. and the low nibble stores the second value 6.

I saw in onnx spec (https://onnx.ai/onnx/technical/float4.html), The first element is stored in the 4 LSB: pack(x,y): y << 4 | x & 0x0F

onnxruntime/test/framework/float_8_test.cc
hariharans29: Pulling this change out as it fails some CUDA float8 tests. It looks like there is some discrepancy between the CPU conversion implementation and the CUDA conversion routines. To be debugged separately. Since, it is unrelated to this PR - I am pulling out this change. CC: @xadupre

include/onnxruntime/core/framework/float4.h
hariharans29: Thanks for catching. I had a typo here. Fixed and added a test for 1.25f.

include/onnxruntime/core/framework/float4.h
hariharans29: Sure, used bit manipulation. But I don't expect the Float4E2M1x2 instance creation to be in the critical perf path - I expect them to be only used to create `OrtValue` initializers and the kernels are the one that are going to process the underlying data.

onnxruntime/test/framework/float_4_test.cc
hariharans29: I think my comment is incorrect but the check is correct. Fixed.

"High nibble stores the second value +6 and low nibble stores the first value -6"

tianleiwu (approved):

PR #25770, opened 2025-08-17
Fix static mutex destruction order issues causing crashes at program termination

un33k:
## Description
  This PR fixes crashes that occur during program termination due to static mutex destruction order issues. When static
  mutexes are destroyed while other static objects still reference them, it causes undefined behavior and crashes.

  ## Motivation and Context
  Static destruction order in C++ is undefined across translation units. This causes crashes when:
  - Logging infrastructure attempts to use mutexes during cleanup
  - Multiple static objects have interdependencies
  - Cleanup order varies between platforms/configurations

  ## Changes
  The fix uses heap-allocated mutexes (intentionally leaked) for critical static mutexes to ensure they remain valid
  throughout program lifetime:

  1. `DefaultLoggerMutex()` in `logging.cc` - Used by logging infrastructure
  2. `GetOrtEnvMutex()` in `ort_env.cc` - Used by ORT environment management
  3. `GetActiveSessionsMutex()` in `inference_session.cc` - Windows-specific session tracking

  ## Technical Details
  - Uses `new std::mutex()` without corresponding `delete` (intentional leak)
  - The leaked memory is negligible (24-40 bytes per mutex)
  - This is a common C++ pattern for avoiding static destruction issues
  - Similar approach used in many production systems (LLVM, Chrome, etc.)

  ## Testing
  - Tested on macOS (ARM64)
  - Existing test suite passes
  - Additional testing on Windows and Linux would be appreciated

  ## Note
  This issue was discovered while working on the Rust-based TTS library which uses ONNX Runtime. The crashes were occurring during cleanup when the library was unloaded.

### DIFF
--- a/onnxruntime/core/common/logging/logging.cc
+++ b/onnxruntime/core/common/logging/logging.cc
@@ -65,8 +65,11 @@ LoggingManager* LoggingManager::GetDefaultInstance() {
 #endif
 
 static std::mutex& DefaultLoggerMutex() noexcept {
-  static std::mutex mutex;
-  return mutex;
+  // Fix for static destruction order issue:
+  // Use a heap-allocated mutex that we intentionally leak to avoid
+  // destruction order problems during program termination
+  static std::mutex* mutex = new std::mutex();
+  return *mutex;
 }
 
 Logger* LoggingManager::s_default_logger_ = nullptr;
--- a/onnxruntime/core/session/inference_session.cc
+++ b/onnxruntime/core/session/inference_session.cc
@@ -253,7 +253,12 @@ Status GetMinimalBuildOptimizationHandling(
 std::atomic<uint32_t> InferenceSession::global_session_id_{1};
 std::map<uint32_t, InferenceSession*> InferenceSession::active_sessions_;
 #ifdef _WIN32
-std::mutex InferenceSession::active_sessions_mutex_;  // Protects access to active_sessions_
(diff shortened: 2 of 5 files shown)


un33k:
@microsoft-github-policy-service agree [company="{your company}"]

> @un33k please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> ```
> 
> > Options:
> > 
> > 
> > ```
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > 
> > ```
> > ```
> 

@microsoft-github-policy-service agree [company="{Neekware Inc.}"]

un33k:
> @un33k the command you issued was incorrect. Please try again.
> 
> Examples are:
> 
> ```
> @microsoft-github-policy-service agree
> ```
> 
> and
> 
> ```
> @microsoft-github-policy-service agree company="your company"
> ```

@microsoft-github-policy-service agree company="Neekware Inc."

a-akoval:
There's no need to leak as we can use the placement `new`:
```cpp
static std::mutex& DefaultLoggerMutex() noexcept {
    alignas(std::mutex) static std::byte storage[sizeof(std::mutex)];
    static std::mutex* const m = ::new (static_cast<void*>(storage)) std::mutex();
    return *m;
}
```

PR #25779, opened 2025-08-18
[VitisAI] Ability to dynamically add kernels to kernel registry

cferry-AMD:
### Description
This PR adds the ability to add operations into the KernelRegistry after VitisAI EP library initialization. It also currently calls it after EP factory creation.

This PR does:
* a refactoring of the `CreateKernelInfo` creation (puts it into a dedicated function)
* add a `refresh_kernel_registry` function and rename `create_kernel_registry` into `initialize_kernel_registry`,
* maintain a list `s_vitisaiep_registered_ops` of ops known to be in the KernelRegistry, to prevent duplicating already registered ops when `refresh_kernel_registry` is called.

### Motivation and Context
The VitisAI library needs to register ops at multiple points past its initialization. This facility lets it do so, and updates the ops once the EP factory exists (before the EPs are actually created).

Sits on top of #25759.


cferry-AMD:
PR not needed any longer.

PR #25783, opened 2025-08-18
[webgpu] Move MatMulNBits to templates

sushraja-msft:
### Description
This change moves the regular matmulnbits shader to templates.


qjia7 (approved):
LGTM, thanks!

PR #25787, opened 2025-08-19
Add cuda graph implementation for NV TRT RTX EP

umangb-09:
### Description
This change adds CUDA Graph support to the NV TensorRT RTX Execution Provider (EP).

### Motivation and Context
Integrating CUDA Graphs into the NV TRT RTX EP provides:
Lower latency by minimizing per-kernel launch overhead.
Better throughput for repeated inference runs.
Improved efficiency on GPUs with high kernel launches overhead sensitivity.


umangb-09:
@jywu-msft can you mark this for 1.23 release?

jywu-mysoft:
please fix ClangFormat errors and also resolve conflicts

umangb-09:
> please fix ClangFormat errors and also resolve conflicts

FIxed

onnxruntime/core/providers/cuda/cuda_graph.cc
```diff
@@ -85,6 +85,21 @@ Status CUDAGraphManager::Replay(CudaGraphAnnotation_t cuda_graph_annotation_id)
   return Status::OK();
 }
 
+Status CUDAGraphManager::Replay(CudaGraphAnnotation_t cuda_graph_annotation_id, bool sync_status_flag) {
+  // Although this function is not thread safe, the lock is not needed here because
```
chilo-ms, line 76: Could we extend the existing `Replay` to have a default parameter rather than duplicating one here?
```c++
CUDAGraphManager::Replay(CudaGraphAnnotation_t cuda_graph_annotation_id, bool sync_status_flag = true)
```

onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
```diff
@@ -1072,41 +1164,82 @@ NvExecutionProvider::~NvExecutionProvider() {
   }
 }
 
+void NvExecutionProvider::HandleCudaGraphStart(cudaStream_t stream, bool require_io_binding,
+                                               CudaGraphAnnotation_t cuda_graph_annotation_id, bool& graph_replay_on_this_run, bool& should_start_capture) {
+  graph_replay_on_this_run = false;
```
chilo-ms, line 1169: Any reason that the CUDA graph capture and reply logic is being implemented inside the EP, instead of in the session? 
(see [here](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/session/inference_session.cc#L2321) and the replay [here](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/session/inference_session.cc#L2969))

Other questions are:

- CUDA graph doesn't support the case where the graph has control flow op due to it's possible that other part of the subgraph in the control flow op is not being captured and might have issue if running that subgraph during replay.
It seems there is no check here.
- Have you tried the case where the graph are partitioned and run by TRT RTX EP and CUDA EP?

onnxruntime/core/providers/cuda/cuda_graph.cc
umangb-09: We can, this was the original solution, but this will require to change in CUDA EP as well. So, we skipped it as part of this push. @gaugarg-nv @gedoensmax any opinion on this?

gaugarg-nv: I'm fine with @chilo-ms 's suggestion.

onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
umangb-09: Any reason that the CUDA graph capture and reply logic is being implemented inside the EP, instead of in the session?
--> Yes, we had enabled this in the first go but seems like the TRT and NV EP has some issues to use this as it is, specifically involving how enqueue calls are made. This is already in our notice and is supposed to be fixed soon. For now, this was the only way we were able to implement by giving this control to EP instead of ORT framework.

CUDA graph doesn't support the case where the graph has control flow op due to it's possible that other part of the subgraph in the control flow op is not being captured and might have issue if running that subgraph during replay.
--> Yes, that's correct. This is why we have not enabled cuda graph support by default and we believe the user/application developer have to be aware of where and when to enable this feature. We are still working on this along with other cases and testing which will be handled or at least shared with ERROR message. 

Have you tried the case where the graph is partitioned and run by TRT RTX EP and CUDA EP
--> No, as mentioned in above step as well many cases CPU fallback etc are not yet fully tested and supported. We are working on this as well to share a documentation on when and how to support CG.

onnxruntime/core/providers/cuda/cuda_graph.h
```diff
@@ -38,7 +38,7 @@ struct CUDAGraphManager {
   void SetStream(cudaStream_t stream);
   void CaptureBegin(CudaGraphAnnotation_t cuda_graph_annotation_id);
   void CaptureEnd(CudaGraphAnnotation_t cuda_graph_annotation_id);
-  Status Replay(CudaGraphAnnotation_t cuda_graph_annotation_id);
+  Status Replay(CudaGraphAnnotation_t cuda_graph_annotation_id, bool sync_status_flag);
```
gaugarg-nv: Does CUDE EP compile with this change? Default value should always be in the header and not in cpp.

umangb-09: Have not compiled CUDA EP. It has some other issues as well AFAIK. Will keep it for compilation.

chilo-ms:
Please help resolve the conflicts

onnxruntime/core/providers/cuda/cuda_graph.h
chilo-ms: What's the issue for putting default value in header file?
The CIs should test building CUDA EP, but it's better to test it locally though. Since this PR touches the file that CUDA EP uses and we don't want it to break anything as this is going to be merged in ORT 1.23.

onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
chilo-ms: We had an internal discussion that the cuda graph logic should stay in the EP, the logic in session is the legacy approach and we will refactor it. So, let's stick to what you did to put the logic in the EP.

onnxruntime/core/providers/cuda/cuda_graph.h
umangb-09: Fixed

chilo-ms:
open and reopen for the specific CI to run

umangb-09:

snnn:

Issue #25778, opened 2025-08-18
[Build] QNN EP asserts in X64/Debug tests


a-jmist:
### Describe the issue

These tests all assert in onnxruntime_test_all when built for rel-1.23.0 branch, with config: Debug / QNN EP / X64:
- QnnCPUBackendTests.AveragePool_3D_AsGlobal
- QnnCPUBackendTests.MaxPool_3D
- QnnCPUBackendTests.GlobalAveragePool_3D

<img width="426" height="312" alt="Image" src="https://github.com/user-attachments/assets/790e7412-8ebe-4809-a199-66dc6cbc5f1b" />

File is \onnxruntime\onnxruntime\core\providers\qnn\builder\opbuilder\pool_op_builder.cc, line 104

Assert is checking input_shape.size(), kernel_shape.size, stride.size() and pads.size(). Setting all checks to >= lets the tests pass, so at least the code isn't accessing out of bounds.

### Urgency

- Debug build is itself not urgent, until we need to actually debug anything more detailed than we can do with RelWithDebug.
- If the assert is indicating a fundamental error that could also manifest in Release builds, then urgency will increase
- Pipeline builds will fail (because there is no user to click on "Ignore").

### Target platform

Windows/x64

### Build script

```
python.exe ./tools/ci_build/build.py --config Debug --cmake_generator "Visual Studio 17 2022" --build_shared_lib --parallel 16 --build_dir build_x64-qnn  --cmake_extra_defines onnxruntime_ENABLE_MEMLEAK_CHECKER=OFF --cmake_extra_defines "onnxruntime_USE_NEURAL_SPEED=OFF" --use_qnn --qnn_home "..\deps\qnn_sdk\build\sdk" --enable_msvc_static_runtime --use_binskim_compliant_compile_flags
```

### Error / output

<img width="426" height="312" alt="Image" src="https://github.com/user-attachments/assets/131461fc-0cdc-4e29-9589-d5231a9684a4" />

### Visual Studio Version

17.14.12

a-jmist:
@HectorSVC as discussed

quic-tirupath:
The asserts are not updated in a recent update to the pool Op builder. We will fix the assert conditions or replace the assert with a return statement with ORT error to gracefully handle this failure condition.

PR #25788, opened 2025-08-19
[QNN EP] Fix Pool builder assert in Debug build.

minfhong-qti:
### Description

In PoolOpBuilder,
- Revise the check to exploit ORT macros.
- Fix invoking the function for 5D cases.

### Motivation and Context

Refer to https://github.com/microsoft/onnxruntime/issues/25778.
Pool builder incorrectly invokes a function calculating 4D shape in 5D input, which originally expects 3D cases only. However, the check used assert to validate the shape, which did not work in Release nor RelWithDebInfo builds.


minfhong-qti:

PR #25789, opened 2025-08-19
Log the illegal test data directory during onnxruntime_perf_test

qti-hungjuiw:
### Description

- This change improves traceability by explicitly logging the names of directories missing `.pb` files.

### Motivation and Context

- Currently, the log only states `there is no test input data for input "input.1" and model "onnx-model" when any of `test_data_set_X` lacks data.
- This makes it difficult to identify and remove invalid directories, especially when there are many `test_data_set_X` folders.

### DIFF
--- a/onnxruntime/test/onnx/TestCase.cc
+++ b/onnxruntime/test/onnx/TestCase.cc
@@ -346,7 +346,6 @@ class OnnxTestCase : public ITestCase {
   }
 
   size_t GetDataCount() const override { return test_data_dirs_.size(); }
-  std::filesystem::path GetDataDir(size_t id) const override { return test_data_dirs_[id]; }
   const std::string& GetNodeName() const override { return model_info_->GetNodeName(); }
   const std::filesystem::path& GetModelUrl() const override { return model_info_->GetModelUrl(); }
   const std::string& GetTestCaseName() const override { return test_case_name_; }
--- a/onnxruntime/test/onnx/TestCase.h
+++ b/onnxruntime/test/onnx/TestCase.h
@@ -44,7 +44,6 @@ class ITestCase {
   virtual std::string GetDatasetDebugInfoString(size_t dataset_id) const = 0;
   // The number of input/output pairs
   virtual size_t GetDataCount() const = 0;
-  virtual std::filesystem::path GetDataDir(size_t id) const = 0;
   virtual ~ITestCase() = default;
   virtual void GetPerSampleTolerance(double* value) const = 0;
   virtual void GetRelativePerSampleTolerance(double* value) const = 0;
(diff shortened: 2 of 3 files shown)


chuteng-quic:
Hi @HectorSVC , this PR is raised by QCOM, could you help review this PR?
Thanks!

onnxruntime/test/onnx/TestCase.cc
```diff
@@ -346,6 +346,7 @@ class OnnxTestCase : public ITestCase {
   }
 
   size_t GetDataCount() const override { return test_data_dirs_.size(); }
+  std::filesystem::path GetDataDir(size_t id) const override { return test_data_dirs_[id]; }
```
edgchen1: there's no validation of `id` being in range

onnxruntime/test/onnx/TestCase.h
```diff
@@ -44,6 +44,7 @@ class ITestCase {
   virtual std::string GetDatasetDebugInfoString(size_t dataset_id) const = 0;
   // The number of input/output pairs
   virtual size_t GetDataCount() const = 0;
+  virtual std::filesystem::path GetDataDir(size_t id) const = 0;
```
edgchen1: is it necessary to add a new virtual method? does the existing `GetDatasetDebugInfoString()` have information that can be used instead?

onnxruntime/test/onnx/TestCase.cc
qti-hungjuiw: Since `GetDatasetDebugInfoString()` is sufficient for the use case, I've removed the `GetDataDir()`.

onnxruntime/test/onnx/TestCase.h
qti-hungjuiw: Hi @edgchen1, thanks for pointing this out. Sure, I think `GetDatasetDebugInfoString()` is sufficient.

onnxruntime/test/perftest/performance_runner.cc
```diff
@@ -325,8 +325,10 @@ bool PerformanceRunner::Initialize() {
     for (int i = 0; i != input_count; ++i) {
       auto iter = feeds.find(test_model_info->GetInputName(i));
       if (iter == feeds.end()) {
-        std::cout << "there is no test input data for input " << test_model_info->GetInputName(i) << " and model "
-                  << test_case_->GetTestCaseName() << std::endl;
+        std::cout << "there is no test input data in " << test_case_->GetDatasetDebugInfoString(test_data_id)
+                  << " for input " << test_model_info->GetInputName(i)
+                  << " and model " << test_case_->GetTestCaseName()
+                  << std::endl;
```
edgchen1, line 331: I believe the debug info may contain more than just the directory path. how about something like this?

```
        std::cout << "there is no test input data for input " << test_model_info->GetInputName(i)
                  << " and model " << test_case_->GetTestCaseName()
                  << " in dataset: " << test_case_->GetDatasetDebugInfoString(test_data_id)
                  << std::endl;
```

Issue #25790, opened 2025-08-19
[AIX] Test failures fixes


ranjitshs:
### Describe the issue

This issue is to track/discuss the AIX (Big Endian system ) test failures introduced recently mainly by https://github.com/microsoft/onnxruntime/pull/23979 ,  along with some additional failures observed.

So Below are the list of test-suites failures .
- onnxruntime_test_all
```
1: [  FAILED  ] 28 tests, listed below:
1: [  FAILED  ] FlatbufferUtilsTest.ExternalWriteReadWithLoadInitializers
1: [  FAILED  ] InternalTestingEP.TestSaveAndLoadOrtModel
1: [  FAILED  ] DequantizeLinearOpTest.DequantizeLinear_per_tensor_float_int16_cpu
1: [  FAILED  ] DequantizeLinearOpTest.DequantizeLinear_per_tensor_float_uint16_cpu
1: [  FAILED  ] DequantizeLinearOpTest.Int16
1: [  FAILED  ] DequantizeLinearOpTest.Uint16
1: [  FAILED  ] QuantizeLinearOpTest.Uint16
1: [  FAILED  ] QuantizeLinearOpTest.Int16
1: [  FAILED  ] MathOpTest.Clip_MLFloat16
1: [  FAILED  ] QuantizeLinearContribOpTest.QuantizeLinear_per_tensor_float_uint16
1: [  FAILED  ] QuantizeLinearContribOpTest.QuantizeLinear_per_tensor_float_int16
1: [  FAILED  ] LayerNormTest.LayerNorm_Scale_Float16InputScaleOutput_Initializers
1: [  FAILED  ] LayerNormTest.LayerNorm_Scale_Bias_Broadcast_Fp16
1: [  FAILED  ] LayerNormTest.LayerNorm_Scale_Bias_Float16InputScaleBiasOutput
1: [  FAILED  ] LayerNormTest.LayerNorm_Scale_Bias_Float16InputScaleBiasOutput_Initializers
1: [  FAILED  ] SparseTensorConversionTests.TestConstantNodeConversion
1: [  FAILED  ] OrtModelOnlyTests.ValidateOrtFormatModelDoesNotRunOptimizersInFullBuild
1: [  FAILED  ] ConvertRawDataInTensorProtoTest.FloatData
1: [  FAILED  ] ConvertRawDataInTensorProtoTest.Int32Data
1: [  FAILED  ] QDQTransformerTests.Clip
[...]

PR #25791, opened 2025-08-19
[AIX] test-suites failure fixes 

ranjitshs:
### Description
This PR is to fix some of the test case failures mentioned in https://github.com/microsoft/onnxruntime/issues/25790

1. cmake/onnxruntime_unittests.cmake
   To fix the onnxruntime_shared_lib_test crash
2. include/onnxruntime/core/providers/utils/ort_graph_to_proto.h
   Added method for handling external data.
   For fixing , EpGraphTest.SerializeToProto_ConstantOfShape , EpGraphTest.SerializeToProto_Mnist,  EpGraphTest.SerializeToProto_InputModelHasExternalIni
3. onnxruntime/core/framework/tensorprotoutils.cc
   For fixing many test failures , related to float16 or Int16 type.
4. onnxruntime/test/framework/endian_test.cc
  For fixing, ConvertRawDataInTensorProtoTest.FloatData and ConvertRawDataInTensorProtoTest.Int32Data
  
### Motivation and Context
To fix the AIX (Big endian) related test failures.


ranjitshs:
@snnn @tianleiwu @yuslepukhin 
FYI. 
Could you please review this PR and let me know your inputs.

Also, please trigger CI workflows for this PR.

onnxruntime/core/framework/tensorprotoutils.cc
```diff
@@ -362,11 +362,13 @@ void ConvertRawDataInTensorProto(TensorProto& tensor) {
     case TensorProto_DataType_INT16:
     case TensorProto_DataType_FLOAT16:
     case TensorProto_DataType_BFLOAT16:
+      bytes = tensor.mutable_int32_data()->mutable_data();
```
yuslepukhin: This would be incorrect. All 4 bytes in should be swapped for a single 32 bit, not just 2. #Resolved

yuslepukhin (changes requested):

onnxruntime/core/framework/tensorprotoutils.cc
ranjitshs: @yuslepukhin 
Some scenarios, I see where sizeof() is used while setting the raw data . 
If we consider those  case, then we need to consider 2 only  as sizeof(BFLOAT16) is 2. 

- Do we need to consider both type of handling in this method. 
    One for data added by set_raw_data() and other for data added by add_int32_data types of API's.
    
- Or since this method is only for handling raw data, Can we first check if raw data is present or not and return if not.  Not sure why we haven't added during first time, so there could be some reason and need to test it out. 

Let me know your input on this.

Below is one example of test case where we use sizeof() for setting the raw data.
```
 template <typename T>
void TestInitializerRawData() {
  std::vector<T> data = GetInitializerData<T>();

  ONNX_NAMESPACE::TensorProto tensor_proto;
  tensor_proto.set_data_type(GetTensorProtoDataType<T>());
  tensor_proto.set_name("OptimizerInitializerTest_RawData");
  tensor_proto.add_dims(3);
  tensor_proto.add_dims(4);
  utils::SetRawDataInTensorProto(tensor_proto, data.data(), data.size() * sizeof(T));

  const Initializer init(tensor_proto, std::filesystem::path());

  for (size_t idx = 0; idx < data.size(); idx++) {
    EXPECT_EQ(data[idx], init.data<T>()[idx]);
  }
}

TEST(OptimizerInitializerTest, RawData) {
  TestInitializerRawData<int8_t>();
  TestInitializerRawData<uint8_t>();
  TestInitializerRawData<int32_t>();
  TestInitializerRawData<int64_t>();
  TestInitializerRawData<MLFloat16>();
  TestInitializerRawData<BFloat16>();
  TestInitializerRawData<float>();
  TestInitializerRawData<double>();
}
```

yuslepukhin: If the data is stored as raw data, and the size of the type is 2 bytes, then they are stored contigiously and for that sizeof() needs to be 2 bytes.
However, if it is *not* stored as raw data, and in this case stored in int32, then all 4 bytes need to be swapped, although only two bytes are meaningful.
Thus we either need two different switches, branch on the presense of raw data and set the sizeof() accordingly OR we need two different sizes for raw data and otherwise.

ranjitshs: Okay. I will put two different switches based on raw data presence.  Let me test it out.

ranjitshs: @yuslepukhin
Updated the change. Please have a look.

ranjitshs:
Also, for handling external data I have added ConvertExternalData()  in ort_graph_to_proto.h which will be called in case be BE system Before writing data to file. 

With this change, all tests for onnxruntime_ep_graph_test are passing.

onnxruntime/core/framework/tensorprotoutils.cc
```diff
@@ -344,66 +344,109 @@ void ConvertRawDataInTensorProto(TensorProto& tensor) {
...
+
+      case TensorProto_DataType_INT64:
+        element_size = sizeof(int64_t);
+        break;
+
+      case TensorProto_DataType_DOUBLE:
+        element_size = sizeof(double);
+        break;
+
+      case TensorProto_DataType_COMPLEX64:
```
yuslepukhin: >   case Tensor [](http://example.com/codeflow?start=3&length=14)

We do not support it and there are no tests for that.

yuslepukhin: >   case Tensor [](http://example.com/codeflow?start=3&length=14)

The function is too long now, hard to read. Need a way to bring size back to be comparable to what it was. #Resolved

ranjitshs: okay. I will remove it.

ranjitshs: okay. I will remove this.

ranjitshs: I can remove the new raw data switch block and use  static unordered map as  this switch only sets the element size. 

Let me know if it's okay to use map instead of switch or 
Any suggestions for  improvement .

```
    static std::unordered_map<size_t,size_t> tensorproto_data_size{{TensorProto_DataType_FLOAT,sizeof(float)},
                                                     {TensorProto_DataType_UINT8,sizeof(uint8_t)},
                                                     {TensorProto_DataType_INT8,sizeof(int8_t)},
                                                     {TensorProto_DataType_UINT16,sizeof(uint16_t)},
                                                     {TensorProto_DataType_INT16,sizeof(int16_t)},
                                                     {TensorProto_DataType_FLOAT16,sizeof(uint16_t)},
                                                     {TensorProto_DataType_BFLOAT16,sizeof(uint16_t)},
                                                     {TensorProto_DataType_INT32,sizeof(int32_t)},
                                                     {TensorProto_DataType_UINT32,sizeof(uint32_t)},
                                                     {TensorProto_DataType_UINT64,sizeof(uint64_t)},
                                                     {TensorProto_DataType_INT64,sizeof(int64_t)},
                                                     {TensorProto_DataType_DOUBLE,sizeof(double)},
                                                     };
```

yuslepukhin: Using a map (actually two maps that are function local statics) is not a bad idea.

ranjitshs:
@yuslepukhin 
Let me know if any further comment in this PR . Can we close this PR if no further comments.

I will spend time for rest failures and will raise separate PR.

```
1: [  FAILED  ] FlatbufferUtilsTest.ExternalWriteReadWithLoadInitializers
1: [  FAILED  ] InternalTestingEP.TestSaveAndLoadOrtModel
1: [  FAILED  ] OrtModelOnlyTests.ValidateOrtFormatModelDoesNotRunOptimizersInFullBuild
```

onnxruntime/core/framework/tensorprotoutils.cc
```diff
@@ -344,66 +344,85 @@ void ConvertRawDataInTensorProto(TensorProto& tensor) {
...
-      break;
-  }
-
-  if (element_size == 1) {
-    return;
-  }
-
+  // For some data_type, element size differs for raw data vs
+  // data set using the add_<data_type>data() API
   if (tensor.has_raw_data()) {
```
yuslepukhin: >tensor.has_raw_data()) { [](http://example.com/codeflow?start=6&length=24)

HasRawData(tensor) #Resolved

yuslepukhin: You might also want to add quantized types:

[https://github.com/onnx/onnx/blob/main/onnx/onnx-ml.proto3#L637](https://github.com/onnx/onnx/blob/main/onnx/onnx-ml.proto3#L637)

onnxruntime/core/framework/tensorprotoutils.cc
```diff
@@ -344,66 +344,85 @@ void ConvertRawDataInTensorProto(TensorProto& tensor) {
...
+        num_elements = tensor.float_data_size();
+        element_size = sizeof(float);
+        break;
+
+      case TensorProto_DataType_UINT8:
+      case TensorProto_DataType_INT8:
+        bytes = tensor.mutable_int32_data()->mutable_data();
+        num_elements = tensor.int32_data_size();
+        element_size = sizeof(uint8_t);
+        break;
```
yuslepukhin: I think all datatypes with size == 1 and stored in int32 data, must be swapped. This includes BOOL.
Reason being, the meaningful byte happens to be on the wrong end. So this case must be merged with the below.

Size == 1 exception is only for raw data where they are stored contiguously.

onnxruntime/core/framework/tensorprotoutils.cc
```diff
@@ -344,66 +344,85 @@ void ConvertRawDataInTensorProto(TensorProto& tensor) {
...
+        {TensorProto_DataType_UINT64, sizeof(uint64_t)},
+        {TensorProto_DataType_INT64, sizeof(int64_t)},
+        {TensorProto_DataType_DOUBLE, sizeof(double)},
+    };
+    auto pos = tensorproto_data_size.find(tensor.data_type());
+    if (pos == tensorproto_data_size.end()) {
+      return;
+    }
+    element_size = pos->second;
+    if (element_size <= 1) {
```
yuslepukhin: > if ( [](http://example.com/codeflow?start=3&length=5)

Cannot be 0 or negative. #Resolved

onnxruntime/core/framework/tensorprotoutils.cc
```diff
@@ -344,66 +344,85 @@ void ConvertRawDataInTensorProto(TensorProto& tensor) {
...
+      case TensorProto_DataType_DOUBLE:
+        bytes = tensor.mutable_double_data()->mutable_data();
+        num_elements = tensor.double_data_size();
+        element_size = sizeof(double);
+        break;
+    }
+
+    if (element_size == 1) {
+      return;
+    }
```
yuslepukhin: This should never be true here. Perhaps removing it is best. #Resolved

onnxruntime/test/ep_graph/test_ep_graph.cc
```diff
@@ -242,6 +242,11 @@ TEST(EpGraphTest, SerializeToProto_InputModelHasExternalIni) {
         return Ort::Status{nullptr};
       }
 
+      // For BE system, Before writing to file, we need to do data coversion.
+      if constexpr (endian::native != endian::little) {
+        OrtEpUtils::ConvertExternalData(value_info, const_cast<void*>(data), bytes);
+      }
```
yuslepukhin, line 231: Please, create a copy, do not  modify the original const data, the consequences of that is not known.
it also serves as a bad example and a source of bugs.
In production, the process may continue with data swapped which would be wrong for AIX anyway. #Resolved

onnxruntime/test/ep_graph/test_ep_graph.cc
```diff
@@ -361,12 +366,16 @@ TEST(EpGraphTest, SerializeToProto_Mnist) {
...
       (void)value_info;
-
       if (bytes <= 127) {
         is_external = false;  // Keep small initializers stored inside the TensorProto.
         return Ort::Status{nullptr};
       }
 
+      // For BE system, Before writing to file, we need to do data coversion.
+      if constexpr (endian::native != endian::little) {
+        OrtEpUtils::ConvertExternalData(value_info, const_cast<void*>(data), bytes);
```
yuslepukhin: >       OrtEpUtils::C [](http://example.com/codeflow?start=1&length=20)

Same here #Resolved

include/onnxruntime/core/providers/utils/ort_graph_to_proto.h
```diff
@@ -864,5 +867,42 @@ static Ort::Status OrtOpAttrToProto(const OrtOpAttr& ort_attr, onnx::AttributePr
   return Ort::Status{nullptr};
 }
 
+static Ort::Status ConvertExternalData(const OrtValueInfo* value_info, void* data, size_t bytes) {
+  if constexpr (onnxruntime::endian::native == onnxruntime::endian::little) {
+    return Ort::Status{nullptr};
+  }
+  using tensor_elem_data_map = std::unordered_map<ONNXTensorElementDataType, size_t>;
```
yuslepukhin: >tensor_elem_data_map [](http://example.com/codeflow?start=8&length=20)

types should have camel notation #Resolved

yuslepukhin:
Added more comments, this also needs to be rebased.

onnxruntime/core/framework/tensorprotoutils.cc
ranjitshs: Okay. Will add all quantized types.

onnxruntime/core/framework/tensorprotoutils.cc
ranjitshs: Right. Will check with 1.

onnxruntime/core/framework/tensorprotoutils.cc
ranjitshs: Will add BOOL and quantized types also .

onnxruntime/test/ep_graph/test_ep_graph.cc
ranjitshs: Done.

onnxruntime/test/ep_graph/test_ep_graph.cc
ranjitshs: Done

include/onnxruntime/core/providers/utils/ort_graph_to_proto.h
ranjitshs: Done

onnxruntime/core/framework/tensorprotoutils.cc
ranjitshs: okay.

ranjitshs:
> Added more comments, this also needs to be rebased.

Will address these comments in next commit. Thanks for reviewing it.

ranjitshs:
I see in some OS Ci's is failing with below . I will change and make this method as API . Wondering why it's not hit in all OS. 
Could be strict compiler flags or latest compiler is  used where it's failing. 

```
/mnt/vss/_work/onnxruntime/onnxruntime/include/onnxruntime/core/providers/utils/ort_graph_to_proto.h:720:20: error: 'static' function 'ConvertExternalData' declared in header file should be declared 'static inline' [-Werror,-Wunneeded-internal-declaration]
  720 | static Ort::Status ConvertExternalData(const OrtValueInfo* value_info, void* data, size_t bytes) {
```

include/onnxruntime/core/providers/utils/ort_graph_to_proto.h
```diff
@@ -123,6 +123,7 @@
 
 #include <functional>
 #include "core/session/onnxruntime_cxx_api.h"
+#include "core/framework/endian.h"
```
adrianlizarraga: Can we avoid including other ORT code files into this file (other than onnxruntime_cxx_api.h)? This file is meant to be easy to copy to other projects/repositories as stated at the top of this file. This may mean that we have to copy/rename the struct in `core/framework/endian.h` in this file.

include/onnxruntime/core/providers/utils/ort_graph_to_proto.h
```diff
@@ -435,7 +438,7 @@ Ort::Status OrtGraphToProto(const OrtGraph& graph,
       } else {
         // User wants to store data inline the TensorProto's raw_data
         tensor_proto->set_data_location(onnx::TensorProto_DataLocation_DEFAULT);
-        tensor_proto->set_raw_data(data, data_bytes);
+        onnxruntime::utils::SetRawDataInTensorProto(*tensor_proto, data, data_bytes);
```
adrianlizarraga: I believe this function comes from an internal ORT file. We do not want other dependencies in this file. Please refer to this comment: https://github.com/microsoft/onnxruntime/pull/25791/files#r2325723465

ranjitshs: @adrianlizarraga 
okay.  In that case , how we can use some ORT utilities for BE support from this file. 
Should we design and expose those utilities using C/C++ API and then use it  here ? 

@yuslepukhin 
FYI.  Let me know if your input/comment regarding above design or any new approach.

ranjitshs:
@yuslepukhin @adrianlizarraga 
To not put any dependency on ort_graph_to_proto.h, I am thinking to add the required BE related method in this  header file as static method . I am testing my changes  and will upload for review.

ranjitshs:
@yuslepukhin @adrianlizarraga 
Could you please review the latest ort_graph_to_proto.h and let me know your input/comments.

ranjitshs:
HI @yuslepukhin @adrianlizarraga 
Could you please let me know if any further changes is needed in this PR.

include/onnxruntime/core/providers/utils/ort_graph_to_proto.h
```diff
@@ -226,8 +236,29 @@ static Ort::Status GetOrtValueInfoTensorTypeShape(Ort::ConstValueInfo vi,
...
+  const int little = __ORDER_LITTLE_ENDIAN__;
+  const int big = __ORDER_BIG_ENDIAN__;
+  const int native = __BYTE_ORDER__;
+#endif
+  if constexpr (native == big) {
+    return true;
+  } else {
+    return false;
+  }
+}
```
yuslepukhin: There is a header in infrastructure that can be copied here to make it universal.

include/onnxruntime/core/providers/utils/ort_graph_to_proto.h
```diff
@@ -695,6 +731,11 @@ static Ort::Status OrtOpAttrToProto(Ort::ConstOpAttr attr, onnx::AttributeProto&
         const size_t data_bytes = tensor.GetTensorSizeInBytes();
 
         // Copy the Ortvalue to TensorProto as raw data
+        if constexpr (isBigEndian()) {
+          size_t element_size = 0;
+          GetTensorElementSize(element_type, element_size);
+          SwapByteOrderInplace(const_cast<void*>(data), data_bytes, element_size);
```
yuslepukhin: >SwapByteOrderInplace(const_cast<void*>(data),  [](http://example.com/codeflow?start=10&length=46)

There is a small utility that can also be copies here.
It would helps us to copy/convert which is what we need and avoid const_cast. It is in framework/endianess.cc

onnxruntime/core/framework/tensorprotoutils.cc
yuslepukhin: The single bytes are still stored in int32_t so these cases must be merged with the below cases and the size will be sizeof(int32_t)

onnxruntime/test/framework/endian_test.cc
```diff
@@ -214,7 +214,7 @@ class ConvertRawDataInTensorProtoTest : public ::testing::Test {
         // Just verify the value is different after one swap on big-endian
         // We can't predict the exact value without manual byte swapping
         if (expected_values[i] != 0) {  // Skip zero values as they're invariant to byte swapping
-          EXPECT_NE(tensor.float_data(i), expected_values[i]);
+          EXPECT_EQ(tensor.float_data(i), expected_values[i]);
```
yuslepukhin: The comment contradicts the assert.
I think this should be NE

onnxruntime/test/framework/endian_test.cc
```diff
@@ -229,7 +229,7 @@ class ConvertRawDataInTensorProtoTest : public ::testing::Test {
         EXPECT_EQ(tensor.int32_data(i), expected_values[i]);
       } else {
         if (expected_values[i] != 0) {
-          EXPECT_NE(tensor.int32_data(i), expected_values[i]);
+          EXPECT_EQ(tensor.int32_data(i), expected_values[i]);
```
yuslepukhin: >XPECT_EQ(tensor.int32_ [](http://example.com/codeflow?start=11&length=22)

Ditto?

include/onnxruntime/core/providers/utils/ort_graph_to_proto.h
ranjitshs: okay. Will update.

include/onnxruntime/core/providers/utils/ort_graph_to_proto.h
ranjitshs: SwapByteOrderInplace implemented in endian_utils.cc is having gsl dependency that's why I thought not to use that. 
is gsl implementation is header only implementation ? Can we include that gsl header and use it here ?
Let me know.

onnxruntime/test/framework/endian_test.cc
ranjitshs: Will update it and remove above platform check and put common EXPECT_EQ.

If we do double swap , we are getting the original value so EQ is valid for BE system also.

include/onnxruntime/core/providers/utils/ort_graph_to_proto.h
ranjitshs: gsl looks header only implementation. So if we include gsl in this file, client of this header will require gsl as dependencies.

ranjitshs: @yuslepukhin 
Could you let me know your opinion on my understanding and provide your comment.

ranjitshs:
@yuslepukhin 
Hoping to get further comments from you so I can proceed further on this.

yuslepukhin:
@ranjitshs General comment. I think what is happening with AIX is trying to hit a moving target. ORT is under active development.
So the best way to approach this is to setup a CI pipeline for AIX. We do not have hardware for this for obvious reasons. However, if you have an ability to provision such a hardware, it would be possible to create a CI pipeline that would run on every PR.

yuslepukhin:
I can see it should be rebased.

onnxruntime/test/framework/endian_test.cc
yuslepukhin: Where is the double swap coming from?

onnxruntime/core/framework/tensorprotoutils.cc
yuslepukhin: This is still not addressed.

include/onnxruntime/core/providers/utils/ort_graph_to_proto.h
adrianlizarraga: For context: This file was meant to an example implementation of how to serialize an OrtGraph to protobuf. It was meant to be easy to copy to other code bases, where the EP author could modify and improve it as needed (e.g., adding BE). Adding include statements for other ORT files makes the above more difficult as now multiple files have to be copied.

To answer the question, I would say that the implementations of the functions that come from a different file have to be pasted into this file to meet the above objectives. Others may disagree.

yuslepukhin: I agree with @adrianlizarraga

onnxruntime/test/framework/endian_test.cc
ranjitshs: In test case ConvertRawDataInTensorProtoTest, we call ConvertRawDataInTensorProto two times. And after that , we call this method for result comparing .

onnxruntime/core/framework/tensorprotoutils.cc
ranjitshs: Done

ranjitshs:
@yuslepukhin 
I did rebasing and address review comments . Because of gsl dependency, I didn't include bytes swapping utilities ,yet Let me know your opinion on this. 

Please trigger CI workflows.

ranjitshs:
> @ranjitshs General comment. I think what is happening with AIX is trying to hit a moving target. ORT is under active development. So the best way to approach this is to setup a CI pipeline for AIX. We do not have hardware for this for obvious reasons. However, if you have an ability to provision such a hardware, it would be possible to create a CI pipeline that would run on every PR.

Sure. Though, I don't have full understanding of how onnxruntime CI is hosted for other OS'es . I guess it's using VM hosted over Azure.   I will start a separate conversation for understanding CI requirements and see whether we can provide the required infra.

ranjitshs:
@yuslepukhin 
I was going through CI failures . There are some python related import failures, not related to this PR. 

In windows debug X64, I see below . But same is passing in release mode.
I don't have windows for analysing it. Any suggestion .
I am thinking to put from line num 762 to 775 under `if !defined(_WIN32)`
```
Error: E:\_work\onnxruntime\onnxruntime\include\onnxruntime\core\providers\utils\ort_graph_to_proto.h(762,1): error C2220: the following warning is treated as an error [E:\_work\onnxruntime\onnxruntime\build\Debug\onnxruntime_ep_graph_test.vcxproj]
Warning: E:\_work\onnxruntime\onnxruntime\include\onnxruntime\core\providers\utils\ort_graph_to_proto.h(762,1): warning C4702: unreachable code [E:\_work\onnxruntime\onnxruntime\build\Debug\onnxruntime_ep_graph_test.vcxproj]
Warning: E:\_work\onnxruntime\onnxruntime\include\onnxruntime\core\providers\utils\ort_graph_to_proto.h(768,1): warning C4702: unreachable code [E:\_work\onnxruntime\onnxruntime\build\Debug\onnxruntime_ep_graph_test.vcxproj]
Warning: E:\_work\onnxruntime\onnxruntime\include\onnxruntime\core\providers\utils\ort_graph_to_proto.h(771,1): warning C4702: unreachable code [E:\_work\onnxruntime\onnxruntime\build\Debug\onnxruntime_ep_graph_test.vcxproj]
Warning: E:\_work\onnxruntime\onnxruntime\include\onnxruntime\core\providers\utils\ort_graph_to_proto.h(773,1): warning C4702: unreachable code [E:\_work\onnxruntime\onnxruntime\build\Debug\onnxruntime_ep_graph_test.vcxproj]
Warning: E:\_work\onnxruntime\onnxruntime\include\onnxruntime\core\providers\utils\ort_graph_to_proto.h(775,1): warning C4702: unreachable code [E:\_work\onnxruntime\onnxruntime\build\Debug\onnxruntime_ep_graph_test.vcxproj]
```

ranjitshs:
Hi @yuslepukhin 
I have updated the code to fix windows debug build warning mentioned above. 
Could you please run CI workflow .

ranjitshs:
@yuslepukhin 
I see  some new wanring(converted to error) under mainly WIN workflows. 

```
Error: D:\a\_work\onnxruntime\onnxruntime\include\onnxruntime\core\providers\utils\ort_graph_to_proto.h(758,84): error C2220: the following warning is treated as an error [D:\a\_work\onnxruntime\onnxruntime\RelWithDebInfo\onnxruntime_ep_graph_test.vcxproj] 
Warning: D:\a\_work\onnxruntime\onnxruntime\include\onnxruntime\core\providers\utils\ort_graph_to_proto.h(758,84): warning C4100: 'bytes': unreferenced formal parameter [D:\a\_work\onnxruntime\onnxruntime\RelWithDebInfo\onnxruntime_ep_graph_test.vcxproj] 
Warning: D:\a\_work\onnxruntime\onnxruntime\include\onnxruntime\core\providers\utils\ort_graph_to_proto.h(758,71): warning C4100: 'data': unreferenced formal parameter [D:\a\_work\onnxruntime\onnxruntime\RelWithDebInfo\onnxruntime_ep_graph_test.vcxproj] 
Warning: D:\a\_work\onnxruntime\onnxruntime\include\onnxruntime\core\providers\utils\ort_graph_to_proto.h(758,53): warning C4100: 'value_info': unreferenced formal parameter [D:\a\_work\onnxruntime\onnxruntime\RelWithDebInfo\onnxruntime_ep_graph_test.vcxproj]
```
Update the code to handle unreferenced function parameters 
Once you get time, please trigger CI once. Hope, this time it will go through.  sorry for inconvenience

yuslepukhin:
You can comment out the names of the unreferenced params

ranjitshs:
@yuslepukhin 
I see 4 pending checks . What's needs to be done for that.

ranjitshs:
@yuslepukhin 
One of check Windows ARM64 QNN CI Pipeline above is timing out ..  I don't see any build issues in log.
Please re-trigger it.

ranjitshs:
@yuslepukhin 
If we are okay with these changes , then please approve and merge.

ranjitshs:
Hi @yuslepukhin @tianleiwu 
BUILD_QNN_EP STATIC_LIB is failed with below error. Please re-trigger this workflow.

```
Starting: NuGetToolInstaller
==============================================================================
Task         : NuGet tool installer
Description  : Acquires a specific version of NuGet from the internet or the tools cache and adds it to the PATH. Use this task to change the version of NuGet used in the NuGet tasks.
Version      : 1.260.0
Author       : Microsoft Corporation
Help         : https://docs.microsoft.com/azure/devops/pipelines/tasks/tool/nuget
==============================================================================
You are using a query match on the version string. Behavior changes or breaking changes might occur as NuGet updates to a new version.
ERR:write EPROTO C81D0000:error:0A000438:SSL routines:ssl3_read_bytes:tlsv1 alert internal error:c:\ws\deps\openssl\openssl\ssl\record\rec_layer_s3.c:1605:SSL alert number 80

Finishing: NuGetToolInstaller
```

include/onnxruntime/core/providers/utils/ort_graph_to_proto.h
yuslepukhin: I do not like const_cast. The copy of the data must be made anyway, so, please, make the copy and then set raw data.

yuslepukhin: Just one more comment above

ranjitshs:
@yuslepukhin 
Update the code as per your suggestion. Please check once and trigger workflows

yuslepukhin (approved):

ranjitshs:
@yuslepukhin @tianleiwu 
Thanks for reviewing/merging this PR.

PR #25797, opened 2025-08-20
Add model compilation in ORT perf test

thevishalagarwal:
This PR adds model compilation support in ort perf test
* Add `--compile_ep_context` flag in onnxruntime_perf_test. This generates EP context model and prints out compilation time and perf statistics for the compiled model
* Prints `Compile time cost` for EP context compilation

Sample usage:
```sh
$ ./onnxruntime_perf_test -e trt-rtx -I -r 1 "/path/to/model.onnx" --compile_ep_context --compile_model_path "/path/to/model_ctx.onnx"
```

Output:
```sh
Compile time cost: 17.8012 s
Session creation time cost: 0.966619 s
First inference time cost: 8151 ms
Total inference time cost: 8.08084 s
Total inference requests: 1
Average inference time cost: 8080.84 ms
Total inference run time: 8.08085 s
Number of inferences per second: 0.123749
Avg CPU usage: 6 %
Peak working set size: 7861874688 bytes
Avg CPU usage:6
Peak working set size:7861874688
Runs:1
Min Latency: 8.08084 s
Max Latency: 8.08084 s
P50 Latency: 8.08084 s
P90 Latency: 8.08084 s
P95 Latency: 8.08084 s
P99 Latency: 8.08084 s
P999 Latency: 8.08084 s
```


thevishalagarwal:
@jywu-msft @chilo-ms Can you please review this?

onnxruntime/test/perftest/command_args_parser.cc
```diff
@@ -163,6 +163,9 @@ ABSL_FLAG(std::string, plugin_ep_options, "",
           "--plugin_ep_options \"ep_1_option_1_key|ep_1_option_1_value ...;;ep_3_option_1_key|ep_3_option_1_value ...;... \"");
 ABSL_FLAG(bool, list_ep_devices, false, "Prints all available device indices and their properties (including metadata). This option makes the program exit early without performing inference.\n");
 ABSL_FLAG(std::string, select_ep_devices, "", "Specifies a semicolon-separated list of device indices to add to the session and run with.");
+ABSL_FLAG(bool, compile_ep_context, DefaultPerformanceTestConfig().run_config.compile_ep_context, "Generate an EP context model");
+ABSL_FLAG(std::string, compile_model_path, "", "The compiled model path for saving EP context model. Overwrites if already exists. Default is model_ctx.onnx");
```
edgchen1: the second parameter of `ABSL_FLAG` specifies a default value. can we just use that instead of setting it later?

onnxruntime/test/perftest/command_args_parser.cc
```diff
@@ -163,6 +163,9 @@ ABSL_FLAG(std::string, plugin_ep_options, "",
...
+ABSL_FLAG(int, compile_embed_mode, DefaultPerformanceTestConfig().run_config.compile_embed_mode, "Embed binary blob within EP context node. Can be 0 or 1. Default is 0");
```
edgchen1: nit: don't need to explicitly mention a default value in the help text. the generated help message will already include the default value.

onnxruntime/test/perftest/main.cc
```diff
@@ -30,7 +33,7 @@ int real_main(int argc, char* argv[]) {
     ORT_TRY {
       OrtLoggingLevel logging_level = test_config.run_config.f_verbose
                                           ? ORT_LOGGING_LEVEL_VERBOSE
-                                          : ORT_LOGGING_LEVEL_WARNING;
+                                          : ORT_LOGGING_LEVEL_ERROR;
```
edgchen1: why this change?

onnxruntime/test/perftest/main.cc
```diff
@@ -107,3 +110,55 @@ int main(int argc, char* argv[]) {
...
+    return -1;
+  }
+
+  perf_runner.SerializeResult();
+  return 0;
+}
+
+void CompileEpContextModel(Ort::Env& env, const perftest::PerformanceTestConfig& test_config) {
+  auto output_ctx_model_path = test_config.run_config.compile_model_path;
+  const auto provider_name_ = test_config.machine_config.provider_type_name;
```
edgchen1: we follow the naming conventions here: https://google.github.io/styleguide/cppguide.html#Variable_Names

trailing underscores are typically only used for private class data members.

onnxruntime/test/perftest/main.cc
```diff
@@ -107,3 +110,55 @@ int main(int argc, char* argv[]) {
...
+
+  Ort::Status status;
+  std::chrono::duration<double> compile_duration;
+  {
+    auto compile_time_start_ = std::chrono::high_resolution_clock::now();
+    status = Ort::CompileModel(env, model_compile_options);
+    auto compile_time_end_ = std::chrono::high_resolution_clock::now();
+    compile_duration = compile_time_end_ - compile_time_start_;
+  }
+  model_compile_options.release();
```
edgchen1: why is release called here?

`Base::release()` does not free the contained pointer. it is more like `std::unique_ptr::release()`.

https://github.com/microsoft/onnxruntime/blob/0482251e6253a76a6cb019514896bb4d79bce3ca/include/onnxruntime/core/session/onnxruntime_cxx_api.h#L640-L646

as we don't store the returned value, it looks like it will be leaked.

the dtor of `Ort::ModelCompilationOptions` should automatically free the contained pointer.

onnxruntime/test/perftest/main.cc
```diff
@@ -107,3 +110,55 @@ int main(int argc, char* argv[]) {
...
+
+  if (!status.IsOK()) {
+    std::cout << "Failed to compile model: " << status.GetErrorMessage() << std::endl;
```
edgchen1, line 161: do we want to continue if there is an error?

onnxruntime/test/perftest/main.cc
```diff
@@ -107,3 +110,55 @@ int main(int argc, char* argv[]) {
...
+  if (!status.IsOK()) {
+    printf("Run failed:%s\n", status.ErrorMessage().c_str());
+    return -1;
+  }
+
+  perf_runner.SerializeResult();
+  return 0;
+}
+
+void CompileEpContextModel(Ort::Env& env, const perftest::PerformanceTestConfig& test_config) {
```
edgchen1: can `env` be `const Ort::Env&`?

onnxruntime/test/perftest/command_args_parser.cc
edgchen1: why not a boolean option?

thevishalagarwal: Will remove the default message. 

> why not a boolean option?

The embed mode is an integer option in the ORT so making it consistent

onnxruntime/test/perftest/command_args_parser.cc
thevishalagarwal: Thanks for the review. Updated it

onnxruntime/test/perftest/main.cc
thevishalagarwal: Unintended changes. Reverted it back

onnxruntime/test/perftest/main.cc
thevishalagarwal: Updated all variables

onnxruntime/test/perftest/main.cc
thevishalagarwal: No, we would like to return with error. I have updated the implementation.

onnxruntime/test/perftest/command_args_parser.cc
edgchen1: > The embed mode is an integer option in the ORT so making it consistent

isn't it a bool?
https://github.com/microsoft/onnxruntime/blob/ce9060936201ef979a5075cd825a5942dec433ee/include/onnxruntime/core/session/onnxruntime_cxx_api.h#L1348

edgchen1:
general question. why is this PR adding two ways to compile the model? i.e., with onnxruntime_perf_test and compile_ep_context_model.py. IMO, onnxruntime_perf_test already does a lot - do we need to add more functionality to it or can we just use the python script to first generate the compiled model?

onnxruntime/test/perftest/command_args_parser.cc
thevishalagarwal: Oh, my bad. It is `int` in the EP so I assumed the same for ORT as well. I will update this to bool

https://github.com/microsoft/onnxruntime/blob/69ec7b17307f2b94e20199220108e7a415377cf1/onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider_info.cc#L78

thevishalagarwal:
> why is this PR adding two ways to compile the model?

The core utility is the same. Just added the same implementation for Python and C++ to reduce dependency overhead and devs can use whichever language they are already working on. Python is quicker to test and prototype.

thevishalagarwal:
@edgchen1 @chilo-ms ping for review

tools/python/compile_ep_context_model.py
```diff
@@ -0,0 +1,77 @@
...
+
+    model_compiler = ort.ModelCompiler(session_options, input_path, embed_compiled_data_into_model=embed_mode)
+
+    start = time.perf_counter()
+    model_compiler.compile_to_file(output_path)
+    stop = time.perf_counter()
+
+    if os.path.exists(output_path):
+        print("> Compiled successfully!")
+        print(f"> Comile time: {stop - start: .3f} sec")
```
edgchen1:
```suggestion
        print(f"> Compile time: {stop - start: .3f} sec")
```

tools/python/compile_ep_context_model.py
```diff
@@ -0,0 +1,77 @@
...
+# EPs
+TRT_RTX_EP = "NvTensorRTRTXExecutionProvider"
+
+
+def compile(input_path, output_path, provider, embed_mode=False):
+    if os.path.exists(output_path):
+        os.remove(output_path)
+
+    session_options = ort.SessionOptions()
+    session_options.add_provider(provider, {})
```
edgchen1: should this support a way to specify provider options?

edgchen1:
> > why is this PR adding two ways to compile the model?
> 
> The core utility is the same. Just added the same implementation for Python and C++ to reduce dependency overhead and devs can use whichever language they are already working on. Python is quicker to test and prototype.

I haven't used this, but a potential concern is having two separate implementations to maintain. is having both paths worth that maintenance overhead?

tools/python/compile_ep_context_model.py
thevishalagarwal: Added provider options support as key-val pair

thevishalagarwal:
> I haven't used this, but a potential concern is having two separate implementations to maintain. is having both paths worth that maintenance overhead?

IMO, it is not a big maintenance overhead. The python script is independent and does just one and only thing - generate EP context model. For python usecases like LLMs with ORT GenAI and Diffusion pipelines, a python based script is desirable for developers to compile model ahead-of-time, without needing to have the C/C++ build.

chilo-ms:
Discussed offline. Let's remove the python script for now and only merge the part in perf test.

chilo-ms (approved):

PR #25801, opened 2025-08-20
[MIGRAPHX EP] Workaround for reducing GPU memory allocations

apwojcik:
### Description
The current design of the MIGraphX graph compiler, which we use for the MIGraphX Execution Provider, requires managing GPU allocations independently from ONNXRuntime. With this PR, we implement a workaround to prevent ONNXRuntime from allocating initializers for MIGraphX EP. For all constant initializer tensors located in an AMD GPU device, the function responsible for allocating memory returns status OK without performing any actual allocation or data read from disk. Otherwise, those are double allocated and consume more memory than required. In the ONNXRuntime memory profiler, we are registering those kinds of allocations as empty blocks for further analysis.

### Motivation and Context
To reduce the memory allocations on the GPU when used with MIGraphX EP.

### DIFF
--- a/onnxruntime/core/framework/memory_info.cc
+++ b/onnxruntime/core/framework/memory_info.cc
@@ -47,7 +47,6 @@ void MemoryInfo::Init(const SequentialExecutionPlan* execution_plan,
     tensor_alloc_info_map_[value_idx] = std::move(mem_info);
     tensors_memory_info_map_[mem_info.location];
   }
-  return;
 }
 
 // Record the planned memory information
@@ -77,6 +76,12 @@ void MemoryInfo::RecordPatternInfo(const MemoryPatternGroup& mem_patterns, const
 // Record the actual allocated tensor in the device
 void MemoryInfo::RecordTensorDeviceAllocInfo(const OrtValueIndex idx, const OrtValue& value, const MapType& type) {
   if (tensor_alloc_info_map_.find(idx) == tensor_alloc_info_map_.end()) return;
+#ifdef USE_MIGRAPHX
+  if (!value.IsAllocated()) {
+    tensors_memory_info_map_.at(AllocPlan(idx)->location)[type].AddAllocMemory(idx, {0, 0});
+    return;
+  }
+#endif
   ORT_ENFORCE(value.IsTensor(), "Memory profiler only supports tensor type.");
   auto& tensor = value.Get<Tensor>();
   auto tensor_size_in_bytes = tensor.SizeInBytes();
--- a/onnxruntime/core/framework/session_state.cc
+++ b/onnxruntime/core/framework/session_state.cc
@@ -469,7 +469,13 @@ Status SessionState::PrepackConstantInitializedTensors(
 
               if (constant_initialized_tensors.count(ort_value_idx)) {
                 bool is_packed = false;
-                const Tensor& const_initialized_tensor = constant_initialized_tensors[ort_value_idx].Get<Tensor>();
+                const auto ort_value = constant_initialized_tensors[ort_value_idx];
+#ifdef USE_MIGRAPHX
+                if (!ort_value.IsAllocated()) {
+                  break;
+                }
+#endif
+                const Tensor& const_initialized_tensor = ort_value.Get<Tensor>();
 
                 auto iter = initializers_to_share_map.find(input_name);
                 bool is_shared_initializer = (iter != initializers_to_share_map.end());
--- a/onnxruntime/core/framework/session_state_utils.cc
+++ b/onnxruntime/core/framework/session_state_utils.cc
@@ -90,14 +90,21 @@ static common::Status DeserializeTensorProto(const Env& env, const std::basic_st
                   "DeserializeTensorProto() takes either pre-allocated buffer or an allocator!");
   }
 
+  const auto& memory_info = (alloc != nullptr) ? alloc->Info() : memory_buffer->GetAllocInfo();
+  const auto device = memory_info.device;
+
+#ifdef USE_MIGRAPHX
+  if (device.Type() == OrtDevice::GPU && device.Vendor() == OrtDevice::VendorIds::AMD) {
+    return common::Status::OK();
+  }
+#endif
+
+  static const auto default_cpu_device = OrtDevice();
+
+  // Get shape and type of the tensor, and allocate the empty tensor
   TensorShape tensor_shape = utils::GetTensorShapeFromTensorProto(tensor_proto);
   const DataTypeImpl* const type = DataTypeImpl::TensorTypeFromONNXEnum(tensor_proto.data_type())->GetElementType();
   Tensor tensor;
-
-  // Get shape and type of the tensor, and allocate the empty tensor
-  static const auto default_cpu_device = OrtDevice();
-  const auto& memory_info = (alloc != nullptr) ? alloc->Info() : memory_buffer->GetAllocInfo();
-  const auto device = memory_info.device;
 
   if (utils::HasExternalData(tensor_proto)) {
     auto external_data_loader = external_data_loader_mgr.GetExternalDataLoader(memory_info);
@@ -122,7 +129,6 @@ static common::Status DeserializeTensorProto(const Env& env, const std::basic_st
       if (utils::HasString(tensor_proto)) {
         return ORT_MAKE_STATUS(ONNXRUNTIME, FAIL, "string tensor is not supported for copying between allocators");
       }
-
       // deserialize to CPU first for non-CPU allocator, then copy to device
       // for external initializer load on non-CPU device:
       // 1. allocate memory on device - tensor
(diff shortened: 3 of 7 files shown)


psakhamoori:
@skottmckay @nieubank  Please help review this

psakhamoori:
@snnn Can you please help review this. Both Scott and Nick are OOF

TedThemistokleous:
@tianleiwu  @snnn  really urgent and important PR required for AMD based builds. We need this to reduce a bunch of memory overhead we're seeing. 

Can you enable CI and ensure we can get this in sooner than later?

snnn (commented):
Added comments

onnxruntime/core/framework/memory_info.cc
```diff
@@ -77,6 +76,12 @@ void MemoryInfo::RecordPatternInfo(const MemoryPatternGroup& mem_patterns, const
 // Record the actual allocated tensor in the device
 void MemoryInfo::RecordTensorDeviceAllocInfo(const OrtValueIndex idx, const OrtValue& value, const MapType& type) {
   if (tensor_alloc_info_map_.find(idx) == tensor_alloc_info_map_.end()) return;
+#ifdef USE_MIGRAPHX
```
snnn: We prefer to not have such macros in the core framework code so that the core framework can work with multiple EPs with problems.  For example, recently we partnered with WinML team to build a new release package  so that the same onnxruntime.dll can work with multiple different EPs for different hardware.

onnxruntime/core/providers/migraphx/migraphx_execution_provider.cc
```diff
@@ -1417,6 +1416,8 @@ Status MIGraphXExecutionProvider::Compile(const std::vector<FusedNodeAndGraph>&
           auto& name = it.first;
           auto& index = it.second;
           auto input_tensor = ctx.GetInput(index);
+          if (!input_tensor.HasValue())
```
snnn, line 1433: So, it only happens when MIGRAPHX EP is enabled? Could it be other cases?

TedThemistokleous: No. We only use this provider code for the MIGraphX EP. This is required since MIGraphX will preallocate I/O so that things are static on the GPU as part of our compile step.

onnxruntime/core/framework/memory_info.cc
TedThemistokleous: Isn't this macro dependent on whether the packages are building MIGraphX EP (AMD-GPU) then? 

Is there an alternative approach you'd like to see here? It doesn't seem you have provisions for EPs/Libraries that handle the allocation leading to this double memory allocation issue.

TedThemistokleous:
Comments inline

> This PR implements a workaround to reduce GPU memory allocations for the MIGraphX EP by preventing onnxruntime from allocating initializers that are already managed by MIGraphX. This is a reasonable approach to address the double allocation issue.
> 
> Here are a few points for consideration:
> 
>     * **Workaround:** The PR is described as a "workaround". It would be helpful to understand if there are any plans to address the root cause of this issue in the MIGraphX graph compiler in the future.
> 
Root cause would be a fundamental rewrite and larger change on our side. Is there any sort of provision or way you'd like us to handle the double alloc here then? This in.turn actually reduces the size of models and batches we can support due to us allocating 2x the memory for input data and weights 

Is the solution here to refactor part of the memory allocation and add a flag to the execution provider code that says "Onnxruntime does I/O alloc" similar to how there's a flag for NHWC memory layouts?  That would likely allow us to drop this in and scale to other EPs who want to follow similar suit along with not changing default behavior of the other EPs compiled in . You would just check if    EP->does_own_io_alloc() is true and stop the additional code from running.  

>     * **Testing:** How is this change tested? Are there any existing tests that cover this scenario, or should a new test be added to prevent regressions?
> 
Tested on our side and build using the unit test suite; onnxruntime_test_all and moving forward we're also going to be using Onnxruntime_provder_tests provided by the OnnxRT build. AMD then runs through our own set of models for validation to ensure no regressions in performance have occurred (accuracy and latency)

> 
> Overall, this is a good workaround for an important issue. I would be ready to approve it after getting some clarification on the points above.

No problem. Feel free to use this ticket or Email Artur and I and we can help work to a better solution if you have further concerns

onnxruntime/core/providers/migraphx/migraphx_execution_provider.cc
snnn: I didn't notice this file was in MigraphX's dir. Sorry.

onnxruntime/core/framework/memory_info.cc
snnn: Going forward we recommend all EPs using the new EP ABI that was just released in ONNX Runtime 1.23.0.  https://github.com/microsoft/onnxruntime/releases/tag/v1.23.0

And the core framework code should not have code like `#ifdef USE_CUDA/#ifdef USE_MIGRAPHX` otherwise the binary won't be compatible with multiple EPs.

causten:
How are the review going?  Any questions?

snnn:
@devang-ml , could you please help review these changes?

onnxruntime/core/framework/session_state.cc
```diff
@@ -478,7 +478,11 @@ Status SessionState::PrepackConstantInitializedTensors(
 
               if (constant_initialized_tensors.count(ort_value_idx)) {
                 bool is_packed = false;
-                const Tensor& const_initialized_tensor = constant_initialized_tensors[ort_value_idx].Get<Tensor>();
+                const auto ort_value = constant_initialized_tensors[ort_value_idx];
```
yuslepukhin, line 488: Need to add comment as to why this may happen.

onnxruntime/core/framework/session_state_utils.cc
```diff
@@ -90,15 +90,20 @@ static common::Status DeserializeTensorProto(const Env& env, const std::basic_st
                   "DeserializeTensorProto() takes either pre-allocated buffer or an allocator!");
   }
 
+  const auto& memory_info = (alloc != nullptr) ? alloc->Info() : memory_buffer->GetAllocInfo();
```
yuslepukhin, line 93: This code should not be reached, and it is not necessary. The function is only used in one place.

onnxruntime/core/framework/session_state_utils.cc
```diff
@@ -381,41 +385,46 @@ common::Status SaveInitializedTensors(
...
-                                             tensor_shape, use_device_allocator_for_initializers,
-                                             alloc));
-          ORT_RETURN_IF_ERROR(CopyTensorFromCPUToDevice(data_transfer_mgr,
-                                                        ort_value_from_graph.Get<Tensor>(),
-                                                        std::move(tensor), ort_value));
-        }
+      const auto& memory_info = (alloc != nullptr) ? alloc->Info() : memory_buffer->GetAllocInfo();
+
+      if (memory_info.device.Type() == OrtDevice::GPU && memory_info.device.Vendor() == OrtDevice::AMD) {
+        // Intentionally left empty
```
yuslepukhin, line 404: Add a comment here.

onnxruntime/core/framework/session_state_utils.cc
```diff
@@ -381,41 +385,46 @@ common::Status SaveInitializedTensors(
...
-                                               ->GetElementType();
-          Tensor tensor;
-          ORT_RETURN_IF_ERROR(AllocateTensor((memory_buffer) ? &*memory_buffer : nullptr, tensor, type,
-                                             tensor_shape, use_device_allocator_for_initializers,
-                                             alloc));
-          ORT_RETURN_IF_ERROR(CopyTensorFromCPUToDevice(data_transfer_mgr,
-                                                        ort_value_from_graph.Get<Tensor>(),
-                                                        std::move(tensor), ort_value));
-        }
+      const auto& memory_info = (alloc != nullptr) ? alloc->Info() : memory_buffer->GetAllocInfo();
```
yuslepukhin, line 401: If I am reading the code correctly, when pre-packing is disabled (rare but possible), the memory planner(allocator) enables memory patterns and that allocates a single big chunk of memory on a device, which would not be used in your case, but would occupy memory. Need to resolve this.

onnxruntime/core/framework/session_state_utils.cc
```diff
@@ -90,15 +90,20 @@ static common::Status DeserializeTensorProto(const Env& env, const std::basic_st
                   "DeserializeTensorProto() takes either pre-allocated buffer or an allocator!");
   }
 
+  const auto& memory_info = (alloc != nullptr) ? alloc->Info() : memory_buffer->GetAllocInfo();
+  const auto device = memory_info.device;
+
+  if (device.Type() == OrtDevice::GPU && device.Vendor() == OrtDevice::VendorIds::AMD) {
+    return Status::OK();
+  }
```
skottmckay, line 98: We need a better way to do this. We shouldn't have EP specific hardcoding in core ORT code. 

Does this potentially break WebGPU EP usage of an AMD GPU if MiGraphX isn't available? AFAIK it's perfectly valid for that EP to be using an AMD GPU device.

Another potential issue (lower probability) would be if the initializer is used by multiple nodes, and not all are assigned to MiGraphX.

Can you clarify how/when/where the EP is loading the initializers and making this call redundant?

TedThemistokleous: Is there a way to tell onnxRT core to not allocate input/outputs tensors then? The intent when using MIGraphX and MIGraphX EP is that an entire model is not split and run entirely through MIGraphX. MIgraphX (which is the backbone of AMD GPU EP and MIGraphX EP) also allocates the input and outputs which is why we're adding code to stop this double alloc of things like weights which limits the batch size available to an inference run.

yuslepukhin: We remove initializers that are not referenced by any node and do not load them from disk. If the EP is monolithic (compiled EP) and internalizes the initializers, then we potentially remove all the external references to the MIGraphX node, then all the weights within the protobuf and anything external would be removed from the model and no memory on device would be allocated.

TedThemistokleous: Would that be through an api call or some other mechanism? In the case of MIGraphX EP, we do this using an MIGraphX program and leverage the MIGraphX CPP API, which we prime the input/output tensors from the serialized onnx program read in before giving it to out compile() and run_async calls respectively. 

If possible, could you do a review of the MIGraphX EP in its current tips and suggest what to exclude/include in this? In most case we want the compile/node to be a monilith and have everything run on MIGraphX. The only case where I believe we need to allocate different inputs/split a graph is when an Onnx operator isn't in the support list and we partition the graph to run on say CPU EP (since no ROCm EP) 

Previously we added ROCm EP as a fallback mechanism to leverage the precompile CK kernels we've run HIPIFY from CUDA side and scale accordingly, but that would still be slower than compiling and optimizing the node that made it to the MIGraphX EP.

yuslepukhin: I can see that MIGraphX is a compile-time, monolithic, provider bridge-based EP within ORT.
When working with a monolithic EP, we typically construct (fuse) an artificial node for the entire subgraph handled by the EP. However, as of my last check, we retain all input references to that node, including initializers—potentially even those referenced by intermediate nodes within the subgraph.
This piece of code is very generic. We do not attempt to remove unnecessary references, and at this time, we cannot determine whether they can be removed. Perhaps we could introduce another virtual interface that indicates whether a particular initializer has been internalized and is no longer needed by the EP for the given subgraph.

TedThemistokleous: That would be ideal to be honest. As an MIGraphX program defines its input and outputs if there was a mechanism to mark a reference we're handling would be extremely useful. That would let us define where we want to get the data from within ORT when a graph is partitioned to say CPU

Right now when a graph is partitioned, we typically only see some nodes up until a point partitioned to MIGraphX and the rest fall onto CPU and prior would fallback to ROCm EP which had a kernel 9/10 and would give an inference. The issue is even though there may be one node thats not supported, the others are and we see a large spike in perf drop and in some cases (like in the contrib ops) a fail since not everything is supported on CPU EP right now.

skottmckay: >> Can you clarify how/when/where the EP is loading the initializers and making this call redundant?

> Is there a way to tell onnxRT core to not allocate input/outputs tensors then? The intent when using MIGraphX and MIGraphX EP is that an entire model is not split and run entirely through MIGraphX. MIgraphX (which is the backbone of AMD GPU EP and MIGraphX EP) also allocates the input and outputs which is why we're adding code to stop this double alloc of things like weights which limits the batch size available to an inference run.

ORT is loading initializers and putting them on the device associated with the EP that is running the node. Is the MiGraphX EP independently loading/copying initializers to the device? Still trying to understand what is causing the duplicate allocation here.

Also not quite understanding how this affects an inference run as all the initializer load/copy to device logic happens during model load and not during inferencing.

yuslepukhin: Can you, please, file an issue with a short repro so we can come up with a solid solution?

skottmckay: If this is compiling, are you returning the constant initializers as inputs in the ComputeCapability? Based on the CoreML EP that is optional as you can get any constant initializer that is in scope from the GraphViewer provided to the Compile.

Our partitioning utils support dropping constant initializers when creating the ComputeCapability.

https://github.com/microsoft/onnxruntime/blob/654137fab69f80265e5195deba209a635b4edb97/onnxruntime/core/providers/partitioning_utils.h#L92-L95

If the fused node for the compiled bits does not have the constant initializers as inputs they should be detected as unused by ORT on the next Graph::Resolve (unless they had multiple consumers), and shouldn't be around when we get to session state creation.

yuslepukhin: @TedThemistokleous Does the above answer the question? The examples of usage for the utility are in the code.

skottmckay: @apwojcik Can we get some input on these two questions?

> Does this potentially break WebGPU EP usage of an AMD GPU if MiGraphX isn't available? AFAIK it's perfectly valid for that EP to be using an AMD GPU device.

> If this is compiling, are you returning the constant initializers as inputs in the ComputeCapability?

tianleiwu:
@apwojcik,
Here is a prototype that might achieve similar goal with simple code change: https://github.com/microsoft/onnxruntime/commit/5219a86e1bf0995573c9bcea28aa20fc187c1185
Please take a look. I do not have AMD device, so I cannot test the code.

apwojcik:
> @apwojcik, Here is a prototype that might achieve similar goal with simple code change: [5219a86](https://github.com/microsoft/onnxruntime/commit/5219a86e1bf0995573c9bcea28aa20fc187c1185) Please take a look. I do not have AMD device, so I cannot test the code.

Thank you for the suggested change. We will test it out and let you know if it resolves the issue.

onnxruntime/core/framework/session_state_utils.cc
apwojcik: > @apwojcik Can we get some input on these two questions?
> 
> > Does this potentially break WebGPU EP usage of an AMD GPU if MiGraphX isn't available? AFAIK it's perfectly valid for that EP to be using an AMD GPU device.
> 
> > If this is compiling, are you returning the constant initializers as inputs in the ComputeCapability?

The only GPU EP from AMD that allocates tensors as device type GPU and vendor AMD is MIGraphX, so, in theory, it should not break the WebGPU EP. We are working on the new MIGraphX plugin EP. It might be that this fix is no longer needed. I will keep you posted.

apwojcik:
I am colsing this one in favour of https://github.com/microsoft/onnxruntime/pull/27551

PR #25804, opened 2025-08-21
webgpu: support bool type for Flatten and Unsqueeze operators

xhcao:

### DIFF
--- a/onnxruntime/core/providers/webgpu/tensor/flatten.cc
+++ b/onnxruntime/core/providers/webgpu/tensor/flatten.cc
@@ -15,7 +15,7 @@ ONNX_OPERATOR_VERSIONED_KERNEL_EX(
     kWebGpuExecutionProvider,
     (*KernelDefBuilder::Create())
         .Alias(0, 0)
-        .TypeConstraint("T", WebGpuSupportedNumberTypes())
+        .TypeConstraint("T", WebGpuSupportedNumberAndBoolTypes())
         .InputMemoryType(OrtMemTypeCPU, 1),
     Flatten);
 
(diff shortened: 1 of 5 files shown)


xhcao:
@guschmue @fs-eire In order to avoid `MemcpyToHost` on the `detr-resnet-50` model, and the root-cause is that 4 operators (Flatten, Unsqueeze, Gather, Expand) of webgpu EP does not support `tensor(bool)` type.
I had written 3 PRs (https://github.com/microsoft/onnxruntime/pull/25804, https://github.com/microsoft/onnxruntime/pull/25808,  https://github.com/microsoft/onnxruntime/pull/25811) to fix the issue. but I did not get performance improvement on the `detr-resnet-50` model after removing `MemcpyToHost`.
Is it still happy to continue to review the PRs, or suspend them?

fs-eire:
I looked at the model. Is the input `pixel_mask` expected to be on CPU or GPU?

xhcao:
> I looked at the model. Is the input `pixel_mask` expected to be on CPU or GPU?

I test the model performance using `onnxruntime_perf_test`, and do not specify the location of the input, only need to supply input shape. Are there big different performance for the input location?

fs-eire:
> > I looked at the model. Is the input `pixel_mask` expected to be on CPU or GPU?
> 
> I test the model performance using `onnxruntime_perf_test`, and do not specify the location of the input, only need to supply input shape. Are there big different performance for the input location?

yes. for example, for a typical language model we are currently putting the input IDs on CPU and the KV cache on GPU to minimize the CPU <-> GPU copy

- If an input is the data that used to calculate, that is usually on GPU (because we do heavy calculations on GPU)

- If an input is for index or something related to a tensor shape, it will need to be accessible for its value on CPU.

xhcao:
The PR fix the situation: node1 -> node2(Flatten(bool)) -> node3. If webgpu ep Flatten operator does not support the bool type tensor, node2 needs to be executed on CPU side, ort will insert `MemcpyToCPU` node before node2 and insert `MemcpyFromCPU` node after node2.

PR #25813, opened 2025-08-21
add c# interface for onnxruntime

wenjieh-amd:
### Description

add C# interface for VitisAI EP

### Motivation and Context

- in order to execute VitisAI EP in onnxruntime for C# users, we commit these changes

### DIFF
--- a/csharp/src/Microsoft.ML.OnnxRuntime/NativeMethods.shared.cs
+++ b/csharp/src/Microsoft.ML.OnnxRuntime/NativeMethods.shared.cs
@@ -417,7 +417,7 @@ namespace Microsoft.ML.OnnxRuntime
             OrtCreateEnvWithGlobalThreadPools = (DOrtCreateEnvWithGlobalThreadPools)Marshal.GetDelegateForFunctionPointer(api_.CreateEnvWithGlobalThreadPools, typeof(DOrtCreateEnvWithGlobalThreadPools));
             OrtCreateEnvWithCustomLoggerAndGlobalThreadPools = (DOrtCreateEnvWithCustomLoggerAndGlobalThreadPools)Marshal.GetDelegateForFunctionPointer(api_.CreateEnvWithCustomLoggerAndGlobalThreadPools, typeof(DOrtCreateEnvWithCustomLoggerAndGlobalThreadPools));
             OrtReleaseEnv = (DOrtReleaseEnv)Marshal.GetDelegateForFunctionPointer(api_.ReleaseEnv, typeof(DOrtReleaseEnv));
-            
+
             OrtEnableTelemetryEvents = (DOrtEnableTelemetryEvents)Marshal.GetDelegateForFunctionPointer(api_.EnableTelemetryEvents, typeof(DOrtEnableTelemetryEvents));
             OrtDisableTelemetryEvents = (DOrtDisableTelemetryEvents)Marshal.GetDelegateForFunctionPointer(api_.DisableTelemetryEvents, typeof(DOrtDisableTelemetryEvents));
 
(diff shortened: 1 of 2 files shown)


Mayaliming:
> @wenjieh-amd please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> ```
> 
> > Options:
> > 
> > 
> > ```
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > 
> > ```
> > ```
> 

@microsoft-github-policy-service agree [company="{amd}"]

wenjieh-amd:

@microsoft-github-policy-service agree [company="{AMD}"]

jywu-mysoft:
can't we use the IHV agnostic interface https://github.com/microsoft/onnxruntime/blob/0c25d1bb2009c78e3a87155c4071faadca71555f/csharp/src/Microsoft.ML.OnnxRuntime/SessionOptions.shared.cs#L468 
rather than adding a new api ?

wenjieh-amd:
> can't we use the IHV agnostic interface
> 
> https://github.com/microsoft/onnxruntime/blob/0c25d1bb2009c78e3a87155c4071faadca71555f/csharp/src/Microsoft.ML.OnnxRuntime/SessionOptions.shared.cs#L468
> 
> 
> rather than adding a new api ?

image
but this interface doesn't work

wenjieh-amd:
> can't we use the IHV agnostic interface
> 
> https://github.com/microsoft/onnxruntime/blob/0c25d1bb2009c78e3a87155c4071faadca71555f/csharp/src/Microsoft.ML.OnnxRuntime/SessionOptions.shared.cs#L468
> 
> 
> rather than adding a new api ?

You know, NVDA has its CUDA interface, MSFT has its own DML interface, I just think AMD should has its VitisAI interface too. 
The code has a comment:
 "// New EPs should use OrtApis, preferably leveraging the generic SessionOptionsAppendExecutionProvider"
 but VAI is not a new EP.

Besides, the test on AppendExecutionProvider(string providerName, Dictionary<string, string> providerOptions = null) interface passes too in this PR.

So, I lean forward keeping AppendExecutionProvider_VitisAI this interface instead of removing it. 

looking forward to your reply

wenjieh-amd:
> can't we use the IHV agnostic interface
> 
> https://github.com/microsoft/onnxruntime/blob/0c25d1bb2009c78e3a87155c4071faadca71555f/csharp/src/Microsoft.ML.OnnxRuntime/SessionOptions.shared.cs#L468
> 
> 
> rather than adding a new api ?

You know, NVDA has its CUDA interface, MSFT has its own DML interface, I just think AMD should has its VitisAI interface too.
The code has a comment:
"// New EPs should use OrtApis, preferably leveraging the generic SessionOptionsAppendExecutionProvider"
but VAI is not a new EP.

Besides, the test on AppendExecutionProvider(string providerName, Dictionary<string, string> providerOptions = null) interface passes too in this PR.

So, I lean forward keeping AppendExecutionProvider_VitisAI this interface instead of removing it.

looking forward to your reply

yuslepukhin:
> > can't we use the IHV agnostic interface
> > https://github.com/microsoft/onnxruntime/blob/0c25d1bb2009c78e3a87155c4071faadca71555f/csharp/src/Microsoft.ML.OnnxRuntime/SessionOptions.shared.cs#L468
> > 
> > rather than adding a new api ?
> 
> You know, NVDA has its CUDA interface, MSFT has its own DML interface, I just think AMD should has its VitisAI interface too. The code has a comment: "// New EPs should use OrtApis, preferably leveraging the generic SessionOptionsAppendExecutionProvider" but VAI is not a new EP.
> 
> Besides, the test on AppendExecutionProvider(string providerName, Dictionary<string, string> providerOptions = null) interface passes too in this PR.
> 
> So, I lean forward keeping AppendExecutionProvider_VitisAI this interface instead of removing it.
> 
> looking forward to your reply

This has escaped my attention. We want to move aware from the practice of adding EP specific endpoint which was necessary under the old schema. Going forward we want this to be the case.

All you need to do is make sure this works with your provider.

wenjieh-amd:
> > > can't we use the IHV agnostic interface
> > > https://github.com/microsoft/onnxruntime/blob/0c25d1bb2009c78e3a87155c4071faadca71555f/csharp/src/Microsoft.ML.OnnxRuntime/SessionOptions.shared.cs#L468
> > > 
> > > rather than adding a new api ?
> > 
> > 
> > You know, NVDA has its CUDA interface, MSFT has its own DML interface, I just think AMD should has its VitisAI interface too. The code has a comment: "// New EPs should use OrtApis, preferably leveraging the generic SessionOptionsAppendExecutionProvider" but VAI is not a new EP.
> > Besides, the test on AppendExecutionProvider(string providerName, Dictionary<string, string> providerOptions = null) interface passes too in this PR.
> > So, I lean forward keeping AppendExecutionProvider_VitisAI this interface instead of removing it.
> > looking forward to your reply
> 
> This has escaped my attention. We want to move aware from the practice of adding EP specific endpoint which was necessary under the old schema. Going forward we want this to be the case.
> 
> All you need to do is make sure this works with your provider.

I understand your concern, but as you know VitisAI is not a new EP, it has been in onnxruntime for a long time.
C++ has an AppendExecutionProvider_VitisAI interface, once the C++ devoloper switch to C#, they may be confused about why this interface doesn't work anymore.
Besides, I've already tested the AppendExecutionProvider interface, and the test passes.
that's why I prefer keeping AppendExecutionProvider_VitisAI
what's your opinion about it ?
looking forward to your reply

yuslepukhin:
The consensus is not to add a new entry point. This is the way forward to cutdown on the binary size and unify the approach. We may deprecate other APIs.

wenjieh-amd:
thanks for your review, since we would not like to introduce the new interface AppendExecutionProvider_VitisAI, this pr is supposed to be closed.
I have already tested the AppendExecutionProvider("VitisAI"), and test passed.
thanks MSFT's help

PR #25814, opened 2025-08-21
Fixes for DynamicQuantizeMatMul and Attention3D tests

JonathanC-ARM:
### Description
This change fixes correctness issues in two areas that were causing failures in onnxruntime_test_all:

- DynamicQuantizeMatMul.WithConstantBInputs
- AttentionTest.Attention3DDefault
- AttentionTest.Attention3DWithPastAndPresentQkMatmul

What was wrong and how it’s fixed
1) DynamicQuantizeMatMul.WithConstantBInputs
- Root cause: The Kleidi dynamic quantization GEMM path could be selected even when the B scales contained values such as (zero, negative, or non-finite). That violates kernel assumptions and can lead to incorrect results.
- Fix: In `onnxruntime/contrib_ops/cpu/quantization/dynamic_quantize_matmul.cc`, we now explicitly validate that all B scales are finite and strictly positive before enabling the Kleidi/MLAS dynamic path. If any scale is invalid, we disable that path.

2) Attention tests (Attention3DDefault, Attention3DWithPastAndPresentQkMatmul)
- Root causes in `onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp`:
  - Incorrect handling of GEMM corner cases for alpha/beta and K==0 (e.g., not respecting C = beta*C when alpha==0 or K==0).
  - Unnecessary or premature fallbacks for small shapes.
- Fixes:
  - Add early-outs for degenerate sizes: if M==0 or N==0, return handled.
  - Correctly implement alpha/beta semantics:

### DIFF
--- a/onnxruntime/contrib_ops/cpu/quantization/dynamic_quantize_matmul.cc
+++ b/onnxruntime/contrib_ops/cpu/quantization/dynamic_quantize_matmul.cc
@@ -206,10 +206,10 @@ class DynamicQuantizeMatMul final : public MatMulIntegerToFloatBase {
         const float* bs_data = b_scale_tensor->Data<float>();
         const size_t bs_size = static_cast<size_t>(b_scale_tensor->Shape().Size());
         for (size_t i = 0; i < bs_size; ++i) {
-            const float s = bs_data[i];
-            if (!std::isfinite(s) || s <= 0.0f) {
-                can_use_dynamic_quant_mlas_ = false;
-                break;
+          const float s = bs_data[i];
+          if (!std::isfinite(s) || s <= 0.0f) {
+            can_use_dynamic_quant_mlas_ = false;
+            break;
           }
         }
       }
@@ -406,8 +406,9 @@ Status DynamicQuantizeMatMul::Compute(OpKernelContext* ctx) const {
     std::vector<MLAS_GEMM_DYN_QUANT_DATA_PARAMS> gemm_data_vec(num_gemms);
 
     for (size_t gemm_idx = 0; gemm_idx < num_gemms; gemm_idx++) {
+
       auto& params = gemm_data_vec[gemm_idx];
-      params.A = reinterpret_cast<const float*>(a_data + helper.LeftOffsets()[gemm_idx]);
+      params.A = reinterpret_cast<const float*>(a_data) + helper.LeftOffsets()[gemm_idx];
       params.lda = gemm_shape.K;
       params.PackedB = packed_b_.get();
       params.C = y_data + helper.OutputOffsets()[gemm_idx];
--- a/onnxruntime/contrib_ops/cpu/quantization/dynamic_quantize_matmul.cc
+++ b/onnxruntime/contrib_ops/cpu/quantization/dynamic_quantize_matmul.cc
@@ -200,6 +200,20 @@ class DynamicQuantizeMatMul final : public MatMulIntegerToFloatBase {
 
       can_use_dynamic_quant_mlas_ = (!b_quantization_might_be_asymmetric && b_scale_available);
 
+      // Kleidi dynamic path requires strictly positive, finite scales.
+      // Disable if any invalid scale is detected.
+      if (can_use_dynamic_quant_mlas_) {
+        const float* bs_data = b_scale_tensor->Data<float>();
+        const size_t bs_size = static_cast<size_t>(b_scale_tensor->Shape().Size());
+        for (size_t i = 0; i < bs_size; ++i) {
+            const float s = bs_data[i];
+            if (!std::isfinite(s) || s <= 0.0f) {
+                can_use_dynamic_quant_mlas_ = false;
+                break;
+          }
+        }
+      }
+
       // Currently, MlasDynamicQGemmBatch() and associated functions require SME or else they are no-ops.
       // We check that here too before attempting to use them.
       if (!CPUIDInfo::GetCPUIDInfo().HasArm_SME()) {
--- a/onnxruntime/contrib_ops/cpu/quantization/dynamic_quantize_matmul.cc
+++ b/onnxruntime/contrib_ops/cpu/quantization/dynamic_quantize_matmul.cc
@@ -203,14 +203,13 @@ class DynamicQuantizeMatMul final : public MatMulIntegerToFloatBase {
       // Kleidi dynamic path requires strictly positive, finite scales.
       // Disable if any invalid scale is detected.
       if (can_use_dynamic_quant_mlas_) {
-        const float* bs_data = b_scale_tensor->Data<float>();
-        const size_t bs_size = static_cast<size_t>(b_scale_tensor->Shape().Size());
-        for (size_t i = 0; i < bs_size; ++i) {
-          const float s = bs_data[i];
-          if (!std::isfinite(s) || s <= 0.0f) {
-            can_use_dynamic_quant_mlas_ = false;
-            break;
-          }
+        const auto bs = b_scale_tensor->DataAsSpan<float>();
+        const bool has_invalid =
+            std::any_of(bs.begin(), bs.end(),
+                        [](float s) { return !std::isfinite(s) || s <= 0.0f; });
+
+        if (has_invalid) {
+          can_use_dynamic_quant_mlas_ = false;
         }
       }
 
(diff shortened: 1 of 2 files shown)


JonathanC-ARM:

onnxruntime/contrib_ops/cpu/quantization/dynamic_quantize_matmul.cc
```diff
@@ -393,7 +407,7 @@ Status DynamicQuantizeMatMul::Compute(OpKernelContext* ctx) const {
 
     for (size_t gemm_idx = 0; gemm_idx < num_gemms; gemm_idx++) {
       auto& params = gemm_data_vec[gemm_idx];
-      params.A = reinterpret_cast<const float*>(a_data + helper.LeftOffsets()[gemm_idx]);
+      params.A = params.A = a_data + helper.LeftOffsets()[gemm_idx];
```
edgchen1: I think `a_data` would need to be a `const float*` here. also, the `params.A = params.A =` assignment is redundant.

JonathanC-ARM: The params.A = params.A was a typo on my part, but I have since change the pointer logic in the latest push.

onnxruntime/contrib_ops/cpu/quantization/dynamic_quantize_matmul.cc
```diff
@@ -392,8 +406,9 @@ Status DynamicQuantizeMatMul::Compute(OpKernelContext* ctx) const {
     std::vector<MLAS_GEMM_DYN_QUANT_DATA_PARAMS> gemm_data_vec(num_gemms);
 
     for (size_t gemm_idx = 0; gemm_idx < num_gemms; gemm_idx++) {
+
       auto& params = gemm_data_vec[gemm_idx];
-      params.A = reinterpret_cast<const float*>(a_data + helper.LeftOffsets()[gemm_idx]);
+      params.A = reinterpret_cast<const float*>(a_data) + helper.LeftOffsets()[gemm_idx];
```
edgchen1: nit: can we just make `a_data` a float* to begin with?

```c++
const float* a_data = ctx->Input<Tensor>(IN_A)->Data<float>();
```

onnxruntime/contrib_ops/cpu/quantization/dynamic_quantize_matmul.cc
```diff
@@ -200,6 +200,20 @@ class DynamicQuantizeMatMul final : public MatMulIntegerToFloatBase {
 
       can_use_dynamic_quant_mlas_ = (!b_quantization_might_be_asymmetric && b_scale_available);
 
+      // Kleidi dynamic path requires strictly positive, finite scales.
+      // Disable if any invalid scale is detected.
+      if (can_use_dynamic_quant_mlas_) {
+        const float* bs_data = b_scale_tensor->Data<float>();
```
edgchen1: nit: consider using Tensor::DataAsSpan() and something like std::any_of. can avoid size_t cast and raw loop.

JonathanC-ARM: I've changed this to use DataAsSpan and Any of as suggested

onnxruntime/contrib_ops/cpu/quantization/dynamic_quantize_matmul.cc
JonathanC-ARM: So this is changed now the type is const float* and we no longer require the reinterpret_cast

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
```diff
@@ -153,30 +153,41 @@ ArmKleidiAI::MlasGemmBatch(
...
-    if(TransA == CblasTrans)
-    {
-        return false;
+    if (M == 0 || N == 0) {
+        return true;
     }
-    if (TransA == CblasNoTrans && K == 0) {
-        if (Data->beta != 1.0f) {
+
+    if (Data->alpha == 0.0f) {
```
edgchen1: nit: can we merge this with the condition below? the contents look the same.

`if (Data->alpha == 0.0f || K == 0) {`

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
```diff
@@ -185,11 +196,9 @@ ArmKleidiAI::MlasGemmBatch(
         auto m_step = kai_get_m_step_matmul_clamp_f32_f32p2vlx1_f32p2vlx1biasf32_sme2_mopa();
         auto n_step = kai_get_n_step_matmul_clamp_f32_f32p2vlx1_f32p2vlx1biasf32_sme2_mopa();
 
-        if (M < m_step || N < n_step) {
-            if (GetMlasPlatform().MlasGemmBatchOverride != ArmKleidiAI::MlasGemmBatch){
-                //Fallback to MLAS
-                return false;
-            }
+        if ((M < m_step && N < n_step) && !Data->BIsPacked) {
```
edgchen1: why did it change from `M < m_step || N < n_step` to `M < m_step && N < n_step`? what do m_step and n_step represent?

nit: there are unnecessary parentheses in the expression `(M < m_step && N < n_step) && !Data->BIsPacked`.

JonathanC-ARM: m_step and n_step are values that represent the number of elements in terms of columns and rows that the kernel being invoked will operate on. 

We have noticed that for very small values where M and N are both below the step values. the overhead of performing the packing negates the speed up of the kernel. So in this case we have said if they are both small and the data isn't already prepacked we should fallback. The reason we have said and this time also and not or is that if one of the values is disproportionately larger than the other it will be worth the overhead i.e. Big M and small N or vice versa

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
JonathanC-ARM: Yeah you are right, will fix this

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
edgchen1: got it, thanks for the explanation

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
JonathanC-ARM: Fixed in latest push

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
```diff
@@ -328,21 +321,37 @@ ArmKleidiAI::MlasGemmBatch(
...
-            }else {
-                // apply alpha scaling and beta to output files
-                for (size_t i = 0; i <  TileSizeM; ++i) {
-                    for (size_t j = 0; j < TileSizeN; ++j) {
-                        const size_t idx = i * TileSizeN + j;
-                        const size_t dst_idx = i * Data[BIdx].ldc + j;
-
-                        float ab = temp_tile[idx];
-                        float c_orig = dst_tile[dst_idx];
+                return true;
```
edgchen1: nit: MlasTrySimpleParallel() takes a `std::function<void(std::ptrdiff_t tid)>` for the parallel work

```suggestion
                return;
```

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
```diff
@@ -328,21 +321,37 @@ ArmKleidiAI::MlasGemmBatch(
...
+                    } else if (alpha == 1.0f) {
+                        dst_tile[dst_idx] = ab + beta * c_orig;
+                    } else if (beta == 0.0f) {
+                        dst_tile[dst_idx] = alpha * ab;
+                    } else {
+                        dst_tile[dst_idx] = alpha * ab + beta * c_orig;
                     }
                 }
             }
+            return true;
```
edgchen1: nit: don't return a value from the lambda passed to MlasTrySimpleParallel
```suggestion
```

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
JonathanC-ARM: Resolved

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
JonathanC-ARM: Resolved this also

edgchen1 (approved):

snnn:

PR #25817, opened 2025-08-21
[VitisAI EP] Add GetLibraryPath API to expose execution provider plugin library paths

psakhamoori:
This PR adds functionality to expose execution provider plugin library filesystem paths through new APIs:

**Changes:**
- Add virtual `GetLibraryPath()` method to `EpLibrary` base class
- Implement `GetLibraryPath()` override in `EpLibraryPlugin` to return `library_path_`
- Add `Environment::GetExecutionProviderLibraryPath()` public API for thread-safe access to EP library paths
- Enable external applications to resolve custom ops libraries relative to EP plugin locations

**Use Case:**
Allows downstream projects (like win-onnxruntime-genai) to intelligently resolve relative custom ops library paths by using the execution provider plugin location as a base directory.

**API Usage:**
```cpp
onnxruntime::Environment& env = GetEnvironment();
std::filesystem::path ep_lib_path;
auto status = env.GetExecutionProviderLibraryPath("QNNExecutionProvider", ep_lib_path);

### DIFF
--- a/include/onnxruntime/core/session/environment.h
+++ b/include/onnxruntime/core/session/environment.h
@@ -154,6 +154,11 @@ class Environment {
   const DataTransferManager& GetDataTransferManager() const {
     return data_transfer_mgr_;
   }
+
+  // Get the filesystem path of the library registered for the given EP registration name.
+  // Returns NOT_FOUND/FAIL if the EP is not registered or does not expose a library path.
+  Status GetExecutionProviderLibraryPath(const std::string& registration_name,
+                                         std::filesystem::path& out) const;
 #endif  // !defined(ORT_MINIMAL_BUILD)
 
   // return a shared allocator from a plugin EP or custom allocator added with RegisterAllocator
--- a/onnxruntime/core/session/environment.cc
+++ b/onnxruntime/core/session/environment.cc
@@ -797,6 +797,31 @@ Environment::EpInfo::~EpInfo() {
   }
 }
 
+Status Environment::GetExecutionProviderLibraryPath(const std::string& registration_name,
+                                                   std::filesystem::path& out) const {
+  std::lock_guard<std::mutex> lock{mutex_};
+
+  auto it = ep_libraries_.find(registration_name);
+  if (it == ep_libraries_.end()) {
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: include/onnxruntime/core/session/environment.h [hunk 1]
```diff
@@ -154,6 +154,11 @@ class Environment {
   const DataTransferManager& GetDataTransferManager() const {
     return data_transfer_mgr_;
   }
+
+  // Get the filesystem path of the library registered for the given EP registration name.
+  // Returns NOT_FOUND/FAIL if the EP is not registered or does not expose a library path.
+  Status GetExecutionProviderLibraryPath(const std::string& registration_name,
```
chrisdMSFT, line 160: Consider adding C++ and C# versions of this method too.

onnxruntime/core/session/environment.cc
```diff
@@ -797,6 +797,31 @@ Environment::EpInfo::~EpInfo() {
...
+  }
+
+  auto& ep_info = it->second;
+  if (!ep_info || !ep_info->library) {
+    return ORT_MAKE_STATUS(ONNXRUNTIME, FAIL, "Execution provider library registered under '", registration_name,
+                           "' has no EpLibrary instance.");
+  }
+
+  const auto& lib_path = ep_info->library->GetLibraryPath();
+  if (lib_path.empty()) {
```
chrisdMSFT, line 816: I appreciate the rich failure strings. 

Does the caller need to pragmatically distinguish between the failure modes?

chrisdMSFT: Looking at the usage, I think this behavior is fine for now.

edgchen1:
> Add Environment::GetExecutionProviderLibraryPath() public API for thread-safe access to EP library paths

The onnxruntime public C API is defined in include/onnxruntime/core/session/onnxruntime_c_api.h.

Despite the path, include/onnxruntime/core/session/environment.h is not a public header.

psakhamoori:
New implementation using existing "LoadPluginOrProviderBridge" https://github.com/microsoft/onnxruntime/pull/25830

PR #25821, opened 2025-08-21
Fix CPU EP Tile 0D overvalidation

### LINKED ISSUE #11523: Tile fails for scalars on CPU
**Describe the bug**
Looking at the code, it seems this works on GPU, but on CPU tile() will give an error if called with a scalar:

  https://github.com/microsoft/onnxruntime/blob/master/onnxruntime/core/providers/cpu/tensor/tile.cc#L149

The spec is clear on this, allowing any shape: "Input tensor of any shape."

  https://github.com/onnx/onnx/blob/main/docs/Operators.md#inputs-124

This is an edge case which would also require the repeats vector to be empty, but there is nothing special about it, it functions consistently with all other cases and would be expected to be another way of spelling the identity function. I didn't run this code, but do not see this error on GPU, so presumably it works correctly there.

**Urgency**
It's a bug.

**System information**
Any CPU

**To Reproduce**
Do tile on a scalar.

**Expected behavior**
Should be the identity.

**Screenshots**

**Additional context**

fdwr:
### Description
Fixes #11523. Scalars should just be nops for tile. Simply removing the overvalidation lets the case work.

### Motivation and Context
Conformance with expectation.

*We should probably sweep more broadly across all related ops to verify they behave rationally...*


fdwr:
Seems the WebGPU EP may not be prepared correctly for this possibility:

```
2025-08-30 02:06:27.346 onnxruntime_test_all[71138:229996] 2025-08-30 02:06:27.346092 [E:onnxruntime:Tile, sequential_executor.cc:572 ExecuteKernel] Non-zero status code returned while running Tile node. Name:'node1' Status Message: /Users/runner/work/onnxruntime/onnxruntime/onnxruntime/core/providers/webgpu/program.cc:46 onnxruntime::webgpu::ProgramUniformVariableValue::ProgramUniformVariableValue(ProgramUniformVariableDataType, const void *, size_t, size_t) length > 0 was false. number of element of uniform variable must be greater than 0
2: /Users/runner/work/onnxruntime/onnxruntime/onnxruntime/test/providers/base_tester.cc:342: Failure
2: Expected equality of these values:
2:   expect_result
2:     Which is: 4-byte object <00-00 00-00>
2:   ExpectResult::kExpectFailure
2:     Which is: 4-byte object <01-00 00-00>
```

https://github.com/microsoft/onnxruntime/actions/runs/17337504737/job/49226161742?pr=25821#step:9:17441

fdwr:
@hariharans29 Have a recommended reviewer for this one?

hariharans29:
LGTM.

The spec says: "Input tensor of any shape." which I guess technically means a scalar is allowed. Curious - does numpy Tile allow scalars since the spec quotes Numpy ?

onnxruntime/test/providers/cpu/tensor/tile_op_test.cc
```diff
@@ -142,6 +144,11 @@ void RunTestWrapper() {
   RunTest<T>({2, 1, 3}, {2, 2, 1});
   RunTest<T>({2, 1, 3}, {2, 2, 1}, true);
 
+#if !defined(USE_WEBGPU)
+  // Tile0D (nop)
```
hariharans29, line 151: Nit: Maybe add a comment/provide context saying WebGPU is not prepared to handle this (for whatever reason) like the PR comment says ?

hariharans29 (dismissed):

fdwr:
> Curious - does numpy Tile allow scalars since the spec quotes Numpy ?

Yep.

```python
# pip freeze numpy==2.3.1

import numpy
x = numpy.tile(42.0, [])
print("value:", x)
print("shape:", x.shape)
print("dtype:", x.dtype)

# value: 42.0
# shape: ()
# dtype: float32
```

hariharans29 (approved):

PR #25827, opened 2025-08-22
add csharp interface for vitisai

wenjieh-amd:
### Description

add csharp interface for vitisai

### DIFF
--- a/csharp/src/Microsoft.ML.OnnxRuntime/NativeMethods.shared.cs
+++ b/csharp/src/Microsoft.ML.OnnxRuntime/NativeMethods.shared.cs
@@ -597,6 +597,8 @@ namespace Microsoft.ML.OnnxRuntime
                 api_.SessionOptionsAppendExecutionProvider_CUDA, typeof(DSessionOptionsAppendExecutionProvider_CUDA));
             SessionOptionsAppendExecutionProvider_CUDA_V2 = (DSessionOptionsAppendExecutionProvider_CUDA_V2)Marshal.GetDelegateForFunctionPointer(
                 api_.SessionOptionsAppendExecutionProvider_CUDA_V2, typeof(DSessionOptionsAppendExecutionProvider_CUDA_V2));
+            SessionOptionsAppendExecutionProvider_VitisAI = (DSessionOptionsAppendExecutionProvider_VitisAI)Marshal.GetDelegateForFunctionPointer(
+                api_.SessionOptionsAppendExecutionProvider_VitisAI, typeof(DSessionOptionsAppendExecutionProvider_VitisAI));
             OrtCreateCUDAProviderOptions = (DOrtCreateCUDAProviderOptions)Marshal.GetDelegateForFunctionPointer(api_.CreateCUDAProviderOptions, typeof(DOrtCreateCUDAProviderOptions));
             OrtUpdateCUDAProviderOptions = (DOrtUpdateCUDAProviderOptions)Marshal.GetDelegateForFunctionPointer(api_.UpdateCUDAProviderOptions, typeof(DOrtUpdateCUDAProviderOptions));
             OrtGetCUDAProviderOptionsAsString = (DOrtGetCUDAProviderOptionsAsString)Marshal.GetDelegateForFunctionPointer(api_.GetCUDAProviderOptionsAsString, typeof(DOrtGetCUDAProviderOptionsAsString));
@@ -1311,6 +1313,20 @@ namespace Microsoft.ML.OnnxRuntime
             IntPtr /*(const OrtTensorRTProviderOptionsV2*)*/ trtProviderOptions);
 
         public static DSessionOptionsAppendExecutionProvider_TensorRT_V2 SessionOptionsAppendExecutionProvider_TensorRT_V2;
+
+        /// <summary>
+        /// Append a VitisAI EP instance (configured based on given provider options) to the native OrtSessionOptions instance
+        /// </summary>
+        /// <param name="options">Native OrtSessionOptions instance</param>
+        /// <param name="keys">Native keys instance</param>
+        /// <param name="values">Native values instance</param>
+        /// <param name="numEntries">Native numEntries instance</param>
+        [UnmanagedFunctionPointer(CallingConvention.Winapi)]
+        public delegate IntPtr /*(OrtStatus*)*/ DSessionOptionsAppendExecutionProvider_VitisAI(
+            IntPtr /*(OrtSessionOptions*)*/ options,
+            IntPtr keys, IntPtr values, UIntPtr numEntries);
+
+        public static DSessionOptionsAppendExecutionProvider_VitisAI SessionOptionsAppendExecutionProvider_VitisAI;
 
         /// <summary>
         /// Append a CUDA EP instance (configured based on given provider options) to the native OrtSessionOptions instance
--- a/csharp/src/Microsoft.ML.OnnxRuntime/SessionOptions.shared.cs
+++ b/csharp/src/Microsoft.ML.OnnxRuntime/SessionOptions.shared.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Runtime.InteropServices;
 using System.Text;
+using System.Linq;
 
 namespace Microsoft.ML.OnnxRuntime
 {
@@ -254,6 +255,56 @@ namespace Microsoft.ML.OnnxRuntime
             throw new NotSupportedException("The CUDA Execution Provider is not supported in this build");
 #else
             NativeApiStatus.VerifySuccess(NativeMethods.SessionOptionsAppendExecutionProvider_CUDA_V2(handle, cudaProviderOptions.Handle));
+#endif
+        }
+
+        /// <summary>
+        /// Append a VitisAI EP instance (configured based on given provider options) to the native OrtSessionOptions instance
+        /// </summary>
+        /// <param name="options">Native OrtSessionOptions instance</param>
+        /// <param name="keys">Native keys instance</param>
+        /// <param name="values">Native values instance</param>
+        /// <param name="numEntries">Native numEntries instance</param>
+        public void AppendExecutionProvider_VitisAI(Dictionary<string, string> config)
+        {
+#if __MOBILE__
+            throw new NotSupportedException("The VitisAI Execution Provider is not supported in this build");
+#else
+            int count = config.Count;
+            IntPtr[] keyPtrs = new IntPtr[count];
+            IntPtr[] valuePtrs = new IntPtr[count];
+
+            string[] keys = config.Keys.ToArray();
+            string[] values = config.Values.ToArray();
+
+            for (int i = 0; i < count; ++i) {
+                keyPtrs[i] = Marshal.StringToHGlobalAnsi(keys[i]);
+                valuePtrs[i] = Marshal.StringToHGlobalAnsi(values[i]);
+            }
+
+            IntPtr keysNative = Marshal.AllocHGlobal(IntPtr.Size * count);
+            IntPtr valuesNative = Marshal.AllocHGlobal(IntPtr.Size * count);
+
+            Marshal.Copy(keyPtrs, 0, keysNative, count);
+            Marshal.Copy(valuePtrs, 0, valuesNative, count);
+
+            try
+            {
+                UIntPtr numKeys = new UIntPtr((uint)count);
+                NativeApiStatus.VerifySuccess(
+                    NativeMethods.SessionOptionsAppendExecutionProvider_VitisAI(
+                        handle, keysNative, valuesNative, numKeys));
+            }
+            finally
+            {
+                for (int i = 0; i < count; ++i)
+                {
+                    Marshal.FreeHGlobal(keyPtrs[i]);
+                    Marshal.FreeHGlobal(valuePtrs[i]);
+                }
+                Marshal.FreeHGlobal(keysNative);
+                Marshal.FreeHGlobal(valuesNative);
+            }
 #endif
         }
 


jywu-mysoft:
just noticed. don't post PR's directly to rel-1.23.0 , they should go into main.

csharp/src/Microsoft.ML.OnnxRuntime/SessionOptions.shared.cs
```diff
@@ -257,6 +258,56 @@ public void AppendExecutionProvider_CUDA(OrtCUDAProviderOptions cudaProviderOpti
...
+
+            string[] keys = config.Keys.ToArray();
+            string[] values = config.Values.ToArray();
+
+            for (int i = 0; i < count; ++i) {
+                keyPtrs[i] = Marshal.StringToHGlobalAnsi(keys[i]);
+                valuePtrs[i] = Marshal.StringToHGlobalAnsi(values[i]);
+            }
+
+            IntPtr keysNative = Marshal.AllocHGlobal(IntPtr.Size * count);
```
yuslepukhin: Global allocations are slow. The only advantage is that they do not need to be pinned.
Pinning is necessary so the allocations is not moved or GCed while being used by native calls.

However, in this case keysNatives aren valuesNative are automatically pinned when being passed to native method
because IntPtr are considered a blittable type so they donot need to be allocated and copied separatly, you can pass keysPtrs and valuesPtrs directly
to the native call even though those are managed arrays. 

So you will not need to allocate and copy to *Native arrays. This will also take care of the memory leak in case Copy() throws.

The issue here, is that we need to convert strings to UTF-8 and not ANSI. Use utility functions to accomplish that.
Those will produce managed byte[] arrays with zero terminator containing UTF-8 bytes.

You will need to PIN those and obtain IntPtr which can be placed to keysPtrs and valuesPtrs.

```.cs
byte[] buffer = new byte[1024];
GCHandle handle = GCHandle.Alloc(buffer, GCHandleType.Pinned);
IntPtr ptr = handle.AddrOfPinnedObject();
// Use ptr...
handle.Free();
```


You will need to unpin those in finally block (Free()). This way Marshal functionality is not longer needed.

csharp/src/Microsoft.ML.OnnxRuntime/SessionOptions.shared.cs
```diff
@@ -257,6 +258,56 @@ public void AppendExecutionProvider_CUDA(OrtCUDAProviderOptions cudaProviderOpti
...
         }
 
+        /// <summary>
+        /// Append a VitisAI EP instance (configured based on given provider options) to the native OrtSessionOptions instance
+        /// </summary>
+        /// <param name="options">Native OrtSessionOptions instance</param>
+        /// <param name="keys">Native keys instance</param>
+        /// <param name="values">Native values instance</param>
+        /// <param name="numEntries">Native numEntries instance</param>
+        public void AppendExecutionProvider_VitisAI(Dictionary<string, string> config)
```
yuslepukhin: We need to add a test to this functionality. #Resolved

csharp/src/Microsoft.ML.OnnxRuntime/SessionOptions.shared.cs
wenjieh-amd: > Global allocations are slow. The only advantage is that they do not need to be pinned. Pinning is necessary so the allocations is not moved or GCed while being used by native calls.
> 
> However, in this case keysNatives aren valuesNative are automatically pinned when being passed to native method because IntPtr are considered a blittable type so they donot need to be allocated and copied separatly, you can pass keysPtrs and valuesPtrs directly to the native call even though those are managed arrays.
> 
> So you will not need to allocate and copy to *Native arrays. This will also take care of the memory leak in case Copy() throws.
> 
> The issue here, is that we need to convert strings to UTF-8 and not ANSI. Use utility functions to accomplish that. Those will produce managed byte[] arrays with zero terminator containing UTF-8 bytes.
> 
> You will need to PIN those and obtain IntPtr which can be placed to keysPtrs and valuesPtrs.
> 
> ```cs
> byte[] buffer = new byte[1024];
> GCHandle handle = GCHandle.Alloc(buffer, GCHandleType.Pinned);
> IntPtr ptr = handle.AddrOfPinnedObject();
> // Use ptr...
> handle.Free();
> ```
> 
> You will need to unpin those in finally block (Free()). This way Marshal functionality is not longer needed.

thanks for your advice, modified following your instructions

csharp/src/Microsoft.ML.OnnxRuntime/SessionOptions.shared.cs
wenjieh-amd: > We need to add a test to this functionality.

I wonder which kind of test do we need? An unittest or system-test?

csharp/src/Microsoft.ML.OnnxRuntime/SessionOptions.shared.cs
```diff
@@ -257,6 +258,63 @@ public void AppendExecutionProvider_CUDA(OrtCUDAProviderOptions cudaProviderOpti
...
+            GCHandle[] pinnedValueHandles = new GCHandle[count];
+
+            string[] keys = config.Keys.ToArray();
+            string[] values = config.Values.ToArray();
+
+            try
+            {
+                for (int i = 0; i < count; i++)
+                {
+                    byte[] keyBytes = System.Text.Encoding.UTF8.GetBytes(keys[i] + "\0");
```
yuslepukhin: >System.Text.Encoding.UTF8.GetBytes(keys[i]  [](http://example.com/codeflow?start=38&length=43)

Use efficient utilities from NativeOnnxValueHelper that reduce garbage #Resolved

csharp/src/Microsoft.ML.OnnxRuntime/SessionOptions.shared.cs
```diff
@@ -257,6 +258,63 @@ public void AppendExecutionProvider_CUDA(OrtCUDAProviderOptions cudaProviderOpti
...
+
+                    pinnedKeyHandles[i] = GCHandle.Alloc(keyBytes, GCHandleType.Pinned);
+                    pinnedValueHandles[i] = GCHandle.Alloc(valueBytes, GCHandleType.Pinned);
+
+                    keyPtrs[i] = pinnedKeyHandles[i].AddrOfPinnedObject();
+                    valuePtrs[i] = pinnedValueHandles[i].AddrOfPinnedObject();
+                }
+
+                GCHandle pinnedKeyArray = GCHandle.Alloc(keyPtrs, GCHandleType.Pinned);
+                GCHandle pinnedValueArray = GCHandle.Alloc(valuePtrs, GCHandleType.Pinned);
```
yuslepukhin: Not needed, pass keyPtrs and valuePtrs direclty to the native calls #Resolved

csharp/src/Microsoft.ML.OnnxRuntime/NativeMethods.shared.cs
```diff
@@ -1312,6 +1314,20 @@ IntPtr[] outputValues /* An array of output value pointers. Array must be alloca
...
+        /// Append a VitisAI EP instance (configured based on given provider options) to the native OrtSessionOptions instance
+        /// </summary>
+        /// <param name="options">Native OrtSessionOptions instance</param>
+        /// <param name="keys">Native keys instance</param>
+        /// <param name="values">Native values instance</param>
+        /// <param name="numEntries">Native numEntries instance</param>
+        [UnmanagedFunctionPointer(CallingConvention.Winapi)]
+        public delegate IntPtr /*(OrtStatus*)*/ DSessionOptionsAppendExecutionProvider_VitisAI(
+            IntPtr /*(OrtSessionOptions*)*/ options,
+            IntPtr keys, IntPtr values, UIntPtr numEntries);
```
yuslepukhin: > IntPtr keys, IntPtr values [](http://example.com/codeflow?start=11&length=27)

keys and values must be arrays  #Resolved

csharp/src/Microsoft.ML.OnnxRuntime/SessionOptions.shared.cs
yuslepukhin: Under test/common/InferenceTest.cs there are examples for other AppendExecutionProvider.

Add an #if section for your provider, build and run the test. You will need to have ORT built with your provider.

csharp/src/Microsoft.ML.OnnxRuntime/SessionOptions.shared.cs
wenjieh-amd: > > System.Text.Encoding.UTF8.GetBytes(keys[i]
> 
> Use efficient utilities from NativeOnnxValueHelper that reduce garbage

updated in the latest commit

csharp/src/Microsoft.ML.OnnxRuntime/SessionOptions.shared.cs
wenjieh-amd: > Not needed, pass keyPtrs and valuePtrs direclty to the native calls


updated in the latest commit

csharp/src/Microsoft.ML.OnnxRuntime/NativeMethods.shared.cs
wenjieh-amd: > > IntPtr keys, IntPtr values
> 
> keys and values must be arrays


updated in the latest commit

csharp/src/Microsoft.ML.OnnxRuntime/SessionOptions.shared.cs
wenjieh-amd: > Under test/common/InferenceTest.cs there are examples for other AppendExecutionProvider.
> 
> Add an #if section for your provider, build and run the test. You will need to have ORT built with your provider.


updated in the latest commit, test passes in my local environment

wenjieh-amd:
let's move on to https://github.com/microsoft/onnxruntime/pull/25813 too

yuslepukhin:
Doc checker fails:

> E:/_work/onnxruntime/onnxruntime/csharp/src/Microsoft.ML.OnnxRuntime/SessionOptions.shared.cs:264: error: argument 'options' of command @param is not found in the argument list of Microsoft.ML.OnnxRuntime.SessionOptions::AppendExecutionProvider_VitisAI(Dictionary< string, string > providerOptions) (warning treated as error, aborting now)

csharp/src/Microsoft.ML.OnnxRuntime/SessionOptions.shared.cs
```diff
@@ -257,6 +258,61 @@ public void AppendExecutionProvider_CUDA(OrtCUDAProviderOptions cudaProviderOpti
...
+
+            var handles = new List<GCHandle>(count * 2);
+
+            try
+            {
+                int i = 0;
+                foreach (var kvp in providerOptions)
+                {
+                    string keyWithNull = kvp.Key + "\0";
+                    string valueWithNull = kvp.Value + "\0";
```
yuslepukhin: UTF-16 C# strings must be converted to UTF-8 and then pinned. Use utilities for conversion #Resolved

csharp/src/Microsoft.ML.OnnxRuntime/SessionOptions.shared.cs
```diff
@@ -257,6 +258,61 @@ public void AppendExecutionProvider_CUDA(OrtCUDAProviderOptions cudaProviderOpti
...
+                    keyPtrs[i] = keyHandle.AddrOfPinnedObject();
+                    valuePtrs[i] = valHandle.AddrOfPinnedObject();
+                    i++;
+                }
+
+                NativeApiStatus.VerifySuccess(NativeMethods.SessionOptionsAppendExecutionProvider_VitisAI(
+                    Handle,
+                    keyPtrs,
+                    valuePtrs,
+                    (UIntPtr)count));
```
yuslepukhin: >                 [](http://example.com/codeflow?start=37&length=16)

Stray spaces #Resolved

yuslepukhin:
Please, merge with main to pickup any pipeline fixes.

wenjieh-amd:
> Doc checker fails:
> 
> > E:/_work/onnxruntime/onnxruntime/csharp/src/Microsoft.ML.OnnxRuntime/SessionOptions.shared.cs:264: error: argument 'options' of command @param is not found in the argument list of Microsoft.ML.OnnxRuntime.SessionOptions::AppendExecutionProvider_VitisAI(Dictionary< string, string > providerOptions) (warning treated as error, aborting now)

fixed thanks

csharp/src/Microsoft.ML.OnnxRuntime/SessionOptions.shared.cs
wenjieh-amd: fixed thanks

csharp/src/Microsoft.ML.OnnxRuntime/SessionOptions.shared.cs
wenjieh-amd: > UTF-16 C# strings must be converted to UTF-8 and then pinned. Use utilities for conversion

fixed thanks

yuslepukhin:
Okay, we need to retarget this PR to main

csharp/src/Microsoft.ML.OnnxRuntime/NativeMethods.shared.cs
```diff
@@ -597,6 +597,8 @@ static NativeMethods()
                 api_.SessionOptionsAppendExecutionProvider_CUDA, typeof(DSessionOptionsAppendExecutionProvider_CUDA));
             SessionOptionsAppendExecutionProvider_CUDA_V2 = (DSessionOptionsAppendExecutionProvider_CUDA_V2)Marshal.GetDelegateForFunctionPointer(
                 api_.SessionOptionsAppendExecutionProvider_CUDA_V2, typeof(DSessionOptionsAppendExecutionProvider_CUDA_V2));
+            SessionOptionsAppendExecutionProvider_VitisAI = (DSessionOptionsAppendExecutionProvider_VitisAI)Marshal.GetDelegateForFunctionPointer(
+                api_.SessionOptionsAppendExecutionProvider_VitisAI, typeof(DSessionOptionsAppendExecutionProvider_VitisAI));
```
skottmckay, line 601: What's the reason for adding this vs. using the existing generic AppendExecutionProvider? That's the preferred way to do things so we don't need to keep adding EP specific code in all the language bindings. 

With EPs becoming plugin-based ORT needs to be agnostic and not require EP specific code for them to work.

wenjieh-amd: You know, NVDA has its CUDA interface, MSFT has its own DML interface, I just think AMD should has its VitisAI interface too.
The code has a comment:
"// New EPs should use OrtApis, preferably leveraging the generic SessionOptionsAppendExecutionProvider"
but VAI is not a new EP.

Besides, the test on AppendExecutionProvider(string providerName, Dictionary<string, string> providerOptions = null) interface passes too in this PR.

So, I lean forward keeping AppendExecutionProvider_VitisAI this interface instead of removing it.

wenjieh-amd:
> Okay, we need to retarget this PR to main

Thanks for the review!
I noticed that this pr is based on rel-1.23.0 branch, which has diverged significantly from main branch.
Since this one and https://github.com/microsoft/onnxruntime/pull/25813 contain the same changes, it might be safer to rely on https://github.com/microsoft/onnxruntime/pull/25813 for merging into main rather than rebasing rel-1.23.0.
Please let me know if you prefer me to attempt a rebase of this pr onto main, or just keep https://github.com/microsoft/onnxruntime/pull/25813 for main.

yuslepukhin:
> > Okay, we need to retarget this PR to main
> 
> Thanks for the review! I noticed that this pr is based on rel-1.23.0 branch, which has diverged significantly from main branch. Since this one and #25813 contain the same changes, it might be safer to rely on #25813 for merging into main rather than rebasing rel-1.23.0. Please let me know if you prefer me to attempt a rebase of this pr onto main, or just keep #25813 for main.

Yes, we cannot merge to 1.23 directly. Only from main.

wenjieh-amd:
let me know if this pr can be merge into branch rel-1.23.0

yuslepukhin:
> let me know if this pr can be merge into branch rel-1.23.0

We do not merge to rel-1.23.0 directly. Instead, we cherry-pick proven and tested changes from main.

wenjieh-amd:
> > let me know if this pr can be merge into branch rel-1.23.0
> 
> We do not merge to rel-1.23.0 directly. Instead, we cherry-pick proven and tested changes from main.

got it, thanks very much
any update on https://github.com/microsoft/onnxruntime/pull/25813?

PR #25830, opened 2025-08-22
Add custom ops library_path to EP metadata

psakhamoori:
## Summary
Adds EP metadata library path support to enable custom ops DLL registration with proper path resolution.

## Changes
- Added `library_path` metadata key to EP metadata infrastructure
- Pass resolved library path directly to `EpLibraryProviderBridge` constructor
- Simplified implementation per reviewer feedback (removed virtual method complexity)
- Added `#include <utility>` for std::move compliance

## Purpose
Enables downstream applications (like onnxruntime-genai) to resolve relative custom ops library paths using EP metadata, improving DLL registration reliability.

## Files Modified
- `plugin_ep/ep_factory_provider_bridge.h`
- `plugin_ep/ep_library.h` 
- `plugin_ep/ep_library_plugin.h`
- `plugin_ep/ep_library_provider_bridge.cc`
- `plugin_ep/ep_library_provider_bridge.h`
- `utils.cc`

### DIFF
--- a/include/onnxruntime/core/session/onnxruntime_ep_device_ep_metadata_keys.h
+++ b/include/onnxruntime/core/session/onnxruntime_ep_device_ep_metadata_keys.h
@@ -9,7 +9,11 @@
 // Key for the execution provider version string. This should be available for all plugin EPs.
 static const char* const kOrtEpDevice_EpMetadataKey_Version = "version";
 
+
 // Prefix for execution provider compatibility information stored in model metadata.
 // Used when generating EP context models to store compatibility strings for each EP.
 // Full key format: "ep_compatibility_info.<EP_TYPE>"
 static const char* const kOrtModelMetadata_EpCompatibilityInfoPrefix = "ep_compatibility_info.";
+
+// Key for the execution provider library path (for dynamically loaded EPs)
+static const char* const kOrtEpDevice_EpMetadataKey_LibraryPath = "library_path";
--- a/include/onnxruntime/core/session/onnxruntime_ep_device_ep_metadata_keys.h
+++ b/include/onnxruntime/core/session/onnxruntime_ep_device_ep_metadata_keys.h
@@ -9,7 +9,6 @@
 // Key for the execution provider version string. This should be available for all plugin EPs.
 static const char* const kOrtEpDevice_EpMetadataKey_Version = "version";
 
-
 // Prefix for execution provider compatibility information stored in model metadata.
 // Used when generating EP context models to store compatibility strings for each EP.
 // Full key format: "ep_compatibility_info.<EP_TYPE>"
--- a/onnxruntime/core/session/plugin_ep/ep_factory_provider_bridge.cc
+++ b/onnxruntime/core/session/plugin_ep/ep_factory_provider_bridge.cc
@@ -4,6 +4,8 @@
 #include "core/session/plugin_ep/ep_factory_provider_bridge.h"
 
 #include "core/providers/shared_library/provider_host_api.h"
+#include "core/session/plugin_ep/ep_library_plugin.h"
+#include "core/session/onnxruntime_ep_device_ep_metadata_keys.h"
 
 namespace onnxruntime {
 OrtStatus* ProviderBridgeEpFactory::GetSupportedDevices(EpFactoryInternal& ep_factory,
@@ -20,6 +22,11 @@ OrtStatus* ProviderBridgeEpFactory::GetSupportedDevices(EpFactoryInternal& ep_fa
     auto* ep_device = ep_devices[i];
     if (ep_device) {
       ep_device->ep_factory = &ep_factory;
+
+      // Add library path to EP metadata if available
+      if (library_path_.has_value()) {
+        ep_device->ep_metadata.Add(kOrtEpDevice_EpMetadataKey_LibraryPath, library_path_->string());
+      }
     }
   }
 
--- a/onnxruntime/core/session/plugin_ep/ep_factory_provider_bridge.h
+++ b/onnxruntime/core/session/plugin_ep/ep_factory_provider_bridge.h
@@ -3,6 +3,9 @@
 
 #pragma once
 
+#include <filesystem>
+#include <optional>
+
 #include "core/framework/error_code_helper.h"
 #include "core/session/abi_devices.h"
 #include "core/session/abi_session_options_impl.h"
@@ -12,12 +15,14 @@
 namespace onnxruntime {
 class ProviderBridgeEpFactory : public EpFactoryInternalImpl {
  public:
-  ProviderBridgeEpFactory(OrtEpFactory& ep_factory, ProviderLibrary& provider_library)
+  ProviderBridgeEpFactory(OrtEpFactory& ep_factory, ProviderLibrary& provider_library,
+                          std::optional<std::filesystem::path> library_path = std::nullopt)
       : EpFactoryInternalImpl(ep_factory.GetName(&ep_factory),
                               ep_factory.GetVendor(&ep_factory),
                               ep_factory.GetVendorId(&ep_factory)),
         ep_factory_{ep_factory},
-        provider_library_{provider_library} {
+        provider_library_{provider_library},
+        library_path_{std::move(library_path)} {
   }
 
  private:
@@ -61,6 +66,7 @@ class ProviderBridgeEpFactory : public EpFactoryInternalImpl {
 
   OrtEpFactory& ep_factory_;           // OrtEpFactory from the provider bridge EP
   ProviderLibrary& provider_library_;  // ProviderLibrary from the provider bridge EP
+  std::optional<std::filesystem::path> library_path_;  // Library path for EP metadata
 };
 
 }  // namespace onnxruntime
--- a/onnxruntime/core/session/plugin_ep/ep_factory_provider_bridge.h
+++ b/onnxruntime/core/session/plugin_ep/ep_factory_provider_bridge.h
@@ -65,9 +65,9 @@ class ProviderBridgeEpFactory : public EpFactoryInternalImpl {
     return ep_factory_.CreateSyncStreamForDevice(&ep_factory_, device, stream_options, stream);
   }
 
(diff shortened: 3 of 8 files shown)


nieubank:
It may be worth noting that this is not Vitis specific, the motivation may be for the custom ops dll loading scenario, but exposing the library path as metadata, generally speaking is goodness here.

onnxruntime/core/session/plugin_ep/ep_library_provider_bridge.cc
```diff
@@ -4,6 +4,7 @@
 #include "core/session/plugin_ep/ep_library_provider_bridge.h"
 
 #include "core/session/plugin_ep/ep_factory_provider_bridge.h"
+#include "core/session/plugin_ep/ep_library_plugin.h"
```
chrisdMSFT, line 7: header still needed?

psakhamoori: @chrisdMSFT 

 Yes, the #include "core/session/plugin_ep/ep_library_plugin.h" header is still needed. 

Here's why:

Member variable usage: The class has a member variable `std::unique_ptr<EpLibrary> ep_library_plugin_` which is declared as `EpLibrary` type in the header, but in the implementation file it's used with specific `EpLibraryPlugin` methods.

Method calls: The implementation calls` ep_library_plugin_->GetFactories() `and `ep_library_plugin_->Unload()`, which are methods that need the complete type definition to be compiled properly.

Type safety: Even though the member is declared as `std::unique_ptr<EpLibrary>`, the actual object passed in is an `EpLibraryPlugin`, so the header is needed for the compiler to understand the virtual method dispatch.

chrisdMSFT (dismissed):
LGTM - and same after rebase.

chrisdMSFT (approved):

jaholme (approved):

onnxruntime/core/session/plugin_ep/ep_factory_provider_bridge.h
```diff
@@ -12,12 +16,14 @@
 namespace onnxruntime {
 class ProviderBridgeEpFactory : public EpFactoryInternalImpl {
  public:
-  ProviderBridgeEpFactory(OrtEpFactory& ep_factory, ProviderLibrary& provider_library)
+  ProviderBridgeEpFactory(OrtEpFactory& ep_factory, ProviderLibrary& provider_library,
+                          std::optional<std::filesystem::path> library_path = std::nullopt)
```
snnn, line 20: Why the new parameter is optional? With this change, sometimes the library_path_ variable is there but is not set, and it could be read.  I worry it could leave hard-to-find bugs. 
Would you please explain when this parameter needs to be set?

snnn:
Could you please add more background what issue you are solving?

snnn:
This needs to be add to the main branch as well, if it is not there yet.

onnxruntime/core/session/plugin_ep/ep_factory_provider_bridge.h
psakhamoori: The `library_path` parameter is intentionally optional for the following reasons:

**When the parameter is set:**
Dynamically loaded EPs: When execution providers are loaded from external DLL/SO files (e.g., custom ops libraries), the library path is available and should be included in EP metadata for debugging and custom ops DLL registration.

**When the parameter is NOT set (nullopt):**
Statically linked EPs: When execution providers are compiled directly into the ORT binary, there's no library path to report.
Built-in EPs: Core execution providers like CPU, CUDA (when statically linked) don't have an external library path

For specific case of this genai_config.json from AMD
https://huggingface.co/amd/Phi-3-mini-4k-instruct-awq-g128-int4-asym-bf16-onnx-ryzen-strix/blob/main/genai_config.json#L8

This `library_path` needs to be set to locate `custom_ops_library` dll path.

psakhamoori:
@snnn 

**Problem Being Solved:**
**Issue:** When ONNX Runtime loads custom execution providers (EPs) from external DLL/SO files, for the EP to determine which custom ops library it should load for extended functionality. This creates a coupling problem where:

Manual Configuration Required: 
1. Users must manually specify custom ops library paths separately from the EP library
2. Discovery Problem: EPs can't automatically discover their associated custom ops libraries
3. Deployment Complexity: Harder to package and deploy EP + custom ops as a cohesive unit


Solution:
Add library path to EP metadata so that:

1. Intelligent Discovery: EPs can access their own library path via `kOrtEpDevice_EpMetadataKey_LibraryPath`
2. Automatic Registration: EPs can dynamically construct paths to co-located custom ops libraries
3. Simplified Deployment: Single EP library can manage its entire ecosystem

snnn:
> When ONNX Runtime loads custom execution providers (EPs) from external DLL/SO files, there's currently no reliable way for the EP to determine which custom ops library it should load for extended functionality

Why?  Could you please show us more details? Why it is not reliable ?

snnn:
Is there any open Github issue for that?

include/onnxruntime/core/session/onnxruntime_ep_device_ep_metadata_keys.h
```diff
@@ -9,7 +9,11 @@
...
+
 // Prefix for execution provider compatibility information stored in model metadata.
 // Used when generating EP context models to store compatibility strings for each EP.
 // Full key format: "ep_compatibility_info.<EP_TYPE>"
-static const char* const kOrtModelMetadata_EpCompatibilityInfoPrefix = "ep_compatibility_info.";
\ No newline at end of file
+static const char* const kOrtModelMetadata_EpCompatibilityInfoPrefix = "ep_compatibility_info.";
+
+// Key for the execution provider library path (for dynamically loaded EPs)
+static const char* const kOrtEpDevice_EpMetadataKey_LibraryPath = "library_path";
```
snnn, line 18: Is it a path to a file or folder?
Is it allowed to be a relative path?

onnxruntime/core/session/plugin_ep/ep_factory_provider_bridge.h
```diff
@@ -59,8 +65,9 @@ class ProviderBridgeEpFactory : public EpFactoryInternalImpl {
     return ep_factory_.CreateSyncStreamForDevice(&ep_factory_, device, stream_options, stream);
   }
 
-  OrtEpFactory& ep_factory_;           // OrtEpFactory from the provider bridge EP
-  ProviderLibrary& provider_library_;  // ProviderLibrary from the provider bridge EP
+  OrtEpFactory& ep_factory_;// OrtEpFactory from the provider bridge EP
```
snnn: Was this change made by clang-format?

snnn:
As it is a bug fix, could you please add some tests? The tests will also be helpful for us to understand the issue that you are solving.

include/onnxruntime/core/session/onnxruntime_ep_device_ep_metadata_keys.h
psakhamoori: Its path to a folder where the custom ops dll file is present. I believe it will be an absolute path.

psakhamoori:
> Is there any open Github issue for that?

This is been tracked internally

psakhamoori:
> As it is a bug fix, could you please add some tests? The tests will also be helpful for us to understand the issue that you are solving.

Have internal bug work-item with issue and repro details, I will share it.

psakhamoori:
> > When ONNX Runtime loads custom execution providers (EPs) from external DLL/SO files, there's currently no reliable way for the EP to determine which custom ops library it should load for extended functionality
> 
> Why? Could you please show us more details? Why it is not reliable ?

Thank you for the clarification request. I need to correct my description - I misstated the problem - the current custom ops loading is reliable and works well.

What this PR actually does: It adds optional library path information to EP metadata. What this PR actually does: It adds the library path to EP metadata so EPs know where they were loaded from.

onnxruntime/core/session/plugin_ep/ep_factory_provider_bridge.cc
```diff
@@ -20,6 +22,11 @@ OrtStatus* ProviderBridgeEpFactory::GetSupportedDevices(EpFactoryInternal& ep_fa
     auto* ep_device = ep_devices[i];
     if (ep_device) {
       ep_device->ep_factory = &ep_factory;
+
+      // Add library path to EP metadata if available
```
snnn, line 26: Who uses this metadata? I only see this key-value pair is set. I don't see how the value is used. And there is no test case or example.

snnn: Users need to be caution that the value string is in ANSI encoding, not UTF-8

include/onnxruntime/core/session/onnxruntime_ep_device_ep_metadata_keys.h
snnn: No. I just read the code and I think it seems to be the full path of the EP DLL. It is not a folder

snnn:
We have an autoep test in the "onnxruntime/test/autoep" folder. Could you please some tests there?

include/onnxruntime/core/session/onnxruntime_ep_device_ep_metadata_keys.h
snnn: The EP should know the full path of itself. Why do we need to pass this information to the EP DLL?

snnn (dismissed):
Talked to the author offline. This metadata information is needed by ORT GenAI DLL.

psakhamoori:
[like]  Pradeep Sakhamoori reacted to your message:
________________________________

chrisdMSFT (approved):

VishalX:
Pls mark this for 1.23 release @chrisdMSFT / @psakhamoori

onnxruntime/core/session/plugin_ep/ep_factory_provider_bridge.h
skottmckay: Statically linked EPs do not use the provider bridge. They use the 'internal' infrastructure.

PR #25833, opened 2025-08-22
Fix a long standing bug on file memory mapping on windows.

yuslepukhin:
### Description

While memory profiling some models I noticed multiple file mapping failures.
`WindowsEnv::MapFileIntoMemory()`  While it properly checks for the mapping offset to be granularity
  aligned, it calculates it as page aligned.
  Also, while saving external tensors we do not need to align big tensors to windows granularity or anything
  that is platform dependent. Set it to 4096 for all platforms.
  Granularity matters only for calculating mapping address.

### Motivation and Context

Multiple failures for file mapping for certain models.
This saves some hundreds of Mbs for some models.

### DIFF
--- a/include/onnxruntime/core/graph/model_saving_options.h
+++ b/include/onnxruntime/core/graph/model_saving_options.h
@@ -9,31 +9,30 @@ class PrepackedWeightsForGraph;
 
 // These options affect how the model initializers are written to the external file.
 // This includes options to align external initializer offset.
-// For models running on CPU, ORT will try to use mmap to load external
-// initializers. To use mmap, external initializer need to be offset aligned.
+// ORT will try to use mmap to load external initializers.
+//
 // ORT saves external initializers into single data file, each initializer is
 // accessed with offset(start position of initializer) and length(byte length of
-// initializer) of the data file. To use mmap, each offset need to be aligned
-// which means offset need to divisible by allocation granularity(64KB for
-// windows and 4K for other OSes). With align_offset to true, ORT will align
-// offset for large initializer when save ONNX model with external data file.
+// initializer) of the data file. With align_offset to true, ORT will align
+// offset for large initializer (larger than  align_threshold)
+// when save ONNX model with external data file. It will align then to
+// on_disk_alignment value.
 struct ModelSavingOptions {
   explicit ModelSavingOptions(size_t size_threshold)
       : initializer_size_threshold(size_threshold) {}
 
   // Minimal initializer size in bytes to be externalized on disk
   size_t initializer_size_threshold;
-  // Offset will always be page aligned and allocation granularity aligned for
-  // mmap support. This is done by padding previous tensor data with zeros
-  // keeping same length.
+  // Offset will always be aligned for mmap support.
+  // This is done by padding previous tensor data with zeros keeping same length.
   bool align_offset = false;
   // Alignment threshold for size of data.
   // Having a low threshold will waste file space for small initializers.
   // Only when tensor's data size is > the page_align_threshold it will be force
   // aligned. Default to 1MB.
   int64_t align_threshold = 1048576;
-  // Alignment factor for big tensors (bigger than align_threshold).
-  int64_t on_disk_alignment_factor = 4096;
+  // Alignment factor for big tensors (bigger than align_threshold). Defaults to 4K.
+  int64_t on_disk_alignment = 4096;
   // Force embed all external initializer into the Onnx file
   // Used for EPContext model generation while some nodes fallback on CPU which has external data dependency
   bool force_embed_external_ini = false;
--- a/include/onnxruntime/core/graph/model_saving_options.h
+++ b/include/onnxruntime/core/graph/model_saving_options.h
@@ -32,13 +32,8 @@ struct ModelSavingOptions {
   // Only when tensor's data size is > the page_align_threshold it will be force
   // aligned. Default to 1MB.
   int64_t align_threshold = 1048576;
-  // The allocation Granularity for mmap() support.
-  // Typically 64KB for Windows & 4KB for other OSes. Default to 64KB.
-#ifdef _WIN32
-  int64_t allocation_granularity = 65536;
-#else
-  int64_t allocation_granularity = 4096;
-#endif
+  // Alignment factor for big tensors (bigger than align_threshold).
+  int64_t on_disk_alignment_factor = 4096;
   // Force embed all external initializer into the Onnx file
   // Used for EPContext model generation while some nodes fallback on CPU which has external data dependency
   bool force_embed_external_ini = false;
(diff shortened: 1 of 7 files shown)


onnxruntime/core/platform/windows/env.cc
```diff
@@ -439,11 +439,11 @@ Status WindowsEnv::MapFileIntoMemory(_In_z_ const ORTCHAR_T* file_path,
   SYSTEM_INFO sysinfo;
   GetSystemInfo(&sysinfo);
 
-  static const DWORD page_size = sysinfo.dwPageSize;
+  // static const DWORD page_size = sysinfo.dwPageSize;
```
yuslepukhin: >// static  [](http://example.com/codeflow?start=2&length=10)

Remove

justinchuby:
Could you say more on the alignment granularity/ why we don't need to align to 64kb for windows? I suppose https://github.com/onnx/ir-py/blob/ce1d0e63c41104b584271090ecd1b4a8f8bec52f/src/onnx_ir/external_data.py#L33 can be changed the same way too if that's the case? Thanks

yuslepukhin:
> Could you say more on the alignment granularity/ why we don't need to align to 64kb for windows? I suppose https://github.com/onnx/ir-py/blob/ce1d0e63c41104b584271090ecd1b4a8f8bec52f/src/onnx_ir/external_data.py#L33 can be changed the same way too if that's the case? Thanks

The allocation granularity requirement applies to a mapping offset of the file, not to the way data is written on disk.
Even today, we only align data sizes above certain thresholds (>1Mb)_on disk and pad them, everything else is just written sequentially. There is not a need to align it to the allocation granularity on disk which is a lot of waste on Windows.
It is important that the mapping offset is aligned, and that was not done.
For that we shift back the offset when mapping, but we return the offset to the beginning of the data.
I tested it and it the mapping now always succeeds (before the fix, it failed most of the time).

yuslepukhin:
There are more ways to improve mappings today. Because of the allocation granularity and default externalizing threshold of 1K, we have many weights fall into the same mapping resulting the same disk area being mapped multiple times, up to 64 times on windows. We can do better here and map certain regions only ones OR we can decide to map the entire file and ref count it.

onnxruntime/core/framework/tensor_external_data_info.h
```diff
@@ -41,15 +41,13 @@ class ExternalDataInfo {
                                          size_t tensor_bytes_size,
                                          ::ONNX_NAMESPACE::TensorProto& proto);
 
-  // Pads the output with zeros according to the specified allocation_granularity
+  // Pads the output with zeros according to the specified alignment_factor
   // It updates external_offset for alignment.
   // need to do padding before write actual tensor data as we do offset alignment at the begin of
-  // large tensors (offset need to be page aligned and allocation granularity aligned) like below:
+  // large tensors (offset need to be page aligned) like below:
```
edgchen1, line 47: are references to "page" in the comments still relevant?

include/onnxruntime/core/graph/model_saving_options.h
```diff
@@ -32,13 +32,8 @@ struct ModelSavingOptions {
...
   int64_t align_threshold = 1048576;
-  // The allocation Granularity for mmap() support.
-  // Typically 64KB for Windows & 4KB for other OSes. Default to 64KB.
-#ifdef _WIN32
-  int64_t allocation_granularity = 65536;
-#else
-  int64_t allocation_granularity = 4096;
-#endif
+  // Alignment factor for big tensors (bigger than align_threshold).
+  int64_t on_disk_alignment_factor = 4096;
```
edgchen1: this is simply an alignment value, right? how about the name `on_disk_alignment`?

is there a good reason to make this configurable? what if someone chooses a value other than the page size?

edgchen1: I think this description above needs to be updated:

```c++
  // Offset will always be page aligned and allocation granularity aligned for
  // mmap support. This is done by padding previous tensor data with zeros
  // keeping same length.
  bool align_offset = false;
```

and this:
```c++
// initializer) of the data file. To use mmap, each offset need to be aligned
// which means offset need to divisible by allocation granularity(64KB for
// windows and 4K for other OSes). With align_offset to true, ORT will align
```

onnxruntime/core/platform/windows/env.cc
```diff
@@ -439,11 +439,10 @@ Status WindowsEnv::MapFileIntoMemory(_In_z_ const ORTCHAR_T* file_path,
...
-  static const DWORD page_size = sysinfo.dwPageSize;
   static const DWORD allocation_granularity = sysinfo.dwAllocationGranularity;
-  const FileOffsetType offset_to_page = offset % static_cast<FileOffsetType>(page_size);
-  const size_t mapped_length = length + static_cast<size_t>(offset_to_page);
-  const FileOffsetType mapped_offset = offset - offset_to_page;
+  const FileOffsetType offset_to_granularity = offset % static_cast<FileOffsetType>(allocation_granularity);
+  const size_t mapped_length = length + static_cast<size_t>(offset_to_granularity);
+  const FileOffsetType mapped_offset = offset - offset_to_granularity;
   if (mapped_offset % allocation_granularity != 0) {
     const auto error_code = GetLastError();
```
edgchen1: what error information is this call to `GetLastError()` supposed to get?

onnxruntime/core/graph/graph.cc
```diff
@@ -4543,7 +4543,7 @@ Status Graph::AddExternalInitializersToGraphProtoImpl(
       // |<---smaller tensor---->|<---padding--->|<------------------large tensor----------------------------->|
```
edgchen1, line 4543: this comment refers to allocation granularity and page alignment. I suppose it needs to be updated.

onnxruntime/core/framework/tensor_external_data_info.h
yuslepukhin: Yes, we are still page aligning large tensors. I will need to investigate the reasoning behind it.

include/onnxruntime/core/graph/model_saving_options.h
yuslepukhin: I do not know at this point if the alignment on disk is required.
We are only doing this for large tensors. The fix contained in this PR indicates that we can live w/o alignment.
To remove it completely would be a subject for more research than I am able to do at this time.
While it exists, a combination of these parameters affects how much space is wasted during alignment, so I will leave it in for now.

onnxruntime/core/framework/tensor_external_data_info.h
edgchen1: as far as I can tell, this function is only aligning to `alignment_factor` and not taking page size into account. perhaps the previously hardcoded 4096 was not a great way to get the page size either.

edgchen1 (approved):

onnxruntime/core/framework/tensor_external_data_info.h
yuslepukhin: The issue was in the code, not on disk

snnn:

Issue #23108, opened 2024-12-14
`mypy` complaints about missing stubs


cospectrum:
## error
```sh 
error: Skipping analyzing "onnxruntime": module is installed, but missing library stubs or py.typed marker  [import-untyped]
```

## deps
```sh
mypy 1.13.0 (compiled: yes)
onnxruntime v1.20.1
```

## host
```sh
sw_vers
ProductName:            macOS
ProductVersion:         13.6.8
BuildVersion:           22G820
```

```sh
sysctl -n machdep.cpu.brand_string
Apple M1
```

cospectrum:
Wait

ikamensh:
I think the solution is just to add file named py.typed into root of the python package. 

https://peps.python.org/pep-0561/
>Package maintainers who wish to support type checking of their code MUST add a marker file named py.typed to their package supporting typing. This marker applies recursively ...

dhorkel:
I have a PR ready. Anyone know who would be a reviewer to tag?

istupakov:
It would be great to solve this problem. Currently, I have to add exceptions for onnxruntime when using the type checker...

PR #25836, opened 2025-08-24
Create py.typed marker file with partial marker

dhorkel:
### Description

Add a py.typed marker file additionally with a `partial` marker as not all python files are fully typed.

This follows PEP561 (original: https://peps.python.org/pep-0561/ up to date: https://typing.python.org/en/latest/spec/distributing.html) to allow users who use type checkers to use the existing types.

### Motivation and Context

- Addresses issue: https://github.com/microsoft/onnxruntime/issues/23108

### DIFF
--- a/onnxruntime/python/py.typed
+++ b/onnxruntime/python/py.typed
@@ -1,0 +1,6 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+# This is a marker file indicating that this package contains types. See: https://typing.python.org/en/latest/spec/distributing.html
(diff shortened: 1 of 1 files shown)


victorpoughon:
Any update on this? This would be very useful for client code trying to typecheck their code with `mypy` or such modern Python type checker.

dhorkel:
> Any update on this? This would be very useful for client code trying to typecheck their code with `mypy` or such modern Python type checker.

Just hoping to get a review. It seems like the project has a large backlog of community contributions. I'm hoping this one can get eyes on it considering it is so trivial.

dhorkel:
Changes made in https://github.com/microsoft/onnxruntime/pull/28438

PR #25841, opened 2025-08-25
Add API for precompiled model compatibility check using just the compat info

adrastogi:
### Description
This PR adds a new API that applications can use to verify compatibility of a precompiled model with the underlying system, using only the compatibility info string from the model's metadata. 

### Motivation and Context

- This is a feature to enable apps to check compatibility of a precompiled model without necessarily having the model locally on the device. This enables precompiled models to be stored remotely and downloaded once the application has been able to confirm the validity of a given model with EPs on the device.

### DIFF
--- a/include/onnxruntime/core/session/onnxruntime_c_api.h
+++ b/include/onnxruntime/core/session/onnxruntime_c_api.h
@@ -33,6 +33,7 @@
 #include <stdint.h>
 #include <stdlib.h>
 #include <string.h>
+#include "onnxruntime_common_enums.h"
 
 /** \brief The API version defined in this header
  *
@@ -6480,6 +6481,20 @@ struct OrtApi {
    * \since Version 1.23.
    */
   ORT_API2_STATUS(Graph_GetModelMetadata, _In_ const OrtGraph* graph, _Outptr_ OrtModelMetadata** out);
+
+  /** \brief Validate a compiled model's compatibility information for a specific EP device.
+   *
+   * \param[in] ep_device The EP device to validate against (e.g., from GetEpDevices
+   * \param[in] compatibility_info The compatibility info string produced when the model was compiled.
+   * \param[out] out_status The resulting compatibility status for the EP device.
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \since Version 1.23.
+   */
+  ORT_API2_STATUS(GetEpCompatibilityForDevice, _In_ const OrtEpDevice* ep_device,
+                  _In_ const char* compatibility_info,
+                  _Out_ OrtCompiledModelCompatibility* out_status);
 };
 
 /*
--- a/include/onnxruntime/core/session/onnxruntime_common_enums.h
+++ b/include/onnxruntime/core/session/onnxruntime_common_enums.h
@@ -1,0 +1,18 @@
+// Shared public enums used by both onnxruntime_c_api.h and onnxruntime_ep_c_api.h
+#pragma once
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// Compatibility state of a compiled model relative to an execution provider.
+typedef enum OrtCompiledModelCompatibility {
+    OrtCompiledModelCompatibility_EP_NOT_APPLICABLE = 0,
+    OrtCompiledModelCompatibility_EP_SUPPORTED_OPTIMAL,
+    OrtCompiledModelCompatibility_EP_SUPPORTED_PREFER_RECOMPILATION,
+    OrtCompiledModelCompatibility_EP_UNSUPPORTED,
+  } OrtCompiledModelCompatibility;
+
+#ifdef __cplusplus
+}
+#endif
--- a/include/onnxruntime/core/session/onnxruntime_ep_c_api.h
+++ b/include/onnxruntime/core/session/onnxruntime_ep_c_api.h
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 // Do not include this file directly. Please include "onnxruntime_c_api.h" instead.
+#include "onnxruntime_common_enums.h"
 
 #ifdef __cplusplus
 extern "C" {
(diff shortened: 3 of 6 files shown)


include/onnxruntime/core/session/onnxruntime_common_enums.h
```diff
@@ -0,0 +1,18 @@
+// Shared public enums used by both onnxruntime_c_api.h and onnxruntime_ep_c_api.h
```
adrianlizarraga: I believe this should not be necessary because onnxruntime_c_api.h includes onnxruntime_ep_c_api.h inline (end of file). The EP author is meant to include onnxruntime_c_api.h.

OrtStatus is another example of a symbol that is defined in onnxruntime_c_api.h but used in onnxruntime_ep_c_api.h #Resolved

adrastogi: I unfortunately couldn't get the branch building until I made this change- the new API is in onnxruntime_c_api.h (maybe that is not the right place for it? In this scenario, the caller of the API is the developer of an application); and since I want the API to return a compatibility status enum value, the symbol needs to already be available.

adrastogi: Per offline conversation, will remove this common header and move the enum into onnxruntime_c_api.h, which should cover the new API as well as the usage in onnxruntime_ep_c_api.h since that header is included at the bottom of onnxruntime_c_api.h

onnxruntime/core/session/onnxruntime_c_api.cc
```diff
@@ -3423,6 +3423,56 @@ ORT_API_STATUS_IMPL(OrtApis::CopyTensors, _In_ const OrtEnv* env,
...
+    if (ep_devices[i] == nullptr) {
+      return OrtApis::CreateStatus(ORT_INVALID_ARGUMENT, "ep_devices contains a null entry.");
+    }
+    const OrtEpFactory* f = ep_devices[i]->GetMutableFactory();
+    if (i == 0) {
+      first_factory = f;
+    } else if (f != first_factory) {
+      return OrtApis::CreateStatus(ORT_INVALID_ARGUMENT, "All ep_devices must be from the same execution provider.");
+    }
+  }
```
adrianlizarraga, line 3449: This is the check we use for AppendEp_V2. Perhaps we can have use same logic: 
https://github.com/microsoft/onnxruntime/blob/ce9060936201ef979a5075cd825a5942dec433ee/onnxruntime/core/session/utils.cc#L458-L467

onnxruntime/core/session/onnxruntime_c_api.cc
```diff
@@ -3445,6 +3495,17 @@ ORT_API_STATUS_IMPL(OrtApis::GetEpDevices, _In_ const OrtEnv* /*env*/,
...
 }
 
+// Minimal build stub for GetEpCompatibilityForDevices to satisfy symbol references from the API table
+ORT_API_STATUS_IMPL(OrtApis::GetEpCompatibilityForDevices,
+                    _In_reads_(num_ep_devices) const OrtEpDevice* const* /*ep_devices*/,
+                    _In_ size_t /*num_ep_devices*/,
+                    _In_ const char* /*compatibility_info*/,
+                    _Out_ OrtCompiledModelCompatibility* /*out_status*/) {
+  API_IMPL_BEGIN
+  return OrtApis::CreateStatus(ORT_NOT_IMPLEMENTED, "This API in not supported in a minimal build.");
```
adrianlizarraga: nit: can include the function name in the error message: `"GetEpCompatibilityForDevices is not supported in a minimal build."`

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -6480,6 +6490,24 @@ struct OrtApi {
...
+   *                        All devices must belong to the same execution provider.
+   * \param[in] num_ep_devices The number of EP devices provided.
+   * \param[in] compatibility_info The compatibility info string produced when the model was compiled.
+   * \param[out] out_status The resulting compatibility status for the EP devices.
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \since Version 1.23.
+   */
+  ORT_API2_STATUS(GetEpCompatibilityForDevices,
```
adrianlizarraga: What do you think of calling this something like `GetModelCompatibilityForEp` or `GetModelCompatibilityForEpDevices`? Otherwise, the name may lead user to believe this function validates an EP or devices (instead of the model compat).

include/onnxruntime/core/session/onnxruntime_common_enums.h
adrianlizarraga, line 1: Should be able to delete this file now

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -901,20 +889,28 @@ struct OrtEpFactory {
    */
   ORT_API_T(const char*, GetVersion, _In_ const OrtEpFactory* this_ptr);
 
-  /** \brief Validate the compatibility of a compiled model with the execution provider.
+  /** \brief Validate the compatibility of a compiled model with the execution provider for one or more devices.
+   *
+   * Given a compatibility info string produced during model compilation, the EP should determine whether the
```
adrianlizarraga: nit: Since this function is on the factory struct, perhaps we use "EP factory" instead of just "EP" for clarify. Ex: `the EP should determine` -> `the EP factory should determine`

onnxruntime/test/framework/ep_compatibility_test.cc
```diff
@@ -408,3 +408,93 @@ TEST_F(EpCompatibilityTest, TestSessionOptionConfiguration) {
...
+  const OrtApi* api = OrtGetApiBase()->GetApi(ORT_API_VERSION);
+  ASSERT_NE(api, nullptr);
+
+  auto [env, device] = CreateEnvAndGetCpuEpDevice(api);
+  ASSERT_NE(env, nullptr);
+  ASSERT_NE(device, nullptr);
+
+  OrtCompiledModelCompatibility out_status = static_cast<OrtCompiledModelCompatibility>(-1);
+  const OrtEpDevice* devices2[] = {device};
+  OrtStatus* st = api->GetEpCompatibilityForDevices(devices2, 1, "arbitrary-compat-string", &out_status);
```
adrianlizarraga: Looks like this is missing a call to release the `OrtStatus`.

Perhaps we can use the C++ API so that the test code doesn't have to manually release the status:
```c++
Ort::Status st{api->GetEpCompatibilityForDevices(devices2, 1, "arbitrary-compat-string", &out_status)};
// no need to manually release the status (done in destructor)
```

onnxruntime/core/session/onnxruntime_c_api.cc
adrastogi: If it's ok by you, I'm going to hold off on this (what I'd ideally want to do is refactor this into a small helper and have it cover both scenarios- I will look at that once we're past the current wave of work and will file an issue to track once this PR goes in).

onnxruntime/test/framework/ep_compatibility_test.cc
adrastogi: Good eyes. For now, I made the surgical fix to release the lingering status.

snnn:

PR #25849, opened 2025-08-26
Enable ABSL_FLAGS flag registration for onnxruntime_perf_test for mobile build

chilo-ms:
### Description
`ABSL_FLAGS_STRIP_NAMES `is set to 1 by default to disable flag registration when building for Android, iPhone, and "embedded devices".
So, running onnxruntime_perf_test on Android will see that flags are not registered.

image (2)

Set `ABSL_FLAGS_STRIP_NAMES ` to 0 by default for all builds.


jywu-msft (approved):

snnn:

PR #25850, opened 2025-08-26
Improve SimplifiedLayerNorm by using same techniques as SkipSimplifiedLayerNorm

sushraja-msft:
### Description
Use similar shaders as SkipSimplifiedLayerNorm in SimplifiedLayerNorm, to fix the performance issues with SimplifiedLayerNorm.

### Motivation and Context
Prior to this change, generation in Bitnet was bottlenecked on SimplifiedLayerNorm
image
with this change performance has now improved to match SkipSimplifiedLayerNorm
image

### DIFF
--- a/onnxruntime/core/providers/webgpu/nn/layer_norm.cc
+++ b/onnxruntime/core/providers/webgpu/nn/layer_norm.cc
@@ -26,7 +26,7 @@ static TensorShape GetOverrideShape(const TensorShape& shape, int components) {
 }
 
 Status LayerNormProgram::GenerateShaderCode(ShaderHelper& shader) const {
-  const auto& x = shader.AddInput("x", ShaderUsage::UseUniform | ShaderUsage::UseValueTypeAlias);
+  const auto& x = shader.AddInput("x", ShaderUsage::UseUniform | ShaderUsage::UseValueTypeAlias | ShaderUsage::UseElementTypeAlias);
   shader.AddInput("scale", ShaderUsage::UseUniform);
   if (has_bias_) {
     shader.AddInput("bias", ShaderUsage::UseUniform);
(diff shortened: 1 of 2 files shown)


onnxruntime/core/providers/webgpu/nn/layer_norm.cc
```diff
@@ -39,35 +39,113 @@ Status LayerNormProgram::GenerateShaderCode(ShaderHelper& shader) const {
...
+        << "let offset = ix * stride + iy * norm_size_vectorized;\n"
+        << "let offset1d = stride * ix;\n"
+        << "sum_shared[ix] = f32_val_t(0);\n"
+        << "sum_squared_shared[ix] = f32_val_t(0);\n"
+        << "if (ix == workgroup_size_x - 1) {\n"
+        << " stride = norm_size_vectorized - stride * ix;\n"
+        << "}\n"
+        << "for (var i: u32 = 0; i < stride; i++) {\n"
+        << " let input_value = x[offset + i];\n"
+        << " y[offset + i] = input_value;\n"
```
qjia7, line 116: It seems that most of the shader code are same with SkipSimplifiedLayerNorm except that one is assign `x[offset + i]` to  `y[offset + i]` and other one is assign `x[offset + i]+ skip_value + bias[offset1d + i]` to `y[offset + i]` and SkipSimplifiedLayerNorm  also has an optional `input_skip_bias_sum` output. Is it possible that merge these two shaders into one template and reuse the code?

sushraja-msft: Yes JiaJia, can be done - I wasn't unsure about having templates for all these simple ops, but yeah we can merge the code with SkipSimplifiedLayerNorm. That can be a follow up change though.

PR #25851, opened 2025-08-26
Fix native module loading and improve React Native package

danemadsen:
- The buildscript block was removed from js/react_native/android/build.gradle to avoid conflicts with Expo prebuild. 
- The react-native entry in js/react_native/package.json was updated to dist/commonjs/index, allowing Expo/Metro to import compiled JavaScript.
- Improved the bindings to prevent crashes


danemadsen:

PR #25856, opened 2025-08-26
[WebNN] Fix the op support limit for batchNormalization

shiyi9801:
### Description

According to the [WebNN spec](https://www.w3.org/TR/webnn/#api-mlgraphbuilder-batchnorm), the batchNorm should have input names "mean" and "variance" instead of "input_mean" and "input_var".

### Motivation and Context
This issue causes any BatchNorm with mean/variance inputs to fall back to wasm.


shiyi9801:
@fdwr PTAL, thanks!

/cc @Honry

PR #25857, opened 2025-08-26
Fix incorrect symbolic shape inference for ConstantOfShape

ankutalev:
### Description
See title.

### Motivation and Context
Current code incorrectly deals with empty default input tensor (evaluates it as empty shape), but should to constant, see [spec](https://onnx.ai/onnx/operators/onnx__ConstantOfShape.html#inputs)

### DIFF
--- a/onnxruntime/python/tools/symbolic_shape_infer.py
+++ b/onnxruntime/python/tools/symbolic_shape_infer.py
@@ -974,6 +974,11 @@ class SymbolicShapeInference:
 
     def _infer_ConstantOfShape(self, node):  # noqa: N802
         sympy_shape = self._get_int_or_float_values(node)[0]
+        # according to spec, ConstantOfShape may have empty input; in this case output must be scalar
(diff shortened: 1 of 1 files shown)


onnxruntime/python/tools/symbolic_shape_infer.py
```diff
@@ -974,6 +974,11 @@ def _infer_Constant(self, node):  # noqa: N802
 
     def _infer_ConstantOfShape(self, node):  # noqa: N802
         sympy_shape = self._get_int_or_float_values(node)[0]
+        # according to spec, ConstantOfShape may have empty input; in this case output must be scalar
+        # also we can infer shape of this fake input for free
+        if sympy_shape == []:
+            sympy_shape = [1]
+            self.sympy_data_[node.input[0]] = [1]
```
xadupre, line 981: A scalar should have an empty shape [] not [1]? Am I missing something?

PR #25860, opened 2025-08-26
Add default constructor to Ort::Status.

yuslepukhin:
### Description

Fix packaging pipelines

### Motivation and Context

  During CIs and local builds Ort::Status() gets inherited from the base due to using directives,
  however, that does not work for packaging pipelines.
  Having default ctor is important for storing Status in containers if needed.

### DIFF
--- a/include/onnxruntime/core/session/onnxruntime_cxx_api.h
+++ b/include/onnxruntime/core/session/onnxruntime_cxx_api.h
@@ -728,6 +728,7 @@ struct Status : detail::Base<OrtStatus> {
   using Base = detail::Base<OrtStatus>;
   using Base::Base;
 
+  Status() = default;
   explicit Status(std::nullptr_t) noexcept {}      ///< Create an empty object, must be assigned a valid one to be used
   explicit Status(OrtStatus* status) noexcept;     ///< Takes ownership of OrtStatus instance returned from the C API.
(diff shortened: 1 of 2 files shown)


include/onnxruntime/core/session/onnxruntime_cxx_api.h
```diff
@@ -728,6 +728,7 @@ struct Status : detail::Base<OrtStatus> {
   using Base = detail::Base<OrtStatus>;
   using Base::Base;
 
+  Status() = default;
```
edgchen1: it might be worth mentioning that `Status()` is the same as `Status(nullptr)`. this is unlike some of the other C++ API types.

also, it might be worth mentioning that such a Status represents "no error".

include/onnxruntime/core/session/onnxruntime_cxx_api.h
```diff
@@ -728,6 +728,7 @@ struct Status : detail::Base<OrtStatus> {
   using Base = detail::Base<OrtStatus>;
   using Base::Base;
```
edgchen1: if we explicitly declare a default ctor now, do we still need `using Base::Base;`?

snnn (approved):

snnn:

PR #25861, opened 2025-08-26
[QNN-EP] Enable einsum with QK equations for QNN.

quic-muchhsu:
### Description

Enable einsum op with QK equations for attention in QNN EP.

### Motivation and Context

Current einsum op in QNN doesn't support equations with capital alphabets. Loose this constraint to allow more usecases.


HectorSVC (approved):
:shipit:

PR #25864, opened 2025-08-27
Increase Qnn Windows build timeout

HectorSVC:
### Description
Increase Qnn Windows build timeout


jywu-mysoft:
we may need to increase some of the timeouts for win arm64 qnn too? I see occasional timeouts there too.

HectorSVC:
> we may need to increase some of the timeouts for win arm64 qnn too? I see occasional timeouts there too.

done

jywu-mysoft:
seems very abnormal that even with 4 hour time out, the build still didn't complete (or it hung)
@snnn any ideas? could it be a faulty compute instance?

jywu-mysoft:
no longer needed. updated VM SKU seems to be running better.

PR #25865, opened 2025-08-27
[QNN EP] Add Case-2 LPBQ pattern support for Gemm and Matmul nodes

quic-tirupath:
### Description

 - Case-2 LPBQ pattern omits QuantizeLinear node in LPBQ packing pattern
 - Modify LPBQ fusion logic in QNN EP implemented for Gemma and MatMul nodes to gracefully handle the optional QuantizeLinear node in LPBQ packing pattern.
 - Add unit tests to verify Case-2 LPBQ pattern fusion for Gemm and MatMul nodes.

### Motivation and Context
- QuantizeLinear node in LowPowerBlockQuantization encoding packing pattern can be optional as it helps to keep the weights in INT datatype and further helps to reduce the size of model.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/qnn_node_group/lpbqgemm_fusion.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_node_group/lpbqgemm_fusion.cc
@@ -16,12 +16,12 @@ namespace onnxruntime {
 namespace qnn {
 
 static Status CreateOrValidateOnQnn(QnnModelWrapper& qnn_model_wrapper,
-                                    const NodeUnit& scale_dql_node_unit,
-                                    const NodeUnit& w_ql_node_unit,
-                                    const NodeUnit& w_dql_node_unit,
-                                    const NodeUnit& act_dql_node_unit,
-                                    const NodeUnit& gemm_node_unit,
-                                    const NodeUnit& output_ql_node_unit,
+                                    const NodeUnit* scale_dql_node_unit,
+                                    const NodeUnit* w_ql_node_unit,
+                                    const NodeUnit* w_dql_node_unit,
+                                    const NodeUnit* act_dql_node_unit,
+                                    const NodeUnit* gemm_node_unit,
+                                    const NodeUnit* output_ql_node_unit,
                                     bool validate);
 
 std::unique_ptr<IQnnNodeGroup> LowPowerBlockQuantizedGemmFusion::TryFusion(
@@ -67,24 +67,41 @@ std::unique_ptr<IQnnNodeGroup> LowPowerBlockQuantizedGemmFusion::TryFusion(
                                                      w_dql_parent_types,
                                                      node_to_node_unit,
                                                      node_unit_to_qnn_node_group);
-  if (p_w_ql_node_unit == nullptr) {
-    return nullptr;
+  if (p_w_ql_node_unit != nullptr) {
+    // Check if input of QuantizeLinear is constant initializer
+    if (!qnn_model_wrapper.IsConstantInput(p_w_ql_node_unit->Inputs()[0].node_arg.Name())) {
+      return nullptr;
+    }
+  } else {
+    // Check if input of DequantizeLinear is constant initializer
+    if (!qnn_model_wrapper.IsConstantInput(p_w_dql_node_unit->Inputs()[0].node_arg.Name())) {
+      return nullptr;
+    }
   }
 
-  // Check if input of QuantizeLinear is constant initializer
-  if (!qnn_model_wrapper.IsConstantInput(p_w_ql_node_unit->Inputs()[0].node_arg.Name())) {
-    return nullptr;
-  }
-
-  // Get DequantizeLinear contains per-block int scales and per-channel float scales
-  const std::array<std::string_view, 1> w_ql_parent_types = {"DequantizeLinear"};
-  const NodeUnit* p_scale_dql_node_unit = GetParentOfType(graph_viewer,
-                                                          *p_w_ql_node_unit,
-                                                          w_ql_parent_types,
-                                                          node_to_node_unit,
-                                                          node_unit_to_qnn_node_group);
-  if (p_scale_dql_node_unit == nullptr) {
-    return nullptr;
+  const NodeUnit* p_scale_dql_node_unit = nullptr;
+  if (p_w_ql_node_unit != nullptr) {
+    const std::array<std::string_view, 1> w_ql_parent_types = {"DequantizeLinear"};
+    p_scale_dql_node_unit = GetParentOfType(graph_viewer,
+                                            *p_w_ql_node_unit,
+                                            w_ql_parent_types,
+                                            node_to_node_unit,
+                                            node_unit_to_qnn_node_group);
+    if (p_scale_dql_node_unit == nullptr) {
+      return nullptr;
+    }
+  } else {
+    // Get DequantizeLinear contains per-block int scales and per-channel float scales
+    // DequantizeLinear connects to scale (input 1) of DequantizeLinear node
+    const std::string& scale_name = (p_w_dql_node_unit->Inputs()[0]).quant_param->scale.Name();
+    p_scale_dql_node_unit = GetParentOfInputByName(graph_viewer,
+                                                   *p_w_dql_node_unit,
+                                                   scale_name,
+                                                   node_to_node_unit,
+                                                   node_unit_to_qnn_node_group);
+    if (p_scale_dql_node_unit == nullptr || p_scale_dql_node_unit->OpType() != "DequantizeLinear") {
+      return nullptr;
+    }
   }
 
   TensorInfo pc_scales_tensor_info = {};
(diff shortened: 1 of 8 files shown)


jywu-msft (dismissed):

minfhong-qti (commented):
https://github.com/microsoft/onnxruntime/blob/bd28495aee356df14ad1ef213d03ae38f2807e89/onnxruntime/core/providers/qnn/builder/qnn_node_group/lpbqgemm_fusion.cc#L246
Should it be
```python
const NodeUnitIODef& per_channel_float_def = scale_dql_node_unit.Inputs()[1]
```
instead?

jywu-mysoft:
> https://github.com/microsoft/onnxruntime/blob/bd28495aee356df14ad1ef213d03ae38f2807e89/onnxruntime/core/providers/qnn/builder/qnn_node_group/lpbqgemm_fusion.cc#L246
> 
> 
> Should it be
> ```python
> const NodeUnitIODef& per_channel_float_def = scale_dql_node_unit.Inputs()[1]
> ```
> 
> instead?

@quic-tirupath ?

quic-tirupath:
> https://github.com/microsoft/onnxruntime/blob/bd28495aee356df14ad1ef213d03ae38f2807e89/onnxruntime/core/providers/qnn/builder/qnn_node_group/lpbqgemm_fusion.cc#L246
> 
> 
> Should it be
> ```python
> const NodeUnitIODef& per_channel_float_def = scale_dql_node_unit.Inputs()[1]
> ```
> 
> instead?

Nope; This is correct. The DQL node_unit contains only one input which is input[0] and other inputs like input[1], input[2] are wrapped as quantization params of input[0]. Here we first get the input[0] IODef, and then get the per_channel_float values from its quantization param: scale.

onnxruntime/core/providers/qnn/builder/qnn_node_group/lpbqgemm_fusion.h
```diff
@@ -46,7 +46,9 @@ class LowPowerBlockQuantizedGemmFusion : public IQnnNodeGroup {
 
  private:
   std::array<const NodeUnit*, 6> node_units_;
+  mutable std::vector<const NodeUnit*> filtered_node_units_;
+  mutable bool filtered_node_units_initialized_ = false;
```
edgchen1: can these not be mutable? it seems that `filtered_node_units_` could be populated in the ctor. same comment for `LowPowerBlockQuantizedMatMulFusion`.

onnxruntime/core/providers/qnn/builder/qnn_node_group/lpbqgemm_fusion.cc
```diff
@@ -178,23 +209,26 @@ Status UnpackWeightTensorData(const QnnModelWrapper& qnn_model_wrapper,
...
+                             const NodeUnit* p_w_dql_node_unit,
+                             const NodeUnit* p_act_dql_node_unit,
+                             const NodeUnit* p_gemm_node_unit,
+                             const NodeUnit* p_output_ql_node_unit,
                              bool validate) {
+  const NodeUnit& scale_dql_node_unit = *p_scale_dql_node_unit;
+  const NodeUnit& w_dql_node_unit = *p_w_dql_node_unit;
+  const NodeUnit& act_dql_node_unit = *p_act_dql_node_unit;
+  const NodeUnit& gemm_node_unit = *p_gemm_node_unit;
+  const NodeUnit& output_ql_node_unit = *p_output_ql_node_unit;
```
edgchen1: why update the parameters to be pointers if they will immediately be unconditionally dereferenced? now there is the possibility of passing in a nullptr by mistake.

onnxruntime/core/providers/qnn/builder/qnn_node_group/lpbqgemm_fusion.cc
```diff
@@ -252,35 +286,52 @@ Status CreateOrValidateOnQnn(QnnModelWrapper& qnn_model_wrapper,
...
+                                                              per_block_int_scale,
+                                                              weight_offset,
+                                                              quant_data,
+                                                              weight_data_type,
+                                                              output_channel_axis,
+                                                              block_scales_axis,
+                                                              block_size,
+                                                              block_scales_shape));
+
+    // Get weight tensor type from input of w_dql_tensor or output_dql_tensor
```
edgchen1: >w_dql_tensor [](http://example.com/codeflow?start=44&length=12)

is this comment up to date? `weight_tensor_name` is coming from `w_ql_input_1_def` now.

onnxruntime/core/providers/qnn/builder/qnn_node_group/lpbqmatmul_fusion.cc
```diff
@@ -215,31 +248,30 @@ Status TwoDimensionTranspose(std::vector<uint8_t>& data,
...
-                         const NodeUnit& matmul_node_unit,
+                         const NodeUnit* scale_dql_node_unit,
+                         const NodeUnit* w_ql_node_unit,
+                         const NodeUnit* matmul_node_unit,
                          std::vector<std::string>& input_names,
                          const logging::Logger& logger) {
   ORT_UNUSED_PARAMETER(logger);
-  const NodeUnitIODef& mm_input_1_def = matmul_node_unit.Inputs()[1];
-  const NodeUnitIODef& w_ql_input_1_def = w_ql_node_unit.Inputs()[0];
+  const NodeUnitIODef& mm_input_1_def = matmul_node_unit->Inputs()[1];
```
edgchen1: >const NodeUnitIODef& mm_input_1_def = matmul_node_unit->Inputs()[1]; [](http://example.com/codeflow?start=2&length=68)

similar comment - why update the parameters from references to pointers if the pointers will be unconditionally dereferenced? using references seems safer.

onnxruntime/core/providers/qnn/builder/qnn_node_group/lpbqmatmul_fusion.cc
```diff
@@ -251,67 +283,130 @@ Status ProcessLPBQWeight(QnnModelWrapper& qnn_model_wrapper,
...
+    for (auto node : matmul_node_unit->GetAllNodesInGroup()) {
+      for (auto node_input : node->InputDefs()) {
+        if (node_input->Name() == weight_tensor_name) {
+          p_w_dql_node = node;
+          break;
+        }
+
+        if (p_w_dql_node != nullptr) {
+          break;
+        }
```
edgchen1: is this meant to be in the outer for loop?

onnxruntime/core/providers/qnn/builder/qnn_node_group/lpbqmatmul_fusion.cc
```diff
@@ -251,67 +283,130 @@ Status ProcessLPBQWeight(QnnModelWrapper& qnn_model_wrapper,
...
+    ORT_RETURN_IF_ERROR(UnpackWeightTensorData(qnn_model_wrapper, weight_tensor_proto, weight_shape, input_channel_axis, quant_data));
+
+    size_t output_channel_axis = 0;  // MatMul requires axis to be rank-1
 
-  // Get weight tensor type from input of w_dql_tensor or output_dql_tensor
-  Qnn_TensorType_t weight_tensor_type = qnn_model_wrapper.GetTensorType(weight_tensor_name);
-  QnnTensorWrapper weight_tensor(weight_tensor_name, weight_tensor_type, QNN_DATATYPE_SFIXED_POINT_8,
-                                 std::move(weight_qparams), std::move(weight_shape),
-                                 std::move(quant_data));
+    // MatMul w/ LPBQ requies MatMul(MxK, KxN) and axis = rank-1 (out channels)
```
edgchen1: >requies [](http://example.com/codeflow?start=22&length=7)

"requies" -> "requires"

onnxruntime/core/providers/qnn/builder/qnn_node_group/utils.cc
```diff
@@ -314,5 +314,62 @@ const NodeUnit* GetOnlyChildOfOutput(const GraphViewer& graph_viewer,
...
+                                 [&](const NodeArg* node_input) {
+                                   return node_input->Name() == input_name;
+                                 });
+    if (!has_input) {
+      continue;
+    }
+
+    const Node& child_node = *node;
+
+    for (auto edge = child_node.InputEdgesBegin(); edge != child_node.InputEdgesEnd(); ++edge) {
```
edgchen1, line 335: there are similarities between this function and `GetParentOfInput()`. can they be refactored to avoid code duplication?

one notable difference is that this function also checks `p_parent_node_unit->UnitType() != NodeUnit::Type::SingleNode`, but `GetParentOfInput()` doesn't. why are they different?

onnxruntime/core/providers/qnn/builder/qnn_node_group/lpbqmatmul_fusion.cc
tirupath-qti: yes. this is correct; will update.

onnxruntime/core/providers/qnn/builder/qnn_node_group/lpbqmatmul_fusion.cc
tirupath-qti: will address it.

onnxruntime/core/providers/qnn/builder/qnn_node_group/lpbqgemm_fusion.cc
tirupath-qti: since QuantizeLinear on LPBQ weights is optional i.e., case-2 pattern (DQ -> DQ) contains LPBQ encodings, instead of making one parameter pointer, we chose modified all parameters to pointer types. We will revert the parameter types to references and keep the w_ql_node_unit parameter in pointer type.

onnxruntime/core/providers/qnn/builder/qnn_node_group/lpbqmatmul_fusion.cc
tirupath-qti: since QuantizeLinear on LPBQ weights is optional i.e., case-2 pattern (DQ -> DQ) contains LPBQ encodings, instead of making one parameter pointer, we chose modified all parameters to pointer types. We will revert the parameter types to references and keep the w_ql_node_unit parameter in pointer type.

onnxruntime/core/providers/qnn/builder/qnn_node_group/lpbqgemm_fusion.cc
tirupath-qti: thanks for catching it. will remove it.

onnxruntime/core/providers/qnn/builder/qnn_node_group/utils.cc
tirupath-qti: there are some similarities between these 2 functions. Since the functions in this file are used in multiple Op builders and requires extensive regression coverage, can we refactor the code in a separate PR.

tirupath-qti:
Hi @edgchen1 
Thanks for providing the review comments. I added my responses to all your comments and address many of them in the new commit added to this PR.

Could you please help to review and approve the latest PR version? Also, please help trigger CI job and merge after the successful review.

onnxruntime/core/providers/qnn/builder/qnn_node_group/utils.cc
edgchen1: only one of the functions has existing usage. what I had in mind for "refactoring" was sharing some of the code between the two implementations. they could still exist with their current signatures.

could you shed some light on the reason for this difference?

> one notable difference is that this function also checks p_parent_node_unit->UnitType() != NodeUnit::Type::SingleNode, but GetParentOfInput() doesn't.

onnxruntime/core/providers/qnn/builder/qnn_node_group/lpbqgemm_fusion.h
```diff
@@ -23,12 +23,12 @@ class QnnModelWrapper;
...
-                                   const NodeUnit& W_DQL_node_unit,
-                                   const NodeUnit& Act_DQL_node_unit,
-                                   const NodeUnit& Gemm_node_unit,
-                                   const NodeUnit& Output_QL_node_unit);
+  LowPowerBlockQuantizedGemmFusion(const NodeUnit* Scale_DQL_node_unit,
+                                   const NodeUnit* W_QL_node_unit,
+                                   const NodeUnit* W_DQL_node_unit,
+                                   const NodeUnit* Act_DQL_node_unit,
+                                   const NodeUnit* Gemm_node_unit,
+                                   const NodeUnit* Output_QL_node_unit);
```
edgchen1: general comment about converting references to pointers in signatures. can that only be done for arguments that are optional? it'd be clearer to leave the required ones as references.

tirupath-qti: sure. We can limit that only for arguments that are optional. I will add a new patch.

onnxruntime/core/providers/qnn/builder/qnn_node_group/utils.cc
tirupath-qti: When I first opened this PR, GetParentOfInput() wasn’t around yet. I looked at how GetParentOfType() was done in this file and tried to keep the logic similar for consistency and a bit of defensive programming.

onnxruntime/core/providers/qnn/builder/qnn_node_group/lpbqgemm_fusion.h
tirupath-qti: Addressed this in new commit added in this PR.

onnxruntime/core/providers/qnn/builder/qnn_node_group/lpbqmatmul_fusion.cc
tirupath-qti: This is addressed in the new commit added in this PR.

onnxruntime/core/providers/qnn/builder/qnn_node_group/utils.cc
tirupath-qti: We will do the suggested refactor in a separate PR instead of blocking the current PR and risking 1.24 release.

edgchen1 (approved):

PR #25877, opened 2025-08-27
Fix MoE CPP tests

apsonawane:
This change adds skip test for QMoE CPU tests when running on TensorRT or CUDA EP.
In the QMoE kernel there was a memory overwrite bug in the accumulate part, updated that and this fixed the python tests back

### DIFF
--- a/onnxruntime/contrib_ops/cpu/moe/moe_quantization_cpu.cc
+++ b/onnxruntime/contrib_ops/cpu/moe/moe_quantization_cpu.cc
@@ -331,7 +331,13 @@ Status QMoECPU<T>::Compute(OpKernelContext* context) const {
         const int64_t token_idx = route_idx / k_;
         const float weight = route_scale[route_idx];
 
-        float* dest = thread_local_outputs + static_cast<size_t>(thread_id) * output_buffer_size + token_idx * hidden_size;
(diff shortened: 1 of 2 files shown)


onnxruntime/contrib_ops/cpu/moe/moe_quantization_cpu.cc
```diff
@@ -331,7 +331,13 @@ Status QMoECPU<T>::Compute(OpKernelContext* context) const {
         const int64_t token_idx = route_idx / k_;
         const float weight = route_scale[route_idx];
 
-        float* dest = thread_local_outputs + static_cast<size_t>(thread_id) * output_buffer_size + token_idx * hidden_size;
+        const size_t buffer_offset = static_cast<size_t>(token_idx) * static_cast<size_t>(hidden_size);
+        if (buffer_offset + static_cast<size_t>(hidden_size) > output_buffer_size) {
```
tianleiwu, line 335: Do you know the root cause of this > output_buffer_size?

apsonawane: It is just a safety measure, there might be a edge case which may cause the output_buffer to overflow. We can remove that if we want

tianleiwu: This is not needed. But it is fine to merge it to unblock the pipeline.

PR #25878, opened 2025-08-27
Initial draft of language bindings for model compatibility API

adrastogi:
### Description
This change builds on top of #25841 , and adds the scaffolding necessary to call into this API from C++ / C# / Python.

### DIFF
--- a/csharp/src/Microsoft.ML.OnnxRuntime/NativeMethods.shared.cs
+++ b/csharp/src/Microsoft.ML.OnnxRuntime/NativeMethods.shared.cs
@@ -368,6 +368,88 @@ namespace Microsoft.ML.OnnxRuntime
         public IntPtr EpDevice_Device;
         public IntPtr GetEpApi;
         public IntPtr GetTensorSizeInBytes;
+
+        public IntPtr AllocatorGetStats;
+
+        public IntPtr CreateMemoryInfo_V2;
+        public IntPtr MemoryInfoGetDeviceMemType;
+        public IntPtr MemoryInfoGetVendorId;
+
+        public IntPtr ValueInfo_GetValueProducer;
+        public IntPtr ValueInfo_GetValueNumConsumers;
+        public IntPtr ValueInfo_GetValueConsumers;
+        public IntPtr ValueInfo_GetInitializerValue;
+        public IntPtr ValueInfo_GetExternalInitializerInfo;
+        public IntPtr ValueInfo_IsRequiredGraphInput;
+        public IntPtr ValueInfo_IsOptionalGraphInput;
+        public IntPtr ValueInfo_IsGraphOutput;
+        public IntPtr ValueInfo_IsConstantInitializer;
+        public IntPtr ValueInfo_IsFromOuterScope;
+        public IntPtr Graph_GetName;
+        public IntPtr Graph_GetModelPath;
+        public IntPtr Graph_GetOnnxIRVersion;
+        public IntPtr Graph_GetNumOperatorSets;
+        public IntPtr Graph_GetOperatorSets;
+        public IntPtr Graph_GetNumInputs;
+        public IntPtr Graph_GetInputs;
+        public IntPtr Graph_GetNumOutputs;
+        public IntPtr Graph_GetOutputs;
+        public IntPtr Graph_GetNumInitializers;
+        public IntPtr Graph_GetInitializers;
+        public IntPtr Graph_GetNumNodes;
+        public IntPtr Graph_GetNodes;
+        public IntPtr Graph_GetParentNode;
+        public IntPtr Graph_GetGraphView;
+        public IntPtr Node_GetId;
+        public IntPtr Node_GetName;
+        public IntPtr Node_GetOperatorType;
+        public IntPtr Node_GetDomain;
+        public IntPtr Node_GetSinceVersion;
+        public IntPtr Node_GetNumInputs;
+        public IntPtr Node_GetInputs;
+        public IntPtr Node_GetNumOutputs;
+        public IntPtr Node_GetOutputs;
+        public IntPtr Node_GetNumImplicitInputs;
+        public IntPtr Node_GetImplicitInputs;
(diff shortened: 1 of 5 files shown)


csharp/src/Microsoft.ML.OnnxRuntime/NativeMethods.shared.cs
```diff
@@ -368,6 +368,88 @@ public struct OrtApi
         public IntPtr EpDevice_Device;
         public IntPtr GetEpApi;
         public IntPtr GetTensorSizeInBytes;
+
```
adrastogi, line 371: It looks like this file has yet to be updated with all the other APIs that have been added since 1.22... I tried adding them (ended up with some extra whitespace) since the comment further up in the file said these need to be in sync. Feedback welcome on whether I should have done something else here.

csharp/src/Microsoft.ML.OnnxRuntime/OrtEnv.shared.cs
```diff
@@ -7,6 +7,14 @@
 
 namespace Microsoft.ML.OnnxRuntime
 {
+    public enum OrtCompiledModelCompatibility
```
adrastogi, line 17: Wasn't too sure about placement of tthe code in this file. Advice welcome.

Craigacp:
Should I add a Java API for this? I can fold it into #25131 or make a fresh PR.

csharp/src/Microsoft.ML.OnnxRuntime/NativeMethods.shared.cs
yuslepukhin: Yes, just keep adding `IntPtr` with names there, you do not need initialize them in your particular PR, but the layout needs to be preserved. And yes, some of them has already been added since 1.22.

csharp/src/Microsoft.ML.OnnxRuntime/OrtEnv.shared.cs
```diff
@@ -361,6 +376,28 @@ public string[] GetAvailableProviders()
             }
         }
 
+        /// <summary>
+        /// Validate a compiled model's compatibility information for one or more EP devices.
+        /// </summary>
+        /// <param name="epDevices">The list of EP devices to validate against.</param>
+        /// <param name="compatibilityInfo">The opaque compatibility information string from the precompiled model to validate.</param>
+        /// <returns>OrtCompiledModelCompatibility enum value denoting the compatibility status</returns>
+        public OrtCompiledModelCompatibility GetModelCompatibilityForEpDevices(IReadOnlyList<OrtEpDevice> epDevices, string compatibilityInfo)
```
yuslepukhin: Let's truncate all the lines to 120. #Resolved

onnxruntime/python/onnxruntime_pybind_state.cc
```diff
@@ -1575,6 +1575,17 @@ void addGlobalMethods(py::module& m) {
       R"pbdoc(Get the list of available OrtEpDevice instances.)pbdoc",
       py::return_value_policy::reference);
 
+  m.def(
+      "get_model_compatibility_for_ep_devices",
+      [](const std::vector<const OrtEpDevice*>& ep_devices,
+         const std::string& compatibility_info) -> OrtCompiledModelCompatibility {
+        OrtCompiledModelCompatibility status = OrtCompiledModelCompatibility_EP_NOT_APPLICABLE;
```
yuslepukhin, line 1582: We actually do not use public API in pybind. Call the internal function directly.

include/onnxruntime/core/session/onnxruntime_cxx_api.h
```diff
@@ -1013,6 +1013,14 @@ struct EpDevice : detail::EpDeviceImpl<OrtEpDevice> {
            ConstKeyValuePairs ep_metadata = {}, ConstKeyValuePairs ep_options = {});
 };
 
+/** \brief Validate a compiled model's compatibility for one or more EP devices.
```
yuslepukhin, line 1016: Need a good doc with all params documented. Or refer to the ABI entry point doc.

csharp/src/Microsoft.ML.OnnxRuntime/OrtEnv.shared.cs
yuslepukhin: I could find a better place, should be fine.

csharp/src/Microsoft.ML.OnnxRuntime/OrtEnv.shared.cs
adrastogi: Did this somewhat tactically (and also by hand, not sure if there's a good automatic way of doing this across the whole file without resorting to extensions).

onnxruntime/python/onnxruntime_pybind_state.cc
adrastogi: @yuslepukhin would you be able to elaborate more? I'm not quite understanding what I need to do here.

yuslepukhin: pybind statically links to onnxruntime, so public C interfaces are not used there.
In your case, I guess, it is fine, since the implementation is actually within the C interface.

csharp/src/Microsoft.ML.OnnxRuntime/OrtEnv.shared.cs
yuslepukhin: I use EditorGuidlines

yuslepukhin (approved):
:shipit:

adrastogi:
Thanks, @Craigacp! Adding a Java wrapper sounds like goodness to me. I don't have a strong opinion on the PR options, but others might have recommendations.  

---
In reply to: [3233960994](https://github.com/microsoft/onnxruntime/pull/25878#issuecomment-3233960994) [](http://example.com/codeflow?ancestors=3233960994)

snnn:

PR #25880, opened 2025-08-28
[GH Pages] Add plugin EP library documentation

adrianlizarraga:
### Description
Add a page on plugin EP libraries to the onnxruntime.ai documentation.

Preview: https://adrianlizarraga.github.io/onnxruntime/docs/execution-providers/plugin-ep-libraries.html


docs/execution-providers/community-maintained/CANN-ExecutionProvider.md
```diff
@@ -389,4 +389,3 @@ Following ops are supported by the CANN Execution Provider in single-operator In
 Additional operator support and performance tuning will be added soon.
 
 * [Ascend](https://www.hiascend.com/en/)
-* [CANN](https://www.hiascend.com/en/software/cann)
```
jywu-mysoft, line 392: removing because of broken link?  can update it to https://www.hiascend.com/cann

jywu-msft (dismissed):

docs/execution-providers/plugin-ep-libraries.md
```diff
@@ -0,0 +1,764 @@
...
+
+## Creating a plugin EP library
+A plugin EP is built as a dynamic/shared library that exports the functions `CreateEpFactories()` and `ReleaseEpFactory()`. ONNX Runtime calls `CreateEpFactories()` to obtain one or more instances of `OrtEpFactory`. An `OrtEpFactory` creates `OrtEp` instances and specifies the hardware devices supported by the EPs it creates.
+
+The ONNX Runtime repository includes a [sample plugin EP library](https://github.com/microsoft/onnxruntime/tree/main/onnxruntime/test/autoep/library), which is referenced in the following sections.
+
+### Defining an OrtEp
+An `OrtEp` represents an instance of an EP that is used by an ONNX Runtime session to identify and execute the model operations supported by the EP.
+
+The following table lists the **required** varibles and functions that an implementer must define for an `OrtEp`.
```
edgchen1:
```suggestion
The following table lists the **required** variables and functions that an implementer must define for an `OrtEp`.
```

docs/execution-providers/plugin-ep-libraries.md
```diff
@@ -0,0 +1,764 @@
...
+Implementation of this function is optional. If an EP prefers a non-default data layout, it may implement this to customize the specific op data layout preferences at a finer granularity.
+</td>
+<td></td>
+</tr>
+
+<tr>
+<td>SetDynamicOptions</td>
+<td>
+Set dynamic options on this EP. Dynamic options can be set by the application at any time after session creation with <a href="https://onnxruntime.ai/docs/api/c/struct_ort_api.html#ab1117a51683e4fbb42687c9db6e8d5fb"><code>OrtApi::SetEpDynamicOptions()</code></a>.<br/><br/>
+Implementation of this function is optional. An EP should only impliment this function if it needs to handle any dynamic options.
```
edgchen1:
```suggestion
Implementation of this function is optional. An EP should only implement this function if it needs to handle any dynamic options.
```

docs/execution-providers/plugin-ep-libraries.md
```diff
@@ -0,0 +1,764 @@
...
+</td>
+<td></td>
+</tr>
+
+<tr>
+<td>OnRunStart</td>
+<td>
+Called by ORT to notify the EP of the start of a run.<br/><br/>
+Implementation of this function is optional. An EP should only impliment this function if it needs to handle application-provided options at the start of a run.
```
edgchen1:
```suggestion
Implementation of this function is optional. An EP should only implement this function if it needs to handle application-provided options at the start of a run.
```

docs/execution-providers/plugin-ep-libraries.md
```diff
@@ -0,0 +1,764 @@
...
+</td>
+<td></td>
+</tr>
+
+<tr>
+<td>OnRunEnd</td>
+<td>
+Called by ORT to notify the EP of the end of a run.<br/><br/>
+Implementation of this function is optional. An EP should only impliment this function if it needs to handle application-provided options at the end of a run.
```
edgchen1:
```suggestion
Implementation of this function is optional. An EP should only implement this function if it needs to handle application-provided options at the end of a run.
```

docs/execution-providers/plugin-ep-libraries.md
```diff
@@ -0,0 +1,764 @@
...
+
+<tr>
+<td>ort_version_supported</td>
+<td>The ONNX Runtime version with which the EP was compiled. Implementation should set this to <code>ORT_API_VERSION</code>.</td>
+<td><a href="https://github.com/microsoft/onnxruntime/blob/3cadbdb495761a6a54845b178f9bdb811a2c8bde/onnxruntime/test/autoep/library/ep_factory.cc#L16">ExampleEpFactory()</a></td>
+</tr>
+
+<tr>
+<td>GetName</td>
+<td>Get the name of the EP that the factory creates. Must match <code>OrtEp::GetName()</code>.</td>
```
edgchen1, line 180: might be late to question this, but why do we have two separate functions for EP name?

docs/execution-providers/plugin-ep-libraries.md
```diff
@@ -0,0 +1,764 @@
...
+// NOTE: Assume plugin EP library has been registered with RegisterExecutionProviderLibrary()
+OrtStatus* status = ort_api->RegisterExecutionProviderLibrary(/*...*/);
+
+size_t num_ep_devices = 0;
+const OrtEpDevice* const* ep_devices = nullptr;
+
+status = ort_api->GetEpDevices(ort_env, &ep_devices, &num_ep_devices);
+
+// Find the OrtEpDevice for "my_ep".
+std::array<const OrtEpDevice*, 1> desired_ep_devices = { nullptr };
```
edgchen1: if we're using C++ already, should we just use the C++ API for simpler example code?

docs/execution-providers/plugin-ep-libraries.md
adrianlizarraga: I think that's because we need the EP name during device discovery (lib registration), and that happens before creating an EP instance. Then, when a session is initialized, we get the EP name from the instance itself.

docs/execution-providers/community-maintained/CANN-ExecutionProvider.md
adrianlizarraga: Thanks. updated

docs/execution-providers/plugin-ep-libraries.md
adrianlizarraga: Yep, updated to use c++

docs/execution-providers/plugin-ep-libraries.md
edgchen1: > Then, when a session is initialized, we get the EP name from the instance itself.

are we dealing with `onnxruntime::PluginExecutionProvider` instances at that point? if so, couldn't that instance get the name from the EP factory?

it seems odd to require two implementations of a GetName function that must return the same value. just wanted to check whether this is necessary.

adrianlizarraga: Hmm, yeah we are dealing with `PluginExecutionProvider` which stores the name from the factory.

PR #25888, opened 2025-08-28
 Migrate model tests to ONNX Model ZOO only

kobby-kobbs:
### Description
- Added `VERIFIED_PUBLIC_MODELS` (20 unique models) and `ONNX_ZOO_MODELS` (158 total)
- Implemented `IsModelAllowed()` function with O(1) hash lookup
- Added comprehensive name mapping for backwards compatibility
- Maintained all existing EP provider configurations

### Motivation and Context
- Part of ONNX Runtime Phase 1 migration initiative
- Addresses security requirements for public CI pipelines
- Prepares for ORT 1.23.0 release compliance


kobby-kobbs:
@microsoft-github-policy-service agree company="Microsoft"

cmake/external/onnxruntime_external_deps.cmake
```diff
@@ -599,7 +599,7 @@ if(NOT (onnx_FOUND OR ONNX_FOUND)) # building ONNX from source
 endif()
 
 if (onnxruntime_RUN_ONNX_TESTS)
-  add_definitions(-DORT_RUN_EXTERNAL_ONNX_TESTS)
+  # add_definitions(-DORT_RUN_EXTERNAL_ONNX_TESTS)
```
snnn: If you have time, please help us clean up the code.

onnxruntime/test/providers/cpu/model_tests.cc
```diff
@@ -750,6 +899,59 @@ ::std::vector<::std::basic_string<ORTCHAR_T>> GetParameterStrings() {
         std::basic_string<PATH_CHAR_TYPE> test_case_name = path.parent_path().filename().native();
         if (test_case_name.compare(0, 5, ORT_TSTR("test_")) == 0)
           test_case_name = test_case_name.substr(5);
+
+        // Check if model is in the public whitelist
+        std::string model_name_str = ToUTF8String(test_case_name);
+        if (!IsModelAllowed(model_name_str)) {
+          // Try alternative names for known models
+          if (model_name_str == "bvlc_alexnet") {
```
adrianlizarraga: Just to double check, are these alternative names still needed given that we're only running models in `/models/zoo`?

cmake/external/onnxruntime_external_deps.cmake
```diff
@@ -599,7 +599,6 @@ if(NOT (onnx_FOUND OR ONNX_FOUND)) # building ONNX from source
 endif()
 
 if (onnxruntime_RUN_ONNX_TESTS)
-  add_definitions(-DORT_RUN_EXTERNAL_ONNX_TESTS)
 endif()
```
adrianlizarraga: I think we can remove all three lines here.

onnxruntime/test/providers/cpu/model_tests.cc
kobby-kobbs: It was intended to match legacy names to model zoo names. Example, "bvlc_alexnet", is a legacy name (VM) being matched to "AlexNet" which is the original name in model zoo.

PR #25895, opened 2025-08-29
Extend ORT perf test to report submit vs inference time

gedoensmax:
This PR adds:
- ORT perf test dynamic option parsing for TRT RTX EP
- ORT perf test can now report submit vs inference time for device bindings

@chilo-ms can you help merge this ?

### DIFF
--- a/onnxruntime/test/perftest/ort_test_session.cc
+++ b/onnxruntime/test/perftest/ort_test_session.cc
@@ -20,10 +20,6 @@
 #include "TestCase.h"
 #include "strings_helper.h"
 
-#if defined(USE_CUDA) || defined(USE_TENSORRT) || defined(USE_NV)
-#include <cuda_runtime.h>
-#endif
-
 #ifdef USE_OPENVINO
 #include "nlohmann/json.hpp"
 #endif
@@ -220,12 +216,17 @@ OnnxRuntimeTestSession::OnnxRuntimeTestSession(Ort::Env& env, std::random_device
     std::string ov_string = ToUTF8String(performance_test_config.run_config.ep_runtime_config_string);
 
     ParseSessionConfigs(ov_string, provider_options);
+    if (performance_test_config.run_config.enable_cuda_io_binding) {
+      device_memory_name_ = CUDA;
+      if (cudaStreamCreate(&stream_) != cudaError_t::cudaSuccess) {
+        ORT_THROW("Unable to create CUDA stream for IOBinding");
+      }
+      auto stream_str = std::to_string(reinterpret_cast<uintptr_t>(stream_));
+      provider_options["user_compute_stream"] = stream_str;
+    }
     cuda_options.Update(provider_options);
 
     session_options.AppendExecutionProvider_CUDA_V2(*cuda_options);
-    if (performance_test_config.run_config.enable_cuda_io_binding) {
-      device_memory_name_ = CUDA;
-    }
 #else
     ORT_THROW("CUDA is not supported in this build\n");
 #endif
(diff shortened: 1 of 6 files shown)


chilo-ms:
synced offline to split the NV RTX EP code to another PR that can be cherry-picked for ORT 1.23

onnxruntime/test/perftest/ort_test_session.cc
```diff
@@ -40,20 +41,39 @@ extern const OrtApi* g_ort;
...
-  session_.Run(Ort::RunOptions{nullptr}, input_names_.data(), input.data(), input_names_.size(),
-               output_names_raw_ptr.data(), outputs_.data(), output_names_raw_ptr.size());
-
-  auto end = std::chrono::high_resolution_clock::now();
-  std::chrono::duration<double> duration_seconds = end - start;
-  return duration_seconds;
+    session_.Run(run_options, io_binding);
+    timing.submit_timing = std::chrono::high_resolution_clock::now() - start;
+    io_binding.SynchronizeOutputs();
+    timing.total_timing = std::chrono::high_resolution_clock::now() - start;
```
gedoensmax: Running async and relying on IO Binding Sync through `SyncProviders`

onnxruntime/test/perftest/ort_test_session.cc
```diff
@@ -40,20 +41,39 @@ extern const OrtApi* g_ort;
...
+    }
 
-  session_.Run(Ort::RunOptions{nullptr}, input_names_.data(), input.data(), input_names_.size(),
-               output_names_raw_ptr.data(), outputs_.data(), output_names_raw_ptr.size());
-
-  auto end = std::chrono::high_resolution_clock::now();
-  std::chrono::duration<double> duration_seconds = end - start;
-  return duration_seconds;
+    session_.Run(run_options, io_binding);
+    timing.submit_timing = std::chrono::high_resolution_clock::now() - start;
```
chilo-ms, line 64: nit: Do we want the submit time to include the time of setting up IO bindings (aka line 51 - 64), even though they might be negligible ?

onnxruntime/test/perftest/ort_test_session.cc
```diff
@@ -238,10 +258,26 @@ OnnxRuntimeTestSession::OnnxRuntimeTestSession(Ort::Env& env, std::random_device
...
+    if (!provider_options.empty()) {
+      std::cout << "Setting NV TensorRT RTX provider options to:\n";
+      for (const auto& provider_option : provider_options) {
+        std::cout << "\t" << provider_option.first << ":" << provider_option.second << "\n";
+      }
+    }
     if (performance_test_config.run_config.enable_cuda_io_binding) {
       device_memory_name_ = CUDA;
+      cudaStream_t stream;
+      cudaStreamCreate(&stream);
```
chilo-ms: Might need to call `cudaStreamDestroy` for cleanup after runs?

onnxruntime/test/perftest/ort_test_session.cc
gedoensmax: My take is that we do not want that since one could actually bind the IO bindings to the same pointers - which is even more optimal and reuse the IO Bindings.

chilo-ms: so i think the submit time for async execution should be recorded as:

```c++    
auto start = std::chrono::high_resolution_clock::now();
session_.Run(run_options, io_binding);
timing.submit_timing = std::chrono::high_resolution_clock::now() - start;
```

chilo-ms: Also, instead of binding the IO binding for every run, is it possible to bind the IO bindings only once and reuse it for the rest of the inference runs?

gedoensmax: With the current implementation it is dependent on how the `test_inputs_` are constructed and this can change on how the user specifies testing. It would be able to cache the IO binding as member and reuse if the selected index does not change, any other assumption does not very stable to me. 
Do you want me to change it in that way or should we stick with the current rebinding ? It seemed pretty fast in my profiling.

chilo-ms: okay, i think we can stick with the current rebinding and refactor it later.

PR #25908, opened 2025-09-01
[webgpu] Manually normalize `Conv-Transpose` dispatch group size for Intel GPUs

daijh:
### Description
This PR improves the performance of `Conv-Transpose` by ~12x on Lunar Lake.

For a specific shape: (3, 3, 2560, 1280), the default normalization produced a dispatch group size of (679, 679, 1) resulted in extremely slow performance on LNL (likely due to a driver issue). By manually normalize the dispatch group size to (5, 640, 160), we achieve a significant ~12x performance improvement on LNL.

### Motivation and Context
See above.

Co-authored-by: Yang, Wenqin <wenqin.yang@intel.com>

### DIFF
--- a/onnxruntime/core/providers/webgpu/nn/conv.cc
+++ b/onnxruntime/core/providers/webgpu/nn/conv.cc
@@ -41,9 +41,9 @@ Status TransposeKernel(ComputeContext& context, const Tensor* kernel, const Tens
   // TODO: Revert this change once the driver issue is fixed.
   if (context.AdapterInfo().vendor == std::string_view{"intel"}) {
     ORT_ENFORCE(rank == static_cast<size_t>(4), "Input tensor must have rank 4.");
-    uint32_t dispatch_x = ceil_div(transposed_kernel_shape_vector[0] * transposed_kernel_shape_vector[1], 2);
-    uint32_t dispatch_y = ceil_div(transposed_kernel_shape_vector[2], 4);
-    uint32_t dispatch_z = ceil_div(transposed_kernel_shape_vector[3], 8);
+    dispatch_x = ceil_div(transposed_kernel_shape_vector[0] * transposed_kernel_shape_vector[1], 2);
+    dispatch_y = ceil_div(transposed_kernel_shape_vector[2], 4);
+    dispatch_z = ceil_div(transposed_kernel_shape_vector[3], 8);
   }
 
   TensorShape transposed_kernel_shape(transposed_kernel_shape_vector);
--- a/onnxruntime/core/providers/webgpu/nn/conv.cc
+++ b/onnxruntime/core/providers/webgpu/nn/conv.cc
@@ -8,6 +8,15 @@
 #include "core/providers/webgpu/nn/grouped_conv.h"
 #include "core/providers/webgpu/webgpu_utils.h"
 #include "core/providers/webgpu/math/matmul.h"
+
+namespace {
+
+inline uint32_t ceil_div(int64_t numerator, int32_t denominator) {
+  return static_cast<uint32_t>((numerator + denominator - 1) / denominator);
+}
+
+}  // namespace
+
 namespace onnxruntime {
 namespace webgpu {
 
@@ -19,6 +28,13 @@ Status TransposeKernel(ComputeContext& context, const Tensor* kernel, const Tens
     transposed_kernel_shape_vector[i] = kernel_shape[perm[i]];
   }
   uint32_t output_size = onnxruntime::narrow<uint32_t>(kernel_shape.Size());
+
+  // Normalize dispatch = ceil_div(output_size, 64)
+  ORT_ENFORCE(rank == static_cast<size_t>(4), "Input tensor must have rank 4.");
+  uint32_t dispatch_x = onnxruntime::narrow<uint32_t>(transposed_kernel_shape_vector[0] * transposed_kernel_shape_vector[1]);
+  uint32_t dispatch_y = ceil_div(transposed_kernel_shape_vector[2], 8);
+  uint32_t dispatch_z = ceil_div(transposed_kernel_shape_vector[3], 8);
+
   TensorShape transposed_kernel_shape(transposed_kernel_shape_vector);
   *transposed_kernel = context.CreateGPUTensor(kernel->DataType(), transposed_kernel_shape);
   bool use_shared = false;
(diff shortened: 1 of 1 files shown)


daijh:
@guschmue @fs-eire @qjia7

onnxruntime/core/providers/webgpu/nn/conv.cc
```diff
@@ -19,6 +28,24 @@ Status TransposeKernel(ComputeContext& context, const Tensor* kernel, const Tens
...
+  // (10x slower) for the shape (3, 3, 2560, 1280) due to an issue with Intel's
+  // GPU drivers. We manually normalize the dispatch group size to restore
+  // performance.
+  //
+  // TODO: Revert this change once the driver issue is fixed.
+  if (context.AdapterInfo().vendor == std::string_view{"intel"}) {
+    ORT_ENFORCE(rank == static_cast<size_t>(4), "Input tensor must have rank 4.");
+    dispatch_x = ceil_div(transposed_kernel_shape_vector[0] * transposed_kernel_shape_vector[1], 2);
+    dispatch_y = ceil_div(transposed_kernel_shape_vector[2], 4);
+    dispatch_z = ceil_div(transposed_kernel_shape_vector[3], 8);
```
qjia7, line 46: I am wondering what kind of pattern of the dispatched workgroups will trigger the driver slow path. Does a dispatch group size of (679, 679, 1) resulting in extremely slow performance on LNL apply for any compute shader's dispatch or only for transpose? If it's not transpose specific, should we put the workaround in a common place for intel device?

daijh: The issue was identified on sd-turbo-unet. By default, the Conv-Transpose with (3, 3, 2560, 1280) 
took 72ms, but with this workaround, the time is reduced to approximately 6ms.

I'd prefer to request an official driver update (if this is a driver issue), instead of broadening the scope.

> what kind of pattern of the dispatched workgroups will trigger the driver slow path.

I'm not certain at this time. The results for various dispatch sizes are included below, and I will provide an update once I receive feedback from the driver.

| Time (us) | Workgroup | Dispatch  |
| :-------- | :-------- | :-------- |
|  5,609.18   | 64x1x1    | 5x640x160 |
|  6,809.02   | 64x1x1    | 1x679x679 |
|  7,271.47   | 64x1x1    | 3x640x320 |
|  7,710.00    | 64x1x1    | 9x320x160 |
| 72,187.40   | 64x1x1    | 320x160x9 |
| 72,685.90   | 64x1x1    | 679x679x1 |

qjia7: Thanks for the background. I just want to make sure we don't missing something. I am ok if this workaround won't bring any regression on existed models on Intel that @guschmue monitored.

BTW, you can do an easy test that only include one op in the model, for example, transpose with the issued shape to see whether it triggers the driver issue. If yes, your workaround should at least also apply to the transpose op in https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/providers/webgpu/tensor/transpose.cc#L103. Or maybe both of the places reuse the same DoTranspose interface.

Using the similar method, you can try another op , like matmul , to see if it has the similar issue to verify that whether other ops have the dispatching layout issue.

daijh: The dispatch normalization affects other shaders as well, though to a less significant degree than this one. 

For now, let's limit the current workaround to Conv-Transpose only, and we'll look forward to a driver-level optimization in the future.

daijh: driver version: 32.0.101.7026

qjia7: Can you refactor the code and directly call `DoTranspose` instead of `TransposeKernel` and put your workaround into `DoTranspose` so that all transpose can use this workaround? And if this's LNL specific issue, please further constraint it to LNL instead of all intel device.

qjia7: `DoTranspose` is in https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/providers/webgpu/tensor/transpose.cc#L103

daijh: Can you refactor the code and directly call DoTranspose instead of TransposeKernel and put your workaround into DoTranspose so that all transpose can use this workaround?

> After looking at the source code, the refactor is more complicated than I expected. I'll handle it in a separate PR.

And if this's LNL specific issue, please further constraint it to LNL instead of all intel device.

> Probably PTL as well. It's much safer to check the `vendor` here.

qjia7: Personally, I still prefer you use the unified DoTranspose interface and move your workaround there. Looking forward your following-up PRs. No more comments for me. Take it over to @guschmue and @fs-eire for review. Thanks.

daijh: Sure.
The refactoring to converge `TransposeKernel` and `DoTranspose` has already started.

guschmue (approved):

daijh:
Looks like a network issue on the CI server.
Could you please try to re-trigger the build?

```
Trying to download onnx-onnx-v1.18.0.tar.gz using asset cache https://vcpkg.storage.devpackages.microsoft.io/artifacts/2f38664947c8d1efc40620a7c1b1953d2aa4b0a37b67c4886b86e77c1d697363c26413413ddda8eabc545892fb1bcb43afc7e93e62f0901527524a2727e1ea8d
error: curl: (92) HTTP/2 stream 0 was not closed cleanly: INTERNAL_ERROR (err 2)
error: there were no asset cache hits, and x-block-origin blocks trying the authoritative source https://github.com/onnx/onnx/archive/v1.18.0.tar.gz
note: If you are using a proxy, please ensure your proxy settings are correct.
Possible causes are:
1. You are actually using an HTTP proxy, but setting HTTPS_PROXY variable to `https//address:port`.
This is not correct, because `https://` prefix claims the proxy is an HTTPS proxy, while your proxy (v2ray, shadowsocksr, etc...) is an HTTP proxy.
Try setting `http://address:port` to both HTTP_PROXY and HTTPS_PROXY instead.
2. If you are using Windows, vcpkg will automatically use your Windows IE Proxy Settings set by your proxy software. See: https://github.com/microsoft/vcpkg-tool/pull/77
The value set by your proxy might be wrong, or have same `https://` prefix issue.
3. Your proxy's remote server is our of service.
If you believe this is not a temporary download server failure and vcpkg needs to be changed to download this file from a different location, please submit an issue to https://github.com/Microsoft/vcpkg/issues
CMake Error at scripts/cmake/vcpkg_download_distfile.cmake:136 (message):
  Download failed, halting portfile.
Call Stack (most recent call first):
  scripts/cmake/vcpkg_from_github.cmake:120 (vcpkg_download_distfile)
  /onnxruntime_src/cmake/vcpkg-ports/onnx/portfile.cmake:3 (vcpkg_from_github)
  scripts/ports.cmake:206 (include)
```

daijh:
@guschmue could you pls apply this PR?

guschmue:
sure

PR #25909, opened 2025-09-01
[TRT RTX EP] Memory map the engine buffer

gedoensmax:
### Description

Change from fread to mmap to save on system memory. This also accelerated the load time of a ~4GB model  in my testing by 1.5X.


gedoensmax:
@chilo-ms we would like to pick this to 1.23

jywu-msft (approved):

PR #25927, opened 2025-09-02
Fix Local Attention off by 1 bug

aciddelgado:
### Description
Previously, local window size of GQA op excluded the current token. This does not match standard HuggingFace implementations where tokens are appended and then local masking occurs; the mismatch can cause the mask to be off by 1 during generation, leading to accuracy issues. This PR corrects this mismatch by including the current token. In practice, this effectively decreases GQA window size by 1.
 

### Motivation and Context
This helps align our models with HuggingFace models.


onnxruntime/contrib_ops/cpu/bert/attention_parameters.h
```diff
@@ -86,7 +86,7 @@ struct GroupQueryAttentionParameters : AttentionParameters {
   int kv_hidden_size;           // hidden size of key or value
   int seqlen_past_kv_cache;     // sequence length of past kv tensor
   int seqlen_present_kv_cache;  // sequence length of present kv tensor
-  int local_window_size;        // The window size excludes current token. It only includes tokens on the left side.
+  int local_window_size;        // The window size includes new token. It only includes tokens on the left side.
```
tianleiwu: Update the comment to avoid confusion for "only part".
Also, in prompt, there is no "new" token.

onnxruntime/contrib_ops/cpu/bert/gqa_attention_base.h
```diff
@@ -299,14 +299,14 @@ class GQAAttentionBase {
 
           // local_window_size does not include the current query token, while window_size includes it.
```
tianleiwu: update this comment

tianleiwu:
There is not backward compatible after this definition change (for all published model). 

@kunal-vaishnavi, Please double check whether our published model uses the old definition or this new one that follows huggingface.

kunal-vaishnavi:
> There is not backward compatible after this definition change (for all published model).
> 
> @kunal-vaishnavi, Please double check whether our published model uses the old definition or this new one that follows huggingface.

Models have the window size attribute in GQA [set](https://github.com/microsoft/onnxruntime-genai/blob/53cf21211c876cc6746cbe3461230ccb0a454869/src/python/py/models/builder.py#L1947) to the value used in the original `config.json` file. In the [published gpt-oss ONNX model](https://huggingface.co/onnxruntime/gpt-oss-20b-onnx/blob/main/cuda/cuda-int4-kquant-block-32-mixed/model.onnx) for example, the window size is set to 128 and matches the window size in its [`config.json` file](https://huggingface.co/openai/gpt-oss-20b/blob/main/config.json#L70).

onnxruntime/contrib_ops/cuda/bert/group_query_attention_impl.cu
```diff
@@ -476,7 +476,7 @@ Status FlashAttention(
       parameters.seqlen_present_kv_cache, kv_sequence_length, parameters.rotary_dim,
       scale, parameters.softcap, is_causal, is_bf16, parameters.use_smooth_softmax, past_bsnh, parameters.num_splits,
       reinterpret_cast<void*>(data.softmax_lse_accum), reinterpret_cast<void*>(data.out_accum),
-      parameters.local_window_size, parameters.rotary_interleaved, parameters.is_packed_qkv));
+      parameters.local_window_size - 1, parameters.rotary_interleaved, parameters.is_packed_qkv));
```
kunal-vaishnavi, line 479: Why is the -1 needed here?

aciddelgado: Flash attention is the kernel that considers current token to be implied, therefore it's window size needs to be reduced by 1 to compensate

kunal-vaishnavi: Thanks for the clarification. I also found that Hugging Face does this subtraction [here](https://github.com/huggingface/transformers/blob/5a468e56b7c04ffe2e4acd4ba0e1b64325e3b83e/src/transformers/modeling_flash_attention_utils.py#L506-L512).

tianleiwu:
@aciddelgado, One required pipeline (license/cla) not triggered. Need close/reopen, or add a new commit to trigger it.

justinchuby:
Should this be part of 1.23?

justinchuby:
Should this be part of 1.23?

tianleiwu:
@aciddelgado, there are two required pipeline not triggered.  You might need rebase your branch or merge main.

kunal-vaishnavi (approved):

PR #25928, opened 2025-09-02
[QNN-EP] Add negtive padding value support.

quic-muchhsu:
### Description

Add negative padding value support for QNN EP by lower padding to pad + stride_slice(if negative padding value exists.)

### Motivation and Context

ONNX and QNN complier in QAIRT both supports negative padding value by lowering to slice/resize_crop. Adding this change to align EP with QNN and ONNX


quic-muchhsu:
@HectorSVC I have fixed this PR according to copilot review. Please start CI. Thanks!

HectorSVC (approved):
:shipit:

PR #25930, opened 2025-09-03
[WebNN] Support IsNaN and IsInf

Honry:


Honry:

onnxruntime/core/providers/webnn/builders/impl/logical_op_builder.cc
```diff
@@ -41,9 +51,62 @@ Status LogicalOpBuilder::AddToModelBuilderImpl(ModelBuilder& model_builder, cons
...
+      const bool detect_negative = helper.Get("detect_negative", 1) != 0;
+
+      emscripten::val inf_constant = emscripten::val::undefined();
+      if (!detect_positive || !detect_negative) {
+        int32_t input0_type;
+        ORT_RETURN_IF_NOT(GetType(*input_defs[0], input0_type, logger), "Cannot get data type of input");
+        const bool is_float16 = (input0_type == ONNX_NAMESPACE::TensorProto_DataType_FLOAT16);
+
+        // WebNN only supports float and float16 data types for isInfinite.
+        ORT_RETURN_IF_NOT(input0_type == ONNX_NAMESPACE::TensorProto_DataType_FLOAT || is_float16,
```
fdwr, line 70: Shouldn't this type detection be earlier in `HasSupportedInputsImpl` so that we fall back to another EP? Also, wouldn't the opSupportLimits catch a non-{float16, float32} case?

fdwr (dismissed):
👍 given one question.

onnxruntime/core/providers/webnn/builders/impl/logical_op_builder.cc
Honry: Yes, the type detection should be covered in `HasSupportedInputsImpl`, I repeated it here to emphasize that the input of `isInfinite` is limited to fp16 and fp32. As you know, from current code in `HasSupportedInputsImpl`, we couldn't explicitly know the limited types.

onnxruntime/core/providers/webnn/builders/impl/logical_op_builder.cc
```diff
@@ -25,10 +25,20 @@ class LogicalOpBuilder : public BaseOpBuilder {
                               const emscripten::val& wnn_limits, const logging::Logger& logger) const override;
 };
 
+const std::unordered_set<std::string> UNARY_LOGICAL_OPS = {
+    "IsInf",
+    "IsNaN",
+    "Not",
+};
+
+bool IsUnaryOp(const std::string& op_type) {
```
fdwr: Can this one be a `string_view` (along with`std::unordered_set<std::string>`), since this function isn't passing the label into the WebNN API?

Honry: Good point, done.

PR #25931, opened 2025-09-03
[WebNN] Fix bug in dequantizeLinear and dequantizeLinear

Honry:
### Description
In latest WebNN spec, it requires the `scale` and `zeroPoint` rank to be same as the input rank for `dequantizeLinear` and `dequantizeLinear` operators.

### Motivation and Context
This PR adjusts the shape of scale and zeroPoint by appending or prepending 1s to their dimensions if they are not equal to the input rank to align with the latest spec.
BTW, this PR also fixes some nits and makes some code improvement.


Honry:

onnxruntime/core/providers/webnn/builders/impl/qdq_op_builder.cc
```diff
@@ -134,21 +149,11 @@ bool QDQOpBuilder::IsOpSupportedImpl(const GraphViewer&,
...
 
-  for (size_t i = 0; i < scale_shape.size(); ++i) {
-    auto scale_dim = scale_shape[scale_shape.size() - i - 1];
-    auto input_dim = input_shape[input_shape.size() - i - 1];
-    if (input_dim % scale_dim != 0) {
-      LOGS(logger, VERBOSE) << "The shape of scale is not a subsample of the shape of input";
-      return false;
-    }
-  }
-
```
Honry, line 151: Remove this because the ONNX spec already requires the scale shape to be a subsample of the input shape.

onnxruntime/core/providers/webnn/builders/impl/gemm_op_builder.cc
```diff
@@ -70,57 +71,102 @@ Status GemmOpBuilder::AddToModelBuilderImpl(ModelBuilder& model_builder, const N
...
       std::vector<int64_t> a_zero_point_shape;
       ORT_RETURN_IF_NOT(GetShape(*input_defs[2], a_zero_point_shape, logger), "Cannot get shape of a_zero_point");
-      // Scale is not used by MatMulInteger but required by DequantizeLinear. So set it to default value 1.0f.
+
+      // Adjust a_zero_point shape based on its rank and the rank of input A.
+      if (a_zero_point_shape.size() == 1) {
+        const uint32_t safe_a_zero_point = SafeInt<uint32_t>(a_zero_point_shape[0]);
+        if (a_rank == 1) {
+          // Input A is 1-D, reshaped to [1, K]. Prepend 1 to a_zero_point shape.
+          target_a_zero_point_shape.back() = safe_a_zero_point;
```
fdwr: Can these two if/else branches just be merged? If a_rank == 1, then is `target_a_zero_point_shape.back()` equivalent to `target_a_zero_point_shape[0]`?

onnxruntime/core/providers/webnn/builders/impl/gemm_op_builder.cc
```diff
@@ -70,57 +71,102 @@ Status GemmOpBuilder::AddToModelBuilderImpl(ModelBuilder& model_builder, const N
...
+      if (b_zero_point_shape.size() == 1) {
+        const uint32_t safe_b_zero_point = SafeInt<uint32_t>(b_zero_point_shape[0]);
+        if (b_rank == 1) {
+          // Input B is 1-D, reshaped to [K, 1]. Append 1 to b_zero_point shape.
+          target_b_zero_point_shape[0] = safe_b_zero_point;
+        } else {
+          // Input B is at least 2-D. Prepend 1s to b_zero_point shape.
+          target_b_zero_point_shape.back() = safe_b_zero_point;
+        }
+      }
```
fdwr: I've seen a few variants of this repeated above. Can we add a helper function that just expands the shape, kinda like the ORT DML EP's helper function that takes an input shape, new rank, and alignment (left or right). e.g.:

```
target_b_zero_point_shape = ExpandShape(b_zero_point_shape, target_rank, AxisAlignment::Right, /*offset*/ 0);
```

I'm not sure if it is possible to cleanly do this for all the cases above, but feels like it should be so for at least a few. These DML EP helpers aren't exactly what's needed, but they might be inspiring:

https://github.com/microsoft/onnxruntime/blob/b49e3b17f583a2f1a68a56c06b2365dc698e13f1/onnxruntime/core/providers/dml/OperatorAuthorHelper/OperatorHelper.h#L101

https://github.com/microsoft/onnxruntime/blob/b49e3b17f583a2f1a68a56c06b2365dc698e13f1/onnxruntime/core/providers/dml/OperatorAuthorHelper/OperatorHelper.cpp#L60

https://github.com/microsoft/onnxruntime/blob/b49e3b17f583a2f1a68a56c06b2365dc698e13f1/onnxruntime/core/providers/dml/OperatorAuthorHelper/Common.h#L51

fdwr (approved):
👍

onnxruntime/core/providers/webnn/builders/impl/gemm_op_builder.cc
Honry: No, they are not equivalent, if `a_rank==1`, the `a_shape` had been reshaped to [1, K] in previous code as WebNN only supports at least 2-D inputs. So now `target_a_zero_point_shape` rank is 2.

onnxruntime/core/providers/webnn/builders/impl/gemm_op_builder.cc
Honry: I refactor the code to make it more clear, PTAL again. :)

onnxruntime/core/providers/webnn/builders/impl/gemm_op_builder.cc
```diff
@@ -30,6 +30,72 @@ class GemmOpBuilder : public BaseOpBuilder {
                                const logging::Logger& logger) const override;
 };
 
+// Helper function
+common::Status ProcessZeroPointAndScale(
+    const std::vector<int64_t>& input_shape,
```
fdwr:
```suggestion
    gsl::span<const int64_t> input_shape,
```
Postel's law, be permissive in what you accept and explicit in what you return.

onnxruntime/core/providers/webnn/builders/impl/gemm_op_builder.cc
```diff
@@ -30,6 +30,72 @@ class GemmOpBuilder : public BaseOpBuilder {
...
+    const size_t zero_point_index,
+    emscripten::val& zero_point,
+    emscripten::val& scale,
+    ModelBuilder& model_builder,
+    const Node& node,
+    const int32_t zero_point_type,
+    const logging::Logger& logger) {
```
fdwr:
```suggestion
    ModelBuilder& model_builder,
    const Node& node,
    int32_t zero_point_type,
    const logging::Logger& logger,
    emscripten::val& zero_point,
    emscripten::val& scale) {
```
Output parameters really should go last.

fdwr (approved):
Small requests else LGTM.

onnxruntime/core/providers/webnn/builders/impl/gemm_op_builder.cc
Honry: Done.

onnxruntime/core/providers/webnn/builders/impl/gemm_op_builder.cc
Honry: Done.

fdwr (approved):
👍

PR #25868, opened 2025-08-27
[Don't review][webgpu] Make graph capture work on LLM

qjia7:
### Description

This PR includes all necessary changes to enable graph capture in LLM.
It mainly introduces support for indirect dispatch in the WebGPU FlashAttention implementation, enabling kernel launches based on runtime total sequence lengths and some necessary changes so that the whole model can run on gpu.

This pull request is intended to facilitate discussion and provide a comprehensive overview of the overall changes. Subsequently, it will be divided into smaller pull requests to make the review process more manageable.

PR #25934, opened 2025-09-03
[webgpu] Add dispatchWorkgroupsIndirect support

qjia7:
### Description
This PR adds the dispatchWorkgroupsIndirect capability for the program. It's part of the work to enable graph capture in phi4 #25868


qjia7:
@fs-eire @guschmue Please take a look, thanks.
cc @sushraja-msft

onnxruntime/core/providers/webgpu/webgpu_context.cc
```diff
@@ -279,7 +279,11 @@ Status WebGpuContext::Run(ComputeContext& context, const ProgramBase& program) {
...
   uint32_t y = program.DispatchGroupSizeY();
   uint32_t z = program.DispatchGroupSizeZ();
-  ORT_RETURN_IF_ERROR(program_mgr_->NormalizeDispatchGroupSize(x, y, z));
+
+  // Skip normalization for indirect dispatch since dimensions are determined by the indirect buffer
+  if (!program.UseIndirectDispatch()) {
+    ORT_RETURN_IF_ERROR(program_mgr_->NormalizeDispatchGroupSize(x, y, z));
+  }
 
   bool is_1d_dispatch = (y == 1 && z == 1);
```
fs-eire, line 268: does "whether use indirect dispatch or not" affect the program key ?

onnxruntime/core/providers/webgpu/webgpu_context.cc
```diff
@@ -762,14 +767,27 @@ void WebGpuContext::LaunchComputePipeline(const wgpu::ComputePassEncoder& comput
...
+                                            indirect_buffer});
   } else {
     compute_pass_encoder.SetPipeline(program_artifact.compute_pipeline);
     wgpuComputePassEncoderSetBindGroup(compute_pass_encoder.Get(), 0, bind_group, 0, nullptr);
-    compute_pass_encoder.DispatchWorkgroups(x, y, z);
+
+    if (indirect_dispatch_tensor != nullptr) {
+      // Use indirect dispatch
+      WGPUBuffer indirect_buffer = reinterpret_cast<WGPUBuffer>(const_cast<void*>(indirect_dispatch_tensor->DataRaw()));
+      compute_pass_encoder.DispatchWorkgroupsIndirect(indirect_buffer, 0);
```
fs-eire, line 765: is there guarantee that the indirect buffer is still alive when the actual flush happens? ( I assume that the buffer is required to be still alive at the flush time. please correct me if I were wrong )

qjia7: The indirect buffer is also a storage buffer. It's managed by the bucket buffer manager. Even the tensor is released, the underlying gpu buffer is still alive and it's content is dynamically updated based on the compute shader's result. Like other storage buffer, it can be reused during a batch.

onnxruntime/core/providers/webgpu/webgpu_context.cc
qjia7: It depends on the op. Theoretically, using indirect dispatch or not won't impact the shader. The GQA's change is because that we want to support graph capture and meanwhile, keep the previously path unchanged. So I updated the cacheKey in the op not here. Do you prefer update here?

guschmue (dismissed):

onnxruntime/core/providers/webgpu/program.h
```diff
@@ -348,6 +351,8 @@ class ProgramBase {
   inline uint32_t DispatchGroupSizeX() const { return dispatch_group_size_x_; }
   inline uint32_t DispatchGroupSizeY() const { return dispatch_group_size_y_; }
   inline uint32_t DispatchGroupSizeZ() const { return dispatch_group_size_z_; }
+  inline const Tensor* IndirectDispatchTensor() const { return indirect_dispatch_tensor_; }
+  inline bool UseIndirectDispatch() const { return indirect_dispatch_tensor_ != nullptr; }
```
fs-eire: it is not necessary to have `UseIndirectDispatch`. this should be able to determine by `IndirectDispatchTensor() == nullptr`

onnxruntime/core/providers/webgpu/webgpu_context.cc
```diff
@@ -279,7 +279,11 @@ Status WebGpuContext::Run(ComputeContext& context, const ProgramBase& program) {
   uint32_t x = program.DispatchGroupSizeX();
   uint32_t y = program.DispatchGroupSizeY();
   uint32_t z = program.DispatchGroupSizeZ();
-  ORT_RETURN_IF_ERROR(program_mgr_->NormalizeDispatchGroupSize(x, y, z));
+
+  // Skip normalization for indirect dispatch since dimensions are determined by the indirect buffer
```
fs-eire, line 259: could check if `SetIndirectDispatchTensor` and `SetDispatchGroupSize` are both called.

if (validation_mode >= basic && dispatch_tensor != nullptr && (dispatch_x != 0 || dispatch_y != 0 || dispatch_z != 0) {
  ERROR
}

qjia7: Done

xiaofeihan1 (approved):

PR #25935, opened 2025-09-03
Fixing the stream parameter in CopyTensors API and passing cudastream by value in createNotification API

praneshgo:
Fixing the stream parameter in CopyTensors API and passing cudastream by value in createNotification API

### Description
Fixing the stream parameter in CopyTensors API to pass the application passed stream instead of nullptr Passing cudastream by value in createNotification API as passing pointer was leading to dangling reference issues.

### Motivation and Context

- Without this code, copy tensors always happens synchronously even when app specifies a different stream to it.
- Passing pointer for cudastream in EP API is leading to dangling reference issues, hence switched to passing value.


praneshgo:
@gaugarg-nv can you please review this PR? Thanks.

PR #25937, opened 2025-09-03
[TRT RTX EP] Fixing the stream parameter in CopyTensors API and passing cudastream by value in createNotification API

praneshgo:
Fixing the stream parameter in CopyTensors API and passing cudastream by value in createNotification API

### Description

Fixing the stream parameter in CopyTensors API to pass the application passed stream instead of nullptr Passing cudastream by value in createNotification API as passing pointer was leading to dangling reference issues.
Can you please make sure that this goes into 1.23? @chilo-ms 

### Motivation and Context

- Without this code, copy tensors always happens synchronously even when app specifies a different stream to it.
- Passing pointer for cudastream in EP API is leading to dangling reference issues, hence switched to passing value.

### DIFF
--- a/onnxruntime/core/providers/cuda/cuda_provider_factory.cc
+++ b/onnxruntime/core/providers/cuda/cuda_provider_factory.cc
@@ -586,7 +586,7 @@ struct CudaSyncNotificationImpl : OrtSyncNotificationImpl {
     Release = ReleaseImpl;
   }
 
-  cudaStream_t& stream_;
+  cudaStream_t stream_;
   cudaEvent_t event_;
 
   const OrtApi& ort_api;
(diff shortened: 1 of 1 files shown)


praneshgo:
@microsoft-github-policy-service agree company="Nvidia"

chilo-ms (approved):

onnxruntime/core/framework/plugin_data_transfer.cc
```diff
@@ -41,7 +41,7 @@ Status DataTransfer::CopyTensors(const std::vector<SrcDstPair>& src_dst_pairs) c
   for (size_t i = 0; i < src_dst_pairs.size(); ++i) {
     src_values.push_back(&values[i * 2]);
     dst_values.push_back(&values[i * 2 + 1]);
-    streams.push_back(nullptr);  // static_cast<OrtSyncStream*>(src_dst_pairs[i].src_stream));
+    streams.push_back(reinterpret_cast<OrtSyncStream*>(src_dst_pairs[i].src_stream));
```
yuslepukhin, line 44: >src_stream [](http://example.com/codeflow?start=72&length=10)

Is src_stream always initialized?

praneshgo: It should always be initialized to the stream parameter of OrtApi::CopyTensors passed in the application. Application can also choose to pass nullptr for the stream.

yuslepukhin: `OrtApi::CopyTensors()` is a recent addition. The code in question is used all over the place internally and previously stream was not used.

praneshgo: Okay. But it looks like the stream is always initialized in all the paths CopyTensors is called. I see copyTensors API available in DataTransferManager and DataTransfer classes and in every path copyTensors is called, the stream parameter is initialized ahead of the call.

Issue #25882, opened 2025-08-28
[Web] WebGPU first‑run model warm‑up causes long GPU‑blocking operations (maxDispatchNumber & synchronous pipeline creation)


grazder:
### Describe the issue

During the very first inference of a model in ONNX Runtime Web, the UI freezes while a large GPU operation is executed.
The freeze comes from two sources:

1. [maxDispatchNumber](https://github.com/microsoft/onnxruntime/blob/e525ea2295c455cbb14840b075a7935a089ecb94/js/web/lib/wasm/jsep/backend-webgpu.ts#L200) is a hard‑coded constant – all dispatches are accumulated until this limit is reached, then a single massive command‑encoder submit is performed. This results in one long GPU dispatch that stalls the render thread.

2. Shader pipelines are created synchronously with [GPUDevice.createComputePipeline](https://github.com/microsoft/onnxruntime/blob/e525ea2295c455cbb14840b075a7935a089ecb94/js/web/lib/wasm/jsep/webgpu/program-manager.ts#L115) – every required compute pipeline is compiled on‑the‑fly. Because the call blocks until compilation finishes, all pipelines are compiled back‑to‑back, which ends up as a huge synchronous GPU operation on the first frame. There is [createComputePipelineAsync](https://developer.mozilla.org/en-US/docs/Web/API/GPUDevice/createComputePipelineAsync) which is async method, and there is recommendation there: 

> It is generally preferable to use this method over [GPUDevice.createComputePipeline()](https://developer.mozilla.org/en-US/docs/Web/API/GPUDevice/createComputePipeline) whenever possible, as it prevents blocking of GPU operation execution on pipeline compilation.

Both issues lead to a noticeable “jank” of animations and page interactions on the first model run.

I think maxDispatchNumber should either be set to 1 or the user should be allowed to set it and regarding the second one - it seems that createComputePipeline needs to be changed to createComputePipelineAsync.

Right now I see during profiling heavy GPU operations that blocks UI.

<img width="527" height="360" alt="Image" src="https://github.com/user-attachments/assets/abb22807-78d7-4665-82f8-aaa1bcac0597" />

### To reproduce

I use multiple small mostly Conv models. But it can be easily reproduced on transformers-js models, for example - https://huggingface.co/spaces/Xenova/realtime-whisper-webgpu

<img width="651" height="345" alt="Image" src="https://github.com/user-attachments/assets/e327fcdd-f14b-4acd-a639-1118ec59e18e" />

[...]

PR #25941, opened 2025-09-03
[WebGPU] allow async shader compilation

fs-eire:
### Description

Reduce the time blocked waiting for the shader to be compiled.

### Motivation and Context

Try to optimize the responsiveness of the application when running ort-web in main thread. See #25882

### DIFF
--- a/onnxruntime/core/providers/webgpu/program_manager.cc
+++ b/onnxruntime/core/providers/webgpu/program_manager.cc
@@ -10,6 +10,7 @@
 
 #include "core/providers/webgpu/program_manager.h"
 #include "core/providers/webgpu/shader_helper.h"
+#include "core/providers/webgpu/webgpu_context.h"
 
 namespace onnxruntime {
 namespace webgpu {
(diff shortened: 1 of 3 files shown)


grazder:
> Try to optimize the responsiveness of the application when running ort-web in main thread

I actually launch ORT-WEB in worker, so these GPU blocks appear regardless of whether it is launched in worker or in main thread

fs-eire:
> > Try to optimize the responsiveness of the application when running ort-web in main thread
> 
> I actually launch ORT-WEB in worker, so these GPU blocks appear regardless of whether it is launched in worker or in main thread

Do you mean that the UI responsiveness problem mentioned in #25882 is caused by GPU exhausted but not caused by the UI threads running JavaScript?

grazder:
> Do you mean that the UI responsiveness problem mentioned in https://github.com/microsoft/onnxruntime/issues/25882 is caused by GPU exhausted but not caused by the UI threads running JavaScript?

Yes, the main problem is that when the model initialized, it causes large GPU operations (not CPU operations in the main thread) that lock up the GPU and prevent the user interface from being rendered, which is also rendered using the GPU.

The image shows that during large GPU-based operations, frames were not rendered.
image

qjia7:
I think the async compilation is resolving the cpu issue that gpu process is occupied a long time due to shader compilation. The UI  threads' render commands have to wait on gpu process until one CreateComputePipeline is finished. So with this change, the CreateComputePipeline is moved into a gpu thread and won't block the gpu main thread so that the ui commands can send to gpu in time.
GPU busy is another issue that one ort task is too big and the ui task has to be wait on gpu. Currently we batch 16 dispatches and submit once to minimize the submit overhead. Too frequently submit will bring gpu bubbles and not friendly for small operations or models. It's challenging to determine an optimal batch size that suits all models. Maybe we could consider exposing the batch size as a session option, allowing users to customize this value to better fit their needs.

grazder:
> Maybe we could consider exposing the batch size as a session option, allowing users to customize this value to better fit their needs.

Yeah, that would be great

guschmue (dismissed):

fs-eire:

guschmue (approved):

PR #25942, opened 2025-09-04
Add onnxruntime_plugin_ep_onnx_test for plugin EPs to run onnx test

chilo-ms:
### Description
Add a separate onnx_test_runner, called it `onnxruntime_plugin_ep_onnx_test`, that supports plugin EPs using public api to compare output values.

What this PR does:

- Based on original onnx_test_runner code and enhances it as a separate program for plugin EPs to run onnx test
- Implements `CompareOrtValue` using public Apis but still keeps the old one using internal Apis.
- Make utilities functions shareable between `onnxruntime_plugin_ep_onnx_test` and `onnxruntime_perf_test`.


snnn:
How about do it in Python? Python code are easier to maintain.

chilo-ms:
> How about do it in Python? Python code are easier to maintain.

Not sure all IHVs have python setup in their CIs/Pipelines though.
Many of them rely on onnx_test_runner to run onnx test, and i think using this tool/program to test with plugin EPs should be easy.

PR #25943, opened 2025-09-04
Remove std::string::data() non-const usage from public headers

yuslepukhin:
### Description

Some compilers in our pipeline do not support string::data() nonconst


include/onnxruntime/core/session/onnxruntime_cxx_inline.h
```diff
@@ -2593,8 +2593,10 @@ struct GetValueImpl<std::string> {
     std::string result;
     if (size > 0) {
       result.resize(size);
+      // some compilers in use do not support std::string::data() non-const
```
edgchen1, line 2596: it's not a compiler support issue. if we compile with `-std=c++17` it should work.

```suggestion
      // The std::string::data() overload returning a non-const char* is only available starting from C++17.
```

PR #25960, opened 2025-09-04
Make name an std::string in OrtMemoryInfo

yuslepukhin:
### Description

OrtMemoryInfo has a name which is pointer.
With the recent changes a user can pass an arbitrary name and then deallocate the string
as the API does not require it. Make the name a `std::strng` to own it and refactor.

### Motivation and Context

We will have a dangling pointer especially when it comes to other languages.
The way it used to work in the past, is that the string would be compared to an internal constant and that constant would be used then.

### DIFF
--- a/include/onnxruntime/core/framework/ortmemoryinfo.h
+++ b/include/onnxruntime/core/framework/ortmemoryinfo.h
@@ -13,13 +13,13 @@ struct OrtMemoryInfo {
   OrtMemoryInfo() = default;  // to allow default construction of Tensor
 
   // use string for name, so we could have customized allocator in execution provider.
-  const char* name = nullptr;
+  std::string name;
   OrtMemType mem_type = OrtMemTypeDefault;
   OrtAllocatorType alloc_type = OrtInvalidAllocator;
   OrtDevice device;
 
-  constexpr OrtMemoryInfo(const char* name_, OrtAllocatorType type_, OrtDevice device_ = OrtDevice(),
-                          OrtMemType mem_type_ = OrtMemTypeDefault)
+  OrtMemoryInfo(const char* name_, OrtAllocatorType type_, OrtDevice device_ = OrtDevice(),
+                OrtMemType mem_type_ = OrtMemTypeDefault)
 #if ((defined(__GNUC__) && __GNUC__ > 4) || defined(__clang__))
       // this causes a spurious error in CentOS gcc 4.8 build so disable if GCC version < 5
       __attribute__((nonnull))
@@ -39,7 +39,7 @@ struct OrtMemoryInfo {
     if (device != other.device)
       return device < other.device;
 
-    return strcmp(name, other.name) < 0;
+    return name < other.name;
   }
 
   // This is to make OrtMemoryInfo a valid key in hash tables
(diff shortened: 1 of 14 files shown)


include/onnxruntime/core/framework/ortmemoryinfo.h
```diff
@@ -13,18 +13,18 @@ struct OrtMemoryInfo {
...
   OrtAllocatorType alloc_type = OrtInvalidAllocator;
   OrtDevice device;
 
-  constexpr OrtMemoryInfo(const char* name_, OrtAllocatorType type_, OrtDevice device_ = OrtDevice(),
-                          OrtMemType mem_type_ = OrtMemTypeDefault)
+  OrtMemoryInfo(std::string name_, OrtAllocatorType type_, OrtDevice device_ = OrtDevice(),
+                OrtMemType mem_type_ = OrtMemTypeDefault)
 #if ((defined(__GNUC__) && __GNUC__ > 4) || defined(__clang__))
       // this causes a spurious error in CentOS gcc 4.8 build so disable if GCC version < 5
       __attribute__((nonnull))
```
snnn: You may remove this annotation.

include/onnxruntime/core/framework/ortmemoryinfo.h
```diff
@@ -13,18 +13,14 @@ struct OrtMemoryInfo {
   OrtMemoryInfo() = default;  // to allow default construction of Tensor
 
   // use string for name, so we could have customized allocator in execution provider.
-  const char* name = nullptr;
+  std::string name;
```
adrianlizarraga, line 16: There's an error for DML, which is trying to use strcmp on this member:

```
Error: D:\a\_work\onnxruntime\onnxruntime\onnxruntime\core\providers\dml\DmlExecutionProvider\src\ExecutionProvider.cpp(930,16): error C2664: 'int strcmp(const char *,const char *)': cannot convert argument 1 from 'const std::string' to 'const char *' 
```

```
D:\a\_work\1\s\onnxruntime\core\providers\dml\DmlExecutionProvider\src\MLOperatorAuthorImpl.cpp(101,16): error C2664: 'int strcmp(const char *,const char *)': cannot convert argument 1 from 'const std::string' to 'const char *'
```

https://github.com/microsoft/onnxruntime/actions/runs/17481477834/job/49652496421?pr=25960#step:14:3224

adrianlizarraga: Same issue for the Win_TRT_Minimal_CUDA_Test_CI:

```
##[error]onnxruntime\test\lora\lora_test.cc(219,5): Error C2664: 'int strcmp(const char *,const char *)': cannot convert argument 1 from 'const std::string' to 'const char *
```

adrianlizarraga: Still missing these two locations:
- https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/providers/dml/DmlExecutionProvider/src/MLOperatorAuthorImpl.cpp#L101
- https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/providers/dml/DmlExecutionProvider/src/MLOperatorAuthorImpl.cpp#L1777

snnn:

PR #25962, opened 2025-09-05
[webgpu] Split large inputs into smaller buffers to bypass maxStorageBufferBindingSize limit

xiaofeihan1:
### Description
When an input is bigger than maxStorageBufferBindingSize, use multiple binding entries for it. We refine the implement for `getByOffset`/`setByOffset` so that let's say, `input_b` is 257MB, but maxStorageBufferBindingSize is 256MB, we can use `b.getByOffset(offset)` to get the correct element and no need to care about the different binding entry. Actually, it will generate shader code like this.
```
var<storage, read> input_b: array<vec4<u32>>; // [0, 256MB) of input_b
var<storage, read> input_b1: array<vec4<u32>>; // [256MB, 257MB) of input_b
```

### Motivation and Context
QC's maxStorageBufferBindingSize is 256MB, which is not enough for phi-4 model. So for QC, we customized a new phi4 model which use `slice` op to split the big matrix. That means we need to keep two different phi4 model for different platform.

This can improve ~2%(130.7->133.2) for prefill on QC X12-core X1E80100

### DIFF
--- a/.github/workflows/mac.yml
+++ b/.github/workflows/mac.yml
@@ -35,7 +35,7 @@ jobs:
       use_coreml: true
       matrix_include: >-
         [
-          {"machine": "x86_64", "target": "x86_64", "build_config": "Release"},
+          {"machine": "arm64", "target": "x86_64", "build_config": "Release"},
           {"machine": "arm64", "target": "arm64", "build_config": "Debug"},
           {"machine": "arm64", "target": "arm64", "build_config": "Release"}
         ]


onnxruntime/core/providers/webgpu/webgpu_context.cc
```diff
@@ -745,12 +754,31 @@ void WebGpuContext::OnRunEnd() {
...
 void WebGpuContext::LaunchComputePipeline(const wgpu::ComputePassEncoder& compute_pass_encoder,
                                           const std::vector<WGPUBuffer>& bind_buffers,
+                                          const std::vector<int>& bind_buffers_segments,
                                           const ProgramArtifact& program_artifact,
                                           uint32_t x, uint32_t y, uint32_t z) {
   uint32_t entry_index = 0;
   std::vector<WGPUBindGroupEntry> bind_group_entries;
-  for (WGPUBuffer buffer : bind_buffers) {
-    bind_group_entries.push_back({nullptr, entry_index++, buffer, 0, WGPU_WHOLE_SIZE, nullptr, nullptr});
+
```
xiaofeihan1, line 753: TODO: Check whether bind_buffers.size exceed limit

onnxruntime/core/providers/webgpu/program.h
```diff
@@ -226,6 +226,7 @@ struct ProgramInput {
   ProgramInput(const Tensor* tensor, ProgramTensorMetadataDependency dependency, const TensorShape& override_shape, int component);
 
   const Tensor* tensor;
+  uint32_t segments;
```
qjia7: Initialize segments to 1 so that the default value is 1 not 0.

onnxruntime/core/providers/webgpu/shader_helper.h
```diff
@@ -172,6 +172,8 @@ class ShaderHelper final {
   OStringStream additional_implementation_ss_;
   std::string body_;
   OStringStream body_ss_;
+  // Count of logical inputs added so far (each may map to multiple segmented shader variables)
+  size_t logical_input_count_ = 0;
```
qjia7: Does anywhere use it?

onnxruntime/core/providers/webgpu/webgpu_context.cc
```diff
@@ -552,7 +561,7 @@ wgpu::Limits WebGpuContext::GetRequiredLimits(const wgpu::Adapter& adapter) cons
   required_limits.maxBindGroups = adapter_limits.maxBindGroups;
   required_limits.maxComputeWorkgroupStorageSize = adapter_limits.maxComputeWorkgroupStorageSize;
   required_limits.maxComputeWorkgroupsPerDimension = adapter_limits.maxComputeWorkgroupsPerDimension;
-  required_limits.maxStorageBufferBindingSize = adapter_limits.maxStorageBufferBindingSize;
+  required_limits.maxStorageBufferBindingSize = 536870912;  // adapter_limits.maxStorageBufferBindingSize;
```
qjia7: You can change it back to adapter_limits.maxStorageBufferBindingSize now for official review :)

onnxruntime/core/providers/webgpu/webgpu_context.cc
```diff
@@ -455,18 +455,27 @@ Status WebGpuContext::Run(ComputeContext& context, const ProgramBase& program) {
...
 
   std::vector<WGPUBuffer> bind_buffers;
+  std::vector<int> bind_buffers_segments;
   bind_buffers.reserve(inputs.size() + outputs.size() + (uniform_buffer ? 1 : 0));
+  bind_buffers_segments.reserve(inputs.size() + outputs.size() + (uniform_buffer ? 1 : 0));
   for (const auto& input : inputs) {
     bind_buffers.push_back(reinterpret_cast<WGPUBuffer>(const_cast<void*>(input.tensor->DataRaw())));
+    const uint64_t size_in_bytes = static_cast<uint64_t>(input.tensor->SizeInBytes());
+    const uint64_t max_binding_size = static_cast<uint64_t>(device_limits_.maxStorageBufferBindingSize);
+    uint64_t segments = (size_in_bytes + max_binding_size - 1) / max_binding_size;
```
qjia7: Maybe directly get the segments from input.segments instead of re-calculating it again.

onnxruntime/core/providers/webgpu/program.h
```diff
@@ -287,6 +288,8 @@ class ProgramBase {
   ProgramBase& AddInput(ProgramInput&& input);
   // add multiple program inputs
   ProgramBase& AddInputs(std::initializer_list<ProgramInput> inputs);
+  // clear all program inputs
+  ProgramBase& ClearInputs();
```
qjia7: Not needed any more?

onnxruntime/core/providers/webgpu/compute_context.h
```diff
@@ -120,7 +120,25 @@ class ComputeContext {
   //
   // Run a compute shader program.
   //
-  inline Status RunProgram(const ProgramBase& program) {
+  inline Status RunProgram(ProgramBase& program) {
+    // Create new inputs with segmentation
+    std::vector<ProgramInput> new_inputs;
```
qjia7: `new_inputs` not needed?

onnxruntime/core/providers/webgpu/compute_context.h
```diff
@@ -120,7 +120,25 @@ class ComputeContext {
...
+    for (int i = 0; i < program.Inputs().size(); ++i) {
+      const auto& input = program.Inputs()[i];
+      uint32_t segments = 1;
+      if (input.tensor && input.tensor->SizeInBytes() > maxStorageBufferBindingSize) {
+        // Calculate number of segments needed
+        segments = static_cast<uint32_t>((input.tensor->SizeInBytes() + maxStorageBufferBindingSize - 1) / maxStorageBufferBindingSize);
+        if (segments == 0) segments = 1;
+      }
+      program.setSegmentsForInput(i, segments);
+    }
```
qjia7: The above code is mainly to set the segments for input, right? The best time should be in the ProgramInput construction. But it doesn't have the device limits information in program.h. Please wrap above code into a function, like updateInputSegments, and put the definition into `ProgramManager` in program_manager.h. This case is similar to `NormalizeDispatchGroupSize`. Call it in WebGpuContext::Run and sit next to `NormalizeDispatchGroupSize` for consistence.

xiaofeihan1: That makes it clearer. I’ve updated it.

onnxruntime/core/providers/webgpu/program_manager.cc
```diff
@@ -38,6 +38,22 @@ Status ProgramManager::NormalizeDispatchGroupSize(uint32_t& x, uint32_t& y, uint
...
+
+  for (size_t i = 0; i < program.Inputs().size(); ++i) {
+    const auto& input = program.Inputs()[i];
+    uint32_t segments = 1;
+    if (input.tensor && input.tensor->SizeInBytes() > maxStorageBufferBindingSize) {
+      // Calculate number of segments needed
+      segments = static_cast<uint32_t>((input.tensor->SizeInBytes() + maxStorageBufferBindingSize - 1) / maxStorageBufferBindingSize);
+      if (segments == 0) segments = 1;
+    }
+    program.setSegmentsForInput(i, segments);
```
qjia7: Move `program.setSegmentsForInput(i, segments);` into above `if` body. If the tensor size doesn't exceed maxStorageBufferBindingSize, it's not necessary to do the update.

onnxruntime/core/providers/webgpu/program_manager.cc
```diff
@@ -38,6 +38,22 @@ Status ProgramManager::NormalizeDispatchGroupSize(uint32_t& x, uint32_t& y, uint
...
+Status ProgramManager::CalculateSegmentsForInputs(ProgramBase& program) {
+  const uint64_t maxStorageBufferBindingSize = limits_.maxStorageBufferBindingSize;
+
+  for (size_t i = 0; i < program.Inputs().size(); ++i) {
+    const auto& input = program.Inputs()[i];
+    uint32_t segments = 1;
+    if (input.tensor && input.tensor->SizeInBytes() > maxStorageBufferBindingSize) {
+      // Calculate number of segments needed
+      segments = static_cast<uint32_t>((input.tensor->SizeInBytes() + maxStorageBufferBindingSize - 1) / maxStorageBufferBindingSize);
+      if (segments == 0) segments = 1;
```
qjia7: It's not possible that segments is 0, right?

onnxruntime/core/providers/webgpu/shader_helper.cc
```diff
@@ -89,9 +89,16 @@ const ShaderVariableHelper& ShaderHelper::AddInput(const std::string& name, Shad
   ORT_ENFORCE(input_index < program_.Inputs().size(),
               "Too many inputs in the program (", program_.Inputs().size(), ")");
 
-  const auto& dims = program_.Inputs()[input_index].use_override_shape ? program_.Inputs()[input_index].override_shape
-                                                                       : program_.Inputs()[input_index].tensor->Shape();
-  return AddVariableImpl(true, name, usage, dims);
+  const auto& prog_input = program_.Inputs()[input_index];
+  const TensorShape& dims = prog_input.use_override_shape ? prog_input.override_shape : prog_input.tensor->Shape();
+
+  ShaderVariableHelper& var_helper = AddVariableImpl(true, name, usage, dims);
```
qjia7: nit: Is it better if we directly use `AddVariableImpl(true, name, usage, dims, prog_input.segments, limits_.maxStorageBufferBindingSize)`?

onnxruntime/core/providers/webgpu/shader_helper.cc
```diff
@@ -418,28 +425,49 @@ Status ShaderHelper::GenerateSourceCode(std::string& code, std::vector<int>& sha
...
+      ss << "@group(0) @binding(" << binding_index++ << ") var<storage, read> ";
+      if (seg == 0) {
+        ss << input->name_;
+      } else {
+        ss << input->name_ << seg;  // naming convention matches ShaderVariableHelper::Impl usage (name + index)
+      }
+      ss << ": array<" << input->StorageType() << ">;\n";
+    }
   }
+  // outputs (currently not segmented, but handle generically if future segmentation added)
```
qjia7: Remove `(currently not segmented, but handle generically if future segmentation added)`

onnxruntime/core/providers/webgpu/shader_variable.cc
```diff
@@ -273,11 +301,32 @@ void ShaderVariableHelper::Impl(std::ostream& ss) const {
       SS_APPEND(ss, "}\n");
     }
   }
+  // Implementation of "fn get_{name}_by_offset" for multi-buffer segmented inputs
```
qjia7, line 306: Should you also support the `fn set_{name}_by_offset` for output?

onnxruntime/core/providers/webgpu/program_manager.cc
xiaofeihan1: I am worried about segment is 0 if tensor's size is 0. So I add that.

xiaofeihan1: I think you're right...The condition is that `input.tensor->SizeInBytes() > maxStorageBufferBindingSize` so it cannot be 0.

onnxruntime/core/providers/webgpu/shader_variable.cc
xiaofeihan1: Thanks. I added it.

onnxruntime/core/providers/webgpu/program_manager.cc
```diff
@@ -38,6 +38,30 @@ Status ProgramManager::NormalizeDispatchGroupSize(uint32_t& x, uint32_t& y, uint
   return Status::OK();
 }
 
+Status ProgramManager::CalculateSegmentsForInputsAndOutputs(ProgramBase& program) {
+  const uint64_t maxStorageBufferBindingSize = limits_.maxStorageBufferBindingSize;
+
+  // Inputs
+  for (size_t i = 0; i < program.Inputs().size(); ++i) {
+    const auto& input = program.Inputs()[i];
+    uint32_t segments = 1;
```
qjia7: nit: Directly assign segments in L49. Similar to L56

onnxruntime/core/providers/webgpu/webgpu_context.cc
```diff
@@ -431,18 +432,24 @@ Status WebGpuContext::Run(ComputeContext& context, const ProgramBase& program) {
   WriteTimestamp(num_pending_dispatches_ * 2);
 
   std::vector<WGPUBuffer> bind_buffers;
+  std::vector<uint32_t> bind_buffers_segments;
   bind_buffers.reserve(inputs.size() + outputs.size() + (uniform_buffer ? 1 : 0));
+  bind_buffers_segments.reserve(inputs.size() + outputs.size() + (uniform_buffer ? 1 : 0));
```
qjia7: nit: Only calculate `inputs.size() + outputs.size() + (uniform_buffer ? 1 : 0)` once and reuse it.

onnxruntime/test/contrib_ops/matmul_4bits_test.cc
```diff
@@ -599,6 +604,26 @@ TEST(MatMulNBits, Float16_Large) {
...
+// output size(float32): M * N * 4 = 4 * 8388612 * 4 = 134217792 bytes > 134217728 bytes (128MB)
+TEST(MatMulNBits, Float32_Large) {
+  // Keep tolerance similar to Float16_Large (float path typically equal or better numerically).
+  constexpr float abs_error = 0.1f;
+  constexpr bool zp_is_4bit = true;
+
+  // Use a single block_size (16) to minimize additional memory/time while still covering the large-output case.
+  for (auto block_size : {16}) {
+    for (auto has_zeropoint : {false}) {
+      RunTest<float>(4 /*M*/, 8388612 /*N*/, 32 /*K*/, block_size, has_zeropoint, zp_is_4bit, abs_error);
```
qjia7: If you only test one case, directly use `RunTest<float>(4 /*M*/, 8388612 /*N*/, 32 /*K*/, 32, false, zp_is_4bit, abs_error);`?

BTW, why block_size (16) can minimize additional memory?

qjia7 (dismissed):
Overall looks good to me with nits. Thanks for your iterations and patience.

onnxruntime/test/contrib_ops/matmul_4bits_test.cc
xiaofeihan1: The comment is wrong. Change the code and done!

xiaofeihan1:
reopen the PR to re-trigger pipeline

qjia7 (approved):

PR #25966, opened 2025-09-05
Fix Attention GQA implementation on CPU

xadupre:
### Description
Attention on CPU is following ONNX specifications. This change replicates the changes introduced by https://github.com/onnx/onnx/pull/7274.


onnxruntime/test/onnx/main.cc
```diff
@@ -795,6 +795,24 @@ select from 'TF8', 'TF16', 'UINT8', 'FLOAT', 'ITENSOR'. \n)");
     // Please make no more changes to the list
     static const ORTCHAR_T* immutable_broken_tests[] =
         {
+            // pending ONNX update
```
titaiwangms, line 798: Is this the right place to disable the tests? See line 795. We should add them into backend test and Test.cc. I never touched this file.

onnxruntime/core/providers/cpu/llm/attention.cc
```diff
@@ -62,14 +62,14 @@ void make_copy<MLFloat16, MLFloat16>(MLFloat16* mask_data, const MLFloat16* mask
 template <>
 void make_copy<float, bool>(float* mask_data, const bool* mask_index, size_t size) {
   for (size_t i = 0; i < size; ++i) {
-    mask_data[i] = mask_index[i] ? 0.0f : std::numeric_limits<float>::lowest();
+    mask_data[i] = mask_index[i] ? 0.0f : -std::numeric_limits<float>::infinity();
```
kunal-vaishnavi: Should these `make_copy` changes also use the template functions (e.g. `negative_infinity`)?

onnxruntime/core/providers/cpu/llm/attention.cc
```diff
@@ -188,6 +188,16 @@ Status Attention<T>::Compute(OpKernelContext* context) const {
   );
 }
 
+template <typename T>
+T negative_infinity() {
+  return -std::numeric_limits<T>::infinity();
+}
+
```
kunal-vaishnavi: Should these functions be in a header file to be leveraged across multiple files and use `inline`?

onnxruntime/test/onnx/main.cc
xadupre: I never remember when it has to be done. This will be removed when onnx is updated anyway.

onnxruntime/core/providers/cpu/llm/attention.cc
xadupre: I forgot one. Thanks

onnxruntime/core/providers/cpu/llm/attention.cc
xadupre: will do

onnxruntime/test/onnx/main.cc
titaiwangms: Could you please move it here: https://github.com/microsoft/onnxruntime/blob/ec3bf7f03d9363ebf5c6c952a7f017fc42d7417f/onnxruntime/test/testdata/onnx_backend_test_series_filters.jsonc#L31

That way, I don't need an extra step to find them when I update ONNX.

tianleiwu (approved):
Any change of attention value from changing masked value from lowest() to -inf? I think the result shall be close enough.

justinchuby:
There is one case where we see NaN from softmax: https://github.com/microsoft/onnxscript/blob/a70ee8d0905f563c840bbd5338595e9ac6b1b5b4/onnxscript/function_libs/torch_lib/ops/nn.py#L2147-L2155

Not sure if it is related

justinchuby:
I wonder if there are benefits in using -inf? Should we change https://github.com/microsoft/onnxscript/blob/a70ee8d0905f563c840bbd5338595e9ac6b1b5b4/onnxscript/function_libs/torch_lib/ops/nn.py#L2141 to use `lowest()`?

xadupre:
The reason why I switch to inifinity is that there is a case where the mask is added to the input. If it is infinity, any value won't change it unless it is infinity itself or nan. That's not true for lowest. I don't think it is likely to happen than the mask is == -lower and the input is +lowest ~highest.

justinchuby:
Question is do we need special handling for softmax, when mask is -inf?

    # When using scaled dot product attention with a boolean mask, the softmax operation might return NaN values
    # due to the presence of -inf in an entire row (padding tokens), resulting in 0/0 (NaN) in the softmax output.
    # This is because there's no safe/masked softmax imp in ONNX, so we need to handle NaN values explicitly to match
    # the behavior of PyTorch with boolean masks.
    # Reference: https://github.com/pytorch/pytorch/issues/103749

xadupre:
> Question is do we need special handling for softmax, when mask is -inf?
> 
> ```
> # When using scaled dot product attention with a boolean mask, the softmax operation might return NaN values
> # due to the presence of -inf in an entire row (padding tokens), resulting in 0/0 (NaN) in the softmax output.
> # This is because there's no safe/masked softmax imp in ONNX, so we need to handle NaN values explicitly to match
> # the behavior of PyTorch with boolean masks.
> # Reference: https://github.com/pytorch/pytorch/issues/103749
> ```

I see your point. We should add a unit test with such a mask.

justinchuby (approved):

PR #25967, opened 2025-09-05
Update React Native iOS CI build.

edgchen1:
### Description

Update React Native iOS CI build.
- Split iOS package build, unit test, and E2E test into separate jobs. The finer granularity allows for more parallelism and the ability to re-run a specific test job in case of failure due to test flakiness.
- Fix issue with detox not actually running the e2e tests on iOS due to an earlier failure. The issue was resolved by upgrading to a more recent version.
- Reduce iOS package build time by not building ORT test code. That's already tested elsewhere.

### Motivation and Context

CI pipeline improvements.


tools/ci_build/github/azure-pipelines/templates/react-native-ci.yml
```diff
@@ -178,50 +176,89 @@ stages:
...
+        TargetPath: '$(Build.BinariesDirectory)/ios_pod'
+
+    - template: use-xcode-version.yml
+
+    - task: NodeTool@0
+      inputs:
+        versionSpec: '22.x'
+
+    - script:
+        brew install npm
```
snnn: The machine should already have npm. Search it in "https://github.com/actions/runner-images/blob/main/images/macos/macos-14-Readme.md"

tools/ci_build/github/js/react_native_e2e_full_ios_framework_build_settings.json
```diff
@@ -10,7 +10,8 @@
             "--use_xcode",
             "--build_apple_framework",
             "--use_coreml",
-            "--skip_tests"
+            "--skip_tests",
+            "--cmake_extra_defines=onnxruntime_BUILD_UNIT_TESTS=OFF"
```
snnn, line 14: Is it for reducing iOS package build time ?

edgchen1: that's right.

PR #25968, opened 2025-09-05
[QNN EP] fix memory leak in OrtEpFactory::GetSupportedDevices()

adrianlizarraga:
### Description
Fixes memory leak in `OrtEpFactory::GetSupportedDevices()`. The `OrtKeyValuePairs` instance created by the factory was not released.

This memory leak can be reproduced by running the unit test [QnnHTPBackendTests.AutoEp_PreferNpu](https://github.com/microsoft/onnxruntime/blob/4e2699fbf24d96d2a0261b30509864da08b701de/onnxruntime/test/providers/qnn/qnn_basic_test.cc#L1426).

### Motivation and Context
Fix memory leak that one encounters when using automatic EP selection (e.g., PREFER_NPU) with QNN EP.

### DIFF
--- a/onnxruntime/core/providers/qnn/qnn_provider_factory.cc
+++ b/onnxruntime/core/providers/qnn/qnn_provider_factory.cc
@@ -219,9 +219,11 @@ struct QnnEpFactory : OrtEpFactory {
         OrtKeyValuePairs* ep_options = nullptr;
         factory->ort_api.CreateKeyValuePairs(&ep_options);
         factory->ort_api.AddKeyValuePair(ep_options, "backend_path", factory->qnn_backend_path.c_str());
-        ORT_API_RETURN_IF_ERROR(
-            factory->ort_api.GetEpApi()->CreateEpDevice(factory, &device, nullptr, ep_options,
-                                                        &ep_devices[num_ep_devices++]));
+        OrtStatus* status = factory->ort_api.GetEpApi()->CreateEpDevice(factory, &device, nullptr, ep_options,
+                                                                        &ep_devices[num_ep_devices++]);
+
+        factory->ort_api.ReleaseKeyValuePairs(ep_options);
+        ORT_API_RETURN_IF_ERROR(status);
       }
     }
 


onnxruntime/core/providers/qnn/qnn_provider_factory.cc
```diff
@@ -219,9 +219,11 @@ struct QnnEpFactory : OrtEpFactory {
         OrtKeyValuePairs* ep_options = nullptr;
         factory->ort_api.CreateKeyValuePairs(&ep_options);
         factory->ort_api.AddKeyValuePair(ep_options, "backend_path", factory->qnn_backend_path.c_str());
-        ORT_API_RETURN_IF_ERROR(
-            factory->ort_api.GetEpApi()->CreateEpDevice(factory, &device, nullptr, ep_options,
-                                                        &ep_devices[num_ep_devices++]));
+        OrtStatus* status = factory->ort_api.GetEpApi()->CreateEpDevice(factory, &device, nullptr, ep_options,
+                                                                        &ep_devices[num_ep_devices++]);
+
+        factory->ort_api.ReleaseKeyValuePairs(ep_options);
```
adrianlizarraga, line 225: The alternative is to use the C++ API but that would require a larger code change (handle exceptions, initialize C++ global api object).

jywu-msft (approved):

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
yuslepukhin: Not really, you could capture status with Ort::Status and the same for keyvalue pairs and then decide what to do next.

adrianlizarraga: I'd prefer to defer that change. QNN EP is built as both a static and shared library, which impacts how the C++ api object should be initialized. Also, while one could mix use of the C API with C++ api to avoid exceptions, I'd prefer to let QC handle that to keep this change small.

snnn:

PR #25971, opened 2025-09-06
Revert accidental memory alignment change for x86 for prepacked weights buffer

hariharans29:
### Description
The memory alignment for the pre-packed weights buffer was accidentally changed for 8-bit Gemms on x86 while supporting the ARM64 equivalent kernel in https://github.com/microsoft/onnxruntime/pull/25110. This change in alignment could either cause perf penalty or seg-fault depending on the platform while the corresponding aligned data load instruction is executed in the Gemm kernel. 

This changes fixes it as well as adds back a couple of tests to the MLAS 8-bit Gemm test suite and fixes a minor nit in the test file.

### Motivation and Context
Resolve packaging pipeline crash


jywu-msft (approved):

snnn:

Issue #25460, opened 2025-07-20
Bug in CreateTensorValueFromSystemNumericsTensorObject


tjwald:
The api CreateTensorValueFromSystemNumericsTensorObject throws the following error: 

```
Unhandled exception. System.MissingMethodException: Method not found: 'Void System.Numerics.Tensors.Tensor`1.CopyTo(System.Numerics.Tensors.TensorSpan`1<!0>)'.
   at Microsoft.ML.OnnxRuntime.OrtValue.CreateTensorValueFromSystemNumericsTensorObject[T](Tensor`1 tensor)
   at Microsoft.ML.OnnxRuntime.OrtValue.CreateTensorValueFromSystemNumericsTensorObject[T](Tensor`1 tensor)
   at Program.<Main>$(String[] args) in D:\Code\NET\ML\fai\Example\FailingTensor.cs:line 15
```

When running:
```C#
#:package System.Numerics.Tensors@10.0.0-preview.6.* 
#:package Microsoft.ML.OnnxRuntime.Gpu@1.22.1

#:property NoWarn=$(NoWarn);SYSLIB5001


using System.Numerics.Tensors;
using Microsoft.ML.OnnxRuntime;

Span<nint> length = [1000, 1];
Span<nint> strides = [1, 0];

Tensor<long> tensor = Tensor.Create<long>(length, strides);

OrtValue.CreateTensorValueFromSystemNumericsTensorObject(tensor);
```

Also, looking at the implementation of CreateTensorValueFromSystemNumericsTensorObject, there is a bug/performance issue with `IsContiguousAndDense` - 

If I have a tensor of shape [1000, 1] then the strides will be [1, 0] and contiguous and dense.
But then both checks fail, and it returns false even though it should return true.

tjwald:
I don't mind fixing these bugs

tjwald:
```csharp
[Experimental("SYSLIB5001")]
public static OrtValue CreateTensorValueFromSystemNumericsTensorObject<T>(SystemNumericsTensors.Tensor<T> tensor) where T : unmanaged
{
    if (!tensor.IsDense)
    {
        tensor = tensor.ToDenseTensor();
    }
    unsafe
    {
        var memHandle = tensor.GetPinnedHandle();

        try
        {
            IntPtr dataBufferPointer = IntPtr.Zero;
            unsafe
            {
                dataBufferPointer = (IntPtr)memHandle.Pointer;
            }

[...]

PR #25972, opened 2025-09-06
fix breaking change

tjwald:
### Description

Fixing a breaking change from the System.Numerics.Tensors package.
See: https://github.com/microsoft/onnxruntime/issues/25460


tjwald:

Issue #25823, opened 2025-08-22
Incorrect calculation of zero point for uint8 symmetric quantization


Akupadhye:
### Describe the issue

zero_point is incorrectly calculated as 127 instead of 128 for uint8 symmetric quantized case in [compute_scale_zp()](https://github.com/microsoft/onnxruntime/blob/17b2725777fdc8d3c639f58985693ebbb3e588b5/onnxruntime/python/tools/quantization/quant_utils.py#L304) function. This seems to be due to the update of qmax in [ONNX_INT_TYPE_SYMMETRIC_RANGE ](https://github.com/microsoft/onnxruntime/blob/17b2725777fdc8d3c639f58985693ebbb3e588b5/onnxruntime/python/tools/quantization/quant_utils.py#L166)done in this [PR](https://github.com/microsoft/onnxruntime/pull/22020) as earlier qmax was 255, but now its updated to 254

### To reproduce

-

### Platform

Windows

### OS Version

22H2

### ONNX Runtime Installation

Built from Source

### ONNX Runtime Version or Commit ID

1.23.0

### ONNX Runtime API

Python

### Architecture

X64

### Execution Provider

Default CPU

PR #25986, opened 2025-09-08
[QNN] Fix multiple issues with per_channel quantization of QNN models

qti-kromero:
### Description
* Remove workaround for input1 of FC not supporting uint quantization
* Remove workaround for PCQ not being supported on bias of layernorm
* Remove workaround for PCQ/uint conv weights unsupported
* Fix issue with zero point not being calculated correctly for uint symmetric quantization (https://github.com/microsoft/onnxruntime/issues/25823)

### Motivation and Context
* Fixes several issues faced by our development team when bringing up a model with PCQ


HectorSVC:
some tests need to be reverted?

qti-kromero:
> some tests need to be reverted?

it seems so yes, do you happen to know if there's a specific build flag needed to trigger these quantization tests? we didn't observe this failure during internal testing and would be good to align

qti-kromero:
> > some tests need to be reverted?
> 
> it seems so yes, do you happen to know if there's a specific build flag needed to trigger these quantization tests? we didn't observe this failure during internal testing and would be good to align

issue with internal invocation of tests - working to enable them and fix this issue

qti-kromero:
@HectorSVC test is updated, would expect this to pass now

PR #25987, opened 2025-09-08
[QNN-EP] Enable Optrace from QNN into QNN EP

quic-calvnguy:
### Description
 - Add optrace profiling level
 - Add profiling to compose graph
 - Add new qnn system profile serializer class
 - Add API versioning safeguards
 - Use current behavior for QNN API < 2.29
 - Use QNN System Profile API for QNN API >= 2.29
 - Check for log file at end of profiling unit test
 - Ensure system libs are loaded when profiling is enabled

### Motivation and Context
Adds optrace level profiling for debugging purposes. Utilizes new QNN System Profile API to generate a binary log file compatible with qnn-profile-viewer executable for all profiling levels when QNN API is >= 2.29 (i.e. QAIRT 2.39 or later).

Current behavior will still persist regardless of QNN API version (i.e. a .csv file will still be generated).

The binary log file naming will be based on the existing profiling_file_path EP option. The value of profiling_file_path is expected to be .csv as before, and the binary log name will remove ".csv" and append "_qnn.log". For example, if profiling_file_path is set to "foo.csv", then both "foo.csv" and "foo_qnn.log" will be generated.

With optrace enablement, qnn-profile-viewer can then be used to generate and view chrometraces with a user-friendly UI.
For more details, see:
https://docs.qualcomm.com/bundle/publicresource/topics/80-63442-50/htp_backend.html#qnn-htp-optrace-profiling
https://docs.qualcomm.com/bundle/publicresource/topics/80-63442-50/htp_backend.html#qnn-htp-analysis-summary-qhas-

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
+++ b/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
@@ -233,7 +233,9 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
   // Resets the context priority to the session default as defined by context_priority_
   Status ResetContextPriority();
 
+#ifdef QNN_SYSTEM_PROFILE_API_ENABLED
   bool ProfilingEnabled() { return profiling_enabled_; }
+#endif
 
  private:
   Status LoadBackend();
@@ -436,7 +438,11 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
   ProfilingLevel profiling_level_merge_;
   const std::string profiling_file_path_;
   bool system_lib_loaded_ = false;
+
+#ifdef QNN_SYSTEM_PROFILE_API_ENABLED
   bool profiling_enabled_ = false;
+#endif
+
   bool backend_initialized_ = false;
   bool device_created_ = false;
   bool context_created_ = false;


onnxruntime/core/providers/qnn/builder/qnn_profile_serializer.h
```diff
@@ -0,0 +1,143 @@
...
+
+#include <QnnInterface.h>
+#include <QnnProfile.h>
+#include <System/QnnSystemProfile.h>
+
+#endif
+
+namespace onnxruntime {
+
+using namespace common;
```
edgchen1: we should not put namespace using directives in header files. they pollute the namespace for any file that includes this file.

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf7-dont-write-using-namespace-at-global-scope-in-a-header-file

onnxruntime/core/providers/qnn/builder/qnn_profile_serializer.h
```diff
@@ -0,0 +1,143 @@
...
+
+  std::string qnn_log_filename_;
+  std::string output_directory_;
+  std::vector<QnnSystemProfile_ProfileEventV1_t> event_data_list_;
+  std::unordered_map<QnnProfile_EventId_t, QnnSystemProfile_ProfileEventV1_t*> system_parent_event_lookup_map_;
+  std::unordered_map<QnnProfile_EventId_t, QnnSystemProfile_ProfileEventV1_t*> event_profile_id_lookup_map_;
+  std::unordered_map<QnnSystemProfile_ProfileEventV1_t*, std::vector<QnnSystemProfile_ProfileEventV1_t> > sub_event_data_map_;
+
+  QNN_SYSTEM_INTERFACE_VER_TYPE qnn_system_interface_;
+#endif
```
edgchen1: general: for ifdefs that span more than a few lines, please put a comment next to associated `#endif` or `#elif` to indicate what condition it is associated with.
```suggestion
#endif  // defined(QNN_SYSTEM_PROFILE_API_ENABLED)
```

onnxruntime/core/providers/qnn/builder/qnn_profile_serializer.h
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  Status InitFile();
+
+ private:
+  void cleanup() {
+    if (outfile_.is_open()) {
+      outfile_.close();
+    }
```
edgchen1: wouldn't this already be handled by the std::ofstream dtor?

onnxruntime/core/providers/qnn/builder/qnn_profile_serializer.cc
```diff
@@ -0,0 +1,445 @@
...
+  status = qnn_system_interface_.systemProfileSerializeEventData(serialization_target_handle,
+                                                                 system_profile_data_list.data(),
+                                                                 1);
+
+  ORT_RETURN_IF(QNN_SYSTEM_PROFILE_NO_ERROR != status, "Failed to create serialize optrace data");
+
+  status = qnn_system_interface_.systemProfileFreeSerializationTarget(serialization_target_handle);
+  ORT_RETURN_IF(QNN_SYSTEM_PROFILE_NO_ERROR != status, "Failed to free serialization target");
+
+  // explicitly call clear on objs so they are not optimized out prematurely during compilation
```
edgchen1: "optimized out prematurely during compilation"

can you elaborate on what the concern is here?

onnxruntime/core/providers/qnn/builder/qnn_profile_serializer.cc
```diff
@@ -0,0 +1,445 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include "qnn_profile_serializer.h"
+#include "core/providers/qnn/qnn_telemetry.h"
+
+using namespace onnxruntime;
```
edgchen1: nit: instead of using namespace directives, could we just put the function definitions into the correct namespace?

onnxruntime/core/providers/qnn/builder/qnn_profile_serializer.cc
```diff
@@ -0,0 +1,445 @@
...
+    it->second.reserve(num_sub_events);
+  }
+}
+
+QnnSystemProfile_ProfileEventV1_t* Serializer::GetSystemEventPointer(QnnProfile_EventId_t event_id) {
+  if (event_profile_id_lookup_map_.find(event_id) == event_profile_id_lookup_map_.end()) {
+    return nullptr;
+  }
+
+  return event_profile_id_lookup_map_.at(event_id);
```
edgchen1: general nit: instead of looking up twice, can just reuse the initial result of `find()`

onnxruntime/core/providers/qnn/builder/qnn_profile_serializer.cc
```diff
@@ -0,0 +1,445 @@
...
+  system_profile_data.v1.numProfilingEvents = static_cast<uint32_t>(event_data_list_.size());
+
+  std::vector<const QnnSystemProfile_ProfileData_t*> system_profile_data_list = {&system_profile_data};
+  status = qnn_system_interface_.systemProfileSerializeEventData(serialization_target_handle,
+                                                                 system_profile_data_list.data(),
+                                                                 1);
+
+  ORT_RETURN_IF(QNN_SYSTEM_PROFILE_NO_ERROR != status, "Failed to create serialize optrace data");
+
+  status = qnn_system_interface_.systemProfileFreeSerializationTarget(serialization_target_handle);
```
edgchen1: consider using RAII to manage the serialization target handle. if this function exits between the call to create it and the call to free it, the handle will get leaked.

onnxruntime/core/providers/qnn/builder/qnn_profile_serializer.cc
```diff
@@ -0,0 +1,445 @@
...
+}
+
+Serializer::Serializer(const ProfilingInfo& profiling_info)
+    : qnn_system_interface_(profiling_info.qnn_system_interface),
+      profiling_info_(profiling_info),
+      tracelogging_provider_ep_enabled_(profiling_info.tracelogging_provider_ep_enabled) {
+  std::filesystem::path output_fs_filepath(profiling_info.output_filepath);
+  qnn_log_filename_ = output_fs_filepath.filename().string();
+  // Remove extension (assumed to be ".csv") then add "_qnn.log"
+  size_t extension_start_idx = qnn_log_filename_.find(".");
```
edgchen1: what if the file doesn't end in `.csv`, or has more than one `.` in the name?

onnxruntime/core/providers/qnn/builder/qnn_profile_serializer.cc
```diff
@@ -0,0 +1,445 @@
...
+  system_profile_data.v1.header.methodType = ParseMethodType(method_type);
+  system_profile_data.v1.profilingEvents = event_data_list_.data();
+  system_profile_data.v1.numProfilingEvents = static_cast<uint32_t>(event_data_list_.size());
+
+  std::vector<const QnnSystemProfile_ProfileData_t*> system_profile_data_list = {&system_profile_data};
+  status = qnn_system_interface_.systemProfileSerializeEventData(serialization_target_handle,
+                                                                 system_profile_data_list.data(),
+                                                                 1);
+
+  ORT_RETURN_IF(QNN_SYSTEM_PROFILE_NO_ERROR != status, "Failed to create serialize optrace data");
```
edgchen1: "create serialize optrace data" - should it be "serialize data"? is it specific to optrace data?

also, is there more information associated with the error code that we can also include in the message?

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1629,35 +1675,38 @@ Status QnnBackendManager::ExtractBackendProfilingInfo() {
...
+          ExtractProfilingEvent(*(profile_events + event_idx), "ROOT", profile_writer.get(),
+                                backendSupportsExtendedEventData));
       ORT_RETURN_IF_ERROR(
-          ExtractProfilingSubEvents(*(profile_events + event_idx), outfile, backendSupportsExtendedEventData,
-                                    tracelogging_provider_ep_enabled));
+          ExtractProfilingSubEvents(*(profile_events + event_idx), profile_writer.get(),
+                                    backendSupportsExtendedEventData));
     }
+#ifdef QNN_SYSTEM_PROFILE_API_ENABLED
+    if (profile_writer) {
```
edgchen1: seems like this condition is not required - `profile_writer` always gets set to `std::make_unique<profile::Serializer>(profiling_info)` above. also, does `profile_writer` need to be a `std::unique_ptr` (with the dynamic allocation) instead of a local variable on the stack?

onnxruntime/test/providers/qnn/qnn_basic_test.cc
```diff
@@ -1108,6 +1108,74 @@ TEST_F(QnnHTPBackendTests, ProfilingTest) {
...
                   ExpectedEPNodeAssignment::All,
                   0.008f);
+
+  // Check that output csv file exists and is non-empty
+  std::ifstream csv_file(provider_options["profiling_file_path"], std::ifstream::binary);
+  EXPECT_TRUE(csv_file.good());
+
+  csv_file.seekg(0, csv_file.end);
+  size_t buffer_size = static_cast<size_t>(csv_file.tellg());
+  EXPECT_FALSE(0 == buffer_size);
```
edgchen1: where possible, use a more specific gtest assertion (e.g., `EXPECT_NE` here) to get a better error message.

onnxruntime/test/providers/qnn/qnn_basic_test.cc
```diff
@@ -1108,6 +1108,74 @@ TEST_F(QnnHTPBackendTests, ProfilingTest) {
                   13,
                   ExpectedEPNodeAssignment::All,
                   0.008f);
+
+  // Check that output csv file exists and is non-empty
```
edgchen1: maybe this logic to check that a file exists and is not empty can be placed in a helper function and reused

onnxruntime/test/providers/qnn/qnn_basic_test.cc
```diff
@@ -1108,6 +1108,74 @@ TEST_F(QnnHTPBackendTests, ProfilingTest) {
...
+
+  std::remove("detailed_profile_qnn.log");
+#endif
+}
+
+TEST_F(QnnHTPBackendTests, OptraceTest) {
+  onnxruntime::ProviderOptions provider_options;
+
+#if defined(_WIN32)
+  provider_options["backend_path"] = "QnnHtp.dll";
```
edgchen1: fyi, there's a "backend_type" option that is a platform-agnostic way of specifying backend

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
```diff
@@ -225,6 +233,8 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
   // Resets the context priority to the session default as defined by context_priority_
   Status ResetContextPriority();
 
+  bool ProfilingEnabled() { return use_system_profiling_api_; }
```
edgchen1: between this and `#ifdef QNN_SYSTEM_PROFILE_API_ENABLED`, why do we need two ways to check?

onnxruntime/core/providers/qnn/builder/qnn_profile_serializer.cc
quic-calvnguy: The output file from ofstream is a .csv file and is defined as a .csv file in the ORT QNN EP documentation (see profiling_file_path):
https://onnxruntime.ai/docs/execution-providers/QNN-ExecutionProvider.html#configuration-options


 Will handle the second case though

onnxruntime/core/providers/qnn/builder/qnn_profile_serializer.cc
quic-calvnguy: By "optimized out" I meant that the containers used to hold all the data for serialization were unused after their original initialization. I was worried that the compiler will see they are unused after initialization and, thus, destruct the contains prematurely before or during the serialization process. A similar issue was found during the VTCM backup buffer verification, so this is why I added this bit of code here.

I've removed this bit of code in the upcoming commit as I've verified lines 378-381 are not needed.

onnxruntime/core/providers/qnn/builder/qnn_profile_serializer.h
```diff
@@ -0,0 +1,155 @@
...
+    event_profile_id_lookup_map_.clear();
+    sub_event_data_map_.clear();
+#endif
+  }
+
+  Status InitFile();
+
+#ifdef QNN_SYSTEM_PROFILE_API_ENABLED
+ public:
+  Status SerializeEvents();
```
edgchen1: can the public methods be grouped together? and then the private methods?

onnxruntime/core/providers/qnn/builder/qnn_profile_serializer.h
```diff
@@ -0,0 +1,155 @@
...
+  uint64_t start_time = 0;
+  uint64_t stop_time = 0;
+  uint32_t num_events = 0;
+
+  ProfilingMethodType method_type = ProfilingMethodType::UNKNOWN;
+  QNN_SYSTEM_INTERFACE_VER_TYPE qnn_system_interface = QNN_SYSTEM_INTERFACE_VER_TYPE_INIT;
+#endif
+};
+
+class Serializer {
```
edgchen1, line 43: it would be helpful to have some documentation for this class and its methods or more descriptive method names. e.g., it wasn't obvious to me that `ProcessEvent()` will to a file but `SerializeEvents()` does not.

onnxruntime/core/providers/qnn/builder/qnn_profile_serializer.cc
```diff
@@ -0,0 +1,466 @@
...
+      return "Memory Allocation Error";
+    default:
+      return "Unknown";
+  }
+}
+
+Serializer::Serializer(const ProfilingInfo& profiling_info)
+    : qnn_system_interface_(profiling_info.qnn_system_interface),
+      profiling_info_(profiling_info),
+      tracelogging_provider_ep_enabled_(profiling_info.tracelogging_provider_ep_enabled) {
```
edgchen1: it seems redundant to save a copy of `profiling_info` in the `profiling_info_` data member and also have data members that are copies of data members of `profiling_info`.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -639,17 +642,44 @@ Status QnnBackendManager::InitializeProfiling() {
...
+    qnn_profile_level = QNN_PROFILE_LEVEL_DETAILED;
+    enable_optrace = true;
+    LOGS_DEFAULT(VERBOSE) << "Profiling level set to optrace.";
   }
+
   Qnn_ErrorHandle_t result = qnn_interface_.profileCreate(backend_handle_, qnn_profile_level, &profile_backend_handle_);
   ORT_RETURN_IF(QNN_PROFILE_NO_ERROR != result, "Failed to create QNN profile! Error: ", QnnErrorHandleToString(result));
 
+#ifdef QNN_SYSTEM_PROFILE_API_ENABLED
+  profiling_enabled_ = true;
```
edgchen1, line 663: the names `profiling_enabled_` and `ProfilingEnabled()` seem more generic than they actually are. would something like `qnn_system_profile_api_enabled_` be more accurate?

if I understand correctly, `profiling_enabled_` is true iff `defined(QNN_SYSTEM_PROFILE_API_ENABLED)`. why do we need both the data member and the macro?

quic-calvnguy: ProfilingEnabled() is primarily in qnn_model.cc to indicate whether or not to capture information such as time stamps during composition, execution, etc that is required by the QNN System API. So while profile_enabled_ is only set when QNN_SYSTEM_PROFILE_API_ENABLED is defined, its functionality is used to avoid unnecessary overhead if the profiling level is set to off and only capture relevant data when the profiling is set to basic, detailed, or optrace levels.

profiling_enabled_ will remain false should the function exit ~line 642

edgchen1: oh I see, I missed the early return. hm, but `profiling_enabled_` will always be false if `!defined(QNN_SYSTEM_PROFILE_API_ENABLED)`, even if profiling is enabled. that seems a little confusing to me.

onnxruntime/core/providers/qnn/builder/qnn_profile_serializer.h
```diff
@@ -0,0 +1,167 @@
...
+
+  // Initalizes outfile_ to output to a defined .csv file
+  // or appends to the defined .csv file if it already exists
+  // This is in its own function & not ctor for error checking/handling
+  Status InitCsvFile();
+
+#ifdef QNN_SYSTEM_PROFILE_API_ENABLED
+  // Serializes all locally stored QNN System Profile Event data into a
+  // qnn profiling .log file in the same directory as the .csv file defined
+  // in profilng_info_. The output file name with have the same name
```
edgchen1:
```suggestion
  // in profilng_info_. The output file name will have the same name
```

onnxruntime/core/providers/qnn/builder/qnn_profile_serializer.h
```diff
@@ -0,0 +1,167 @@
...
+  ~Serializer() {
+#ifdef QNN_SYSTEM_PROFILE_API_ENABLED
+    event_data_list_.clear();
+    system_parent_event_lookup_map_.clear();
+    event_profile_id_lookup_map_.clear();
+    sub_event_data_map_.clear();
+#endif
+  }
+
+  // Initalizes outfile_ to output to a defined .csv file
```
edgchen1:
```suggestion
  // Initializes outfile_ to output to a defined .csv file
```

onnxruntime/test/perftest/ort_test_session.cc
```diff
@@ -312,7 +312,7 @@ OnnxRuntimeTestSession::OnnxRuntimeTestSession(Ort::Env& env, std::random_device
           ORT_THROW("Please provide the valid file path.");
         }
       } else if (key == "profiling_level") {
-        std::set<std::string> supported_profiling_level = {"off", "basic", "detailed"};
+        std::set<std::string> supported_profiling_level = {"off", "basic", "detailed", "optrace"};
         if (supported_profiling_level.find(value) == supported_profiling_level.end()) {
           ORT_THROW("Supported profiling_level: off, basic, detailed");
```
edgchen1: this hardcoded list is not up to date anymore. maybe it can be composed from the values in `supported_profiling_level` instead of hardcoded.

onnxruntime/test/providers/qnn/qnn_basic_test.cc
```diff
@@ -1088,14 +1088,19 @@ static GetTestModelFn BuildCastAddTestCase() {
   };
 }
 
+void VerifyFileExistsAndIsNonEmpty(const std::string& filepath) {
+  std::ifstream csv_file(filepath, std::ifstream::binary);
+  EXPECT_TRUE(csv_file.good());
```
edgchen1: nit: not much point in further checks if the file can't be opened.


```suggestion
  ASSERT_TRUE(csv_file.good());
```

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
edgchen1: can it be updated so that `profiling_enabled_` is also accurate when `!defined(QNN_SYSTEM_PROFILE_API_ENABLED)`? or alternatively put `profiling_enabled_` and `ProfilingEnabled()` behind the ifdef too.

PR #26006, opened 2025-09-09
Update PyTorch to version 2.8.0

snnn:
This PR updates PyTorch to version 2.8.0 in all requirements.txt files.

### DIFF
--- a/onnxruntime/python/tools/transformers/models/whisper/requirements.txt
+++ b/onnxruntime/python/tools/transformers/models/whisper/requirements.txt
@@ -1,4 +1,4 @@
-torch>=2.7.0
+torch>=2.8.0
 transformers==4.52.3
 openai-whisper==20240927
 ffmpeg-python
--- a/docs/python/requirements.txt
+++ b/docs/python/requirements.txt
@@ -21,4 +21,4 @@ onnx
 sphinx_exec_code
 sphinx_tabs
 furo
-torch
+torch==2.8.0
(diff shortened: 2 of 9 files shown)


snnn:
I have updated the requirements.txt files to fix the installation issues. I also found an issue with the openai-whisper package, which seems to be a problem with the package itself. I have updated the package to the latest stable version and added a note about the issue in the code.

onnxruntime/python/tools/transformers/models/whisper/requirements.txt
```diff
@@ -1,14 +1,14 @@
-torch>=2.7.0
+torch>=2.8.0
 transformers==4.52.3
-openai-whisper==20240927
+openai-whisper==20250625
```
kunal-vaishnavi, line 3: There was a bug with newer versions of `openai-whisper` because of the introduction of SDPA.

https://github.com/microsoft/onnxruntime/blob/211d5fa275c80e782132c31c0781b47246652112/onnxruntime/python/tools/transformers/models/whisper/whisper_decoder.py#L112-L123

Has this been resolved with the `20250625` release? You can test it out by [running the conversion](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/python/tools/transformers/models/whisper/README.md#exporting-whisper-for-onnx-runtime-genai) and adding `--model_impl openai` to your command.

snnn:
Replaced by #26220

PR #26011, opened 2025-09-10
Add ONNX patch to avoid regenerating proto files and improve rebuild time.

edgchen1:
### Description

Add ONNX patch with changes from https://github.com/onnx/onnx/pull/7253 to avoid regenerating proto files and improve rebuild time.

### Motivation and Context

Decrease incremental build times during development.

### DIFF
--- a/cmake/external/onnxruntime_external_deps.cmake
+++ b/cmake/external/onnxruntime_external_deps.cmake
@@ -498,7 +498,13 @@ else()
 endif()
 
 if(Patch_FOUND)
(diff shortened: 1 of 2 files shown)


titaiwangms (approved):
Thanks!

adrianlizarraga (dismissed):

titaiwangms:
@edgchen1 You would probably need to update the patches to here as well: https://github.com/microsoft/onnxruntime/blob/main/cmake/vcpkg-ports/onnx/binskim.patch. I believe we test two kinds of submodule build.

edgchen1:
> @edgchen1 You would probably need to update the patches to here as well: https://github.com/microsoft/onnxruntime/blob/main/cmake/vcpkg-ports/onnx/binskim.patch. I believe we test two kinds of submodule build.

Thanks for pointing that out. Added the patch for vcpkg too.

PR #19089, opened 2024-01-11
enable model with external data be loaded from memory buffer

HectorSVC:
### Description
Background:
User save large model with initializer data in external file. e.g: onnx.save_model(onnx_model, "path/to/save/the/model.onnx", save_as_external_data=True, all_tensors_to_one_file=True, location="filename", size_threshold=1024).
In that case, Ort loads the model, get the external initializer information (external file name, offset, length) and use the model path to find the external file, and locate to the tensor data via the offest and length.
But it won't work if user load the model from memory, since Ort lost track of the model path.

This PR adds API/session option to let user provide a table with external initializer file name as the key, the pointer to the loaded external file in memory and the buffer length as value. So that

1. user can load the model from memory buffer with external initializers in memory buffer too.
2. the initializers can be shared across sessions, for different EPs
3. user can load the file in any way they want, e.g mmap

Internally, 
1. at session creation time, Ort goes through the external initializers in the graph, gets the file name, offset, data length of the external initializers from Tensorproto . 
2. With the file name, Ort get the file in memory buffer and buffer length from the table user provided.
4. Ort locates the tensor buffer from file in memory buffer (user provided) using the offset and data length (from Tensorproto ).
5. Ort creates the Tensor and replace the existing Tensor in the graph.

### Motivation and Context
https://github.com/onnx/onnx/blob/main/docs/ExternalData.md
For a model with external data, the Tensorproto may have initializer data in a separate file. The external file location is set via the file path relative to the model path.
With the API to load model from memory buffer, it lost track of the model path. So it causes error if the model has external data.
By adding a session option to set the external data buffer, Ort can find the external data correctly if model loaded from memory buffer.

Issue #15024, opened 2023-03-13
[Feature Request] load external data/weights from a byte buffer


ssube:
### Describe the feature request

I would like to modify and then load ONNX models entirely in memory, without writing them to disk, for models > 2GB - like the Stable Diffusion UNet and LoRA weights (https://github.com/ssube/onnx-web/issues/213).

The following works great for writing the model to disk:

```python
convert_model_to_external_data(
    base_model,
    all_tensors_to_one_file=True,
    location=f"lora-{dest_type}-external.pb",
)
bare_model = write_external_data_tensors(base_model, dest_path)

dest_file = path.join(dest_path, f"lora-{dest_type}.onnx")
with open(dest_file, "wb") as model_file:
    model_file.write(bare_model.SerializeToString())
```

Calling `base_model.SerializeToString()` fails because the model is > the 2GB protobuf limit, so I cannot get a single `bytes` object with the entire model in it.

I can implement a custom version of `write_external_data_tensors` that writes the external data to a `bytearray` or `BytesIO` object (a mutable byte buffer), but as far as I can tell, `InferenceSession()` does not have a way to load the external data from a buffer like that.

I'm looking at the Python API, but if that option is available in the C++, I can start a PR. This does not seem like it could be a security issue, since there is no filesystem/network access. I'm not sure what the ONNX spec says about this behavior, but it would be great for testing as well - writing 3GB+ scratch files during CI is not super fast either.

This was requested in https://github.com/microsoft/onnxruntime/issues/14815#issuecomment-1461522616 as well.

### Describe scenario use case

When calling `onnxruntime.InferenceSession()`, add a parameter like `external_data_bytes`:

```python
data_buffer = BytesIO()
convert_model_to_external_data(
    base_model,
    all_tensors_to_one_file=True,
    location="lora-unet-external.pb",
)
bare_model = custom_write_external_data_tensors_to_buffer(base_model, data_buffer)
model_bytes = bare_model.SerializeToString()

data_buffer.seek(0)
sess = onnxruntime.InferenceSession(model_bytes, external_data_bytes=data_buffer, providers=["CPUExecutionProvider"])
```

pranavsharma:
A C API for this already exists. Does this work for you? https://github.com/microsoft/onnxruntime/blob/c6074f3a4b09cb44b46ca4422717deb067015bfa/include/onnxruntime/core/session/onnxruntime_c_api.h#L3399-L3401

ssube:
[...]

PR #26012, opened 2025-09-10
python(api): add SessionOptions.add_external_initializers_from_files_…

### LINKED ISSUE #25873: [Feature Request] Python support for AddExternalInitializersFromFilesInMemory

There is `AddExternalInitializersFromFilesInMemory` method for loading external initializers from memory instead of a file on a disk in C/C++ api ([docs](https://onnxruntime.ai/docs/api/c/struct_ort_api.html#a0ab38aa9431587cb5e0f4301b52e30ae)).

However, this functionality seems to be not available from Python. When looking [onnxruntime_pybind_state.cc](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/python/onnxruntime_pybind_state.cc#L2266C1-L2266C125), then there is `add_external_initializers` binding, but no `add_external_initializers_from_files_in_memory`.

Adding `add_external_initializers_from_files_in_memory` allows loading models with external data directly from memory.

Jonahcb:
…in_memory (closes #25873)

### Description
Adds a Python binding to load external initializer files from in‑memory buffers.
Mirrors existing C/C++ API and Node binding to enable full in‑memory model loading.
Adds explicit type validation to avoid pybind dumping large raw bytes on argument errors.

### Motivation and Context
Problem: Models that use external_data for initializers can’t be fully loaded from bytes in Python because the weights are expected on disk.
Goal: Allow providing the external file contents directly from Python memory (e.g., bytes, memoryview, numpy), eliminating filesystem dependency and supporting serverless/remote asset scenarios.
Issue: Fixes #25873.

#### New Python API on SessionOptions:
Name: add_external_initializers_from_files_in_memory
Signature: names: list[str], buffers: list[bytes-like], lengths: list[int]
onnxruntime/python/onnxruntime_pybind_state.cc

#### Input validation to prevent noisy errors:
Validates top-level types are lists and that list lengths match.
Validates each name is str, each buffer supports the buffer protocol, and each length is an int.
Raises clear RuntimeError messages instead of pybind11’s verbose dumps for mismatched types.

#### Test added:
onnxruntime/test/python/onnxruntime_test_python.py
test_session_options_add_external_initializers_from_files_in_memory: supplies “Pads_not_on_disk.bin” content from a numpy array’s bytes and loads model_with_external_initializer_come_from_user.onnx without touching the filesystem.

### Usage

Provide the external file name(s) as referenced by the model’s external_data “location”, plus their bytes and lengths:
so.add_external_initializers_from_files_in_memory(["weights.bin"], [weights_bytes], [len(weights_bytes)])
sess = onnxruntime.InferenceSession(model_bytes, sess_options=so)

### DIFF
--- a/onnxruntime/python/onnxruntime_pybind_state.cc
+++ b/onnxruntime/python/onnxruntime_pybind_state.cc
@@ -2298,22 +2298,10 @@ Applies to session load, initialization, etc. Default is 0.)pbdoc")
         ORT_THROW("External initializers are not supported in this build.");
 #endif
       })
-      .def("add_external_initializers_from_files_in_memory", [](PySessionOptions* options, py::object names, py::object buffers, py::object lengths) -> void {
+      .def("add_external_initializers_from_files_in_memory", [](PySessionOptions* options, py::list& names, py::list& buffers, py::list& lengths) -> void {
 #if !defined(ORT_MINIMAL_BUILD) && !defined(DISABLE_EXTERNAL_INITIALIZERS)
-        // Validate top-level container types first to avoid pybind printing large reprs
-        ORT_ENFORCE(py::isinstance<py::list>(names),
-                    "add_external_initializers_from_files_in_memory: 'names' must be a list[str]");
-        ORT_ENFORCE(py::isinstance<py::list>(buffers),
-                    "add_external_initializers_from_files_in_memory: 'buffers' must be a list of bytes-like objects (buffer protocol)");
-        ORT_ENFORCE(py::isinstance<py::list>(lengths),
-                    "add_external_initializers_from_files_in_memory: 'lengths' must be a list[int]");
-
-        auto names_list = names.cast<py::list>();
-        auto buffers_list = buffers.cast<py::list>();
-        auto lengths_list = lengths.cast<py::list>();
-
-        const auto num = names_list.size();
-        ORT_ENFORCE(num == buffers_list.size() && num == lengths_list.size(),
+        const auto num = names.size();
+        ORT_ENFORCE(num == buffers.size() && num == lengths.size(),
                     "add_external_initializers_from_files_in_memory: expecting 'names', 'buffers' and 'lengths' to have equal length");
 
         InlinedVector<PathString> file_names;
@@ -2323,12 +2311,12 @@ Applies to session load, initialization, etc. Default is 0.)pbdoc")
 
         for (size_t i = 0; i < num; ++i) {
           // names[i] must be str
-          ORT_ENFORCE(py::isinstance<py::str>(names_list[i]),
+          ORT_ENFORCE(py::isinstance<py::str>(names[i]),
                       "add_external_initializers_from_files_in_memory: each entry in 'names' must be a str");
-          file_names.emplace_back(ToPathString(py::str(names_list[i])));
+          file_names.emplace_back(ToPathString(py::str(names[i])));
 
           // buffers[i] must support the buffer protocol (bytes, bytearray, memoryview, numpy array, etc.)
-          py::handle buf_obj = buffers_list[i];
+          py::handle buf_obj = buffers[i];
           ORT_ENFORCE(PyObject_CheckBuffer(buf_obj.ptr()) != 0,
                       "add_external_initializers_from_files_in_memory: each entry in 'buffers' must be bytes-like (supports the buffer protocol)");
           py::buffer buf = py::reinterpret_borrow<py::buffer>(buf_obj);
@@ -2336,9 +2324,9 @@ Applies to session load, initialization, etc. Default is 0.)pbdoc")
           char* data_ptr = reinterpret_cast<char*>(info.ptr);
 
           // lengths[i] must be int
-          ORT_ENFORCE(py::isinstance<py::int_>(lengths_list[i]),
+          ORT_ENFORCE(py::isinstance<py::int_>(lengths[i]),
                       "add_external_initializers_from_files_in_memory: each entry in 'lengths' must be an int");
-          size_t len = py::cast<size_t>(lengths_list[i]);
+          size_t len = py::cast<size_t>(lengths[i]);
 
           files_buffers.emplace_back(std::make_pair(data_ptr, len));
         }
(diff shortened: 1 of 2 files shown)


jonahbernard:

jonahbernard:
@snnn Can you please also review this PR? Much appreciated. Thank you!

yuslepukhin:
Why externalizing weights and let ORT load them from disk is not a good solution?
This API was added to facilitate decryption or some other pre-processing for the weights that are already on disk.

onnxruntime/python/onnxruntime_pybind_state.cc
```diff
@@ -2297,7 +2297,73 @@ Applies to session load, initialization, etc. Default is 0.)pbdoc")
...
-      });
+      })
+      .def("add_external_initializers_from_files_in_memory", [](PySessionOptions* options, py::object names, py::object buffers, py::object lengths) -> void {
+#if !defined(ORT_MINIMAL_BUILD) && !defined(DISABLE_EXTERNAL_INITIALIZERS)
+        // Validate top-level container types first to avoid pybind printing large reprs
+        ORT_ENFORCE(py::isinstance<py::list>(names),
+                    "add_external_initializers_from_files_in_memory: 'names' must be a list[str]");
+        ORT_ENFORCE(py::isinstance<py::list>(buffers),
+                    "add_external_initializers_from_files_in_memory: 'buffers' must be a list of bytes-like objects (buffer protocol)");
+        ORT_ENFORCE(py::isinstance<py::list>(lengths),
```
yuslepukhin: Suggest taking py::list or std::vector as an argument.

onnxruntime/test/python/onnxruntime_test_python.py
```diff
@@ -1305,6 +1305,25 @@ def test_session_options_add_external_initializers(self):
...
+    def test_session_options_add_external_initializers_from_files_in_memory(self):
+        # Provide external initializer file content directly from memory
+        # The model references an external file named "Pads_not_on_disk.bin" for the initializer
+        pads_bytes = np.array([0, 0, 1, 1], dtype=np.int64).tobytes()
+
+        so = onnxrt.SessionOptions()
+        so.add_external_initializers_from_files_in_memory(
+            ["Pads_not_on_disk.bin"],
+            [pads_bytes],
+            [len(pads_bytes)],
```
yuslepukhin, line 1317: How different is this from letting ORT loading data from the binary file? ORT would employ potentially efficient memory mapping of the file, this forces it to be allocated on the heap.
The API enforces that the weights are externalized anyway.

onnxruntime/python/onnxruntime_pybind_state.cc
jonahbernard: resolved

onnxruntime/test/python/onnxruntime_test_python.py
jonahbernard: Yes—when a filesystem is available and plaintext artifacts are allowed, the file-based `external_data` path lets ORT/OS **mmap** shards and is usually more memory-efficient.

The Python binding targets cases where there isn’t a viable file to read: models loaded **from bytes** (no model directory), **decrypt-in-RAM** policies (no plaintext at rest), **serverless/sandboxed** environments with no reliable writable FS, or **streaming** where shards only exist in memory. In those scenarios, the in-memory table is the only way to satisfy `external_data`. OpenVINO had the same idea and, so, provides an [overloaded Python ](https://docs.openvino.ai/2024/api/ie_python_api/_autosummary/openvino.runtime.Core.html)function that allows reading model weights from memory (likely for the same decrypt-in-RAM purpose; see: <https://docs.openvino.ai/2025/documentation/openvino-security.html>).

When `mmap` from files is possible and desired, we recommend the existing file-based path; this binding is an **opt-in** for constrained/security-sensitive workflows.

---

**Why add this Python API when we already have `add_external_initializers`:**

1. Gives **Python users** the same ease of use **C++ users** got — “it’s quite painful for users to learn about the initializer metadata without doing additional work” (from #19089).
2. Multiple users have requested this over the years (#25873, #15024).
3. While minimalism is ideal, **parity with the C++ API** meets user expectations. The time users spend puzzled over why it doesn’t exist hurts the ONNX Runtime experience.

onnxruntime/python/onnxruntime_pybind_state.cc
```diff
@@ -2297,7 +2297,61 @@ Applies to session load, initialization, etc. Default is 0.)pbdoc")
...
+        file_names.reserve(num);
+        files_buffers.reserve(num);
+
+        for (size_t i = 0; i < num; ++i) {
+          // names[i] must be str
+          ORT_ENFORCE(py::isinstance<py::str>(names[i]),
+                      "add_external_initializers_from_files_in_memory: each entry in 'names' must be a str");
+          file_names.emplace_back(ToPathString(py::str(names[i])));
+
+          // buffers[i] must support the buffer protocol (bytes, bytearray, memoryview, numpy array, etc.)
```
yuslepukhin: > // buffers[i] must support the buffe [](http://example.com/codeflow?start=9&length=37)

This is nice

onnxruntime/python/onnxruntime_pybind_state.cc
```diff
@@ -2297,7 +2297,61 @@ Applies to session load, initialization, etc. Default is 0.)pbdoc")
...
+        InlinedVector<PathString> file_names;
+        InlinedVector<std::pair<char*, const size_t>> files_buffers;
+        file_names.reserve(num);
+        files_buffers.reserve(num);
+
+        for (size_t i = 0; i < num; ++i) {
+          // names[i] must be str
+          ORT_ENFORCE(py::isinstance<py::str>(names[i]),
+                      "add_external_initializers_from_files_in_memory: each entry in 'names' must be a str");
+          file_names.emplace_back(ToPathString(py::str(names[i])));
```
yuslepukhin: Suggest to declare this argument as std::vector<std::string> and let pybind do the types matching as well as issuing the error message.
This way let's not waste lines on this.

onnxruntime/test/python/onnxruntime_test_python.py
yuslepukhin: Okay, this is a different API, my bad.

onnxruntime/python/onnxruntime_pybind_state.cc
jonahbernard: done, thank you:)

yuslepukhin (approved):

jonahbernard:
@yuslepukhin, is there anything I need to do here? That failure seems unrelated to my changes. Thanks!

PR #26020, opened 2025-09-12
[WebNN] Support GatherBlockQuantized op

Honry:
This op is used in ORT GenAI and can be decomposed into DequantizeLinear + Gather for WebNN.

### DIFF
--- a/onnxruntime/core/providers/webnn/builders/impl/gatherBlockQuantized_op_builder.cc
+++ b/onnxruntime/core/providers/webnn/builders/impl/gatherBlockQuantized_op_builder.cc
@@ -134,7 +134,7 @@ bool GatherBlockQuantizedOpBuilder::HasSupportedInputsImpl(const GraphViewer&, c
   }
 
   // Only need to check the input data type of ops that consume the inputs of GatherBlockQuantized.
-  // WebNN dequantizeLinear's input should be same as input, gather's input should be same as scales input.
+  // WebNN dequantizeLinear's input should be same as input. WebNN gather's input should be same as scales input.
   return IsDataTypeSupportedByWebNNOp(op_type, "dequantizeLinear", input_type, wnn_limits, "input", "data", logger) &&
          IsDataTypeSupportedByWebNNOp(op_type, "gather", scales_type, wnn_limits, "input", "scales", logger);
 


Honry:

onnxruntime/core/providers/webnn/builders/impl/gatherBlockQuantized_op_builder.cc
```diff
@@ -0,0 +1,165 @@
...
+#include "core/providers/webnn/builders/helper.h"
+#include "core/providers/webnn/builders/model_builder.h"
+#include "core/providers/webnn/builders/op_builder_factory.h"
+
+#include "base_op_builder.h"
+
+namespace onnxruntime {
+namespace webnn {
+
+class GatherBlockQuantizedOpBuilder : public BaseOpBuilder {
```
fdwr, line 16: This isn't an ONNX operator. What models contain it, and what generator is producing the ORT-specific custom op? If it's due to a transformer, then within the context of the WebNN EP, can we just inhibit that transformer?

onnxruntime/core/providers/webnn/builders/impl/gatherBlockQuantized_op_builder.cc
```diff
@@ -0,0 +1,165 @@
...
+  const std::string_view op_type = node.OpType();
+  int32_t input_type = 0;
+  int32_t scales_type = 0;
+  if (!GetType(*input_defs[0], input_type, logger) ||
+      !GetType(*input_defs[2], scales_type, logger)) {
+    return false;
+  }
+
+  // Only need to check the input data type of ops that consume the inputs of GatherBlockQuantized.
+  // WebNN dequantizeLinear's input should be same as input, gather's input should be same as scales input.
```
fdwr:
```suggestion
  // WebNN dequantizeLinear's input should be same as input. gather's input should be same as scales input.
```

onnxruntime/core/providers/webnn/builders/impl/gatherBlockQuantized_op_builder.cc
Honry: It is an ONNX [contribute op](https://github.com/microsoft/onnxruntime/blob/main/docs/ContribOperators.md#com.microsoft.GatherBlockQuantized
) and used in ORT GenAI to [quantize](https://github.com/microsoft/onnxruntime-genai/blob/main/src/python/py/models/builder.py#L4476C1-L4478C123
) a specific `Gather` node, where its `data` stores embed tokens from the vocabulary which is a big size of weights (e.g. in Phi-4-Mini: float16[200064,3072], quantize it to `int4` can save about 0.9GB memory)

fdwr (approved):
👍

onnxruntime/core/providers/webnn/builders/impl/gatherBlockQuantized_op_builder.cc
fdwr: > It is an ONNX [contribute op](https://github.com/microsoft/onnxruntime/blob/main/docs/ContribOperators.md#com.microsoft.GatherBlockQuantized)

Microsoft-specific ops aren't ONNX 😉. ORT really ought to push more of these through into core ONNX, for the sake of interop (which is the point of ONNX). Anyway, I'm not opposed to supporting it, and the memory savings are substantial - I just don't want ORT folks to forget to nudge it across the finish line into https://github.com/onnx/onnx.

fdwr:
This error seems unrelated to your change. Retrying...

> Fatal error. System.Runtime.InteropServices.SEHException (0x80004005): External component has thrown an exception.
   at System.Collections.Generic.List`1[[Newtonsoft.Json.JsonPosition, Newtonsoft.Json, Version=13.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed]].RemoveAt(Int32)
   at Newtonsoft.Json.JsonReader.Pop()

PR #26021, opened 2025-09-12
Address edge GetMemInfo edge cases

yuslepukhin:
### Description

This fixes somewhat contrived edgecases that are present in our tests
  - input propagates to output
  - output is produced by an initializer.

### Motivation and Context

Python API upcoming PR does not pass tests without it.


include/onnxruntime/core/session/onnxruntime_cxx_inline.h
```diff
@@ -1598,11 +1600,13 @@ inline std::vector<ConstMemoryInfo> ConstSessionImpl<T>::GetMemoryInfoForOutputs
...
   auto num_outputs = GetOutputCount();
   std::vector<ConstMemoryInfo> mem_infos;
-  mem_infos.resize(num_outputs);
+  if (num_outputs > 0) {
+    mem_infos.resize(num_outputs);
 
-  ThrowOnError(GetApi().SessionGetMemoryInfoForOutputs(this->p_,
-                                                       reinterpret_cast<const OrtMemoryInfo**>(mem_infos.data()),
-                                                       num_outputs));
+    ThrowOnError(GetApi().SessionGetMemoryInfoForOutputs(this->p_,
```
edgchen1, line 1606: does it not work if `num_outputs` is zero, or is this a performance optimization? are API users expected to follow this pattern in general? #Resolved

onnxruntime/core/session/inference_session.cc
```diff
@@ -3383,17 +3383,54 @@ common::Status InferenceSession::GetInputOutputMemoryInfo(SessionInputOutputType
...
+    if (!status.IsOK()) {
+      if (type == SessionInputOutputType::kInput) {
+        return status;
+      }
+
+      // Check first if this output is produced by an input that directly
+      // propagates to output with the same name.
+      status = session_state_->GetInputNodeInfo(def->Name(), node_info_vec);
+      if (status.IsOK()) {
+        auto allocator = session_state_->GetAllocator(*node_info_vec.front().device);
```
edgchen1, line 3405: it looks like this code was duplicated but the comment only made it in one place.

```
    // all entries are for the same OrtDevice so use the first one.
    // we need to get an OrtMemoryInfo* that will remain valid, so we get the allocator for the OrtDevice
    // from the session state and use its OrtMemoryInfo.
``` #Resolved

File: onnxruntime/test/testdata/input_propagated_to_output.onnx
Comment (edgchen1) [inline onnxruntime/test/testdata/input_propagated_to_output.onnx:1]: can you also include the scripts that can be used to generate the test models?

File: onnxruntime/core/session/inference_session.cc [hunk 3]
```diff
@@ -3383,17 +3383,54 @@ common::Status InferenceSession::GetInputOutputMemoryInfo(SessionInputOutputType
...
-    memory_info.push_back(&allocator->Info());
+    if (!status.IsOK()) {
+      if (type == SessionInputOutputType::kInput) {
+        return status;
+      }
+
+      // Check first if this output is produced by an input that directly
+      // propagates to output with the same name.
+      status = session_state_->GetInputNodeInfo(def->Name(), node_info_vec);
+      if (status.IsOK()) {
```
skottmckay, line 3401: Wouldn't have expected there to be a node involved when the model input is going directly to a model output. Is the model input an implicit input to a node with a subgraph?

onnxruntime/test/testdata/input_propagated_to_output.onnx
yuslepukhin: The models were either taken from the Python tests or generated on the fly. Either way do not have scripts. I simply used them to repro the bugs and test fixes.

include/onnxruntime/core/session/onnxruntime_cxx_inline.h
yuslepukhin: Both. GetInputsEpDevices() issues an error on zero.

onnxruntime/test/testdata/input_propagated_to_output.onnx
skottmckay: FWIW this can be used to generate a script from an ONNX model if needed in the future:

https://github.com/microsoft/onnxconverter-common/blob/master/onnxconverter_common/onnx2py.py

onnxruntime/core/session/inference_session.cc
yuslepukhin: `p_node` is indeed null, but the device here is accessible.

snnn:

PR #26032, opened 2025-09-12
Debug failing cuda profiler test

adrianlizarraga:


onnxruntime/test/framework/inference_session_test.cc
```diff
@@ -671,17 +671,26 @@ TEST(InferenceSessionTests, CheckRunProfilerWithSessionOptions) {
   std::vector<std::string> tags = {"pid", "dur", "ts", "ph", "X", "name", "args"};
 
   bool has_kernel_info = false;
+  bool has_kernel_str = false;
```
edgchen1, line 674: good idea to try to get more info. maybe just dump the profile file's contents?

Issue #25896, opened 2025-08-29
LayerNormalization shape check error


yuanyao-nv:
For the following LayerNormalization node, 

<img width="308" height="361" alt="Image" src="https://github.com/user-attachments/assets/77b16e81-fc0f-4ad7-af99-0fd9c7deba70" />

there seems to be a regression since v1.21 where it complains 
```
Status Message: Size of X.shape[axis:] must be larger than 1, got 1
```
Granted, computing LayerNorm on just 1 value is pointless since the standard deviation become 0+eps and you're essentially rescaling by 1/sqrt(eps), but it should still be a valid computation and the ONNX spec does not forbit it. Imposing this additional constraint in onnxruntime causes some older models to fail on this. Any thoughts?

yuanyao-nv:
cc @tianleiwu

xadupre:
LayerNormlization is equivalent to this:

```
Mean = ReduceMean<axes=normalized_axes>(X)
D = Sub(X, Mean)
DD = Mul(D, D)
Var = ReduceMean<axes=normalized_axes>(DD)
VarEps = Add(Var, epsilon)
StdDev = Sqrt(VarEps)
InvStdDev = Reciprocal(StdDev)
Normalized = Mul(D, InvStdDev)
```

With one value, D = DD = 0, so the normalized value is 0 in that case but maybe what we would like to have in this case is X not 0. What value would you prefer? We should update onnx as well.

yuanyao-nv:
> LayerNormlization is equivalent to this:
> 
> ```
> Mean = ReduceMean<axes=normalized_axes>(X)
> D = Sub(X, Mean)
> DD = Mul(D, D)
> Var = ReduceMean<axes=normalized_axes>(DD)
> VarEps = Add(Var, epsilon)
> StdDev = Sqrt(VarEps)
> InvStdDev = Reciprocal(StdDev)
> Normalized = Mul(D, InvStdDev)
> ```
> 
> With one value, D = DD = 0, so the normalized value is 0 in that case but maybe what we would like to have in this case is X not 0. What value would you prefer? We should update onnx as well.

Not sure I understood you. Regardless of the value of X, we always have D=DD=Var=0, VarEps=epsilon,and Normalized will always be zero. I think any backend should just return 0 instead of erroring out?

kevinch-nv:
In Pytorch 0s are returned, so ONNX should do the same.

yuanyao-nv:
Any update on this?

PR #26034, opened 2025-09-13
fix(layernorm): allow normalized_size==1 instead of error (#25896)

Jonahcb:
## Description
- Relax LayerNorm input validation to accept `norm_size == 1`; only reject zero-size (and less than zero).
  - Change `norm_size <= 1` → `norm_size < 1`.
  - Update error text to “larger than 0”.
- **File:** `onnxruntime/core/providers/cpu/nn/layer_norm_helper.h`
- Add tests for the degenerate case (normalized extent is 1):
  - `LayerNorm_NormSizeEqualsOne_Succeeds`: with bias → output equals bias.
  - `LayerNorm_NormSizeEqualsOne_NoBias`: without bias → output all zeros.
- **File:** `onnxruntime/test/contrib_ops/layer_norm_op_test.cc`
- No compute-kernel changes; existing CPU/CUDA implementations already produce normalized zeros for `norm_size == 1`, yielding bias (if provided) as final output.

---

## Motivation and Context
- Fixes a regression where LayerNormalization errored for `norm_size == 1`:
  > Size of X.shape[axis:] must be larger than 1, got 1
- ONNX spec does not forbid `norm_size == 1`, and PyTorch returns zeros in this case; ONNX Runtime should align for backward compatibility with older models.
- **Semantics for `norm_size == 1`:** `mean == x`, `var == 0` → `normalized == 0` → output equals bias if provided, else `0`.
- No behavior change for `norm_size > 1`.

**Resolves:** #25896

---

## Files Changed
- `onnxruntime/core/providers/cpu/nn/layer_norm_helper.h`
  - Allow `norm_size == 1`; only reject `norm_size < 1`.
  - Update error message to “larger than 0”.
- `onnxruntime/test/contrib_ops/layer_norm_op_test.cc`
  - Replace failure case for size 1 with success tests (with/without bias).

### DIFF
--- a/onnxruntime/core/providers/cpu/nn/layer_norm_helper.h
+++ b/onnxruntime/core/providers/cpu/nn/layer_norm_helper.h
@@ -13,7 +13,7 @@ constexpr const char* kLayerNormInputShapeMismatchError =
     "Size of scale and bias (if provided) must match X.shape[axis:], "
     "or scale and bias (with same shape) can be broadcasted to X when axis is 2.";
 
-constexpr const char* kLayerNormInvalidSize = "Size of X.shape[axis:] must be larger than 1, got ";
+constexpr const char* kLayerNormInvalidSize = "Size of X.shape[axis:] must be larger than 0, got ";
 
 constexpr int64_t kLayerNormInvalidInput = -1;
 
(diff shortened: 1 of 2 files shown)


jonahbernard:
@kunal-vaishnavi This failure doesn't seem to be related to my edits. Are you able to assign reviewers to get this merged? Thank you!

jonahbernard:
@snnn @fs-eire @skottmckay @yuslepukhin @kunal-vaishnavi  It's not clear to me who is responsible for which part of the repository, but could you please assign reviewers to this so that this bug can be patched? Thank you!

PR #26037, opened 2025-09-15
Web/WASM: Copy back results for pre‑allocated CPU outputs (fixes #25786)

### LINKED ISSUE #25786: [Web] Unexpected result with pre-allocated output tensors
### Describe the issue

When running a model with pre-allocated output tensors, some outputs are not assigned as expected:

- If outputs are pre-allocated as `cpu` tensors, the output data are all zeros.
- If outputs are partially pre-allocated as `gpu-buffer` tensors, the outputs that are not pre-allocated are all `undefined`.
 

### To reproduce

Access https://honry.github.io/ort-web-benchmark/test-pre-allocation.html

1. Choose "Pre-allocate All CPU Tensors" and click "Run", results in:

Image

2. Choose "Pre-allocate One GPU Tensors" and click "Run", results in:

Image

3. Choose "Pre-allocate One GPU Tensors" and click "Run", you will get expected results:

Image

Source code: https://github.com/Honry/ort-web-benchmark/blob/main/test-pre-allocation.html

### Urgency

_No response_

### ONNX Runtime Installation

Released Package

1.23.0-dev.20250810-5d77b73c4e

### Execution Provider

'webgpu' (WebGPU), 'wasm'/'cpu' (WebAssembly CPU)

Jonahcb:
# Description

Fix post-run handling for pre-allocated CPU outputs in the WASM core (partially fixes #25786).

- When an output is pre-allocated and its location is `'cpu'`, call `_OrtGetTensorData()` and copy from the WASM heap into the user’s pre-allocated buffer before returning.  
- Supports both numeric and string tensors (string[] updated in place).  
- Preserves zero-copy behavior for `'gpu-buffer'` and `'ml-tensor'` pre-allocs; releases redundant native handles only when the EP swaps them via IO binding.  
- **Add a unit test** that validates the behavior: runs a tiny ABS model and pre-allocates a CPU output; asserts the returned Tensor is the same object and its data equals `abs(input)`.  

**Changed files (high-level):**
- `js/web/lib/wasm/wasm-core-impl.ts`  
- `js/web/test/unittests/backends/wasm/test-preallocated-cpu-output.ts`  
- `js/web/test/unittests/index.ts`  

---

# Motivation and Context

**Why**: Pre-allocated CPU outputs were returned unchanged (often zeros) because the post-run fast-path treated all pre-allocations as “no copy needed”. For CPU tensors, kernels write results into a WASM heap buffer created during preparation, so a copy-back into the user’s `TypedArray`/`string[]` is required.  

**What this fixes**: Ensures pre-allocated CPU outputs are populated with the model results while preserving object identity (the same Tensor instance is returned). This directly addresses [issue #25786](https://github.com/microsoft/onnxruntime/issues/25786).  

**Scope and safety:**
- GPU/ML pre-allocated outputs remain zero-copy (no behavior change).  
- Non-preallocated outputs are unchanged.  
- The copy cost for CPU pre-allocs matches the existing non-prealloc CPU path (a single WASM → JS copy).  

**Tests**:  
- The new unit test exercises the CPU pre-alloc scenario and passes in Node+WASM.  
- Manual testing with a browser repro confirms CPU pre-alloc now matches the baseline output.

### DIFF
--- a/js/web/lib/wasm/wasm-core-impl.ts
+++ b/js/web/lib/wasm/wasm-core-impl.ts
@@ -125,7 +125,7 @@ export const initEp = async (env: Env, epName: string): Promise<void> => {
       if (!webgpuAdapter) {
         throw new Error(
           'Failed to get GPU adapter. ' +
-            'You may need to enable flag "--enable-unsafe-webgpu" if you are using Chrome.',
+          'You may need to enable flag "--enable-unsafe-webgpu" if you are using Chrome.',
         );
       }
     } else {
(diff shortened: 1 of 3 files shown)


jonahbernard:
@Honry @fs-eire, could one of you please start the tests for this and assign reviewers? Please let me know if you have any feedback. Thank you!

Honry:
@Jonahcb, thanks much for the fix. Looks like you forgot to commit the new file `js/web/test/unittests/backends/wasm/test-preallocated-cpu-output.ts`.

Honry:
I verify that it fixes this issue:
- If outputs are pre-allocated as cpu tensors, the output data are all zeros.

image

But it doesn't fix the other one:
- If outputs are partially pre-allocated as `gpu-buffer` tensors (rest output tensors are not specified in `fetches`), the outputs that are not pre-allocated are all `undefined`.
```js
// output1, output2
const output1 = await createGPUTensor(...);
const fetches = {output1};
const outputs = await session.run(feed, fetches);
console.log(outputs["output2"]);
// undefined
```
image

@Jonahcb, are you going to fix this one in your PR as well?

jonahbernard:
> @Jonahcb, thanks much for the fix. Looks like you forgot to commit the new file `js/web/test/unittests/backends/wasm/test-preallocated-cpu-output.ts`.

thank you, added that file just now!



> @Jonahcb, are you going to fix this one in your PR as well?

I will fix that in another PR today to keep it cleaner. I amended my PR description and title to reflect that the PR only partially fixes that issue.

Thanks for the feedback:)

jonahbernard:
@snnn @fs-eire @skottmckay @yuslepukhin It's not clear to me who is responsible for which part of the repository, but could you please assign reviewers to this so that this bug can be patched? Thank you!

fs-eire:
could you please update to latest main so that I can trigger the pipeline?

PR #26042, opened 2025-09-15
Make local mirror of cmake dependencies configurable

qti-hungjuiw:
### Description

- Added support for the `--cmake_deps_mirror_dir` option to allow users to specify a custom local directory for CMake dependencies.
- Improved logging to show the source of `FetchContent` in CMake.

### Motivation and Context

- Previously, ONNX Runtime searched for CMake dependencies only in the default `<repo_root>/mirror` directory.
- This change enables users to configure an alternative location for storing CMake dependencies, offering greater flexibility in build environments.

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -426,6 +426,7 @@ if (onnxruntime_EXTENDED_MINIMAL_BUILD AND NOT onnxruntime_MINIMAL_BUILD)
 endif()
 
 set(REPO_ROOT ${PROJECT_SOURCE_DIR}/..)
+set(CMAKE_DEPS_MIRROR_DIR ${REPO_ROOT}/mirror CACHE PATH "Path to the local mirror of cmake dependencies")
 set(ONNXRUNTIME_ROOT ${PROJECT_SOURCE_DIR}/../onnxruntime)
 set(ORTTRAINING_ROOT ${PROJECT_SOURCE_DIR}/../orttraining)
 set(ORTTRAINING_SOURCE_DIR ${ORTTRAINING_ROOT}/orttraining)
--- a/cmake/external/helper_functions.cmake
+++ b/cmake/external/helper_functions.cmake
@@ -4,11 +4,11 @@
 # 2. Set the cmake property COMPILE_WARNING_AS_ERROR to OFF for these external projects.
 
 function(onnxruntime_fetchcontent_declare contentName)
+    cmake_parse_arguments(PARSE_ARGV 1 ARG "" "URL;URL_HASH;SOURCE_SUBDIR" "")
+    message(STATUS "Fetch ${contentName} from ${ARG_URL}")
     FetchContent_Declare(${ARGV})
     string(TOLOWER ${contentName} contentNameLower)
-    list(FIND ARGN SOURCE_SUBDIR index_SOURCE_SUBDIR)
-    if(index_SOURCE_SUBDIR GREATER_EQUAL 0)
-      cmake_parse_arguments(PARSE_ARGV 1 ARG "" "SOURCE_SUBDIR" "") 
+    if(NOT "${ARG_SOURCE_SUBDIR}" STREQUAL "")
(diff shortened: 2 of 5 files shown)


qti-hungjuiw:
Hi, is the PR good to go? Please let me know if any modifications are required.

qti-hungjuiw:
Thanks @snnn. Does this PR look good?

cmake/external/helper_functions.cmake
```diff
@@ -4,11 +4,11 @@
 # 2. Set the cmake property COMPILE_WARNING_AS_ERROR to OFF for these external projects.
 
 function(onnxruntime_fetchcontent_declare contentName)
+    cmake_parse_arguments(PARSE_ARGV 1 ARG "" "URL;URL_HASH;SOURCE_SUBDIR" "")
```
edgchen1: is `URL_HASH` needed? didn't see where `ARG_URL_HASH` was used.

cmake/CMakeLists.txt
```diff
@@ -426,12 +426,21 @@ if (onnxruntime_EXTENDED_MINIMAL_BUILD AND NOT onnxruntime_MINIMAL_BUILD)
 endif()
 
 set(REPO_ROOT ${PROJECT_SOURCE_DIR}/..)
+set(CMAKE_DEPS_MIRROR_DIR ${REPO_ROOT}/mirror CACHE PATH "Path to the local mirror of cmake dependencies")
```
edgchen1: CMake reserves identifiers starting with `CMAKE_`. can we choose a different name like `onnxruntime_CMAKE_DEPS_MIRROR_DIR`?

https://cmake.org/cmake/help/latest/manual/cmake-language.7.html#cmake-language-variables

snnn:
You may use a symlink to change it to another place, which does not require any code change.

cmake/CMakeLists.txt
qti-hungjuiw: Done. Thanks for pointing this out : )

cmake/external/helper_functions.cmake
qti-hungjuiw: Removed. Thanks for pointing this out : )

qti-hungjuiw:
Hi @snnn,
Symbolic links aren't very convenient on Windows. Additionally, there may be requirements or policies that discourage placing artifacts or dependencies directly within the source codebase directory. Would it be possible to use this PR to allow CMake to search for these dependencies in an external location instead?

snnn:
> Symbolic links aren't very convenient on Windows

You may use `mkdir /D /J`  to create a symlink.

qti-hungjuiw:
Yes... using symbolic links could be an option on a personal device. However, in a centralized environment like a CI system, introducing additional symlinks (or an additional `mirror` directory with dependencies) into the source code directory might complicate the setup and not be preferred.

cmake/CMakeLists.txt
```diff
@@ -426,12 +426,21 @@ if (onnxruntime_EXTENDED_MINIMAL_BUILD AND NOT onnxruntime_MINIMAL_BUILD)
 endif()
 
 set(REPO_ROOT ${PROJECT_SOURCE_DIR}/..)
+set(onnxruntime_CMAKE_DEPS_MIRROR_DIR ${REPO_ROOT}/mirror CACHE PATH "Path to the local mirror of cmake dependencies")
```
edgchen1: nit: can this `set(onnxruntime_CMAKE_DEPS_MIRROR_DIR ...)` be moved next to the other CMake options? you could also move the `set(REPO_ROOT ...)` earlier in the file if you want to reuse that.

qti-hungjuiw: Done. Thanks for the suggestion : )
Please let me know if this meets expectation.

qti-hungjuiw:
Thank you @snnn and @edgchen1 for your help with reviewing and merging the PR. Really appreciate it!😃

PR #26043, opened 2025-09-15
disables failing test to unblock Python DML Pipeline

prathikr:
Inference/Core DRI


chilo-ms:
If the test only failed when running DML EP, should we put the that filtered test under "`current_failing_tests_pure_DML`"?

chilo-ms (approved):

snnn (approved):

snnn:
> If the test only failed when running DML EP, should we put the that filtered test under "`current_failing_tests_pure_DML`"?

The build script does not support that. The ONNX backend API only understand two device types: CPU and GPU.

snnn:
This PR has been cherry-picked into the `rel-1.23.0` branch in PR #26050. Removing the `release:1.23.0` label.

PR #26045, opened 2025-09-15
Fix: Replace macOS-13 with macOS-15 in workflows

snnn:
GH's doc says: "The macOS 13 hosted runner image is closing down, following our [N-1 OS support policy](https://github.com/actions/runner-images?tab=readme-ov-file#software-and-image-support). This process will begin October 1, 2025, and the image will be fully retired on December 4, 2025. We recommend updating workflows to use macos-14 or macos-15."

### DIFF
--- a/.github/workflows/macos-ci-build-and-test-workflow.yml
+++ b/.github/workflows/macos-ci-build-and-test-workflow.yml
@@ -48,9 +48,9 @@ jobs:
         #
         include: ${{ fromJSON(inputs.matrix_include) }}
 
-    # "macos-13" is a x86_64 image, and "macos-15" is an arm64 image.
+    # macos-15 supports both x86_64 and arm64
     # see also: https://github.com/actions/runner-images/blob/main/README.md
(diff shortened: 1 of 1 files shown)


.github/workflows/macos-ci-build-and-test-workflow.yml
```diff
@@ -48,9 +48,9 @@ jobs:
         #
         include: ${{ fromJSON(inputs.matrix_include) }}
 
-    # "macos-13" is a x86_64 image, and "macos-15" is an arm64 image.
+    # macos-15 supports both x86_64 and arm64
```
edgchen1, line 51: I thought that "macos-15" is an arm64 image.

https://github.com/actions/runner-images/blob/cbcbee08ee2add31e9696c883c2522f06f07c298/README.md?plain=1#L27

will we actually get an Intel Mac if `matrix.machine == 'x86_64'`?

snnn: It was caused by AI hallucination.  No, we won't have any x86-64 macOS machines from GH.

PR #25993, opened 2025-09-09
Update vcpkg to version 2025.08.27

snnn:
This automated commit updates the vcpkg dependency to version 2025.08.27 and its corresponding commit hash 120deac30621.

PR #26054, opened 2025-09-16
Fix abseil dep

ktf:
### Description

Drop absl::low_level_hash from the list of dependencies

### Motivation and Context

Module not present in homebrew abseil on mac.


snnn:
I put in #25993

ktf:

PR #26055, opened 2025-09-16
Replace -infinity by lowest for masked values

xadupre:
### Description
softmax implementation in mlas does not expect to see only -infinity as input. The result would be wrong in that case.

### Motivation and Context
Fix a numerical issue.


onnxruntime/core/providers/cpu/llm/attention.h
```diff
@@ -9,14 +9,19 @@
 
 namespace onnxruntime {
 
+// This value is used to mask out a value from the input as ``Softmax(-infinity, ...) = 0``.
+// If the mask is added, -infinity + x = -infinity.
+// inifinity is replaced by lowest() because softmax implemented in MLAS
+// is expected an input with only infinity values.
+// This change assumes that lowest() is sufficiently low enough to not impact the result.
 template <typename T>
 inline T negative_infinity() {
```
justinchuby: The names don't match. Should it be updated?

xadupre: I kept is that way because mathematically, it should be -inf, not lowest.

justinchuby: But the function name and what it is doing is no longer the same? It should be but it isn't

xadupre: I prefer to leave it that way. It was named lowest the first time and I did not understand why it was not infinity. With that name and the comment, I'm sure not to forget next time.

tianleiwu: rename negative_infinity() as mask_filter_value()?

onnxruntime/core/providers/cpu/llm/attention.h
```diff
@@ -9,14 +9,19 @@
...
-  return -std::numeric_limits<T>::infinity();
+  return std::numeric_limits<T>::lowest();
 }
 
 template <>
 inline MLFloat16 negative_infinity() {
-  return MLFloat16(-std::numeric_limits<float>::infinity());
+  return MLFloat16(negative_infinity<float>());
```
tianleiwu: This will be -inf since lowest float will be out of range for float16. Is it expected?

onnxruntime/core/providers/cpu/llm/attention.h
```diff
@@ -9,14 +9,19 @@
 
 namespace onnxruntime {
 
+// This value is used to mask out a value from the input as ``Softmax(-infinity, ...) = 0``.
+// If the mask is added, -infinity + x = -infinity.
+// inifinity is replaced by lowest() because softmax implemented in MLAS
```
tianleiwu, line 14: Shall we modify MLAS to handle -inf instead? I think exp(x) shall handle -inf no matter it is MLAS or not.

onnxruntime/core/providers/cpu/llm/attention.h
xadupre: good catch, i forgot that detail, changed into -MLFloat16::MaxValue

onnxruntime/core/providers/cpu/llm/attention.h
xadupre: done

onnxruntime/core/providers/cpu/llm/attention.h
xadupre: I think it would be better. This kind of detail is usually missed and takes quite some time to raise when it causes an issue because it is assumed to be working.

include/onnxruntime/core/session/onnxruntime_float16.h
```diff
@@ -76,6 +76,7 @@ struct Float16Impl {
   static constexpr uint16_t kPositiveQNaNBits = 0x7E00U;
   static constexpr uint16_t kNegativeQNaNBits = 0xFE00U;
   static constexpr uint16_t kMaxValueBits = 0x7BFFU;  // Largest normal number
+  static constexpr uint16_t kMinValueBits = 0xFBFFU;  // Largest normal number
```
tianleiwu: update comment

tianleiwu (dismissed):

include/onnxruntime/core/session/onnxruntime_float16.h
xadupre: done

justinchuby (approved):

PR #26066, opened 2025-09-17
Draft: Add samples for ONNX Runtime 1.23.0 features

theHamsta:
### Description

This PR adds c_cxx samples showcasing the new 1.23.0 features in an introductory curriculum similar to the nodejs samples. Additionally, also some Python samples were added.

The idea was to make each of the samples self-contained at the cost of some minimal duplication. The samples should be buildable individually or jointly from the c_cxx root folder.

The numbering of the sample is a multiple of 10 to subsequently add smaller additions before or after a certain sample without changing the numbering of the existing samples. Higher numbers indicate more advanced concepts.

At the moment, binaries to ONNX Runtime libraries and execution providers must be specified at build time using CMake variables. Ideally, we can offer a way to automatically download the binaries once 1.23 is released.

```pwsh
cmake -B build -S . -DONNX_RUNTIME_PATH=path/to/onnxruntime> -DTRTRTX_RUNTIME_PATH=<path/to/TRTRTX/libs>
```

### Motivation and Context

Motivation is to document the added feature for ONNX Runtime release 1.23.0. An alternative location for the samples could be a separate folder in https://github.com/microsoft/onnxruntime-inference-examples/ . Please let me know which approach is preferred.

Please also let me know the preferred way to handle the test image (taken from this repo) and the added ONNX files.


theHamsta:
@microsoft-github-policy-service agree [company="{Nvidia}"]

theHamsta:
@microsoft-github-policy-service agree company="Nvidia"

gedoensmax:
@jywu-msft These are the promised GA samples. We are purely using ORT APIs to make the samples cross platform, as soon as there are 1.23 binaries we also want to update the main README with some information on where to best download libraries and that one could also use Windows ML for that purpose and point to a sample.

We have a few questions:
- Since we believe these samples do not fit the pattern of samples in the inference samples repo well we decided to file the PR here. What is your guidance on that ?
- Does MSFT want to add guidance for other EPs here as well ? Feel free to involve other IHVs to add their setup guidance on the README as well.

snnn (changes requested):
Please put them in http://github.com/microsoft/onnxruntime-inference-examples. A lot of teams build ORT from source and they do not need to build the samples.  We put samples in  a different repos so that these customers won't be bothered with doing compliance work for the samples.

PR #26069, opened 2025-09-17
Disable some of rotary embedding tests as they are failing on Win GPU CUDA

titaiwangms:


onnxruntime/test/testdata/onnx_backend_test_series_filters.jsonc
```diff
@@ -127,6 +127,9 @@
         "^test_rotary_embedding_no_position_ids_rotary_dim",
         "^test_rotary_embedding_with_interleaved_rotary_dim",
         "^test_rotary_embedding_with_rotary_dim",
+        "^test_rotary_embedding_3d_input_expanded",
```
hariharans29, line 130: Maybe add a comment describing why the tests are failing ? Are they broken tests ? If so, should they be tracked as an issue in the ONNX repo and linked here with a TODO on re-enabling once the fix is consumed ?

titaiwangms: https://github.com/microsoft/onnxruntime/issues/26070

hariharans29 (approved):

PR #26071, opened 2025-09-17
[WebGPU] add unidirectional support for MHA

fs-eire:
### Description

Add unidirectional support  for MHA. Also updated `GetCapability` to make sure only run Attention on WebGPU under certain conditions

### DIFF
--- a/onnxruntime/core/providers/webgpu/webgpu_execution_provider.cc
+++ b/onnxruntime/core/providers/webgpu/webgpu_execution_provider.cc
@@ -850,6 +850,42 @@ std::vector<std::unique_ptr<ComputeCapability>> WebGpuExecutionProvider::GetCapa
       LOGS(*GetLogger(), INFO) << "Force CPU execution for node: " << node.Name();
       continue;
     }
+
+    //
+    // The following code checks if the node is really supported by webgpu EP
+    //
+
+#define FALLBACK_TO_CPU_IF_EXIST_INPUT(idx)           \
+  if (inputs.size() > idx && inputs[idx]->Exists()) { \
+    continue;                                         \
+  }
+
+#define FALLBACK_TO_CPU_IF_EXIST_OUTPUT(idx)            \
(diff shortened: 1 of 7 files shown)


onnxruntime/core/providers/webgpu/webgpu_execution_provider.cc
```diff
@@ -850,6 +850,42 @@ std::vector<std::unique_ptr<ComputeCapability>> WebGpuExecutionProvider::GetCapa
...
+    continue;                                         \
+  }
+
+#define FALLBACK_TO_CPU_IF_EXIST_OUTPUT(idx)            \
+  if (outputs.size() > idx && outputs[idx]->Exists()) { \
+    continue;                                           \
+  }
+
+    // Check for Attention
+    if (node.OpType() == "Attention" && node.Domain() == kMSDomain) {
```
guschmue: I think it would be better to throw instead of falling back.

daijh:
While optimizing Stable Diffusion Turbo, we found the `Attention` operation is currently falling back to the CPU.
This PR could help to resolve that issue, and we look forward to this PR being merged.

onnxruntime/core/providers/webgpu/webgpu_execution_provider.cc
fs-eire: will do this change

fs-eire:
> While optimizing Stable Diffusion Turbo, we found the `Attention` operation is currently falling back to the CPU. This PR could help to resolve that issue, and we look forward to this PR being merged.

I am going to merge this PR. However, this change only run `Attention` operator on WebGPU when certain condition is fulfilled. Please check https://github.com/microsoft/onnxruntime/pull/26071/files#diff-904fb03958cf9430020305a356202c6af249df571e90bd9b13bab7989e6f8dacR876-R895.

If any model still fallback to CPU because of that, this should be followed up to add implementation.

daijh:
Thanks. I will follow-up with it.

PR #26074, opened 2025-09-18
[webgpu] Use DoTranspose directly in Conv Op

wenqinI:
### Description

This refactors the `TransposeKernel` to call `Transpose::DoTranspose` directly.

### Motivation and Context

See above.

### DIFF
--- a/onnxruntime/core/providers/webgpu/nn/conv.cc
+++ b/onnxruntime/core/providers/webgpu/nn/conv.cc
@@ -21,7 +21,8 @@ Status TransposeKernel(ComputeContext& context, const Tensor* kernel, const Tens
   }
   TensorShape transposed_kernel_shape(transposed_kernel_shape_vector);
   *transposed_kernel = context.CreateGPUTensor(kernel->DataType(), transposed_kernel_shape);
-  return Transpose::DoTranspose(context, perm, *kernel, *transposed_kernel, &kernel_shape, &transposed_kernel_shape);
+  const Tensor reshaped_kernel(kernel->DataType(), kernel_shape, const_cast<void*>(kernel->DataRaw()), kernel->Location());
+  return Transpose::DoTranspose(context, perm, reshaped_kernel, *transposed_kernel);
 }
 
 template <bool is_channels_last, bool is_fused>
--- a/onnxruntime/core/providers/webgpu/tensor/transpose.cc
+++ b/onnxruntime/core/providers/webgpu/tensor/transpose.cc
@@ -110,22 +110,15 @@ Status TransposeProgram::GenerateShaderCode(ShaderHelper& shader) const {
 
 Status Transpose::DoTranspose(onnxruntime::webgpu::ComputeContext& context,
                               gsl::span<const size_t> permutations,
-                              const Tensor& input, Tensor& output, const TensorShape* input_shape_ptr, TensorShape* output_shape_ptr) {
-  const auto& input_shape = input_shape_ptr ? *input_shape_ptr : input.Shape();
+                              const Tensor& input, Tensor& output) {
+  const auto& input_shape = input.Shape();
   const auto& input_dims = input_shape.GetDims();
   int32_t rank = static_cast<int32_t>(input_shape.NumDimensions());
 
   TensorShapeVector output_dims(rank);
 
-  if (output_shape_ptr) {
-    auto dims = output_shape_ptr->GetDims();
-    for (int32_t i = 0; i < rank; ++i) {
-      output_dims[i] = dims[i];
-    }
-  } else {
-    for (int32_t i = 0; i < rank; i++) {
-      output_dims[i] = input_dims[permutations[i]];
-    }
+  for (int32_t i = 0; i < rank; i++) {
+    output_dims[i] = input_dims[permutations[i]];
   }
 
   TensorShapeVector new_shape{};
--- a/onnxruntime/core/providers/webgpu/tensor/transpose.h
+++ b/onnxruntime/core/providers/webgpu/tensor/transpose.h
@@ -16,7 +16,7 @@ class Transpose final : public WebGpuKernel, public TransposeBase {
   Transpose(const OpKernelInfo& info) : WebGpuKernel{info}, TransposeBase{info} {
   }
   Status ComputeInternal(ComputeContext& context) const override;
-  static Status DoTranspose(onnxruntime::webgpu::ComputeContext& context, gsl::span<const size_t> permutations, const Tensor& input, Tensor& output, const TensorShape* input_shape_ptr = nullptr, TensorShape* output_shape_ptr = nullptr);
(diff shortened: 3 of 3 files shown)


guschmue (dismissed):

onnxruntime/core/providers/webgpu/tensor/transpose.cc
```diff
@@ -147,9 +162,28 @@ Status Transpose::DoTranspose(onnxruntime::webgpu::ComputeContext& context,
           {static_cast<uint32_t>(output_size)},
       });
 
-  use_shared ? program.SetDispatchGroupSize(static_cast<uint32_t>((new_output_shape[1] + TILE_SIZE - 1) / TILE_SIZE),
-                                            static_cast<uint32_t>(((new_output_shape[0] + TILE_SIZE - 1) / TILE_SIZE)))
-             : program.SetDispatchGroupSize((output_size + WORKGROUP_SIZE - 1) / WORKGROUP_SIZE);
+  if(use_shared) {
```
daijh:
```
  program
      .CacheHint(absl::StrJoin(permutations, "-"))
      .AddInputs({{&input, ProgramTensorMetadataDependency::TypeAndRank, new_input_shape, 1}})
      .AddOutputs({{&output, ProgramTensorMetadataDependency::None, new_output_shape, 1}})
      .SetDispatchGroupSize(static_cast<uint32_t>((new_output_shape[1] + TILE_SIZE - 1) / TILE_SIZE),
                            static_cast<uint32_t>(((new_output_shape[0] + TILE_SIZE - 1) / TILE_SIZE)))
```

The `SetDispatchGroupSize` is called multiple times. 
It's not relevant to current PR, but nice to fix.

daijh:
```
if (use_shared) {
    program.SetWorkgroupSize(TILE_SIZE, TILE_SIZE, 1);
  }
```

Should the conditional logic (if (use_shared)) be merged with below? 

Additionally, should SetWorkgroupSize be called explicitly?

wenqinI: done.

wenqinI: done

wenqinI (commented):
Address comments, PTAL, thanks!

wenqinI:
Hi @guschmue @qjia7,

This PR is ready to review, PTAL, thanks!

onnxruntime/core/providers/webgpu/nn/conv.cc
```diff
@@ -27,37 +19,9 @@ Status TransposeKernel(ComputeContext& context, const Tensor* kernel, const Tens
...
-  TransposeProgram program{perm, use_shared};
-  program
-      .CacheHint(absl::StrJoin(perm, "-"))
-      .AddInput({kernel, ProgramTensorMetadataDependency::TypeAndRank, kernel_shape, 1})
-      .AddOutput({transposed_kernel, ProgramTensorMetadataDependency::TypeAndRank})
-      .AddUniformVariable({output_size})
-      .SetWorkgroupSize(64)
-      .SetDispatchGroupSize(dispatch_x, dispatch_y, dispatch_z);
-  return context.RunProgram(program);
+  return Transpose::DoTranspose(context, perm, *kernel, *transposed_kernel, &kernel_shape, &transposed_kernel_shape);
```
qjia7: How about not changing the original DoTranspose's parameter list and reshape the kernel to the expected kernel_shape so that we don't need to modify the original logic in DoTranspose?
For example:
```
  kernel->Reshape(kernel_shape);
  return Transpose::DoTranspose(context, perm, *kernel, *transposed_kernel);
```

wenqinI: I think there are two problems if we want to directly reshape the input tensor:
1. the `kernel_shape` is not always same as the kernel's shape, which means the `Conv` OP may change the shape of the input tensor, it may not be a good way, please see below code, it shows how `kernel_shape` was modified:
https://github.com/microsoft/onnxruntime/blob/94de31fa124c29e79c9583e3273a1a60a7e8f7b0/onnxruntime/core/providers/webgpu/nn/conv.cc#L113-L118

2. the `kernel` is a `const Tensor*`, so we may have to do some `const_cast` before we reshape it, is it ok in this semantic?

qjia7: I think `const_cast` is safe since we're not modifying the data, just creating a different view of it. I prefer `DoTranspose` to transpose an input tensor with the right shape. @fs-eire What's your opinion on this?
```suggestion
  // Create a reshaped view of the kernel tensor (same data, different shape)
  Tensor reshaped_kernel(kernel->DataType(), kernel_shape, const_cast<void*>(kernel->DataRaw()), kernel->Location());
  
  return Transpose::DoTranspose(context, perm, reshaped_kernel, *transposed_kernel);
```

wenqinI: Updated this PR, PTAL, thanks!

qjia7 (approved):

onnxruntime/core/providers/webgpu/nn/conv.cc
```diff
@@ -27,37 +19,10 @@ Status TransposeKernel(ComputeContext& context, const Tensor* kernel, const Tens
...
-  TransposeProgram program{perm, use_shared};
-  program
-      .CacheHint(absl::StrJoin(perm, "-"))
-      .AddInput({kernel, ProgramTensorMetadataDependency::TypeAndRank, kernel_shape, 1})
-      .AddOutput({transposed_kernel, ProgramTensorMetadataDependency::TypeAndRank})
-      .AddUniformVariable({output_size})
-      .SetWorkgroupSize(64)
-      .SetDispatchGroupSize(dispatch_x, dispatch_y, dispatch_z);
-  return context.RunProgram(program);
+  const Tensor reshaped_kernel(kernel->DataType(), kernel_shape, const_cast<void*>(kernel->DataRaw()), kernel->Location());
```
fs-eire, line 24: can use `MutableDataRaw` which is better than using `const_cast`. (we know that we don't modify it any way)

wenqinI: It seems `kernel` is a `const Tensor`, so it's not allow to call `MutableDataRaw`.

daijh:
@guschmue @fs-eire 

May we move forward with this PR, if there are no additional comments.

fs-eire (approved):

fs-eire:
Sorry for the delay. resubmitted failed jobs and wait for the pass to merge.

fs-eire:
The jobs keep failing and I think the error is unrelated to this change. Could you please merge to latest main branch?

wenqinI:
Hi @fs-eire ,

Thank you for your review, I rebased this PR, hope it could pass the CI test.

fs-eire:

PR #26075, opened 2025-09-18
Declare Shape, Reshape, Transpose for opsets 21, 23 on CUDA

xadupre:


yuslepukhin:
Does it fix it?

xadupre:
It works fine on the model I used to test. I did not check the whole list of missing kernels declarations. We could also change the logic behind the kernel selection. An operator is often upgraded because it supports more types but onnxruntime does not always implements the kernel for the new types.

PR #26085, opened 2025-09-18
[QNN-EP] Add GPU to QNN EP factory

qti-mattsinc:
## Description

Make the GPU backend visible to users of the WinML API.

The V2 EP selection mechanism will now expose the GPU device for QNN. The GPU can be picked when:

* The device list contains only GPU
* The execution policy is set to PREFER_GPU

If multiple devices (e.g. HTP and GPU) are provided to `AppendExecutionProviders_V2`, whichever device was provided last in the list will be used.

## Motivation and Context
Required to enable WinML usage with the QNN GPU backend

### DIFF
--- a/onnxruntime/core/providers/qnn/qnn_provider_factory.cc
+++ b/onnxruntime/core/providers/qnn/qnn_provider_factory.cc
@@ -82,15 +82,11 @@ struct QNN_Provider : Provider {
 
   Status CreateIExecutionProvider(const OrtHardwareDevice* const* /*devices*/,
                                   const OrtKeyValuePairs* const* /*ep_metadata*/,
-                                  size_t num_devices,
+                                  size_t /*num_devices*/,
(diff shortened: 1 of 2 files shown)


johnpaultaken:
@adrianlizarraga  @HectorSVC  This is the GPU enablement for V2 EP selection that I mentioned in last weeks meeting. Please approve. Thanks.

qti-mattsinc:
CI was failing for one of the added autoep tests. Looks like the CI environment does not have the appropriate hardware for running tests on QNN HTP / GPU (judging from [default_providers.cc:L253](https://github.com/microsoft/onnxruntime/blob/f6b405c9c50fe855e8745b5774d385de7238ec84/onnxruntime/test/util/default_providers.cc#L253) and from all the Qnn{HTP,GPU}BackendTests skipping), so I've removed it.

PR #26088, opened 2025-09-18
MemcpyFromHost and MemcpyToHost support for plugin EPs

edgchen1:
### Description

Add support for `MemcpyFromHost` and `MemcpyToHost` ops with plugin EPs.

- Add CPU EP fallback kernels for the memcpy ops. These are generic implementations using a data transfer manager.
- Update `SessionState::PopulateKernelCreateInfo()` to fall back to CPU memcpy kernels if a node's assigned provider doesn't have them.
- Update `MemcpyTransformer` to determine whether providers are CPU-based or compatible with other providers by looking at the device type instead of matching against a hardcoded list of provider types. This accommodates plugin EPs, where the provider type can't be hardcoded.

### Motivation and Context

Allow plugin EPs to work with models where memcpy ops are required (i.e., models where connected nodes are not fully assigned to the plugin EP).

### DIFF
--- a/onnxruntime/test/autoep/library/ep_stream_support.cc
+++ b/onnxruntime/test/autoep/library/ep_stream_support.cc
@@ -61,12 +61,11 @@ OrtStatus* ORT_API_CALL NotificationImpl::ActivateImpl(_In_ OrtSyncNotificationI
 /*static*/
 OrtStatus* ORT_API_CALL NotificationImpl::WaitOnDeviceImpl(_In_ OrtSyncNotificationImpl* this_ptr,
                                                            _In_ OrtSyncStream* stream) noexcept {
-  auto& impl = *static_cast<NotificationImpl*>(this_ptr);
-
-  // TODO is this the right place to check `stream != nullptr` or should that be done elsewhere?
   if (stream == nullptr) {
     return nullptr;
   }
+
+  auto& impl = *static_cast<NotificationImpl*>(this_ptr);
 
   void* handle = impl.ort_api.SyncStream_GetHandle(stream);
   static_cast<void>(handle);


onnxruntime/core/optimizer/transformer_memcpy.h
```diff
@@ -19,13 +23,17 @@ Transformer that inserts nodes to copy memory between devices when needed.
 */
 class MemcpyTransformer : public GraphTransformer {
  public:
-  MemcpyTransformer(const std::vector<std::string>& provider_types, const KernelRegistryManager& registry_manager)
-      : GraphTransformer("MemcpyTransformer"), provider_types_(provider_types), registry_manager_(std::cref(registry_manager)) {}
+  MemcpyTransformer(InlinedVector<gsl::not_null<const IExecutionProvider*>> providers,
```
adrianlizarraga, line 28: You probably already saw this but there's a test that is still using the old `MemcpyTransformer()` constructor:

> test/framework/memcpy_transformer_test.cc:115:93: error: no matching function for call to ‘onnxruntime::MemcpyTransformer::MemcpyTransformer(<brace-enclosed initializer list>, onnxruntime::KernelRegistryManager&)’
  115 |   MemcpyTransformer transformer({onnxruntime::kCudaExecutionProvider}, test_registry_manager);

onnxruntime/test/autoep/library/ep.cc
```diff
@@ -248,6 +248,27 @@ OrtStatus* ORT_API_CALL ExampleEp::GetCapabilityImpl(OrtEp* this_ptr, const OrtG
...
+          const auto is_static_shape = [](gsl::span<const int64_t> shape) -> bool {
+            return std::all_of(shape.begin(), shape.end(), [](int64_t dim) { return dim >= 0; });
+          };
+
+          if (!is_static_shape(*input_0_shape) || !is_static_shape(*input_1_shape)) {
+            continue;  // input shape has dynamic dimensions
+          }
+
+          if (*input_0_shape != *input_1_shape) {
+            continue;  // input shapes do not match (no broadcasting support for now)
```
edgchen1, line 268: note: added these additional GetCapability checks to pass this test when run with this plugin EP.

https://github.com/microsoft/onnxruntime/blob/f6b405c9c50fe855e8745b5774d385de7238ec84/onnxruntime/test/providers/cpu/nn/flatten_op_test.cc#L79

it has a Mul with broadcasting.

onnxruntime/core/framework/utils.h
```diff
@@ -52,12 +52,8 @@ void DestroyStrings(void* p_data, int64_t elements);
 
 const std::string& GetNodeInputProviderType(const SessionState::NodeInfo& info);
 
-// EP used for internal testing. We define it here as it's used in ProviderIsCpuBased, but we don't want
-// it to be in the public header include/onnxruntime/core/graph/constants.h as it's purely internal.
-constexpr const char* kInternalTestingExecutionProvider = "InternalTestingExecutionProvider";
```
edgchen1, line 57: note: moved to internal_testing_execution_provider.h

onnxruntime/test/autoep/library/ep_stream_support.cc
```diff
@@ -62,6 +62,12 @@ OrtStatus* ORT_API_CALL NotificationImpl::ActivateImpl(_In_ OrtSyncNotificationI
 OrtStatus* ORT_API_CALL NotificationImpl::WaitOnDeviceImpl(_In_ OrtSyncNotificationImpl* this_ptr,
                                                            _In_ OrtSyncStream* stream) noexcept {
   auto& impl = *static_cast<NotificationImpl*>(this_ptr);
+
+  // TODO is this the right place to check `stream != nullptr` or should that be done elsewhere?
+  if (stream == nullptr) {
```
edgchen1, line 64: I ran into `stream` being nullptr in a test. not sure if that should have been caught earlier or if implementors are expected to handle a possibly null stream.

onnxruntime/core/optimizer/transformer_memcpy.cc
```diff
@@ -218,10 +223,11 @@ void TransformerMemcpyImpl::ProcessDefs(onnxruntime::Node& node,
...
   auto node_provider_type = node.GetExecutionProviderType();
-  if ((node_provider_type == provider_) ||
-      (node_provider_type == kCudaExecutionProvider && kTensorrtExecutionProvider == provider_) ||
-      (node_provider_type == kCudaExecutionProvider && kNvTensorRTRTXExecutionProvider == provider_) ||
-      (node_provider_type == kRocmExecutionProvider && kMIGraphXExecutionProvider == provider_)) {
+  // TODO consider using info from provider device to detect compatibility instead of checking provider types
+  if ((node_provider_type == provider_.Type()) ||
+      (node_provider_type == kCudaExecutionProvider && kTensorrtExecutionProvider == provider_.Type()) ||
+      (node_provider_type == kCudaExecutionProvider && kNvTensorRTRTXExecutionProvider == provider_.Type()) ||
+      (node_provider_type == kRocmExecutionProvider && kMIGraphXExecutionProvider == provider_.Type())) {
```
skottmckay: Can we get rid of this by using the OrtDevice from each EP for comparisons instead of the EP name?

onnxruntime/core/providers/cpu/cpu_execution_provider.cc
```diff
@@ -39,6 +67,8 @@ std::vector<AllocatorPtr> CPUExecutionProvider::CreatePreferredAllocators() {
 }
 
 // Forward declarations of op kernels
+class ONNX_OPERATOR_KERNEL_CLASS_NAME(kCpuExecutionProvider, kOnnxDomain, 1, MemcpyFromHost);
+class ONNX_OPERATOR_KERNEL_CLASS_NAME(kCpuExecutionProvider, kOnnxDomain, 1, MemcpyToHost);
```
skottmckay, line 71: nit: can we use a different domain given these aren't onnx ops?

onnxruntime/test/autoep/library/ep_stream_support.cc
skottmckay: Stream support is optional for an EP to implement. Not sure it's expected as we wouldn't have registered any handlers in `PluginExecutionProvider::RegisterStreamHandlers`, so I wouldn't have thought there was a path to call to this function given that. Could just be the test setup though.

onnxruntime/core/providers/cpu/cpu_execution_provider.cc
edgchen1: I looked into this, and it seemed a bit more involved if we use a separate domain for these new kernels. it would require a new op schema. also, in `SessionState::PopulateKernelCreateInfo()` (where the fallback currently happens), it seems tricky to update the node's domain.

the original choice of using the onnx domain for these memcpy ops is not ideal but it seems simpler to be consistent.

onnxruntime/test/autoep/library/ep_stream_support.cc
edgchen1: looks like the test EP does claim to be stream-aware.

https://github.com/microsoft/onnxruntime/blob/1c436de5ccd05181a28cc6e8d5f04661ceab0e0e/onnxruntime/test/autoep/library/ep_factory.cc#L294-L297

I'll remove the TODO comment.

snnn:

PR #26092, opened 2025-09-19
Write parameters based on output shapes not the amount of names

TedThemistokleous:
Running into a case where the amount of names being larger than inputs from the MIGraphX program causes us to go out of bounds and try to index outputs that don't exist. This leads to undefined/broken behavior (faults)

### Description

This fixes an out of bounds condition that occurs when output_names() errornoursly is used to bind size of the amount of inputs used for setting parameters in MIGraphX EP

### Motivation and Context

Seeing this in some cases for user data

### DIFF
--- a/onnxruntime/core/providers/migraphx/migraphx_execution_provider.cc
+++ b/onnxruntime/core/providers/migraphx/migraphx_execution_provider.cc
@@ -1359,7 +1359,7 @@ Status MIGraphXExecutionProvider::Compile(const std::vector<FusedNodeAndGraph>& 
       }
 
       auto prog_output_shapes = prog.get_output_shapes();
-      for (std::size_t i = 0; i < output_names.size(); ++i) {
+      for (std::size_t i = 0; i < prog_output_shapes.size(); ++i) {
         auto out_len = prog_output_shapes[i].lengths();
         options.set_input_parameter_shape(output_names[i], out_len);
       }


snnn (approved):
This looks like a good bug fix. Using `prog_output_shapes.size()` as the loop bound is the correct approach to prevent the out-of-bounds access.

I have a question about testing: How is this change tested? Are there any existing tests that cover this scenario, or should a new test be added to prevent regressions?

I am approving this PR, but it would be great to get some information on the testing strategy.

onnxruntime/core/providers/migraphx/migraphx_execution_provider.cc
```diff
@@ -1359,7 +1359,7 @@ Status MIGraphXExecutionProvider::Compile(const std::vector<FusedNodeAndGraph>&
       }
 
       auto prog_output_shapes = prog.get_output_shapes();
-      for (std::size_t i = 0; i < output_names.size(); ++i) {
+      for (std::size_t i = 0; i < prog_output_shapes.size(); ++i) {
         auto out_len = prog_output_shapes[i].lengths();
         options.set_input_parameter_shape(output_names[i], out_len);
```
snnn, line 1364: How do we ensure that output_names[i] won't have an out-of-bound access error?

TedThemistokleous:
> This looks like a good bug fix. Using `prog_output_shapes.size()` as the loop bound is the correct approach to prevent the out-of-bounds access.
> 
> I have a question about testing: How is this change tested? Are there any existing tests that cover this scenario, or should a new test be added to prevent regressions?
> 
> I am approving this PR, but it would be great to get some information on the testing strategy.

Appreciate it

We were seeing this in a customer model. There's a case we're hitting right now where it appears between the GetCapability() call in the EP and Compile() steps, the entire graph, including dangling outputs of some operators are being fused and used as final outputs instead of dropped.

MIGraphX by default will prune dangling/dead outputs as part of our compile and what happens in the EP is since we were traversing by output_names originally, would get an out of bounds index error as MIGraphX is seeing 1 output, wheras Onnxruntime is seeing 17 (8 ops with 2 dangling outputs + 1 real output).

Seems like the result from the graph_viewer is errornously including the output of intermediary nodes in the final output list which is what's also causing this. This fix ultimately solves correctness withing the MIGraphX EP here since we should only be iterating over the known amount of inputs/outputs from the MIGraphX program.

@snnn if you have any idea how to do the output prune I'm all ears. Its leading to anther what appears failure/signal during runs through the EP wheras the same model (even after compiling through the EP) appears fine using MIGraphX tooling (migraphx-driver)

To further add to this we've dumped the resulting onnx file after we get it before MIGraphX does a compile() and Onnxruntime is fundamentally changing the model by adding the additional dangling operator outputs as final outputs.  I was able to catch and see this visually using a tool like netron

TedThemistokleous:
@snnn  refer to the issue I just opened - https://github.com/microsoft/onnxruntime/issues/26176

Seeing weird behavior or I'm just not drinking enough coffee to determine where or why this is occurring.

onnxruntime/core/providers/migraphx/migraphx_execution_provider.cc
TedThemistokleous: output_names is gathered from graph viewer and it has a list of all the output names

TedThemistokleous:
@devang-ml  can we get this in since this is approved?

TedThemistokleous:
ping @devang-ml @snnn  @tianleiwu  this is approved and all green. Can we get this merged?

PR #26103, opened 2025-09-20
[DO NOT REVIEW OR MERGE] Test PR

hariharans29:
### Description
TBD

### Motivation and Context
TBD

### DIFF
--- a/onnxruntime/core/mlas/lib/convolve.cpp
+++ b/onnxruntime/core/mlas/lib/convolve.cpp
@@ -729,6 +729,61 @@ Return Value:
     }
 }
 
+void
+MlasConvExpandThenGemmSegmentedThreaded(
+    void* Context,
+    ptrdiff_t Index
+)
+{
+    MLAS_CONV_WORK_BLOCK* WorkBlock = (MLAS_CONV_WORK_BLOCK*)Context;
+
+    const MLAS_CONV_PARAMETERS* Parameters = WorkBlock->Parameters;
+
+    const size_t GroupCount = Parameters->GroupCount;
+    const size_t BatchGroupCount = Parameters->BatchCount * GroupCount;
+
+    const size_t TargetThreadCount = WorkBlock->TargetThreadCount;
+
+    const size_t BatchGroupCountPerThread = BatchGroupCount / TargetThreadCount;
+    const size_t BatchGroupCountExtra = BatchGroupCount % TargetThreadCount;
+
+    size_t BatchGroupStart;
+    size_t BatchGroupEnd;
+
+    if (uint32_t(Index) < BatchGroupCountExtra) {
+        BatchGroupStart = (BatchGroupCountPerThread + 1) * Index;
+        BatchGroupEnd = BatchGroupStart + BatchGroupCountPerThread + 1;
+    } else {
+        BatchGroupStart = BatchGroupCountPerThread * Index + BatchGroupCountExtra;
+        BatchGroupEnd = BatchGroupStart + BatchGroupCountPerThread;
+    }
+
+    const size_t FilterCount = Parameters->FilterCount;
+    const size_t OutputSize = Parameters->OutputSize;
+    const size_t K = Parameters->K;
+
+    const size_t InputGroupSize = Parameters->InputChannels * Parameters->InputSize;
+    const size_t OutputGroupSize = FilterCount * OutputSize;
+    const size_t FilterGroupSize = FilterCount * K;
+
+    // std::cout << "Address of WorkBlock->WorkingBuffer" << WorkBlock->WorkingBuffer << std::endl;
+
+    for (size_t bg = BatchGroupStart; bg < BatchGroupEnd; bg++) {
+        size_t group = bg % GroupCount;
+
+        const float* input = WorkBlock->Input + bg * InputGroupSize;
+        const float* filter = WorkBlock->Filter + group * FilterGroupSize;
+        float* output = WorkBlock->Output + bg * OutputGroupSize;
+        const float* bias = WorkBlock->Bias;
+        if (bias != nullptr) {
+            bias += group * FilterCount;
+        }
+        float* ColumnBuffer = WorkBlock->WorkingBuffer + Index * OutputSize * K;
+
+        MlasConvOperation(Parameters, input, filter, bias, ColumnBuffer, output, 0, OutputSize);
+    }
+}
+
 inline
 bool
 MlasConvTryMultithread(
@@ -918,6 +973,32 @@ Return Value:
     }
 
 #endif
+
+    if (Algorithm == MlasConvAlgorithmExpandThenGemmSegmented && ((BatchCount > 1) || (GroupCount > 1))) {
+        const size_t BatchGroupCount = BatchCount * GroupCount;
+
+        ptrdiff_t TargetThreadCount = MlasGetMaximumThreadCount(ThreadPool);
+        // TargetThreadCount = 16;
+
+
+        if (size_t(TargetThreadCount) >= BatchGroupCount) {
+            TargetThreadCount = int32_t(BatchGroupCount);
+        }
+
+        MLAS_CONV_WORK_BLOCK WorkBlock;
+
+        WorkBlock.Parameters = Parameters;
+        WorkBlock.Input = Input;
+        WorkBlock.Filter = Filter;
+        WorkBlock.Bias = Bias;
+        WorkBlock.WorkingBuffer = WorkingBuffer;
+        WorkBlock.Output = Output;
+        WorkBlock.TargetThreadCount = TargetThreadCount;
+
+        MlasExecuteThreaded(MlasConvExpandThenGemmSegmentedThreaded, &WorkBlock, TargetThreadCount, ThreadPool);
+
+        return;
+    }
 
     //
     // Iterate over each batch and group.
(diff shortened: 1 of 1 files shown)


onnxruntime/core/mlas/lib/convolve.cpp
```diff
@@ -729,6 +729,59 @@ Return Value:
...
+
+    const size_t TargetThreadCount = WorkBlock->TargetThreadCount;
+
+    const size_t BatchGroupCountPerThread = BatchGroupCount / TargetThreadCount;
+    const size_t BatchGroupCountExtra = BatchGroupCount % TargetThreadCount;
+
+    size_t BatchGroupStart;
+    size_t BatchGroupEnd;
+
+    if (static_cast<uint32_t>(Index) < BatchGroupCountExtra) {
```
edgchen1:
```suggestion
    if (static_cast<size_t>(Index) < BatchGroupCountExtra) {
```

onnxruntime/core/mlas/lib/convolve.cpp
```diff
@@ -729,6 +729,59 @@ Return Value:
     }
 }
 
+void
+MlasConvExpandThenGemmSegmentedThreaded(
+    void* Context,
+    ptrdiff_t Index
+)
+{
```
edgchen1, line 760: could we add some documentation for this function? in particular, it might be useful to describe how it differs from `MlasConvTryMultithread()`

onnxruntime/core/mlas/lib/convolve.cpp
```diff
@@ -1308,6 +1385,16 @@ Return Value:
         Parameters->u.ExpandThenGemmSegmented.ThreadStrideN = StrideN;
 
         *WorkingBufferSize = TargetThreadCount * MLAS_CONV_WORKING_BUFFER_SIZE_PER_THREAD;
+
+        if (Parameters->BatchCount > 1 || Parameters->GroupCount > 1) {
+
+            size_t WorkingBufferSizePerThread = std::max(Parameters->OutputSize * Parameters->K, std::max(Parameters->FilterCount * Parameters->OutputSize, static_cast<size_t>(MLAS_CONV_WORKING_BUFFER_SIZE_PER_THREAD)));
```
edgchen1: nit: line length

edgchen1: also, there's an `std::max` overload which takes more than one argument

onnxruntime/core/mlas/lib/convolve.cpp
```diff
@@ -919,6 +972,30 @@ Return Value:
 
 #endif
 
+    if (Algorithm == MlasConvAlgorithmExpandThenGemmSegmented && ((BatchCount > 1) || (GroupCount > 1))) {
+        const size_t BatchGroupCount = BatchCount * GroupCount;
+
+        ptrdiff_t TargetThreadCount = MlasGetMaximumThreadCount(ThreadPool);
+
+        if (size_t(TargetThreadCount) >= BatchGroupCount) {
+            TargetThreadCount = ptrdiff_t(BatchGroupCount);
```
edgchen1: nit: `static_cast` instead of function-style casts

hariharans29: I thought of that - but on line 947 - we seem to apply C style cast. So, I kept it consistent.

onnxruntime/core/mlas/lib/convolve.cpp
hariharans29: I am not quite sure of that myself. From what I understand, MlasConvTryMultithread() tries to split the work across the `N` dimension (i.e.) the output columns, whereas this implementation spreads the work across the batch size and group count of the convolution operation. It is not clear to me which approach is better when we have minimal parallelization opportunity across the batch and group dimensions. Let us say when BS = 2 and group count = 1, we still will use this new path but we won't invoke the `MlasConvTryMultithread()` which may be quite detrmental. Perhaps, we should only use the new approach when BS * group_count >= threshold. The threshold will be tuneable.

onnxruntime/core/mlas/lib/convolve.cpp
hariharans29: Fixed here and the other instance of the same issue too

onnxruntime/core/mlas/lib/convolve.cpp
hariharans29: Then again, usually people pass in a batch of images == cpu core count - so I guess it should be okay.

Maybe another PR can introduce "online" algorithm selection for the Conv algorithm (i.e.) run all possible algorithms for a provided {activation, filter} combination and cache the best algorithm to be used.

onnxruntime/core/mlas/lib/convolve.cpp
hariharans29: Thanks for the tip. Used the initializer list and reduced line length.

onnxruntime/test/mlas/bench/bench_sconv.cpp
```diff
@@ -138,6 +139,113 @@ void SCONV_NCHW(benchmark::State& state, const char* /*dummy*/) {
   }
 }
 
+MLAS_THREADPOOL* GetMlasThreadPool(void) {
+  static auto threadpool = std::make_unique<onnxruntime::concurrency::ThreadPool>(
+      &onnxruntime::Env::Default(), onnxruntime::ThreadOptions(), nullptr, 4, true);
+  return threadpool.get();
+}
```
edgchen1: `GetMlasThreadPool(void)` is already defined here:

https://github.com/microsoft/onnxruntime/blob/cd4ac4943465fbbadbbc622dcfd11f453a27a2c1/onnxruntime/test/mlas/unittest/test_main.cpp#L10-L24

instead of providing another definition, can we maybe refactor the existing one to allow customization?

e.g., maybe have a helper `std::unique_ptr<MLAS_THREADPOOL> CreateMlasThreadPool(int degree_of_parallelism)` and update `GetMlasThreadPool()` to call that and also call that here directly.

edgchen1: also, it would be good to limit inclusion of ORT headers (like `#include "core/util/thread_utils.h"`) to fewer files.

onnxruntime/core/mlas/lib/convolve.cpp
```diff
@@ -891,7 +967,7 @@ Return Value:
         ptrdiff_t TargetThreadCount = MlasGetMaximumThreadCount(ThreadPool);
 
         if (size_t(TargetThreadCount) >= BatchGroupCount) {
```
edgchen1: I know this is how it was before this PR, but see https://google.github.io/styleguide/cppguide.html#Casting
```suggestion
        if (static_cast<size_t>(TargetThreadCount) >= BatchGroupCount) {
```

onnxruntime/core/mlas/lib/convolve.cpp
```diff
@@ -919,6 +995,30 @@ Return Value:
 
 #endif
 
+    if (Algorithm == MlasConvAlgorithmExpandThenGemmSegmented && ((BatchCount > 1) || (GroupCount > 1))) {
+        const size_t BatchGroupCount = BatchCount * GroupCount;
+
+        ptrdiff_t TargetThreadCount = MlasGetMaximumThreadCount(ThreadPool);
+
+        if (size_t(TargetThreadCount) >= BatchGroupCount) {
```
edgchen1:
```suggestion
        if (static_cast<size_t>(TargetThreadCount) >= BatchGroupCount) {
```

onnxruntime/test/mlas/bench/bench_sconv.cpp
hariharans29: Can this be done in another PR ? There are quite a few call sites that need modifying. Can I make this a file scoped function for now and rename it to indicate isolation to this benchmark ?

hariharans29: Made it static and renamed it. WIll make the change to remove the duplicate definition in another PR later.

edgchen1: renaming it sounds good. the important thing is that we don't have two definitions of the same function.

onnxruntime/core/mlas/lib/convolve.cpp
```diff
@@ -1308,6 +1408,18 @@ Return Value:
         Parameters->u.ExpandThenGemmSegmented.ThreadStrideN = StrideN;
 
         *WorkingBufferSize = TargetThreadCount * MLAS_CONV_WORKING_BUFFER_SIZE_PER_THREAD;
+
+        if (Parameters->BatchCount > 1 || Parameters->GroupCount > 1) {
+
+            size_t WorkingBufferSizePerThread = std::max({Parameters->OutputSize * Parameters->K, 
+                                                          Parameters->FilterCount * Parameters->OutputSize, 
+                                                          static_cast<size_t>(MLAS_CONV_WORKING_BUFFER_SIZE_PER_THREAD)});
+            TargetThreadCount = MaximumThreadCount;
```
edgchen1, line 1417: `TargetThreadCount` is getting set again here, but it is not being used to update `Parameters->ThreadCount`, as it is above. is this intentional?

onnxruntime/test/mlas/bench/bench_sconv.cpp
edgchen1: > There are quite a few call sites that need modifying.

to clarify, what I wanted to suggest was something like this:

```c++
std::unique_ptr<MLAS_THREADPOOL> CreateMlasThreadPool(int degree_of_parallelism) {
  return std::make_unique<onnxruntime::concurrency::ThreadPool>( 
             &onnxruntime::Env::Default(), onnxruntime::ThreadOptions(), nullptr, degree_of_parallelism, true);
}

MLAS_THREADPOOL* GetMlasThreadPool(void) { 
  static auto threadpool = CreateMlasThreadPool(2);
  return threadpool.get();
}
```

so the signature of `GetMlasThreadPool()` is unchanged. I don't think it should require modifying existing call sites.

onnxruntime/core/mlas/lib/convolve.cpp
hariharans29: Hmm - good point. I don't know if it was intentional or not. In any case, I think the TargetThreadCount here is only for finding the size of the woking buffer.

As an aside, I don't see any execution path within MlasConv() actually reference Parameters->ThreadCount !!? Each implementation seems to disregard the the Parameters->ThreadCount and comes up with its own target thread count based on what the implementatation is parallelizing on which it then adds to its WorkBlock instance.

edgchen1: ok, if `Parameters->ThreadCount` is not actually used, perhaps it would be worth cleaning up later

onnxruntime/test/mlas/bench/bench_sconv.cpp
hariharans29: Ah ok. Makes sense. If I push any more changes, I ll incorporate it.

apsonawane:

Issue #24003, opened 2025-03-12
[Build] Missmatch between CMake config and folder structure of  onnxruntime-linux-x64-1.21.0.tgz


Zyrin:
### Describe the issue

Just unpacking `onnxruntime-linux-x64-1.21.0.tgz` (the gpu tgz has the same problem) to where the system expects libraries and include files (in my case `/usr`) and then trying to build against this release by using the provided CMake config fails with two errors:

1. `onnxruntimeTargets-release.cmake` sets `IMPORTED_LOCATION_RELEASE` to`.../lib64/libonnxruntime.so.1.21.0`. This does not correspond with the folder structure for the libraries in the tgz being `lib`, so the so-file can not be found.
2. `onnxruntimeTargets.cmake` sets `INTERFACE_INCLUDE_DIRECTORIES` to `.../include/onnxruntime`. This does not correspond with the folder structure for includes in the tgz being `include`, so the headers can not be found.

If I my voice my opinion: the CMake files expecting the headers in `include/onnxruntime` is a good thing as copying headers from different libraries directly into `include` has  an overall high likelihood of name clashes between libraries.

### Target platform

Linux (Ubuntu 22.04 in my case)

### Build script

```
cmake_minimum_required(VERSION 3.22)
project(myproject)

find_package(onnxruntime REQUIRED)

add_executable(${PROJECT_NAME})
target_sources(${PROJECT_NAME} PRIVATE main.cpp)
target_link_libraries(${PROJECT_NAME} PRIVATE onnxruntime::onnxruntime)
```

### Error / output

1. 
```
CMake Error at /usr/lib/cmake/onnxruntime/onnxruntimeTargets.cmake:83 (message):
  The imported target "onnxruntime::onnxruntime" references the file

     "/usr/lib64/libonnxruntime.so.1.21.0"

  but this file does not exist.  Possible reasons include:

  * The file was deleted, renamed, or moved to another location.

  * An install or uninstall procedure did not complete successfully.

  * The installation package was faulty and contained

     "/usr/lib/cmake/onnxruntime/onnxruntimeTargets.cmake"

  but not all the files it references.
```

2. After fixing the wrong library path manually:
```
CMake Error in CMakeLists.txt:
  Imported target "onnxruntime::onnxruntime" includes non-existent path

    "/usr/include/onnxruntime"

  in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:

  * The path was deleted, renamed, or moved to another location.

[...]

PR #26104, opened 2025-09-20
Consistent with the configuration in the packaged cmake

WangFengtu1996:
Shared library path and include path are not   same  in release file `onnxruntime-linux-x64-gpu-1.22.0.tgz` .

1. The library file named libonnxruntime.so.1.22.0 in `onnxruntimeTargets-release.cmake`  is at `lib64/libonnxruntime.so.1.22.0`, but file in `lib` path. 
2.  The include path in `onnxruntimeTargets.cmake`  is in `${_IMPORT_PREFIX}/include/onnxruntime`, but path in `include` path.

### Description

1. 
image

`  IMPORTED_LOCATION_RELEASE "${_IMPORT_PREFIX}/lib64/libonnxruntime.so.1.22.0"`

2. 
image
`  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include/onnxruntime"`

### Motivation and Context

https://github.com/microsoft/onnxruntime/issues/24003#issue-2913355372

### DIFF
--- a/tools/ci_build/github/linux/copy_strip_binary.sh
+++ b/tools/ci_build/github/linux/copy_strip_binary.sh
@@ -18,8 +18,6 @@ EXIT_CODE=1
 uname -a
 cd "$BINARY_DIR"
 mv installed/usr/local $ARTIFACT_NAME
-mv $ARTIFACT_NAME/include/onnxruntime/* $ARTIFACT_NAME/include
-rmdir $ARTIFACT_NAME/include/onnxruntime
 # Do not ship onnx_test_runner
 rm -rf $ARTIFACT_NAME/bin
 echo "Copy debug symbols in a separate file and strip the original binary."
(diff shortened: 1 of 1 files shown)


WangFengtu1996:
* The config file named `onnxruntime-linux-x64-gpu-1.22.0\lib\pkgconfig\libonnxruntime.pc` is also `lib64` and  `include/onnxruntime`
```
prefix=/usr/local
bindir=${prefix}/bin
mandir=${prefix}/share/man
docdir=${prefix}/share/doc/onnxruntime
libdir=${prefix}/lib64
includedir=${prefix}/include/onnxruntime

Name: onnxruntime
Description: ONNX runtime
URL: https://github.com/microsoft/onnxruntime
Version: 1.22.0
Libs: -L${libdir} -lonnxruntime
Cflags: -I${includedir}

```

snnn:
The "optional Lint" can be ignored. Sometimes the pipeline does not work.

hariharans29:
> The "optional Lint" can be ignored. Sometimes the pipeline does not work.

Cool - FWIW- I fixed it in the other PR - the issue was that reviewdog had hit a threshold of warning annotations - I reduced it down to 3 which is within the acceptable threshold and keeps the check happy. But we can merge if you think this is okay.

WangFengtu1996:
This PR benefits all CPP users, thank you！！！

PR #26115, opened 2025-09-22
[NV TensorRT RTX EP] Package CUDA Runtime and TRT RTX with NV TensorRT RTX EP python wheel

keshavv27:
### Description
- CUDA Runtime added as a dependency for NV EP python wheel
- TRT RTX dlls and license copied to wheel as public wheels for TRT RTX are currently unavailable
-  Use onnxruntime.preload_dlls() to load CUDA Runtime DLL

### Motivation and Context
Enables out-of-the-box usage for the NV TensorRT RTX EP python wheel

### DIFF
--- a/cmake/onnxruntime_python.cmake
+++ b/cmake/onnxruntime_python.cmake
@@ -962,13 +962,31 @@ if (onnxruntime_USE_TENSORRT)
 endif()
 
 if (onnxruntime_USE_NV)
+  if ( OR ${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
+      file(GLOB NV_LIB_FILES LIST_DIRECTORIES false "${TENSORRT_RTX_ROOT}/lib/tensorrt_*.dll"
+                                             "${TENSORRT_RTX_ROOT}/lib/libtensorrt_*.so.*")
+    message(STATUS "NV lib files: " ${NV_LIB_FILES})
+  endif()
   add_custom_command(
     TARGET onnxruntime_pybind11_state POST_BUILD
     COMMAND ${CMAKE_COMMAND} -E copy
+        ${NV_LIB_FILES}
         $<TARGET_FILE:onnxruntime_providers_nv_tensorrt_rtx>
         $<TARGET_FILE:onnxruntime_providers_shared>
         $<TARGET_FILE_DIR:${build_output_target}>/onnxruntime/capi/
   )
+  if (EXISTS "${TENSORRT_RTX_ROOT}/doc/LICENSE.txt")
+    add_custom_command(
+      TARGET onnxruntime_pybind11_state POST_BUILD
+        COMMAND ${CMAKE_COMMAND} -E copy "${TENSORRT_RTX_ROOT}/doc/LICENSE.txt" $<TARGET_FILE_DIR:${build_output_target}>/onnxruntime/TRT_RTX_LICENSE.txt
+    )
+  endif()
+  if (EXISTS "${TENSORRT_RTX_ROOT}/doc/Acknowledgements.txt")
+    add_custom_command(
+      TARGET onnxruntime_pybind11_state POST_BUILD
+        COMMAND ${CMAKE_COMMAND} -E copy "${TENSORRT_RTX_ROOT}/doc/Acknowledgements.txt" $<TARGET_FILE_DIR:${build_output_target}>/onnxruntime/TRT_RTX_Acknowledgements.txt
+    )
+  endif()
 endif()
 
 if (onnxruntime_USE_MIGRAPHX)
--- a/setup.py
+++ b/setup.py
@@ -374,6 +374,12 @@ if platform.system() == "Linux" or platform.system() == "AIX":
         "libHtpPrepare.so",
     ]
     dl_libs.extend(qnn_deps)
+    #NV TensorRT RTX
+    nv_tensorrt_rtx_deps = [
+        "libtensorrt_rtx.so.*",
+        "libtensorrt_onnxparser_rtx.so.*"
+    ]
+    dl_libs.extend(nv_tensorrt_rtx_deps)
     if nightly_build:
         libs.extend(["libonnxruntime_pywrapper.so"])
 elif platform.system() == "Darwin":
@@ -454,6 +460,12 @@ else:
         "migraphx_tf.dll",
     ]
     libs.extend(migraphx_deps)
+    #NV TensorRT RTX Libs
+    nv_tensorrt_rtx_deps = [
+        "tensorrt_onnxparser_rtx_*.dll",
+        "tensorrt_rtx_*.dll"
+    ]
+    libs.extend(nv_tensorrt_rtx_deps)
 
 if is_manylinux:
     if is_openvino:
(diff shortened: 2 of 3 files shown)


keshavv27:
@chilo-ms @jywu-msft @snnn @tianleiwu pinging for review, Thanks!

hlevring:
Would be very convenient to get this landed asap

tianleiwu:
Please run lintrunner to format: https://github.com/microsoft/onnxruntime/blob/main/docs/Coding_Conventions_and_Standards.md#linting

setup.py
```diff
@@ -374,6 +374,13 @@ def finalize_options(self):
         "libHtpPrepare.so",
     ]
     dl_libs.extend(qnn_deps)
+    #NV TensorRT RTX
+    nv_tensorrt_rtx_deps = [
+        "libtensorrt_rtx.so.*",
+        "libtensorrt_onnxparser_rtx.so.*"
```
gedoensmax: Is this correct or are there also libraries without a version at the end ? E.g. `libtensorrt_rtx.so` as a symlink to `libtensorrt_rtx.so.1`.

cmake/onnxruntime_python.cmake
```diff
@@ -962,13 +962,33 @@ if (onnxruntime_USE_TENSORRT)
 endif()
 
 if (onnxruntime_USE_NV)
+  if (WIN32 OR ${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
+      file(GLOB NV_LIB_FILES LIST_DIRECTORIES false "${TENSORRT_RTX_ROOT}/lib/tensorrt_*.dll"
+                                             "${TENSORRT_RTX_ROOT}/bin/tensorrt_*.dll"
+                                             "${TENSORRT_RTX_ROOT}/lib/libtensorrt_*.so.*"
+                                             "${TENSORRT_RTX_ROOT}/bin/libtensorrt_*.so.*")
+    message(STATUS "NV lib files: " ${NV_LIB_FILES})
+  endif()
```
gedoensmax, line 971: This will not work if TRT RTX is installed as a system package. I think this is currently not easily supported via `apt install` or something. That being said if you use [`find_path`](https://cmake.org/cmake/help/latest/command/find_path.html) instead of a file GLOB you can include system install paths and so on.

setup.py
keshavv27: Yes, there are two symlinks libtensorrt_rtx.so and libtensorrt_rtx.so.1 for libtensorrt_rtx.so.1.1.1. In the wheel the libs are getting duplicated in place of the symlink.

keshavv27: Will update the path here to just use the symlink libtensorrt_rtx.so

gedoensmax: It depends on which library the actual pybind lib is using. You can check that with `ldd`.

setup.py
```diff
@@ -496,8 +509,8 @@ def finalize_options(self):
 examples_names = ["mul_1.onnx", "logreg_iris.onnx", "sigmoid.onnx"]
 examples = [path.join("datasets", x) for x in examples_names]
 
-# Extra files such as EULA and ThirdPartyNotices (and Qualcomm License, only for QNN release packages)
-extra = ["LICENSE", "ThirdPartyNotices.txt", "Privacy.md", "Qualcomm_LICENSE.pdf"]
+# Extra files such as EULA and ThirdPartyNotices (and Qualcomm License, only for QNN release packages)(TRT RTX and CUDA License for NV TRT RTX EP release packages)
+extra = ["LICENSE", "ThirdPartyNotices.txt", "Privacy.md", "Qualcomm_LICENSE.pdf", "TRT_RTX_LICENSE.txt", "TRT_RTX_Acknowledgements.txt"]
```
snnn: ONNX Runtime is released under MIT license, which cannot be changed. TRT's license should be put in https://github.com/microsoft/onnxruntime/blob/main/ThirdPartyNotices.txt .

ThirdPartyNotices.txt
```diff
@@ -6154,3 +6154,4814 @@ Redistribution and use in source and binary forms, with or without modification,
...
+2.12    Reverse engineer, decompile or disassemble any portion of the output generated using Software elements for the purpose of translating such output artifacts to target a non-NVIDIA platform; or
+2.13    Distribute or disclose to third parties the output of the Software where the output reveals functionality or performance data pertinent to NVIDIA hardware or software products, results of benchmarking, competitive analysis, or regression or performance data relating to the Software or NVIDIA GPUs without the prior written permission from NVIDIA.
+3.  Authorized Users. You may allow employees and contractors of your entity or of your subsidiary(ies), and for educational institutions also enrolled students, to internally access and use the Software as authorized by this Agreement from your secure network to perform the work authorized by this Agreement on your behalf. You are responsible for the compliance with the terms of this Agreement by your authorized users. Any act or omission that if committed by you would constitute a breach of this Agreement will be deemed to constitute a breach of this Agreement if committed by your authorized users.
```
snnn: This license agreement seems very different than what open source software should have.

ThirdPartyNotices.txt
```diff
@@ -6154,3 +6154,4814 @@ Redistribution and use in source and binary forms, with or without modification,
...
+1.2 License Grant to NVIDIA. Subject to the terms of this Agreement, you grant NVIDIA and its affiliates a non-exclusive, perpetual, irrevocable, sublicensable, worldwide, royalty-free, fully paid-up and transferable license, under your intellectual property rights, to publicly perform, publicly display, reproduce, use, make, have made, sell, offer for sale, distribute (through multiple tiers of distribution), import, create derivative works of and otherwise commercialize and exploit at NVIDIA's discretion any Derivatives created by or for you. You may, but are not required to, deliver any Derivatives to NVIDIA.
+2. License Restrictions. Your license to use the Software and Derivatives is restricted as stated in this Section 2 ("License Restrictions"). You will cooperate with NVIDIA and, upon NVIDIA's written request, you will confirm in writing and provide reasonably requested information to verify your compliance with the terms of this Agreement.  You may not:
+2.1 Use the Software or Derivatives for any purpose other than the Purpose;
```
snnn: I am concerned about the restrictions. ONNX Runtime is an open source project. We encourage people to create derived work from it and modify it.  They can sell it or redistribute it .

setup.py
snnn: I understand TRT RTX is a commercial software. So, if the ORT package redistribute TRT RTX DLLs, we need to have this license. But we don't. So there should be something different here.

snnn: Qualcomm_LICENSE.pdf was added here because ORT's nuget package and python package contains parts for QNN SDK.

snnn:
The [ThirdPartyNotices.txt](https://github.com/microsoft/onnxruntime/pull/26115/files/4f29ba5a94b7ff818631469f867774a9f60f5e10#diff-39b094970d276cb11c8bf4cdd1284b2a92cb8cf8ab503c72009eea75bf3ae492) is for third-party components that we distribute. i.e. what ORT packages contain.

snnn:
I am asking my colleagues regarding to the license updates.

ThirdPartyNotices.txt
snnn: @keshavv27 , can you explain why is this this license notice needed here? Is it for which library?

snnn: The "NVIDIA" license.

snnn:
Who will build and publish this package, Microsoft or Nvidia?

ThirdPartyNotices.txt
keshavv27: This is copied from the TensorRT RTX license

keshavv27:
> Who will build and publish this package, Microsoft or Nvidia?
Nvidia will release the package, but this change will enable developers to build the package from source also

setup.py
snnn: Sorry, would you mind reverting the change back to have a separately TRT_RTX_LICENSE.txt file there? That will be easier to maintain. 

I have no other comments to add.  I will approve and merge this PR short after.

snnn (approved):

setup.py
keshavv27: Thanks!

Issue #26079, opened 2025-09-18
IO Bindings in python API do not allow binding raw values using vendor_id


gedoensmax:
The API used here: https://github.com/microsoft/onnxruntime-inference-examples/blob/main/python/api/onnxruntime-python-api.py#L67-L81
Does not yet support the string "gpu" and a vendor ID. 
Internally the python bindings do support this but the python wrapper is lacking support:
https://github.com/microsoft/onnxruntime/blob/8705c68c66f5009339ce2b99570ff23246449ef0/onnxruntime/python/onnxruntime_inference_collection.py#L835

PR #26119, opened 2025-09-22
Allow vendor IDs in Python IOBinding device binds

Jonahcb:
### Description
- Extend IOBinding.bind_input and IOBinding.bind_output to accept an optional vendor_id
- Funnel both methods through OrtDevice.make(...) so all device creation shares the same validation and vendor-handling logic

### Motivation and Context
- Without this change, Python callers cannot forward a PCI vendor id when binding buffers
- Fixes issue #26079


snnn (approved):
Thanks!

jonahbernard:
> Thanks!

@snnn will this get merged automatically? Thank you!

PR #26132, opened 2025-09-23
[TRT RTX EP] Fix bug for generating the correct subgraph in GetCapability

chilo-ms:
### Description
In current TRT RTX EP/ TRT EP implementation, when constructing the `IndexedSubGraph`, for some cases, it will include the node's unused output as the SubGraph's output. So, it will return the incorrect `IndexedSubGraph` from its GetCapability to ORT.
Add the logic to prevent adding the unused node's output.

With this fix, we can avoid generating the incorrect EPContext model where the EPContext node has unused output.


yuslepukhin:
I am not seeing a test that demos that this is fixed.

jywu-mysoft:
@chilo-ms add a test in a separate PR.  this PR needs to get into rel-1.23.1 as soon as possible.

chilo-ms:
Here is the separate PR to add the unit test.
https://github.com/microsoft/onnxruntime/pull/26139

Unfortunately, at this point, we don't have a CI to run TRT RTX EP unit tests, will have to run it locally.

snnn:

PR #26146, opened 2025-09-24
Add logging macros for KleidiAI and fix todos from previous PR

orlmon01:
### Key changes
This patch contains logging macros for the KleidiAI kernels
It also contains changes / todos from a previous PR:
https://github.com/microsoft/onnxruntime/pull/25187

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -678,25 +678,14 @@ if(onnxruntime_USE_SVE)
   endif()
 endif()
 
-if (onnxruntime_USE_KLEIDIAI AND (
-      (onnxruntime_target_platform STREQUAL "aarch64") OR
-      (onnxruntime_target_platform STREQUAL "ARM64") OR
-      (APPLE AND CMAKE_SYSTEM_PROCESSOR STREQUAL "arm64")))
-
-  # TODO Add checks for MSVC Compilation
-  if(NOT MSVC)
-    check_cxx_compiler_flag(-march=armv8.2-a+dotprod HAS_ARM64_DOTPROD)
-    check_cxx_compiler_flag(-march=armv8.2-a+i8mm HAS_ARM64_I8MM)
-    if (NOT HAS_ARM64_DOTPROD)
-      message(FATAL_ERROR "The compiler doesn't support dotprod")
-    endif()
-    if (NOT HAS_ARM64_I8MM)
-      message(FATAL_ERROR "The compiler doesn't support i8mm")
-    endif()
-  else()
-    message(STATUS "Skipping -march= checks on MSVC (not supported), assuming dotprod/i8mm support manually.")
-    set(HAS_ARM64_DOTPROD TRUE)
-    set(HAS_ARM64_I8MM TRUE)
+if(onnxruntime_USE_KLEIDIAI)
+  # If KLEIDIAI_DEBUG is enabled that implies both DEBUG and KERNEL messages.
+  if(KLEIDIAI_DEBUG)
+    add_definitions(-DKLEIDIAI_DEBUG=1)
+    add_definitions(-DKLEIDIAI_KERNEL=1)
+  endif()
+  if(KLEIDIAI_KERNEL)
+    add_definitions(-DKLEIDIAI_KERNEL=1)
   endif()
 endif()
 
--- a/onnxruntime/contrib_ops/cpu/quantization/dynamic_quantize_matmul.cc
+++ b/onnxruntime/contrib_ops/cpu/quantization/dynamic_quantize_matmul.cc
@@ -222,9 +222,13 @@ class DynamicQuantizeMatMul final : public MatMulIntegerToFloatBase {
       // Only handle the common case of a 2D weight matrix. Additional matrices
       // could be handled by stacking the packed buffers.
       b_shape_ = tensor.Shape();
-      // TO DO: handle b_shape_.NumDimensions() > 2 and all dimension values but the last two being 1.
-      if (!(b_shape_.NumDimensions() == 2 || (b_shape_.NumDimensions() == 3 && b_shape_[0] == 1))) {
-        can_use_dynamic_quant_mlas_ = false;
+      if (b_shape_.NumDimensions() > 2) {
+        for (size_t i = 0; i < (b_shape_.NumDimensions() - 2); ++i) {
+          if (b_shape_[i] != 1) {
+            can_use_dynamic_quant_mlas_ = false;
+            break;
+          }
+        }
       }
 
       // Can we use the mlas dynamic Q gemm interface supported with float output ?
@@ -302,8 +306,10 @@ class DynamicQuantizeMatMul final : public MatMulIntegerToFloatBase {
   int GetBIdx() const override { return IN_B; }
 
  private:
+  //Indicates when dynamic quantization is available so we can use the KleidiAI dynamic qgemm kernel
   bool can_use_dynamic_quant_mlas_{false};
 #if defined(USE_KLEIDIAI) && !defined(_MSC_VER)
+  //Indicates that the biases are a constant input and thus already quantized / packed
   bool dynamic_quant_mlas_bias_data_was_packed_{false};
 #endif
 };
--- a/onnxruntime/core/mlas/lib/kleidiai/mlasi_kleidiai.h
+++ b/onnxruntime/core/mlas/lib/kleidiai/mlasi_kleidiai.h
@@ -7,6 +7,7 @@
 #pragma once
 
 #include "mlasi.h"
+#include <iostream>
 
 // Fix to ensure compatibility with MSVC build
 #if defined(_MSC_VER)
@@ -14,6 +15,35 @@
 #else
   #define RESTRICT __restrict__
 #endif
+
+// Logging macros.
+#ifndef KLEIDIAI_DEBUG
+#define KLEIDIAI_DEBUG 0
+#endif
+#ifndef KLEIDIAI_KERNEL
+#define KLEIDIAI_KERNEL 0
+#endif
+
+// General logging. "tag" is expected to quality the type of message.
+#define KLEIDIAI_LOG(tag, msg) std::cout << "[KLEIDIAI " << tag << "]: " << msg << std::endl;
+
+#if KLEIDIAI_DEBUG
+    // General debug messages.
+    #define KLEIDIAI_DEBUG_LOG(msg) KLEIDIAI_LOG("DEBUG", msg)
+#else
+    #define KLEIDIAI_DEBUG_LOG(msg)
+#endif
+
+#if KLEIDIAI_KERNEL
+    // Messages specifically written before a call to kai_run.
+    // Note: In cases where a kernel is called in multiple threads, for example MlasTrySimpleParallel,
+    // the output order can be inconsistient. The solution is to set the intra-node thread size to 1.
+    // If using onnxruntime_perf_test this is done with "--x 1".
+    #define KLEIDIAI_KERNEL_LOG(kernel_name) KLEIDIAI_LOG("KERNEL", kernel_name)
+#else
+    #define KLEIDIAI_KERNEL_LOG(msg)
+#endif
+
 namespace ArmKleidiAI {
 // By default we should try for SME2 first before falling back to SME.
 inline const bool UseSME2 = MLAS_CPUIDINFO::GetCPUIDInfo().HasArm_SME2();
--- a/onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
+++ b/onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
@@ -45,6 +45,7 @@ Return Value:
 --*/
 {
     if (TransA != CblasNoTrans ||  N == 0  || K == 0) {
+        KLEIDIAI_DEBUG_LOG("MlasGemmPackBSize returning 0 size. N=" << N << " K=" << K);
         return 0;
     }
     //
@@ -61,9 +62,11 @@ Return Value:
                 bytes = kai_get_rhs_packed_size_rhs_pack_nxk_f32p2vlx1biasf32_f32_f32_sme(N, K);
                 break;
             default:
+                KLEIDIAI_DEBUG_LOG("MlasGemmPackBSize TransB is neither CblasNoTrans nor CblasTrans, returning 0.");
                 return 0;
         }
     } else {
+        KLEIDIAI_DEBUG_LOG("MlasGemmPackBSize TransA is CblasTrans, returning 0.");
         return 0;
     }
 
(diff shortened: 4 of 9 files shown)


orlmon01:

cmake/CMakeLists.txt
```diff
@@ -678,25 +678,14 @@ if(onnxruntime_USE_SVE)
...
-    if (NOT HAS_ARM64_I8MM)
-      message(FATAL_ERROR "The compiler doesn't support i8mm")
-    endif()
-  else()
-    message(STATUS "Skipping -march= checks on MSVC (not supported), assuming dotprod/i8mm support manually.")
-    set(HAS_ARM64_DOTPROD TRUE)
-    set(HAS_ARM64_I8MM TRUE)
+if(onnxruntime_USE_KLEIDIAI)
+  # If KLEIDIAI_DEBUG is enabled that implies both DEBUG and KERNEL messages.
+  if(KLEIDIAI_DEBUG)
```
edgchen1: are these supposed to be defined via `cmake -D...`? can they have the prefix `onnxruntime_`? and maybe a more specific name that indicates they are logging-related?

cmake/CMakeLists.txt
```diff
@@ -678,25 +678,14 @@ if(onnxruntime_USE_SVE)
...
+    add_definitions(-DKLEIDIAI_DEBUG=1)
```
edgchen1: can these definitions be scoped to the onnxruntime_mlas target? e.g., `target_compile_definitions(onnxruntime_mlas PRIVATE ...)`

onnxruntime/contrib_ops/cpu/quantization/dynamic_quantize_matmul.cc
```diff
@@ -302,8 +306,10 @@ class DynamicQuantizeMatMul final : public MatMulIntegerToFloatBase {
   int GetBIdx() const override { return IN_B; }
 
  private:
+  //Indicates when dynamic quantization is available so we can use the KleidiAI dynamic qgemm kernel
```
edgchen1: nit: would prefer to keep mention of KleidiAI within the USE_KLEIDIAI ifdefs if possible.

```suggestion
  // Indicates when MlasDynamicQGemmBatch() can be used
```

onnxruntime/core/mlas/lib/qgemm.cpp
```diff
@@ -406,7 +406,9 @@ Return Value:
     const size_t AlignedBytesRequired = (BytesRequired + BufferAlignment - 1) &
         ~(BufferAlignment - 1);
     //If this gemm B argument is used in a dynamically quantization gemm operation we can optimize for
-    //this use case. Concat both packed representations for later decision.
+    //this use case. Concat both packed representations for later decision. This allows for cases later
+    //where can_use_dynamic_mlas is false we still have the prepack at the cost of some memory otherwise
```
edgchen1: this comment about a private data member in another source file outside of MLAS seems out of place. it could be documented more at the call site if needed.

tools/ci_build/build.py
```diff
@@ -887,7 +887,6 @@ def generate_build_tree(
     # * Finally enable if platform.machine contains "arm64" and not a WebAssembly build. This should cover the following cases:
     #     *  Linux on Arm
     #     *  MacOs (case must be ignored)
```
edgchen1: it doesn't seem like this TODO was addressed

orlmon01: This TODO wasn't necessary. build.py is the correct place to determine if Onnxruntime_USE_KLEIDIAI should be ON

edgchen1: I think it would be better in the CMake logic. I have a separate PR for that.
https://github.com/microsoft/onnxruntime/pull/26178

cmake/CMakeLists.txt
orlmon01: Yes, they are very explicitly debug only and should never be used outside of that as it will murder your performance. I'll update the prefixes and names to reflect this.

hariharans29:
Can you please resolve the conflicts ?

cmake/CMakeLists.txt
```diff
@@ -713,6 +713,32 @@ else()
   endif()
 endif()
 
+if(onnxruntime_USE_SVE)
```
edgchen1: FYI, this code got moved earlier by another PR:

https://github.com/orlmon01/onnxruntime/blob/6f14c5fc0e3210fd4646e4bbda0eadee3413a470/cmake/CMakeLists.txt#L510-L523

I think this copy can be deleted.

onnxruntime/contrib_ops/cpu/quantization/dynamic_quantize_matmul.cc
```diff
@@ -222,9 +222,13 @@ class DynamicQuantizeMatMul final : public MatMulIntegerToFloatBase {
       // Only handle the common case of a 2D weight matrix. Additional matrices
       // could be handled by stacking the packed buffers.
       b_shape_ = tensor.Shape();
-      // TO DO: handle b_shape_.NumDimensions() > 2 and all dimension values but the last two being 1.
-      if (!(b_shape_.NumDimensions() == 2 || (b_shape_.NumDimensions() == 3 && b_shape_[0] == 1))) {
-        can_use_dynamic_quant_mlas_ = false;
+      if (b_shape_.NumDimensions() > 2) {
```
edgchen1: thanks for making the dimension check more generic.

the old version set `can_use_dynamic_quant_mlas_` to false if `b_shape_.NumDimensions() < 2` but the new version does not. do we want to allow 0 or 1 dimensions?

orlmon01: Good spot, I don't think it will cause any issues. Or at least it wasn't failing any tests. But we also only want to use this for 2d cases, and cases where the NumDimensions is > 2 but the it's still effectively 2d. Adding an else to be safe. :)

hariharans29: Sorry if I am mis-reading - shouldn't the if condition read >= 2 though (i.e.) `can_use_dynamic_quant_mlas_` needs to remain true when rank is 2 ?

cmake/CMakeLists.txt
hariharans29: Just thinking out loud - should these be made available to targets related to MLAS as well - https://github.com/microsoft/onnxruntime/pull/26225 ? (i.e.) should logging be exposed to benchmarking and unit-test projects too ?

onnxruntime/contrib_ops/cpu/quantization/dynamic_quantize_matmul.cc
hariharans29: Thanks for updating the condition. Don't we want to keep the else to keep the old behavior ?

cmake/CMakeLists.txt
orlmon01: That's a good question. I'm leaning towards no. Definitely not for benchmarking as enabling / using these will murder performance. And probably not for unit-test projects as they can be added in explicitly when needed for debugging. They're just intended for when you need more information on why something is failing / not working.

hariharans29:
LGTM. @edgchen1 - Do you have any additional comments ?

hariharans29:
The build failures seem unrelated to this PR. Might be something else going on.

tools/ci_build/build.py
edgchen1: this comment is not applicable now and it was removed in main. looks like it may have been added back in a merge.

onnxruntime/core/mlas/lib/qgemm.cpp
hariharans29: @orlmon01 - Sorry, is this comment still pending ?

orlmon01: Apologies, I had tweaked it but it must have gotten back in with all of the various merges / updates. Fixed :)

orlmon01:
How does it look? Any chance of this going in today? :)

hariharans29:
> How does it look? Any chance of this going in today? :)

Thanks for the updates. I will check again with @edgchen1 and @devang-ml  on this.

hariharans29 (approved):

onnxruntime/core/mlas/lib/kleidiai/mlasi_kleidiai.h
```diff
@@ -7,13 +7,43 @@
...
+    #define KLEIDIAI_DEBUG_LOG(msg) KLEIDIAI_LOG("DEBUG", msg)
+#else
+    #define KLEIDIAI_DEBUG_LOG(msg)
+#endif
+
+#if onnxruntime_KLEIDIAI_KERNEL_LOGGING
+    // Messages specifically written before a call to kai_run.
+    // Note: In cases where a kernel is called in multiple threads, for example MlasTrySimpleParallel,
+    // the output order can be inconsistient. The solution is to set the intra-node thread size to 1.
+    // If using onnxruntime_perf_test this is done with "--x 1".
```
edgchen1, line 46: can this interleaved output be avoided if we only write to `cout` once?

```c++
std::ostringstream s;
s << ...;

std::cout << s.str();
```

onnxruntime/core/mlas/lib/kleidiai/mlasi_kleidiai.h
```diff
@@ -7,13 +7,43 @@
...
+// Logging macros.
+#ifndef onnxruntime_KLEIDIAI_DEBUG_LOGGING
+#define onnxruntime_KLEIDIAI_DEBUG_LOGGING 0
+#endif
+#ifndef onnxruntime_KLEIDIAI_KERNEL_LOGGING
+#define onnxruntime_KLEIDIAI_KERNEL_LOGGING 0
+#endif
+
+// General logging. "tag" is expected to quality the type of message.
+#define KLEIDIAI_LOG(tag, msg) std::cout << "[KLEIDIAI " << tag << "]: " << msg << std::endl;
```
edgchen1: nit: can the macro code be within its own scope and require callers to follow it with a semi-colon?

```c++
#define LOG(...) \
  do { \
    std::cout << ... \
  } while(false)
```

edgchen1: if we're already using macros, it might be nice to include the file and line number in the output too

cmake/CMakeLists.txt
```diff
@@ -716,6 +716,17 @@ else()
   endif()
 endif()
 
+if(onnxruntime_USE_KLEIDIAI)
+  # If KLEIDIAI_DEBUG is enabled that implies both DEBUG and KERNEL messages.
+  if(onnxruntime_KLEIDIAI_DEBUG_LOGGING)
+    target_compile_definitions(onnxruntime_mlas PRIVATE -DKLEIDIAI_DEBUG=1)
```
edgchen1: nit: don't need to specify the leading "-D"
```suggestion
    target_compile_definitions(onnxruntime_mlas PRIVATE KLEIDIAI_DEBUG=1)
```

cmake/CMakeLists.txt
```diff
@@ -716,6 +716,17 @@ else()
...
+    target_compile_definitions(onnxruntime_mlas PRIVATE -DKLEIDIAI_KERNEL=1)
+  endif()
+  if(onnxruntime_KLEIDIAI_KERNEL_LOGGING)
+    target_compile_definitions(onnxruntime_mlas PRIVATE -DKLEIDIAI_KERNEL=1)
```
edgchen1: I believe this will be run before the onnxruntime_mlas target has been created in onnxruntime_mlas.cmake which is included below.

maybe it would be simpler to move this code to onnxruntime_mlas.cmake.

onnxruntime/core/mlas/lib/kleidiai/mlasi_kleidiai.h
```diff
@@ -7,13 +7,43 @@
...
+
+// Logging macros.
+#ifndef onnxruntime_KLEIDIAI_DEBUG_LOGGING
+#define onnxruntime_KLEIDIAI_DEBUG_LOGGING 0
+#endif
+#ifndef onnxruntime_KLEIDIAI_KERNEL_LOGGING
+#define onnxruntime_KLEIDIAI_KERNEL_LOGGING 0
+#endif
+
+// General logging. "tag" is expected to quality the type of message.
```
edgchen1:
```suggestion
// General logging. "tag" is expected to qualify the type of message.
```

onnxruntime/core/mlas/lib/kleidiai/mlasi_kleidiai.h
edgchen1: can `KLEIDIAI_LOG` only be defined if logging is enabled?

onnxruntime/core/mlas/lib/kleidiai/mlasi_kleidiai.h
orlmon01: I'm not sure. If I understand what you're suggesting correctly, you mean to have the KLEIDIAI_KERNEL_LOG just append the messages on to the end of a stringstream rather than logging them all as they happen. And then print them all out together? I think the resulting message would still be interleaved, but as a block instead of seperate messages. 
In any case, we're not really worried about the order 99% of the time as it's just for debugging and trying to see what is being called / used. For cases where the order matters you can specify the thread size to be 1. These messages should never be used when testing for performance anyways.

cmake/onnxruntime_mlas.cmake
```diff
@@ -5,6 +5,19 @@ set(MLAS_ROOT ${ONNXRUNTIME_ROOT}/core/mlas)
 set(MLAS_SRC_DIR ${MLAS_ROOT}/lib)
 set(MLAS_INC_DIR ${MLAS_ROOT}/inc)
 
+
+if(onnxruntime_USE_KLEIDIAI)
+  # If KLEIDIAI_DEBUG is enabled that implies both DEBUG and KERNEL messages.
+  if(onnxruntime_KLEIDIAI_DEBUG_LOGGING)
```
edgchen1: nit: perhaps these could go into `setup_kleidiai()` so that we can consolidate the KleidiAI-specific code.

onnxruntime/core/mlas/lib/kleidiai/mlasi_kleidiai.h
edgchen1: > If I understand what you're suggesting correctly, you mean to have the KLEIDIAI_KERNEL_LOG just append the messages on to the end of a stringstream rather than logging them all as they happen. And then print them all out together? I think the resulting message would still be interleaved, but as a block instead of seperate messages.

yes, that's what I'm suggesting. I think it would be more legible for complete log lines from different threads to be interleaved vs. fragments of them.

but if this logging is only intended to be used in a single threaded context, I suppose it doesn't matter.

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
```diff
@@ -70,9 +71,11 @@ Return Value:
                 bytes = kai_get_rhs_packed_size_rhs_pack_nxk_f32p2vlx1biasf32_f32_f32_sme(N, K);
                 break;
             default:
+                KLEIDIAI_DEBUG_LOG("MlasGemmPackBSize TransB is neither CblasNoTrans nor CblasTrans, returning 0.");
                 return 0;
         }
     } else {
+        KLEIDIAI_DEBUG_LOG("MlasGemmPackBSize TransA is CblasTrans, returning 0.");
```
edgchen1: wouldn't the check for `TransA != CblasNoTrans` above already have caused the function to return early?

https://github.com/microsoft/onnxruntime/blob/edfb6f599e8ac7ddaf672037d95404eb0a473c96/onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp#L57-L59

seems like this could be simplified by removing the second `if (TransA == CblasNoTrans)` check entirely.

onnxruntime/core/mlas/lib/kleidiai/mlasi_kleidiai.h
```diff
@@ -6,14 +6,46 @@
...
+#ifndef onnxruntime_KLEIDIAI_DEBUG_LOGGING
+#define onnxruntime_KLEIDIAI_DEBUG_LOGGING 0
+#endif
+#ifndef onnxruntime_KLEIDIAI_KERNEL_LOGGING
+#define onnxruntime_KLEIDIAI_KERNEL_LOGGING 0
+#endif
+
+// General logging. "tag" is expected to qualify the type of message.
+#if onnxruntime_KLEIDIAI_DEBUG_LOGGING
+#define KLEIDIAI_LOG(tag, msg) \
```
edgchen1, line 28: should `KLEIDIAI_LOG` also be defined if `onnxruntime_KLEIDIAI_KERNEL_LOGGING` is non-zero?

orlmon01: Good spot, that was silly of me. Strange that it still seemed to work locally. I have moved it back out and added an if statement so it should be defined for either debug or kernel logging.

hariharans29 (approved):

onnxruntime/core/mlas/lib/kleidiai/mlasi_kleidiai.h
```diff
@@ -6,14 +6,49 @@
...
+#define onnxruntime_KLEIDIAI_DEBUG_LOGGING 0
+#endif
+#ifndef onnxruntime_KLEIDIAI_KERNEL_LOGGING
+#define onnxruntime_KLEIDIAI_KERNEL_LOGGING 0
+#endif
+
+#if onnxruntime_KLEIDIAI_DEBUG_LOGGING || onnxruntime_KLEIDIAI_KERNEL_LOGGING
+#define KLEIDIAI_LOG(tag, msg) \
+    do { \
+        std::cout << __FILE__ << __LINE__ << "[KLEIDIAI " << tag << "]: " << msg << std::endl; \
```
edgchen1: nit: IMO, it would be easier to spot these log lines visually if `KLEIDIAI` was at the start of the line. also, it might be good to have some delimiters like `__FILE__ << ':' << __LINE__`

orlmon01: Done :)

PR #26148, opened 2025-09-24
[QNN-EP] Fix logic flow bug

quic-calvnguy:
### Description
Fix logic flow bug where rpc polling interval is set to 9999 when perf performance is NOT burst. The interval should be set to 9999 when the perf performance is burst


apsonawane:

PR #26149, opened 2025-09-24
Regenerate test model with ONNX IR < 12

adrianlizarraga:
### Description
Regenerates the test model used in [this unit test](https://github.com/microsoft/onnxruntime/blob/35dcab5088118117acc6086c9b6dd6dd92c7060f/onnxruntime/test/shared_lib/test_inference.cc#L497-L506) so that it uses an ONNX IR version compatible with ONNX 1.18.0.

### Motivation and Context
The main branch is using ONNX 1.19. However, this unit test also needs to pass in the `rel-1.23.1` branch, which is still using ONNX 1.18.0. So, by downgrading the model's IR version, the unit test can run in both branches.


onnxruntime/test/testdata/input_propagated_to_output.py
```diff
@@ -0,0 +1,87 @@
+"""
```
adrianlizarraga, line 1: Note: I tried to use the [onnx2py](https://github.com/microsoft/onnxconverter-common/blob/master/onnxconverter_common/onnx2py.py) tool to create this script from the model, but it threw some errors. So, I created this script manually.

adrianlizarraga:
> We do not need any .npy files, just remove.

Both python scripts will fail to run if the .npy files are missing. If we remove them, we should embed the data in the python file.

yuslepukhin (approved):
:shipit:

snnn:

PR #26152, opened 2025-09-24
Normalize to positive axis for comparison on GEMM

qti-yuduo:
GEMM has 2D input, when given `axis=-1` in DequantizeLinear node attributes. The check `axis == expected_axis` would fail. However, it sematically the same (-1 and 1 for 2D GEMM's input). Normalize the check so we can still perform the quantization for this case.

### DIFF
--- a/onnxruntime/core/optimizer/qdq_transformer/weight_bias_quantization.cc
+++ b/onnxruntime/core/optimizer/qdq_transformer/weight_bias_quantization.cc
@@ -128,8 +128,15 @@ Status WeightBiasQuantization::ApplyImpl(Graph& graph, bool& modified, int graph
 
       int64_t axis = 1;
       if (auto axis_iter = dq_attrs.find("axis"); axis_iter != dq_attrs.end()) {
+        axis = axis_iter->second.i();
         const ONNX_NAMESPACE::TensorShapeProto* weight_shape = weight_arg->Shape();
-        axis = HandleNegativeAxis(axis_iter->second.i(), weight_shape->dim_size());
+        if (!weight_shape && dq_1 && dq_1->InputDefs()[0]) {
+          weight_shape = dq_1->InputDefs()[0]->Shape();
+        }
+        if (axis < 0 && !weight_shape) {
+          continue;
+        }
+        axis = HandleNegativeAxis(axis, weight_shape->dim_size());
       }
 
       int64_t expected_axis = 0;
--- a/onnxruntime/core/optimizer/qdq_transformer/weight_bias_quantization.cc
+++ b/onnxruntime/core/optimizer/qdq_transformer/weight_bias_quantization.cc
@@ -129,8 +129,7 @@ Status WeightBiasQuantization::ApplyImpl(Graph& graph, bool& modified, int graph
       int64_t axis = 1;
       if (auto axis_iter = dq_attrs.find("axis"); axis_iter != dq_attrs.end()) {
         const ONNX_NAMESPACE::TensorShapeProto* weight_shape = weight_arg->Shape();
-        int64_t weight_rank = weight_shape->dim_size();
-        axis = HandleNegativeAxis(axis_iter->second.i(), weight_rank);
+        axis = HandleNegativeAxis(axis_iter->second.i(), weight_shape->dim_size());
       }
 
       int64_t expected_axis = 0;
--- a/onnxruntime/core/optimizer/qdq_transformer/weight_bias_quantization.cc
+++ b/onnxruntime/core/optimizer/qdq_transformer/weight_bias_quantization.cc
@@ -138,6 +138,7 @@ Status WeightBiasQuantization::ApplyImpl(Graph& graph, bool& modified, int graph
           transB = trans_b_iter->second.i();
         }
         expected_axis = transB == 0 ? 1 : 0;
+        axis = (axis == -1) ? 1 : axis;  // normalize for comparison
       }
 
       if (axis != expected_axis) {
--- a/onnxruntime/core/optimizer/qdq_transformer/weight_bias_quantization.cc
+++ b/onnxruntime/core/optimizer/qdq_transformer/weight_bias_quantization.cc
@@ -4,6 +4,7 @@
 #include "core/optimizer/qdq_transformer/weight_bias_quantization.h"
 
 #include "core/common/common.h"
+#include "core/providers/common.h"
 #include "core/util/qmath.h"
 #include "core/graph/graph_utils.h"
 #include "core/graph/graph_viewer.h"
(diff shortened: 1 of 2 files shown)


onnxruntime/core/optimizer/qdq_transformer/weight_bias_quantization.cc
```diff
@@ -138,6 +138,7 @@ Status WeightBiasQuantization::ApplyImpl(Graph& graph, bool& modified, int graph
           transB = trans_b_iter->second.i();
         }
         expected_axis = transB == 0 ? 1 : 0;
+        axis = (axis == -1) ? 1 : axis;  // normalize for comparison
```
edgchen1: can the normalization be done more generally like this?

https://github.com/microsoft/onnxruntime/blob/88f2652dee465faa485cf68cb39c4489f28945bb/onnxruntime/core/providers/common.h#L24-L35

maybe it can also move out of this Gemm-only block.

qti-yuduo: Thanks for the reference to the utility. I did this simple way due to the transform was only designed to handle GEMM (always 2D) and Conv would always have axis = 0. So make it general sounds like overkill due to the simplicity of the use case. 

I'll switch to the utility.

qti-yuduo: Done.

onnxruntime/core/optimizer/qdq_transformer/weight_bias_quantization.cc
```diff
@@ -127,7 +128,9 @@ Status WeightBiasQuantization::ApplyImpl(Graph& graph, bool& modified, int graph
 
       int64_t axis = 1;
       if (auto axis_iter = dq_attrs.find("axis"); axis_iter != dq_attrs.end()) {
-        axis = axis_iter->second.i();
+        const ONNX_NAMESPACE::TensorShapeProto* weight_shape = weight_arg->Shape();
+        int64_t weight_rank = narrow<int64_t>(weight_shape->dim_size());
```
edgchen1: I believe protobuf `x_size()` returns `int`, right? that doesn't need a call to `narrow` to convert it to `int64_t`.
```suggestion
        int64_t weight_rank = weight_shape->dim_size();
```

also, is it possible for `weight_shape` to be nullptr?

qti-yuduo: Weight is check as initializer and early returned when not.

edgchen1: > Weight is check as initializer and early returned when not.

I'm not very familiar with this code and may be missing something. but in this branch:
https://github.com/microsoft/onnxruntime/blob/b160e8c6b4f0d34d42b85146e1d5973a5d6224b1/onnxruntime/core/optimizer/qdq_transformer/weight_bias_quantization.cc#L113

doesn't `dq_1 != nullptr` imply that `weight_arg` is not an initializer?

qti-yuduo: The check I was referring to was on line 90-93: 

```
    } else if (!graph_utils::IsInitializer(graph, weight_arg->Name(), true) ||
               !graph.GetInitializedTensor(weight_arg->Name(), weight_proto) ||
               weight_proto->data_type() != ONNX_NAMESPACE::TensorProto_DataType_FLOAT) {
      // Support float32 weight initializer only for now.
      continue;
    }
```

This should guarantee we're having concrete shape for weighs?

edgchen1: I see - I think that should guarantee `weight_arg` is a constant initializer, but only if that `else if` branch is run. in the case where `dq_1` gets assigned (is not nullptr), that `else if` branch is not run.

https://github.com/microsoft/onnxruntime/blob/b160e8c6b4f0d34d42b85146e1d5973a5d6224b1/onnxruntime/core/optimizer/qdq_transformer/weight_bias_quantization.cc#L85-L94

qti-yuduo: Yup, missed that. I'll update to make it more conservative and skip when impossible to fetch weight shape here.

qti-yuduo: Updated. Lemme know anything else.

onnxruntime/core/optimizer/qdq_transformer/weight_bias_quantization.cc
```diff
@@ -128,6 +129,14 @@ Status WeightBiasQuantization::ApplyImpl(Graph& graph, bool& modified, int graph
       int64_t axis = 1;
       if (auto axis_iter = dq_attrs.find("axis"); axis_iter != dq_attrs.end()) {
         axis = axis_iter->second.i();
+        const ONNX_NAMESPACE::TensorShapeProto* weight_shape = weight_arg->Shape();
+        if (!weight_shape && dq_1 && dq_1->InputDefs()[0]) {
```
edgchen1: nit: it is redundant to check for `dq_1 != nullptr` here as that ensured earlier.

https://github.com/CodeLinaro/onnxruntime/blob/01112314cc770c15db297b8a37b31ea9110dcb04/onnxruntime/core/optimizer/qdq_transformer/weight_bias_quantization.cc#L114

onnxruntime/core/optimizer/qdq_transformer/weight_bias_quantization.cc
```diff
@@ -128,6 +129,14 @@ Status WeightBiasQuantization::ApplyImpl(Graph& graph, bool& modified, int graph
...
+          weight_shape = dq_1->InputDefs()[0]->Shape();
+        }
+        if (axis < 0 && !weight_shape) {
+          continue;
+        }
+        axis = HandleNegativeAxis(axis, weight_shape->dim_size());
```
edgchen1, line 139: would it be possible to add a test for the additional Gemm case that this change enables?

qti-yuduo: Added

edgchen1: thanks!

qti-yuduo:
@edgchen1 could you help to check the CI runner status here? Thanks!

edgchen1:
@qti-yuduo Have you agreed to the CLA? It looks like the "license/cla" check is pending.

edgchen1 (approved):

qti-yuduo:
> @qti-yuduo Have you agreed to the CLA? It looks like the "license/cla" check is pending.

I think so? Unless it's changed and I need to do it again. I've commited changes in the past: https://github.com/microsoft/onnxruntime/pulls?q=is%3Apr+author%3Aqti-yuduo+is%3Aclosed

edgchen1:
> > @qti-yuduo Have you agreed to the CLA? It looks like the "license/cla" check is pending.
> 
> I think so? Unless it's changed and I need to do it again. I've commited changes in the past: https://github.com/microsoft/onnxruntime/pulls?q=is%3Apr+author%3Aqti-yuduo+is%3Aclosed

Ok, let me try to close and re-open the PR.

PR #26155, opened 2025-09-25
Add basic detection of whether a GPU is discrete on Linux

edgchen1:
### Description

Update Linux device discovery to add rudimentary detection of whether a GPU is discrete and include that in the device metadata.

### Motivation and Context

Include more potentially useful metadata.

### DIFF
--- a/onnxruntime/core/platform/linux/device_discovery.cc
+++ b/onnxruntime/core/platform/linux/device_discovery.cc
@@ -6,6 +6,7 @@
 #include <filesystem>
 #include <fstream>
 #include <iterator>
+#include <optional>
 #include <string_view>
 
 #include "core/common/common.h"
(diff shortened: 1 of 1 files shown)


snnn:
If we can use Vulcan, we can get the information from VkPhysicalDeviceProperties, but, in more general case I have no good idea. 
However, if we focus on Android or macOS, I think the system should be able to provide something.

edgchen1:
> How do you plan to use the 'Discrete' metadata? This will help in understanding the impact of the current assumption.

@skottmckay would you mind elaborating on how this would be used?

edgchen1:
> Not all NVIDIA GPUs are discrete; for example, some Tegra SoCs have integrated GPUs.

good to know. I was misled by an AI suggestion...

skottmckay:
Used in auto EP selection to prefer discrete GPU over integrated.
Will be used for WebGPU to differentiate between dGPU and iGPU when both are present as there's a different setting for each (high perf vs low power) and no way (afaik) to directly select a device.

snnn:
I think WebGPU intentionally hided the hardware detail to protect users' privacy, but we also need to run the same code outside of browsers  so the privacy concern does not hold.

snnn (approved):

PR #26156, opened 2025-09-25
[EP ABI] Check if nodes specified in GetCapability() have already been assigned

adrianlizarraga:
### Description
Fixes segfault in `PluginExecutionProvider::GetCapability()` when the underlying `OrtEp` tries to claim nodes that have already been assigned to another EP.

### Motivation and Context
Should log a warning (instead of crashing or throwing an exception) when a plugin EP tries to claim a node that is already assigned to another EP.

### DIFF
--- a/onnxruntime/core/session/plugin_ep/ep_plugin_provider_interfaces.cc
+++ b/onnxruntime/core/session/plugin_ep/ep_plugin_provider_interfaces.cc
@@ -5,6 +5,7 @@
 
 #include <memory>
 #include <string>
+#include <sstream>
 #include <unordered_set>
 #include <utility>
 #include <vector>
@@ -158,9 +159,11 @@ PluginExecutionProvider::GetCapability(const onnxruntime::GraphViewer& graph_vie
   ORT_UNUSED_PARAMETER(resource_accountant);       // TODO: Add support? Not used by prioritized EPs
   ORT_UNUSED_PARAMETER(kernel_lookup);             // TODO: Add support? Not used by prioritized EPs, so probably not needed?
 
+  const logging::Logger& logger = GetLogger() != nullptr ? *GetLogger() : logging::LoggingManager::DefaultLogger();
+
   std::unique_ptr<EpGraph> ep_graph = nullptr;
   if (Status status = EpGraph::Create(graph_viewer, ep_graph); !status.IsOK()) {
-    LOGS_DEFAULT(ERROR) << "Failed to create OrtGraph: " << status.ToString();
+    LOGS(logger, ERROR) << "Failed to create OrtGraph for " << Type() << ": " << status.ToString();
     return {};
   }
 
@@ -168,7 +171,7 @@ PluginExecutionProvider::GetCapability(const onnxruntime::GraphViewer& graph_vie
   Status status = ToStatusAndRelease(ort_ep_->GetCapability(ort_ep_.get(), ep_graph->ToExternal(), &api_graph_support_info));
 
   if (!status.IsOK()) {
-    LOGS_DEFAULT(ERROR) << "OrtEp::GetCapability() failed with error: " << status.ToString();
+    LOGS(logger, ERROR) << "OrtEp::GetCapability() for " << Type() << " failed with error: " << status.ToString();
     return {};
   }
 
@@ -188,6 +191,14 @@ PluginExecutionProvider::GetCapability(const onnxruntime::GraphViewer& graph_vie
       indexed_sub_graph->nodes.push_back(node_grouping.nodes[0]->GetInternalNode().Index());
       result.push_back(std::make_unique<ComputeCapability>(std::move(indexed_sub_graph)));
     } else if (node_grouping.kind == OrtEpGraphSupportInfo::NodeGroupingKind::kFusedNode) {
+      if (node_grouping.nodes.empty()) {
+        // The EpGraphSupportInfo_AddNodesToFuse() C API should already return an error if the EP tries to provide
+        // an empty array of nodes from OrtEp::GetCapability(). However, we check here too just in case this changes.
+        LOGS(logger, WARNING) << "OrtEp::GetCapability() for " << Type() << " set an empty array of nodes "
+                              << "when specifying supported nodes.";
+        return {};
+      }
+
       std::unordered_set<const Node*> node_set;
       node_set.reserve(node_grouping.nodes.size());
 
@@ -207,9 +218,33 @@ PluginExecutionProvider::GetCapability(const onnxruntime::GraphViewer& graph_vie
           this->Type(), this->Type(), /*node_unit_map*/ nullptr,
           node_grouping.fusion_options.drop_constant_initializers);
 
+      // Check if utils::CreateSupportedPartitions returned zero results.
+      // Happens if nodes have already been assigned to another EP.
+      if (capabilities.empty()) {
+        const std::string& ep_type = Type();
+        std::ostringstream oss;
+
+        oss << "OrtEp::GetCapability() specified nodes that cannot be assigned to " << ep_type << ". ";
+        auto node_iter = std::find_if(node_grouping.nodes.begin(), node_grouping.nodes.end(),
+                                      [&ep_type](const EpNode* node) -> bool {
+                                        return node->GetInternalNode().GetExecutionProviderType() != ep_type;
+                                      });
+
+        if (node_iter != node_grouping.nodes.end()) {
+          const Node& node_assigned_to_other_ep = (*node_iter)->GetInternalNode();
+          oss << "Found one or more nodes that were already assigned to a different EP named '"
+              << node_assigned_to_other_ep.GetExecutionProviderType() << "'. Ex: "
+              << node_assigned_to_other_ep.OpType() << " node with name '"
+              << node_assigned_to_other_ep.Name() << "'.";
+        }
+
+        LOGS(logger, WARNING) << oss.str();
+        return {};
+      }
+
       if (capabilities.size() > 1) {
-        LOGS_DEFAULT(ERROR) << "OrtEp::GetCapability() set nodes that cannot be fused together. "
-                            << "Please ensure that the nodes provided to EpGraphSupportInfo_AddFusedNodes() do not "
+        LOGS(logger, ERROR) << "OrtEp::GetCapability() for " << Type() << " set nodes that cannot be fused together. "
+                            << "Please ensure that the nodes provided to EpGraphSupportInfo_AddNodesToFuse() do not "
                             << "have an unsupported node in any path between two of the supported nodes.";
         return {};
       }
@@ -227,7 +262,7 @@ PluginExecutionProvider::GetCapability(const onnxruntime::GraphViewer& graph_vie
 
       result.push_back(std::move(capabilities[0]));
     } else {
-      LOGS_DEFAULT(ERROR) << "PluginExecutionProvider::GetCapability() has invalid NodeGroupingKind: "
+      LOGS(logger, ERROR) << "PluginExecutionProvider::GetCapability() has invalid NodeGroupingKind: "
                           << static_cast<int>(node_grouping.kind);
       return {};
     }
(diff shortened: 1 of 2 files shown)


onnxruntime/core/session/plugin_ep/ep_plugin_provider_interfaces.cc
```diff
@@ -117,6 +120,16 @@ static OrtDevice GetOrtDeviceForPluginEp(gsl::span<const OrtEpDevice* const> ep_
   return device_memory_info != nullptr ? device_memory_info->device : OrtDevice();
 }
 
+static const Node* FindFirstNodeAssignedToOtherEP(const std::string& ep_type,
+                                                  gsl::span<const EpNode* const> ep_nodes) {
+  auto node_iter = std::find_if(ep_nodes.begin(), ep_nodes.end(),
+                                [&ep_type](const EpNode* node) -> bool {
+                                  return node->GetInternalNode().GetExecutionProviderType() != ep_type;
```
edgchen1: should it also check that `node->GetInternalNode().GetExecutionProviderType()` is non-empty (`node` is assigned to an EP)?

```suggestion
                                  const auto& node_ep_type = node->GetInternalNode().GetExecutionProviderType();
                                  return !node_ep_type.empty() && node_ep_type != ep_type;
```

onnxruntime/test/framework/ep_plugin_provider_test.cc
```diff
@@ -317,4 +334,121 @@ TEST(PluginExecutionProviderTest, InferOrtDeviceFromDeviceMemoryInfo) {
...
+
+  // Load a model and forcibly assign only the first Add node to another EP named 'OtherEp'.
+  // Other nodes are unassigned and should be taken by the test plugin EP.
+  std::shared_ptr<Model> model;
+  Status status = Model::Load(ORT_TSTR("testdata/add_mul_add.onnx"), model, nullptr, DefaultLoggingManager().DefaultLogger());
+  ASSERT_TRUE(status.IsOK());
+
+  Graph& graph = model->MainGraph();
+  for (Node& node : graph.Nodes()) {
+    if (node.Name() == "add_0") {
```
edgchen1: maybe assign the second add node to "OtherEp" to also test handling of the first one being unassigned.

onnxruntime/test/framework/ep_plugin_provider_test.cc
```diff
@@ -317,4 +334,121 @@ TEST(PluginExecutionProviderTest, InferOrtDeviceFromDeviceMemoryInfo) {
...
+// nodes that are already assigned to another EP.
+TEST(PluginExecutionProviderTest, GetCapability_ClaimSomeNodesAssignedToOtherEP) {
+  auto [ep, ort_ep] = test_plugin_ep::MakeTestOrtEp();
+  ort_ep->GetCapability = GetCapabilityTakeAllNodes;
+
+  // Load a model and forcibly assign only the first Add node to another EP named 'OtherEp'.
+  // Other nodes are unassigned and should be taken by the test plugin EP.
+  std::shared_ptr<Model> model;
+  Status status = Model::Load(ORT_TSTR("testdata/add_mul_add.onnx"), model, nullptr, DefaultLoggingManager().DefaultLogger());
+  ASSERT_TRUE(status.IsOK());
```
edgchen1: nit: ASSERT_STATUS_OK to also output any error message.

onnxruntime/core/session/plugin_ep/ep_plugin_provider_interfaces.cc
```diff
@@ -188,6 +218,14 @@ PluginExecutionProvider::GetCapability(const onnxruntime::GraphViewer& graph_vie
       indexed_sub_graph->nodes.push_back(node_grouping.nodes[0]->GetInternalNode().Index());
       result.push_back(std::make_unique<ComputeCapability>(std::move(indexed_sub_graph)));
     } else if (node_grouping.kind == OrtEpGraphSupportInfo::NodeGroupingKind::kFusedNode) {
+      if (node_grouping.nodes.empty()) {
+        // The EpGraphSupportInfo_AddNodesToFuse() C API should already return an error if the EP tries to provide
+        // an empty array of nodes from OrtEp::GetCapability(). However, we check here too just in case this changes.
+        LOGS(logger, WARNING) << "OrtEp::GetCapability() for " << Type() << " set an empty array of nodes "
```
chilo-ms: nit: Should this check and warning also apply to the `NodeGroupingKind::kSingleAssignedNode` case above?

onnxruntime/test/framework/ep_plugin_provider_test.cc
adrianlizarraga: I added more test cases (like this one).

onnxruntime/test/framework/ep_plugin_provider_test.cc
adrianlizarraga: Updated, thanks

onnxruntime/core/session/plugin_ep/ep_plugin_provider_interfaces.cc
adrianlizarraga: Added. Thanks.

onnxruntime/core/session/plugin_ep/ep_plugin_provider_interfaces.cc
adrianlizarraga: good catch, thanks

onnxruntime/core/session/plugin_ep/ep_plugin_provider_interfaces.cc
```diff
@@ -183,11 +213,27 @@ PluginExecutionProvider::GetCapability(const onnxruntime::GraphViewer& graph_vie
   // Create ComputeCapability instances from OrtEpGraphSupportInfo::NodeGrouping instances.
   for (const OrtEpGraphSupportInfo::NodeGrouping& node_grouping : api_graph_support_info.node_groupings) {
     if (node_grouping.kind == OrtEpGraphSupportInfo::NodeGroupingKind::kSingleAssignedNode) {
+      if (node_grouping.nodes.empty()) {
```
edgchen1: hm, it _seems_ safe to assume that calling `EpGraphSupportInfo_AddSingleNode` will add one node. if we do want to check, checking for `node_grouping.nodes.size() == 1` would be more exact.

however, perhaps we also want to check that the EP didn't provide a node that was already assigned to another EP here.

onnxruntime/test/framework/ep_plugin_provider_test.cc
```diff
@@ -317,4 +334,129 @@ TEST(PluginExecutionProviderTest, InferOrtDeviceFromDeviceMemoryInfo) {
 #endif  // !defined(ORT_NO_EXCEPTIONS)
 }
 
+static void LoadModelAndAssignNodesToEp(const ORTCHAR_T* model_path,
+                                        const char* ep_name,
+                                        std::unordered_set<std::string> ep_node_names,
+                                        /*out*/ std::shared_ptr<Model>& model) {
+  ASSERT_STATUS_OK(Model::Load(model_path, model, nullptr,
```
edgchen1, line 341: note that a failure here will only exit early from this helper function. might want to use ASSERT_NO_FATAL_FAILURE at the call site.

https://google.github.io/googletest/advanced.html#asserting-on-subroutines

onnxruntime/core/session/plugin_ep/ep_plugin_provider_interfaces.cc
adrianlizarraga: Good call. Added both checks (size and assigned ep) and a unit test.

onnxruntime/test/framework/ep_plugin_provider_test.cc
adrianlizarraga: Thanks!

onnxruntime/core/session/plugin_ep/ep_plugin_provider_interfaces.cc
```diff
@@ -207,27 +261,34 @@ PluginExecutionProvider::GetCapability(const onnxruntime::GraphViewer& graph_vie
...
+        log_unsupported_node_info(node_grouping.nodes);
+        continue;
+      }
+
       if (capabilities.size() > 1) {
-        LOGS_DEFAULT(ERROR) << "OrtEp::GetCapability() set nodes that cannot be fused together. "
-                            << "Please ensure that the nodes provided to EpGraphSupportInfo_AddFusedNodes() do not "
+        LOGS(logger, ERROR) << "OrtEp::GetCapability() for " << Type() << " set nodes that cannot be fused together. "
+                            << "Please ensure that the nodes provided to EpGraphSupportInfo_AddNodesToFuse() do not "
                             << "have an unsupported node in any path between two of the supported nodes.";
```
edgchen1, line 254: I think the `capabilities.size() > 1` case could also happen with a node that is assigned to another EP. e.g., if such a node is in between two supported nodes.

a few options I can think - treat that as an error, ignore a node_grouping like that, or add a capability (or capabilities) comprised of nodes that aren't assigned to another EP. not sure which is best.

it might be more consistent to warn and then ignore the entire node_grouping if it contains any nodes assigned to another EPs in all cases (single node, group with node assigned to another EP that partitions the group, and group with node assigned to another EP that doesn't).

onnxruntime/test/framework/ep_plugin_provider_test.cc
```diff
@@ -317,4 +334,165 @@ TEST(PluginExecutionProviderTest, InferOrtDeviceFromDeviceMemoryInfo) {
 #endif  // !defined(ORT_NO_EXCEPTIONS)
 }
 
+static void LoadModelAndAssignNodesToEp(const ORTCHAR_T* model_path,
+                                        const char* ep_name,
+                                        std::unordered_set<std::string> ep_node_names,
```
edgchen1:
```suggestion
                                        const std::unordered_set<std::string>& ep_node_names,
```

onnxruntime/core/session/plugin_ep/ep_plugin_provider_interfaces.cc
adrianlizarraga: As you suggested, I'd prefer to log a warning and skip any node grouping that has any node assigned to another EP. I went ahead and just checked that once at the beginning of the for loop.

Afterwards, the error checking for the other cases is simplified since we now know that all nodes are available. If we then encounter something that could be considered a hard error, we log it and return immediately.

Let me know what you think.

onnxruntime/core/session/plugin_ep/ep_plugin_provider_interfaces.cc
```diff
@@ -207,27 +248,29 @@ PluginExecutionProvider::GetCapability(const onnxruntime::GraphViewer& graph_vie
...
+      if (capabilities.size() != 1) {
+        LOGS(logger, ERROR) << "OrtEp::GetCapability() for " << Type() << " set nodes that cannot be fused together. "
+                            << "Please ensure that the nodes provided to EpGraphSupportInfo_AddNodesToFuse() do not "
                             << "have an unsupported node in any path between two of the supported nodes.";
         return {};
       }
 
-      // Enforce that the nodes in node_set match the nodes in capabilities[0]
+      // Log an error if the nodes in node_set do not match the nodes in capabilities[0]. We expect this to always
+      // be true because we've already checked that the EP did not try to claim nodes already assigned to another EP.
```
edgchen1, line 259: hm, I wonder if the original `ORT_ENFORCE` checks are ok to keep. if those conditions aren't met something is not as expected, right? I suppose logging an error is ok too, but ORT_ENFORCE is louder.

onnxruntime/core/session/plugin_ep/ep_plugin_provider_interfaces.cc
edgchen1: cool, that looks good to me. nice that it simplified the error handling too.

edgchen1 (approved):

onnxruntime/core/session/plugin_ep/ep_plugin_provider_interfaces.cc
skottmckay: It does cost binary size though and causes an abort in a minimal build.

I'm not sure I'd pay for an abort if CreateSupportedPartitions split things up (EP might make a mistake and include nodes that aren't in a single partition but we can gracefully continue even if it did), or for CreateSupportedPartitions randomly adding a new node in the returned values (if we want to check that an `assert` may be a better balance).

edgchen1: sure, `assert` would be fine too. the EP making that mistake should be handled with the `capabilities.size() != 1` check above. but having mismatched node sets at this point is just something that, afaik, we really don't expect to happen.

snnn (approved):
Thanks

snnn:

PR #26166, opened 2025-09-25
Reworked sgemm_kleidi memory allocations to reuse memory buffers

JonathanC-ARM:
### **Key changes**
This PR makes changes to KleidiAI integration within the existing sgemm_kleidiai.cpp implementation.

It was noted that during internal testing that memory allocation overhead due to repeated allocations of vectors was having a negative impact on performance figures. 

The changes introduce thread local buffers for reusing memory during inference.

Android platforms are particularly sensitive to this, we have observed inference times being significantly impacted due to memory allocation overheads
### Example performance
All runs were captured using onnxruntime_perf_test
e.g. onnxruntime_perf_test -v -e cpu -I -m times -x 1 -y 1 -r 1000
**Android Platform**
image

In addition to this on M4 we have also observed slight improvements on models, however its the gain is not as significant as the allocation overhead is lower in terms of total time on that platform 

**Mac Mini M4**
image

**Onnxruntime Mlas Benchmark**
Mlas Benchmark was executed on a Mac Mini M4 with SME2 instructions
Tested code with and without changes in pr and observed the following results (subset shown) comparison generated using compare.py located in google benchmark repo tools
`./onnxruntime_mlas_benchmark --benchmark_filter="SGEMM/NORMAL*" --benchmark_repetitions=100`

```

Benchmark                                                             Time             CPU      Time Old      Time New       CPU Old       CPU New
--------------------------------------------------------------------------------------------------------------------------------------------------
SGEMM/NORMAL_NoTrans/M:63/N:63/K:63/real_time                      -0.1897         -0.1897          3270          2650          3270          2650
SGEMM/NORMAL_NoTrans/M:255/N:63/K:63/real_time                     -0.1468         -0.1469          8383          7152          8382          7151
SGEMM/NORMAL_NoTrans/M:1023/N:63/K:63/real_time                    -0.1506         -0.1506         19072         16200         19072         16200
SGEMM/NORMAL_NoTrans/M:63/N:255/K:63/real_time                     -0.1957         -0.1957          7742          6227          7742          6227
SGEMM/NORMAL_NoTrans/M:255/N:255/K:63/real_time                    -0.1032         -0.1032         14323         12845         14322         12845
SGEMM/NORMAL_TransB/M:63/N:63/K:63/real_time                       -0.2221         -0.2221          3356          2611          3356          2610
SGEMM/NORMAL_TransB/M:255/N:63/K:63/real_time                      -0.0439         -0.0438          8602          8224          8601          8224
SGEMM/NORMAL_TransB/M:1023/N:63/K:63/real_time                     +0.0436         +0.0436         16488         17206         16487         17206
SGEMM/NORMAL_TransB/M:63/N:255/K:63/real_time                      -0.2000         -0.1999          8046          6437          8046          6437
SGEMM/NORMAL_TransB/M:255/N:255/K:63/real_time                     -0.0979         -0.0979         14131         12747         14130         12747
SGEMM/NORMAL_TransB/M:1023/N:255/K:63/real_time                    -0.2836         -0.2836         62540         44802         62540         44802
SGEMM/NORMAL_TransB/M:63/N:1023/K:63/real_time                     -0.2183         -0.2183         15342         11993         15342       
```

Some small regressions have been seen but are difficult to explain, suspected machine variance during run could account for things like 
```
SGEMM/NORMAL_TransB/M:1023/N:63/K:63/real_time                     +0.0436         +0.0436         16488         17206         16487         17206
```
For example, as part of testing these results sgemm_kleidi.cpp was instrumented (after the previous benchmark results) with timer code, in MlasGemmBatch, MlasGemmPackB, and MlasGemmPackBSize.
Which produced the following, indicating that the code performs better in this case on average than baseline which is currently in main
```
Head of main
Function           Count         Avg (ns)     Avg (pretty)
----------------------------------------------------------
MlasGemmBatch      42664        19601.015     19.601 us
MlasGemmPackB      42664          373.943    373.943 ns
MlasGemmPackBSize  42664           17.179     17.179 ns

TLB changes
Function           Count         Avg (ns)     Avg (pretty)
----------------------------------------------------------
MlasGemmBatch      55492        16985.256     16.985 us
MlasGemmPackB      55492          344.800    344.800 ns
MlasGemmPackBSize  55492           16.788     16.788 ns
```

### DIFF
--- a/onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
+++ b/onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
@@ -247,7 +247,7 @@ Return Value:
     size_t lhs_resize = 0;
     if(mul_overflow_size_t_builtin(LhsPackedStride, BatchSize, &lhs_resize))
     {
-        // size_t wraparound detected, exit
+        // size_t wraparound detected for LhsPackedStride, fallback to MLAS
         return false;
     }
 
@@ -271,7 +271,7 @@ Return Value:
         size_t rhs_resize = 0;
         if (mul_overflow_size_t_builtin(RhsPackedStride, BatchSize, &rhs_resize))
         {
-            // size_t wraparound detected, exit
+            // size_t wraparound detected for RhsPackedStride, fallback to MLAS
             return false;
         }
 
--- a/onnxruntime/core/mlas/lib/kleidiai/mlasi_kleidiai.h
+++ b/onnxruntime/core/mlas/lib/kleidiai/mlasi_kleidiai.h
@@ -116,7 +116,28 @@ MlasConv(
     );
 }
 
+/*++
 
+Routine Description:
+
+    This routine determines if a wraparound will occur when multiplying two size_t variables
+    Uses __builtin_mul_overflow if available on the current system and if not falls back
+    to a default implementation to check this wraparound.
+
+Arguments:
+
+    a - Supplies the first number to be muliplied.
+
+    b - Supplies the second number to be muliplied.
+
+    out - pointer to a size_t which acts as the return value in success cases.
+
+Return Value:
+
+    Returns false if the operation was successful
+    Returns true if wraparound of size_t was detected
+
+--*/
 inline bool mul_overflow_size_t_builtin(size_t a, size_t b, size_t* out) {
 #if defined(__has_builtin)
 #  if __has_builtin(__builtin_mul_overflow)


snnn (commented):
This PR introduces a good performance optimization by reusing memory buffers through `thread_local` storage. The approach seems sound, especially for reducing allocation overhead on platforms like Android, as highlighted in the description.

The use of `std::vector::reserve` and `std::vector::resize` for managing the thread-local buffers is appropriate for this kind of optimization.

One minor question regarding the `lhs_packed` and `rhs_packed` buffers:

```cpp
if (g_kai_tls.lhs_packed.capacity() < LhsPackedStride * BatchSize) {
    g_kai_tls.lhs_packed.reserve(LhsPackedStride * BatchSize);
}
g_kai_tls.lhs_packed.resize(LhsPackedStride * BatchSize);
```

This logic ensures the capacity grows as needed. Is there any scenario where the `LhsPackedStride * BatchSize` could become significantly smaller than the current capacity, leading to a large amount of unused reserved memory? While `std::vector` doesn't shrink its capacity automatically, it's generally fine for performance-critical buffers that tend to grow or stabilize at a certain size. Just confirming this is the intended behavior and not a concern for this specific use case.

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
```diff
@@ -230,54 +242,50 @@ Return Value:
...
-
     size_t LhsPackedStride = 0;
     std::byte* LhsPackedData = nullptr;
 
     LhsPackedStride = kai_get_lhs_packed_size_lhs_pack_f32p2vlx1_f32_sme(M, K, mr, kr, sr);
-    auto LhsPacked = std::make_unique<std::byte[]>(LhsPackedStride * BatchSize);
-    LhsPackedData = LhsPacked.get();
 
-    std::unique_ptr<std::byte[]> RhsPacked{nullptr};
+    if (g_kai_tls.lhs_packed.capacity() < LhsPackedStride * BatchSize) {
```
snnn: Is there any measure to avoid LhsPackedStride * BatchSize having integer overflow?

JonathanC-ARM:
> This PR introduces a good performance optimization by reusing memory buffers through `thread_local` storage. The approach seems sound, especially for reducing allocation overhead on platforms like Android, as highlighted in the description.
> 
> The use of `std::vector::reserve` and `std::vector::resize` for managing the thread-local buffers is appropriate for this kind of optimization.
> 
> One minor question regarding the `lhs_packed` and `rhs_packed` buffers:
> 
> ```c++
> if (g_kai_tls.lhs_packed.capacity() < LhsPackedStride * BatchSize) {
>     g_kai_tls.lhs_packed.reserve(LhsPackedStride * BatchSize);
> }
> g_kai_tls.lhs_packed.resize(LhsPackedStride * BatchSize);
> ```
> 
> This logic ensures the capacity grows as needed. Is there any scenario where the `LhsPackedStride * BatchSize` could become significantly smaller than the current capacity, leading to a large amount of unused reserved memory? While `std::vector` doesn't shrink its capacity automatically, it's generally fine for performance-critical buffers that tend to grow or stabilize at a certain size. Just confirming this is the intended behavior and not a concern for this specific use case.

Yes, this is the intended behavior. You are correct that if the required buffer size later becomes smaller, the capacity may remain larger than strictly necessary, leading to some unused reserved memory. In practice, we expect the footprint to plateau based on the model/matrices under test, and the maximum allocation would be the same with or without this approach.

We deliberately avoid shrinking the vectors, since at this level we cannot predict future operations, and frequent shrink/grow cycles could be costly. The trade-off is slightly higher sustained memory usage, but with more predictable performance.

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
JonathanC-ARM: That is a fair point, I can add some checks before resizing to try protect against overflow.

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
```diff
@@ -14,6 +14,16 @@
...
 
+
+// Thread-local reusable buffers to reduce allocation overhead across tiles.
+struct KaiTlsBuffers {
+    std::vector<float> output_tile;
+    std::vector<float> bias_zero;
+    std::vector<std::byte> rhs_packed;
+    std::vector<std::byte> lhs_packed;
+};
+static thread_local KaiTlsBuffers g_kai_tls;
```
snnn, line 25: Please note that ORT as a library could not control when a thread_local variable is deallocated.  As the destructor of this var does not depend on anything except the C++ runtime, I feel it's fine.

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
JonathanC-ARM: I've added a new function to mlasi_kleidi.h which uses a gcc/clang built in for checking the wrap around for size_t, I also added a generic implementation for systems without the builtin.

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
```diff
@@ -336,9 +358,23 @@ Return Value:
...
-        // Allocate temporary buffer for raw A*B result
-        std::vector<float> OutputTile(TileSizeM * TileSizeN, 0.0f);
-        float* temp_tile = OutputTile.data();
+        // Allocate temporary buffer for raw A*B result (TLS reusable buffer)
+
+        size_t tile_elems = 0;
+        if (mul_overflow_size_t_builtin(TileSizeM, TileSizeN, &tile_elems))
+        {
+            // size_t wraparound detected, exit
+            return
```
edgchen1: missing semi-colon?

also, this will just return early from the worker thread and not actually exit the function. perhaps this check should be moved before the `MlasTrySimpleParallel` call using the maximum possible values of `TileSizeM` and `TileSizeN`.

JonathanC-ARM: Apologies, that semi-colon was a typo on my part. 

I have moved that logic to outside the MlasTrySimpleParallel so we can return based on the max size

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
```diff
@@ -125,15 +134,18 @@ Return Value:
         const size_t sr = UseSME2 ? kai_get_sr_matmul_clamp_f32_f32p2vlx1_f32p2vlx1biasf32_sme2_mopa()
                                   : kai_get_sr_matmul_clamp_f32_f32p2vlx1_f32p2vlx1b_2vlx2vl_sme_mopa();
 
-        // pass zeroed bias values
-        const std::vector<float> bias(N);
+        // Ensure size and zero the used span (only newly added elements are value-initialized by resize).
+        if (g_kai_tls.bias_zero.size() != N) {
+            g_kai_tls.bias_zero.resize(N);
+        }
+        std::fill_n(g_kai_tls.bias_zero.data(), N, 0.0f);
```
edgchen1: is it only ever set to zeros? it was a const vector before, so I assume nothing else is writing to it?

can it be simplified to something like this? seems like we could avoid repeatedly writing the initial zeros.

```suggestion
        g_kai_tls.bias_zero.resize(N, 0.0f);
```

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
```diff
@@ -225,59 +237,69 @@ Return Value:
...
+    {
+        // size_t wraparound detected, exit
+        return false;
+    }
+
+    if (g_kai_tls.lhs_packed.capacity() < lhs_resize) {
+
+        g_kai_tls.lhs_packed.reserve(lhs_resize);
+    }
+    g_kai_tls.lhs_packed.resize(lhs_resize);
```
edgchen1: general: what is the benefit of explicitly calling `capacity()` and `reserve()`, if `resize()` will take care of additional allocation as needed?

edgchen1:
> All runs were captured using onnxruntime_perf_test
> e.g. onnxruntime_perf_test -v -e cpu -I -m times -x 1 -y 1 -r 1

with `-r 1`, it is running one iteration, right? measuring just a single iteration might be noisy. could you please try measuring more iterations?

also, could you share some relevant measurements from the MLAS SGEMM microbenchmarks?
- build with `--build_micro_benchmarks`
- run `onnxruntime_mlas_benchmark`, e.g. `onnxruntime_mlas_benchmark --benchmark_filter="SGEMM*"`

JonathanC-ARM:
> > All runs were captured using onnxruntime_perf_test
> > e.g. onnxruntime_perf_test -v -e cpu -I -m times -x 1 -y 1 -r 1
> 
> with `-r 1`, it is running one iteration, right? measuring just a single iteration might be noisy. could you please try measuring more iterations?
> 
> also, could you share some relevant measurements from the MLAS SGEMM microbenchmarks?
> 
>     * build with `--build_micro_benchmarks`
> 
>     * run `onnxruntime_mlas_benchmark`, e.g. `onnxruntime_mlas_benchmark --benchmark_filter="SGEMM*"`

Yeah thats yet another typo on my part, this would have been with 1000 runs. So noise shouldn't be a factor. 

I can definitely look and put in some of those numbers from the micro benchmark also

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
JonathanC-ARM: Yeah thats a fair comment, can definitely just be resized to N with 0's

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
JonathanC-ARM: Maybe this was over thought,it makes sense to call resize here same as the bias you pointed out above.

JonathanC-ARM:
@edgchen1 I have added a subset of benchmark results to description

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
```diff
@@ -225,59 +234,61 @@ Return Value:
     size_t n_step = UseSME2 ? kai_get_n_step_matmul_clamp_f32_f32p2vlx1_f32p2vlx1biasf32_sme2_mopa()
                             : kai_get_n_step_matmul_clamp_f32_f32p2vlx1_f32p2vlx1b_2vlx2vl_sme_mopa();
 
-    if (M < m_step && N < n_step && !Data->BIsPacked) {
+    if ((M < m_step || N < n_step) && !Data->BIsPacked) {
```
hariharans29, line 237: Are there no tests to exercise the either M or N are lesser than their respective KAI steps ? 
If so, could we please add them now ?

I am guessing with the current version of main - if we ran into a problem shape where either M or N were less than their respective steps (not both), it would crash in KAI given that the bailout logic had a bug ?

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
```diff
@@ -225,59 +234,61 @@ Return Value:
...
-        size_t RhsPackedStride = 0;
-        std::byte* RhsPackedData = nullptr;
-
         RhsPackedStride = ArmKleidiAI::MlasGemmPackBSize(TransA, TransB, N, K);
-        RhsPacked = std::make_unique<std::byte[]>(RhsPackedStride * BatchSize);
-        RhsPackedData = RhsPacked.get();
+        size_t rhs_resize = 0;
+        if (mul_overflow_size_t_builtin(RhsPackedStride, BatchSize, &rhs_resize))
+        {
+            // size_t wraparound detected, exit
```
hariharans29: Very minor nit: Consistency.

I think the other places call out that we fallback to MLAS when we return false

JonathanC-ARM: Agreed should say specifically it will fall back to mlas

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
JonathanC-ARM: In test_fgemm_fixture.h

A series of tests are already there
 ```
for (size_t b = 16; b <= 256; b <<= 1) {
      test_registered += RegisterTestTransposeABProduct(b, b, b, 1, 1.0f, 0.0f);
    }
```
In builds with kleidi enabled these would trigger this fallback both before and after this change. 

From what I've seen the kernel will still work when M or N are less than the respective step sizes, it just starts to perform sub optimally. So after some tweaking we settled on this OR Vs the AND. So in terms of what's there Vs the change, neither will crash, it just might perform differently before Vs after.

Comment (hariharans29) [reply onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp:237]: Thank you for the context

File: onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp [hunk 8]
```diff
@@ -225,59 +234,61 @@ Return Value:
...
-        RhsPackedData = RhsPacked.get();
+        size_t rhs_resize = 0;
+        if (mul_overflow_size_t_builtin(RhsPackedStride, BatchSize, &rhs_resize))
+        {
+            // size_t wraparound detected for RhsPackedStride, fallback to MLAS
+            return false;
+        }
+
+        g_kai_tls.rhs_packed.resize(rhs_resize);
+        RhsPackedData = g_kai_tls.rhs_packed.data();
```
edgchen1, line 279: just to confirm, the RHS packed buffer doesn't need to be initialized to zero, right?

edgchen1:
/azp run Linux QNN CI Pipeline,Win_TRT_Minimal_CUDA_Test_CI,Windows ARM64 QNN CI Pipeline,
Windows GPU Doc Gen CI Pipeline

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
JonathanC-ARM: No it doesn't require this, the ZA registers are zero'd by the kleidi kernel prior to the matrices being multiplied. So the result just overrides whatever is currently stored, the most important thing is that the rhs_packed buffer is the correct size for the result.

PR #26102, opened 2025-09-20
Add BF16 kernels in several ops for Gemma-3

kunal-vaishnavi:
### Description

This PR adds missing kernels for bfloat16 precision across several ops in both the `ai.onnx` and `com.microsoft` domains.

1. `SkipLayerNormalization` [contrib op](https://github.com/microsoft/onnxruntime/blob/main/docs/ContribOperators.md#commicrosoftskiplayernormalization)
2. `Conv` for [opset 22](https://onnx.ai/onnx/operators/onnx__Conv.html)
3. `Pow` for [opset 15](https://onnx.ai/onnx/operators/onnx__Pow.html)
4. `AveragePool` for [opset 22](https://onnx.ai/onnx/operators/onnx__AveragePool.html)

This PR also enables weight-only quantization of a bfloat16 `MatMul` op to a bfloat16 `MatMulNBits` [contrib op](https://github.com/microsoft/onnxruntime/blob/main/docs/ContribOperators.md#commicrosoftmatmulnbits).

### Motivation and Context

This PR enables running ONNX models from the Gemma-3 family that are generated with bfloat16 precision.

PR #26168, opened 2025-09-26
Fix Windows non-CUDA build break

hariharans29:
### Description
Fix warnings in non-CUDA EP Windows builds

### Motivation and Context

 
Minor fix from https://github.com/microsoft/onnxruntime/pull/26102

### DIFF
--- a/onnxruntime/test/providers/cpu/nn/conv_fp16_test.cc
+++ b/onnxruntime/test/providers/cpu/nn/conv_fp16_test.cc
@@ -273,14 +273,13 @@ TEST(ConvFp16Test, Conv1D_Bias) {
 }
 
 TEST(ConvBF16Test, Conv2D_1) {
-#ifdef USE_CUDA
+#ifndef USE_CUDA
+  GTEST_SKIP() << "BFloat16 tests are only enabled on CUDA builds";
+#else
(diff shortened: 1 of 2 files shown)


onnxruntime/test/providers/cpu/nn/conv_fp16_test.cc
```diff
@@ -282,6 +282,7 @@ TEST(ConvBF16Test, Conv2D_1) {
   return;
 #endif
 
+#ifdef USE_CUDA
```
edgchen1: is the warning for unreachable code after the unconditional `return`? would `GTEST_SKIP()` work any better without an additional ifdef?

kunal-vaishnavi (dismissed):

onnxruntime/test/providers/cpu/nn/conv_fp16_test.cc
hariharans29: Even with the GTEST_SKIP() - I get the same warning.

hariharans29: Re-worked it a bit

tianleiwu (approved):

PR #26169, opened 2025-09-26
[WebNN] Fix bug in int64 data type support validation

Honry:
Only allows int64 data type when current WebNN context totally doesn't support int64 but support int32.

### DIFF
--- a/onnxruntime/core/providers/webnn/builders/helper.cc
+++ b/onnxruntime/core/providers/webnn/builders/helper.cc
@@ -226,7 +226,9 @@ bool AreDataTypesSame(const std::string_view op_type,
   return true;
 }
 
-bool IsSupportedDataType(const int32_t onnx_data_type, const emscripten::val& webnn_supported_data_types) {
+bool IsSupportedDataType(const int32_t onnx_data_type,
+                         const emscripten::val& webnn_supported_data_types,
+                         const emscripten::val& wnn_limits) {
   auto it = onnx_to_webnn_data_type_map.find(static_cast<ONNX_NAMESPACE::TensorProto_DataType>(onnx_data_type));
   if (it == onnx_to_webnn_data_type_map.end())
     return false;
@@ -240,7 +242,8 @@ bool IsSupportedDataType(const int32_t onnx_data_type, const emscripten::val& we
 
   if (webnn_data_type == "int64" &&
       !is_supported &&
-      webnn_supported_data_types.call<emscripten::val>("includes", emscripten::val("int32")).as<bool>()) {
+      webnn_supported_data_types.call<emscripten::val>("includes", emscripten::val("int32")).as<bool>() &&
+      !wnn_limits["constant"]["dataTypes"].call<emscripten::val>("includes", emscripten::val("int64")).as<bool>()) {
     // Current context doesn't support int64, but int32 is supported.
     // We can use int32 as a workaround.
     is_supported = true;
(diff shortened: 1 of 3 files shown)


Honry:

onnxruntime/core/providers/webnn/builders/helper.cc
```diff
@@ -280,8 +283,9 @@ bool IsDataTypeSupportedByWebNNOp(const std::string_view onnx_op_type,
                           << webnn_input_output_name << "]";
     return false;
   }
-  if (!IsSupportedDataType(
-          onnx_data_type, wnn_limits[std::string(webnn_op_type)][std::string(webnn_input_output_name)]["dataTypes"])) {
+  if (!IsSupportedDataType(onnx_data_type,
+                           wnn_limits[std::string(webnn_op_type)][std::string(webnn_input_output_name)]["dataTypes"],
+                           wnn_limits)) {
```
fdwr: It feels odd to pass a subfield of `wnn_limits` but then also to pass the entire `wnn_limits` in the very next parameter, because then why not just pass the full `wnn_limits` along with the `webnn_op_type` and `webnn_input_output_name`?

fdwr (approved):
> Only allows int64 data type when current WebNN context totally doesn't support int64 but support int32

Can you explain the nature of the bug more, and which Chromium backend or ORT EP encounters the issue?

Honry:
> > Only allows int64 data type when current WebNN context totally doesn't support int64 but support int32
> 
> Can you explain the nature of the bug more, and which Chromium backend or ORT EP encounters the issue?

Currently we only fallback `int64` data type to `int32` for the backend that totally doesn't support `int64`. But for some other backends, e.g. TFLite, there're partial ops that don't support int64 (e.g. ScatterND's indices only supports `int32`), current code doesn't fallback these ops and still passes `int64` indices to `scatterND`, then it will throw error.

onnxruntime/core/providers/webnn/builders/helper.cc
Honry: Good point, while in following usage, the `constant` doesn't have an input/output name, or shall we just pass empty string as its input/output name and skip it if it is empty?

```js
if (IsSupportedDataType(data_type, wnn_limits_["constant"]["dataTypes"], wnn_limits_)) {
```

fdwr: (🤔 heading home for the night - will reply tomorrow, but likely you'll come up with a nice clean idea before then 😉)

Honry: I made some improvement, PTAL again. Thanks!

onnxruntime/core/providers/webnn/builders/helper.cc
```diff
@@ -226,21 +226,36 @@ bool AreDataTypesSame(const std::string_view op_type,
...
+                         const std::string_view webnn_op_type,
+                         const std::string_view webnn_input_output_name) {
   auto it = onnx_to_webnn_data_type_map.find(static_cast<ONNX_NAMESPACE::TensorProto_DataType>(onnx_data_type));
   if (it == onnx_to_webnn_data_type_map.end())
     return false;
 
   const std::string_view webnn_data_type = it->second;
 
+  // MLOpSupportLimits has different structure. Certain WebNN ops have input and output name,
+  // special cases like 'constant', 'input' and 'output' have no input or output name.
```
fdwr, line 240: 🤔 Reviewing @BruceDai's test CR recently, it made we wonder if it would be more useful for callers to have `constant` and `input` placed at the same level in the hierarchy as all the other operators. The `constant` operator is essentially a generator operator (kinda like the `fillSequence` that never made it into WebNN) which takes 0 inputs and yields 1 output. e.g.

```diff
dictionary MLOpSupportLimits
{
    MLInputOperandLayout preferredInputLayout;
    [EnforceRange] unsigned long long maxTensorByteLength;
-   MLTensorLimits input;
-   MLTensorLimits constant;
    MLTensorLimits output;
};

+ dictionary MLSingleOutputSupportLimits
+ {
+     MLTensorLimits output;
+ };

partial dictionary MLOpSupportLimits
{
+   MLSingleOutputSupportLimits input;
+   MLSingleOutputSupportLimits constant;
};
```

The oddity though is the `output` field, which doesn't fit nicely into this pattern, because there is no 'output' operator (and if there was, then it would have a single input, but the existing `MLSingleInputSupportLimits` actually has an input *and* an output, which would make it awkward). 🤷‍♂️

Anyway, thinking aloud - resolve me after reading.

fdwr (approved):
👍

fdwr:
Retrying failed "Windows ARM64 QNN CI Pipeline".

PR #26173, opened 2025-09-26
Add trt-rtx install command

anujj:
Add trt-rtx install command

### DIFF
--- a/docs/genai/howto/install.md
+++ b/docs/genai/howto/install.md
@@ -30,6 +30,12 @@ pip install onnxruntime-genai
 pip install onnxruntime-genai-directml
 ```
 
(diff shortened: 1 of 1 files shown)


snnn (commented):
Hi @anujj, thanks for adding the installation instructions for the new TRT-RTX package.

I noticed that the `onnxruntime-genai-trt-rtx` package is not yet available on PyPI. I assume this documentation is in preparation for a future release.

To avoid confusion for users who might try to install this package before it's released, could we add a note to the documentation? Something like:

> **Note:** This package is not yet available on PyPI and will be included in a future release.

This would help manage user expectations and prevent unnecessary bug reports.

PR #26178, opened 2025-09-26
[WIP] Move KleidiAI support detection to CMake logic.

edgchen1:

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -90,11 +90,6 @@ option(onnxruntime_USE_JSEP "Build with JavaScript implemented kernels support" 
 option(onnxruntime_USE_SVE "Build with SVE support in MLAS" OFF)
 
 option(onnxruntime_USE_KLEIDIAI "Build with KleidiAI integration in MLAS" OFF)
-# iOS simulator build explicitly builds targets with USE_KLEIDIAI=ON so attempting to force override if so
-if(APPLE AND CMAKE_OSX_ARCHITECTURES MATCHES "x86_64")
-    message(WARNING "Disabling KleidiAI: not supported on Apple x86_64 platforms")
-  set(onnxruntime_USE_KLEIDIAI OFF CACHE BOOL "" FORCE)
-endif()
 option(onnxruntime_BUILD_UNIT_TESTS "Build ONNXRuntime unit tests" ON)
 option(onnxruntime_BUILD_CSHARP "Build C# library" OFF)
 option(onnxruntime_BUILD_OBJC "Build Objective-C library" OFF)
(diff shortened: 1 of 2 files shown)


cmake/detect_onnxruntime_target_platform.cmake
```diff
@@ -0,0 +1,80 @@
...
+          onnxruntime_target_platform STREQUAL "i686")
+    set(onnxruntime_target_platform "x86")
+  else()
+    message(FATAL_ERROR "Unknown target platform: ${onnxruntime_target_platform}")
+  endif()
+elseif(APPLE)
+  if(DEFINED CMAKE_OSX_ARCHITECTURES)
+    # We'll only set onnxruntime_target_platform when CMAKE_OSX_ARCHITECTURES specifies a single architecture.
+    list(LENGTH CMAKE_OSX_ARCHITECTURES CMAKE_OSX_ARCHITECTURES_LEN)
+    if(CMAKE_OSX_ARCHITECTURES_LEN EQUAL 1)
```
snnn, line 38: We stop supporting the case that CMAKE_OSX_ARCHITECTURES_LEN is greater than 1, as a lot of core ORT features are not supported in this case. Our build pipelines have been refactored to build the two architectures separately then combine them.

snnn (commented):
Hello! I've reviewed the changes, and this is a great refactoring that simplifies the build logic. I have a few technical suggestions for improvement.

---

### 1. Potential bug in KleidiAI platform detection for Apple

In `cmake/CMakeLists.txt`, the platform check for Apple is:
```cmake
if(NOT (... OR (APPLE AND CMAKE_OSX_ARCHITECTURES STREQUAL "arm64")))
```
This check depends on `CMAKE_OSX_ARCHITECTURES` being defined. According to the logic in the new `cmake/detect_onnxruntime_target_platform.cmake` file, if `CMAKE_OSX_ARCHITECTURES` is *not* defined during a build on an Apple machine, `onnxruntime_target_platform` is set from `CMAKE_SYSTEM_PROCESSOR`. For a native build on Apple Silicon, this means `onnxruntime_target_platform` would be `arm64`, but `CMAKE_OSX_ARCHITECTURES` would be undefined.

In this case, the check above would evaluate to false, and KleidiAI would be incorrectly disabled.

A more robust condition would be to check `onnxruntime_target_platform` directly:
```cmake
if(NOT (... OR (APPLE AND onnxruntime_target_platform STREQUAL "arm64")))
```
This would correctly enable KleidiAI for native Apple Silicon builds, regardless of whether `CMAKE_OSX_ARCHITECTURES` is explicitly set.

### 2. Potential issues in `cmake/adjust_global_compile_flags.cmake`

I noticed two small things in this file:

*   The `if/elseif` chain appears to have a syntax error based on the indentation in the diff. It looks like an `if` statement was added after an `elseif`, but before the final `else`, which is not valid CMake syntax. The logic for enabling `SAFESEH` for x86 builds should likely be nested inside the `elseif` block that checks for `x64` or `x86`, or placed after the entire `if-elseif-else` block.

*   The `FATAL_ERROR` message at the end of that block seems to have a copy-paste error:
    ```cmake
    message(FATAL_ERROR "Unknown CMAKE_SYSTEM_PROCESSOR: ${CMAKE_SYSTEM_PROCESSOR}")
    ```
    Since the variable being checked throughout this block is `onnxruntime_target_platform`, the error message should probably refer to that variable for clearer debugging:
    ```cmake
    message(FATAL_ERROR "Unknown onnxruntime_target_platform: ${onnxruntime_target_platform}")
    ```

---
*This review comment was generated by an AI assistant.*

cmake/detect_onnxruntime_target_platform.cmake
edgchen1: does that mean we should consider it an error if `CMAKE_OSX_ARCHITECTURES` has more than one element?

snnn (approved):

cmake/detect_onnxruntime_target_platform.cmake
edgchen1: might treat that as an error in another PR.

PR #26185, opened 2025-09-27
Add CastLike grad

youzow3:
### Description

Add CastLike grad implementation.

### Motivation and Context

Some models use CastLike instead of Cast, which prevents on-device training artifacts from being generated.


youzow3:

youzow3:
Updated the test code but I'm not sure this is correct code to test cast between float-double.
Also, I think integer types cannot have gradient and gradient for "target_type" is always 0, so I don't think we need tests for cast between integer to float or float to integer.

PR #26196, opened 2025-09-29
[WebNN] Rename outputs for gru and lstm

Honry:
WebNN spec changed the output names of `gru` and `lstm` for `OpSupportLimits` in https://github.com/webmachinelearning/webnn/pull/857, renamed them in WebNN EP as well.


Honry:

PR #26200, opened 2025-09-29
[CUDA] Update GroupQueryAttention to support quantized KV cache

tianleiwu:
This updates GroupQueryAttention operator to suppport KV cache quantization. The KV cache can store in data types of float8, int8 or uint8  (for 4 bits).

### Background

A center piece of LLM inference is the Key-Value (KV) cache, a critical optimization that accelerates text generation by removing computational redundancy. However, this introduces a new, formidable bottleneck related to memory. 

#### Related Research
(1) Early work on quantized KV cache uses **per-tensor** post-training quantization.

(2) KIVI: Asymmetric Per-Channel/Per-Token Quantization. The core observation behind KIVI is that the distribution of outliers in the Key and Value caches is fundamentally different. So it uses asymmetric quantization strategy:
 * Key Cache: Quantized on a per-channel basis. This allows each channel to have its own unique scale and zero-point, effectively isolating the high-magnitude outlier channels and preventing them from skewing the quantization range for the other, smaller-valued channels.
 * Value Cache: Quantized on a per-token basis, which was found to be more effective for its data distribution.
 * <1% accuracy drop on average at 2-bit.
 
(3) NQKV: Normal Distribution Quantile Quantization. The observation when grouped into blocks along the token dimension, elements of kv cache closely approximate a normal distribution. It divides each token into smaller blocks and quantizes each block in 4-bits independently.

(4) KVQuant: It uses Pre-RoPE Per-Channel Key Quantization. It also uses per-vector Dense-and-Sparse Quantization.

#### Related Implementation
* Huggingface Transformers follow the idea of KIVI. To improve performance, it retains a fixed size (like 128) residual cache to store keys and values in their original precision. When the residual cache reaches its maximum capacity the stored keys and values are quantized and the cache content is discarded. See https://huggingface.co/blog/kv-cache-quantization for detail.
* TensorRT-LLM supports INT8 and FP8 KV cache implementations. It shows up to a 1.45x throughput improvement over FP16 baseline.

Based on the related work, quantized kv cache is better to support per-tensor (like TensorRT-LLM) and KIVI style (per-channel quantization for key, and per token dynamic quantization for value).

In this pull request, we will focus on per-tensor and per-channel static quantization.

The following are not supportted, but can be extended in the future if needed:
 * per token dynamic quantization: This is complicated since it need integration with genai.
 * 2-bit quantization: Current design can be easily extended to 2-bit. It is not implemented in this PR.
 * zero-points:  Assume that symmetric quantization is enough, zero-point support is lower priority.
 * block-quantization: seems not very popular since per-channel can have better accuracy.
 * per-head quantization: seems not very popular since per-channel can have better accuracy.
 * fp4: lower priority than 2-bit support.
 * Two other float8 data types: right now only supports float8 data types having better match with CUDA.

Note: int2 data type is not supported by onnx (see https://github.com/onnx/onnx/pull/7163) right now.

### Acknowledgment

This work started from a draft proposal from Nachiket Khasbag (Nvidia) and Gaurav Garg (Nvidia) about supporting Quantized GQA.

### DIFF
--- a/onnxruntime/core/graph/contrib_ops/bert_defs.cc
+++ b/onnxruntime/core/graph/contrib_ops/bert_defs.cc
@@ -235,7 +235,22 @@ void BaseGroupQueryAttentionTypeAndShapeInference(ONNX_NAMESPACE::InferenceConte
                                                   int past_key_index = -1,
                                                   int use_max_past_present_buffer = -1,
                                                   int output_qk_index = -1) {
-  ONNX_NAMESPACE::propagateElemTypeFromInputToOutput(ctx, 0, 0);
+  // Type inference for outputs
+  ONNX_NAMESPACE::propagateElemTypeFromInputToOutput(ctx, 0, 0);  // output
+
+  if (ctx.getNumOutputs() >= 3) {  // has present output
+    const auto* past_key_type = ctx.getInputType(past_key_index);
+    if (past_key_type != nullptr) {
+      // present_key and present_value have the same type as past_key/past_value.
+      // This allows them to be int8 or packed uint8 when quantization is enabled.
+      ONNX_NAMESPACE::propagateElemTypeFromInputToOutput(ctx, past_key_index, 1);      // present_key
+      ONNX_NAMESPACE::propagateElemTypeFromInputToOutput(ctx, past_key_index + 1, 2);  // present_value
+    } else {
+      // If no past state, present is the same type as query.
+      ONNX_NAMESPACE::propagateElemTypeFromInputToOutput(ctx, 0, 1);
+      ONNX_NAMESPACE::propagateElemTypeFromInputToOutput(ctx, 0, 2);
+    }
+  }
 
   int64_t kv_sequence_length = -1;
   if (hasInputShape(ctx, 0)) {
@@ -280,12 +295,6 @@ void BaseGroupQueryAttentionTypeAndShapeInference(ONNX_NAMESPACE::InferenceConte
   }
 
   if (ctx.getNumOutputs() >= 3) {  // has present output
-    // copy the type from query to present key
-    ONNX_NAMESPACE::propagateElemTypeFromInputToOutput(ctx, 0, 1);
-
-    // copy the type from query to present value
-    ONNX_NAMESPACE::propagateElemTypeFromInputToOutput(ctx, 0, 2);
-
     int64_t total_sequence_length_value = 0;
     const auto* total_sequence_length_data = ctx.getInputData(6);
     if (total_sequence_length_data != nullptr) {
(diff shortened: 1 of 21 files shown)


onnxruntime/core/graph/contrib_ops/bert_defs.cc
```diff
@@ -1237,19 +1277,23 @@ ONNX_MS_OPERATOR_SET_SCHEMA(
...
         .Output(3,
                 "output_qk",
                 "Values of QK matrix multiplication, either before or after softmax normalization",
                 "T",
                 OpSchema::Optional)
+        .Output(4, "present_k_scale", "Scale tensor for present_key.", "T_SCALE", OpSchema::Optional)
+        .Output(5, "present_v_scale", "Scale tensor for present_value.", "T_SCALE", OpSchema::Optional)
         .TypeConstraint("T", {"tensor(float16)", "tensor(bfloat16)", "tensor(float)"}, "Constrain input and output to float tensors.")
+        .TypeConstraint("T_CACHE", {"tensor(float)", "tensor(float16)", "tensor(bfloat16)", "tensor(uint8)", "tensor(int8)", "tensor(int4)", "tensor(float8e4m3fn)", "tensor(float8e5m2)"}, "Constrain KV cache types.")
+        .TypeConstraint("T_SCALE", {"tensor(float)"}, "Constrain scale types.")
```
vishalpandya1990: Should we also have 16-bit floating point formats for the quantization scales?

onnxruntime/core/graph/contrib_ops/bert_defs.cc
```diff
@@ -1166,6 +1201,9 @@ ONNX_MS_OPERATOR_SET_SCHEMA(
               "Output values of QK matrix multiplication before (1) or after (2) softmax normalization. Default value is 0 (don't output).",
               AttributeProto::INT,
               static_cast<int64_t>(QKOutputType::NO_OUTPUT))
+        .Attr("k_quant_type", "Quantization type for K cache. One of 'NONE', 'PER_TENSOR', 'PER_CHANNEL', 'PER_TOKEN'.", AttributeProto::STRING, std::string("NONE"))
+        .Attr("v_quant_type", "Quantization type for V cache. One of 'NONE', 'PER_TENSOR', 'PER_CHANNEL', 'PER_TOKEN'.", AttributeProto::STRING, std::string("NONE"))
+        .Attr("kv_cache_bit_width", "Bit width of quantized KV cache when its data type is uint8. Supported values are 4 and 2.", AttributeProto::INT, OPTIONAL_VALUE)
```
vishalpandya1990, line 1200: Just to be clear, this field will not be used (unset) when KV cache is in FP8 or INT8 format?

tianleiwu: Right. It will be ignored when kv cache is not uint8, since other data type has implicit bit width information.

onnxruntime/core/graph/contrib_ops/bert_defs.cc
tianleiwu: Good suggestion. I've updated to use T for scale data.

onnxruntime/core/graph/contrib_ops/bert_defs.cc
```diff
@@ -1237,19 +1278,22 @@ ONNX_MS_OPERATOR_SET_SCHEMA(
...
+                "T_CACHE")
         .Output(3,
                 "output_qk",
                 "Values of QK matrix multiplication, either before or after softmax normalization",
                 "T",
                 OpSchema::Optional)
+        .Output(4, "present_k_scale", "Scale tensor for present_key.", "T", OpSchema::Optional)
+        .Output(5, "present_v_scale", "Scale tensor for present_value.", "T", OpSchema::Optional)
         .TypeConstraint("T", {"tensor(float16)", "tensor(bfloat16)", "tensor(float)"}, "Constrain input and output to float tensors.")
+        .TypeConstraint("T_CACHE", {"tensor(float)", "tensor(float16)", "tensor(bfloat16)", "tensor(uint8)", "tensor(int8)", "tensor(int4)", "tensor(float8e4m3fn)", "tensor(float8e5m2)"}, "Constrain KV cache types.")
```
kunal-vaishnavi: Should we add `uint4` as well?

tianleiwu: It's symmetric quantization so the type need to be signed. uint8 is added here since there is no data type for int2, so we have to use uint8.
For consistency, I removed int4 as well.

PR #26201, opened 2025-09-29
Fix(build): Correct maybe-uninitialized and range-loop-construct warnings

snnn:
This pull request addresses two build warnings that were treated as errors:

1.  **`maybe-uninitialized` warning in `nchwc_transformer.cc`**:
    A `maybe-uninitialized` warning was triggered when accessing `nchwc_inputs[0]` without checking if `nchwc_inputs` was empty. This could lead to a crash if a binary node had no inputs. The fix adds a check to ensure the vector is not empty before accessing the first element.

2.  **`range-loop-construct` warnings in `transformer_memcpy.cc`**:
    The compiler warned about creating copies of objects in range-based for loops. This was resolved by using references (`const auto&`) instead of copies (`const auto`) for the loop variables, which is more efficient.

This fix was provided by an AI bot.

### DIFF
--- a/onnxruntime/core/optimizer/nchwc_transformer.cc
+++ b/onnxruntime/core/optimizer/nchwc_transformer.cc
@@ -609,6 +609,10 @@ void NchwcTransformerImpl::TransformBinary(Node& node, bool add_node) {
     nchwc_inputs.push_back(nchwc_input);
   }
 
+  if (nchwc_inputs.empty()) {
+    return;
+  }
+
   auto* nchwc_input_0 = nchwc_inputs[0];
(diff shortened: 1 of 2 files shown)


edgchen1 (approved):
looks fine. FWIW, I think `gsl::not_null` is not very expensive to copy. in which build are these showing up as errors?

snnn:
> in which build are these showing up as errors?

Linux GCC 15. Build command was: 
```
python3 tools/ci_build/build.py --update --build --build_dir b2 --skip_submodule_sync --parallel 16  --build_shared_lib --config Release --cmake_extra_defines protobuf_WITH_ZLIB=OFF
```

PR #26202, opened 2025-09-29
Bump onnx to 1.19.1

titaiwangms:
~~Test rel-1.19.1~~

Bump to ONNX==1.19.1

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -287,8 +287,12 @@ if (onnxruntime_ENABLE_TRAINING_APIS)
 endif()
 
 
-# Single output director for all binaries
+# Single output directory for all binaries
 set(RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin CACHE PATH "Single output directory for all binaries.")
+
+# Local mirror directory of cmake dependencies
+set(REPO_ROOT ${PROJECT_SOURCE_DIR}/..)
+set(onnxruntime_CMAKE_DEPS_MIRROR_DIR ${REPO_ROOT}/mirror CACHE PATH "Path to the local mirror of cmake dependencies")
 
 
 include(FetchContent)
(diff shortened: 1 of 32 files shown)


tools/ci_build/github/linux/docker/scripts/manylinux/requirements.txt
```diff
@@ -3,7 +3,7 @@ mypy
 pytest
 setuptools>=68.2.2
 wheel
-onnx==1.19.0
+onnx @ git+https://github.com/onnx/onnx@e11938a232432630e1ff8904bb80091af6823021
```
justinchuby: ==1.19.1rc1

titaiwangms: Commit is preferred for vcpkg and deps.txt

justinchuby: But this is requirements? Specifying a commit will mean the env has to do a source build instead of installing from wheel. But it doesn't matter if it works I guess

justinchuby: However, we do want to verify that the publish wheels are verified. So if it can be done here it would be nice too

titaiwangms: Eventually, this is going to be the official package you release, aka 1.19.1. the commit can give me more precise info on where are we at right now in terms of the build system ort is using.

tools/ci_build/requirements/pybind/requirements.txt
```diff
@@ -2,9 +2,10 @@ setuptools
 flatbuffers
 wheel==0.45.1
 pytest
-numpy>=1.19.0
+onnx==1.19.1rc1
```
adrianlizarraga: Is removing numpy as a dependency intentional?

adrianlizarraga: Also, this will now install this specific onnx python package on every build of ORT, right? I don't think we should do that.

tools/ci_build/requirements/pybind/requirements.txt
```diff
@@ -2,9 +2,10 @@ setuptools
...
 sympy>=1.10
 packaging
 cerberus
 psutil
 flatbuffers
+onnx-ir
```
adrianlizarraga: Is this a necessary dependency for pybind?

tools/ci_build/requirements/pybind/requirements.txt
titaiwangms: Thank you!

PR #26203, opened 2025-09-29
[MLAS] Fix SVE enabled pipeline break

hariharans29:
### Description
The `MLAS_USE_SVE` macro was missing for some unittests/benchmark targets. In the original PR, it was scoped down to just the mlas target and this resulted in different mlas platform struct definitions across targets.

### Motivation and Context
Fix pipeline crash and unblock daily pipeline run
#25238 (change that introduced the issue)


cmake/CMakeLists.txt
```diff
@@ -1066,11 +1066,17 @@ function(onnxruntime_set_compile_flags target_name)
...
-    # TODO: Narrow scope for Kleidiai compile
+
+    # TODO: Narrow scope for MLAS Kleidiai compile
     if (onnxruntime_USE_KLEIDIAI)
       target_compile_definitions(${target_name} PRIVATE USE_KLEIDIAI)
     endif()
 
+    # TODO: Narrow scope for MLAS SVE compile
+    if (onnxruntime_USE_SVE)
+      target_compile_definitions(${target_name} PRIVATE MLAS_USE_SVE)
```
snnn: But, I didn't find any other code use this macro except mlas. So, what was the problem?

hariharans29: That's what I am unsure of too. 

I recall the ARM team made a similar comment about it being challenging to scope down the KAI flag - https://github.com/microsoft/onnxruntime/pull/25187/files#r2228743076. 

It needs some more investigation. But, let us unblock the pipeline first - otherwise more issues could go in uncaught.

hariharans29: Passing pipeline runs:
Python packaging pipeline: https://aiinfra.visualstudio.com/Lotus/_build/results?buildId=955491&view=results
Python test pipeline: https://aiinfra.visualstudio.com/Lotus/_build/results?buildId=955639&view=results

hariharans29: One more sanity check:
In the above successful test runs, it is building SVE code:

image

It is important to check this as the tests had no issue with the previous setup and when `--no_sve` was turned on (i.e.) SVE code was being ignored

hariharans29: Just chatted with @edgchen1 - It is probably because the mlas test target doesn't get the compile definition in the previous version, and it would see a different platform struct definition

PR #26207, opened 2025-09-30
[webgpu] Enable indirect dispatch for flash attention

qjia7:
This pull request introduces support for indirect dispatch in the WebGPU FlashAttention implementation, enabling more dynamic and efficient kernel launches based on runtime sequence lengths. The changes add new logic and parameters to propagate sequence length information and indirect dispatch buffers through the attention pipeline, with conditional code paths to maintain compatibility with the existing direct dispatch approach.

### DIFF
--- a/onnxruntime/contrib_ops/webgpu/bert/flash_attention.h
+++ b/onnxruntime/contrib_ops/webgpu/bert/flash_attention.h
@@ -17,19 +17,24 @@ using namespace onnxruntime::webgpu;
 
 class CopyKVCacheProgram final : public Program<CopyKVCacheProgram> {
  public:
-  CopyKVCacheProgram(const std::string& kernel_name, bool has_past, bool kv_BNSH, bool past_present_share_buffer)
-      : Program{kernel_name}, has_past_(has_past), kv_BNSH_(kv_BNSH), past_present_share_buffer_(past_present_share_buffer) {
+  CopyKVCacheProgram(const std::string& kernel_name, bool has_past, bool kv_BNSH, bool past_present_share_buffer,
+                     bool prepare_indirect_dispatch = false)
+      : Program{kernel_name}, has_past_(has_past), kv_BNSH_(kv_BNSH), past_present_share_buffer_(past_present_share_buffer), prepare_indirect_dispatch_(prepare_indirect_dispatch) {
   }
 
   Status GenerateShaderCode(ShaderHelper& sh) const override;
 
   WEBGPU_PROGRAM_DEFINE_UNIFORM_VARIABLES({"copy_size", ProgramUniformVariableDataType::Uint32},
-                                          {"past_sequence_length", ProgramUniformVariableDataType::Uint32});
+                                          {"total_sequence_length", ProgramUniformVariableDataType::Uint32},
+                                          {"kv_sequence_length", ProgramUniformVariableDataType::Uint32},
+                                          {"tile_size", ProgramUniformVariableDataType::Uint32},
+                                          {"num_heads", ProgramUniformVariableDataType::Uint32});
 
  private:
   bool has_past_;
   bool kv_BNSH_;
   bool past_present_share_buffer_;
+  bool prepare_indirect_dispatch_;
 };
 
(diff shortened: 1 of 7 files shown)


onnxruntime/contrib_ops/webgpu/bert/flash_attention.h
```diff
@@ -107,19 +112,19 @@ class FlashAttentionDecodeSplitVxProgram final : public Program<FlashAttentionDe
...
                                           {"present_sequence_length", ProgramUniformVariableDataType::Uint32},
                                           {"n_reps", ProgramUniformVariableDataType::Uint32},
-                                          {"num_total_seq_length_tile", ProgramUniformVariableDataType::Uint32},
                                           {"num_present_sequence_length_tile", ProgramUniformVariableDataType::Uint32},
                                           {"num_heads", ProgramUniformVariableDataType::Uint32});
 
  private:
   uint32_t tile_size_;
   int head_size_vec_;
+  bool use_indirect_dispatch_;
```
sushraja-msft, line 121: Could you leave a comment here as to what indirect_dispatch means ? Is this an existing concept in ORT ?

onnxruntime/contrib_ops/webgpu/bert/flash_attention.h
```diff
@@ -132,11 +137,13 @@ class FlashAttentionDecodeVxReduceProgram final : public Program<FlashAttentionD
...
  private:
   uint32_t tile_size_;
+  uint32_t seq_tile_size_;
+  bool use_indirect_dispatch_;
 };
 
 Status ApplyFlashAttention(const Tensor* Q, const Tensor* K, const Tensor* V, const Tensor* attention_bias,
                            Tensor* output, const Tensor* past_key, Tensor* present_key, const Tensor* past_value, Tensor* present_value,
-                           const WebgpuAttentionParameters& parameters, onnxruntime::webgpu::ComputeContext& context);
+                           const WebgpuAttentionParameters& parameters, onnxruntime::webgpu::ComputeContext& context, const Tensor* seqlen_k = nullptr);
```
sushraja-msft, line 146: why is the tensor seqlen_k needed, doesnt the parameters already have total_sequence_length, kv_sequence_length. What new information does seqlen_k have ?

sushraja-msft: should you pass total_seqlen_tensor in as well ?

qjia7: Current GQA is registered with total_seqlen_tensor in cpu. See [.InputMemoryType(OrtMemTypeCPUInput, 6),](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/contrib_ops/webgpu/bert/group_query_attention.cc#L31). `total_sequence_length` is dynamic during decoding. But for graph capture, all commands and dispatched workgroups are only recorded once for the first token in generation. Then directly call replay for the following tokens in generation. In order to get the correct token sequence length when graph capture is enabled, I get it from `seqlen_k` by `let total_sequence_length = u32(seqlens_k[0]) + 1u` in shader since `seqlen_k` is in gpu and got updated correctly.

In the next PR, I will support dynamically register GQA to specify total_seqlen_tensor to gpu when graph capture enabled and do the corresponding refactors. I prefer this PR focus on 1) creating the indirect buffer and filling it during copykvcache 2) avoiding depending on dynamic total_sequence_length in decoding shaders and recalculating it from `seqlen_k`.

onnxruntime/contrib_ops/webgpu/bert/flash_attention.h
qjia7: indirect_dispatch means dispatch work is performed with the current GPUComputePipeline using parameters read from a GPUBuffer. That is to say we use `dispatchWorkgroupsIndirect` instead of `dispatchWorkgroups` to launch the pipeline. See https://www.w3.org/TR/webgpu/#dom-gpucomputepassencoder-dispatchworkgroupsindirect

PR #26210, opened 2025-09-30
Linux device discovery for TRT-RTX Ep

theHamsta:
### Description

This change adds vendor id and PCIe bus_id to the properties detected during Linux device discovery. 

Those properties are used to enable device discovery on Linux for the TRT-RTX execution provider.

### Motivation and Context

I want to use device discovery for TRT-EP also on Linux.

This changes have already been tested with the newly added inference samples https://github.com/microsoft/onnxruntime-inference-examples/pull/529 . Some further testing is required to see whether this repo's CI passes (therefore setting draft for now)

@gedoensmax for visibilty

### DIFF
--- a/onnxruntime/core/platform/linux/device_discovery.cc
+++ b/onnxruntime/core/platform/linux/device_discovery.cc
@@ -7,6 +7,7 @@
 #include <fstream>
 #include <iterator>
 #include <optional>
+#include <regex>
 #include <string_view>
 
 #include "core/common/common.h"
@@ -114,6 +115,28 @@ std::optional<bool> IsGpuDiscrete(uint16_t vendor_id, uint16_t device_id) {
   return std::nullopt;
 }
 
+Status GetPciBusId(const std::filesystem::path& sysfs_path, std::optional<std::string>& pci_bus_id) {
+  constexpr const char* regex_pattern{R"([0-9a-f]+:[0-9a-f]+:[0-9a-f]+[.][0-9a-f]+)"};
+  static const std::regex pci_bus_id_regex(regex_pattern);
+
+  std::error_code error_code;
+  auto pci_bus_id_path = std::filesystem::canonical(sysfs_path / "device", error_code);  // resolves symlink to PCI bus id, e.g. 0000:65:00.0
+  ORT_RETURN_IF_ERROR(ErrorCodeToStatus(error_code));
+
+  auto pci_bus_id_filename = pci_bus_id_path.filename();
+  if (std::regex_match(pci_bus_id_filename.string(), pci_bus_id_regex)) {
+    pci_bus_id = pci_bus_id_filename.string();
+  } else {
+    pci_bus_id = {};
+    LOGS_DEFAULT(WARNING) << MakeString("Skipping pci_bus_id for PCI path at \"",
+                                        pci_bus_id_path.string(),
+                                        "\" because filename \"", pci_bus_id_filename, "\" dit not match expected pattern of ",
+                                        regex_pattern);
+  };
+
+  return Status::OK();
+}
+
 Status GetGpuDeviceFromSysfs(const GpuSysfsPathInfo& path_info, OrtHardwareDevice& gpu_device_out) {
   OrtHardwareDevice gpu_device{};
   const auto& sysfs_path = path_info.path;
@@ -138,6 +161,12 @@ Status GetGpuDeviceFromSysfs(const GpuSysfsPathInfo& path_info, OrtHardwareDevic
   if (const auto is_gpu_discrete = IsGpuDiscrete(vendor_id, device_id);
       is_gpu_discrete.has_value()) {
     gpu_device.metadata.Add("Discrete", (*is_gpu_discrete ? "1" : "0"));
+  }
+
+  std::optional<std::string> pci_bus_id;
+  ORT_RETURN_IF_ERROR(GetPciBusId(sysfs_path, pci_bus_id));
+  if (pci_bus_id) {
+    gpu_device.metadata.Add("pci_bus_id", std::move(*pci_bus_id));
   }
 
   gpu_device.type = OrtHardwareDeviceType_GPU;
--- a/onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc
+++ b/onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc
@@ -584,6 +584,7 @@ struct NvTensorRtRtxEpFactory : OrtEpFactory {
    * @return True if the device is a supported NVIDIA GPU, false otherwise.
    */
   bool IsOrtHardwareDeviceSupported(const OrtHardwareDevice& device) {
+#if _WIN32
     const auto& metadata_entries = device.metadata.Entries();
     const auto it = metadata_entries.find("LUID");
     if (it == metadata_entries.end()) {
@@ -625,6 +626,25 @@ struct NvTensorRtRtxEpFactory : OrtEpFactory {
     }
 
     return false;
+#else
+    const auto& metadata_entries = device.metadata.Entries();
+    const auto it = metadata_entries.find("pci_bus_id");
+    if (it == metadata_entries.end()) {
+      return false;
+    }
+    auto& target_id = it->second;
+    int cuda_device_idx = 0;
+    if (cudaDeviceGetByPCIBusId(&cuda_device_idx, target_id.c_str()) != cudaSuccess) {
+      return false;
+    }
+
+    cudaDeviceProp prop;
+    if (cudaGetDeviceProperties(&prop, cuda_device_idx) != cudaSuccess) {
+      return false;
+    }
+    // Ampere architecture or newer is required.
+    return prop.major >= 8;
+#endif
   }
 
   // Creates and returns OrtEpDevice instances for all OrtHardwareDevices that this factory supports.
--- a/onnxruntime/test/providers/nv_tensorrt_rtx/nv_basic_test.cc
+++ b/onnxruntime/test/providers/nv_tensorrt_rtx/nv_basic_test.cc
@@ -278,7 +278,6 @@ INSTANTIATE_TEST_SUITE_P(NvExecutionProviderTest, TypeTests,
                                            ),
                          [](const testing::TestParamInfo<TypeTests::ParamType>& info) { return getTypeAsName(info.param); });
 
-#ifdef _WIN32
 static bool SessionHasEp(Ort::Session& session, const char* ep_name) {
   // Access the underlying InferenceSession.
   const OrtSession* ort_session = session;
@@ -295,7 +294,6 @@ static bool SessionHasEp(Ort::Session& session, const char* ep_name) {
 }
 
 // Tests autoEP feature to automatically select an EP that supports the GPU.
-// Currently only works on Windows.
 TEST(NvExecutionProviderTest, AutoEp_PreferGpu) {
   PathString model_name = ORT_TSTR("nv_execution_provider_auto_ep.onnx");
   std::string graph_name = "test";
@@ -305,7 +303,11 @@ TEST(NvExecutionProviderTest, AutoEp_PreferGpu) {
   CreateBaseModel(model_name, graph_name, dims);
 
   {
+#if _WIN32
     ort_env->RegisterExecutionProviderLibrary(kNvTensorRTRTXExecutionProvider, ORT_TSTR("onnxruntime_providers_nv_tensorrt_rtx.dll"));
+#else
+    ort_env->RegisterExecutionProviderLibrary(kNvTensorRTRTXExecutionProvider, ORT_TSTR("libonnxruntime_providers_nv_tensorrt_rtx.so"));
+#endif
 
     Ort::SessionOptions so;
     so.SetEpSelectionPolicy(OrtExecutionProviderDevicePolicy_PREFER_GPU);
@@ -601,8 +603,6 @@ TEST(NvExecutionProviderTest, FP4CustomOpModel) {
 
   LOGS_DEFAULT(INFO) << "[NvExecutionProviderTest] TRT FP4 dynamic quantize model run completed successfully";
 }
-
-#endif
 
 }  // namespace test
 }  // namespace onnxruntime
--- a/onnxruntime/core/providers/nv_tensorrt_rtx/version_script.lds
+++ b/onnxruntime/core/providers/nv_tensorrt_rtx/version_script.lds
@@ -2,6 +2,8 @@
 VERS_1.0 {
   global:
     GetProvider;    
+    CreateEpFactories;
+    ReleaseEpFactory;
 
   # Hide everything else.
   local:
--- a/onnxruntime/test/providers/nv_tensorrt_rtx/nv_ep_context_test.cc
+++ b/onnxruntime/test/providers/nv_tensorrt_rtx/nv_ep_context_test.cc
@@ -14,7 +14,6 @@ namespace test {
 
 RegisteredEpDeviceUniquePtr AppendTrtEtxEP(Ort::SessionOptions& session_options, std::unordered_map<std::string, std::string>& option_map) {
   RegisteredEpDeviceUniquePtr nv_tensorrt_rtx_ep;
-#ifdef _WIN32
   /// Since this test runs after other tests that use registration interface this test has to use it as well
   /// windows as otherwise the kernel registry inside the EP will not be populated. The legacy APis ony call the initialize once.
   Utils::RegisterAndGetNvTensorRtRtxEp(*ort_env, nv_tensorrt_rtx_ep);
@@ -26,9 +25,6 @@ RegisteredEpDeviceUniquePtr AppendTrtEtxEP(Ort::SessionOptions& session_options,
     }
   }
   session_options.AppendExecutionProvider_V2(*ort_env, {selected_device}, option_map);
-#else
-  session_options.AppendExecutionProvider(onnxruntime::kNvTensorRTRTXExecutionProvider, option_map);
-#endif
   return nv_tensorrt_rtx_ep;
 }
 
--- a/onnxruntime/test/providers/nv_tensorrt_rtx/test_nv_trt_rtx_ep_util.cc
+++ b/onnxruntime/test/providers/nv_tensorrt_rtx/test_nv_trt_rtx_ep_util.cc
@@ -24,7 +24,6 @@
 
 namespace onnxruntime {
 namespace test {
-#ifdef _WIN32
 
 Utils::NvTensorRtRtxEpInfo Utils::nv_tensorrt_rtx_ep_info;
 
@@ -61,7 +60,6 @@ void Utils::RegisterAndGetNvTensorRtRtxEp(Ort::Env& env, RegisteredEpDeviceUniqu
     c_api.UnregisterExecutionProviderLibrary(env, nv_tensorrt_rtx_ep_info.registration_name.c_str());
   });
 }
-#endif  // _WIN32
 
 void CreateBaseModel(const PathString& model_name,
                      std::string graph_name,


onnxruntime/core/platform/linux/device_discovery.cc
```diff
@@ -120,6 +121,7 @@ Status GetGpuDeviceFromSysfs(const GpuSysfsPathInfo& path_info, OrtHardwareDevic
 
   // metadata
   gpu_device.metadata.Add("card_idx", MakeString(path_info.card_idx));
+  gpu_device.metadata.Add("bus_id", std::filesystem::read_symlink(sysfs_path / "device").filename().string());  // e.g. 0000:65:00.0
```
edgchen1: is it possible to infer whether the GPU is discrete from the bus_id? I'm trying to figure out how to determine that - any pointers would be appreciated.

theHamsta: Hi! I don't know whether this is reliable, but it seems that the parent bus of a iGPU seems to have a parent bus with  id of pattern `picXXXX:00`. dGPU seems to have names like `card1` or higher while iGPU would be named `card0`. So it's rather the topology of the bus than the bus id of the GPU itself

dGPU
```
realpath /sys/class/drm/card1/device 
/sys/devices/pci0000:64/0000:64:00.0/0000:65:00.0
```

iGPUs would often have the parent path and be named card0

```
/sys/devices/picXXXX:00/XXXX:00:00.0/
```

I don't have a iGPU so my node 
```lspci
00:00.0 Host bridge: Intel Corporation Sky Lake-E DMI3 Registers (rev 07)
```
or 
```
/sys/devices/pci0000:00/
```
wouldn't have a any drm nodes

onnxruntime/test/providers/nv_tensorrt_rtx/nv_basic_test.cc
```diff
@@ -233,7 +232,6 @@ static bool SessionHasEp(Ort::Session& session, const char* ep_name) {
 }
 
 // Tests autoEP feature to automatically select an EP that supports the GPU.
-// Currently only works on Windows.
 TEST(NvExecutionProviderTest, AutoEp_PreferGpu) {
   PathString model_name = ORT_TSTR("nv_execution_provider_auto_ep.onnx");
```
theHamsta, line 298: this test is currently still failing: NvTensorRtRtxEpFactory::CreateEp is called but not implemented https://github.com/theHamsta/onnxruntime/blob/bc7f4bbc92ed0609e18e1c4c80c3c8d560ce1729/onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc#L698. The other test passes

onnxruntime/core/platform/linux/device_discovery.cc
edgchen1: good to know, thanks for sharing that observation

onnxruntime/test/providers/nv_tensorrt_rtx/nv_basic_test.cc
theHamsta: Where would would the internal factories on Windows come from
https://github.com/theHamsta/onnxruntime/blob/20d2c8912708b37fb914fbb0c719914a4040830c/onnxruntime/core/session/provider_policy_context.cc#L351-L358 ? I don't have them on Linux when no execution providers are appended to the session options. My desired device get selected, but NvEp::CreateEp is called.

NvEp like other EPs does not implement CreateEp.  I guess IExecutionProvider.CreateProviders is the modern version of it?

onnxruntime/core/platform/posix/env.cc
```diff
@@ -592,6 +593,25 @@ class PosixEnv : public Env {
...
+    } else {
+#if __linux__
+      return PathString(std::filesystem::read_symlink(std::filesystem::path("/proc/self/exe")).parent_path()) + "/";
+#else
+      // TODO: MacOS could use _NSGetExecutablePath, but this needs to be tested!
+      return PathString();
+#endif
+    }
+  }
+
```
theHamsta: @edgchen1 This resolves the failing test under Linux. I could alternatively put `std::filesystem::read_symlink(std::filesystem::path("/proc/self/exe")).parent_path()` in the unit test instead and move `PosixEnv::GetRuntimePath` to a separate PR.

I'm not sure whether the Windows behavior to set runtime path to current exe's dir is always desired (but it would be needed to behave the same in tests like on Windows)

onnxruntime/core/platform/posix/env.cc
```diff
@@ -592,6 +593,25 @@ class PosixEnv : public Env {
     return val == NULL ? std::string() : std::string(val);
   }
 
+  // Return the path of the executable/shared library for the current running code. This is to make it
+  // possible to load other shared libraries installed next to our core runtime code.
+  PathString GetRuntimePath() const override {
+    Dl_info dl_info{};
+    // Must be one of the symbols exported in libonnxruntime.{so,dynlib}.
+    void* symbol_from_this_library = dlsym(RTLD_DEFAULT, "OrtGetApiBase");
```
theHamsta: needs to be an exported symbol. onnxruntim_core.a does not have exported symbols by itself (we could add one though), only with onnxruntime.so

onnxruntime/core/platform/posix/env.cc
```diff
@@ -591,6 +592,46 @@ class PosixEnv : public Env {
...
+      // TODO: MacOS could use _NSGetExecutablePath, but this needs to be tested!
+      return PathString();
+#endif
+    }
+  }
+
+
+  // Return the path of the executable/shared library for the current running code. This is to make it
+  // possible to load other shared libraries installed next to our core runtime code.
+  PathString GetRuntimePath() const override {
```
edgchen1: there are two definitions of `GetRuntimePath()`

onnxruntime/core/platform/posix/env.cc
edgchen1: > I could alternatively put std::filesystem::read_symlink(std::filesystem::path("/proc/self/exe")).parent_path() in the unit test instead and move PosixEnv::GetRuntimePath to a separate PR.

maybe for this PR, just have the unit test pass in an absolute path to the plugin EP library.

onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc
```diff
@@ -616,6 +618,25 @@ struct NvTensorRtRtxEpFactory : OrtEpFactory {
...
     return false;
+#else
+    const auto& metadata_entries = device.metadata.Entries();
+    const auto it = metadata_entries.find("bus_id");
+    if (it == metadata_entries.end()) {
+      return false;
+    }
+    auto& target_id = it->second;
+    int cuda_device_idx = 0;
+    if (cudaSuccess != cudaDeviceGetByPCIBusId(&cuda_device_idx, target_id.c_str())) {
```
edgchen1: nit: for consistency
```suggestion
    if (cudaDeviceGetByPCIBusId(&cuda_device_idx, target_id.c_str()) != cudaSuccess) {
```

onnxruntime/core/platform/linux/device_discovery.cc
```diff
@@ -2,11 +2,13 @@
 // Licensed under the MIT License.
 
 #include "core/platform/device_discovery.h"
+#include "core/framework/ortdevice.h"
```
edgchen1: usually, we want to stick to this dependency ordering:

https://github.com/microsoft/onnxruntime/blob/5ef8ce2103e862a1c14166ed7a406f88b990927e/cmake/onnxruntime.cmake#L238-L254

so onnxruntime_common (which contains core/platform/linux/device_discovery.cc) should not depend on onnxruntime_framework (core/framework/ortdevice.h).

we can move the vendor id enum and helper function to be a separate header, in onnxruntime_common perhaps, or just move the helper function to this file for now.

onnxruntime/test/providers/nv_tensorrt_rtx/nv_basic_test.cc
edgchen1: a plugin EP should implement CreateEp. the internal factory is not used for plugin EPs.

onnxruntime/core/platform/posix/env.cc
edgchen1: actually, I think it would be useful to implement `PosixEnv::GetRuntimePath()` too. it would also help fix another test on Linux (only enabled for Windows now).

https://github.com/microsoft/onnxruntime/blob/b3ba580346293ed1cab35b59cf93ae1da56e65f7/onnxruntime/test/autoep/test_selection.cc#L171-L175

I think we can just include it in this PR.

theHamsta: I created https://github.com/microsoft/onnxruntime/pull/26424 since enabling the tests with CUDA EP required another small tweak. Let's do this PR after enabling the PosixEnv::GetRuntimePath

onnxruntime/core/platform/linux/device_discovery.cc
theHamsta: I've put the helper to the linux file. I will try to generalize in a follow-up.

theHamsta:
With https://github.com/microsoft/onnxruntime/pull/26344 merged, @edgchen1  is there anything missing here?

onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc
```diff
@@ -616,6 +618,25 @@ struct NvTensorRtRtxEpFactory : OrtEpFactory {
     }
 
     return false;
+#else
+    const auto& metadata_entries = device.metadata.Entries();
+    const auto it = metadata_entries.find("bus_id");
```
edgchen1: can we name it "pci_bus_id"?

onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc
```diff
@@ -616,6 +618,25 @@ struct NvTensorRtRtxEpFactory : OrtEpFactory {
...
+    if (it == metadata_entries.end()) {
+      return false;
+    }
+    auto& target_id = it->second;
+    int cuda_device_idx = 0;
+    if (cudaDeviceGetByPCIBusId(&cuda_device_idx, target_id.c_str()) != cudaSuccess) {
```
edgchen1, line 637: just wondering, is it possible to get the PCI bus ID on Windows too? if so, could we just have a single code path here?

onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc
theHamsta: I just fixed that.

onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc
theHamsta: I assume this would be possible. I asked my local AI that one get the ids that form this string individually, starting from getting the LUID.

LUID is just the gold standard on Windows and determined with a single API call. This PCI string with its formatting seems to be Linux-specific. The CUDA call should work the same on Windows though

onnxruntime/core/platform/linux/device_discovery.cc
```diff
@@ -114,6 +115,27 @@ std::optional<bool> IsGpuDiscrete(uint16_t vendor_id, uint16_t device_id) {
   return std::nullopt;
 }
 
+constexpr static const char* VendorIdToString(OrtDevice::VendorId vendorId) {
```
edgchen1: this is implicitly depending on `include/onnxruntime/core/framework/ortdevice.h` which is not ideal (see this [comment](https://github.com/microsoft/onnxruntime/pull/26210#discussion_r2450054530)). is setting the vendor name necessary for this PR? if not, maybe we can refine the vendor ID/name setup in another PR.

onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc
edgchen1: from the [cudaDeviceGetByPCIBusId() docs](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DEVICE.html#group__CUDART__DEVICE_1g65f57fb8d0981ca03f6f9b20031c3e5d), the pciBusId argument is a "String in one of the following forms: [domain]:[bus]:[device].[function] [domain]:[bus]:[device] [bus]:[device].[function]".

do you think it would be worth trying to do this device lookup in a cross-platform way? can consider it for another PR.

onnxruntime/core/platform/linux/device_discovery.cc
theHamsta: It's not necessary. I stripped it out from this PR. This line is now the only change outside of NV-EP https://github.com/microsoft/onnxruntime/pull/26210/commits/0f89e15b8ce1dc139dfa997572596fcf7e17c97f

onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc
theHamsta: What would the benefit of having this in a cross-platform way? The goal should be to have the most reliable way which is easy with LUID. LUID exists in DXGI, CUDA, Vulkan and it guaranteed to be always available on Windows, for virtual and physical devices.

The usual replacement on Linux is "UUID" which is Vulkan/CUDA/NVML concept https://registry.khronos.org/VulkanSC/specs/1.0-extensions/man/html/VkPhysicalDeviceIDProperties.html. Opening a Vulkan device would be too much overhead for device discovery, especially since we already have the sysfs infos.

I found this piece of info from the Microsoft documentation 
https://learn.microsoft.com/en-us/windows-hardware/drivers/pci/obtaining-pci-segment-values so it should be possible to parse and reformat this string. I find this more brittle than using LUID. We could still add this to add the `pci_bus_id` on Windows, too (e.g. if user want to match a OrtHardwareDevice with a NVML device https://docs.nvidia.com/deploy/nvml-api/index.html, even on Windows)

edgchen1: > What would the benefit of having this in a cross-platform way?

generally speaking, less code to maintain and fewer ifdefs.

I'm not familiar with this area. if it is much less convenient to get the same information on Windows then perhaps it is not worth it.

onnxruntime/core/platform/linux/device_discovery.cc
```diff
@@ -140,6 +140,8 @@ Status GetGpuDeviceFromSysfs(const GpuSysfsPathInfo& path_info, OrtHardwareDevic
     gpu_device.metadata.Add("Discrete", (*is_gpu_discrete ? "1" : "0"));
   }
 
+  gpu_device.metadata.Add("pci_bus_id", std::filesystem::read_symlink(sysfs_path / "device").filename().string());  // e.g. 0000:65:00.0
```
edgchen1: is it possible for `std::filesystem::read_symlink` to fail here? if so, could we use the non-throwing overload and have this function return a failure status instead?

we have some tolerance of GPU device discovery failure here, but it doesn't handle exceptions.
https://github.com/microsoft/onnxruntime/blob/a19954cd9a0d666dcb438e364ce1758f3548369c/onnxruntime/core/platform/linux/device_discovery.cc#L176-L183

theHamsta: I added error checking by extracting this to a separate function. I made the error non-fatal for device discovery and just warn when the PCI bus id can't determined.

onnxruntime/core/platform/linux/device_discovery.cc
```diff
@@ -140,6 +152,11 @@ Status GetGpuDeviceFromSysfs(const GpuSysfsPathInfo& path_info, OrtHardwareDevic
     gpu_device.metadata.Add("Discrete", (*is_gpu_discrete ? "1" : "0"));
   }
 
+  std::string pci_bus_id;
+  if (GetPciBusId(sysfs_path, pci_bus_id).IsOK()) {
```
edgchen1: why does the failing status not get propagated to the caller here? this is inconsistent with other calls to status-returning functions in this function. if there's a good reason for it, it would be worth a comment.

theHamsta: We could propagate the failure upwards. My reasoning was that pci_bus_id is currently an optional property only used for trt-rtx at the moment. Failing pci_bus_id should therefore only fail for trt-rtx and not impact other EPs. We could consider it as something that we expect for every device and consider it an error condition when it fails.

What is your preference?

onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc
theHamsta: > I'm not familiar with this area. if it is much less convenient to get the same information on Windows then perhaps it is not worth it.

I mostly worried about reliability in edgecases. We will never test this code for all setups (HW/OS version/VM<->physical), so I'm trying to call as few functions as possible. The LUID matching is the recommend approach in the CTK programming, sadly it's not available for Linux. PCI bus ID is just something that I felt that we already have with our current Linux device discovery.

onnxruntime/core/platform/linux/device_discovery.cc
edgchen1: I see. I suppose it depends on whether we consider it to be optional. if so, maybe having `GetPciBusId()` return a `std::optional<std::string>` would be appropriate.

> We could consider it as something that we expect for every device and consider it an error condition when it fails.

I don't have much experience here. do you think that this is a valid expectation? this would only be for `/sys/class/drm/cardN` paths.

onnxruntime/core/platform/linux/device_discovery.cc
```diff
@@ -114,6 +114,18 @@ std::optional<bool> IsGpuDiscrete(uint16_t vendor_id, uint16_t device_id) {
   return std::nullopt;
 }
 
+Status GetPciBusId(const std::filesystem::path& sysfs_path, std::string& pci_bus_id) {
+  std::error_code error_code;
+  auto pci_bus_id_path = std::filesystem::read_symlink(sysfs_path / "device", error_code);  // e.g. 0000:65:00.0
```
edgchen1: > // e.g. 0000:65:00.0

would it be worth validating that the path filename matches the expected format?

theHamsta: I added a regex-based validation. if validation fails, I considered it an error, so we would know which other unknown cases there could be for this sysfs path.

onnxruntime/core/platform/linux/device_discovery.cc
theHamsta: I currently have two error conditions:
- sysfs_path / device is not readable (path existing already assumed in other device discovery checks)
- read_link(sysfs_path / device) does not match our expected pattern (we don't know under which circumstances this could occur.

I currently don't use the `std::optional::null_opt` but made the function return an optional. We assume that there might be legitimate reasons a device might not have a pci_bus_id, but we don't know yet under which conditions this can occur. So not sure whether the function should return `std::string` or `optional<std::string>`

onnxruntime/core/platform/linux/device_discovery.cc
```diff
@@ -114,6 +115,23 @@ std::optional<bool> IsGpuDiscrete(uint16_t vendor_id, uint16_t device_id) {
...
 }
 
+Status GetPciBusId(const std::filesystem::path& sysfs_path, std::optional<std::string>& pci_bus_id) {
+  constexpr const char* regex_pattern{R"(\d+:\d+:\d+[.]\d+)"};
+  static const std::regex pci_bus_id_regex(regex_pattern);
+
+  std::error_code error_code;
+  auto pci_bus_id_path = std::filesystem::read_symlink(sysfs_path / "device", error_code);  // e.g. 0000:65:00.0
+
+  ORT_RETURN_IF_ERROR(ErrorCodeToStatus(error_code));
```
theHamsta, line 124: ErrorCodeToStatus is very unspecific. People getting this status would have no idea where this is coming from.

I suggest adding at least the filesystem path as a context `ErrorCodeToStatus` and even better additionally a context why this file was accessed to format a better error message. Possible follow-up for this PR

onnxruntime/core/platform/linux/device_discovery.cc
```diff
@@ -114,6 +115,23 @@ std::optional<bool> IsGpuDiscrete(uint16_t vendor_id, uint16_t device_id) {
   return std::nullopt;
 }
 
+Status GetPciBusId(const std::filesystem::path& sysfs_path, std::optional<std::string>& pci_bus_id) {
+  constexpr const char* regex_pattern{R"(\d+:\d+:\d+[.]\d+)"};
```
edgchen1: I think this would need to match hex digits, i.e., `[0-9a-f]` instead of `\d` (perhaps also upper-case A-F).

I encountered this on one system: `000e:00:00.0`

onnxruntime/core/platform/linux/device_discovery.cc
```diff
@@ -114,6 +115,23 @@ std::optional<bool> IsGpuDiscrete(uint16_t vendor_id, uint16_t device_id) {
...
+  static const std::regex pci_bus_id_regex(regex_pattern);
+
+  std::error_code error_code;
+  auto pci_bus_id_path = std::filesystem::read_symlink(sysfs_path / "device", error_code);  // e.g. 0000:65:00.0
+  ORT_RETURN_IF_ERROR(ErrorCodeToStatus(error_code));
+
+  if (!std::regex_match(pci_bus_id_path.string(), pci_bus_id_regex)) {
```
edgchen1: `std::regex_match` matches against the whole string, right? but the pattern is only for the last path component. should it match against `pci_bus_id_path.filename().string()` instead?

onnxruntime/core/platform/linux/device_discovery.cc
yuslepukhin: Would using std::filesystem::canonical or weakly_canonical be a better way to validate the path?

onnxruntime/core/platform/linux/device_discovery.cc
theHamsta: oh yes, stupid mistake! Also we want to validate the filename, not the whole path!

onnxruntime/core/platform/linux/device_discovery.cc
theHamsta: sounds like it does exactly what we want. I changed it.

read_symlink just seems to be a bit more self-documenting since resolving that specifc symlink was the actual intention. but a comment can bring this intention back.

onnxruntime/core/platform/linux/device_discovery.cc
theHamsta: yes, just saw that. Sorry, I should do this PR as side task while being distracted by something else.

gedoensmax:
This fixes #26705.

gedoensmax:
@yuslepukhin can we still get this into 1.24 ? Seems like it has been waiting for quite some time.

onnxruntime/core/platform/linux/device_discovery.cc
```diff
@@ -114,6 +115,29 @@ std::optional<bool> IsGpuDiscrete(uint16_t vendor_id, uint16_t device_id) {
...
+
+  auto pci_bus_id_filename = pci_bus_id_path.filename();
+  if (!std::regex_match(pci_bus_id_filename.string(), pci_bus_id_regex)) {
+    return Status{common::StatusCategory::ONNXRUNTIME, common::StatusCode::FAIL,
+                  MakeString("Expected PCI path at \"",
+                             pci_bus_id_path.string(),
+                             "\" did not match expected pattern of ",
+                             regex_pattern,
+                             " for the filename \"",
+                             pci_bus_id_filename, "\"")};
```
edgchen1: nit: [`path::operator<<`](https://en.cppreference.com/w/cpp/filesystem/path/operator_ltltgtgt.html) should handle quoting already, so this could simply be
```suggestion
                  MakeString("Expected PCI path at ",
                             pci_bus_id_path,
                             " did not match expected pattern of ",
                             regex_pattern,
                             " for the filename ",
                             pci_bus_id_filename)};
```

edgchen1 (dismissed):

yuslepukhin (dismissed):
:shipit:

onnxruntime/core/platform/linux/device_discovery.cc
```diff
@@ -114,6 +115,29 @@ std::optional<bool> IsGpuDiscrete(uint16_t vendor_id, uint16_t device_id) {
...
+  constexpr const char* regex_pattern{R"([0-9a-f]+:[0-9a-f]+:[0-9a-f]+[.][0-9a-f]+)"};
+  static const std::regex pci_bus_id_regex(regex_pattern);
+
+  std::error_code error_code;
+  auto pci_bus_id_path = std::filesystem::canonical(sysfs_path / "device", error_code);  // resolves symlink to PCI bus id, e.g. 0000:65:00.0
+  ORT_RETURN_IF_ERROR(ErrorCodeToStatus(error_code));
+
+  auto pci_bus_id_filename = pci_bus_id_path.filename();
+  if (!std::regex_match(pci_bus_id_filename.string(), pci_bus_id_regex)) {
+    return Status{common::StatusCategory::ONNXRUNTIME, common::StatusCode::FAIL,
```
edgchen1: from a failed CI build:
`1: 2026-01-20 20:00:44.635509407 [W:onnxruntime:Default, device_discovery.cc:212 DiscoverDevicesForPlatform] GPU device discovery failed: Expected PCI path at "/sys/devices/LNXSYSTM:00/LNXSYBUS:00/ACPI0004:00/VMBUS:00/5620e0c7-8062-4dce-aeb7-520c7ef76171" did not match expected pattern of [0-9a-f]+:[0-9a-f]+:[0-9a-f]+[.][0-9a-f]+ for the filename ""5620e0c7-8062-4dce-aeb7-520c7ef76171""`

maybe we make the `regex_match` failure return an empty optional instead of failing? good error message though, maybe log that instead.

tianleiwu:
I saw a test failure:
9: [ RUN      ] AutoEpSelection.CudaEP
9: /onnxruntime_src/onnxruntime/test/autoep/test_selection.cc:48: Failure
9: Value of: !devices.empty()
9:   Actual: false
9: Expected: true
9: No devices found with EP name of CUDAExecutionProvider
9: 
9: unknown file: Failure
9: C++ exception with description "Must provide one or more OrtEpDevice instances." thrown in the test body.
9: 
9: [  FAILED  ] AutoEpSelection.CudaEP (144 ms)

theHamsta:
> I saw a test failure: 9: [ RUN ] AutoEpSelection.CudaEP 9: /onnxruntime_src/onnxruntime/test/autoep/test_selection.cc:48: Failure 9: Value of: !devices.empty() 9: Actual: false 9: Expected: true 9: No devices found with EP name of CUDAExecutionProvider 9: 9: unknown file: Failure 9: C++ exception with description "Must provide one or more OrtEpDevice instances." thrown in the test body. 9: 9: [ FAILED ] AutoEpSelection.CudaEP (144 ms)

Thank you 

> I saw a test failure: 9: [ RUN ] AutoEpSelection.CudaEP 9: /onnxruntime_src/onnxruntime/test/autoep/test_selection.cc:48: Failure 9: Value of: !devices.empty() 9: Actual: false 9: Expected: true 9: No devices found with EP name of CUDAExecutionProvider 9: 9: unknown file: Failure 9: C++ exception with description "Must provide one or more OrtEpDevice instances." thrown in the test body. 9: 9: [ FAILED ] AutoEpSelection.CudaEP (144 ms)

Thank you for reviewing! I'll have a look onto the test failures next week.

onnxruntime/core/platform/linux/device_discovery.cc
theHamsta: I made it optional and will ask about the specific of this PCI bus format. It still seems to be some kind of bus although virtual

edgchen1 (approved):

onnxruntime/core/platform/linux/device_discovery.cc
edgchen1: we can fix this later.

edgchen1:
> > I saw a test failure: 9: [ RUN ] AutoEpSelection.CudaEP 9: /onnxruntime_src/onnxruntime/test/autoep/test_selection.cc:48: Failure 9: Value of: !devices.empty() 9: Actual: false 9: Expected: true 9: No devices found with EP name of CUDAExecutionProvider 9: 9: unknown file: Failure 9: C++ exception with description "Must provide one or more OrtEpDevice instances." thrown in the test body. 9: 9: [ FAILED ] AutoEpSelection.CudaEP (144 ms)
> 
> Thank you
> 
> > I saw a test failure: 9: [ RUN ] AutoEpSelection.CudaEP 9: /onnxruntime_src/onnxruntime/test/autoep/test_selection.cc:48: Failure 9: Value of: !devices.empty() 9: Actual: false 9: Expected: true 9: No devices found with EP name of CUDAExecutionProvider 9: 9: unknown file: Failure 9: C++ exception with description "Must provide one or more OrtEpDevice instances." thrown in the test body. 9: 9: [ FAILED ] AutoEpSelection.CudaEP (144 ms)
> 
> Thank you for reviewing! I'll have a look onto the test failures next week.

I think the test failure was due to the PCI id regex match failure which prevented it from detecting the device. the CI builds have passed now since you've made the regex match optional.

edgchen1:
@theHamsta thanks for the contribution!

bwakeford9-cyber (commented):
ok

PR #26220, opened 2025-10-02
upgrade torch version for component  governance

vraspar:
### Description
Upgrade torch version to address component governance alert

### DIFF
--- a/docs/python/requirements.txt
+++ b/docs/python/requirements.txt
@@ -21,4 +21,4 @@ onnx
 sphinx_exec_code
 sphinx_tabs
 furo
-torch
+torch >= 2.6.0


docs/python/requirements.txt
```diff
@@ -21,4 +21,4 @@ onnx
 sphinx_exec_code
 sphinx_tabs
 furo
-torch
+torch >= 2.6.0
```
snnn, line 24: Why is this change helpful? Would it by default install the latest version?

vraspar: We are still seeing older version of torch in some pipeline stage. I thought updating it would fix the issue https://aiinfra.visualstudio.com/Lotus/_componentGovernance/218239/alert/13216243?typeId=25422597

snnn (approved):

docs/python/requirements.txt
snnn: You are right.  It's because we fetch pytorch from an ADO feed. And that version is not updated. I am not sure why.

PR #26222, opened 2025-10-02
Fix minimal build (make it build with the recent changes)

yuslepukhin:
This pull request introduces a minor update to the test infrastructure and test files, focusing on improving build configuration and test utility usage.

Build configuration:

* Updated `cmake/onnxruntime_test_pch.cmake` to only enable precompiled headers for test targets when not performing a minimal build, preventing unnecessary PCH usage in minimal build scenarios.

Test utility improvements:

* Added an include for `asserts.h` in `onnxruntime/test/platform/file_io_test.cc`, ensuring test assertions are available and improving test reliability.

### DIFF
--- a/cmake/onnxruntime_test_pch.cmake
+++ b/cmake/onnxruntime_test_pch.cmake
@@ -2,12 +2,15 @@
 
 if(CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
   # Visual Studio PCH
-  target_precompile_headers(onnxruntime_test_all PRIVATE
-    "${CMAKE_CURRENT_SOURCE_DIR}/test_pch.h"
-  )
-  target_precompile_headers(onnxruntime_provider_test PRIVATE
-    "${CMAKE_CURRENT_SOURCE_DIR}/test_pch.h"
-  )
+  if(NOT onnxruntime_MINIMAL_BUILD)
+    target_precompile_headers(onnxruntime_test_all PRIVATE
+      "${CMAKE_CURRENT_SOURCE_DIR}/test_pch.h"
+    )
+
+    target_precompile_headers(onnxruntime_provider_test PRIVATE
+      "${CMAKE_CURRENT_SOURCE_DIR}/test_pch.h"
+   )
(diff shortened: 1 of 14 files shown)


snnn (approved):

cmake/onnxruntime_test_pch.cmake
```diff
@@ -2,12 +2,14 @@
...
-    "${CMAKE_CURRENT_SOURCE_DIR}/test_pch.h"
-  )
+  if(NOT onnxruntime_MINIMAL_BUILD)
+    target_precompile_headers(onnxruntime_test_all PRIVATE
+      "${CMAKE_CURRENT_SOURCE_DIR}/test_pch.h"
+    )
+    target_precompile_headers(onnxruntime_provider_test PRIVATE
+      "${CMAKE_CURRENT_SOURCE_DIR}/test_pch.h"
+   )
+  endif()
```
edgchen1: It looks like the issue is that `onnxruntime_provider_test` might not exist as a target. whether that target is added also depends on `onnxruntime_REDUCED_OPS_BUILD`.

https://github.com/microsoft/onnxruntime/blob/8af9f580860b2bd4ecd3df242ad6e1650e8b9066/cmake/onnxruntime_unittests.cmake#L1202

it would be more robust to check for the existence of the target.

```suggestion
  target_precompile_headers(onnxruntime_test_all PRIVATE
    "${CMAKE_CURRENT_SOURCE_DIR}/test_pch.h"
  )
  if(TARGET onnxruntime_provider_test)
    target_precompile_headers(onnxruntime_provider_test PRIVATE
      "${CMAKE_CURRENT_SOURCE_DIR}/test_pch.h"
    )
  endif()
```

edgchen1: I tried this and made some other fixes along the way to get it building.

https://github.com/microsoft/onnxruntime/commit/3915d7291de5b9ab8254f0e28a79c26a1dd0533d

edgchen1 (approved):

PR #26223, opened 2025-10-02
Build WebGPU on Windows arm64

baijumeswani:
Adds ability to build webgpu ep on win-arm64 for the custom nuget package pipeline.

### DIFF
--- a/tools/ci_build/github/azure-pipelines/templates/jobs/download_win_webgpu_build_tools.yml
+++ b/tools/ci_build/github/azure-pipelines/templates/jobs/download_win_webgpu_build_tools.yml
@@ -1,0 +1,20 @@
+steps:
+  - powershell: |
+      azcopy.exe cp --recursive https://lotusscus.blob.core.windows.net/models/webgpu-build-tools $(Agent.TempDirectory)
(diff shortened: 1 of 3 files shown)


tools/ci_build/github/azure-pipelines/templates/jobs/download_win_webgpu_build_tools.yml
```diff
@@ -0,0 +1,20 @@
+steps:
+  - powershell: |
+      azcopy.exe cp --recursive https://lotusscus.blob.core.windows.net/models/webgpu-build-tools $(Agent.TempDirectory)
```
snnn, line 3: Let's find a better way to do this.

fs-eire: I think it should be good for now. the table gen executable is not version sensitive anyway.

In long term, it would be great if we can use the dawn's version (commit SHA1) as key (eg. path part) to avoid any potential inconsistency.

baijumeswani:
This PR is not needed anymore since we're not intending to support webgpu ep on win-arm64 for fl at the moment.

Issue #26218, opened 2025-10-02
ConstantOfShape on scalar input


justinchuby:
https://onnx.ai/onnx/operators/onnx__ConstantOfShape.html#attributes states that ConstantOfShape takes a single element value, but ORT seems to require that the value to be 1d.

yuslepukhin:
Would you like to make a PR and modify the check for num of dims to be 0 or 1?

justinchuby:
Yes I can do that. Could you point me to where I should change? Do we need to update every ep or is there a central check?

yuslepukhin:
https://github.com/microsoft/onnxruntime/blob/8af9f580860b2bd4ecd3df242ad6e1650e8b9066/onnxruntime/core/providers/cpu/generator/constant_of_shape_base.h?plain=1#L80C1-L80C80

justinchuby:
Thanks - created https://github.com/microsoft/onnxruntime/pull/26227

PR #26227, opened 2025-10-03
Allow any dimensional tensor for ConstantOfShape

justinchuby:
Updated enforcement conditions to allow any dimensional tensor with a single element according to the ONNX spec.

Fix https://github.com/microsoft/onnxruntime/issues/26218

### DIFF
--- a/onnxruntime/core/providers/cpu/generator/constant_of_shape_base.h
+++ b/onnxruntime/core/providers/cpu/generator/constant_of_shape_base.h
@@ -78,8 +78,7 @@ class ConstantOfShapeBase {
     auto* t_proto_p = t_proto.get();
 #endif
     if (info.GetAttr<ONNX_NAMESPACE::TensorProto>("value", t_proto_p).IsOK()) {
-      ORT_ENFORCE(t_proto_p->dims_size() == 1, "Must have a single dimension");
-      ORT_ENFORCE(t_proto_p->dims()[0] == 1, "Must have a single dimension of 1");
+      // value can be any dimentional. It only needs to be a single-element tensor.
       SetValueFromTensorProto(*t_proto_p);
     } else {
       float f_value = 0.f;


onnxruntime/core/providers/cpu/generator/constant_of_shape_base.h
```diff
@@ -78,8 +78,7 @@ class ConstantOfShapeBase {
     auto* t_proto_p = t_proto.get();
 #endif
     if (info.GetAttr<ONNX_NAMESPACE::TensorProto>("value", t_proto_p).IsOK()) {
-      ORT_ENFORCE(t_proto_p->dims_size() == 1, "Must have a single dimension");
-      ORT_ENFORCE(t_proto_p->dims()[0] == 1, "Must have a single dimension of 1");
+      // value can be any dimensional. It only needs to be a single-element tensor.
```
yuslepukhin: The value should be either scalar or a dim[0] == 1.
The spec says: a single element tensor. #Closed

justinchuby: Updated check

yuslepukhin: Okay, but the check cannot be just a comment. Perhaps, we can compute the size and ensure it is 1.

justinchuby: I updated to use utils::GetTensorShapeFromTensorProto. Does that look right?

onnxruntime/core/providers/cpu/generator/constant_of_shape_base.h
```diff
@@ -78,8 +78,7 @@ class ConstantOfShapeBase {
     auto* t_proto_p = t_proto.get();
 #endif
     if (info.GetAttr<ONNX_NAMESPACE::TensorProto>("value", t_proto_p).IsOK()) {
-      ORT_ENFORCE(t_proto_p->dims_size() == 1, "Must have a single dimension");
-      ORT_ENFORCE(t_proto_p->dims()[0] == 1, "Must have a single dimension of 1");
+      ORT_ENFORCE(utils::GetTensorShapeFromTensorProto(t_proto_p).Size(), "The value attribute of ConstantOfShape must be a single-element tensor");
```
yuslepukhin: The size must be equal to 1.
Also I would actually adjust the title of the PR. #Resolved

justinchuby: Sorry thanks - for some reason I missed the comparison when editing

onnxruntime/core/providers/cpu/generator/constant_of_shape_base.h
```diff
@@ -78,8 +78,7 @@ class ConstantOfShapeBase {
     auto* t_proto_p = t_proto.get();
 #endif
     if (info.GetAttr<ONNX_NAMESPACE::TensorProto>("value", t_proto_p).IsOK()) {
-      ORT_ENFORCE(t_proto_p->dims_size() == 1, "Must have a single dimension");
-      ORT_ENFORCE(t_proto_p->dims()[0] == 1, "Must have a single dimension of 1");
+      ORT_ENFORCE(utils::GetTensorShapeFromTensorProto(t_proto).Size() == 1, "The value attribute of ConstantOfShape must be a single-element tensor");
```
justinchuby: @yuslepukhin the line below de-references t_proto_p but I think we can pass in t_proto directly here because the method takes a reference and t_proto_p points to t_proto? Is there a preference? Thanks #Resolved

yuslepukhin: The reason for this `#ifdef` is that the header is being used in both CPU code and DLL based providers, and they always have to created the instance on the heap. Thus, the correct way to pass this would be to `*t_proto_p`.
However, we need to verify that the above utlity function is present in the provider bridge. I ran a quick search and it is not present. So, the choice is either to wire it in the provider bridge, which is more code, or insert a simple std::accumulate on dims from tensor_proto. I suggest you check that this tensor_proto has_shape() first to account for scalars.

onnxruntime/core/providers/cpu/generator/constant_of_shape_base.h
```diff
@@ -78,8 +78,14 @@ class ConstantOfShapeBase {
     auto* t_proto_p = t_proto.get();
 #endif
     if (info.GetAttr<ONNX_NAMESPACE::TensorProto>("value", t_proto_p).IsOK()) {
-      ORT_ENFORCE(t_proto_p->dims_size() == 1, "Must have a single dimension");
-      ORT_ENFORCE(t_proto_p->dims()[0] == 1, "Must have a single dimension of 1");
+      ORT_ENFORCE(
+          t_proto_p->dims_size() == 0 ||
+              std::accumulate(
+                  t_proto_p->dims().begin(),
```
justinchuby: @yuslepukhin could you suggest what the proper way of accumulating t_proto_p->dims() is? It seems dims is `onnx::int64s` and I am not sure what is the conventional way of working with it. #Resolved

yuslepukhin: int64_t size = 1;
      for (auto dim : t_proto_p->dims())
          size *= dim;

onnxruntime/core/providers/cpu/generator/constant_of_shape_base.h
```diff
@@ -78,8 +78,14 @@ class ConstantOfShapeBase {
...
+                  t_proto_p->dims().end(),
+                  1,
```
snnn: The latest change correctly handles both scalar tensors (dims_size() == 0) and tensors with a single element by calculating the product of dimensions. This aligns with the ONNX spec and addresses the feedback from the review comments. The use of `std::accumulate` is functionally correct.

snnn:
I agree with the suggestion to use a range-based for loop. Since we are using C++17, a range-based for loop would be more modern and readable than `std::accumulate` for calculating the product of dimensions. 

For example:
```cpp
int64_t size = 1;
for (auto dim : t_proto_p->dims()) {
    size *= dim;
}
```
This is also in line with suggestions from other reviewers.

onnxruntime/core/providers/cpu/generator/constant_of_shape_base.h
justinchuby: Thanks

onnxruntime/core/providers/cpu/generator/constant_of_shape_base.h
```diff
@@ -78,8 +78,11 @@ class ConstantOfShapeBase {
     auto* t_proto_p = t_proto.get();
 #endif
     if (info.GetAttr<ONNX_NAMESPACE::TensorProto>("value", t_proto_p).IsOK()) {
-      ORT_ENFORCE(t_proto_p->dims_size() == 1, "Must have a single dimension");
-      ORT_ENFORCE(t_proto_p->dims()[0] == 1, "Must have a single dimension of 1");
+      int64_t size = 1;
```
edgchen1: should we account for the possibility of overflow here (e.g., with `SafeInt<int64_t>`? alternatively, if we want to verify that the product is 1, we could just check that all dimensions are equal to 1.

justinchuby: Thanks - that sounds like a good idea: hopefully ORT_ENFORCE doesn't have much of an overhead?

edgchen1: > hopefully ORT_ENFORCE doesn't have much of an overhead?

I think `ORT_ENFORCE` is fine there, it's basically `if(!condition) { throw ...; }`

onnxruntime/core/providers/cpu/generator/constant_of_shape_base.h
```diff
@@ -78,8 +78,9 @@ class ConstantOfShapeBase {
     auto* t_proto_p = t_proto.get();
 #endif
     if (info.GetAttr<ONNX_NAMESPACE::TensorProto>("value", t_proto_p).IsOK()) {
-      ORT_ENFORCE(t_proto_p->dims_size() == 1, "Must have a single dimension");
-      ORT_ENFORCE(t_proto_p->dims()[0] == 1, "Must have a single dimension of 1");
+      for (auto dim : t_proto_p->dims()) {
```
edgchen1, line 81: hm, the issue is that the shared provider version of `onnx::int64s` doesn't define enough to be used in the range for. that would require `begin()` and `end()`.

https://github.com/microsoft/onnxruntime/blob/535f1f37c410575ecbbb47de8f902a137e498869/onnxruntime/core/providers/shared_library/provider_wrappedtypes.h#L75-L82

alternatives - make shared provider version of `onnx::int64s` more ergonomic, or do things the old-fashioned way...
```c++
for (int i = 0; i < t_proto_p->dims().size(); ++i) {
  auto dim = t_proto_p->dims()[i];
  ...
}
```

justinchuby: Is the idea implementing begin and end in struct int64s?

Issue #26181, opened 2025-09-26
no kernel image is available for execution on the device [rtx 5090 laptop, wan2.2 animate, DWPreprocessor, onnxruntime-gpu]


igor-iki:
Hello. 
at first I've created thread on ComfyUI [https://github.com/comfyanonymous/ComfyUI/issues/10028](url)
There I found out that other people have the same issue and it's related to onnxruntime

**error**
` [ONNXRuntimeError] : 1 : FAIL : Non-zero status code returned while running Slice node. Name:'Slice_34' Status Message: CUDA error cudaErrorNoKernelImageForDevice:no kernel image is available for execution on the device`

So I have installed:
- latest pytorch with cu129 (tried also 128) 
- cuda toolkit
- latest onnxruntime-gpu
- latest nvidia gpu driver

Some people said that it's probably problem between onnxruntime and rtx50  GPU series

Comfy support also replied me with next words 

> means that ONNX Runtime’s GPU kernels are not fully compatible with your current GPU/driver/PyTorch setup.

`# ComfyUI Error Report
## Error Details
- **Node ID:** 233
- **Node Type:** DWPreprocessor
- **Exception Type:** onnxruntime.capi.onnxruntime_pybind11_state.Fail
- **Exception Message:** [ONNXRuntimeError] : 1 : FAIL : Non-zero status code returned while running Slice node. Name:'Slice_34' Status Message: CUDA error cudaErrorNoKernelImageForDevice:no kernel image is available for execution on the device

## Stack Trace

  File "C:\comfy\ComfyUI\execution.py", line 496, in execute
    output_data, output_ui, has_subgraph, has_pending_tasks = await get_output_data(prompt_id, unique_id, obj, input_data_all, execution_block_cb=execution_block_cb, pre_execute_cb=pre_execute_cb, hidden_inputs=hidden_inputs)
                                                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "C:\comfy\ComfyUI\execution.py", line 315, in get_output_data
[...]

PR #26002, opened 2025-09-09
Reduce Python and Nuget GPU package size

chilo-ms:
### Description
The package size limit for PyPI and Nuget are:
- python package size under 300MB
- Nuget package size under 250MB

To meet the size limit, 
this PR firstly removes some old GPU arch support in CMAKE_CUDA_ARCHITECTURE.
Secondly, it removes the FPA_INTB_GEMM support in Linux Python wheel.

#### Python wheel

| OS      | cmake_cuda_architecture                                | CUDA kernel removal |Package size | Under 300MB|
|---------|--------------------------------------------------------|-|-------------|---|
| Linux   | 60-real;70-real;75-real;80-real;86-real;90a-real;90a-virtual | |341 MB       |No (original)|
| Linux   | 70-real;75-real;80-real;86-real;90a-real;90a-virtual   | | 329 MB       |No|
| Linux   | 75-real;80-real;86-real;90a-real;90a-virtual   | |319 MB       |No|
| Linux   | 80-real;86-real;90a-real;90a-virtual   | |304 MB       |No|
| Linux   | 60-real;70-real;75-real;80-real;86-real;90a-real;90a-virtual. | FPA_INTB_GEMM|287 MB       |Yes|
| Windows | 52-real;61-real;75-real;86-real;89-real;90a-virtual    | | 272 MB       |Yes (original)|

#### Nuget

| OS      | cmake_cuda_architecture                                | CUDA kernel removal |Package size |Under 250MB|
|---------|--------------------------------------------------------|---|--------------|---|
| Linux   | 60-real;70-real;75-real;80-real;90a-real;90a-virtual | |276 MB       |No (original)|
| Linux   | 75-real;80-real;90a-real;90a-virtual   | |253 MB       |No|
| Linux   |  60-real;70-real;75-real;80-real;90a-real;90a-virtual    |FPA_INTB_GEMM| 230 MB       |Yes|
| Windows | 52-real;61-real;75-real;86-real;89-real;90a-virtual    || 264 MB       |No (original)|
| Windows | 61-real;75-real;86-real;89-real;90a-virtual    || 254 MB       |No|
| Windows | 75-real;86-real;89-real;90a-virtual    || 242 MB       |Yes|

PR #26230, opened 2025-10-03
[CUDA] replace 90a-virtual by 90-virtual for forward compatible

tianleiwu:
Users with RTX 5090 GPUs are experiencing runtime errors when using onnxruntime-gpu:
```
[ONNXRuntimeError] : 1 : FAIL : Non-zero status code returned while running Slice node. 
Name:'Slice_34' Status Message: CUDA error cudaErrorNoKernelImageForDevice:
no kernel image is available for execution on the device
```
This occurs because RTX 5090 uses CUDA compute architecture 12.0 (SM 12.0). The incompatibility of `onnxruntime-gpu` 1.23 was built with `90a-virtual`. The `90a` architecture is a specialized, non-forward-compatible version of the Hopper architecture, making it incompatible with future GPU generations like Blackwell.

This change will revert `90a-virtual` back to `90-virtual` as used in 1.22. This shall bring back the compatibility in Blackwell GPU.

Related:
https://github.com/microsoft/onnxruntime/pull/26002
https://github.com/microsoft/onnxruntime/pull/26226
https://github.com/microsoft/onnxruntime/issues/26181


hariharans29:
Just curious - What is the binary size impact for this ?

tianleiwu:
> Just curious - What is the binary size impact for this ?

Binary will be smaller (for example 4MB is reduced for Linux wheel).

onnxruntime/contrib_ops/cuda/llm/fpA_intB_gemm_profiler.cc
```diff
@@ -14,6 +14,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+#if USE_FPA_INTB_GEMM
```
nenad1002, line 17: No need for device specific checks?

tianleiwu: Right. The file has no device code.

JulienMaille:
Does this affect both the cuda backend and the dml backend or only the former?

snnn:
Sorry it did not get into the 1.23.1 patch release.

apsonawane:

PR #26234, opened 2025-10-03
[DRAFT] [Compile API] Support for offline/off-target compile for plugin EPs

adrianlizarraga:
### Description
DRAFT. Not ready for detailed review.

### DIFF
--- a/include/onnxruntime/core/session/onnxruntime_ep_c_api.h
+++ b/include/onnxruntime/core/session/onnxruntime_ep_c_api.h
@@ -465,6 +465,15 @@ struct OrtEpApi {
    */
   ORT_API_T(uint64_t, GetSyncIdForLastWaitOnSyncStream,
             _In_ const OrtSyncStream* producer_stream, _In_ const OrtSyncStream* consumer_stream);
+
+  ORT_API2_STATUS(CreateHardwareDevice, _In_ OrtHardwareDeviceType type,
+                  _In_ uint32_t vendor_id,
+                  _In_ uint32_t device_id,
+                  _In_ const char* vendor_name,
+                  _In_opt_ const OrtKeyValuePairs* metadata,
+                  _Out_ OrtHardwareDevice** hardware_device);
+
+  ORT_CLASS_RELEASE(HardwareDevice);
 };
 
 /**
@@ -981,6 +990,13 @@ struct OrtEpFactory {
                   _In_ const OrtMemoryDevice* memory_device,
                   _In_opt_ const OrtKeyValuePairs* stream_options,
                   _Outptr_ OrtSyncStreamImpl** stream);
+
+  ORT_API2_STATUS(GetAdditionalHardwareDevices, _In_ OrtEpFactory* this_ptr,
+                  _In_reads_(num_devices) const OrtHardwareDevice* const* found_devices,
+                  _In_ size_t num_found_devices,
+                  _Inout_ OrtHardwareDevice** additional_devices,
+                  _In_ size_t max_additional_devices,
+                  _Out_ size_t* num_additional_devices);
 };
 
 #ifdef __cplusplus
--- a/onnxruntime/core/session/environment.cc
+++ b/onnxruntime/core/session/environment.cc
@@ -746,6 +746,63 @@ std::vector<const OrtHardwareDevice*> SortDevicesByType() {
 
   return sorted_devices;
 }
+
+std::vector<OrtHardwareDevice*> FilterEpHardwareDevices(const OrtEpFactory& ep_factory,
+                                                        gsl::span<const OrtHardwareDevice* const> ort_hw_devices,
+                                                        gsl::span<OrtHardwareDevice* const> ep_hw_devices,
+                                                        const char* lib_registration_name) {
+  // ORT is not required to use all hw devices provided by the EP factory.
+  // This function filters out the following hw devices:
+  //  - HW devices that were already found during ORT's device discovery.
+  //  - HW devices with vendor information that does not match the EP factory.
+
+  if (ep_hw_devices.empty()) {
+    return {};
+  }
+
+  auto have_ort_hw_device = [&ort_hw_devices](const OrtHardwareDevice* candidate) -> bool {
+    return std::find_if(ort_hw_devices.begin(), ort_hw_devices.end(),
+                        [&candidate](const OrtHardwareDevice* ort_device) {
+                          return candidate->device_id == ort_device->device_id &&
+                                 candidate->vendor_id == ort_device->vendor_id &&
+                                 candidate->type == ort_device->type;
+                        }) != ort_hw_devices.end();
+  };
+
+  std::vector<OrtHardwareDevice*> result;
+  result.reserve(ep_hw_devices.size());
+
+  const char* ep_factory_name = ep_factory.GetName(&ep_factory);
+  const uint32_t ep_vendor_id = ep_factory.GetVendorId(&ep_factory);
+  const std::string ep_vendor = ep_factory.GetVendor(&ep_factory);
+
+  for (OrtHardwareDevice* candidate : ep_hw_devices) {
+    if (candidate == nullptr) {
+      continue;  // EP library provided a NULL hw device. Skip it.
+    }
+
+    if (candidate->vendor_id != ep_vendor_id ||
+        candidate->vendor != ep_vendor) {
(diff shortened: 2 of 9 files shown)


include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -465,6 +465,33 @@ struct OrtEpApi {
...
+   * \note Called within OrtEpFactory::GetAdditionalHardwareDevices to augment the list of devices discovered by ORT.
+   *
+   * \param[in] type The hardware device type.
+   * \param[in] vendor_id The hardware device's vendor identifier.
+   * \param[in] device_id The hardware device's identifier.
+   * \param[in] vendor_name The hardware device's vendor name as a null-terminated string. Copied by ORT.
+   * \param[in] metadata Optional OrtKeyValuePairs instance for hardware device metadata that may be queried by
+   *                     applications via OrtApi::GetEpDevices() or the EP factory that receives this hardware device
+   *                     instance as input to OrtEpFactory::GetSupportedDevices().
+   *                     Refer to onnxruntime_ep_device_ep_metadata_keys.h for common OrtHardwareDevice metadata keys.
```
skottmckay, line 479: nit: does ORT take ownership and EP does not need to release metadata KVP instance if provided?

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -981,6 +1008,39 @@ struct OrtEpFactory {
...
                   _Outptr_ OrtSyncStreamImpl** stream);
+
+  /** \brief Get additional hardware devices from the execution provider to augment the devices discovered by ORT.
+   *
+   * \note Any returned devices that have already been found by ORT are ignored.
+   *
+   * \note New additional devices created by this EP factory are not provided to other EP factories. Only this
+   *       EP factory receives the new additional hardware devices via OrtEpFactory::GetSupportedDevices().
+   *       Any OrtEpDevice instances that this EP factory creates with an additional hardware device are visible to
+   *       applications that call OrtApi::GetEpDevices().
```
skottmckay: Do we need this complexity?

Alternatively, if we're only going to show the OrtEpDevice to the EP that created it, can the EP create it inside of GetSupportedDevices? 

e.g. inside GetSupportedDevices
- EP determines a virtual device is required
- EP creates OrtHardwareDevice. 
- EP uses that to create OrtEpDevice.

EP could register the OrtHardwareDevice with ORT so we can release it when the EP is unregistered, or EP could own it if we provided a ReleaseHardwareDevice function. Only place the OrtHardwareDevice shows up is indirectly in the OrtEpDevice.

onnxruntime/core/platform/device_discovery_common.cc
```diff
@@ -48,6 +49,8 @@ OrtHardwareDevice DeviceDiscovery::GetCpuDeviceFromCPUIDInfo() {
   cpu_device.vendor_id = cpuid_info.GetCPUVendorId();
   cpu_device.device_id = 0;
   cpu_device.type = OrtHardwareDeviceType_CPU;
+  cpu_device.metadata.Add(kOrtHardwareDevice_MetadataKey_DiscoveredBy, "ONNX Runtime");
```
skottmckay: Do we need this as well as IsVirtual? Wary of having too much stuff in metadata, and possibly the IsVirtual implies it was not discovered by ORT unless there's some other feature that will rely on an explicit 'discovered by' value.

adrianlizarraga: My thought is that:
- `DiscoveredBy` tells the application who discovered the device. May be useful if the app only wants stuff discovered by ORT.
- `IsVirtual` is "1" if it does not correspond to actual hardware. It's the negation of `IsHardware` used by [dxcore api](https://learn.microsoft.com/en-us/windows/win32/api/dxcore_interface/ne-dxcore_interface-dxcoreadapterproperty). This is a separate concept from `DiscoveredBy`. This would be useful for off-target compilation, for example.

skottmckay: If we only show the virtual device to the EP that created it, is there value in DiscoveredBy?

skottmckay:
How does an EP know what sort of device it should create? e.g. an arm64 vs x64 virtual device.

Would it create a generic virtual device so we have an OrtEpDevice for it, that gets explicitly selected using SessionOptionsAppendExecutionProvider_V2, and the EP options that can be provided there are used to specify things like the target architecture etc. if this is a cross compiling scenario?

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
skottmckay: I think that helps with complexity, but it lacks a signal for the EP to know it should create a virtual device. 

`GetAdditionalHardwareDevices` is better, but that doesn't have a signal for device type (e.g. create a GPU or NPU or both). 

Or do we expect it creates all possible virtual devices by default?

adrianlizarraga: >Or do we expect it creates all possible virtual devices by default?

Yeah, that was my original idea. The EP library knows 
1. The platform/arch the Ep library was built for (e.g., win x64).
2. The set of actual `OrtHardwareDevice` instances that ORT found.
3. The set of virtual devices that the EP library wants to expose to the application for that specific platform/arch.

So, based on the above, the EP library would create all possible/necessary virtual devices. The EP library also knows the correct provider options to associate with each virtual device.

onnxruntime/core/platform/device_discovery_common.cc
adrianlizarraga: The application, which can see all OrtEpDevices via `GetEpDevices`, would be able to see who created the hw device.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
adrianlizarraga: As per our offline discussion, this PR now allows app to signal the EP (via ".virtual" suffix to EP registration name) that virtual devices are allowed.

onnxruntime/core/platform/device_discovery_common.cc
adrianlizarraga: I removed `DiscoveredBy` and retained `IsVirtual`.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
adrianlizarraga: As per our discussion, the EP library now owns the virtual `OrtHardwareDevice`, which it releases via `ReleaseHardwareDevice`.

onnxruntime/core/session/environment.cc
```diff
@@ -746,6 +746,28 @@ std::vector<const OrtHardwareDevice*> SortDevicesByType() {
...
+bool AreVirtualDevicesAllowed(std::string_view lib_registration_name) {
+  constexpr std::string_view suffix{".virtual"};
+
+  return lib_registration_name.size() >= suffix.size() &&
+         lib_registration_name.compare(lib_registration_name.size() - suffix.size(),
+                                       suffix.size(), suffix) == 0;
+}
+
+Status SetEpFactoryEnvironmentOptions(OrtEpFactory& factory, std::string_view lib_registration_name) {
+  if (factory.SetEnvironmentOptions == nullptr) {
```
skottmckay: needs version check as well. could be random bits if the EP was built with ORT 1.23

onnxruntime/core/session/plugin_ep/ep_api.cc
```diff
@@ -230,6 +258,9 @@ static constexpr OrtEpApi ort_ep_api = {
     &OrtExecutionProviderApi::SyncStream_GetImpl,
     &OrtExecutionProviderApi::SyncStream_GetSyncId,
     &OrtExecutionProviderApi::GetSyncIdForLastWaitOnSyncStream,
+
```
skottmckay, line 262:
```suggestion
    // End of Version 23 - DO NOT MODIFY ABOVE

```

adrianlizarraga: This was added in a separate PR today: https://github.com/microsoft/onnxruntime/pull/26418

onnxruntime/core/session/environment.cc
adrianlizarraga: Thanks!

skottmckay (approved):

PR #26240, opened 2025-10-04
Switch to a new macOS pool

snnn:


snnn:
The new machine does  not have Java. I need to solve this problem first.

PR #26250, opened 2025-10-06
Ignore deprecation warnings when building TRT/NV EP

kevinch-nv:
### Description
Add -Wno-deprecated-declarations when compiling unit test with TRT/NV EP Support

### Motivation and Context
Fixes source build for these EPs.


kevinch-nv:
@chilo-ms for viz.

PR #26252, opened 2025-10-06
Move macOS build jobs to a dedicated pool

snnn:
### Description
Move packaging pipelines' macOS build jobs to a new machine pool that is much faster, which can reduce the build time by 85%.
However, it has some drawbacks:
1. It does not have Java installed. Also, ADO's JavaToolInstaller task does not support macOS. So I have to remove these things from the pipeline. We will stop providing java binaries for macOS until this issue is resolved.
2. It does not Cocoapod. But Cocoapod itself is being deprecated.  So it is not a big deal.


baijumeswani (approved):

snnn:
There are some test failures in the Nuget Test pipeline and Jar Packaging Test pipeline, because these pipelines have not been migrated to the new machine yet. I will solve it later.

PR #26255, opened 2025-10-07
Add support to filter_ep_devices using ep or device metadata in perftest

jnagi-intel:


jnagi-intel:
@chilo-ms Can you please re-run the CI ? I pushed a fix for lint error

onnxruntime/test/perftest/strings_helper.cc
```diff
@@ -137,5 +137,22 @@ void ParseEpDeviceIndexList(const std::string& input, std::vector<int>& result)
     }
   }
 }
+
+void ParseEpDeviceFilterKeyValuePairs(const std::string& input, std::vector<std::pair<std::string, std::string>>& result) {
```
chilo-ms, line 141: Do we want to add another helper function when the existing `ParseSessionConfigs()` can achieve the same goal? 
I suggest we can rename `ParserSessionConfigs()` to a more generic name.

jnagi-intel: ParseSessionConfigs returns an unordered map and will not work for 
`--filter_ep_devices "ov_device|NPU ov_device|CPU"`

In this case, we would like to append NPU and CPU devices in the same order.

chilo-ms:
Reopen this PR to trigger "license/cla" pipeline

chilo-ms:
Please help merge main to include the [fix ](https://github.com/microsoft/onnxruntime/pull/26315)to pass the failing pipelines

jnagi-intel:
> Please help merge main to include the [fix ](https://github.com/microsoft/onnxruntime/pull/26315)to pass the failing pipelines

done

PR #26257, opened 2025-10-07
Add libbacktrace in Linux Debug build

tianleiwu:
Show filename and line number in stack trace in Linux debug build


snnn:
abseil has this functionality. Or we may enable C++20 which has other benefits as well.

tianleiwu:
> abseil has this functionality. Or we may enable C++20 which has other benefits as well.

You are right. Abseil uses libunwind and libbacktrace internally, so use abseil backtrace is a better solution.

PR #26269, opened 2025-10-09
Add partial data propagation to enhance shape inference

chilo-ms:
### Description
Calling an operator's `TypeAndShapeInferenceFunction()` alone is sometimes insufficient for complete shape inference. 

For example, the `Shape` operator only infers the output’s rank (a 1-dimensional tensor) but not its actual dimension values.
For instance, given an input of shape [1, 3, 64, 64], the Shape operator's `TypeAndShapeInferenceFunction()` produces an output shape tensor with 1-dimension as int64[4], representing the rank of the input tensor.

Therefore, as you can imagine, the below graph's output shape can't be properly inferred (even though the input shape is known) because the shape data is lost at the `Shape `operator.
image

To solve the issue, the `PartialDataPropagationFunction()`, defined in the ONNX operator schema, must also be executed to obtain the concrete output shape values, allowing accurate propagation of shape information throughout the graph.

This PR adds the support of executing operator's  `PartialDataPropagationFunction()` in ORT, and makes sure the shape values is properly propagated throughout the graph.

### Motivation and Context

When using the Compile API to generate an EPContext model, all graph optimizations are disabled by default except for free dimension overrides. However, for certain models, such as a VAE decoder, the output shape may still fail to be properly inferred even when free dimension override values are provided beforehand.

However, you won't hit this issue if enabling all the graph optimizations as some nodes, e.g. `Shape`, `Reshape `.. will be constant folded.

### DIFF
--- a/onnxruntime/core/graph/graph.cc
+++ b/onnxruntime/core/graph/graph.cc
@@ -2947,7 +2947,7 @@ Status Graph::SaveShapeValuesFromDataPropagation(const Node& node,
       // Get shape from TensorProto as well as element counts.
       // If shape has dimension size equals zero, it means it's a scalar and has only one element.
       auto tensor_shape = utils::GetTensorShapeFromTensorProto(*initializer);
-      size_t element_cnt = static_cast<size_t>(tensor_shape.Size());
+      size_t element_cnt = narrow<size_t>(tensor_shape.Size());
 
       // Check if this is in-memory external data (data stored in OrtValue)
       if (utils::HasExternalDataInMemory(*initializer)) {
--- a/onnxruntime/test/framework/shape_inference_test.cc
+++ b/onnxruntime/test/framework/shape_inference_test.cc
@@ -86,11 +86,9 @@ TEST(ShapeInferenceV2Test, PartialDataPropagationTest) {
 
     Ort::SessionOptions session_options{};
     session_options.SetGraphOptimizationLevel(ORT_DISABLE_ALL);
-
-    const OrtApi* g_ort = OrtGetApiBase()->GetApi(ORT_API_VERSION);
-    ORT_ENFORCE(g_ort->AddFreeDimensionOverrideByName(session_options, "batch", 1) == nullptr);
-    ORT_ENFORCE(g_ort->AddFreeDimensionOverrideByName(session_options, "width", 64) == nullptr);
-    ORT_ENFORCE(g_ort->AddFreeDimensionOverrideByName(session_options, "height", 64) == nullptr);
+    session_options.AddFreeDimensionOverrideByName("batch", 1);
+    session_options.AddFreeDimensionOverrideByName("width", 64);
+    session_options.AddFreeDimensionOverrideByName("height", 64);
 
     // Even though all graph optimizations are disabled, the free dimension override is still enabled by default.
     // The shape of graph's output should be correctly inferred by shape inference and data propagation.
@@ -116,11 +114,9 @@ TEST(ShapeInferenceV2Test, PartialDataPropagationTest) {
 
     Ort::SessionOptions session_options{};
     session_options.SetGraphOptimizationLevel(ORT_DISABLE_ALL);
-
-    const OrtApi* g_ort = OrtGetApiBase()->GetApi(ORT_API_VERSION);
-    ORT_ENFORCE(g_ort->AddFreeDimensionOverrideByName(session_options, "batch", 1) == nullptr);
-    ORT_ENFORCE(g_ort->AddFreeDimensionOverrideByName(session_options, "width", 64) == nullptr);
-    ORT_ENFORCE(g_ort->AddFreeDimensionOverrideByName(session_options, "height", 64) == nullptr);
+    session_options.AddFreeDimensionOverrideByName("batch", 1);
+    session_options.AddFreeDimensionOverrideByName("width", 64);
+    session_options.AddFreeDimensionOverrideByName("height", 64);
 
     // Even though all graph optimizations are disabled, the free dimension override is still enabled by default.
     // The shape of graph's output should be correctly inferred by shape inference and data propagation.
@@ -145,11 +141,9 @@ TEST(ShapeInferenceV2Test, PartialDataPropagationTest) {
 
     Ort::SessionOptions session_options{};
     session_options.SetGraphOptimizationLevel(ORT_DISABLE_ALL);
-
-    const OrtApi* g_ort = OrtGetApiBase()->GetApi(ORT_API_VERSION);
-    ORT_ENFORCE(g_ort->AddFreeDimensionOverrideByName(session_options, "batch", 1) == nullptr);
-    ORT_ENFORCE(g_ort->AddFreeDimensionOverrideByName(session_options, "width", 64) == nullptr);
-    ORT_ENFORCE(g_ort->AddFreeDimensionOverrideByName(session_options, "height", 64) == nullptr);
+    session_options.AddFreeDimensionOverrideByName("batch", 1);
+    session_options.AddFreeDimensionOverrideByName("width", 64);
+    session_options.AddFreeDimensionOverrideByName("height", 64);
 
     // Even though all graph optimizations are disabled, the free dimension override is still enabled by default.
     // The shape of graph's output should be correctly inferred by shape inference and data propagation.
--- a/include/onnxruntime/core/session/onnxruntime_cxx_api.h
+++ b/include/onnxruntime/core/session/onnxruntime_cxx_api.h
@@ -1441,6 +1441,12 @@ struct SessionOptionsImpl : ConstSessionOptionsImpl<T> {
 
   ///< Wraps OrtApi::SessionOptionsAppendExecutionProvider_VitisAI
   SessionOptionsImpl& AppendExecutionProvider_VitisAI(const std::unordered_map<std::string, std::string>& provider_options = {});
+
+  ///< Wraps OrtApi::AddFreeDimensionOverride
+  SessionOptionsImpl& AddFreeDimensionOverride(const char* dim_denotation, int64_t dim_value);
+
+  ///< Wraps OrtApi::AddFreeDimensionOverrideByName
+  SessionOptionsImpl& AddFreeDimensionOverrideByName(const char* dim_name, int64_t dim_value);
 };
 }  // namespace detail
 
--- a/include/onnxruntime/core/session/onnxruntime_cxx_inline.h
+++ b/include/onnxruntime/core/session/onnxruntime_cxx_inline.h
@@ -1503,6 +1503,18 @@ inline SessionOptionsImpl<T>& SessionOptionsImpl<T>::RegisterCustomOpsUsingFunct
   return *this;
 }
 
+template <typename T>
+inline SessionOptionsImpl<T>& SessionOptionsImpl<T>::AddFreeDimensionOverride(const char* dim_denotation, int64_t dim_value) {
+  ThrowOnError(GetApi().AddFreeDimensionOverrideByName(this->p_, dim_denotation, dim_value));
+  return *this;
+}
+
+template <typename T>
+inline SessionOptionsImpl<T>& SessionOptionsImpl<T>::AddFreeDimensionOverrideByName(const char* dim_name, int64_t dim_value) {
+  ThrowOnError(GetApi().AddFreeDimensionOverrideByName(this->p_, dim_name, dim_value));
+  return *this;
+}
+
 /// Session
 template <typename T>
 inline size_t ConstSessionImpl<T>::GetInputCount() const {


include/onnxruntime/core/graph/graph.h
```diff
@@ -1760,6 +1760,12 @@ class Graph {  // NOLINT(clang-analyzer-optin.performance.Padding): preserve exi
                                                     std::vector<const ONNX_NAMESPACE::TypeProto*>& output_types,
                                                     const Graph::ResolveOptions& options);
 
+  // If the dimension values are inferred after executing ONNX operator's PartialDataPropagationFunction(),
+  // save them to the output NodeArg as a TensorShapeProto or a scalar value so that downstream (consumer) nodes
+  // can use them later for their PartialDataPropagationFunction() and PartialDataPropagationFunction().
+  common::Status SaveValuesFromDataPropagation(Node& node, NodeArg& output_def,
```
yuslepukhin: We are saving shapes? I would suggest the name has something to do with shapes, also, NodeArg Get method returns TensorShapeProto and here we are saving TypeProto? Can we make it uniform? #Resolved

include/onnxruntime/core/graph/node_arg.h
```diff
@@ -107,6 +107,9 @@ class NodeArg {
   /** Gets this NodeArg as a NodeArgInfo, AKA ValueInfoProto. */
   const NodeArgInfo& ToProto() const noexcept { return node_arg_info_; }
 
+  /** Gets the inferred shape as a TensorShapeProto. */
+  const ONNX_NAMESPACE::TensorShapeProto& GetValuesAfterDataPropagation() const noexcept { return values_after_data_propagation_; }
```
yuslepukhin: Can we come up with a better name? #Resolved

onnxruntime/core/graph/graph.cc
```diff
@@ -2675,10 +2676,48 @@ class InferenceContextImpl : public ONNX_NAMESPACE::InferenceContext {
     if (!def)
       return nullptr;
 
-    // only return data if it's for a constant initializer. checks for outer scope initializers
-    // if this is a subgraph and the name isn't found locally.
+    // Returns if it's a constant initializer.
```
yuslepukhin, line 2722: Needs to be rebased from main to pickup the latest. #Resolved

include/onnxruntime/core/graph/node_arg.h
```diff
@@ -128,6 +131,22 @@ class NodeArg {
...
+  // This variable stores the inferred shape data as a TensorShapeProto after executing
+  // the ONNX operator's PartialDataPropagationFunction().
+  //
+  // Calling an operator's TypeAndShapeInferenceFunction() alone is sometimes insufficient
+  // for complete shape inference. For example, the Shape operator only provides the
+  // output's rank (1-dimensional) but not its actual dimension values.
+  // The PartialDataPropagationFunction(), defined in the ONNX operator schema, must also
+  // be executed to obtain the concrete output shape values, allowing accurate propagation
+  // of shape information throughout the graph.
+  ONNX_NAMESPACE::TensorShapeProto values_after_data_propagation_;
```
yuslepukhin: Should this be std::optional to add clarity if this actually present? Checking for number of dims is not that good given that empty shape is a valid scalar. #Resolved

onnxruntime/core/graph/graph.cc
```diff
@@ -2675,10 +2676,48 @@ class InferenceContextImpl : public ONNX_NAMESPACE::InferenceContext {
...
+        if (dim.has_dim_value()) {
+          tensor_proto.add_int64_data(dim.dim_value());
+        } else {
+          all_values = false;
+          break;
+        }
+      }
+
+      if (all_values) {
+        tensor_proto_for_shape_.push_back(std::make_unique<TensorProto>(std::move(tensor_proto)));
```
yuslepukhin: Should we store it inside def as TensorProto to avoid conversion here or is better as TensorShape? #Resolved

onnxruntime/core/graph/graph.cc
```diff
@@ -2675,10 +2676,48 @@ class InferenceContextImpl : public ONNX_NAMESPACE::InferenceContext {
...
+    if (tensor_shape_proto.dim_size() > 0) {
+      TensorProto tensor_proto;
+      tensor_proto.set_data_type(TensorProto_DataType_INT64);
+      tensor_proto.add_dims(tensor_shape_proto.dim_size());
+      bool all_values = true;
+      for (const auto& dim : tensor_shape_proto.dim()) {
+        if (dim.has_dim_value()) {
+          tensor_proto.add_int64_data(dim.dim_value());
+        } else {
+          all_values = false;
```
yuslepukhin, line 2770: Should this be saved inside the def in the first place if it is not good for nothing? #Resolved

include/onnxruntime/core/graph/graph.h
chilo-ms: We save either a `TensorShapeProto` or a scalar in the NodeArg to keep the inferred shape values.
The `TypeProto` is the result from executing data propagation and we only parse the `TypeProto` but not save it to get the tensor that represents the shape values out of it. 

Regarding the name, i will rename it to `SaveShapeValuesFromDataPropagation`

onnxruntime/test/framework/shape_inference_test.cc
```diff
@@ -76,6 +78,63 @@ TEST_F(ShapeInferenceTest, BasicTest) {
   CheckShapeEquality(InputShape(node), OutputShape(node));
 }
 
+TEST(ShapeInferenceV2Test, PartialDataPropagationTest) {
+  {
```
yuslepukhin, line 82: What is the difference between this test case and the one after that? Can you add comments? #Resolved

onnxruntime/core/graph/graph.cc
```diff
@@ -2698,8 +2699,45 @@ class InferenceContextImpl : public ONNX_NAMESPACE::InferenceContext {
...
+
+    // The following code handles cases where a node stores the previously inferred output shape values in its NodeArg.
+    //
+    // For example, the Reshape operator, its shape input may come from a producer node such as a Shape operator,
+    // and the inferred output shape value is already stored as a TensorShapeProto in corresponding NodeArg.
+    //
+    // In such cases, the Reshape operator should convert this TensorShapeProto into a TensorProto.
+    // The resulting TensorProto will then be treated as an initializer during ONNX shape inference,
+    // allowing the real dimension values to be correctly used.
+    // See https://github.com/onnx/onnx/blob/main/onnx/defs/shape_inference.cc#L467 for details.
```
yuslepukhin: I am not seeing how this link is helpful. Perhaps, we should put a link that describes the Propagate function and how it can be used? #Resolved

include/onnxruntime/core/graph/graph.h
yuslepukhin: I have a trouble finding where this interface is used.

onnxruntime/core/graph/graph.cc
yuslepukhin: Can we re-use `temp_tensor_protos_`?

include/onnxruntime/core/graph/graph.h
chilo-ms: It's used after ORT gets the `TypeProto` output from executing data propagation.
https://github.com/microsoft/onnxruntime/blob/chi/fix_shape_inference/onnxruntime/core/graph/graph.cc#L3345

include/onnxruntime/core/graph/node_arg.h
chilo-ms: Changed to `GetInferredShapeValues`.

include/onnxruntime/core/graph/node_arg.h
chilo-ms: good suggestion. I used std::optional for it.

onnxruntime/core/graph/graph.cc
chilo-ms: `onnx::InferenceContext::getInputData()` asks for a `TensorProto`  and `onnx::DataPropagationContext::getInputData()` asks for a `TensorShapeProto`.
So i think the conversion is inevitable.

But yes we can re-use `temp_tensor_protos_`

onnxruntime/core/graph/graph.cc
chilo-ms: One reason i don't think we should save the `TensorProto` in the def is that it's waste of memory because it's going to stay with the def if we don't clean it up.

Saving this `TensorProto` in inference context saves more memory as the inference context will be cleaned up for every node iteration.

onnxruntime/test/framework/shape_inference_test.cc
chilo-ms: comments added.

onnxruntime/core/graph/graph.cc
chilo-ms: I removed the link and added the [link ](https://github.com/onnx/onnx/blob/main/onnx/defs/shape_inference.h#L117-L127)to the description of data propagation in 
```
// An implementation of the DataPropagationContext interface optional by operator-specific
// shape inference for onnxruntime graphs.
// Please see the description and usage of ONNX's data propagation here:
// https://github.com/onnx/onnx/blob/main/onnx/defs/shape_inference.h#L117-L127
class DataPropagationContextImpl : public ONNX_NAMESPACE::DataPropagationContext {
...
```

yuslepukhin (dismissed):

onnxruntime/test/framework/shape_inference_test.cc
```diff
@@ -76,6 +78,109 @@ TEST_F(ShapeInferenceTest, BasicTest) {
...
+    // Model #1
+    // This model contains "Shape" and "Reshape" operators.
+    auto model_path = ORT_TSTR("testdata/test_shape_data_propagation_with_shape_related_nodes.onnx");
+
+    Ort::SessionOptions session_options{};
+    session_options.SetGraphOptimizationLevel(ORT_DISABLE_ALL);
+
+    const OrtApi* g_ort = OrtGetApiBase()->GetApi(ORT_API_VERSION);
+    ORT_ENFORCE(g_ort->AddFreeDimensionOverrideByName(session_options, "batch", 1) == nullptr);
```
yuslepukhin: This should be C++ #Resolved

onnxruntime/core/graph/graph.cc
```diff
@@ -2742,9 +2778,474 @@ class InferenceContextImpl : public ONNX_NAMESPACE::InferenceContext {
...
+          }
+        }
+        return true;
+      }
+
+      return false;
+    };
+
+    // Try to get the previously inferred shape values that stored in NodeArg's inferred_shape_values_
+    auto& tensor_shape_proto = def->GetInferredShapeValues();
```
yuslepukhin: const auto& #Resolved

onnxruntime/core/graph/graph.cc
```diff
@@ -2742,9 +2778,474 @@ class InferenceContextImpl : public ONNX_NAMESPACE::InferenceContext {
...
+    if (tensor_shape_proto.has_value() && has_shape_values(*tensor_shape_proto)) {
```
yuslepukhin: Need an explanation for scalars. This is a bit confusing. It has a value, but has_shape_values returns false, it means it is a scalar. Should we handle differently? #Resolved

onnxruntime/core/graph/graph.cc
```diff
@@ -2742,9 +2778,474 @@ class InferenceContextImpl : public ONNX_NAMESPACE::InferenceContext {
...
+  }
+
+  void RunInferencing() {
+    auto schema = node_.Op();
+    if (nullptr != schema) {
+      schema->GetDataPropagationFunction()(*this);
+    }
+  }
+
+  std::vector<TypeProto> InferredOutputTypes() const { return node_output_types_; }
```
yuslepukhin: Why by value? #Resolved

onnxruntime/core/graph/graph.cc
```diff
@@ -2742,9 +2778,474 @@ class InferenceContextImpl : public ONNX_NAMESPACE::InferenceContext {
...
+
+  void addOutputData(size_t index, TensorShapeProto&& tsp) override {
+    if (index >= node_output_types_.size()) return;
+
+    TypeProto& type_proto = node_output_types_[index];
+    *type_proto.mutable_tensor_type()->mutable_shape() = std::move(tsp);
+  }
+
+  void RunInferencing() {
+    auto schema = node_.Op();
```
yuslepukhin: `auto*` #Resolved

onnxruntime/core/graph/graph.cc
```diff
@@ -2698,8 +2699,43 @@ class InferenceContextImpl : public ONNX_NAMESPACE::InferenceContext {
...
+    // and the inferred output shape value is already stored as a TensorShapeProto in corresponding NodeArg.
+    //
+    // In such cases, the Reshape operator should convert this TensorShapeProto into a TensorProto.
+    // The resulting TensorProto will then be treated as an initializer during ONNX shape inference,
+    // allowing the real dimension values to be correctly used.
+    const auto& tensor_shape_proto = def->GetInferredShapeValues();
+
+    // Make sure the returning shape tensor as a TensorProto has rank > 0 and all the dimensions
+    // have values (not symbolic)
+    if (tensor_shape_proto.has_value() && tensor_shape_proto->dim_size() > 0) {
```
yuslepukhin: We need to add an explanation why scalars are ignored here and below. #Resolved

onnxruntime/core/graph/graph.cc
```diff
@@ -2742,9 +2778,474 @@ class InferenceContextImpl : public ONNX_NAMESPACE::InferenceContext {
...
+// Please see the description and usage of ONNX's data propagation here:
+// https://github.com/onnx/onnx/blob/main/onnx/defs/shape_inference.h#L117-L127
+class DataPropagationContextImpl : public ONNX_NAMESPACE::DataPropagationContext {
+ public:
+  DataPropagationContextImpl(Node& node) noexcept : node_(node) {
+    node_output_types_.resize(node.OutputDefs().size());
+  }
+
+  const AttributeProto* getAttribute(const std::string& name) const override {
+    auto& attribute_value_map = node_.GetAttributes();
```
yuslepukhin, line 2856: `const auto&` #Resolved

onnxruntime/core/graph/graph.cc
```diff
@@ -2742,9 +2778,474 @@ class InferenceContextImpl : public ONNX_NAMESPACE::InferenceContext {
...
+  }
+
+  void addOutputData(size_t index, TensorShapeProto&& tsp) override {
+    if (index >= node_output_types_.size()) return;
+
+    TypeProto& type_proto = node_output_types_[index];
+    *type_proto.mutable_tensor_type()->mutable_shape() = std::move(tsp);
+  }
+
+  void RunInferencing() {
```
yuslepukhin, line 2924: Should this be an override? #Resolved

onnxruntime/core/graph/graph.cc
```diff
@@ -2742,9 +2778,474 @@ class InferenceContextImpl : public ONNX_NAMESPACE::InferenceContext {
...
+  Node& node_;
+  std::vector<TypeProto> node_output_types_;
+};
+
+Status Graph::SaveShapeValuesFromDataPropagation(Node& node,
+                                                 NodeArg& output_def,
+                                                 const TypeProto& onnx_inferred_type_after_data_propagation) const {
+  auto dim_size = onnx_inferred_type_after_data_propagation.tensor_type().shape().dim_size();
+
+  // Helper function to get the input value if it's a initializer.
```
yuslepukhin, line 2941: Should we check that there is a tensor_type (since support other types) and shape is present? #Resolved

onnxruntime/core/graph/graph.cc
```diff
@@ -2742,9 +2778,474 @@ class InferenceContextImpl : public ONNX_NAMESPACE::InferenceContext {
...
+          if (initializer->data_type() == TensorProto_DataType_INT32) {
+            input_values.resize(element_cnt);
+            for (size_t i = 0; i < element_cnt; ++i) {
+              input_values[i] = static_cast<int64_t>(tensor.Data<int32_t>()[i]);
+            }
+          } else if (initializer->data_type() == TensorProto_DataType_INT64) {
+            input_values.resize(element_cnt);
+            for (size_t i = 0; i < element_cnt; ++i) {
+              input_values[i] = tensor.Data<int64_t>()[i];
+            }
```
yuslepukhin: Should this be a memcpy? #Resolved

onnxruntime/core/graph/graph.cc
```diff
@@ -2742,9 +2778,474 @@ class InferenceContextImpl : public ONNX_NAMESPACE::InferenceContext {
...
+      // Check if this is in-memory external data (data stored in OrtValue)
+      if (utils::HasExternalDataInMemory(*initializer)) {
+        // Try to get the OrtValue for this initializer
+        OrtValue ort_value;
+        if (this->GetOrtValueInitializer(input_name, ort_value, true)) {
+          const Tensor& tensor = ort_value.Get<Tensor>();
+          if (initializer->data_type() == TensorProto_DataType_INT32) {
+            input_values.resize(element_cnt);
+            for (size_t i = 0; i < element_cnt; ++i) {
+              input_values[i] = static_cast<int64_t>(tensor.Data<int32_t>()[i]);
```
yuslepukhin: can we put resize() outside if/else since it is a common thing #Resolved

onnxruntime/core/graph/graph.cc
```diff
@@ -3280,6 +3798,26 @@ Status Graph::VerifyNodeAndOpMatch(const ResolveOptions& options) {
...
+
+Status Graph::CleanUpShapeValuesFromDataPropagation() {
+  for (auto node_index : nodes_in_topological_order_) {
+    auto& node = *GetNode(node_index);
+
+    for (auto node_arg : node.MutableInputDefs()) {
+      node_arg->inferred_shape_values_.reset();
+      node_arg->inferred_scalar_value_.reset();
+    }
+
```
yuslepukhin, line 3623: scalars do not allocate memory it is simply an int64_t in this case. #Resolved

onnxruntime/core/graph/graph.cc
```diff
@@ -2742,9 +2778,474 @@ class InferenceContextImpl : public ONNX_NAMESPACE::InferenceContext {
...
+        }
+      }
+
+      if (tensor_shape_proto.dim_size() == 1) {
+        output_def.inferred_scalar_value_ = tensor_shape_proto.dim(0).dim_value();
+      } else if (tensor_shape_proto.dim_size() > 1) {
+        // Get axes value
+        std::vector<int64_t> axes;
+        std::unordered_set<int64_t> axes_set;
+
```
yuslepukhin: Here and everywhere we should use TensorShapeVector for shapes. In general, we should not use std::vector unless required by some onnx interface. #Resolved

onnxruntime/core/graph/graph.cc
```diff
@@ -2742,9 +2778,474 @@ class InferenceContextImpl : public ONNX_NAMESPACE::InferenceContext {
...
+              input_values[i] = static_cast<int64_t>(tensor.Data<int32_t>()[i]);
+            }
+          } else if (initializer->data_type() == TensorProto_DataType_INT64) {
+            input_values.resize(element_cnt);
+            for (size_t i = 0; i < element_cnt; ++i) {
+              input_values[i] = tensor.Data<int64_t>()[i];
+            }
+          }
+        } else {
+          // If we can't get the OrtValue, it is a bug
```
yuslepukhin, line 2975: What happens if this is neither 32 or 64? #Resolved

onnxruntime/core/graph/graph.cc
```diff
@@ -2742,9 +2778,474 @@ class InferenceContextImpl : public ONNX_NAMESPACE::InferenceContext {
...
+                    " has in-memory external data but cannot get OrtValue during shape inference");
+        }
+      } else if (utils::HasRawData(*initializer)) {
+        const std::string& raw = initializer->raw_data();
+        const int64_t* data = reinterpret_cast<const int64_t*>(raw.data());
+        input_values.resize(element_cnt);
+        for (size_t i = 0; i < element_cnt; ++i) {
+          input_values[i] = data[i];
+        }
+      } else {
```
yuslepukhin: memcpy? #Resolved

onnxruntime/core/graph/graph.cc
```diff
@@ -2742,9 +2778,474 @@ class InferenceContextImpl : public ONNX_NAMESPACE::InferenceContext {
...
+        if (initializer->data_type() == TensorProto_DataType_INT32) {
+          std::vector<int32_t> values;
+          values.assign(initializer->int32_data().begin(), initializer->int32_data().end());
+          input_values.resize(element_cnt);
```
yuslepukhin, line 2990: I believe you should declare it as vector<int64_t>, 32 will be converted implicitely on insert. #Resolved

onnxruntime/core/graph/graph.cc
```diff
@@ -2742,9 +2778,474 @@ class InferenceContextImpl : public ONNX_NAMESPACE::InferenceContext {
...
+            }
+          }
+        } else {
+          // If we can't get the OrtValue, it is a bug
+          ORT_THROW("Initializer ", input_name,
+                    " has in-memory external data but cannot get OrtValue during shape inference");
+        }
+      } else if (utils::HasRawData(*initializer)) {
+        const std::string& raw = initializer->raw_data();
+        const int64_t* data = reinterpret_cast<const int64_t*>(raw.data());
```
yuslepukhin: From here until the end, you need to use one of the tensorproto utilities to propertly unpack the data because it is error prone. #Resolved

onnxruntime/core/graph/graph.cc
```diff
@@ -2742,9 +2778,474 @@ class InferenceContextImpl : public ONNX_NAMESPACE::InferenceContext {
...
+  size_t getNumOutputs() const noexcept override {
+    return node_output_types_.size();
+  }
+
+  const TypeProto* getOutputType(size_t index) const override {
+    if (index >= getNumOutputs()) {
+      return nullptr;
+    }
+
+    return &node_output_types_[index];
```
yuslepukhin, line 2891: If elements are added the vector may be reallocated and the address will become dangling. Is this possible? #Resolved

onnxruntime/core/graph/graph.cc
```diff
@@ -2742,9 +2778,474 @@ class InferenceContextImpl : public ONNX_NAMESPACE::InferenceContext {
...
+      }
+
+      if (tensor_shape_proto.dim_size() == 1) {
+        output_def.inferred_scalar_value_ = tensor_shape_proto.dim(0).dim_value();
+      } else if (tensor_shape_proto.dim_size() > 1) {
+        // Get axes value
+        std::vector<int64_t> axes;
+        std::unordered_set<int64_t> axes_set;
+
+        // Note: Starting from opset 13, "axes" is provided as a second input to the Squeeze operator.
```
yuslepukhin: InlinedHashSet #Resolved

onnxruntime/core/graph/graph.cc
```diff
@@ -2742,9 +2778,474 @@ class InferenceContextImpl : public ONNX_NAMESPACE::InferenceContext {
...
+            for (const auto& i : axes_attr.ints()) {
+              axes.push_back(i);
+            }
+          }
+        }
+
+        for (size_t i = 0; i < axes.size(); ++i) {
+          // Negative value means counting dimensions from the back.
+          if (axes[i] < 0) {
+            axes_set.insert(axes[i] + tensor_shape_proto.dim_size());
```
yuslepukhin: There is a utility function HandleNegativeDimensions or something like that. #Resolved

onnxruntime/core/graph/graph.cc
```diff
@@ -2742,9 +2778,474 @@ class InferenceContextImpl : public ONNX_NAMESPACE::InferenceContext {
...
+
+        int64_t dim_index = 0;
+        for (const auto& dim : tensor_shape_proto.dim()) {
+          auto value = dim.dim_value();
+          if (axes_set.size() > 0) {
+            if (axes_set.find(dim_index) == axes_set.end()) {
+              output_def.inferred_shape_values_->add_dim()->set_dim_value(value);
+            }
+          } else {
+            if (value != 1) {
```
yuslepukhin: I think we are loosing a dimension here #Resolved

onnxruntime/core/graph/graph.cc
```diff
@@ -2742,9 +2778,474 @@ class InferenceContextImpl : public ONNX_NAMESPACE::InferenceContext {
...
+        }
+        output_def.inferred_shape_values_->clear_dim();
+
+        int64_t dim_index = 0;
+        for (const auto& dim : tensor_shape_proto.dim()) {
+          auto value = dim.dim_value();
+          if (axes_set.size() > 0) {
+            if (axes_set.find(dim_index) == axes_set.end()) {
+              output_def.inferred_shape_values_->add_dim()->set_dim_value(value);
+            }
```
yuslepukhin: Not clear what we have in `axes_set` is it a dimension or a dimension index? #Resolved

onnxruntime/core/graph/graph.cc
```diff
@@ -2742,9 +2778,474 @@ class InferenceContextImpl : public ONNX_NAMESPACE::InferenceContext {
...
+  //   1. The inferred output is a scalar.
+  //   2. The node's input is a scalar, and the operator's PartialDataPropagationFunction() doesn't handle it.
+  //
+  // In other words, some operators' PartialDataPropagationFunction() implementations do not support
+  // scalar inputs or outputs. In such cases, attempt data propagation manually and store the inferred
+  // scalar value in the NodeArg if any.
+  if (dim_size == 0) {
+    if (node.OpType() == "Gather") {
+      // Following code extracts an element from a 1D array if all conditions are met.
+      // e.g.
```
yuslepukhin: I suggest make different operators logic into a separate function. #Resolved

onnxruntime/test/framework/shape_inference_test.cc
chilo-ms: It seems we don't have a C++ wrapper for `AddFreeDimensionOverrideByName()` C API.

onnxruntime/core/graph/graph.cc
yuslepukhin: Actually, this whole context may warrant separate header and .cc files.

onnxruntime/test/framework/shape_inference_test.cc
yuslepukhin: Let's add one and this test would provider coverage.

onnxruntime/core/graph/graph.cc
chilo-ms: The explanation is added

onnxruntime/core/graph/graph.cc
chilo-ms: added

onnxruntime/core/graph/graph.cc
chilo-ms: The reallocation won't happen as we `node_output_types_.resize(node.OutputDefs().size())` in constructor.

onnxruntime/core/graph/graph.cc
chilo-ms: added

onnxruntime/core/graph/graph.cc
chilo-ms: The explanation is added

onnxruntime/core/graph/graph.cc
chilo-ms: no, this is not a virtual function in `DataPropagationContext`

onnxruntime/core/graph/graph.cc
chilo-ms: added

onnxruntime/core/graph/graph.cc
chilo-ms: changed to reference

onnxruntime/core/graph/graph.cc
chilo-ms: Do we really need to?

GetConstantInitializer() returns a pointer to a TensorProto. My understanding is that TensorProto already implies a tensor type.
As for whether to check shape is preset, the initializer could be a tensor or a scalar, for a scalar, there is no shape.

onnxruntime/core/graph/graph.cc
chilo-ms: Hmm, the reason i'm not putting the resize() outside if/else is that there are still cases that it doesn't meet all the "if" and "else if" condition and ended up not getting the values.

Basically, making sure we will get the values and then resize the vector.

onnxruntime/core/graph/graph.cc
chilo-ms: I think above code still works, as it dereferences the "i-th element" of the tensor's internal data and assigns that value to input_values[i]. Correct me if i'm wrong.

onnxruntime/core/graph/graph.cc
chilo-ms: I think that's okay, as this helper function aims to get the "axes" or "indices" inputs of the operates that are all INT32/INT64 data type. 
If it's neither 32 or 64, we end up won't get the values, and won't shape inference it and won't save the inferred shape values. It won't make the whole shape inference fail.

onnxruntime/core/graph/graph.cc
chilo-ms: Add separate headers and .cc files for each OP's custom data propagation.

onnxruntime/core/graph/graph.cc
chilo-ms: It's a dimension index or an axis.

onnxruntime/core/graph/graph.cc
yuslepukhin: Here we are dealing with `TypeProto` which can contain anything.

onnxruntime/core/graph/graph.cc
```diff
@@ -2742,9 +2784,215 @@ class InferenceContextImpl : public ONNX_NAMESPACE::InferenceContext {
...
+      -> Status {
+    const TensorProto* initializer = this->GetConstantInitializer(input_name, true);
+
+    if (initializer) {
+      // Get shape from TensorProto and calculate element counts.
+      // If shape has dimension size equals zero, it means it's a scalar and has only one element.
+      size_t element_cnt = 1;
+      for (auto& dim : initializer->dims()) {
+        element_cnt *= dim;
+      }
```
yuslepukhin: Bad merge here? #Resolved

onnxruntime/core/graph/graph.cc
yuslepukhin: I am still not getting it.

onnxruntime/core/graph/graph.cc
yuslepukhin: Data() is expensive in the loop

onnxruntime/core/graph/graph.cc
yuslepukhin: resize() is also not ideal but in this case shape is small

onnxruntime/core/graph/data_propagation/unsqueeze_op_data_propagation.cc
```diff
@@ -0,0 +1,110 @@
...
+          return Status::OK();
+        }
+      }
+    }
+
+    if (tensor_shape_proto.dim_size() > 0) {
+      // Get axes value
+      TensorShapeVector axes;
+#if !defined(ORT_MINIMAL_BUILD) && !defined(ORT_EXTENDED_MINIMAL_BUILD)
+      InlinedHashSet<int64_t> axes_set;
```
yuslepukhin, line 50: This `ifdef` should already be present in the inlined_containers header. #Resolved

onnxruntime/core/graph/data_propagation/unsqueeze_op_data_propagation.cc
```diff
@@ -0,0 +1,110 @@
...
+          }
+
+          auto value = dim.dim_value();
+          inferred_shape_values->add_dim()->set_dim_value(value);
+
+          axis += 1;
+        }
+      }
+      ORT_CATCH(const std::exception& ex) {
+        ORT_HANDLE_EXCEPTION([&]() {
```
yuslepukhin, line 88: Just curious what can possibly throw here, we are not calling onnx API? #Resolved

onnxruntime/test/framework/shape_inference_test.cc
yuslepukhin: This is not addressed

onnxruntime/core/graph/graph.cc
chilo-ms: i was checking the wrong code snippet previously and was confused by your comment.
Now i got it, you are referring to the TypeProto and i added the check now.

onnxruntime/core/graph/graph.cc
chilo-ms: converted dim which is an int64 to size_t first and then do the computation.

onnxruntime/core/graph/graph.cc
chilo-ms: then i think we can keep using resize() here?
What is the ideal solution here?

onnxruntime/core/graph/graph.cc
chilo-ms: changed to use memcpy

onnxruntime/test/framework/shape_inference_test.cc
chilo-ms: Ah, forgot to reply here. How about i will create another PR to add the C++ wrapper for it?

onnxruntime/core/graph/data_propagation/unsqueeze_op_data_propagation.cc
chilo-ms: only DISABLE_ABSEIL is present in inlined_containers.h but in minimal build the pipeline failed to build `InlinedHashSet`.

````
/onnxruntime_src/onnxruntime/core/graph/data_propagation/squeeze_op_data_propagation.cc:34:31: error: aggregate ‘onnxruntime::InlinedHashSet<long int> axes_set’ has incomplete type and cannot be defined
   34 |       InlinedHashSet<int64_t> axes_set;
      |                               ^~~~~~~~
````

File: onnxruntime/core/graph/data_propagation/unsqueeze_op_data_propagation.cc (hunk 34 above)
Comment (chilo-ms) [reply onnxruntime/core/graph/data_propagation/unsqueeze_op_data_propagation.cc:88]: It's `HandleNegativeAxis()` which could throw.

File: onnxruntime/core/graph/data_propagation/unsqueeze_op_data_propagation.cc (hunk 33 above)
Comment (yuslepukhin) [reply onnxruntime/core/graph/data_propagation/unsqueeze_op_data_propagation.cc:50]: Why we do not do #ifdefs in other parts of the code for this?

File: onnxruntime/test/framework/shape_inference_test.cc (hunk 9 above)
Comment (yuslepukhin) [reply onnxruntime/test/framework/shape_inference_test.cc]: It is just a few lines of code, why not here?

File: onnxruntime/core/graph/data_propagation/data_propagation.h [hunk 35]
```diff
@@ -0,0 +1,40 @@
...
+
+ protected:
+  CustomDataPropagation(const Node& node,
+                        NodeArg& output_def,
+                        std::function<Status(const std::string&, TensorShapeVector&)> func,
+                        const ONNX_NAMESPACE::TypeProto& output_from_onnx_op_data_propagation,
+                        const logging::Logger& logger) noexcept
+      : node_(node),
+        output_def_(output_def),
+        get_initialized_input_values_func_(func),
```
yuslepukhin: std::move() #Resolved

onnxruntime/core/graph/data_propagation/data_propagation.h
```diff
@@ -0,0 +1,40 @@
...
+#include "core/common/logging/logging.h"
+#include <onnx/onnx-ml.pb.h>
+
+namespace onnxruntime {
+
+/**
+ * @class CustomDataPropagation
+ * Custom data propagation for the operator to help enhance shape inference.
+ */
+class CustomDataPropagation {
```
yuslepukhin: Here in other classes prohibit copying usig pre-defined macros. #Resolved

onnxruntime/core/graph/data_propagation/add_op_data_propagation.cc
```diff
@@ -0,0 +1,25 @@
...
+namespace onnxruntime {
+
+Status AddOpDataPropagation::infer() {
+  // Get "A" input
+  const auto* input_0 = node_.InputDefs()[0];
+  // Get "B" input
+  const auto* input_1 = node_.InputDefs()[1];
+
+  if (input_0->GetInferredShapeScalarValue().has_value() && input_1->GetInferredShapeScalarValue().has_value()) {
+    output_def_.SetInferredShapeScalarValue(input_0->GetInferredShapeScalarValue().value() + input_1->GetInferredShapeScalarValue().value());
```
yuslepukhin: Not clear here. `Add` is an elementwise operation. I would think the output shape either the same as the shape of two inputs or the result of broadcasting.
If both inputs are scalars, then the output is also a scalar with a single element? #Resolved

onnxruntime/core/graph/data_propagation/data_propagation.h
yuslepukhin: Suggest renaming the class as CustomDataPropagationBase as it is done for kernels.
Also, suggest merging the base header with the factory header and .cc file.
Also rename the file to custom_data_propagation.h

onnxruntime/core/graph/data_propagation/div_op_data_propagation.h
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include "data_propagation.h"
+#include "core/graph/graph.h"
+namespace onnxruntime {
+
+class DivOpDataPropagation : public CustomDataPropagation {
```
yuslepukhin: Need doc that is specific for Div #Resolved

onnxruntime/core/graph/data_propagation/add_op_data_propagation.h
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include "data_propagation.h"
+#include "core/graph/graph.h"
+namespace onnxruntime {
+
+class AddOpDataPropagation : public CustomDataPropagation {
```
yuslepukhin: Add doc about what this does explaing the logic and why only scalars are handled. #Resolved

onnxruntime/core/graph/data_propagation/div_op_data_propagation.cc
```diff
@@ -0,0 +1,25 @@
...
+namespace onnxruntime {
+
+Status DivOpDataPropagation::infer() {
+  // Get "A" input
+  const auto* input_0 = node_.InputDefs()[0];
+  // Get "B" input
+  const auto* input_1 = node_.InputDefs()[1];
+
+  if (input_0->GetInferredShapeScalarValue().has_value() && input_1->GetInferredShapeScalarValue().has_value()) {
+    output_def_.SetInferredShapeScalarValue(input_0->GetInferredShapeScalarValue().value() / input_1->GetInferredShapeScalarValue().value());
```
yuslepukhin: I think this is similar to `Add`. Div applies to data, not to shapes unless I totally misunderstand it. Well, the docs would hopefully clear it. #Resolved

onnxruntime/core/graph/data_propagation/gather_op_data_propagation.h
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include "data_propagation.h"
+#include "core/graph/graph.h"
+namespace onnxruntime {
+
+class GatherOpDataPropagation : public CustomDataPropagation {
```
yuslepukhin: Docs #Resolved

onnxruntime/core/graph/data_propagation/gather_op_data_propagation.cc
```diff
@@ -0,0 +1,68 @@
...
+
+namespace onnxruntime {
+
+Status GatherOpDataPropagation::infer() {
+  int dim_size = 0;
+  if (output_from_onnx_op_data_propagation_.has_tensor_type() &&
+      output_from_onnx_op_data_propagation_.tensor_type().has_shape()) {
+    dim_size = output_from_onnx_op_data_propagation_.tensor_type().shape().dim_size();
+  }
+  ORT_ENFORCE(dim_size == 0);
```
yuslepukhin: Not clear here, comments would help #Resolved

onnxruntime/core/graph/data_propagation/data_propagation_factory.cc
```diff
@@ -0,0 +1,71 @@
...
+ * @param node The ORT's node
+ * @param output_def The node's output NodeArg to save the inferred shape values if needed
+ * @param func Helper function to get the input value if it's a initializer
+ * @param output_from_onnx_op_data_propagation The result from executing ONNX operator's data propagation
+ * @param logger The reference to a logger
+ * @return std::unique_ptr<CustomDataPropagation> Returns a CustomDataPropagation object if available
+ */
+std::unique_ptr<CustomDataPropagation> CreateCustomDataPropagation(const Node& node,
+                                                                   NodeArg& output_def,
+                                                                   std::function<Status(const std::string&, TensorShapeVector&)> func,
```
yuslepukhin: This function must be moved into each of the constructors since they are by value.

onnxruntime/core/graph/data_propagation/gather_op_data_propagation.cc
```diff
@@ -0,0 +1,68 @@
...
+  // Index value is expected to be within bounds [-s, s-1] along axis of size s
+  if (input_0->GetInferredShapeValues().has_value()) {
+    const auto& tensor_shape_proto = input_0->GetInferredShapeValues().value();
+
+    // If "indices" input is a scalar, then the size of indices is 1.
+    if (indices.size() == 1) {
+      ORT_TRY {
+        auto& dim = tensor_shape_proto.dim(static_cast<int32_t>(HandleNegativeAxis(indices[0], tensor_shape_proto.dim_size())));
+        if (dim.has_dim_value()) {
+          output_def_.SetInferredShapeScalarValue(dim.dim_value());
```
yuslepukhin, line 69: What is the meaning of the result here? If indices are scalar, the output is not scalar according to the docs. #Resolved

onnxruntime/core/graph/graph.cc
```diff
@@ -2698,8 +2700,58 @@ class InferenceContextImpl : public ONNX_NAMESPACE::InferenceContext {
...
+
+    // Converts the inferred shape scalar value if any to a TensorProto and returns the TensorProto.
+    //
+    // Note: ONNX's getShapeInput() internally calls getInputData() to retrieve a TensorProto (if available)
+    //       and then extracts shape/dimension values from it. As a result, the scalar value may not be
+    //       properly handled and propagated in ONNX's shape inference.
+    //       However, Graph::SaveShapeValuesFromDataPropagation() properly handles data propagation for
+    //       some operators.
+    if (inferred_shape_scalar_value.has_value()) {
+      TensorProto tensor_proto;
```
yuslepukhin, line 2791: It is not clear how we use this? #Resolved

onnxruntime/core/graph/data_propagation/add_op_data_propagation.cc
chilo-ms: the Add's custom data propagation here is not doing the shape inference, it's actually doing the inference on the "shape inputs" and save the output values/scalar in node_arg for later use.

Same for all other operators' custom data propagation.

I updated the class description and hopefully that help understand the code.

onnxruntime/core/graph/data_propagation/data_propagation.h
chilo-ms: updated

onnxruntime/core/graph/data_propagation/data_propagation.h
chilo-ms: updated

onnxruntime/core/graph/data_propagation/div_op_data_propagation.cc
chilo-ms: doc also added.
As explained above, the op's custom data propagation here is not doing the shape inference, it's actually doing the inference on the "shape inputs" and save the output values/scalar in node_arg for later use.

onnxruntime/core/graph/data_propagation/gather_op_data_propagation.cc
chilo-ms: Re: If indices are scalar, the output is not scalar according to the docs.
Hmm, i think the output is scalar.

As per ONNX doc, Gather 13:
```
..
Inputs
data (heterogeneous) - T:

Tensor of rank r >= 1.

indices (heterogeneous) - Tind:

Tensor of int32/int64 indices, of any rank q. All index values are expected to be within bounds [-s, s-1] along axis of size s. It is an error if any of the index values are out of bounds.

Outputs
output (heterogeneous) - T:

Tensor of rank q + (r - 1).
```

If 'data' input has rank 1 and 'indices' input is scalar (rank 0) and output will be scalar as rank is 0.

onnxruntime/core/graph/graph.cc
chilo-ms: Good catch, update the code.

include/onnxruntime/core/graph/node_arg.h
```diff
@@ -128,6 +142,23 @@ class NodeArg {
...
+  // This variable stores the actual tensor data of the shape as a TensorShapeProto after executing
+  // the ONNX operator's PartialDataPropagationFunction(). It's used for shape inference purpose.
+  //
+  // Calling an operator's TypeAndShapeInferenceFunction() alone is sometimes insufficient
+  // for complete shape inference. For example, the Shape operator's TypeAndShapeInferenceFunction()
+  // only provides the output's rank which is 1 but not its actual shape values.
+  //
+  // The PartialDataPropagationFunction(), defined in the ONNX operator schema, must also
+  // be executed to obtain the concrete output shape values, allowing accurate propagation
+  // of shape information throughout the graph.
```
yuslepukhin: Add when the concrete output is not computed and that is why this is optional. #Resolved

include/onnxruntime/core/graph/graph.h
```diff
@@ -1760,6 +1760,15 @@ class Graph {  // NOLINT(clang-analyzer-optin.performance.Padding): preserve exi
                                                     std::vector<const ONNX_NAMESPACE::TypeProto*>& output_types,
                                                     const Graph::ResolveOptions& options);
 
+  // If the shape values are inferred after executing ONNX operator's PartialDataPropagationFunction(),
```
yuslepukhin: It is not always shape of the output, but the shape is in the output. I think the wording may need to be updated to make it clear. #Resolved

onnxruntime/core/graph/data_propagation/gather_op_data_propagation.cc
yuslepukhin: This throws. Should this return an error code? In any case, the error message would help.
Also, I understand why I raised this question in the first place. The var is declared as if it is going to be used later. But it is not necessary. The condition can be checked inside the above if and error is raised if necessary.

onnxruntime/core/graph/data_propagation/gather_op_data_propagation.cc
```diff
@@ -0,0 +1,71 @@
...
+  ORT_ENFORCE(dim_size == 0);
+
+  // Following code extracts an element from a 1D array if all conditions are met.
+  // e.g.
+  // shape data is [1, 3, 64, 64] -> gets 64 if the index is 2.
+  // shape data is [1, 3, 64, 64] -> gets 3 if the index is 1.
+
+  // Get "data" input
+  // Note: The "data" input should be a one dimension array in this case.
+  const auto* input_0 = node_.InputDefs()[0];
```
yuslepukhin, line 34: Here an all other derived classes. Should we check if the input exists? #Resolved

onnxruntime/core/graph/data_propagation/gather_op_data_propagation.cc
```diff
@@ -0,0 +1,71 @@
...
+
+  // Get "indices" input
+  // Note: The "indices" input could be one of the three cases:
+  //       1. A tensor with rank > 0 and all tensor values are known.
+  //       2. A tensor with rank > 0 but not all tensor values are know.
+  //       3. A scalar.
+  //
+  //       If it's case #1, ONNX operator's PartialDataPropagationFunction()
+  //       should have inferred the output shape value.
```
yuslepukhin, line 43: Mention what happens in case #2. #Resolved

onnxruntime/core/graph/data_propagation/size_op_data_propagation.h
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#pragma once
+
+#include "custom_data_propagation.h"
+#include "core/graph/graph.h"
+
+namespace onnxruntime {
+
```
yuslepukhin, line 10: Lacks docs. Just a few sentenses would be helpful. Is this handling outputs of the Shape()? #Resolved

onnxruntime/core/graph/data_propagation/squeeze_op_data_propagation.cc
```diff
@@ -0,0 +1,100 @@
...
+        }
+      }
+    }
+
+    if (tensor_shape_proto.dim_size() == 1) {
+      output_def_.SetInferredShapeScalarValue(tensor_shape_proto.dim(0).dim_value());
+    } else if (tensor_shape_proto.dim_size() > 1) {
+      // Get axes value
+      TensorShapeVector axes;
+#if !defined(ORT_MINIMAL_BUILD) && !defined(ORT_EXTENDED_MINIMAL_BUILD)
```
yuslepukhin: Stil not happy with that. Let's find out why in other parts of the code this is not required. May be you should just include `inlined_containers.h here`? #Resolved

onnxruntime/core/graph/data_propagation/gather_op_data_propagation.cc
```diff
@@ -0,0 +1,71 @@
...
+
+    if (indices.size() == 1) {
+      ORT_TRY {
+        // Note: Index value is expected to be within bounds [-s, s-1] along axis of size s
+        auto& dim = tensor_shape_proto.dim(static_cast<int32_t>(HandleNegativeAxis(indices[0], tensor_shape_proto.dim_size())));
+        if (dim.has_dim_value()) {
+          output_def_.SetInferredShapeScalarValue(dim.dim_value());
+        }
+      }
+      ORT_CATCH(const std::exception& ex) {
```
yuslepukhin: Here and other places. I think the most likely source of exception is the function that is passed used, so it would make sense to include it under the try. Other things are mostly exception free. #Resolved

onnxruntime/core/graph/data_propagation/sub_op_data_propagation.cc
```diff
@@ -0,0 +1,25 @@
...
+namespace onnxruntime {
+
+Status SubOpDataPropagation::infer() {
+  // Get "A" input
+  const auto* input_0 = node_.InputDefs()[0];
+  // Get "B" input
+  const auto* input_1 = node_.InputDefs()[1];
+
+  if (input_0->GetInferredShapeScalarValue().has_value() && input_1->GetInferredShapeScalarValue().has_value()) {
+    output_def_.SetInferredShapeScalarValue(input_0->GetInferredShapeScalarValue().value() - input_1->GetInferredShapeScalarValue().value());
```
yuslepukhin: Here and other places, need to honor 120 chars limit. #Resolved

onnxruntime/core/graph/graph.cc
```diff
@@ -2742,9 +2796,218 @@ class InferenceContextImpl : public ONNX_NAMESPACE::InferenceContext {
...
+  // Helper function to get the input value if it's a initializer.
+  auto get_initialized_input_values_func = [&](const std::string& input_name, TensorShapeVector& input_values)
+      -> Status {
+    const TensorProto* initializer = this->GetConstantInitializer(input_name, true);
+
+    if (initializer) {
+      // Get shape from TensorProto and calculate element counts.
+      // If shape has dimension size equals zero, it means it's a scalar and has only one element.
+      size_t element_cnt = 1;
+      for (auto& dim : initializer->dims()) {
```
yuslepukhin: Can we use `utils::TensorShapeFromTensorProto` #Resolved

onnxruntime/core/graph/graph.cc
```diff
@@ -2742,9 +2796,218 @@ class InferenceContextImpl : public ONNX_NAMESPACE::InferenceContext {
...
+        } else {
+          // If we can't get the OrtValue, it is a bug
+          ORT_THROW("Initializer ", input_name,
+                    " has in-memory external data but cannot get OrtValue during shape inference");
+        }
+      }
+      // Unpack tensor from raw data, external data (not in memory) or the type specific data field
+      else {
+        if (initializer->data_type() == TensorProto_DataType_INT32) {
+          std::vector<int32_t> tmp_values;
```
yuslepukhin: InlinedVector #Resolved

onnxruntime/core/graph/graph.cc
```diff
@@ -2742,9 +2796,218 @@ class InferenceContextImpl : public ONNX_NAMESPACE::InferenceContext {
...
+      if (utils::HasExternalDataInMemory(*initializer)) {
+        // Try to get the OrtValue for this initializer
+        OrtValue ort_value;
+        if (this->GetOrtValueInitializer(input_name, ort_value, true)) {
+          const Tensor& tensor = ort_value.Get<Tensor>();
+          if (initializer->data_type() == TensorProto_DataType_INT32) {
+            input_values.resize(element_cnt);
+            for (size_t i = 0; i < element_cnt; ++i) {
+              input_values[i] = static_cast<int64_t>(tensor.Data<int32_t>()[i]);
+            }
```
yuslepukhin: This still calls Data<>() in a loop. DataAsSpan() first, then you can use index. #Resolved

onnxruntime/core/graph/graph.cc
```diff
@@ -2742,9 +2796,218 @@ class InferenceContextImpl : public ONNX_NAMESPACE::InferenceContext {
...
+                                        get_initialized_input_values_func,
+                                        onnx_inferred_type_after_data_propagation,
+                                        logger_);
+  if (dp) {
+    ORT_RETURN_IF_ERROR(dp->infer());
+    return Status::OK();
+  }
+
+  // If no custom data propagation is defined for the operator,
+  // fall back to using ONNX's PartialDataPropagationFunction(), if available.
```
yuslepukhin: It is using the results of ONNX propagation, not the function itself. #Resolved

include/onnxruntime/core/graph/node_arg.h
chilo-ms: Added

include/onnxruntime/core/graph/graph.h
chilo-ms: updated the wording

onnxruntime/core/graph/data_propagation/gather_op_data_propagation.cc
chilo-ms: Added the check for input existence

onnxruntime/core/graph/data_propagation/gather_op_data_propagation.cc
chilo-ms: Added

onnxruntime/core/graph/data_propagation/size_op_data_propagation.h
chilo-ms: Added

onnxruntime/core/graph/data_propagation/squeeze_op_data_propagation.cc
chilo-ms: Good catch, `inlined_containers.h` added

onnxruntime/core/graph/data_propagation/gather_op_data_propagation.cc
chilo-ms: The function that is passed in and `HandleNegativeAxis()` can throw, and i put try/catch around them.

onnxruntime/core/graph/graph.cc
chilo-ms: Updated

onnxruntime/core/graph/graph.cc
chilo-ms: updated

onnxruntime/core/graph/graph.cc
chilo-ms: updated

onnxruntime/core/graph/graph.cc
chilo-ms: updated

onnxruntime/core/graph/data_propagation/gather_op_data_propagation.cc
```diff
@@ -0,0 +1,85 @@
...
+  //       should have inferred the output shape value.
+  //       If it's case #2, neither ONNX operator's PartialDataPropagationFunction()
+  //       nor Gather's custom data propagation can handle it.
+  //       This Gather's custom data propagation handles case #3.
+  const auto* input_1 = node_.InputDefs()[1];
+
+  // Return and do nothing if input doesn't exist
+  if (!input_0 || !input_1) {
+    return Status::OK();
+  }
```
yuslepukhin, line 52: Here and in other classes.
While checking for null is good, NodeArg has a method Exist(). This is for inputs that has empty names.
This is the way ONNX standard designates inputs, that were not supplied within the model.
We need to call and check for that. #Resolved

onnxruntime/core/graph/data_propagation/mul_op_data_propagation.cc
```diff
@@ -0,0 +1,32 @@
...
+  // Get "A" input
+  const auto* input_0 = node_.InputDefs()[0];
+  // Get "B" input
+  const auto* input_1 = node_.InputDefs()[1];
+
+  // Return and do nothing if input doesn't exist
+  if (!input_0 || !input_1) {
+    return Status::OK();
+  }
+
```
yuslepukhin, line 22: Same here

onnxruntime/core/graph/data_propagation/size_op_data_propagation.cc
```diff
@@ -0,0 +1,41 @@
...
+
+Status SizeOpDataPropagation::infer() {
+  // Size operator generates a scalar output
+  const auto* input_0 = node_.InputDefs()[0];
+
+  // Return and do nothing if input doesn't exist
+  if (!input_0) {
+    return Status::OK();
+  }
+
```
yuslepukhin, line 19: Same here

onnxruntime/core/graph/data_propagation/unsqueeze_op_data_propagation.cc
```diff
@@ -0,0 +1,119 @@
...
+
+namespace onnxruntime {
+
+Status UnsqueezeOpDataPropagation::infer() {
+  const auto* input_0 = node_.InputDefs()[0];
+
+  // Return and do nothing if input doesn't exist
+  if (!input_0) {
+    return Status::OK();
+  }
```
yuslepukhin, line 19: Same here

onnxruntime/core/graph/data_propagation/sub_op_data_propagation.cc
```diff
@@ -0,0 +1,32 @@
...
+Status SubOpDataPropagation::infer() {
+  // Get "A" input
+  const auto* input_0 = node_.InputDefs()[0];
+  // Get "B" input
+  const auto* input_1 = node_.InputDefs()[1];
+
+  // Return and do nothing if input doesn't exist
+  if (!input_0 || !input_1) {
+    return Status::OK();
+  }
```
yuslepukhin, line 21: Same here #Resolved

onnxruntime/core/graph/graph.cc
```diff
@@ -2784,9 +2838,223 @@ class InferenceContextImpl : public ONNX_NAMESPACE::InferenceContext {
...
+  // Helper function to get the input value if it's a initializer.
+  auto get_initialized_input_values_func = [&](const std::string& input_name, TensorShapeVector& input_values)
+      -> Status {
+    const TensorProto* initializer = this->GetConstantInitializer(input_name, true);
+
+    if (initializer) {
+      // Get shape from TensorProto as well as element counts.
+      // If shape has dimension size equals zero, it means it's a scalar and has only one element.
+      auto tensor_shape = utils::GetTensorShapeFromTensorProto(*initializer);
+      size_t element_cnt = static_cast<size_t>(tensor_shape.Size());
```
yuslepukhin: >static_cast [](http://example.com/codeflow?start=27&length=11)

use narrow<>, important on 32-bit systems

PR #26273, opened 2025-10-10
[WebNN] Remove constraints for Gemm's C input

Honry:
Now WebNN implementation for gemm's C operand has supported unidirectional broadcasting, which is align with ONNX spec. Removing constraints for Gemm's C input as which should be covered in ORT kernel.


Honry:

PR #26280, opened 2025-10-12
[LARCH64 CPU]Provide inference acceleration optimization for Loongson CPU with 4-bit quantized models

movedancer:
### Description
This submission is a 4-bit quantized matrix multiplication operator suitable for the Loongson platform. It has passed the internal test checks of ONNX and has been successfully deployed for actual inference on the Loongson platform. It includes five modifications:
(1) **sqnbitgemm_kernel_lasx.cpp**: Acceleration of inference for 4-bit quantized models on the LoongArch64 architecture, utilizing lasx/lsx vector instruction sets;
(2) **sqnbitgemm_kernel_lasx_common.h**: Implementation of auxiliary functions used by **sqnbitgemm_kernel_lasx.cpp**`;
(3) **cmake**: Added compilation options for **sqnbitgemm_kernel_lasx.cpp** under the LoongArch64 architecture;
(4) **mlasi.h**: Added interface for calling the operator in **sqnbitgemm_kernel_lasx.cpp** under the LoongArch64 architecture;
(5) **platform.cpp**: Added calls to the operators in **sqnbitgemm_kernel_lasx.cpp** under the LoongArch64 architecture.

### Motivation and Context
Loongson has a critical lack of key operations in ONNX quantized model inference tasks. 
The issue of poor inference performance for 4-bit quantized models on the Loongson platform has been addressed. In tests using the Deepseek-R1-1.5B model, our operators have increased TPS by more than 7 times, with the speed of quantization matrix dequantization improving by up to 3 times.

### Pictures
Dequantization Acceleration：
In the chart, the vertical axis represents time in milliseconds (ms), the horizontal axis represents the number of test matrices, and the size of the quantized matrix is rows × columns, such as the 1536*256.
反量化加速

### DIFF
--- a/onnxruntime/core/mlas/lib/sqnbitgemm_kernel_lasx_common.h
+++ b/onnxruntime/core/mlas/lib/sqnbitgemm_kernel_lasx_common.h
@@ -7,6 +7,49 @@
 #pragma once
 #include "qnbitgemm.h"
 #include "core/common/safeint.h"
+#include <memory>
+
+template<typename T, size_t RequiredAlignment = 0>
+struct MlasAlignedAllocator {
+    using value_type = T;
+
+    MlasAlignedAllocator() = default;
+
+    template<typename U, size_t A>
+    MlasAlignedAllocator(const MlasAlignedAllocator<U, A>&) {}
+
+    T* allocate(size_t n) {
+        // If RequiredAlignment > 0, use the required value
+        // Otherwise, use the value of MlasGetPreferredBufferAlignment()
+        size_t alignment = RequiredAlignment > 0 ?
+                          RequiredAlignment :
+                          MlasGetPreferredBufferAlignment();
+
+        size_t size = n * sizeof(T);
+        if (size % alignment != 0)  // check the size
+            size = ((size + alignment - 1) / alignment) * alignment;
+        #if defined(_MSC_VER)
+            void* ptr = _aligned_malloc(size, alignment);
+        #else
+            void* ptr = aligned_alloc(alignment, size);
+        #endif
+        if (!ptr) throw std::bad_alloc();
+        return static_cast<T*>(ptr);
+    }
+
+    void deallocate(T* ptr, size_t) {
+        #if defined(_MSC_VER)
+            _aligned_free(ptr);
+        #else
+            free(ptr);
+        #endif
+    }
+
+    template<typename U>
+    struct rebind {
+        using other = MlasAlignedAllocator<U, RequiredAlignment>;
+    };
+};
 
 static MLAS_FORCEINLINE __m256
 __lasx_xvzero()
@@ -310,7 +353,7 @@ Q8ComputePackBlkSum(
 )
 {
     SafeInt<size_t> size =  SafeInt<size_t>(N) * BlockCountK;
-    std::vector<float> QuantBScaleBeginCopy(size.Value());
+    std::vector<float, MlasAlignedAllocator<float, 32>> QuantBScaleBeginCopy(size.Value());
     std::copy(QuantBScaleBegin, QuantBScaleBegin + N * BlockCountK, QuantBScaleBeginCopy.begin());
 
     MlasTrySimpleParallel(ThreadPool, N * BlockCountK, [&](ptrdiff_t tid) {


movedancer:

onnxruntime/core/mlas/lib/sqnbitgemm_kernel_lasx.cpp
```diff
@@ -0,0 +1,1066 @@
...
+            size_t kklen = std::min((int)SubBlkLen32, (int)(ck - kk));
+
+            __m256 av0_8_ps = load_float_n_lasx(ARowPtr + k + kk, std::min<size_t>(kklen, 8));
+            __m256 av1_8_ps = load_float_n_lasx(ARowPtr + k + kk + 8, std::min<size_t>(kklen > 8 ? kklen - 8 : 0, 8));
+            __m256 av2_8_ps = load_float_n_lasx(ARowPtr + k + kk + 16, std::min<size_t>(kklen > 16 ? kklen - 16 : 0, 8));
+            __m256 av3_8_ps = load_float_n_lasx(ARowPtr + k + kk + 24, std::min<size_t>(kklen > 24 ? kklen - 24 : 0, 8));
+
+            if constexpr (IsBlkLen64Layout) {
+                count_half_4 = 4 * (int)((kk % (2 * SubBlkLen32)) / SubBlkLen32);
+            }
```
snnn, line 307: The line `float sum = acc_buf[0] + acc_buf[1] + acc_buf[2] + acc_buf[3] + acc_buf[4] + acc_buf[5] + acc_buf[6] + acc_buf[7];` could be replaced with a loop or an intrinsic for horizontal sum to improve readability and maintainability. The `hsum_float_8_lasx` function you've defined in `sqnbitgemm_kernel_lasx_common.h` seems perfect for this.

onnxruntime/core/mlas/lib/sqnbitgemm_kernel_lasx.cpp
```diff
@@ -0,0 +1,1066 @@
...
+    constexpr size_t SubBlkLen32 = 32;
+    constexpr size_t SubBlkStep16 = MlasQNBitBlkDataSizeInBytes(BlkBitWidth4, SubBlkLen32);
+    static_assert(SubBlkStep16 == 16);
+
+    __m256 acc[NCols];
+
+    alignas(32) static const float zero_array[8] = {0};
+    UnrolledLoop<NCols>([&](size_t i) {
+        acc[i] = (__m256)__lasx_xvld((void*)&zero_array, 0);
+    });
```
snnn, line 257: Please remove the `// DEBUG` comment.

onnxruntime/core/mlas/lib/sqnbitgemm_kernel_lasx.cpp
```diff
@@ -0,0 +1,1066 @@
...
+        BiasPtr += BiasPtr != nullptr ? 1 : 0;
+        SumPtr += 1;
+    }
+}
+
+// SQ4BitGemmM1Kernel_BlkLen32Plus_CompFp32_lasx
+template <bool HasZeroPoint>
+MLAS_FORCEINLINE void
+SQ4BitGemmM1Kernel_BlkLen32Plus_CompFp32_lasx(
+    size_t BlkLen,
```
snnn, line 602: The `constexpr` is commented out. Please either remove it or add a comment explaining why it's not used.

onnxruntime/core/mlas/lib/sqnbitgemm_kernel_lasx.cpp
```diff
@@ -0,0 +1,1066 @@
...
+
+        BiasPtr += BiasPtr != nullptr ? 1 : 0;
+        SumPtr += 1;
+    }
+}
+
+// SQ4BitGemmM1Kernel_BlkLen32Plus_CompFp32_lasx
+template <bool HasZeroPoint>
+MLAS_FORCEINLINE void
+SQ4BitGemmM1Kernel_BlkLen32Plus_CompFp32_lasx(
```
snnn, line 601: There is a typo in the comment `// TODO: constexpr use temaplte parameter`. It should be `template`.

onnxruntime/core/mlas/lib/sqnbitgemm_kernel_lasx.cpp
```diff
@@ -0,0 +1,1066 @@
...
+                acc[i] = __lasx_xvfmadd_s(fbv_24_31, av3_8_ps, acc[i]);
+            });
+        }
+
+        b_blk_data_ptr += MlasQNBitBlkDataSizeInBytes(BlkBitWidth4, BlkLen);
+        ++s;
+        if constexpr (HasZeroPoint) {
+            QuantBZeroPointIdx += 1;
+        }
+    }
```
snnn, line 369: Please remove the commented-out block of code for `zp_array`.

onnxruntime/core/mlas/lib/sqnbitgemm_kernel_lasx_common.h
```diff
@@ -0,0 +1,463 @@
...
+    size_t SubBlkLen,
+    size_t N,
+    float* QuantBScaleBegin,
+    const std::byte* QuantBZPBegin,
+    float* BlockSumBegin,
+    MLAS_THREADPOOL* ThreadPool,
+    const size_t BlockCountK
+)
+{
+    MlasTrySimpleParallel(ThreadPool, N * BlockCountK, [&](ptrdiff_t tid) {
```
snnn, line 105: For consistency, please use lowercase for hexadecimal literals: `0x0f`.

onnxruntime/core/mlas/lib/sqnbitgemm_kernel_lasx.cpp
```diff
@@ -0,0 +1,1066 @@
...
+    multiplication kernels for loongarch64. Accelerate inference
+    optimization using lasx/lsx vector instruction sets.
+
+--*/
+
+#include <lasxintrin.h>
+#include <lsxintrin.h>
+
+#include <algorithm>
+#include <cassert>
```
snnn, line 19: It seems that `<iostream>`, `<fstream>`, `<chrono>`, `<ctime>`, and `<mutex>` are included but not used in the final code. They might be from a debug build. Please remove them.

onnxruntime/core/mlas/lib/sqnbitgemm_kernel_lasx.cpp
```diff
@@ -0,0 +1,1066 @@
...
+    const size_t StrideQuantBZeroPoint = MlasQNBitZeroPointsForBlksSizeInBytes<BlkBitWidth4>(BlockCountK);
+
+    const float* BiasPtr = Bias;
+
+    const std::byte* QuantBDataColPtr = QuantBData;
+    const float* QuantBScaleColPtr = QuantBScale;
+    const std::byte* QuantBZeroPointColPtr = QuantBZeroPoint;
+
+    float* SumPtr = CRowPtr;
+
```
snnn, line 633: The `alignas(32)` specifier on a scalar local variable `zp` is unusual. Could you please explain why this alignment is necessary here?

onnxruntime/core/mlas/lib/sqnbitgemm_kernel_lasx_common.h
```diff
@@ -0,0 +1,463 @@
...
+static void
+Q8PackQuantBDataAndBlkSum_lasx(
+    size_t N,
+    size_t BlockCountK,
+    size_t BlkLen,
+    size_t SubBlkLen,
+    const std::byte* QuantBDataBegin,
+    const float* QuantBScaleBegin,
+    bool HasZeroPoint,
+    const std::byte* QuantBZPBegin,
```
snnn, line 409: Please translate the comments `// 如果你有这个 instr` and `// 向量全部 0` (on line 365) to English to make them accessible to all contributors.

onnxruntime/core/mlas/lib/sqnbitgemm_kernel_lasx_common.h
```diff
@@ -0,0 +1,463 @@
...
+    std::byte* PackedQuantBDataBegin,
+    MLAS_THREADPOOL* ThreadPool,
+    const size_t N,
+    const size_t BlockCountK,
+    const size_t BlkLen,
+    const size_t SubBlkLen
+)
+{
+    constexpr size_t BlkBitWidth = 8;
+    const size_t StrideN = BlockCountK * BlkLen;
```
snnn, line 296: For buffers that are used in performance-critical code, it is recommended to use an allocator that can provide aligned memory. Please consider using the default CPU allocator for this vector to ensure proper alignment, instead of relying on the default `std::allocator`.

snnn:
From a security perspective, an AI bot has reviewed the code and found several potential vulnerabilities, primarily related to integer overflows when calculating buffer sizes and memory offsets. The project's coding conventions mandate the use of `SafeInt` for such calculations to prevent these issues, but this practice is not followed in the new code.

Here is a summary of the findings, from most to least critical:

### Critical Vulnerabilities

1.  **Heap Buffer Overflow in `Q8ComputePackBlkSum`**:
    *   **File**: `onnxruntime/core/mlas/lib/sqnbitgemm_kernel_lasx_common.h`
    *   **Line**: 252
    *   **Issue**: The allocation `std::vector<float> QuantBScaleBeginCopy(N * BlockCountK);` is vulnerable to an integer overflow if `N * BlockCountK` exceeds `SIZE_MAX`. This would cause `std::vector` to allocate a much smaller buffer than required. The following `std::copy` would then write out of bounds, leading to heap corruption.
    *   **Recommendation**: Use `SafeInt<size_t>` to calculate the size of the vector to prevent overflow.
        ```cpp
        SafeInt<size_t> size = static_cast<SafeInt<size_t>>(N) * BlockCountK;
        std::vector<float> QuantBScaleBeginCopy(size.Value());
        ```

2.  **Integer Overflow in Buffer Size Calculation**:
    *   **File**: `onnxruntime/core/mlas/lib/sqnbitgemm_kernel_lasx.cpp`
    *   **Function**: `QNBitGemmPackQuantBDataSize_Lasx`
    *   **Issue**: This function calculates the total size required for a packed buffer. The calculations involve multiplications like `N * BlockCountK * ...` which can overflow. If this function returns a size that is smaller than the actual required size due to an overflow, the caller will allocate an undersized buffer, leading to heap overflows when this buffer is written to by packing functions.
    *   **Recommendation**: All size calculations in this function must use `SafeInt`. For example:
        ```cpp
        SafeInt<size_t> packed_quant_b_data_size = static_cast<SafeInt<size_t>>(N) * BlockCountK * MlasQNBitBlkDataSizeInBytes(BlkBitWidth, BlkLen);
        ```
        Apply this to all multiplications for `PackedQuantBDataSize`, `ScaleSize`, and `BlkSumSize`.

### High-Risk Vulnerabilities

1.  **Integer Overflows in Offset Calculations**:
    *   **Files**: `sqnbitgemm_kernel_lasx.cpp`, `sqnbitgemm_kernel_lasx_common.h`
    *   **Issue**: Throughout the code, there are numerous calculations for offsets into buffers, such as `n * BlockCountK`, `c * StrideN`, `BlockCountK * blk_data_size_in_bytes`. These are performed using raw `size_t` arithmetic and are prone to overflow. An attacker who can control tensor shapes in a model could potentially trigger an overflow, causing pointers to be miscalculated, which would lead to out-of-bounds reads or writes.
    *   **Examples**:
        *   `sqnbitgemm_kernel_lasx_common.h:55`: `const size_t src_blk_offset = n * BlockCountK + k_blk;`
        *   `sqnbitgemm_kernel_lasx_common.h:92`: `const size_t src_data_offset = n * BlockCountK * BlkDataSize + k_subblk * SubBlkDataSize;`
        *   `sqnbitgemm_kernel_lasx.cpp:600`: `const size_t b_data_col_stride_in_bytes = BlockCountK * blk_data_size_in_bytes;`
    *   **Recommendation**: All offset calculations involving multiplications should be performed using `SafeInt<size_t>` to prevent overflows.

### Medium-Risk Vulnerabilities

1.  **Unchecked `std::memcpy`**:
    *   **File**: `onnxruntime/core/mlas/lib/sqnbitgemm_kernel_lasx.cpp`
    *   **Lines**: e.g., 260, 366
    *   **Issue**: Functions like `ComputeDotProducts_BlkLen32Plus_CompFp32_lasx` use `std::memcpy` with a hardcoded size (e.g., 16 or 8 bytes) from a source pointer that is calculated with offsets. While the logic appears correct for valid inputs, there are no explicit checks to ensure that the source buffer is actually large enough to read from. If an integer overflow occurs during the offset calculation, these `memcpy` operations could read from arbitrary memory locations.
    *   **Recommendation**: While using `SafeInt` for offset calculations will mitigate most of this risk, it is good practice to be mindful of buffer boundaries when dealing with raw pointers.

### Low-Risk/Code Hygiene

1.  **Fragile Stack Buffer Usage in `load_float_n_lasx`**:
    *   **File**: `onnxruntime/core/mlas/lib/sqnbitgemm_kernel_lasx.cpp`
    *   **Line**: 189
    *   **Issue**: The function `load_float_n_lasx` uses a fixed-size stack buffer `buf[8]`. The loop `for (int i = 0; i < n; ++i)` writes to this buffer. While the current callers in this file all ensure `n <= 8`, this creates a fragile design that could be vulnerable to future changes.
    *   **Recommendation**: Add a check like `assert(n <= 8);` at the beginning of the function to make the precondition explicit and prevent future misuse.

movedancer:
Thank you for pointing out the flaws! I'll check and fix them right away.

onnxruntime/core/mlas/lib/sqnbitgemm_kernel_lasx_common.h
snnn: Check the usages of "MlasGetPreferredBufferAlignment" and think if we need to make any change to the function for Loongson .

movedancer: Thank you for your suggestions! I have checked the usage of the function MlasGetPreferredBufferAlignment and found that it is used to allocate memory aligned to 64 size_t units. However, the size of the quantization data blocks that this  function may need to calculate at once far exceeds this size. Therefore, I followed the example of the official operator **Q8ComputePackBlkSum** function in the existing **onnxruntime/core/mlas/lib/sqnbitgemm_kernel_avx_common.h** file(on line 353) and used a vector to allocate memory.
To test whether there might be any misaligned hardware errors, I used the built-in test library of ONNX to conduct operator tests. Under stress tests with different matrix sizes, our operator performed well. Therefore, for now, I still choose to use vector for memory allocation to avoid potential errors from modifying MlasGetPreferredBufferAlignment.   However, I will improve this method in the future.

movedancer:
Hello! I have made code modifications according to your suggestions, corrected all risks and errors, and look forward to your review.

onnxruntime/core/mlas/lib/sqnbitgemm_kernel_lasx_common.h
```diff
@@ -18,32 +19,32 @@ GetContinueLayoutOffsetSubBlk(size_t N, const size_t n, const size_t SubOrBlkCou
 {
     size_t T = n / 4, t = n % 4;
     bool te = T == N / 4;
-    size_t scale_dst_offset = T * 4 * SubOrBlkCountK;
+    SafeInt<size_t> scale_dst_offset = static_cast<SafeInt<size_t>>(T) * 4 * SubOrBlkCountK;
```
snnn: You don't need to have the static_cast. SafeInt<size_t>(T) should be enough.

onnxruntime/core/mlas/lib/sqnbitgemm_kernel_lasx_common.h
snnn: So, there is no alignment requirement for SIMD instructions for this CPU architecture?

movedancer: Sorry I missed this message earlier. The vector aligned allocation feature has already been added.

snnn (approved):

movedancer:
Thank you for the approval! I'm glad my contribution was accepted.

PR #26289, opened 2025-10-13
[WebNN] Fix some issues in reduction ops

Honry:
- Allow empty axes input
- When axes is empty and ‘noop_with_empty_axes’ is true, WebNN should set axes to []
- Simplify the code

### DIFF
--- a/onnxruntime/core/providers/webnn/builders/impl/reduction_op_builder.cc
+++ b/onnxruntime/core/providers/webnn/builders/impl/reduction_op_builder.cc
@@ -85,16 +85,12 @@ Status ReductionOpBuilder::AddToModelBuilderImpl(ModelBuilder& model_builder,
     }
   }
 
-  if (axes_data.empty()) {
-    // When axes is not provided or is empty, check the 'noop_with_empty_axes' attribute:
-    // - If it is false, perform reduction over all dimensions.
-    //   (In WebNN, this means the 'axes' option is not set.)
-    // - If it is true, no reduction is applied, but other operations are still performed.
-    //   (In WebNN, this requires setting 'axes' to an empty array.)
-    if (helper.Get("noop_with_empty_axes", 0) == 1) {
-      options.set("axes", emscripten::val::array());
-    }
-  } else {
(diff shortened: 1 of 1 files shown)


Honry:

onnxruntime/core/providers/webnn/builders/impl/reduction_op_builder.cc
```diff
@@ -53,71 +56,54 @@ Status ReductionOpBuilder::AddToModelBuilderImpl(ModelBuilder& model_builder,
...
+    //   (In WebNN, this means the 'axes' option is not set.)
+    // - If it is true, no reduction is applied, but other operations are still performed.
+    //   (In WebNN, this requires setting 'axes' to an empty array.)
+    if (helper.Get("noop_with_empty_axes", 0) == 1) {
+      options.set("axes", emscripten::val::array());
+    }
   } else {
-    return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT, "ReductionOpBuilder, unknown op: ", op_type);
+    options.set("axes", emscripten::val::array(GetNarrowedIntFromInt64<uint32_t>(axes_data)));
   }
```
fdwr: Would this work?

```c++
  // When axes is not provided or is empty, check the 'noop_with_empty_axes' attribute:
  // - If it is false, perform reduction over all dimensions.
  //   (In WebNN, this means the 'axes' option is not set.)
  // - If it is true, no reduction is applied, but other operations are still performed.
  //   (In WebNN, this requires setting 'axes' to an empty array.)
  if (!axes_data.empty() || helper.Get("noop_with_empty_axes", 0) == 1) {
    options.set("axes", emscripten::val::array(GetNarrowedIntFromInt64<uint32_t>(axes_data)));
  }
```

Honry: It should work. 👍

Honry: Addressed in new commit.

PR #26297, opened 2025-10-14
[webgpu] Support RotaryEmbedding in flash attention

xiaofeihan1:
### Description
We enabled Rotary Embedding (ROE) support in the Flash Attention code and introduced a fused operator, `FusedQKRotaryEmbeddingProgram`, which combines two ROE operations into a single kernel.

### Motivation and Context

Previously, Flash Attention for WebGPU did not support ROE, while the CUDA version did. As a result, different operators were used for the Phi-4 model. With this PR, both CUDA and WebGPU can share the same Phi-4 model implementation.

Fusing the ROE operations within the attention module helps reduce CPU-side overhead by:
	1.	Decreasing the number of MatMulNbits operations.
	2.	Reducing memcpyFromHost calls.

This optimization improves **token generation speed** on high-performance GPUs — **achieving over 5% speedup on NVIDIA 5080**

Test prompt: ~800tokens. 
prompt, gen(tps) | Before | Now(use phi4 model which is fused RoE into GQA)
-- | -- | --
NV 5080 | 2215, 119 | 2225, 126
Mac M3Max | 866, 69.4 | 868, 70.7
Intel | 360, 15.3 | 354, 15.5

We believe the generation phase is CPU-bound on such high-end GPUs, so reducing CPU time leads to noticeable gains.
However, the improvement is negligible on low-performance GPUs.

image

### DIFF
--- a/onnxruntime/contrib_ops/webgpu/bert/group_query_attention.cc
+++ b/onnxruntime/contrib_ops/webgpu/bert/group_query_attention.cc
@@ -146,14 +146,11 @@ Status RunFusedQKRotaryEmbedding(onnxruntime::webgpu::ComputeContext& context,
                                     hidden_size_k / head_size,
                                     static_cast<int64_t>(head_size - half_rotary_embedding_dim)});
   std::vector<uint32_t> k_global_dims(rank);
-  std::vector<uint32_t> k_global_strides(rank);
   for (size_t j = 0; j < rank; ++j) {
     k_global_dims[j] = gsl::narrow_cast<uint32_t>(k_global_shape[j]);
-    k_global_strides[j] = gsl::narrow_cast<uint32_t>(k_global_shape.SizeFromDimension(j + 1));
   }
 
   const auto q_domain_size = gsl::narrow_cast<uint32_t>(q_global_shape.Size());
-  const auto k_domain_size = gsl::narrow_cast<uint32_t>(k_global_shape.Size());
 
   const auto q_input_output_strides = std::vector<uint32_t>(
       {gsl::narrow_cast<uint32_t>(query_in->Shape().SizeFromDimension(1)),
@@ -189,10 +186,8 @@ Status RunFusedQKRotaryEmbedding(onnxruntime::webgpu::ComputeContext& context,
           {gsl::make_span(q_global_strides)},
           {gsl::make_span(q_input_output_strides)},
           {gsl::make_span(k_global_dims)},
-          {gsl::make_span(k_global_strides)},
           {gsl::make_span(k_input_output_strides)},
           {q_domain_size},
-          {k_domain_size},
       })
       .AddIndices(TensorShape{1, 1});
 
--- a/onnxruntime/contrib_ops/webgpu/bert/rotary_embedding.h
+++ b/onnxruntime/contrib_ops/webgpu/bert/rotary_embedding.h
@@ -37,17 +37,14 @@ class FusedQKRotaryEmbeddingProgram final : public Program<FusedQKRotaryEmbeddin
 
   // q_* describes query rotation domain (same definition as existing program)
   // k_* describes key rotation domain
-  // q_domain_size, k_domain_size let shader decide which branch to execute
   WEBGPU_PROGRAM_DEFINE_UNIFORM_VARIABLES(
       {"scale", ProgramUniformVariableDataType::Float32},
       {"q_global_shape", ProgramUniformVariableDataType::Uint32},
       {"q_global_stride", ProgramUniformVariableDataType::Uint32},
       {"q_input_output_stride", ProgramUniformVariableDataType::Uint32},
       {"k_global_shape", ProgramUniformVariableDataType::Uint32},
-      {"k_global_stride", ProgramUniformVariableDataType::Uint32},
       {"k_input_output_stride", ProgramUniformVariableDataType::Uint32},
-      {"q_domain_size", ProgramUniformVariableDataType::Uint32},
-      {"k_domain_size", ProgramUniformVariableDataType::Uint32});
+      {"q_domain_size", ProgramUniformVariableDataType::Uint32});
 
  private:
   const bool interleaved_;


guschmue:
super, wanted that one for some time :)
Can you resolve this small merge conflict ?

onnxruntime/contrib_ops/webgpu/bert/group_query_attention.cc
```diff
@@ -110,35 +110,93 @@ Status GeneratePositionIDs(onnxruntime::webgpu::ComputeContext& context, const W
...
+          {key_in, ProgramTensorMetadataDependency::Rank},
+          {&pos_ids, ProgramTensorMetadataDependency::Rank},
+          {cos_cache, ProgramTensorMetadataDependency::Rank},
+          {sin_cache, ProgramTensorMetadataDependency::Rank},
+      })
+      .AddOutputs({
+          {query_out, ProgramTensorMetadataDependency::Rank},
+          {key_out, ProgramTensorMetadataDependency::Rank},
+      })
+      .SetDispatchGroupSize((total + WORKGROUP_SIZE - 1) / WORKGROUP_SIZE)
```
qjia7: Maybe we can optimize it by `total = q_domain_size` not `total = q_domain_size + k_domain_size`. And get the updated Q and K in one pass not one after another in shader since Q has shape [batch_size, sequence_length, num_head, head_size] and K has shape [batch_size, sequence_length, num_head_k, head_size] and num_head_k is always less than or equal to num_head.
Please see comments in shader part.

onnxruntime/contrib_ops/webgpu/bert/rotary_embedding.cc
```diff
@@ -50,6 +50,71 @@ Status RotaryEmbeddingProgram::GenerateShaderCode(ShaderHelper& shader) const {
...
+      << "      let pos_ids_idx = " << position_ids.BroadcastedIndicesToOffset("bsnh.xy", dummy_indices) << ";\n"
+      << "      let position_id = u32(" << position_ids.GetByOffset("pos_ids_idx") << ") + select(0u, bsnh[1], pos_ids_idx == 0u);\n"
+      << "      let i = dot(bsnh, uniforms.q_input_output_stride) + select(0u, bsnh[3], " << interleaved_str << ");\n"
+      << "      let j = i + select(half_rotary_dim, 1u, " << interleaved_str << ");\n"
+      << "      let cos_v = " << cos_cache.GetByIndices("vec2<u32>(position_id, bsnh[3])") << ";\n"
+      << "      let sin_v = " << sin_cache.GetByIndices("vec2<u32>(position_id, bsnh[3])") << ";\n"
+      << "      let re = " << q_input.GetByOffset("i") << " * cos_v - " << q_input.GetByOffset("j") << " * sin_v;\n"
+      << "      " << q_output.SetByOffset("i", "re") << "\n"
+      << "      let im = " << q_input.GetByOffset("i") << " * sin_v + " << q_input.GetByOffset("j") << " * cos_v;\n"
+      << "      " << q_output.SetByOffset("j", "im") << "\n"
```
qjia7: When the number head index is smaller than num_head_k, also update the k_output. Then the Key branch can be removed.
```
if (bsnh[2] < uniforms.num_head_k) {
   << " " << k_output.SetByOffset("j", "im") << "\n"
}
```

onnxruntime/contrib_ops/webgpu/bert/rotary_embedding.cc
```diff
@@ -50,6 +50,71 @@ Status RotaryEmbeddingProgram::GenerateShaderCode(ShaderHelper& shader) const {
...
+      << "    } else {\n"
```
qjia7: I think the `if - else` is not necessary since `bsnh[3]` must be smaller than `half_rotary_dim` here because that q_global_shape[3] = half_rotary_dim = head_size / 2 in GQA.

onnxruntime/contrib_ops/webgpu/bert/group_query_attention.cc
xiaofeihan1: Good suggesttion!

onnxruntime/contrib_ops/webgpu/bert/rotary_embedding.cc
xiaofeihan1: `Else` is necessary because bsnh[3] = half_rotary_dim + copy_numbers. Else branch is to handle copy

onnxruntime/contrib_ops/webgpu/bert/rotary_embedding.cc
xiaofeihan1: Changed the code.

onnxruntime/contrib_ops/webgpu/bert/rotary_embedding.cc
qjia7: I mean `else` will be never reached since  bsnh[3] < half_rotary_dim is always true. bsnh's shape is as below:
```
onst TensorShape q_global_shape({params.batch_size_, params.sequence_length_,
                                    hidden_size_q / head_size,
                                    static_cast<int64_t>(head_size - half_rotary_embedding_dim)});
```
half_rotary_embedding_dim = head_size / 2 = head_size - half_rotary_embedding_dim.

I think copy_numbers is only used when rotary_embedding_dim != head_size. With this, you can further simplify your shader.

xiaofeihan1: We need to keep this code logic because GQA fused ROE also support Partial Rotary Embedding(i.g. rotary_embedding_dim<head size). For Phi4, the rotary_embedding_dim is 96, the head size is 128.

Although [GQA doc](https://github.com/microsoft/onnxruntime/blob/main/docs/ContribOperators.md#commicrosoftgroupqueryattention) says that cos_cache (optional) : T 2D tensor with shape (max_sequence_length, **head_size / 2)**. but actually it also support rotary_embedding_dim. :)
image

onnxruntime/contrib_ops/webgpu/bert/rotary_embedding.cc
```diff
@@ -50,6 +50,58 @@ Status RotaryEmbeddingProgram::GenerateShaderCode(ShaderHelper& shader) const {
...
+  const auto interleaved_str = interleaved_ ? "true" : "false";
+
+  shader.MainFunctionBody()
+      << "  if (global_idx >= uniforms.q_domain_size) { return; }\n"
+      << "  let half_rotary_dim = uniforms.cos_cache_shape[1];\n"
+      << "  let bsnh = global_idx / uniforms.q_global_stride % uniforms.q_global_shape;\n"
+      << "  let pos_ids_idx = " << position_ids.BroadcastedIndicesToOffset("bsnh.xy", dummy_indices) << ";\n"
+      << "  let position_id = u32(" << position_ids.GetByOffset("pos_ids_idx") << ") + select(0u, bsnh[1], pos_ids_idx == 0u);\n"
+      << "  let cos_v = " << cos_cache.GetByIndices("vec2<u32>(position_id, bsnh[3])") << ";\n"
+      << "  let sin_v = " << sin_cache.GetByIndices("vec2<u32>(position_id, bsnh[3])") << ";\n"
```
qjia7: Move above four lines into if (bsnh[3] < half_rotary_dim)'s body. Otherwise, `bsnh[3]` may be out of bounds.
```
if (bsnh[3] < half_rotary_dim) {
   ...
let cos_v = " << cos_cache.GetByIndices("vec2<u32>(position_id, bsnh[3])")
let sin_v = " << sin_cache.GetByIndices("vec2<u32>(position_id, bsnh[3])")
...
// update q_output
  if (bsnh[2] < uniforms.k_global_shape[2]) {
    // update k_output
  }
} else {
  // update q_output
  if (bsnh[2] < uniforms.k_global_shape[2]) {
    // update k_output
  }
}

Comment (xiaofeihan1) [reply onnxruntime/contrib_ops/webgpu/bert/rotary_embedding.cc]: Good catch! Thanks!

File: onnxruntime/contrib_ops/webgpu/bert/group_query_attention.cc [hunk 5]
```diff
@@ -110,35 +110,92 @@ Status GeneratePositionIDs(onnxruntime::webgpu::ComputeContext& context, const W
...
+      .AddUniformVariables({
+          {params.scale_},
+          {gsl::make_span(q_global_dims)},
+          {gsl::make_span(q_global_strides)},
+          {gsl::make_span(q_input_output_strides)},
+          {gsl::make_span(k_global_dims)},
+          {gsl::make_span(k_global_strides)},
+          {gsl::make_span(k_input_output_strides)},
+          {q_domain_size},
+          {k_domain_size},
```
qjia7: nit: `k_domain_size` not needed anymore?

guschmue:
There are some unrelated UT errors (ie GatherBlockQuantized / cpu). 
Let me check what change came in.

qjia7 (approved):

PR #25760, opened 2025-08-15
Integrate SME1 SGEMM KleidiAI kernels

patryk-kaiser-ARM:
**Key changes**
This PR integrates KleidiAI SME1 FP32 kernels into the existing kleidiai_sgemm.cpp implementation.

Fixing the following test failures:
- DynamicQuantizeMatMul.WithConstantBInputs
- AttentionTest.Attention3DDefault
- AttentionTest.Attention3DWithPastAndPresentQkMatmul

Adding SME2 flag in onnxruntime/core/common/cpuid_info.h & onnxruntime/core/common/cpuid_info.cc
Previous SME2 kernels integrated were using SME(1) check, this change will correctly distinguish between when SME1 and SME2 kernels are to be used.

Bumping KleidiAI version to 1.10.0

**Indicative performance data**
Single thread Mac Mini M4 runs on various models using: onnxruntime_perf_test -v -e cpu -I -m times -x 1 -y 1 -r 1
image

**Next steps**
Additional commits to come will address outstanding to-do issues from previous PR linked below:
[ KleidiAI SGEMM/IGEMM/Quantized MatMul - Modular MLAS API Changes for KleidiAI #25187](https://github.com/microsoft/onnxruntime/pull/25187)

PR #26301, opened 2025-10-14
Implement multithreading in qgemm_kleidi

melkap01-Arm:
**Key changes**

This PR makes changes to improve the performance on Dynamic Qgemms by implementing tiling and threading across operations. 

The changes introduce thread local buffers for reusing memory during inference. And utilizes those in Dynamic Quantised Matmul operations using Kleidiai kernels.

And updating KleidiAI version to 1.15.0

**Example performance**

single thread : 
ort_ops_compare_encoder_1_2025-10-02_17-21-32_vs_encoder_1_2025-10-02_16-54-55

2 threads :
ort_ops_compare_encoder_2_2025-10-02_17-21-47_vs_encoder_2_2025-10-02_16-55-13

### DIFF
--- a/cmake/deps.txt
+++ b/cmake/deps.txt
@@ -56,5 +56,5 @@ extensions;https://github.com/microsoft/onnxruntime-extensions/archive/c24b7bab0
 directx_headers;https://github.com/microsoft/DirectX-Headers/archive/refs/tags/v1.613.1.zip;47653509a3371eabb156360f42faf582f314bf2e
 cudnn_frontend;https://github.com/NVIDIA/cudnn-frontend/archive/refs/tags/v1.12.0.zip;7e733cfdc410d777b76122d64232499205589a96
 dawn;https://github.com/google/dawn/archive/13c1635a14574ebb7116b56a69f5519301417fda.zip;0aadd28fc385cf7d657d5fc70a352372d2d3c76a
-kleidiai;https://github.com/ARM-software/kleidiai/archive/refs/tags/v1.15.0.tar.gz;62ccd24ab60bcef68766440fb42d79071ac2a5d2
+kleidiai;https://github.com/ARM-software/kleidiai/archive/refs/tags/v1.20.0.tar.gz;6895e72b3d5cf1173358164cb3d64c9d7d33cc84
 duktape;https://github.com/svaarala/duktape/releases/download/v2.7.0/duktape-2.7.0.tar.xz;8200c8e417dbab7adcc12c4dbdef7651cfc55794
--- a/onnxruntime/contrib_ops/cpu/quantization/dynamic_quantize_matmul.cc
+++ b/onnxruntime/contrib_ops/cpu/quantization/dynamic_quantize_matmul.cc
@@ -263,8 +263,8 @@ Status DynamicQuantizeMatMul::Compute(OpKernelContext* ctx) const {
       ScaleOutput(*b_scale_tensor, *ctx->Output<Tensor>(0));
     }
   }
-  // Guard against KleidiAI functions being called in non kleidi builds
-  // migrate to a suitable override function call for kelidiai dynamic qgemm function calls
+  // Guard against KleidiAI functions being called in non-Kleidi builds
+  // migrate to a suitable override function call for KleidiAI dynamic QGEMM function calls
 #if defined(USE_KLEIDIAI)
   else {
     MatMulComputeHelper helper;
--- a/onnxruntime/core/mlas/lib/kleidiai/qgemm_kleidiai.cpp
+++ b/onnxruntime/core/mlas/lib/kleidiai/qgemm_kleidiai.cpp
@@ -23,7 +23,7 @@ struct KaiTlsBuffersQgemm {
 };
 static thread_local KaiTlsBuffersQgemm g_kai_tls_qgemm;
 
-//Matmul with float output of dynamic quantized A and symmetric quantized B.
+// Matmul with float output of dynamic-quantized A and symmetric-quantized B.
 
 size_t
 MLASCALL
@@ -36,12 +36,12 @@ ArmKleidiAI::MlasDynamicQgemmPackBSize(
         return 0;
     }
 
-    //Default to sme2_mopa but this may not awalys be the most optimal kernel variant to use
+    // Default to sme2_mopa, but this may not always be the most optimal kernel variant to use.
     auto nr = kai_get_nr_matmul_clamp_f32_qai8dxp1vlx4_qsi8cxp4vlx4_1vlx4vl_sme2_mopa();
     auto kr = kai_get_kr_matmul_clamp_f32_qai8dxp1vlx4_qsi8cxp4vlx4_1vlx4vl_sme2_mopa();
     auto sr = kai_get_sr_matmul_clamp_f32_qai8dxp1vlx4_qsi8cxp4vlx4_1vlx4vl_sme2_mopa();
 
-    //regardless of kernel variant use neon packing variant
+    // Regardless of kernel variant, use the NEON packing variant.
     KLEIDIAI_KERNEL_LOG("kai_run_rhs_pack_kxn_qsi8cxp_qsi8cx_neon Groups=1"
                         << " N="<< N << " K=" << K << " nr=" << nr << " kr=" << kr << " sr=" << sr);
     return kai_get_rhs_packed_size_rhs_pack_kxn_qsi8cxp_qsi8cx_neon(N, K, nr, kr, sr);
@@ -62,7 +62,7 @@ ArmKleidiAI::MlasDynamicQgemmPackB(
         return;
     }
 
-    // Default to sme2_mopa but this may not awalys be the most optimal kernel variant to use
+    // Default to sme2_mopa, but this may not always be the most optimal kernel variant to use.
     auto nr = kai_get_nr_matmul_clamp_f32_qai8dxp1vlx4_qsi8cxp4vlx4_1vlx4vl_sme2_mopa();
     auto kr = kai_get_kr_matmul_clamp_f32_qai8dxp1vlx4_qsi8cxp4vlx4_1vlx4vl_sme2_mopa();
     auto sr = kai_get_sr_matmul_clamp_f32_qai8dxp1vlx4_qsi8cxp4vlx4_1vlx4vl_sme2_mopa();
@@ -75,9 +75,9 @@ ArmKleidiAI::MlasDynamicQgemmPackB(
     // lhs_zp - lhs zero point
     // y = (1/(scale_factor_lhs * scale_factor_rhs) * sum( (lhs_q + lhs_zp)*rhs_q )) + bias
 
-    // rhs packing requires lhs_zp because it will perform lhs_zp*rhs_q during rhs packing
-    // because lhs quantization is hidden from us, by lhs quant packing, we don't have a value for lhs_zp it is
-    // lhs dynamic quantization
+    // RHS packing requires lhs_zp because it will perform lhs_zp*rhs_q during RHS packing.
+    // Because LHS quantization is hidden from us by LHS quant packing, we don't have a value for lhs_zp.
+    // LHS uses dynamic quantization.
 
     kai_rhs_pack_qsi8cx_params params{
         1,  // lhs_zp - set to 1 so it becomes sum((lhs_q + 1)*rhs_q )),
@@ -85,7 +85,7 @@ ArmKleidiAI::MlasDynamicQgemmPackB(
         1.f  // it is not used
     };
 
-    //regardless of kernel variant use neon packing variant
+    // Regardless of kernel variant, use the NEON packing variant.
     kai_run_rhs_pack_kxn_qsi8cxp_qsi8cx_neon(1, N, K, nr, kr, sr, B,
                                              // N bias values
                                              Bias,
@@ -118,23 +118,38 @@ ArmKleidiAI::MlasDynamicQGemmBatch(
         return;
     }
 
-    //We are required to enforce errors when we reach this stage as we will not be able
-    //to reverse the packing decision that was made for RHS.
+    // We are required to fail fast when we reach this stage as we will not be able
+    // to reverse the packing decision that was made for RHS.
 
-    ORT_ENFORCE(DataParams != nullptr, "Dynamic QGEMM requires valid DataParams.");
+    if (DataParams == nullptr) {
+        MLAS_THROW_EX(std::runtime_error, "Dynamic QGEMM requires valid DataParams.");
+    }
 
     for (size_t batch_idx = 0; batch_idx < BatchSize; ++batch_idx) {
         const auto& params = DataParams[batch_idx];
-        ORT_ENFORCE(params.A != nullptr, "Dynamic QGEMM requires non-null A pointer for batch ", batch_idx);
-        ORT_ENFORCE(params.C != nullptr, "Dynamic QGEMM requires non-null C pointer for batch ", batch_idx);
-        ORT_ENFORCE(params.PackedB != nullptr, "Dynamic QGEMM requires non-null PackedB pointer for batch ", batch_idx);
+
+        if (params.A == nullptr) {
+            MLAS_THROW_EX(std::runtime_error, "Dynamic QGEMM requires non-null A pointer.");
+        }
+        if (params.C == nullptr) {
+            MLAS_THROW_EX(std::runtime_error, "Dynamic QGEMM requires non-null C pointer.");
+        }
+        if (params.PackedB == nullptr) {
+            MLAS_THROW_EX(std::runtime_error, "Dynamic QGEMM requires non-null PackedB pointer.");
+        }
+
         const size_t lda = params.lda != 0 ? params.lda : Shape.K;
         const size_t ldc = params.ldc != 0 ? params.ldc : Shape.N;
-        ORT_ENFORCE(lda >= Shape.K, "lda (", lda, ") must be >= Shape.K (", Shape.K, ") for batch ", batch_idx);
-        ORT_ENFORCE(ldc >= Shape.N, "ldc (", ldc, ") must be >= Shape.N (", Shape.N, ") for batch ", batch_idx);
+
+        if (lda < Shape.K) {
+            MLAS_THROW_EX(std::runtime_error, "Dynamic QGEMM requires lda >= K.");
+        }
+        if (ldc < Shape.N) {
+            MLAS_THROW_EX(std::runtime_error, "Dynamic QGEMM requires ldc >= N.");
+        }
     }
 
-    //Dynamic Quantize A - lhs
+    // Dynamic-quantize A (LHS).
     const size_t LhsPackedStride = kai_get_lhs_packed_size_lhs_quant_pack_qai8dxp_f32(Shape.M, Shape.K, mr, kr, sr);
     std::byte* LhsPackedData = nullptr;
 
@@ -145,7 +160,7 @@ ArmKleidiAI::MlasDynamicQGemmBatch(
     g_kai_tls_qgemm.lhs_packed.resize(LhsPackedStride * BatchSize);
     LhsPackedData = g_kai_tls_qgemm.lhs_packed.data();
 
-    //Per-batch table of lhs
+    // Per-batch table of LHS base pointers.
     if (g_kai_tls_qgemm.lhs_base_table.capacity() < BatchSize) {
 
         g_kai_tls_qgemm.lhs_base_table.reserve(BatchSize);
@@ -153,8 +168,8 @@ ArmKleidiAI::MlasDynamicQGemmBatch(
     g_kai_tls_qgemm.lhs_base_table.resize(BatchSize);
     // Capture the shared batch table pointer so worker threads use the same backing storage.
     const std::byte** tls_lhs_base = g_kai_tls_qgemm.lhs_base_table.data();
-    // B batches require no packing
-    // We have already decided the matmul variant we are using, before having values for M,N,K
+    // B batches require no packing.
+    // We have already decided the matmul variant we are using before having values for M, N, and K.
     MlasTrySimpleParallel(ThreadPool, BatchSize, [&](ptrdiff_t batch_idx) {
 
         std::byte* lhs = nullptr;
@@ -169,30 +184,30 @@ ArmKleidiAI::MlasDynamicQGemmBatch(
         tls_lhs_base[batch_idx] = lhs;
     });
 
-    // tile iteration dimensions
+    // Tile iteration dimensions.
     std::array<size_t, 3> dim;
     dim[0] = BatchSize;                  // B
     dim[1] = MlasDivRoundup(Shape.M, m_step);  // M
     dim[2] = MlasDivRoundup(Shape.N, n_step);  // N
 
-    // Minimize the kernel call count for the number of available threads
+    // Minimize the kernel call count for the number of available threads.
     auto RequiredTiles = std::min(static_cast<size_t>(MlasGetMaximumThreadCount(ThreadPool)), dim[0] * dim[1] * dim[2]);
 
-    // scale required tiles over available tile processors
+    // Scale required tiles over available tile processors.
     dim[1] = MlasDivRoundup(RequiredTiles * dim[1], dim[1] * dim[2]);
     dim[2] = MlasDivRoundup(RequiredTiles * dim[2], dim[1] * dim[2]);
 
-    // compute new step sizes
+    // Compute new step sizes.
     m_step *= MlasDivRoundup(MlasDivRoundup(Shape.M, dim[1]), m_step);
     n_step *= MlasDivRoundup(MlasDivRoundup(Shape.N, dim[2]), n_step);
 
-    // update tile iterations
+    // Update tile iterations.
     dim[1] = MlasDivRoundup(Shape.M, m_step);
     dim[2] = MlasDivRoundup(Shape.N, n_step);
 
     MlasTrySimpleParallel(ThreadPool, static_cast<ptrdiff_t>(dim[0] * dim[1] * dim[2]), [=](ptrdiff_t tid) {
 
-        // compute B,M,N index from iteration index
+        // Compute B, M, N indices from the iteration index.
         ptrdiff_t BIdx = tid / (dim[1] * dim[2]);
         ptrdiff_t MIdx = (tid % (dim[1] * dim[2])) / dim[2];
         ptrdiff_t NIdx = (tid % (dim[1] * dim[2])) % dim[2];
--- a/onnxruntime/core/providers/cpu/quantization/matmul_integer_base.h
+++ b/onnxruntime/core/providers/cpu/quantization/matmul_integer_base.h
@@ -131,10 +131,10 @@ class MatMulIntegerBase : public OpKernel {
   /*
     Helper method to pre-pack Matrix B using Arm® KleidiAI™ packing if eligible.
 
-    Returns false if KleidiAI dynamic qantization is not supported or the index of the input tensor is not input B's index.
-    If these checks passes, prepares a dynamic quantization pack content and calls PrepareKleidiaiDynamicPack for futher policies.
+    Returns false if KleidiAI dynamic quantization is not supported or the index of the input tensor is not input B's index.
+    If these checks pass, prepares a dynamic quantization pack context and calls PrepareKleidiaiDynamicPack for further policies.
     If those policies also satisfy, it calls the helper to execute the pre-packing in KleidiAI context.
-    Returns true of pre-packing was performed and false otherwise.
+    Returns true if pre-packing was performed and false otherwise.
   */
   bool TryKleidiaiDynamicPrePack(const Tensor& tensor, int input_idx, AllocatorPtr alloc,
                                  bool& is_packed,
@@ -153,10 +153,10 @@ class MatMulIntegerBase : public OpKernel {
   /*
       Helper method to determine if Arm® KleidiAI™ dynamic quantization pre-packing policies are satisfied.
 
-      Checks for the presence of the constant input tensor B, symmetricity on the zero point and validity of the scales.
+      Checks for the presence of the constant input tensor B, symmetry on the zero point and validity of the scales.
       Also checks if the shape of the tensor B is supported by KleidiAI and if bias tensor is also a constant input.
       Makes B transposition if necessary.
-      Sets can_use_dynamic_quant_mlas_ flag accordingly nd returns true if all policies are satisfied.
+      Sets can_use_dynamic_quant_mlas_ flag accordingly and returns true if all policies are satisfied.
   */
   bool PrepareKleidiaiDynamicPack(const Tensor& tensor,
                                   AllocatorPtr alloc,
@@ -182,9 +182,6 @@ class MatMulIntegerBase : public OpKernel {
     }
 
     ctx.bias = GetConstantInputTensor(GetBiasIdx());
-    if (ctx.bias != nullptr) {
-      dynamic_quant_mlas_bias_data_was_packed_ = true;
-    }
 
     ctx.K = static_cast<size_t>(b_shape_[0]);
     ctx.N = static_cast<size_t>(b_shape_[1]);
@@ -199,6 +196,10 @@ class MatMulIntegerBase : public OpKernel {
     // If K==0 there is nothing to reduce over, and the RHS packer may dereference invalid memory.
     if (ctx.K == 0 || ctx.N == 0) {
       return false;
+    }
+
+    if (ctx.bias != nullptr) {
+      dynamic_quant_mlas_bias_data_was_packed_ = true;
     }
 
     can_use_dynamic_quant_mlas_ = true;
@@ -255,9 +256,9 @@ class MatMulIntegerBase : public OpKernel {
   /*
   Helper for checking the zero points tensor of the input. Arm® KleidiAI™ supports symmetric zero points.
 
-  This helper method checks if zero point tensor , if its present in the inputs with its index, it checks it the data type aither uint8_t or int8_t.
-  It also checks if all the zero point values are zeros. If not sets the can_use_dynamic_quant_mlas_ flag to false.
-  If zero point tensor is not present, it sets the falg true as symmetric zero point is assumed.
+  This helper method checks if the zero point tensor, if it's present in the inputs with its index, it checks the data type either uint8_t or int8_t.
+  It also checks if all the zero point values are zeros. If not, sets the can_use_dynamic_quant_mlas_ flag to false.
+  If zero point tensor is not present, it sets the flag true as symmetric zero point is assumed.
   Returns the flag.
   */
   bool IsZeroPointSymmetric() {
@@ -278,7 +279,7 @@ class MatMulIntegerBase : public OpKernel {
     return !b_zp_input_exists;
   }
   /*
-  Heper method to check the validity of the scales tensor for Arm® KleidiAI™ dynamic qantization.
+  Helper method to check the validity of the scales tensor for Arm® KleidiAI™ dynamic quantization.
   Scales are invalid and can_use_dynamic_quant_mlas_ flag is false returns if the float scales are non-finite or non-positive.
   Otherwise can_use_dynamic_quant_mlas_ flag returned true.
   */
@@ -291,7 +292,7 @@ class MatMulIntegerBase : public OpKernel {
     return !has_invalid;
   }
   /*
-    Helper to promote a 1D tensor to 2D, for Arm® KleidiAI™ dynamic qantization, if necessary. Returns false if the tensor rank is 0.
+    Helper to promote a 1D tensor to 2D, for Arm® KleidiAI™ dynamic quantization, if necessary. Returns false if the tensor rank is 0.
   */
   bool PromoteBShapeIfNeeded() {
     if (b_shape_.NumDimensions() == 0) {
@@ -307,7 +308,7 @@ class MatMulIntegerBase : public OpKernel {
   }
   /*
     Helper method to check the shape policy of the tensor B is passes for Arm® KleidiAI™ dynamic quantization.
-    The shape should be at least 2D and all the dimentions except the last two should be 1. 1D tensor is promoted to 2D.
+    The shape should be at least 2D and all the dimensions except the last two should be 1. 1D tensor is promoted to 2D.
   */
   bool IsBShapeSupportedForDynamicQuant(const TensorShape& tensor_shape) {
     b_shape_ = tensor_shape;
@@ -324,8 +325,8 @@ class MatMulIntegerBase : public OpKernel {
     return true;
   }
   /*
-    Checks against the constant initilized tensor index and returns the constant tensor if present.
-    Returns nullptr if index is invalid or the tensor is not hold by the kernel instance.
+    Checks against the constant initialized tensor index and returns the constant tensor if present.
+    Returns nullptr if index is invalid or the tensor is not held by the kernel instance.
   */
   const Tensor* GetConstantInputTensor(int input_idx) const {
     if (input_idx < 0) {


melkap01-Arm:

cmake/deps.txt
```diff
@@ -56,5 +56,5 @@ extensions;https://github.com/microsoft/onnxruntime-extensions/archive/c24b7bab0
 directx_headers;https://github.com/microsoft/DirectX-Headers/archive/refs/tags/v1.613.1.zip;47653509a3371eabb156360f42faf582f314bf2e
 cudnn_frontend;https://github.com/NVIDIA/cudnn-frontend/archive/refs/tags/v1.12.0.zip;7e733cfdc410d777b76122d64232499205589a96
 dawn;https://github.com/google/dawn/archive/13c1635a14574ebb7116b56a69f5519301417fda.zip;0aadd28fc385cf7d657d5fc70a352372d2d3c76a
-kleidiai;https://github.com/ARM-software/kleidiai/archive/refs/tags/v1.10.0.tar.gz;11b62149cb2514b3b9069cc435c3aa7a4e82b97a
+kleidiai;https://github.com/ARM-software/kleidiai/archive/refs/tags/v1.15.0.tar.gz;62ccd24ab60bcef68766440fb42d79071ac2a5d2
```
hariharans29: With this update in the KAI version from 1.10 to 1.15, can SME/SME2 detection be enabled on Windows too to leverage the kernels ?

https://github.com/microsoft/onnxruntime/pull/25187/files#r2223006773
https://github.com/microsoft/onnxruntime/pull/25760/files#r2325260570

patryk-kaiser-ARM:
Can we get workflows ran please

onnxruntime/core/mlas/lib/kleidiai/qgemm_kleidiai.cpp
```diff
@@ -78,39 +87,147 @@ MLASCALL
...
+
+    //Dynamic Quantize A - lhs
+    const size_t LhsPackedStride = kai_get_lhs_packed_size_lhs_quant_pack_qai8dxp_f32(Shape.M, Shape.K, mr, kr, sr);
+    std::byte* LhsPackedData = nullptr;
+
+    if (g_kai_tls_qgemm.lhs_packed.capacity() < LhsPackedStride * BatchSize) {
+
+        g_kai_tls_qgemm.lhs_packed.reserve(LhsPackedStride * BatchSize);
+    }
+    g_kai_tls_qgemm.lhs_packed.resize(LhsPackedStride * BatchSize);
```
hariharans29, line 160: Can't we just do the resizing directly instead of reserve + resize ?

onnxruntime/core/mlas/lib/kleidiai/qgemm_kleidiai.cpp
```diff
@@ -78,39 +87,147 @@ MLASCALL
...
-    }
+        // Allocate temporary buffer for raw A*B result (TLS reusable buffer)
+        {
+            const size_t tile_elems = TileSizeM * TileSizeN;
+            if (g_kai_tls_qgemm.output_tile.capacity() < tile_elems) {
+                // reserve more memory if required
+                g_kai_tls_qgemm.output_tile.reserve(tile_elems);
+            }
+            // resize the tile to the required size (doesn't effect memory)
+            g_kai_tls_qgemm.output_tile.resize(tile_elems);
```
hariharans29: Ditto - Is Reserve + Resize necessary ?

onnxruntime/core/mlas/lib/kleidiai/qgemm_kleidiai.cpp
```diff
@@ -11,10 +11,19 @@
...
 #include "kai/ukernels/matmul/matmul_clamp_f32_qai8dxp_qsi8cxp/kai_matmul_clamp_f32_qai8dxp1vlx4_qsi8cxp4vlx4_1vlx4vl_sme2_mopa.h"
+#include "kai/ukernels/matmul/matmul_clamp_f32_qai8dxp_qsi8cxp/kai_matmul_clamp_f32_qai8dxp1vlx4_qsi8cxp4vlx4_1vlx4vl_sme_mopa.h"
 #include "kai/ukernels/matmul/matmul_clamp_f32_qai8dxp_qsi8cxp/kai_matmul_clamp_f32_qai8dxp1x4_qsi8cxp4vlx4_1x4vl_sme2_dot.h"
 
 #include "mlasi_kleidiai.h"
 
+// Thread-local reusable buffers to reduce allocation overhead across tiles.
+struct KaiTlsBuffersQgemm {
+    std::vector<float> output_tile;
+    std::vector<float> bias_zero;
```
hariharans29: Is `bias_zero` used somewhere ?

onnxruntime/core/mlas/lib/kleidiai/qgemm_kleidiai.cpp
```diff
@@ -78,39 +87,147 @@ MLASCALL
...
+            const size_t tile_elems = TileSizeM * TileSizeN;
+            if (g_kai_tls_qgemm.output_tile.capacity() < tile_elems) {
+                // reserve more memory if required
+                g_kai_tls_qgemm.output_tile.reserve(tile_elems);
+            }
+            // resize the tile to the required size (doesn't effect memory)
+            g_kai_tls_qgemm.output_tile.resize(tile_elems);
+        }
+        float* temp_tile = g_kai_tls_qgemm.output_tile.data();
+        std::fill_n(temp_tile, TileSizeM * TileSizeN, 0.0f);
```
hariharans29: Is this buffer zeroing absolutely needed (i.e.) Does the micro-kernel accumulate into the existing contents ?

Is there a concept of dis-reagrding existing contents in the output buffer in the micro-kernel's interface ?

onnxruntime/core/mlas/lib/kleidiai/qgemm_kleidiai.cpp
```diff
@@ -78,39 +87,147 @@ MLASCALL
...
+
+    if (g_kai_tls_qgemm.lhs_packed.capacity() < LhsPackedStride * BatchSize) {
+
+        g_kai_tls_qgemm.lhs_packed.reserve(LhsPackedStride * BatchSize);
+    }
+    g_kai_tls_qgemm.lhs_packed.resize(LhsPackedStride * BatchSize);
+    LhsPackedData = g_kai_tls_qgemm.lhs_packed.data();
+
+    //Per-batch table of lhs
+    std::vector<const std::byte*> LhsBase(BatchSize);
```
hariharans29: Just a thought - Can this vector containing the per-batch address be moved into the `KaiTlsBuffersQgemm` struct and be re-sized when it's size is less than the BatchSize ? 

The pro of that approach:
1. We generally expect the BatchSize to be stable across runs and that will mean we can do away with the dynamic memory allocation latency variance that comes with using std::vector

The con of that approach:
1. The size of that caching vector will be bound by the highest batch size that the kernel will encounter. 

Given that the batch sizes are **generally** stable across different runs, I am thinking the pro might outweight the con ? 

What are your thoughts on this ?

onnxruntime/core/mlas/lib/kleidiai/qgemm_kleidiai.cpp
```diff
@@ -78,39 +87,147 @@ MLASCALL
...
+                            : kai_get_n_step_matmul_clamp_f32_qai8dxp1vlx4_qsi8cxp4vlx4_1vlx4vl_sme_mopa();
 
-        //Dynamic Quantize A - lhs
-        auto lhs_size = kai_get_lhs_packed_size_lhs_quant_pack_qai8dxp_f32(Shape.M, Shape.K, mr, kr, sr);
-        std::byte* lhs = nullptr;
-        std::unique_ptr<std::byte[]> fallback;
 
-        if (DataParams->Workspace && DataParams->WorkspaceSize >= lhs_size) {
-            lhs = static_cast<std::byte*>(DataParams->Workspace);
+    if (Shape.M == 0 || Shape.N == 0) {
```
hariharans29: Should there be a Shape.K check for completeness ?

hariharans29:
General sanity check question: Are there enough tests that trigger all the nuances of the multi-threaded implementation - Are there enough tests with multiple batch sizes, M, and N dimensions that exercise all aspects of the multi-threaded implementation ?

onnxruntime/core/mlas/lib/kleidiai/qgemm_kleidiai.cpp
```diff
@@ -78,39 +87,147 @@ MLASCALL
...
+        return;
+    }
+    if ((Shape.M < m_step || Shape.N < n_step) && !DataParams->PackedB) {
+        // Fallback to MLAS
```
edgchen1: there is no fallback implementation of `MlasDynamicQGemmBatch()`.

https://github.com/microsoft/onnxruntime/blob/0f6cffc140211dd5766aff1c052e8bc57e3f954e/onnxruntime/core/mlas/lib/qgemm.cpp#L212-L222

if we get to this point, the computation should happen or (maybe less preferably) it should be a hard error.

onnxruntime/core/mlas/lib/kleidiai/qgemm_kleidiai.cpp
```diff
@@ -78,39 +87,147 @@ MLASCALL
...
+                temp_tile,
+                TileSizeN * sizeof(float),
+                sizeof(float),
+                -std::numeric_limits<float>::max(), std::numeric_limits<float>::max()
+                );
+        }
+
+        // Final output tile pointer
+        float* dst_tile = reinterpret_cast<float*>(CTile);
+        std::memcpy(dst_tile, temp_tile, TileSizeM * TileSizeN * sizeof(float));
```
edgchen1: what's the benefit of writing to a temporary buffer (`temp_tile`) and then copying it to `dst_tile` instead of directly writing to `dst_tile`?

onnxruntime/core/mlas/lib/kleidiai/qgemm_kleidiai.cpp
melkap01-Arm: Yes, reserve() + resize() or using only resize() cases both end up with one allocation + one initialisation. But somehow there is a very very little performance difference in the case allocation and initialisation separated or done at once with resize().  (after: is the case reserve() calls removed and only resize() is used.)
ort_ops_compare_2_thread_before_2025-10-29_13-08-56_vs_2_thread_after_2025-10-29_13-32-05

onnxruntime/core/mlas/lib/kleidiai/qgemm_kleidiai.cpp
melkap01-Arm: same above

onnxruntime/core/mlas/lib/kleidiai/qgemm_kleidiai.cpp
melkap01-Arm: This is an idea worth to try and measure the impact. 
I implemented it and the results with single thread: 

ort_ops_compare_single_thread_before_2025-10-30_10-17-10_vs_single_thread_after_2025-10-30_14-24-15


and 2 threads : 
ort_ops_compare_2_thread_before_2025-10-30_10-17-38_vs_2_thread_after_2025-10-30_14-24-29


After: Lhsbase is moved inside the TLS structure. Before: LhaBase is a local buffer shared with the threads.  


 here is the implementation: 

```
      //Per-batch table of lhs
    if (g_kai_tls_qgemm.LhsBase.capacity() < BatchSize) {
        g_kai_tls_qgemm.LhsBase.reserve(BatchSize);
    }
      g_kai_tls_qgemm.LhsBase.resize(BatchSize);
    // Capture the shared batch table pointer so worker threads use the same backing storage.
    const std::byte** tls_lhs_base = g_kai_tls_qgemm.LhsBase.data();
      // B batches require no packing
        ⋮
          kai_run_lhs_quant_pack_qai8dxp_f32(Shape.M, Shape.K, mr, kr, sr, 0, DataParams[batch_idx].A, DataParams[batch_idx].lda*sizeof(float), lhs);

        tls_lhs_base[batch_idx] = lhs;
      });
        ⋮

        const std::byte* A_base = tls_lhs_base[BIdx]; // LhsPackedData + LhsPackedStride * BIdx; OR DataParams[batch_idx].Workspace;
         auto ATile = reinterpret_cast<const std::byte*>(A_base + lhs_packed_offset);
```

onnxruntime/core/mlas/lib/kleidiai/qgemm_kleidiai.cpp
JonathanC-ARM: We can remove the fill_n, the kernel handles zeroing of the tile

onnxruntime/core/mlas/lib/kleidiai/qgemm_kleidiai.cpp
melkap01-Arm: The idea behind it was making the arithmetics on the temporary tile to be error prone as it was implemented on the sgemms. But I see making the calculations on the destination and writing directly is lowering the complexity. 

instead of having the result in each TLS and copying to the destination tile, destination tile can have the result directly. 
Measuring the impact :
single thread: 

ort_ops_compare_single_thread_no_fill_after_2025-10-30_15-28-05_vs_single_thread_no_temp_buffer_after_2025-10-30_16-39-27

2 threads : 
ort_ops_compare_2_thread_no_fill_after_2025-10-30_15-28-19_vs_2_thread_no_temp_buffer_after_2025-10-30_16-39-43

onnxruntime/core/mlas/lib/kleidiai/qgemm_kleidiai.cpp
melkap01-Arm: addressed in the new commit

onnxruntime/core/mlas/lib/kleidiai/qgemm_kleidiai.cpp
hariharans29: I suspect perf-wise there isn't much difference but it is coming from a performance variance POV. If we performed dynamic memory allocations on every Run(), I suspect we may see some latency variance. I was just wonderinf if this can be avoided as in most cases, usually the Gemm problem shapes stay the same across invocations. Let us dynamically resize only when we encounter a change of shape (batch size). Hope the motivation of the comment is clear now.

hariharans29:
Will trigger CI once you push commits addressing the PR feedback (right now I only see a rebase). Thanks.

onnxruntime/core/mlas/lib/kleidiai/qgemm_kleidiai.cpp
melkap01-Arm: We will investigate the fallback case further and try to provide better implementation. 
Until then, would like to get your opinion on using ORT_ENFORCE 

ORT_ENFORCE(false, "ArmKleidiAI::MlasDynamicQGemmBatch(): unsupported small-shape case (M < m_step or N < n_step)");

onnxruntime/core/mlas/lib/kleidiai/qgemm_kleidiai.cpp
melkap01-Arm: addressed in the newest commit.

onnxruntime/core/mlas/lib/kleidiai/qgemm_kleidiai.cpp
melkap01-Arm: Motivation behind the comment is clear, if we expect generally stable batches, reusing its capacity across calls is making sense. If the performance results also acceptable we are all good with this idea. Please find the implementation in the latest commit.

melkap01-Arm:
> General sanity check question: Are there enough tests that trigger all the nuances of the multi-threaded implementation - Are there enough tests with multiple batch sizes, M, and N dimensions that exercise all aspects of the multi-threaded implementation ?

We checked the existing tests for qgemm. In current implementation tests are supported for thread pool = null. We created a follow up ticket for test coverage.

hariharans29:
> > General sanity check question: Are there enough tests that trigger all the nuances of the multi-threaded implementation - Are there enough tests with multiple batch sizes, M, and N dimensions that exercise all aspects of the multi-threaded implementation ?
> 
> We checked the existing tests for qgemm. In current implementation tests are supported for thread pool = null. We created a follow up ticket for test coverage.

If all the tests are with ThreadPool == null, does that mean the new threadpool based parallel code path(s) are not exercised ?

onnxruntime/core/mlas/lib/kleidiai/qgemm_kleidiai.cpp
hariharans29: Could we instead implement @edgchen1's suggestion in the other PR: https://github.com/microsoft/onnxruntime/pull/26302#discussion_r2476120222 to have a universal check that can be used in all places to check if MLAS supports QGemm for that problem shape, platform, etc. ? 

Also since we have a check on the `M` dimension, this might need some thinking - In the current setup, we turn off MLAS usage for QGemm in PrePack() if we don't detect SME or the weight's shape don't match requirements in `PrePack()`. See [here](https://github.com/microsoft/onnxruntime/blob/fc3bc097c941c812f481cff42594f2da1ae79097/onnxruntime/contrib_ops/cpu/quantization/dynamic_quantize_matmul.cc#L219) and [here](https://github.com/microsoft/onnxruntime/blob/fc3bc097c941c812f481cff42594f2da1ae79097/onnxruntime/contrib_ops/cpu/quantization/dynamic_quantize_matmul.cc#L227). The `M` dimension won't be known in `PrePack()`. 

Just curious - what would happen if the M was < m_step ? Would there be a crash or would the perf be sub-optimal ? If so, we need to add a runtime check in the CPU kernel's Run() function which means we may need to perform pre-packing for both KAI and the "regular" path. See [here](https://github.com/microsoft/onnxruntime/blob/fc3bc097c941c812f481cff42594f2da1ae79097/onnxruntime/contrib_ops/cpu/quantization/dynamic_quantize_matmul.cc#L233).

melkap01-Arm:
> > > General sanity check question: Are there enough tests that trigger all the nuances of the multi-threaded implementation - Are there enough tests with multiple batch sizes, M, and N dimensions that exercise all aspects of the multi-threaded implementation ?
> > 
> > 
> > We checked the existing tests for qgemm. In current implementation tests are supported for thread pool = null. We created a follow up ticket for test coverage.
> 
> If all the tests are with ThreadPool == null, does that mean the new threadpool based parallel code path(s) are not exercised ?

It means it was not exercised on the onnxruntime_mlas_test run, but it is on the onnxruntime_perf_test. However, unit tests for the multithreaded code added now, in the latest commit. Both cases can use multiple threads in the latest situation.

melkap01-Arm:
Latest commit (b6ff3be3c63885f9f6d6e00a4f9e9940ff4d9f0a) includes the changes:

-The Kleidiai-specific policy in DynamicQuantizeMatMul::Prepack() has been moved into a separate function that is now called inside MatMulIntegerBase::Prepack(), placing the logic higher in the inheritance hierarchy.
-The refactoring separates each policy check into its own call, making the logic more modular and easier to understand. This also better aligns with how SGEMM implementations handle prepacking at upper levels of the inheritance tree.
-The policy checks are now grouped in a clear order, improving readability and making it explicit that the Kleidiai packing kernel is invoked only if all checks pass, and that no fallback is needed after matrix B is packed.
-In the core GEMM implementation (MlasDynamicQGemmBatch), additional checks are added:
M and N values are validated to skip no-op cases.
K == 0 is explicitly checked to ensure an error is reported, since this condition indicates a failure discovered after packing was completed.

onnxruntime/contrib_ops/cpu/quantization/dynamic_quantize_matmul.cc
```diff
@@ -164,132 +165,23 @@ class DynamicQuantizeMatMul final : public MatMulIntegerToFloatBase {
...
-
-      const auto* b_data = static_cast<const uint8_t*>(tensor.DataRaw());
-
-      std::optional<Tensor> b_trans_buffer;
-      if (IsBTransposed()) {
-        std::swap(K, N);
-        b_data = quantization::TransPoseInputData(b_data, b_trans_buffer, alloc, N, K);
-      }
+  bool SupportsKleidiaiDynamicQuant() const override {
+    if (!CPUIDInfo::GetCPUIDInfo().HasArm_SME()) {
```
hariharans29: Can the API `MlasIsDynamicQGemmAvailable()` be modified to return true if either SME1/SME2 are now available and can that be called here on line 169 to ensure MLAS supports DynamicQGemm ?

onnxruntime/contrib_ops/cpu/quantization/dynamic_quantize_matmul.cc
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+#include "core/common/cpuid_info.h"  // for CPUIDInfo::GetCPUIDInfo().HasArm_SME()
```
hariharans29: Can get rid of this include if we invoke MlasIsDynamicQGemmAvailable() API

onnxruntime/contrib_ops/cpu/quantization/dynamic_quantize_matmul.cc
```diff
@@ -306,11 +198,7 @@ class DynamicQuantizeMatMul final : public MatMulIntegerToFloatBase {
 
  private:
   // Indicates when MlasDynamicQGemmBatch() can be used
-  bool can_use_dynamic_quant_mlas_{false};
-#if defined(USE_KLEIDIAI) && !defined(_MSC_VER)
-  // Indicates that the biases are a constant input and thus already quantized / packed
-  bool dynamic_quant_mlas_bias_data_was_packed_{false};
-#endif
+  // Flag storage is handled by MatMulIntegerBase.
```
hariharans29: Comment on line 200 is dangling ("Indicates when....") - I guess it is no longer relevant given that the flags have moved....

onnxruntime/core/mlas/lib/kleidiai/qgemm_kleidiai.cpp
```diff
@@ -80,42 +88,148 @@ MLASCALL
...
+    });
+
+    // tile iteration dimensions
+    std::array<size_t, 3> dim;
+    dim[0] = BatchSize;                  // B
+    dim[1] = MlasDivRoundup(Shape.M, m_step);  // M
+    dim[2] = MlasDivRoundup(Shape.N, n_step);  // N
+
+    // Minimize the kernel call count for the number of available threads
+    auto RequiredTiles = std::min(static_cast<size_t>(MlasGetMaximumThreadCount(ThreadPool)), dim[0] * dim[1] * dim[2]);
```
hariharans29, line 194: Is there room for tuning this heuristic ? What if dim[0] * dim[1] * dim[2] is closer to 2 * Thread_count ? In that case, does it make sense to keep the required tiles as is as the tail processing is quite less or does it make sense to process bigger tiles and keep tile count smaller ?

onnxruntime/core/mlas/lib/kleidiai/qgemm_kleidiai.cpp
```diff
@@ -80,42 +88,148 @@ MLASCALL
...
-            Shape.N * sizeof(float),
-            sizeof(float),
-            -std::numeric_limits<float>::max(), std::numeric_limits<float>::max()
+        kai_run_lhs_quant_pack_qai8dxp_f32(Shape.M, Shape.K, mr, kr, sr, 0, DataParams[batch_idx].A, DataParams[batch_idx].lda*sizeof(float), lhs);
+        tls_lhs_base[batch_idx] = lhs;
+    });
+
+    // tile iteration dimensions
+    std::array<size_t, 3> dim;
+    dim[0] = BatchSize;                  // B
```
hariharans29, line 189: Is there room for optimization in any of the logic below if BatchSize == 1 ?

onnxruntime/core/providers/cpu/quantization/matmul_integer_base.h
```diff
@@ -18,8 +25,13 @@ class MatMulIntegerBase : public OpKernel {
                  /*out*/ PrePackedWeights* prepacked_weights) override {
     is_packed = false;
 
-    // only pack Matrix B
+    // only pack Matrix B++
```
hariharans29: Nit: Is ++ a typo ?

onnxruntime/core/providers/cpu/quantization/matmul_integer_base.h
```diff
@@ -89,6 +101,192 @@ class MatMulIntegerBase : public OpKernel {
...
+    if (!can_use_dynamic_quant_mlas_) {
+      return false;
+    }
+
+    is_packed = false;
+
+    const size_t packed_b_size = MlasDynamicQgemmPackBSize(ctx.N, ctx.K);
+    if (packed_b_size == 0) {
+      can_use_dynamic_quant_mlas_ = false;
+      return true;
```
hariharans29: Should this return false if `can_use_dynamic_quant_mlas_` = false ?

onnxruntime/core/providers/cpu/quantization/matmul_integer_base.h
```diff
@@ -89,6 +101,192 @@ class MatMulIntegerBase : public OpKernel {
...
+    if (has_invalid) {
+      can_use_dynamic_quant_mlas_ = false;
+    }
+    return can_use_dynamic_quant_mlas_;
+  }
+
+  bool IsBShapeSupportedForDynamicQuant(const TensorShape& tensor_shape) {
+    b_shape_ = tensor_shape;
+    if (b_shape_.NumDimensions() < 2) {
+      return false;
```
hariharans29, line 316: Low priority question: Can 1-D shapes be promoted to 2-D shapes by pre-pending or appending 1 ?

onnxruntime/core/providers/cpu/quantization/matmul_integer_base.h
```diff
@@ -89,6 +101,192 @@ class MatMulIntegerBase : public OpKernel {
...
+  struct KleidiaiDynamicPackContext {
+    const Tensor* scale{nullptr};
+    const Tensor* bias{nullptr};
+    const uint8_t* b_data{nullptr};
+    size_t K{0};
+    size_t N{0};
+    std::optional<Tensor> transposed_buffer;
+  };
+
+  bool TryKleidiaiDynamicPrePack(const Tensor& tensor, int input_idx, AllocatorPtr alloc,
```
hariharans29, line 139: A brief description of what each of the following helper methods do and look for and when it returns true/false will help the reader.

onnxruntime/test/mlas/unittest/test_dynamic_qgemm.cpp
```diff
@@ -96,45 +142,73 @@ class MlasDynamicQgemmTest {
...
+
+    validate(run_tag);
+  }
+};
+
+class MlasDynamicQgemmSingleThreadTest : public MlasDynamicQgemmTestBase {
+ public:
+  void Test(size_t M, size_t N, size_t K, size_t BatchSize) {
+    // Currently, MlasDynamicQGemmBatch() and associated functions require SME or else they are no-ops.
+    if (!MLAS_CPUIDINFO::GetCPUIDInfo().HasArm_SME())
```
hariharans29: Could we use the MLAS APi to check if dynamic Q Gemm functionality is available ?

onnxruntime/test/mlas/unittest/test_dynamic_qgemm.cpp
```diff
@@ -96,45 +142,73 @@ class MlasDynamicQgemmTest {
...
+  static const char* GetTestSuiteName() { return "DynamicQgemmSingleThread"; }
+};
 
-  static const char* GetTestSuiteName() {
-    return "DynamicQgemm";
+class MlasDynamicQgemmThreadPoolTest : public MlasDynamicQgemmTestBase {
+ public:
+  void Test(size_t M, size_t N, size_t K, size_t BatchSize) {
+    // Currently, MlasDynamicQGemmBatch() and associated functions require SME or else they are no-ops.
+    if (!MLAS_CPUIDINFO::GetCPUIDInfo().HasArm_SME())
```
hariharans29: Same as above

onnxruntime/core/mlas/lib/kleidiai/qgemm_kleidiai.cpp
melkap01-Arm: In this line minimum of the dim[0] * dim[1] * dim[2] vs Thread_count is taken account in order to keep as is or enlarge the tile size accordingly. If we see the later lines code updates the m_step/n_step by the scale calculated in the middle. A rebalancing work going on here in order to minimise the kernel call. For example m_step/n_step = 16/64 initially and after the scaling according to the required tiles new m_step/n_step = 16/256 for a C matrix 1x512 when #threads=2 and becomes 16/192 when #threads=3 and 16/128 when #threads=6...
I believe this logic here both minimises the kernel call & leaves no room for tail processing. I am sorry if I didn't  clearly understand the question but I feel like this logic is better for reducing tail processing as it tries to fit the tiles into the C tensor cleanly. Please highlight any point if this does not answer your question.

onnxruntime/contrib_ops/cpu/quantization/dynamic_quantize_matmul.cc
melkap01-Arm: will be addressed in the new commit

onnxruntime/contrib_ops/cpu/quantization/dynamic_quantize_matmul.cc
melkap01-Arm: will be addressed in the new commit

onnxruntime/contrib_ops/cpu/quantization/dynamic_quantize_matmul.cc
melkap01-Arm: will be addressed in the new commit

onnxruntime/core/providers/cpu/quantization/matmul_integer_base.h
melkap01-Arm: yes it is , will be addressed in the new commit

onnxruntime/core/providers/cpu/quantization/matmul_integer_base.h
melkap01-Arm: will be addressed in the new commit

onnxruntime/core/mlas/lib/kleidiai/qgemm_kleidiai.cpp
melkap01-Arm: I am not sure if I understood this clearly. BatchSize, dim[0], is contributes the multiplication on RequiredTiles. In later lines all the calculation goes over other dimensions not the BatchSize.  I would argue the cost of the multiply of the other two dim by 1 is negligible and changing the code to treat this differently ,e.g checking against the BatchSize ==1,  would complicate the code for no substantive gain.

hariharans29: Makes sense, thanks

onnxruntime/core/mlas/lib/kleidiai/qgemm_kleidiai.cpp
hariharans29: Fair enough, thanks

onnxruntime/core/providers/cpu/quantization/matmul_integer_base.h
melkap01-Arm: provided

onnxruntime/contrib_ops/cpu/quantization/dynamic_quantize_matmul.cc
melkap01-Arm: done

onnxruntime/contrib_ops/cpu/quantization/dynamic_quantize_matmul.cc
melkap01-Arm: done

onnxruntime/contrib_ops/cpu/quantization/dynamic_quantize_matmul.cc
melkap01-Arm: done

onnxruntime/core/providers/cpu/quantization/matmul_integer_base.h
melkap01-Arm: corrected

onnxruntime/test/mlas/unittest/test_dynamic_qgemm.cpp
melkap01-Arm: done

onnxruntime/test/mlas/unittest/test_dynamic_qgemm.cpp
melkap01-Arm: done

onnxruntime/core/providers/cpu/quantization/matmul_integer_base.h
melkap01-Arm: done

onnxruntime/core/providers/cpu/quantization/matmul_integer_base.h
melkap01-Arm: It is implemented in the latest commit.

onnxruntime/core/mlas/lib/kleidiai/qgemm_kleidiai.cpp
```diff
@@ -80,42 +88,148 @@ MLASCALL
...
-        std::unique_ptr<std::byte[]> fallback;
+    if (BatchSize == 0 || Shape.M == 0 || Shape.N == 0 ) {
+        return;
+    }
+
+    //We are required to enforce errors when we reach this stage as we will not be able
+    //to reverse the packing decision that was made for RHS.
+
+    ORT_ENFORCE(DataParams != nullptr, "Dynamic QGEMM requires valid DataParams.");
+    ORT_ENFORCE(Shape.K > 0, "Dynamic QGEMM requires Shape.K to be non-zero.");
```
hariharans29: Nit: technically is K is 0, we can bail out like on line 107 (when M or N are 0)

onnxruntime/test/mlas/unittest/test_dynamic_qgemm.cpp
```diff
@@ -96,45 +142,73 @@ class MlasDynamicQgemmTest {
...
-      float diff = std::abs(C[i] - CRef[i]);
-      ASSERT_LE(diff, allowed);
-    }
+    auto validate = [&](const char* tag) {
+      SCOPED_TRACE(tag);
+      for (size_t i = 0; i < M * N * BatchSize; ++i) {
+        float abs_c_ref = std::abs(CRef[i]);
+        float dynamic_rel_tol = (K <= 4) ? 0.05f : 0.03f;
+        float rel_tol = dynamic_rel_tol * std::max(abs_c_ref, 1.0f);
+        float abs_tol = 3.0f;
```
hariharans29: Is an absolute tolerance of 3.0f acceptable ?

onnxruntime/test/contrib_ops/matmul_integer_to_float_test.cc
```diff
@@ -489,5 +490,163 @@ TEST(MatMulIntegerToFloat, MatMulInteger_With_ZeroPoint) {
   test_case({15, 14, 13}, {15, 13, 27}, {15, 1, 27});
 }
 
+#if defined(USE_KLEIDIAI) && !defined(_MSC_VER)
```
hariharans29: General comment: I guess we can remove the MSC_VER guards everywhere in this PR now that https://github.com/microsoft/onnxruntime/pull/26995 is in place ?

onnxruntime/test/mlas/unittest/test_dynamic_qgemm.cpp
```diff
@@ -96,45 +142,73 @@ class MlasDynamicQgemmTest {
...
-      float abs_tol = 3.0f;
-      float allowed = std::max(rel_tol, abs_tol);
-      float diff = std::abs(C[i] - CRef[i]);
-      ASSERT_LE(diff, allowed);
-    }
+    auto validate = [&](const char* tag) {
+      SCOPED_TRACE(tag);
+      for (size_t i = 0; i < M * N * BatchSize; ++i) {
+        float abs_c_ref = std::abs(CRef[i]);
+        float dynamic_rel_tol = (K <= 4) ? 0.05f : 0.03f;
```
hariharans29: Some commentary around why we need multiple relative tolerances based on K would help.

onnxruntime/core/providers/cpu/quantization/matmul_integer_base.h
```diff
@@ -89,6 +101,248 @@ class MatMulIntegerBase : public OpKernel {
...
+    if (ctx.scale == nullptr) {
+      return false;
+    }
+
+    if (!IsZeroPointSymmetric()) {
+      return false;
+    }
+
+    if (!AreScalesValid(*ctx.scale)) {
+      can_use_dynamic_quant_mlas_ = false;
```
hariharans29: Isn't can_use_dynamic_quant_mlas_ false to begin with ?

onnxruntime/core/providers/cpu/quantization/matmul_integer_base.h
```diff
@@ -89,6 +101,248 @@ class MatMulIntegerBase : public OpKernel {
...
+      return false;
+    }
+
+    if (!IsBShapeSupportedForDynamicQuant(tensor.Shape())) {
+      can_use_dynamic_quant_mlas_ = false;
```
hariharans29: Ditto

onnxruntime/core/providers/cpu/quantization/matmul_integer_base.h
hariharans29: Oh I know why - This is because IsZeroPointSymmetric() could potentially have set can_use_dynamic_quant_mlas_ to true. 

In my opinion, it would be cleaner for all the validating routines to just return true/false and the can_use_dynamic_quant_mlas_ get set after all the validation has been done - I feel it is much cleaner and less error prone that way...

onnxruntime/test/contrib_ops/matmul_integer_to_float_test.cc
```diff
@@ -489,5 +490,163 @@ TEST(MatMulIntegerToFloat, MatMulInteger_With_ZeroPoint) {
...
+  auto expected = data.Reference(bias[0], bias[1], bias[2]);
+
+  OpTester test("DynamicQuantizeMatMul", 1, kMSDomain);
+  test.AddInput<float>("A", {data.M, data.K}, data.a);
+  test.AddInput<int8_t>("B", {data.K, data.N}, data.b, true /*initializer*/);
+  test.AddInput<float>("b_scale", {data.N}, data.b_scale, true);
+  test.AddInput<int8_t>("b_zero_point", {data.N}, data.b_zp, true /*initializer*/);
+  test.AddInput<float>("bias", {data.N}, bias, true /*initializer*/);
+  test.AddOutput<float>("Y", {data.M, data.N}, expected);
+  test.SetOutputAbsErr("Y", 0.2f);
```
hariharans29: Is such a high output absolute error tolerance required ?

onnxruntime/test/contrib_ops/matmul_integer_to_float_test.cc
```diff
@@ -489,5 +490,163 @@ TEST(MatMulIntegerToFloat, MatMulInteger_With_ZeroPoint) {
...
+    out.reserve(leading * M * N);
+    for (int64_t i = 0; i < leading; ++i) {
+      out.insert(out.end(), base.begin(), base.end());
+    }
+    return out;
+  }
+};
+
+// 1. Bias provided as initializer -> Kleidi packs bias and skips runtime add.
+TEST(DynamicQuantizeMatMul, KleidiBiasInitializer) {
```
hariharans29: Isn't there a separate test file for DynamicQuantizematmul - https://github.com/microsoft/onnxruntime/blob/4b50b2f12a94ace49afe2acf1dbb075606645b45/onnxruntime/test/contrib_ops/dynamic_quantize_matmul_test.cc#L1 ? Any reason why these tests need to live here ?

onnxruntime/test/mlas/unittest/test_dynamic_qgemm.cpp
JonathanC-ARM: Hi Hari, I originally added the K-dependent relative/absolute tolerances months ago because we were seeing larger numerical deltas and wanted to avoid flaky test failures. The kernel has been updated since due to some issues being resolved with the assembly. 

I’ve re-evaluated this now and changed the test to use a single constant absolute tolerance of 0.001f (removing the K-based relative tolerance).  
The worst-case difference I’m seeing is ~0.00025 (give or take depending on the shape/run), so 0.001f gives a small safety margin while still being quite tight. These deltas are consistent with floating-point round-off / accumulation-order differences rather than functional errors.

onnxruntime/test/mlas/unittest/test_dynamic_qgemm.cpp
JonathanC-ARM: This will be changed as per the other reply

onnxruntime/core/mlas/lib/kleidiai/qgemm_kleidiai.cpp
JonathanC-ARM: Yeah thats entirely fair, we shouldnt need ort enforce if we return on K==0
, I wound up making a few changes in other places to handle K and N of 0 particularly around packing.

onnxruntime/core/providers/cpu/quantization/matmul_integer_base.h
JonathanC-ARM: Yeah thats fair, I am going to change IsZeroPOintSymmetric to not mutate that value at all, and just set can_use_dynamic_quant_mlas_ to begin with and only update to true if all checks pass

onnxruntime/test/contrib_ops/matmul_integer_to_float_test.cc
JonathanC-ARM: Yeah I should be able to do that

onnxruntime/test/contrib_ops/matmul_integer_to_float_test.cc
JonathanC-ARM: I didn’t originally add these tests, but from reviewing them they are operator-level integration tests for DynamicQuantizeMatMul. They validate the KleidiAI prepack gating and runtime branching (bias packed vs deferred runtime add, and fallback when zero-points/scales/shapes are unsupported).

Happy to move these if you think this isn’t the right location?
Also open to temporary removal of these tests until we can clarify the correct location.

onnxruntime/test/contrib_ops/matmul_integer_to_float_test.cc
JonathanC-ARM: I suspect that this threshold was set based on the other tests perhaps, when I'm testing it here I can see a simiarly low threshold is achievable, so i will lower the tolerance accordingly

hariharans29: Can we please see if we can match the error tolerance for this op before/without the KleidiAI integration ? If we are going to have to raise the tolerance due to some implementation quirk of KAI, we may have to document this as users may not like that (re-inforcing the need for a way for users to opt-out of KleidiAI at runtime if that's the case)

onnxruntime/test/contrib_ops/matmul_integer_to_float_test.cc
hariharans29: Thanks Jonathan. I am not an expert in this area either - but I would say that given that there is a file where op tests correspodnign to these ops live, maybe we could try moving it there for consistency (if it is not too hard) ?

onnxruntime/test/contrib_ops/matmul_integer_to_float_test.cc
JonathanC-ARM: I'll check the tolerance of this when its ran without kleidi guards. I was able to see it passing with tolerances of 0.03f but yeah would be good to be sure that is within the expected tolerance of the non kleidi op

onnxruntime/test/contrib_ops/matmul_integer_to_float_test.cc
JonathanC-ARM: Sure that makes sense to me Hari, I'll look at moving these to a more appropriate location

onnxruntime/test/contrib_ops/matmul_integer_to_float_test.cc
JonathanC-ARM: I checked the tolerances for clarity sake, it seems that without kleidi the difference is worse in some scenarios i.e. kleidi is closer to the reference values, and in 3 of the 5 test cases the are identical. So its at worst as good as mlas but at best its producing a more accurate result.

| Test case | Max abs diff (non‑Kleidi) | Max abs diff (Kleidi) | Which is closer to expected? | Notes |
|---|---:|---:|---|---|
| `DynamicQuantizeMatMul.KleidiBiasInitializer` | `0.0254901648` | `0.0098038912` | **Kleidi** | Kleidi path exercised; Kleidi ~2.6× smaller max error |
| `DynamicQuantizeMatMul.KleidiBiasRuntime` | `0.0254902840` | `0.0098038912` | **Kleidi** | Kleidi path exercised; Kleidi ~2.6× smaller max error |
| `DynamicQuantizeMatMul.KleidiRejectsNonZeroZeroPoint` | `0.0215682983` | `0.0215682983` | tie | Outputs identical → fallback path (Kleidi gated off) |
| `DynamicQuantizeMatMul.KleidiRejectsInvalidScale` | `0.0254901648` | `0.0254901648` | tie | Outputs identical → fallback path |
| `DynamicQuantizeMatMul.KleidiRejectsUnsupportedBShape` | `0.0254901648` | `0.0254901648` | tie | Outputs identical → fallback path |

hariharans29: Thanks Jonathan. So, to summarize the diff for both the KleidiAI path and the non-KAI path are higher than the default tolerance of the OpTester and that in general KAI does better or on par with the non-KAI path ? Is that why we still need to set the tolerance in the OpTester to 0.03f ? Just curious - do other tests for the same op do the same ?

onnxruntime/core/mlas/lib/kleidiai/qgemm_kleidiai.cpp
```diff
@@ -80,42 +98,148 @@ MLASCALL
...
-        std::unique_ptr<std::byte[]> fallback;
+    if (BatchSize == 0 || Shape.M == 0 || Shape.N == 0 || Shape.K == 0) {
+        return;
+    }
+
+    //We are required to enforce errors when we reach this stage as we will not be able
+    //to reverse the packing decision that was made for RHS.
+
+    ORT_ENFORCE(DataParams != nullptr, "Dynamic QGEMM requires valid DataParams.");
+   // ORT_ENFORCE(Shape.K > 0, "Dynamic QGEMM requires Shape.K to be non-zero.");
```
hariharans29: Nit: Can we remove the ORT_ENFORCE comment ?

JonathanC-ARM: Yep will do, left this by mistake

JonathanC-ARM: also will fix the build problem in next push

onnxruntime/test/contrib_ops/matmul_integer_to_float_test.cc
JonathanC-ARM: Yeah so thats it exactly the kleidi and non kleidi path for this op are at worst equivalent but kleidi is closer to a reference value typically in some cases for this op type. 

No other testcase within this particular file actually seem to have a tolerance set whatsoever, so I can't comment really as to whether its to be expect, I just set this slightly above the worst case which should help prevent failures right on the edge of the threshold, I could have went further for example 0.026 or something to that effect. 

The tolerance set throughout other files for various op's seems to vary wildly some being as low as 0.001 and others approaching up to 0.5

onnxruntime/core/providers/cpu/quantization/matmul_integer_base.h
```diff
@@ -89,6 +101,247 @@ class MatMulIntegerBase : public OpKernel {
...
+
+    if (IsBTransposed()) {
+      std::swap(ctx.K, ctx.N);
+      ctx.b_data = quantization::TransPoseInputData(ctx.b_data, ctx.transposed_buffer, alloc, ctx.N, ctx.K);
+    }
+
+    // KleidiAI dynamic-qgemm packing is not expected to handle degenerate shapes.
+    // If K==0 there is nothing to reduce over, and the RHS packer may dereference invalid memory.
+    if (ctx.K == 0 || ctx.N == 0) {
+      return false;
```
hariharans29, line 198: It would be nice if all the validations that may return false be done before we set either `can_use_dynamic_quant_mlas_` or `dynamic_quant_mlas_bias_data_was_packed_` to true. In that vein, can this be moved above the point where `dynamic_quant_mlas_bias_data_was_packed_` may be set to true ?

hariharans29:
Please rebase with main and the CUDA / TensorRT issues should go away

hariharans29:
May have some conflicts with https://github.com/microsoft/onnxruntime/pull/26849

hariharans29 (approved):

PR #26302, opened 2025-10-14
Implement FP32 kleidiai Gemv

JonathanC-ARM:
### Description
Implementation of special sgemm path which uses GEMV kernels in cases where M or N are 1

Additionally this pr introduces the usage of a microkernel interface which utilizes typedef's provided by KleidiAI such that we can simplify the code and remove things such as ternary operations for SME1 vs SME2 kernels

### Indicative Performance
In Lieu of any production models where gemv was a large contributor of the network. I opted to create a mini model to test which contains thousands of randomized matmul variants. With a distribution of GEMV cases throughout 
image (6)

Using onnxruntime perf test I was able to half the total inference time vs mlas with this model
ort_ops_compare_gemv_no_2025-10-07_19-40-30_vs_gemv_2025-10-07_19-40-58

**_More Benchmarks to come shortly_**

### DIFF
--- a/onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
+++ b/onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
@@ -27,8 +27,8 @@ struct KaiTlsBuffers {
 };
 static thread_local KaiTlsBuffers g_kai_tls;
 
-const kai_matmul_clamp_f32_f32p_f32p_ukernel sgemm_gemm = GetKleidiAISGemmUKernel();
-const kai_matmul_clamp_f32_f32_f32p_ukernel sgemm_gemv = GetKleidiAISGemvUKernel();
+const kai_matmul_clamp_f32_f32p_f32p_ukernel& sgemm_gemm = GetKleidiAISGemmUKernel();
+const kai_matmul_clamp_f32_f32_f32p_ukernel& sgemm_gemv = GetKleidiAISGemvUKernel();
 
 
 // Helpers for GEMV
@@ -141,33 +141,44 @@ ArmKleidiAI::MlasGemvBatch(
     const MLAS_SGEMM_DATA_PARAMS* Data,
     size_t BatchSize
 ) {
-        // Edge case where M and N are both one so we can take the simpler M == 1 Path
-        bool m_n_both_one = (M == 1 && N == 1);
-        if (N == 1 && Data->BIsPacked && !m_n_both_one)
-        {
-            // Exit early because we cannot support cases where N is 1 and B is already packed
+        // Only two paths: M-path (M == 1, also covers M == N == 1) or N-path (N == 1).
+        if (M != 1 && N != 1) {
+            return false;
+        }
+    
+        const bool m_path = (M == 1);
+    
+        // We cannot support cases where N == 1 and B is already packed.
+        // When both are 1, we route through the M-path, so this naturally doesn't trigger.
+        if (!m_path && Data->BIsPacked) {
             return false;
         }
 
+        // Decide RHS and transposition once based on the path
+        CBLAS_TRANSPOSE tb = m_path ? TransB : CblasTrans;
+        size_t rhs_shape = m_path ? N : M;
+
         for (size_t b = 0; b < BatchSize; ++b) {
-            // Depending on the value of M or N we might transpose when packing
-            const bool using_m_path = (M == 1);
-            const bool using_n_path = (N == 1 && !m_n_both_one);
-            CBLAS_TRANSPOSE tb = TransB;
-            size_t rhs_shape = N;
-            size_t rhs_ld = Data[b].ldb;
-            size_t lhs_ld = using_m_path ? Data[b].lda : Data[b].ldb;
-            const float* rhs_base = using_m_path ? reinterpret_cast<const float*>(Data[b].B) : reinterpret_cast<const float*>(Data[b].A);
-            const float* lhs_base = using_m_path ? static_cast<const float*>(Data[b].A) : static_cast<const float*>(Data[b].B);
-            if (using_n_path)
-            {
-                tb = CblasTrans;
-                rhs_shape = M;
-                rhs_ld = Data[b].lda;
-                lhs_ld = Data[b].ldb;
-            }
+
+            size_t rhs_ld    = m_path ? Data[b].ldb : Data[b].lda;
+            // LHS is the vector row we feed to the GEMV microkernel
+            // - M-path: LHS is A, stride = lda
+            // - N-path: LHS is B, stride = ldb
+            size_t lhs_ld = m_path ? Data[b].lda : Data[b].ldb;
+            
+            const float* rhs_base = m_path ? static_cast<const float*>(Data[b].B)
+                                           : static_cast<const float*>(Data[b].A);
+            const float* lhs_base = m_path ? static_cast<const float*>(Data[b].A) 
+                                           : static_cast<const float*>(Data[b].B);
+
             // Prepare packed RHS if needed
             const void* rhs_packed_ptr = nullptr;
+            
+            // The if branch can only be taken in cases where we are dealing with M == 1
+            // We previously reject any prepacked B where N == 1
+            // In cases where N == 1 we Pack A Matrix as the RHS using tb = CBlasTrans
+            // After which the rhs_packed_ptr points to Packed A not B
+            // rhs_packed_ptr = Data[b].B only when M == 1
             if (Data[b].BIsPacked) {
                 rhs_packed_ptr = Data[b].B;
             } else {
@@ -186,9 +197,7 @@ ArmKleidiAI::MlasGemvBatch(
             }
             // Ensure LHS is a contiguous K-length row for the GEMV microkernel.
             // Compute once whether we need to gather based on which side is LHS.
-            const bool needs_gather =
-                (using_m_path ? (TransA == CblasTrans)
-                              : (using_n_path ? (TransB == CblasNoTrans) : false));
+            const bool needs_gather = m_path ? (TransA == CblasTrans) : (TransB == CblasNoTrans);
             if (needs_gather) {
                 g_kai_tls.gemv_lhs_row_tmp.resize(K);
                 for (size_t k = 0; k < K; ++k) {
@@ -215,8 +224,8 @@ ArmKleidiAI::MlasGemvBatch(
                 std::numeric_limits<float>::max()
             );
             // Apply alpha/beta to destination C row
-            bool allowMemCopy = (M == 1) ? (Data[b].ldc == N) : (Data[b].ldc == 1);
-            size_t destStride = (M == 1) ? 1 : Data[b].ldc;
+            bool allowMemCopy = m_path ? (Data[b].ldc == N) : (Data[b].ldc == 1);
+            size_t destStride = m_path ? 1 : Data[b].ldc;
             ApplyAlphaBetaStrided(g_kai_tls.output_tile.data(), rhs_shape, Data[b].alpha, Data[b].beta, Data[b].C, destStride, allowMemCopy);
         }
         return true;


onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
```diff
@@ -21,9 +23,200 @@ struct KaiTlsBuffers {
     std::vector<float> bias_zero;
     std::vector<std::byte> rhs_packed;
     std::vector<std::byte> lhs_packed;
+    std::vector<float> gemv_lhs_row_tmp;
 };
 static thread_local KaiTlsBuffers g_kai_tls;
 
+kai_matmul_clamp_f32_f32p_f32p_ukernel sgemm_gemm = GetKleidiAISGemmUKernel();
+kai_matmul_clamp_f32_f32_f32p_ukernel sgemm_gemv = GetKleidiAISGemvUKernel();
```
edgchen1: `GetKleidiAIXUKernel()` returns `const&`. do we need to make a copy here?

onnxruntime/core/mlas/lib/qgemm.cpp
```diff
@@ -211,8 +211,8 @@ MlasDynamicQGemmBatch (
 ) {
 #if defined(USE_KLEIDIAI) && !defined(_MSC_VER)
     //No fallback and putting in guards
-    if(MLAS_CPUIDINFO::GetCPUIDInfo().HasArm_SME()){
-    ArmKleidiAI::MlasDynamicQGemmBatch(Shape, DataParams, BatchN, ThreadPool);
+    if(ArmKleidiAI::SMEInfo::CanUseSME2){
```
edgchen1: there are other places that need to be updated, like:

https://github.com/microsoft/onnxruntime/blob/b3ba580346293ed1cab35b59cf93ae1da56e65f7/onnxruntime/contrib_ops/cpu/quantization/dynamic_quantize_matmul.cc#L218

https://github.com/microsoft/onnxruntime/blob/b3ba580346293ed1cab35b59cf93ae1da56e65f7/onnxruntime/test/mlas/unittest/test_dynamic_qgemm.cpp#L24

I might be missing some.

I think it would be worth making a helper function like `MlasIsDynamicQGemmAvailable` that has the appropriate checks and using that instead.

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
patryk-kaiser-ARM:
```suggestion
const kai_matmul_clamp_f32_f32p_f32p_ukernel& sgemm_gemm = GetKleidiAISGemmUKernel();
const kai_matmul_clamp_f32_f32_f32p_ukernel& sgemm_gemv = GetKleidiAISGemvUKernel();
```

JonathanC-ARM: updated to const in the latest push

onnxruntime/core/mlas/lib/qgemm.cpp
JonathanC-ARM: Added in the updated checks in various places like these in the latest push

edgchen1: > I think it would be worth making a helper function like `MlasIsDynamicQGemmAvailable` that has the appropriate checks and using that instead.

to clarify, this was the main suggestion.

onnxruntime/test/mlas/unittest/test_dynamic_qgemm.cpp
```diff
@@ -21,7 +22,7 @@ class MlasDynamicQgemmTest {
  public:
   void Test(size_t M, size_t N, size_t K, size_t BatchSize) {
     // Currently, MlasDynamicQGemmBatch() and associated functions require SME or else they are no-ops.
-    if (!MLAS_CPUIDINFO::GetCPUIDInfo().HasArm_SME()) {
+    if (!ArmKleidiAI::SMEInfo::CanUseSME2) {
```
hariharans29: Nit: I guess the Gtest skip comment needs corresponding update too.

onnxruntime/core/mlas/lib/qgemm.cpp
hariharans29: I guess after merging https://github.com/microsoft/onnxruntime/pull/26301, the checks looking for SME2 will go away (i.e.) it can be run on both SME1 and SME2 then ?

JonathanC-ARM: Yes thats correct

JonathanC-ARM: So one change I've made in the latest push is to remove this structure from our kleidi code specifically and put it into mlasi.h removing the armkleidiai namespacing around it, seemed like a sensible place to put it given that other similar code exists in terms of cpu features

hariharans29:
Could you please rebase this @JonathanC-ARM  ?

JonathanC-ARM:
Hi @hariharans29 I've updated the branch now, thanks!

hariharans29:
Please rebase to include this : https://github.com/microsoft/onnxruntime/pull/26559

JonathanC-ARM:
Hi @hariharans29  I've gone ahead and synced my fork now so this branch should include that change now , thanks for letting me know!

onnxruntime/core/mlas/lib/mlasi.h
```diff
@@ -158,6 +158,11 @@ MLAS_FORCEINLINE void
 // of the ONNX Runtime source tree. OpenMP may or may not be enabled in this
 // configuration.
 //
+struct SMEInfo {
```
hariharans29: Can we guard this by #if defined(MLAS_TARGET_ARM64) as this is ARM specific?

onnxruntime/core/mlas/lib/mlasi.h
```diff
@@ -210,9 +229,19 @@ class MLASCPUIDInfo
...
     bool has_arm_sve_i8mm_{false};
     bool has_arm_neon_bf16_{false};
+    bool has_arm_sme_{false};
+    bool has_arm_sme2_{false};
 };
 using MLAS_CPUIDINFO = MLASCPUIDInfo;
 
+// Boolean condition to determine if we can use SME2
+// By default we should try for SME2 first before falling back to SME.
+inline const bool SMEInfo::CanUseSME2 = MLAS_CPUIDINFO::GetCPUIDInfo().HasArm_SME2();
```
hariharans29: Can we avoid duplicate initializations by just moving the initializations out of the `BUILD_MLAS_NO_ONNXRUNTIME` guarded sections ?

cmake/deps.txt
```diff
@@ -47,7 +47,7 @@ protoc_mac_universal;https://github.com/protocolbuffers/protobuf/releases/downlo
 psimd;https://github.com/Maratyszcza/psimd/archive/072586a71b55b7f8c584153d223e95687148a900.zip;1f5454b01f06f9656b77e4a5e2e31d7422487013
 pthreadpool;https://github.com/google/pthreadpool/archive/dcc9f28589066af0dbd4555579281230abbf74dd.zip;533a77943203ef15ca608bcd9dbe2c94da7451d2
 pybind11;https://github.com/pybind/pybind11/archive/refs/tags/v2.13.6.zip;f780292da9db273c8ef06ccf5fd4b623624143e9
-pytorch_cpuinfo;https://github.com/pytorch/cpuinfo/archive/877328f188a3c7d1fa855871a278eb48d530c4c0.zip;9152d4bf6b8bde9f19b116de3bd8a745097ed9df
+pytorch_cpuinfo;https://github.com/pytorch/cpuinfo/archive/de0ce7c7251372892e53ce9bc891750d2c9a4fd8.zip;c45b8d3619b9bccbd26dc5f657959aee38b18b7a
```
hariharans29: Can you please remind me - why do we need this cupinfo dependency update for this PR ?

onnxruntime/contrib_ops/cpu/quantization/dynamic_quantize_matmul.cc
```diff
@@ -215,7 +219,7 @@ class DynamicQuantizeMatMul final : public MatMulIntegerToFloatBase {
 
       // Currently, MlasDynamicQGemmBatch() and associated functions require SME2 or else they are no-ops.
       // We check that here too before attempting to use them.
-      if (!CPUIDInfo::GetCPUIDInfo().HasArm_SME2()) {
+      if (!SMEInfo::CanUseSME2) {
```
hariharans29: Can we skip the qgemm.cpp / dynamic_quantize_matmul.cc / test_dynamic_qgemm.cpp  changes in this PR ? https://github.com/microsoft/onnxruntime/pull/26598 is taking care of it with a new MLAS API for this.

onnxruntime/core/mlas/lib/kleidiai/mlasi_kleidiai.h
```diff
@@ -51,9 +51,7 @@
 
 namespace ArmKleidiAI {
 
-// By default we should try for SME2 first before falling back to SME.
-inline const bool UseSME2 = MLAS_CPUIDINFO::GetCPUIDInfo().HasArm_SME2();
-
+//
```
hariharans29: Stray change ?

onnxruntime/core/mlas/lib/kleidiai/mlasi_kleidiai.h
```diff
@@ -77,6 +75,19 @@ MlasGemmPackB(
     void* PackedB
     );
 
+bool
+MLASCALL
+MlasFp32Gemv(
```
hariharans29: Consider renaming to `MlasGemvBatch` to be consistent with `MlasGemmBatch` ? Thoughts on this ?

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
```diff
@@ -226,23 +417,26 @@ Return Value:
...
-    const size_t mr = UseSME2 ? kai_get_mr_matmul_clamp_f32_f32p2vlx1_f32p2vlx1biasf32_sme2_mopa()
-                              : kai_get_mr_matmul_clamp_f32_f32p2vlx1_f32p2vlx1b_2vlx2vl_sme_mopa();
-    const size_t kr = UseSME2 ? kai_get_kr_matmul_clamp_f32_f32p2vlx1_f32p2vlx1biasf32_sme2_mopa()
-                              : kai_get_kr_matmul_clamp_f32_f32p2vlx1_f32p2vlx1b_2vlx2vl_sme_mopa();
-    const size_t sr = UseSME2 ? kai_get_sr_matmul_clamp_f32_f32p2vlx1_f32p2vlx1biasf32_sme2_mopa()
-                              : kai_get_sr_matmul_clamp_f32_f32p2vlx1_f32p2vlx1b_2vlx2vl_sme_mopa();
+    // Attempt GEMV (M==1 or N==1)
+    if (M == 1 || N == 1)
+    {
+        if (ArmKleidiAI::MlasFp32Gemv(TransA, TransB, M, N, K, Data, BatchSize)) {
```
hariharans29: Any scope for using multiple threads in the Gemv implementation (I see that the Gemv routing doesn't take in the ThreadPool param) ?

If there are plans to add the multi-threaded implementation in the future for Gemv, can you please add a TODO for that ?

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
```diff
@@ -226,23 +417,26 @@ Return Value:
...
+            return true;
```
hariharans29, line 439: Just checking - In case the Gemv execution flow returns false for some reason, I am guessing you want to try KleidiAI's Gemm before falling back to MLAS ? That is how it is right now but I wanted to check if that is the intended flow.

onnxruntime/core/mlas/lib/mlasi.h
JonathanC-ARM: Sure makes sense, will make the change

onnxruntime/core/mlas/lib/mlasi.h
JonathanC-ARM: Not 100% sure but I can check and I'd prefer that to be honest

cmake/deps.txt
JonathanC-ARM: I don't think I made this change, it might be somehow an older version of the deps file. Will look into it.

onnxruntime/core/mlas/lib/kleidiai/mlasi_kleidiai.h
JonathanC-ARM: Yes that seems to be the case, removed the above code but looks like I left the initial //

onnxruntime/contrib_ops/cpu/quantization/dynamic_quantize_matmul.cc
JonathanC-ARM: Sure I can revert this change altogether in favor of #26598

onnxruntime/core/mlas/lib/kleidiai/mlasi_kleidiai.h
JonathanC-ARM: Seems like a sensible suggestion, doesn't really impact anything other than making things consistent. So I'd be happy to make the change

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
JonathanC-ARM: I will likely add a todo it's not something that we tested to be honest, so would need to investigate whether threaded implementation provided any benefit to performance given that the kernels in question can handle the entire matrix in a single execution as is. But maybe there are cases where splitting the workload would lead to be benefits. But without testing I'm just speculating.

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
JonathanC-ARM: That's the intention yes, attempt op and fallback to mlas if we cannot proceed for some reason.

hariharans29:
Can you please resolve conflicts ?

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
```diff
@@ -21,9 +23,200 @@ struct KaiTlsBuffers {
...
+
+                for (size_t k = 0; k < K; ++k) {
+                    g_kai_tls.gemv_lhs_row_tmp[k] = lhs_base[k * Data[b].ldb];
+                }
+                lhs_base = g_kai_tls.gemv_lhs_row_tmp.data();
+            }
+
+            // Temporary buffer for output row
+            g_kai_tls.output_tile.resize(rhs_shape);
+            std::fill_n(g_kai_tls.output_tile.data(), rhs_shape, 0.0f);
```
hariharans29: Is zeroing here necessary or something that the ukernel can take care of ?

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
```diff
@@ -370,29 +561,15 @@ Return Value:
...
-            kai_run_matmul_clamp_f32_f32p2vlx1_f32p2vlx1b_2vlx2vl_sme_mopa(
-                TileSizeM,
-                TileSizeN,
-                K,
-                ATile, BTile, temp_tile,
-                TileSizeN * sizeof(float), sizeof(float),
-                -std::numeric_limits<float>::max(), std::numeric_limits<float>::max()
-            );
-        }
+
```
hariharans29, line 576: Is zeroing on line 562 necessary ?

JonathanC-ARM: No its not required

JonathanC-ARM:
Hi @hariharans29 I've removed all extraneous changes from the latest commits. So its just GEMV now as such so should be easier to review. Also addressed the fill_n stuff you were right that wasn't required. Thanks for the comments!

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
```diff
@@ -21,9 +23,200 @@ struct KaiTlsBuffers {
...
+            // We have to handle the case where we transpose the data to the correct format as we used a transpose packing kernel
+            if (N == 1 && TransB == CblasNoTrans)
+            {
+                g_kai_tls.gemv_lhs_row_tmp.resize(K);
+
+                for (size_t k = 0; k < K; ++k) {
+                    g_kai_tls.gemv_lhs_row_tmp[k] = lhs_base[k * Data[b].ldb];
+                }
+                lhs_base = g_kai_tls.gemv_lhs_row_tmp.data();
+            }
```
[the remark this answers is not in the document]
hariharans29: @JonathanC-ARM : Is this comment by Copilot relevant ?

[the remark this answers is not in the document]
JonathanC-ARM: Hi Hari, somehow I missed this comment. I think it could be onto something here. I should be able to check this and address fairly quickly

hariharans29:
Apart from the comment from Copilot, the rest of the PR looks good to me. Thanks.

JonathanC-ARM:
Hi @hariharans29 I did make a change around the copilot suggestion. But saying that I don't think it was actually that big of a deal so to speak. Technically copilot was correct in isolation, but within ORT, the testing fixtures, and mlas this wasn't really possible. I've made the change on the off chance this could be called in isolation somehow and added tests that go down this path. 

But ORT/MLAS always use minimal leading dimensions: `lda = (TransA ? M : K)` and `ldb = (TransB ? K : N)`.
- For the degenerate GEMV shapes:
  - If `M == 1` and `TransA == CblasTrans`, then `lda == 1` -  A’s LHS row is contiguous.
  - If `N == 1` and `TransB == CblasNoTrans`, then `ldb == 1` - B’s LHS row is contiguous.
- Since the LHS vector is always contiguous under these rules, the gather is never required, so the original issue could never taken.

This could only occur if someone called MLAS with padded/non-minimal `lda`/`ldb`  which ORT doesn’t do today. But as it stands now if that was to ever happen we should be ok also.

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
```diff
@@ -21,9 +23,205 @@ struct KaiTlsBuffers {
...
+        for (size_t b = 0; b < BatchSize; ++b) {
+            // Depending on the value of M or N we might transpose when packing
+            const bool using_m_path = (M == 1);
+            const bool using_n_path = (N == 1 && !m_n_both_one);
+            CBLAS_TRANSPOSE tb = TransB;
+            size_t rhs_shape = N;
+            size_t rhs_ld = Data[b].ldb;
+            size_t lhs_ld = using_m_path ? Data[b].lda : Data[b].ldb;
+            const float* rhs_base = using_m_path ? reinterpret_cast<const float*>(Data[b].B) : reinterpret_cast<const float*>(Data[b].A);
+            const float* lhs_base = using_m_path ? static_cast<const float*>(Data[b].A) : static_cast<const float*>(Data[b].B);
```
edgchen1: nit: can we just use static_cast here?

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
```diff
@@ -21,9 +23,205 @@ struct KaiTlsBuffers {
...
+            {
+                tb = CblasTrans;
+                rhs_shape = M;
+                rhs_ld = Data[b].lda;
+                lhs_ld = Data[b].ldb;
+            }
+            // Prepare packed RHS if needed
+            const void* rhs_packed_ptr = nullptr;
+            if (Data[b].BIsPacked) {
+                rhs_packed_ptr = Data[b].B;
```
edgchen1, line 183: if `using_m_path` is false, would it still be correct to assign `rhs_packed_ptr` to `Data[b].B` here? wouldn't "RHS" be A in that case?

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
```diff
@@ -21,9 +23,205 @@ struct KaiTlsBuffers {
...
+
+        for (size_t b = 0; b < BatchSize; ++b) {
+            // Depending on the value of M or N we might transpose when packing
+            const bool using_m_path = (M == 1);
+            const bool using_n_path = (N == 1 && !m_n_both_one);
+            CBLAS_TRANSPOSE tb = TransB;
+            size_t rhs_shape = N;
+            size_t rhs_ld = Data[b].ldb;
+            size_t lhs_ld = using_m_path ? Data[b].lda : Data[b].ldb;
+            const float* rhs_base = using_m_path ? reinterpret_cast<const float*>(Data[b].B) : reinterpret_cast<const float*>(Data[b].A);
```
edgchen1: nit: use `static_cast` instead of `reinterpret_cast` for cast from `void*`
```suggestion
            const float* rhs_base = using_m_path ? static_cast<const float*>(Data[b].B) : static_cast<const float*>(Data[b].A);
```

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
```diff
@@ -21,9 +23,205 @@ struct KaiTlsBuffers {
...
+        if (N == 1 && Data->BIsPacked && !m_n_both_one)
+        {
+            // Exit early because we cannot support cases where N is 1 and B is already packed
+            return false;
+        }
+
+        for (size_t b = 0; b < BatchSize; ++b) {
+            // Depending on the value of M or N we might transpose when packing
+            const bool using_m_path = (M == 1);
+            const bool using_n_path = (N == 1 && !m_n_both_one);
```
edgchen1: how many paths are there? it seems like there would only be two valid ones, if M=1 or N=1 is a prerequisite. is it possible to just use one bool?

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
```diff
@@ -21,9 +23,205 @@ struct KaiTlsBuffers {
     std::vector<float> bias_zero;
     std::vector<std::byte> rhs_packed;
     std::vector<std::byte> lhs_packed;
+    std::vector<float> gemv_lhs_row_tmp;
 };
 static thread_local KaiTlsBuffers g_kai_tls;
 
+const kai_matmul_clamp_f32_f32p_f32p_ukernel sgemm_gemm = GetKleidiAISGemmUKernel();
+const kai_matmul_clamp_f32_f32_f32p_ukernel sgemm_gemv = GetKleidiAISGemvUKernel();
```
edgchen1:
```suggestion
const kai_matmul_clamp_f32_f32p_f32p_ukernel& sgemm_gemm = GetKleidiAISGemmUKernel();
const kai_matmul_clamp_f32_f32_f32p_ukernel& sgemm_gemv = GetKleidiAISGemvUKernel();
```

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
```diff
@@ -21,9 +23,205 @@ struct KaiTlsBuffers {
...
+            // Exit early because we cannot support cases where N is 1 and B is already packed
+            return false;
+        }
+
+        for (size_t b = 0; b < BatchSize; ++b) {
+            // Depending on the value of M or N we might transpose when packing
+            const bool using_m_path = (M == 1);
+            const bool using_n_path = (N == 1 && !m_n_both_one);
+            CBLAS_TRANSPOSE tb = TransB;
+            size_t rhs_shape = N;
```
edgchen1: would it be possible to move some of the setup that is not dependent on `Data[b]` outside of the loop?

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
JonathanC-ARM: Yeah thats completely fair, should really only be controlled by one bool, the m_n_both one case kinda threw me off somewhat but objectively yeah a single bool can suffice

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
JonathanC-ARM: This is correct, albeit maybe not immediately clear looking at the code, I will leave a comment to clarify in the code. But essentially
On the N path, we reject prepacked B, then pack A as the RHS with tb = CblasTrans. So rhs_packed_ptr points to packed A, not B.
The rhs_packed_ptr = Data[b].B fast-path is only taken on the M path.

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
JonathanC-ARM: Yeah I can move some of this outside of the loop for sure

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
JonathanC-ARM: Addressed in latest push

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
JonathanC-ARM: Fixed this inconsistency in the latest push

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
edgchen1: thanks for the update. it's clearer to me now.

JonathanC-ARM: No problem Edward glad to clear things up. And as always thanks for reviewing and the comments much appreciated!

hariharans29 (approved):

Issue #26174, opened 2025-09-26
[Build]  Fixed ORT model (platform = amd64) broken when compiled on an AMD CPU but not when compiled on an Intel CPU


wbudd:
### Describe the issue

I've been successfully using the `convert_onnx_models_to_ort.py` utility of ONNX Runtime for a number of years now in accordance with the following link:

https://onnxruntime.ai/docs/performance/model-optimizations/ort-format-models.html#convert-onnx-models-to-ort-format-script-usage

* I've been converting ResNets from a standard ONNX model file without any errors, both on an AMD machine and on an Intel machine.
* However when converting a certain GNN-variety of ONNX model, inference-time results are different when the build was carried out on an AMD machine compared to when built on an Intel machine (same two machine as the aforementioned ResNets).
    * When compiled on Intel, the build completes without any errors and so does any inference performed thereafter.
    * When compiled on AMD, the build completes without any errors, but when I try inference with ONNX Runtime on the resulting ORT model file the error listed below occurs.

I will follow up with hardware details, build flags, the applicable ONNX model file, and more when/if needed.

What is the recommended way to share an ONNX file? A direct attachment to this thread? The file size is only 13MB for this ONNX file.

### Urgency

I no longer have the Intel machine available to me, so this bug is starting to hamper my workflow a lot more than before...

### Target platform

amd64 (both AMD and Intel machines)

### Build script

```
python /onnxruntime/tools/python/convert_onnx_models_to_ort.py \
        --enable_type_reduction \
        --optimization_style Fixed \
        --save_optimized_onnx_model \
        --target_platform amd64 \
        some_MS-G3D_model.onnx
```
I can also share the subsequent C++ inference program compilation details if needed.

### Error / output

```
2025-09-26 03:50:39.118182196 [E:onnxruntime:, sequential_executor.cc:572 ExecuteKernel]
Non-zero status code returned while running Conv node. Name:'/sgcn1/sgcn1.0/mlp/layers.2/Relu_output_0_nchwc'
Status Message: Input channels C is not equal to kernel channels * group.
C: 24 kernel channels: 32 group: 1
```
On older versions of ONNX Runtime the location of the error seems to have been slightly difference, but the error itself was the same (this bug has been present for a few years at least!):
```
Non-zero status code returned while running Conv node. Name:'878_nchwc'
Status Message: Input channels C is not equal to kernel channels * group.
C: 24 kernel channels: 32 group: 1
``` 
[...]

PR #26306, opened 2025-10-14
Address MLAS NchwcBlockSize for AMD64 platforms behavior in the minimal builds

yuslepukhin:
### Fix MLAS issue that affects minimal build behavior
This PR addresses https://github.com/microsoft/onnxruntime/issues/26174

When ONNX Runtime optimizes a model that involves an NCHWC layout transformation, it inserts `ReorderInput` and `ReorderOutput` nodes. The resulting input shapes must be compatible with the W input of the subsequent Conv node. The `ReorderInput` node calculates its output shape based on the `NchwcBlockSize` for the given platform, which is expected because minimal-build models are platform-dependent.

However, MLAS has incorrectly placed the `#ifdef ORT_MINIMAL_BUILD` directive, resulting in a different value for the `NchwcBlockSize` constant between minimal and full builds—even though both are intended to run on the same platform.
This discrepancy causes inference to fail.

### NCHWC Transformer Test Additions

* Added multiple new tests to `graph_transform_test.cc` that verify `NchwcTransformer` correctly transforms ONNX graphs containing `Conv`, `Conv+Relu`, `Conv+unsupported activation`, and `MaxPool` nodes into their NCHWC equivalents, including checks for correct operator replacement and attribute handling.
* Included the `nchwc_transformer.h` header in test builds when contrib ops are enabled, ensuring the transformer is available for testing.

### Platform and Alignment Updates

* Refined logic in `platform.cpp` to set `NchwcBlockSize` and `PreferredBufferAlignment` only when specific CPU features are detected, improving accuracy of platform-specific optimizations. [[1]](diffhunk://#diff-2f1ef5fcaaa83277145acadea37a3d9da65e9605664f03efd4ca6208fa191e3bR445-R449) [[2]](diffhunk://#diff-2f1ef5fcaaa83277145acadea37a3d9da65e9605664f03efd4ca6208fa191e3bL468-L469)

### Test Infrastructure Improvements

* Updated graph transformer test builder to register the NCHWC domain, enabling proper opset versioning for NCHWC operators in test models.
* Enabled model dumping for debugging by setting `SAVE_TEST_GRAPH` to 1 in `graph_transform_test_builder.cc`.### Description

### DIFF
--- a/onnxruntime/core/mlas/lib/platform.cpp
+++ b/onnxruntime/core/mlas/lib/platform.cpp
@@ -442,7 +442,8 @@ Return Value:
                     this->QNBitGemmDispatch = &MlasSQNBitGemmDispatchAvx2vnni;
                 }
 
-                if (((Cpuid7[1] & 0x10000) != 0) && ((xcr0 & 0xE0) == 0xE0)) {
+                const bool AVX512FSupported = (Cpuid7[1] & 0x10000) != 0 && (xcr0 & 0xE0) == 0xE0;
+                if (AVX512FSupported) {
                     this->NchwcBlockSize = 16;
                     this->PreferredBufferAlignment = 64;
                 }
@@ -454,7 +455,7 @@ Return Value:
                 // operating system supports saving AVX512F state.
                 //
 
-                if (((Cpuid7[1] & 0x10000) != 0) && ((xcr0 & 0xE0) == 0xE0)) {
+                if (AVX512FSupported) {
 
                     this->GemmFloatKernel = MlasGemmFloatKernelAvx512F;
                     this->GemmDoubleKernel = MlasGemmDoubleKernelAvx512F;


onnxruntime/core/mlas/lib/platform.cpp
```diff
@@ -442,6 +442,11 @@ Return Value:
                     this->QNBitGemmDispatch = &MlasSQNBitGemmDispatchAvx2vnni;
                 }
 
+                if (((Cpuid7[1] & 0x10000) != 0) && ((xcr0 & 0xE0) == 0xE0)) {
```
hariharans29: I am not an expert on minimal builds and their use-case but something doesn't add up for me or maybe I am not understanding things right.
I would have thought the NCHWcBlockSize goes hand-in-hand with the **kernel** that actually uses it which is the NCHWcConvKernel. Why would there be a (minimal) build flavor that updates the NCHWcBlockSize when the kernel that supposedly uses it is not part of the build?

hariharans29:
Could the issue be something else - (i.e.) the ORT model is produced on an AMD64 platform that doesn't support AVX512F and is tried to run on an AMD64 platform that supports AVX512F or vice-versa ?

onnxruntime/core/mlas/lib/platform.cpp
skottmckay: IIRC AVX512 kernels are excluded in a minimal build due to the binary size cost, so unless that has changed the ifdef here was to match that the kernels weren't available.

hariharans29: Exactly. I think AVX512 kernels seem excluded from minimal builds - that was my understanding as well. So, if they are excluded, metadata of the kernels like the NCHWcBlockSize (which basically support the excluded NCHWc AVX512 Conv kernel) should also be excluded from minimal builds right? (i.e.) isn't the current setup correct?

yuslepukhin: Please, re-read the PR summary. Revisiting `ReorderInput` implementation may also be helpful. `NCHWcBlockSize ` is *always* present in the build with the default value and it can be used at any time. However, the default value is different in minimal build for the same platform from that of the full build for AMD64. This results in optimizations for NCHWc done with one value and runtime mismatch with another value.

yuslepukhin:
> Could the issue be something else - (i.e.) the ORT model is produced on an AMD64 platform that doesn't support AVX512F and is tried to run on an AMD64 platform that supports AVX512F or vice-versa ?

We can chat offline

onnxruntime/core/mlas/lib/platform.cpp
hariharans29: Not sure if this comment was before or after our call, but I understand that `NCHWcBlockSize` is always present. I just felt dicey about potentially matching the NCHWc block size that was clearly set for the AVX512 NCHWc Conv kernel with (let's pick an example) the SSE2 NCHWc Conv kernel. Its correctness and perf implications are not very clear.

edgchen1: this appears to be copied from here:

https://github.com/microsoft/onnxruntime/blob/878863d00ae7151e6965b3938fd049cad0a6f346/onnxruntime/core/mlas/lib/platform.cpp#L447-L452

the expression is kind of cryptic. can we instead store the result in a constant with a meaningful name and reuse that?

yuslepukhin: The block size for AVX512 works with SSE. The concern was about the perf. I did run some perf with the full build (with AVX512) and optimized model, and a minimal build with ORT optimized model (uses SSE) and the numbers look good.
The model was customer provided which was convolution heavy.

Normal Release build on optimized model
=========================================
onnxruntime_perf_test -s -o 0 -e cpu -I -S 0 -m times -r 100 "C:\Users\dmitrism\OneDrive - Microsoft\dev\AMDOrtFormat_26174\repro\msg3d.optimized.onnx"

Session creation time cost: 0.0414746 s
First inference time cost: 10 ms
Total inference time cost: 0.895583 s
Total inference requests: 100
Average inference time cost total: 8.955835 ms
Total inference run time: 0.895989 s
Number of inferences per second: 111.609
Avg CPU usage: 49 %
Peak working set size: 55029760 bytes
Avg CPU usage:49
Peak working set size:55029760
Runs:100
Min Latency: 0.0075418 s
Max Latency: 0.0109003 s
P50 Latency: 0.008976 s
P90 Latency: 0.0101723 s
P95 Latency: 0.0107656 s
P99 Latency: 0.0109003 s
P999 Latency: 0.0109003 s

ORT optimized minimal build that uses SSE only (with a fix)
===============================
onnxruntime_perf_test -s -e cpu -I -S 0 -m times -r 100 "C:\Users\dmitrism\OneDrive - Microsoft\dev\AMDOrtFormat_26174\repro\msg3d.ort"

Session creation time cost: 0.0718184 s
First inference time cost: 11 ms
Total inference time cost: 0.829151 s
Total inference requests: 100
Average inference time cost total: 8.291509 ms
Total inference run time: 0.829581 s
Number of inferences per second: 120.543
Avg CPU usage: 49 %
Peak working set size: 40538112 bytes
Avg CPU usage:49
Peak working set size:40538112
Runs:100
Min Latency: 0.0068583 s
Max Latency: 0.0098785 s
P50 Latency: 0.00821 s
P90 Latency: 0.009382 s
P95 Latency: 0.0095542 s
P99 Latency: 0.0098785 s
P999 Latency: 0.0098785 s

Issue #26282, opened 2025-10-12
Reduce ONNX Runtime GPU wheel size using `fatbin` compression


XXXXRT666:
### Describe the issue

Following [PR #26002](https://github.com/microsoft/onnxruntime/pull/26002), certain GPU architectures were removed due to the overall wheel size exceeding GitHub and PyPI limits.  

However, newer versions of `nvcc` (introduced since CUDA 12.8) support `fatbin` compression (`-Xfatbin=-compress-all -compress-mode=MODE`), which can significantly reduce binary size without affecting functionality.

Below are my test results comparing different compression modes (`size`, `balance`, `speed`) under both CUDA 12.8 and CUDA 13.0.  

`default` indicates the mode automatically applied by the corresponding `nvcc` version

**Note: enabling `--compress-mode` requires a driver version at least as new as the one shipped with CUDA 12.4, which is why `PyTorch` only enables it for wheels built with CUDA 13.0 and later.**

| CUDA Ver | Mode       | Wheel Size (MB) | Compile Time (min:s) | 
|-----------|-------------|-----------------|----------------------|
| 13.0      | Speed       | 923.9           | 187m51.531s          | 
| 13.0      | Balance (Default)     | 516.1           | 179m6.947s           | 
| 13.0      | Size        | 360.0           | 191m42.614s          | 
| 12.8      | Speed (Default)       | 689.5           | 159m1.095s           | 
| 12.8      | Balance     | 435.5           | 185m53.233s          | 
| 12.8      | Size        | 309.3           | 182m14.164s          |

### Build Script

First replace `set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} -Xfatbin=-compress-all")` with `set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} -Xfatbin=-compress-all -compress-mode=YOUR_MODE")`

```bash
docker run -it -v ~/onnxruntime:/root/onnxruntime --name ort continuumio/miniconda3

cd ~/onnxruntime
conda create -n onnx python=3.12 -y && conda activate onnx
pip install -r requirements.txt

conda install cuda-nvcc=13.0/12.8 cuda-toolkit cudnn -c nvidia -y
conda install gcc gxx cmake ninja -c conda-forge

ln -s /opt/conda/envs/onnx/targets/sbsa-linux/include/cccl/cuda /opt/conda/envs/onnx/targets/sbsa-linux/include/cuda
[...]

PR #26267, opened 2025-10-09
[CUDA] Upgrade build pipelines to use CUDA 12.8 + cuDNN 9.8

tianleiwu:
This upgrades CUDA 12.2 + cuDNN 9.5 to CUDA 12.8 + cuDNN 9.8 in CI pipelines, so that we can build 120-real to support Blackwell GPU.

To speed up build, we also disable relocatable-device-code.

MSVC is updated to latest for some windows build pipelines.

#### Known issues

Some onnx models (yolo v3, yolo v4, mobilenet v1)  failed to run due to cudnn frontend failed to find engine plan. We will try upgrade cudnn frontend later. Related failed tests are disabled for now.

PR #26307, opened 2025-10-14
[CUDA] Adjust package build settings for Blackwell GPU

tianleiwu:
### Description

We recently upgraded CI pipeline to CUDA 12.8 (See https://github.com/microsoft/onnxruntime/pull/26267), which allows us to build native code for Blackwell GPU of [compute capability](https://developer.nvidia.com/cuda-gpus) 120. Here we update the CMAKE_CUDA_ARCHITECTURES settings.

Cloud GPUs GB200, GB300 etc have compute capability of 100 and 103, which are excluded due to package size limit.

Since CUDA 12.4, there is a new nvcc option for compress mode:
https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/#compress-mode-default-size-speed-balance-none-compress-mode
Note that this option is not compatible with drivers released before CUDA 12.4 Release so we only apply it to CUDA 13 build. That could reduce package size for cuda 13 significantly.

Python wheel size:
* Linux: 322 MB
* Windows: 270 MB

### Motivation and Context

Binary size can be reduced as reported in https://github.com/microsoft/onnxruntime/issues/26282


tianleiwu:
Similar change is merged: https://github.com/microsoft/onnxruntime/commit/8fceca0d853090b0cde8ce939802a4f8d79f0800

PR #26308, opened 2025-10-15
[WebNN] Fallback int64 indices to int32

Honry:
ONNX's ScatterND and ScatterElements limit their indices input to int64, but some WebNN backends only support int32 indices. As a workaround for such backends, we can insert a Cast operation to convert the data type.

### DIFF
--- a/onnxruntime/core/providers/webnn/builders/impl/scatterElements_op_builder.cc
+++ b/onnxruntime/core/providers/webnn/builders/impl/scatterElements_op_builder.cc
@@ -102,8 +102,8 @@ bool ScatterElementsOpBuilder::HasSupportedInputsImpl(const GraphViewer&, const 
   // ONNX specifies that indices must use int64, but some WebNN backends only support int32.
   // Allows to use int32 as a workaround for such backends.
   can_fallback_int64_to_int32_ = CanFallbackInt64ToInt32(wnn_limits, "scatterElements", "indices");
-  return IsDataTypeSupportedByOp(op_type, indices_type, wnn_limits, "indices", "indices", logger) ||
-         can_fallback_int64_to_int32_;
+  return can_fallback_int64_to_int32_ ||
+         IsDataTypeSupportedByOp(op_type, indices_type, wnn_limits, "indices", "indices", logger);
 }
 
 void CreateScatterElementsOpBuilder(const std::string& op_type, OpBuilderRegistrations& op_registrations) {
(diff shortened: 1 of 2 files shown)


Honry:
@fdwr, @guschmue, PTAL, thanks! This will make Whisper Base demo on WebNN Developer Preview can fully run behind TFLite backend without any fallback.

onnxruntime/core/providers/webnn/builders/impl/scatterElements_op_builder.cc
```diff
@@ -86,9 +94,16 @@ bool ScatterElementsOpBuilder::HasSupportedInputsImpl(const GraphViewer&, const
...
+  if (!IsDataTypeSupportedByOp(op_type, data_type, wnn_limits, "input", "data", logger) ||
+      !IsInputRankSupportedByOp(node, wnn_limits, logger)) {
+    return false;
+  }
+
+  // ONNX specifies that indices must use int64, but some WebNN backends only support int32.
+  // Allows to use int32 as a workaround for such backends.
+  can_fallback_int64_to_int32_ = CanFallbackInt64ToInt32(wnn_limits, "scatterElements", "indices");
+  return IsDataTypeSupportedByOp(op_type, indices_type, wnn_limits, "indices", "indices", logger) ||
+         can_fallback_int64_to_int32_;
```
fdwr:
```c++
  return can_fallback_int64_to_int32_ ||
      IsDataTypeSupportedByOp(op_type, indices_type, wnn_limits, "indices", "indices", logger);
```
Suggest reversing the test, since `can_fallback_int64_to_int32_` already checked some data type support, that might skip the `IsDataTypeSupportedByOp` call.

fdwr (dismissed):
👍 Minor thought, else LGTM.

onnxruntime/core/providers/webnn/builders/impl/scatterElements_op_builder.cc
Honry: Good point. Fixed.

fdwr (approved):
👍

fdwr:
Hmm, I've retried the Linux CI build and CUDA build twice now. Wonder if this happens in main now.

image

Honry:
> Hmm, I've retried the Linux CI build and CUDA build twice now. Wonder if this happens in main now.
> 
> image

Let me rebase it to main and pls. help retrigger the CI.

guschmue (approved):

Honry:
@guschmue, all CLs passed, could you please merge this one? Thanks!

PR #26313, opened 2025-10-15
Use time_costs_total for output_stats

keshavv27:
### Description
- Use total inference time instead of the submission time for output statistics calculation

### Motivation and Context
- The min, max, and other statistics reported for inference were using device submission time instead of the inference time.

@ishwar-raut1 @gaugarg-nv @thevishalagarwal @umangb-09 @gedoensmax


keshavv27:
Can the pipelines be retriggered? The failures look unrelated to this change. Pinging @snnn @chilo-ms , thank you

onnxruntime/test/perftest/performance_runner.cc
```diff
@@ -70,7 +70,7 @@ void PerformanceResult::DumpToFile(const std::basic_string<ORTCHAR_T>& path, boo
 
   if (have_file) {
     for (size_t runs = 0; runs < time_costs_submission.size(); runs++) {
-      outfile << model_name << "," << time_costs_submission[runs] << "," << peak_workingset_size << ","
+      outfile << model_name << "," << time_costs_submission[runs] << "," << time_costs_total[runs] << "," << peak_workingset_size << ","
```
edgchen1: is there any indication in the output of what this extra value represents?

keshavv27: I wanted to maintain the format of the output (as the other parameters are also not described while being written to outfile)
so didn't explicitly add what the value represents here which is the total inference time

edgchen1: I'd say that the format has been updated because it now has an additional value. if maintaining the old format is a concern, you could replace the old "submission time" with the "total time".

keshavv27: That makes sense, changing in next revision. Thanks!

edgchen1 (approved):

PR #26334, opened 2025-10-17
[webgpu] Don't use num_workgroups when use indirect dispatch

qjia7:
This pull request updates the FlashAttention WebGPU implementation to improve support for indirect dispatch. The main changes ensure that when indirect dispatch is used, the shader receives the actual workgroup dimensions from an input buffer rather than relying on built-in variables, which avoids duplication overhead in Dawn/WebGPU. See https://source.chromium.org/chromium/chromium/src/+/main:third_party/dawn/src/dawn/native/ComputePassEncoder.cpp;l=275.
This PR fixes the issue that indirect dispatch is slower than normal dispatch for the same program. 
With this change, the phi4 with graph capture enabled can run 145 tps from 125 tps on NV 5080.

### DIFF
--- a/onnxruntime/core/providers/webgpu/program.cc
+++ b/onnxruntime/core/providers/webgpu/program.cc
@@ -362,9 +362,6 @@ ProgramBase& ProgramBase::SetDispatchGroupSize(uint32_t x, uint32_t y, uint32_t 
 
 ProgramBase& ProgramBase::SetIndirectDispatchTensor(const Tensor* indirect_dispatch_tensor) {
   indirect_dispatch_tensor_ = indirect_dispatch_tensor;
-  // Automatically add the indirect buffer as an input so the shader can read dispatch dimensions.
-  // This avoids the need to manually add it in every program that uses indirect dispatch.
-  AddInput({indirect_dispatch_tensor, ProgramTensorMetadataDependency::None});
   return *this;
 }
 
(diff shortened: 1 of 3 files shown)


onnxruntime/contrib_ops/webgpu/bert/flash_attention.cc
```diff
@@ -287,8 +293,10 @@ Status ComputeFlashAttentionDecodeSplitVxScore(onnxruntime::webgpu::ComputeConte
                      {present_value, ProgramTensorMetadataDependency::TypeAndRank, components}});
   program.AddOutputs({{out_split_vx, ProgramTensorMetadataDependency::TypeAndRank, components}});  // [B, N, split_k, head_size]
   if (use_indirect_dispatch) {
-    program.AddInput({seqlen_k, ProgramTensorMetadataDependency::None})
-        .SetIndirectDispatchTensor(indirect_buffer);
+    program.AddInput({seqlen_k, ProgramTensorMetadataDependency::None});
+    // Add indirect buffer as input so shader can read workgroup dimensions
+    program.AddInput({indirect_buffer, ProgramTensorMetadataDependency::None});
```
fs-eire: what about doing this in a shared place so that no need to manually add this every time?

The program object has the reference to the indirect buffer because `SetIndirectDispatchTensor` is called.

qjia7: Great suggestion! Done.

PR #26336, opened 2025-10-17
[QNN EP] Skip inputs/outputs shape validation for QNN Batch Multiple

chunghow-qti:
### Description

* QNN EP supports batch multiplier during inference, which allows the compile batch size is different to running batch size. 
* However, InferenceSession::ValidateInputs & InferenceSession::ValidateOutputs validates the input_output_shape against the expected_shape.
* We avoid these validations if qnn batch multiplier options is used.
* p.s. QNN EP does not support dynamic shapes. We cannot convert ONNX graphs with dynamic shapes into static shapes inside QNN EP. There is no internal mechanism to infer or materialize static shapes from dynamic ones.
* A separate PR will be submitted for the implementation of batch multiplier support in QNN EP.

### Motivation and Context

* This change supports batch multiplier in QNN API to ORT as described in this page: https://docs.qualcomm.com/bundle/publicresource/topics/80-63442-10/function_QnnGraph_8h_1a3ea05f42a9295f9a74a2e3a0cdd64228.html

### DIFF
--- a/onnxruntime/core/session/inference_session.cc
+++ b/onnxruntime/core/session/inference_session.cc
@@ -2606,30 +2606,12 @@ common::Status InferenceSession::CheckShapes(const std::string& input_output_nam
                            " Please fix either the inputs/outputs or the model.");
   }
 
-#ifdef USE_QNN
-  // Helper function to check whether QNN EP is used & all nodes are assigned to QNN EP,
-  // and relax the constraint to support batch multiplier on the first dimension.
-  // We will check whether only the Htp backend is used inside QnnModel::ExecuteGraph.
-  auto is_valid_qnn_batch_multiplier = [this](int64_t input_dim, int64_t expected_dim, const Graph& graph) -> bool {
-    if (!AreAllNodesInMainGraphAssignedToOneEp(graph, kQnnExecutionProvider)) {
-      LOGS_IF(input_dim != expected_dim, *session_logger_, WARNING) << "input batch size and expected batch size are different. "
-                                                                    << "Batch multiplier is only supported on the QNN EP, "
-                                                                    << "but some nodes in the graph are assigned to other EPs";
-      return false;
-    }
-    return expected_dim > 0 && input_dim % expected_dim == 0;
-  };
-#endif
-
   InlinedVector<size_t> invalid_dim_indices;
   for (size_t i = 0; i < shape_size; ++i) {
     if (expected_shape[i] < 0) {
       continue;  // this represents a symbolic shape dimension
-#ifdef USE_QNN
-    } else if (i == 0 && is_valid_qnn_batch_multiplier(input_output_shape[i], expected_shape[i], model_->MainGraph())) {
-      continue;  // Qnn API supports batch multiplier, but the running batch size must be divisible by the original batch size.
-#endif
-    } else if (input_output_shape[i] != expected_shape[i]) {
+    }
+    if (input_output_shape[i] != expected_shape[i]) {
       invalid_dim_indices.push_back(i);
     }
   }
@@ -2999,10 +2981,17 @@ Status InferenceSession::Run(const RunOptions& run_options,
 
       // log evaluation start to trace logging provider
       env.GetTelemetryProvider().LogEvaluationStart(session_id_);
-
+#ifdef USE_QNN
+      const bool batch_multiplier = session_options_.config_options.GetConfigOrDefault(kOrtSessionOptionsQnnHtpBatchMultiplier, "0") == "1";
+      if (!batch_multiplier) {
+        LOGS(*session_logger_, INFO) << "Enable QNN HTP batch mutliplier. Don't validate the inputs/outptus";
+        ORT_RETURN_IF_ERROR_SESSIONID_(ValidateInputs(feed_names, feeds));
+        ORT_RETURN_IF_ERROR_SESSIONID_(ValidateOutputs(output_names, p_fetches));
+      }
+#else
       ORT_RETURN_IF_ERROR_SESSIONID_(ValidateInputs(feed_names, feeds));
       ORT_RETURN_IF_ERROR_SESSIONID_(ValidateOutputs(output_names, p_fetches));
-
+#endif
       // shrink certain default memory arenas if the user has requested for it
       const std::string& shrink_memory_arenas =
           run_options.config_options.GetConfigOrDefault(kOrtRunOptionsConfigEnableMemoryArenaShrinkage, "");
--- a/include/onnxruntime/core/session/onnxruntime_session_options_config_keys.h
+++ b/include/onnxruntime/core/session/onnxruntime_session_options_config_keys.h
@@ -415,3 +415,10 @@ static const char* const kOrtSessionOptionsFailOnSuboptimalCompiledModel =
 // "high_power_saver", "low_balanced", "extreme_power_saver", "low_power_saver", "power_saver",
 // "sustained_high_performance". Default to "default".
 static const char* const kOrtEpDynamicOptionsQnnHtpPerformanceMode = "ep.dynamic.qnn_htp_performance_mode";
+
+// Enable QNN HTP batch multiplier
+//
+// Option values:
+// - "0": QNN htp batch multiplier is disabled. [DEFAULT]
+// - "1": QNN htp batch multiplier is enabaled.
+static const char* const kOrtSessionOptionsQnnHtpBatchMultiplier = "ep.qnn.enable_htp_batch_multiplier";


chunghow-qti:

onnxruntime/core/session/inference_session.cc
```diff
@@ -2606,12 +2606,34 @@ common::Status InferenceSession::CheckShapes(const std::string& input_output_nam
...
+  };
+
   InlinedVector<size_t> invalid_dim_indices;
   for (size_t i = 0; i < shape_size; ++i) {
     if (expected_shape[i] < 0) {
       continue;  // this represents a symbolic shape dimension
-    }
-    if (input_output_shape[i] != expected_shape[i]) {
+    } else if (i == 0 && is_qnn_batch_multiplier_valid(input_output_shape[i], expected_shape[i], model_->MainGraph())) {
+      continue;  // Qnn API supports batch multiplier, but the running batch size must be divisible by the original batch size.
```
edgchen1: I don't think we want to add QNN EP-specific relaxing of shape validation here. if the graph was not running on the QNN EP, would it be considered invalid? can you elaborate on what you are trying to do?

chunghow-qti: Hi, thanks for the question. 

The shape only be considered valid under QNN EP due to its support for batch multiplier.
If the graph is not assigned to QNN EP, the original shape check logic applies, and the shape would be considered invalid.

QNN EP supports **batch multiplier**, which allow the model to be compiled with smaller batch size (e.g. 2), and then run inference with a larger batch size (e.g. 128), as long as the inference batch size is divisible by the compile batch size.

This can help reduce compile time (session creation time), while still supporting larger inference batches. 
What we're trying to do is ensure that this flexibility is preserved during shape validation when the graph is intended to run on QNN EP.

chunghow-qti:
Hi @yuslepukhin , this PR will impact our developing feature, it would be great if you can help review and share the thought on this PR, thanks!

onnxruntime/core/session/inference_session.cc
edgchen1: > The shape only be considered valid under QNN EP due to its support for batch multiplier.
> If the graph is not assigned to QNN EP, the original shape check logic applies, and the shape would be considered invalid.

thanks for clarifying.

I still think that we don't want to add QNN EP-specific handling here. looking ahead, consider that for plugin EPs, it would be even less desirable to have similar hardcoded logic at this point.

> QNN EP supports batch multiplier, which allow the model to be compiled with smaller batch size (e.g. 2), and then run inference with a larger batch size (e.g. 128), as long as the inference batch size is divisible by the compile batch size.

is it possible for the QNN EP to manage this optimization internally? e.g., identify that a smaller batch size than the one in the actual shape can be used for compilation.

qti-chuteng: Hi @edgechen1, here is more detailed explanation about this PR.
 
When using onnxruntime_perf_test.exe, onnx_test_runner.exe, or any application that calls InferenceSession, it will use InferenceSession::Initiatization() and InferenceSession::Run().
 
**Current Flow**

- InferenceSession::Initiatization()
    - Uses the source onnx model's input shape as the expected dimension. If a dimension in the input shape is dynamic, it is labeled as -1. Otherwise, it is a positive integer.
- InferenceSession::Run()
    - Checks the input data shape.
        - If a dimension in the input shape is dynamic, it ignores the check.
        - Otherwise, the dimension must match.
    - After checking, it calls the QNN function to execute the graph.
 
**Our Usecase**
- InferenceSession::Initiatization()
    - The source onnx model's input shape is used, and we assign the batch dimension with a positive integer as base batch (ex: 2)
- InferenceSession::Run()
    - Checks the input data shape.
        - If a dimension in the input shape is dynamic, it ignores the check. → Won't happen in our usecase
        - Otherwise, the dimension must match. → This is what we don't want, because QNN supports "batch multiplier", which allows the running input batch to be divisible by the assigned batch number. (ex: 4, 6, 8)
After checking, it calls the QNN function to execute the graph.

**Reason of Change**
- As in our Usecase 2.1.2, this check prevents us from enabling the batch multiplier, which would allow us to improve context preparation efficiency. For example, we could compile the model with batch=2 and support different input batch sizes, rather than being limited to a fixed batch size. 
- If we do not relax this check, there is no alternative to adjust it on the QNN side since this check occurs at the very first phase of inference.

edgchen1: > Otherwise, the dimension must match. → This is what we don't want, because QNN supports "batch multiplier", which allows the running input batch to be divisible by the assigned batch number. (ex: 4, 6, 8)

I'm not convinced that what you're proposing is the behavior that we (ORT) would want. My understanding is that the input dimension at runtime should match what is specified in the static shape. If one wants to set different values for an input dimension, then it should be a dynamic dimension.

Could the QNN EP support this by allowing the "batch multiplier" dimension to be dynamic?

onnxruntime/core/session/inference_session.cc
```diff
@@ -2606,12 +2606,30 @@ common::Status InferenceSession::CheckShapes(const std::string& input_output_nam
...
+
   InlinedVector<size_t> invalid_dim_indices;
   for (size_t i = 0; i < shape_size; ++i) {
     if (expected_shape[i] < 0) {
       continue;  // this represents a symbolic shape dimension
-    }
-    if (input_output_shape[i] != expected_shape[i]) {
+#ifdef USE_QNN
+    } else if (i == 0 && is_valid_qnn_batch_multiplier(input_output_shape[i], expected_shape[i], model_->MainGraph())) {
+      continue;  // Qnn API supports batch multiplier, but the running batch size must be divisible by the original batch size.
```
skottmckay: Can you explain the logic more. Naively if I have a shape of {N, C, H, W} with N being the batch and say {4, 3, 128, 128}. If the batch multiplier is 4 what does the shape change to? 

It feels off to be providing an input shape that does not match what the ONNX model has defined as the required input shape, and to have EP specific logic in ORT code (especially with the move to plugin EPs where the ORT implementation should be EP agnostic). Why can't the QNN EP do that shape re-interpretation internally?

onnxruntime/core/session/inference_session.cc
qti-chuteng: > Could the QNN EP support this by allowing the "batch multiplier" dimension to be dynamic?

For QNN-EP, dynamic dimension support has not been implemented yet.
[https://github.com/microsoft/onnxruntime/blob/615c22bf6ba5d259c49484416d0a87a91d936e13/onnxruntime/…](https://github.com/microsoft/onnxruntime/blob/615c22bf6ba5d259c49484416d0a87a91d936e13/onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.cc#L335)

The "batch multiplier" helps save preparation time by compiling the smallest batch, allowing different batch sizes to run without recompiling. However, even we enable dynamic dimensions in QNN-EP, it would not help reduce preparation time in this case.

onnxruntime/core/session/inference_session.cc
yuslepukhin: I agree that the EP specific logic must be inside the EP.

onnxruntime/core/session/inference_session.cc
skottmckay: > **Our Usecase**
> 
> * InferenceSession::Initiatization()
>   
>   * The source onnx model's input shape is used, and we assign the batch dimension with a positive integer as base batch (ex: 2)
> * InferenceSession::Run()
>   
>   * Checks the input data shape.
>     
>     * If a dimension in the input shape is dynamic, it ignores the check. → Won't happen in our usecase
>     * Otherwise, the dimension must match. → This is what we don't want, because QNN supports "batch multiplier", which allows the running input batch to be divisible by the assigned batch number. (ex: 4, 6, 8)
>       After checking, it calls the QNN function to execute the graph.

A shape can be dynamic or fixed. If it's fixed, the data must match. Creating a new "dimension must be a multiple of this value" state would invalidate a lot of assumptions in the code and it is not part of the ONNX spec. 

Having an EP specific "trust us it's okay" piece of code in ORT core doesn't really work either. What happens if another EP wants a slightly different new state? Do we accumulate a bunch of special cases that potentially conflict with each other?

Can you set some other metadata in the model to indicate the smallest batch size? e.g. put entries in the model metadata like "batch_multiplier":"4" and "batch_multiplier_dim_name":"N" if the symbolic dimension name for the batch size is 'N'. that lets you specify the value and the dim_name it applies to.

chunghow-qti: Hi @skottmckay @edgchen1 , 

In the batch multiplier scenario (where the compile-time batch size may differ from the runtime batch size), it appears that only ONNX models with dynamic batch size can pass the checkShape validation. However, since QNN EP does not support dynamic shapes, we require a static ONNX model inside the EP.
Would it be feasible to apply a **free dimension override** to ONNX Graph within QNN EP to convert the dynamic shape of the ONNX Graph into a static shape, and then perform **shape inference** to ensure that the output shape is also fully static? 

Given that GraphViewer or Graph cannot be constructed inside QNN EP, we cannot replicate the ONNX graph internally within the EP.

qti-chuteng:
Hi @edgchen1, @yuslepukhin @skottmckay,
Thank you for reviewing. We've updated this PR with a new solution that reduces the specific logic in the common code. Could you please take another look?

onnxruntime/core/session/inference_session.cc
```diff
@@ -2958,10 +2981,17 @@ Status InferenceSession::Run(const RunOptions& run_options,
 
       // log evaluation start to trace logging provider
       env.GetTelemetryProvider().LogEvaluationStart(session_id_);
-
+#ifdef USE_QNN
+      const bool batch_multiplier = session_options_.config_options.GetConfigOrDefault(kOrtSessionOptionsQnnHtpBatchMultiplier, "0") == "1";
+      if (!batch_multiplier) {
```
edgchen1, line 2986: disabling input/output validation entirely seems a bit drastic. while it would allow your use case, it would also let other invalid cases through and potentially lead to runtime errors that are harder to debug.

quic-tirupath: Agree with @edgchen1. We should not disable the input/output validation entirely.

@qti-chuteng,
We should still validate the input and output shapes with batch_dim at run is integral multiplier of original dimension while other dimensions should match the original shape as shown in below example:

original shape: [2, 224, 224, 3]
input shape at run inference: [10, 224, 224, 3]
where, batch(run_input_shape) % batch(orig_input_shape) == 0 and other dim values should match between both orig_input_shape and run_input_shape.

yuslepukhin:
We should not disable this validation.

PR #26341, opened 2025-10-17
[QNN EP] Enable v81 devices

qti-jkilpatrick:
### Description

This change adds support for the latest Qualcomm NPUs with the v81 architecture.

Specifically:

* Include v81 stub/skel/cat files in build artifacts.
* Disable tests that are broken on v79 and v81.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
@@ -32,6 +32,11 @@
 // Flag to determine if Backend should do node validation for each opNode added
 #define DO_GRAPH_NODE_VALIDATIONS 1
 
+// Ensure that we have a recent enough version of QNN
+static_assert(QNN_API_VERSION_MAJOR > 2 ||
+                  (QNN_API_VERSION_MAJOR == 2 && QNN_API_VERSION_MINOR >= 29),
+              "Minimum required QAIRT SDK version is 2.39.0");
+
 namespace onnxruntime {
 namespace qnn {
 


tools/ci_build/github/azure-pipelines/templates/c-api-artifacts-package-and-publish-steps-windows-qnn.yml
```diff
@@ -86,6 +86,7 @@ steps:
           copy $(Build.BinariesDirectory)\${{parameters.buildConfig}}\${{parameters.buildConfig}}\QnnHtpPrepare.dll $(Build.BinariesDirectory)\${{parameters.artifactName}}\lib
           copy $(Build.BinariesDirectory)\${{parameters.buildConfig}}\${{parameters.buildConfig}}\QnnHtpV68Stub.dll $(Build.BinariesDirectory)\${{parameters.artifactName}}\lib
           copy $(Build.BinariesDirectory)\${{parameters.buildConfig}}\${{parameters.buildConfig}}\QnnHtpV73Stub.dll $(Build.BinariesDirectory)\${{parameters.artifactName}}\lib
+          copy $(Build.BinariesDirectory)\${{parameters.buildConfig}}\${{parameters.buildConfig}}\QnnHtpV81Stub.dll $(Build.BinariesDirectory)\${{parameters.artifactName}}\lib
```
qti-jkilpatrick, line 89: From here to the end, I'm unable to test this change locally so please look closely. I found these files via vigorous `git grep` and hunting through history. Unfortunately, it was a bit murky due to refactors and no analogous change adding a new architecture version (v68 and v73 were added together).

edgchen1: how should we test these changes?

onnxruntime/test/providers/qnn/gemm_op_test.cc
```diff
@@ -447,8 +447,19 @@ TEST_F(QnnHTPBackendTests, Gemm_Static_B_And_Bias) {
                                         ExpectedEPNodeAssignment::All);
 }
 
+// Broken on v79 and v81 devices:
```
edgchen1, line 450: where should the fix be? is the test invalid, or is it a known issue for v79/v81 that will be fixed in future versions?

tools/ci_build/github/azure-pipelines/templates/c-api-artifacts-package-and-publish-steps-windows-qnn.yml
qti-jkilpatrick: I am not familiar enough with your build to answer this question authoritatively. Code inspection suggests that this logic is hit when `tools/ci_build/github/azure-pipelines/custom-nuget-packaging-pipeline.yml` is run, though I cannot find any references to that file (`git grep custom-nuget-packaging-pipeline.yml` returns nothing).

At a high level though, the hope is that all Windows packages you create should now include the v81 files. I've confirmed this is the case for the Python wheel, but it would also be good to build nuget packages and make sure they're there. I'm not aware of other places this is needed, but it's possible that I just don't know all the binary distribution formats.

onnxruntime/test/providers/qnn/gemm_op_test.cc
qti-jkilpatrick: These are known issues that are being tracked internally at Qualcomm.

edgchen1: is the plan to re-enable these tests at some point?

edgchen1:
was v81 support only added in a recent version of the QNN SDK? if so, should we attempt to keep supporting building with older SDK versions or just increase the minimum required SDK version?

onnxruntime/test/providers/qnn/gemm_op_test.cc
qti-jkilpatrick: Yes, of course

qti-jkilpatrick:
> was v81 support only added in a recent version of the QNN SDK? if so, should we attempt to keep supporting building with older SDK versions or just increase the minimum required SDK version?

Based on what I know, I think bumping the minimum required version is the way to go. A hard build failure is preferable to accidentally incomplete packages. Given that there isn't any testing of this repo or its build products on real Qualcomm devices, such an error would likely go undetected during the release process (similar to the 1.22.2 wheels reverting to x64 from ARM64x). Of course we test on real hardware on our internal fork, but the packaging parts of the build are not shared.

An alternative is to make more parts of the build aware of what's in the various QNN versions, but this further complicates an already complicated system. This is especially challenging/risky because I can't test any of the affected code. What I do not know are the consequences for you of raising the minimum version other than having to revert two changes instead of one if the latest QNN version ends up needing to be yanked. If they're dire, I would be happy to investigate a more complex change.

edgchen1:
> Based on what I know, I think bumping the minimum required version is the way to go. A hard build failure is preferable to accidentally incomplete packages.

this sounds good to me. could we make it an explicit hard failure with a clear error message? e.g.,
```c++
static_assert(QNN_API_VERSION_MAJOR > 2 ||
              (QNN_API_VERSION_MAJOR == 2 && QNN_API_VERSION_MINOR >= <required minor version>),
              "Minimum required QNN SDK version is ...")
```

qti-jkilpatrick:
> > Based on what I know, I think bumping the minimum required version is the way to go. A hard build failure is preferable to accidentally incomplete packages.
> 
> this sounds good to me. could we make it an explicit hard failure with a clear error message? e.g.,
> 
> ```c++
> static_assert(QNN_API_VERSION_MAJOR > 2 ||
>               (QNN_API_VERSION_MAJOR == 2 && QNN_API_VERSION_MINOR >= <required minor version>),
>               "Minimum required QNN SDK version is ...")
> ```

Great idea, @edgchen1. Done.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -32,6 +32,11 @@
 // Flag to determine if Backend should do node validation for each opNode added
 #define DO_GRAPH_NODE_VALIDATIONS 1
 
+// Ensure that we have a recent enough version of QNN
+static_assert(QNN_API_VERSION_MAJOR > 2 ||
+                  (QNN_API_VERSION_MAJOR == 2 && QNN_API_VERSION_MINOR >= 29),
```
edgchen1, line 37: should the minor version value be 39?
```suggestion
                  (QNN_API_VERSION_MAJOR == 2 && QNN_API_VERSION_MINOR >= 39),
```

qti-jkilpatrick: Interestingly, no. This is what I find in `QnnCommon.h` in the `2.39.0` release:

```c++
// Provide values to use for API version.
#define QNN_API_VERSION_MAJOR 2
#define QNN_API_VERSION_MINOR 29
#define QNN_API_VERSION_PATCH 0
```

edgchen1: hm, will this detect the minimum version that we require (and not also an earlier version)?

qti-jkilpatrick: Yes, it'll work. Here's what's in 2.38.0:

```c++
// Provide values to use for API version.
#define QNN_API_VERSION_MAJOR 2
#define QNN_API_VERSION_MINOR 28
#define QNN_API_VERSION_PATCH 1
```

I don't pretend to understand why these differ from the QAIRT package version, but do claim that it suffices for this task.

edgchen1: cool, sounds good to me then. thanks for checking.

PR #26342, opened 2025-10-17
VitisAI bugfix model clone optimization2

abalasa-amd:
### Description
VitisAI bugfix model clone optimization2


abalasa-amd:
@microsoft-github-policy-service agree company="AMD"

PR #26344, opened 2025-10-17
Enable onnxruntime_autoep_test on non-Windows platforms

edgchen1:
### Description

Enable onnxruntime_autoep_test on non-Windows platforms. Made some fixes as needed to build and run tests.

### Motivation and Context

Verify plugin EP support for Android.

### DIFF
--- a/onnxruntime/test/autoep/library/example_plugin_ep_virt_gpu/ep.cc
+++ b/onnxruntime/test/autoep/library/example_plugin_ep_virt_gpu/ep.cc
@@ -274,7 +274,7 @@ void ORT_API_CALL EpVirtualGpu::ReleaseNodeComputeInfosImpl(OrtEp* this_ptr,
                                                             size_t num_node_compute_infos) noexcept {
   (void)this_ptr;
   for (size_t i = 0; i < num_node_compute_infos; i++) {
-    delete node_compute_infos[i];
+    delete static_cast<AddNodeComputeInfo*>(node_compute_infos[i]);
   }
 }
 


onnxruntime/test/autoep/test_registration.cc
```diff
@@ -35,7 +32,7 @@ TEST(OrtEpLibrary, LoadUnloadPluginLibrary) {
   ASSERT_ORTSTATUS_OK(Ort::GetApi().GetEpDevices(*ort_env, &ep_devices, &num_devices));
   // should be one device for the example EP
   auto num_test_ep_devices = std::count_if(ep_devices, ep_devices + num_devices,
-                                           [&registration_name, &c_api](const OrtEpDevice* device) {
+                                           [&](const OrtEpDevice* device) {
```
edgchen1, line 36: note: this change and other similar changes were made to address a warning from clang like this:

```
onnxruntime/test/autoep/test_registration.cc:35:46: error: lambda capture 'registration_name' is not required to be captured for this use [-Werror,-Wunused-lambda-capture]
   35 |                                            [&registration_name, &c_api](const OrtEpDevice* device) {
      |
```

see: https://stackoverflow.com/questions/52416362/unused-lambda-capture-warning-when-capture-is-actually-used

onnxruntime/test/platform/env_test.cc
```diff
@@ -53,5 +54,25 @@ TEST(PlatformEnvTest, GetErrnoInfo) {
 #pragma warning(pop)
 #endif
 }
+
+TEST(PlatformEnvTest, GetRuntimePath) {
```
edgchen1: it might be better to test a call to `Env::GetRuntimePath()` from a shared library instead of directly from the test program. I'm considering whether it's worth the adding the extra test infrastructure to do so.

onnxruntime/core/platform/posix/env.cc
```diff
@@ -554,6 +555,22 @@ class PosixEnv : public Env {
     return common::Status::OK();
   }
 
+  PathString GetRuntimePath() const override {
+    // Use dladdr() to look up the file that contains an address from this binary.
+    const void* const address_from_this_binary = reinterpret_cast<const void*>(Env::Default);
```
theHamsta, line 560: I settled on `OrtGetApiBase` as a string because most symbols from this static library that isn't exported might not survive LTO. Might be different for `Env::Default`. `OrtGetApiBase` is unfortunately not from this module so Env::Default might be better.

Will this work when doing static linking in the unit tests (so no libonnxruntime.so)?

edgchen1: good to know. it does work when statically linked to the unit test executable. but I've only tested it locally with a debug build of libonnxruntime.so, so it's possible the current implementation is not sufficient for the shared library case. will look into it - maybe it will require looking up an exported symbol.

onnxruntime/test/platform/env_test.cc
edgchen1: added a shared library test

onnxruntime/core/platform/posix/env.cc
edgchen1: @theHamsta I've tested a call to `PosixEnv::GetRuntimePath()` from a shared library and it seems to work as expected. there's a [test](https://github.com/microsoft/onnxruntime/blob/6d25cb11cf94f44b3fdb456c457671dc2a3d03b5/onnxruntime/test/shared_lib/test_runtime_path.cc#L24-L48) that calls a shared library function which returns the value of `GetRuntimePath()`.

do you think that this is enough to show that the LTO behavior is not problematic for `GetRuntimePath()`, or is there a better way to check?

skottmckay (approved):

onnxruntime/core/platform/posix/env.cc
edgchen1: I'll merge this PR first. we can refine it later as needed.

PR #26345, opened 2025-10-17
Save much memory at model loading time by converting weights to OrtValues early

yuslepukhin:
### Description
Converts weights early and revert "Properly remove in-memory references (#25652)"
This reverts commit 3ca49d88017cfbf90f23b7d85426ab2da0a9e1a9 and makes appropriate adjustments for the current state of the code.

This PR is made possible and on the heels of:
https://github.com/microsoft/onnxruntime/pull/26263
https://github.com/microsoft/onnxruntime/pull/25833.

Previous history:
https://github.com/microsoft/onnxruntime/pull/23979
https://github.com/microsoft/onnxruntime/pull/25320
https://github.com/microsoft/onnxruntime/pull/25626
https://github.com/microsoft/onnxruntime/pull/25652

The first change (https://github.com/microsoft/onnxruntime/pull/26263) allows us to convert initializers to OrtValues early and save lots of memory at model loading time.

Specifically, for Phi-4-mini-instruct-INT4 model before and after looks like this:

**Before**
Before change DEBUG 2025-10-16 144819

**After**

After change DEBUG 2025-10-16 144819

The two peaks represent memory usage at optimization time (8.1Gb before) and after weights memory mapping (6.5Gb)
After this change corresponding numbers look 3.5Gb and 4.7Gb respectively.
Most of the savings during optimization phase come from `ConstantFolding` where we are able to reuse the resulting OrtValues directly for the new initializers.

This PR concludes a series of PRs converting initializers to OrtValues.

Memory consumption before the conversion began was 9.3Gb and 6.7Gb respectively. We are saving almost 6Gb during optimization and 2Gb for the steady state. 
 
 
image

The model also loads about 12 seconds faster.

Example of ConstantFolding being one of the top contributors where we duplicate memory for higher peak before Resolve takes care of no longer used initializers.
Sanpshot 3 Peak on ConstantFolding Transpose Optimizer

Snapshot 4 Peak AddInitializer from ConstantFolding

image

### Motivation and Context

Reduce memory usage.

### DIFF
--- a/onnxruntime/core/graph/graph.cc
+++ b/onnxruntime/core/graph/graph.cc
@@ -1232,9 +1232,10 @@ Graph::Graph(const Model& owning_model,
   const auto& model_path = ModelPath();
 
   // If the tensor proto data is large enough, move data from TensorProto to an OrtValue
-  // Add external data reference to TensorProto that points to an OrtValue.
+  // - Add external data reference to TensorProto that points to an OrtValue.
   // This lambda should not be used on initializers that already have external data reference.
-  auto put_weight_maybe_in_ortvalue = [this, &model_path](ONNX_NAMESPACE::TensorProto& tensor_proto) {
+  // Otherwise, this function does nothing.
+  auto put_large_tensor_in_ort_value = [this, &model_path](ONNX_NAMESPACE::TensorProto& tensor_proto) {
     size_t size_in_bytes = 0;
     ORT_THROW_IF_ERROR(utils::GetSizeInBytesFromTensorProto<0>(tensor_proto, &size_in_bytes));
     if (size_in_bytes > utils::kSmallTensorExternalDataThreshold) {
@@ -1271,7 +1272,7 @@ Graph::Graph(const Model& owning_model,
       }
     }
 
-    put_weight_maybe_in_ortvalue(*tensor);
+    put_large_tensor_in_ort_value(*tensor);
 
     // Ensure initializers are also graph inputs.
     if (ir_version_ < 4) {
@@ -1355,11 +1356,14 @@ Graph::Graph(const Model& owning_model,
     // or during session state finalization.
     // If data is already in memory, do nothing.
     if (!utils::HasExternalData(tensor)) {
+      // sparse_tensor_names_ contain references to strings to save memory
+      // in case we replace the tensor_proto, we want to make sure we remove
+      // the old reference first, and then add a new one.
       const bool is_sparse = sparse_tensor_names_.count(tensor.name());
       if (is_sparse) {
         sparse_tensor_names_.erase(tensor.name());
       }
-      put_weight_maybe_in_ortvalue(tensor);
+      put_large_tensor_in_ort_value(tensor);
       if (is_sparse) {
         sparse_tensor_names_.emplace(tensor.name());
       }


onnxruntime/core/framework/session_state_utils.cc
```diff
@@ -346,6 +352,13 @@ common::Status SaveInitializedTensors(
                        << i.second << " bytes for " << i.first.ToString() << std::endl;
   }
 
+  // ??? Should we ignore this session option if the EP is explicitly providing the read only allocator?
+  // bool have_readonly_initializer_allocator = alloc->Info().alloc_type == OrtReadOnlyAllocator;
+  // This option also means to ignore arena if present and use Reserve().
+  const bool use_device_allocator_for_initializers =
```
skottmckay, line 358: nit: why did this move up to a wider scope?

onnxruntime/core/graph/graph.cc
```diff
@@ -1231,6 +1231,26 @@ Graph::Graph(const Model& owning_model,
   ArgNameToTypeMap name_to_type_map;
   const auto& model_path = ModelPath();
 
+  // If the tensor proto data is large enough, externalize it and replace with a tensor_proto
+  // with external data reference pointing to an OrtValue, otherwise do nothing.
+  auto put_data_maybe_in_memory = [this, &model_path](ONNX_NAMESPACE::TensorProto& tensor_proto) {
```
skottmckay: nit: can you clarify what this is doing? it seems to only process tensors `> utils::kSmallTensorExternalDataThreshold` (the opposite of what I would expect to be 'put in memory' and I don't see anything in `utils::TensorProtoToOrtValue` that moves something to being external so the 'externalize' aspect is unclear.

onnxruntime/core/optimizer/attention_fusion.cc
```diff
@@ -111,7 +111,7 @@ static NodeArg& MergeQkvWeights(Graph& graph, int64_t hidden_size,
     utils::SetRawDataInTensorProto(initializer, result.data(), gsl::narrow<size_t>(element_count) * sizeof(MLFloat16));
   }
 
-  return graph_utils::AddInitializer(graph, initializer);
+  return graph_utils::AddInitializerWithExternalData(graph, initializer);
```
skottmckay: what ensures that the data is external given `graph_utils::AddInitializerWithExternalData` has

`  ORT_ENFORCE(!has_external_data, "Expecting an initializer that contains data inline");`

onnxruntime/test/framework/session_state_test.cc
```diff
@@ -392,20 +396,25 @@ TEST(SessionStateTest, TestInitializerMemoryAllocatedUsingNonArenaMemory) {
...
-    ASSERT_TRUE(status.IsOK()) << status;
+    auto tmp_cpu_execution_provider = DefaultCudaExecutionProvider();
+    tmp_cpu_execution_provider->SetLogger(&default_logger);
+    ASSERT_STATUS_OK(dtm.RegisterDataTransfer(tmp_cpu_execution_provider->GetDataTransfer()));
+    ASSERT_STATUS_OK(execution_providers.Add(kCudaExecutionProvider, std::move(tmp_cpu_execution_provider)));
+    // Make sure CPU allocator is registered
+    auto cpu_execution_provider = DefaultCpuExecutionProvider();
+    cpu_execution_provider->SetLogger(&default_logger);
+    ASSERT_STATUS_OK(dtm.RegisterDataTransfer(cpu_execution_provider->GetDataTransfer()));
+    ASSERT_STATUS_OK(execution_providers.Add(kCpuExecutionProvider, std::move(cpu_execution_provider)));
```
skottmckay: nit: can this setup be in a helper function or does it differ from TestInitializerMemoryAllocatedUsingNonArenaMemory?

onnxruntime/test/providers/tensorrt/tensorrt_basic_test.cc
```diff
@@ -299,7 +299,9 @@ void RunWithOneSessionMultiThreadsInference(PathString model_name, std::string s
   ASSERT_TRUE(HasCacheFileWithPrefix(params.trt_engine_cache_prefix));
 }
 
-TEST(TensorrtExecutionProviderTest, SessionCreationWithMultiThreadsAndInferenceWithMultiThreads) {
+// The test is disabled due to the issue described at
+// https://github.com/microsoft/onnxruntime/issues/26366
+TEST(TensorrtExecutionProviderTest, DISABLED_SessionCreationWithMultiThreadsAndInferenceWithMultiThreads) {
```
skottmckay, line 304: is there a plan to fix this? is it a test issue or does ORT need updates?

onnxruntime/core/optimizer/attention_fusion.cc
tianleiwu: I suggest to change graph_utils::AddInitializer(graph, initializer) to call graph_utils::AddInitializerWithExternalData(graph, initializer) internally when there is external data. In this way, the caller can always call graph_utils::AddInitializer.

onnxruntime/core/framework/session_state_utils.cc
yuslepukhin: It did not make sense for it to be in the loop

onnxruntime/core/optimizer/attention_fusion.cc
yuslepukhin: This initializer has just been created with RawData set. It has no external data. The utility function externalizes the data into OrtValue. Perhaps, the name of the utility function could be changed to `ExternalizeDataAndAddInitializer()`?

onnxruntime/test/providers/tensorrt/tensorrt_basic_test.cc
yuslepukhin: This is a test issue, and the owners are aware of it.

onnxruntime/core/optimizer/attention_fusion.cc
yuslepukhin: Changing to `AddInitializerWithOrtValue`

onnxruntime/core/graph/graph_utils.cc
```diff
@@ -285,19 +285,19 @@ NodeArg& AddInitializer(Graph& graph, const ONNX_NAMESPACE::TensorProto& new_ini
   return GetOrCreateNodeArg(graph, new_initializer);
 }
 
-NodeArg& AddInitializerWithExternalData(Graph& graph, const ONNX_NAMESPACE::TensorProto& new_initializer) {
+NodeArg& AddInitializerWithOrtValue(Graph& graph, const ONNX_NAMESPACE::TensorProto& new_initializer) {
   const bool has_external_data = utils::HasExternalData(new_initializer);
   ORT_ENFORCE(!has_external_data, "Expecting an initializer that contains data inline");
```
tianleiwu, line 290: Could we add a fallback when there is external data? Otherwise, either we add a central step to convert external data to raw data during model loading, or each caller has to check/handle the external data before calling this function.

onnxruntime/core/graph/graph.cc
```diff
@@ -1231,6 +1231,27 @@ Graph::Graph(const Model& owning_model,
   ArgNameToTypeMap name_to_type_map;
   const auto& model_path = ModelPath();
 
+  // If the tensor proto data is large enough, move data from TensorProto to an OrtValue
+  // Add external data reference to TensorProto that points to an OrtValue.
+  // This lambda should not be used on initializers that already have external data reference.
```
tianleiwu, line 1236: Regarding the comment: where do we handle the initializers with external data reference? 

For external data, I assume that we can use OrtValue to point to a buffer that is memory mapped to the external data file.

onnxruntime/core/graph/graph.cc
```diff
@@ -1250,6 +1271,8 @@ Graph::Graph(const Model& owning_model,
       }
     }
 
+    put_weight_maybe_in_ortvalue(*tensor);
```
tianleiwu: Check there is no external data before calling put_weight_maybe_in_ortvalue?

onnxruntime/core/graph/graph.cc
```diff
@@ -1231,6 +1231,27 @@ Graph::Graph(const Model& owning_model,
   ArgNameToTypeMap name_to_type_map;
   const auto& model_path = ModelPath();
 
+  // If the tensor proto data is large enough, move data from TensorProto to an OrtValue
+  // Add external data reference to TensorProto that points to an OrtValue.
+  // This lambda should not be used on initializers that already have external data reference.
+  auto put_weight_maybe_in_ortvalue = [this, &model_path](ONNX_NAMESPACE::TensorProto& tensor_proto) {
```
tianleiwu: nit: suggest to change the name to like put_large_tensor_in_ort_value.

onnxruntime/core/graph/graph.cc
```diff
@@ -1326,7 +1349,22 @@ Graph::Graph(const Model& owning_model,
...
   // Copy initial tensors to a map.
-  for (auto& tensor : graph_proto_->initializer()) {
+  for (int i = 0, lim = graph_proto_->initializer_size(); i < lim; ++i) {
+    auto& tensor = *graph_proto_->mutable_initializer(i);
+    // If data is on disk, it will be loaded either by optimizers
+    // or during session state finalization.
+    // If data is already in memory, do nothing.
+    if (!utils::HasExternalData(tensor)) {
+      const bool is_sparse = sparse_tensor_names_.count(tensor.name());
+      if (is_sparse) {
```
tianleiwu, line 1363: Add some comments about the reason of special handling of sparse tensor names here.

onnxruntime/core/graph/graph_utils.cc
yuslepukhin: No fallback. This interface is specifically designed take a tensor proto with inline data.
Other overloads take both tensorproto and OrtValue which would be the case when tensorproto has data in memory. These do not deal with external data on disk.

onnxruntime/core/graph/graph.cc
yuslepukhin: When we load protobuf, we ignore initializers with external data on disk. They are loaded on demand during optimization phase using `Initializer` class that can handle all kinds of initializers.
Initializers produced by transformers do not go back on disk, they exist only in memory.
During session state finalization external data is memory mapped from disk.

onnxruntime/core/graph/graph.cc
yuslepukhin: There is no need. This initializer is a newly created one from a ConstantNode.

tianleiwu (approved):

Issue #26189, opened 2025-09-28
[Build] [CUDA] [FlashAttention] Compilation fails with “template‑id not allowed for constructor in C++20” (cutlass/exmy_base.h)


ProPatte:
### Describe the issue

While building ONNX Runtime with CUDA support on Linux, the compilation stops at flash_api.cc at 51 % with errors from CUTLASS headers indicating that a template‑id is used for a constructor, which is disallowed in C++20. All warnings are treated as errors (-Werror).

Steps to reproduce:
1. git clone --recursive https://github.com/microsoft/onnxruntime
2. cd onnxruntime
3. ./build.sh   --use_cuda   --cudnn_home /usr   --cuda_home /opt/cuda   --cmake_extra_defines     CMAKE_CUDA_ARCHITECTURES=120

_/usr and /opt/cuda are home dirs of cudnn and cuda on Arch._

### Target platform

Arch Linux (6.16.8-arch3-1)

### Build script

> #!/bin/bash
> #Copyright (c) Microsoft Corporation. All rights reserved.
> #Licensed under the MIT License.
> 
> #Get directory this script is in
> DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
> OS=$(uname -s)
> 
> if [ "$OS" = "Darwin" ]; then
>     DIR_OS="MacOS"
> else
>     DIR_OS="Linux"
> fi
> 
> if [[ "$*" == *"--ios"* ]]; then
>     DIR_OS="iOS"
> elif [[ "$*" == *"--android"* ]]; then
>     DIR_OS="Android"
> fi
> 
> python3 $DIR/tools/ci_build/build.py --build_dir $DIR/build/$DIR_OS "$@"

### Error / output

> [ 51%] Building CXX object CMakeFiles/onnxruntime_providers_cuda.dir/home/patte/onnxruntime/onnxruntime/contrib_ops/cuda/bert/flash_attention/flash_api.cc.o
> In file included from /home/patte/onnxruntime/build/Linux/Debug/_deps/cutlass-src/include/cutlass/float8.h:99,
>                  from /home/patte/onnxruntime/build/Linux/Debug/_deps/cutlass-src/include/cutlass/half.h:66,
>                  from /home/patte/onnxruntime/build/Linux/Debug/_deps/cutlass-src/include/cutlass/numeric_types.h:41,
[...]

PR #26369, opened 2025-10-20
[CUDA] patch cutlass for C++20

tianleiwu:
Fix build error with C++ 20 (See issue https://github.com/microsoft/onnxruntime/issues/26189)


tianleiwu:
Included in https://github.com/microsoft/onnxruntime/pull/26349.

PR #26372, opened 2025-10-21
use std::move(model_proto) to avoid extra copy

mingyueliuh:
Previously, model_proto was passed by name, which triggered a copy constructor call instead of move construction. Using std::move(model_proto) ensures that the object is constructed via move semantics, reducing unnecessary memory allocation and copy overhead.


mingyueliuh:
hi @yuslepukhin @skottmckay please take a look when you are available. Thanks.

mingyueliuh:
@tianleiwu  please help review , Thanks

PR #26373, opened 2025-10-21
[EP API] Check for out-of-range node in EpGraph::IndexToEpNodeMap::GetEpNode

minfhong-qti:
### Description

Check whether given `node_index` is within the target range of nodes during `EpGraph::IndexToEpNodeMap::GetEpNode`. Since EpGraph may be created from a subset of nodes, it should be checked that queried node is inside current EpGraph.

### Motivation and Context

While operating on partitioned graphs, SegFault occurrs if tries to acquire a node that is partitioned to another EpGraph (i.e., different EP).

### DIFF
--- a/onnxruntime/core/graph/ep_api_types.cc
+++ b/onnxruntime/core/graph/ep_api_types.cc
@@ -543,6 +543,10 @@ void EpGraph::IndexToEpNodeMap::Resize(NodeIndex min_node_index, NodeIndex max_n
 }
 
 EpNode* EpGraph::IndexToEpNodeMap::GetEpNode(NodeIndex node_index) const {
+  if (node_index < min_node_index_ || node_index >= min_node_index_ + nodes_.size()) {
+    return nullptr;
+  }
+
   size_t i = node_index - min_node_index_;
   assert(i < nodes_.size());
   return nodes_[i];


minfhong-qti:
Hi @edgchen1,
Not sure whether I'm looking for the correct one for reviewing this. Could you kindly check whether the issue fixed in this PR is valid? Or help redirect to other reviewers.
Thanks!

edgchen1:
> Hi @edgchen1, Not sure whether I'm looking for the correct one for reviewing this. Could you kindly check whether the issue fixed in this PR is valid? Or help redirect to other reviewers. Thanks!

@adrianlizarraga mind taking a look?

onnxruntime/core/graph/ep_api_types.cc
```diff
@@ -543,6 +543,10 @@ void EpGraph::IndexToEpNodeMap::Resize(NodeIndex min_node_index, NodeIndex max_n
 }
 
 EpNode* EpGraph::IndexToEpNodeMap::GetEpNode(NodeIndex node_index) const {
+  if (node_index < min_node_index_ || node_index >= min_node_index_ + nodes_.size()) {
```
adrianlizarraga, line 546: qq: Is this necessary due to the `Graph_GetGraphView` API?

adrianlizarraga (commented):
Looks good to me. Should merge in the latest main branch though.

minfhong-qti:
> Looks good to me. Should merge in the latest main branch though.

Yes, it's already been fixed in a0d4439. Will close this PR.

onnxruntime/core/graph/ep_api_types.cc
minfhong-qti: Not actually. I discovered this bug when enabling ABI version of QNN-EP which was before `Graph_GetGraphView` API introduced.

PR #26376, opened 2025-10-21
Add MulAddFusion in L1 Transformer

qti-hungjuiw:
### Description

- Added a core optimizer, MulAddFusion, to the L1 Transformer. This optimization fuses Mul and Add operations into a single BatchNormalization node when they are mathematically equivalent.
- Included corresponding unit tests to validate the optimization logic.

### Motivation and Context

- By identifying and fusing patterns that are functionally equivalent to BatchNormalization, we streamline the graph and potentially enhance performance.

### DIFF
--- a/onnxruntime/core/optimizer/mul_add_fusion.cc
+++ b/onnxruntime/core/optimizer/mul_add_fusion.cc
@@ -17,13 +17,18 @@ namespace onnxruntime {
 
 bool IsPatternBatchnorm(const NodeArg* inp,
                         const ONNX_NAMESPACE::TensorProto* scale,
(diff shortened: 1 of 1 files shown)


qti-hungjuiw:
Hi @tianleiwu, @edgchen1, I've resolved the crash issue on Roberta. Could you please help retrigger the pipeline?

edgchen1:
> By identifying and fusing patterns that are functionally equivalent to BatchNormalization, we streamline the graph and potentially enhance performance.

it's an interesting fusion. do you have a particular scenario where this improves the performance, and if so, can you share some details and measurements?

PR #26392, opened 2025-10-23
webgpu / nbitmm support for bias and weight_index

guschmue:
add support for bias and weight_index, move subgroup_matrix_matmul_nbits to template and make program callable from other ops.

### DIFF
--- a/onnxruntime/contrib_ops/webgpu/quantization/dp4a_matmul.wgsl.template
+++ b/onnxruntime/contrib_ops/webgpu/quantization/dp4a_matmul.wgsl.template
@@ -140,7 +140,6 @@ fn loadSHMA(a_global_base:u32, kidx_v:u32, row: u32, col: u32)
         const b_weight_offset : u32 = 0;
         let b_value = b.getByOffset(b_global * uniforms.K16 + kidx_v + col);
 #endif
-        let b_weight_offset = uniforms.weight_idx * uniforms.N * uniforms.K16;
         tile_B[col][row] = DequantizedFrom2BitsTo8Bits(b_value);
         let block_idx = kidx_v/(block_size/16);
         let b_scale_offset = uniforms.weight_idx * uniforms.N * (uniforms.K/block_size);


onnxruntime/contrib_ops/webgpu/quantization/matmul_nbits_wide_tile.wgsl.template
```diff
@@ -190,7 +194,15 @@ $MAIN {
   }
 
   // Write the results.
+  #if has_bias
+    let b_bias_offset = uniforms.weight_idx * uniforms.N;
+    let bias_value = bias[b_bias_offset + col + local_idx];
```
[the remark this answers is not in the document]
guschmue: hm, that comment is incorrect, the assignment is already outside of the loop

onnxruntime/contrib_ops/webgpu/quantization/dp4a_matmul_common.wgsl.template
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 #param n_bits
+#param has_bias
```
qjia7: Is this needed to avoid the template's grammar issue?

onnxruntime/contrib_ops/webgpu/quantization/dp4a_matmul_small_m.wgsl.template
```diff
@@ -73,7 +78,8 @@ $MAIN {
 #endif
 #if single_scale_weights
     let zero = mm_read_zero(0, 0, uniforms.N, uniforms.zero_blocks_per_col);
-    let own_scale_b = scales_b.getByOffset(0);
+    let b_scale_offset = uniforms.weight_idx * uniforms.N * (uniforms.K / uniforms.block_size);
+    let own_scale_b = scales_b.getByOffset(b_scale_offset);
```
qjia7: I think for single_scale_weights, you can directly use `let own_scale_b = scales_b.getByOffset(uniforms.weight_idx);`.

onnxruntime/contrib_ops/webgpu/quantization/matmul_nbits.cc
```diff
@@ -124,23 +131,76 @@ Status MatMulNBits::ComputeInternal(onnxruntime::webgpu::ComputeContext& context
...
+ * @param scales         Pointer to the tensor containing scaling factors for quantization.
+ * @param zero_points    Pointer to the zero-point tensor for quantization; must be of type uint8 if provided.
+ * @param bias           Pointer to the bias tensor; optional.
+ * @param K_op           The K dimension of the operation (number of columns in 'a' and rows in 'b' before quantization).
+ * @param N_op           The N dimension of the operation (number of columns in 'b').
+ * @param block_size_op  The block size used for quantization partitioning.
+ * @param accuracy_level Accuracy level influencing the choice of optimized kernel.
+ * @param nbits          Number of bits used for quantization.
+ * @param context        Compute context for WebGPU, providing device-specific information and execution facilities.
+ * @param y              Pointer to the output tensor that will hold the result.
```
qjia7, line 170: Please add description for param `weight_index` which is used to specify which index of batch in b to participant into the calculation.

onnxruntime/contrib_ops/webgpu/quantization/matmul_nbits.cc
```diff
@@ -124,23 +131,76 @@ Status MatMulNBits::ComputeInternal(onnxruntime::webgpu::ComputeContext& context
...
+ *
+ * This function computes the matrix multiplication of the quantized tensor inputs with multiple
+ * optional optimizations tailored to the GPU backend. Depending on the provided parameters and GPU
+ * capabilities, it selects one of several optimized kernels (such as subgroup matrix multiplication,
+ * DP4A, wide tile programs, or the default matmul program) to perform the computation.
+ * It can be called by the MatMulNBits operator or directly for custom scenarios like QMoe.
+ *
+ * @param a              Pointer to the left-hand side (activation) tensor.
+ * @param b              Pointer to the quantized weight tensor.
+ * @param scales         Pointer to the tensor containing scaling factors for quantization.
```
qjia7, line 158: I assume the b's shape will be  (weight_batch, N, k_blocks, blob_size) instead of  (N, k_blocks, blob_size). For `MatMulNBits` operator, weight_batch is 1. And for custom scenarios like QMoe, weight_batch is num_experts. So it will be good to have some description about this parameter in case others modify this file in future.
And for scales, similarly, the shape of it is (weight_batch, N) instead of N.

onnxruntime/contrib_ops/webgpu/quantization/matmul_nbits.cc
```diff
@@ -124,23 +131,76 @@ Status MatMulNBits::ComputeInternal(onnxruntime::webgpu::ComputeContext& context
...
+ * @param zero_points    Pointer to the zero-point tensor for quantization; must be of type uint8 if provided.
```
qjia7, line 160: For zero point, currently, `weight_index` is not used since `QMoe` is using symmetric quantization. Please add comment here that `weight_batch` is not supported in zero_points.

onnxruntime/contrib_ops/webgpu/quantization/matmul_nbits.cc
```diff
@@ -204,6 +264,9 @@ Status MatMulNBits::ComputeInternal(onnxruntime::webgpu::ComputeContext& context
                         {ceil_div(zero_points->Shape().Size(), static_cast<int64_t>(4))},
                         4});
     }
+    if (has_bias) {
+      program.AddInput({bias, ProgramTensorMetadataDependency::TypeAndRank});
```
qjia7:
```suggestion
      program.AddInput({bias, ProgramTensorMetadataDependency::None});
```

onnxruntime/contrib_ops/webgpu/quantization/matmul_nbits.cc
```diff
@@ -227,19 +291,32 @@ Status MatMulNBits::ComputeInternal(onnxruntime::webgpu::ComputeContext& context
...
+                            {zero_blocks_per_col},
+                            {num_N_tile},
+                            {batch_count},
+                            {weight_index}})
+      .CacheHint(nbits, has_zero_points, single_scale_weights, has_bias);
   if (has_zero_points) {
     program.AddInput({zero_points, ProgramTensorMetadataDependency::None, {(zero_points->Shape().Size() + 3) / 4}, 4});
   }
+  if (has_bias) {
+    program.AddInput({bias, ProgramTensorMetadataDependency::TypeAndRank});
```
qjia7:
```suggestion
    program.AddInput({bias, ProgramTensorMetadataDependency::None});
```

onnxruntime/contrib_ops/webgpu/quantization/matmul_nbits.wgsl.template
```diff
@@ -60,10 +65,10 @@ $MAIN {
       {
 #if !single_scale_weights
         let block_idx = (kidx + idx * elements_in_value_b) / uniforms.block_size;
-        let scale_b = scales_b.getByOffset(b_global * uniforms.blocks_per_col + block_idx);
+        let scale_b = scales_b.getByOffset(b_global * uniforms.blocks_per_col + block_idx + b_scale_offset);
```
qjia7, line 79: Please also update the scale_b in `single_scale_weights` path in line 48 (`let scale_b = scales_b.getByOffset(uniforms.weight_idx);`).

onnxruntime/contrib_ops/webgpu/quantization/dp4a_matmul.wgsl.template
```diff
@@ -76,14 +77,16 @@ fn loadSHMA(a_global_base:u32, kidx_v:u32, row: u32, col: u32)
             return;
         }
 
-        let b_value = b.getByOffset(b_global*uniforms.K16+kidx_v+col);
+        let b_weight_offset = uniforms.weight_idx * uniforms.N * uniforms.K16;
```
qjia7, line 81: nit: Since `b_weight_offset` and `b_scale_offset` are constant value, will it be better to calculate them in cpu and write into uniform? In shader, we can always read them from uniform.

onnxruntime/test/contrib_ops/matmul_4bits_test.cc
```diff
@@ -342,6 +342,14 @@ void TestMatMulNBitsTyped(std::optional<float> abs_error = std::nullopt,
     RunTest<AType>(opts);
   }
 #endif  // !defined(USE_DML) && !defined(USE_WEBGPU)
+#if defined(USE_WEBGPU)
```
qjia7, line 345: Also enable the bias test for matmul_8bits_tests.cc?

guschmue:
> In current change, it seems that ApplyMatMulNBits calculate a with one weight index of b. For QMoE case, it only compute one expert. I remember you said you need 4 selected experts. So will ApplyMatMulNBits be called 4 times to get the up projection results? Why not directly generate the result by calling ApplyMatMulNBits once?
Earlier on I has a QMoE implementation that does 4 experts in 1 shot but in that case I need to go token by token and the prefill would get pretty costly.
The current implementation (https://github.com/microsoft/onnxruntime/blob/3c65aaa50f8ce8a7d73e502549371e6433282a49/onnxruntime/contrib_ops/webgpu/moe/qmoe.cc#L198)
does it different. I looks at all tokens and assigns them to experts, than walks over the experts and runs them 1 by one if there are tokens assigned to them. I think this is more in line with other implementations and should be better for prefill performance. 
If you have say 1000 tokens it is most likely we need to run all experts, for generation we'd run 4. 
Also helps with memory usage.

onnxruntime/contrib_ops/webgpu/quantization/dp4a_matmul.wgsl.template
guschmue: could do that but I intentionally did not because there are 32 experts so we'd need to compile 32 shaders.
But I was thinking for weight_idx == 0 I could do a #if in the template and 'const b_weight_offset = 0' so everything not QMoE would benefit of const and for QMoE we'd need to compile 2 shaders.

onnxruntime/contrib_ops/webgpu/quantization/dp4a_matmul_common.wgsl.template
guschmue: compile is happy, let me take this out.

onnxruntime/contrib_ops/webgpu/quantization/dp4a_matmul_small_m.wgsl.template
guschmue: done

onnxruntime/contrib_ops/webgpu/quantization/matmul_nbits.wgsl.template
guschmue: done

onnxruntime/contrib_ops/webgpu/quantization/matmul_nbits.cc
guschmue: done

onnxruntime/contrib_ops/webgpu/quantization/matmul_nbits.cc
guschmue: done.

onnxruntime/contrib_ops/webgpu/quantization/matmul_nbits.cc
guschmue: done

onnxruntime/contrib_ops/webgpu/quantization/matmul_nbits.cc
guschmue: done

onnxruntime/contrib_ops/webgpu/quantization/dp4a_matmul.wgsl.template
guschmue: I made a change to use const for weight_idx related offset if weight_idx == 0. So only QMoE takes a tiny hit for the  weight_idx.

onnxruntime/test/contrib_ops/matmul_4bits_test.cc
guschmue: done

onnxruntime/contrib_ops/webgpu/quantization/dp4a_matmul.wgsl.template
```diff
@@ -119,10 +133,18 @@ fn loadSHMA(a_global_base:u32, kidx_v:u32, row: u32, col: u32)
...
         }
-        let b_value = b.getByOffset(b_global*uniforms.K16+kidx_v+col);
+#if has_weight_idx
+        let b_weight_offset = uniforms.weight_idx * uniforms.N * uniforms.K16;
+        let b_value = b.getByOffset(b_weight_offset + b_global * uniforms.K16 + kidx_v + col);
+#else
+        const b_weight_offset : u32 = 0;
+        let b_value = b.getByOffset(b_global * uniforms.K16 + kidx_v + col);
+#endif
+        let b_weight_offset = uniforms.weight_idx * uniforms.N * uniforms.K16;
```
qjia7: nit: Delete line 143?

onnxruntime/contrib_ops/webgpu/quantization/dp4a_matmul.wgsl.template
```diff
@@ -411,6 +430,7 @@ $MAIN {
  #endif
 #else
  #if has_bias
+        // TODO: wanted to use vec4 for bias but for some reason that fails ut. Later.
```
qjia7, line 432: Using vec4 for bias, you need to make sure N % 4 == 0 or it will be very complicated to re-arrange to get the correct vec4 data.

onnxruntime/contrib_ops/webgpu/quantization/dp4a_matmul.wgsl.template
guschmue: done

onnxruntime/contrib_ops/webgpu/quantization/dp4a_matmul.wgsl.template
guschmue: we can sit it out

PR #26394, opened 2025-10-23
[Web] Add wasm and mjs files to package.json exports to improve bundler support

KingSora:
### Description
Currently the files which are needed for  wasm aren't exported from the package at all.
The files in question are:
```
ort-wasm-simd-threaded.wasm
ort-wasm-simd-threaded.jsep.wasm
ort-wasm-simd-threaded.asyncify.wasm
ort-wasm-simd-threaded.mjs
ort-wasm-simd-threaded.jsep.mjs
ort-wasm-simd-threaded.asyncify.mjs
```

This PR changes that and adds them to `exports` field in the `package.json`.

### Motivation and Context

Bundlers like `webpack` use the `copyPlugin` to move those files into the `public` directory, so the files can be accessed by a stable url. However more advanced and "state of the art" bundlers like `vite` are able to [import asset urls directly](https://vite.dev/guide/assets.html#explicit-url-imports). Vite takes the asset, moves it to to public assets folder (possibily renames the asset and adds a hash etc.). The imported value then is the bundled assets final url.

Those urls can then be used in the `env.wasm.wasmPaths` directly. In vites case the full code example is:
```js
import wasmUrl from 'onnxruntime-web/ort-wasm-simd-threaded.wasm?url';
import mjsUrl from 'onnxruntime-web/ort-wasm-simd-threaded.mjs?url';

env.wasm.wasmPaths = {
  wasm: wasmUrl,
  mjs: mjsUrl,
};
```

With those added exports we can leverage more of the bundlers capabilities and in vites case there isn't any need to add any additional configs. It would just work.

When importing we also get proper suggestions:
imports

----

I would like additional tests to ensure that the exports are available, but I couldn't make the `e2e` tests work on my system. I would appreciate some guidance on that topic.

### DIFF
--- a/js/web/test/e2e/package.json
+++ b/js/web/test/e2e/package.json
@@ -20,6 +20,7 @@
     "rollup": "^4.13.2",
     "rollup-plugin-copy": "^3.5.0",
     "tree-kill": "^1.2.2",
+    "vite": "^7.1.12",
     "webpack-cli": "^5.1.4"
   },
   "scripts": {
@@ -32,7 +33,10 @@
     "build:p:esmjs": "parcel build --no-autoinstall --target esm && node ./bundler.esm.postprocess.js ./dist/parcel_esm_js/main.js",
     "build:p:umdjs": "parcel build --no-autoinstall --target umd",
     "build:p": "npm run build:p:esmjs && npm run build:p:umdjs",
-    "build": "npm run build:w && npm run build:r && npm run build:p"
+    "build:v:esmjs": "vite build -c vite.config.esm-js.js",
+    "build:v:umdjs": "vite build -c vite.config.umd-js.js",
+    "build:v": "npm run build:v:esmjs && npm run build:v:umdjs",
+    "build": "npm run build:v && npm run build:w && npm run build:r && npm run build:p"
   },
   "@parcel/resolver-default": {
     "packageExports": true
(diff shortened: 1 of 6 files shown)


KingSora:
@fs-eire when I'm trying to run `npm run test:e2e` in the `js/web` directory I'm getting the following error: `cannot find exactly single package for onnxruntime-web`. This error prevents me to write an meaningful test.

fs-eire:
> @fs-eire when I'm trying to run `npm run test:e2e` in the `js/web` directory I'm getting the following error: `cannot find exactly single package for onnxruntime-web`. This error prevents me to write an meaningful test.

Looks like the CI passes, so this may indicate that you are not running `npm run test:e2e` in the expected order. you need run `npm pack` in /js/common/ and /js/web/ to create the 2 .tgz files and then run the `npm run test:e2e`

js/web/package.json
```diff
@@ -115,7 +115,13 @@
         "default": "./dist/ort.jspi.bundle.min.mjs"
       },
       "require": "./dist/ort.jspi.min.js"
-    }
+    },
+    "./ort-wasm-simd-threaded.wasm": "./dist/ort-wasm-simd-threaded.wasm",
+    "./ort-wasm-simd-threaded.jsep.wasm": "./dist/ort-wasm-simd-threaded.jsep.wasm",
+    "./ort-wasm-simd-threaded.asyncify.wasm": "./dist/ort-wasm-simd-threaded.asyncify.wasm",
```
fs-eire, line 122: could you please also add .jspi

KingSora: done :)

KingSora:
@fs-eire Thanks for the info, `npm pack` was missing - might be worth adding it to the docs.

I've added some e2e tests which test the said exports with `vite` and its `?url` import feature.

KingSora:

fs-eire (approved):

PR #26395, opened 2025-10-23
Fixing dump model ops feature on Windows for MGX EP

urpetkov-amd:
Fixing dump model ops feature for MIGraphX EP on Windows. The feature wasn't functional because of saving format rules on Windows which are opposite from Linux.

Current state of the feature gives us opportunity to generate and save onnx model after subgraph optimizations before compiling it. On this way we can look how model graph looks like after optimizations and we can use the optimized model.

### DIFF
--- a/onnxruntime/core/providers/migraphx/migraphx_execution_provider.cc
+++ b/onnxruntime/core/providers/migraphx/migraphx_execution_provider.cc
@@ -1341,11 +1341,11 @@ Status MIGraphXExecutionProvider::Compile(const std::vector<FusedNodeAndGraph>& 
       std::string onnx_name = fused_node.Name() + ".onnx";
       std::ofstream ofs(onnx_name, std::ios::binary);
       if (!ofs.is_open()) {
-        ORT_THROW("Failed to open file to dump ONNX model: " + model_name);
+        ORT_THROW("Failed to open file to dump ONNX model: " + onnx_name);
       }
       ofs.write(onnx_string_buffer.data(), onnx_string_buffer.size());
       ofs.close();
-      LOGS_DEFAULT(INFO) << "ONNX model dumped to " << model_name;
+      LOGS_DEFAULT(INFO) << "ONNX model dumped to " << onnx_name;
     }
 
     std::vector<std::string> input_names, output_names;


urpetkov-amd:
Thanks.
Okay, should we merge this and do that in another PR or add to this one? @snnn

urpetkov-amd:
@snnn 

I added changes for this. I added check if file is open, if it's not successfully opened then we can throw an error, otherwise print a log. In this way check for closing is not necessary, it will be closed in the ofstream destructor after leaving the block.

onnxruntime/core/providers/migraphx/migraphx_execution_provider.cc
```diff
@@ -1335,9 +1339,13 @@ Status MIGraphXExecutionProvider::Compile(const std::vector<FusedNodeAndGraph>&
 
     if (dump_model_ops_) {
       std::string onnx_name = fused_node.Name() + ".onnx";
-      std::ofstream ofs(onnx_name);
+      std::ofstream ofs(onnx_name, std::ios::binary);
```
snnn, line 1342: There seems to be a small typo here. The variable should probably be `onnx_name` instead of `model_name` to match the variable defined a few lines above.

onnxruntime/core/providers/migraphx/migraphx_execution_provider.cc
```diff
@@ -1335,9 +1339,13 @@ Status MIGraphXExecutionProvider::Compile(const std::vector<FusedNodeAndGraph>&
...
+      if (!ofs.is_open()) {
+        ORT_THROW("Failed to open file to dump ONNX model: " + model_name);
+      }
```
snnn, line 1345: Similar to the `ORT_THROW` above, this should likely use `onnx_name` to log the correct filename.

onnxruntime/core/providers/migraphx/migraphx_execution_provider.cc
```diff
@@ -1335,9 +1339,13 @@ Status MIGraphXExecutionProvider::Compile(const std::vector<FusedNodeAndGraph>&
 
     if (dump_model_ops_) {
       std::string onnx_name = fused_node.Name() + ".onnx";
-      std::ofstream ofs(onnx_name);
+      std::ofstream ofs(onnx_name, std::ios::binary);
+      if (!ofs.is_open()) {
+        ORT_THROW("Failed to open file to dump ONNX model: " + model_name);
```
snnn: This is a good fix. For even more robust error handling, you might consider checking the stream's state after closing the file (e.g., `if (!ofs)`). This would catch potential errors during the final flush to disk, which could result in a corrupt or incomplete file. However, I see that not checking the return of `close()` is a common pattern in this codebase, so this is more of a suggestion for hardening the code rather than a required change for this PR.

snnn:
This PR looks good and fixes the issue on Windows.

From a security perspective, there's a potential path traversal vulnerability here. The filenames for the dumped models are created from the graph/node names in the ONNX model without sanitization. An attacker could craft a model with a name like `../../etc/passwd` to write files to arbitrary locations.

This feature is for debugging, so the risk is lower, but it's still a potential issue if used in a shared environment.

I'd recommend sanitizing the names before using them as filenames. A simple sanitization function could be something like this:

```cpp
#include <string>
#include <string_view>

std::string SanitizeFileName(std::string_view name) {
    std::string newName(name);
    for (char& c : newName) {
        switch (c) {
            case '\\':
            case '/':
            case '\"':
            case '|':
            case '<':
            case '>':
            case ':':
            case '?':
            case '*':
                c = '_';
                break;
        }
    }
    return newName;
}

// ... usage ...
std::string model_name = SanitizeFileName(graph_viewer.Name()) + ".onnx";
```

I've noticed that other parts of onnxruntime that write files might also not have strong path traversal prevention, so this might be a broader issue. Given that, I'll leave it to you to decide if this is something you want to address in this PR or if it should be handled separately as a larger effort to improve security across the project.

Other than this security consideration, the PR looks good to me.

snnn (dismissed):

onnxruntime/core/providers/migraphx/migraphx_execution_provider.cc
urpetkov-amd: Yes, this is a typo I missed accidentally, I'm changing it now.

onnxruntime/core/providers/migraphx/migraphx_execution_provider.cc
urpetkov-amd: Yes, it's typo because I just copied it from above example, it should be onnx_name, I'll change it.

urpetkov-amd:
@snnn 
I have made changes of typo errors and other recommendations.  I saw it's a common pattern not to use stream check for closing in a codebase, so I left it without it. If that's okay with you, I would leave it like this.
Also, for model naming, it's catching just the name of main graph (etc. torch_jit, main_graph, ...) and saving it in a working directory and it's used just for debugging. So, I would leave it like this and keep the opportunity to speak with the team to change it everywhere if it's needed on a global side and make a separate PR for that. Thanks!

snnn (approved):

PR #26397, opened 2025-10-23
Nightly python packages for python 3.14

tianleiwu:
Update python packaging pipeline to support python 3.14. This is to add python 3.14 wheels to both CUDA 12 and CUDA 13 nightly packaging pipelines. The CUDA 13 pipelines were added in https://github.com/microsoft/onnxruntime/pull/26420.

For linux, two wheels are added for python 3.14 and 3.14t.
For Windows, a wheel is added for pyhthon 3.14.

Known issues:
1. onnx does not have prebuild package for python 3.14, which causes some pipelines failed. 
2.  onnxscript and onnx-ir depends on onnx.
3. The CI machine for Windows cuda python packaging has cuda driver 553.24, which does not support CUDA 13.0. TODO: Update cuda driver to 580 or later to test properly.

Walkarounds for issues 1 and 2:
* For python 3.14, we skip onnx, onnxscript and onnx-ir in requirements.txt. 
* Disable related tests.

When onnx package supports python 3.14 in the future, we can undo the walkarounds and re-enable those tests.

### DIFF
--- a/tools/ci_build/github/linux/docker/inference/aarch64/python/cpu/scripts/install_deps.sh
+++ b/tools/ci_build/github/linux/docker/inference/aarch64/python/cpu/scripts/install_deps.sh
@@ -12,13 +12,5 @@ export CMAKE_ARGS="-DONNX_GEN_PB_TYPE_STUBS=ON -DONNX_WERROR=OFF"
 
 for PYTHON_EXE in "${PYTHON_EXES[@]}"
 do
-
-  # For python 3.14, install onnxscript without dependencies (It depends on onnx, which does not suppport python 3.14 yet)
-  # This can be removed once onnx supports python 3.14: update requirements.txt to replace onnx_weekly with onnx.
-  if [[ "$PYTHON_EXE" == *cp314* ]]; then
-    ${PYTHON_EXE} -m pip install --upgrade pip setuptools wheel
-    ${PYTHON_EXE} -m pip install onnxscript ml_dtypes onnx_ir onnx_weekly typing_extensions packaging numpy --no-deps
-  fi
-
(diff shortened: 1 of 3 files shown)


tools/ci_build/github/azure-pipelines/templates/py-package-smoking-test.yml
```diff
@@ -32,6 +32,8 @@ jobs:
         PythonVersion: '3.12'
       Python313:
         PythonVersion: '3.13'
+      Python310:
```
tianleiwu: 314

snnn:
> The CI machine for Windows cuda python packaging has cuda driver 553.24, which does not support CUDA 13.0.

If you mean the A10 machines, they use a special driver from https://learn.microsoft.com/en-us/azure/virtual-machines/windows/n-series-driver-setup

Now the driver is preinstalled in the VM images. But, now our infra has the ability to help us auto install the driver when VM starts up. We can use that feature then we won't need to remember to update the drivers. So, I need to create a new VM image without driver and test it.

tianleiwu:
> > The CI machine for Windows cuda python packaging has cuda driver 553.24, which does not support CUDA 13.0.
> 
> If you mean the A10 machines, they use a special driver from https://learn.microsoft.com/en-us/azure/virtual-machines/windows/n-series-driver-setup
> 
> Now the driver is preinstalled in the VM images. But, now our infra has the ability to help us auto install the driver when VM starts up. We can use that feature then we won't need to remember to update the drivers. So, I need to create a new VM image without driver and test it.

Right. Updating driver is independent of python 3.14. I think we can update the driver in another PR.

Issue #26377, opened 2025-10-21
KleidiAI library usage needs to check for SME2, not SME


john-dance:
### Describe the issue

Commit cd450d1 introduced the usage of the KleidiAI library. However, the test to enable it in platform.cpp just checks for SME, rather than SME2.

```
    if(MLAS_CPUIDINFO::GetCPUIDInfo().HasArm_SME()){
        this->MlasGemmBatchOverride = ArmKleidiAI::MlasGemmBatch;
        this->MlasGemmPackBSizeOverride = ArmKleidiAI::MlasGemmPackBSize;
        this->MlasGemmPackBOverride = ArmKleidiAI::MlasGemmPackB;
        this->MlasConvPrepareOverride = ArmKleidiAI::MlasConvPrepare;
        this->MlasConvOverride = ArmKleidiAI::MlasConv;
    }
```

This allows the library to be used on systems that support SME but not SME2, leading to illegal instructions during a run.

The newest pytorch/cpuinfo implementation has support for cpuinfo_has_arm_sme2(). https://github.com/search?q=repo%3Apytorch%2Fcpuinfo%20cpuinfo_has_arm_sme2&type=code

This needs to be added to cpuid_info.cc and exposed with a new function. Meanwhile, the default build should include "--no_kleidiai".

### To reproduce

Run a simple convolution using the CPU execution provider on a device that supports SME1, but not SME2.

### Platform

Android

### OS Version

16

### ONNX Runtime Installation

Built from Source

### ONNX Runtime Version or Commit ID

1.23

### ONNX Runtime API

C++

### Architecture

ARM64

### Execution Provider

Default CPU

snnn:
Oh, then I think it probably is Android and Linux only. 
Since the detection logic isn't fully implemented for the other operating systems. See https://github.com/pytorch/cpuinfo/issues/317 and https://github.com/pytorch/cpuinfo/issues/279 . Thanks @edgchen1  for pointing it out.

patryk-kaiser-ARM:
Hi @john-dance can you apply the patch attached in this comment, it should fix the issue for you:
https://github.com/microsoft/onnxruntime/pull/26399#issuecomment-3442557132

laurenspriem:
Any update on this issue? It's causing crashes on new devices such as OnePlus15 #26678

chengxinping:
Any update on this issue?

john-dance:
[...]

Issue #26361, opened 2025-10-20
[Mobile] crashes using Xiaomi 17 Pro Max


3xiaoxiao3:
### Describe the issue

报错及闪退log如下
`2025-10-20 16:00:41.825 20353-20353 MIUIInput               com.delicloud.app.ai.demo            I  [MotionEvent] ViewRootImpl windowName 'com.delicloud.app.ai.demo/com.delicloud.app.ai.demo.MainActivity', { action=ACTION_DOWN, id[0]=0, pointerCount=1, eventTime=261484241, downTime=261484241, phoneEventTime=16:00:41.823 } moveCount:0
2025-10-20 16:00:41.826 20353-20353 VRI[MainActivity]       com.delicloud.app.ai.demo            D  getMiuiFreeformStackInfo mTmpFrames.miuiFreeFormStackInfo: null
2025-10-20 16:00:41.858 20353-21653 oud.app.ai.demo         com.delicloud.app.ai.demo            I  This is non sticky GC, maxfree is 33554432 minfree is 8388608
2025-10-20 16:00:41.861 20353-21655 System                  com.delicloud.app.ai.demo            W  A resource failed to call AbstractCursor.close. 
2025-10-20 16:00:41.861 20353-21655 System                  com.delicloud.app.ai.demo            W  A resource failed to call CursorWrapperInner.close. 
2025-10-20 16:00:41.875 20353-20353 MIUIInput               com.delicloud.app.ai.demo            I  [MotionEvent] ViewRootImpl windowName 'com.delicloud.app.ai.demo/com.delicloud.app.ai.demo.MainActivity', { action=ACTION_UP, id[0]=0, pointerCount=1, eventTime=261484293, downTime=261484241, phoneEventTime=16:00:41.875 } moveCount:0
2025-10-20 16:00:41.883 20353-21721 SKIA                    com.delicloud.app.ai.demo            I  $$$$$$mi shader key lengh is 136.
2025-10-20 16:00:41.884 20353-21721 SKIA                    com.delicloud.app.ai.demo            I  CreateGraphicsPipeline pipeline cache hit. elpased time: 0.33 ms.
[...]

PR #26399, opened 2025-10-24
Fix: Disable KleidiAI on systems with SME1 but not SME2

snnn:
This PR fixes an issue where KleidiAI was incorrectly enabled on systems that support SME1 but not SME2, leading to crashes.

The issue was reported in #26377.

The fix involves changing the check in `onnxruntime/core/mlas/lib/platform.cpp` to use `HasArm_SME2()` instead of `HasArm_SME()`. This ensures that KleidiAI is only enabled on systems that have SME2 support.


patryk-kaiser-ARM:
[sme_check_fix.patch](https://github.com/user-attachments/files/23123824/sme_check_fix.patch)

We are aware that SME2 kernels can be executed on SME only hardware in qgemm.cpp and convolve.cpp, https://github.com/microsoft/onnxruntime/pull/26301/ fixes the issue for qgemm and another pr will be put in place soon to fix the checks in convolve.
For now it would be suggested to apply the attached patch to fix this issue and leave the check in platform.cpp for SME only as sgemm supports both (SME2 feature present implies SME).
Future PRs from us will have made these checks more robust also as we are moving to use an interface for all the kernels starting with sgemm (https://github.com/microsoft/onnxruntime/pull/26302)

Craigacp:
I think this might fix the crash reported in #26361 as well.

snnn:
@patryk-kaiser-ARM , then I will close this pull request and wait your PRs.

snnn:
I extracted the other part to a new PR: #26405

PR #26402, opened 2025-10-24
Adding SME1 Convolution Kernel to convole_kleidiai.cpp

JonathanC-ARM:
### Description
- Integration of SME1 Variant of existing SME2 convolution Kernel, kai_run_imatmul_clamp_f32_f32p2vlx1_f32p2vlx1b_2vlx2vl_sme_mopa and associated packing functions
- Formatting changes in convolve_kleidiai.cpp
- Addition of proper sme2 gate for dynamic qgemm
- Updating of kleidiai version to 1.14 (first version which contains the appropriate kernel)

### DIFF
--- a/onnxruntime/contrib_ops/cpu/quantization/dynamic_quantize_matmul.cc
+++ b/onnxruntime/contrib_ops/cpu/quantization/dynamic_quantize_matmul.cc
@@ -213,7 +213,7 @@ class DynamicQuantizeMatMul final : public MatMulIntegerToFloatBase {
         }
       }
 
-      // Currently, MlasDynamicQGemmBatch() and associated functions require SME or else they are no-ops.
+      // Currently, MlasDynamicQGemmBatch() and associated functions require SME2 or else they are no-ops.
       // We check that here too before attempting to use them.
       if (!CPUIDInfo::GetCPUIDInfo().HasArm_SME2()) {
         can_use_dynamic_quant_mlas_ = false;
--- a/onnxruntime/test/mlas/unittest/test_dynamic_qgemm.cpp
+++ b/onnxruntime/test/mlas/unittest/test_dynamic_qgemm.cpp
@@ -20,9 +20,9 @@ class MlasDynamicQgemmTest {
 
  public:
   void Test(size_t M, size_t N, size_t K, size_t BatchSize) {
-    // Currently, MlasDynamicQGemmBatch() and associated functions require SME or else they are no-ops.
-    if (!MLAS_CPUIDINFO::GetCPUIDInfo().HasArm_SME()) {
-      GTEST_SKIP() << "MlasDynamicQGemmBatch() requires ARM64 SME but it was not detected. Skipping test.";
+    // Currently, MlasDynamicQGemmBatch() and associated functions require SME2 or else they are no-ops.
+    if (!MLAS_CPUIDINFO::GetCPUIDInfo().HasArm_SME2()) {
+      GTEST_SKIP() << "MlasDynamicQGemmBatch() requires ARM64 SME2 but it was not detected. Skipping test.";
     }
 
     // Setup buffers for holding various data


cmake/deps.txt
```diff
@@ -56,5 +56,5 @@ extensions;https://github.com/microsoft/onnxruntime-extensions/archive/c24b7bab0
 directx_headers;https://github.com/microsoft/DirectX-Headers/archive/refs/tags/v1.613.1.zip;47653509a3371eabb156360f42faf582f314bf2e
 cudnn_frontend;https://github.com/NVIDIA/cudnn-frontend/archive/refs/tags/v1.12.0.zip;7e733cfdc410d777b76122d64232499205589a96
 dawn;https://github.com/google/dawn/archive/13c1635a14574ebb7116b56a69f5519301417fda.zip;0aadd28fc385cf7d657d5fc70a352372d2d3c76a
-kleidiai;https://github.com/ARM-software/kleidiai/archive/refs/tags/v1.10.0.tar.gz;11b62149cb2514b3b9069cc435c3aa7a4e82b97a
+kleidiai;https://github.com/ARM-software/kleidiai/archive/refs/tags/v1.14.0.tar.gz;161cce94808f1141b08e32096ccb1f294aa901c5
```
hariharans29: Looks like this can be bumped up to 1.15 now ? Given that https://github.com/microsoft/onnxruntime/pull/26301 gets that update anyway ?

onnxruntime/core/mlas/lib/kleidiai/convolve_kleidiai.cpp
```diff
@@ -599,10 +581,19 @@ static void ConvolveSme(const size_t co, //channels out
                 MIdx * m_step * co * sizeof(float) +
                 NIdx * n_step * sizeof(float)];
 
-            kai_run_imatmul_clamp_f32_f32p2vlx1_f32p2vlx1b_2vlx2vl_sme2_mopa(
-                TileSizeM, TileSizeN, d_kh*d_kw, ci, ATile, BTile, CTile, co * sizeof(float),
-                -std::numeric_limits<float>::max(), std::numeric_limits<float>::max()
-            );
+            if (ArmKleidiAI::UseSME2) {
+                KLEIDIAI_KERNEL_LOG("kai_run_imatmul_clamp_f32_f32p2vlx1_f32p2vlx1b_2vlx2vl_sme2_mopa" << " M=" << TileSizeM << " N=" << TileSizeN << " k_chunk_count=" << (d_kh * d_kw) << " k_chunk_length=" << ci);
```
hariharans29, line 593: I guess the usage of the logging macros here mean that we need to wait for the logging PR to be merged ?

hariharans29:
Can you please re-base ? @JonathanC-ARM

onnxruntime/core/mlas/lib/kleidiai/convolve_kleidiai.cpp
JonathanC-ARM: That's now resolved since the other pr merged

onnxruntime/core/mlas/lib/qgemm.cpp
```diff
@@ -210,9 +210,9 @@ MlasDynamicQGemmBatch (
     MLAS_THREADPOOL* ThreadPool
 ) {
 #if defined(USE_KLEIDIAI) && !defined(_MSC_VER)
-    //No fallback and putting in guards
-    if(MLAS_CPUIDINFO::GetCPUIDInfo().HasArm_SME()){
-    ArmKleidiAI::MlasDynamicQGemmBatch(Shape, DataParams, BatchN, ThreadPool);
+    //No fallback and putting in guards. This implementation is SME2 specific.
+    if(ArmKleidiAI::UseSME2){
```
hariharans29, line 214: Does this need a corresponding change here too - https://github.com/microsoft/onnxruntime/blob/c30905d638418383b8d83b3b1bb65b7b42226f5a/onnxruntime/contrib_ops/cpu/quantization/dynamic_quantize_matmul.cc#L218 ?

Seeing as there is no MLAS fallback, we would like to gate things appropriately in the upper CPU EP kernel.

hariharans29:
Can you please confirm if the corresponding tests passed on an SME1 machine or using SME1 mode on an SME2 machine ?

JonathanC-ARM:
Hi Hari, yeah so I've one push to come fixing the sme2 checks. These are going to be changed subsequently via the gemv pr anyway but in the meantime you are correct its right to fix them now. 

The sme1 tests pass also, we test on both sme2 and sme1. We typically force sme2 to be false to verify these kernels during development. Please see attached output of the tests for ort test all and mlas tests. Both of which pass and in the case of mlas tests we see that there are a few hundred skipped tests with a message "MlasDynamicQGemmBatch() requires ARM64 SME2 but it was not detected. Skipping test."
[ort_mlas_test.txt](https://github.com/user-attachments/files/23524401/ort_mlas_test.txt)
[ort_test_all.txt](https://github.com/user-attachments/files/23524402/ort_test_all.txt)

PR #26403, opened 2025-10-24
[MIGraphX EP] Bugfix for toVector Function causing rollover with int64 inputs

TedThemistokleous:
Refactors the toVector function which causes narrowing of the datatype for vector inputs such as in the slice operator

### Description

Removes the casting done for toVector as that causes rollovers due to some Onnx Operators using int64_t inputs rather than int32_t. This case causes us to actually incorrectly represent axis, start and end values in the slice operator

### Motivation and Context

Fixes a bug since this causes some of our Onnx operators to not operate in the full correct data range for input vectors

### DIFF
--- a/onnxruntime/core/providers/migraphx/migraphx_execution_provider.cc
+++ b/onnxruntime/core/providers/migraphx/migraphx_execution_provider.cc
@@ -364,11 +364,11 @@ static bool getMIGraphXType(ONNXTensorElementDataType type,
   return true;
 }
 
-std::vector<int> toVector(const ONNX_NAMESPACE::int64s& nums) {
-  std::vector<int> result;
-  int num = nums.size();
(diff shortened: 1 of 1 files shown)


TedThemistokleous:
Used to fix models in benchmarking in ROCm - https://github.com/ROCm/onnxruntime/pull/188

@snnn @cloudhan @tianleiwu  @ytaous

TedThemistokleous:
cc @devang-ml @tianleiwu @cloudhan can we get this looked at/reviewed? This is a bit of an insidious bug since it will clamp addresses on the slice operator due to the casting done in the toVector call. This won't break but will cause roll overs which will produce undefined behavior in our slice operator on the MIGraphX side.

tianleiwu:
@TedThemistokleous, 
Please rebase/merge latest main branch. The CI pipelines are not triggered.

PR #25776, opened 2025-08-18
python GPU IO Bindings for NVIDIA 

ishwar-raut1:
### Description

1. A Small change to use the shared allocator in Python binding. 
2. Remove the FP64 support from the EP. 

### Motivation and Context

The Python GPU IO binding is necessary for performance.  The change will enable the shared allocator for GPU allocation. 
The FP64 was using the FP32 inference—aligned WRT TRT RTX support.

PR #26423, opened 2025-10-28
[TRT RTX EP] Enable double datatype for intermediate nodes

thevishalagarwal:
* Enable double datatype for intermediate nodes
* add check for unsupported input output nodes in GetCapability()
* add free dims override for ort perf test in ep context compilation

### DIFF
--- a/onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
+++ b/onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
@@ -85,6 +85,25 @@ struct ShutdownProtobuf {
 
 namespace onnxruntime {
 
+// Helper function to check if a data type is supported by input output nodes ofNvTensorRTRTX EP
+static bool IsSupportedInputOutputDataType(ONNXTensorElementDataType data_type) {
+  switch (data_type) {
+    case ONNX_TENSOR_ELEMENT_DATA_TYPE_FLOAT:         // kFLOAT - 32-bit floating point
+    case ONNX_TENSOR_ELEMENT_DATA_TYPE_FLOAT16:       // kHALF - IEEE 16-bit floating-point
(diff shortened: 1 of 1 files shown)


onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
```diff
@@ -98,6 +98,7 @@ static bool IsSupportedDataType(ONNXTensorElementDataType data_type) {
     case ONNX_TENSOR_ELEMENT_DATA_TYPE_INT32:         // kINT32 - 32-bit signed integer
     case ONNX_TENSOR_ELEMENT_DATA_TYPE_INT64:         // kINT64 - 64-bit signed integer
     case ONNX_TENSOR_ELEMENT_DATA_TYPE_FLOAT8E4M3FN:  // kFP8 - 8-bit floating point
+    case ONNX_TENSOR_ELEMENT_DATA_TYPE_DOUBLE:        // kDOUBLE - 64-bit floating point
```
gaugarg-nv, line 120: I think the double support was disabled intentionally as TRT doesn't support fp64 datatype.

Check this PR where @ishwar-raut1 had removed fp64 support: https://github.com/microsoft/onnxruntime/pull/25776

ishwar-raut1: The fp64 needs to be converted into fp32 . But this is only true for the input and output nodes. There is no casting needed for intermediate node.
Please see this PR, which removed the support of the fp64
https://github.com/microsoft/onnxruntime/pull/25776

thevishalagarwal: Enabling double is intended only for intermediate nodes, not for input/output

thevishalagarwal:
@chilo-ms for review

PR #26429, opened 2025-10-29
[webgpu] revise implementation of buffer split support

fs-eire:
### Description

This PR addresses a few concerns:
- revert `const ProgramBase&` -> `ProgramBase&`: this itself is not doing something wrong but gives much more pressure for who reads the code to understand whether/where the program object is modified. It also can introduce further unexpected modifications to the program object (for example the indirect dispatch code)
- change bool option `"ep.webgpuexecutionprovider.smallStorageBufferBindingSizeForTesting"` to `"ep.webgpuexecutionprovider.maxStorageBufferBindingSize"` so now it's possible to set any value in option. (setting to <128MB will cause an assert failure)
- segments are optional in cache key (only present when not equals to 1, which is the common case)
- avoid some unnecessary API calls, which is OK for native but may affect web perf.
- clean up the code a little bit and add a few comments

### DIFF
--- a/onnxruntime/core/providers/webgpu/webgpu_context.cc
+++ b/onnxruntime/core/providers/webgpu/webgpu_context.cc
@@ -199,7 +199,8 @@ Status WebGpuContext::Run(ComputeContext& context, const ProgramBase& program) {
 
     if (program.IndirectDispatchTensor() != nullptr) {
       ORT_ENFORCE(!inputs.empty() && inputs.back().tensor == program.IndirectDispatchTensor(),
-                  "Indirect dispatch tensor must be the last input.");
+                  "The indirect dispatch tensor must be the last input. "
+                  "Ensure no call to program.AddInput() occurs after program.SetIndirectDispatchTensor().");
     }
 
     ORT_ENFORCE(std::all_of(outputs.begin(), outputs.end(), [](const ProgramOutput& output) {


onnxruntime/core/providers/webgpu/program.cc
```diff
@@ -362,6 +362,7 @@ ProgramBase& ProgramBase::SetDispatchGroupSize(uint32_t x, uint32_t y, uint32_t
 
 ProgramBase& ProgramBase::SetIndirectDispatchTensor(const Tensor* indirect_dispatch_tensor) {
   indirect_dispatch_tensor_ = indirect_dispatch_tensor;
+  AddInput({indirect_dispatch_tensor, ProgramTensorMetadataDependency::None});
```
qjia7, line 365: I considered this path. But it can't ensure that indirect_dispatch_tensor is added as the last input since we bind it as the last input in shader side.

onnxruntime/core/providers/webgpu/webgpu_context.cc
```diff
@@ -550,11 +564,11 @@ wgpu::Limits WebGpuContext::GetRequiredLimits(const wgpu::Adapter& adapter) cons
   required_limits.maxComputeWorkgroupsPerDimension = adapter_limits.maxComputeWorkgroupsPerDimension;
   required_limits.maxStorageBuffersPerShaderStage = adapter_limits.maxStorageBuffersPerShaderStage;
 
-  if (small_storage_buffer_binding_size_for_testing_) {
-    // No matter how small it is set, the minimum storage buffer binding size in WebGPU is 128 MB.
-    required_limits.maxStorageBufferBindingSize = 134217728;
-  } else {
+  if (max_storage_buffer_binding_size_ == 0) {
```
qjia7, line 568: Please also update the test case config https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/test/contrib_ops/matmul_4bits_test.cc#L551

onnxruntime/core/providers/webgpu/webgpu_context.cc
```diff
@@ -756,24 +770,24 @@ void WebGpuContext::LaunchComputePipeline(const wgpu::ComputePassEncoder& comput
...
       uint64_t offset = 0;
+      uint64_t buffer_size = wgpuBufferGetSize(buffer);
       for (uint32_t segment = 0; segment < total_segments; ++segment) {
         uint64_t segment_size = std::min(kMaxBufferSize, buffer_size - offset);
         bind_group_entries.push_back({nullptr, entry_index++, buffer, offset, segment_size, nullptr, nullptr});
         offset += segment_size;
       }
     } else {
-      bind_group_entries.push_back({nullptr, entry_index++, buffer, 0, std::min(kMaxBufferSize, buffer_size), nullptr, nullptr});
+      bind_group_entries.push_back({nullptr, entry_index++, buffer, 0, WGPU_WHOLE_SIZE, nullptr, nullptr});
```
xiaofeihan1, line 791: I wonder that is there a specific reason to use WGPU_WHOLE_SIZE (UINT64_MAX)? Intuitively, using the actual buffer size be more appropriate?

fs-eire: the reason is I try to avoid perform one call to `wgpuBufferGetSize`, which is trivial in native but can be expensive in web because it involves an inter-op call.

onnxruntime/core/providers/webgpu/program.cc
fs-eire: currently it is checked [here](https://github.com/microsoft/onnxruntime/pull/26429/files#diff-19e068e3cc8d76b34fab02a26f3bbdffe722e5fbdba92fa0617ef1c22a02ad14R200-R203).

do you prefer to support indirect_dispatch_tensor being able to be added in any order?

qjia7: The simpler the better. I just want to call out that we can't make sure the `SetIndirectDispatchTensor` is called after AddInputs. In this way, it's possible that `indirect_dispatch_tensor` is not the last input. The checking does can check the order errors. But it's not clear to tell developers where the error happens since indirect_disptach_tensor is added implicitly by  `SetIndirectDispatchTensor`. Maybe update the check message like please make sure `SetIndirectDispatchTensor` is called after AddInputs so that Indirect dispatch tensor is the last input.

onnxruntime/core/providers/webgpu/webgpu_context.cc
qjia7: Will it be better to change it only bind the normalized tensor size (align with 16) which is the actual size we need and no extra memory is bound and no wgpuBufferGetSize is called? Anyway, I think it's ok to leave it as `WGPU_WHOLE_SIZE`. I am just curious whether it helps the performance if we only bind the needed size instead of the buffer size which is the bucket size.

onnxruntime/core/providers/webgpu/program.cc
fs-eire: updated comments

onnxruntime/core/providers/webgpu/webgpu_context.cc
fs-eire: I will think about how to do this in another PR. Ideally, all necessary information for creating binding group can be determined earlier.

qjia7 (approved):

PR #26433, opened 2025-10-29
webgpu: optimize Gemm and MatMul using subgroup feature

xhcao:


xhcao:
@guschmue @fs-eire @qjia7 @jchen10 @Jiawei-Shao 
Hi, all, I want to discuss with you whether we could optimize Gemm, MatMul and Conv operators on some special vendor and special architectures as this PR.
Reasons:
1. It is difficult to design an algorithm that benifits all vendors and all architectures.
2. Even for the same vendor, but different architectures, it is also difficult.
3. Maintaining and reviewing the code is also difficult if adding some vendor and architecture information in common files.
4. There are not enough devices to test the correctness and performance.
...

jchen10:
@xhcao It seems you didn't include your test cases in this PR. What's your concern? BTW, you'd better provide some performance data for reference!

xhcao:
> @xhcao It seems you didn't include your test cases in this PR. What's your concern? BTW, you'd better provide some performance data for reference!

Firstly, I want being consented to create a `vendor` directory and make some optimizations on some special platforms, especially from Microsoft reviewers' consent.
If so, I will add the test cases in the other PR and apply performance data here.

jchen10:
> > @xhcao It seems you didn't include your test cases in this PR. What's your concern? BTW, you'd better provide some performance data for reference!
> 
> Firstly, I want being consented to create a `vendor` directory and make some optimizations on some special platforms, especially from Microsoft reviewers' consent. If so, I will add the test cases in the other PR and apply performance data here.

You don't need to depend on it. You can just best improve your PR with more test cases, better perf data, and easier to review.

fs-eire:
could you please help to merge to latest main and push?

The pipeline failures should be unrelated, and may be fixed by rerun.

guschmue:
I understand the desire for a vendor directory. 
But I'm concerned that we'll have trouble with testing and CI if we have too many vendor specific paths - it will get fragmented and very hard to test all code path variations.
We already head this problem with subgroup.matrixmultiply where I could not test a change for the intel specific code path so I had to let it fall back to different path.

But we can try and see how that works, just need to be careful with the testing

jchen10:
@guschmue Thank you for your understanding and support. We will fully test our changes and closely monitor the status on LNL/BMG. Once mature enough we could roll out to more devices.

qjia7:
Haven't closely look at all code. But several high level comments.
1. Please use workgroup_idx instead workgroup_id.x/y/z to recalculate the right offset.
2. Please use the help function .setByOffset/getByOffset/setByIndices/getByindices to load/store data.
3. Does your shader support any size of subgroup? I don't see sg_size related checking in your shader. I suppose intel's subgroup size range 8-32.
4. Please use template to write the shader if possible.

xhcao:
@fs-eire Had addressed your comments. @qjia7 Except for the fourth comment (Because I will continue to optimize the PR, and address the comment in future), all the comments have been addressed.
PTAL, thank you.

xhcao:
Attach shder files

[gemm_subgroup_shader.txt](https://github.com/user-attachments/files/23847599/gemm_subgroup_shader.txt)
[gemm_subgroup_vec4_shader.txt](https://github.com/user-attachments/files/23847600/gemm_subgroup_vec4_shader.txt)

xhcao:
@qjia7 @fs-eire PTAL, thanks.

onnxruntime/core/providers/webgpu/math/gemm_utils.cc
```diff
@@ -123,59 +123,63 @@ void MatMulReadFnSource(ShaderHelper& shader,
                         const ShaderIndicesHelper* batch_dims,
                         bool transA,
                         bool transB,
-                        bool is_vec4) {
-  int components = is_vec4 ? 4 : 1;
+                        bool is_vec4,
+                        bool use_subgroup) {
+  // Always read A with 1-component when using subgroup.
```
qjia7: Maybe both is_vec4 and use_subgroup are not necessary. In the body, we can use a|b.NumComponents() to get the right components.

onnxruntime/core/providers/webgpu/vendor/intel/math/gemm.cc
```diff
@@ -0,0 +1,121 @@
...
+  if (need_handle_matmul) {
+    program.AddInputs({{a, ProgramTensorMetadataDependency::TypeAndRank, a_components},
+                       {b, ProgramTensorMetadataDependency::TypeAndRank, b_components}});
+  }
+
+  if (need_handle_bias) {
+    program.AddInput({c, ProgramTensorMetadataDependency::TypeAndRank, c_components});
+  }
+
+  program.CacheHint(alpha, transA, transB, c_is_scalar, is_vec4, absl::StrJoin(elements_per_thread, "-"))
```
qjia7: nit: is_vec4 is not needed since they are already integrated into  the common key part by `ProgramTensorMetadataDependency::TypeAndRank` (Type will includes the components).

xhcao: Done

onnxruntime/core/providers/webgpu/math/gemm_utils.cc
xhcao: Done. Also removed `output_components` in `MatMulWriteFnSource` function

onnxruntime/core/providers/webgpu/vendor/intel/math/gemm_subgroup.cc
```diff
@@ -0,0 +1,184 @@
...
+  }
+
+  return SS_GET(cal_acc_ss);
+}
+
+}  // namespace
+
+bool CanApplySubgroup(const ComputeContext& context, int64_t M, int64_t N, int64_t K, bool transA, bool transB) {
+  if (context.AdapterInfo().vendor == std::string_view{"intel"}) {
+    bool use_subgroup = (context.AdapterInfo().architecture == std::string_view{"xe-2lpg"} ||
```
qjia7: Please also add check `context.HasFeature(wgpu::FeatureName::Subgroups)`.
Can we apply it to all intel devices with subgroups support?

onnxruntime/core/providers/webgpu/vendor/intel/math/gemm.cc
```diff
@@ -0,0 +1,121 @@
...
+  if (need_handle_matmul) {
+    program.AddInputs({{a, ProgramTensorMetadataDependency::TypeAndRank, a_components},
+                       {b, ProgramTensorMetadataDependency::TypeAndRank, b_components}});
+  }
+
+  if (need_handle_bias) {
+    program.AddInput({c, ProgramTensorMetadataDependency::TypeAndRank, c_components});
+  }
+
+  program.CacheHint(alpha, transA, transB, c_is_scalar, absl::StrJoin(elements_per_thread, "-"))
```
qjia7: Since you add alpha as uniform, why still use it as const value and cache it in CacheHint?

xhcao:
@qjia7 I had addressed your comments. @fs-eire @guschmue Please take a look, thanks.

onnxruntime/core/providers/webgpu/vendor/intel/math/gemm_subgroup.cc
xhcao: Had added the subgroup support check. I had only tuned the parameters on LNL machines, not tuned on the other intel devices. I will continue to optimize the code, if it could get good performance on the other devices, I will apply it in future.

xhcao:
@qjia7 I tested the PR using ~20 models on Rocket-Lake, several models got performance improvement, but was less than Lunar-Lake.
There was no regression on any models, so applied the PR to all intel devices support subgroup feature. Please take a look again.

qjia7:
The shader changes look good to me. I’ll hand off the remaining structural review to @fs-eire and @guschmue. Thanks!

xhcao:
@fs-eire @guschmue PTAL

xhcao:
The performance data of some models is shown as below on Lunar-lake.
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/xcao10/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/xcao10/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
<style>

</style>
</head>

<body link="#467886" vlink="#96607D">


Model | Performance
-- | --
depth-anything-base-fp16-pow-fp32 | 10%
depth-anything-base-fp32 | 15%
florence-2-base-decoder-fp16 | 15%
florence-2-base-encoder-fp16 | 35%
florence-2-base-vision-encoder-fp16-0.5-fp32 | 15%
jina-clip-v1-version | 10%
moondream2-vision-encoder-fp16 | 25%
sam-vit-b-encoder-fp16-demo | 20%
sd-turbo-safety-checker-int32-reduceSum | 25%
sd-turbo-unet-fp16-demo-layernorm | 10%
whisper-base-encoder-lm-fp16-layernorm | 20%



</body>

</html>

guschmue (approved):

PR #26438, opened 2025-10-29
Allow get clip constant min/max from DequantizeLinear

qti-yuduo:
When detecting QDQ node group. Clip would usually be selected part of Conv node group when redundant. This change allow Clip min/max value from DQ node to be considered and performing same check. Allowing more use cases to be tagged as valid QDQ node group, result in substantial performance boost for EPs.

### DIFF
--- a/onnxruntime/core/optimizer/utils.cc
+++ b/onnxruntime/core/optimizer/utils.cc
@@ -161,11 +161,6 @@ bool IsAttributeWithExpectedValues(const Node& node, const std::string& attr_nam
   return true;
 }
 
-bool IsScalarOr1Element1DTensor(gsl::span<const int64_t> tensor_shape) {
-  const size_t rank = tensor_shape.size();
-  return (rank == 0) || ((rank == 1) && (tensor_shape[0] == 1));
-}
-
 bool AppendTensorFromInitializer(const Graph& graph, const NodeArg& input_arg, InlinedVector<int64_t>& data, bool require_constant) {
   if (require_constant && !graph_utils::IsConstantInitializer(graph, input_arg.Name(), true)) {
     return false;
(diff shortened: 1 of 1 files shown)


onnxruntime/core/optimizer/utils.cc
```diff
@@ -330,28 +330,112 @@ bool GetClipConstantMinMax(const Graph& graph, const Node& node, float& min, flo
...
+            const ONNX_NAMESPACE::TensorProto* dq_input = graph.GetConstantInitializer(dq_inputs[0]->Name(), true);
+            const ONNX_NAMESPACE::TensorProto* dq_scale = graph.GetConstantInitializer(dq_inputs[1]->Name(), true);
+            const ONNX_NAMESPACE::TensorProto* dq_zero_point = graph.GetConstantInitializer(dq_inputs[2]->Name(), true);
+            if (!dq_input || !dq_scale || !dq_zero_point) {
+              return false;
+            }
+            // Check scale and zero_point are scalar
+            Initializer scale_initializer(graph, *dq_scale, graph.ModelPath());
+            Initializer zero_point_initializer(graph, *dq_zero_point, graph.ModelPath());
+            if (scale_initializer.dims().size() != 0 || zero_point_initializer.dims().size() != 0) {
```
adrianlizarraga: Should this also handle a shape of `(1,)` for the scale? I believe a scale of this shape is also considered per-tensor quantization.

Refer to:
https://github.com/microsoft/onnxruntime/blob/fc3bc097c941c812f481cff42594f2da1ae79097/onnxruntime/core/optimizer/transpose_optimization/onnx_transpose_optimization.cc#L29C16-L32

adrianlizarraga (commented):
One minor comment

onnxruntime/core/optimizer/utils.cc
qti-yuduo: Done.

adrianlizarraga (approved):

Issue #26294, opened 2025-10-13
[Performance] EPCTX file check doesn't happen for default model path


preetha-intel:
### Describe the issue

The check if an Epctx file alreay exists, is carried out at the beginning of session creation even before graph_partitioning is called to fail early when session_options.epctx_file_path key is set. 

As per this PR :
https://github.com/microsoft/onnxruntime/commit/8705c68c66f5009339ce2b99570ff23246449ef0#r167792340 

The check is not happening for default model paths. If epctx file path is not explicitly set then EP compile happens and failure for a exisiting epctx file is thrown in dump epctx model. 

@adrianlizarraga I have added the comments part of the PR as well 

 

### To reproduce

onnxruntime_perf_test.exe -v -m times -r 1 -C "ep.context_enable|1 ep.context_embed_mode|0" -i "device_type|CPU" <path_to_model>

 During first run epctx.onnx model will be generated. 

When the same command is run again, the process should fail early that epctx.onnx model exists.

### Platform

Windows

### OS Version

Windows

### ONNX Runtime Installation

Built from Source

### ONNX Runtime Version or Commit ID

8705c68

### ONNX Runtime API

C++

### Architecture

X64

### Execution Provider

Default CPU

### Is this a quantized model?

No

adrianlizarraga:
Hi @preetha-intel,
Thank you for reporting this. Here is a PR with the fix: https://github.com/microsoft/onnxruntime/pull/26439

MayureshV1:
@adrianlizarraga .. verified that the PR resolves the issue !

PR #26439, opened 2025-10-29
Fix early check that prevents a compiled model from being overwritten

adrianlizarraga:
### Description
Fixes https://github.com/microsoft/onnxruntime/issues/26294

When using the old model compilation approach (session option configuration), ORT should verify that the generated output model does not already exist. Importantly, this check should be done _before_ calling an EP's compile() method. This PR fixes this check, which was unintentionally disabled by a [previous PR.](https://github.com/microsoft/onnxruntime/pull/25455).

Note that this check also (correctly) happens _after_ calling the EP's compile() method, but it is better to catch it early if we can.

### Motivation and Context
Fixes a regression in the older compilation workflow.

### DIFF
--- a/include/onnxruntime/core/graph/graph.h
+++ b/include/onnxruntime/core/graph/graph.h
@@ -1454,13 +1454,6 @@ class Graph {  // NOLINT(clang-analyzer-optin.performance.Padding): preserve exi
     return Resolve(default_options);
   }
 
-  /// <summary>
-  /// This function converts all the graph TensorProto initializers into OrtValues
-  /// and creates a in-memory external data reference for each OrtValue.
-  /// </summary>
-  /// <returns></returns>
-  Status ConvertInitializersIntoOrtValues();
-
   /**
    * @brief Converts a subset of graph TensorProto initializers into OrtValues and updates the graph proto.
    *
--- a/onnxruntime/core/framework/session_state_utils.cc
+++ b/onnxruntime/core/framework/session_state_utils.cc
@@ -140,12 +140,12 @@ static common::Status DeserializeTensorProto(const Env& env, const std::basic_st
                                        std::move(tensor), ort_value);
     }
   } else {
-    // for internal initializer, always allocate memory on device - tensor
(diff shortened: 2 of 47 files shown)


adrianlizarraga:
Close and re-open to re-run checks

MayureshV1:
Thanks Adrian. This PR resolves [issue:26294](https://github.com/microsoft/onnxruntime/issues/26294)

chilo-ms (approved):

onnxruntime/core/framework/ep_context_options.cc
```diff
@@ -15,15 +15,15 @@ namespace epctx {
...
 
-  std::string output_model_path = config_options.GetConfigOrDefault(kOrtSessionOptionEpContextFilePath, "");
-  if (!output_model_path.empty()) {
-    output_model_location = std::filesystem::path(output_model_path);
-  } else {
-    output_model_location = std::monostate{};
-  }
+  // Note: the older compilation approach only supports compiling to an output file.
+  output_model_location = std::filesystem::path(
+      config_options.GetConfigOrDefault(kOrtSessionOptionEpContextFilePath, ""));
```
edgchen1, line 26: what does it mean for the path to be empty? it looks like the check in `CreateSessionAndLoadModelImpl()` in utils.cc treats that as invalid. is it also invalid here?

adrianlizarraga: If the output path is empty, then it is auto generated based on the input path name. 

It is invalid in utils.cc only when the input path is null (i.e., input model loaded from bytes). In this case, ORT has no way to auto generate the output path.

adrianlizarraga: I'll clarify this in utils.cc

Edit: Added comments in utils.cc

edgchen1 (approved):

onnxruntime/core/framework/ep_context_options.cc
edgchen1: got it. thanks for explaining that and adding the comments.

PR #26444, opened 2025-10-29
[TRT/TRT RTX EP] Fix bug for missing outputs in the returning ComputeCapability/IndexedSubGraph

chilo-ms:
### Description
For TRT EP's `GetCapability()`, in some case, the `GetSubGraph()` won't add graph's output to the `ComputeCapability/IndexedSubGraph` returning to ORT.

The issue if from following code:
````c++
...
if (node->GetOutputEdgesCount() > node->OutputDefs().size()) {
 ... // execute here
} else {
  ...
          if (graph_output_names.find(output->Name()) != graph_output_names.end()) {
            graph_outputs_to_add[output] = output_order; // missing this
          }
}
````

Update TRT RTX EP as well.

### Motivation and Context
https://github.com/microsoft/onnxruntime/issues/25373

### DIFF
--- a/onnxruntime/test/testdata/node_output_not_used.py
+++ b/onnxruntime/test/testdata/node_output_not_used.py
@@ -4,9 +4,9 @@ from onnx import TensorProto, helper
 
 def create_model_with_node_output_not_used(model_path):
     # Create graph
-    X = helper.make_tensor_value_info("X", TensorProto.FLOAT, [3, 2])
-    W = helper.make_tensor_value_info("W", TensorProto.FLOAT, [2, 3])
-    Y = helper.make_tensor_value_info("Y", TensorProto.FLOAT, [2, 3])
+    x = helper.make_tensor_value_info("X", TensorProto.FLOAT, [3, 2])
+    w = helper.make_tensor_value_info("W", TensorProto.FLOAT, [2, 3])
+    y = helper.make_tensor_value_info("Y", TensorProto.FLOAT, [2, 3])
 
     # Dropout node (two outputs)
     dropout_node = helper.make_node(
@@ -27,8 +27,8 @@ def create_model_with_node_output_not_used(model_path):
     graph = helper.make_graph(
         nodes=[dropout_node, matmul_node],
         name="DropoutMatMulGraph",
-        inputs=[X, W],
-        outputs=[Y],
+        inputs=[x, w],
+        outputs=[y],
     )
 
     model = helper.make_model(graph, opset_imports=[helper.make_operatorsetid("", 13)])
--- a/onnxruntime/test/testdata/topk_and_multiple_graph_outputs.py
+++ b/onnxruntime/test/testdata/topk_and_multiple_graph_outputs.py
@@ -11,7 +11,7 @@ def create_model_with_topk_graph_output(model_path):
     # ======================
     # ---- Initializers ----
     # ======================
-    K = helper.make_tensor("K", TensorProto.INT64, dims=[1], vals=[300])
+    k = helper.make_tensor("K", TensorProto.INT64, dims=[1], vals=[300])
     zero = helper.make_tensor("zero", TensorProto.INT64, dims=[], vals=[0])
     twenty_six = helper.make_tensor("twenty_six", TensorProto.INT64, dims=[], vals=[26])
 
@@ -62,7 +62,7 @@ def create_model_with_topk_graph_output(model_path):
         name="TopKGraph",
         inputs=[input_tensor],
         outputs=[scores_out, less_out, div_out, labels_out],
-        initializer=[K, zero, twenty_six],
+        initializer=[k, zero, twenty_six],
     )
 
     model = helper.make_model(graph, opset_imports=[helper.make_operatorsetid("", 13)])

### CODE REVIEW

**Comment by gcunhase** [comment]
@chilo-ms any more AIs needed for this to be merged? Thanks.

File: onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc [hunk 1]
```diff
@@ -1387,9 +1387,25 @@ std::unique_ptr<IndexedSubGraph> NvExecutionProvider::GetSubGraph(SubGraph_t gra
   }
 
   // Find inputs and outputs of the subgraph
+
   std::unique_ptr<IndexedSubGraph> sub_graph = onnxruntime::IndexedSubGraph::Create();
-  std::unordered_map<const NodeArg*, int> original_inputs, fused_inputs, fused_outputs, fused_outputs_to_add, graph_outputs_to_add;
+  std::unordered_map<const NodeArg*, int> original_inputs;
+
+  // These maps store the inputs and outpus of the subgraph.
```
edgchen1:
```suggestion
  // These maps store the inputs and outputs of the subgraph.
```

onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
```diff
@@ -1451,32 +1467,38 @@ std::unique_ptr<IndexedSubGraph> NvExecutionProvider::GetSubGraph(SubGraph_t gra
             erased.insert(output);
           } else if (erased.find(output) == erased.end()) {
             if (graph_output_names.find(output->Name()) != graph_output_names.end()) {
-              graph_outputs_to_add[output] = output_order;
+              // This output is the graph's output.
+              // So the output should be put into the subgraph's output list.
+              graph_outputs_to_add.insert({output, output_order});
```
yuslepukhin: > graph_outputs_to_add.insert({output, output_order}); [](http://example.com/codeflow?start=13&length=53)

It would not insert as before if the entry existed

onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
```diff
@@ -1451,32 +1467,38 @@ std::unique_ptr<IndexedSubGraph> NvExecutionProvider::GetSubGraph(SubGraph_t gra
...
-        else if (erased.find(output) == erased.end()) {
-          if (graph_output_names.find(output->Name()) != graph_output_names.end()) {
-            graph_outputs_to_add[output] = output_order;
-          }
+    }
 
-          if (graph.GetGraph().GetConsumerNodes(output->Name()).size() > 0) {
-            fused_outputs[output] = output_order++;
-          }
+    for (const auto& output : node->OutputDefs()) {
```
yuslepukhin, line 1497: >for (const auto& output : node->OutputDefs()) { [](http://example.com/codeflow?start=4&length=47)

General comment for both inputs and outputs.
Should we check at some point if an optional input/output Exists()?

onnxruntime/test/providers/nv_tensorrt_rtx/nv_basic_test.cc
```diff
@@ -203,6 +203,56 @@ TEST_P(TypeTests, IOTypes) {
...
+   * Model #2:
+   *
+   * "X" ---> Dropout ---> MatMul ---> "Y"
+   *          ^     |
+   *          |     |
+   * "W" ------     ----> Can't be graph's output
+   *
+   */
+  {
+    Ort::Env env{ORT_LOGGING_LEVEL_WARNING, "test"};
```
yuslepukhin: > Ort::Env env{ORT_LOGGING_LEVEL_WARNING, "test"}; [](http://example.com/codeflow?start=3&length=49)

Just instantiate it once

onnxruntime/test/providers/nv_tensorrt_rtx/nv_basic_test.cc
```diff
@@ -203,6 +203,56 @@ TEST_P(TypeTests, IOTypes) {
...
+   * Model #1:
+   *
+   * "input" ---> TopK ---
+   *                     |---> "scores"
+   *                     |--- Less ---> "Less_output_0"
+   *                     |--- Div ---> "Div_output_0"
+   *                     |--- Mod ---> "labels"
+   */
+  {
+    Ort::Env env{ORT_LOGGING_LEVEL_WARNING, "test"};
```
edgchen1: there's already a global `Ort::Env` instance [here](https://github.com/microsoft/onnxruntime/blob/186c938c014019b2ca5d80320631db6419bc687c/onnxruntime/test/unittest_main/test_main.cc#L40) - why do we need to define another one? the underlying `OrtEnv` is a singleton so this would refer to the same instance.

onnxruntime/test/providers/nv_tensorrt_rtx/test_nv_trt_rtx_ep_util.cc
```diff
@@ -464,5 +472,266 @@ Ort::IoBinding generate_io_binding(
...
+                                            const std::vector<float>& vals) {
+  ONNX_NAMESPACE::TensorProto t;
+  t.set_name(name);
+  t.set_data_type(ONNX_NAMESPACE::TensorProto::FLOAT);
+  for (auto d : dims) t.add_dims(d);
+  for (auto v : vals) t.add_float_data(v);
+  return t;
+}
+
+OrtStatus* CreateModelWithNodeOutputNotUsed(const PathString& model_name) {
```
edgchen1: for other tests that use models, we have checked in .onnx files and a script to generate them. is there a good reason to do it differently here? defining the model in a Python script (e.g., with onnxscript) could be more concise.

onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
chilo-ms: If the key already exists in any of the those maps, i.e. `fused_inputs`,
`fused_outputs`, `fused_outputs_to_add` and `graph_outputs_to_add`, it's not necessarily to override it.

input_order/output_order is simply a relative ordering that associated with the input/output, so that when constructing final sub_graph's input of output lists from above maps, input/output with smaller order index will appear before those with larger order indices.

So exact order index is not necessary, as long as the order index for an output that should appear before another output has smaller order index is sufficient.

BTW, i added the comments for input_order/output_order to explain the usage of them.

onnxruntime/test/providers/nv_tensorrt_rtx/nv_basic_test.cc
chilo-ms: fixed

onnxruntime/test/providers/nv_tensorrt_rtx/nv_basic_test.cc
chilo-ms: changed to use  global `ort_env`

onnxruntime/test/providers/nv_tensorrt_rtx/test_nv_trt_rtx_ep_util.cc
chilo-ms: okay, i added the python scripts as well as the models

onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
chilo-ms: Yes, as you scroll down and will see the code calls Exists() for not including the optional input/output

````c++
..
  for (const auto& input : inputs) {
    if (input.second->Exists()) {
      meta_def->inputs().push_back(input.second->Name());
    }
  }

  for (const auto& initializer : initializers) {
    meta_def->constant_initializers().push_back(initializer);
  }

  for (const auto& output : outputs) {
    if (output.second->Exists()) {
      meta_def->outputs().push_back(output.second->Name());
    }
  }
  ..
````

PR #26456, opened 2025-10-30
WeightBiasQuantization: fix expected axis for ConvTranspose

qti-yuduo:
`WeightBiasQuantization` not properly quantize bias when we have `ConvTranspose` op with float bias. This is due to we're not correctly set proper expected filter axis.

### DIFF
--- a/onnxruntime/test/optimizer/qdq_transformer_test.cc
+++ b/onnxruntime/test/optimizer/qdq_transformer_test.cc
@@ -5593,6 +5593,54 @@ TEST(QDQTransformerTests, WeightBiasQuantization_ConvTranspose_Weight) {
 #endif
 }
 
+TEST(QDQTransformerTests, WeightBiasQuantization_ConvTransposePerChannelWeight) {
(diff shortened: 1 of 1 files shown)


onnxruntime/core/optimizer/qdq_transformer/weight_bias_quantization.cc
```diff
@@ -147,6 +147,8 @@ Status WeightBiasQuantization::ApplyImpl(Graph& graph, bool& modified, int graph
           transB = trans_b_iter->second.i();
         }
         expected_axis = transB == 0 ? 1 : 0;
+      } else if (node.OpType() == "ConvTranspose") {
+        expected_axis = 1;
```
adrianlizarraga, line 151: Would be good to add a unit test for this use case to prevent future regressions.

qti-yuduo: Added.

adrianlizarraga (approved):

PR #26457, opened 2025-10-30
[QNN-EP] - Tie HTP power config id lifetime to session

quic-calvnguy:
### Description

- Remove PerThreadContext (Only used for id management)
- Create ManagedHtpPowerConfigId to manage destruction of id
- Only create one htp power config id per session (previously was one per thread)

### Motivation and Context
One session can potentially be used on multiple threads for execution, and there is a set max number of htp power config ids that can be used at any given time. If enough sessions and enough threads are created, then the maximum number of power config ids can be reached easily (see ticket).

Additionally, all power configurations are available on a per-session basis. Therefore, there is no reason to have more than one power config id per session.

Removal of PerThreadContext is due to the fact that it was only used to contain and destroy the power config ids on thread termination. As such, there is no more need for PerThreadContext.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
@@ -1564,16 +1564,16 @@ Status QnnBackendManager::SetPerThreadHtpPowerConfigs(const std::thread::id& thr
   auto htp_power_config_id = htp_power_configs.power_config_id;
   if (pre_run) {
     if (htp_power_configs.pre_run_perf_mode.has_value()) {
-      ORT_RETURN_IF_ERROR(SetHtpPowerConfig(htp_power_config_id, htp_power_configs.pre_run_perf_mode.value()));
+      ORT_RETURN_IF_ERROR(SetHtpPowerConfig(htp_power_config_id, *htp_power_configs.pre_run_perf_mode));
     }
 
     if (htp_power_configs.rpc_configs.has_value()) {
       ORT_RETURN_IF_ERROR(SetRpcPowerConfigs(htp_power_config_id,
-                                             htp_power_configs.rpc_configs.value().rpc_control_latency,
-                                             htp_power_configs.rpc_configs.value().rpc_polling_time));
+                                             htp_power_configs.rpc_configs->rpc_control_latency,
+                                             htp_power_configs.rpc_configs->rpc_polling_time));
     }
   } else if (htp_power_configs.post_run_perf_mode.has_value()) {
-    ORT_RETURN_IF_ERROR(SetHtpPowerConfig(htp_power_config_id, htp_power_configs.post_run_perf_mode.value()));
+    ORT_RETURN_IF_ERROR(SetHtpPowerConfig(htp_power_config_id, *htp_power_configs.post_run_perf_mode));
   }
 
   return Status::OK();
--- a/onnxruntime/core/providers/qnn/qnn_execution_provider.cc
+++ b/onnxruntime/core/providers/qnn/qnn_execution_provider.cc
@@ -655,7 +655,7 @@ QNNExecutionProvider::~QNNExecutionProvider() {
 
     std::lock_guard<std::mutex> lock(config_id_mutex_);
     if (htp_power_config_id_.has_value()) {
-      ORT_IGNORE_RETURN_VALUE(qnn_backend_manager_->DestroyHTPPowerConfigID(htp_power_config_id_.value()));
+      ORT_IGNORE_RETURN_VALUE(qnn_backend_manager_->DestroyHTPPowerConfigID(*htp_power_config_id_));
     }
   }
 
@@ -1543,7 +1543,7 @@ bool QNNExecutionProvider::GetHtpPowerConfigId(uint32_t& htp_power_config_id) {
     return false;
   }
 
-  htp_power_config_id = htp_power_config_id_.value();
+  htp_power_config_id = *htp_power_config_id_;
   return true;
 }
 
--- a/onnxruntime/core/providers/qnn/qnn_execution_provider.h
+++ b/onnxruntime/core/providers/qnn/qnn_execution_provider.h
@@ -89,6 +89,9 @@ class QNNExecutionProvider : public IExecutionProvider {
   // Will return false if htp_power_config_id_ has no value
   bool GetHtpPowerConfigId(uint32_t& htp_power_config_id);
 
+  // htp_power_config_id_ must be created during GetCapability() as it is the only
+  // step during setup/initialization in which QNNBackendManager is setup and ready.
+  // GetCapability() is a const function, so these options must be mutable
   mutable std::optional<uint32_t> htp_power_config_id_;
   mutable std::mutex config_id_mutex_;
 


onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1479,14 +1390,14 @@ Status QNNExecutionProvider::OnRunStart(const onnxruntime::RunOptions& run_optio
     rpc_polling_time = 9999;
   }
 
-  if (GetPerThreadContext().IsHtpPowerConfigIdValid()) {
+  if (IsHtpPowerConfigIdValid()) {
     if (qnn::HtpPerformanceMode::kHtpDefault != htp_performance_mode) {
-      ORT_RETURN_IF_ERROR(qnn_backend_manager_->SetHtpPowerConfig(GetPerThreadContext().GetHtpPowerConfigId(),
+      ORT_RETURN_IF_ERROR(qnn_backend_manager_->SetHtpPowerConfig(GetHtpPowerConfigId(),
```
adrianlizarraga: When a session is used across multiple threads, is it possible that they can interfere with each other here (and in other places where the EP calls `qnn_backend_manager_->Set*PowerConfig`.)?

The mutex is only locked during the call to `GetHtpPowerConfigId()` but there is no synchronization here.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1555,6 +1555,63 @@ Status QnnBackendManager::SetRpcPowerConfigs(uint32_t htp_power_config_client_id
...
+
+bool QnnBackendManager::GetPerThreadHtpPowerConfigMapping(const std::thread::id& thread_id,
+                                                          PerThreadHtpPowerConfigs_t& htp_power_configs) {
+  std::lock_guard<std::mutex> lock(per_thread_power_configs_mutex_);
+
+  if (per_thread_power_configs_.find(thread_id) == per_thread_power_configs_.end()) {
+    return false;
+  }
+
+  htp_power_configs = per_thread_power_configs_.at(thread_id);
```
edgchen1: don't need to look up again - just use the result of `unordered_map::find()`

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1555,6 +1555,63 @@ Status QnnBackendManager::SetRpcPowerConfigs(uint32_t htp_power_config_client_id
...
+  return true;
+}
+
+void QnnBackendManager::RemovePerThreadHtpPowerConfigs(const std::thread::id& thread_id) {
+  std::lock_guard<std::mutex> lock(per_thread_power_configs_mutex_);
+  auto res = per_thread_power_configs_.find(thread_id);
+  if (res != per_thread_power_configs_.end()) {
+    per_thread_power_configs_.erase(thread_id);
```
edgchen1: can just directly call `unordered_map::erase(key)` without first checking that `key` is present.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
```diff
@@ -172,6 +173,13 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
                             uint32_t rpc_control_latency,
                             uint32_t rpc_polling_time);
 
+  Status SetPerThreadHtpPowerConfigs(const std::thread::id& theard_id, bool post_run);
+
+  Status AddPerThreadHtpPowerConfigMapping(const std::thread::id& thread_id,
+                                           const PerThreadHtpPowerConfigs_t& htp_power_configs);
+
+  void RemovePerThreadHtpPowerConfigs(const std::thread::id& thread_id);
```
edgchen1: maybe naming this `RemovePerThreadHtpPowerConfigMapping` would be more consistent with the other functions that manipulate `per_thread_power_configs_`

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
```diff
@@ -172,6 +173,13 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
                             uint32_t rpc_control_latency,
                             uint32_t rpc_polling_time);
 
+  Status SetPerThreadHtpPowerConfigs(const std::thread::id& theard_id, bool post_run);
```
edgchen1:
```suggestion
  Status SetPerThreadHtpPowerConfigs(const std::thread::id& thread_id, bool pre_run);
```

onnxruntime/core/providers/qnn/builder/qnn_def.h
```diff
@@ -71,6 +71,19 @@ enum class HtpPerformanceMode : uint8_t {
   kHtpExtremePowerSaver,
 };
 
+typedef struct PerThreadHtpPowerConfigs {
+  HtpPerformanceMode pre_run_perf_mode = HtpPerformanceMode::kHtpDefault;
+  HtpPerformanceMode post_run_perf_mode = HtpPerformanceMode::kHtpDefault;
+  uint32_t rpc_control_latency = 0;
+  uint32_t rpc_polling_time = 0;
+
+  bool pre_run_perf_mode_set = false;
```
edgchen1: instead of an extra bool to indicate whether something is set, consider std::optional

onnxruntime/core/providers/qnn/qnn_execution_provider.h
```diff
@@ -82,6 +83,32 @@ class QNNExecutionProvider : public IExecutionProvider {
...
+    uint32_t GetHtpPowerConfigId();
+
+   private:
+    bool is_valid_ = false;
+    std::mutex config_id_mutex_;
+
+    uint32_t htp_power_config_id_;
+    std::shared_ptr<qnn::QnnBackendManager> qnn_backend_manager_;
+  };
+
```
edgchen1: why does this need to be a shared_ptr?

I wonder if this can be simplified a bit - can `ManagedHtpPowerConfigId` be replaced with an `optional<uint32_t>` to hold the HTP power config ID? the "management" of cleaning up the power config ID can just be done in the QNN EP dtor, where `managed_htp_power_config_id_` is being reset now.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
```diff
@@ -464,6 +475,10 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
   uint32_t soc_model_ = QNN_SOC_MODEL_UNKNOWN;
   const std::vector<OpPackage> op_packages_;
   bool skip_qnn_version_check_ = false;
+
+  // Mapping of thread id to on-run-start/end power configs
+  std::mutex per_thread_power_configs_mutex_;
+  std::unordered_map<std::thread::id, PerThreadHtpPowerConfigs_t> per_thread_power_configs_;
```
edgchen1, line 481: is thread ID sufficient as a key? can one QnnBackendManager be shared between multiple EP instances (i.e., ORT sessions)? if so, it's possible that a single thread could be used to call `Run()` on multiple sessions.

quic-calvnguy: If a single thread is calling Run() on multiple sessions with shared context enabled (single QnnBackendManager across multiple sessions), then the calls will be sequential, so there should be no conflicts with the power configs for each Run() call using the same thread id. If Run() happens to be asynchronous, then I expect there to be a new thread spawned for that separate Run call. In this case, there should be no conflicts due to a unique key.

For the sake of clarity, the power configs stored in this map are those associated with the keys:
- qnn.htp_perf_mode
- qnn.htp_perf_mode_post_run
- qnn.rpc_control_latency

Previously, these values were meant to be set on OnRunStart() and OnRunEnd(). Now the values are stored during OnRunStart() but actually set directly before and after the actual execution within an execution mutex (also existed previously). From what I can tell, if there is an issue with this new implementation, there should also have been an issue with the previous implementation as well.

Additionally, these keys are meant to be thread-specific for multithreaded Run() calls on the same session. They mutually exclusive from the EP options used during session creation as well as the EP dynamic option:
- htp_performance_mode
- htp_graph_finalization_optimization_mode
- ep.dynamic.qnn_htp_performance_mode

edgchen1: ok, I think the handling in OnRunStart() and OnRunEnd() should ensure that the right power config is used. thanks for the clarification.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1555,6 +1555,61 @@ Status QnnBackendManager::SetRpcPowerConfigs(uint32_t htp_power_config_client_id
...
+    }
+
+    if (htp_power_configs.rpc_configs.has_value()) {
+      ORT_RETURN_IF_ERROR(SetRpcPowerConfigs(htp_power_config_id,
+                                             htp_power_configs.rpc_configs.value().rpc_control_latency,
+                                             htp_power_configs.rpc_configs.value().rpc_polling_time));
+    }
+  } else if (htp_power_configs.post_run_perf_mode.has_value()) {
+    ORT_RETURN_IF_ERROR(SetHtpPowerConfig(htp_power_config_id, htp_power_configs.post_run_perf_mode.value()));
+  }
```
edgchen1, line 1577:
```suggestion
    if (htp_power_configs.pre_run_perf_mode.has_value()) {
      ORT_RETURN_IF_ERROR(SetHtpPowerConfig(htp_power_config_id, *htp_power_configs.pre_run_perf_mode));
    }

    if (htp_power_configs.rpc_configs.has_value()) {
      ORT_RETURN_IF_ERROR(SetRpcPowerConfigs(htp_power_config_id,
                                             htp_power_configs.rpc_configs->rpc_control_latency,
                                             htp_power_configs.rpc_configs->rpc_polling_time));
    }
  } else if (htp_power_configs.post_run_perf_mode.has_value()) {
    ORT_RETURN_IF_ERROR(SetHtpPowerConfig(htp_power_config_id, *htp_power_configs.post_run_perf_mode));
  }
```

nit: optional::value() checks whether it contains a value. that's not necessary here because of the earlier call to optional::has_value().

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -649,12 +649,14 @@ QNNExecutionProvider::QNNExecutionProvider(const ProviderOptions& provider_optio
...
-    const auto cache = cache_weak.lock();
-    if (!cache) continue;
-    ORT_IGNORE_RETURN_VALUE(cache->erase(this));
+  if (qnn_backend_manager_) {
+    auto thread_id = std::this_thread::get_id();
+    qnn_backend_manager_->RemovePerThreadHtpPowerConfigMapping(thread_id);
+
+    std::lock_guard<std::mutex> lock(config_id_mutex_);
+    if (htp_power_config_id_.has_value()) {
+      ORT_IGNORE_RETURN_VALUE(qnn_backend_manager_->DestroyHTPPowerConfigID(htp_power_config_id_.value()));
```
edgchen1:
```suggestion
      ORT_IGNORE_RETURN_VALUE(qnn_backend_manager_->DestroyHTPPowerConfigID(*htp_power_config_id_));
```

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1609,4 +1537,40 @@ Status QNNExecutionProvider::SetEpDynamicOptions(gsl::span<const char* const> ke
   return Status::OK();
 }
 
+bool QNNExecutionProvider::GetHtpPowerConfigId(uint32_t& htp_power_config_id) {
+  std::lock_guard<std::mutex> lock(config_id_mutex_);
+  if (!htp_power_config_id_.has_value()) {
+    return false;
+  }
+
+  htp_power_config_id = htp_power_config_id_.value();
```
edgchen1:
```suggestion
  htp_power_config_id = *htp_power_config_id_;
```

onnxruntime/core/providers/qnn/qnn_execution_provider.h
```diff
@@ -82,6 +83,15 @@ class QNNExecutionProvider : public IExecutionProvider {
   bool IsHtpSharedMemoryAllocatorAvailable() const { return rpcmem_library_ != nullptr; }
 
  private:
+  qnn::PerThreadHtpPowerConfigs_t GetPerThreadHtpPowerConfigs(const ConfigOptions& config_options);
+
+  void CreateHtpPowerConfigId() const;
+  // Will return false if htp_power_config_id_ has no value
+  bool GetHtpPowerConfigId(uint32_t& htp_power_config_id);
+
+  mutable std::optional<uint32_t> htp_power_config_id_;
```
edgchen1, line 95: maybe it would be good to have a comment explaining why this is mutable

quic-tirupath:
@edgchen1 
Calvin updated the PR with addressing the comments posted on previous version. Could you please help to review, approve and trigger CI job?

PR #26461, opened 2025-10-31
[WebGPU] Implement Split-K on Conv|MatMul

Jiawei-Shao:
### Description
This patch implements the `Split-K` optimization on `Conv|MatMul`. With `Split-K` we can re-arrange the computation into multiple workgroups when `K` is large to increase the parallelism on the platforms that `Split-K` is confirmed to be useful.

1. Support `Split-K` in `MakeMatMulPackedVec4Source()` to split a workgroup with large K into smaller ones. In this patch we only support `Split-K` with `batch_size == 1` and `vec4` on `Conv|MatMul`.
2. Support `Split-K` in `MatMulWriteFnSource()` (add the partial result to output with atomic built-in functions)
3. Implement `SplitKConfig` to decide whether `Split-K` should be used or not, and all the related thresholds.
4. Implement `MatMulFillBiasBeforeSplitKProgram` to initialize the output with `bias` or 0 when `Split-K` is used.

### Motivation and Context
In current implementation, when `K` or `dim_inner` is large, in each invocation we always do the computation one by one in a very large loop, which may not make full use of all EUs on a GPU.

With `Split-K` we can split such large amount of computation (`K`) into multiple workgroups with less computation (`kSplitK`, smaller than K), which can greatly improve the parallelism.

With this patch we can get about 15% performance improvement on `efficientnet-lite-f16-demo` and 9% improvement on `mobilenetv2-12-f16-demo` on Lunar Lake and Meteor Lake.

### DIFF
--- a/onnxruntime/test/providers/cpu/nn/conv_fp16_test.cc
+++ b/onnxruntime/test/providers/cpu/nn/conv_fp16_test.cc
@@ -45,7 +45,8 @@ void TestConvFp16Op(const ConvOpAndTestAttributes& attributes,
                     bool weight_is_initializer = false,
                     OpTester::ExpectResult expect_result = OpTester::ExpectResult::kExpectSuccess,
                     const std::string& err_str = "",
-                    int opset = 11) {
+                    int opset = 11,
+                    float rel_error = 0.002f) {
   std::unique_ptr<OpTester> tester;
   if (!attributes.activation.empty()) {
     tester = std::make_unique<OpTester>("NhwcFusedConv", 1, onnxruntime::kMSDomain);
@@ -85,7 +86,7 @@ void TestConvFp16Op(const ConvOpAndTestAttributes& attributes,
   if (inputs.size() >= 4)
     tester->AddInput<MLFloat16>(szNames[3], input_shapes[3], inputs[3]);
 
-  tester->AddOutput<MLFloat16>("Y", expected_output_shape, expected_output, /*no sort*/ false, 0.002f, 0.0f);
+  tester->AddOutput<MLFloat16>("Y", expected_output_shape, expected_output, /*no sort*/ false, rel_error, 0.0f);
 
   std::unordered_set<std::string> excluded_providers(attributes.excluded_providers);
   // Disable TensorRT because weight as input is not supported
@@ -525,14 +526,16 @@ TEST(ConvFp16Test, Conv2D_MatMul_SplitK_With_Bias) {
   }
   vector<MLFloat16> expected_vals = FloatsToMLFloat16s(expected_vals_float32);
 
+  // Using a higher relative error threshold for the Linux arm64 bots
+  constexpr float rel_error = 0.02f;
   TestConvFp16Op(
       attrs, {X, W, B}, {X_shape, W_shape, B_shape}, expected_vals, Y_shape, false,
-      OpTester::ExpectResult::kExpectSuccess, "", 11);
+      OpTester::ExpectResult::kExpectSuccess, "", 11, rel_error);
 
   // NNAPI/CoreML EP requires weight to be an initializer
   TestConvFp16Op(
       attrs, {X, W, B}, {X_shape, W_shape, B_shape}, expected_vals, Y_shape, true,
-      OpTester::ExpectResult::kExpectSuccess, "", 11);
+      OpTester::ExpectResult::kExpectSuccess, "", 11, rel_error);
 }
 
 TEST(ConvFp16Test, Conv2D_Bias_1) {


Jiawei-Shao:
@jchen10 @xhcao 
This is may first PR to support `Split-K` in `Conv|MatMul`. PTAL, thanks!

onnxruntime/core/providers/webgpu/math/gemm_utils.cc
```diff
@@ -208,14 +278,51 @@ Status MakeMatMulPackedVec4Source(ShaderHelper& shader,
...
-      << "  let batch = i32(global_id.z);\n"
-      << (nullptr != batch_dims ? "  let batchIndices = " + batch_dims->OffsetToIndices("u32(batch)") + ";\n" : "")
       << "  let globalRowStart = i32(workgroup_id.y) * " << tile_a_outer << ";\n"
       << "  let globalColStart = i32(workgroup_id.x) * " << tile_b_outer << ";\n"
-      << "  let num_tiles = (uniforms.dim_inner - 1) / tileInner + 1;\n"
-      << "  var kStart = 0;\n"
       << "  var acc: array<vec4<" << data_type << ">, rowPerThread>;\n";
 
+  if (split_k) {
+    // With Split-K, the original "workgroup" (with dispatch_z == 1 in API side) into multiple
```
jchen10: the original "workgroup" (with dispatch_z == 1 in API side) **is split** into multiple

onnxruntime/core/providers/webgpu/math/gemm_utils.cc
```diff
@@ -208,14 +278,51 @@ Status MakeMatMulPackedVec4Source(ShaderHelper& shader,
...
+    // ones, and in the current workgroup we only compute `kSplitK` elements starting from
+    // `kSplitK * i32(global_id.z)`.
+    //
+    //  For example: considering computing Y = (X * W + B) in one workgroup.
+    //  Let kSplitk = 2, B = [d1, d2]
+    //  Let X = [[a1 a1 b1 b1 c1 c1]  = [ A1 B1 C1 ] W = [[a2 a2]  = [ A2
```
jchen10:
```suggestion
    //  Let X = [[a1 a1 b1 b1 c1 c1]  = [ A1 B1 C1 ], W = [[a2 a2]  = [ A2
```

jchen10:
LGTM, thanks!

fs-eire:
I have a general question regarding "reduce" operation in WebGPU. In my understanding, there are general 2 ways to implement a reduce operation:
- use atomic premitive, specifically `atomicCompareExchangeWeak` in this case, just like this PR
- use tree reduction (implemented in [reduction_ops.cc](https://github.com/microsoft/onnxruntime/blob/1fb03cc45844e0d463cf0355ade2be4566558dd3/onnxruntime/core/providers/webgpu/reduction/reduction_ops.cc#L265-L276)) (is it possible to apply this algo to Split-K?)

In general, what is the pros vs. cons for the 2 different approaches?

Jiawei-Shao:
> I have a general question regarding "reduce" operation in WebGPU. In my understanding, there are general 2 ways to implement a reduce operation:
> 
> * use atomic primitive, specifically `atomicCompareExchangeWeak` in this case, just like this PR
> * use tree reduction (implemented in [reduction_ops.cc](https://github.com/microsoft/onnxruntime/blob/1fb03cc45844e0d463cf0355ade2be4566558dd3/onnxruntime/core/providers/webgpu/reduction/reduction_ops.cc#L265-L276)) (is it possible to apply this algo to Split-K?)
> 
> In general, what is the pros vs. cons for the 2 different approaches?

I think `tree reduction` works best when we only need to increase parallelism by increasing the number of invocations in one workgroup, while `Split-K` is used when we not only want to increase the number of invocations, but also want to use more workgroups and shared local memory on the fly. On modern Intel GPUs dispatching more workgroups means utilizing more Xe cores (each Xe core has its own vector engines and Shared Local Memory), and that's why we see performance improvements with `Split-K` on the matrix multiplications with large K.

Jiawei-Shao:
@fs-eire @guschmue @qjia7 PTAL, thanks!

onnxruntime/core/providers/webgpu/math/matmul_packed.cc
```diff
@@ -23,16 +31,61 @@ Status MatMulProgram::GenerateShaderCode(ShaderHelper& shader) const {
...
+  let global_row = global_id.y;
+  if (global_row >= uniforms.dim_a_outer) {
+    return;
+  }
+  let dim_b_outer = i32(uniforms.dim_b_outer) / output_components;
+  let batch = 0;
+  let row = i32(global_row);
+  let value = output_value_t();
+  let start_col = i32(global_id.x) * elements_per_thread;
+  for (var col = start_col; col < start_col + elements_per_thread; col++) {
```
qjia7: Why not just simple write one vec4 result for each thread and use the 1D dispatch in program side? I think it has better performance for small output.

onnxruntime/core/providers/webgpu/math/matmul_packed.h
```diff
@@ -13,24 +13,52 @@ namespace onnxruntime {
...
   const InlinedVector<int64_t> elements_per_thread_;
   bool is_channels_last_ = false;
+  uint32_t split_dim_inner_ = 1;
+};
+
+// The program to initialize the output with 0 or bias before doing MatMul with Split-K. In Split-K,
+// we set the output values with `atomicLoad` and `atomicCompareExchangeWeak` instead of a direct
+// assignment (see the function `HandleMatMulWithSplitK()` in `gemm_utils.cc`), so we must initialize
+// the output with 0 or bias first to make sure `atomicLoad` won't return garbage data.
+class MatMulFillBiasBeforeSplitKProgram final : public Program<MatMulFillBiasBeforeSplitKProgram> {
```
qjia7: Rename to MatMulFillBiasOrZeroBeforeSplitKProgram

onnxruntime/core/providers/webgpu/nn/conv.cc
```diff
@@ -200,7 +200,18 @@ Status Conv<is_channels_last, is_fused>::ComputeInternal(ComputeContext& context
...
+      // TODO: enable Split-K in all the places that call `CreateMatMulProgram()`.
+      SplitKConfig split_K_config = SplitKConfig::GetSplitKConfig(context);
+      MatMulProgram program = CreateMatMulProgram(activation_, matmul_inputs, output, is_channels_last, split_K_config, matmul_input_reshapes[0], matmul_input_reshapes[1]);
+      if (program.NeedSplitK()) {
+        // Currently we only support bias in vec4 and channels last format for Split-K MatMul.
+        assert(is_channels_last);
+        MatMulFillBiasBeforeSplitKProgram fill_bias_program = CreateMatMulFillBiasBeforeSplitKProgram(
+            bias, output, matmul_input_reshapes[0], matmul_input_reshapes[1]);
+        ORT_RETURN_IF_ERROR(context.RunProgram(fill_bias_program));
+      }
```
qjia7: I prefer keeping this part unchanged and do the fill_bias_program part in `CreateMatMulProgram`'s implementation if possible. Maybe pass context into CreateMatMulProgram. Rename CreateMatMulProgram to ComputeMatMul so that you can directly call `context.RunProgram(program)` at the end of `ComputeMatMul`.

onnxruntime/core/providers/webgpu/shader_helper.cc
```diff
@@ -465,6 +465,11 @@ Status ShaderHelper::GenerateSourceCode(std::string& code, std::vector<int>& sha
   for (size_t i = 0; i < output_vars_.size(); ++i) {
     const auto& output = output_vars_[i];
     bool is_atomic = program_.Outputs()[i].is_atomic;
+    ProgramVariableDataType atomic_type = output->type_;
+    if (output->usage_ & ShaderUsage::UseAtomicU32ForSplitK) {
+      is_atomic = true;
+      atomic_type = ProgramVariableDataType::Uint32;
+    }
```
qjia7: why can't reuse the existed atomic support by 'program.AddOutput({output_tensor, ProgramTensorMetadataDependency::TypeAndRank, ProgramOutput::Atomic});' in program?

onnxruntime/core/providers/webgpu/math/matmul.cc
```diff
@@ -254,5 +319,52 @@ MatMulProgram CreateMatMulProgram(const Activation& activation, std::vector<cons
...
+  const uint32_t dispatch_x = (output_col + workgroup_size_x * elements_per_thread - 1) / (workgroup_size_x * elements_per_thread);
+  const uint32_t dispatch_y = (output_row + workgroup_size_y - 1) / workgroup_size_y;
+
+  // TODO: support batch_size > 1
+  constexpr uint32_t batch_size = 1;
+  TensorShape output_shape_temp = TensorShape({batch_size, output_row, output_col});
+
+  const uint32_t data_type = output->GetElementType();
+  program.CacheHint(has_bias, data_type)
+      .AddOutput({output, ProgramTensorMetadataDependency::Rank, output_shape_temp, static_cast<int32_t>(bias_components)})
```
qjia7: If the `output_shape_temp` is only used to dividing the components in the last dimension. You can directly use `.AddOutput({output, ProgramTensorMetadataDependency::TypeAndRank, static_cast<int32_t>(bias_components)})`. It will help you do it underlying. `TypeAndRank` can let you avoid adding `data_type` in the `CacheHint`.  Similar for other places.

onnxruntime/core/providers/webgpu/math/matmul.cc
```diff
@@ -222,21 +274,34 @@ MatMulProgram CreateMatMulProgram(const Activation& activation, std::vector<cons
...
   const TensorShape output_shape_temp = TensorShape({batch_size, dim_a_outer, dim_b_outer / components});
 
-  MatMulProgram program{activation, has_bias, is_vec4, elements_per_thread, is_channels_last};
+  MatMulProgram program{activation, has_bias, is_vec4, elements_per_thread, is_channels_last, split_dim_inner};
   program
-      .CacheHint(activation.ToString(), absl::StrJoin(elements_per_thread, "-"), std::to_string(is_vec4), components, is_channels_last)
+      .CacheHint(activation.ToString(), absl::StrJoin(elements_per_thread, "-"), std::to_string(is_vec4), components, is_channels_last, split_dim_inner)
       .AddInputs({{a, ProgramTensorMetadataDependency::TypeAndRank, a_shape_temp, components},
                   {b, ProgramTensorMetadataDependency::TypeAndRank, b_shape_temp, components}})
       .AddOutputs({{output_tensor, ProgramTensorMetadataDependency::Rank, output_shape_temp, components}})
```
Jiawei-Shao: [1]

onnxruntime/core/providers/webgpu/math/matmul.cc
Jiawei-Shao: Oh I find we must use `output_shape_temp` here because the output tensor here is reshaped as what we do in [1].

onnxruntime/core/providers/webgpu/shader_helper.cc
Jiawei-Shao: Done

onnxruntime/core/providers/webgpu/nn/conv.cc
Jiawei-Shao: Done

onnxruntime/core/providers/webgpu/math/matmul_packed.h
Jiawei-Shao: Done

onnxruntime/core/providers/webgpu/math/matmul_packed.cc
Jiawei-Shao: Done

onnxruntime/core/providers/webgpu/nn/conv.cc
```diff
@@ -200,8 +200,11 @@ Status Conv<is_channels_last, is_fused>::ComputeInternal(ComputeContext& context
           .AddUniformVariables({{output_size}, {static_cast<uint32_t>(matmul_output_shape[1])}, {static_cast<uint32_t>(matmul_output_shape[2])}, {static_cast<uint32_t>(K)}});
       return context.RunProgram(program);
     } else {
-      MatMulProgram program = CreateMatMulProgram(activation_, matmul_inputs, output, is_channels_last, matmul_input_reshapes[0], matmul_input_reshapes[1]);
-      return context.RunProgram(program);
+      // Explicitly pass `SplitKConfig` to `ComputeMatMul()` to enable Split-K. Now it is not
+      // used in any other places that call `ComputeMatMul()` (e.g. in `MatMul::ComputeInternal()`).
+      // TODO: enable Split-K in all the places that call `ComputeMatMul()`.
+      SplitKConfig split_K_config = SplitKConfig::GetSplitKConfig(context);
+      return ComputeMatMul(&context, activation_, matmul_inputs, output, is_channels_last, split_K_config, matmul_input_reshapes[0], matmul_input_reshapes[1]);
```
qjia7: Can we move `SplitKConfig split_K_config = SplitKConfig::GetSplitKConfig(context);` into the body of `ComputeMatMul`?

onnxruntime/core/providers/webgpu/math/matmul.cc
```diff
@@ -222,35 +279,114 @@ MatMulProgram CreateMatMulProgram(const Activation& activation, std::vector<cons
                                                    ? InlinedVector<int64_t>({4, 1, 1})
                                                    : InlinedVector<int64_t>({4, 4, 1});
 
+  bool need_split_k = split_k_config.UseSplitK(is_vec4, activation.activation_kind_, batch_size, is_channels_last, dim_a_outer, dim_b_outer, dim_inner);
+
+  bool use_bias_in_matmul = has_bias;
+
+  // When Split-K is used, bias will be handled in `MatMulFillBiasBeforeSplitKProgram`
+  // instead of `MatMulProgram`.
+  if (need_split_k) {
```
qjia7: Merge all things together when need_split_k is true as possible? Currently, you have 4 `if (need_split_k)`.

onnxruntime/core/providers/webgpu/math/matmul.cc
```diff
@@ -222,35 +279,114 @@ MatMulProgram CreateMatMulProgram(const Activation& activation, std::vector<cons
...
   const uint32_t dispatch_y = narrow<uint32_t>((dim_a_outer + MatMul::MATMUL_PACKED_WORKGROUP_SIZE_Y * elements_per_thread[1] - 1) /
                                                (MatMul::MATMUL_PACKED_WORKGROUP_SIZE_Y * elements_per_thread[1]));
-  const uint32_t dispatch_z = narrow<uint32_t>((static_cast<uint32_t>(batch_size) + MatMul::MATMUL_PACKED_WORKGROUP_SIZE_Z * elements_per_thread[2] - 1) /
-                                               (MatMul::MATMUL_PACKED_WORKGROUP_SIZE_Z * elements_per_thread[2]));
+  uint32_t dispatch_z = narrow<uint32_t>((static_cast<uint32_t>(batch_size) + MatMul::MATMUL_PACKED_WORKGROUP_SIZE_Z * elements_per_thread[2] - 1) /
+                                         (MatMul::MATMUL_PACKED_WORKGROUP_SIZE_Z * elements_per_thread[2]));
+  uint32_t split_dim_inner = 1;
+  if (need_split_k) {
+    split_dim_inner = split_k_config.GetSplitDimInner();
+    dispatch_z = (dim_inner + split_dim_inner - 1) / split_dim_inner;
```
qjia7: Not introduced by your PR. But please be careful that when the dispatch size is normalized (ProgramManager::NormalizeDispatchGroupSize) all things will be wrong. Maybe add todo here to have a comprehensive solution to fix it in follow-ups. Currently, flash attention programs and matmulnbits programs are all trying to avoid using `global_id.xxx` or `workgroup_id.xxx`.

onnxruntime/core/providers/webgpu/math/matmul.cc
```diff
@@ -222,35 +279,114 @@ MatMulProgram CreateMatMulProgram(const Activation& activation, std::vector<cons
...
+  }
+
+  return context->RunProgram(matmul_program);
+}
+
+MatMulFillBiasOrZeroBeforeSplitKProgram CreateMatMulFillBiasOrZeroBeforeSplitKProgram(
+    const Tensor* bias,
+    Tensor* output,
+    const TensorShape& input_a_shape,
+    const TensorShape& input_b_shape) {
```
qjia7: Maybe directly pass the output_tensor_reshaped_shape instead of input_a_shape/input_b_shape to simplify this function.

onnxruntime/core/providers/webgpu/math/matmul_packed.cc
```diff
@@ -14,25 +14,77 @@ namespace webgpu {
...
+  // Here `use_split_k` is false because we just initialize `output` with bias.
+  // `use_split_k` is true only when we do the actual MatMul with Split-K.
+  // Currently we only support bias in vec4 and channels last format for Split-K MatMul.
+  MatMulWriteFnSource(
+      shader, output, has_bias_, /*is_gemm*/ false, /*c_components*/ 4, /*output_components*/ 4, /*c_is_scalar*/ false,
+      /*activation_snippet*/ "", /*is_channels_last*/ true, /*use_split_k*/ false);
+
+  shader.MainFunctionBody() << R"(
+  let output_components = 4;
+  let output_id = i32(global_id.x);
```
qjia7: Use `global_idx` instead of `global_id.x`

onnxruntime/core/providers/webgpu/math/matmul.cc
```diff
@@ -46,6 +46,62 @@ static std::string CalcResult(int64_t components, int64_t a_components, int64_t
   return oss.str();
 }
 
+SplitKConfig SplitKConfig::GetSplitKConfig(const ComputeContext& context) {
```
qjia7: SplitKConfig::GetSplitKConfig is called repeatedly with the same result for every MatMul operation and also by Conv operations. Maybe cache it in WebGpuExecutionProvider and do lazy initialization.

onnxruntime/core/providers/webgpu/math/matmul.cc
```diff
@@ -46,6 +46,62 @@ static std::string CalcResult(int64_t components, int64_t a_components, int64_t
...
+
+bool SplitKConfig::UseSplitK(
+    bool is_vec4,
+    ActivationKind activation_kind,
+    uint64_t batch_size,
+    bool is_channels_last,
+    uint32_t dim_a_outer,
+    uint32_t dim_b_outer,
+    uint32_t dim_inner) const {
+  bool use_split_k = enable_split_k_;
```
qjia7: early return if enable_split_k_ is false.

onnxruntime/core/providers/webgpu/math/matmul.cc
```diff
@@ -226,31 +282,99 @@ MatMulProgram CreateMatMulProgram(const Activation& activation, std::vector<cons
...
+  constexpr uint32_t bias_components = 4;
+  MatMulFillBiasOrZeroBeforeSplitKProgram program(has_bias);
+
+  const uint32_t dim_a_outer = narrow<uint32_t>(output_shape_vec4[output_shape_vec4.NumDimensions() - 2]);
+  const uint32_t dim_b_outer_vec4 = narrow<uint32_t>(output_shape_vec4[output_shape_vec4.NumDimensions() - 1]);
+
+  // Fill one value (currently only vec4) per invocation.
+  constexpr uint32_t workgroup_size_x = MatMulFillBiasOrZeroBeforeSplitKProgram::WORKGROUP_SIZE_X;
+  const uint32_t total_outputs_vec4 = dim_a_outer * dim_b_outer_vec4;
+  const uint32_t dispatch_x = (total_outputs_vec4 + workgroup_size_x - 1) / workgroup_size_x;
```
qjia7: You can directly use `WORKGROUP_SIZE` which is defined in the base Program. So there is no need to add another variable workgroup_size_x/MatMulFillBiasOrZeroBeforeSplitKProgram::WORKGROUP_SIZE_X

onnxruntime/core/providers/webgpu/math/matmul.cc
```diff
@@ -226,31 +282,99 @@ MatMulProgram CreateMatMulProgram(const Activation& activation, std::vector<cons
...
+
+  // To reuse `MatMulWriteFnSource()` we need to set `dim_a_outer` and `dim_b_outer` in scalar
+  // instead of vec4, while use `output_shape_vec4` directly as the output shape.
+  const uint32_t dim_b_outer = narrow<uint32_t>(dim_b_outer_vec4 * bias_components);
+  program.CacheHint(has_bias)
+      .AddOutput({output, ProgramTensorMetadataDependency::TypeAndRank, output_shape_vec4, static_cast<int32_t>(bias_components)})
+      .AddUniformVariables({{dim_a_outer}, {dim_b_outer}})
+      .SetDispatchGroupSize(dispatch_x, 1, 1)
+      .SetWorkgroupSize(workgroup_size_x, 1, 1);
```
qjia7: 64 is the default one. So you don't need to set it if you use 64.

fs-eire:
May need merge or rebase to latest main branch to include the fix to the CI pipeline.

Jiawei-Shao:
> No more comments for me. Please refactor the related shaders to eliminate any reliance on global_id.xxx or workgroup_id.xxx in your follow-up PRs. This dependency presents a potential risk and could lead to unforeseen issues.

Got it. We will refactor this in the next PR.

guschmue:
ran some test. lgtm. I see gains on prefill for nvidia. I'll capture complete numbers later today or tomorrow.

Jiawei-Shao:
Thanks for your review! Can we merge this PR?

onnxruntime/core/providers/webgpu/compute_context.h
```diff
@@ -139,6 +139,13 @@ class ComputeContext final {
     return webgpu_context_.Run(*this, program);
   }
 
+  //
+  // Get Split-K configuration.
+  //
+  // `split_k_config_` won't be initialized until the first call to this method.
```
fs-eire, line 145: Is there a special consideration for lazy-initializing the split_k_config_?

Jiawei-Shao: Currently `split_k_config_` is only used in a few operators on certain platform. With lazy initialization we can avoid allocating it when it is not used (for example, when we only want to execute LLMs).

PR #26462, opened 2025-10-31
coreml: Add a few missing operators

borg323:
### Description
Adds `Elu`, `Exp` and `Softplus` to coreml.

### Motivation and Context

We use them in LeelaChessZero.


borg323:

snnn:
Cool project!

snnn:
Some tests failed:
```
 Initialize failed but expected success: Error compiling model: Failed to parse the model specification. Error: Unable to parse ML Program: in operation node1: Required param 'epsilon' is missing
```

borg323:
I removed `Log` for now as it was for future extensions, will make a better implementation when we need it.

PR #26463, opened 2025-10-31
Update out-of-support net8.0 MAUI targets to net9.0

edgchen1:
### Description

.NET MAUI 8 is out of support. See here: https://dotnet.microsoft.com/en-us/platform/support/policy/maui

We started seeing errors about this in the NuGet packaging pipeline.
```
##[error]C:\Program Files\dotnet\sdk\9.0.306\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.EolTargetFrameworks.targets(38,5): Error NETSDK1202: The workload 'net8.0-ios' is out of support and will not receive security updates in the future. Please refer to https://aka.ms/maui-support-policy for more information about the support policy.
```

This change updates net8.0 mobile target framework monikers to net9.0.

### Motivation and Context

Fix packaging pipeline.

### DIFF
--- a/tools/nuget/generate_nuspec_for_native_nuget.py
+++ b/tools/nuget/generate_nuspec_for_native_nuget.py
@@ -8,6 +8,8 @@ import shutil
 import sys
 from pathlib import Path
 
+# These platform version values are the default target platform versions for .NET 9 from the table here:
+# https://learn.microsoft.com/en-us/dotnet/standard/frameworks#os-version-in-tfms
 platform_version_android = "35.0"
 platform_version_ios = "18.0"
 platform_version_maccatalyst = "18.0"


tools/nuget/generate_nuspec_for_native_nuget.py
```diff
@@ -4,9 +4,14 @@
 import argparse
 import os
 import re
+import shutil
 import sys
 from pathlib import Path
 
+platform_version_android = "35.0"
+platform_version_ios = "18.0"
+platform_version_maccatalyst = "18.0"
```
edgchen1, line 15: these values are the default versions for .NET 9 in the table here:
https://learn.microsoft.com/en-us/dotnet/standard/frameworks#os-version-in-tfms

skottmckay: Excellent info. Can we put it in a comment so the next developer to read the file knows how/why these values were picked without having to trawl the commit history, find this PR and look for comments explaining the choice?

edgchen1:
> LGTM. But, actually I am not able to build the code with .net9 on Linux ( I didn't enable mobile). It's due to the *.snk files . The files were generated years ago with a weak RSA key. Now openssl rejects such low-strength RSA algorithms. So, it cannot sign the files. So, `dotnet build` cannot generate the managed DLLs.

I suppose we'll have to deal with it when upgrading the other target frameworks to net9.0. currently, only the MAUI net8.0 ones are out of support.

snnn (approved):

PR #26464, opened 2025-10-31
[QNN-EP] Add skip_qnn_version_check option to QNN EP

quic-muchhsu:
### Description

Add `skip_qnn_version_check` provider_option to QNN EP to allow user to skip the check and use different QNN libraries.

### Motivation and Context

This will allow QNN EP users to provide different QNN backend library than the default ones and run ONNX Runtime with them.


quic-muchhsu:
I don't see merge bottom from my end, can anyone help me merge it? @edgchen1 Thanks!

PR #26465, opened 2025-10-31
[QNN-EP] Use same power votes as burst for SHP

quic-calvnguy:
### Description
Use same HTP voltage corner votes as burst for sustained_high_performance

### Motivation and Context
Increases performance without DSPQ polling and utilizes the expected voltage mappings

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
@@ -1383,6 +1383,7 @@ Status QnnBackendManager::SetHtpPowerConfig(uint32_t htp_power_config_client_id,
   // choose performance mode
   switch (htp_performance_mode) {
     case HtpPerformanceMode::kHtpBurst:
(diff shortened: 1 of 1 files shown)


onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1383,6 +1383,7 @@ Status QnnBackendManager::SetHtpPowerConfig(uint32_t htp_power_config_client_id,
   // choose performance mode
   switch (htp_performance_mode) {
     case HtpPerformanceMode::kHtpBurst:
+    case HtpPerformanceMode::kHtpSustainedHighPerformance:
```
edgchen1, line 1386: this seems to be changing the behavior of kHtpSustainedHighPerformance. was it incorrect before?

quic-calvnguy: With recent updates for V81, the NPU bus clock frequency is not set appropriately unless sustained high performance uses the same power voting as burst. This is a temporary fix for now--an official fix is still in discussion

PR #26468, opened 2025-11-01
[QNN-EP] Disable DSPQ polling when needed

quic-calvnguy:
### Description
Create new class to handle HTP power config updates
Only update if there are changes to power config settings
On dynamic htp perf mode updates, disable DSPQ polling if perf mode is not burst

### Motivation and Context
Currently, if a session has set the performance mode to burst then changed the performance mode to anything else, DSPQ polling will be enabled and never disabled. This change is to allow disabling of DSPQ polling when the performance mode is not burst, even on updates.

### DIFF
--- a/onnxruntime/core/providers/qnn/qnn_execution_provider.cc
+++ b/onnxruntime/core/providers/qnn/qnn_execution_provider.cc
@@ -1573,7 +1573,7 @@ void QNNExecutionProvider::CreateHtpPowerConfigId() const {
 
   Status rt = qnn_backend_manager_->CreateHtpPowerCfgId(device_id_, core_id, htp_power_config_id);
 
-  if (rt == Status::OK()) {
+  if (rt.IsOK()) {
     htp_power_config_id_ = htp_power_config_id;
 
     rt = qnn_backend_manager_->SetHtpPowerConfigs(htp_power_config_id,
@@ -1581,7 +1581,7 @@ void QNNExecutionProvider::CreateHtpPowerConfigId() const {
                                                   default_rpc_polling_time_,
                                                   default_rpc_control_latency_);
 
-    if (rt.IsOK()) {
+    if (!rt.IsOK()) {
       LOGS_DEFAULT(ERROR) << "Unable to set HTP power configurations.";
     }
   } else {


quic-calvnguy:
This PR is based on another PR:
https://github.com/microsoft/onnxruntime/pull/26457

The reason for this is that PR 26457 removes per-thread HTP power config ids and creates these ids on a per-session basis. Implementing this PR's changes without PR 26457's changes would require unnecessary work that would otherwise be removed by PR 26457 anyway.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1561,15 +1576,16 @@ void QNNExecutionProvider::CreateHtpPowerConfigId() const {
...
-    if (default_rpc_control_latency_ > 0 || default_rpc_polling_time_ > 0) {
-      ORT_IGNORE_RETURN_VALUE(qnn_backend_manager_->SetRpcPowerConfigs(htp_power_config_id,
-                                                                       default_rpc_control_latency_,
-                                                                       default_rpc_polling_time_));
+    rt = qnn_backend_manager_->SetHtpPowerConfigs(htp_power_config_id,
+                                                  default_htp_performance_mode_,
+                                                  default_rpc_polling_time_,
+                                                  default_rpc_control_latency_);
+
+    if (rt != Status::OK()) {
```
yuslepukhin: >if (rt != Status::OK())  [](http://example.com/codeflow?start=4&length=24)

nit: IsOK(). #Resolved

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1365,192 +1365,19 @@ Status QnnBackendManager::CreateHtpPowerCfgId(uint32_t device_id, uint32_t core_
...
-
-    std::vector<const QnnHtpPerfInfrastructure_PowerConfig_t*> perf_power_configs_ptr =
-        ObtainNullTermPtrVector(rpc_power_configs);
-    status = htp_perf_infra.setPowerConfig(htp_power_config_client_id, perf_power_configs_ptr.data());
-    ORT_RETURN_IF(QNN_SUCCESS != status, "setPowerConfig failed for RPC control latency.");
-  }
+  ORT_RETURN_IF_ERROR(htp_power_config_manager_.AddRpcPollingTime(rpc_polling_time));
+  ORT_RETURN_IF_ERROR(htp_power_config_manager_.AddRpcControlLatency(rpc_control_latency));
+  ORT_RETURN_IF_ERROR(htp_power_config_manager_.AddHtpPerformanceMode(htp_performance_mode, htp_power_config_client_id));
+  ORT_RETURN_IF_ERROR(htp_power_config_manager_.SetPowerConfig(htp_power_config_client_id, GetQnnInterface()));
```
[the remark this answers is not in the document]
quic-calvnguy: Flags are already set only when Add methods are successful and are cleared only if setpowerconfig is successful (in which case power_configs_ is also empty).

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
yuslepukhin: This is missing negation. `!rt.IsOK()`

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1581,7 +1581,7 @@ void QNNExecutionProvider::CreateHtpPowerConfigId() const {
                                                   default_rpc_polling_time_,
                                                   default_rpc_control_latency_);
 
-    if (rt != Status::OK()) {
+    if (rt.IsOK()) {
```
quic-calvnguy: Bug--will fix

PR #26485, opened 2025-11-04
[build] set ONNX_USE_LITE_PROTO=ON only for wasm when using vcpkg

fs-eire:
### Description

Before this change, ONNX_USE_LITE_PROTO is always set to OFF for all vcpkg builds, including web assembly build. This PR modifies the behavior to use lite proto when building targetting wasm.

No change to non-vcpkg build.


cmake/vcpkg-ports/onnx/portfile.cmake
```diff
@@ -36,7 +36,6 @@ vcpkg_cmake_configure(
         "-DProtobuf_PROTOC_EXECUTABLE:FILEPATH=${PROTOC}"
         -DONNX_ML=ON
         -DONNX_USE_PROTOBUF_SHARED_LIBS=${USE_PROTOBUF_SHARED}
-        -DONNX_USE_LITE_PROTO=OFF
```
snnn, line 39: Good. Since the default value is OFF.

snnn (approved):

snnn:
This PR is a good step towards fixing our vcpkg build configuration for the `onnx` dependency.

For context for everyone, here's the situation this PR addresses:

1.  **The Problem:** Our overlay port in `cmake/vcpkg-ports/onnx/portfile.cmake` incorrectly hardcodes the build option `-DONNX_USE_LITE_PROTO=OFF`. This forces the `onnx` dependency to always be built with the full protobuf library, which breaks ONNX Runtime's default build, as it expects lite protobuf.

2.  **The Fix in this PR:** This PR modifies `tools/python/util/vcpkg_helpers.py` to correctly configure the `onnx` port via the generated vcpkg triplet file.
    *   For a **lite protobuf build** (the default), this PR ensures that `-DONNX_USE_LITE_PROTO=ON` is added to the `onnx` port's configuration. This overrides the incorrect `OFF` setting from the `portfile.cmake` and fixes the default build.
    *   As part of the cleanup, this PR also removes a redundant `-DONNX_USE_LITE_PROTO=OFF` that was being generated for other build configurations.

This correctly addresses the immediate issue for the default build path.

snnn:
To build on the cleanup in this PR, we should consider a more comprehensive fix to make our vcpkg build more robust and transparent.

The current implementation has a known issue where the `--use_full_protobuf` flag in `build.py` does not explicitly control the `onnx` dependency's build configuration. The build works due to a combination of overrides and default behaviors, which is not ideal.

A suggested path for improvement would be:

1.  **Decouple `portfile.cmake` from Configuration:** Remove the hardcoded `-DONNX_USE_LITE_PROTO=OFF` from our overlay port at `cmake/vcpkg-ports/onnx/portfile.cmake`. The portfile should be generic and not enforce a specific build configuration.

2.  **Plumb Configuration from `build.py`:** The `--use_full_protobuf` setting should be passed from `build.py` down into the triplet generation functions in `tools/python/util/vcpkg_helpers.py`.

3.  **Drive Configuration from Triplets:** The triplet generation logic should then use this setting to explicitly add either `-DONNX_USE_LITE_PROTO=ON` (for lite builds) or `-DONNX_USE_LITE_PROTO=OFF` (for full builds) to the `onnx` port's CMake options.

These steps would create a clear and explicit chain of configuration, making the build system easier to maintain and debug. I can create a follow-up issue to track this work.

PR #26491, opened 2025-11-04
[quant] Fix QuantizeLinear opset-23 kernel registration to match ONNX schema

### LINKED ISSUE #25932: Could not find an implementation for QuantizeLinear(23)
### Describe the issue

When exporting a model that contains an Attention block I deliberately target **ONNX opset 23** so that the single `Attention` operator (introduced in opset 23) is kept intact instead of being decomposed into many small primitives. The exported FP32 model runs correctly with the current ONNX Runtime CPU build.  
Afterwards I apply **static INT8 quantization** via `onnxruntime.quantization.quantize_static`. The resulting graph contains `QuantizeLinear` nodes at opset 23. At runtime the session fails to initialize with

```
[ONNXRuntimeError] : 9 : NOT_IMPLEMENTED :
Could not find an implementation for QuantizeLinear(23)
```

The same workflow using **opset 21** works without error.  
Inspection of the ORT source tree shows that the kernels

```
ONNX_OPERATOR_TYPED_KERNEL_CLASS_NAME(kCpuExecutionProvider, kOnnxDomain, 23, uint8_t, QuantizeLinear)
ONNX_OPERATOR_TYPED_KERNEL_CLASS_NAME(kCpuExecutionProvider, kOnnxDomain, 23, int8_t, QuantizeLinear)
```

are indeed registered for opset 23, yet the binary that is loaded at runtime appears to lack them.

### To reproduce

```
import os, torch, numpy as np, onnxruntime as ort
from torch import nn
from onnxruntime.quantization import quantize_static, QuantFormat, QuantType, CalibrationMethod

class TinyConv(nn.Module):
    def __init__(self):
        super().__init__()
        self.conv = nn.Conv2d(1, 8, 3, padding=1)

    def forward(self, x):
        return self.conv(x.unsqueeze(1).unsqueeze(2)).squeeze(2)

os.makedirs('tmp_model', exist_ok=True)
model_path = 'tmp_model/contentvec.onnx'
int8_path  = 'tmp_model/contentvec_int8.onnx'

model = TinyConv().eval()
dummy = torch.randn(1, 128)
torch.onnx.export(model, dummy, model_path,
                  opset_version=23, dynamo=True,
                  input_names=['input_values'], output_names=['hidden_states'])

class DummyReader:
    def __init__(self):
        self.data = [np.random.randn(1, 128).astype(np.float32) for _ in range(4)]
        self.idx = 0
    def get_next(self):
        if self.idx >= len(self.data): return None
        out = {'input_values': self.data[self.idx]}
        self.idx += 1
        return out

quantize_static(model_path, int8_path, DummyReader(),
                quant_format=QuantFormat.QOperator,
                activation_type=QuantType.QUInt8,
                weight_type=QuantType.QInt8,
                calibrate_method=CalibrationMethod.MinMax)

sess = ort.InferenceSession(int8_path, providers=['CPUExecutionProvider'])
print(sess.run(None, {'input_values': np.random.randn(1, 128).astype(np.float32)})[0].shape)
```

### Urgency

[HIGH PRIORITY] Missing CPU Kernel for QuantizeLinear at ONNX Opset 23 After Static Quantization

### Platform

Windows

### OS Version

26100.4946

### ONNX Runtime Installation

Released Package

1.23.0.dev20250902003

### ONNX Runtime API

Python

### Architecture

X64

### Execution Provider

Default CPU

_No response_

rachelElfenbein-dev:
### Description

This PR fixes the type-constraint mapping for QuantizeLinear in opset 23 in the CPU EP kernel registration.
The kernel registration for opset 23 now matches the schema names and allowed types.
[Type Constraints (opset 23)](https://onnx.ai/onnx/operators/onnx__QuantizeLinear.html#type-constraints_23)

### Motivation and Context

Fixes #25932

**Problem**: Models with QuantizeLinear at opset 23 failed session initialization with “Could not find an implementation for QuantizeLinear(23)”, even though a CPU kernel exists.
The previous kernel registration for opset 23 incorrectly constrained the second input (y_scale, input index 1) to the quantized/output type (T).
As a result, at runtime KernelRegistry::TryFindKernel failed to match the kernel .

**Root cause**: The previous opset-23 registration did not mirror the schema’s per-parameter type variables. During KernelRegistry::TryFindKernel, the constraint-to-arg mapping (via KernelTypeStrResolver) compared Y_scale against the quantized type set (or otherwise mismatched sets), causing combinations to be rejected.

**Solution**: Register opset-23 with schema-accurate constraints (T1, T2, T3) so each argument is validated against the correct constraint . It restores the ability to find a matching kernel for valid models and removes the initialization failure.

### DIFF
--- a/onnxruntime/test/providers/cpu/tensor/quantize_linear_test.cc
+++ b/onnxruntime/test/providers/cpu/tensor/quantize_linear_test.cc
@@ -6,6 +6,10 @@
 #include "test/providers/provider_test_utils.h"
 #include "test/util/include/default_providers.h"
 #include "core/framework/int4.h"
+#include "test/util/include/asserts.h"
+#include "core/session/inference_session.h"
+#include "test/util/include/test_environment.h"
+#include "test/util/include/inference_session_wrapper.h"
 
 namespace onnxruntime {
 namespace test {
@@ -3390,6 +3394,16 @@ TEST(QuantizeLinearOp21BlockedTest, Float8_NoZeroPoint_LastAxis) {
   }
 }
 #endif
+
+TEST(QuantizeLinearOpTest, Opset23_FloatScale_PerTensor_Int8) {
+  OpTester t("QuantizeLinear", 23);
+  t.AddInput<float>("x", {4}, {-1.f, 0.f, 3.f, 10.f});
+  t.AddInput<float>("y_scale", {}, {2.f});
+  t.AddInput<int8_t>("y_zero_point", {}, {0});
+  t.AddOutput<int8_t>("y", {4}, {0, 0, 2, 5});
+  t.Run(OpTester::ExpectResult::kExpectSuccess, {kCpuExecutionProvider});
+}
+
 }  // namespace blocked_quantization
 }  // namespace test
 }  // namespace onnxruntime


rachelElfenbein-dev:
@microsoft-github-policy-service agree company="KamaTech"

xadupre (dismissed):

rachelElfenbein-dev:
@xadupre
I have fixed the lint errors.
Could you please review and merge my PR?
Thanks!

rachelElfenbein-dev:
@yuslepukhin
Could you please review the PR? 
Thanks!

onnxruntime/core/providers/cpu/quantization/quantize_linear.cc
```diff
@@ -599,6 +599,20 @@ Status DequantizeLinear<T>::Compute(OpKernelContext* ctx) const {
           .TypeConstraint("T2", DataTypeImpl::GetTensorType<T>()),          \
       QuantizeLinear<T>);
 
+#define REGISTER_QUANTIZELINEAR_VERSIONED_POST_22(T)                        \
+  ONNX_CPU_OPERATOR_VERSIONED_TYPED_KERNEL(                                 \
+      QuantizeLinear,                                                       \
+      23,                                                                   \
+      23,                                                                   \
```
yuslepukhin, line 606: We do not want macros contain fixed versions.
The latest kernel should have open ended version

onnxruntime/core/providers/cpu/quantization/quantize_linear.cc
```diff
@@ -599,6 +599,20 @@ Status DequantizeLinear<T>::Compute(OpKernelContext* ctx) const {
...
+      T,                                                                    \
+      KernelDefBuilder()                                                    \
+          .TypeConstraint("T1", {DataTypeImpl::GetTensorType<float>(),      \
+                                 DataTypeImpl::GetTensorType<MLFloat16>()}) \
+          .TypeConstraint("T2", {DataTypeImpl::GetTensorType<float>(),      \
+                                 DataTypeImpl::GetTensorType<MLFloat16>()}) \
```
yuslepukhin, line 612: >DataTypeImpl::GetTensorType<MLFloat16>()})  [](http://example.com/codeflow?start=33&length=43)

This enables MLFloat16 for y_scale, but there is not test for it. Typically float16 handling also requires adding code to handle it.

yuslepukhin (changes requested):
:clock1:

PR #26492, opened 2025-11-04
[Build] Fix Docker for Nuget_Test_Linux_GPU

tianleiwu:
(1) Not install cudnn9-cuda-12 since the base image has cudnn so no need to install again.
(2) Upgrade ubuntu to 24.04

This fixes the following error:
```
#11 315.1  cudnn9-cuda-12 : Depends: cudnn9-cuda-12-9 (>= 9.10.2.21) but it is not going to be installed
#11 315.1 E: Unable to correct problems, you have held broken packages.
```

### DIFF
--- a/tools/ci_build/github/linux/docker/Dockerfile.package_ubuntu_2404_gpu
+++ b/tools/ci_build/github/linux/docker/Dockerfile.package_ubuntu_2404_gpu
@@ -62,15 +62,17 @@ ARG BUILD_UID=1000
 
 RUN if getent passwd $BUILD_UID > /dev/null; then \
         echo "User with UID $BUILD_UID exists. Modifying." && \
-        export EXISTING_USER=$(id -un $BUILD_UID) && \
+        EXISTING_USER=$(id -un $BUILD_UID) && \
+        EXISTING_GROUP=$(id -gn $EXISTING_USER) && \
         usermod -l $BUILD_USER $EXISTING_USER && \
-        groupmod -n $BUILD_USER $(id -gn $EXISTING_USER) && \
+        groupmod -n $BUILD_USER $EXISTING_GROUP && \
         usermod -d /home/$BUILD_USER -m $BUILD_USER && \
         chown -R $BUILD_USER:$BUILD_USER /home/$BUILD_USER; \
     else \
         echo "User with UID $BUILD_UID does not exist. Creating." && \
         adduser --uid $BUILD_UID $BUILD_USER --disabled-password --gecos "ONNX Runtime Developer"; \
     fi
+# --- End of updated logic ---
 
 WORKDIR /home/$BUILD_USER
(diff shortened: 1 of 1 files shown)


snnn (approved):
Going forward we should consider using uv and uv's venv to manage dependencies, then we won't have this global pip installation problem.

snnn:
Example docker:

```
FROM nvidia/cuda:12.9.1-cudnn-devel-ubuntu24.04

RUN apt-get update && apt-get install -y git curl libc6-dev git unzip zip python3-dev gdb gcc g++ gh pkg-config libssl-dev libstdc++6-13-dbg python3-pip python3-pkg-resources python3-setuptools python3-wheel valgrind wget patch patchutils binutils-dev
RUN cd /tmp && curl -o cmake.tar.gz -sSL https://github.com/Kitware/CMake/releases/download/v4.1.2/cmake-4.1.2-linux-x86_64.tar.gz && tar -zxf cmake.tar.gz -C /usr --strip=1

USER ubuntu

RUN cd /tmp && curl -o /tmp/1.sh  --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs && bash 1.sh -y

ENV PATH=/home/ubuntu/.cargo/bin:/home/ubuntu/.local/bin:$PATH

RUN cd /tmp && git clone https://github.com/astral-sh/uv.git  && cd uv/crates/uv && cargo install --path .  && \
      cd $HOME && uv venv --python 3.13 --seed && . .venv/bin/activate && uv pip install onnx-ir onnx vllm --torch-backend=auto && uv tool install "huggingface_hub" 
```

tools/ci_build/github/linux/docker/Dockerfile.package_ubuntu_2404_gpu
```diff
@@ -46,17 +49,17 @@ RUN apt-key adv --fetch-keys https://developer.download.nvidia.com/compute/cuda/
...
+COPY scripts /tmp/scripts
 RUN cd /tmp/scripts && /tmp/scripts/install_dotnet.sh && rm -rf /tmp/scripts
 
 # Build final image from base.
-FROM base as final
+FROM base AS final
 ARG BUILD_USER=onnxruntimedev
 ARG BUILD_UID=1000
-RUN adduser --uid $BUILD_UID $BUILD_USER
+RUN adduser --uid $BUILD_UID $BUILD_USER --disabled-password --gecos "ONNX Runtime Developer" \
```
snnn: Ubuntu 24.04's base image already has a builtin user(the previous versions did not have). So, if the builtin user's uid is 1000 and here if BUILD_UID is also 1000, this line of code will fail.

tianleiwu: fixed

snnn (approved):

PR #26497, opened 2025-11-05
Skip setup-python for arm64 Linux builds

snnn:
## Problem
The setup-python action does not support AzureLinux 3.0.20251030 for arm64 architecture, causing workflow failures.

## Solution
Skip the setup-python step for arm64 builds since Python is already available in the Docker containers used for those builds.

## Error Details
```
Version 3.x was not found in the local cache
(node:4478) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
Error: The version '3.x' with architecture 'arm64' was not found for azurelinux 3.0.20251030 undefined.
The list of all available versions can be found here: https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json
```

## Changes
- Added conditional `if: inputs.architecture != 'arm64'` to the setup-python step in `.github/workflows/reusable_linux_build.yml`
- This affects `build-linux-arm64-debug` and `build-linux-arm64-release` jobs which recently migrated from Ubuntu to AzureLinux3

### DIFF
--- a/.github/workflows/reusable_linux_build.yml
+++ b/.github/workflows/reusable_linux_build.yml
@@ -78,6 +78,7 @@ jobs:
         uses: actions/checkout@v5
 
       - name: Set up Python ${{ inputs.python_version }}
+        if: inputs.architecture != 'arm64'
         uses: actions/setup-python@v6
         with:
           python-version: ${{ inputs.python_version }}


snnn:
## Test Failure Update

The arm64 builds are failing with this test error:
```
CApiTest.input_output_type_info - Load model from ../models/opset8/test_squeezenet/model.onnx failed: File doesn't exist
```

This appears to be unrelated to the setup-python change, as:
1. Python is still available from the Docker container (`/usr/bin/python3`)
2. The test failure is about missing model files, not Python
3. The docker volumes for `/data/models` are still being mounted during the test step

Investigating if this is a pre-existing flaky test or if there's an unexpected side effect from skipping the setup-python step that affects the test environment.

tianleiwu (approved):

PR #26501, opened 2025-11-05
webgpu: fix dispatch size issue of Transpose operator

xhcao:

### DIFF
--- a/.github/workflows/windows_x86.yml
+++ b/.github/workflows/windows_x86.yml
@@ -61,7 +61,7 @@ jobs:
         working-directory: ${{ github.workspace }}
 
       - name: Use .NET 8.x
-        uses: actions/setup-dotnet@v5
+        uses: actions/setup-dotnet@v3
         with:
           dotnet-version: '8.x'
         env:
--- a/cmake/deps.txt
+++ b/cmake/deps.txt
@@ -56,5 +56,5 @@ extensions;https://github.com/microsoft/onnxruntime-extensions/archive/c24b7bab0
 directx_headers;https://github.com/microsoft/DirectX-Headers/archive/refs/tags/v1.613.1.zip;47653509a3371eabb156360f42faf582f314bf2e
 cudnn_frontend;https://github.com/NVIDIA/cudnn-frontend/archive/refs/tags/v1.12.0.zip;7e733cfdc410d777b76122d64232499205589a96
 dawn;https://github.com/google/dawn/archive/13c1635a14574ebb7116b56a69f5519301417fda.zip;0aadd28fc385cf7d657d5fc70a352372d2d3c76a
-kleidiai;https://github.com/ARM-software/kleidiai/archive/refs/tags/v1.10.0.tar.gz;11b62149cb2514b3b9069cc435c3aa7a4e82b97a
+kleidiai;https://github.com/ARM-software/kleidiai/archive/refs/tags/v1.15.0.tar.gz;62ccd24ab60bcef68766440fb42d79071ac2a5d2
 duktape;https://github.com/svaarala/duktape/releases/download/v2.7.0/duktape-2.7.0.tar.xz;8200c8e417dbab7adcc12c4dbdef7651cfc55794
--- a/js/common/lib/inference-session.ts
+++ b/js/common/lib/inference-session.ts
@@ -259,6 +259,20 @@ export declare namespace InferenceSession {
     forceCpuNodeNames?: readonly string[];
 
     /**
+     * Specify the validation mode for WebGPU execution provider.
+     * - 'disabled': Disable all validation.
+     * When used in Node.js, disable validation may cause process crash if WebGPU errors occur. Be cautious when using
+     * this mode.
+     * When used in web, this mode is equivalent to 'wgpuOnly'.
+     * - 'wgpuOnly': Perform WebGPU internal validation only.
+     * - 'basic': Perform basic validation including WebGPU internal validation. This is the default mode.
+     * - 'full': Perform full validation. This mode may have performance impact. Use it for debugging purpose.
+     *
+     * @default 'basic'
+     */
+    validationMode?: 'disabled' | 'wgpuOnly' | 'basic' | 'full';
+
+    /**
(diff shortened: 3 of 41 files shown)


onnxruntime/core/providers/webgpu/tensor/transpose.cc
```diff
@@ -168,12 +168,10 @@ Status Transpose::DoTranspose(onnxruntime::webgpu::ComputeContext& context,
     //
     // TODO: Revert this change once the driver issue is fixed.
     if (context.AdapterInfo().vendor == std::string_view{"intel"}) {
-      // Only adjusted the dispatch size when rank is 4 yet.
-      if (rank == static_cast<size_t>(4)) {
-        dispatch_x = ceil_div(input_shape[0] * input_shape[1], 2);
-        dispatch_y = ceil_div(input_shape[2], 4);
-        dispatch_z = ceil_div(input_shape[3], 8);
-      }
```
xhcao, line 176: When running `detr-resnet-50` model, I found there are redundant/invalid work groups. For example, the input shape is [512, 512, 3, 3], the original dispatch size is [36864, 1, 1]. After these codes, the dispatch size is [131072, 1, 1],  nearly ~3x invalid work groups. The PR changed the dispatch size to [4, 8, 1152], was still 36864 work groups. And the worst situation only has 31 invalid work groups.
And The Transpose with [512, 512, 3, 3] shape cost from 3.86 ms to 1.29 ms.

xhcao:
@daijh @guschmue  @jchen10 PTAL.

onnxruntime/core/providers/webgpu/tensor/transpose.cc
daijh: @xhcao 
Nice fixing. 
Let me check on previous shape introduced this WA.

daijh: I am thinking to rewrite the transpose shaders later, as the performance of current shader is not good for many large tensors.

daijh: > @xhcao Nice fixing. Let me check on previous shape introduced this WA.

This CL brings back previous performance issue.

fs-eire: I think we probably need to have a general function to "reshape" the dispatch size so that it helps with most cases, but not a single case.

daijh: @fs-eire we drafted a new transpose shader to address this issue. Will submit in a few days.

xhcao: I think this PR is not good. For large input shape of Transpose, finally the context will normalize 1D dispatch size to 2D or 3D https://github.com/microsoft/onnxruntime/blob/2214aabc04a7c80e7a514937aff98937741748eb/onnxruntime/core/providers/webgpu/webgpu_context.cc#L280, so need to verify whether the work around code is valid here.

xhcao: Unless the driver requires 3D dispatch size here

xhcao: I had tested `sd-turbo-unet-fp16-demo` model. Current status is ~1529ms. if removing the work around code, the time is ~1342ms. The PR time is ~950ms. So the 3D dispatch size is needed.

wenqinI: Hi @xhcao ,

Thank you for this fix!

I also looked into this issue:
1. This regression probably came from this [PR](https://github.com/microsoft/onnxruntime/pull/26074).
2. Before this PR, this dispatch adjustment workaround just works for Conv weight transpose, but if we move it here, there may be some potential issue (We don't know the size of each dims).
3. The current PR looks good to mitigate the regression and good for me before the new shader is ready to totally remove this workaround.

xhcao: > I think we probably need to have a general function to "reshape" the dispatch size so that it helps with most cases, but not a single case.

@fs-eire I had tested many models, and did not find any obvious performance loss on other operators, there is even a slight performance loss if the dispatch size is relatively small.
The issue may be related with Transpose naive implementation, but the root-cause is unknown, may be a driver issue. Currently, we can only apply it to Transpose operator before driver gives us the cause.

onnxruntime/core/providers/webgpu/tensor/transpose.cc
```diff
@@ -162,18 +162,16 @@ Status Transpose::DoTranspose(onnxruntime::webgpu::ComputeContext& context,
...
+      uint32_t dispatch_size = dispatch_x;
```
daijh, line 171: We could further restrict this WA:

```
      if (rank == static_cast<size_t>(4) && dispatch_size > kMaxDispatchPerDim) {
        dispatch_x = 4;
        dispatch_y = 8;
        dispatch_z = ceil_div(dispatch_size, dispatch_x * dispatch_y);
      }
```

xhcao: I had restrict `rank == 4` because cannot get performance data from `rank == 3`.
I had tested that no `kMaxDispatchPerDim` has better performance, take `detr-resnet-50` for an example.

onnxruntime/core/providers/webgpu/tensor/transpose.cc
fs-eire: > > @xhcao Nice fixing. Let me check on previous shape introduced this WA.
> 
> This CL brings back previous performance issue.

Is the latest change resolves this issue?

fs-eire:
The CI failures - should be fixed after sync up to latest main branch

xhcao:
> The CI failures - should be fixed after sync up to latest main branch

Done

onnxruntime/core/providers/webgpu/tensor/transpose.cc
daijh: Current PR fixes the regression. 
It looks good before we rewrite the shaders.

PR #26503, opened 2025-11-05
Adjust Quantization Preprocessing on Opset Handling

qti-hungjuiw:
### Description

- Moved `ReplaceUpsampleWithResize` in the quantization preprocessing pipeline to occur before `SymbolicShapeInference`, due to the current lack of shape inference support for the `Upsample` operator.
Prevented unnecessary modifications to `model.opset_import`. Setting `model.opset_import` prior to invoking `onnx.version_converter` can interfere with successful opset conversion.

### Motivation and Context

- This change ensures the quantization preprocessing functions correctly by addressing limitations in shape inference for `Upsample`.
- It also avoids potential issues with opset conversion caused by premature modification of `model.opset_import`.


qti-hungjuiw:
Hi @xadupre, there seems to be no failure on CI. Is the PR good to go?

PR #26506, opened 2025-11-05
GatherBlockQuantized: Fix 4 bit uint8 case

jambayk:
### Description
- When uint8 packing is used for 4 bits, the packing happens along the quantization axis. 
  - For cases where the number of blocks is odd, there is an additional padding block per row of the zero-point tensor. The indexing for zero points is updated to handle this.
  - For the data tensor, there appears to be some assumption that the quantization dim is divisible by block size (this packing is supported to share weights with the lm head which uses matmulnbits but that only works if quant dim is divisible by blocksize. otherwise, there is extra padding in the final block per data row). Since block size is a power of 2, there is no padding. Without this assumption, the data indexing logic would need to be updated as well.
    - even if the above assumption is not true, there is an assumption that the quantization dim is even 
- Fixed the default zero-point value for uint8 case in CUDA implementation

### DIFF
--- a/onnxruntime/contrib_ops/cuda/quantization/gather_block_quantized.cu
+++ b/onnxruntime/contrib_ops/cuda/quantization/gather_block_quantized.cu
@@ -91,6 +91,8 @@ void LaunchGatherBlockQuantizedKernel(const T1* data,
   bool uint8_storage = std::is_same<T1, uint8_t>::value;
   int64_t uint8_padding = 0;
   if (uint8_storage && param.bits != 8) {
+    // for uint8_t with bits==4, packing happens along last dimensions, i.e. quantize_axis
+    // for odd num_quantize_blocks, we have a padding block
     const int64_t elems_per_byte = 8 / param.bits;
     const int64_t num_blocks_per_row = (param.quantize_dim + param.block_size - 1) / param.block_size;
     uint8_padding = (elems_per_byte - (num_blocks_per_row % elems_per_byte)) % elems_per_byte;


onnxruntime/contrib_ops/cpu/quantization/gather_block_quantized.cc
```diff
@@ -193,7 +193,16 @@ Status GatherBlockQuantized<T1, Tind>::CopyDataAndDequantize(const T1* data_ptr,
...
-  auto scale_full_block = (quantize_axis_dim + block_size_ - 1) / block_size_ * quantize_N;
+  auto num_quantize_blocks = (quantize_axis_dim + block_size_ - 1) / block_size_;
+  auto scale_full_block = num_quantize_blocks * quantize_N;
+  auto zp_full_block = scale_full_block;
+  if constexpr (std::is_same_v<T1, uint8_t>) {
+    // for uint8_t with bits==4, packing happens along last dimensions, i.e. quantize_axis
+    // for odd num_quantize_blocks, we have a padding block
+    int32_t components = 8 / static_cast<int>(bits_);
+    auto quantize_block_padding = (components - (num_quantize_blocks % components)) % components;
+    zp_full_block += quantize_block_padding * quantize_N;
```
yuslepukhin, line 204: Do we want to use SafeInt and narrow<> casts here?

onnxruntime/contrib_ops/cuda/quantization/gather_block_quantized.cu
```diff
@@ -80,9 +88,18 @@ void LaunchGatherBlockQuantizedKernel(const T1* data,
   // Require quant_axis is last dim
   int blocksPerGrid = (int)(ceil(static_cast<float>(param.N) / GridDim::maxThreadsPerBlock));
   bool sign = std::is_same<T1, Int4x2>::value;
+  bool uint8_storage = std::is_same<T1, uint8_t>::value;
```
yuslepukhin, line 91: >bool u [](http://example.com/codeflow?start=2&length=6)

constexpr

yuslepukhin (changes requested):
:clock1:

Issue #26502, opened 2025-11-05
[Quantization][Refactor][Internal Naming] The naming of functions that perform registration implies that they actually  perform quantization


rivkastroh:
In the quantize_model function, which is called from quantize_static, the first part deals with registering the tensors for quantization and the second part with the actual quantization of each tensor, but the meaning of the function names in the first part is the actual execution of quantization.
For example, 
CreateQDQQuantizer(onnx_quantizer, node)
mismatch: this is a factory that selects and returns a per-node handler used in the registration phase, it does not perform quantization. The name suggests creating a “quantizer” that actually  quantizes.
quantize_activation_tensor(self, tensor_name: str)
mismatch: this method only registers an activation tensor for later quantization, it does not insert Q/DQ nodes or modify tensor values. The name implies performing quantization of the activation tensor.
and so forth.

[https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/python/tools/quantization/qdq_quantizer.py#L551](url)

PR #26511, opened 2025-11-06
QDQ refactor: rename registration-phase functions to reflect their true meaning

rivkastroh:
### Description

This PR refactors the QDQ quantization pipeline to use registration-oriented names for the early pass that only marks tensors, without changing runtime behavior. It clarifies the flow (registration → materialization of quantization) and aligns the factory utility with PEP8.
Additionally, a concise term reg2quant_* (“register to quantize”) is introduced to label registration-only identifiers.

### Motivation and Context

Fixes https://github.com/microsoft/onnxruntime/issues/26502

The early pass was using quantize* names while only registering/marking tensors for later quantization. This caused confusion in code reviews, and led to incorrect assumptions that quantization was already applied at that point.

#### Problem

Methods and variables named quantize* in the early pass imply immediate graph mutations (adding Q/DQ), but they only record intent.
for example,
The factory name CreateQDQQuantizer uses CamelCase and implies it creates a “quantizer” rather than a registrar for node tensors.
is_tensor_quantized actually checks if a tensor is marked (registered), not whether it has been quantized.

#### Solution
a concise reg2quant (“register to quantize”)
Rename registration-phase symbols, No functional changes.

#### Scope of Changes (renames):

functions:

quantize_activation_tensor → reg2quant _activation_tensor

quantize_weight_tensor → reg2quant _weight_tensor

quantize_weight_tensor_per_channel → reg2quant _weight_tensor_per_channel

quantize_output_same_as_input → reg2quant _output_same_as_input

__quantize_tensor → __reg2quant _tensor

is_tensor_quantized → is_tensor_reg2quant

CreateQDQQuantizer → create_node_tensor_reg2quant

Per-operator method quantize()  →reg2quant()

variable:
op_quantizer  →op_reg2quanter

Additionally, change the calls to the functions.

No behavioral changes. Graph outputs remain identical.

### DIFF
--- a/onnxruntime/python/tools/quantization/operators/activation.py
+++ b/onnxruntime/python/tools/quantization/operators/activation.py
@@ -99,11 +99,11 @@ class QDQRemovableActivation(QDQOperatorBase):
     def __init__(self, onnx_quantizer, onnx_node):
         super().__init__(onnx_quantizer, onnx_node)
 
-    def quantize(self):
+    def reg2quant(self):
         node = self.node
 
         # If input to this node is not quantized then keep this node
-        if not self.quantizer.is_tensor_quantized(node.input[0]):
+        if not self.quantizer.is_tensor_reg2quant(node.input[0]):
             return
 
         if (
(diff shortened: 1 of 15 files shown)


rivkastroh:
@microsoft-github-policy-service agree company="KamaTech"

rivkastroh:
@xadupre, @chenfucn, you can review the PR ? Thanks.

PR #26513, opened 2025-11-06
[AIX]Blocking the call of dladdr under _AIX

ranjitshs:
### Description
In AIX,  dladdr() is not supported so blocking the call of dladdr API under _AIX. 
we don't have support of cpuinfo pkg also which generates a warning at runtime. 

This PR is to fox the issues mentioned above.

### Motivation and Context

1. Fix for below compilation error
```
/home/buildusr/jenkins/workspace/onnxruntime-openxl/onnxruntime/onnxruntime/core/platform/posix/env.cc:562:9: error: unknown type name 'Dl_info'
  562 |     if (Dl_info dl_info{};
```

2. Fix for below warning during test application executions.

`2025-11-06 07:23:44.176700000 [W:onnxruntime:Default, cpuid_info.cc:95 LogEarlyWarning] Unknown CPU vendor. cpuinfo_vendor value: 0`

### DIFF
--- a/cmake/onnxruntime_unittests.cmake
+++ b/cmake/onnxruntime_unittests.cmake
@@ -1530,11 +1530,11 @@ endif()
 
       if (CMAKE_SYSTEM_NAME MATCHES "AIX")
         target_link_libraries(onnxruntime_runtime_path_test_shared_library PRIVATE
-                            onnxruntime_common ${CMAKE_DL_LIBS})
+                              onnxruntime_common ${CMAKE_DL_LIBS})
         set_target_properties(onnxruntime_runtime_path_test_shared_library PROPERTIES AIX_SHARED_LIBRARY_ARCHIVE OFF)
       else()
         target_link_libraries(onnxruntime_runtime_path_test_shared_library PRIVATE
-                            onnxruntime_common cpuinfo ${CMAKE_DL_LIBS})
+                              onnxruntime_common cpuinfo ${CMAKE_DL_LIBS})
       endif()
       target_include_directories(onnxruntime_runtime_path_test_shared_library PRIVATE ${ONNXRUNTIME_ROOT})
 
--- a/onnxruntime/core/common/cpuid_info_vendor.cc
+++ b/onnxruntime/core/common/cpuid_info_vendor.cc
@@ -190,9 +190,6 @@ struct CpuVendorInfo {
 };
 
 constexpr auto kUnknownCpuVendorInfo = CpuVendorInfo{cpuinfo_vendor_unknown, "unknown", 0x0000};
-#if defined(_AIX)
-constexpr auto kIBMCpuVendorInfo = CpuVendorInfo{cpuinfo_vendor_ibm, "IBM", 0x1014};
-#endif
 
 constexpr std::array kCpuVendorInfos{
     CpuVendorInfo{cpuinfo_vendor_amd, "AMD", 0x1022},
@@ -232,7 +229,7 @@ void CPUIDInfo::VendorInfoInit() {
     }
 #endif  // defined(CPUINFO_SUPPORTED)
 #if defined(_AIX)
-    result = kIBMCpuVendorInfo.vendor;
+    result = cpuinfo_vendor_ibm;
 #endif
     return result;
   }();


ranjitshs:
@tianleiwu @snnn @edgchen1 @yuslepukhin 
Please review the changes in this PR. 
Let me know your comments/suggestions.

onnxruntime/core/platform/posix/env.cc
```diff
@@ -556,6 +556,8 @@ class PosixEnv : public Env {
   }
 
   PathString GetRuntimePath() const override {
+// In AIX, dladdr is not supported.
```
edgchen1, line 559: is there another way to get the current shared library path on AIX?

onnxruntime/core/common/cpuid_info_vendor.cc
```diff
@@ -232,11 +232,15 @@ void CPUIDInfo::VendorInfoInit() {
   }();
 
   const auto* vendor_info = FindCpuVendorInfo(vendor);
+// Do below logging only if CPUINFO_SUPPORTED by the OS and still vendor_info is not available.
```
edgchen1: I think that it is an implementation detail that the vendor info lookup depends on cpuinfo. I also think that if the lookup is unsuccessful, it is worth logging.

do you think that this information should be logged at a different level?

cmake/onnxruntime_unittests.cmake
```diff
@@ -1528,8 +1528,14 @@ endif()
       onnxruntime_add_shared_library(onnxruntime_runtime_path_test_shared_library
                                      ${onnxruntime_runtime_path_test_shared_library_src})
 
+      if (CMAKE_SYSTEM_NAME MATCHES "AIX")
+      target_link_libraries(onnxruntime_runtime_path_test_shared_library PRIVATE
```
edgchen1: please indent the `if()/else()` blocks

onnxruntime/test/shared_lib/test_runtime_path.cc
```diff
@@ -46,5 +47,28 @@ TEST(GetRuntimePathFromSharedLibraryTest, Basic) {
 
   ASSERT_TRUE(std::filesystem::is_regular_file(canonical_shared_library_path));
 }
+#else
+TEST(GetRuntimePathFromSharedLibraryTest, Basic) {
```
edgchen1: if GetRuntimePath() returns an empty path, it is not a valid one. perhaps the documentation needs to be updated, but I believe at least one place is relying on the returned path to be absolute.

https://github.com/microsoft/onnxruntime/blob/760eea48e1870c3c8164c50e806c6ec7ae49954f/onnxruntime/core/session/utils.cc#L419-L427

if it is unsupported on AIX, I think it's not worth testing that it returns an empty path and we can just skip this test.

cmake/onnxruntime_unittests.cmake
ranjitshs: I will update in next commit.

onnxruntime/test/shared_lib/test_runtime_path.cc
ranjitshs: Okay. I will update and skip this test.

onnxruntime/core/platform/posix/env.cc
ranjitshs: No. I don't see any way to get these shared library paths in AIX.

onnxruntime/core/common/cpuid_info_vendor.cc
ranjitshs: Are you suggesting to use the same code but with INFO  level of message ?

is it okay to define like below and use in case of AIX and if vendor_info is NULL after calling FindCpuVendorInfo()

`constexpr auto kIBMCpuVendorInfo = CpuVendorInfo{cpuinfo_vendor_ibm, "IBM", 0x0000};`

edgchen1: > Are you suggesting to use the same code but with INFO level of message ?

I'm suggesting keeping the logging. I am fine with changing the level.

> is it okay to define like below and use in case of AIX and if vendor_info is NULL after calling FindCpuVendorInfo()

> `constexpr auto kIBMCpuVendorInfo = CpuVendorInfo{cpuinfo_vendor_ibm, "IBM", 0x0000};`

does `defined(_AIX)` only apply to IBM CPUs? if so, then I think doing something like this is fine. perhaps an entry can be added to the `kCpuVendorInfos` table, and `vendor` in this function can be initialized to `cpuinfo_vendor_ibm` if `defined(_AIX)`. also, the vendor ID can be the IBM PCI vendor ID value of [0x1014](https://admin.pci-ids.ucw.cz/read/PC/1014) instead of 0x0000.

ranjitshs: _AIX macro is not for IBM CPU. It's AIX OS level check.

ranjitshs:
Hi @edgchen1 
I have updated the code . Please have a look .

onnxruntime/core/common/cpuid_info_vendor.cc
```diff
@@ -228,6 +231,9 @@ void CPUIDInfo::VendorInfoInit() {
       }
     }
 #endif  // defined(CPUINFO_SUPPORTED)
+#if defined(_AIX)
+    result = kIBMCpuVendorInfo.vendor;
+#endif
```
edgchen1, line 233:
```suggestion
#else
#if defined(_AIX)
    result = cpuinfo_vendor_ibm;
#endif
#endif
```

and then we also don't need to define `kIBMCpuVendorInfo` above.

cmake/onnxruntime_unittests.cmake
```diff
@@ -1528,8 +1528,14 @@ endif()
       onnxruntime_add_shared_library(onnxruntime_runtime_path_test_shared_library
                                      ${onnxruntime_runtime_path_test_shared_library_src})
 
-      target_link_libraries(onnxruntime_runtime_path_test_shared_library PRIVATE
+      if (CMAKE_SYSTEM_NAME MATCHES "AIX")
+        target_link_libraries(onnxruntime_runtime_path_test_shared_library PRIVATE
+                            onnxruntime_common ${CMAKE_DL_LIBS})
```
edgchen1: nit: indentation of the next line of arguments
```suggestion
        target_link_libraries(onnxruntime_runtime_path_test_shared_library PRIVATE
                              onnxruntime_common ${CMAKE_DL_LIBS})
```
here and in the `else()`

ranjitshs: Updated

onnxruntime/core/common/cpuid_info_vendor.cc
ranjitshs: Yes. We don't need kIBMCpuVendorInfo.  Updated the code.

edgchen1 (approved):

ranjitshs:
Thanks @edgchen for merging this.

PR #26516, opened 2025-11-06
[Build] update cuda 13 package: fatbin compress mode and cuda archs

tianleiwu:
### Changes

Update cuda 13 python packaging pipeline:
(1) Use fatbin compress mode = size to reduce package size. This could significantly reduce package size. 
(2) Update CMAKE_CUDA_ARCHITECTURES for cuda 13. Since we reduced package size, we are able to add more architectures.
(3) Fix cuda 13 packaging pipeline: 
- use correct (cuda13 instead of cuda12) manylinux docker. The new linux docker has cuda 13.0.2 and cuDNN 9.14.
- pass cuda version properly to run build_linux_python_package.sh in docker. (CUDA_VERSION in docker was 12.8.1, and now we pass "12.8" from yml to be consistent).

Note that the compress mode and cuda archs settings are not changed for CUDA 12.8, so cuda 12 wheel size is larger than cuda 13 wheel size. We can update settings for cuda 12.8 in a separated PR if needed.

The nuget pipeline for cuda 13 need extra code change, and this PR only fixes python packaging pipeline.

### Python GPU Wheel Size (Cuda Architectures + PTX)

CUDA | Windows  | Linux
----|---|---
12.8 | 221 MB  (52;61;75;86;89+90) | 271 MB (60;70;75;80;86;90a+90)  
13.0 | 186 MB (75;80;86;89;90a;100a;120a+120) |  191 MB (75;80;86;89;90a;100a;120a+120)


GrigoryEvko:
Yes please, nightly feeds are not available publicly anymore and nobody raised this issue and all cuda 13 builds hardcode cuda 12 cublas and other libraries

GrigoryEvko:
Also this pr is required to fix cuda 13 builds https://github.com/microsoft/onnxruntime/pull/26518

snnn (approved):

tools/ci_build/github/azure-pipelines/py-cuda-packaging-pipeline.yml
```diff
@@ -50,3 +50,6 @@ extends:
         parameters:
           cmake_build_type: ${{ parameters.cmake_build_type }}
           cuda_version: '12.8'
+          cudnn_folder: ''
+          cmake_cuda_archs: '52-real;61-real;75-real;86-real;89-real;90-virtual'
```
johannes-rehm-snkeos, line 54: Why these CUDA architectures? Why `61-real` and `75-real` and not `70-real`?
CUDA kernels are only forward compatible within minor versions. That means it does not run e.g. on a Tesla V100 with SM 7.0

PR #26518, opened 2025-11-07
Fix CUDA version hardcoding to support CUDA 13+ dynamically

GrigoryEvko:
## Description
Fixes runtime library loading failures when building with CUDA 13 by replacing hardcoded CUDA 12 references with dynamic version detection.

Related to #26516 which updates CUDA 13 build pipelines, but this PR fixes the Python runtime code that was still hardcoded to CUDA 12.

## Problem
The build system correctly detects CUDA 13 via CMake, but the runtime Python code had CUDA 12 hardcoded in multiple locations, causing "CUDA 12 not found" errors on CUDA 13 systems.

## Solution
Modified onnxruntime/__init__.py and setup.py to dynamically use the detected CUDA version instead of hardcoded "12" strings.

## Changes
- Dynamic CUDA version extraction from build info
- Library paths now use f-strings with cuda_major_version
- Added CUDA 13 support to extras_require and dependency exclusions
- Fixed TensorRT RTX package to use correct CUDA version
- Updated version validation to accept CUDA 12+
- Fixed PyTorch compatibility checks to compare versions dynamically

## Impact
- CUDA 13 builds now load correct libraries
- Backward compatible with CUDA 12
- Forward compatible with future CUDA versions

## Testing
Verified with CUDA 13.0 build that library paths resolve correctly and preload_dlls() loads CUDA 13 libraries without errors.

### DIFF
--- a/onnxruntime/__init__.py
+++ b/onnxruntime/__init__.py
@@ -144,13 +144,7 @@ def _extract_cuda_major_version(version_str: str) -> str:
     Returns:
         Major version as string, or "12" if parsing fails
     """
-    if not version_str:
-        return "12"
-
-    with contextlib.suppress(AttributeError, IndexError):
-        return version_str.split(".")[0]
-
-    return "12"
+    return version_str.split(".")[0] if version_str else "12"
 
 
 def _get_cufft_version(cuda_major: str) -> str:
@@ -329,16 +323,17 @@ def preload_dlls(cuda: bool = True, cudnn: bool = True, msvc: bool = True, direc
             print("It can be downloaded at https://aka.ms/vs/17/release/vc_redist.x64.exe.")
 
     # Check if CUDA version is supported (12.x or 13.x+)
+    ort_cuda_major = None
     if cuda_version:
         try:
-            cuda_major = int(cuda_version.split(".")[0])
-            if cuda_major < 12 and (cuda or cudnn):
+            ort_cuda_major = int(cuda_version.split(".")[0])
+            if ort_cuda_major < 12 and (cuda or cudnn):
                 print(
                     f"\033[33mWARNING: {package_name} is built with CUDA {cuda_version}, which is not supported for preloading. "
                     f"CUDA 12.x or newer is required. Call preload_dlls with cuda=False and cudnn=False.\033[0m"
                 )
                 return
-        except (ValueError, IndexError, AttributeError):
+        except ValueError:
             print(
                 f"\033[33mWARNING: Unable to parse CUDA version '{cuda_version}'. "
                 "Skipping DLL preloading. Call preload_dlls with cuda=False and cudnn=False.\033[0m"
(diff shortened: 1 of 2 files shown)


GrigoryEvko:

onnxruntime/__init__.py
```diff
@@ -134,13 +134,21 @@ def _get_package_root(package_name: str, directory_name: str | None = None):
...
+            pass  # Use default if parsing fails
+
     if is_windows:
         # Path is relative to site-packages directory.
         cuda_dll_paths = [
-            ("nvidia", "cublas", "bin", "cublasLt64_12.dll"),
-            ("nvidia", "cublas", "bin", "cublas64_12.dll"),
+            ("nvidia", "cublas", "bin", f"cublasLt64_{cuda_major_version}.dll"),
+            ("nvidia", "cublas", "bin", f"cublas64_{cuda_major_version}.dll"),
             ("nvidia", "cufft", "bin", "cufft64_11.dll"),
```
xadupre: what about this one?

setup.py
```diff
@@ -783,8 +797,15 @@ def reformat_run_count(count_str):
 
 # Adding CUDA Runtime as dependency for NV TensorRT RTX python wheel
 if package_name == "onnxruntime-trt-rtx":
-    install_requires.append("nvidia-cuda-runtime-cu12~=12.0")
     cuda_version = parse_arg_remove_string(sys.argv, "--cuda_version=")
+    # Determine CUDA major version for correct runtime package
+    cuda_runtime_major = "12"  # Default to CUDA 12
```
xadupre: Maybe it is worth adding a function which extracts the major version, this code is repeated a couple of times.

setup.py
```diff
@@ -823,18 +844,31 @@ def save_build_and_package_info(package_name, version_number, cuda_version, rocm
...
-        ],
-        "cudnn": [
-            "nvidia-cudnn-cu12~=9.0",
-        ],
-    }
+if package_name == "onnxruntime-gpu":
+    if is_cuda_version_12:
+        extras_require = {
+            "cuda": [
+                "nvidia-cuda-nvrtc-cu12~=12.0",
```
xadupre: why not using something like ``f"....{cuda_major_version}"`` ?

GrigoryEvko: Updated

setup.py
GrigoryEvko: Updated

onnxruntime/__init__.py
GrigoryEvko: Missed this one, for reasons unknown to me it lags one version behind toolkit major version
```/usr/local/cuda-13.0/targets/x86_64-linux/lib/libcufft.so
/usr/local/cuda-13.0/targets/x86_64-linux/lib/libcufftw.so
/usr/local/cuda-13.0/targets/x86_64-linux/lib/libcufft.so.12
/usr/local/cuda-13.0/targets/x86_64-linux/lib/libcufft.so.12.0.0.61
/usr/local/cuda-13.0/targets/x86_64-linux/lib/libcufftw.so.12
/usr/local/cuda-13.0/targets/x86_64-linux/lib/libcufftw.so.12.0.0.61
/usr/local/cuda-13.0/targets/x86_64-linux/lib/stubs/libcufft.so
/usr/local/cuda-13.0/targets/x86_64-linux/lib/stubs/libcufftw.so
/usr/local/cuda-13.0/targets/x86_64-linux/lib/libcufft_static.a
/usr/local/cuda-13.0/targets/x86_64-linux/lib/libcufftw_static.a```

File: setup.py (hunk 3 above)
Comment (GrigoryEvko) [reply setup.py]: oh sorry, moment please

**Comment by tianleiwu** [comment]
Please format the code with lintrunner: https://github.com/microsoft/onnxruntime/blob/main/docs/Coding_Conventions_and_Standards.md#Linting

File: setup.py [hunk 4]
```diff
@@ -54,18 +55,26 @@ def parse_arg_remove_string(argv, arg_name_equal):
...
 is_openvino = False
 is_qnn = False
 qnn_version = None
 # The following arguments are mutually exclusive
 if wheel_name_suffix == "gpu":
-    # TODO: how to support multiple CUDA versions?
     cuda_version = parse_arg_remove_string(sys.argv, "--cuda_version=")
     if cuda_version:
-        is_cuda_version_12 = cuda_version.startswith("12.")
+        try:
```
tianleiwu: nit: can be simplified as
```
   is_cuda_version_12 = cuda_version.startswith("12.")
   is_cuda_version_13 = cuda_version.startswith("13.")
```

onnxruntime/__init__.py
```diff
@@ -133,14 +135,49 @@ def _get_package_root(package_name: str, directory_name: str | None = None):
...
+    Returns:
+        Major version as string, or "12" if parsing fails
+    """
+    if not version_str:
+        return "12"
+
+    with contextlib.suppress(AttributeError, IndexError):
+        return version_str.split(".")[0]
+
+    return "12"
```
tianleiwu: nit: the following might be more robust:
```
def _extract_cuda_major_version(version_str: str, default_value:str = "12") -> str:
    if not isinstance(version_str, str):
        return default_value
        
    major, separator, _ = version_str.partition(".")
    return major if major else default_value
```

GrigoryEvko:
Maybe this way? deduplicated some code

PR #26519, opened 2025-11-07
[VitisAI] Remove unused function body handling in graph fusion

mingyueliuh:
### Description
Vitis AI EP graph_fuse memory optimize.
This PR removes unused function body handling code in the VitisAI graph fusion implementation.

### Changes
- Removed unused function body handling in graph fusion (`onnxruntime/core/providers/vitisai/imp/graph.cc`)

### Context
The function body handling code in the graph fusion logic was not being used and can be safely removed to simplify the implementation.

### DIFF
--- a/onnxruntime/core/providers/vitisai/imp/graph.cc
+++ b/onnxruntime/core/providers/vitisai/imp/graph.cc
@@ -248,12 +248,6 @@ Node& graph_fuse(Graph& graph, const std::string& name,
   indexed_subgraph->SetMetaDef(std::move(meta_def));
 
   auto& fused_node = graph.FuseSubGraph(*indexed_subgraph, name);
-  auto function_body = fused_node.GetFunctionBody();
-  if (function_body) {
-    auto proto = function_body->Body().ToGraphProto();
-    *proto->mutable_name() = name;
-    fused_node.AddAttribute("body", *proto);
-  }
   for (auto&& o : fused_node.OutputDefs()) {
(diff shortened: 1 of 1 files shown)


mingyueliuh:
@adrianlizarraga please help reivew the PR , Thanks

onnxruntime/core/providers/vitisai/imp/graph.cc
```diff
@@ -248,12 +248,6 @@ Node& graph_fuse(Graph& graph, const std::string& name,
   indexed_subgraph->SetMetaDef(std::move(meta_def));
 
   auto& fused_node = graph.FuseSubGraph(*indexed_subgraph, name);
-  auto function_body = fused_node.GetFunctionBody();
-  if (function_body) {
-    auto proto = function_body->Body().ToGraphProto();
-    *proto->mutable_name() = name;
-    fused_node.AddAttribute("body", *proto);
-  }
```
yuslepukhin, line 256: I think this code is being used by Context Dependent Kernels.

https://github.com/onnx/onnx/blob/main/onnx/defs/sequence/defs.cc#L425

wcy123: @yuslepukhin This function is used internally by a compiler backend behind VitisAI EP. Actually, a compiler backend modifies the cloned graph via this function, and the attribute "body" is not really in use. It causes high peak memory usage. This can reduce ~100M memory footprint.

yuslepukhin (approved):
:shipit:

PR #26526, opened 2025-11-07
[webgpu] Fix `GatherBlockQuantized` on Intel ADL/TGL platforms

daijh:
### Description
The `GatherBlockQuantized` operation was using incorrect `data_indices` during execution on Intel Alder Lake (ADL) and Tiger Lake (TGL) platforms.

This change sets the proper `data_indices`, resolving correctness issues encountered with the Phi-4-mini model on these architectures.

### Motivation and Context
See above.

### DIFF
--- a/onnxruntime/contrib_ops/webgpu/quantization/gather_block_quantized.cc
+++ b/onnxruntime/contrib_ops/webgpu/quantization/gather_block_quantized.cc
@@ -17,7 +17,7 @@ using onnxruntime::webgpu::ComputeContext;
 Status GatherBlockQuantizedProgram::GenerateShaderCode(ShaderHelper& shader) const {
   const auto& x = shader.AddInput("input", ShaderUsage::UseElementTypeAlias);
   const auto& x_shape = shader.AddIndices("input_shape", ShaderUsage::UseUniform | ShaderUsage::UseIndicesTypeAlias);
-  const auto& indices = shader.AddInput("indices", ShaderUsage::UseUniform | ShaderUsage::UseIndicesTypeAlias | ShaderUsage::UseIndicesToOffset);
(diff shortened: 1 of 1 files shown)


daijh:
The first commit is only the early draft to demonstrate the fixing.
To be refined in following commits.

daijh:
@guschmue @fs-eire PTAL.

fs-eire:
Thanks for the fix! LGTM

onnxruntime/contrib_ops/webgpu/quantization/gather_block_quantized.cc
```diff
@@ -38,17 +38,23 @@ Status GatherBlockQuantizedProgram::GenerateShaderCode(ShaderHelper& shader) con
     shader.MainFunctionBody()
         << "let indices_indices = " << output.IndicesGet("output_indices", "uniforms.gather_axis") << ";\n";
   }
+
+  shader.MainFunctionBody()
+      << "var index = " << indices.GetByIndices("indices_indices") << ";\n"
+      << "if (index < 0) { index += " << x_shape_[gather_axis_] << ";}\n"
```
fs-eire: Not related to this PR (because this part is not changed):

The current implementation of `GatherBlockQuantizedProgram::GenerateShaderCode` seems to have dependency on `X`'s shape (value of `x_shape_[gather_axis_]` is directly written into the shader) but the value of `x_shape_[gather_axis_]` is not a part of cache hint. It looks to me like this may potentially cause an inconsistency when a new input `X` with different shape reuses the same program.

daijh: Thanks. Get gather axis dim from uniforms fixing the cache issue.

fs-eire (approved):

guschmue:
tested on tgl with qwen3-0.6b and int4 embeddings - works!

daijh:
Not sure why CI failed. Probably kick off a retry.

PR #26534, opened 2025-11-07
[QNN-EP] Restrict zero pads validation check to NPU backend

qti-mattsinc:
### Description
- Pad ops with all zero padding values are allowed in the QNN GPU backend.
- Restricting a recently-added validation check in the op builder to apply only to the NPU.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/pad_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/pad_op_builder.cc
@@ -191,8 +191,9 @@ Status PadOpBuilder::ProcessAttributesAndOutputs(QnnModelWrapper& qnn_model_wrap
   bool has_negative = std::any_of(tensor_data, tensor_data + size, [](int64_t item) { return item < 0; });
   bool has_positive = std::any_of(tensor_data, tensor_data + size, [](int64_t item) { return item > 0; });
 
-  // Zero padding value gives 3110 error on QNN.
-  if (!has_positive && !has_negative) {
+  // Zero padding value gives 3110 error on QNN on NPU.
+  const bool is_npu_backend = IsNpuBackend(qnn_model_wrapper.GetQnnBackendType());
+  if (is_npu_backend && !has_positive && !has_negative) {
     return ORT_MAKE_STATUS(ONNXRUNTIME, FAIL, "Got QNN invalid zero only padding value.");
   }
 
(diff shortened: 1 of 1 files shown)


onnxruntime/core/providers/qnn/builder/opbuilder/pad_op_builder.cc
```diff
@@ -236,7 +237,7 @@ Status PadOpBuilder::ProcessAttributesAndOutputs(QnnModelWrapper& qnn_model_wrap
   qnn_model_wrapper.AddParamWrapper(std::move(mode_param));
 
   QnnParamWrapper pad_amount_param(node_unit.Index(), node_unit.Name(), QNN_OP_PAD_PARAM_PAD_AMOUNT,
-                                   std::move(pad_amount_dim), std::move(pad_amount));
+                                   std::move(pad_amount_dim), std::vector<uint32_t>(pad_amount));
```
yuslepukhin, line 240: >td::vector<uint32_t>(pad_amount) [](http://example.com/codeflow?start=63&length=32)

Why do you need to copy pad_amount vector?

onnxruntime/core/providers/qnn/builder/opbuilder/pad_op_builder.cc
```diff
@@ -274,19 +277,18 @@ Status PadOpBuilder::ProcessAttributesAndOutputs(QnnModelWrapper& qnn_model_wrap
                                   std::vector<uint32_t>(pad_output_shape));
       ORT_RETURN_IF_NOT(qnn_model_wrapper.AddTensorWrapper(std::move(pad_output)),
                         "Failed to add Pad output tensor.");
+
       ORT_RETURN_IF_NOT(qnn_model_wrapper.CreateQnnNode(pad_name,
                                                         QNN_OP_PACKAGE_NAME_QTI_AISW,
                                                         GetQnnOpType(node_unit.OpType()),
-                                                        std::move(input_names),
+                                                        std::vector<std::string>(input_names),
```
yuslepukhin, line 284: >input_names [](http://example.com/codeflow?start=81&length=11)

Why input_names copy is necessary?

onnxruntime/core/providers/qnn/builder/opbuilder/pad_op_builder.cc
qti-mattsinc: The vector is used again on line 262, and it shouldn't be moved from before then.

onnxruntime/core/providers/qnn/builder/opbuilder/pad_op_builder.cc
qti-mattsinc: It's used again on line 336 and shouldn't be moved from before then.

edgchen1:
will close and reopen to try to restart CI workflows

PR #26535, opened 2025-11-07
Create a CUDA based memory arena instead of Cuda Allocator wrapped into BFCArena

yuslepukhin:
### Description
This change allows users to better control GPU memory in shared environments with multiple tenants or multiple inference sessions per process.

Cuda based memory pool features native allocations on streams, allows trimming the memory on Shrink if enabled and releases memory back to the system based on the user specified parameters. 

In my limited testing latencies were comparable with running on BFCArena, although your milage and requirements may vary.

CudaMemoryPoolArena is enabled via OrtArenaCfg with introducing 3 new parameters:
- `use_cuda_mempool` set to 1 to enable
- `cuda_mempool_release_threshold` amount of memory to keep cached
- `cuda_mempool_bytes_to_keep_on_shrink` the amount of memory to keep on Shrink when being trimmed, allocated memory is not affected.

### Motivation and Context
Better GPU memory control in multitenant environments.

There are some new options for `onnxruntime_perf_test` introduced in this PR so they may assist clients to figure out the best settings for they case:
- `--enable_cuda_mempool 209715200;1048576` with first parameter being `cuda_mempool_release_threshold`. The second `cuda_mempool_bytes_to_keep_on_shrink` can be zero if shrink is not enabled.
- `--shrink_arena_between_runs gpu:0` measure perf and memory consumption with shrink. This new allocator strictly speaking does not need `Shrink()` since cuda mempool may release memory on the go according to `cuda_mempool_release_threshold`.

Here is some performance numbers gathered when running HF_Bart model.

If the CudaMempool release threshold is set too low, latency increases because the system ends up constantly allocating and releasing memory. But as we raise the threshold and allow more memory to stay allocated, latency improves—and we end up using only about half as much memory between runs compared to BFCArena.

Running default setup with BFCArena
> onnxruntime_perf_test -s -e cuda -I -S 10 -m times -r 100 "hf_Bart_torchscript.onnx"
Average inference time cost total: 66.493545 ms
P99 Latency: 0.0805385 s
Total memory allocated: 1,409,286,144

200 MB release threshold
> onnxruntime_perf_test -s -e cuda --enable_cuda_mempool 209715200;0 -I -S 10 -m times -r 100 hf_Bart_torchscript.onnx
Average inference time cost total: 77.367473 ms
P99 Latency: 0.0931895 s

0.5Gb release threshold
> onnxruntime_perf_test -s -e cuda --enable_cuda_mempool 536870912;0 -I -S 10 -m times -r 100 hf_Bart_torchscript.onnx
Average inference time cost total: 75.112840 ms
P99 Latency: 0.0910992 s

1Gb release threshold
> onnxruntime_perf_test -s -e cuda --enable_cuda_mempool 1073741824;0 -I -S 10 -m times -r 100 hf_Bart_torchscript.onnx
Average inference time cost total: 66.533892 ms
P99 Latency: 0.0761336 s

Enabling shrink show we’re retaining only half the memory compared to BFCArena in between inference runs.

>CudaMempoolArena::Shrink: pool current_in_use: 709,603,688 reserved size after trim : 738,197,504 bytes.

### DIFF
--- a/js/common/lib/inference-session.ts
+++ b/js/common/lib/inference-session.ts
@@ -259,6 +259,20 @@ export declare namespace InferenceSession {
     forceCpuNodeNames?: readonly string[];
 
     /**
+     * Specify the validation mode for WebGPU execution provider.
+     * - 'disabled': Disable all validation.
+     * When used in Node.js, disable validation may cause process crash if WebGPU errors occur. Be cautious when using
+     * this mode.
+     * When used in web, this mode is equivalent to 'wgpuOnly'.
+     * - 'wgpuOnly': Perform WebGPU internal validation only.
+     * - 'basic': Perform basic validation including WebGPU internal validation. This is the default mode.
+     * - 'full': Perform full validation. This mode may have performance impact. Use it for debugging purpose.
+     *
+     * @default 'basic'
+     */
+    validationMode?: 'disabled' | 'wgpuOnly' | 'basic' | 'full';
+
+    /**
      * Specify an optional WebGPU device to be used by the WebGPU execution provider.
      */
     device?: TryGetGlobalType<'GPUDevice'>;
--- a/js/package-lock.json
+++ b/js/package-lock.json
@@ -4020,9 +4020,9 @@
       }
     },
     "node_modules/js-yaml": {
-      "version": "4.1.0",
-      "resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-4.1.0.tgz",
-      "integrity": "sha512-wpxZs9NoxZaJESJGIZTyDEaYpl0FKSA+FB9aJiyemKhMwkxQg63h4T1KJgUGHpTqPDNRcmmYLugrRjJlBtWvRA==",
+      "version": "4.1.1",
+      "resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-4.1.1.tgz",
+      "integrity": "sha512-qQKT4zQxXl8lLwBtHMWwaTcGfFOZviOJet3Oy/xmGk2gZH677CJM9EvtfdSkgWcATZhj/55JZ0rmy3myCT5lsA==",
       "dev": true,
       "dependencies": {
         "argparse": "^2.0.1"
@@ -8555,9 +8555,9 @@
       }
     },
     "js-yaml": {
-      "version": "4.1.0",
-      "resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-4.1.0.tgz",
-      "integrity": "sha512-wpxZs9NoxZaJESJGIZTyDEaYpl0FKSA+FB9aJiyemKhMwkxQg63h4T1KJgUGHpTqPDNRcmmYLugrRjJlBtWvRA==",
+      "version": "4.1.1",
+      "resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-4.1.1.tgz",
+      "integrity": "sha512-qQKT4zQxXl8lLwBtHMWwaTcGfFOZviOJet3Oy/xmGk2gZH677CJM9EvtfdSkgWcATZhj/55JZ0rmy3myCT5lsA==",
       "dev": true,
       "requires": {
         "argparse": "^2.0.1"
--- a/js/web/lib/wasm/session-options.ts
+++ b/js/web/lib/wasm/session-options.ts
@@ -129,6 +129,11 @@ const setExecutionProviders = async (
 
               appendEpOption(epOptions, 'forceCpuNodeNames', names.join('\n'), allocs);
             }
+
+            // set validation mode
+            if (webgpuOptions.validationMode) {
+              appendEpOption(epOptions, 'validationMode', webgpuOptions.validationMode, allocs);
+            }
           }
 
           const info = getInstance().webgpuRegisterDevice!(customDevice);
--- a/onnxruntime/contrib_ops/webgpu/bert/bias_gelu.cc
+++ b/onnxruntime/contrib_ops/webgpu/bert/bias_gelu.cc
@@ -1,0 +1,95 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include "core/providers/webgpu/shader_helper.h"
+#include "core/providers/webgpu/webgpu_supported_types.h"
+#include "core/providers/webgpu/math/unary_elementwise_ops.h"
+#include "contrib_ops/webgpu/bert/bias_gelu.h"
+#include "contrib_ops/webgpu/webgpu_contrib_kernels.h"
+
+namespace onnxruntime {
+namespace contrib {
+namespace webgpu {
+
+ONNX_OPERATOR_KERNEL_EX(
+    BiasGelu,
+    kMSDomain,
+    1,
+    kWebGpuExecutionProvider,
+    (*KernelDefBuilder::Create())
+        .TypeConstraint("T", WebGpuSupportedFloatTypes()),
+    BiasGelu);
+
+Status BiasGeluProgram::GenerateShaderCode(ShaderHelper& shader) const {
+  const auto& x = shader.AddInput("x", ShaderUsage::UseUniform | ShaderUsage::UseValueTypeAlias);
+  const auto& bias = shader.AddInput("bias", ShaderUsage::UseUniform | ShaderUsage::UseShapeAndStride);
+  const auto& y = shader.AddOutput("y", ShaderUsage::UseUniform);
+
+  shader.AdditionalImplementation() << onnxruntime::webgpu::ErfImpl;
+  shader.MainFunctionBody() << shader.GuardAgainstOutOfBoundsWorkgroupSizes("uniforms.vec_size")
+                            << "  var a = " << x.GetByOffset("global_idx") << ";\n";
+
+  // Add bias to input
+  if (bias_components_ == 1) {
+    shader.MainFunctionBody() << "  let bias_offset = global_idx * 4;\n"
+                                 "  a += x_value_t("
+                              << bias.GetByOffset("bias_offset % uniforms.bias_shape") << ", "
+                              << bias.GetByOffset("(bias_offset + 1) % uniforms.bias_shape") << ", "
+                              << bias.GetByOffset("(bias_offset + 2) % uniforms.bias_shape") << ", "
+                              << bias.GetByOffset("(bias_offset + 3) % uniforms.bias_shape") << ");\n";
+  } else {
+    shader.MainFunctionBody() << "  a += " << bias.GetByOffset("global_idx % uniforms.bias_shape") + ";\n";
+  }
+
+  // Apply GELU activation: 0.5 * a * (1.0 + erf(a * 0.7071067811865475))
+  shader.MainFunctionBody() << y.SetByOffset("global_idx", onnxruntime::webgpu::GeluExpr);
+
+  return Status::OK();
+}
+
+Status BiasGelu::ComputeInternal(onnxruntime::webgpu::ComputeContext& context) const {
+  const auto* input = context.Input(0);
+  const auto* bias = context.Input(1);
+  auto* output = context.Output(0, input->Shape());
+
+  uint32_t data_size = onnxruntime::narrow<uint32_t>(output->Shape().Size());
+  if (data_size == 0) {
+    return Status::OK();
+  }
+
+  const auto& input_shape = input->Shape();
+  const auto& bias_shape = bias->Shape();
+
+  // Validate inputs
+  if (input_shape.NumDimensions() < 1 || bias_shape.NumDimensions() != 1) {
+    return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT,
+                           "BiasGelu: input must have at least 1 dimension and bias must be 1-dimensional.");
+  }
+
+  if (input_shape.GetDims().back() != bias_shape.GetDims().back()) {
+    return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT,
+                           "BiasGelu: bias must match the last dimension of input.");
+  }
+
+  const auto vec_size = (data_size + 3) / 4;
+  uint32_t bias_size = onnxruntime::narrow<uint32_t>(bias->Shape().Size());
+  int bias_components = 1;
+
+  if (bias_size % 4 == 0) {
+    bias_components = 4;
+    bias_size = bias_size / 4;
+  }
+
+  BiasGeluProgram program{bias_components};
+  program.AddInput({input, ProgramTensorMetadataDependency::Type, {vec_size}, 4})
+      .AddInput({bias, ProgramTensorMetadataDependency::TypeAndRank, {bias_size}, bias_components})
+      .AddOutput({output, ProgramTensorMetadataDependency::None, {vec_size}, 4})
+      .SetDispatchGroupSize((vec_size + WORKGROUP_SIZE - 1) / WORKGROUP_SIZE)
+      .AddUniformVariable({vec_size});
+
+  return context.RunProgram(program);
+}
+
+}  // namespace webgpu
+}  // namespace contrib
+}  // namespace onnxruntime
--- a/onnxruntime/contrib_ops/webgpu/bert/bias_gelu.h
+++ b/onnxruntime/contrib_ops/webgpu/bert/bias_gelu.h
@@ -1,0 +1,38 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#pragma once
+
+#include "core/providers/webgpu/program.h"
+#include "core/providers/webgpu/webgpu_kernel.h"
+
+namespace onnxruntime {
+namespace contrib {
+namespace webgpu {
+
+using namespace onnxruntime::webgpu;
+using onnxruntime::webgpu::ComputeContext;
+
+class BiasGeluProgram final : public Program<BiasGeluProgram> {
+ public:
+  BiasGeluProgram(int bias_components) : Program{"BiasGelu"}, bias_components_{bias_components} {
+  }
+
+  Status GenerateShaderCode(ShaderHelper& sh) const override;
+
+  WEBGPU_PROGRAM_DEFINE_UNIFORM_VARIABLES({"vec_size", ProgramUniformVariableDataType::Uint32});
+
+ private:
+  int bias_components_;
+};
+
+class BiasGelu final : public WebGpuKernel {
+ public:
+  BiasGelu(const OpKernelInfo& info) : WebGpuKernel(info) {}
+
+  Status ComputeInternal(ComputeContext& context) const override;
+};
+
+}  // namespace webgpu
+}  // namespace contrib
+}  // namespace onnxruntime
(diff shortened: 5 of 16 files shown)


include/onnxruntime/core/framework/allocator.h
```diff
@@ -38,6 +38,13 @@ struct OrtArenaCfg {
   int max_dead_bytes_per_chunk;           // use -1 to allow ORT to choose the default
   int initial_growth_chunk_size_bytes;    // use -1 to allow ORT to choose the default
   int64_t max_power_of_two_extend_bytes;  // use -1 to allow ORT to choose the default
+  // Use CudaMemPool based arena if available (starting with cuda 11.2)
+  int use_cuda_mempool = -1;
```
adrianlizarraga, line 42: It's my understanding that ORT core should become more EP-agnostic to facilitate, for example, a single build of ORT that works for multiple (plugin) EPs.

Does having CUDA-specific options deviate from this direction/goal? Do you know if there's a way to potentially pass these EP-specific through other means (.e.g, provider options)?

yuslepukhin: I have not found a way to do so. It is true that OrtArenaCfg is currently not EP specific, however, it is `BFCArena` specific. `OrtCUDAProviderOptionsV2` is on the other hand is cuda specific and it points to the arena config.

onnxruntime/core/providers/cuda/cuda_mempool_arena.cc
```diff
@@ -0,0 +1,236 @@
...
+    pool_ = nullptr;
+  }
+}
+
+void* CudaMempoolArena::Alloc(size_t size) {
+  if (size == 0) return nullptr;
+
+  void* p = nullptr;
+  constexpr const cudaStream_t kDefaultStream = static_cast<cudaStream_t>(0);
+  cudaError_t err = cudaMallocFromPoolAsync(&p, size, pool_, kDefaultStream);
```
gedoensmax, line 86: Since this alloc function guarantees that memory is resent after the call i think a `cudaStreamSynchonize` is missing. If the default stream is busy this can be delayed significantly.

gedoensmax (commented):
Thanks this looks like a good addition. Did you capture any performance numbers ?

onnxruntime/core/providers/cuda/cuda_mempool_arena.cc
yuslepukhin: What is better: 1) cudaStreamSynchonize 2) create and event and wait on it? Is 1 and 2 the same thing?

include/onnxruntime/core/framework/allocator.h
adrianlizarraga: What about adding a new member/field to OrtCudaProviderOptionsV2 that holds these cuda-specific arena configuration options?
 
https://github.com/microsoft/onnxruntime/blob/8fe48049c089ec50f6f915ca97c37969b7194a96/include/onnxruntime/core/providers/cuda/cuda_provider_options.h#L29

include/onnxruntime/core/framework/allocator.h
```diff
@@ -55,6 +62,9 @@ struct OrtArenaCfg {
     static constexpr const char* InitialGrowthChunkSizeBytes = "arena.initial_growth_chunk_size_bytes";
     static constexpr const char* MaxPowerOfTwoExtendBytes = "arena.max_power_of_two_extend_bytes";
     static constexpr const char* MaxMem = "arena.max_mem";
+    static constexpr const char* UseCudaMemPool = "arena.use_cuda_mempool";
```
hariharans29, line 65: Should there be a distinguishable trait between the arena and the memory pool in the string ? (i.e.) should "arena." be the prefix for memory pool related configs too ?

yuslepukhin: https://github.com/microsoft/onnxruntime/issues/26545

onnxruntime/core/providers/cuda/cuda_execution_provider.cc
```diff
@@ -152,19 +154,34 @@ AllocatorPtr CUDAExecutionProvider::CreateCudaAllocator(OrtDevice::DeviceId devi
...
-        [](OrtDevice::DeviceId id) {
-          return std::make_unique<CUDAAllocator>(id, CUDA);
-        },
-        device_id,
-        true,
-        {default_memory_arena_cfg ? *default_memory_arena_cfg
-                                  : OrtArenaCfg(gpu_mem_limit, static_cast<int>(arena_extend_strategy), -1, -1, -1, -1L)},
-        // make it stream aware
-        true);
+    // Check if we are running against older version of CUDA runtime
```
hariharans29: Is the logic to check the cuda runtime version missing here (There is a comment) ?

yuslepukhin: No, it was related to cuda version with respect to Cuda Mempool availability. We are long past 11.2 so no longer needed.

include/onnxruntime/core/framework/allocator.h
yuslepukhin: Looks like, the consensus for now is to stick with the old config since it is a legacy CUDA EP. There are several public APIs that take only OrtArenaCfg and we cannot change it.

include/onnxruntime/core/framework/allocator.h
hariharans29: Not following how that issue is relevant to this. That is an issue no doubt but this is a separate question/thought.

onnxruntime/core/providers/cuda/cuda_execution_provider.cc
hariharans29: The comment about checking cuda runtime version is dangling then

include/onnxruntime/core/framework/allocator.h
yuslepukhin: We are creating an arena by means of CudaMempool. It is an ORT arena allocator.

onnxruntime/core/providers/cuda/cuda_mempool_arena.cc
yuslepukhin: FWIW, chat gpt answer:

Prefer event sync (cudaEventRecord + cudaEventSynchronize) over cudaStreamSynchronize() if more work may be queued to that stream after the allocation—events avoid waiting for unrelated later work.
Use cudaStreamSynchronize() only when you truly want to wait for all prior work enqueued to that stream, not just the allocation.

onnxruntime/core/framework/bfc_arena.h
```diff
@@ -100,7 +87,13 @@ class BFCArena : public IAllocator {
 
   size_t AllocatedSize(const void* ptr);
 
-  ArenaType GetArenaType() const { return arena_type_; }
+  // Frees all allocation regions in which no chunk is in use.
```
hariharans29, line 90: The documentary around `Shrink()` needs an update to call out that they are for BFCArena specific ?

For example, `initial_growth_chunk_size_bytes_` is BFCArena specific and there is no guarantee that memory allocated through `Reserve()` is not de-allocated if cuda mempool is used (as Reserve() basically calls Allocate() for the cuda mempool)

onnxruntime/core/providers/cuda/cuda_mempool_arena.h
```diff
@@ -0,0 +1,169 @@
...
+/**
+ * @brief Stream-aware CUDA allocator implemented on top of a private `cudaMemPool_t`.
+ *        The purpose of this arena is to assist with memory allocations in environments where
+ *        a single process is hosting more than one cuda session. This arena hosts cuda memory pool
+ *        which has some tunable parameters to control its memory usage and de-allocates memory back to
+ *        the device according to the specified params. This is opposite to the BFCArena which only
+ *        attempts to free memory on Shrink() at the end of the run.
+ *
+ * ### Behavior
+ * - Creates a **process-local** CUDA mempool for a specific device (from `OrtMemoryInfo`).
```
hariharans29, line 29: Can you please explain the how "process-local" is achieved? If a user had multiple sessions in the same process with all  of them using the CUDA EP, wouldn't they all instantiate their own memory pool via `cudaMemPoolCreate()` ?

onnxruntime/core/providers/cuda/cuda_mempool_arena.h
```diff
@@ -0,0 +1,169 @@
...
+namespace logging {
+class Logger;
+}
+namespace cuda {
+/**
+ * @brief Stream-aware CUDA allocator implemented on top of a private `cudaMemPool_t`.
+ *        The purpose of this arena is to assist with memory allocations in environments where
+ *        a single process is hosting more than one cuda session. This arena hosts cuda memory pool
+ *        which has some tunable parameters to control its memory usage and de-allocates memory back to
+ *        the device according to the specified params. This is opposite to the BFCArena which only
```
hariharans29, line 25: Not understanding this comment: "This is opposite to the BFCArena...."

IIUC - this mempool also gets trimmed only during `Shrink()` which gets called only on `OnRunEnd()`. Wouldn't that behavior be exactly the same as the BFCArena's ?

onnxruntime/core/providers/cuda/cuda_mempool_arena.cc
```diff
@@ -0,0 +1,239 @@
...
+  // All allocations explicitly target this pool via cudaMallocFromPoolAsync.
+}
+
+CudaMempoolArena::~CudaMempoolArena() {
+  // 1) Best-effort: enqueue frees for any remaining allocations on their recorded streams.
+  //    No locking by design: destruction implies no concurrent access.
+  for (auto& kv : alloc_map_) {
+    void* p = kv.first;
+    const cudaStream_t s = kv.second.stream;
+    (void)cudaFreeAsync(p, s);  // ignore errors in destructor
```
hariharans29: Nit: Should we use ORT_IGNORE_RETURN_VALUE() in all place where we ignore return values instead of (void)(...)

onnxruntime/core/providers/cuda/cuda_mempool_arena.cc
```diff
@@ -0,0 +1,239 @@
...
+  }
+
+  // Now it is safe to drop our bookkeeping.
+  alloc_map_.clear();
```
hariharans29, line 67: We clear `stream_map_` here but it is actually used in the next operation - `SyncAllKnownStreams_NoThrow()` ?

onnxruntime/core/providers/cuda/cuda_mempool_arena.h
```diff
@@ -0,0 +1,169 @@
...
+  /**
+   * @brief Enqueue `cudaFreeAsync()` for all allocations made on @p stream.
+   * we intentionally do not implement this method. The call to this method
+   * will yank memory from under live OrtValues such as allocated for output
+   * bound and the resulting output OrtValue will not be valid.
+   * Then when the OrtValues attempt to release memory those entries are not found
+   * in the map: CudaMempoolArena::Free: pointer 0000000203800400 not found in allocation map; ignoring
+   * The reason this works with BFCArena is because it does not really release memory.
+   */
+  // void ReleaseStreamBuffers(Stream* stream) override;
```
hariharans29, line 125: Can ReleaseStreamBuffers() just queue cudaSyncFrees on the stream like mentioned several lines above:

 * - `Free()` and `ReleaseStreamBuffers()` enqueue **`cudaFreeAsync()`** on the recorded stream to
 *   respect CUDA's stream-ordered semantics.

This would ensure that memory doesn't get yanked from live OrtValues as the freed cuda memory just goes back to the memory pool ?

onnxruntime/core/framework/bfc_arena.h
```diff
@@ -525,11 +517,7 @@ class StreamAwareArena : public BFCArena {
   // Standard alloc behavior. Returns valid pointer if size > 0 and memory was available. Otherwise returns nullptr.
   void* AllocOnStream(size_t size, Stream* current_stream_id) override;
 
-  void ReleaseStreamBuffers(Stream* stream);
-
-  static StreamAwareArena* FromBFCArena(BFCArena& arena) {
-    return arena.GetArenaType() == ArenaType::StreamAwareArena ? reinterpret_cast<StreamAwareArena*>(&arena) : nullptr;
-  }
+  void ReleaseStreamBuffers(Stream* stream) override;
```
hariharans29, line 520: Should `StreamAwareArena` be renamed to something like `StreamAwareBFCArena` to show its connection with `BFCArena` and to avoid confusion with the new `CudaMempoolArena` which is also a stream aware arena ?

onnxruntime/core/providers/cuda/cuda_mempool_arena.cc
gedoensmax: In this exact case i do not expect a difference in both implementations. Both will block CPU execution. Chat GPT would be correct if you would wait outside this function: (malloc -> event record -> other submits -> event wait) but we need to wait inside to guarantee correct execution order. I would put a stream sync and put a lot of emphasis on not using this variant :)

yuslepukhin: I will publish the performance numbers shortly.

onnxruntime/core/framework/bfc_arena.h
yuslepukhin: Shrink() is not `BFCArena` specific. Both BFCArena and the new arena now implement IArena interface. `Shrink()` is called on either of them. 

Memory allocated with `Reserve()` on Cuda mempool is allocated from the same pool now and is subject to shrink. And this is actually one of the advantages of the new pool. Still allocated memory is obviously not affected.

Yes, only `cuda*` settings apply to cuda mempool.

onnxruntime/core/providers/cuda/cuda_mempool_arena.h
yuslepukhin: Cuda has a default memory pool that is global to the host. This is where the allocations come from.
To have control over the allocations one needs to instantiate its own mem pool and tune the parameters.
Such a mem pool is process local. Yes, if you have multiple sessions if the same process you would have more than one process local mempool. In both cases, we allow better memory sharing on the box by releasing memory that is not necessary. I will publish the numbers.

onnxruntime/core/providers/cuda/cuda_mempool_arena.h
yuslepukhin: `Shrink()` is called from InferenceSession at the end of the Run() if configured. In that sense the behavior is the same.
BFCArena marks the memory as deallocated, but it releases it back to the OS only if the entire region is free. 

Cuda mempool operates in separate allocations which can be released on Shrink().
However, in both cases `Shrink()` is expensive in terms of latency and customers do like it. Especially, if BFCArena does not release anything.

Cuda mempool has an advantage that it does not need `Shrink()`. Using the release threshold setting it can trim the pool at on stream synchronizations and at some points may re-use the same memory. See the documentation.

onnxruntime/core/providers/cuda/cuda_mempool_arena.h
yuslepukhin: Yes, this is what it does. All allocs and frees are async. And they get executed before the customer gets access to the output OrtValue.
The issue here is not that is not queued. It is queued before OrtValue gets destroyed which would be the right point to queue the Free() and this is what is now takes place with Cuda mempool not implementing `ReleaseStreamBuffers()`. I personally think interface is not needed.
Even better, we should implement FreeOnStream and retain stream inside either a smart ptr or Tensor deleter.

onnxruntime/core/framework/bfc_arena.h
yuslepukhin: I think it is a good idea. It is ugly. BFCArena is actually stream aware.

onnxruntime/core/framework/bfc_arena.h
hariharans29: Thank you. I got all that looking at the code. I was just pointing out this commentery before `Shrink()` needs a corresponding update to call out all that you just explained.

onnxruntime/core/framework/bfc_arena.h
hariharans29: If I understand correctly, BFCArena (in itself default state) itself is **not** stream aware ? It is the `StreamAwareArena` which inherits from `BFCArena` that is stream aware. 

And yes, I too think their connection with each other can be marked explicitly by renaming StreamAwareArena as StreamAwareBFCArena

yuslepukhin: > If I understand correctly, BFCArena (in itself default state) itself is **not** stream aware ? It is the `StreamAwareArena` which inherits from `BFCArena` that is stream aware.

Not really. All StreamAwareArena does returns true on `IsStreamAware()` interface, all the implementation is in BFCArena.

onnxruntime/core/providers/cuda/cuda_mempool_arena.h
```diff
@@ -0,0 +1,169 @@
...
+ *        a single process is hosting more than one cuda session. This arena hosts cuda memory pool
+ *        which has some tunable parameters to control its memory usage and de-allocates memory back to
+ *        the device according to the specified params. This is opposite to the BFCArena which only
+ *        attempts to free memory on Shrink() at the end of the run.
+ *
+ * ### Behavior
+ * - Creates a **process-local** CUDA mempool for a specific device (from `OrtMemoryInfo`).
+ * - All allocations use **`cudaMallocFromPoolAsync()`** on either the legacy default stream (0) or a
+ *   caller-provided stream. The allocation stream is recorded for ordered free.
+ * - `Free()` and `ReleaseStreamBuffers()` enqueue **`cudaFreeAsync()`** on the recorded stream to
```
hariharans29: Nit: Misleading comment - `ReleaseStreamBuffers()` is default to the base implementation (which is a no-op)

onnxruntime/core/providers/cuda/cuda_mempool_arena.h
hariharans29: Thanks for the explanation.

In that vein - I guess the comment above needs an update to reflect this (https://github.com/microsoft/onnxruntime/pull/26535/files#r2528785161)

onnxruntime/test/providers/cuda/cuda_mempool_arena_test.cc
```diff
@@ -25,26 +25,33 @@ namespace test {
 static bool IsCudaMemPoolSupported() {
   int ort_cuda_rt_version = 0;
   cudaError_t cuda_status = cudaRuntimeGetVersion(&ort_cuda_rt_version);
-  bool version_supported = (cuda_status == cudaSuccess && ort_cuda_rt_version >= 11020);
-  if (!version_supported) {
+  if (cuda_status != cudaSuccess) {
```
hariharans29, line 28: Nit: I guess we can just re-use IsCudaVersionSupported() from the memory pool class ?

yuslepukhin: Not really, the impl in the separate DLL

hariharans29:
Side question: How does the new memory pool play with cuda graphs ? Is it expected to work ?

hariharans29 (approved):

yuslepukhin:
> Side question: How does the new memory pool play with cuda graphs ? Is it expected to work ?

I think the difference is that we have a single stream. I would think this is just an allocator doing async allocations on a stream, as before.

tianleiwu (dismissed):

hariharans29 (approved):

hariharans29:
> > Side question: How does the new memory pool play with cuda graphs ? Is it expected to work ?
> 
> I think the difference is that we have a single stream. I would think this is just an allocator doing async allocations on a stream, as before.

FWIW - I think it is not safe to use cuda graphs with this allocator unless we are absolutely certain we have picked a high enough `cuda_mempool_release_threshold` such that no memory will get returned back to the system on a synchronization event. 

Cuda graphs require same memory as the graph capture step for the replays to happen. If we can't guarantee that the same memory will be held by the process, using cuda graphs are unsafe for usage in that scenario. BFCArena is safe in that sense because it never really released memory back to the system.

PR #26542, opened 2025-11-10
[TRT RTX EP] Don't register set device function when we use existing stream

hadiFute:
### Description
- Don't register set device function when we use existing stream
- Fix bug nv_execution_provider.cc : set device only if user did not provide existing stream

### Motivation and Context
In some use cases, we push a user generated CUDA context, create streams using this context, and then provide these streams to TRT-RTX.
However, we noticed that after calling Run(), the custom context is replaced by another CUDA context created by ORT. This means that TRT-RTX is no longer using the original CUDA context.
After investigating further, we found that the new context is being created in onnxruntime/core/framework/stream_execution_context.cc.
The solution we propose is to not register set device function if we provide the stream. 

Also there is a bug in onnxruntime\core\providers\nv_tensorrt_rtx\nv_execution_provider.cc.
We should set the device only if the user has not provided any stream. (coherent with the original comment)


hadiFute:

hadiFute:
For review @chilo-ms @edgchen1 @ishwar-raut1

anskumar01:
@chilo-ms , can you please review and merge it?

yuslepukhin:
Looks good to me. @tianleiwu, please, take a look.

tianleiwu (approved):

yuslepukhin (approved):
:shipit:

PR #26543, opened 2025-11-10
Graphics ORT interop design and implementation

praneshgo:
### Description

- Introduced external GPU synchronization primitives enabling interoperability with Direct3D 12 and Vulkan graphics APIs.
- Added APIs to acquire, wait on, and signal external GPU synchronization fences or semaphores in execution providers.
- Enabled setup of CUDA Interop Graphics (CIG) contexts to synchronize CUDA streams with graphics API command queues/devices.
- Extended NVIDIA TensorRT RTX execution provider with support for importing and synchronizing external semaphores from graphics APIs.
- Added execution provider interface methods for managing external synchronization primitives and interop synchronization.
- Added build options to enable DirectX and Vulkan interoperability features.
- Added comprehensive GPU interoperability test validating D3D12 and ONNX Runtime synchronization with TensorRT RTX.

### Motivation and Context
<!-- - Why is this change required? What problem does it solve?
The proposal aims to address interoperability challenges between Graphics APIs like DX/Vulkan and ORT with specific focus on introducing new APIs that enable DX/Vulkan and ORT workloads to synchronize seamlessly, ensuring tasks are scheduled coherently in the hardware timeline—both in the same TSG (Thread Scheduling Group) whenever possible for maximum scheduling efficiency.
One of our design goals is to have no EP specific APIs visible at the application level. The APIs are designed to be general enough to cater to all possible execution providers, with a CPU fallback in place. 
This interoperability enables even non-game developers to use a standard GPU compute API (DX12/Vulkan) to perform pre/post-processing in a IHV agnostic way.

### DIFF
--- a/include/onnxruntime/core/session/onnxruntime_c_api.h
+++ b/include/onnxruntime/core/session/onnxruntime_c_api.h
@@ -6631,21 +6631,19 @@ struct OrtApi {
 
   /** \brief Setup Graphics Interopcontext for an execution provider device.
    *
-   * This function enables D3D12/Vulkan interoperability by creating a Graphics context
-   * associated with a graphics API command queue/device. Once setup, any OrtSyncStream
-   * created for this ep_device via CreateSyncStreamForEpDevice will be created on the
-   * Graphics context, enabling efficient GPU-side synchronization.
+   * This function enables D3D12/Vulkan interoperability with a graphics API command queue/device. Once setup, any OrtSyncStream
+   * created for this ep_device via CreateSyncStreamForEpDevice will be created, enabling efficient GPU-side synchronization.
    *
    * This must be called BEFORE CreateSyncStreamForEpDevice for the same ep_device.
    *
-   * \param[in] ep_device The OrtEpDevice to setup Graphics context for.
+   * \param[in] ep_device The OrtEpDevice to setup Graphics interop for.
    * \param[in] graphicsInteropParams Pointer to struct containing D3D12 command queue or Vulkan device info.
    *
    * \snippet{doc} snippets.dox OrtStatus Return Value
    *
    * \since Version 1.24
    */
-  ORT_API2_STATUS(SetupGraphicsInteropContextForEpDevice, _In_ const OrtEpDevice* ep_device,
+  ORT_API2_STATUS(SetupGraphicsInteropForEpDevice, _In_ const OrtEpDevice* ep_device,
                   _In_ const struct GraphicsInteropParams* graphicsInteropParams);
 
   /**
@@ -6659,7 +6657,7 @@ struct OrtApi {
    * GPU-side synchronization between ONNX Runtime inference and external graphics workloads.
    *
    * Prerequisites:
-   * - SetupGraphicsInteropContextForEpDevice must be called first for the relevant ep_device.
+   * - SetupGraphicsInteropForEpDevice must be called first for the relevant ep_device.
    * - The external fence/semaphore provided must be compatible with the graphics API specified in graphicsInteropParams.
    *
    * \param[in] session An OrtSession instance whose execution providers participate in the graphics interop.
@@ -6688,7 +6686,6 @@ struct OrtApi {
    *
    * The implementation and support for this is execution provider-specific.
    *
-   * \param[in] session The OrtSession instance for which synchronization is required.
    * \param[in] extSemFence The handle to the external synchronization primitive, as returned from GetOrtFenceForGraphicsInterop.
    * \param[in] stream The OrtSyncStream instance on which the synchronization will be performed.
    * \param[in] fenceValue The fence value for synchronization (if required for the specific graphics API/interop scenario).
@@ -6699,7 +6696,7 @@ struct OrtApi {
    *
    * \since Version 1.24
    */
-  ORT_API2_STATUS(InteropEpWait, _In_ OrtSession* session, _In_ OrtFence* ortFence, _In_ OrtSyncStream* stream, _In_ uint64_t fenceValue);
+  ORT_API2_STATUS(InteropEpWait, _In_ OrtFence* ortFence, _In_ OrtSyncStream* stream, _In_ uint64_t fenceValue);
 
   /**
    * \brief Signal a graphics interop external fence/semaphore using an ONNX Runtime execution provider.
@@ -6713,7 +6710,6 @@ struct OrtApi {
    *
    * The behavior and support for this operation is execution provider-specific.
    *
-   * \param[in] session   The OrtSession instance corresponding to the computation that should be synchronized.
    * \param[in] extSemFence The handle to the external synchronization primitive, as returned from GetOrtFenceForGraphicsInterop.
    * \param[in] stream    The OrtSyncStream on which the signal operation should occur.
    * \param[in] fenceValue The fence value to signal (if required for the specific graphics API or interop scenario).
@@ -6724,7 +6720,7 @@ struct OrtApi {
    *
    * \since Version 1.24
    */
-  ORT_API2_STATUS(InteropEpSignal, _In_ OrtSession* session, _In_ OrtFence* ortFence, _In_ OrtSyncStream* stream, _In_ uint64_t fenceValue);
+  ORT_API2_STATUS(InteropEpSignal, _In_ OrtFence* ortFence, _In_ OrtSyncStream* stream, _In_ uint64_t fenceValue);
 };
 
 /*
--- a/include/onnxruntime/core/session/onnxruntime_ep_c_api.h
+++ b/include/onnxruntime/core/session/onnxruntime_ep_c_api.h
@@ -1010,7 +1010,6 @@ struct OrtEpFactory {
    * This is used to create a synchronization stream for the memory device that can be used for operations outside of
    * a session.
    *
-   * If SetupGraphicsContext was called previously for this memory device, the stream will be created on the Graphics context.
    *
    * \param[in] this_ptr The OrtEpFactory instance.
    * \param[in] memory_device The OrtMemoryDevice to create the synchronization stream for.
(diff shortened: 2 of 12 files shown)


cmake/CMakeLists.txt
```diff
@@ -244,6 +244,31 @@ option(onnxruntime_USE_AZURE "Build with azure inferencing support" OFF)
 option(onnxruntime_USE_LOCK_FREE_QUEUE "Build with lock-free task queue for threadpool." OFF)
 option(onnxruntime_FORCE_GENERIC_ALGORITHMS "Disable optimized arch-specific algorithms. Use only for testing and debugging generic algorithms." OFF)
 
+# DX for interop feature option
+option(onnxruntime_USE_DX_FOR_INTEROP "Build with the DX for Interop feature." OFF)
+
+if (onnxruntime_USE_DX_FOR_INTEROP)
+  add_compile_definitions(DX_FOR_INTEROP=1)
+else()
+  add_compile_definitions(DX_FOR_INTEROP=0)
```
skottmckay: Does adding this make it harder to use as now I have to check the value and not just that it's defined?

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -507,6 +508,43 @@ typedef enum OrtExecutionProviderDevicePolicy {
...
+    void* pVkFence;
+    void* VkSemaphore;
+  } FencePtr;
+  VulkanDeviceParams VulkanDeviceParams;
+} FenceInteropParams;
+
+typedef struct SemaphoreEpMap {
+  void* extSemFence;
+  void* selectedEp;
+} SemaphoreEpMap;
```
skottmckay, line 546: The usage of 'fence' vs 'semaphore' in a number of places is confusing/inconsistent. Would be good to explain what exactly is used for DX vs VK and why. I expected these changes were only for device to device synchronization yet based on the VK doco a vkFence is for CPU to GPU sync so it's not clear to me why we need that.

We have a *Semaphore*EpMap with an `extSemFence` which I assume could be either but the type name doesn't indicate that. 
Conflicting with that is *Fence*InteropParams that contains fences or a semaphore.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -6590,6 +6628,99 @@ struct OrtApi {
    * \since Version 1.24
    */
   ORT_API_T(bool, TensorTypeAndShape_HasShape, _In_ const OrtTensorTypeAndShapeInfo* info);
+
+  /** \brief Setup Graphics Interopcontext for an execution provider device.
+   *
+   * This function enables D3D12/Vulkan interoperability by creating a CUDA context
```
skottmckay: Anything in the ORT API needs to be EP agnostic. I think it's ok for the documentation to give an example that's CUDA specific, but the general documentation on what the function does should be generic.

include/onnxruntime/core/framework/execution_provider.h
```diff
@@ -92,6 +100,86 @@ class IExecutionProvider {
  public:
   virtual ~IExecutionProvider() = default;
 
+  virtual Status GetExtSemaphore(const struct GraphicsInteropParams* graphicsInteropParams, struct FenceInteropParams* fenceInteropParams, void** extSemFence) {
```
skottmckay: IExecutionProvider is an internal type. As such I wouldn't expect it's API to be using things like `void** extSemFence` as conversion from opaque API types to the 'real' internal types should be handled by the `PluginExecutionProvider` class which implements IExecutionProvider for plugin EPs.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -6590,6 +6628,99 @@ struct OrtApi {
...
+   *
+   * \param[in] ep_device The OrtEpDevice to setup CIG context for.
+   * \param[in] graphicsInteropParams Pointer to struct containing D3D12 command queue or Vulkan device info.
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \since Version 1.24
+   */
+  ORT_API2_STATUS(SetupGraphicsInteropContextForEpDevice, _In_ const OrtEpDevice* ep_device,
+                  _In_ const struct GraphicsInteropParams* graphicsInteropParams);
```
skottmckay: Naively I would have expected this to be setup on a per-Session basis. An OrtEpDevice is an Environment level type so used for all sessions.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -1052,6 +1054,25 @@ struct OrtEpFactory {
...
+   *
+   * \param[in] this_ptr The OrtEpFactory instance.
+   * \param[in] memory_device The OrtMemoryDevice to setup CIG context for.
+   * \param[in] graphicsInteropParams Graphics API parameters (D3D12 command queue or Vulkan device info).
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \since Version 1.24.
+   */
+  ORT_API2_STATUS(SetupCigContext, _In_ OrtEpFactory* this_ptr,
```
skottmckay: This functionality needs to be made generic.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -6590,6 +6628,99 @@ struct OrtApi {
...
+   * \param[in] stream    The OrtSyncStream on which the signal operation should occur.
+   * \param[in] fenceValue The fence value to signal (if required for the specific graphics API or interop scenario).
+   *
+   * \retval ORT_OK                On successful signal operation.
+   * \retval ORT_NOT_IMPLEMENTED   If the current execution provider does not support graphics interop signaling.
+   * \retval ORT_FAIL or provider-specific error if signaling fails for another reason.
+   *
+   * \since Version 1.24
+   */
+  ORT_API2_STATUS(InteropEpSignal, _In_ OrtSession* session, _In_ OrtFence* ortFence, _In_ OrtSyncStream* stream, _In_ uint64_t fenceValue);
```
skottmckay: Not sure about the naming of InteropEpWait/InteropEpSignal as the usage of EP/s within a session is somewhat opaque.

I would have expected the wait/signal to be around a call to OrtSession Run, with ORT figuring out when/where an EP is connected to a fence/semaphore for input/output. Should the wait/signal instead by coupled to the Run? The current setup also means you could not have concurrent calls to Run which feels like it could be overly restrictive.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -6590,6 +6628,99 @@ struct OrtApi {
...
+   * \param[in] graphicsInteropParams Pointer to a struct defining which graphics API interop is desired, providing descriptor handles, device pointers, etc.
+   * \param[out] extSemFence Pointer to a variable that will receive the external semaphore/fence handle. The handle is suitable for passing to external graphics APIs.
+   *
+   * \retval ORT_OK On success.
+   * \retval ORT_NOT_IMPLEMENTED If none of the active providers support graphics-fence interop.
+   * \retval ORT_FAIL or provider-specific error if the operation fails for another reason.
+   *
+   * \since Version 1.24
+   */
+  ORT_API2_STATUS(GetOrtFenceForGraphicsInterop, _In_ OrtSession* session, _In_ const struct GraphicsInteropParams* graphicsInteropParams, _In_ struct FenceInteropParams* fenceInteropParams, _Outptr_ OrtFence** ortFence);
```
skottmckay: Who owns `ortFence` and how/when should it be released?

Comments seem to be out of date.

Why do fenceInteropParams need to be non-const?

onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
```diff
@@ -104,6 +111,112 @@ static bool IsSupportedInputOutputDataType(ONNXTensorElementDataType data_type)
   }
 }
 
+Status NvExecutionProvider::GetExtSemaphore(const struct GraphicsInteropParams* graphicsInteropParams, struct FenceInteropParams* fenceInteropParams, void** extSemFence)
+{
+  if (!info_.has_user_compute_stream)
```
skottmckay: `has_user_compute_stream` is very specific to this EP. how will this work for an EP that doesn't currently support user provided streams? would they create a stream using the ORT API and pass it in somehow (e.g. in the params to one of the new API calls)? we don't really want to manually have to add an option to all the EPs for this sort of thing.

include/onnxruntime/core/session/onnxruntime_c_api.h
praneshgo: Application owns ortFence and it is responsible for releasing it when done.
Updated comments in the latest commit.
fenceInteropParams is made const struct.

include/onnxruntime/core/session/onnxruntime_c_api.h
praneshgo: Updated the comments and API to be EP agnostic.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
praneshgo: Updated the API name to SetupGraphicsContext to make it more generic. The functionality is generic already, EPs are expected to have their implementation in EP code files. In case of NV TRT RTX EP, SetupGraphicsContext = SetupCigContextImpl. This assignment and implementation of SetupCigContextImpl is present in onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc

include/onnxruntime/core/session/onnxruntime_c_api.h
praneshgo: Renamed to SetupGraphicsInteropForEpDevice to better reflect its usage. The idea is that this API must be called at the beginning in order to setup graphics interop on all the sessions associated with an OrtEpDevice correctly.

include/onnxruntime/core/session/onnxruntime_c_api.h
praneshgo: > Not sure about the naming of InteropEpWait/InteropEpSignal as the usage of EP/s within a session is somewhat opaque.

InteropEpWait and InteropEpSignal are designed as generic synchronization APIs that users can invoke on demand. They operate independently of the session context. To make this explicit and remove ambiguity, I have updated both API signatures to remove the OrtSession parameter.

> The current setup also means you could not have concurrent calls to Run which feels like it could be overly restrictive.

Can you kindly provide an example of this scenario, so I may better understand?

onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
praneshgo: There seems to be no way today to associate an external OrtSyncStream with a session in an EP-agnostic way—something like sessionOptions.setOrtSyncStream. If there is an option to do that, we can certainly avoid passing EP-specific options like has_user_compute_stream.

include/onnxruntime/core/framework/execution_provider.h
praneshgo: That makes sense. I will modify the code in the next commit based on this suggestion.

skottmckay:
Would be great to get your initial thoughts on #26821. I think that's more consistent in approach to other things the plugin EP is providing, as well as with the standard Create/Release setup the ORT API uses. The proposed approach also includes memory handling which other IHVs have asked for. I think there are some gaps, but that structure feels a lot cleaner to me with more explicit usage.

I have a lot of comments about the PR having spent the last day or so going through it, but if we want to move to something closer to that setup quite a few will be invalidated.

Some apply here as well though
- should the OrtSyncStream to use for an inference be passed in as part of the call to Run so that there's potentially more than one stream that can be used for different inferences/sessions
- would be better to get the OrtEpDevice for an input from SessionGetEpDeviceForInputs so you know for sure that the EP will be consuming the input. 
- all structs need to be versioned

Are the Vulkan changes here tested? I didn't see anything Vulkan specific in the unit test.

skottmckay:
Can you take a look at the PRs Nick has added with initial implementation to see if there are any gaps in what that provides?

#26828
#26829

praneshgo:
> Can you take a look at the PRs Nick has added with initial implementation to see if there are any gaps in what that provides?
> 
> #26828 #26829

Hi Scott
Sure. I will take a look at these PRs and provide feedback. Thanks.

skottmckay:
> > Can you take a look at the PRs Nick has added with initial implementation to see if there are any gaps in what that provides?
> > #26828 #26829
> 
> Hi Scott Sure. I will take a look at these PRs and provide feedback. Thanks.

Important that the setup works for Level Zero as well, which I don't believe has an equivalent to cuCtxCreate_v4 (please correct me if I'm wrong). 

Possibly we could support using external semaphores as the general-purpose usage and maybe have an 'attach to device' option to enable using cuCtxCreate_v4 as an additional optimization. If you attached to the device first, would the pattern of calling 'import semaphore' followed by wait/signal with an OrtSyncStream and the result of the import still be viable? Possibly the implementation would use different types if attached to device, but if the API is generic enough that should all be opaque to ORT.

PR #26544, opened 2025-11-10
[TRT/TRT RTX EP] EP gets weights as OrtValues at compile time to refit weightless engine

chilo-ms:
### Description
When TRT/TRT RTX EP trying to refit the weightless engine using external data in memory, currently, passing external data bytes stream via ep options is the only option. This PR makes provider-bridge TRT/TRT RTX EP be able to get the external data in memory via OrtValues.
Please see `TensorrtExecutionProvider::GetInMemoryInitializers` for implementation.

Either ORT converts initializers to OrtValues during session creation, or EP explicitly calls related API to convert initializers to OrtValues.

In some cases, e.g. using Compile API to compile the model, the initializers (still in form of TensorProto) haven't been converted to OrtValues by ORT yet at EP Compile(). Therefore, EP has to call `LoadExternalInitializerAsOrtValue` on its own to get the external data in memory for refitting the weightless engine.

### DIFF
--- a/onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
+++ b/onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
@@ -2506,9 +2506,14 @@ common::Status NvExecutionProvider::Compile(const std::vector<FusedNodeAndGraph>
       status = CreateNodeComputeInfoFromGraph(graph_body_viewer, fused_node, input_map, output_map, node_compute_funcs);
     }
     if (status != Status::OK()) {
+      initializer_values_.clear();
       return ORT_MAKE_STATUS(ONNXRUNTIME, FAIL, status.ErrorMessage());
     }
   }
+
+  // Release the OrtValues that are cached in TRT EP
+  initializer_values_.clear();
+
   return Status::OK();
 }
 
--- a/onnxruntime/core/providers/tensorrt/tensorrt_execution_provider.cc
+++ b/onnxruntime/core/providers/tensorrt/tensorrt_execution_provider.cc
@@ -2348,9 +2348,10 @@ SubGraphCollection_t TensorrtExecutionProvider::GetSupportedList(SubGraphCollect
         }
 #else
         if (load_user_initializer_) {
-          LOGS_DEFAULT(VERBOSE) << "TensorRT EP is built agasint TRT version < 10.12 which doesn't support separating out initializer data from model proto. 
-                                   "Initializer data will be included in model proto.;
-              load_user_initializer_ = false;
+          std::string warning_msg = "TensorRT EP is built agasint TRT version < 10.12 which doesn't support separating out initializer data from model proto."
+                                    "Initializer data will be included in model proto.";
+          LOGS_DEFAULT(WARNING) << warning_msg;
+          load_user_initializer_ = false;
         }
 #endif
 
@@ -3159,9 +3160,14 @@ common::Status TensorrtExecutionProvider::Compile(const std::vector<FusedNodeAnd
       status = CreateNodeComputeInfoFromGraph(graph_body_viewer, fused_node, input_map, output_map, node_compute_funcs);
     }
     if (status != Status::OK()) {
+      initializer_values_.clear();
       return ORT_MAKE_STATUS(ONNXRUNTIME, FAIL, status.ErrorMessage());
     }
   }
+
+  // Release the OrtValues that are cached in TRT EP
+  initializer_values_.clear();
+
   return Status::OK();
 }
 
@@ -3186,9 +3192,11 @@ Status TensorrtExecutionProvider::CreateNodeComputeInfoFromGraph(const GraphView
   }
 #else
   if (load_user_initializer_) {
-    LOGS_DEFAULT(VERBOSE) << "TensorRT EP is built agasint TRT version < 10.12 which doesn't support separating out initializer data from model proto. 
-                             "Initializer data will be included in model proto.;
-        load_user_initializer_ = false;
+    std::string warning_msg =
+        "TensorRT EP is built agasint TRT version < 10.12 which doesn't support separating out initializer data from model proto."
+        "Initializer data will be included in model proto.";
+    LOGS_DEFAULT(WARNING) << warning_msg;
+    load_user_initializer_ = false;
   }
 #endif
 
--- a/onnxruntime/core/providers/tensorrt/tensorrt_execution_provider.h
+++ b/onnxruntime/core/providers/tensorrt/tensorrt_execution_provider.h
@@ -369,7 +369,7 @@ class TensorrtExecutionProvider : public IExecutionProvider {
   bool engine_hw_compatible_ = false;
   std::string op_types_to_exclude_;
   std::vector<nvinfer1::PreviewFeature> preview_features_;
-  bool load_user_initializer_ = false;
+  mutable bool load_user_initializer_ = false;
 
   // Cache initializer's external data as an OrtValue
   mutable std::unordered_map<std::string_view, std::unique_ptr<OrtValue>> initializer_values_;


chilo-ms:
@gaugarg-nv @thevishalagarwal 
This PR updates both TRT EP and TRT RTX EP with the same function to fetch external weights as OrtValues in memory so that the weights can be fed to TRT parser to refit the weightless engine.
Please help take a look, thanks!

onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
```diff
@@ -2222,6 +2222,65 @@ NvExecutionProvider::GetCapability(const GraphViewer& graph,
...
+                    " has in-memory external data but cannot get OrtValue.");
+        }
+        continue;
+      }
+
+      // If external data is not in memory, meaning ORT hasn't converted it to a OrtValue yet at this moment,
+      // then loads this initializer with data in an external file into an OrtValue.
+      // Note: The OrtValue is not cached in the graph, so we need to copy data.
+      ORT_RETURN_IF_ERROR(graph_body_viewer.GetGraph().LoadExternalInitializerAsOrtValue(tp->name(), ort_value));
+      populate_tensorrt_weight_with_ort_value(ort_value, true);
```
adrianlizarraga: Note that instead of copying the data, the EP _could_ store/cache the `OrtValue` itself. This would prevent a copy. I'm not saying this is the right approach for this EP, but just noting it just in case.

onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
```diff
@@ -2222,6 +2222,65 @@ NvExecutionProvider::GetCapability(const GraphViewer& graph,
...
+
+    }
+    // Initializer has raw data
+    else if (utils::HasRawData(*tp)) {
```
adrianlizarraga, line 2258: I think there is a third case that is missed: tensor proto has data in `int32_data()`, for example. Is this omission intentional?

onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
```diff
@@ -2222,6 +2222,65 @@ NvExecutionProvider::GetCapability(const GraphViewer& graph,
...
+common::Status NvExecutionProvider::GetInMemoryInitializers(const GraphViewer& graph_body_viewer,
+                                                                  std::unordered_map<std::string, TensorrtUserWeights>& user_weights) const {
+  const InitializedTensorSet& all_initializers = graph_body_viewer.GetAllInitializedTensors();
+  user_weights.reserve(all_initializers.size());
+
+  for (auto& entry : all_initializers) {
+    auto* tp = entry.second;
+
+    // Initializer has external data
+    if (utils::HasExternalData(*tp)) {
```
yuslepukhin: Does this EP requires a copy or you could just store the OrtValue and retain pointers where needed?

onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
```diff
@@ -2328,23 +2390,36 @@ common::Status NvExecutionProvider::RefitEngine(std::string onnx_model_filename,
...
+                sizes.push_back(weight.Size());
+                continue;
+              }
+            }
+
+            if (refit_with_external_data_bytestream) {
+              // Default values for reading into external_data blob.
+              int64_t offset = 0;
+              size_t length = 0;
+              auto external_data = proto.mutable_external_data();
```
yuslepukhin, line 2390: I would check the ExternalData API presense in the bridge, so you do not need to write parsing code again. One has to do some overflow checks too.

onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
```diff
@@ -2232,10 +2291,13 @@ common::Status NvExecutionProvider::RefitEngine(std::string onnx_model_filename,
                                                 size_t onnx_model_bytestream_size,
                                                 const void* onnx_external_data_bytestream,
                                                 size_t onnx_external_data_bytestream_size,
+                                                std::unordered_map<std::string, TensorrtUserWeights>& in_memory_weights,
```
yuslepukhin, line 2276: here and below, coding standards require use of InlinedHashMap

onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
```diff
@@ -3208,6 +3287,11 @@ Status NvExecutionProvider::CreateNodeComputeInfoFromPrecompiledEngine(const Gra
   std::unordered_map<std::string, size_t> output_indexes;  // TRT engine output name -> ORT kernel context output index
   std::unordered_map<std::string, size_t> output_types;    // TRT engine output name -> ORT output tensor type
 
+  std::unordered_map<std::string, TensorrtUserWeights> in_memory_weights;
```
yuslepukhin, line 3260: std containers allocate a node even if not used, so InlinedHashMap is always better

onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
```diff
@@ -2903,13 +2978,17 @@ Status NvExecutionProvider::CreateNodeComputeInfoFromGraph(const GraphViewer& gr
 
   if (weight_stripped_engine_refit_) {
     LOGS_DEFAULT(VERBOSE) << "[NvTensorRTRTX EP] Refit engine from main ONNX file after engine build";
+    std::unordered_map<std::string, TensorrtUserWeights> in_memory_weights;
```
yuslepukhin, line 2951: What would be different between on disk weights and in-memory weights if they are both represented by an OrtValue? Should it matter to differentiate>?

thevishalagarwal:
++ @gedoensmax @kevinch-nv

onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
chilo-ms: The omission is intentional.
My understanding is that when building TRT weightless engine from ONNX model, the data in type specific data filed in the TensorProto, e.g. `int32_data()`, will still be included in the engine, only those TensorProtos have external data will be excluded. 
So, we don't need to include those type specific data here for refitting.

@kevinch-nv, correct me if my understanding is wrong.

onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
chilo-ms: yea, i thought about caching OrtValue in EP but EP might want to release the cache once it's not used anymore. But still, that's another option.

onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
chilo-ms: TRT EP could store the OrtValue and retain the pointer, copy is not needed.
Let me try this approach.

onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
chilo-ms: Let me try this EP caching OrtValue approach.

onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
kevinch-nv: From TRT's point of view all weights are handled the same since the parser resolves all different types of initializers (external, raw data, int32_data, user-provided, etc) into a raw pointer.

So we should be adding logic to handle int32_data() here

PR #26549, opened 2025-11-11
[QNN-EP] SSR Handling Enablement

qti-hungjuiw:
### Description

- Introduced a new provider option: `enable_ssr_handling`
- Handle SSR before graphFinalize
    - `SetupBackend`: Re-run SetupBackend to re-create context as
    - `CompileFromOrtGraph`: Release invalid Context API objects and re-create Graph and Context
- Handle SSR after graphFinalize
    - `ExecuteGraph`: Release invalid Context, load the Context from the pre-saved binary and re-execute the graph
- Minor fix to function return values to propagate SSR-specific error codes
- Add test framework of QNNEP SSR Handling
    - Introduce `QnnMockSSR.dll`, which mimics the functionality of `QnnHtp.dll` but consistently triggers SSR to validate QNNEP behavior

### Motivation and Context

- QNN APIs support SSR by allowing users to release and recreate specific handles when a subsystem restart occurs.
- This PR enhances ORT's robustness by enabling it to handle SSR scenarios gracefully.
- It also introduces a mock implementation to facilitate testing and validation of SSR handling across different stages of the QNN workflow.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
@@ -999,6 +999,18 @@ std::unique_ptr<unsigned char[]> QnnBackendManager::GetContextBinaryBuffer(uint6
   return context_buffer;
 }
 
+Status QnnBackendManager::SaveContextToBinary(const logging::Logger& logger) {
+  LOGS(logger, VERBOSE) << "[SSR Handle] SaveContextToBinary for recover";
+  uint64_t writtenBufferSize{0};
+  qnn_save_buffer_ = GetContextBinaryBuffer(writtenBufferSize);
+  qnn_save_buffer_size_ = writtenBufferSize;
+
+  ORT_RETURN_IF(qnn_save_buffer_ == nullptr || writtenBufferSize == 0,
+                "Failed to get valid context binary buffer.");
+
+  return Status::OK();
+}
+
 Status QnnBackendManager::GetMaxSpillFillBufferSize(unsigned char* buffer,
                                                     uint64_t buffer_length,
                                                     uint64_t& max_spill_fill_buffer_size) {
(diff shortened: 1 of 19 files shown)


onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -2094,5 +2108,72 @@ Status QnnBackendManager::GetOrRegisterContextMemHandle(Qnn_ContextHandle_t cont
   return Status::OK();
 }
 
+Status QnnBackendManager::InvokeWithSSRHandle(
+    const std::function<Status()>& operation,
+    const std::function<Status()>& ssr_recover,
```
yuslepukhin, line 2113: Can these functions throw exceptions and what effect it would have on the state machine?

onnxruntime/core/providers/qnn/builder/qnn_model.h
```diff
@@ -105,6 +105,8 @@ class QnnModel {
     return outputs_info_;
   }
 
+  const std::unique_ptr<GraphInfo>& GetGraphInfo() const { return graph_info_; }
```
yuslepukhin, line 108: >std::unique_ptr<GraphInfo>& GetGraphInfo()  [](http://example.com/codeflow?start=8&length=43)

Would returning a const reference to GraphInfo be more logical since the function is const? Why a user needs this smart ptr?
Returning const ref to smart ptr is not the same.

onnxruntime/core/providers/qnn/builder/qnn_model.cc
```diff
@@ -132,9 +130,7 @@ Status QnnModel::ComposeGraph(const GraphViewer& graph_viewer,
   }
 #endif
```
yuslepukhin, line 131: Do you really need to time failed calls?

onnxruntime/core/providers/qnn/builder/qnn_model.cc
```diff
@@ -331,7 +324,7 @@ Status QnnModel::ExecuteGraph(const Ort::KernelContext& context,
         qnn_outputs.back()));
   }
 
-  Qnn_ErrorHandle_t execute_status = QNN_GRAPH_NO_ERROR;
+  Status execute_status = Status::OK();
```
yuslepukhin, line 327: >Status::OK(); [](http://example.com/codeflow?start=26&length=13)

default __ctor__ will do the right thing quickly

PR #26557, opened 2025-11-12
Tries to enable BFloat16 for PackedMultiHeadAttention

xadupre:
### Description

Enables PackedMultiHeadAttention with bfloat16.
Does not fix discrepancies issues with bfloat16

### Motivation and Context

1. PackedMultiHeadAttention is 3x faster than a loop over multiple MultiHeadAttention.
2. PackedMultiHeadAttention with bfloat16 is as fast as float16.
3. A loop over multiple MultiHeadAttention with float16 is 2x slower than PackedMultiHeadAttention.

But:

torch implementation and onnxruntime implementation match with float16 (for both MHA), but they don't match with bfloat16.

Results:

image

Measurements obtained with this file:

[test_packed_attention_configs.py](https://github.com/user-attachments/files/23527539/test_packed_attention_configs.py)


tianleiwu:
There are CI pipeline failures. Please take a look. Try merge latest main if the error is not caused by this PR.

PR #26558, opened 2025-11-12
Adds automatic loading of the external `tensorrt_plugins` library in the TensorRT-RTX Execution Provider to support custom transformer plugins

anujj:
External TensorRT plugins for transformer models were previously only available as internal TensorRT components. To enable flexible deployment these plugins have been externalized into a separate library.


anujj:
@gaugarg-nv @ishwar-raut1 @chilo-ms @edgchen1 for review

anujj:
@chilo-ms : can we merge this change, please ?

chilo-ms (approved):

PR #26561, opened 2025-11-13
[build] fix use dotnet

fs-eire:
### Description

The CI pipeline fails because although the `actions/setup-dotnet@v5` step uses `8.x`, the actual build process may still possibly run a latest version.

according to https://learn.microsoft.com/en-us/dotnet/core/tools/global-json, we need to prepare a `global.json` file in the working directory to ensure the correct version of dotnet is used.


fs-eire:
The "fix" does not work.

PR #26565, opened 2025-11-13
[WebNN EP] Support local attention feature for GQA

peishenyan:
### Description

Support the `local_window_size` attribute in **GroupQueryAttention** Operator, which is designed for sliding window attention and may influence the attention mask pattern.

For local window size not equal to -1, new attention mask pattern will be created as follows for applying sliding window.
```
     condition_1 (old attn_mask) ---> CumSum (axis=3, exclusive=true, reversed=true)
          |                             |
          |                           Lesser <--- local_window_size
          |                             |
      LogicalAnd <----------------- condition_2
          |
    new attn_mask
```

### DIFF
--- a/onnxruntime/core/providers/webnn/builders/impl/gqa_op_builder.cc
+++ b/onnxruntime/core/providers/webnn/builders/impl/gqa_op_builder.cc
@@ -295,14 +295,13 @@ Status GroupQueryAttentionOpBuilder::AddToModelBuilderImpl(ModelBuilder& model_b
                                                 |
                                           attention_bias
   */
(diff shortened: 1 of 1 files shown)


peishenyan:
PTAL, thanks.  @Honry

Honry (approved):
Thanks, LGTM, pls. also mention the needs to use 'expand' for `mask_shape_ones_shape_constant` in your commit message.

Honry:
+ @fdwr, please take another look, thanks!

fdwr (approved):
:shipit:

PR #26566, opened 2025-11-13
[WebNN] Update unit tests list

Honry:
Only list no fallback and pass tests for WebNN, as a pre-requisite for enabling WebNN CI tests in future.


Honry:

js/web/test/suite-test-list.jsonc
```diff
@@ -2365,14 +2372,15 @@
...
       // // "test_size",
-      "test_slice_default_axes",
-      "test_slice_default_steps",
-      "test_slice_end_out_of_bounds",
-      "test_slice_neg_steps",
-      "test_slice_neg",
-      "test_slice_negative_axes",
-      "test_slice_start_out_of_bounds",
-      "test_slice",
+      // tests "test_slice_*" on opset > 9 are excluded because starts, ends, axes and steps are non-constant inputs.
```
fdwr, line 2375: In practice they are constant inputs for most models. Even if the input tensor itself is technically *capable* of being dynamic, ORT almost always resolves them to constants via constant folding. So I wonder if the issue is that this test case binds the input dynamically. Anyway, keep as-is - just seeking understanding.

Honry: These unit tests make starts, ends, axes, etc. as non-constant inputs (model inputs), i.e. we can only get their values during inference.
But WebNN `slice` needs to know the starts and sizes for operand construction, that's why they'll fallback.

image

fdwr:
> [Windows CPU CI Pipeline / build_x86_release (pull_request)](https://github.com/microsoft/onnxruntime/actions/runs/19325004555/job/55365721524?pr=26566)Failing after 19m

> Failed to load the dll from [C:\Program Files\dotnet\host\fxr\10.0.0\hostfxr.dll], HRESULT: 0x800700C1
> The library hostfxr.dll was found, but loading it from C:\Program Files\dotnet\host\fxr\10.0.0\hostfxr.dll failed
>  - Installing .NET prerequisites might help resolve this problem.
>     https://go.microsoft.com/fwlink/?linkid=798306
> Failed to resolve hostfxr.dll [C:\Program Files\dotnet\host\fxr\10.0.0\hostfxr.dll]. Error code: 0x80008082

Err, retrying again in case, but this is unrelated to your change.

Honry:
> Err, retrying again in case, but this is unrelated to your change.

@fdwr, it passes now, can we land it?

PR #26572, opened 2025-11-14
webgpu: add MatMul and Gemm cases with large shapes

xhcao:


xhcao:
The PR add cases for https://github.com/microsoft/onnxruntime/pull/26433 and https://github.com/microsoft/onnxruntime/pull/26461/

@jchen10 @Jiawei-Shao PTAL

onnxruntime/test/providers/cpu/math/matmul_large_test.cc
```diff
@@ -0,0 +1,192 @@
...
+  RunTestTyped<float, 13>({511, 1024}, {1024, 1023});
+  RunTestTyped<float, 13>({512, 1024}, {1024, 1024});
+  RunTestTyped<float, 13>({1, 512, 1024}, {1024, 1024});
+  RunTestTyped<float, 13>({2, 512, 1024}, {1024, 1024});
+  RunTestTyped<float, 13>({2, 512, 1024}, {2, 1024, 1024});
+  RunTestTyped<float, 13>({2, 2, 512, 1024}, {2, 1024, 1024});
+}
+
+TEST(MatMul_Large, Float16_Subgroup) {
+  RunTestTyped<float, 13>({512, 1024}, {1024, 1024});
```
jchen10: `MLFloat16`

onnxruntime/test/providers/cpu/math/gemm_large_test.cc
```diff
@@ -0,0 +1,239 @@
...
+  RunTestTyped<MLFloat16, 13>({16, 1024}, 0, {1024, 192}, 0, {1, 192}, 1.3f, 1.5f);
+  RunTestTyped<MLFloat16, 13>({16, 1024}, 0, {1024, 192}, 0, {192}, 1.3f, 1.5f);
+  RunTestTyped<MLFloat16, 13>({16, 1024}, 0, {1024, 192}, 0, {16, 192}, 1.3f, 1.5f);
+  RunTestTyped<MLFloat16, 13>({1024, 16}, 1, {1024, 192}, 0, {192});
+  RunTestTyped<MLFloat16, 13>({16, 1024}, 0, {192, 1024}, 1, {192});
+  RunTestTyped<MLFloat16, 13>({1024, 16}, 1, {192, 1024}, 1, {192});
+}
+
+}  // namespace test
+}  // namespace onnxruntime
```
jchen10: Too many duplicated code lines between the Float32 and Float16 cases. You can have one case covering both Float32 and Float16.
Split_Dim_Inner and Subgroup are not good names for the cases. They don't help me understand the rationale behind the tested various shapes.

xhcao:
@fs-eire All test cases nearly cost ~120s, including CPU EP. Should I let CPU EP not execute these cases , make these case only for WebGPU EP, and in a WebGPU EP special directory,  not test on the other EPs?
Is there a way that not need to provide expected results, comparing CPU EP and WebGPU EP results directly, and let CPU EP result as a reference result?

xiaofeihan1:
I added GenerateGemmParams in gemm_test.cc to cover many shapes and different biasType/transpose/type combinations. Would that be useful for you?

xhcao:
> I added GenerateGemmParams in gemm_test.cc to cover many shapes and different biasType/transpose/type combinations. Would that be useful for you?

Yes, thanks

xhcao:
There are some bots failed for precision issues. Special machine issue? Our code issue? Or set the tolerance to small?

fs-eire:
Sorry for the late reply.

I am thinking about whether enabling a ~120sec test case by default is a good idea. Is it OK that the test case disabled by default? (can use commandline to run explicitly)

xhcao:
> Sorry for the late reply.
> 
> I am thinking about whether enabling a ~120sec test case by default is a good idea. Is it OK that the test case disabled by default? (can use commandline to run explicitly)

Hi, @fs-eire,  no problem. I will follow your comments and try to modify the code.

fs-eire:
One way is to use `DISABLED_` as prefix to the test case name so that gtest won't run it by default. Can manually run by `onnxruntime_provider_test --gtest_filter=<FULLNAME>`.

This may not be a very good option to be honest, because it's very easy to forget to run the test for a long time. But it's at least than not having the tests at all. I am open to other options.

qjia7:
> One way is to use `DISABLED_` as prefix to the test case name so that gtest won't run it by default. Can manually run by `onnxruntime_provider_test --gtest_filter=<FULLNAME>`.
> 
> This may not be a very good option to be honest, because it's very easy to forget to run the test for a long time. But it's at least than not having the tests at all. I am open to other options.

@fs-eire
Can we add some webgpu ep options specific for tests so that some specific matmul can be tested directly no matter it satisfies the shapes requirement? In this way, we can use very small tests to cover different matmul algorithms if the purpose of this PR is to test different matmul path.

fs-eire:
> > One way is to use `DISABLED_` as prefix to the test case name so that gtest won't run it by default. Can manually run by `onnxruntime_provider_test --gtest_filter=<FULLNAME>`.
> > This may not be a very good option to be honest, because it's very easy to forget to run the test for a long time. But it's at least than not having the tests at all. I am open to other options.
> 
> @fs-eire Can we add some webgpu ep options specific for tests so that some specific matmul can be tested directly no matter it satisfies the shapes requirement? In this way, we can use very small tests to cover different matmul algorithms if the purpose of this PR is to test different matmul path.

this is a good idea. However, we need to be very careful designing the flags. Otherwise it can turn out to be too many or inconsistent to make it difficult to maintain.

PR #26573, opened 2025-11-14
webgpu: split transpose perm{2310} in two steps

xhcao:
In order to use transpose-shared instead transpose-naive, we could split transpose perm{2310} in two steps, which benifits Conv operator.


xhcao:
The PR gets performance on `sdunet-v1.5-demo-layernorm` model,  all Conv|Transpose time is from 224ms to 135ms

xhcao:
@jchen10 @daijh PTAL

jchen10:
Looks great. As we discussed in https://github.com/microsoft/onnxruntime/pull/26554#issuecomment-3530476273, we are going to cache the transposed kernel. This PR could be less beneficial for Conv|Transpose. Maybe we could find other place to apply this optimization later.

xhcao:
After the feature that prepacks conv weights on `session` initializing stage, there is no poor transpose performance issue, so I close  the PR.

PR #26575, opened 2025-11-14
KFI-203 Improve thread safety of packing in convolve_kleidiai.cpp

Colm-in-Arm:
### Description
Making cache objects of packed data thread_local rather than static.

### Motivation and Context
Both LHS and RHS packing utilize a cache mechanism based on a static unordered map. There's the potential for interference between parallel inference sessions. Made both structures thread_local.


hariharans29:
It will need this to fix the failing pipeline: https://github.com/microsoft/onnxruntime/pull/26559

edgchen1 (approved):

edgchen1 (approved):

Issue #24183, opened 2025-03-26
[Feature Request] How should I use symmetric quantization to quantify weight and obtain the correct quantization model?


cosonjia:
### Describe the feature request

Describe the bug:

I am able to do quantization with:
```python
quantized_model = quantize_dynamic(
        model_input=model_input,
        model_output=model_output,
        # optimize_model=False,
        weight_type=QuantType.QInt8, 
        # activation_type=QuantType.QInt8,
        # op_types_to_quantize=["Conv", "MatMul", "Add","Relu"], 
        extra_options={"UseSymmetric": True, "ActivationSymmetric": True, 
                       "EnableSubgraph": True, "ForceQuantizeNoInputCheck": True}  
    )
```
but  I got the following error while doing the inference:

File "/home/pc10/anaconda3/envs/evo-venv/lib/python3.11/site-packages/onnxruntime/capi/onnxruntime_inference_collection.py", line 465, in __init__
    self._create_inference_session(providers, provider_options, disabled_optimizers)
  File "/home/pc10/anaconda3/envs/evo-venv/lib/python3.11/site-packages/onnxruntime/capi/onnxruntime_inference_collection.py", line 537, in _create_inference_session
    sess.initialize_session(providers, provider_options, disabled_optimizers)
onnxruntime.capi.onnxruntime_pybind11_state.NotImplemented: [ONNXRuntimeError] : 9 : NOT_IMPLEMENTED : Could not find an implementation for ConvInteger(10) node with name 'Conv__116_quant'

System information

OS Platform and Distribution: Ubuntu Ubuntu 20.04.5 LTS(5.15.0-48-generic)
CPU	 12th Gen Intel(R) Core(TM) i9-12900K
ONNX Runtime installed by pip:
ONNX Runtime version: 1.20.1
Python version: 3.11.11

How should I use symmetric quantization to quantify weight and obtain the correct quantization model?

### Describe scenario use case

```python
quantized_model = quantize_dynamic(
        model_input=model_input,
        model_output=model_output,
        # optimize_model=False,
        weight_type=QuantType.QInt8, 
        # activation_type=QuantType.QInt8,
        # op_types_to_quantize=["Conv", "MatMul", "Add","Relu"], 
        extra_options={"UseSymmetric": True, "ActivationSymmetric": True, 
                       "EnableSubgraph": True, "ForceQuantizeNoInputCheck": True}  
    )
```

Issue #15888, opened 2023-05-10
NOT_IMPLEMENTED : Could not find an implementation for ConvInteger(10) node with name 'Conv_0_quant'


chinmayjog13:
### Describe the issue

Following the documentation, I dynamically quantiized a resnet based model. The model is quantized and saved without error. However, when I try to create an inference session using the quantized model, the code crashes with the following error.

```
>>> ort_session = ort.InferenceSession(int8_path, providers=['CPUExecutionProvider'])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/chinmay/anaconda3/envs/v_pytorch2/lib/python3.8/site-packages/onnxruntime/capi/onnxruntime_inference_collection.py", line 360, in __init__
    self._create_inference_session(providers, provider_options, disabled_optimizers)
  File "/home/chinmay/anaconda3/envs/v_pytorch2/lib/python3.8/site-packages/onnxruntime/capi/onnxruntime_inference_collection.py", line 408, in _create_inference_session
    sess.initialize_session(providers, provider_options, disabled_optimizers)
onnxruntime.capi.onnxruntime_pybind11_state.NotImplemented: [ONNXRuntimeError] : 9 : NOT_IMPLEMENTED : Could not find an implementation for ConvInteger(10) node with name 'Conv_0_quant'
```

This is a duplicate of #12558 , which was closed a few months ago so I guess the support should have been added in onnxruntime, but I am still getting the same error.

### To reproduce

```
import onnx
from onnxruntime.quantization import quantize_dynamic, QuantType

model_fp32 = 'path/to/the/model.onnx'
model_quant = 'path/to/the/model.quant.onnx'
quantized_model = quantize_dynamic(model_fp32, model_quant)

import onnxruntime as ort
ort_session = ort.InferenceSession(model_quant, providers=['CPUExecutionProvider'])
```

[fp32_model](https://drive.google.com/file/d/1AfrxsOZ_zaslAFnkjuvl_gYAxb-WtSRr/view?usp=share_link)
[quantized_model](https://drive.google.com/file/d/1tb4xt93I-kjdYmHYJ50BbDV2hRFrg57l/view?usp=share_link)

### Urgency

Not very urgent, but not too low priority also.

### Platform

Linux

### OS Version

Ubuntu 20.04

### ONNX Runtime Installation

Released Package

### ONNX Runtime Version or Commit ID

1.14.1

### ONNX Runtime API

Python

### Architecture

X64

[...]

Issue #12558, opened 2022-08-11
Could not find an implementation for ConvInteger(10) node with name 'Conv_59_quant'


talhaanwarch:
I am trying to quantize slow_r50 model, but when i load it, I got this error
```
[ONNXRuntimeError] : 9 : NOT_IMPLEMENTED : Could not find an implementation for ConvInteger(10) node with name 'Conv_59_quant'
```
Code to reproduce
```
class OurModel(LightningModule):
    def __init__(self):
        super(OurModel,self).__init__()
        self.model = torch.hub.load('facebookresearch/pytorchvideo', 'slow_r50', pretrained=False)
    def forward(self,x):
        return self.model(x)

model=OurModel()


input_sample = torch.randn((1,3,30, 256,256))
model.to_onnx('onnx.onnx', input_sample,opset_version=11, export_params=True)

from onnxruntime.quantization import quantize_dynamic, QuantType
model_fp32 = 'onnx/{}.onnx'.format(vital)
model_quant = 'onnx/{}-quantized.onnx'.format(vital)
quantized_model = quantize_dynamic(model_fp32, model_quant)
```
Full Error
```
---------------------------------------------------------------------------
NotImplemented                            Traceback (most recent call last)
Input In [88], in <cell line: 1>()
----> 1 onnxruntime.InferenceSession('onnx/{}-quantized.onnx'.format(vital))

File ~/venv/lib/python3.8/site-packages/onnxruntime/capi/onnxruntime_inference_collection.py:347, in InferenceSession.__init__(self, path_or_bytes, sess_options, providers, provider_options, **kwargs)
    344 disabled_optimizers = kwargs["disabled_optimizers"] if "disabled_optimizers" in kwargs else None
    346 try:
--> 347     self._create_inference_session(providers, provider_options, disabled_optimizers)
    348 except ValueError:
    349     if self._enable_fallback:

File ~/venv/lib/python3.8/site-packages/onnxruntime/capi/onnxruntime_inference_collection.py:395, in InferenceSession._create_inference_session(self, providers, provider_options, disabled_optimizers)
    392     disabled_optimizers = set(disabled_optimizers)
    394 # initialize the C++ InferenceSession
--> 395 sess.initialize_session(providers, provider_options, disabled_optimizers)
    397 self._sess = sess
    398 self._sess_options = self._sess.session_options

[...]

Issue #12362, opened 2022-07-28
Integer quantization fails on Transformer-based vision model


sayakpaul:
Here's the Colab to reproduce the issue: https://colab.research.google.com/gist/sayakpaul/7009b366f4e77df684dba3070067664c/tfsegformer_onnx.ipynb.

Issue:

```
NotImplemented                            Traceback (most recent call last)
<ipython-input-13-3e9283520bd1> in <module>()
----> 1 sess = ort.InferenceSession(quant_model_path)
      2 quant_ort_outputs = sess.run(None, {"pixel_values": dummy_inputs_numpy})

1 frames
/usr/local/lib/python3.7/dist-packages/onnxruntime/capi/onnxruntime_inference_collection.py in __init__(self, path_or_bytes, sess_options, providers, provider_options, **kwargs)
    345 
    346         try:
--> 347             self._create_inference_session(providers, provider_options, disabled_optimizers)
    348         except ValueError:
    349             if self._enable_fallback:

/usr/local/lib/python3.7/dist-packages/onnxruntime/capi/onnxruntime_inference_collection.py in _create_inference_session(self, providers, provider_options, disabled_optimizers)
    393 
    394         # initialize the C++ InferenceSession
--> 395         sess.initialize_session(providers, provider_options, disabled_optimizers)
    396 
    397         self._sess = sess

NotImplemented: [ONNXRuntimeError] : 9 : NOT_IMPLEMENTED : Could not find an implementation for ConvInteger(10) node with name 'Conv__9597_quant'
```

@fatcat-z Cc

PR #26585, opened 2025-11-16
Add int8 support to ConvInteger

rivkastroh:
### Description

This change extends the `ConvInteger` implementation to match the [ONNX operator spec](https://onnx.ai/onnx/operators/onnx__ConvInteger.html), which allows both `int8` and `uint8` for the input tensors:

- The ONNX `ConvInteger` schema defines:
  - `T1`: `tensor(int8)` or `tensor(uint8)`
  - `T2`: `tensor(int8)` or `tensor(uint8)`
  - `T3`: `tensor(int32)`
- Previously, only the `uint8` × `uint8` combination was supported.
- This PR adds support for all 8-bit combinations:
  - `uint8` × `uint8` (existing behavior)
  - `uint8` × `int8`
  - `int8` × `uint8`
  - `int8` × `int8`

### Motivation and Context

Fixes https://github.com/microsoft/onnxruntime/issues/24183
Fixes https://github.com/microsoft/onnxruntime/issues/15888
Fixes https://github.com/microsoft/onnxruntime/issues/12558
Fixes https://github.com/microsoft/onnxruntime/issues/3130
Fixes https://github.com/microsoft/onnxruntime/issues/12362

The ONNX ConvInteger operator schema allows both int8 and uint8 element types for its inputs, but the current implementation only supports uint8 × uint8. This leads to a gap where valid ONNX models using ConvInteger with int8 tensors cannot be executed.
This PR closes that gap by:
Aligning the implementation with the official ConvInteger type constraints.
Enabling models that use int8 (or mixed int8/uint8) for X and W to run without needing operator rewrites or additional custom kernels.
Keeping existing uint8 behavior unchanged, so the change is backwards compatible for current users.

### Implementation details

1. Templated core implementation (ComputeInner)
The core logic of ConvInteger::Compute is moved into a templated helper:
```text
class ConvInteger : public OpKernel {
 public:
       ...
 private:
  template <typename XT, typename WT>
  Status ComputeInner(OpKernelContext* context) const
};
```
XT is the element type of X (uint8_t or int8_t).
WT is the element type of W (uint8_t or int8_t).

2. Zero-point handling
Zero points are still treated as per-tensor scalar values, with the same validation,
The values are read via `DataRaw()` and stored as 8-bit scalars, preserving the previous behavior.
Interpretation of these raw bytes as signed or unsigned is delegated to the GEMM implementation via explicit signedness flags (see below).

3. Im2col templated on XT
The Im2col call now uses the runtime input type XT.

4. Quantized GEMM with signedness flags:
```text
gemm_shape.AIsSigned = W->IsDataType<int8_t>();
gemm_shape.BIsSigned = X->IsDataType<int8_t>();
```
AIsSigned and BIsSigned are derived from the runtime types of W and X.
Data for A and B is passed as raw bytes, the GEMM implementation uses the signedness flags to interpret them correctly (In a manner similar to the implementation in `MatMulInteger`).

5. Runtime dispatch in Compute()
The public Compute method becomes a thin dispatcher that selects the appropriate ComputeInner<XT, WT> instantiation based on the actual input types.

In addition, a small set of unit tests is added on top of the existing ConvInteger tests to cover the new type combinations, including cases where the first input tensor contains negative values (for the int8 × int8 path).

### DIFF
--- a/docs/OperatorKernels.md
+++ b/docs/OperatorKernels.md
@@ -88,7 +88,7 @@ Do not modify directly.*
 |Conv|*in* X:**T**<br> *in* W:**T**<br> *in* B:**T**<br> *out* Y:**T**|22+|**T** = tensor(float)|
 |||[11, 21]|**T** = tensor(float)|
 |||[1, 10]|**T** = tensor(float)|
-|ConvInteger|*in* x:**T1**<br> *in* w:**T2**<br> *in* x_zero_point:**T1**<br> *in* w_zero_point:**T2**<br> *out* y:**T3**|10+|**T1** = tensor(uint8)<br/> **T2** = tensor(uint8)<br/> **T3** = tensor(int32)|
+|ConvInteger|*in* x:**T1**<br> *in* w:**T2**<br> *in* x_zero_point:**T1**<br> *in* w_zero_point:**T2**<br> *out* y:**T3**|10+|**T1** = tensor(int8), tensor(uint8)<br/> **T2** = tensor(int8), tensor(uint8)<br/> **T3** = tensor(int32)|
 |ConvTranspose|*in* X:**T**<br> *in* W:**T**<br> *in* B:**T**<br> *out* Y:**T**|22+|**T** = tensor(float)|
 |||[11, 21]|**T** = tensor(float)|
 |||[1, 10]|**T** = tensor(float)|


rivkastroh:
@skottmckay, @yuslepukhin: could you please review the PR? Thanks.

onnxruntime/test/providers/cpu/nn/conv_integer_test.cc
```diff
@@ -254,6 +254,425 @@ TEST(ConvIntegerTest, WithStride3_2D_u8u8) {
...
+
+  std::vector<int64_t> y_dims{1, 1, 3, 3};
+  test.AddOutput<int32_t>("y", y_dims,
+                          {0, 8, 20,
+                           80, 330, 375,
+                           200, 780, 825});
+  // Exercise the (stride_w > 2) path inside Math::Im2col.
+
+  test.Run();
+}
```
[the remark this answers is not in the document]
rivkastroh: I added tests

onnxruntime/core/providers/cpu/quantization/conv_integer.cc
```diff
@@ -20,161 +20,183 @@ class ConvInteger : public OpKernel {
...
+    ConvInteger,
+    kOnnxDomain,
+    10,
+    kCpuExecutionProvider,
+    KernelDefBuilder()
+        .TypeConstraint("T1", {DataTypeImpl::GetTensorType<uint8_t>(),
+                               DataTypeImpl::GetTensorType<int8_t>()})
+        .TypeConstraint("T2", {DataTypeImpl::GetTensorType<uint8_t>(),
+                               DataTypeImpl::GetTensorType<int8_t>()})
+        .TypeConstraint("T3", DataTypeImpl::GetTensorType<int32_t>()),
```
yuslepukhin: Please, move it up top to simplify the review and according to the convention #Resolved

rivkastroh: I couldn’t move the `ONNX_OPERATOR_KERNEL_EX` macro all the way to the top of the file as in some other kernels because `ConvInteger` doesn’t have a separate header file – the class is only declared in this `.cc` file. The macro needs the full `ConvInteger` type, so it has to appear after the class definition.

Instead, I placed the templated `ComputeInner` implementation below the macro. That way, the registration stays as high as possible in the file.

If you prefer, I can also split `ConvInteger` into a header (`conv_integer.h`) and `.cc` file. Please let me know if that structure would be preferable.

onnxruntime/core/providers/cpu/quantization/conv_integer.cc
```diff
@@ -28,12 +32,15 @@ ONNX_OPERATOR_KERNEL_EX(
...
-        .TypeConstraint("T2", DataTypeImpl::GetTensorType<uint8_t>())
+        .TypeConstraint("T1", {DataTypeImpl::GetTensorType<uint8_t>(),
+                               DataTypeImpl::GetTensorType<int8_t>()})
+        .TypeConstraint("T2", {DataTypeImpl::GetTensorType<uint8_t>(),
+                               DataTypeImpl::GetTensorType<int8_t>()})
         .TypeConstraint("T3", DataTypeImpl::GetTensorType<int32_t>()),
     ConvInteger);
 
-Status ConvInteger::Compute(OpKernelContext* context) const {
+template <typename XT, typename WT>
```
yuslepukhin: >emplate <typename XT, typename WT> [](http://example.com/codeflow?start=1&length=34)

This location is fine. No need in header.
I noticed that a large portion of this function is actually type independent.
Would it be possible to split this function in two parts so no code is duplicated unless necessary?
We are very sensitive to binary size increase due to some constrained environments.

rivkastroh: You’re right.

I considered two options to address this. The first was to remove the templated function entirely and instead add a small number of `if` branches in the existing code. The second was to extract the `math::Im2col` call into a separate templated helper, but that would require passing around about 9 parameters to that helper.

I chose the first option, since we would still need `if` branches at the call site of the templated helper.

If you think the second option would be preferable, please let me know.

( I also noticed now that there was another place where I previously used `XT` just to advance the input pointer. I removed that, because this is only a pointer offset and does not depend on whether the element type is `uint8` or `int8`.)

yuslepukhin: Looks good to me.
One of our pipelines needs a fix, but is made optional in the interim. You many need to rebase from main to see it optional.

yuslepukhin (dismissed):
:shipit:

yuslepukhin:
The documentation pipeline failed. Here is the output:
> 2025-11-19 20:41:16,441 build [WARNING] - The updated document D:\a\_work\1\s\docs\OperatorKernels.md is different from the checked in version. Please regenerate the file with CPU, CUDA and DML execution providers enabled, or copy the updated version from the CI build's published artifacts if applicable.
2025-11-19 20:41:16,441 build [DEBUG] - diff:
diff --git a/docs/OperatorKernels.md b/docs/OperatorKernels.md
index 92093ec546..d7be243323 100644
--- a/docs/OperatorKernels.md
+++ b/docs/OperatorKernels.md
@@ -88,7 +88,7 @@ Do not modify directly.*
 |Conv|*in* X:**T**<br> *in* W:**T**<br> *in* B:**T**<br> *out* Y:**T**|22+|**T** = tensor(float)|
 |||[11, 21]|**T** = tensor(float)|
 |||[1, 10]|**T** = tensor(float)|
-|ConvInteger|*in* x:**T1**<br> *in* w:**T2**<br> *in* x_zero_point:**T1**<br> *in* w_zero_point:**T2**<br> *out* y:**T3**|10+|**T1** = tensor(uint8)<br/> **T2** = tensor(uint8)<br/> **T3** = tensor(int32)|
+|ConvInteger|*in* x:**T1**<br> *in* w:**T2**<br> *in* x_zero_point:**T1**<br> *in* w_zero_point:**T2**<br> *out* y:**T3**|10+|**T1** = tensor(int8), tensor(uint8)<br/> **T2** = tensor(int8), tensor(uint8)<br/> **T3** = tensor(int32)|
 |ConvTranspose|*in* X:**T**<br> *in* W:**T**<br> *in* B:**T**<br> *out* Y:**T**|22+|**T** = tensor(float)|
 |||[11, 21]|**T** = tensor(float)|
 |||[1, 10]|**T** = tensor(float)|

rivkastroh:
The failure in both tests is: "Non-zero status code returned while running Conv node. Name:'node1' Status Message: CUDA error cudaErrorNotSupported: operation not supported" — i.e., a CUDA runtime error (cudaErrorNotSupported) during Conv kernel execution.
The latest commit only changes `OperatorKernels.md` (documentation) and doesn’t modify any CUDA/kernel code. The same tests passed on the previous commit, so the `cudaErrorNotSupported` failures are likely related to the CI environment rather than this PR – could you please re-run CI?

Thanks!

rivkastroh:
@yuslepukhin Could you please re-approve the PR? Thank you for the review!

yuslepukhin (approved):
:shipit:

vasanthzoh:
Since this PR is merged, does that mean dynamic quantized Int8 Conv and ConvTranspose layers are now supported in the master branch? @yuslepukhin @rivkastroh

rivkastroh:
Yes – this PR adds dynamic quantized Int8 Conv and ConvTranspose support to the main branch.

Issue #26568, opened 2025-11-13
Cannot install nightly onnxruntime-gpu for CUDA 13.0: "User for aiinfra.pkgs.visualstudio.com"


NullDev:
I tried to follow the nightly installation guide for CUDA 13.0 here:
https://onnxruntime.ai/docs/install/#nightly-for-cuda-13x

```sh
$ pip install flatbuffers numpy packaging protobuf sympy
```
> All good

---

But then:

```sh
$ pip install --pre --index-url https://aiinfra.pkgs.visualstudio.com/PublicPackages/_packaging/ORT-Nightly/pypi/simple/ onnxruntime-gpu
```

```
...
Requirement already satisfied: mpmath<1.4,>=1.1.0 in c:\users\shad0w\anaconda3\envs\kr3\lib\site-packages (from sympy->onnxruntime-gpu) (1.3.0)
...
User for aiinfra.pkgs.visualstudio.com: 
WARNING: 401 Error, Credentials not correct for https://aiinfra.pkgs.visualstudio.com/PublicPackages/_packaging/ort-cuda-13-nightly/pypi/simple/pip/
```

Why is it asking for a username?
After just pressing enter it aborted with `401 Error, Credentials not correct`

On: https://aiinfra.pkgs.visualstudio.com/PublicPackages/_packaging/ort-cuda-13-nightly/pypi/simple/pip/

<img width="362" height="324" alt="Image" src="https://github.com/user-attachments/assets/2a05c5e5-3b9b-426c-be80-54f723757580" />

---

Using a conda venv with Python 3.13.5, in case that's relevant:

```
$ python --version
Python 3.13.5

$ pip --version
pip 25.2 from C:\Users\shad0w\anaconda3\envs\kr3\Lib\site-packages\pip (python 3.13)

$ nvcc --version
nvcc: NVIDIA (R) Cuda compiler driver
Copyright (c) 2005-2025 NVIDIA Corporation
Built on Wed_Aug_20_13:58:20_Pacific_Daylight_Time_2025
Cuda compilation tools, release 13.0, V13.0.88
Build cuda_13.0.r13.0/compiler.36424714_0
```

---

Related Issue: https://github.com/microsoft/onnxruntime/issues/26547 ([comment](https://github.com/microsoft/onnxruntime/issues/26547#issuecomment-3528456615))

tianleiwu:
Please try again.

The URL is not for browser. You can view it in browser with the following link:
[...]

Issue #26547, opened 2025-11-11
Nightly packages for Python 3.14 or CUDA 13.x available for testing


tianleiwu:
We added a new nightly python package feed for CUDA 13.x. See Installation section below.

## Python 3.14

Nightly packages now support python 3.14:

Device | OS | cp314 | cp314t 
-- | -- | -- | --
CUDA 13.x | Linux x64 | Available | Available 
CUDA 13.x | Windows | Available | None
CUDA 12.x | Linux x64 | Available | Available 
CUDA 12.x | Windows | Available  | None
CPU | Linux x64 | Available | Available
CPU | Linux ARM64 | Available| Available
CPU | Windows | Available | None
CPU | MacOS| Available | None

Tag	| CPython Version & Build	| Key Feature
-- | -- | --
cp314 | CPython 3.14 (Standard)	| Includes the Global Interpreter Lock (GIL).
cp314t | CPython 3.14 (Free-Threaded) | Built without the GIL (or with a free-threading alternative).

## Installation

First, you might need uninstall existed packages like
```
pip uninstall onnxruntime onnxruntime-gpu -y
```

Then choose only one of  the following based on cuda version (or CPU only) to install:

### CUDA 13.x
```
pip install flatbuffers numpy packaging protobuf sympy coloredlogs
pip install --pre --index-url https://aiinfra.pkgs.visualstudio.com/PublicPackages/_packaging/ort-cuda-13-nightly/pypi/simple/ onnxruntime-gpu --no-deps
```

### CUDA 12.x
```
pip install flatbuffers numpy packaging protobuf sympy coloredlogs
pip install --pre --index-url https://aiinfra.pkgs.visualstudio.com/PublicPackages/_packaging/ORT-Nightly/pypi/simple/ onnxruntime-gpu --no-deps
```

### CPU Only
```
pip install flatbuffers numpy packaging protobuf sympy coloredlogs
pip install --pre --index-url https://aiinfra.pkgs.visualstudio.com/PublicPackages/_packaging/ORT-Nightly/pypi/simple/ onnxruntime --no-deps
```

If you encountered issues, please reply to this issue, or create a new issue and reference this for tracking.

---
Related issues: 
https://github.com/microsoft/onnxruntime/issues/26388
https://github.com/microsoft/onnxruntime/issues/26473
https://github.com/microsoft/onnxruntime/issues/26238

NullDev:
Thank you! It's asking for a username/password on my end though:
[...]

PR #26588, opened 2025-11-17
[Doc] Fix nightly installation: add coloredlogs

tianleiwu:
coloredlogs was added to requirements.txt.in (later renamed to requirements.txt) since 1.12.0 by https://github.com/microsoft/onnxruntime/pull/11522.

This updates the installation command lines for nightly packages to avoid installation failure like https://github.com/microsoft/onnxruntime/issues/26568. 

The dependency on coloredlogs was also reported by user: https://github.com/microsoft/onnxruntime/issues/26547#issuecomment-3533435073.

### DIFF
--- a/docs/install/index.md
+++ b/docs/install/index.md
@@ -51,7 +51,7 @@ pip install onnxruntime
 #### Install nightly
 
 ```bash
-pip install flatbuffers numpy packaging protobuf sympy
+pip install coloredlogs flatbuffers numpy packaging protobuf sympy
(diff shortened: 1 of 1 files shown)


nenad1002 (approved):

stonebig:
coloredlogs is a problem: 
- it makes a crash on python console, because it embarks pyreadline3
- pyreadline3 is no more friendly with python-3.13+ 

Can it be removed/replaced ?


image

PR #26595, opened 2025-11-18
Add weight layout transformation cache for Conv operator

jchen10:
Implement lazy weight layout transformation for WebGPU Conv kernel to avoid redundant GPU transposes on every inference.

Key changes:
- Add WeightLayoutTransformCache to cache transformed weights by name and format
- Implement TransformWeightLayout() helper using existing TransposeKernel for OIHW->HWIO transformation
- Cache stored in WebGpuExecutionProvider, shared across all kernels

### DIFF
--- a/onnxruntime/core/providers/webgpu/compute_context.h
+++ b/onnxruntime/core/providers/webgpu/compute_context.h
@@ -139,6 +139,13 @@ class ComputeContext final {
     return webgpu_context_.Run(*this, program);
   }
 
+  //
+  // Get the execution provider.
+  //
+  inline const WebGpuExecutionProvider& GetExecutionProvider() const {
+    return ep_;
+  }
+
  private:
   WebGpuContext& webgpu_context_;
   OpKernelContext& kernel_context_;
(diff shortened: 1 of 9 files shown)


jchen10:
Follow-up for https://github.com/microsoft/onnxruntime/pull/26554

jchen10:
@fs-eire PTAL

jchen10:
I am still looking into the PrePack approach, which seems more appealing as it does release the original tensors.

fs-eire:
> I am still looking into the PrePack approach, which seems more appealing as it does release the original tensors.

Please take a look at https://github.com/microsoft/onnxruntime/pull/26602. However I didn't finish all validation yet.

jchen10:
> > I am still looking into the PrePack approach, which seems more appealing as it does release the original tensors.
> 
> Please take a look at #26602. However I didn't finish all validation yet.

Great. That's exactly what I wanted. One pity of PrePack is that we couldn't know the runtime input/output shapes which may impact how we choose the optimal blocked format for weight. Let's see if this issue will come up in the future. So far so good.

PR #26598, opened 2025-11-18
Upgrade cpuinfo version and update some CPU feature detection code

edgchen1:
### Description

Upgrade cpuinfo version.
Update ARM64 Windows feature detection to use cpuinfo functions which are now implemented.

### Motivation and Context

Additional ARM64 Windows feature detection was implemented in cpuinfo in [this commit](https://github.com/pytorch/cpuinfo/commit/f7b233b2755235c97ef1c2e650820ece1b2869a4). Update ORT to use the newly implemented feature detection.


onnxruntime/core/common/cpuid_info.cc
```diff
@@ -308,22 +298,15 @@ void CPUIDInfo::ArmWindowsInit() {
...
   if (pytorch_cpuinfo_init_) {
+    has_arm_neon_dot_ = cpuinfo_has_arm_neon_dot();
     has_fp16_ = cpuinfo_has_arm_neon_fp16_arith();
-    // cpuinfo_has_arm_i8mm() doesn't work on Windows yet. See https://github.com/pytorch/cpuinfo/issues/279.
-    // has_arm_neon_i8mm_ = cpuinfo_has_arm_i8mm();
-    has_arm_sve_i8mm_ = cpuinfo_has_arm_sve() && has_arm_neon_i8mm_;
+    has_arm_neon_i8mm_ = cpuinfo_has_arm_i8mm();
+    has_arm_sve_i8mm_ = cpuinfo_has_arm_sve() && cpuinfo_has_arm_i8mm();
     has_arm_neon_bf16_ = cpuinfo_has_arm_neon_bf16();
+    has_arm_sme_ = cpuinfo_has_arm_sme();
```
hariharans29, line 308: Should we leave these defaulted to false for now? Or are the KleidiAI kernels enabled only on non-Windows platforms (as opposed to being enabled in platform.cpp if SME/SME2 is available) ?

edgchen1: I think that it is fine to set them here so that the `CPUIDInfo` info is accurate. if KleidiAI SME/SME2 kernels aren't supported on Windows then we should refine the conditions that enable them.

onnxruntime/core/mlas/lib/qgemm.cpp
```diff
@@ -201,6 +201,17 @@ MlasGemmBatch(
...
 }
 
+bool
+MLASCALL
+MlasIsDynamicQGemmAvailable()
+{
+#if defined(USE_KLEIDIAI) && !defined(_MSC_VER)
+  return ArmKleidiAI::UseSME2;
+#else
+  return false;
```
hariharans29: Could we add a TODO here with relevant commentary and a reminder to adjust these checks once KAI support for Windows is added ?

onnxruntime/core/common/cpuid_info.cc
edgchen1: added a helper function `MlasIsDynamicQGemmAvailable()` to replace some of the existing checks of `CPUIDInfo::HasArm_SME2()`. I think the original checks were already in ifdefs to ensure they weren't on Windows, but the helper function should be easier to use.

onnxruntime/core/common/cpuid_info.cc
```diff
@@ -308,22 +298,31 @@ void CPUIDInfo::ArmWindowsInit() {
...
+    has_arm_neon_dot_ = cpuinfo_has_arm_neon_dot();
     has_fp16_ = cpuinfo_has_arm_neon_fp16_arith();
-    // cpuinfo_has_arm_i8mm() doesn't work on Windows yet. See https://github.com/pytorch/cpuinfo/issues/279.
-    // has_arm_neon_i8mm_ = cpuinfo_has_arm_i8mm();
-    has_arm_sve_i8mm_ = cpuinfo_has_arm_sve() && has_arm_neon_i8mm_;
+
+    // Note:
+    // cpuinfo is using IsProcessorFeaturePresent(PF_ARM_V82_FP16_INSTRUCTIONS_AVAILABLE):
+    // https://github.com/pytorch/cpuinfo/blob/403d652dca4c1046e8145950b1c0997a9f748b57/src/arm/windows/init.c#L224-L225
+    // However, on some systems (notably, a Windows ARM64 CI build agent), cpuinfo_has_arm_neon_fp16_arith() started to
```
hariharans29: Should cpuinfo be notified to treat this as a regression in detection of fp16 support in their newer implementation if their new way (seemingly) doesn't detect fp16 reliably on certain Windows versions ?

edgchen1: yeah, it would be good to fix it there

edgchen1: created a PR: https://github.com/pytorch/cpuinfo/pull/348

edgchen1: the cpuinfo PR isn't merged yet. added the corresponding patch for cpuinfo in this repo.

onnxruntime/core/mlas/lib/qgemm.cpp
edgchen1: we discussed it. it seemed inconsistent to add a new comment here given the other existing ifdefs with no mention about future Windows support. so I'll leave it as is for now.

also, this got moved to another [PR](https://github.com/microsoft/onnxruntime/pull/26668)

PR #26602, opened 2025-11-19
[webgpu] refactor a few "context" classes

fs-eire:
### Description

This PR refactors a few "context" classes to make it clearer and support new features.

PR #26603, opened 2025-11-19
[webgpu] Optimize Conv by im2col-matmul

daijh:
### Description
This PR optimizes the `Conv` operation by implementing two new compute shaders: `oihw_to_ohwi` and `im2col-matmul`.

`oihw_to_ohwi`: 
Improves performance over the default Transpose shader by utilizing workgroup memory to ensure continuous memory read/write patterns.

`im2col-matmul`:
- Employs a workgroup size of 64.
- Dynamically selects tile sizes (32x64 or 16x64) based on the source/weight shape.
- Each invocation handles a dedicated weight element.
- Uses subgroupShuffle to efficiently access the source tile, leveraging k_vec4 vectorization for better memory throughput.

Testing on Lunar Lake demonstrated **up to an 87%** performance improvement in Conv_2D operations.

### Motivation and Context
See above.

### DIFF
--- a/onnxruntime/core/providers/webgpu/nn/im2col_matmul.cc
+++ b/onnxruntime/core/providers/webgpu/nn/im2col_matmul.cc
@@ -47,7 +47,6 @@ std::pair<uint32_t, uint32_t> ChooseTileSize(uint32_t im2col_m, uint32_t im2col_
 
 // Add support for more devices.
 bool IsDeviceSupported(ComputeContext& context) {
-#if 0
   const wgpu::AdapterInfo& adapter_info = context.AdapterInfo();
 
   if (adapter_info.vendor == std::string_view("intel")) {
@@ -57,10 +56,6 @@ bool IsDeviceSupported(ComputeContext& context) {
   }
 
   return false;
-#else
-  (void)context;  // <-- This line suppresses the unused variable warning
-  return true;
-#endif
 }
 
 }  // namespace


daijh:
**Lunar Lake**
onnxruntime commit d55ade03897350f1f2b51b26ba298a789278dab6

## Operation

| Milliseconds | conv2d-mm | im2col-matmul |
| :--- | :---: | :---: |
| src: 1x128x512x512<br>weight: 128x128x3x3 | 56.071 | **42.824** |
| src: 1x2560x8x8<br>weight: 1280x2560x3x3 | 21.066 | **11.263** |
| src: 1x1280x8x8<br>weight: 1280x1280x3x3 | 10.384 | **6.357** |

## sd-turbo

| Milliseconds | conv2d-mm | im2col-matmul |
| :--- | :--- | :--- |
| sd-turbo-unet-fp16-demo.onnx | 1010.245 | **612.092** |
| sd-turbo-vae-decoder-fp16-demo.onnx | 2317.391 | **1848.545** |

daijh:
@guschmue @fs-eire @qjia7 PTAL.

onnxruntime/core/providers/webgpu/nn/im2col_matmul.cc
```diff
@@ -0,0 +1,223 @@
...
+  const auto* bias = has_bias ? context.Input<Tensor>(2) : nullptr;
+
+  // Transpose OIHW Weight to OHWI
+  TensorShape kernel_shape = kernel->Shape();
+  const uint32_t channel_output = onnxruntime::narrow<uint32_t>(kernel_shape[0]);
+  const uint32_t channel_input = onnxruntime::narrow<uint32_t>(kernel_shape[1]);
+  const uint32_t kernel_height = onnxruntime::narrow<uint32_t>(kernel_shape[2]);
+  const uint32_t kernel_width = onnxruntime::narrow<uint32_t>(kernel_shape[3]);
+
+  TensorShape nhwc_kernel_shape{channel_output, kernel_height, kernel_width, channel_input};
```
qjia7:
```suggestion
  TensorShape ohwi_kernel_shape{channel_output, kernel_height, kernel_width, channel_input};
```

onnxruntime/core/providers/webgpu/nn/im2col_matmul.cc
```diff
@@ -0,0 +1,223 @@
...
+  Tensor nhwc_kernel = context.CreateGPUTensor(kernel->DataType(), nhwc_kernel_shape);
```
qjia7:
```suggestion
  Tensor ohwi_kernel = context.CreateGPUTensor(kernel->DataType(), nhwc_kernel_shape);
```

onnxruntime/core/providers/webgpu/nn/im2col_matmul.cc
```diff
@@ -0,0 +1,223 @@
...
+
+  const auto [tile_m, tile_n] = ChooseTileSize(im2col_m, im2col_n);
+  const uint32_t workgroup_size = tile_n;
+  const bool use_subgroup = true;
+  Im2ColMatMulProgram im2col_mm_program{has_bias, tile_m, tile_n, use_subgroup};
+  im2col_mm_program.SetWorkgroupSize(workgroup_size);
+
+  const uint32_t M_tiles = ceil_div(im2col_m, tile_m);
+  const uint32_t N_tiles = ceil_div(im2col_n, tile_n);
+  im2col_mm_program.SetDispatchGroupSize(M_tiles, N_tiles, batch);
```
qjia7, line 161: How about enhancing the current TransposeProgram with shared path instead of adding a new one?
You are doing transpose from perm [0, 1, 2, 3] to perm [0, 2, 3, 1].  It equals that we are transposing from [o, i, hw] to [o, hw, i]. You can simply extend the DoTranspose with shared path to support any shape that only transpose the last two dimensions and keep the previous dimensions unchanged. Currently, the shared path only supports 2d transpose from new shape from perm [0, 1] to new shape with perm [1, 0]. We can extend it to transpose from [0, 1, 2] to [0, 2, 1] if the transpose meets the requirement that only transpose the last two dimensions by reshape it into 3d tensor [d0 * d1*...*dn-3, dn-2, dn-1]

onnxruntime/core/providers/webgpu/nn/im2col_matmul.wgsl.template
```diff
@@ -0,0 +1,139 @@
...
+
+      weight_tile[load_weight_k][load_weight_n] = load_weight(n_global_base + load_weight_n,
+                                                              k_idx * TILE_K_VEC_SIZE + load_weight_k);
+    }
+    workgroupBarrier();
+
+    for (var inner_k_idx = 0u; inner_k_idx < TILE_K_VEC_SIZE; inner_k_idx++) {
+      let weight_data = weight_tile[inner_k_idx][local_idx];
+#if use_subgroup
+      let src_data = src_tile[inner_k_idx][sg_id];
```
qjia7, line 118: What if the sg_size is larger than or less than TILE_M_SIZE?

daijh: Currently, Lunar Lake devices support a subgroup size of 32. 
We must carefully add support for devices with different subgroup sizes, passing this value as a parameter to the template.

daijh: BTW, using `subgroupShuffle` improves the performance 5%~10%, comparing no `subgroupShuffle`.

qjia7: If you only want to support sg_size = 32, I prefer you write it like blow. Otherwise, you can't make sure the sg_size must be 32 unless your check the supported subgroup range with subgroupMinSize = subgroupMaxSize = 32.
```
if (sg_size == 32) {
      let src_data = src_tile[inner_k_idx][sg_id];
      for (var m_idx = 0u; m_idx < TILE_M_SIZE; m_idx++) {
        results[m_idx] += output_element_t(dot(weight_data, subgroupShuffle(src_data, m_idx)));
      }
}
else {
      for (var m_idx = 0u; m_idx < TILE_M_SIZE; m_idx++) {
        results[m_idx] += output_element_t(dot(weight_data, src_tile[inner_k_idx][m_idx]));
      }
}

Comment (qjia7) [reply onnxruntime/core/providers/webgpu/nn/im2col_matmul.wgsl.template]: And the TILE_M_SIZE can 16 or 32 in your code. When TILE_M_SIZE = 16 and sg_size = 32, src_tile[inner_k_idx][sg_id] will be out of bound?

Comment (daijh) [reply onnxruntime/core/providers/webgpu/nn/im2col_matmul.wgsl.template:118]: My understanding is that most modern GPUs typically support a subgroup size of 32 or greater, which this shader is designed to support. 

The code's purpose is to mitigate potential performance penalties associated with the runtime conditional check (if (sg_size == 32)) within the shader. 

```
if (sg_size == 32) {
  // do somethings
}
```

Testing on Lunar Lake showed these penalties to be high.

Add comment in the code about it. Or I can simply make `use_subgroup=false` by default.

onnxruntime/core/providers/webgpu/nn/im2col_matmul.cc
daijh: Understood. 
I intend to improve the current Transpose path discussed in the previous PR #26501. 
Could I handle this as a separate task for a following PR?

onnxruntime/core/providers/webgpu/nn/im2col_matmul.cc
daijh: Done

onnxruntime/core/providers/webgpu/nn/im2col_matmul.wgsl.template
qjia7: My previous knowledge is that `if (sg_size == 32)` won't bring much perf penalty since sg_size is uniform. If it's the case in lunar lake, I prefer you disable it by default in this PR and re-enable it when you can ensure the correctness in any sg_size in your supported devices.

onnxruntime/core/providers/webgpu/nn/im2col_matmul.cc
qjia7: FYI, prepack is supported with this PR https://github.com/microsoft/onnxruntime/pull/26602. You can move the weights transpose to prepack to get better performance.

daijh:
Certainly. 
I should note, however, that performance may be negatively impacted on un-tested devices. 
Given that the CI process is focused on functional correctness, this impact may be acceptable.

daijh:
Fix `defined but not used [-Werror=unused-function]` for Release build

daijh:
Suppresses the unused variable warning, please help to re-trigger. 
Thanks.

qjia7 (dismissed):

daijh:
Thanks.
The WebGPU/Web CI completed successfully. The failure noted in the logs is not relevant to this PR's changes.

Revert previous change to prevent performance regression on untested devices.

daijh:
The failure in the CI is not relevant to this PR's changes.
Let me rebase the main if fixing the CUDA/TensorRT CI.

PR #23297, opened 2025-01-09
LayerNormalization broadcast (limited support for axis=2)

tianleiwu:
### Description

Spec of LayerNormalization supports broadcasting (tensors Scale and B should be unidirectional broadcastable to tensor X).
https://onnx.ai/onnx/operators/onnx__LayerNormalization.html
However, current implementation only allow scale and bias size to be X.shape()[axis:].

Example of input tensors that normalized with axis=2:

| X shape |  Scale shape | B shape | Before | After |
| - | - | - | - | - |
| (B, S, D) | (D) | (D) | Supported | Supported |
| (B, S, D) | (1, 1, D) | (1, 1, D) | Supported | Supported |
| (B, S, D) | (B, 1, D) | (B, 1, D) | Not Supported | Supported |
| (B, S, D) | (1, S, D) | (1, S, D) | Not Supported | Supported |
| (B, S, D) | (B, S, D) | (B, S, D) | Not Supported | Supported |

Here we add limited support: axis=2; scale/bias has same shape; scale/bias/X have same number of dimensions. It could support common use case in LLM and vision models.

### Motivation and Context

Support Stable Diffusion 3.x and Flux model.

PR #26613, opened 2025-11-19
Add full broadcasting support to LayerNormalization and RMSNormalization

### LINKED ISSUE #18184: [LayerNormalization Operator] Why LayerNormalization do not support weight and bias auto broadcast.
### Describe the issue

Input is [1, 16, 256, 256]
I want to do layernorm in axes 1 and the weight and bias shape is [1, 16, 1, 1].
It will broadcast to [1, 16, 256, 256].
But the error is:
``` text
onnxruntime.capi.onnxruntime_pybind11_state.InvalidArgument: [ONNXRuntimeError] : 2 : INVALID_ARGUMENT : Non-zero status code returned while running LayerNormalization node. Name:'/LayerNormalization' Status Message: Size of X.shape()[axis:] == 1048576. Size of scale and bias (if provided) must match this. Got scale size of 16 and bias size of 16
```
So what can i do?

### Urgency

_No response_

### Platform

Linux

### OS Version

ubuntu 20.04

### ONNX Runtime Installation

Released Package

1.16.0

### ONNX Runtime API

Python

### Architecture

X64

### Execution Provider

Default CPU

cpu

### Model File

_No response_

No

### LINKED ISSUE #26432: Issue with RMS Normalization when scale is a scalar
ONNX spec states "Scale tensor shape should be broadcastable to the normalized shape," which should mean a scalar weight should be acceptable as the scale input. ORT rejects this input with error

`onnxruntime.capi.onnxruntime_pybind11_state.InvalidArgument: [ONNXRuntimeError] : 2 : INVALID_ARGUMENT : Non-zero status code returned while running RMSNormalization node. Name:'node_rms_norm' Status Message: Size of scale and bias (if provided) must match X.shape[axis:], or scale and bias (with same shape) can be broadcasted to X when axis is 2. X.shape={2,5,3} scale.shape={} bias.shape={} and axis=2`

Ref: https://github.com/microsoft/onnxruntime/blob/0f6cffc140211dd5766aff1c052e8bc57e3f954e/onnxruntime/core/providers/cpu/nn/rms_norm.h#L13

Ref: https://github.com/microsoft/onnxruntime/blob/0f6cffc140211dd5766aff1c052e8bc57e3f954e/onnxruntime/core/providers/cpu/nn/layer_norm_helper.h#L62

cc @titaiwangms @gramalingam 

Also @gramalingam : would it be a good idea to make `scale` optional? We can do a batch update on the normalization ops to make scale and bias optional in opset 25: https://github.com/onnx/onnx/issues/6552

naomiOvad:
### Description

This PR adds full and spec-compliant broadcasting support to both LayerNormalization and RMSNormalization.

Previously, onnxruntime supported only a partial set of broadcasting cases (based on the logic introduced in this PR #23297 ).
That implementation handled several cases but did not cover all valid broadcasting scenarios.

This PR introduces a complete generic broadcasting path, following the [ONNX specification rules](https://github.com/onnx/onnx/blob/main/docs/Broadcasting.md).
The previous implementation is preserved as a fast-path and is still used whenever the Scale/Bias shapes match directly.

Main changes:

- Extended broadcasting logic in:
     layer_norm_helper.h
     layer_norm_impl.cc

- Added full support for all valid broadcasting configurations of Scale and Bias.

- Preserved previous partial logic as a fast-path for exact-match cases.

- Added comprehensive tests to:
      layer_norm_op_test.cc
      rms_norm_op_test.cc

### Motivation and Context

Before this fix, some valid ONNX broadcasting shapes were rejected  in LayerNormalization and RMSNormalization.
This PR brings the operators into full alignment with the ONNX specification and fixes models that previously failed due to incomplete broadcasting support.

Fixes #26432
Fixes #18184

### DIFF
--- a/onnxruntime/core/providers/cpu/nn/layer_norm_helper.h
+++ b/onnxruntime/core/providers/cpu/nn/layer_norm_helper.h
@@ -3,12 +3,12 @@
 
 #pragma once
 
+#include <algorithm>
 #include "core/framework/tensor_shape.h"
 #include "core/common/status.h"
 #include "core/common/narrow.h"
 #include "core/common/inlined_containers.h"
 #include "core/providers/cpu/nn/layer_norm_macro.h"
-#include <algorithm>
 
 namespace onnxruntime {
 
--- a/onnxruntime/core/providers/cpu/nn/layer_norm_macro.h
+++ b/onnxruntime/core/providers/cpu/nn/layer_norm_macro.h
@@ -8,4 +8,3 @@
   ((broadcast_param == 0) ? 0                                           \
                           : norm_size * (broadcast_param > 0 ? x_row / broadcast_param : x_row % (-broadcast_param)))
 #endif
-


naomiOvad:
cc @justinchuby, @amarin16, @tianleiwu Could you please review my PR?
Thanks!

tianleiwu:
Please fix build errors in CI pipelines.

Overall logic aligns with ONNX broadcasting rules. Preserves optimized "fast path" for standard shapes; no regression for existing models.

Below are AI analysis:
---------------------
Here are recommendations to improve the code quality, performance, and testing of your implementation.

### 1\. Performance: Optimize the Generic Inner Loop (Critical)

**The Issue:**
The current `ComputeJobGeneric` implementation updates a multi-dimensional index array (`idx`) and calculates offsets in the innermost loop for *every single element*.

```cpp
// Current Inner Loop
for (int64_t h = 0; h < norm_size; ++h) {
  // ... calculate offsets ...
  // ... math ...
  // ... heavy logic to update idx array ...
}
```

This logic prevents the compiler from auto-vectorizing the normalization math (SIMD), which is crucial for performance, especially for `MLFloat16`.

**Recommendation:**
Hoist the iteration of the **innermost dimension** out of the generic index tracking logic. Since the input `X` is always contiguous in the last dimension, and `Scale`/`Bias` will have a constant stride (either 0 or 1) for that dimension, exposing this allows the compiler to generate efficient vector code.

**Proposed Change (Concept):**
Modify `ComputeJobGeneric` to iterate in chunks of the last dimension.

```cpp
// In ComputeJobGeneric:

// 1. Identify properties of the innermost dimension
int64_t last_dim_size = params.x_inner_dims.back();
int64_t sc_last_stride = params.sc_inner_inc.back();
int64_t bi_last_stride = has_bias ? params.bi_inner_inc.back() : 0; // Check has_bias first

// 2. Outer loop iterates over the "counts" of chunks
// Total iterations = norm_size / last_dim_size
int64_t num_chunks = norm_size / last_dim_size;

for (int64_t c = 0; c < num_chunks; ++c) {
    // Calculate base offsets for the start of this chunk using the generic logic
    int64_t off_sc = off_sc_row;
    int64_t off_bi = off_bi_row;
    for (size_t d = 0; d < static_cast<size_t>(last_rank) - 1; ++d) { // Note: -1 to skip last dim
        off_sc += idx[d] * sc_inner_inc[d];
        if (bias_data) off_bi += idx[d] * bi_inner_inc[d];
    }

    // 3. Inner loop: Tight, vectorized loop over the last dimension
    // The compiler can now verify that strides are constant (0 or 1) and vectorize this.
    int64_t base_h = c * last_dim_size;
    for (int64_t i = 0; i < last_dim_size; ++i) {
        int64_t h = base_h + i;
        // Use simple stride addition
        T s = scale_data[off_sc + i * sc_last_stride];
        // ... Perform Math ...
    }

    // 4. Update the multi-dimensional index 'idx' for the *next* chunk
    // Iterate backwards from the second-to-last dimension
    for (int64_t d = last_rank - 2; d >= 0; --d) {
        if (++idx[static_cast<size_t>(d)] < params.x_inner_dims[static_cast<size_t>(d)]) break;
        idx[static_cast<size_t>(d)] = 0;
    }
}
```

*Note: This requires handling the edge case where `last_rank == 0` or `1` separately or ensuring `last_rank - 1` logic holds.*

### 2\. Code Quality: Reduce Duplication via Templates

**The Issue:**
You have two copies of `ComputeJobGeneric`: one for `T` (float/double) and one for `MLFloat16`. The broadcasting logic and loop structure are identical; only the load/store/cast mechanics differ.

**Recommendation:**
Refactor the broadcasting logic into a single function that accepts a **Calculator Policy** or a **Functor**.

```cpp
// Define a functor for Float/Double
template <typename T>
struct NormalizationMath {
    static T Load(const T* ptr, int64_t offset) { return ptr[offset]; }
    static void Store(T* ptr, int64_t offset, float val) { ptr[offset] = static_cast<T>(val); }
    // ... other helpers
};

// Define a functor for MLFloat16
struct HalfMath {
    static float Load(const MLFloat16* ptr, int64_t offset) { return static_cast<float>(ptr[offset]); }
    static void Store(MLFloat16* ptr, int64_t offset, float val) { ptr[offset] = MLFloat16(val); }
};

// Single Generic Implementation
template <typename DataT, typename MathPolicy>
void ComputeJobGenericShared(...) {
    // ... unified broadcasting logic ...
    // float s = MathPolicy::Load(scale_data, off_sc);
    // ...
    // MathPolicy::Store(Y_data, h, result);
}
```

This reduces code size and ensures that fixes to the complex broadcasting logic (like the optimization in point \#1) automatically apply to all data types.

### 3\. Validation: Refine `CheckInputs` Constraints

**The Issue:**
The constraint `params.norm_size <= 1` returns an error. While mathematically valid (variance is 0), standard implementations often handle scalar normalization (returning 0) rather than erroring, or `norm_size` might technically be 1 if `axis` is the last dim and dim size is 1.

**Recommendation:**
Ensure that `norm_size == 1` behaves gracefully. If the input is `[Batch, 1]`, normalization usually results in zeros (since mean=value, value-mean=0).
Also, in `LayerNormHelper::CheckInputs`:

```cpp
// While strictly correct for variance calc, consider if this check is too aggressive for edge cases 
// where dimensions might be dynamically 1.
if (params.norm_size < 1) { ... } // Change <= to < ?
```

*Note: If your specific kernel requires variance calculation that divides by `norm_size`, 1 is fine. If it divides by `norm_size - 1` (unbiased), 1 is bad.*

### 4\. Testing: Add "Mixed" Broadcasting Cases

**The Issue:**
The current tests cover:

1.  No broadcasting.
2.  Outer broadcasting (Batch broadcast).
3.  Inner broadcasting (scalar scale).

**Recommendation:**
Add a test case for **Mixed/Strided Broadcasting** to ensure the generic loop truly handles non-contiguous strides correctly.

**New Test Case Example:**

  * **Input:** `[1, 2, 4]` (Batch=1, Seq=2, Feature=4)
  * **Axis:** 1 (Normalize over `[2, 4]`)
  * **Scale:** `[1, 4]` (Broadcast over Seq dim 2).
      * Here, `Scale` has shape `[1, 4]` matching the inner dims `[2, 4]` via broadcasting `1 -> 2`.
      * The stride for `Scale` on the Seq dimension is **0**, but for Feature dimension is **1**.
      * This exercises the specific logic where one inner stride is 0 and another is 1.



```cpp
TEST(LayerNormTest, LayerNorm_Scale_Broadcast_Inner_Mixed) {
  OpTester test("LayerNormalization", 17);
  test.AddAttribute<float>("epsilon", 1e-05f);
  test.AddAttribute<int64_t>("axis", 1); // Normalize over last 2 dims (2, 4)

  // Input: [1, 2, 4]
  std::vector<int64_t> dims{1, 2, 4};
  std::vector<float> x = {
      0, 1, 2, 3,   // Row 0
      4, 5, 6, 7    // Row 1
  }; 
  test.AddInput<float>("X", dims, x);

  // Scale: [1, 4]. Broadcasts to [2, 4]
  // Stride on dim '2' is 0. Stride on dim '4' is 1.
  std::vector<float> scale = {1.0f, 0.5f, 1.0f, 0.5f}; 
  test.AddInput<float>("Scale", {1, 4}, scale); 

  // Expected output calculation required...
  test.Run();
}
```

### 5\. Pre-calculation Optimization

**The Issue:**
Inside `CheckInputs`, you calculate `sc_outer_inc`, `bi_outer_inc` etc.

```cpp
params.sc_outer_inc.push_back(params.sc_strides[static_cast<size_t>(i)]);
```

This allocates memory (`push_back`) every time `Compute` is called (since `LayerNormParams` is stack-allocated per Compute).

**Recommendation:**
Since `LayerNormParams` is recreated every call, avoid `InlinedVector::push_back` inside the loop if possible, or `reserve` size upfront.
More importantly, `CheckInputs` runs on *every inference*.

  * Ensure `InlinedVector` capacity (8) is sufficient for common models (it usually is).
  * Consider calculating `sc_inner_inc` and `bi_inner_inc` only if `use_generic_broadcast` becomes true, or lazily. Currently, you calculate them always, even for the fast path which doesn't use them.

**Move heavy setup inside the conditional:**

```cpp
// In LayerNormHelper::CheckInputs
// ... detect if fast path is possible ...

if (params.broadcast_param == kLayerNormInvalidInput || outer_dep) {
   params.use_generic_broadcast = true;
   // ONLY compute expensive generic strides (sc_inner_inc, etc.) HERE
   // This saves overhead for the 99% case (BERT/Transformer fast path).
}
```

naomiOvad:
@tianleiwu Thanks for the detailed review!

Regarding the CI failures — they seem unrelated to this PR.
The failing pipelines are due to transient infrastructure issues (GitHub 503
errors when downloading googletest). Once GitHub is back to normal, a rerun
should resolve them.

I’ve applied all 5 suggested improvements:
1) optimized the generic inner loop,
2) unified the generic implementation using MathPolicy,
3) refined norm_size validation,
4) added the mixed-broadcast test,
5) moved generic stride setup inside the generic-path branch.

Updated code pushed. Please let me know if anything else needs adjustment.

naomiOvad:
@tianleiwu I reviewed the CI failures and it looks like the 4 failing CI jobs are unrelated to the changes here.
The macOS job fails in vcpkg (ranlib I/O error), Windows CUDA has a generic build failure without compiler errors, and the QNN jobs fail due to unsupported ops/layouts inside the QNN EP.
Let me know if there’s anything specific I should adjust. Thanks!

tianleiwu:
The latest commit is a massive improvement. It elegantly addresses the performance concerns by vectorizing the inner loop and significantly improves code maintainability via the `NormalizationMath` policy functors. The generic path is now both robust and performant.

I have just one critical safety finding regarding input validation that could cause a crash (segfault) if not addressed.

### 1\. Critical Safety Fix: Potential Segfault in Helper

In `layer_norm_helper.h`, the logic assumes the rank of `Scale` (and `Bias`) is less than or equal to the rank of `X`. While the ONNX spec implies this for unidirectional broadcasting, passing a malformed model where `Scale.ndim > X.ndim` will currently cause a **segfault** (buffer overflow) instead of a graceful error.

**The Issue:**

```cpp
// layer_norm_helper.h
const size_t xr = x_shape.NumDimensions();
const size_t sr = scale_shape.NumDimensions();

// ...

// If sr > xr, then (xr - 1 - i) wraps around to SIZE_MAX when i >= xr
for (size_t i = 0; i < sr; ++i) {
  params.sc_dims[xr - 1 - i] = scale_shape.GetDims()[sr - 1 - i]; 
}
```

**Recommended Fix:**
Add a check at the top of `CheckInputs` to validate ranks.

```cpp
// layer_norm_helper.h -> CheckInputs

const size_t xr = x_shape.NumDimensions();
const size_t sr = scale_shape.NumDimensions();
const size_t br = has_bias ? bias_shape.NumDimensions() : 0;

// ADD THIS CHECK:
if (sr > xr || (has_bias && br > xr)) {
  return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT, 
                         kLayerNormInputShapeMismatchError,
                         " Scale/Bias rank cannot exceed Input rank.");
}

params.x_dims.clear();
// ...
```

onnxruntime/core/providers/cpu/nn/layer_norm_helper.h
```diff
@@ -23,15 +23,29 @@ struct LayerNormParams {
   int64_t scale_size;
   int64_t bias_size;
   int64_t broadcast_param;
+  bool use_generic_broadcast{false};  // true: full NumPy-style broadcast; false: legacy broadcast_param path
+  onnxruntime::InlinedVector<int64_t, 8> x_dims;
+  onnxruntime::InlinedVector<int64_t, 8> x_inner_dims;  // X.shape[axis:]
+  onnxruntime::InlinedVector<int64_t, 8> sc_dims;
```
tianleiwu: nit: use full name "scale_" instead of "sc_"

onnxruntime/core/providers/cpu/nn/layer_norm_helper.h
```diff
@@ -23,15 +23,29 @@ struct LayerNormParams {
...
+  onnxruntime::InlinedVector<int64_t, 8> bi_dims;
```
tianleiwu: nit: use full name "bias_" instead of "bi_"

naomiOvad: Renamed to scale_ / bias_ as suggested.

naomiOvad:
> The latest commit is a massive improvement. It elegantly addresses the performance concerns by vectorizing the inner loop and significantly improves code maintainability via the `NormalizationMath` policy functors. The generic path is now both robust and performant.
> 
> I have just one critical safety finding regarding input validation that could cause a crash (segfault) if not addressed.
> 
> ### 1. Critical Safety Fix: Potential Segfault in Helper
> In `layer_norm_helper.h`, the logic assumes the rank of `Scale` (and `Bias`) is less than or equal to the rank of `X`. While the ONNX spec implies this for unidirectional broadcasting, passing a malformed model where `Scale.ndim > X.ndim` will currently cause a **segfault** (buffer overflow) instead of a graceful error.
> 
> **The Issue:**
> 
> ```c++
> // layer_norm_helper.h
> const size_t xr = x_shape.NumDimensions();
> const size_t sr = scale_shape.NumDimensions();
> 
> // ...
> 
> // If sr > xr, then (xr - 1 - i) wraps around to SIZE_MAX when i >= xr
> for (size_t i = 0; i < sr; ++i) {
>   params.sc_dims[xr - 1 - i] = scale_shape.GetDims()[sr - 1 - i]; 
> }
> ```
> 
> **Recommended Fix:** Add a check at the top of `CheckInputs` to validate ranks.
> 
> ```c++
> // layer_norm_helper.h -> CheckInputs
> 
> const size_t xr = x_shape.NumDimensions();
> const size_t sr = scale_shape.NumDimensions();
> const size_t br = has_bias ? bias_shape.NumDimensions() : 0;
> 
> // ADD THIS CHECK:
> if (sr > xr || (has_bias && br > xr)) {
>   return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT, 
>                          kLayerNormInputShapeMismatchError,
>                          " Scale/Bias rank cannot exceed Input rank.");
> }
> 
> params.x_dims.clear();
> // ...
> ```

Added the rank validation to CheckInputs as suggested. Thanks!

naomiOvad:
@tianleiwu I investigated the 3 failing CI jobs:

• **Windows GPU CUDA CI** – nvcc fails when compiling
  `core/providers/cuda/nn/layer_norm_impl.cu`. The diagnostics come from
  Abseil headers under vcpkg (hash.h, raw_hash_set.h, etc.) and are treated
  as errors due to `-Werror all-warnings`. This PR doesn’t modify any CUDA
  kernels or Abseil usage.

• **Windows x86 CPU pipeline** – the failure is in the existing test
  `SessionStatePrepackingTest.PrePackingTest/0` (“schema … already registered”
  in `session_state_test.cc`), which is unrelated to the RMSNorm/LayerNorm
  changes here.

• **Windows GPU TensorRT pipeline** – it fails in the same CUDA build step
  for `onnxruntime_providers_cuda.vcxproj` (again when compiling
  `layer_norm_impl.cu` with the Abseil warnings promoted to errors), not in
  the CPU RMSNorm/LayerNorm code or tests.

From what I can see, these look like CI/toolchain issues rather than something
caused by this PR. Please let me know how you’d like to proceed.

tianleiwu:
@naomiOvad, your branch is 68 commits behind. Please merge latest main branch. I guess some commits are needed to pass CI pipelines.

naomiOvad:
@tianleiwu I’ve merged the latest main into this branch. Thanks!

tianleiwu:
@naomiOvad, there is build error (likely caused by adding `#include "core/common/inlined_containers.h"`):
```
2025-12-08T19:47:36.5681890Z E:\_work\_temp\build\RelWithDebInfo\vcpkg_installed\x64-windows-static-md\include\absl/hash/internal/hash.h(491): error #68-D: integer conversion resulted in a change of sign [E:\_work\_temp\build\RelWithDebInfo\onnxruntime_providers_cuda.vcxproj]
2025-12-08T19:47:36.5755054Z           sizeof(T) == -1,
2025-12-08T19:47:36.5774213Z                        ^
2025-12-08T19:47:36.5774545Z   
2025-12-08T19:47:36.5795020Z   Remark: The warnings can be suppressed with "-diag-suppress <warning-number>"
2025-12-08T19:47:36.5795559Z   
2025-12-08T19:47:36.6055585Z E:\_work\_temp\build\RelWithDebInfo\vcpkg_installed\x64-windows-static-md\include\absl/hash/hash.h(337): error #549-D: variable "s" is used before its value is set [E:\_work\_temp\build\RelWithDebInfo\onnxruntime_providers_cuda.vcxproj]
2025-12-08T19:47:36.6057008Z         return s;
2025-12-08T19:47:36.6191830Z                ^
2025-12-08T19:47:36.6387620Z   
2025-12-08T19:47:36.6428327Z E:\_work\_temp\build\RelWithDebInfo\vcpkg_installed\x64-windows-static-md\include\absl/container/internal/raw_hash_set.h(473): error #69-D: integer conversion resulted in truncation [E:\_work\_temp\build\RelWithDebInfo\onnxruntime_providers_cuda.vcxproj]
2025-12-08T19:47:36.6429803Z           static_cast<uint16_t>(reinterpret_cast<uintptr_t>(&seed));
2025-12-08T19:47:36.6430193Z                       ^
2025-12-08T19:47:36.6430420Z   
2025-12-08T19:47:36.6431116Z   3 errors detected in the compilation of "E:/_work/onnxruntime/onnxruntime/onnxruntime/core/providers/cuda/nn/layer_norm_impl.cu".
2025-12-08T19:47:36.7420671Z E:\_work\_temp\v12.8\extras\visual_studio_integration\MSBuildExtensions\CUDA 12.8.targets(800,9): error MSB3721: The command ""E:\_work\_temp\v12.8\bin\nvcc.exe"  --use-local-env -ccbin "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.44.35207\bin\HostX64\x64" -x cu   -IE:\_work\onnxruntime\onnxruntime\include\onnxruntime -IE:\_work\onnxruntime\onnxruntime\include\onnxruntime\core\session -IE:\_work\_temp\build\RelWithDebInfo -IE:\_work\onnxruntime\onnxruntime\onnxruntime -I"E:\_work\_temp\build\RelWithDebInfo\_deps\cutlass-src\include" -I"E:\_work\_temp\build\RelWithDebInfo\_deps\cutlass-src\examples" -I"E:\_work\_temp\build\RelWithDebInfo\_deps\cutlass-src\tools\util\include" -I"E:\_work\_temp\build\RelWithDebInfo\_deps\cudnn_frontend-src\include" -I"E:\_work\_temp\build\RelWithDebInfo\vcpkg_installed\x64-windows-static-md\include" -IE:\_work\_temp\v12.8\include -I"E:\_work\_temp\build\RelWithDebInfo\vcpkg_installed\x64-windows-static-md\include\eigen3" -IE:\_work\_temp\v12.8\include     --keep-dir onnxrunt.C920140F\x64\RelWithDebInfo  -maxrregcount=0    --machine 64 --compile -cudart shared -allow-unsupported-compiler -Xfatbin=-compress-all --expt-relaxed-constexpr --Werror default-stream-launch -Xcudafe --diag_suppress=bad_friend_decl -Xcudafe --diag_suppress=unsigned_compare_with_zero -Xcudafe --diag_suppress=expr_has_no_effect -include algorithm -std=c++17 --generate-code=arch=compute_86,code=[sm_86] -Xcudafe --diag_suppress=conversion_function_not_usable --threads 3 --static-global-template-stub=false --diag-suppress=221 -Werror all-warnings -Xcompiler="/MP /guard:cf /Qspectre /Ob1 /EHsc -Ob1 -Zi /utf-8 /sdl /experimental:external /external:W0 /external:IE:/_work/onnxruntime/onnxruntime/cmake /external:IE:/_work/_temp/build/RelWithDebInfo /wd4251 /wd4201 /wd4324 /wd5054 /w15038 /wd4251 /wd4201 /wd4324 /wd5054 /w15038 /wd4505 /wd4834 /wd4127 /Zc:__cplusplus /bigobj"   -D_WINDOWS -DNDEBUG -DWIN32 -DWINAPI_FAMILY=100 -DWINVER=0x0A00 -D_WIN32_WINNT=0x0A00 -DNTDDI_VERSION=0x0A000000 -DVER_MAJOR=1 -DVER_MINOR=24 -DVER_BUILD=0 -DVER_PRIVATE=0 -D"VER_STRING=\"1.24.0\"" -DCPUINFO_SUPPORTED_PLATFORM=1 -DORT_ENABLE_STREAM -DEIGEN_USE_THREADS -DDISABLE_CUSPARSE_DEPRECATED -DPLATFORM_WINDOWS -DNOGDI -DNOMINMAX -D_USE_MATH_DEFINES -D_SILENCE_ALL_CXX17_DEPRECATION_WARNINGS -DONNXRUNTIME_ENABLE_MEMLEAK_CHECK -DUSE_CUDA=1 -DUSE_FLASH_ATTENTION=1 -DUSE_MEMORY_EFFICIENT_ATTENTION=1 -DUSE_TENSORRT=1 -D"FILE_NAME=\"onnxruntime_providers_cuda.dll\"" -DONLY_C_LOCALE=0 -DHAS_STRING_VIEW=1 -DONNX_NAMESPACE=onnx -DONNX_ML=1 -D__ONNX_DISABLE_STATIC_REGISTRATION -DWIN32_LEAN_AND_MEAN -DEIGEN_MPL2_ONLY -DEIGEN_HAS_CONSTEXPR -DEIGEN_HAS_VARIADIC_TEMPLATES -DEIGEN_HAS_CXX11_MATH -DEIGEN_HAS_CXX11_ATOMIC -DEIGEN_STRONG_INLINE=inline -DCPUINFO_SUPPORTED -DEXCLUDE_SM_80 -DEXCLUDE_SM_89 -DEXCLUDE_SM_90 -DEXCLUDE_SM_100 -DEXCLUDE_SM_120 -DENABLE_FP8 -DENABLE_FP4 -DENABLE_DLPACK -DENABLE_CUDA_NHWC_OPS -D"CMAKE_INTDIR=\"RelWithDebInfo\"" -Donnxruntime_providers_cuda_EXPORTS -D_WINDLL -D_MBCS -DWIN32 -D_WINDOWS -DWINAPI_FAMILY=100 -DWINVER=0x0A00 -D_WIN32_WINNT=0x0A00 -DNTDDI_VERSION=0x0A000000 -DNDEBUG -DEIGEN_HAS_C99_MATH -DVER_MAJOR=1 -DVER_MINOR=24 -DVER_BUILD=0 -DVER_PRIVATE=0 -D"VER_STRING=\"1.24.0\"" -DCPUINFO_SUPPORTED_PLATFORM=1 -DORT_ENABLE_STREAM -DEIGEN_USE_THREADS -DDISABLE_CUSPARSE_DEPRECATED -DPLATFORM_WINDOWS -DNOGDI -DNOMINMAX -D_USE_MATH_DEFINES -D_SILENCE_ALL_CXX17_DEPRECATION_WARNINGS -DONNXRUNTIME_ENABLE_MEMLEAK_CHECK -DUSE_CUDA=1 -DUSE_FLASH_ATTENTION=1 -DUSE_MEMORY_EFFICIENT_ATTENTION=1 -DUSE_TENSORRT=1 -D"FILE_NAME=\"onnxruntime_providers_cuda.dll\"" -DONLY_C_LOCALE=0 -DHAS_STRING_VIEW=1 -DONNX_NAMESPACE=onnx -DONNX_ML=1 -D__ONNX_DISABLE_STATIC_REGISTRATION -DWIN32_LEAN_AND_MEAN -DEIGEN_MPL2_ONLY -DEIGEN_HAS_CONSTEXPR -DEIGEN_HAS_VARIADIC_TEMPLATES -DEIGEN_HAS_CXX11_MATH -DEIGEN_HAS_CXX11_ATOMIC -DEIGEN_STRONG_INLINE=inline -DCPUINFO_SUPPORTED -DEXCLUDE_SM_80 -DEXCLUDE_SM_89 -DEXCLUDE_SM_90 -DEXCLUDE_SM_100 -DEXCLUDE_SM_120 -DENABLE_FP8 -DENABLE_FP4 -DENABLE_DLPACK -DENABLE_CUDA_NHWC_OPS -D"CMAKE_INTDIR=\"RelWithDebInfo\"" -Donnxruntime_providers_cuda_EXPORTS -Xcompiler "/EHsc /W4 /nologo /O2 /FS /Zi  /MD /GR" -Xcompiler "/Fdonnxruntime_providers_cuda.dir\RelWithDebInfo\vc143.pdb" -o onnxruntime_providers_cuda.dir\RelWithDebInfo\layer_norm_impl.obj "E:\_work\onnxruntime\onnxruntime\onnxruntime\core\providers\cuda\nn\layer_norm_impl.cu"" exited with code -1. [E:\_work\_temp\build\RelWithDebInfo\onnxruntime_providers_cuda.vcxproj]
```

The proposed fix is to do some refactoring tomove the following code to a new file like layer_norm_macro.h:
```
#pragma once

// Below is a macro to compute the offset for scale and bias data for a row of X.
#ifndef LAYER_NORM_SCALE_BIAS_OFFSET
#define LAYER_NORM_SCALE_BIAS_OFFSET(broadcast_param, x_row, norm_size) \
  ((broadcast_param == 0) ? 0                                           \
                          : norm_size * (broadcast_param > 0 ? x_row / broadcast_param : x_row % (-broadcast_param)))
#endif
```

Then modify onnxruntime\core\providers\cuda\nn\layer_norm_impl.cu to include layer_norm_macro.h instead of layer_norm_helper.h.

naomiOvad:
@tianleiwu Thanks for the suggestion!  
I applied the refactor:

• Moved LAYER_NORM_SCALE_BIAS_OFFSET into a new header  
  core/providers/cpu/nn/layer_norm_macro.h  
• layer_norm_helper.h now includes this header  
• layer_norm_impl.cu includes layer_norm_macro.h instead of 
  layer_norm_helper.h, so CUDA no longer pulls in inlined_containers

Changes pushed.

onnxruntime/core/providers/cpu/nn/layer_norm_macro.h
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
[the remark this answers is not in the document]
tianleiwu: Please see https://github.com/microsoft/onnxruntime/blob/main/docs/Coding_Conventions_and_Standards.md#linting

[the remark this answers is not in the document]
naomiOvad: Formatting updated using lintrunner.

tianleiwu (approved):

PR #26616, opened 2025-11-20
Adding candidate metadata key for tracking EP's OS driver version

adrastogi:
### Description
This change adds a well-known key name (`os_driver_version`) corresponding to the OS driver version associated with an EP. We will eventually flesh this out to enable retrieving it from the `OrtEpDevice` if it's been populated, but for starters we reserve the name. 

### Motivation and Context
We have a scenario in WebNN where the browser would like to get the driver version associated with a given EP (this is to enable policy against the driver, e.g. for maintaining a blocklist if a particular driver version has a scenario-blocking bug in it). Having a mechanism to retrieve the driver version via ORT would help with implementing this feature.

### DIFF
--- a/include/onnxruntime/core/session/onnxruntime_ep_device_ep_metadata_keys.h
+++ b/include/onnxruntime/core/session/onnxruntime_ep_device_ep_metadata_keys.h
@@ -9,7 +9,7 @@
 // Key for the execution provider version string. This should be available for all plugin EPs.
 static const char* const kOrtEpDevice_EpMetadataKey_Version = "version";
 
-// Key for the execution provider OS driver version. This should be available for all plugin EPs.
+// Key for the execution provider OS driver version.
 static const char* const kOrtEpDevice_EpMetadataKey_OSDriverVersion = "os_driver_version";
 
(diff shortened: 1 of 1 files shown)


include/onnxruntime/core/session/onnxruntime_ep_device_ep_metadata_keys.h
```diff
@@ -9,6 +9,9 @@
 // Key for the execution provider version string. This should be available for all plugin EPs.
 static const char* const kOrtEpDevice_EpMetadataKey_Version = "version";
 
+// Key for the execution provider OS driver version. This should be available for all plugin EPs.
```
adrianlizarraga: I'm not sure if we should say it should always be available. In general, an EP may not be associated with a particular driver (e.g., some cpu-based EP implementation).

Also, we don't enforce it within ORT.

adrastogi: Great points, I removed that part of the comment.

adrianlizarraga (approved):

PR #26617, opened 2025-11-20
[core] allow using initializer allocator for prepack

fs-eire:
### Description

This PR makes ORT to prefer initializer allocator when calling `OpKernel::PrePack`.

If an EP does not register an initializer allocator (currently only WebGPU does this), the behavior is kept unchanged.

### Motivation and Context

Helps to improve the memory usage when doing prepack.


onnxruntime/core/framework/session_state.cc
```diff
@@ -588,15 +588,15 @@ Status SessionState::PrepackConstantInitializedTensors(
                   // within this session. Or if the weight is not present on disk,
                   // we store the newly minted pre-packed data.
 
-                  AllocatorPtr session_cpu_alloc = GetAllocator(kernel->Info().GetDevice(OrtMemType::OrtMemTypeDefault));
+                  AllocatorPtr session_initializer_alloc = GetInitializerAllocator(kernel->Info().GetDevice(OrtMemType::OrtMemTypeDefault));
```
yuslepukhin, line 591: >AllocatorPtr session_initializer_allo [](http://example.com/codeflow?start=18&length=37)

Q. Is it guaranteed accessible from CPU code? We need to compute the container key. Sorry, I do not remember the properties of the an allocator.

fs-eire: The original variable name is confusing. It actually returns the registered allocator by looking up the map using the OrtDevice as key. for a node on webgpu EP, this returns a webgpu allocator instead of a CPU one.

PR #26619, opened 2025-11-20
[QNN EP] Support batch multiple on HTP backend

chunghow-qti:
### Description

This change adds support for the batch multiplier in the QNN HTP backend, allowing the compiled batch size to differ from the runtime batch size, as long as the runtime batch size is divisible by the compiled batch size.

Included in this change:
* Skip input/output validation in the inference session to accommodate varying batch sizes when the batch multiplier option is used.
* Verify that disable_cpu_ep_fallback and the HTP backend are used with the batch multiplier option, since batch multiplier is only supported when the entire graph runs on QNN.
* Ensure that the HTP backend is used inside QnnModel::ExecuteGraph.
* Modify the batch dimension of qnn_inputs and qnn_outputs when the batch multiplier is triggered.

### Motivation and Context

* This change supports batch multiplier in QNN API to ORT as described in this page: https://docs.qualcomm.com/bundle/publicresource/topics/80-63442-10/function_QnnGraph_8h_1a3ea05f42a9295f9a74a2e3a0cdd64228.html
* relevant PR: https://github.com/microsoft/onnxruntime/pull/26336

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/qnn_model.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_model.cc
@@ -174,6 +174,46 @@ Status QnnModel::FinalizeGraphs(const logging::Logger& logger) {
   return Status::OK();
 }
 
+// Helper function to check if ORT tensor shape matches expected QNN tensor shape (excluding batch dimension at index 0)
+template <typename OrtValueType>
+static Status CheckShape(const OrtValueType& ort_tensor,
+                         const QnnTensorInfo& qnn_io_info) {
+  const auto input_output_shape = ort_tensor.GetTensorTypeAndShapeInfo().GetShape();
+  const auto shape_size = input_output_shape.size();
+  const auto& expected_shape = qnn_io_info.ori_dimensions_;
+  const auto expected_shape_size = expected_shape.size();
+  const auto& tensor_name = qnn_io_info.tensor_wrapper->GetName();
+
+  ORT_RETURN_IF_NOT(shape_size == expected_shape_size,
+                    "Invalid rank for tensor: ", tensor_name,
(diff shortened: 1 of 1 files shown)


onnxruntime/core/session/inference_session.cc
```diff
@@ -2981,10 +2981,17 @@ Status InferenceSession::Run(const RunOptions& run_options,
 
       // log evaluation start to trace logging provider
       env.GetTelemetryProvider().LogEvaluationStart(session_id_);
-
+#ifdef USE_QNN
```
adrianlizarraga, line 2984: Considering that we're moving towards plugin EPs, we should avoid EP-specific code in the core onnxruntime library. Otherwise, we would need a special build of onnxruntime.dll that works with the plugin QNN EP.

qti-hungjuiw: Hi @adrianlizarraga, would it make sense to follow the approach of `kOrtSessionOptionsConfigStrictAllowReleasedOpsetsOnly` and introduce a new session option, such as `kOrtSessionOptionsConfigStrictAllowConsistentShapeOnly` (defaulting to "1")?

This option could be **EP-agnostic** and would allow skipping `InferenceSession::CheckShapes` based on user intent. Necessary warnings can be provided when the option is set, similar to the approach in https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/graph/model_load_utils.h#L58.

I believe this approach could help us avoid relying on the `USE_QNN` macro.

adrianlizarraga: Yes, that moves us toward a more ep-agnostic approach. Thanks.

PR #26624, opened 2025-11-21
[Quantization] Fix static quantize runner usage.

minfhong-qti:
### Description

- Input pb files were read in incorrect order.
  - Cause: Python `sorted` was used to acquire sorted input files in order. However, the input files would be in incorrect order since "10" is lexicographically smaller than "2".
  - Fix: Revise to enumerating indices to read input files.
- CumSum's output wasn't quantized.
  - Cause: CumSum wasn't registered into QDQ registry.
  - Fix: Register CumSum with QDQDirect8bitOp.

### Motivation and Context

Fix two issues in `static_quantize_runner` usage.

### DIFF
--- a/onnxruntime/python/tools/quantization/registry.py
+++ b/onnxruntime/python/tools/quantization/registry.py
@@ -87,6 +87,7 @@ QDQRegistry = {
     "LayerNormalization": QDQNormalization,
     "BatchNormalization": QDQNormalization,
     "TopK": QDQDirect8BitOp,
+    "CumSum": QDQOperatorBase,
 }
 
 
--- a/onnxruntime/python/tools/quantization/static_quantize_runner.py
+++ b/onnxruntime/python/tools/quantization/static_quantize_runner.py
@@ -20,7 +20,7 @@ class OnnxModelCalibrationDataReader(CalibrationDataReader):
         name2tensors = []
         for data_dir in data_dirs:
             name2tensor = {}
-            data_paths = [os.path.join(data_dir, a) for a in sorted(os.listdir(data_dir))]
+            data_paths = [os.path.join(data_dir, f"input_{input_idx}.pb") for input_idx in range(len(model_inputs))]
             data_ndarrays = [self.read_onnx_pb_data(data_path) for data_path in data_paths]
             for model_input, data_ndarray in zip(model_inputs, data_ndarrays, strict=False):
                 name2tensor[model_input.name] = data_ndarray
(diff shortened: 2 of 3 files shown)


minfhong-qti:
Hi @edgchen1, please kindly review this PR or reassign to proper reviewer.
Really appreciate your time and effort.

onnxruntime/python/tools/quantization/static_quantize_runner.py
```diff
@@ -20,7 +20,7 @@ def __init__(self, model_path):
         name2tensors = []
         for data_dir in data_dirs:
             name2tensor = {}
-            data_paths = [os.path.join(data_dir, a) for a in sorted(os.listdir(data_dir))]
+            data_paths = [os.path.join(data_dir, f"input_{input_idx}.pb") for input_idx in range(len(model_inputs))]
```
adrianlizarraga, line 23: Should this change also require an update to the [README](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/python/tools/quantization/README.md)?

onnxruntime/python/tools/quantization/registry.py
```diff
@@ -87,6 +87,7 @@
     "LayerNormalization": QDQNormalization,
     "BatchNormalization": QDQNormalization,
     "TopK": QDQDirect8BitOp,
+    "CumSum": QDQDirect8BitOp,
```
adrianlizarraga: The `QDQDirect8BitOp` is for operators that can use the same scale/zero-point at the inputs and outputs (e.g., Transpose, Reshape, etc.). Is this the appropriate base class for `CumSum` given that the cumulative sum value range could potentially be very different from that of the inputs, which necessitates a different output scale?

onnxruntime/python/tools/quantization/static_quantize_runner.py
minfhong-qti: I believe the naming `input_{idx}.pb` is already implicitly assumed. I've updated the README in the latest commit to explicitly point out.

onnxruntime/python/tools/quantization/registry.py
minfhong-qti: I've changed it to `QDQOperatorBase` to separately quantize input and output.

In addition, I've found out some conflicts in current registry mechanism.
- https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/python/tools/quantization/registry.py#L109 The look-up here fallbacks unregistered op to `QDQOperatorBase` which makes sense in my opinion.
- However, this fallback would never be used in `quantize_static` as it explicitly specifies `op_types_to_quantize` https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/python/tools/quantization/quantize.py#L645 if not given by config. Then, op with types not in current registry will never be quantized due to the check https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/python/tools/quantization/base_quantizer.py#L180.

Not sure whether it is intentional but I'll leave it as future work if it requires any modification.

PR #26625, opened 2025-11-21
[QNN EP] Support NonZero.

minfhong-qti:
### Description

- Implement NonZero op builder and regsiter QDQ selector.
- Implement ShapeNonZero QNN preprocess to fix shape. Test: UTs.

### Motivation and Context

Unlike ONNX NonZero which normally has dynamic output shape, QNN defines NonZero to always have maximum output size to avoid the dynamic shape. In order to meet QNN's expectation, the model with NonZero node must be preprocessed first to fix the shape beforehand. Additionally, QNN NonZero has shape in `[#elements, rank]` and thus an extra Transpose is required.

### DIFF
--- a/onnxruntime/core/optimizer/qdq_transformer/selectors_actions/shared/utils.cc
+++ b/onnxruntime/core/optimizer/qdq_transformer/selectors_actions/shared/utils.cc
@@ -48,7 +48,8 @@ static const OpVersionsAndSelector::OpVersionsMap GetMiscOpVersionsMap() {
 // These produce int64 indices output, which can't be quantized, so there's no downstream Q node.
 static const OpVersionsAndSelector::OpVersionsMap GetDropDQOpVersionsMap() {
   return {{"ArgMax", {}},
-          {"ArgMin", {}}};
+          {"ArgMin", {}},
+          {"NonZero", {}}};
 }
 
 static const OpVersionsAndSelector::OpVersionsMap GetUnaryOpVersionsMap() {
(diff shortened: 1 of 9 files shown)


skottmckay:
Doesn't this just delay the need to deal with a dynamic shape? Let's assume you use -1 for an invalid index for the empty entries. That gives the NonZero node a fixed output shape, but how do downstream operators know how to handle an invalid index correctly? How do you track what's valid/invalid across all downstream operators? 

If QNN can handle that tracking internally, what ensures that when returning output from QNN to ORT that only valid values are returned? 
- Simplest example is a NonZero node being at the boundary of nodes QNN takes. The output of that can't be directly returned to ORT if there are invalid indices in it.
- More complicated is when there are multiple operators downstream of the QNN NonZero. How do you know how to truncate the output that ORT receives to take into account which indices returned by NonZero were invalid?

onnxruntime/core/optimizer/qdq_transformer/selectors_actions/shared/utils.cc
```diff
@@ -48,7 +48,8 @@ static const OpVersionsAndSelector::OpVersionsMap GetMiscOpVersionsMap() {
 // These produce int64 indices output, which can't be quantized, so there's no downstream Q node.
 static const OpVersionsAndSelector::OpVersionsMap GetDropDQOpVersionsMap() {
   return {{"ArgMax", {}},
-          {"ArgMin", {}}};
+          {"ArgMin", {}},
+          {"NonZero", {}}};
```
yuslepukhin, line 52: >{"NonZero", {}}} [](http://example.com/codeflow?start=10&length=16)

This would apply to all NonZero then, not just QNN?

skottmckay: Is it safe to drop the DQ? Wouldn't the zero point and scale affect which values are zero?

minfhong-qti:
Close this PR as no actual request.

PR #26628, opened 2025-11-21
[QNN EP] Optimize Session Creation Time

chunghow-qti:
### Description

qnn::utils::TwoDimensionTranspose makes the bottleneck during session creation, because there is a double for loop memcpy. If the weight is quite large, it will be very slow. And it's called a total of 3 times by ReshapeGemmFusion.

```
QnnModel::ComposeGraph → ReshapeGemmFusion::AddToModelBuilder → CreateOrValidateOnQnn → qnn::utils::TwoDimensionTranspose
QNNExecutionProvider::GetCapability → QNNExecutionProvider::GetSupportedNodes → ReshapeGemmFusion::IsSupported → CreateOrValidateOnQnn → qnn::utils::TwoDimensionTranspose (do QNN OP validation)
QNNExecutionProvider::GetCapability → QNNExecutionProvider::GetSupportedNodes → onnxruntime::qnn::ReshapeGemmFusion::IsSupported → CreateOrValidateOnQnn → qnn::utils::TwoDimensionTranspose (do QNN OP validation)
```
This change avoid heavy memcpy by using a dummy tensor when only shape validation is required.

### Motivation and Context

Below tested on a tensor with shape [2304, 700]

| Function                   | TwoDimensionTranspose_1 | TwoDimensionTranspose_2 | TwoDimensionTranspose_3 | SessionCreationTime
|---------------------------|--------------------------|--------------------------|--------------------------|--------------------------|
| original                  | 88.39 ms             | 57.80 ms            | 53.09 m             |               9.41871 s |
| avoid 2 memcpy                 | 51.52 ms              |12.00 m               | 8.05 ms                  | 9.05975 s               |

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/qnn_utils.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_utils.cc
@@ -1255,13 +1255,13 @@ static Status TransposeDataRank5(const TensorShape& input_shape,
   return Status::OK();
 }
 
-// Use use_dummy_tensor flag when performing only QNN op validation and no real tensor data is required.
+// Use skip_output_data_copy flag when performing only QNN op validation and no real tensor data is required.
 Status TwoDimensionTranspose(const QnnModelWrapper& qnn_model_wrapper,
                              std::vector<uint32_t>& data_shape,
                              const onnx::TensorProto& initializer,
                              std::vector<uint8_t>& transposed_data,
                              const logging::Logger& logger,
-                             bool use_dummy_tensor) {
+                             bool skip_output_data_copy) {
   ORT_RETURN_IF_NOT(data_shape.size() == 2, "Expected shape of rank 2");
 
   std::array<size_t, 2> perm = {1, 0};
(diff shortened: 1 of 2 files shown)


chunghow-qti:
Hi @edgchen1 , could you please help review this PR? Thanks!

onnxruntime/core/providers/qnn/builder/qnn_utils.cc
```diff
@@ -1255,10 +1255,13 @@ static Status TransposeDataRank5(const TensorShape& input_shape,
   return Status::OK();
 }
 
+// Use use_dummy_tensor flag when performing only QNN op validation and no real tensor data is required.
```
edgchen1: this `use_dummy_tensor` flag feels a bit unintuitive to me. what is the "dummy tensor"?

if you want to skip the population of `transposed_data`, perhaps make it optional instead? e.g., `std::vector<uint8_t>* transposed_data` and skip the population if it is nullptr.

kuanyul-qti: Hi, the dummy tensor here refers to a tensor that does not contain actual data, i.e., `transposed_data` at Line 1277, and we skip the population for the `transposed_data` in this case.


> if you want to skip the population of `transposed_data`, perhaps make it optional instead? e.g., `std::vector<uint8_t>* transposed_data` and skip the population if it is nullptr.

Regarding this, when determining whether to skip the population, the `transposed_data` is always nullptr because the memory copy for `transposed_data` occurs at L1299.

edgchen1: I see that this function still resizes and zeros `transposed_data` if `use_dummy_tensor` is true. perhaps that is important. I'd think that would have some overhead though.

maybe a name like `skip_output_data_copy` would be clearer than `use_dummy_tensor`.

kuanyul-qti: Yes. The function still resizes `transposed_data` because we need to know its size in the following stage, which is the same as in the original code. What we do in the PR is to avoid data coping and give it an initial value.


> maybe a name like `skip_output_data_copy` would be clearer than `use_dummy_tensor`.

Done.

edgchen1 (approved):

PR #26629, opened 2025-11-21
Clear cuda error on unsupported CudaMemPool test

yuslepukhin:
### Description

CudaMemPool test checks if it is supported in a given environment.
We need to clear the error not to affect subsequent tests.

### Motivation and Context

Potential test failure.

### DIFF
--- a/onnxruntime/test/providers/cuda/cuda_mempool_arena_test.cc
+++ b/onnxruntime/test/providers/cuda/cuda_mempool_arena_test.cc
@@ -89,7 +89,9 @@ static ::cudaStream_t NewCudaStream() {
 }
 
 static void DestroyCudaStream(::cudaStream_t s) {
-  if (s) EXPECT_EQ(cudaSuccess, ::cudaStreamDestroy(s));
+  if (s) {
+    EXPECT_EQ(cudaSuccess, ::cudaStreamDestroy(s));
+  }
 }
 
 static void TouchDevice(void* p, size_t bytes, ::cudaStream_t s, unsigned char value = 0xAB) {


onnxruntime/test/providers/cuda/cuda_mempool_arena_test.cc
```diff
@@ -65,9 +73,10 @@ static bool IsCudaMemPoolSupported() {
   cudaMemPool_t pool;
   auto cuda_error = cudaMemPoolCreate(&pool, &props);
   if (cuda_error != cudaSuccess) {
+    ClearCudaError();
```
hariharans29, line 76: Should we peek at the error before trying to create the mempool to check if there were issues from running other tests **before** trying to attempt mem pool creation and subsequently resetting the error if mempool creation failed?

onnxruntime/test/providers/cuda/cuda_mempool_arena_test.cc
```diff
@@ -22,10 +22,17 @@ namespace test {
...
+// we want to clear the error if when supported check fails.
+void ClearCudaError() {
+  ORT_IGNORE_RETURN_VALUE(::cudaGetLastError());
+}
+
 static bool IsCudaMemPoolSupported() {
   int ort_cuda_rt_version = 0;
   cudaError_t cuda_status = cudaRuntimeGetVersion(&ort_cuda_rt_version);
   if (cuda_status != cudaSuccess) {
+    ClearCudaError();
```
hariharans29, line 35: Just curious - do we get cuda errors just by querying runtime and driver versions on some machines ?

tianleiwu: https://docs.nvidia.com/cuda/archive/13.0.0/cuda-runtime-api/group__CUDART____VERSION.html

This function automatically returns if driverVersion is NULL.


Note:
Note that this function may also return error codes from previous, asynchronous launches.

Note that this function may also return [cudaErrorInitializationError](https://docs.nvidia.com/cuda/archive/13.0.0/cuda-runtime-api/group__CUDART__TYPES.html#group__CUDART__TYPES_1gg3f51e3575c2178246db0a94a430e0038ce7993a88ecf2c57b8102d55d997a18c), [cudaErrorInsufficientDriver](https://docs.nvidia.com/cuda/archive/13.0.0/cuda-runtime-api/group__CUDART__TYPES.html#group__CUDART__TYPES_1gg3f51e3575c2178246db0a94a430e0038f5e52d1774934b77ba55d2aa2c063067) or [cudaErrorNoDevice](https://docs.nvidia.com/cuda/archive/13.0.0/cuda-runtime-api/group__CUDART__TYPES.html#group__CUDART__TYPES_1gg3f51e3575c2178246db0a94a430e0038e942e4cbbd2bef6e92e293253f055613) if this call tries to initialize internal CUDA RT state.

Note that as specified by [cudaStreamAddCallback](https://docs.nvidia.com/cuda/archive/13.0.0/cuda-runtime-api/group__CUDART__STREAM.html#group__CUDART__STREAM_1g74aa9f4b1c2f12d994bf13876a5a2498) no CUDA function may be called from callback. [cudaErrorNotPermitted](https://docs.nvidia.com/cuda/archive/13.0.0/cuda-runtime-api/group__CUDART__TYPES.html#group__CUDART__TYPES_1gg3f51e3575c2178246db0a94a430e003867b6095ab719b21659a400b553963eb3) may, but is not guaranteed to, be returned as a diagnostic in such case.

onnxruntime/test/providers/cuda/cuda_mempool_arena_test.cc
yuslepukhin: The rationale here is as follows: 1) Typical CUDA code makes use of a macro that clears the error automatically. 2) the issue did not occur before my check-in where I use cuda RT directly, and the subsequent test started failing then.

I may still be wrong, but I think this is what is happening now.

fs-eire:
The pipelines run now, but there are failures in the build

fs-eire:
All tests are passing now.

PR #26630, opened 2025-11-21
[QNN EP] Reshape: relax check on allowzero when concrete shape without 0

qti-yuduo:
For QNN Reshape op. 

Relax check to support some cases of `allowzero=1` when we have concrete static shape without 0. Previous check is too limited.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/reshape_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/reshape_op_builder.cc
@@ -38,7 +38,7 @@ Status ReshapeOpBuilder::ProcessInputs(QnnModelWrapper& qnn_model_wrapper,
     NodeAttrHelper node_helper(node_unit);
     auto allowzero = node_helper.Get("allowzero", static_cast<int64_t>(0));
 
-    // Only reject allowzero=1 if dynamic shape or the shape actually contains zeros
+    // Only reject allowzero=1 if the shape input is not constant or it actually contains zeros
     if (0 != allowzero) {
       const auto& inputs = node_unit.Inputs();
       const auto& initializer_tensors = qnn_model_wrapper.GetInitializerTensors();


onnxruntime/core/providers/qnn/builder/opbuilder/reshape_op_builder.cc
```diff
@@ -37,8 +37,33 @@ Status ReshapeOpBuilder::ProcessInputs(QnnModelWrapper& qnn_model_wrapper,
   if (do_op_validation) {
     NodeAttrHelper node_helper(node_unit);
     auto allowzero = node_helper.Get("allowzero", static_cast<int64_t>(0));
+
+    // Only reject allowzero=1 if dynamic shape or the shape actually contains zeros
```
edgchen1: what does "dynamic shape" mean here?

qti-yuduo: unknown at compile/lowering time. Feel free to recommend better wording.

edgchen1: oh, does it mean that the shape input is not a constant? might be clearer to put it that way. I thought that "dynamic shape" meant a tensor shape such as `[batch_dim, 42]`.

edgchen1: I see. would you mind explaining it a bit more in the current context? I'm not sure how dynamic shapes would apply here if the QNN EP does not support them.

with my current understanding, suggested wording would be:

```suggestion
    // Only reject allowzero=1 if the shape input is not constant or it actually contains zeros
```

onnxruntime/core/providers/qnn/builder/opbuilder/reshape_op_builder.cc
```diff
@@ -37,8 +37,33 @@ Status ReshapeOpBuilder::ProcessInputs(QnnModelWrapper& qnn_model_wrapper,
...
+      if (shape_tensor_iter == initializer_tensors.end()) {
+        return ORT_MAKE_STATUS(ONNXRUNTIME, FAIL,
+                               "QNN Reshape requires a constant shape input");
+      }
+
+      // Check if the constant shape contains any zeros
+      const auto* shape_tensor = shape_tensor_iter->second;
+      std::vector<uint8_t> unpacked_tensor;
+      ORT_RETURN_IF_ERROR(qnn_model_wrapper.UnpackInitializerData(*shape_tensor, unpacked_tensor));
+
```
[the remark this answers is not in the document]
edgchen1: validation of ONNX type constraints should already have been done by this point, so I don't think additional type checking is necessary here.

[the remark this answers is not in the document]
edgchen1: `shape_data` is only dereferenced in the for loop which checks `shape_size`.

onnxruntime/core/providers/qnn/builder/opbuilder/reshape_op_builder.cc
```diff
@@ -37,8 +37,33 @@ Status ReshapeOpBuilder::ProcessInputs(QnnModelWrapper& qnn_model_wrapper,
...
+      const int64_t* shape_data = reinterpret_cast<const int64_t*>(unpacked_tensor.data());
+      size_t shape_size = unpacked_tensor.size() / sizeof(int64_t);
+
+      for (size_t i = 0; i < shape_size; ++i) {
+        if (shape_data[i] == 0) {
+          return ORT_MAKE_STATUS(ONNXRUNTIME, FAIL,
+                                 "QNN Reshape does not support shapes with zero dimensions. "
+                                 "The 'allowzero' attribute is not supported by QNN.");
```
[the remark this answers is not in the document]
edgchen1: FWIW, I think the original message is fine.

PR #26633, opened 2025-11-21
[js] fix NPM package dependency `glob`

fs-eire:
### Description

see https://github.com/microsoft/onnxruntime/security/dependabot/468


fs-eire:
replaced by #26632

PR #26634, opened 2025-11-22
[QNN EP] Add support for building aarch64 wheels

qti-jkilpatrick:
### Description

This change adds support for building a wheel with QNN EP support targeting ARM64 on `manylinux_2_34`. Specifically:

* Include skel libs and other files necessary to use HTP and GPU on Linux.
* Provide a sensible default `ADSP_LIBRARY_PATH` if one hasn't been set by the user.

### Motivation and Context

This is in support of Qualcomm IoT devices such as the QCS6490 running Ubuntu.

Please note that GLIBC usage in QNN dictates the use of a newer `manylinux` image (`2_34` instead of `2.28`).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -304,7 +304,11 @@ try:
                 logger.info("repairing %s for manylinux1", file)
                 auditwheel_cmd = ["auditwheel", "-v", "repair", "-w", self.dist_dir, file]
                 for i in (
-                    cuda_dependencies + migraphx_dependencies + tensorrt_dependencies + cann_dependencies + qnn_dependencies
+                    cuda_dependencies
+                    + migraphx_dependencies
+                    + tensorrt_dependencies
+                    + cann_dependencies
+                    + qnn_dependencies
                 ):
                     auditwheel_cmd += ["--exclude", i]
                 logger.info("Running %s", " ".join([shlex.quote(arg) for arg in auditwheel_cmd]))


cmake/CMakeLists.txt
```diff
@@ -897,6 +897,9 @@ if (onnxruntime_USE_QNN OR onnxruntime_USE_QNN_INTERFACE)
                                                "${onnxruntime_QNN_HOME}/lib/hexagon-v81/unsigned/libQnnHtpV81Skel.so"
                                                "${onnxruntime_QNN_HOME}/lib/hexagon-v81/unsigned/libqnnhtpv81.cat")
         list(APPEND QNN_LIB_FILES ${EXTRA_HTP_LIB})
+      elseif(${QNN_ARCH_ABI} STREQUAL "aarch64-ubuntu-gcc9.4")
+        file(GLOB EXTRA_HTP_LIB LIST_DIRECTORIES false "${onnxruntime_QNN_HOME}/lib/hexagon-v68/unsigned/libQnnHtpV68Skel.so")
+        list(APPEND QNN_LIB_FILES ${EXTRA_HTP_LIB})
```
qti-jkilpatrick: v68 is the only HTP arch supported from Ubuntu.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -293,6 +293,25 @@ void QnnBackendManager::SetQnnBackendType(uint32_t backend_id) {
 }
 
 Status QnnBackendManager::LoadBackend() {
+#if defined(__aarch64__) && defined(__linux__)
+  // QNN requires ADSP_LIBRARY_PATH to be set in order to find skel libs on Linux
```
qti-jkilpatrick, line 297: Just to underscore this comment: the strategy used to find HTP skel libraries is different on Linux than it is on Windows. On the latter, it appears to be inspired by the dll search strategy in that the executable's directory is searched. Not so on Linux where `ADSP_LIBRARY_PATH` *must* be set.

setup.py
```diff
@@ -435,6 +446,7 @@ def finalize_options(self):
         "QnnCpu.dll",
         "QnnGpu.dll",
         "QnnHtp.dll",
+        "QnnIr.dll",
```
qti-jkilpatrick, line 452: This should have been included previously; I suspect that dumping `.dlc` files from Python on Windows is broken.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
qti-jkilpatrick: This has not been an issue so far because we've only been using emulated HTP on [x86_64] Linux. Therefore, there has been no need to locate the skel files to send them over to the HTP.

cmake/CMakeLists.txt
edgchen1: nit: don't need to glob for an exact file path.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -293,6 +293,25 @@ void QnnBackendManager::SetQnnBackendType(uint32_t backend_id) {
...
+  static std::once_flag set_adsp_path_once;
+
+  std::call_once(set_adsp_path_once, []() {
+    constexpr std::string_view kAdspLibraryPathEnvVar{"ADSP_LIBRARY_PATH"};
+    const char* existingPath = getenv(kAdspLibraryPathEnvVar.data());
```
edgchen1: nit: variable naming - `existing_path`. `kAdspLibraryPathEnvVar` is fine.

https://google.github.io/styleguide/cppguide.html#Variable_Names

https://google.github.io/styleguide/cppguide.html#Constant_Names

cmake/CMakeLists.txt
qti-jkilpatrick: Would you prefer that I follow the existing pattern or rewrite all of them to not glob for an exact path?

edgchen1: actually, I realize that `file(GLOB ...)` would also allow for the specified file to not exist. if we do expect it to exist, I'll stand by my original suggestion.

updating the existing code is optional but would be nice.

PR #26638, opened 2025-11-23
webgpu: add FormatTransform kernel and tests

jchen10:
- add the WebGPU FormatTransform kernel, headers, and WGSL template, supporting Plain <-> nChw4c/ABcd16a4b conversions
- register the FormatTransform schema in the internal NHWC domain with padding-aware shape inference
- hook the kernel into the WebGPU execution provider and add WebGPU tests covering both blocked formats, padding cases, and round trips


jchen10:
@qjia7 @fs-eire @guschmue PTAL

PR #26644, opened 2025-11-24
[QNN EP] Handle 0-dim tensor for Concat.

minfhong-qti:
### Description

Define a util function `IsNullNodeUnitIODef` to check whether given tensor has 0-dim in shape. To restrict the impact, current implementation only skip constructing such tensor for Concat.

### Motivation and Context

There may exsit tensor with 0-dim in shape, especially for Concat's inputs. Such null inputs often exist in models saved from PyTorch. Since QNN could not execute empty tensors, they must be explicitly handled.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/base_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/base_op_builder.cc
@@ -13,6 +13,34 @@ bool IsOptionalNodeUnitIODef(const NodeUnitIODef& node_io_def) {
   const NodeArg& arg = node_io_def.node_arg;
   return !arg.Exists() || arg.Name().empty();
 }
+
+// Function to check whether we should skip processing null input which has 0 dim in shape.
+// Such null inputs often exist in models saved from PyTorch, especially for Concat.
+bool DoesConcatInputShapeContainZero(QnnModelWrapper& qnn_model_wrapper,
+                                     const NodeUnit& node_unit,
+                                     const NodeUnitIODef& node_io_def,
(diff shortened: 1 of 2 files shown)


onnxruntime/core/providers/qnn/builder/opbuilder/base_op_builder.cc
```diff
@@ -13,6 +13,33 @@ bool IsOptionalNodeUnitIODef(const NodeUnitIODef& node_io_def) {
   const NodeArg& arg = node_io_def.node_arg;
   return !arg.Exists() || arg.Name().empty();
 }
+
+// Function to check whether we should skip processing null input which has 0 dim in shape.
+// Such null inputs often exist in models saved from PyTorch, especially for Concat.
+bool IsNullNodeUnitIODef(QnnModelWrapper& qnn_model_wrapper,
```
edgchen1: it's not obvious to me what "null node unit I/O def" means. looking at the code, it is checking if any dimensions are zero. can the name be made clearer?

for example, here's a shape-related helper which does the same thing:
https://github.com/microsoft/onnxruntime/blob/db6d83b615d06eb8e4c777385085006a13f1322e/onnxruntime/core/providers/coreml/shape_utils.cc#L56-L58

also, the check for `node_unit.OpType() != "Concat"` feels out of place in a function with a more general name.

minfhong-qti: In the latest commit, I've renamed the function to `DoesConcatInputShapeContainZero`. Please kindly review again.

edgchen1 (approved):

PR #26645, opened 2025-11-24
add LogEvaluationStart for ReplayGraph

xieofxie:
### Description

add LogEvaluationStart for ReplayGraph to match LogEvaluationStop

### Motivation and Context

So by using ETW, could capture run time correctly


onnxruntime/core/session/inference_session.cc
xieofxie, line 2956: not sure why not in function entry to cover both scenarios, https://github.com/microsoft/onnxruntime/commit/60acfd3dd8dabe082f96fa49fab7278a6fd33eb4#diff-3e2227e1225091e8b74c02688e23b21630d1393dd395e15966558901538dd2c7 kept as it is

hanbitmyths (approved):

PR #26652, opened 2025-11-25
Expose start profiling API 

xiaofeihan1:
### Description

Add start profiling API in ORT. With this, we can profile for a time span. Based on this, we have [another genAI PR](https://github.com/microsoft/onnxruntime-genai/pull/1898) to support start/end profiling in genai.

ps: When `enable_profiling` of session option is true and `start_profiling` is called, we always profiling from the beginning of the session

enable_profiling | start_profiling | end_profiling | Result
-- | -- | -- | --
0 | 0 | 0 | no trace
0 | 0 | 1 | no trace
0 | 1 | 0 | trace from start profiling to the end of the session
0 | 1 | 1 | trace from start profiling to the end profiling
1 | 0 | 0 | trace the entire session
1 | 0 | 1 | trace from the begginng of the session to the end profiling
1 | 1 | 0 | trace the entire session
1 | 1 | 1 | trace from the beggining of the session to the end profiling

### Motivation and Context
Previously, we only enable profiling from the beginning of session. There is no way to start profiling in the middle of a session.

With this PR, we can profiling any time span.

### DIFF
--- a/onnxruntime/core/session/onnxruntime_c_api.cc
+++ b/onnxruntime/core/session/onnxruntime_c_api.cc
@@ -3721,10 +3721,10 @@ ORT_API_STATUS_IMPL(OrtApis::SessionGetMemoryInfoForOutputs, _In_ const OrtSessi
   API_IMPL_END
 }
 
-ORT_API_STATUS_IMPL(OrtApis::SessionStartProfiling, _In_ OrtSession* sess) {
+ORT_API_STATUS_IMPL(OrtApis::SessionStartProfiling, _In_ OrtSession* sess, _In_ const char* file_prefix) {
   API_IMPL_BEGIN
   auto session = reinterpret_cast<::onnxruntime::InferenceSession*>(sess);
-  session->StartProfiling("onnxruntime_profile_");
+  session->StartProfiling(file_prefix);
   return nullptr;
   API_IMPL_END
 }
--- a/include/onnxruntime/core/session/onnxruntime_c_api.h
+++ b/include/onnxruntime/core/session/onnxruntime_c_api.h
@@ -6612,11 +6612,12 @@ struct OrtApi {
   /** \brief Start profiling for this session
    *
    * \param[in] session
+   * \param[in] file_prefix
    *
    * \snippet{doc} snippets.dox OrtStatus Return Value
    * \since Version 1.24
    */
-  ORT_API2_STATUS(SessionStartProfiling, _In_ OrtSession* session);
+  ORT_API2_STATUS(SessionStartProfiling, _In_ OrtSession* session, _In_ const char* file_prefix);
 };
 
 /*
--- a/include/onnxruntime/core/session/onnxruntime_cxx_api.h
+++ b/include/onnxruntime/core/session/onnxruntime_cxx_api.h
@@ -1723,7 +1723,7 @@ struct SessionImpl : ConstSessionImpl<T> {
    *
    * Wraps OrtApi::SessionStartProfiling
    */
-  void StartProfiling();  ///< Wraps OrtApi::SessionStartProfiling
+  void StartProfiling(const std::string& file_prefix);  ///< Wraps OrtApi::SessionStartProfiling
 };
 
 }  // namespace detail
--- a/include/onnxruntime/core/session/onnxruntime_cxx_inline.h
+++ b/include/onnxruntime/core/session/onnxruntime_cxx_inline.h
@@ -1778,8 +1778,8 @@ inline void SessionImpl<T>::FinalizeModelEditorSession(const Model& model, const
 #endif  // #if !defined(ORT_MINIMAL_BUILD)
 
 template <typename T>
-inline void SessionImpl<T>::StartProfiling() {
-  ThrowOnError(GetApi().SessionStartProfiling(this->p_));
+inline void SessionImpl<T>::StartProfiling(const std::string& file_prefix) {
+  ThrowOnError(GetApi().SessionStartProfiling(this->p_, file_prefix.c_str()));
 }
 
 }  // namespace detail
--- a/onnxruntime/core/session/ort_apis.h
+++ b/onnxruntime/core/session/ort_apis.h
@@ -754,6 +754,6 @@ ORT_API_STATUS_IMPL(CopyTensors, _In_ const OrtEnv* env,
 
 ORT_API_STATUS_IMPL(KernelInfo_GetConfigEntries, _In_ const OrtKernelInfo* info, _Outptr_ OrtKeyValuePairs** out);
 
-ORT_API_STATUS_IMPL(SessionStartProfiling, _In_ OrtSession* sess);
+ORT_API_STATUS_IMPL(SessionStartProfiling, _In_ OrtSession* sess, _In_ const char* file_prefix);
 
 }  // namespace OrtApis
--- a/onnxruntime/python/onnxruntime_inference_collection.py
+++ b/onnxruntime/python/onnxruntime_inference_collection.py
@@ -366,14 +366,14 @@ class Session:
                 return invoke(self._sess, output_names, input_dict_ort_values, run_options)
             raise
 
-    def start_profiling(self):
+    def start_profiling(self, file_prefix="onnxruntime_profile_"):
         """
         Start profiling.
 
         Note that if the option :meth:`onnxruntime.SessionOptions.enable_profiling` is true,
         profiling will always start at the beginning of this session.
         """
-        self._sess.start_profiling()
+        self._sess.start_profiling(file_prefix)
 
     def end_profiling(self):
         """
--- a/onnxruntime/python/onnxruntime_pybind_state.cc
+++ b/onnxruntime/python/onnxruntime_pybind_state.cc
@@ -2722,8 +2722,8 @@ including arg name, arg type (contains both type and shape).)pbdoc")
         py::gil_scoped_release release;
         OrtPybindThrowIfError(sess->GetSessionHandle()->Run(run_options, feed_names, feeds, fetch_names, &fetches, &fetch_devices));
       })
-      .def("start_profiling", [](const PyInferenceSession* sess) -> void {
-        sess->GetSessionHandle()->StartProfiling("onnxruntime_profile_");
+      .def("start_profiling", [](const PyInferenceSession* sess, const std::string& file_prefix) -> void {
+        sess->GetSessionHandle()->StartProfiling(file_prefix);
       })
       .def("end_profiling", [](const PyInferenceSession* sess) -> std::string {
         return sess->GetSessionHandle()->EndProfiling();


xiaofeihan1:
Make this as draft because we might need other changes to support GenAI support start/end profiling.

qjia7:
@yuslepukhin @guschmue @fs-eire
We’re eager to have this feature because we’ve observed that when the context is long, profiling events exceed the maximum supported limit if profiling is enabled from the start of the session. As a result, no profiling data is generated, which makes debugging difficult. Without profiling data, it’s hard to investigate why generation becomes very slow when the total sequence length exceeds a certain threshold.
We’d like to enhance the current profiling interfaces in ORT and ORT-GenAI. Please take a look at this PR and guide us in the right direction. Thanks!

onnxruntime/core/session/inference_session.cc
```diff
@@ -3486,6 +3486,10 @@ common::Status InferenceSession::Run(IOBinding& io_binding) {
 
 template <typename T>
 void InferenceSession::StartProfiling(const std::basic_string<T>& file_prefix) {
+  if (session_profiler_.IsEnabled()) {
+    LOGS(*session_logger_, WARNING) << "Profiler is already running.";
+    return;
+  }
```
feich-ms, line 3492: Do you consider the situation that users call session.StartProfling multiple times without calling session.EndProfiling, like below, will that cause any errors or crashes?
```
sess.StartProfiling()
xxx
xxx
sess.StartProfiling()
xxx
xxx
sess.EndProfiling()
```

xiaofeihan1: I believe it doesn't cause any error and it will profile the span between first StartProfiling and endProfiling. Because I use the following code and the second start profiling is ignored.
```
    LOGS(*session_logger_, WARNING) << "Profiler is already running.";
    return;
  }
```

onnxruntime/core/session/onnxruntime_c_api.cc
```diff
@@ -3721,6 +3721,14 @@ ORT_API_STATUS_IMPL(OrtApis::SessionGetMemoryInfoForOutputs, _In_ const OrtSessi
   API_IMPL_END
 }
 
+ORT_API_STATUS_IMPL(OrtApis::SessionStartProfiling, _In_ OrtSession* sess) {
+  API_IMPL_BEGIN
+  auto session = reinterpret_cast<::onnxruntime::InferenceSession*>(sess);
+  session->StartProfiling("onnxruntime_profile_");
```
feich-ms: Is the "onnxruntime_profile_" the value by default? Is it possible that the value need to be specified by users? Looks like not necessary.

yuslepukhin: This would be a good feature as the disk may be polluted by many similar files.

yuslepukhin: Are multi-threading issues considered? Typically, there is a single thread that invokes Run() and no other calls except another Run are called.

xiaofeihan1: Thanks for the suggestion! I have changed it to the following shape. Only when developer call `start_profiling()`, we will use the default value.
```
  def start_profiling(self, file_prefix="onnxruntime_profile_"):
```

xiaofeihan1: Hi @yuslepukhin I didn’t consider the multi-threaded scenario. We previously exposed [end_profiling API](https://onnxruntime.ai/docs/api/python/api_summary.html#onnxruntime.InferenceSession.end_profiling). So I added start_profiling to keep it consistent.
 I think you are right, calling session.run in one thread while calling session.start/end_profiling in another thread may lead to unexpected issues. E.g. If thread 1 call `session.startprofiling` and just marked `enabled_` as true but not intiliazed profiler yet. and thread 2 call `session. run` which will try to record data because `session_profiler_.IsEnabled()` is true. This might cause unexpected issue.
Do you think the following pattern is a good way to address this? But it might cause some perf overhead when profiling is turned on,
```
Status InferenceSession::Run(...) {
  TimePoint tp = std::chrono::high_resolution_clock::now();
  
  // Add double-Checked Locking to minimize perf overhead
  if (session_profiler_.IsEnabled()) { // first check
    std::lock_guard<std::mutex> lock(profiler_mutex_);
    if (session_profiler_.IsEnabled()) { // second check
    }
  }
  ...
}

template <typename T>
void InferenceSession::StartProfiling(const std::basic_string<T>& file_prefix) {
  // Add
  std::lock_guard<std::mutex> lock(profiler_mutex_);

    LOGS(*session_logger_, WARNING) << "Profiler is already running.";
    return;
  }
  ...
}

```

yuslepukhin: > Hi @yuslepukhin I didn’t consider the multi-threaded scenario. We previously exposed [end_profiling API]

We certainly do not want to do it Run(), besides, is it not what is taking place now already when profiling is enabled?

xiaofeihan1: Our intention is to allow developers to start profiling at any point during the session. Could we document that this API is not thread-safe in API doc? 

(Note: Enabling profiling via session options starts profiling at the beginning of the session. This may fail for large contexts if the number of profiling events exceeds the maximum supported limit.)

yuslepukhin: > Our intention is to allow developers to start profiling at any point during the session. Could we document that this API is not thread-safe in API doc?
> 
How would anyone use it w/o being thread-safe?

xiaofeihan1: Hi @yuslepukhin  I might be misunderstanding something, so let me clarify a few points to make sure we’re on the same page:
1. **InferenceSession is designed as thread-safe**.
    This means developers can safely invoke its member functions from multiple threads. For example:

    1.1 Multiple threads calling session.Run() concurrently.

    1.2 Some threads calling session.Run() while others call session.GetOutputs().

    All of these are supported.
2. **However, StartProfiling and EndProfiling are not thread-safe in the current implementation.**
    This can lead to two categories of issues:

    2.1 **Multiple threads calling StartProfiling / EndProfiling simultaneously.**
    This is straightforward to fix by protecting these APIs with a dedicated mutex_.

    2.2 **Race conditions between Run() and profiling control APIs (StartProfiling / EndProfiling).**
    Run() checks session_profiler_.IsEnabled(), while the profiling-control APIs modify this state.
This can cause several problems:

    a. session_profiler_.Start() may be invoked before the profiler is fully initialized.

    b. Or session_profiler_.Start() might run even though profiling has already been disabled by EndProfiling.

To address both of these issues listed in 2, **my proposal is to introduce a dedicated `profiler_mutex_`**.

This ensures safe coordination between Run() and profiling state changes.
Importantly, this approach **adds zero overhead to the hot Run() path** when developers do not call profiling APIs at runtime. Do you have any suggestions on this? 🤔
```
// Current code
InferenceSession::Run{
  }
  ....
}

void InferenceSession::StartProfiling(const std::basic_string<T>& file_prefix) {
  session_profiler_.StartProfiling(ss.str());
}

```

```
// My proposoal
InferenceSession::Run{
   // Add double-Checked Locking to minimize perf overhead
    if (session_profiler_.IsEnabled()) { // first check
      std::lock_guard<std::mutex> lock(profiler_mutex_);
      if (session_profiler_.IsEnabled()) { // second check
      }
    }
  ....
}

void InferenceSession::StartProfiling(const std::basic_string<T>& file_prefix) {
  std::lock_guard<std::mutex> lock(profiler_mutex_);
  session_profiler_.StartProfiling(ss.str());
}
```

feich-ms (approved):

yuslepukhin:
Here are my thoughts on the API that you have proposed:
The API's purpose seems unclear to me. The statement: “to be invoked at any time” taken at its face value would mean the following:

- Any time may mean in the middle of the Run() call which would mean getting partial profiling data.
- Data from concurrent Run() calls would intersperse.
- The profiling data would not be reliable due to a potentially variable load on the system during concurrent Run() calls.
- Invocation would have to be from a different thread.
- All the multi-threaded issues we talked about due to all the above.

Today profiling is enabled by means of session options and, therefore, applies to all the sessions that are instantiated with it. Here are the reasons for the current design:

ONNXRuntime is designed to be stateless and lock free at the time of the Run() calls due the performance reasons. To that end:

- We do not employ locks at any time during Run() calls. There is a onetime initialization exception with CUDA EP. 
- All the required optimizations and pre-modifications are done at session loading time. We then make SessionState object constant. 
- All the kernels are stateless.
- This enables us to achieve latencies that are in the range of 300usec for some rather big models.

The profiling feature is a debugging feature not designed to be ran in production environments, and certainly, inserting some locks in the Run() call path of execution would simply not be acceptable.

It would be more reasonable to restate the aim for your API as follows: enable profiling for certain Run() calls at user discretion.

To that end, I suggest you design something along the lines of RunOptions{} which is a structure that is employed by individual Run() calls. In this case the profiling switch would be per Run() call. Furthermore, RunOptions{} may contain profiling data OR a pointer to a profiling data for that particular call where Run() would accumulate the information and no multi-threaded issues would need to be considered.

You would also need to add new APIs (both C and C++, possibly python) to enable this feature.

xiaofeihan1:
> Here are my thoughts on the API that you have proposed: The API's purpose seems unclear to me. The statement: “to be invoked at any time” taken at its face value ...

Hi @yuslepukhin, thanks for your detailed explanation! I think there may be a misunderstanding, so I’d like to first clarify our requirements, then raise a question we’ve had for some time, and finally outline the solution we are leaning toward.
# 1. Requirements
Our scenario targets LLM models. Our goal is to allow developers to start profiling after some tokens have already been generated, rather than always profiling from the beginning, as `enable_profiling` does. This is because profiling the entire session can fail when the number of generated tokens is large.

As you know, LLM inference involves calling `session.run()` repeatedly until the final token is generated. For example, in the following code, we want to profile only between the 290th and 300th calls to session.run(), which corresponds to collecting performance data for tokens generated in that range.

```
  while last_token not in self.eos and seqlen < max_tokens:
    if seqlen == 290:
        self.sess.start_profiling()
        print("started profiling")
    if seqlen == 300:
        self.sess.end_profiling()
        print("ended profiling")

    feed["attention_mask"] = np.empty([1, seqlen], dtype=np.int64)
    feed["attention_mask"].fill(1)
    outputs = self.sess.run([])
    logits = outputs[self.logits_idx]
    self.update_kv_cache(outputs)
    ...
```
# 2. Question

ONNX Runtime has exposed the [`end_profiling` python API](https://onnxruntime.ai/docs/api/python/api_summary.html#onnxruntime.InferenceSession.end_profiling) to developers for several years. It can be called at any time and from any thread.

Does this mean that the same multithreading concerns you mentioned also apply to `end_profiling`? So far, we haven’t added any explicit locking mechanism around its usage.


# 3. Proposed Solution

Preferred solution: **Add a start_profiling API without introducing additional locking to match end_profiling API**  for the following two reasons.

1. Profiling is primarily a debugging feature, and `end_profiling` has already been exposed and used for many years without requiring explicit synchronization. We believe it should be acceptable and natural to introduce a corresponding `start_profiling` API with the same assumptions.

2. I think the run options approach you mentioned might be a bit complex. We also plan to add a start_profiling API in onnxruntime-genai later, but run_options is not exposed in onnxruntime-genai yet.

xiaofeihan1:
Close this PR because we will use another solution. See https://github.com/microsoft/onnxruntime/pull/26846

PR #26396, opened 2025-10-23
[QNN-EP] Enable verbose and artifacts saving in onnxruntime_provider_test.exe

qti-hungjuiw:
### Description

- The change allows users to better debug unit tests by adding the following environment variables:
    - `QNN_DUMP_ONNX`: Dump input onnx model
    - `QNN_DUMP_JSON`: Dump json qnn graph with provider_option `dump_json_qnn_graph`
    - `QNN_DUMP_DLC`: Dump dlc with provider_option `qnn_ir_backend_path`
    - `QNN_VERBOSE`: Use the log level `ORT_LOGGING_LEVEL_VERBOSE`
- Developers can use the environment variables above to save the artifacts of QNN-EP testcases to a directory named with `<TestSuite>_<TestName>`
    ```
        .
        ├── QnnCPUBackendTests_BatchNorm2D_fp32         # RunQnnModelTest
        │   ├── dumped_f32_model.onnx                   # float32 ONNX model
        │   ├── QNNExecutionProvider_QNN_XXXX_X_X.dlc
        │   └── QNNExecutionProvider_QNN_XXXX_X_X.json
        ├── QnnHTPBackendTests_BatchNorm_FP16           # TestFp16ModelAccuracy
        │   ├── dumped_f16_model.onnx                   # float16 ONNX model
        │   ├── dumped_f32_model.onnx                   # float32 ONNX model
        │   ├── QNNExecutionProvider_QNN_XXXX_X_X.dlc
        │   └── QNNExecutionProvider_QNN_XXXX_X_X.json
        └── QnnHTPBackendTests_BatchNorm2D_U8U8S32      # TestQDQModelAccuracy
            ├── dumped_f32_model.onnx                   # float32 ONNX model
            ├── dumped_qdq_model.onnx                   # QDQ ONNX model
            ├── QNNExecutionProvider_QNN_XXXX_X_X.dlc
            └── QNNExecutionProvider_QNN_XXXX_X_X.json

        # All artifact files are placed under the current working directory from which the test binary is invoked.
    ```

### Motivation and Context

- The Json qnn graph/dlc are helpful for backend to debug performance/accuracy issues
- By comparing the onnx and Json qnn graph/dlc, we can locate the issue about graph manipulation.

PR #26660, opened 2025-11-26
[QNN EP] Update tensor naming logic in DebugNodeInputsOutputs

yuhuchua-qti:
### Description
Introduced a new environment variable ORT_DEBUG_NODE_IO_PREPEND_EP_TO_FILE_NAME to control whether the Execution Provider (EP) type is prepended to tensor file names.
Updated MakeTensorFileName() and related logic to include the EP type in the format: `<EPType>_<TensorName>` when the variable is set to a non-zero value. This change helps avoid file name conflicts and improves clarity when analyzing outputs from multi-EP runs.

### Motivation and Context
Accuracy unit tests may run the same model multiple times using different EPs, such as ORT CPU and QNN HTP. Previously, dumped tensor files used identical names when tensor names matched across sessions, causing file overwrites.

### Example
When ORT_DEBUG_NODE_IO_PREPEND_EP_TO_FILE_NAME=1, dumped tensors include EP-specific prefixes:
```
CPU_input.tensorproto
CPU_input_token_0.tensorproto
CPU_output.tensorproto
QNN_constant.tensorproto
QNN_input.tensorproto
QNN_input_token_0.tensorproto
QNN_output.tensorproto
```

### DIFF
--- a/onnxruntime/core/framework/debug_node_inputs_outputs_utils.cc
+++ b/onnxruntime/core/framework/debug_node_inputs_outputs_utils.cc
@@ -99,8 +99,22 @@ struct TensorMetadata {
   std::string producer;
   std::string consumer;
   std::string device_type;
+  std::string ep_type;
   size_t step;
 };
+
+std::string GetCleanedEpType(const Node& node) {
+  std::string ep_type = node.GetExecutionProviderType();
+
+  // Remove "ExecutionProvider" suffix from ep type to reduce length.
+  const std::string suffix_to_remove = "ExecutionProvider";
+  size_t pos = ep_type.find(suffix_to_remove);
+  if (pos != std::string::npos) {
+    ep_type.erase(pos, suffix_to_remove.length());
+  }
+
+  return ep_type;
+}
 
 bool FilterNode(const NodeDumpOptions& dump_options, const Node& node) {
   auto match_pattern =
@@ -139,14 +153,18 @@ void DumpTensorToStdOut(const Tensor& tensor, const NodeDumpOptions& dump_option
   }
 }
 
-PathString MakeTensorFileName(const std::string& tensor_name, const NodeDumpOptions& dump_options) {
+PathString MakeTensorFileName(const TensorMetadata& tensor_metadata, const NodeDumpOptions& dump_options) {
   auto make_valid_name = [](std::string name) {
     std::replace_if(
         name.begin(), name.end(), [](char c) { return !std::isalnum(c); }, '_');
     return name;
   };
 
-  return path_utils::MakePathString(make_valid_name(tensor_name), dump_options.file_suffix, ".tensorproto");
+  if (dump_options.prepend_ep_to_file_name) {
+    return path_utils::MakePathString(make_valid_name(tensor_metadata.ep_type + "_" + tensor_metadata.name), dump_options.file_suffix, ".tensorproto");
+  } else {
+    return path_utils::MakePathString(make_valid_name(tensor_metadata.name), dump_options.file_suffix, ".tensorproto");
+  }
 }
 
 void DumpTensorToFile(const Tensor& tensor, const std::string& tensor_name, const std::filesystem::path& file_path) {
@@ -375,7 +393,7 @@ void DumpCpuTensor(
       break;
     }
     case NodeDumpOptions::DataDestination::TensorProtoFiles: {
-      const std::filesystem::path tensor_file = dump_options.output_dir / MakeTensorFileName(tensor_metadata.name, dump_options);
+      const std::filesystem::path tensor_file = dump_options.output_dir / MakeTensorFileName(tensor_metadata, dump_options);
       DumpTensorToFile(tensor, tensor_metadata.name, tensor_file);
       break;
     }
@@ -485,6 +503,8 @@ const NodeDumpOptions& NodeDumpOptionsFromEnvironmentVariables() {
                 debug_node_inputs_outputs_env_vars::kHalfOverflowThreshold, " shall be a positive integer <= ", kMaxHalfThreshold);
     opts.half_overflow_threshold = static_cast<float>(threshold);
 
+    opts.prepend_ep_to_file_name = ParseEnvironmentVariableWithDefault<bool>(env_vars::kPrependEpToFileName, false);
+
     if (ParseEnvironmentVariableWithDefault<bool>(env_vars::kAppendRankToFileName, false)) {
       std::string rank = Env::Default().GetEnvironmentVar("OMPI_COMM_WORLD_RANK");
       if (rank.empty()) {
@@ -582,6 +602,7 @@ void DumpNodeInputs(
               tensor_metadata.name = input_defs[i]->Name();
               tensor_metadata.step = dump_context.iteration;
               tensor_metadata.consumer = node.Name() + ":" + std::to_string(i);
+              tensor_metadata.ep_type = GetCleanedEpType(node);
 
               TensorStatisticsData tensor_statistics;
               DumpTensor(dump_options, *tensor, tensor_metadata, tensor_statistics, session_state);
@@ -678,6 +699,7 @@ void DumpNodeOutputs(
               tensor_metadata.name = output_defs[i]->Name();
               tensor_metadata.step = dump_context.iteration;
               tensor_metadata.producer = node.Name() + ":" + std::to_string(i);
+              tensor_metadata.ep_type = GetCleanedEpType(node);
 
               TensorStatisticsData tensor_statistics;
               DumpTensor(dump_options, *tensor, tensor_metadata, tensor_statistics, session_state);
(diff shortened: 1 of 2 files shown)


edgchen1:
I believe something similar was implemented in https://github.com/microsoft/onnxruntime/pull/26396. if additional functionality is needed, can the infrastructure from that PR be extended?

yuhuchua-qti:
Hi @edgchen1,
The functionality introduced in this PR is not specific to QNN-EP.
It provides the ability to dump input and output data (input, output, and expected output) from the common accuracy tests in `test_utils.cc`, which can benefit all ORT providers when debugging accuracy issues.
PR https://github.com/microsoft/onnxruntime/pull/26396, which is already merged, only updated QNN-EP-specific logic to dump QNN-specific materials such as DLC files.
If we were to add this functionality by modifying only QNN-EP-specific logic, we would likely need to re-implement the `RunAndVerifyOutputsWithEP()` function within QNN-EP, resulting in duplicated work.
For this reason, we chose to add the dumping functionality in the common accuracy test framework, so it can also benefit other ORT providers.
Thanks.

edgchen1:
I see, I missed the update to onnxruntime/test/util/test_utils.cc.

are you aware of the [debug node I/O dumping infrastructure](https://onnxruntime.ai/docs/build/inferencing.html#debugnodeinputsoutputs)? I'm wondering whether that would work, or whether it is worth having an additional input/output dumping mechanism.

yuhuchua-qti:
Hi @edgchen1 ,

I was not aware of the existing I/O dumping infrastructure earlier, but I have now tried this functionality and confirmed that it successfully dumps input and output data (including input, output, and expected output) from accuracy tests.
To better leverage this infrastructure in the current ORT, I have one discussion point:
 
**Update logic in debug_node_inputs_outputs_utils.cc**
In accuracy unit tests, the UT model may be executed two or three times with different EPs (e.g., ORT CPU and QNN HTP). Currently, the naming logic for dumped tensors is identical if the tensor names match across different inference sessions, which leads to file overwriting.
To avoid this, we propose updating the naming logic in the MakeTensorFileName() function indebug_node_inputs_outputs_utils.cc so that tensors from different EP runs are saved in distinct files.
Would this change be acceptable to you?
 
Thanks.

edgchen1:
> Hi @edgchen1 ,
> 
> I was not aware of the existing I/O dumping infrastructure earlier, but I have now tried this functionality and confirmed that it successfully dumps input and output data (including input, output, and expected output) from accuracy tests. To better leverage this infrastructure in the current ORT, I have one discussion point:
> 
> **Update logic in debug_node_inputs_outputs_utils.cc** In accuracy unit tests, the UT model may be executed two or three times with different EPs (e.g., ORT CPU and QNN HTP). Currently, the naming logic for dumped tensors is identical if the tensor names match across different inference sessions, which leads to file overwriting. To avoid this, we propose updating the naming logic in the MakeTensorFileName() function indebug_node_inputs_outputs_utils.cc so that tensors from different EP runs are saved in distinct files. Would this change be acceptable to you?
> 
> Thanks.

glad to know it may be reusable. sure, updating the file naming logic sounds reasonable to me.

yuhuchua-qti:
Hi @edgchen1,
I’ve updated the PR based on our previous discussion. 
Could you please review it again when you have a chance?
Thank you!

onnxruntime/core/framework/debug_node_inputs_outputs_utils.cc
```diff
@@ -139,14 +140,14 @@ void DumpTensorToStdOut(const Tensor& tensor, const NodeDumpOptions& dump_option
...
-PathString MakeTensorFileName(const std::string& tensor_name, const NodeDumpOptions& dump_options) {
+PathString MakeTensorFileName(const TensorMetadata& tensor_metadata, const NodeDumpOptions& dump_options) {
   auto make_valid_name = [](std::string name) {
     std::replace_if(
         name.begin(), name.end(), [](char c) { return !std::isalnum(c); }, '_');
     return name;
   };
 
-  return path_utils::MakePathString(make_valid_name(tensor_name), dump_options.file_suffix, ".tensorproto");
+  return path_utils::MakePathString(tensor_metadata.ep_type, "_", make_valid_name(tensor_metadata.name), dump_options.file_suffix, ".tensorproto");
```
edgchen1: nit: we can run `make_valid_name()` on more of the file name
```suggestion
  return path_utils::MakePathString(make_valid_name(tensor_metadata.ep_type + "_" + tensor_metadata.name), dump_options.file_suffix, ".tensorproto");
```

tianleiwu (approved):
LGTM

onnxruntime/core/framework/debug_node_inputs_outputs_utils.cc
tianleiwu: Add an environment variable like [this](https://github.com/CodeLinaro/onnxruntime/blob/98d16dd6db6d1943f9c84102927e87037a68114c/onnxruntime/core/framework/debug_node_inputs_outputs_utils.h#L56) to control whether we shall add ep type to filename.

Optionally, we can remove "ExecutionProvider" suffix from ep type to reduce length.

yuhuchua-qti: Hi @edgchen1,
I've updated it in latest PR.
Thanks.

yuhuchua-qti: Hi @tianleiwu,
I've added an environment variable ORT_DEBUG_NODE_IO_PREPEND_EP_TO_FILE_NAME to control whether the EP type is prepended to file names.
"ExecutionProvider" suffix is also removed from ep type.
Thanks.

yuslepukhin:
Please, address and/or comment on all of the Copilot comments.

onnxruntime/core/framework/debug_node_inputs_outputs_utils.cc
```diff
@@ -582,6 +602,7 @@ void DumpNodeInputs(
               tensor_metadata.name = input_defs[i]->Name();
               tensor_metadata.step = dump_context.iteration;
               tensor_metadata.consumer = node.Name() + ":" + std::to_string(i);
+              tensor_metadata.ep_type = GetCleanedEpType(node);
```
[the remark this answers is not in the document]
yuhuchua-qti: Thanks for the suggestion. In this case, the conditional check for prepend_ep_to_file_name is already encapsulated inside the MakeTensorFileName() function. The EP type is computed outside because the node object (needed to determine the EP type) isn't accessible from within that function.
To avoid adding additional parameters to MakeTensorFileName()—and to avoid propagating extra arguments through several call sites—I placed the EP-type computation at the current location, even though it may not always be used.
I agree this creates some minor unnecessary work when prepend_ep_to_file_name is false, but it helps keep the function interface cleaner and avoids broader changes to the call stack. If you think passing node (or EP type) down is preferable despite that, I’m open to discussing alternatives.

onnxruntime/core/framework/debug_node_inputs_outputs_utils.cc
```diff
@@ -678,6 +699,7 @@ void DumpNodeOutputs(
               tensor_metadata.name = output_defs[i]->Name();
               tensor_metadata.step = dump_context.iteration;
               tensor_metadata.producer = node.Name() + ":" + std::to_string(i);
+              tensor_metadata.ep_type = GetCleanedEpType(node);
```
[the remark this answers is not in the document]
yuhuchua-qti: Thanks for the suggestion. In this case, the conditional check for prepend_ep_to_file_name is already encapsulated inside the MakeTensorFileName() function. The EP type is computed outside because the node object (needed to determine the EP type) isn't accessible from within that function.
To avoid adding additional parameters to MakeTensorFileName()—and to avoid propagating extra arguments through several call sites—I placed the EP-type computation at the current location, even though it may not always be used.
I agree this creates some minor unnecessary work when prepend_ep_to_file_name is false, but it helps keep the function interface cleaner and avoids broader changes to the call stack. If you think passing node (or EP type) down is preferable despite that, I’m open to discussing alternatives.

onnxruntime/core/framework/debug_node_inputs_outputs_utils.h
```diff
@@ -117,6 +119,7 @@ struct NodeDumpOptions {
     SqliteDb
   } data_destination{DataDestination::StdOut};
 
+  bool prepend_ep_to_file_name{false};
   std::string file_suffix;
```
yuslepukhin, line 125: nit; separate by a comment or empty line #Resolved

tianleiwu (dismissed):

onnxruntime/core/framework/debug_node_inputs_outputs_utils.h
yuhuchua-qti: Hi @yuslepukhin,
I've updated it in latest PR.
Thanks.

tianleiwu (approved):

PR #26665, opened 2025-11-26
[QNN_EP] ConvTranspose not calculating "pad" if "output_shape" is given.

ankus-qti:
### Description
Given "output_shape" in ConvTranspose, appropriate "pad" should be calculated as per the formula mentioned in: https://onnx.ai/onnx/operators/onnx__ConvTranspose.html#convtranspose

### Motivation and Context
Current QNN_EP implementation does not handle this case and sets pad to default value (zeroes, no padding).

### DIFF
--- a/onnxruntime/test/providers/qnn/conv_test.cc
+++ b/onnxruntime/test/providers/qnn/conv_test.cc
@@ -30,9 +30,10 @@ static GetTestModelFn BuildF32ConvTestCase(const std::string& conv_op_type, cons
                                            const std::vector<int64_t>& dilations,
                                            std::optional<int64_t> group,
                                            const std::string& auto_pad = "NOTSET",
-                                           std::optional<OutputActivationInfo> output_activation = std::nullopt) {
+                                           std::optional<OutputActivationInfo> output_activation = std::nullopt,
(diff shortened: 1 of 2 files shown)


onnxruntime/test/providers/qnn/conv_test.cc
```diff
@@ -30,9 +30,10 @@ static GetTestModelFn BuildF32ConvTestCase(const std::string& conv_op_type, cons
                                            const std::vector<int64_t>& dilations,
                                            std::optional<int64_t> group,
                                            const std::string& auto_pad = "NOTSET",
-                                           std::optional<OutputActivationInfo> output_activation = std::nullopt) {
+                                           std::optional<OutputActivationInfo> output_activation = std::nullopt,
+                                           std::optional<std::vector<int64_t>> output_shape = std::nullopt) {
```
yuslepukhin: Passing by value is expensive. For optional args like this pointers are better #Resolved

ankus-qti: Done.

yuslepukhin (approved):
:shipit:

PR #26666, opened 2025-11-26
[MLAS] Rework the mechanism for checking SME capabilities

JonathanC-ARM:
### Description
In Various parts of Onnxruntime and in particular MLAS with kleidiai enabled. We rely on cpu info checks for SME and/or SME2 this patch aims to simplify that process by adding a struct which gives easy to understand information and adding it to the relevant places in the code.

Additionally it gates qgemm_kleidiai.cpp properly against non sme2 hardware gaining access. This will address https://github.com/microsoft/onnxruntime/issues/26377 which raises one such instance of this occurring on a device with has SME1 support but not SME2

### DIFF
--- a/onnxruntime/core/mlas/lib/mlasi.h
+++ b/onnxruntime/core/mlas/lib/mlasi.h
@@ -317,13 +317,21 @@ operator!=(const MLFloat16& left, const MLFloat16& right)
 
 #endif  // BUILD_MLAS_NO_ONNXRUNTIME
 
-#if defined(MLAS_TARGET_ARM64)
-
 struct SMEInfo {
     static const bool CanUseSME2;
     static const bool CanUseSME;
     static const bool IsSMEAvailable;
 };
+
+#if !defined(MLAS_TARGET_ARM64)
+// SME is only available on arm64 systems but the structure is available in places which can be accessed by all system types
+inline const bool SMEInfo::CanUseSME2 = false;
+inline const bool SMEInfo::CanUseSME  = false;
+inline const bool SMEInfo::IsSMEAvailable = false;
+
+#endif  // !MLAS_TARGET_ARM64
+
+#if defined(MLAS_TARGET_ARM64)
 
 // Boolean condition to determine if we can use SME2
 // By default we should try for SME2 first before falling back to SME.
--- a/onnxruntime/core/mlas/lib/mlasi.h
+++ b/onnxruntime/core/mlas/lib/mlasi.h
@@ -201,6 +201,10 @@ class MLASCPUIDInfo
 
     bool HasArmNeon_BF16() const { return has_arm_neon_bf16_; }
 
+    bool HasArm_SME() const { return has_arm_sme_; }
+
+    bool HasArm_SME2() const { return has_arm_sme2_; }
+
    private:
     MLASCPUIDInfo();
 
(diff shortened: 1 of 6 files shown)


onnxruntime/core/mlas/lib/mlasi.h
```diff
@@ -311,6 +317,24 @@ operator!=(const MLFloat16& left, const MLFloat16& right)
...
+
+struct SMEInfo {
+    static const bool CanUseSME2;
+    static const bool CanUseSME;
+    static const bool IsSMEAvailable;
+};
+
+// Boolean condition to determine if we can use SME2
+// By default we should try for SME2 first before falling back to SME.
+inline const bool SMEInfo::CanUseSME2 = MLAS_CPUIDINFO::GetCPUIDInfo().HasArm_SME2();
```
edgchen1, line 338: `CPUIDInfo::HasArm_SME2()` already returns a cached bool. is it worth having another constant for the same value? admittedly, `SMEInfo::CanUseSME2` is less to type, but it may be simpler to just have a single way of getting this information.

JonathanC-ARM: Thanks Edward, yeah so the reason I kept this as a seperate SMEInfo::CanUseSME2 as a seperate constant is sort of twofold.
 1) As you say it makes the code shorter to write and in places such as sgemm_kleidiai.cpp that became really useful, we were tending to be calling the MLAS_CPUIDINFO::GetCPUIDInfo().HasArm_SME2(); per file and often sometimes within files more than once.

2) having a seperate flag like this sort of separates out the behavior at runtime i.e. what we are choosing to do vs the hardware capabilities and coupled with that we are planning at introducing a follow on from this which would be a build define which can disable or enable support for sme1 or 2. To enable more dedicated testing paths particularly on sme2 machines which can support both but would default to sme2.

onnxruntime/core/mlas/lib/qgemm.cpp
```diff
@@ -223,8 +223,10 @@ MlasDynamicQGemmBatch (
     assert(MlasIsDynamicQGemmAvailable());
 
 #if defined(USE_KLEIDIAI) && !defined(_MSC_VER)
-    //No fallback
-    ArmKleidiAI::MlasDynamicQGemmBatch(Shape, DataParams, BatchN, ThreadPool);
+    //No fallback and putting in guards. This implementation is SME2 specific.
+    if(SMEInfo::CanUseSME2){
```
edgchen1, line 227: I think we don't need this check. now, the expectation is that a user will not call this function unless `MlasIsDynamicQGemmAvailable()` returns true. there's also an `assert` for that at the beginning of this function.

similar comment for `MlasDynamicQgemmPackBSize()` and `MlasDynamicQgemmPackB()`.

onnxruntime/core/mlas/lib/mlasi.h
edgchen1: IMO, having to type less is not as compelling a reason. this can be mitigated by local variables or local helper functions.

regarding separation of behavior at runtime vs. hardware capabilities, there's currently no difference. I might be more open to introducing another set of constants if there was (perhaps when the rest of that change is ready).

JonathanC-ARM:
Yeah I suppose that's fair, originally https://github.com/microsoft/onnxruntime/pull/[26784](https://github.com/microsoft/onnxruntime/pull/26784) was going to be something we then use this for to change behavior, but Hari's comment leads me to believe this might not be something that provides value to MLAS outside of a certain small edge case. I might convert this to draft for now and leave it here incase the use ever arises

PR #26668, opened 2025-11-26
Add `MlasIsDynamicQGemmAvailable()` helper and use that in place of platform-specific checks

edgchen1:
### Description

Add `MlasIsDynamicQGemmAvailable()` helper function and use that in place of platform-specific checks.

### Motivation and Context

Try to reduce platform-specific code.


onnxruntime/core/mlas/lib/qgemm.cpp
```diff
@@ -210,10 +221,8 @@ MlasDynamicQGemmBatch (
     MLAS_THREADPOOL* ThreadPool
 ) {
 #if defined(USE_KLEIDIAI) && !defined(_MSC_VER)
-    //No fallback and putting in guards. This implementation is SME2 specific.
-    if(ArmKleidiAI::UseSME2){
```
edgchen1, line 214: removing `if (SME2)` here because the expectation is that this will only be called if `MlasIsDynamicQGemmAvailable()` returns true.

PR #26673, opened 2025-11-27
[webgpu] Support int64 for range

qjia7:
### Description  
 - Add new registerInt64Ops option to WebGpuExecutionProviderConfig
 - Int64 support now enabled when enable_graph_capture OR register_int64_ops is true
 - Refactor Range kernel registration to support conditional int64 registration
  - Update kernel registry caching to handle all 4 combinations of flags
  - Rename parameters from enable_graph_capture to enable_int64 for clarity
  - Add config parsing in webgpu_provider_factory.cc for registerInt64Ops option

### Motivation
Needed by updating position id with an onnx model in genai.

Continuous decoding mode: `position_ids[i] = i + total_length - new_kv_length`

We can use an onnx model which includes a Range op to implement update the position ids:
    Inputs: start (total_length - new_kv_length), limit (total_length), delta (1)
    Output: position_ids (1D tensor of size new_kv_length)

### DIFF
--- a/onnxruntime/core/providers/webgpu/generator/range.cc
+++ b/onnxruntime/core/providers/webgpu/generator/range.cc
@@ -25,22 +25,31 @@ Status Range<T>::ComputeInternal(ComputeContext& context) const {
 
   uint32_t output_size = onnxruntime::narrow<uint32_t>(n);
   RangeProgram program{output_tensor->GetElementType()};
-#if defined(__GNUC__)
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wstrict-aliasing"
-#endif
+
+  // For int64, we need to ensure values fit in int32 range since we use 4 bytes in uniforms
+  uint32_t start_u32, delta_u32;
+  if constexpr (std::is_same_v<T, int64_t>) {
+    // Check if values fit in int32 range
+    ORT_ENFORCE(start >= std::numeric_limits<int32_t>::min() && start <= std::numeric_limits<int32_t>::max(),
+                "Range start value ", start, " is out of int32 range");
+    ORT_ENFORCE(delta >= std::numeric_limits<int32_t>::min() && delta <= std::numeric_limits<int32_t>::max(),
+                "Range delta value ", delta, " is out of int32 range");
+    int32_t start_i32 = static_cast<int32_t>(start);
+    int32_t delta_i32 = static_cast<int32_t>(delta);
+    start_u32 = std::bit_cast<uint32_t>(start_i32);
+    delta_u32 = std::bit_cast<uint32_t>(delta_i32);
+  } else {
+    start_u32 = std::bit_cast<uint32_t>(start);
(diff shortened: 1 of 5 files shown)


qjia7:
Ping @fs-eire @guschmue, Please take a look, thanks.

onnxruntime/core/providers/webgpu/webgpu_provider_options.h
```diff
@@ -11,6 +11,7 @@ namespace options {
 
 constexpr const char* kPreferredLayout = "ep.webgpuexecutionprovider.preferredLayout";
 constexpr const char* kEnableGraphCapture = "ep.webgpuexecutionprovider.enableGraphCapture";
+constexpr const char* kRegisterInt64Ops = "ep.webgpuexecutionprovider.registerInt64Ops";
```
fs-eire: "ops" is a little bit wired in option key. some candidates:
- registerInt64Operators (full for "ops")
- enableInt64

onnxruntime/core/providers/webgpu/webgpu_execution_provider.cc
```diff
@@ -932,11 +932,18 @@ std::vector<std::unique_ptr<ComputeCapability>> WebGpuExecutionProvider::GetCapa
 }
 
 std::shared_ptr<KernelRegistry> WebGpuExecutionProvider::GetKernelRegistry() const {
-  if (enable_graph_capture_) {
-    static std::shared_ptr<KernelRegistry> registry = webgpu::RegisterKernels(true);
+  // Cache all 4 combinations to preserve exact flag values for potential future use
+  if (enable_graph_capture_ && register_int64_ops_) {
```
fs-eire: I think we can make it simpler: we just check `register_int64_ops_`. and when graph capture is on, `register_int64_ops_` should always be true.

not sure if this may cause other issues but in my understanding there is no scenario where `enable_graph_capture_ == true && register_int64_ops_ == false`

onnxruntime/core/providers/webgpu/generator/range.cc
fs-eire: This implementation may no longer be safe with the change because previously `sizeof(T)` is always 4 but now T can be int64. since we use 4 bytes to represent the value, we need to do extra checks to ensure the value of start and delta falls into range of int32 representation.

also, to avoid strict aliasing violation we can use std::bit_cast

onnxruntime/core/providers/webgpu/webgpu_provider_options.h
qjia7: Done with `enableInt64`.

onnxruntime/core/providers/webgpu/webgpu_execution_provider.cc
qjia7: Done

onnxruntime/core/providers/webgpu/generator/range.cc
qjia7: Done.

PR #26675, opened 2025-11-28
WebGPU: Transpose Conv kernels in Prepack

jchen10:
Prepack Conv kernels with path-aware transpose decisions, store the transposed kernels for reuse, and add ComputeContextBase helpers for node access and GPU buffer unmapping.

### DIFF
--- a/onnxruntime/core/providers/webgpu/nn/conv.cc
+++ b/onnxruntime/core/providers/webgpu/nn/conv.cc
@@ -1,7 +1,6 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 #include "core/providers/webgpu/nn/conv.h"
-#include "core/graph/node_arg.h"
 #include "core/providers/webgpu/nn/conv2d_mm.h"
 #include "core/providers/webgpu/nn/im2col_matmul.h"
 #include "core/providers/webgpu/shader_helper.h"
@@ -303,41 +302,17 @@ Status Conv<is_channels_last, is_fused>::PrePackInternal(ComputeContextBase& con
     return Status::OK();
   }
 
-  // Get kernel spatial dimensions
-  const int64_t kernel_height = dims[2];
-  const int64_t kernel_width = dims[3];
-
-  // Get input shape to check same_size condition
-  const auto& input_defs = context.GetNode().InputDefs();
-  const auto* input_arg = input_defs[0];
-  int64_t input_height = -1;
-  int64_t input_width = -1;
-  if (input_arg && input_arg->Exists()) {
-    const auto* input_shape_proto = input_arg->Shape();
-    if (input_shape_proto && input_shape_proto->dim_size() >= 3) {
-      if constexpr (is_channels_last) {
-        // For channels_last: [N, H, W, C] or [N, W, C] for Conv1D
-        if (input_shape_proto->dim_size() == 4) {
-          if (input_shape_proto->dim(1).has_dim_value()) {
-            input_height = input_shape_proto->dim(1).dim_value();
-          }
-          if (input_shape_proto->dim(2).has_dim_value()) {
-            input_width = input_shape_proto->dim(2).dim_value();
-          }
-        }
-      } else {
-        // For channels_first: [N, C, H, W] or [N, C, W] for Conv1D
-        if (input_shape_proto->dim_size() == 4) {
-          if (input_shape_proto->dim(2).has_dim_value()) {
-            input_height = input_shape_proto->dim(2).dim_value();
-          }
-          if (input_shape_proto->dim(3).has_dim_value()) {
-            input_width = input_shape_proto->dim(3).dim_value();
-          }
-        }
-      }
-    }
-  }
+  // Analyze execution paths in ComputeInternal to determine if kernel transpose is needed:
+  //
+  // 1. Im2ColMatMul path: handled above (skip prepacking)
+  // 2. Grouped conv (group > 1): handled above (skip if !is_channels_last)
+  // 3. MatMul optimization (same_size || is_1x1_conv):
+  //    - is_channels_last: transposes
+  //    - !is_channels_last: does NOT transpose (uses kernel directly)
+  // 4. General conv (fallback): ALWAYS transposes
+  //
+  // The ONLY path that doesn't transpose is: is_1x1_conv && !is_channels_last
+  // (same_size requires is_channels_last, so it's always false for channels_first)
 
   // Get pads and strides (mirroring ComputeInternal logic)
   std::vector<int64_t> pads(conv_attrs_.pads.begin(), conv_attrs_.pads.end());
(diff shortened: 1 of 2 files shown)


jchen10:
Perf data on LNL:



model | variance(%)
-- | --
sd-turbo-unet-fp16-demo-layernorm | -23.72%
modnet-fp32 | -22.99%
sd-turbo-text-encoder-fp16-demo-layernorm | -17.58%
efficientnet-lite-f16-demo | -15.28%
mobilenetv2-12-f16-demo | -14.18%
jina-clip-v1-version | -12.61%
gazenet | -12.22%
sdunet-v1.5-demo-layernorm | -11.43%
modnet-fp16 | -10.06%
resnet50-v1-f16-demo | -8.14%
florence-2-base-decoder-fp16 | -7.95%
movenet-singlepose-thunder-fp32 | -7.61%
jina-clip-v1-version-fp16 | -7.54%
depth-anything-base-fp32 | -7.45%
detr-resnet-50-fp16 | -6.55%
detr-resnet-50 | -6.33%
jina-clip-v1-text | -6.32%
movenet-singlepose-thunder-fp16 | -6.04%
mobileclip_s0_vision_fp32 | -5.14%

jchen10:
@fs-eire @qjia7 @guschmue PTAL

jchen10:
Found the CI error log below. Not quite sure if it is really caused by this PR.

2025-12-02T20:34:21.9671092Z 2: [  FAILED  ] CudaNhwcTypedTest/0.ConvNhwcBias, where TypeParam = float (186 ms)

```
2025-12-02T20:34:21.8768402Z 2: [1;31m2025-12-02 20:34:21.8759375 [E:onnxruntime:Conv, sequential_executor.cc:572 onnxruntime::ExecuteKernel] Non-zero status code returned while running Conv node. Name:'node1' Status Message: CUDA error cudaErrorNotSupported:operation not supported[m
2025-12-02T20:34:21.8769974Z 2: E:\_work\onnxruntime\onnxruntime\onnxruntime\test\providers\compare_provider_test_utils.cc(172): error: Value of: _tmp_status.IsOK()
2025-12-02T20:34:21.8770496Z 2:   Actual: false
2025-12-02T20:34:21.8770652Z 2: Expected: true
2025-12-02T20:34:21.8771227Z 2: [ONNXRuntimeError] : 1 : FAIL : Non-zero status code returned while running Conv node. Name:'node1' Status Message: CUDA error cudaErrorNotSupported:operation not supported
2025-12-02T20:34:21.8771728Z 2: 

```

jchen10:
> Found the CI error log below. Not quite sure if it is really caused by this PR.
> 
> 2025-12-02T20:34:21.9671092Z 2: [ FAILED ] CudaNhwcTypedTest/0.ConvNhwcBias, where TypeParam = float (186 ms)
> 
> ```
> 2025-12-02T20:34:21.8768402Z 2: �[1;31m2025-12-02 20:34:21.8759375 [E:onnxruntime:Conv, sequential_executor.cc:572 onnxruntime::ExecuteKernel] Non-zero status code returned while running Conv node. Name:'node1' Status Message: CUDA error cudaErrorNotSupported:operation not supported�[m
> 2025-12-02T20:34:21.8769974Z 2: E:\_work\onnxruntime\onnxruntime\onnxruntime\test\providers\compare_provider_test_utils.cc(172): error: Value of: _tmp_status.IsOK()
> 2025-12-02T20:34:21.8770496Z 2:   Actual: false
> 2025-12-02T20:34:21.8770652Z 2: Expected: true
> 2025-12-02T20:34:21.8771227Z 2: [ONNXRuntimeError] : 1 : FAIL : Non-zero status code returned while running Conv node. Name:'node1' Status Message: CUDA error cudaErrorNotSupported:operation not supported
> 2025-12-02T20:34:21.8771728Z 2: 
> ```

Tried the case locally with CUDA EP. It didn't reproduce with this PR.

jchen10:
@fs-eire PTAL

onnxruntime/core/providers/webgpu/nn/conv.cc
```diff
@@ -203,56 +221,162 @@ Status Conv<is_channels_last, is_fused>::ComputeInternal(ComputeContext& context
...
-      //       - in this case, use `transposed_kernel_` instead.
+  const auto& kernel_shape = tensor.Shape();
+  const auto& dims = kernel_shape.GetDims();
 
-      // // Step.1 - calculate transposed weight shape
-      // TensorShape transposed_kernel_shape{tensor.Shape()[2],
-      //                                     tensor.Shape()[3],
-      //                                     tensor.Shape()[1],
-      //                                     tensor.Shape()[0]};
+  // Conv kernels must be 4D: [O, I, H, W]
```
qjia7, line 274: How about conv1d's kernel?

onnxruntime/core/providers/webgpu/nn/conv.cc
```diff
@@ -203,56 +221,162 @@ Status Conv<is_channels_last, is_fused>::ComputeInternal(ComputeContext& context
...
+  if (dims.size() != 4) {
+    return Status::OK();
+  }
 
-      // // Step.2 - create transposed weight tensor
-      // transposed_kernel_ = std::make_unique<Tensor>(tensor.DataType(), transposed_kernel_shape, alloc);
+  // Get group attribute
+  int64_t group = conv_attrs_.group;
```
qjia7: Please early return when group > 1 && !is_channels_last.

For below, can I simply understand that transpose will always happen except for point-wise conv (is_1x1_conv = true)? It seems that there is no need to check that same_size is true or false.

onnxruntime/core/providers/webgpu/nn/conv.cc
```diff
@@ -203,56 +221,162 @@ Status Conv<is_channels_last, is_fused>::ComputeInternal(ComputeContext& context
...
+        }
+      }
     }
   }
 
+  // Get pads and strides
+  const auto& pads_vec = conv_attrs_.pads;
+  const auto& strides_vec = conv_attrs_.strides;
+
+  std::vector<int64_t> pads(pads_vec.begin(), pads_vec.end());
```
qjia7: In Conv, it got `pads` by `InferPadsAndOutputShape`. Are they equal?

onnxruntime/core/providers/webgpu/nn/conv.cc
```diff
@@ -203,56 +221,162 @@ Status Conv<is_channels_last, is_fused>::ComputeInternal(ComputeContext& context
...
                                                          /*out*/ bool& is_packed) {
   is_packed = false;
 
-  if constexpr (is_channels_last) {
-    if (input_idx == 1 && tensor.Shape().NumDimensions() == 4) {
-      // only deal with 4D NHWC weights
+  // Only prepack kernel weights (input_idx == 1)
+  if (input_idx != 1) {
+    return Status::OK();
+  }
```
qjia7, line 269: Jianhui added the CanApplyIm2ColMatMulProgram path. Should it be considered since once the prepack is done the original kernel may be released?

onnxruntime/core/providers/webgpu/nn/conv.cc
jchen10: For conv1d, I suppose the performance shouldn't be too much a problem. As the code paths are already complicated enough, maybe it doesn't deserve further complexity.

onnxruntime/core/providers/webgpu/nn/conv.cc
jchen10: Done with early return.
same_size and is_1x_conv are not exactly same according to ComputeInternal. Frankly I'm not firmly sure about them. The current conv.cc is not easy to read. Hopefully we could cleanup it someday.

onnxruntime/core/providers/webgpu/nn/conv.cc
jchen10: Good catch. They are not same. Fixed it by checking autopad is not set.

onnxruntime/core/providers/webgpu/nn/conv.cc
jchen10: Good point. Added that check.

jchen10:
@qjia7 Really sorry. I totally missed your review comments.

onnxruntime/core/providers/webgpu/nn/conv.cc
jchen10: > there is no need to check that same_size is true or false.
You are right.

channel_last | same_size | 1x1 | transpose
-- | -- | -- | --
0 | 0 | 0 | T
0 | 0 | 1 | F
1 | 0 | 0 | T
1 | 0 | 1 | T
1 | 1 | 0 | T
1 | 1 | 1 | T


Examining the code paths in ComputeInternal, I found there is only one condition where transpose is not needed. The condition is "is_1x1_conv && !is_channel_last". So we don't need check input_width and input_height at all. This could simply the node checking code lines in this PR. However, I still would like to have a straightforward mapping between ComputeInternal and PrePackInternal. It would be easy to maintain and keep them in sync. How do you like it?

qjia7: I’d prefer to keep the code as simple as possible in this PR. We can then refactor ComputeInternal so that it syncs to the prepack path, rather than the other way around.

jchen10: Sounds fair, done!

PR #26654, opened 2025-11-25
ARM KleidiAI Micro-Kernel: Depthwise Convolution (3x3 kernel, Stride of 1)

damdoo01-arm:
### Description
This PR introduces a dedicated kernel to perform Depthwise Convolution for specific cases (3x3 kernel, stride of 1).

### Motivation and Context
We have identified opportunities for uplift for this layer type and several customer models consist of layers which would benefit from this kernel being included.

image

PR #26688, opened 2025-12-01
[MLAS/NEON] Add dedicated kernel for depthwise convolution for ARM64 using NEON intrinsics

hariharans29:
### Description

**Motivation and approach taken:**

Add a dedicated depthwise convolution kernel for the most common depthwise convolution configuration (3x3 filter, stride = 1, pad <= 1, dilation = 1) using NEON intrinsics. This does significantly better than the current approach of `Im2Col + SGemm`. The Im2Col step extracts convolution patches and this is a wasteful step and for a 3x3 filter, K would be 9 for the SGemm and usually Gemms are not optimized for such small `K` values. Hence, a dedicated kernel works much better.

Initially, I ported over the Winograd based NEON accelerated depthwise convolution kernel from PyTorch but I found that its performance is not very good. It's poor performance  is probably due to applying the Winograd transformation for the filter repeatedly. A better approach may be to tranform the filter offline and this approach can be considered for later (I reverted the PyTorch Winograd implementation in this commit: https://github.com/microsoft/onnxruntime/pull/26688/commits/2820a84261123499e6ddb03e734810d8f6ad98ed). 

The current depthwise kernel added in this PR was authored by GPT5.1-Codex and with some minor bug fixes it seems to be functionally correct now and also provides the perf boost we are seeking.

**Unit tests:**
There are already depthwise convolution tests already existing in the codebase. I don't see a need for new ones at this point.

**Kernel benchmarking:**
This is the kernel level perf improvement from MLAS Conv benchmarks (About 50% kernel latency improvements):

image

### Motivation and Context
A key customer model had a few depthwise conolution operations and this change provides a **non-negligible ~3% throughput improvement** using the customer provided benchmarking setup

For those interested, https://github.com/microsoft/onnxruntime/pull/26654 adds support for the same type of convolution variant but that leverages SME1/SME2 through KleidiAI. This PR is conceptually the same but targeting NEON only platforms.


onnxruntime/core/mlas/lib/sconv_nchw_kernel_neon.cpp
```diff
@@ -0,0 +1,297 @@
...
+    }
+
+    const float* r = row + base;
+    const float32x4_t c0 = vld1q_f32(r);
+    const float32x4_t c1 = vld1q_f32(r + 1);
+    const float32x4_t c2 = vld1q_f32(r + 2);
+
+    acc = vmlaq_n_f32(acc, c0, w0);
+    acc = vmlaq_n_f32(acc, c1, w1);
+    acc = vmlaq_n_f32(acc, c2, w2);
```
Rohanjames1997: Consider using `MlasLoadFloat32x4` & `MlasMultiplyAddFloat32x4`? The latter has an overloaded implementation for a scalar multiplier as well. 

And in other parts of the file, you can use `MlasZeroFloat32x4` & `MlasStoreFloat32x4`

hariharans29:
> Interesting PR!
> 
> Can you also tell me the command you executed for the benchmark in [bench_sconv.cpp](https://github.com/microsoft/onnxruntime/pull/26688/changes#diff-bb6b724400d6118441dcfa69d185d664cbbe793f8a5e60a937dddb40bc1ea9bf)

I think I just ran all the benchmarks in that file (Something like the below)

./onnxruntime_mlas_benchmark --benchmark_filter=TeamsModel
./onnxruntime_mlas_benchmark --benchmark_filter=2d
./onnxruntime_mlas_benchmark --benchmark_filter=ResNet50

and compared before and after numbers. I ensured that I saw gains pertaining to the depthwise conv alone and that I didn't see regressions in the other cases. In the PR description, I just noted the perf improvements of the threaded versions of the depthwise convs.

onnxruntime/core/mlas/lib/sconv_nchw_kernel_neon.cpp
hariharans29: Thanks for the suggestion. I will look into this.

onnxruntime/core/mlas/lib/convolve.cpp
```diff
@@ -805,6 +805,90 @@ Return Value:
...
+    const size_t K = Parameters->K;
+
+    const size_t InputGroupSize = Parameters->InputChannels * Parameters->InputSize;
+    const size_t OutputGroupSize = FilterCount * OutputSize;
+    const size_t FilterGroupSize = FilterCount * K;
+
+    for (size_t bg = BatchGroupStart; bg < BatchGroupEnd; bg++) {
+        size_t group = bg % GroupCount;
+
+        const float* input = WorkBlock->Input + bg * InputGroupSize;
```
yuslepukhin, line 878: >bg * InputGroupSize; [](http://example.com/codeflow?start=48&length=20)

Since bg is incrementing, would it be better to increment the product by InputGroupSize instead of performing multiplication twice each iteration and then clamp it?

hariharans29: Are you referring to `bg * InputGroupSize` and `bg * OutputGroupSize` as the 2 multiplications ?
If so, I guess some pointer arithmetic can be re-worked to save state and keep incrementing the input and output group sizes every bg iteration. But do you think this will have perf gains ?

hariharans29: Resolved.

Rohanjames1997:
Since this PR and https://github.com/microsoft/onnxruntime/pull/26838 were merged in a short time window, we may have hit an edge case where the CI for #26838 succeeded, but it was last run before this PR merged.*

So, we still have a reference to `sconv.h` lying around here which is failing the mainline build when run with `--enable_arm_neon_nchwc` https://github.com/microsoft/onnxruntime/blob/01f40e6367fa2c6c3a23ab5fda48dbd70b07ad81/onnxruntime/core/mlas/lib/sbconv_kernel_neon.cpp#L20


*This is my hypothesis. I do not know if the CI is extensive enough to cover `--enable_arm_neon_nchwc`

hariharans29:
> Since this PR and #26838 were merged in a short time window, we may have hit an edge case where the CI for #26838 succeeded, but it was last run before this PR merged.*
> 
> So, we still have a reference to `sconv.h` lying around here which is failing the mainline build when run with `--enable_arm_neon_nchwc`
> 
> https://github.com/microsoft/onnxruntime/blob/01f40e6367fa2c6c3a23ab5fda48dbd70b07ad81/onnxruntime/core/mlas/lib/sbconv_kernel_neon.cpp#L20
> 
> *This is my hypothesis. I do not know if the CI is extensive enough to cover `--enable_arm_neon_nchwc`

Oh ! I think the reason is NCHWC is only enabled on MacOS now. With th BF16 feature being turned on only on Linux, I think we may have to enable it in the Linux CI as well. We are a bit busy with the 1.24 release and I will add it to the backlog and re-visit later. 

Can you please propose a fix ? FWIW - This is a PR that might interest you AND have the fix - https://github.com/microsoft/onnxruntime/pull/27099. The PR may not go in 1.24 at this point and so if you could please provide an isolated fix - I will ensure the isolated fix gets cherry-picked into 1.24. Thanks!

Rohanjames1997:
Cool np!

Yeah, I did see #27099 and it prompted me to check why it wasn't caught earlier.

I will raise a fix for it now.

Rohanjames1997:
https://github.com/microsoft/onnxruntime/pull/27101

Issue #26690, opened 2025-12-01
[Web] WebGPU FP16 Slice regression in 1.24-dev


RReverser:
### Describe the issue

Getting this error on a Slice node that slices a FP16 tensor when running with WebGPU backend.

> Error processing: failed to call OrtRun(). ERROR_CODE: 6, ERROR_MESSAGE: Non-zero status code returned while running Slice node. Name:'' Status Message: /mnt/vss/_work/1/s/include/onnxruntime/core/framework/tensor.h:229 gsl::span<const T> onnxruntime::Tensor::DataAsSpan() const [T = long long] utils::IsPrimitiveDataType<T>(dtype_) was false. Tensor type mismatch. T !=(unknown type)

This worked fine in the latest stable version (1.23) and works fine on CPU and WebNN-GPU backends.

**Off-topic:** In fact, WebNN backend seems to have gotten almost unbelievably faster - over 2x faster with chrome://flags/#webnn-directml and almost 6x faster with chrome://flags/#webnn-onnxruntime 🤯 Still looking for the catch, because previously it was actually a little slower than WebGPU. If this is what WebNN is going to look like when it's stable, it's pretty amazing.

### To reproduce

Can't provide full model, but if it's difficult to reproduce, happy to whip one up together with onnxscript or something. I'm pretty sure having a `Slice` node on fp16 input should be enough.

### ONNX Runtime Installation

Released Package

### ONNX Runtime Version or Commit ID

1.24.0-dev.20251116-b39e144322

### Execution Provider

'webgpu' (WebGPU)

guschmue:
I think this is when start, stop, step and axis is int32 - need to check how this broken, possible that was never implemented.

RReverser:
> I think this is when start, stop, step and axis is int32 - need to check how this broken, possible that was never implemented.

Ah, never thought about adding that detail, sorry. Indeed, I've been always using int32 for these params.

RReverser:
Should I close this now that PR was merged or wait for a new `@dev` release that I could test locally?

PR #26693, opened 2025-12-01
add support for int32_t in webgpu / slice

guschmue:
fix for https://github.com/microsoft/onnxruntime/issues/26690

### DIFF
--- a/onnxruntime/core/providers/webgpu/tensor/slice.cc
+++ b/onnxruntime/core/providers/webgpu/tensor/slice.cc
@@ -92,14 +92,28 @@ Status SliceProgram::GenerateShaderCode(ShaderHelper& shader) const {
   return Status::OK();
 }
 
+static std::vector<int64_t> getInt64Input(const Tensor* tensor) {
+  if (tensor->IsDataType<int64_t>()) {
+    return std::vector<int64_t>(tensor->DataAsSpan<int64_t>().begin(), tensor->DataAsSpan<int64_t>().end());
+  }
+  ORT_ENFORCE(tensor->IsDataType<int32_t>(), "Expected tensor of type int32 or int64");
+  std::vector<int64_t> result;
+  auto span = tensor->DataAsSpan<int32_t>();
+  result.reserve(span.size());
+  for (auto v : span) {
+    result.push_back(static_cast<int64_t>(v));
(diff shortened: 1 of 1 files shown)


onnxruntime/core/providers/webgpu/tensor/slice.cc
```diff
@@ -92,14 +92,28 @@ Status SliceProgram::GenerateShaderCode(ShaderHelper& shader) const {
...
+
 Status Slice::ComputeInternal(ComputeContext& context) const {
   // READ INPUTS
   const Tensor* input_tensor = context.Input(0);
   const TensorShape& input_shape = input_tensor->Shape();
   auto input_rank = input_shape.NumDimensions();
 
-  auto starts_raw = attr_starts_.empty() ? context.Input(1)->DataAsSpan<int64_t>() : gsl::make_span(attr_starts_);
-  auto ends_raw = attr_ends_.empty() ? context.Input(2)->DataAsSpan<int64_t>() : gsl::make_span(attr_ends_);
+  auto starts_raw = attr_starts_.empty() ? getInt64Input(context.Input(1)) : attr_starts_;
```
RReverser, line 115: Given that below most of these arrays are being converted back to `int32`/`uint32`, wouldn't it be better to have `getInt32Input` here to avoid extra vectors?

guschmue: I think most models use int64_t (you are the first that runs into this :)).
But same thing is bugging me for int64_t too, I don't like the vectors.
cpu does it the same way.
The way to fix it would be to move the different handling for int32/int64 into the op itself but this would be hard to read, and you'd need to do it 4 times.

Slice never shows up in a profile in a significant way while already doing bunch of this vector stuff - don't think it will be noticeable.

RReverser: > I think most models use int64_t (you are the first that runs into this :)).

Yeah it just seems slightly... suboptimal (incl. for readability) to convert int32->int64->int32. But you're right, it's unlikely to show up anywhere.

RReverser: And thanks for such a quick turnaround!

PR #26695, opened 2025-12-01
Implement new experimental lookup-based matrix multiplication method(TMAC)

vraspar:
### Description

This PR introduces a new experimental lookup-table(LUT) based matrix multiplication method for 2-bit MatMulNBits on x64 AVX2 inspired from [T-MAC paper](https://arxiv.org/abs/2407.00088) and [T-MAC repository](https://github.com/microsoft/T-MAC) to speed up low bit LLM inference. 

Unlike the existing quant-dequant methods, the LUT-based method directly supports mixed-precision-GEMM without dequantization. It uses bit-wise table lookup to eliminate multiplications and reduce additions required in matrix multiplication.

image
 

This PR:
 - Add` mlas.use_lut_gemm` session option allowing use of LUT GEMM inside matmulnbits when it is available (2-bit, BlkLen multiple of 32, K multiple of 32, N multiple of 128, AVX2 present).
 - Introduces LUT packing + kernel config cache (packs bitplanes, scales, ZP) and the main `MlasLUTGemm` entry that generates per-row LUTs and calls the AVX2 kernel.
 - Implements AVX2 LUT generation `GenerateLUT_avx2` and GEMM compute `TMACComputeGemm_avx2` and wires dispatch in MLAS platform init.
- Updates MatMulNBits PrePack/Compute to use LUT packing/compute when opted-in; keeps existing quant-dequant path as fallback.
- Extends Python quant bindings with 2-bit QDQ helper for parity with the new path.
- Adds MLAS unit tests covering LUT GEMM across symmetric/asymmetric quant and multiple shapes/block sizes.

 
 ### Main components:
 
 - `MlasInitLUTGemmKernelConfig`: Config for LUT kernels
 - `MlasLUTGemmPackQuantBData`: Pre Packing of quantized weight
 - `MlasLUTPackScalesAndZeroPoints`: Pre Packing of qunatized scales and zero points
 
 - `MlasLUTGemm`: Main Entry point
 - `GenerateLUT_avx2`:  LUT construction from activations
 - `TMACComputeGemm_avx2`: AVX2 LUT GEMM kernel
 - Session option: mlas.use_lut_gemm

### How to test
- MLAS LUT GEMM unit tests: see `test_sqlutgemm.cpp`
-  Run MatMulNBits models with session option `mlas.use_lut_gemm=1` on AVX2 machines; expect fallback to existing path if availability checks fail.

### Perf
Focus of this PR is functional + kernel bring-up; perf to be reported separately once broader profiling is done.

### Future Work
- Support MLFloat16 (FP16 scales and zero points)
- Add neon kernel for ARM.
- Add kernels for 4 bit weights and bitnet kernels
- Broader batch (N>1) support and additional shape coverage.


onnxruntime/contrib_ops/cpu/quantization/matmul_nbits.cc
```diff
@@ -167,6 +176,15 @@ class MatMulNBits final : public OpKernel {
                         AllocatorPtr& allocator,
                         concurrency::ThreadPool* thread_pool,
                         const MatMulComputeHelper& helper) const;
+
+  Status ComputeBPackedLUT(const Tensor* a,
+                           const Tensor* scales,
```
jambayk: scales, zero_points, bias and allocator are not used in the implementation so could be removed from the signature and call.

onnxruntime/python/onnxruntime_pybind_quant.cc
```diff
@@ -130,10 +143,14 @@ void CreateQuantPybindModule(py::module& m) {
...
   m.def("quantize_matmul_4bits", &QuantizeMatMulNBitsBlockwise<float, 4>);
   m.def("quantize_matmul_4bits", &QuantizeMatMulNBitsBlockwise<MLFloat16, 4>);
+  m.def("quantize_matmul_nbits", &QuantizeMatMulNBitsBlockwise<MLFloat16, 4>);
+  m.def("quantize_matmul_nbits", &QuantizeMatMulNBitsBlockwise<float, 4>);
   m.def("quantize_matmul_8bits", &QuantizeMatMulNBitsBlockwise<float, 8>);
   m.def("quantize_matmul_8bits", &QuantizeMatMulNBitsBlockwise<MLFloat16, 8>);
   m.def("quantize_matmul_bnb4", &QuantizeMatMulBnb4Blockwise<float>);
   m.def("quantize_matmul_bnb4", &QuantizeMatMulBnb4Blockwise<MLFloat16>);
+  m.def("quantize_qdq_matmul_2bits", &QuantizeQDQMatMulNBitsBlockwise<float, 2>);
+  m.def("quantize_qdq_matmul_2bits", &QuantizeQDQMatMulNBitsBlockwise<MLFloat16, 2>);
```
jambayk, line 151: this one is added here but QuantizeQDQMatMulNBitsBlockwise<MFloat16, 2> is not initialized in onnxruntime/core/mlas/lib/q4_dq.cpp

include/onnxruntime/core/session/onnxruntime_session_options_config_keys.h
```diff
@@ -374,6 +374,12 @@ static const char* const kOrtSessionOptionsEpContextModelExternalInitializersFil
 // - "1": Gemm FastMath mode is enabled.
 static const char* const kOrtSessionOptionsMlasGemmFastMathArm64Bfloat16 = "mlas.enable_gemm_fastmath_arm64_bfloat16";
 
+// Use LUT based GEMM for quantized models when available.
```
edgchen1: maybe define what "LUT" stands for at least once. this might be a good place.

include/onnxruntime/core/session/onnxruntime_session_options_config_keys.h
```diff
@@ -374,6 +374,12 @@ static const char* const kOrtSessionOptionsEpContextModelExternalInitializersFil
...
+// Option values:
+// - "0": Do not use LUT based GEMM. [DEFAULT]
+// - "1": Use LUT based GEMM when available.
+static const char* const kOrtSessionOptionsMlasLUTGemm = "mlas.use_lut_gemm";
```
edgchen1: general naming nit: the abbreviation "LUT" can be spelled "Lut" in names. e.g., `kOrtSessionOptionsMlasLutGemm`.

see https://google.github.io/styleguide/cppguide.html#General_Naming_Rules

> When an abbreviation is used, prefer to capitalize it as a single "word", e.g., StartRpc() rather than StartRPC().

onnxruntime/contrib_ops/cpu/quantization/matmul_nbits.cc
```diff
@@ -116,6 +121,7 @@ class MatMulNBits final : public OpKernel {
                 "Only 2b, 4b and 8b quantization is supported for MatMulNBits op, additional bits support is planned.");
     const Tensor* tensor_zero_point = nullptr;
     has_zp_input_ = info.TryGetConstantInput(InputIndex::zero_points, &tensor_zero_point);
+    prefer_lut_gemm_ = prefer_lut_gemm_ && MlasIsLUTGemmAvailable(N_, K_, nbits_, block_size_);
```
edgchen1: can we put all the initialization logic for `prefer_lut_gemm_` in the member initializer? then it could also be made const.

onnxruntime/contrib_ops/cpu/quantization/matmul_nbits.cc
```diff
@@ -179,22 +193,54 @@ Status MatMulNBits<T1>::PrePack(const Tensor& tensor, int input_idx, /*out*/ All
     return Status::OK();
   }
 
-  if (!MlasIsQNBitGemmAvailable(nbits_, block_size_, compute_type_)) {
+  if (!MlasIsQNBitGemmAvailable(nbits_, block_size_, compute_type_) && !prefer_lut_gemm_) {
     return Status::OK();
   }
+
+  // Create a temporary threadpool for parallel packing
+  // This is used during model load time to speed up weight prepacking
```
edgchen1, line 202: what is the overhead like for creating a new threadpool in each call to `PrePack()`?

I wonder if we should make an existing threadpool available to this code. perhaps we can pass in the threadpool from SessionState. something to consider, and maybe for a future PR.

onnxruntime/core/mlas/inc/mlas_qnbit.h
```diff
@@ -232,3 +232,96 @@ MlasQNBitGemmScalesPacked(
     MLAS_QNBIT_GEMM_COMPUTE_TYPE ComputeType,
     bool HasZeroPoint
 );
+
+size_t MLASCALL
+MlasLUTGemmPackQuantBDataSize(
```
edgchen1: please document all the new public functions

onnxruntime/core/mlas/lib/qlutgemm.cpp
```diff
@@ -0,0 +1,615 @@
...
+    int M,  // batch size (number of rows in activation)
+    int N,
+    MLAS_THREADPOOL* threadpool
+)
+{
+    // adapted from ggml_backend_tmac_mul_mat
+    const auto* Dispatch = GetMlasPlatform().LUTGenKernel;
+    if (!Dispatch || !Dispatch->GenerateLUT) {
+        ORT_THROW("TMAC not supported in this configuration.");
+    }
```
edgchen1: this should be ensured by only calling `MlasLUTGemm()` if a previous call to `MlasIsLUTGemmAvailable()` returned true, right? maybe make it an `assert` instead?

onnxruntime/core/mlas/lib/qlutgemm.cpp
```diff
@@ -0,0 +1,615 @@
...
+                    // TODO(vrapsar): fix this M and N swapp mess
+                    Dispatch->ComputeGemm(
+                        packed_weights + w_offset,       // Weight tile
+                        QuantBScale + scales_offset,     // Weight scales for this tile
+                        qlut + qlut_offset,              // LUT for this batch row
+                        lut_scales + lut_scales_offset,  // LUT scales
+                        lut_biases + lut_scales_offset,  // LUT biases
+                        act_output + dst_offset,         // Output location
+                        static_cast<int>(K),             // K dimension
+                        static_cast<int>(N),             // K dimension
```
edgchen1: should the comment say "N"?

onnxruntime/core/mlas/lib/qlutgemm.cpp
```diff
@@ -0,0 +1,615 @@
...
+
+    // search space
+    std::vector<size_t> bms;
+    if (nbits == 1 || nbits == 2 || nbits == 4) {
+        bms = {256, 512, 1024, 2048, 320, 640, 1280};
+    } else if (nbits == 3) {
+        bms = {192, 384, 576, 758};
+    }
+
+    std::vector<size_t> bns = {8, 16, 32, 64};
```
edgchen1: unused?

onnxruntime/core/mlas/lib/qlutgemm.cpp
```diff
@@ -0,0 +1,615 @@
...
+
+#include <cassert>
+#include <cstring>
+#include <memory>
+#include <string>
+#include <thread>
+#include <unordered_map>
+
+/** T-MAC GEMM kernel Config */
+static std::unordered_map<std::string, struct MlasTMACKernelParams> tmac_kernel_configs;
```
edgchen1, line 31: is global state necessary? global state requires some care, e.g., with multithreading.

is it feasible to just compute the `MlasTMACKernelParams` as needed, or is it worth keeping a global cache for it?

onnxruntime/core/mlas/lib/qlutgemm.cpp
```diff
@@ -0,0 +1,615 @@
...
+void MLASCALL
+MlasClearLUTGemmKernelConfig()
+{
+    tmac_kernel_configs.clear();
+}
+
+void MLASCALL
+MlasInitLUTGemmKernelConfig(size_t M, size_t N, size_t nbits, size_t block_size, bool has_zp_point)
+{
+    std::string key = std::to_string(M) + "_" + std::to_string(N) + "_" + std::to_string(nbits) + "_" + std::to_string(block_size);
```
edgchen1: this key doesn't account for `has_zp_point`. two calls to `MlasInitLUTGemmKernelConfig()` that only differ in `has_zp_point` would only initialize one set of parameters.

onnxruntime/core/mlas/lib/qlutgemm.cpp
```diff
@@ -0,0 +1,615 @@
...
+}
+
+void MLASCALL
+MlasClearLUTGemmKernelConfig()
+{
+    tmac_kernel_configs.clear();
+}
+
+void MLASCALL
+MlasInitLUTGemmKernelConfig(size_t M, size_t N, size_t nbits, size_t block_size, bool has_zp_point)
```
edgchen1: nit: `has_zp_point` -> `has_zp` or `has_zero_point`

onnxruntime/core/mlas/inc/mlas_qnbit.h
```diff
@@ -232,3 +232,96 @@ MlasQNBitGemmScalesPacked(
...
+void MLASCALL
+MlasLUTGemm(
+    const void* A,
+    size_t BlkLen,
+    const void* QuantBData,
+    const float* QuantBScale,
+    void* C,
+    int K,
+    int M,
+    int N,
```
edgchen1: elsewhere, M, K, and N are size_t. why the inconsistency here?

onnxruntime/python/onnxruntime_pybind_quant.cc
```diff
@@ -130,10 +143,14 @@ void CreateQuantPybindModule(py::module& m) {
   m.def("quantize_matmul_2bits", &QuantizeMatMulNBitsBlockwise<MLFloat16, 2>);
   m.def("quantize_matmul_4bits", &QuantizeMatMulNBitsBlockwise<float, 4>);
   m.def("quantize_matmul_4bits", &QuantizeMatMulNBitsBlockwise<MLFloat16, 4>);
+  m.def("quantize_matmul_nbits", &QuantizeMatMulNBitsBlockwise<MLFloat16, 4>);
+  m.def("quantize_matmul_nbits", &QuantizeMatMulNBitsBlockwise<float, 4>);
```
edgchen1: why are these ones added? aren't they the same as the `quantize_matmul_4bits` versions above?

onnxruntime/core/mlas/lib/sqnbitgemm_lut_kernel_avx2.cpp
```diff
@@ -0,0 +1,684 @@
...
+void
+QuantizeARow_CompInt8(
+    size_t /*BlkLen*/,
+    const float* /*A*/,
+    size_t /*CountK*/,
+    std::byte* /*QuantA*/
+)
+{
+    // placeholder so that dispatch doesn't break
+    // TODO: figure out a way that we can omit this altogether
```
edgchen1: seems an odd place for this empty definition. what's the error?

onnxruntime/core/mlas/lib/platform.cpp
```diff
@@ -37,11 +35,6 @@ Module Name:
 #endif
 #endif
```
edgchen1, line 41: some of the changes in this file don't seem related to the rest of this PR. are they intentional?

onnxruntime/contrib_ops/cpu/quantization/matmul_nbits_impl.cc
```diff
@@ -16,6 +16,15 @@
 namespace onnxruntime {
 namespace contrib {
 
+template <class T, class zeroT>
+void Dequantize2BitsKernelReOrder(
+    T* /*output*/, const uint8_t* /*quant_data*/, const T* /*scale_data*/,
+    const zeroT* /*zero_points*/, const int32_t* /*reorder_idx*/, int /*block_size*/,
+    int /*groups_per_threadblock*/, int /*total_groups*/, int /*out_rows*/, int /*out_cols*/,
+    int /*blockIdx_x*/, int /*threadIdx_x*/) {
+  assert(false);
```
edgchen1: why the `assert(false)` here?

onnxruntime/contrib_ops/cpu/quantization/matmul_nbits.cc
vraspar: I agree, passing thread pool to `PrePack` would be clean. I am planning to create second PR improving Prepacking logic in general, I will include this along with this :)

onnxruntime/core/mlas/lib/qlutgemm.cpp
vraspar: My reason for using global state was to avoid calculation of `kfactor` and `bm`. At that time, I was also looking into having profile based policy for determining those values. 

If you think global state would be messy, I can switch to computing `MlasTMACKernelParams` as needed, and switch back if we decide to have profile based policy

onnxruntime/contrib_ops/cpu/quantization/matmul_nbits_impl.cc
vraspar: I removed the function and other stubs that are not needed

cmake/onnxruntime_mlas.cmake
```diff
@@ -209,6 +211,7 @@ function(setup_mlas_source_for_windows)
       ${MLAS_SRC_DIR}/qgemm_kernel_sse.cpp
       ${MLAS_SRC_DIR}/qgemm_kernel_sse41.cpp
       ${MLAS_SRC_DIR}/intrinsics/avx512/quantize_avx512f.cpp
+      ${MLAS_SRC_DIR}/sqnbitgemm_lut_kernel_avx2.cpp
```
hariharans29, line 215: Should the .h file be included here ?

cmake/onnxruntime_mlas.cmake
```diff
@@ -693,6 +696,7 @@ else()
           ${MLAS_SRC_DIR}/intrinsics/avx2/qdwconv_avx2.cpp
           ${MLAS_SRC_DIR}/intrinsics/avx2/saturation_check_avx2.cpp
           ${MLAS_SRC_DIR}/sqnbitgemm_kernel_avx2.cpp
+          ${MLAS_SRC_DIR}/sqnbitgemm_lut_kernel_avx2.cpp
```
hariharans29, line 701: Ditto

onnxruntime/contrib_ops/cpu/quantization/matmul_nbits.cc
```diff
@@ -179,22 +194,75 @@ Status MatMulNBits<T1>::PrePack(const Tensor& tensor, int input_idx, /*out*/ All
...
+  std::unique_ptr<concurrency::ThreadPool> temp_threadpool;
+  concurrency::ThreadPool* threadpool_ptr = nullptr;
+
+  // Only create threadpool for operations that can benefit from it
+  if (prefer_lut_gemm_ || compute_type_ == SQNBIT_CompInt8) {
+    OrtThreadPoolParams tpo;
+    tpo.thread_pool_size = 4;    // Use default (typically number of cores)
```
hariharans29: Maybe actually use physical number of cores available (or add a TODO to the same effect) ?

onnxruntime/contrib_ops/cpu/quantization/matmul_nbits.cc
```diff
@@ -179,22 +194,75 @@ Status MatMulNBits<T1>::PrePack(const Tensor& tensor, int input_idx, /*out*/ All
...
+    } else {
+      packed_b_size_ = MlasQNBitGemmPackQuantBDataSize(N_, K_, nbits_, block_size_, has_zp_input_, compute_type_);
+      if (packed_b_size_ == 0) {
+        return Status::OK();
+      }
+      auto qptr = tensor.DataRaw();
+      auto scale_ptr = scales ? scales->DataRaw() : nullptr;
+      packed_b_ = IAllocator::MakeUniquePtr<void>(alloc, packed_b_size_, true);
+      MlasQNBitGemmPackQuantBData(N_, K_, nbits_, block_size_, compute_type_, qptr, packed_b_.get(), scale_ptr,
+                                  has_zp_input_, nullptr, threadpool_ptr);
```
hariharans29, line 265: IIUC - The usage of threadpool in the existing non-LUT path seems like a new addition - is that intentaional (and come with apprioriate tests) ?

onnxruntime/contrib_ops/cpu/quantization/matmul_nbits.cc
```diff
@@ -179,22 +194,75 @@ Status MatMulNBits<T1>::PrePack(const Tensor& tensor, int input_idx, /*out*/ All
...
+  concurrency::ThreadPool* threadpool_ptr = nullptr;
+
+  // Only create threadpool for operations that can benefit from it
+  if (prefer_lut_gemm_ || compute_type_ == SQNBIT_CompInt8) {
+    OrtThreadPoolParams tpo;
+    tpo.thread_pool_size = 4;    // Use default (typically number of cores)
+    tpo.allow_spinning = false;  // Don't spin during model load
+    tpo.auto_set_affinity = false;
+
+    temp_threadpool = concurrency::CreateThreadPool(
```
hariharans29, line 214: Could we add matmulNBits level kernel tests in addition to the MLAS library tests ?

onnxruntime/python/onnxruntime_pybind_quant.cc
```diff
@@ -134,6 +147,8 @@ void CreateQuantPybindModule(py::module& m) {
   m.def("quantize_matmul_8bits", &QuantizeMatMulNBitsBlockwise<MLFloat16, 8>);
   m.def("quantize_matmul_bnb4", &QuantizeMatMulBnb4Blockwise<float>);
   m.def("quantize_matmul_bnb4", &QuantizeMatMulBnb4Blockwise<MLFloat16>);
+  m.def("quantize_qdq_matmul_2bits", &QuantizeQDQMatMulNBitsBlockwise<float, 2>);
```
hariharans29, line 150: Can Python tests be added to exercise these new routines (and test them) ?

cmake/onnxruntime_mlas.cmake
vraspar: Thanks, I fixed it

onnxruntime/contrib_ops/cpu/quantization/matmul_nbits.cc
vraspar: Thanks! I started using physical core

onnxruntime/contrib_ops/cpu/quantization/matmul_nbits.cc
vraspar: Initially, I thought tests in `test_sqnbitgemm.cpp` should suffice since they already test it with thread pool. I applied changes to only use thread pool for LUT path now. 

Once we add tests, I think it might be beneficial to use thread pool for pre packing for other paths

onnxruntime/core/mlas/lib/sqnbitgemm_lut_kernel_avx2.cpp
```diff
@@ -0,0 +1,671 @@
...
+{
+    __m128 res = _mm256_extractf128_ps(v, 1);
+    res = _mm_add_ps(res, _mm256_castps256_ps128(v));
+    res = _mm_add_ps(res, _mm_movehl_ps(res, res));
+    res = _mm_add_ss(res, _mm_movehdup_ps(res));
+    return _mm_cvtss_f32(res);
+}
+
+// Conditional pragma unroll for compiler compatibility
+#if defined(__INTEL_COMPILER) || defined(__clang__)
```
hariharans29, line 43: Why is this complier dependent ? Is this implementation from the T-MAC library as is ?

onnxruntime/core/mlas/lib/qlutgemm.cpp
```diff
@@ -0,0 +1,686 @@
...
+
+    const auto* a_float = reinterpret_cast<const float*>(A);  // Activation data
+
+    // const int num_groups = static_cast<int>(K / BlkLen);
+
+    // Parallelize over M (batch dimension)
+    // Each iteration processes one row of the activation matrix
+    // TODO(vraspar): Ideally we have to do block parallelism here
+
+    MlasTrySimpleParallel(
```
hariharans29, line 555: If M == 1, can we parallelize on N ?

hariharans29 (approved):

onnxruntime/core/mlas/lib/qlutgemm.cpp
```diff
@@ -0,0 +1,686 @@
...
+LutGemmPackQuantBDataSize(
+    size_t N,
+    size_t K,
+    size_t BlkBitWidth,
+    size_t BlkLen,
+    bool HasZeroPoint
+)
+{
+    const MlasTMACKernelParams& tmac_params = MlasGetLutGemmKernelParams(N, K, BlkBitWidth, BlkLen, HasZeroPoint);
+    const size_t PackedQuantBDataSize = (N * BlkBitWidth) * (K / tmac_params.g / tmac_params.ngroups_per_elem);
```
hariharans29: Can you please explain what num_groups_per_elem is  ?

onnxruntime/core/mlas/lib/qlutgemm.cpp
```diff
@@ -0,0 +1,686 @@
...
+    if (BlkBitWidth != 2 || BlkLen == 0 || (BlkLen % 32) != 0) {
+        return false;
+    }
+
+    if (K % 32 != 0) {
+        return false;
+    }
+
+    size_t n_div = 0;
+    switch (BlkBitWidth) {
```
hariharans29, line 448: Why have this switch if BlkBitWidth is guaranteed to be 2 at this stage ?

onnxruntime/contrib_ops/cpu/quantization/matmul_nbits.cc
jambayk: closing comment for now to merge as discussed offline

onnxruntime/core/mlas/lib/qlutgemm.cpp
vraspar: I decided to have it generalized for when we add int4 kernels

onnxruntime/core/mlas/lib/qlutgemm.h
```diff
@@ -0,0 +1,84 @@
...
+#pragma once
+
+#include "mlas_qnbit.h"
+#include "mlasi.h"
+
+/**
+ * @brief Parameters for TMAC kernel
+ */
+struct MlasTMACKernelParams {
+    size_t g;
```
hariharans29, line 26: A brief comment describing what each config is and what it is used for will help

onnxruntime/core/mlas/lib/qlutgemm.cpp
hariharans29, line 139: Is there an alignment requirement for the packed weights ?

onnxruntime/core/mlas/lib/qlutgemm.cpp
```diff
@@ -0,0 +1,686 @@
...
+    const size_t kfactor = tmac_params.kfactor;
+
+    assert(BlkLen % g == 0);
+    assert((BlkLen / g) % kfactor == 0);
+
+    const size_t mgroup = ngroups_per_elem * simd_n_in;  // 32
+    assert(bm % mgroup == 0);
+    assert(bm % bits == 0);
+
+    std::unique_ptr<uint8_t[]> buf(new uint8_t[N * bits * (K / g)]);
```
hariharans29, line 173: For the purpose of what is being done here, a standard RAII containers like vector would do, do we really need a unique_ptr here ?

PR #26696, opened 2025-12-02
[QNN EP] Default to HTP instead of using last provided device

qti-mattsinc:
### Description
When multiple devices are provided in `AppendExecutionProviders_V2`, default to the NPU device, instead of picking the last device in the list.

### DIFF
--- a/onnxruntime/core/providers/qnn/qnn_provider_factory.cc
+++ b/onnxruntime/core/providers/qnn/qnn_provider_factory.cc
@@ -7,6 +7,17 @@
 #include "core/providers/qnn/qnn_provider_factory_creator.h"
 #include "core/providers/qnn/qnn_execution_provider.h"
 #include "core/providers/qnn/builder/qnn_utils.h"
+#include "core/session/abi_devices.h"
+
+static const std::unordered_map<OrtHardwareDeviceType, std::string> kDefaultBackends = {
+#if defined(_WIN32)
+    {OrtHardwareDeviceType_NPU, "QnnHtp.dll"},
+    {OrtHardwareDeviceType_GPU, "QnnGpu.dll"},
+#else
+    {OrtHardwareDeviceType_NPU, "libQnnHtp.so"},
+    {OrtHardwareDeviceType_GPU, "libQnnGpu.so"},
+#endif
+};
 
 namespace onnxruntime {
 struct QNNProviderFactory : IExecutionProviderFactory {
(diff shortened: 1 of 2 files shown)


qti-mattsinc:
Ready for MSFT review. This change was previously cherry-picked onto the branch for the 1.8.30 EP build and can be mainlined now.

@edgchen1 / others, PTAL

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
```diff
@@ -195,27 +252,24 @@ struct QnnEpFactory : OrtEpFactory {
...
-  // to pick the last specified backend only.
+  // to default to npu.
   static OrtStatus* GetSupportedDevicesImpl(OrtEpFactory* this_ptr,
                                             const OrtHardwareDevice* const* devices,
                                             size_t num_devices,
                                             OrtEpDevice** ep_devices,
                                             size_t max_ep_devices,
                                             size_t* p_num_ep_devices) noexcept {
     size_t& num_ep_devices = *p_num_ep_devices;
+    num_ep_devices = 0;
```
qti-mattsinc, line 263: The reason for this line is that the caller of GetSupportedDevices [0] does initialize `num_ep_devices` to 0, but it's unclear if the contract across the provider bridge [1] guarantees this behavior. So, we should initialize it to 0 explicitly just to be safe.

[0] https://github.com/microsoft/onnxruntime/blob/v1.23.2/onnxruntime/core/session/environment.cc#L776-L779
[1] https://github.com/microsoft/onnxruntime/pull/24568/files#diff-5845a5c76fb64abdc8f0cffe21b37f8da1712674eb3abc4cd87190891be1bd48R6037

edgchen1: right, there is no guarantee that the caller will initialize `*p_num_ep_devices` to 0. it is an output parameter.

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
```diff
@@ -80,13 +120,61 @@ struct QNN_Provider : Provider {
...
+            LOGS_DEFAULT(WARNING)
+                << "QNN EP only supports one device. An NPU device was not provided, so only the GPU device will be used.";
+            device_to_use = *device_it;
+          } else {
+            return Status(common::ONNXRUNTIME, ORT_EP_FAIL,
+                          "Multiple devices were provided to QNN EP, but neither an NPU nor a GPU was included.");
+          }
+        }
+      }
+      ORT_RETURN_IF(device_to_use == nullptr, "Failed to select device for QNN EP!");
```
edgchen1, line 160: nit: I don't believe the condition `device_to_use == nullptr` will ever be true, given the current logic above. perhaps this selection logic can be simplified.

qti-mattsinc: Right, it's a redundant check at the moment, but having it here is defensive in case the above logic changes. I was debating including "(this should never happen)" in the log message for that and can add it if you think it would helpful for code-clarity.

edgchen1: I see, maybe `assert` or `ORT_ENFORCE` would be more appropriate for this kind of programming logic check

qti-mattsinc: Got it. In the interest of not having to re-run CI, we can keep it as `ORT_RETURN_IF` for now, although I do like `assert` better. Thanks for bringing it up

PR #26706, opened 2025-12-02
[NV TRT RTX] Set AllocStream to nullptr to omit undefined usage

gedoensmax:
@yuslepukhin am i missing somthing or is there some kind of bridge missing to safely convert the OrtSyncStream pointer ?

### DIFF
--- a/onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc
+++ b/onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc
@@ -2,19 +2,22 @@
 // SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 // Licensed under the MIT License.
 
-#include "core/providers/shared_library/provider_api.h"
-#include "nv_provider_factory.h"
+#include <string.h>
 #include <atomic>
+
+#include "core/providers/shared_library/provider_api.h"
+#include "core/framework/provider_options.h"
+#include "core/framework/plugin_ep_stream.h"
+#include "core/providers/nv_tensorrt_rtx/nv_provider_options.h"
+#include "core/providers/nv_tensorrt_rtx/nv_execution_provider_custom_ops.h"
+#include "core/providers/cuda/shared_inc/cuda_call.h"
+#include "core/providers/cuda/cuda_stream_handle.h"
+
+#include "nv_provider_factory.h"
 #include "nv_execution_provider.h"
 #include "nv_provider_factory_creator.h"
 #include "nv_data_transfer.h"
 #include "nv_allocator.h"
-#include "core/framework/provider_options.h"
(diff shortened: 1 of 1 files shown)


onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc
```diff
@@ -223,6 +224,13 @@ struct NvTrtRtxOrtAllocator : OrtAllocator {
     return impl.allocator_->Alloc(size);
   }
 
+  // TODO missing a bridge for this
```
yuslepukhin: OrtSyncStream is directly inherited from Stream so no cast should be necessary.

The definition is in core/framework/plugin_ep_stream.h

I think it would worked if you included the header.

gedoensmax: If i include the header it has many conflicts with the logging header since it does not use the provider wrapped types, but the private logging inside ORT main lib. I tried using the provider wrapped types and using OrtLogger instead of logger but no success so far. Can you help me understand how i can achieve a clear separation of the compile units here ? Do i really need to split all logging messages into the `.cc` file ?

gedoensmax: I implemented with a separation to .cc but let me know if there is a nicer way.

onnxruntime/core/framework/plugin_ep_stream.cc
```diff
@@ -3,10 +3,40 @@
...
+}
+void Notification::WaitNotificationOnHost(onnxruntime::Stream*, synchronize::Notification& notification) {
+  auto* this_ptr = static_cast<Notification*>(&notification);
+  LOG_AND_RETURN_IF_ORT_ERROR(this_ptr->impl_.WaitOnHost(&this_ptr->impl_), *this_ptr->logger_.ToInternal());
+}
+void Notification::Activate() {
+  LOG_AND_RETURN_IF_ORT_ERROR(impl_.Activate(&impl_), *logger_.ToInternal());
+}
+void Stream::Flush() {
+  LOG_AND_RETURN_IF_ORT_ERROR(impl_.Flush(&impl_), *logger_.ToInternal());
```
adrianlizarraga, line 38: What is the reason for updating the `logger_` member to be a `OrtLogger`? Maybe I missed it, but it seems like we're always just converting it back to a `logging::Logger` on use via `ToInternal()`.

gedoensmax: I did that to not have to forward declare the internal logger in the header. If needed i can also forward declare it and store a `Logger*` in the class.

adrianlizarraga: I see. I was missing the context from your previous comment here: https://github.com/microsoft/onnxruntime/pull/26706#discussion_r2584288534

I think this should be ok.

adrianlizarraga (approved):

PR #26711, opened 2025-12-03
[EP ABI] Update Graph_GetGraphView() implementation

chilo-ms:
### Description

This PR mainly modifies the followings:

- Update `Graph_GetGraphView()` implementation.
- Make sure EpGraph maintains the min/max node index, so that when querying node outside that range, it can return null.
- Provide option to create an EpGraph that contains its parent node when the graph is the subgraph of a control flow op.

#### Update Graph_GetGraphView() implementation
In some cases, e.g. when model has a node that produces output consumed by multiple nodes, calling the current implementation of `Graph_GetGraphView()` to get a subgraph returns incorrect `OrtGraph.`

- Original graph:

  image

- Incorrect graph after calling `Graph_GetGraphView()` to get the subgraph:

  It includes three of the nodes from the original graph.
  The `topk_indices` is the output of the `TopK` and it shouldn't be added as a graph input shown in the graph below.
  The API implementation has issue handling this case.
  If we feed this subgraph into TRT parser, it would fail to parse the graph.
  
  image

- Correct graph after calling `Graph_GetGraphView()` to get the subgraph:

  It includes three of the nodes from the original graph.
  The `topk_indices` now is not added as a graph input. Instead, the `topk_indices` is added as a graph output which is expected as `Mod` is in another subgraph that consumes it, so this subgraph has to make `topk_indices` a graph output.

  image

### DIFF
--- a/onnxruntime/core/session/onnxruntime_c_api.cc
+++ b/onnxruntime/core/session/onnxruntime_c_api.cc
@@ -2893,6 +2893,10 @@ ORT_API_STATUS_IMPL(OrtApis::Graph_GetGraphView, _In_ const OrtGraph* src_graph,
     indexed_sub_graph->nodes.push_back(node.Index());
 
     for (const auto& input : node.InputDefs()) {
+      if (!input->Exists()) {
+        continue;
+      }
+
       if (graph_viewer.IsConstantInitializer(input->Name(), true)) {
         initializers.push_back(input->Name());
         continue;
@@ -2908,6 +2912,10 @@ ORT_API_STATUS_IMPL(OrtApis::Graph_GetGraphView, _In_ const OrtGraph* src_graph,
     }
 
     for (const auto& input : node.ImplicitInputDefs()) {
+      if (!input->Exists()) {
+        continue;
+      }
+
       if (graph_viewer.IsConstantInitializer(input->Name(), true)) {
         initializers.push_back(input->Name());
         continue;
@@ -2929,6 +2937,10 @@ ORT_API_STATUS_IMPL(OrtApis::Graph_GetGraphView, _In_ const OrtGraph* src_graph,
     // if the output is connected to nodes that don't belong to the subgraph, the output need to be added
     // to the output list
     for (const auto& output : node.OutputDefs()) {
+      if (!output->Exists()) {
+        continue;
+      }
+
       const auto& it = subgraph_inputs.find(output);
       if (it != subgraph_inputs.end()) {
         subgraph_inputs.erase(it);


onnxruntime/core/session/onnxruntime_c_api.cc
```diff
@@ -2818,72 +2818,203 @@ ORT_API_STATUS_IMPL(OrtApis::Graph_GetGraphView, _In_ const OrtGraph* src_graph,
...
+
+  // This is the relative ordering that ensures node's input or output being added to the 'subgraph_inputs',
+  // 'subgraph_outputs', 'subgraph_outputs_to_add' and 'graph_outputs_to_add' maps is associated with a relative order index.
+  // Items added earlier receive a smaller order index than items added later.
+  // When constructing the final subgraph's input or output lists, entries with smaller
+  // order indices will appear before those with larger indices.
+  int input_order = 0;
+  int output_order = 0;
+
+  std::vector<std::string> initializers;
```
adrianlizarraga, line 2882: nit: Could these be `std::string_view` to avoid copies? Would probably need to convert to `std::string` only when pushing into `meta_def->constant_initializers.` (or keep it as string and do a move).

onnxruntime/core/session/onnxruntime_c_api.cc
```diff
@@ -2818,72 +2818,228 @@ ORT_API_STATUS_IMPL(OrtApis::Graph_GetGraphView, _In_ const OrtGraph* src_graph,
...
+    const EpNode* ep_node = EpNode::ToInternal(ort_node);
+    if (ep_node == nullptr) {
+      return OrtApis::CreateStatus(OrtErrorCode::ORT_INVALID_ARGUMENT,
+                                   "node is a ModelEditorNode which doesn't support Graph_GetGraphView.");
+    }
+    const Node& node = ep_node->GetInternalNode();
+    indexed_sub_graph->nodes.push_back(node.Index());
+
+    for (const auto& input : node.InputDefs()) {
+      if (graph_viewer.IsConstantInitializer(input->Name(), true)) {
```
adrianlizarraga, line 2900: Do we need to skip this input if `!input->Exists()` (i.e., for missing optional inputs).

onnxruntime/core/session/onnxruntime_c_api.cc
```diff
@@ -2818,72 +2818,228 @@ ORT_API_STATUS_IMPL(OrtApis::Graph_GetGraphView, _In_ const OrtGraph* src_graph,
...
+  // 'subgraph_outputs', 'subgraph_outputs_to_add' and 'graph_outputs_to_add' maps is associated with a relative order index.
+  // Items added earlier receive a smaller order index than items added later.
+  // When constructing the final subgraph's input or output lists, entries with smaller
+  // order indices will appear before those with larger indices.
+  int input_order = 0;
+  int output_order = 0;
+
+#if defined(ORT_MINIMAL_BUILD) && !defined(ORT_EXTENDED_MINIMAL_BUILD)
+  // node arg to its consumer nodes
+  // It's only used in ORT_MINIMAL_BUILD, otherwise we use GetConsumerNodes() instead.
```
edgchen1: it's a little unusual to have an alternate implementation for a minimal build instead of just disabling it. there's more limited testing of minimal builds so I'm a bit wary about test coverage for the alternate implementation. do we need this function to be implemented in a minimal build? if so, consider using a unified implementation that works in both minimal and non-minimal builds.

onnxruntime/core/session/onnxruntime_c_api.cc
```diff
@@ -2818,72 +2818,228 @@ ORT_API_STATUS_IMPL(OrtApis::Graph_GetGraphView, _In_ const OrtGraph* src_graph,
...
+  // Get the input order of the original graph
+  std::unordered_map<const NodeArg*, int> original_inputs;
+  int order = 0;
+  for (const auto* input : graph_viewer.GetInputs()) {
+    original_inputs[input] = order++;
+  }
+
+  // input order needs to be consistent with original graph's input order
+  for (auto it = subgraph_inputs.begin(), end = subgraph_inputs.end(); it != end; ++it) {
+    const auto& iter = original_inputs.find(it->first);
```
edgchen1: can the names `it` and `iter` be more descriptive?

onnxruntime/core/graph/ep_api_types.h
```diff
@@ -269,8 +270,16 @@ struct EpGraph : public OrtGraph {
   /// </summary>
   /// <param name="graph_viewer"></param>
   /// <param name="result"></param>
+  /// <param name="create_parent_node">If the `graph_viewer` is a subgraph of a control flow op,
+  ///                                  e.g. For/If/Scan op, and `create_parent_node` is set to true,
```
edgchen1:
```suggestion
  ///                                  e.g. Loop/If/Scan op, and `create_parent_node` is set to true,
```

onnxruntime/core/graph/ep_api_types.h
```diff
@@ -364,16 +373,21 @@ struct EpGraph : public OrtGraph {
...
   /// <returns></returns>
-  static Status CreateImpl(std::unique_ptr<EpGraph> ep_graph, const GraphViewer& graph_viewer, /*out*/ std::unique_ptr<EpGraph>& result);
+  static Status CreateImpl(std::unique_ptr<EpGraph> ep_graph, const GraphViewer& graph_viewer,
+                           /*out*/ std::unique_ptr<EpGraph>& result, bool create_parent_node = false);
 
   const GraphViewer& graph_viewer_;
-  const EpNode* parent_node_ = nullptr;
+  const EpNode* parent_node_ = nullptr;                  // Keep the pointer to the parent node that
+                                                         // is not owned by this graph
+  std::unique_ptr<EpNode> parent_node_owned_ = nullptr;  // Hold the parent node created and owned by this graph
```
edgchen1: what's the relation between `parent_node_` and `parent_node_owned_`? does `EpGraph::SetParentNode()` need to account for `parent_node_owned_`?

chilo-ms: `parent_node_` is set to the created EpNode for the subgraph by calling `EpGraph::SetParentNode()` when the graph traversal starts from the top-level graph to the inner-most subgraph.

However, ORT's GetCapability() starts from inner-most subgraph to the top-level graph, so when constructing the inner-most subgraph it needs to additionally create the `parent_node_owned_` and holds it.

chilo-ms: EpGraph::SetParentNode() doesn't need to account for `parent_node_owned_`, because `parent_node_` has higher priority than `parent_node_owned_` when calling `EpGraph::GetParentNode`

onnxruntime/core/session/onnxruntime_c_api.cc
chilo-ms: I updated to use the unified implementation across all the builds.

onnxruntime/core/graph/ep_api_types.cc
```diff
@@ -539,10 +547,14 @@ void EpGraph::IndexToEpNodeMap::Resize(NodeIndex min_node_index, NodeIndex max_n
   size_t num_elems = (max_node_index - min_node_index) + 1;
 
   min_node_index_ = min_node_index;
+  max_node_index_ = max_node_index;
```
edgchen1: is an additional `max_node_index_` data member necessary? it seems like the value can be derived from `min_node_index_` and `nodes_.size()`

onnxruntime/core/graph/ep_api_types.h
edgchen1: I was asking about the relation more in terms of class invariants. should `parent_node_` and `parent_node_owned_` point to the same node? or should at most one of them be set at a time? or if both are set and they point to different nodes, what does that mean?

include/onnxruntime/core/providers/utils/ort_graph_to_proto.h
```diff
@@ -413,6 +417,13 @@ Ort::Status OrtGraphToProto(const OrtGraph& graph,
       ORT_EP_UTILS_CXX_RETURN_IF_ERROR(OrtValueInfoToProto(value_info, *value_info_proto));
     }
 
+    // There might be some initializers in the original OrtGraph that are not added yet.
```
adrianlizarraga: qq: Just curious. Would they be missing because the initializers are not used in the current graph? Maybe they're used in a nested graph?

onnxruntime/core/graph/ep_api_types.cc
```diff
@@ -687,13 +700,31 @@ Status EpGraph::CreateImpl(std::unique_ptr<EpGraph> ep_graph, const GraphViewer&
...
+    // If the subgraph of a control-flow op is created before its parent node (for example, when constructing
+    // the graph during ORT's GetCapability() in a bottom-up manner), the parent node must also be created.
+    if (create_parent_node) {
+      std::unique_ptr<EpNode> ep_node = nullptr;
+
+      std::unordered_map<std::string, std::unique_ptr<EpValueInfo>> value_infos_map_tmp;  // won't be used
+
+      // At this point, the EpGraph that contains the parent node hasn't been created yet.
+      // It's not needed to create that EpGraph here, so just pass nullptr.
+      ORT_RETURN_IF_ERROR(EpNode::Create(*parent_node, /*ep_graph*/ nullptr, value_infos_map_tmp, ep_node));
```
adrianlizarraga: I believe that passing in a `value_infos_map_tmp` that is not stored somewhere will create issues if the EP tries to get the inputs/outputs of the parent node. The EpNode only stores pointers to EpValueInfos for its inputs and outputs. The true owner of the EpValueInfos is the `value_infos_map_tmp`. However, those EpValueInfos will be immediately deleted as soon as `value_infos_map_tmp` is destroyed.

If the EP tries to get the inputs/outputs, they'll get pointers that point to freed memory.

include/onnxruntime/core/providers/utils/ort_graph_to_proto.h
chilo-ms: The initializer is not used by any of the nodes, but the initializer is the graph's output and that's the only usage of that initializer.

So iterating through all the nodes to get  nitializer_value_info is not enough, need to get initializers from ort_graph.GetInitializers() as well.

onnxruntime/core/graph/ep_api_types.cc
chilo-ms: Good catch, i changed the code to add `EpGraph::parent_node_value_infos_map_` to cache the value_infos for parent node.

onnxruntime/core/graph/ep_api_types.h
chilo-ms: i make `parent_node_` and `parent_node_owned_` be mutually exclusive and will never both be set.

include/onnxruntime/core/providers/utils/ort_graph_to_proto.h
adrianlizarraga: Got it thanks.

onnxruntime/core/graph/ep_api_types.cc
adrianlizarraga: Thanks. The user will not be able to use the parent node's inputs/outputs to traverse the parent graph, but that's ok for now.

I think that proper longer-term solution would be to completely create the entire model's OrtGraph chain before starting partitioning so that we don't have to dynamically create things like parent nodes. But ... that's for later.

onnxruntime/core/session/onnxruntime_c_api.cc
adrianlizarraga: bump on this one. It could potentially lead to a bug if the input does not actually exist.

include/onnxruntime/core/providers/utils/ort_graph_to_proto.h
edgchen1: > The initializer is not used by any of the nodes, but the initializer is the graph's output and that's the only usage of that initializer.

this may be good info to include in a comment here.

include/onnxruntime/core/providers/utils/ort_graph_to_proto.h
```diff
@@ -320,6 +321,9 @@ Ort::Status OrtGraphToProto(const OrtGraph& graph,
       // For values defined in an outer scope, just add the value info but not the initializer.
```
edgchen1: "For values defined in an outer scope, just add the value info but not the initializer."

does this comment accurately reflect the new setup?

include/onnxruntime/core/providers/utils/ort_graph_to_proto.h
```diff
@@ -492,8 +503,6 @@ Ort::Status OrtGraphToProto(const OrtGraph& graph,
   try {
     // Check that OrtGraph is a top-level graph (no parent node).
```
edgchen1: "Check that OrtGraph is a top-level graph (no parent node)."

should this comment be updated? looks like this check is no longer done.

onnxruntime/core/graph/ep_api_types.cc
chilo-ms: good point, i removed `max_node_index_` and use `min_node_index` and `node.size() `to get the maximum node index.

onnxruntime/core/graph/ep_api_types.cc
```diff
@@ -872,11 +903,21 @@ Status EpGraph::GetNodes(gsl::span<const OrtNode*> dst) const {
...
+  }
+
   return Status::OK();
 }
 
-void EpGraph::SetParentNode(const EpNode* node) { parent_node_ = node; }
+void EpGraph::SetParentNode(const EpNode* node) {
+  parent_node_ = node;
+  parent_node_owned_ = nullptr;
+}
```
edgchen1, line 916: it looks like `parent_node_value_infos_map_` is associated with `parent_node_owned_`. should that be cleared too?

onnxruntime/core/graph/ep_api_types.h
```diff
@@ -364,17 +375,33 @@ struct EpGraph : public OrtGraph {
...
+  /// <param name="create_parent_node"></param>
   /// <returns></returns>
-  static Status CreateImpl(std::unique_ptr<EpGraph> ep_graph, const GraphViewer& graph_viewer, /*out*/ std::unique_ptr<EpGraph>& result);
+  static Status CreateImpl(std::unique_ptr<EpGraph> ep_graph, const GraphViewer& graph_viewer,
+                           /*out*/ std::unique_ptr<EpGraph>& result, bool create_parent_node = false);
 
   const GraphViewer& graph_viewer_;
+
+  // parent_node_ and parent_node_owned_ are mutually exclusive and will never both be set.
+  // One of the following cases applies:
```
edgchen1: thanks for documenting this.

one other approach to consider - `parent_node_` could always point to the parent node, if one is set. this could be the same node as `parent_node_owned_`. `EpGraph::GetParentNode()` could simply check `parent_node_`.

alternatively, to make the mutual exclusivity clearer, consider renaming `parent_node_` to `parent_node_unowned_`.

onnxruntime/core/graph/graph_viewer.cc
```diff
@@ -203,14 +203,16 @@ GraphViewer::GraphViewer(const Graph& graph, const IndexedSubGraph* filter_info)
       ORT_ENFORCE(node, "Mismatch between Graph and IndexedSubGraph. Node not found: ", node_idx);
       const ONNX_NAMESPACE::TensorProto* tensor = nullptr;
       for (const auto* node_input : node->InputDefs()) {
-        if (graph.GetInitializedTensor(node_input->Name(), tensor)) {
+        tensor = graph.GetInitializer(node_input->Name(), true);
```
edgchen1: `GraphViewer::GetInitializedTensor()` also makes a call to `Graph::GetInitializedTensor()` after checking `filtered_initializers_`. would that need to be updated to account for parent graphs too?

include/onnxruntime/core/providers/utils/ort_graph_to_proto.h
adrianlizarraga: It can probably be removed. I think we can allow converting a `OrtGraph` to a `onnx::ModelProto` even if that graph is not a top-level graph.

onnxruntime/core/graph/graph_viewer.cc
adrianlizarraga: This is now changing GraphViewer so that it stores outer-scope initializers, which could have unexpected effects. Is this change necessary given that there's logic in `EpGraph::Create()` to add initializers from the outer scope here:
https://github.com/microsoft/onnxruntime/blob/e7dfd69dce15ae44e6d64b6566af67cf8ef16ffa/onnxruntime/core/graph/ep_api_types.cc#L690-L710

If the above is not working for some reason, can it be fixed locally in there?

onnxruntime/core/session/onnxruntime_c_api.cc
```diff
@@ -2818,72 +2818,227 @@ ORT_API_STATUS_IMPL(OrtApis::Graph_GetGraphView, _In_ const OrtGraph* src_graph,
...
-        } else {
-          // consumed by node so no longer subgraph output
-          // NOTE: Ignoring edge case where a node output is an overall graph output AND a node input
-          outputs.erase(def->Name());
+
+  // Following data structures help determine the final inputs/outputs of the subgraph.
+  // Note: The 'subgraph' here refers to a graph contains a subset of nodes in the 'src_graph'.
+
+  // Subgraph's node set
+  std::unordered_set<size_t> node_set(num_nodes);
```
edgchen1: general - good to make things const when possible. makes it easier to reason about (for readers of the code too), especially in this context, where there are some other dynamically updated data structures in scope.

non-trivial initialization can be done with a lambda, e.g.:
```c++
const auto x = [&](){
  std::unordered_set<size_t> x{};
  // set up x...

  return x;
}();
```

onnxruntime/core/session/onnxruntime_c_api.cc
```diff
@@ -2818,72 +2818,227 @@ ORT_API_STATUS_IMPL(OrtApis::Graph_GetGraphView, _In_ const OrtGraph* src_graph,
...
+    indexed_sub_graph->nodes.push_back(node.Index());
+
+    for (const auto& input : node.InputDefs()) {
+      if (graph_viewer.IsConstantInitializer(input->Name(), true)) {
+        initializers.push_back(input->Name());
+        continue;
+      }
+      const auto& it = subgraph_outputs.find(input);
+      if (it != subgraph_outputs.end()) {
+        subgraph_outputs.erase(it);
```
adrianlizarraga, line 2906: If I'm not mistaken, this assumes that a node input cannot also be a (sub)graph output. However, I think it is valid to have a graph where one of its outputs is also an input to a node.

onnxruntime/core/session/onnxruntime_c_api.cc
```diff
@@ -2818,72 +2818,227 @@ ORT_API_STATUS_IMPL(OrtApis::Graph_GetGraphView, _In_ const OrtGraph* src_graph,
...
-    return OrtApis::CreateStatus(OrtErrorCode::ORT_INVALID_ARGUMENT, "src_graph is a ModelEditorGraph which doesn't support Graph_GetSubGraph.");
+    return OrtApis::CreateStatus(OrtErrorCode::ORT_INVALID_ARGUMENT,
+                                 "src_graph is a ModelEditorGraph which doesn't support Graph_GetGraphView.");
   }
-  const Graph& graph = ep_graph->GetGraphViewer().GetGraph();
+  const GraphViewer& graph_viewer = ep_graph->GetGraphViewer();
+  const Graph& graph = graph_viewer.GetGraph();
 
   // Create a GraphViewer with filtered info
   std::unique_ptr<IndexedSubGraph> indexed_sub_graph = std::make_unique<IndexedSubGraph>();
```
edgchen1, line 2829: could we add a TODO to investigate whether `utils::MakeComputeCapability` can be extended and reused instead, as @adrianlizarraga suggested [here](https://github.com/microsoft/onnxruntime/pull/26711#pullrequestreview-3537268034)?

onnxruntime/core/graph/graph_viewer.cc
chilo-ms: I later checked that as long as the `ort_graph_to_proto.h` has following code to include outer-scope initializer value info which is added in this PR:
````c++
     ... 
      if (is_from_outer_scope) {
        value_infos.emplace(value_name, ort_value_info);
        if (is_constant_initializer) {
          initializer_value_infos.emplace(value_name, ort_value_info);
        }
````
and EpGraph::Create() did add initializers from the outer scope as Adrian mentioned above, then the outer-scope initializer TensorProto will be added to the GraphProto.

The change in graph_viewer.cc is not necessary and i removed them.

File: onnxruntime/core/session/onnxruntime_c_api.cc (hunk 17 above)
Comment (chilo-ms) [reply onnxruntime/core/session/onnxruntime_c_api.cc:2829]: added

File: onnxruntime/core/session/onnxruntime_c_api.cc (hunk 16 above)
Comment (chilo-ms) [reply onnxruntime/core/session/onnxruntime_c_api.cc:2906]: As we discussed offline, the case you mentioned above will be covered later in this implementation that if the node's output is consumed by the node outside of this subgraph, this node's output will be added to subgraph's outputs.

File: onnxruntime/core/session/onnxruntime_c_api.cc (hunk 15 above)
Comment (chilo-ms) [reply onnxruntime/core/session/onnxruntime_c_api.cc]: Good suggestion, i used a lambda instead and use const as well.

File: onnxruntime/core/graph/ep_api_types.h (hunk 13 above)
Comment (chilo-ms) [reply onnxruntime/core/graph/ep_api_types.h]: the approach you suggested is much better and i updated to that.
i.e. - `parent_node_` could always point to the parent node, if one is set. this could be the same node as `parent_node_owned_`. `EpGraph::GetParentNode()` could simply check `parent_node_`.

File: onnxruntime/core/graph/ep_api_types.cc (hunk 12 above)
Comment (chilo-ms) [reply onnxruntime/core/graph/ep_api_types.cc:916]: added

File: include/onnxruntime/core/providers/utils/ort_graph_to_proto.h (hunk 10 above)
Comment (chilo-ms) [reply include/onnxruntime/core/providers/utils/ort_graph_to_proto.h]: removed.

File: onnxruntime/core/session/onnxruntime_c_api.cc (hunk 2 above)
Comment (chilo-ms) [reply onnxruntime/core/session/onnxruntime_c_api.cc:2900]: good catch, added. Thanks.

File: include/onnxruntime/core/providers/utils/ort_graph_to_proto.h (hunk 11 above)
Comment (chilo-ms) [reply include/onnxruntime/core/providers/utils/ort_graph_to_proto.h]: reomved.

File: onnxruntime/core/graph/ep_api_types.cc [hunk 18]
```diff
@@ -543,6 +551,9 @@ void EpGraph::IndexToEpNodeMap::Resize(NodeIndex min_node_index, NodeIndex max_n
 }
 
 EpNode* EpGraph::IndexToEpNodeMap::GetEpNode(NodeIndex node_index) const {
+  if (node_index < min_node_index_ || node_index > (min_node_index_ + nodes_.size() - 1)) {
+    return nullptr;
+  }
   size_t i = node_index - min_node_index_;
   assert(i < nodes_.size());
```
edgchen1, line 558: nit: this assertion is now covered by the check above

onnxruntime/core/session/onnxruntime_c_api.cc
```diff
@@ -2818,72 +2818,242 @@ ORT_API_STATUS_IMPL(OrtApis::Graph_GetGraphView, _In_ const OrtGraph* src_graph,
...
+  // Note: graph.GetConsumerNodes() is not available in minimal build, in order to use unified implementation across
+  //       all builds, this map is needed to determine if node arg is consumed by other nodes.
+  std::unordered_map<std::string, std::unordered_set<NodeIndex>> node_arg_to_consumer_nodes;
+
+  std::vector<std::string> initializers;
 
   // Add nodes
-  for (size_t node_idx = 0; node_idx < num_nodes; node_idx++) {
-    const OrtNode* ort_node = nodes[node_idx];
+  for (size_t i = 0; i < num_nodes; i++) {
```
edgchen1, line 2885: does this algorithm rely on the nodes being provided in topological order, or does it handle an arbitrary order?

chilo-ms: This algorithm doesn't rely on the nodes being provided in topological order, nodes in arbitrary order can be handled.

edgchen1 (approved):

onnxruntime/core/graph/ep_api_types.cc
chilo-ms: will address this later

PR #26723, opened 2025-12-04
[webgpu] Add `CeilDiv` into webgpu utils

daijh:
### Description
The `CeilDiv` function, which is heavily used by operations to determine dispatch sizes, was duplicated in multiple files. This PR moves the implementation into the common WebGPU utils to ensure a single source and reduce code duplication.

### Motivation and Context
See above.

### DIFF
--- a/onnxruntime/core/providers/webgpu/webgpu_utils.h
+++ b/onnxruntime/core/providers/webgpu/webgpu_utils.h
@@ -13,12 +13,12 @@
 namespace onnxruntime {
 namespace webgpu {
 
+class ShaderVariableHelper;
+
 template <typename T>
 inline T CeilDiv(T numerator, T denominator) {
   return (numerator + denominator - 1) / denominator;
 }
-
-class ShaderVariableHelper;
 
 /**
  * Returns the maximum number of components `N` to be used as `vecN` for the given size.


daijh:
@guschmue @fs-eire @qjia7 PTAL.

onnxruntime/core/providers/webgpu/tensor/transpose.cc
```diff
@@ -155,9 +147,9 @@ Status Transpose::DoTranspose(onnxruntime::webgpu::ComputeContextBase& context,
     program.SetDispatchGroupSize(static_cast<uint32_t>((new_output_shape[1] + TILE_SIZE - 1) / TILE_SIZE),
                                  static_cast<uint32_t>(((new_output_shape[0] + TILE_SIZE - 1) / TILE_SIZE)));
   } else {
-    program.SetWorkgroupSize(WORKGROUP_SIZE);
+    program.SetWorkgroupSize(64u);
```
qjia7, line 148: nit: why not use `WORKGROUP_SIZE`?
In line 134, please help correct it to `uint32_t output_size = onnxruntime::narrow<uint32_t>(input_shape.Size());`

daijh: The reasons:
1. Specify workgroup_size=64 explicitly improve readability.
2. `WORKGROUP_SIZE` is defined as `SafeInt<uint32_t>`, it's not easy to cast back into `uint32_t` which is required by `CeilDiv`.

qjia7: It seems that it's not necessary to use SafeInt<uint32_t> for `WORKGROUP_SIZE`. `WORKGROUP_SIZE` is just a default value. If you are using 64 as the workgroup size, you can skip to call `program.SetWorkgroupSize(64u);` unless you are using a different value. Currently, a lot of ops are using the default `WORKGROUP_SIZE`. If you search, there should be many 'SetDispatchGroupSize((XXX + WORKGROUP_SIZE - 1) / WORKGROUP_SIZE). @fs-eire Do you think it's better to change the type of WORKGROUP_SIZE to uint32_t from SafeInt<uint32_t>? Or in all places, we explicitly specify the workgroup size.

daijh: I  personally prefer to explicitly specify the workgroup_size. This practice improves code readability and saves time for developers read the source first time, as it eliminates the need to navigate to another file to figure out the default WORKGROUP_SIZE; 

I have no objection if someone wishes to retain the implicit style.

onnxruntime/core/providers/webgpu/webgpu_utils.h
```diff
@@ -12,6 +12,11 @@
 
 namespace onnxruntime {
 namespace webgpu {
+template <typename T>
+
```
tianleiwu, line 15: remove empty line here.

daijh: Thanks for identifying it.

guschmue (approved):

PR #26724, opened 2025-12-04
webgpu: fix shader compiling error for HardSigmoid operator

xhcao:


xhcao:
When running `selfie_segmentation_general_fp16` and `selfie_segmentation_landscape_fp16` models, throw shader compiling error, is shown as below.
Error while parsing WGSL: :19:16 error: no matching constructor for 'vec4<f16>(f32)'

struct Uniforms {
  vec_size: u32,
  attr: vec2<f32>
};
@group(0) @binding(2) var<uniform> uniforms: Uniforms;

alias x_element_t = f16;


fn hard_sigmoid_v(v: vec4<x_element_t>) -> vec4<x_element_t> {
  let alpha = x_element_t(uniforms.attr[0]);
  let beta_v = vec4<x_element_t>(uniforms.attr[1]);
  return max(vec4<x_element_t>(0.0),
             min(vec4<x_element_t>(1.0), alpha * v + beta_v));
}

@fs-eire @guschmue @jchen10

jchen10:
Good catch!

guschmue (approved):

xenova:
Looks like the same issue is present in QuickGelu

PR #26725, opened 2025-12-04
[webgpu] Fix the check of `ValidateVariableDataType()` with Split-K

Jiawei-Shao:
### Description
This patch adds the support of `float16x4` and `float32x4` in `ValidateVariableDataType()` as we may declare them as atomic variables when Split-K is used.

### Motivation and Context
Previously we failed to discover this issue because `ValidateVariableDataType()` is only used in the Debug build.


Jiawei-Shao:
@qjia7 @xhcao PTAL, thanks!

Jiawei-Shao:
@guschmue @fs-eire PTAL, thanks!

PR #26742, opened 2025-12-07
TypeScript: Allow constructing float16 tensors using Float16Array

### LINKED ISSUE #26741: [Web] TypeScript definitions incorrectly forbid Float16Array
### Describe the issue

When trying to create tensors from Float16Array data, TS complains incorrectly that this type is not supported as an input, even though it is in fact accepted at runtime.

This is likely a result of it being a relatively new type, as in the past onnxruntime-web has only accepted Uint16Array representation of fp16 tensors.

### To reproduce

```ts
import * as ort from 'onnxruntime-web';

const data = new Float16Array(9);
const tensor = new ort.Tensor(data, [3, 3]);
```

Results in:

Image

### Urgency

_No response_

### ONNX Runtime Installation

Released Package

1.23.2

### Execution Provider

Other / Unknown

rockygeekz:
Fixes #26741

This change updates the TypeScript definitions to allow constructing `float16` tensors using `Float16Array` in environments where it is available. Runtime behavior remains unchanged (`float16` is still represented internally as `Uint16Array`).

- Introduces a `GlobalFloat16Array` helper type to safely detect `Float16Array` without requiring global polyfills.
- Adds type-specific and inferred constructor overloads for `float16`.
- No changes to runtime logic or public C APIs.

This resolves compile-time errors when passing `Float16Array` to the `Tensor` constructor in the `onnxruntime-web` package.

---

### **Description**

This PR enhances the TypeScript typings for `float16` tensors within the ONNX Runtime JavaScript API:

- Adds `GlobalFloat16Array`, a conditional utility type that resolves to the instance type of `Float16Array` only when available.
- Updates constructor definitions to accept either:
  - `Uint16Array` (existing behavior),
  - `Float16Array` (new behavior, when supported by the JS environment),
  - or `readonly number[]`.
- Extends inferred-type constructors to support `new Tensor(new Float16Array(...))`.
- Ensures TypeScript consumers can pass `Float16Array` without encountering type errors.

Internally, ONNX Runtime continues to treat `float16` data as `Uint16Array`, so runtime behavior is unchanged.

---

### **Motivation and Context**

Modern JavaScript runtimes (browsers and Node versions) have begun introducing native `Float16Array` support. Developers using ONNX Runtime in TypeScript projects may attempt to construct `float16` tensors using:

```ts
new Tensor(new Float16Array(784), [28, 28]);

### DIFF
--- a/js/common/lib/tensor.ts
+++ b/js/common/lib/tensor.ts
@@ -82,7 +82,7 @@ export declare namespace Tensor {
     int64: BigInt64Array;
     string: string[];
     bool: Uint8Array;
-    float16: Uint16Array | (TryGetGlobalType<'Float16Array'> extends { prototype: infer P } ? P : never);
+    float16: Uint16Array | TryGetGlobalType<'Float16Array', never>;
     float64: Float64Array;
     uint32: Uint32Array;
     uint64: BigUint64Array;

### CODE REVIEW

**Comment by rockygeekz** [comment]

**Comment by RReverser** [comment]
I think you could reuse the https://github.com/microsoft/onnxruntime/blob/790018d2fbcfcd1128352a7289c076ffa4c63120/js/common/lib/type-helper.ts#L27 helper instead of rolling custom check.

**Comment by rockygeekz** [comment]
Updated to use `TryGetGlobalType` for `Float16Array`.  
Let me know if you'd like any further adjustments!

File: js/common/lib/tensor.ts [hunk 1]
```diff
@@ -6,6 +6,10 @@ import { Tensor as TensorImpl } from './tensor-impl.js';
 import { TypedTensorUtils } from './tensor-utils.js';
 import { TryGetGlobalType } from './type-helper.js';
 
+// Helper type: resolves to the instance type of `Float16Array` if it exists in the global scope,
+// or `never` otherwise. Uses the shared TryGetGlobalType helper.
+export type GlobalFloat16Array = TryGetGlobalType<'Float16Array', never>;
```
RReverser: I don't think it needs to be exported / part of the public API, just a local helper.

```suggestion
type GlobalFloat16Array = TryGetGlobalType<'Float16Array', never>;
```

js/common/lib/tensor.ts
```diff
@@ -240,6 +244,19 @@ export interface TensorConstructor extends TensorFactory {
...
+  /**
+   * Construct a new float16 tensor object from the given type, data and dims.
+   *
+   * @param type - Specify the element type.
+   * @param data - Specify the CPU tensor data.
+   * @param dims - Specify the dimension of the tensor. If omitted, a 1-D tensor is assumed.
+   */
+  new (
+    type: 'float16',
+    data: Tensor.DataTypeMap['float16'] | GlobalFloat16Array | readonly number[],
```
RReverser: Actually, wouldn't it be easier to inline this helper into the DataTypeMap?

rockygeekz: Updated `DataTypeMap.float16` to inline `TryGetGlobalType<'Float16Array'>` and map it to the instance type via `prototype`, so `Float16Array` instances are accepted where supported without changing runtime behavior.

RReverser:
Thanks for working on this! Just to be clear, I'm not a maintainer on this repo, we'll need for one of them to review too.

js/common/lib/tensor.ts
```diff
@@ -82,7 +82,7 @@ export declare namespace Tensor {
     int64: BigInt64Array;
     string: string[];
     bool: Uint8Array;
-    float16: Uint16Array; // Keep using Uint16Array until we have a concrete solution for float 16.
+    float16: Uint16Array | (TryGetGlobalType<'Float16Array'> extends { prototype: infer P } ? P : never);
```
RReverser: TryGetGlobalType does the fallback for you btw, so this can be done much simpler.

```suggestion
    float16: Uint16Array | TryGetGlobalType<'Float16Array', never>
```

rockygeekz: Thanks for the suggestion! Updated `DataTypeMap['float16']` to use `TryGetGlobalType<'Float16Array', never>` directly.

RReverser: I see you removed manual overloads when converting to change to the DataMap, does the code from the original issue still work as expected without TS errors?

rockygeekz: Just double-checked locally using the public `dist/cjs` entrypoint,
`new Tensor(new Float16Array(...))` typechecks correctly with no TS errors.
Everything works as expected.

RReverser: Huh I wonder why they even need all those manual overloads if datatypemap is enough.

https://github.com/microsoft/onnxruntime/blob/c6c72e3b9ecd4363d43f116d34d0c0fc0dde0f5c/js/common/lib/tensor.ts#L257-L361

Oh well, perhaps an opportunity for future cleanup. Thanks for double-checking!

rockygeekz: Tested locally and saw no errors. If I missed something, please let me know!

RReverser: No I believe you :)

RReverser:
Oh, why did you close this?

PR #26745, opened 2025-12-07
Log filesystem path instead of string for C++20 compliance

JanSellner:
### Description
Log the `std::filesystem::path` object instead of the `ORTCHAR_T` string to avoid C++20 compliance issues.

### Motivation and Context
`ORTCHAR_T` is a `wchar_t` on Windows and it is not allowed to pass that to `std::basic_ostream`.

<details>
<summary>Example</summary>

([example.cpp](https://godbolt.org/z/1W4K6ojrY))
```C++
#include <iostream>
#include <filesystem>

int main() {
    const wchar_t* s = L"my/path";
    std::filesystem::path path(s);

    std::cout << s << std::endl;

    return 0;
}
```

Output on latest MSVC with `/std:c++20`:
```
<source>(8): error C2280: 'std::basic_ostream<char,std::char_traits<char>> &std::operator <<<std::char_traits<char>>(std::basic_ostream<char,std::char_traits<char>> &,const wchar_t *)': attempting to reference a deleted function
Z:\compilers\msvc\14.44.35207-14.44.35219.0\include\__msvc_ostream.hpp(972): note: see declaration of 'std::operator <<'
<source>(8): error C2088: built-in operator '<<' cannot be applied to an operand of type 'std::ostream'
```
</details>


JanSellner:
@skottmckay Anything I can do to move forward with this (small) PR?

skottmckay:
Can you rebase your branch to pickup the latest `main`? We're not able to trigger the CIs which are required before checkin and that might be the cause as the problem only seems to affect your PR.

image

JanSellner:
Hmm, worth a try :P
Branch is up-to-date now.

PR #26751, opened 2025-12-08
[WebGPU] Implement Split-K on GEMM 

Jiawei-Shao:
### Description
This patch implements the `Split-K` optimization on `GEMM`. 

1. Support handling `GEMM` in `MatMulFillBiasOrZeroBeforeSplitKProgram`.
    We need to add `beta` as a new uniform value and all the parameters that are used to handle all the cases of `GEMM` in `MatMulWriteFnSource()` (including the broadcast of `beta` on both dimensions).
2. Support `Split-K` in `GemmProgram::GenerateShaderCode()`.
3. Add cases to `GemmOptimizePackedTest` to test `Split-K` in `GEMM`.

### Motivation and Context
With this PR we can achieve about 20% improvement in `florence-2-base-decoder-with-past-fp16` and 10% improvement in `detr-resnet-50-fp16` on Lunar Lake iGPU.

### DIFF
--- a/onnxruntime/core/providers/webgpu/math/matmul.cc
+++ b/onnxruntime/core/providers/webgpu/math/matmul.cc
@@ -304,20 +304,11 @@ MatMulFillBiasOrZeroBeforeSplitKProgram CreateMatMulFillBiasOrZeroBeforeSplitKPr
   const uint32_t total_outputs = dim_a_outer * dim_b_outer;
   const uint32_t dispatch_x = (total_outputs + WORKGROUP_SIZE - 1) / WORKGROUP_SIZE;
 
-  TensorShape output_shape_temp;
-  if (is_gemm) {
-    // GEMM doesn't have `batch` in its output shape.
-    output_shape_temp = TensorShape({dim_a_outer, dim_b_outer});
-  } else {
-    const uint32_t batch_size = 1;
-    output_shape_temp = TensorShape({batch_size, dim_a_outer, dim_b_outer});
-  }
-
   // To reuse `MatMulWriteFnSource()` we need to set `dim_b_outer` in components when `output_shape`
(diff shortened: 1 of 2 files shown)


Jiawei-Shao:
@qjia7 @jchen10 @xhcao PTAL, thanks!

Jiawei-Shao:
The failure on Windows ARM64 QNN CI Pipeline is because of a build failure (as below) and it is not related to this PR:

```
##[error]The Operation will be canceled. The next steps may not contain expected logs.

##[error]The process 'C:\ToolCache\Python\3.13.11\arm64\python.exe' failed with exit code null
##[error]The operation was canceled.
```

onnxruntime/core/providers/webgpu/math/matmul.cc
```diff
@@ -287,32 +287,44 @@ Status ComputeMatMul(ComputeContext* context,
...
-      .AddOutput({output, ProgramTensorMetadataDependency::TypeAndRank, output_shape_vec4, static_cast<int32_t>(bias_components)})
-      .AddUniformVariables({{dim_a_outer}, {dim_b_outer}})
+  TensorShape output_shape_temp;
+  if (is_gemm) {
+    // GEMM doesn't have `batch` in its output shape.
+    output_shape_temp = TensorShape({dim_a_outer, dim_b_outer});
+  } else {
+    const uint32_t batch_size = 1;
+    output_shape_temp = TensorShape({batch_size, dim_a_outer, dim_b_outer});
+  }
```
qjia7: Why not directly use `output_shape` since splitK has ensured that batch_size must be 1? And pass the expected 2d shape for is_gemm=true?

Jiawei-Shao: Done

PR #26206, opened 2025-09-30
[EP ABI] [DRAFT] Support kernel-based EPs

adrianlizarraga:
### Description
**DRAFT**

PR #26754, opened 2025-12-08
[EP ABI] Add weight pre-packing support to kernel-based plugin EPs

adrianlizarraga:
### Description
- Adds C APIs to support pre-packing of const weights for `OrtKernelImpl` implementations.
  - APIs optionally support sharing of pre-packed weight data (for cpu-accessible memory).
- Updates example kernel (Mul) to use new pre-packing API. Tested by existing unit test: https://github.com/microsoft/onnxruntime/blob/549d7415e26e2b3f86c42f86e135bb746caa37b4/onnxruntime/test/autoep/test_execution.cc#L242-L256

### Motivation and Context
The [previous PR](https://github.com/microsoft/onnxruntime/pull/26206) added the base APIs that support kernel-based plugin EPs. This PR adds an additional feature that was identified as necessary for the port of WebGPU EP.

### DIFF
--- a/include/onnxruntime/core/session/onnxruntime_ep_c_api.h
+++ b/include/onnxruntime/core/session/onnxruntime_ep_c_api.h
@@ -368,8 +368,11 @@ struct OrtKernelImpl {
 
   /** \brief Optional function that receives data for a shared pre-packed weight from ORT.
    *
-   * This function is called after a prior call to OrtKernelImpl::PrePackWeight for a specific `input_index` set
-   * `is_packed` to true and stored weight data (to share) into the provided OrtSharedPrePackedWeightCache instance.
+   * ORT calls this function after calling OrtKernelImpl::PrePackWeight for a specific `input_index` if:
+   *   - OrtKernelImpl::PrePackWeight set the output parameter `is_packed` to true.
+   *   - OrtKernelImpl::PrePackWeight stored weight data to share into the provided OrtSharedPrePackedWeightCache
+   *     parameter (`prepacked_weight_cache`) via the API SharedPrePackedWeightCache_StoreWeightData.
+   *
    * Refer to the description of the "sharing-mode" in the documentation for OrtKernelImpl::PrePackWeight().
    *
    * \note ORT will not call this function for an `input_index` that a previous call to
--- a/onnxruntime/test/autoep/library/example_plugin_ep_kernel_registry/kernels/mul.cc
+++ b/onnxruntime/test/autoep/library/example_plugin_ep_kernel_registry/kernels/mul.cc
@@ -135,6 +135,9 @@ OrtStatus* ORT_API_CALL Mul::PrePackWeightImpl(OrtKernelImpl* this_ptr, const Or
     std::array<size_t, 1> buffer_data_sizes = {weight_info.num_bytes};
 
     Ort::UnownedSharedPrePackedWeightCache weight_cache(prepacked_weight_cache);
+
+    // weight_cache takes ownership of the data. As the API documentation states, this requires that the
+    // weight data is allocated with the OrtAllocator provided as a parameter to OrtKernelImpl::PrePackWeight.
     RETURN_IF_ERROR(weight_cache.StoreWeightData(buffer_data_ptrs.data(),
                                                  buffer_data_sizes.data(),
                                                  buffer_data_ptrs.size()));


include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -308,6 +308,32 @@ struct OrtKernelImpl {
...
+   * \param[in] tensor The OrtValue instance representing the constant tensor. Do not cache in the kernel.
+   * \param[in] input_index The input index of the tensor in this kernel.
+   * \param[in] alloc Implementation should use this allocator for allocating the packed data. It will either be
+   *                  an allocator provided by the EP (default or read-only) or a shared allocator overridden by
+   *                  the application via OrtApi::CreateSharedAllocator().
+   * \param[out] is_packed Output parameter that the implementation sets to true if the kernel packed the tensor data.
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \note Implementation of this function is optional (set to NULL). If not implemented, ORT assumes the kernel
```
adrianlizarraga:
```suggestion
   * \note Implementation of this function is optional. If not implemented (set to NULL), ORT assumes the kernel
```

onnxruntime/core/session/plugin_ep/ep_kernel_registration.cc
```diff
@@ -37,8 +40,60 @@ class PluginEpOpKernel final : public OpKernel {
     return ToStatusAndRelease(kernel_impl_->Compute(kernel_impl_, reinterpret_cast<OrtKernelContext*>(ctx)));
   }
 
+  Status PrePack(const Tensor& tensor, int input_idx, AllocatorPtr alloc,
+                 /*out*/ bool& is_packed, /*out*/ PrePackedWeights* /*prepacked_weights*/) override {
+    // Note: The `prepacked_weights` parameter is not used because sharing of prepacked weights is only
+    // really supported for ORT's CPU EP.
+
+    assert(kernel_impl_ != nullptr);  // Should be ensured by PluginEpOpKernel::Create().
```
adrianlizarraga:
```suggestion
    // Note: The `prepacked_weights` parameter is not used because sharing of prepacked weights is only
    // really supported for ORT's CPU EP.
    assert(kernel_impl_ != nullptr);  // Should be ensured by PluginEpOpKernel::Create().
```

onnxruntime/core/session/plugin_ep/ep_kernel_registration.cc
```diff
@@ -37,8 +40,60 @@ class PluginEpOpKernel final : public OpKernel {
...
+    }
+
+    auto empty_tensor_deleter = [](void* /*data*/) -> void {
+      // Do not delete Tensor (as we do not own it).
+    };
+
+    // Create a non-owning OrtValue that wraps the const Tensor& with an empty deleter.
+    // This is passed to OrtKernelImpl::PrePackConstantTensor() as a const OrtValue*.
+    // The above reasons make the const_cast relatively "safe".
+    const OrtValue ort_value(const_cast<Tensor*>(&tensor), DataTypeImpl::GetType<Tensor>(), empty_tensor_deleter);
```
adrianlizarraga, line 98:
```suggestion
    // The above reasons make the const_cast relatively "safe".
    // Note: Documentation for OrtKernelImpl::PrePackConstantTensor disallows caching the OrtValue pointer.
    const OrtValue ort_value(const_cast<Tensor*>(&tensor), DataTypeImpl::GetType<Tensor>(), empty_tensor_deleter);
```

onnxruntime/core/session/plugin_ep/ep_kernel_registration.cc
```diff
@@ -37,8 +40,60 @@ class PluginEpOpKernel final : public OpKernel {
     return ToStatusAndRelease(kernel_impl_->Compute(kernel_impl_, reinterpret_cast<OrtKernelContext*>(ctx)));
   }
 
+  Status PrePack(const Tensor& tensor, int input_idx, AllocatorPtr alloc,
+                 /*out*/ bool& is_packed, /*out*/ PrePackedWeights* /*prepacked_weights*/) override {
+    // Note: The `prepacked_weights` parameter is not used because sharing of prepacked weights is only
+    // really supported for ORT's CPU EP.
```
adrianlizarraga: My understanding is that sharing of prepacked weights is only for CPU EP: https://github.com/microsoft/onnxruntime/blob/549d7415e26e2b3f86c42f86e135bb746caa37b4/onnxruntime/core/framework/session_state.cc#L486-L489

The other (non-cpu) EPs that use PrePack() also do not use sharing: cuda, js, webgpu, xnnpack.

Therefore, this PR does not currently pass in the `PrePackedWeights` parameter to the `OrtKernelImpl::PrePackConstantTensor` function as it would not be used. However, I'm curious if anyone thinks we should we pass a parameter anyway just to reserve a parameter in case it is used in the future? I defaulted to not doing that, but I'm curious what others think.

fs-eire (dismissed):

onnxruntime/core/session/plugin_ep/ep_kernel_registration.cc
yuslepukhin: I was thinking about it, and one hand Prepacking is not that popular among non-CPU EPs, on the other hand, if some kernels decided to share, let them do it. Space is usually constrained on devices.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -308,6 +308,33 @@ struct OrtKernelImpl {
    * \since Version 1.24.
    */
   ORT_API_T(void, Release, _In_ OrtKernelImpl* this_ptr);
+
+  /** \brief Optional function to pack an initialized constant tensor to the kernel's preferred data layout.
+   *
+   * For example, a Conv kernel can define this function to pack input W to the channel-last data layout.
+   *
+   * \note The kernel is responsible for storing/owning the packed data and related metadata if it sets `is_packed`
```
yuslepukhin: Sharing is big thing on CPU especially between the sessions. Intra session sharing was introduced a little later. So it would make sense to allow it.

onnxruntime/core/session/plugin_ep/ep_kernel_registration.cc
```diff
@@ -37,8 +40,64 @@ class PluginEpOpKernel final : public OpKernel {
...
+    assert(kernel_impl_ != nullptr);  // Should be ensured by PluginEpOpKernel::Create().
```
yuslepukhin: This is true, but only because we can not calculate the hash of the buffer on CPU, so non-CPU kernels (and at least 1 CPU kernel) choose not to share.
But I would think sharing would be very much helpful given the sizes of the models are growing.
For the same reason, we now can save pre-packs externally and memory map them.

onnxruntime/test/autoep/library/example_plugin_ep_kernel_registry/kernels/mul.cc
```diff
@@ -26,16 +26,21 @@ OrtStatus* Mul::Create(const OrtKernelInfo* info, void* state,
 
 OrtStatus* Mul::DoCompute(OrtKernelContext* kernel_ctx) {
   Ort::KernelContext kernel_context(kernel_ctx);
-  static_cast<void>(this->state_);  // NOTE: Unused in this example.
-  static_cast<void>(this->info_);   // NOTE: Unused in this example.
+  static_cast<void>(this->data_transfer_impl_);  // NOTE: Unused in this example.
```
yuslepukhin: Can Create() above throw? #Resolved

onnxruntime/test/autoep/library/example_plugin_ep_kernel_registry/kernels/base.h
```diff
@@ -17,14 +17,24 @@ class BaseKernelImpl : public OrtKernelImpl {
...
+                                                           /*out*/ bool* is_packed) noexcept;
 
  private:
   // Derived classes implement DoCompute.
   // DoCompute is called by BaseKernelImpl::ComputeImpl, which also catches exceptions thrown by DoCompute
   // implementations and converts them into OrtStatus*.
   virtual OrtStatus* DoCompute(OrtKernelContext* kernel_ctx) = 0;
 
+  // Optional. Derived classes may choose to override. Otherwise, the default implementation sets `is_packed` to false.
+  virtual OrtStatus* DoPrePackConstantTensor(const OrtValue* tensor, int input_index, OrtAllocator* alloc,
```
yuslepukhin: From the description it is not clear how this is invoked and how this works together.
This looks like a template method pattern, but not quite (As in Transformers Apply()). If the user simply overrides this DoPrepack() nothing happens, since nothing is invoking it.
I would expect the default PrepacConstnatTensorImpl to invoke Do*() which may by default choose to do nothing.
As a potential user, I am confused what to override and when.

onnxruntime/test/autoep/library/example_plugin_ep_kernel_registry/kernels/squeeze.cc
```diff
@@ -107,15 +110,9 @@ OrtStatus* Squeeze::DoCompute(OrtKernelContext* kernel_ctx) {
...
   Ort::UnownedValue output = kernel_context.GetOutput(0, output_shape);
-  float* output_data = output.GetTensorMutableData<float>();
-  size_t num_bytes = output.GetTensorSizeInBytes();
-
-  if (input0.data() != output_data) {  // Don't copy if src == dst
-    // This uses a memcpy because the input and output are both located in the EP's device memory (i.e., cpu memory).
-    // Normally, an EP would use a OrtDataTransferImpl to generically handle copies where the source and destination
-    // could be on different devices.
-    memcpy(output_data, input0.data(), num_bytes);
-  }
```
yuslepukhin, line 126: I would expect this to have a MayInplace() option

yuslepukhin (changes requested):

onnxruntime/core/session/plugin_ep/ep_kernel_registration.cc
adrianlizarraga: I now added APIs to allow sharing of pre-packed weights.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
adrianlizarraga: I now added APIs to allow sharing of pre-packed weights.

onnxruntime/core/session/plugin_ep/ep_kernel_registration.cc
adrianlizarraga: I now added APIs to allow sharing of pre-packed weights.

onnxruntime/test/autoep/library/example_plugin_ep_kernel_registry/kernels/mul.cc
adrianlizarraga: Only `std::make_unique` would throw (which would be hard to handle correctly). However, a more complicated kernel implementation may add some error checking that could conceivably throw, so I added try/catch that converts any exceptions to an OrtStatus.

onnxruntime/test/autoep/library/example_plugin_ep_kernel_registry/kernels/base.h
adrianlizarraga: I removed the base class and virtual functions as they are not really needed and only obfuscate things. An EP author, just needs to fill out an `OrtKernelImpl`.

onnxruntime/test/autoep/library/example_plugin_ep_kernel_registry/kernels/squeeze.cc
adrianlizarraga: The squeeze kernel uses aliasing (not MayInPlace). Please refer to the macro calls at the top of the file.

I added a comment here that says that if the source and destination buffers are the same, the `CopyTensor` function does not actually do a copy, which is the case here due to the aliasing.

onnxruntime/core/session/plugin_ep/ep_kernel_registration.cc
```diff
@@ -37,8 +51,124 @@ class PluginEpOpKernel final : public OpKernel {
...
+      // OrtKernelImpl did not define a PrePack implementation.
+      is_packed = false;
+      return Status::OK();
+    }
+
+    // Only allow kernel to store/share pre-packed weights if the weight data will be stored in cpu-accessible memory.
+    // ORT requires that the data reside in cpu memory to be able to compute the hash of the weight's contents.
+    //
+    // If the allocator does not use CPU memory, we pass a NULL OrtPrePackWeightCache instance to the kernel to indicate
+    // that storing/sharing is not allowed and the kernel should manage the memory for the pre-packed weight.
```
adrianlizarraga:
```suggestion
    // If the allocator does not use CPU memory, we pass a NULL OrtSharedPrePackedWeightCache instance to the kernel
    // to indicate that storing/sharing is not allowed and the kernel should manage the memory for the pre-packed weight.
```

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -308,6 +309,94 @@ struct OrtKernelImpl {
...
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \note Implementation of this function is generally optional. It is only required if OrtKernelImpl::PrePack()
+   *       elects to share pre-packed weights.
+   *
+   * \since Version 1.24.
+   */
+  ORT_API2_STATUS(SetSharedPrePackedWeight, _In_ OrtKernelImpl* this_ptr,
+                  _In_reads_(num_buffers) const void* const* buffer_data_ptrs,
+                  _In_ size_t num_buffers, _In_ int input_index);
```
adrianlizarraga, line 406: Note: the internal version of this function ([OpKernel::UseSharedPrePackedBuffers](https://github.com/microsoft/onnxruntime/blob/19ee5716092ce5cbf8ac5cb8a89bf587a1c123c0/include/onnxruntime/core/framework/op_kernel.h#L126-L128)) has an output parameter called `used_shared_buffers` that the implementation sets to true if it used the shared weight data.

This output parameter is omitted from this public API because (imo) it is unnecessary:
- Not setting `used_shared_buffers` to true always results in any error anyway: https://github.com/microsoft/onnxruntime/blob/19ee5716092ce5cbf8ac5cb8a89bf587a1c123c0/onnxruntime/core/framework/session_state.cc#L435-L440
- A kernel's UseSharedPrePackedBuffers() function should never be called for a weight that the kernel is not expecting to share. If that happens, something is really wrong.

For the above reasons, the documentation for this API function instructs the implementer to return an error if it is called for an unexpected weight (instead of setting a boolean output to false).

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -308,6 +309,94 @@ struct OrtKernelImpl {
...
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \note Implementation of this function is generally optional. It is only required if OrtKernelImpl::PrePack()
+   *       elects to share pre-packed weights.
+   *
+   * \since Version 1.24.
+   */
+  ORT_API2_STATUS(SetSharedPrePackedWeight, _In_ OrtKernelImpl* this_ptr,
+                  _In_reads_(num_buffers) const void* const* buffer_data_ptrs,
```
adrianlizarraga, line 404: Another difference between this function and the internal [OpKernel::UseSharedPrePackedBuffers](https://github.com/microsoft/onnxruntime/blob/19ee5716092ce5cbf8ac5cb8a89bf587a1c123c0/include/onnxruntime/core/framework/op_kernel.h#L126-L128) is that the provided `buffer_data_ptrs` are fully const. This makes sense because the data is now owned by ORT, it may be shared with other kernels, and thus should not be modified by the kernel.

onnxruntime/core/session/plugin_ep/ep_kernel_registration.cc
```diff
@@ -17,8 +21,18 @@ namespace onnxruntime {
 /// </summary>
 class PluginEpOpKernel final : public OpKernel {
  private:
+  // Prevents calling constructor directly without having to make it private (required by make_unique_ptr).
```
edgchen1:
```suggestion
  // Prevents calling constructor directly without having to make it private (required by std::make_unique).
```

onnxruntime/core/session/plugin_ep/ep_kernel_registration.cc
```diff
@@ -37,8 +51,124 @@ class PluginEpOpKernel final : public OpKernel {
...
+                    "required OrtAllocator.");
+
+      // Fill out onnxruntime::PrePackedWeights from OrtSharedPrePackedWeightCache.
+      for (size_t i = 0; i < shared_weight_cache.buffer_data_ptrs.size(); i++) {
+        void* data_ptr = shared_weight_cache.buffer_data_ptrs[i].release();
+        size_t num_bytes = shared_weight_cache.buffer_sizes[i];
+
+        // Note: using the AllocatorPtr as the "deleter" (instead of the wrapping OrtAllocator cached by this kernel)
+        // because the AllocatorPtr lifetime likely exceeds the lifetime of the session and kernel instance.
+        prepacked_weights->buffers_.push_back(IAllocatorUniquePtr<void>(data_ptr, BufferDeleter(alloc)));
```
edgchen1: does this mean that the user must pass the same allocator in `OrtAllocator` form to `SharedPrePackedWeightCache_StoreWeightData()`? if so, maybe we don't need to expose the allocator as a parameter there. perhaps we can just internally use that same allocator as a deleter and simplify the API call for the user.

onnxruntime/test/autoep/library/example_plugin_ep_kernel_registry/kernels/mul.cc
```diff
@@ -46,4 +90,90 @@ OrtStatus* Mul::DoCompute(OrtKernelContext* kernel_ctx) {
...
+    RETURN_IF_ERROR(Ort::GetEpApi().SharedPrePackedWeightCache_StoreWeightData(prepacked_weight_cache,
+                                                                               buffer_data_ptrs.data(),
+                                                                               buffer_data_sizes.data(),
+                                                                               buffer_data_ptrs.size(),
+                                                                               allocator));
+
+    // IMPORTANT: This kernel no longer owns the packed weight data.
+    // weight_info.shared_data will be initialized in the call to SetSharedPrePackedWeightImpl.
+    weight_info.owned_data.release();
+    weight_info.owned_data = nullptr;
```
edgchen1: nit: don't need to assign nullptr after unique_ptr::release()

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -308,6 +309,94 @@ struct OrtKernelImpl {
...
+   *       within ORT.
+   *
+   * \param[in] this_ptr The OrtKernelImpl instance.
+   * \param[in] buffer_data_ptrs An array of buffer data pointers that collectively hold the pre-packed data for a
+   *                             single shared weight. Note that sometimes a single weight may have multiple pre-packed
+   *                             buffers and it is up to the kernel implementation to determine how to split the data
+   *                             into multiple buffers (if desired) in the call to OrtKernelImpl::PrePack(). Each
+   *                             buffer element in this array is provided in the same order defined by the kernel
+   *                             implementation in the call to OrtKernelImpl::PrePack().
+   * \param[in] buffer_data_sizes An array of buffer byte sizes, one per element in `buffer_data_ptrs`.
```
edgchen1, line 390: the `buffer_data_sizes` parameter is not present in the signature, but I think it would be good to have. #Resolved

onnxruntime/core/session/plugin_ep/ep_kernel_registration.cc
adrianlizarraga: >does this mean that the user must pass the same allocator in OrtAllocator form to SharedPrePackedWeightCache_StoreWeightData()?

Yes, but only if the ep kernel chooses to share the weight data. There's a check a few lines above that verifies that the expected allocator was used.

If the ep kernel doesn't want to share the weight data, then they could technically use any allocator they want.

>if so, maybe we don't need to expose the allocator as a parameter there. perhaps we can just internally use that same allocator as a deleter and simplify the API call for the user.

In "sharing mode", the ep kernel still has to allocate the data within the call to `OrtKernelImpl::PrePackWeight()`. The allocator provided by ORT could be one of the EP's preferred allocators (IExecutionProvider::GetPreferredAllocators) or an allocator overridden by app (OrtApi::RegisterAllocator). If we don't provide the allocator as an argument, then ORT would have to copy the data. I'm not sure if we want to have that extra copy.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
adrianlizarraga: I agree. Unfortunately, the sizes are not provided to the internal `OpKernel::UseSharedPrePackedBuffers` (which then calls this API). We would have to cache the sizes (per kernel input) within the PluginEpOpKernel class to make that work, but I'm hesitant to make the PluginEpOpKernel class bigger because its size impacts all kernels (even those that do not use prepacking).

 I'll have to remove this param from the description.

onnxruntime/core/session/plugin_ep/ep_kernel_registration.cc
edgchen1: > Yes, but only if the ep kernel chooses to share the weight data. There's a check a few lines above that verifies that the expected allocator was used.

> If the ep kernel doesn't want to share the weight data, then they could technically use any allocator they want.

if they are not sharing, they would not be calling `SharedPrePackedWeightCache_StoreWeightData()`, right? so if there's only one correct allocator to pass to it (i.e., the one provided to `OrtKernelImpl::PrePackWeight()`), what's the value of having it as a parameter? regarding an extra copy, can that be avoided if we just assume that the correct allocator for allocating shared weights is used by the kernel implementation?

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
yuslepukhin: I think the kernel actually returns the sizes when prepacking.

onnxruntime/core/session/plugin_ep/ep_api.cc
```diff
@@ -582,6 +583,81 @@ ORT_API_STATUS_IMPL(EpGraphSupportInfo_LookUpKernel, _In_ OrtEpGraphSupportInfo*
...
+    }
+
+    prepacked_weight_cache->allocator = deleter;
+  }
+  ORT_CATCH(const std::exception& ex) {
+    ORT_HANDLE_EXCEPTION([&]() {
+      // This API function promises that ORT will take ownership of the data only if it returns successfully.
+      // If any exception occurred while filling out `prepacked_weight_cache`, we try to release ownership so that
+      // the caller retains ownership of all of the original data and can delete it.
+
```
yuslepukhin: Nice!

onnxruntime/core/session/plugin_ep/ep_kernel_registration.cc
yuslepukhin: Here is a thought. If the kernel creates a pre-packed weight in a non-CPU memory, it can also compute for it a unique hash/key. This can, in theory, enable sharing?

onnxruntime/core/session/plugin_ep/ep_kernel_registration.cc
adrianlizarraga: >if they are not sharing, they would not be calling SharedPrePackedWeightCache_StoreWeightData(), right?

That's right.

>what's the value of having it [allocator] as a parameter?

Do you mean having it as a parameter to `OrtKernelImpl::PrePackWeight` or `SharedPrePackedWeightCache_StoreWeightData`?

If you mean `OrtKernelImpl::PrePackWeight`, then I would say we need the allocator parameter because the kernel impl doesn't necessarily know where the allocator came from. Most likely the allocator came from the ep itself, but it could also come from the app.

If you mean `SharedPrePackedWeightCache_StoreWeightData`, then there is some value:
- Allows ORT to actually verify that the correct allocator was used. (not just trust the ep). This verification happens within `SharedPrePackedWeightCache_StoreWeightData` (check that the allocation is cpu-accessible) and within `PluginEpOpKernel::PrePack` to verify that the expected allocator was used.
- It is an explicitly signal to the API caller that the ownership of the data will be transferred. By forcing the kernel impl to pass along an allocator to be used as a "deleter", it makes this intention clear (imo).

If we're ok with trusting that the EP's kernel impl will just use the correct allocator, then perhaps we could remove the `allocator` parameter from `SharedPrePackedWeightCache_StoreWeightData`. As you point out, the benefit would be a simpler API call (pass one less param).

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
adrianlizarraga: > I think the kernel actually returns the sizes when prepacking.

Yes, but the subsequent call to `OpKernel::UseSharedPrePackedWeightBuffers()` does not have the buffer sizes as a parameter (only pointer to data):

https://github.com/microsoft/onnxruntime/blob/07724d5c3277d2b2bdaad7687272a9beb6bd2ef7/include/onnxruntime/core/framework/op_kernel.h#L126-L131

I cannot change that interface at this point. The only way to provide the `OrtKernelImpl::SetSharedWeightData` with the buffer sizes is to cache the buffer sizes within the internal `PluginEpOpKernel` instance (inherits from `OpKernel`). As I mentioned in my above comment, this would introduce a size overhead for every single kernel instance used by a plugin EP (even those that do not use pre-packing).

I don't know if I can justify that given that the original API used by provider-bridge EPs currently expects the kernel implementation to remember what the sizes were.

onnxruntime/core/session/plugin_ep/ep_kernel_registration.cc
adrianlizarraga: >If we're ok with trusting that the EP's kernel impl will just use the correct allocator, then perhaps we could remove the allocator parameter from SharedPrePackedWeightCache_StoreWeightData. As you point out, the benefit would be a simpler API call (pass one less param).

Thinking about this a bit more. There will always be some level of trust required. An ep kernel impl could just as easily allocate the data with the **wrong** allocator but then pass the expected allocator to `SharedPrePackedWeightCache_StoreWeightData`. ORT would not really be able to do anything about that ....

Perhaps that sways me a bit towards removing the allocator parameter from `SharedPrePackedWeightCache_StoreWeightData`.

edgchen1: > If we're ok with trusting that the EP's kernel impl will just use the correct allocator

I think that's ok. the expectation that the allocator passed to `OrtKernelImpl::PrePackWeight()` is the one that is used for allocating shared weights is documented.

adrianlizarraga: Pushed a commit that removes this parameter.

onnxruntime/test/autoep/library/example_plugin_ep_kernel_registry/kernels/mul.cc
adrianlizarraga: yep, removed.

onnxruntime/core/session/plugin_ep/ep_kernel_registration.h
```diff
@@ -4,12 +4,49 @@
...
 #include "core/framework/kernel_registry.h"
 #include "core/framework/op_kernel.h"
+#include "core/framework/prepacked_weights.h"
+
+/// <summary>
+/// Implementation of the public C API opaque type OrtSharedPrePackedWeightCache used by plugin EP kernels.
+/// This wraps and fills out an instance of onnxruntime::PrePackedWeights.
+/// </summary>
+struct OrtSharedPrePackedWeightCache {
+  OrtSharedPrePackedWeightCache(onnxruntime::PrePackedWeights& container, onnxruntime::AllocatorPtr allocator);
```
yuslepukhin, line 28: >OrtShared [](http://example.com/codeflow?start=2&length=9)

nit: docs for consistency

onnxruntime/core/session/plugin_ep/ep_kernel_registration.h
```diff
@@ -4,12 +4,49 @@
...
+
+  /// <summary>
+  /// Sets data buffers for the shared weight. The data is expected to be allocated with this->allocator_.
```
yuslepukhin: >/// Sets dat [](http://example.com/codeflow?start=2&length=12)

Need a note on the buffer ownership.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -308,6 +309,93 @@ struct OrtKernelImpl {
    * \since Version 1.24.
    */
   ORT_API_T(void, Release, _In_ OrtKernelImpl* this_ptr);
+
+  /** \brief Optional function to pre-pack a constant tensor (i.e., a weight) to the kernel's preferred data layout.
```
yuslepukhin, line 313: >** \brief Optional  [](http://example.com/codeflow?start=3&length=19)

When the dust settles on C API, lets add CXX API?

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -308,6 +309,93 @@ struct OrtKernelImpl {
...
+  ORT_API2_STATUS(PrePackWeight, _In_ OrtKernelImpl* this_ptr, _In_ const OrtValue* tensor,
+                  _In_ int input_index, _Inout_ OrtAllocator* allocator,
+                  _In_opt_ OrtSharedPrePackedWeightCache* prepacked_weight_cache, _Out_ bool* is_packed);
+
+  /** \brief Optional function that receives data for a shared pre-packed weight from ORT.
+   *
+   * This function is called after a prior call to OrtKernelImpl::PrePackWeight set `is_packed` to true and
+   * stored weight data (that the kernel wants to share) into the provided `OrtSharedPrePackedWeightCache` instance.
+   * If this function is called for an unexpected shared weight (as identified by `input_index`), the
+   * implementation should return an error OrtStatus with an error code of ORT_EP_FAIL.
```
edgchen1: will ORT call this function with an input index for which a previous call to `PrePackWeight()` did not specify sharing mode prepacking? if so, I wonder if we should change that instead of asking the EP implementer to handle this "unexpected" case.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -308,6 +309,93 @@ struct OrtKernelImpl {
...
+   * \note This function is based on the internal OpKernel::UseSharedPrePackedBuffers() virtual function used
+   *       within ORT.
+   *
+   * \param[in] this_ptr The OrtKernelImpl instance.
+   * \param[in] buffer_data_ptrs An array of buffer data pointers that collectively hold the pre-packed data for a
+   *                             single shared weight. Note that sometimes a single weight may have multiple pre-packed
+   *                             buffers and it is up to the kernel implementation to determine how to split the data
+   *                             into multiple buffers (if desired) in the call to OrtKernelImpl::PrePack(). Each
+   *                             buffer element in this array is provided in the same order defined by the kernel
+   *                             implementation in the call to OrtKernelImpl::PrePack().
```
edgchen1: are these buffers expected to contain the same contents and be the same lengths as the ones passed into the last call to `SharedPrePackedWeightCache_StoreWeightData` within the `PrePack` call for the same `input_index`?

this is my understanding, and it might be wrong. I think it'd be worth making this clear in the docs.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -308,6 +309,93 @@ struct OrtKernelImpl {
...
+   * \param[in] num_buffers The number of buffers used to store the data for the shared pre-packed weight.
+   *                        Specifies the number of elements in the `buffer_data_ptrs` and `buffer_sizes` arrays.
```
edgchen1: there's no `buffer_sizes` parameter now.
```suggestion
   *                        Specifies the number of elements in the `buffer_data_ptrs` array.
```

onnxruntime/core/session/plugin_ep/ep_kernel_registration.cc
```diff
@@ -3,22 +3,68 @@
...
+}
+
+bool OrtSharedPrePackedWeightCache::HasData() const noexcept {
+  return !container_.buffers_.empty();
+}
+
+void OrtSharedPrePackedWeightCache::ReleaseAllData() noexcept {
+  for (onnxruntime::IAllocatorUniquePtr<void>& data_unique_ptr : container_.buffers_) {
+    data_unique_ptr.release();
+  }
```
edgchen1, line 46: this kind of leaves the instance in a weird state. `container_.buffers_` is `nullptr`s, but `container_.buffer_sizes_` still has non-zero sizes.

maybe zero `container_.buffer_sizes_`, or clear both `container_.buffers_` and `container_.buffer_sizes_` to leave things in a more consistent state?

onnxruntime/core/session/plugin_ep/ep_kernel_registration.cc
```diff
@@ -37,8 +83,108 @@ class PluginEpOpKernel final : public OpKernel {
...
+    if (prepacked_weights != nullptr && alloc->Info().device.UsesCpuMemory()) {
+      ORT_RETURN_IF(!prepacked_weights->buffers_.empty() || !prepacked_weights->buffer_sizes_.empty(),
+                    "PluginEpOpKernel::PrePack() expected PrePackedWeights instance to be initially empty");
+      shared_weight_cache.emplace(OrtSharedPrePackedWeightCache(*prepacked_weights, alloc));
+    }
+
+    ORT_RETURN_IF_ERROR(ToStatusAndRelease(
+        kernel_impl_->PrePackWeight(kernel_impl_, &ort_value, input_idx,
+                                    ort_allocator,
+                                    shared_weight_cache.has_value() ? &shared_weight_cache.value() : nullptr,
```
edgchen1: nit: `optional::operator*()` doesn't check if there is a value, but `optional::value()` does. we already check for a value earlier.
```suggestion
                                    shared_weight_cache.has_value() ? &*shared_weight_cache : nullptr,
```

onnxruntime/core/session/plugin_ep/ep_api.cc
```diff
@@ -10,6 +10,7 @@
 #include <vector>
 
 #include "core/common/semver.h"
+#include "core/framework/allocator.h"
```
edgchen1: is this still needed?

onnxruntime/core/session/plugin_ep/ep_kernel_registration.cc
```diff
@@ -3,22 +3,68 @@
...
  private:
+  // Prevents calling constructor directly without having to make it private (required by std::make_unique).
   struct PrivateTag {};
 
+  // Stores a mapping between a IAllocator* and the OrtAllocator* that wraps it.
+  struct PrePackAllocatorMapping {
+    PrePackAllocatorMapping(IAllocator* i_alloc, std::unique_ptr<OrtAllocatorImplWrappingIAllocator> ort_alloc)
+        : i_allocator(i_alloc), ort_allocator(std::move(ort_alloc)) {}
+
+    IAllocator* i_allocator{};                                            // Original IAllocator passed to PrePack()
```
edgchen1: do we need to maintain this mapping, or can we use `OrtAllocatorImplWrappingIAllocator::GetWrappedIAllocator()`?

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
adrianlizarraga: >will ORT call this function with an input index for which a previous call to PrePackWeight() did not specify sharing mode prepacking?

Currently, ORT will never call this function unnecessarily.

That said, this leaves a hole in the specification of this API function. What should the implementer return if a future version of ORT mistakenly calls this function for an input that the kernel did not want to share? This doc comment provides the answer: return an error because it should not happen.

I'll modify to explicitly state that it should not happen, but if it does, it is an error.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
adrianlizarraga: It may not have the same data pointers, but the content (`memcmp`) will be the same. This is briefly mentioned in the doc for `OrtKernelImpl::PrePackWeight` but I'll expand here.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
adrianlizarraga: Yep, added cxx api.

onnxruntime/core/session/plugin_ep/ep_kernel_registration.cc
adrianlizarraga: Thanks, updated to clear both.

onnxruntime/core/session/plugin_ep/ep_kernel_registration.h
adrianlizarraga: Added doc comments to constructor.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
adrianlizarraga: Updated. Please let me know if you think this is ok or if I should just remove the instruction to return an error.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
adrianlizarraga: Updated.

onnxruntime/core/session/plugin_ep/ep_kernel_registration.cc
adrianlizarraga: Thanks good catch

onnxruntime/core/session/plugin_ep/ep_api.cc
adrianlizarraga: not anymore. removed.

onnxruntime/core/session/plugin_ep/ep_kernel_registration.cc
adrianlizarraga: Good call! Removed that struct in favor of just using `OrtAllocatorImplWrappingIAllocator`

onnxruntime/core/session/plugin_ep/ep_kernel_registration.h
adrianlizarraga: Added a note on buffer ownership.

include/onnxruntime/core/session/onnxruntime_cxx_api.h
```diff
@@ -3379,5 +3379,22 @@ struct KernelRegistry : detail::Base<OrtKernelRegistry> {
...
+
+  //< Wraps SharedPrePackedWeightCache_StoreWeightData
+  Status StoreWeightData(void** buffer_data_ptrs, size_t* buffer_sizes, size_t num_buffers);
+};
+}  // namespace detail
+
+/** \brief Wrapper around ::OrtSharedPrePackedWeightCache
+ * \remarks OrtSharedPrePackedWeightCache is always unowned, but mutable, for EpApi users.
+ */
+using UnownedSharedPrePackedWeightCache =
```
yuslepukhin, line 3409: If this is Unowned, it must be returned by another API call that would not relinquish the ownership, but I am not seeing another CXX API that returns it.

adrianlizarraga: Not really. An unowned (but mutable) `OrtSharedPrePackedWeightCache*` instance is passed into `OrtKernelImpl::PrePackWeight` as a function **argument** that the kernel implementation can use. Please see the example Mul kernel implementation: 

https://github.com/microsoft/onnxruntime/blob/e94c0aab4408c8353548216be7293e50b07b3424/onnxruntime/test/autoep/library/example_plugin_ep_kernel_registry/kernels/mul.cc#L97-L152

adrianlizarraga: My documentation comment was not good. I updated it to hopefully make it clearer that this is a convenience c++ wrapper class.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
edgchen1:
```
   * \note ORT will not call this function for an `input_index` that a previous call to
   *       OrtKernelImpl::PrePackWeight() did not elect to pre-pack and share. If this function receives an
   *       unexpected `input_index`, the implementation should return an error OrtStatus with code ORT_EP_FAIL.
```

hm, I think it'd be ok for the EP implementer to handle that however they would like. IMO, we don't need to specify how the EP implementer should handle a case that shouldn't happen. maybe we can skip the instruction to return `ORT_EP_FAIL`, but I think the first part of that note is good.

adrianlizarraga: Updated.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -846,6 +939,35 @@ struct OrtEpApi {
...
+   * \note Subsequent calls with the same OrtSharedPrePackedWeightCache instance release and replace the old data.
+   *
+   * \param[in] this_ptr The OrtKernelImpl instance.
+   * \param[in] buffer_data_ptrs An array of buffer data pointers that collectively hold the pre-packed data for a
+   *                             single shared weight. Note that sometimes a single weight may have multiple pre-packed
+   *                             buffers and it is up to the kernel implementation to determine how to split the data
+   *                             into multiple buffers (if desired).
+   * \param[in] buffer_data_sizes An array of buffer byte sizes, one per element in `buffer_data_ptrs`.
+   * \param[in] num_buffers The number of buffers used to store the data for the shared pre-packed weight.
+   *                        Specifies the number of elements in the `buffer_data_ptrs` and `buffer_sizes` arrays.
```
edgchen1:
```suggestion
   *                        Specifies the number of elements in the `buffer_data_ptrs` and `buffer_data_sizes` arrays.
```

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -327,24 +329,25 @@ struct OrtKernelImpl {
...
    *
    * \param[in] this_ptr The OrtKernelImpl instance.
    * \param[in] tensor The OrtValue instance representing the constant tensor (weight). Do not cache in the kernel.
    * \param[in] input_index The input index of the tensor in this kernel.
-   * \param[in] allocator Implementation can use this allocator for allocating the packed data. Use of this allocator
-   *                      is only required in "sharing mode" (see above). This allocator will either be
-   *                      an allocator originally provided by the EP (default or read-only) or a shared allocator
-   *                      overridden by the application via OrtApi::CreateSharedAllocator(). The allocator remains
-   *                      valid throughout the lifetime of the OrtKernelImpl instance.
+   * \param[in] allocator Implementation must use this allocator for allocating the packed data in sharing and
```
edgchen1: the additional information about `allocator` is good.

I wonder if we should recommend instead of require usage of `allocator` in non-sharing mode? then the EP implementer has flexibility to do what they want.

include/onnxruntime/core/session/onnxruntime_cxx_api.h
```diff
@@ -3379,5 +3379,34 @@ struct KernelRegistry : detail::Base<OrtKernelRegistry> {
...
+};
+}  // namespace detail
+
+/** \brief Convenience C++ wrapper class around a ::OrtSharedPrePackedWeightCache instance owned by ORT.
+ *
+ * An `OrtSharedPrePackedWeightCache*` instance is passed as an argument to OrtKernelImpl::PrePackWeight.
+ * Example use:
+ *   OrtStatus* MyKernel::PrePackWeightImpl(OrtKernelImpl*, ..., OrtSharedPrePackedWeightCache* c_cache, ...) {
+ *     ...
+ *     if (cache != nullptr) {
```
edgchen1:
```suggestion
 *     if (c_cache != nullptr) {
```

edgchen1 (dismissed):

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
adrianlizarraga: yeah, updated.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -308,6 +309,98 @@ struct OrtKernelImpl {
...
+   * \since Version 1.24.
+   */
+  ORT_API2_STATUS(PrePackWeight, _In_ OrtKernelImpl* this_ptr, _In_ const OrtValue* tensor,
+                  _In_ int input_index, _Inout_ OrtAllocator* allocator,
+                  _In_opt_ OrtSharedPrePackedWeightCache* prepacked_weight_cache, _Out_ bool* is_packed);
+
+  /** \brief Optional function that receives data for a shared pre-packed weight from ORT.
+   *
+   * This function is called after a prior call to OrtKernelImpl::PrePackWeight for a specific `input_index` set
+   * `is_packed` to true and stored weight data (to share) into the provided OrtSharedPrePackedWeightCache instance.
```
fs-eire: > ... and stored weight data (to share) into the provided OrtSharedPrePackedWeightCache instance

Does it mean the implementation called `SharedPrePackedWeightCache_StoreWeightData()` on the passed-in parameter `prepacked_weight_cache` in `PrePackWeight()`

adrianlizarraga: Yes, that's what this is. Would you like me to rephrase?

onnxruntime/test/autoep/library/example_plugin_ep_kernel_registry/kernels/mul.cc
```diff
@@ -46,4 +90,96 @@ OrtStatus* Mul::DoCompute(OrtKernelContext* kernel_ctx) {
...
+  Ort::ConstValue original_weight(tensor);
+  auto type_shape_info = original_weight.GetTensorTypeAndShapeInfo();
+  size_t num_bytes = original_weight.GetTensorSizeInBytes();
+
+  PackedWeightInfo weight_info = {};
+  weight_info.mem_info = Ort::ConstMemoryInfo(allocator->Info(allocator));
+  weight_info.shape = type_shape_info.GetShape();
+  weight_info.elem_type = type_shape_info.GetElementType();
+  weight_info.num_bytes = num_bytes;
+  weight_info.owned_data = AllocateBytes(allocator, num_bytes);
```
fs-eire, line 122: if I understand correctly, `AllocateBytes` returns a `unique_ptr`, which contains the deleter so that it knows how to destruct the buffer.

Later in the `sharing_allowed == true` condition, the data is passed as `void *` to parameter of function call `weight_cache.StoreWeightData()`, and the ownership is transferred too. How the data will be destroyed after use?

adrianlizarraga: The `OrtSharedPrePackedWeightCache` internally contains a reference to the allocator passed to `PrePackWeight`. The internal implementation of `SharedPrePackedWeightCache_StoreWeightData` uses this same allocator to create unique ptr.

The documentation to these APIs require that the `PrePackWeight` use the provided allocator if sharing is desired.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
adrianlizarraga: Rephrased.

onnxruntime/test/autoep/library/example_plugin_ep_kernel_registry/kernels/mul.cc
adrianlizarraga: Added a comment to line that calls `weight_cache.StoreWeightData` to clarify.

fs-eire (approved):
Approve for the API and examples.

Issue #434, opened 2019-02-02
Build error "Could not copy the file ..."


Victel:
**Describe the bug**
Build error: “Could not copy the file “project\packages\Microsoft.ML.OnnxRuntime.0.2.1\runtimes\win10-x64\native\mkldnn.dll" because it was not found.”.  Indeed, there’s no “win10-x64” in runtimes but “win-x64” and “linux-x64”, only.
**System information**
- Windows 10, VS 2017, Windows Desktop project
**To Reproduce**
Build project with Microsoft.ML.OnnxRuntime.0.2.1 installed
**Expected behavior**
Build w/o errors
**Screenshots**
**Additional context**

shahasad:
Fixed by
https://github.com/Microsoft/onnxruntime/pull/325

shahasad:
https://github.com/Microsoft/onnxruntime/pull/435

PR #26756, opened 2025-12-09
[QNN] Add Genie libraries to QNN EP (#434)

qti-kromero:
### Description
* Adds Genie libraries to onnxruntime-qnn for future GenAI improvements accessible through Genie API

### Motivation and Context
* Lays groundwork for eventual integration of Genie API with onnxruntime-qnn


edgchen1:
why not add the additional libraries at the time they are used?

qti-kromero:
> why not add the additional libraries at the time they are used?

admittedly thought the above was going to be more work than it was so had separated this into its own unit of work - we could defer merging this till other work is ready if that's the preference

edgchen1:
> > why not add the additional libraries at the time they are used?
> 
> admittedly thought the above was going to be more work than it was so had separated this into its own unit of work - we could defer merging this till other work is ready if that's the preference

I see, I think it would be preferable to add them together with the code that uses them.

PR #26759, opened 2025-12-09
[EP ABI] Add CreateCustomOpDomains() API for plugin EP to register custom ops 

chilo-ms:
### Description

The newly added two APIs, `CreateCustomOpDomains()` and `GetNumCustomOpDomains`,  are used when running inference on a model that contains EP-specific custom operations.
  
 Workflow:
 1. The EP implements these functions to supply a list of `OrtCustomOpDomain` instances.
 2. The application either 1) calls `SessionOptionsAppendExecutionProvider_V2()` with an `OrtEpDevice` containing
     the plugin EP's factory or 2) enables auto ep selection.
 3. Then ORT either 1) `SessionOptionsAppendExecutionProvider_V2()` appends the provided OrtCustomOpDomains to the
     session options or 2) registers the OrtCustomOpDomains from the selected EP devices.
  
 As a result, any session created from these session options will have these custom op domains registered
 in ORT, ensuring that the custom ops are properly recognized and validated when the model is loaded.

Plugin EPs can provide two types of custom ops:
   1. A full OrtCustomOp with a concrete kernel implementation
      - This Example EP demonstrates this approach.
      - In GetCapability(), it calls EpGraphSupportInfo_AddSingleNode() to inform ORT
        that the custom node should NOT be fused or compiled. Instead, ORT should invoke
        the custom node's Compute() function at runtime.

   2. A "placeholder" OrtCustomOp with an empty kernel implementation
      - A compile-based Plugin EP can supply an OrtCustomOp whose CustomKernel::Compute()
        does nothing. The purpose is to satisfy model validation during model loading by
        registering the custom op as a valid operator in the session.
      - In GetCapability(), the EP should call EpGraphSupportInfo_AddNodesToFuse() to
        notify ORT that this custom node should be fused and compiled by the EP.
      - In Compile(), the EP executes its compiled bits to perform inference for
      the fused custom node.

### Motivation and Context

Currently, the provider-bridge TRT RTX EP and TRT EP supports registering custom op domain list in session option so
that it can run model contains TRT specific custom ops.

This PR adds the same feature for plugin EP.

### DIFF
--- a/onnxruntime/test/testdata/custom_mul.py
+++ b/onnxruntime/test/testdata/custom_mul.py
@@ -3,15 +3,15 @@ import onnx
 
 def create_custom_mul_model():
     # === Inputs ===
-    x = onnx.helper.make_tensor_value_info("X", TensorProto.FLOAT, [3, 2])
-    w = helper.make_tensor_value_info("W", TensorProto.FLOAT, [3, 2])
+    x = onnx.helper.make_tensor_value_info("X", onnx.TensorProto.FLOAT, [3, 2])
+    w = onnx.helper.make_tensor_value_info("W", onnx.TensorProto.FLOAT, [3, 2])
 
     # === Output ===
-    y = helper.make_tensor_value_info("Y", TensorProto.FLOAT, [3, 2])
+    y = onnx.helper.make_tensor_value_info("Y", onnx.TensorProto.FLOAT, [3, 2])
 
     # === Custom Node: Custom_Mul ===
     # Replace "Mul" with your custom op name and domain
-    custom_node = helper.make_node(
+    custom_node = onnx.helper.make_node(
         op_type="Custom_Mul",  # <-- custom op name
         inputs=["X", "W"],
         outputs=["Y"],
@@ -19,7 +19,7 @@ def create_custom_mul_model():
     )
 
     # === Graph ===
-    graph = helper.make_graph(
+    graph = onnx.helper.make_graph(
         nodes=[custom_node],
         name="CustomMulGraph",
         inputs=[x, w],
@@ -27,11 +27,11 @@ def create_custom_mul_model():
     )
 
     # === Model (opset version 13 or later is fine) ===
-    model = helper.make_model(
+    model = onnx.helper.make_model(
         graph,
         opset_imports=[
-            helper.make_opsetid("", 13),  # standard ONNX domain
-            helper.make_opsetid("com.example", 1),
+            onnx.helper.make_opsetid("", 13),  # standard ONNX domain
+            onnx.helper.make_opsetid("com.example", 1),
         ],  # your custom domain
         producer_name="custom_mul_builder",
     )


include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -1413,6 +1413,32 @@ struct OrtEpFactory {
...
+   *    the plugin EP's factory.
+   * 3. SessionOptionsAppendExecutionProvider_V2() appends the provided OrtCustomOpDomain list to the
+   *    session options.
+   *
+   * As a result, any session created from these session options will have these custom op domains registered
+   * in ORT, ensuring that the custom ops are properly recognized and validated when the model is loaded.
+   *
+   * Note: EP has the responsibility to release OrtCustomOpDomain instances it creates.
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
```
adrianlizarraga, line 2104: reminder to add the missing param descriptions

onnxruntime/core/session/utils.cc
```diff
@@ -505,6 +506,38 @@ Status AddEpOptionsToSessionOptions(gsl::span<const OrtEpDevice* const> ep_devic
...
+      auto is_already_in_domains = [&](std::string& domain_name, std::vector<OrtCustomOpDomain*>& domains) {
+        for (auto ptr : domains) {
+          if (domain_name == ptr->domain_) {
+            return true;
+          }
+        }
+        return false;
+      };
+
+      if (ep_factory->CreateCustomOpDomains == nullptr) {
```
adrianlizarraga: `ep_factory->CreateCustomOpDomains` does not exist for ORT 1.23. Think we need a check here that skips a factory with version older than 24.

onnxruntime/core/session/utils.cc
```diff
@@ -505,6 +506,38 @@ Status AddEpOptionsToSessionOptions(gsl::span<const OrtEpDevice* const> ep_devic
 
       ORT_RETURN_IF_ERROR(config_options.AddConfigEntry((prefix + ep_option_keys[j]).c_str(), ep_option_vals[j]));
     }
+
+    // add custom op domain provided by EP to the session options if any
+    OrtEpFactory* ep_factory = ep_device->ep_factory;
+    if (ep_factory) {
+      auto is_already_in_domains = [&](std::string& domain_name, std::vector<OrtCustomOpDomain*>& domains) {
```
adrianlizarraga:
```suggestion
      auto is_already_in_domains = [&](const std::string& domain_name, const std::vector<OrtCustomOpDomain*>& domains) {
```

onnxruntime/test/autoep/library/example_plugin_ep/ep_factory.cc
```diff
@@ -308,3 +313,22 @@ OrtStatus* ORT_API_CALL ExampleEpFactory::CreateSyncStreamForDeviceImpl(OrtEpFac
 
   return nullptr;
 }
+
+/*static*/
+OrtStatus* ORT_API_CALL ExampleEpFactory::CreateCustomOpDomainsImpl(OrtEpFactory* this_ptr,
+                                                                    _Outptr_result_maybenull_ OrtCustomOpDomain** out,
```
adrianlizarraga: I don't think a `OrtCustomOpDomain**` parameter allows this function to return more than 1 domain, right?

Maybe it would be easier to have a separate API that returns the number of domains (e.g., `OrtEpFactory::GetNumberOfCustomOpDomains()`), then ORT allocates an array of the right size, ORT calls `OrtEpFactory::CreateCustomOpDomains` with the array it allocated, and then the EP factory just fills the array.

The EP is responsible for calling `ReleaseCustomOpDomain` (happens automatically if using C++ api). ORT is responsible for managing the memory for the array of domains (can just be a `std::vector<OrtCustomOpDomain*>`)

onnxruntime/test/autoep/library/example_plugin_ep/ep_factory.h
```diff
@@ -5,10 +5,77 @@
...
+ private:
+  void* compute_stream_;
+};
+
+struct PluginEpCustomOp : Ort::CustomOpBase<PluginEpCustomOp, PluginEpCustomKernel> {
+  explicit PluginEpCustomOp(const char* provider, void* compute_stream) : provider_(provider),
+                                                                          compute_stream_(compute_stream) {
+  }
+
+  void* CreateKernel(const OrtApi& /* api */, const OrtKernelInfo* info) const {
```
adrianlizarraga: where is this called?

onnxruntime/test/autoep/library/example_plugin_ep/ep_factory.h
```diff
@@ -5,10 +5,77 @@
...
+    return new PluginEpCustomKernel(info, compute_stream_);
```
adrianlizarraga: Does this lead to a memory leak?

onnxruntime/test/autoep/library/example_plugin_ep/ep_factory.cc
chilo-ms: Good catch!
I update the `CreateCustomOpDomains()` and added `GetNumCustomOpDomains()` so that it now can support more than one domain.

onnxruntime/test/autoep/library/example_plugin_ep/ep_factory.h
chilo-ms: I updated the unit test, so now the `CreateKernel` will be called by ORT.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
chilo-ms: added.

onnxruntime/core/session/utils.cc
chilo-ms: Added the check.

onnxruntime/test/autoep/library/example_plugin_ep/ep_factory.h
chilo-ms: It won't leak.
`ExampleEpCustomOp` derives from `CustomOpBase` so the KernelDestroy will be called during clean up.

````c++
struct CustomOpBase : OrtCustomOp {
   ...
    OrtCustomOp::KernelDestroy = [](void* op_kernel) { delete static_cast<TKernel*>(op_kernel); };
  ... 
````

File: include/onnxruntime/core/session/onnxruntime_ep_c_api.h [hunk 7]
```diff
@@ -1413,6 +1413,49 @@ struct OrtEpFactory {
...
+   * \param[out] domains Pre-allocated array of `num_domains` elements by ORT that should be filled with
+                         OrtCustomOpDomain created by the EP.
+   * \param[in] num_domains The size of the `domains` array pre-allocated by ORT.
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \since Version 1.24.
+   */
+  ORT_API2_STATUS(CreateCustomOpDomains, _In_ OrtEpFactory* this_ptr,
+                  _Outptr_result_maybenull_ OrtCustomOpDomain** domains, _Out_ size_t num_domains);
```
ishwar-raut1: OrtCustomOpDomain! 
This structure is not ABI safe or stable across boundaries.

chilo-ms: The `OrtCustomOpDomain` is an opaque struct to EP, and EP can create it by calling C API `CreateCustomOpDomain()` to get a pointer of an `OrtCustomOpDomain` instance and then add `OrtCustomOp` instances to it.
I think it's ABI stable as it's just a C pointer.

Please see the implementation in Example EP in the unit test as a reference.

ynankani: What is the expectation for OrtCustomOpDomain** domains this will contain a deep copy or shallow copy of the OrtCustomOp that ep provides?

chilo-ms: The `OrtCustomOpDomain` contains a shallow copy (a pointer) of the `OrtCustomOp`.

The reason is:

EP should call `OrtApis::CustomOpDomain_Add()` to add `OrtCustomOp` to `OrtCustomOpDomain`.
As per api implementation, it simply stores the pointer of  the `OrtCustomOp` in the vector, not constructing a new `OrtCustomOp`
````c++
ORT_API_STATUS_IMPL(OrtApis::CustomOpDomain_Add, _Inout_ OrtCustomOpDomain* custom_op_domain, _In_ const OrtCustomOp* op) {
  API_IMPL_BEGIN
  custom_op_domain->custom_ops_.emplace_back(op);
  return nullptr;
  API_IMPL_END
}
````
Then, inside ORT, it simply stores the pointer of `OrtCustomOpDomain` returned from EP into session options.
````c++
  Status AddEpOptionsToSessionOptions() {
  ...
      ORT_RETURN_IF_ERROR(ToStatusAndRelease(ep_factory->CreateCustomOpDomains(ep_factory,
                                                                               domains.data(),
                                                                               domains.size())));

      const auto domains_span = gsl::span<OrtCustomOpDomain*>(domains.data(), domains.size());
      for (auto domain : domains_span) {
        if (!is_already_in_domains(domain->domain_, ort_session_options.custom_op_domains_) &&
            domain->custom_ops_.size() > 0) {
          ort_session_options.custom_op_domains_.push_back(domain);
        } else {
          LOGS_DEFAULT(WARNING) << "The custom op domain name "
                                << domain->domain_ << " is already in the session option. Skip it.";
        }
      }
    ...
  }  
````

File: onnxruntime/core/session/utils.cc [hunk 8]
```diff
@@ -505,6 +506,45 @@ Status AddEpOptionsToSessionOptions(gsl::span<const OrtEpDevice* const> ep_devic
...
       ORT_RETURN_IF_ERROR(config_options.AddConfigEntry((prefix + ep_option_keys[j]).c_str(), ep_option_vals[j]));
     }
+
+    // Add custom op domain provided by EP to the session options if any.
+    // OrtEpFactory::GetNumCustomOpDomains and OrtEpFactory::CreateCustomOpDomains
+    // were added in ORT 1.24.
+    OrtEpFactory* ep_factory = ep_device->ep_factory;
+    if (ep_factory &&
+        ep_factory->ort_version_supported >= 24 &&
+        ep_factory->CreateCustomOpDomains != nullptr) {
```
adrianlizarraga: also need to check `ep_factory->GetNumCustomOpDomains` just in case?

onnxruntime/core/session/plugin_ep/ep_factory_internal_impl.h
```diff
@@ -88,6 +88,13 @@ class EpFactoryInternalImpl {
     return nullptr;
   }
 
+  virtual OrtStatus* CreateCustomOpDomains(_Outptr_result_maybenull_ OrtCustomOpDomain** out,
```
adrianlizarraga: Do we also need `GetNumCustomOpDomains` in here?

onnxruntime/test/autoep/library/example_plugin_ep/ep_factory.h
```diff
@@ -5,9 +5,90 @@
...
+//      does nothing. The purpose is to satisfy model validation during model loading by
+//      registering the custom op as a valid operator in the session.
+//    - In GetCapability(), the EP should call EpGraphSupportInfo_AddNodesToFuse() to
+//      notify ORT that this custom node should be fused and compiled by the EP.
+//    - In Compile(), the EP executes its compiled bits to perform inference for
+//      the fused custom node.
+//
+// Note: Approach #2 is suitable for plugin TRT RTX EP to support TRT plugins.
+
+struct CustomMulKernel : MulKernel {
```
adrianlizarraga: nit: could these custom op classes be placed in a separate file?

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -1413,6 +1413,62 @@ struct OrtEpFactory {
...
+   *    - A compile-based Plugin EP can supply an OrtCustomOp whose CustomKernel::Compute()
+   *      does nothing. The purpose is to satisfy model validation during model loading by
+   *      registering the custom op as a valid operator in the session.
+   *    - In GetCapability(), the EP should call EpGraphSupportInfo_AddNodesToFuse() to
+   *      notify ORT that this custom node should be fused and compiled by the EP.
+   *    - In Compile(), the EP executes its compiled bits to perform inference for
+   *      the fused custom node.
+   *
+   * Note: EP has the responsibility to release OrtCustomOpDomain instances it creates. It happens
+   *       automatically if using ORT C++ api.
```
edgchen1: IMO, with a name like "**Create**CustomOpDomains", there's some expectation of ownership transfer. E.g., how `OrtApi::CreateCustomOpDomain()` creates a new domain that the user is responsible for releasing.

since this function does not transfer ownership to the caller, maybe a name like `GetCustomOpDomains()` would be better?

onnxruntime/test/autoep/library/example_plugin_ep/ep_custom_op.h
```diff
@@ -0,0 +1,87 @@
...
+  CustomMulKernel(const OrtApi& ort_api,
+                  const OrtLogger& logger,
+                  const std::unordered_map<std::string, FloatInitializer>& float_initializers,
+                  std::string input0_name,
+                  std::string input1_name) : MulKernel(ort_api, logger, float_initializers,
+                                                       input0_name, input1_name) {
+  }
+};
+
+struct ExampleEpCustomOp : Ort::CustomOpBase<ExampleEpCustomOp, CustomMulKernel> {
```
edgchen1: `CustomMulKernel::Compute()` returns `OrtStatus*`. should this inherit from `CustomOpBase<ExampleEpCustomOp, CustomMulKernel, /*WithStatus*/ true>` instead so that the OrtStatus will be checked?

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -1413,6 +1413,62 @@ struct OrtEpFactory {
...
+   * \param[out] domains Pre-allocated array of `num_domains` elements by ORT that should be filled with
+                         OrtCustomOpDomain created by the EP.
+   * \param[in] num_domains The size of the `domains` array pre-allocated by ORT.
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \since Version 1.24.
+   */
+  ORT_API2_STATUS(CreateCustomOpDomains, _In_ OrtEpFactory* this_ptr,
+                  _Outptr_result_maybenull_ OrtCustomOpDomain** domains, _In_ size_t num_domains);
```
edgchen1: why is `_Outptr_result_maybenull_` used for `domains`?

onnxruntime/core/session/utils.cc
```diff
@@ -482,7 +482,8 @@ Status CreateIExecutionProviderFactoryForEpDevices(const Environment& env,
 Status AddEpOptionsToSessionOptions(gsl::span<const OrtEpDevice* const> ep_devices,
```
ynankani, line 596: Is this not called during auto ep selection path, when the session_options.set_provider_selection_policy(ort.OrtExecutionProviderDevicePolicy.PREFER_GPU) is set?
For auto ep path when it creates plugin ep it goes for inference_session.cc - RegisterExecutionProvider() where it calls GetCustomOpDomainList? Is this understanding correct?

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
chilo-ms: It makes sense to me to have a name `GetCustomOpDomains()`. Changed.

onnxruntime/test/autoep/library/example_plugin_ep/ep_custom_op.h
chilo-ms: Good catch. I also used `CreateKernelV2` and `ComputeKernerV2` so that OrtStatus will be checked.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
chilo-ms: I think EP can still implement this function and creates no custom op domain.
Or do you think `_Outptr_result_` makes more sense?

onnxruntime/core/session/utils.cc
chilo-ms: In the case of EP using `GetCustomOpDomains` to register custom op, only this EP can run the model contains that custom op.
IMO, the application should explicitly call `SessionOptionsAppendExecutionProvider_V2()` and specify that ep device, rather than using auto ep selection as other devices might not be able to run that custom op.

onnxruntime/test/autoep/library/example_plugin_ep/ep_factory.cc
```diff
@@ -308,3 +311,63 @@ OrtStatus* ORT_API_CALL ExampleEpFactory::CreateSyncStreamForDeviceImpl(OrtEpFac
...
+
+/*static*/
+OrtStatus* ORT_API_CALL ExampleEpFactory::GetCustomOpDomainsImpl(
+    OrtEpFactory* this_ptr,
+    _Outptr_result_maybenull_ OrtCustomOpDomain** domains,
+    _Out_ size_t num_domains) noexcept {
+  auto* factory = static_cast<ExampleEpFactory*>(this_ptr);
+
+  // Custom Op Domains
+  factory->custom_op_domains_[0] = Ort::CustomOpDomain{"test"};
```
edgchen1: would it be better to only create the custom op domains once, e.g., in the factory ctor?

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
edgchen1: hm, would `_Out_writes_all_(num_domains)` make sense here? assuming that `num_domains` is the value returned by `GetNumCustomOpDomains()` - perhaps that should also be documented. the implementation is expected to treat `domains` as a buffer.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -1564,6 +1564,62 @@ struct OrtEpFactory {
...
+   * \since Version 1.24.
+   */
+  ORT_API2_STATUS(GetNumCustomOpDomains, _In_ OrtEpFactory* this_ptr, _Out_ size_t* num_domains);
+
+  /** \brief Gets the EP-specific OrtCustomOpDomains.
+   *
+   * This function is used when running inference on a model that contains EP-specific custom operations.
+   *
+   * Workflow:
+   * 1. The EP implements this function to supply a list of OrtCustomOpDomain instances.
```
edgchen1: should clarify that the custom op domains are provided by the EP factory, not the EP.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -1564,6 +1564,62 @@ struct OrtEpFactory {
...
+   * 2. A "placeholder" OrtCustomOp with an empty kernel implementation
+   *    - A compile-based Plugin EP can supply an OrtCustomOp whose CustomKernel::Compute()
+   *      does nothing. The purpose is to satisfy model validation during model loading by
+   *      registering the custom op as a valid operator in the session.
+   *    - In GetCapability(), the EP should call EpGraphSupportInfo_AddNodesToFuse() to
+   *      notify ORT that this custom node should be fused and compiled by the EP.
+   *    - In Compile(), the EP executes its compiled bits to perform inference for
+   *      the fused custom node.
+   *
+   * Note: EP has the responsibility to release OrtCustomOpDomain instances it creates. It happens
```
edgchen1: it would be the EP factory and not the EP that has this responsibility, right? and the instances must be valid while any session is using them

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -1564,6 +1564,62 @@ struct OrtEpFactory {
    * \since Version 1.24.
    */
   ORT_API2_STATUS(SetEnvironmentOptions, _In_ OrtEpFactory* this_ptr, _In_ const OrtKeyValuePairs* options);
+
+  /** \brief Returns the number of OrtCustomOpDomains that this factory creates.
```
edgchen1: since we renamed it to `GetCustomOpDomains`, maybe replace "creates" with "provides" or "supplies".

onnxruntime/test/autoep/library/example_plugin_ep/ep.cc
```diff
@@ -262,26 +242,35 @@ OrtStatus* ORT_API_CALL ExampleEp::GetCapabilityImpl(OrtEp* this_ptr, const OrtG
 
         supported_nodes.push_back(node);  // Only support a single Mul for now.
         break;
+      } else if (op_type == "Custom_Mul") {
```
edgchen1: should we also check the domain?

onnxruntime/core/session/utils.cc
ynankani: Registering custom ops should be independent of EP , so it should be available in auto ep selectin path as well

onnxruntime/test/autoep/library/example_plugin_ep/ep_factory.cc
chilo-ms: Ah, that's good suggestion. Moved to factor ctor.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
chilo-ms: yes, modified to EP factory

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
chilo-ms: updated.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
chilo-ms: replaced with "provides"

onnxruntime/test/autoep/library/example_plugin_ep/ep.cc
chilo-ms: I think we don't need to. As when the model is loading, ORT already checks the domain.
Also, we can't really get the domain info given only an op, can we? i might be wrong

edgchen1: I mean the op domain, like `node.GetDomain()`. domain + op type identifies the op. here we are matching an op from a custom domain.

chilo-ms: Ah, okay, i added the domain check.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -2015,6 +2015,64 @@ struct OrtEpFactory {
   ORT_API2_STATUS(CreateExternalResourceImporterForDevice, _In_ OrtEpFactory* this_ptr,
                   _In_ const OrtEpDevice* ep_device,
                   _Outptr_result_maybenull_ OrtExternalResourceImporterImpl** out_importer);
+
+  /** \brief Returns the number of OrtCustomOpDomains that this factory provides.
+   *
+   * \param[in] this_ptr The OrtEpFactory instance.
+   * \param[out] num_domains Output parameter set to the number of created OrtCustomOpDomain instances.
```
edgchen1:
```suggestion
   * \param[out] num_domains Output parameter set to the number of provided OrtCustomOpDomain instances.
```

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -2015,6 +2015,64 @@ struct OrtEpFactory {
...
+   *    the plugin EP's factory.
+   * 3. SessionOptionsAppendExecutionProvider_V2() appends the provided OrtCustomOpDomain list to the
+   *    session options.
+   *
+   * As a result, any session created from these session options will have these custom op domains registered
+   * in ORT, ensuring that the custom ops are properly recognized and validated when the model is loaded.
+   *
+   * Plugin EPs can provide two types of custom ops:
+   *  1. A full OrtCustomOp with a concrete kernel implementation
+   *    - This Example EP demonstrates this approach.
```
edgchen1: "This Example EP" does not seem relevant in the context of the API docs

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -2015,6 +2015,64 @@ struct OrtEpFactory {
...
+   *      does nothing. The purpose is to satisfy model validation during model loading by
+   *      registering the custom op as a valid operator in the session.
+   *    - In GetCapability(), the EP should call EpGraphSupportInfo_AddNodesToFuse() to
+   *      notify ORT that this custom node should be fused and compiled by the EP.
+   *    - In Compile(), the EP executes its compiled bits to perform inference for
+   *      the fused custom node.
+   *
+   * Note: The OrtCustomOpDomain instances must be valid while any session is using them.
+           EP factory has the responsibility to release OrtCustomOpDomain instances it creates. It happens
+   *       automatically if using ORT C++ api.
```
edgchen1:
```suggestion
   *       automatically if using the C++ Ort::CustomOpDomain class.
```

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -2015,6 +2015,64 @@ struct OrtEpFactory {
...
+   *
+   * \param[in] this_ptr The OrtEpFactory instance.
+   * \param[out] domains Pre-allocated array of `num_domains` elements by ORT that should be filled with
```
edgchen1:
```suggestion
   * \param[out] domains Array of `num_domains` elements pre-allocated by ORT that should be filled with
```

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -2015,6 +2015,64 @@ struct OrtEpFactory {
...
+                         OrtCustomOpDomain created by the EP. The `num_domains` is the value returned by
```
edgchen1:
```suggestion
                         OrtCustomOpDomain instances created by the EP. The `num_domains` is the value returned by
```

onnxruntime/core/graph/model.h
```diff
@@ -374,4 +374,7 @@ class Model {
 
   CheckLoadCancellationFn check_load_cancellation_fn_;
 };
+
+template <typename T>
+Status LoadModel(const T& file_path, ONNX_NAMESPACE::ModelProto& model_proto);
```
edgchen1: can we just use [`Model::Load()`](https://github.com/microsoft/onnxruntime/blob/4b50b2f12a94ace49afe2acf1dbb075606645b45/onnxruntime/core/graph/model.h#L244-L245) instead of adding this to the header? typically, the template implementation is provided in the header as well, unless we are providing explicit instantiations for known types.

onnxruntime/core/session/provider_policy_context.cc
```diff
@@ -317,6 +254,89 @@ Status ProviderPolicyContext::SelectEpsForSession(const Environment& env, const
   return Status::OK();
 }
 
+Status ProviderPolicyContext::SelectEpDevices(const OrtSessionOptions& options, std::vector<const OrtEpDevice*>& execution_devices,
+                                              std::vector<const OrtEpDevice*>& devices_selected,
+                                              OrtKeyValuePairs* metadata_from_model,
+                                              InferenceSession& sess) {
```
edgchen1: can `execution_devices`, `metadata_from_model`, and `sess` be const?

onnxruntime/core/session/utils.cc
```diff
@@ -98,9 +100,99 @@ Status TestAutoSelectEPsImpl(const Environment& env, InferenceSession& sess, con
...
+      ep_factory->GetCustomOpDomains != nullptr) {
+    size_t num_domains = 0;
+    ORT_RETURN_IF_ERROR(ToStatusAndRelease(ep_factory->GetNumCustomOpDomains(ep_factory, &num_domains)));
+
+    domains.resize(num_domains);
+    ORT_RETURN_IF_ERROR(ToStatusAndRelease(ep_factory->GetCustomOpDomains(ep_factory,
+                                                                          domains.data(),
+                                                                          domains.size())));
+  }
+
```
edgchen1, line 119: nit: ensure `domains` is set in all successful paths. it can be written right before returning `Status::OK()`.

```c++
Status GetCustomOpDomainsFromEpDevice(const OrtEpDevice& ep_device, InlinedVector<OrtCustomOpDomain*>& domains_out) {
  InlinedVector<OrtCustomOpDomain*> domains{};

  // fill in domains as needed

  domains_out = std::move(domains);
  return Status::OK();
}
```

onnxruntime/core/session/utils.cc
```diff
@@ -98,9 +100,101 @@ Status TestAutoSelectEPsImpl(const Environment& env, InferenceSession& sess, con
...
+  return Status::OK();
+}
+
+bool IsDomainExisted(const std::string& domain_name, const std::vector<OrtCustomOpDomain*>& domains) {
```
edgchen1: name suggestion - `DoesDomainWithNameExist`

onnxruntime/core/session/utils.cc
```diff
@@ -195,6 +289,56 @@ static OrtStatus* CreateSessionAndLoadModelImpl(_In_ const OrtSessionOptions* op
...
+#if !defined(ORT_MINIMAL_BUILD)
+  // Add custom domains if the selected ep from auto ep selection has custom domains to register.
+  // The custom domains should be registered before model load for ORT to validate the custom ops.
+  if (options != nullptr &&
+      options->provider_factories.empty() &&
+      options->value.ep_selection_policy.enable) {
+    ProviderPolicyContext context;
+
+    // Following code calls the same ep selection functions that InitializeSession() calls as well
+    // to get `execution_devices` and `devices_selected`.
```
edgchen1: is it running the same code twice? is it possible to store and reuse the result?

onnxruntime/core/session/utils.cc
chilo-ms: I updated the code and now auto ep selection also support EP specific custom ops.

onnxruntime/core/session/utils.cc
```diff
@@ -98,9 +100,101 @@ Status TestAutoSelectEPsImpl(const Environment& env, InferenceSession& sess, con
...
+    for (auto& prop : model_proto.metadata_props()) {
+      metadata[prop.key()] = prop.value();
+    }
+    model_metadata.custom_metadata_map = metadata;
+
+    if (model_proto.has_graph() && model_proto.graph().has_name()) {
+      model_metadata.graph_name = model_proto.graph().name();
+    }
+
+    return;
```
edgchen1: nit: don't need `return;` at the end of a function.

onnxruntime/core/session/utils.cc
```diff
@@ -98,9 +100,101 @@ Status TestAutoSelectEPsImpl(const Environment& env, InferenceSession& sess, con
...
+  for (auto ptr : domains) {
+    if (domain_name == ptr->domain_) {
+      return true;
+    }
+  }
+  return false;
+}
+}  // namespace
+
+Status static GetModelMetadata(const ORTCHAR_T* model_path,
```
edgchen1: nit: this static function is right next to an anonymous namespace which would also make it scoped to this file. can it just not be static and go in there?

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
chilo-ms: Good catch, I rewrite the description.

onnxruntime/core/session/provider_policy_context.cc
chilo-ms: yes, make them all be const

onnxruntime/core/session/utils.cc
chilo-ms: The `execution_devices` and `devices_selected` will be cached in session and reused if needed.

onnxruntime/core/session/inference_session.h
```diff
@@ -662,6 +662,22 @@ class InferenceSession {
     return session_id_;
   }
 
+  void SetExecutionDevices(std::vector<const OrtEpDevice*>& execution_devices) {
```
edgchen1: if we intend to move from it, the parameter should be `vector&&` so this intent is clear to the caller. `vector` (by value) would also work and allow the caller to choose between moving or copying. otherwise, copying from a `const vector&` would work too. `vector&` is a bit unintuitive as it's unclear why/how the function will change the parameter.

onnxruntime/core/session/inference_session.h
```diff
@@ -662,6 +662,22 @@ class InferenceSession {
...
+    execution_devices_ = std::move(execution_devices);
+  }
+
+  const std::vector<const OrtEpDevice*>& GetExecutionDevices() noexcept {
+    return execution_devices_;
+  }
+
+  void SetSelectedDevices(std::vector<const OrtEpDevice*> selected_devices) {
```
edgchen1: having public setters (`SetSelectedDevices` and `SetExecutionDevices`) allows them to be set anywhere. is this necessary, or is it possible to initialize them in one place and only expose a getter?

onnxruntime/core/session/inference_session.h
chilo-ms: Thanks for the detailed explanations.
My intention is to use `vector` (by value) and like you mentioned it's more flexible, caller can choose between moving or copying.

onnxruntime/core/session/inference_session.h
chilo-ms: It seems hard to initialize them in `InferenceSession` ctor as it may not take `model_path`, `model_data`, `model_data_length` that are needed to get model proto and model metadata.

onnxruntime/core/session/utils.cc
```diff
@@ -542,6 +697,21 @@ Status AddEpOptionsToSessionOptions(gsl::span<const OrtEpDevice* const> ep_devic
 
       ORT_RETURN_IF_ERROR(config_options.AddConfigEntry((prefix + ep_option_keys[j]).c_str(), ep_option_vals[j]));
     }
+
+    // Add custom domains if EP factory has any.
```
adrianlizarraga, line 630: The purpose of this function is to add options from selected `OrtEpDevice` instances into the session options. Handling domains here too seems unexpected imo.

Could this be moved into a dedicated function? We'll end up iterating over the same `OrtEpDevice` instances again, but there shouldn't be that many.

onnxruntime/core/session/inference_session.h
```diff
@@ -1054,6 +1070,13 @@ class InferenceSession {
   // Enable nodestats collection
   std::optional<NodeStatsRecorder> node_stats_recorder_;
 #endif
+
+  // Holds the list of devices from the environment, ordered via OrderDevices().
+  // It's used for auto ep selection.
+  std::vector<const OrtEpDevice*> execution_devices_;
```
adrianlizarraga: Is the purpose of caching these in the session to allow running auto ep selection multiple times? That is, you may have to run it once before loading the model to process custom op domains for the selected OrtEpDevices?

onnxruntime/core/session/utils.cc
```diff
@@ -195,6 +289,66 @@ static OrtStatus* CreateSessionAndLoadModelImpl(_In_ const OrtSessionOptions* op
...
+    }
+
+    ORT_API_RETURN_IF_STATUS_NOT_OK(context.SelectEpDevices(*options, execution_devices,
+                                                            devices_selected, &model_metadata_key_value_pairs, *sess));
+
+    InlinedVector<OrtCustomOpDomain*> all_ep_custom_op_domains;
+
+    for (const OrtEpDevice* ep_device : devices_selected) {
+      InlinedVector<OrtCustomOpDomain*> domains;
+      ORT_API_RETURN_IF_STATUS_NOT_OK(GetCustomOpDomainsFromEpDevice(*ep_device, domains));
```
adrianlizarraga, line 260: Instead of running auto ep logic twice and having to cache stuff across various systems, would it be possible to get each EP library's custom op domains during ep library registration? We could cache those custom op domains in the environment (maybe environment::EpInfo). Then, when we initialize a session, we do a check:
 - If auto ep is enabled, just register all of the ep custom op domains before loading the model. No need to make sure that the OrtEpDevice will be used. There probably shouldn't be that many libraries registered and each library probably wont have many custom ops.
 - If auto ep is not used, only register the custom op domains of the OrtEpDevice instances that are explicitly used.

adrianlizarraga: Actually, maybe caching the ep custom op domains in the environment is unnecessary. Perhaps, just try to register all custom op domains for all OrtEpDevice instances here (no need to make sure they would be selected)?

chilo-ms: > Perhaps, just try to register all custom op domains for all OrtEpDevice instances here (no need to make sure they would be selected)?

One scenario that i can think about is if the custom op domains provided by one OrtEpDevice and that OrtEpDevice is not selected, then the model (contains custom op) might pass the model/graph validation, but the custom op will never be assigned to any EPs and ORT will fail. But that still seems okay to me.

onnxruntime/core/session/utils.cc
chilo-ms: That makes sense, i added `AddEpCustomDomainsToSessionOptions()` to do so.

onnxruntime/core/session/utils.cc
adrianlizarraga: Yeah, ORT should fail with some message that says that the custom_op node was not assigned to any EP, which seems fine.

chilo-ms: I updated the code to simply register all custom op domains for all OrtEpDevice instances and remove the changes for running auto ep logic and caching the results.

onnxruntime/core/session/utils.cc
```diff
@@ -546,5 +607,27 @@ Status AddEpOptionsToSessionOptions(gsl::span<const OrtEpDevice* const> ep_devic
...
+
+    const auto domains_span = gsl::span<OrtCustomOpDomain*>(domains.data(), domains.size());
+    for (auto domain : domains_span) {
+      if (!DoesDomainWithNameExist(domain->domain_, ort_session_options.custom_op_domains_) &&
+          domain->custom_ops_.size() > 0) {
+        ort_session_options.custom_op_domains_.push_back(domain);
+      } else {
+        LOGS_DEFAULT(WARNING) << "The custom op domain name "
+                              << domain->domain_ << " is already in the session option. Skip it.";
+      }
```
edgchen1: nit: the warning about a duplicate domain is a little misleading since that branch also gets taken for domains with no custom ops.

also, there's some similar code to add a custom op domain to a vector above. maybe consider refactoring into a helper function.

onnxruntime/python/onnxruntime_pybind_state.cc
```diff
@@ -1348,6 +1348,9 @@ static Status AddEpFactoryFromEpDevices(PySessionOptions& py_sess_options,
                                                    ep_option_vals,
                                                    py_sess_options.value));
 
+  ORT_RETURN_IF_ERROR(AddEpCustomDomainsToSessionOptions(ep_devices,
+                                                         py_sess_options.value));
```
adrianlizarraga: I think `py_sess_options.GetOrtSessionOptions()`

onnxruntime/test/unittest_util/test_dynamic_plugin_ep.cc
```diff
@@ -186,7 +186,7 @@ std::unique_ptr<IExecutionProvider> MakeEp(const logging::Logger* logger) {
   ORT_THROW_IF_ERROR(AddEpOptionsToSessionOptions(state.selected_c_ep_devices,
                                                   default_ep_option_key_cstrs,
                                                   default_ep_option_value_cstrs,
-                                                  ort_session_options.value));
+                                                  ort_session_options));
```
adrianlizarraga: I think this needs to be `ort_session_options.value`

onnxruntime/core/session/utils.cc
chilo-ms: >nit: the warning about a duplicate domain is a little misleading since that branch also gets taken for domains with no custom ops.

Changed the warning message.

>there's some similar code to add a custom op domain to a vector above. maybe consider refactoring into a helper function.

If i understand your meaning correctly I think it's a bit hard to add a helper function for both places to use this helper, as one is to add custom domain to the session options, another one is to call sess->AddCustomOpDomains() to add custom domain.

edgchen1: it can be something like this:

```c++
bool ShouldAddDomain(const OrtCustomOpDomain* domain_to_add,
                     gsl::span<const OrtCustomOpDomain* const> existing_domains) {
  if (domain_to_add->custom_ops_.size() == 0) {
    // log empty
    return false;
  }

  if (DoesDomainWithNameExist(domain_to_add->domain_, existing_domains)) {
    // log domain already exists
    return false;
  }

  return true;
}
```

chilo-ms: Ah, that's even better, i added this helper.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -2048,6 +2048,64 @@ struct OrtEpFactory {
...
+  /** \brief Gets the EP-specific OrtCustomOpDomains.
+   *
+   * This function is used when running inference on a model that contains EP-specific custom operations.
+   *
+   * Workflow:
+   * 1. The EP factory implements this function to supply a list of OrtCustomOpDomain instances.
+   * 2. The application either 1) calls SessionOptionsAppendExecutionProvider_V2() with an OrtEpDevice containing
+   *    the plugin EP's factory or 2) enables auto ep selection.
+   * 3. 1) SessionOptionsAppendExecutionProvider_V2() appends the provided OrtCustomOpDomains to the
+   *    session options or 2) ORT registers the OrtCustomOpDomains provided by the selected EP devices.
```
adrianlizarraga:
```suggestion
   *    session options or 2) ORT registers the OrtCustomOpDomains provided by the EP devices
   *    that could be potentially selected.
```

onnxruntime/core/session/utils.cc
```diff
@@ -195,6 +247,34 @@ static OrtStatus* CreateSessionAndLoadModelImpl(_In_ const OrtSessionOptions* op
...
+
+    for (const OrtEpDevice* ep_device : env.GetOrtEpDevices()) {
+      InlinedVector<OrtCustomOpDomain*> domains;
+      ORT_API_RETURN_IF_STATUS_NOT_OK(GetCustomOpDomainsFromEpDevice(*ep_device, domains));
+
+      const auto domains_span = gsl::span<OrtCustomOpDomain*>(domains.data(), domains.size());
+      const auto existing_domains = gsl::span<const OrtCustomOpDomain* const>(options->custom_op_domains_.data(),
+                                                                              options->custom_op_domains_.size());
+      for (auto domain : domains_span) {
+        if (ShouldAddDomain(domain, existing_domains)) {
```
adrianlizarraga: nit: (optional) I think that explicitly creating these spans is not needed. Containers like `InlinedVector` or `std::vector` should be implicitly converted to a `gsl::span` when passed as arguments to a function that accepts a span.

```suggestion
      for (auto* domain : domains) {
        if (ShouldAddDomain(domain, options->custom_op_domains_)) {
```

onnxruntime/core/session/utils.cc
```diff
@@ -546,5 +626,25 @@ Status AddEpOptionsToSessionOptions(gsl::span<const OrtEpDevice* const> ep_devic
...
+  for (const OrtEpDevice* ep_device : ep_devices) {
+    // Add custom domains if EP factory has any.
+    InlinedVector<OrtCustomOpDomain*> domains;
+    ORT_RETURN_IF_ERROR(GetCustomOpDomainsFromEpDevice(*ep_device, domains));
+
+    const auto domains_span = gsl::span<OrtCustomOpDomain*>(domains.data(), domains.size());
+    const auto existing_domains = gsl::span<OrtCustomOpDomain*>(ort_session_options.custom_op_domains_.data(),
+                                                                ort_session_options.custom_op_domains_.size());
+    for (auto domain : domains_span) {
+      if (ShouldAddDomain(domain, existing_domains)) {
```
adrianlizarraga: nit: (optional) Same here about the spans not really being necessary.
```suggestion
    for (auto* domain : domains) {
      if (ShouldAddDomain(domain, ort_session_options.custom_op_domains_)) {
```

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -2048,6 +2048,64 @@ struct OrtEpFactory {
...
+   *      the fused custom node.
+   *
+   * Note: The OrtCustomOpDomain instances must be valid while any session is using them.
+           EP factory has the responsibility to release OrtCustomOpDomain instances it creates. It happens
+   *       automatically if using the C++ Ort::CustomOpDomain class.
+   *
+   * \param[in] this_ptr The OrtEpFactory instance.
+   * \param[out] domains Array of `num_domains` elements pre-allocated by ORT that should be filled with
+                         OrtCustomOpDomain instances created by the EP. The `num_domains` is the value returned by
+                         GetNumCustomOpDomains(). The implementation is expected to treat `domains` as a buffer.
```
edgchen1: nit: sorry, this part was just from my comment and not intended as suggested documentation. I think it's obvious from the rest of the documentation that `domains` is an array.

```suggestion
                         GetNumCustomOpDomains().
```

adrianlizarraga (dismissed):

edgchen1 (approved):

chilo-ms:
It seems i need to address Copilot comment in order to merge.
@edgchen1 @adrianlizarraga please help sign off again, thanks.

chilo-ms:
I think all the comments are resolved and can't find the one that is unresolved (might be outdated and not show), so this PR still can't be merged.

image



i created another identical PR [here](https://github.com/microsoft/onnxruntime/pull/27050) and can merge that one if we still can't fix this issue.

PR #26762, opened 2025-12-09
Validate in-memory references in the incoming customer models

yuslepukhin:
### Description
We validate that the in-memory references in model initializers are valid and point to an existing OrtValues as opposed to arbitrary memory addresses

### Motivation and Context
MS FTE see ICM #31000000506659

### DIFF
--- a/onnxruntime/test/testdata/test_evil_weights.py
+++ b/onnxruntime/test/testdata/test_evil_weights.py
@@ -1,5 +1,4 @@
 import onnx
-from onnx import TensorProto, helper
 
 
 def create_exp_model():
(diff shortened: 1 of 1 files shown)


onnxruntime/test/testdata/test_evil_weights.py
```diff
@@ -0,0 +1,51 @@
...
+    )
+    tensors.append(evil_tensor)
+    evil_tensor.data_location = TensorProto.EXTERNAL
+    entry1 = evil_tensor.external_data.add()
+    entry1.key = "location"
+    entry1.value = "*/_ORT_MEM_ADDR_/*"
+    entry2 = evil_tensor.external_data.add()
+    entry2.key = "offset"
+    entry2.value = "4194304"
+    entry2.value = "12230656"
```
adrianlizarraga, line 28: Looks like there's a redundant `entry2.value` statement

apsonawane (approved):

Issue #26186, opened 2025-09-28
CMake configuration files in release assets `onnxruntime-*.tgz` are problematic


wangtz0607:
### Describe the issue

CMake configuration files in release assets `onnxruntime-*.tgz` are problematic.

## Download a release from GitHub

```sh
wget https://github.com/microsoft/onnxruntime/releases/download/v1.23.0/onnxruntime-linux-x64-1.23.0.tgz
```

Please note that this issue is not specific to v1.23.0, x64 and Linux. `onnxruntime-linux-x64-1.23.0.tgz` is used here only as an example.

## Extract the release

```sh
mkdir /tmp/onnxruntime
tar xzf onnxruntime-linux-x64-1.23.0.tgz --strip-components=1 -C /tmp/onnxruntime
```

## Use the release in a CMake project

Create `CMakeLists.txt` with the following content:

```cmake
cmake_minimum_required(VERSION 3.16)

project(example LANGUAGES CXX)

list(APPEND CMAKE_PREFIX_PATH /tmp/onnxruntime)

find_package(onnxruntime CONFIG REQUIRED)

add_executable(example main.cpp)
target_link_libraries(example PRIVATE onnxruntime::onnxruntime)
```

Create `main.cpp` with the following content:

```cpp
int main() {
    return 0;
}
```

Configure the project:

```sh
mkdir build && cd build
cmake ..
```

## Expected behavior

The project should be configured successfully.

## Actual behavior

```plaintext
CMake Error at /tmp/onnxruntime/lib/cmake/onnxruntime/onnxruntimeTargets.cmake:83 (message):
  The imported target "onnxruntime::onnxruntime" references the file

     "/tmp/onnxruntime/lib64/libonnxruntime.so.1.23.0"

  but this file does not exist.  Possible reasons include:

  * The file was deleted, renamed, or moved to another location.

  * An install or uninstall procedure did not complete successfully.

  * The installation package was faulty and contained

     "/tmp/onnxruntime/lib/cmake/onnxruntime/onnxruntimeTargets.cmake"

  but not all the files it references.

Call Stack (most recent call first):
  /tmp/onnxruntime/lib/cmake/onnxruntime/onnxruntimeConfig.cmake:26 (include)
  CMakeLists.txt:7 (find_package)
```

I tried:

```sh
ln -s /tmp/onnxruntime/lib /tmp/onnxruntime/lib64
```

but another error occurred:

```plaintext
CMake Error in CMakeLists.txt:
[...]

Issue #25279, opened 2025-07-03
[Build] CMake configurations files for bin release 1.22.0 are broken for Linux


g-gisbert:
### Describe the issue

Linux binaries provide CMake files to configure the project. However, the path to the lib files is wrong.
The SO is expected to be in the lib64 folder while the actual name is lib. As a workaround I renamed the folder but the CMake file could also be updated.

### Target platform

Linux

### Build script

find_package to onnxruntime

### Error / output

CMake Error at /home/someone/Documents/onnx/onnxruntime-linux-x64-1.22.0/lib/cmake/onnxruntime/onnxruntimeTargets.cmake:83 (message):
   The imported target "onnxruntime::onnxruntime" references the file "/home/someone/Documents/onnx/onnxruntime-linux-x64-1.22.0/lib64/libonnxruntime.so.1.22.0" but this file does not exist.

limann77:
@snnn Please reopen this issue.

mwestphal:
not stale

mwestphal:
nope

hariharans29:
The tar structure in the next release should be corrected with this: https://github.com/microsoft/onnxruntime/pull/26104

mwestphal:
nope

mwestphal:
nope, and thats a stupid policy

mwestphal:
still stupid

mwestphal:
gtfo

mwestphal:
W

mwestphal:
h

mwestphal:
a

mwestphal:
t

mwestphal:
t

Issue #25242, opened 2025-07-01
[Build] CMake configurations files for bin release 1.22.0 are broken


g-gisbert:
### Describe the issue

Linux and OSX binaries provide CMake files to configure the project. However, the path to the header files is wrong.
First, they expect the header files to be located in include/onnxruntime/ while they actually are in include/. As a workaround, I created the missing directory and moved the files there. The path in the CMake file could also be updated.

### Target platform

Linux, MacOS

### Build script

find_package to onnxruntime

### Error / output

CMake Error in CMakeLists.txt:
   Imported target "onnxruntime::onnxruntime" includes non-existent path "/home/someone/Documents/onnx/onnxruntime-linux-x64-1.22.0/include/onnxruntime" in its INTERFACE_INCLUDE_DIRECTORIES.

NicolasVuaille:
We are still facing this issue (@mwestphal FYI)

mwestphal:
not stale

mwestphal:
nope

hariharans29:
The tar structure in the next release should be corrected with this: https://github.com/microsoft/onnxruntime/pull/26104

mwestphal:
nope

mwestphal:
nope, and thats a stupid policy

mwestphal:
still stupid

mwestphal:
NO STALE

mwestphal:
This

hariharans29:
not stale

mwestphal:
no

mwestphal:
no

mwestphal:
no

PR #26767, opened 2025-12-11
Revert "Consistent with the configuration in the packaged cmake"

hariharans29:
Reverts microsoft/onnxruntime#26104

It seems like this fix although is correct and necessary, it needs some patches in some other places (like release pipelines and onnxruntime inference examples). We will try and address the actual issue the PR was addressing in a subsequent release. I will re-open all the GH issues that the fix PR closed out so that the actual issue is still tracked.

Will reopen:

https://github.com/microsoft/onnxruntime/issues/24003
https://github.com/microsoft/onnxruntime/issues/26186
https://github.com/microsoft/onnxruntime/issues/23642
https://github.com/microsoft/onnxruntime/issues/25279
https://github.com/microsoft/onnxruntime/issues/25242

### DIFF
--- a/tools/ci_build/github/linux/copy_strip_binary.sh
+++ b/tools/ci_build/github/linux/copy_strip_binary.sh
@@ -18,6 +18,8 @@ EXIT_CODE=1
 uname -a
 cd "$BINARY_DIR"
 mv installed/usr/local $ARTIFACT_NAME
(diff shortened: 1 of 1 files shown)


hariharans29:
The optional Lint failure can be ignored - IIRC the original PR had the same issue.

PR #26771, opened 2025-12-11
[webgpu] support auto pad for im2col-matmul

wenqinI:
### Description

This PR addressed the TODO about auto_pad for `im2col-matmul`, because the auto_pad attribute was used for computing the padding for the Conv2d before launching the kernel, so we don't need to do extra work to support it in the logic of the new `im2col-matmul` kernel.

### Motivation and Context

Please see above.


wenqinI:

onnxruntime/core/providers/webgpu/nn/im2col_matmul.cc
```diff
@@ -190,17 +190,15 @@ bool CanApplyIm2ColMatMulProgram(ComputeContext& context,
...
   if (!IsDeviceSupported(context)) {
     return false;
   }
 
   // TODO: Support !is_channels_last
   // TODO: Support fuse
-  // TODO: Support auto pad
   // TODO: Support group conv
-  if (!is_channels_last || activation_kind != ActivationKind::None || auto_pad != AutoPadType::NOTSET || group != 1) {
+  if (!is_channels_last || activation_kind != ActivationKind::None || group != 1) {
```
qjia7: I think the original checking 'auto_pad != AutoPadType::NOTSET' is because that 'updated_pads' will not equal to pads. It seems like a bug that group > 1 and group = 1 uses different pads. Could you please give more investigation why it is? For this PR, looks good for me. Thanks.

qjia7 (approved):

onnxruntime/core/providers/webgpu/nn/im2col_matmul.cc
wenqinI: Thank you for your review! I will take a look for the auto pad for "group > 1" case and update it here later.

guschmue:
run `lintrunner -a`

onnxruntime/core/providers/webgpu/nn/im2col_matmul.cc
wenqinI: Hi @qjia7,

I didn't deep dive into the implement of `group conv`, but I guess there may be some slight bugs for it.

The `group conv` just take **2 pads** in webgpu, within both heigh and weight direction, it's allitle weird (the `updated_pads` in below is what it used):

https://github.com/microsoft/onnxruntime/blob/07724d5c3277d2b2bdaad7687272a9beb6bd2ef7/onnxruntime/core/providers/webgpu/nn/conv.cc#L122-L125

But in the cpu backend, we could see it uses **4 pads** (the `pads[0-3]` in below):

https://github.com/microsoft/onnxruntime/blob/07724d5c3277d2b2bdaad7687272a9beb6bd2ef7/onnxruntime/core/providers/cpu/nn/conv.cc#L95-L114

So I guess why we do the check for `group > 1` on webgpu may just be a workaround for the wrong `group conv`?

wenqinI:
Hi @guschmue ,

Thank you for the reminder! I try to run lint for it, PTAL, thanks!

guschmue:
one more change to make CI happy please:

error: cannot convert ‘onnxruntime::AutoPadType’ to ‘uint32_t’ {aka ‘unsigned int’}

wenqinI:
Thank you for the review, I addressed the CI issue, PTAL, thanks!

wenqinI:
Hi @guschmue , 

This PR is updated, could you please run CI again and take a look, thanks!

daijh:
Ping @guschmue to help move this PR forward.

PR #24346, opened 2025-04-08
Enable SME for sgemm and sbgemm through KleidiAI

MichaelTylerArm:
### Description
Enables Arm® KleidiAI™ SME kernels for MLAS sgemm and sbgemm functions.

### Motivation and Context
These kernels provide performance improvements on SME-enabled devices.
We see a performance improvement of 1.2x-1.8x on onnxruntime_perf_test for the following Geekbench models on M4:

```
Model                           Speedup
deeplabv3_mobilenetv2_f16.onnx    1.79x
bert_tiny_f16.onnx                1.47x
deeplabv3_mobilenetv2_f32.onnx    1.43x
mobilenetv1_ssd_f16.onnx          1.29x
mobilenet_v1_f32.onnx             1.28x
mobilenetv1_ssd_f32.onnx          1.26x
de_efficientnetlitev3_f16.onnx    1.25x
mobilenet_v1_f16.onnx             1.23x
```

PR #26773, opened 2025-12-11
[MLAS] Integrate KleidiAI BF16 SME2 Kernel Through Mlas SBGEMM Path

patryk-kaiser-ARM:
**Description**
This PR integrates Arm® KleidiAI™ SME2 BF16 kernel through MLAS SBGEMM path.

Rework of https://github.com/microsoft/onnxruntime/pull/24346

**Motivation and Context**
This kernel provides performance improvements on SME-enabled devices.

### DIFF
--- a/onnxruntime/core/mlas/lib/kleidiai/sbgemm_kleidiai.cpp
+++ b/onnxruntime/core/mlas/lib/kleidiai/sbgemm_kleidiai.cpp
@@ -377,7 +377,7 @@ Return Value:
         const size_t lhs_packed_offset = sbgemm_gemm.ukernel.get_lhs_packed_offset(MIdx * m_step, K);
 
         const std::byte* A_base = LhsPackedData + LhsPackedStride * BIdx;
-        auto ATile = reinterpret_cast<const float*>(A_base + lhs_packed_offset);
+        auto ATile = reinterpret_cast<const void*>(A_base + lhs_packed_offset);
 
         auto TileSizeM = (MIdx + 1) * m_step > M ? (M - MIdx * m_step) : m_step;
         auto TileSizeN = (NIdx + 1) * n_step > N ? (N - NIdx * n_step) : n_step;
@@ -413,7 +413,6 @@ Return Value:
             g_kai_tls_sbgemm.output_tile.resize(tile_elems);
             out_tile = g_kai_tls_sbgemm.output_tile.data();
             out_row_stride_bytes = TileSizeN * sizeof(float);
-            std::fill_n(out_tile, tile_elems, 0.0f);
         }
 
         KLEIDIAI_KERNEL_LOG(sbgemm_gemm.name
--- a/onnxruntime/test/mlas/unittest/test_sbgemm.h
+++ b/onnxruntime/test/mlas/unittest/test_sbgemm.h
@@ -68,7 +68,7 @@ class MlasSBGemmTest : public MlasTestBase {
     }
     void* PackedB = BufferBPacked.GetBuffer(PackedBSize);
     if (std::is_same<BType, float>::value) {
-      MlasSBGemmConvertPackB(TransA, TransB, true, N, K, (const float*)B, ldb, PackedB, nullptr);
+      MlasSBGemmConvertPackB(TransA, TransB, BIsfp32, N, K, (const float*)B, ldb, PackedB, nullptr);
     } else {
     }
     return PackedB;


patryk-kaiser-ARM:

hariharans29:
Hi @patryk-kaiser-ARM / @damdoo01-arm - Can you please resolve conflicts for this PR if it is still on the agenda ? We can target merging this PR next. Thanks.

patryk-kaiser-ARM:
Hi @hariharans29 I resolved conflicts. This one is still on the agenda - I am currently investigating adding support for fastmath to more operators so that this change can have a larger impact, however it would be a good idea to get this one in first and then open up consequent PRs to bring more ops down this path for fastmath.

patryk-kaiser-ARM:
Can workflows be approved please

onnxruntime/core/mlas/lib/kleidiai/sbgemm_kleidiai.cpp
```diff
@@ -0,0 +1,370 @@
+//
+// SPDX-FileCopyrightText: Copyright 2025 Arm Limited and/or its affiliates <open-source-office@arm.com>
+//
+// SPDX-License-Identifier: MIT
+//
+
+#if defined(__aarch64__) && defined(__linux__)
```
hariharans29, line 7: Is this feature Linux only ?

onnxruntime/core/providers/cpu/math/matmul.cc
```diff
@@ -272,6 +272,7 @@ Status MatMul<float>::Compute(OpKernelContext* ctx) const {
       data[i].ldc = N;
       data[i].Bias = nullptr;
       data[i].OutputProcessor = nullptr;
+      data[i].BIsPacked = bool(packed_b_);
```
hariharans29: Please use static_cast<bool> ?

onnxruntime/core/mlas/lib/kleidiai/sbgemm_kleidiai.cpp
patryk-kaiser-ARM: The `#if defined(__aarch64__) && defined(__linux__)` guard was added for consistency with the existing sbgemm.h and related BF16/SBGEMM paths in the codebase. The `__linux__` define also covers Android.

onnxruntime/core/mlas/inc/mlas.h
```diff
@@ -1985,31 +1988,61 @@ struct MLAS_SBGEMM_DATA_PARAMS {
  * @return
  */
 void MLASCALL
-MlasSBGemmBatch(const size_t M, const size_t N, const size_t K, const size_t BatchN, const MLAS_SBGEMM_DATA_PARAMS* DataParams, MLAS_THREADPOOL* ThreadPool);
+MlasSBGemmBatch(
```
hariharans29, line 2034: Can you please rebase and add support for the backend selector that allows users to opt out of KleidiAI. Please see https://github.com/microsoft/onnxruntime/pull/27136. It should be trivial to follow the SGemm pattern in that PR for the SBGemm use-case that this PR is updating I think.

onnxruntime/core/mlas/lib/sbconv_kernel_neon.cpp
```diff
@@ -91,7 +91,7 @@ MlasConvPointwiseBf16KernelNeon(
         }
     }
 
-    MlasSBGemmBatch(OutputCount, BlockSize, BlockSize, idx, gemm_params, nullptr);
+    MlasSBGemmBatch(CblasNoTrans, CblasNoTrans, OutputCount, BlockSize, BlockSize, idx, gemm_params, nullptr, nullptr);
```
hariharans29: This should account for the backend selector as well ?

Can you please refer to my PR -  I think I had it plumbed through via KernelFlags for FP32 Conv and I used the kernel flags to pass in the appropriate backend selector instance to the Gemm call

cmake/onnxruntime_mlas.cmake
```diff
@@ -550,6 +551,23 @@ else()
           set_source_files_properties(${MLAS_SRC_DIR}/eltwise_kernel_neon_fp16.cpp PROPERTIES COMPILE_FLAGS " -march=armv8.2-a+fp16 ")
         endif()
 
+        #
+        # Mac arm64 (M-series) supports BF16 instructions, and ORT's SBGEMM
+        # entrypoints (MlasSBGemm*) are provided by the AArch64 SBGEMM sources.
+        # Add minimal changes required to enable sbgemm on this platform
+        #
+        if (APPLE AND CMAKE_OSX_ARCHITECTURES MATCHES "arm64")
```
hariharans29: Sorry, I am a bit confused by this - The cmake changes seem to indicate this is enabled on MacOS, but the source files have __linux__ in them. Is there an oversight in one of the 2 places ?

hariharans29: EDIT: I see that Copilot has a similar comment too

onnxruntime/core/mlas/lib/sbconv_kernel_neon.cpp
patryk-kaiser-ARM: This one won't go to KleidiAI anyway as it is part of MLAS_USE_ARM_NEON_NCHWC

cmake/onnxruntime_mlas.cmake
patryk-kaiser-ARM: Yes this was an oversight on my end, I was planning on enabling this for MacOS but so far I have only been able to test it on M4 so don't want to introduce problems in other platforms.
I’ll revert this to Linux only for now and remove the CMake change. Enabling this properly on Mac can be addressed as part of future work once it’s been properly validated.

onnxruntime/core/mlas/lib/sbconv_kernel_neon.cpp
hariharans29: But what if MLAS_USE_ARM_NEON_NCHWC was set to true and people wanted to opt out of KAI for the BLAS math ?

patryk-kaiser-ARM: The KAI path is guarded by `!defined(MLAS_USE_ARM_NEON_NCHWC)` as the kernel introduced in this PR is SME2 only, MLAS_USE_ARM_NEON_NCHWC should go down the default Neon SBGEMM path instead.
I agree we should still honor the backend selector so I will include proper handling here.

onnxruntime/core/providers/cpu/math/matmul.cc
```diff
@@ -202,7 +202,7 @@ Status MatMul<float>::PrePack(const Tensor& tensor, int input_idx, /*out*/ Alloc
       dim2 = static_cast<size_t>(b_shape[1]);
     }
 
-    if (use_fastmath_mode_ && (trans_b_attr_ == 0) && ((dim1 * dim2) >= kFastMathModeKernelsizeThreshold)) {
+    if (use_fastmath_mode_ && (trans_a_attr_ == 0) && (trans_b_attr_ == 0) && ((dim1 * dim2) >= kFastMathModeKernelsizeThreshold)) {
```
hariharans29, line 205: Is the no A transpose requirement only for the KAI path or applicable even for the fallback path ?

patryk-kaiser-ARM: I added this check in response to the Copilot comment for lines 289 to 290:
> This call can pass `TransA=CblasTrans` when `trans_a` is true, but the current SBGEMM implementations only support `CblasNoTrans` (and the default SBGEMM dispatch path doesn’t implement transpose handling). Consider preventing entry into the SBGEMM path when `trans_a` is true (or add proper transpose support) to avoid incorrect MatMul results.

The fallback for SBGEMM path (default path) would be through sbgemm_kernel_neon.cpp and there does not appear to be any transpose handling for A or B, I had to add the TransA and TransB params to the SBGEMM calls so they are only consumed by KAI at the moment, therefore I think it's safe to assume that the no transpose A requirement is applicable even for the fallback path.

hariharans29:
Hi @patryk-kaiser-ARM - Thanks for patiently incorporating the changes for all comments. I plan to take one last look at this today. Meanwhile, Copilot has some more comments. Please see if these are relevant. Hopefully, we can close this in a day or two.

onnxruntime/core/mlas/lib/sbgemm.h
```diff
@@ -299,11 +299,36 @@ MlasSBGemmGetDispatch()
...
+    bool BIsfp32,
+    size_t N,
+    size_t K,
+    const MLAS_BACKEND_KERNEL_SELECTOR_CONFIG* BackendKernelSelectorConfig
+)
 {
     //
     // Compute the number of bytes required to hold the packed buffer.
     //
+#if defined(USE_KLEIDIAI) && !defined(MLAS_USE_ARM_NEON_NCHWC)
```
hariharans29, line 314: Sorry - I am a bit confused with this. Why/How is the guard MLAS_USE_ARM_NEON_NCHWC relevant here ? (i.e.) Why is there code conditioned upon whether we support NCHWC Conv for ARM in an SBGemm implementation file ? What is the connection ?

hariharans29: Is it because the KAI route isn't able to handle some things from the Gemm params ?

hariharans29:
LGTM- Left behind some last set of comments + some comments from Copilot. Can be merged after that. Thanks !!

onnxruntime/core/mlas/lib/sbgemm.h
patryk-kaiser-ARM: The guard is relevant here because the NCHWC BF16 conv kernel (in sbconv_kernel_neon.cpp) calls MlasSBGemmBatch. So any SBGEMM override affects that conv path too. I added `!defined(MLAS_USE_ARM_NEON_NCHWC)` to avoid routing those Neon specific conv kernels through the KleidiAI override.

onnxruntime/core/providers/cpu/math/matmul.cc
patryk-kaiser-ARM: Hey Hari, if you think the explanation here is reasonable we can resolve this.

onnxruntime/core/mlas/lib/sbgemm.h
hariharans29: I see. Is the work to support the Conv path trivial or non-trivial (either way, no pressure in getting those changes into this PR)? I only ask because we may enable NCHWC ARM by default in the future if it meets all the performance bars and we may miss out on this optimization if it is turned ON.

hariharans29:
You may need this - https://github.com/microsoft/onnxruntime/pull/27618/

onnxruntime/core/mlas/lib/sbgemm.h
hariharans29, line 314: Any chance we can make the call from the SBConv to explicitly choose non-KAI path via the backend selector and just remove the compile time guard here so that the optimization itself is available on builds with NCHWC is enabled but the Conv call gets routed via the generic path ?

patryk-kaiser-ARM: Instead of reading the flag in sbconv we could just:
```
MLAS_BACKEND_KERNEL_SELECTOR_CONFIG mlas_backend_kernel_selector_config;
mlas_backend_kernel_selector_config.use_kleidiai = false;
```
Makes sense and it won't completely disable the KAI BF16 path for other fastmath operators.
Let me know what you think?

patryk-kaiser-ARM: You're right to raise this point and my main concern is that we need to test first to see how the KAI path will react to the Neon workload. I'd see it as possible future work.

onnxruntime/core/mlas/lib/sbconv_kernel_neon.cpp
```diff
@@ -44,6 +44,9 @@ MlasConvPointwiseBf16KernelNeon(
     const bool BiasAddition = (KernelFlags & MLAS_CONV_KERNEL_FLAG_BIAS_ADDITION) != 0;
     const bool ReluActivation = (KernelFlags & MLAS_CONV_KERNEL_FLAG_RELU_ACTIVATION) != 0;
 
+    MLAS_BACKEND_KERNEL_SELECTOR_CONFIG mlas_backend_kernel_selector_config;
+    mlas_backend_kernel_selector_config.use_kleidiai = ((KernelFlags & MLAS_CONV_KERNEL_MLAS_ARM_USE_KLEIDIAI) != 0);
```
patryk-kaiser-ARM, line 48: @hariharans29 Would this solution be acceptable:
` mlas_backend_kernel_selector_config.use_kleidiai = false;`
That way we won't allow this sbconv to go down KAI path, but it won't restrict other operators going down the MlasSBGemmBatch path from making use of the KAI acceleration.

hariharans29: Yes, that is what I had in mind too !

onnxruntime/core/mlas/lib/sbgemm.h
hariharans29: Yes, that is what I had in mind too !

PR #26774, opened 2025-12-11
Engine compatibility validity API implementation

umangb-09:
Added support for engine validation check for EP Context models.

### Motivation and Context
We wanted to implement the GetModelCompatibilityForEpDevices() API support and thus have an end user available API for the engine validation check for EP context models. Added this support and the necessary function implementation

### DIFF
--- a/onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
+++ b/onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
@@ -2585,6 +2585,12 @@ std::string NvExecutionProvider::GetCompiledModelCompatibilityInfo(
   // Protect read access to engine_headers_ for thread safety
   auto lock = GetApiLock();
 
+  // Compatibility info is only supported when there is exactly one engine.
+  // If multiple EPContext nodes/engines exist, return empty so validation is not applicable.
+  if (engine_headers_.size() > 1) {
+    return std::string();
+  }
+
   // If we have stored engine headers, return the first one found
   // (typically there's only one per EP context)
   if (!engine_headers_.empty()) {


umangb-09:
@ishwar-raut1 @gedoensmax Can you give this a go?

onnxruntime/core/providers/nv_tensorrt_rtx/onnx_ctx_model_helper.h
```diff
@@ -24,6 +24,11 @@ static const std::string PARTITION_NAME = "partition_name";
 static const std::string SDK_VERSION = "ep_sdk_version";
 static const std::string EPCONTEXT_OP_DOMAIN = "com.microsoft";
 
+constexpr size_t kTensorRTEngineHeaderSize = 64;
```
gedoensmax, line 28: Would it be possible to have this define in a TRT header ? This seems a bit risky to maintain if things change.

onnxruntime/core/providers/nv_tensorrt_rtx/onnx_ctx_model_helper.cc
```diff
@@ -14,6 +14,53 @@
 namespace onnxruntime {
 extern TensorrtLogger& GetTensorrtLogger(bool verbose_log);
 
+/*
+ * Convert binary data to hex string
+ */
+std::string BinaryToHexString(const void* data, size_t size) {
```
gedoensmax, line 20: Is the motivation to storing this as a a hex string that it is safer depending on how the string is implemented ? I mean we could have a char string with just the bytes inside right ? just curious :)

gedoensmax (commented):
Looks good but we are missing testing for this change.

umangb-09:
Moving this to draft as there is new approach recommended by Microsoft. Will raise new PR changes with that

onnxruntime/core/providers/nv_tensorrt_rtx/onnx_ctx_model_helper.h
umangb-09: Should be. Will check with TRT team. Will add a risk command meanwhile.

onnxruntime/core/providers/nv_tensorrt_rtx/onnx_ctx_model_helper.cc
umangb-09: yes, hex is safer for storage/transport as a string attribute.
We’re putting the header into the EPContext node as a string attribute. That path expects UTF‑8/printable data and can be sensitive to embedded nulls or non‑printable bytes. 
Let me know if this makes sense. Seems reasonable to me. ;)

onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
```diff
@@ -2578,6 +2578,80 @@ const InlinedVector<const Node*> NvExecutionProvider::GetEpContextNodes() const
...
+    return Status::OK();
+  }
+
+  // Use TensorRT RTX's getEngineValidity to check compatibility
+  uint64_t diagnostics = 0;
+  nvinfer1::EngineValidity validity = runtime_->getEngineValidity(
+      engine_header.data(),
+      engine_header.size(),
+      &diagnostics
+  );
```
adrianlizarraga: Looks like CI is catching some linter errors. Could you please fix those with the lintrunner tool? 
https://github.com/microsoft/onnxruntime/blob/main/docs/Coding_Conventions_and_Standards.md#linting
https://github.com/microsoft/onnxruntime/actions/runs/21199581486/job/61300326074?pr=26774

onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc
```diff
@@ -2578,6 +2578,78 @@ const InlinedVector<const Node*> NvExecutionProvider::GetEpContextNodes() const
...
 
+std::string NvExecutionProvider::GetCompiledModelCompatibilityInfo(
+    const onnxruntime::GraphViewer& graph_viewer) const {
+  ORT_UNUSED_PARAMETER(graph_viewer);
+
+  // Protect read access to engine_headers_ for thread safety
+  auto lock = GetApiLock();
+
+  // If we have stored engine headers, return the first one found
+  // (typically there's only one per EP context)
```
adrastogi, line 2595: @umangb-09 I was thinking more about your question related to handling multiple EPContext nodes; could the implementation enumerate those from the graph (i.e. via the map you're populating), and combine the resulting headers to form the single string? And then in the validation routine, each engine header could be processed and take the "worst-case" as the overall result? e.g., if any one EPContext node is marked as `kInvalid`, then the validation routine would return unsupported to the caller. Would an approach like this work?

adrianlizarraga: Perhaps to be safe, this implementation should not return compatibility info for models with multiple EPContext nodes (i..e, EP_NOT_APPLICABLE). The reason being that this is not completely supported by this implementation at the moment.

Would this work and would it be a simple update to this PR?

Edit: If adrastogi's suggestion to combine headers is feasible, that would probably be preferred.

umangb-09: @adrastogi yes, seems this would be a feasible solution. But this will require some intensive check and changes which I think is out of scope for this PR. For now, we can either go with the approach that we do 'one engine luck' based check, if the engine is incompatible then we return failure else we will have a failure in session time.

Edit : The suggestion by @adrianlizarraga is feasible for this PR change with minimal effect because of time constraint.

PR #26776, opened 2025-12-12
Address security issue of loading arbitrary files as external data

yuslepukhin:
### Description
Verify external data references in TensorProto specify data location that is under the model directory structure, reject absolute paths and paths that escape the model path.
Make the validation function available to bridge based EPs.
Expose ExternalDataInfo via a bridge to some EPs that choose to handle the data itself.

### Motivation and Context
This is a security concern.

### DIFF
--- a/onnxruntime/core/framework/tensorprotoutils.cc
+++ b/onnxruntime/core/framework/tensorprotoutils.cc
@@ -311,15 +311,19 @@ Status ValidateExternalDataPath(const std::filesystem::path& base_dir,
   // Reject absolute paths
   ORT_RETURN_IF(location.is_absolute(),
                 "Absolute paths not allowed for external data location");
-  // Resolve and verify the path stays within model directory
-  auto resolved = std::filesystem::weakly_canonical(base_dir / location);
-  auto base_canonical = std::filesystem::weakly_canonical(base_dir);
-  // Check that resolved path starts with base directory
-  auto [base_end, resolved_it] = std::mismatch(
-      base_canonical.begin(), base_canonical.end(),
-      resolved.begin(), resolved.end());
-  ORT_RETURN_IF(base_end != base_canonical.end(),
-                "External data path: ", location, " escapes model directory: ", base_dir);
+  if (!base_dir.empty()) {
+    // Resolve and verify the path stays within model directory
+    auto base_canonical = std::filesystem::weakly_canonical(base_dir);
+    // If the symlink exists, it resolves to the target path;
+    // so if the symllink is outside the directory it would be caught here.
+    auto resolved = std::filesystem::weakly_canonical(base_dir / location);
+    // Check that resolved path starts with base directory
+    auto [base_end, resolved_it] = std::mismatch(
+        base_canonical.begin(), base_canonical.end(),
+        resolved.begin(), resolved.end());
+    ORT_RETURN_IF(base_end != base_canonical.end(),
+                  "External data path: ", location, " escapes model directory: ", base_dir);
+  }
   return Status::OK();
 }
 
--- a/onnxruntime/core/framework/tensorprotoutils.h
+++ b/onnxruntime/core/framework/tensorprotoutils.h
@@ -523,6 +523,7 @@ Status TensorProtoWithExternalDataToTensorProto(
 
 /// <summary>
 /// The functions will make sure the 'location' specified in the external data is under the 'base_dir'.
+/// If the `base_dir` is empty, the function only ensures that `location` is not an absolute path.
 /// </summary>
 /// <param name="base_dir">model location directory</param>
 /// <param name="location">location is a string retrieved from TensorProto external data that is not
--- a/onnxruntime/core/providers/shared_library/provider_interfaces.h
+++ b/onnxruntime/core/providers/shared_library/provider_interfaces.h
@@ -38,11 +38,6 @@ struct ProviderHost;
 struct ProviderHostCPU;
 
 class ExternalDataInfo;
-#ifdef _WIN32
-using OFFSET_TYPE = int64_t;
-#else
-using OFFSET_TYPE = off_t;
-#endif
 
 class PhiloxGenerator;
 using ProviderType = const std::string&;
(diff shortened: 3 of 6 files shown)


onnxruntime/core/graph/graph.cc
```diff
@@ -3724,9 +3724,14 @@ Status Graph::ConvertInitializersIntoOrtValues() {
   std::vector<Graph*> all_subgraphs;
   FindAllSubgraphs(all_subgraphs);
 
+  const auto& model_path = GetModel().ModelPath();
+  PathString model_dir;
+  if (!model_path.empty()) {
```
tianleiwu, line 3729: Suggest to add a comment:
When model path is empty (for example, the model proto is from memory instead of file), the model_dir will be empty, then weakly_canonical will be current working directory.

tianleiwu:
From AI:
---

### **Summary**

This is a critical security patch that mitigates **Path Traversal (CWE-22)** vulnerabilities when loading ONNX models. By adding a validation step for external data paths, the changes ensure that referenced files must reside within the model's directory tree, preventing malicious models from accessing arbitrary system files (e.g., `/etc/passwd`).

The implementation correctly uses C++17 filesystem capabilities to canonicalize paths (resolving `..` and symlinks) before verification. Additionally, the changes expose these validation mechanisms and the `ExternalDataInfo` structure to Execution Providers (EPs) via the shared library bridge, ensuring EPs can also handle external data securely.

### **Technical Analysis**

**1. Path Validation Logic (`tensorprotoutils.cc`)**

  * **Correct Canonicalization:** The use of `std::filesystem::weakly_canonical` is the correct approach here. It resolves dot components (`.` and `..`) and symbolic links. `weakly_canonical` is preferred over `canonical` because the latter throws an error if the file doesn't exist, whereas we might want to validate the path structure even if the file is missing (or to provide a better error message).
  * **Directory Containment Check:** The logic using `std::mismatch` to compare the `base_canonical` iterator with the `resolved` iterator effectively verifies that the resolved path is a subdirectory (or the same directory) of the base model path.
  * **Absolute Path Rejection:** Explicitly rejecting absolute paths (`location.is_absolute()`) is a good defense-in-depth measure, as external data in ONNX is defined to be relative.

**2. Integration into Graph Resolution (`graph.cc`)**

  * The check is injected into `Graph::ConvertInitializersIntoOrtValues`. This is the standard path where initializers are processed. By placing the check here, it protects the default ORT loading mechanism.
  * **Logic Check:**
    ```cpp
    auto st = utils::ValidateExternalDataPath(model_dir, location);
    ```
    This ensures validation happens *before* any attempt to open or memory-map the file implies the check works as a gatekeeper.

**3. Provider Bridge Updates (`provider_*.h/cc`)**

  * **Exposing Validation:** Adding `Utils__ValidateExternalDataPath` to the `ProviderHost` interface allows shared library Execution Providers to perform the same security checks if they implement custom model loading logic.
  * **`ExternalDataInfo` Wrapper:** Exposing the `ExternalDataInfo` class allows EPs to access parsed external data properties (offset, length, checksum) safely without parsing the raw Protobuf fields manually, which reduces the risk of parsing inconsistencies or errors in EPs.

**4. Testing (`test_inference.cc`, `test_arbitrary_external_file.py`)**

  * The test coverage is appropriate.
      * **Python Generation:** The script correctly generates a "malicious" model using standard ONNX tools but manually injecting the path traversal payload (`../../../../../../../etc/passwd`).
      * **C++ Test:** The test loads this model and asserts that it fails specifically with the expected security error message ("escapes model directory").

### **Specific Code Feedback**

**`onnxruntime/core/framework/tensorprotoutils.cc`**

```cpp
Status ValidateExternalDataPath(const std::filesystem::path& base_dir,
                                const std::filesystem::path& location) {
  // ...
  // Resolve and verify the path stays within model directory
  auto resolved = std::filesystem::weakly_canonical(base_dir / location);
  auto base_canonical = std::filesystem::weakly_canonical(base_dir);
  // ...
}
```

  * **Note on Empty `base_dir`:** If `base_dir` is empty (e.g., for an in-memory model where `GetModel().ModelPath()` returned empty), `weakly_canonical("")` typically resolves to the Current Working Directory (CWD). This implies that for in-memory models, external data references will be restricted to the CWD. This is generally the expected behavior but is worth noting.

**`onnxruntime/core/graph/graph.cc`**

```cpp
+          const auto& location = external_data_info->GetRelPath();
+          auto st = utils::ValidateExternalDataPath(model_dir, location);
```

  * **Optimization:** The validation is performed inside the loop iterating over initializers. `std::filesystem::weakly_canonical` can be expensive (filesystem calls). Since `base_canonical` (`model_dir`) is constant for the whole graph, you might consider computing `base_canonical` once outside the loop and passing it in, rather than re-computing it for every external initializer. However, given this is model load time, the performance impact is likely negligible unless there are thousands of external files.

**`onnxruntime/test/testdata/test_arbitrary_external_file.py`**

```python
    # Location entry - attempts to access system passwd file
    entry1 = evil_tensor.external_data.add()
    entry1.key = "location"
    entry1.value = "../../../../../../../etc/passwd"
```

  * **Cross-Platform Tests:** While `/etc/passwd` is specific to \*nix, the test expects a generic exception regarding path traversal. Ensure this test passes on Windows CI as well (Windows might interpret this path differently or `weakly_canonical` behaves differently regarding drive letters, but usually `..` traversal works similarly enough for this negative test).

### **Conclusion**

The change is well-implemented and addresses the security vulnerability effectively. The additions to the bridge API ensure that the security posture extends to plugins/EPs.

**Action Item:**

  * **Approve**.
  * (Optional) If performance on model load with many external files is a concern, verify if caching `base_canonical` in `ConvertInitializersIntoOrtValues` yields benefits.

tianleiwu (dismissed):

cau-git:
Since the release of onnxruntime following this PR, we are observing failure modes in very commonplace practices of using HuggingFace models. HuggingFace models, some of which provide ONNX exports, download to the HF cache (`~/.cache/huggingface` as default). This now breaks on onnxruntime, because the model versioning is realized with internal symlinks that reference to blobs outside the model's base directory in the cache folder. It would be great if this can be relaxed or optionally disabled.

istupakov:
> Since the release of onnxruntime following this PR, we are observing failure modes in very commonplace practices of using HuggingFace models. HuggingFace models, some of which provide ONNX exports, download to the HF cache (`~/.cache/huggingface` as default). This now breaks on onnxruntime, because the model versioning is realized with internal symlinks that reference to blobs outside the model's base directory in the cache folder. It would be great if this can be relaxed or optionally disabled.

I have the same problem. This PR broke loading of large models in my onnx-asr library!
CC: @yuslepukhin, @tianleiwu

yuslepukhin:
@istupakov would add a list of allowed folders to session options be an acceptable solution? Reverting is not an option.

cau-git:
@yuslepukhin  it would definitely help if one could provide a root path in which loading model files is allowed.

istupakov:
I agree with @cau-git – being able to specify a root path (e.g., `~/.cache/huggingface`) solves the problem. Specifying every possible allowed folder would be quite inconvenient – ​​I wouldn't want to mess around with resolving symlinks.

PR #26779, opened 2025-12-12
Prevent concurrent OM read/write race by introducing file-based locking during dump and load (#26778)

### LINKED ISSUE #26778: [Build] NPU Multi-Card Concurrent Compilation Causes OM File Race Condition (Read-While-Write Error)
### Describe the issue

When compiling an ONNX model to OM on multiple NPUs concurrently, we observed a reproducible race conditon: the faster card finished OM generation first and starts writing the file, while slower cards detect that the OM file has appeared and immediately attempt to load it. 

We also created a script where card 0 compiles first and all other cards continuously poll the directory; as soon as the OM file shows up, the other cards start inference, and the error reproduces reliably. Because the file is still being written by the fast card, the slower cards read an incomplete OM, leading to model-load failures or accuracy issues. 

### Urgency

_No response_

### Target platform

Linux x86_64
CANN:8.3.RC1
torch:2.5.1+cpu
torch_npu:2.5.1
onnxruntime-cann:1.24.0 (This requires building ONNX Runtime from source)

### Build script
inswapper onnx :https://huggingface.co/Devia/G/blob/main/inswapper_128.onnx

python script:
```
import os
import time
import glob
import numpy as np
import torch
import torch_npu
import onnxruntime as ort

RESWAPPER_PATH = "./insightface/inswapper_128.onnx"
OM_DIR = "./"
POLL_INTERVAL = 0.2
MAX_WAIT = 300

def list_om():
    return set(glob.glob(os.path.join(OM_DIR, "*.om")))

def main():
    local_rank = int(os.environ.get("LOCAL_RANK", "0"))
    torch.npu.set_device(local_rank)
    
    sess = ort.InferenceSession(
        RESWAPPER_PATH,
        providers=[
            (
                "CANNExecutionProvider",
                {
                    "device_id":local_rank ,
                    "arena_extend_strategy": "kNextPowerOfTwo",
                    "enable_cann_graph": True,
                    "precision_mode": "must_keep_orgin_dtype",
                    "op_select_impl_mode": "high_precision",
                },
            ),
        ],
    )
    
    target = np.random.randn(1, 3, 128, 128).astype(np.float32)
    source = np.random.randn(1, 512).astype(np.float32)
    feed = {sess.get_inputs()[0].name: target, sess.get_inputs()[0].name: source}
    
    init = list_om()
    
    if local_rank == 0:
        sess.run(None, feed)
    else:
        start = time.time()
        while True:
            now = list_om()
            if now - init:
                break
            if time.time() - start > MAX_WAIT:
                break
        time.sleep(POLL_INTERVAL)
    
    out = sess.run(None, feed)
    
    
if __name__ == "__main__":
    main()
```

shell script:
``` torchrun --nproc_per_node=8 xxx.py```

### Error / output

Status Message: CANN error executing aclmdLoadFromFile()

### Visual Studio Version

_No response_

### GCC / Compiler Version

_No response_

Xiangyanglikecode:
### Description
This PR introduces a file-based locking mechanism using `flock` to serialize OM file access:
**1.  Exclusive lock during OM dump**
When dumping the OM model:
- Open the target `.om` file
- Acquire an exlusive lock (`LOCK_EX`)
- Perform `Save`
- Release the lock after the dump completes

**2.  Shared lock during OM load**
When loading an existing OM model from disk:
- Open the same `.om` file
- Acquire an exlusive lock (`LOCK_SH`)
- Perform `Load`
- Release the lock after the load completes

### Motivation and Context
Fixed #26778

### DIFF
--- a/onnxruntime/core/providers/cann/cann_execution_provider.cc
+++ b/onnxruntime/core/providers/cann/cann_execution_provider.cc
@@ -1403,9 +1403,9 @@ Status CANNExecutionProvider::Compile(const std::vector<FusedNodeAndGraph>& fuse
         std::lock_guard<std::mutex> lock(g_mutex);
         auto filename_with_suffix = cann::MatchFile(filename);
         if (!filename_with_suffix.empty()) {
-          std::string om_flie = filename + ".om";
+          std::string om_file = filename + ".om";
 
-          int fd = open(om_file.c_str(), O_RDWR | O_CREATE, 0666);
+          int fd = open(om_file.c_str(), O_RDWR | O_CREAT, 0666);
 
           if (fd >= 0){
             flock(fd, LOCK_SH);
--- a/onnxruntime/core/providers/cann/cann_graph.cc
+++ b/onnxruntime/core/providers/cann/cann_graph.cc
@@ -122,7 +122,7 @@ Status BuildONNXModel(ge::Graph& graph, std::string input_shape, const char* soc
   if (info.dump_om_model) {
     std::string final_file = file_name + ".om";
 
-    int lock_fd = open(final_flie.c_str(), O_CREATE | O_RDWR, 0666);
+    int lock_fd = open(final_file.c_str(), O_CREAT | O_RDWR, 0666);
 
     if (flock(lock_fd, LOCK_EX) == 0){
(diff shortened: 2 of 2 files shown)


Xiangyanglikecode:
@microsoft-github-policy-service agree company="HUAWEI"

PR #26781, opened 2025-12-12
Add API to get ep graph partitioning info

adrianlizarraga:
### Description
- Adds API functions to get information about the subgraphs/nodes assigned to the EPs in the session.
  - `Session_GetEpGraphAssignmentInfo`: Returns a list of "subgraphs", each with information about the assigned EP and nodes.
    - Note:  App must enable session configuration `"session.record_ep_graph_assignment_info"` to signal ORT to collect this information. If not enabled, API returns empty results.
  - `EpAssignedSubgraph_GetEpName`: Returns the name of the EP to which the subgraph is assigned
  - `EpAssignedSubgraph_GetNodes`: Returns a list of assigned nodes
  - `EpAssignedNode_GetName`: Returns the assigned node's name
  - `EpAssignedNode_GetDomain`: Returns the assigned node's domain
  - `EpAssignedNode_GetOperatorType`: Returns the assigned node's operator type
- Also adds C++ and Python bindings

#### Structure of returned information
The API returns a list of "subgraphs". Each subgraph has the following information:
- Subgraph info:
  - EP name: The name of the execution provider to which this subgraph is assigned.
  - nodes: Name and operator type of each node. Ex: `[{"multiply", "Mul"}, ...]`

Python example program (taken from unit tests):
```python
    def test_get_graph_provider_assignment_info(self):
        """
        Tests querying for information about the nodes assigned to the CPU EP.
        """

        # Create session options that enables recording EP graph partitioning info.
        session_options = onnxrt.SessionOptions()
        session_options.add_session_config_entry("session.record_ep_graph_assignment_info", "1")

        session = onnxrt.InferenceSession(get_name("add_mul_add.onnx"), sess_options=session_options)

        # Query session for information on each subgraph assigned to an EP.
        ep_subgraphs = session.get_provider_graph_assignment_info()

        # Check that all 3 nodes are assigned to CPU EP (each in its own subgraph)
        self.assertEqual(len(ep_subgraphs), 3)
        for ep_subgraph in ep_subgraphs:
            self.assertEqual(ep_subgraph.ep_name, "CPUExecutionProvider")
            self.assertEqual(len(ep_subgraph.get_nodes()), 1)

        # Serialize each node to an identifier (concatenates operator type and node name)
        node_ids: list[str] = [f"{n.op_type}/{n.name}" for s in ep_subgraphs for n in s.get_nodes()]

        # Should have 1 Mul and 2 Adds.
        self.assertEqual(len(node_ids), 3)
        self.assertIn("Add/add_0", node_ids)
        self.assertIn("Add/add_1", node_ids)
        self.assertIn("Mul/mul_0", node_ids)
```

C++ program (taken from unit test):
```c++
  // Check the ep graph partitioning (Mul on plugin EP, others on CPU EP).
  // Model has 3 subgraphs (in no particular order):
  // - Subgraph 1: Add assigned to CPU EP.
  // - Subgraph 2: Mul assigned to plugin EP.
  // - Subgraph 3: Add assigned to CPU EP.
  std::vector<Ort::ConstEpAssignedSubgraph> ep_subgraphs = session.GetEpGraphAssignmentInfo();
  ASSERT_EQ(ep_subgraphs.size(), 3);

  for (Ort::ConstEpAssignedSubgraph ep_subgraph : ep_subgraphs) {
    std::string ep_name = ep_subgraph.EpName();
    ASSERT_TRUE(ep_name == Utils::example_ep_info.ep_name || ep_name == kCpuExecutionProvider);

    const std::vector<Ort::ConstEpAssignedNode> ep_nodes = ep_subgraph.GetNodes();
    ASSERT_GE(ep_nodes.size(), 1);  // All of these subgraphs just have one node.

    if (ep_name == kCpuExecutionProvider) {
      std::string op_type = ep_nodes[0].OpType();
      std::string node_name = ep_nodes[0].Name();

      ASSERT_EQ(op_type, "Add");
      ASSERT_TRUE(node_name == "add_0" || node_name == "add_1");
    } else {
      ASSERT_TRUE(ep_name == Utils::example_ep_info.ep_name);

      std::string op_type = ep_nodes[0].OpType();
      std::string node_name = ep_nodes[0].Name();
      ASSERT_EQ(op_type, "Mul");
      ASSERT_EQ(node_name, "mul_0");
    }
  }
```

### DIFF
--- a/include/onnxruntime/core/session/onnxruntime_c_api.h
+++ b/include/onnxruntime/core/session/onnxruntime_c_api.h
@@ -7052,11 +7052,15 @@ struct OrtApi {
   /** \brief Get the name of the execution provider to which the subgraph was assigned.
    *
    * \param[in] ep_subgraph The OrtEpAssignedSubgraph instance.
-   * \return The execution provider's name as UTF-8 null-terminated string.
+   * \param[out] out Output parameter set to the execution provider's name as a UTF-8 null-terminated string.
+   *                 Owned by the OrtEpAssignedSubgraph instance (do not free).
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
    *
    * \since Version 1.24.
    */
-  ORT_API_T(const char*, EpAssignedSubgraph_GetEpName, _In_ const OrtEpAssignedSubgraph* ep_subgraph);
+  ORT_API2_STATUS(EpAssignedSubgraph_GetEpName, _In_ const OrtEpAssignedSubgraph* ep_subgraph,
+                  _Outptr_ const char** out);
 
   /** \brief Get the nodes in a subgraph assigned to a specific execution provider.
    *
@@ -7074,29 +7078,38 @@ struct OrtApi {
   /** \brief Get the name of the node assigned to an execution provider.
    *
    * \param[in] ep_node The OrtEpAssignedNode instance.
-   * \return The node's name as a UTF-8 null-terminated string.
+   * \param[out] out Output parameter set to the node's name as a UTF-8 null-terminated string.
+   *                 Owned by the OrtEpAssignedNode instance (do not free).
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
    *
    * \since Version 1.24.
    */
-  ORT_API_T(const char*, EpAssignedNode_GetName, _In_ const OrtEpAssignedNode* ep_node);
+  ORT_API2_STATUS(EpAssignedNode_GetName, _In_ const OrtEpAssignedNode* ep_node, _Outptr_ const char** out);
 
   /** \brief Get the domain of the node assigned to an execution provider.
    *
    * \param[in] ep_node The OrtEpAssignedNode instance.
-   * \return The node's domain as a UTF-8 null-terminated string.
+   * \param[out] out Output parameter set to the node's domain as a UTF-8 null-terminated string.
+   *                 Owned by the OrtEpAssignedNode instance (do not free).
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
    *
    * \since Version 1.24.
    */
-  ORT_API_T(const char*, EpAssignedNode_GetDomain, _In_ const OrtEpAssignedNode* ep_node);
+  ORT_API2_STATUS(EpAssignedNode_GetDomain, _In_ const OrtEpAssignedNode* ep_node, _Outptr_ const char** out);
 
   /** \brief Get the operator type of the node assigned to an execution provider.
    *
    * \param[in] ep_node The OrtEpAssignedNode instance.
-   * \return The node's operator type as a UTF-8 null-terminated string.
+   * \param[out] out Output parameter set to the node's operator type as a UTF-8 null-terminated string.
+   *                 Owned by the OrtEpAssignedNode instance (do not free).
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
    *
    * \since Version 1.24.
    */
-  ORT_API_T(const char*, EpAssignedNode_GetOperatorType, _In_ const OrtEpAssignedNode* ep_node);
+  ORT_API2_STATUS(EpAssignedNode_GetOperatorType, _In_ const OrtEpAssignedNode* ep_node, _Outptr_ const char** out);
 };
 
 /*
--- a/include/onnxruntime/core/session/onnxruntime_cxx_inline.h
+++ b/include/onnxruntime/core/session/onnxruntime_cxx_inline.h
@@ -750,7 +750,11 @@ inline EpDevice::EpDevice(OrtEpFactory& ep_factory, ConstHardwareDevice& hardwar
 namespace detail {
 template <typename T>
 inline std::string EpAssignedSubgraphImpl<T>::GetEpName() const {
-  return std::string(GetApi().EpAssignedSubgraph_GetEpName(this->p_));
+  const char* ep_name = nullptr;
+
+  // Returned null-terminated string will not be null if API function returns successfully.
+  ThrowOnError(GetApi().EpAssignedSubgraph_GetEpName(this->p_, &ep_name));
+  return std::string(ep_name);
 }
 
 template <typename T>
@@ -772,17 +776,29 @@ inline std::vector<ConstEpAssignedNode> EpAssignedSubgraphImpl<T>::GetNodes() co
 
 template <typename T>
 inline std::string EpAssignedNodeImpl<T>::GetName() const {
-  return std::string(GetApi().EpAssignedNode_GetName(this->p_));
+  const char* node_name = nullptr;
+
+  // Returned null-terminated string will not be null if API function returns successfully.
+  ThrowOnError(GetApi().EpAssignedNode_GetName(this->p_, &node_name));
+  return std::string(node_name);
 }
 
 template <typename T>
 inline std::string EpAssignedNodeImpl<T>::GetDomain() const {
-  return std::string(GetApi().EpAssignedNode_GetDomain(this->p_));
+  const char* domain = nullptr;
+
+  // Returned null-terminated string will not be null if API function returns successfully.
+  ThrowOnError(GetApi().EpAssignedNode_GetDomain(this->p_, &domain));
+  return std::string(domain);
 }
 
 template <typename T>
 inline std::string EpAssignedNodeImpl<T>::GetOperatorType() const {
-  return std::string(GetApi().EpAssignedNode_GetOperatorType(this->p_));
+  const char* op_type = nullptr;
+
+  // Returned null-terminated string will not be null if API function returns successfully.
+  ThrowOnError(GetApi().EpAssignedNode_GetOperatorType(this->p_, &op_type));
+  return std::string(op_type);
 }
 }  // namespace detail
 
--- a/onnxruntime/core/session/onnxruntime_c_api.cc
+++ b/onnxruntime/core/session/onnxruntime_c_api.cc
@@ -968,14 +968,24 @@ ORT_API_STATUS_IMPL(OrtApis::Session_GetEpGraphAssignmentInfo, _In_ const OrtSes
   API_IMPL_END
 }
 
-ORT_API(const char*, OrtApis::EpAssignedSubgraph_GetEpName, _In_ const OrtEpAssignedSubgraph* ep_subgraph) {
+ORT_API_STATUS_IMPL(OrtApis::EpAssignedSubgraph_GetEpName, _In_ const OrtEpAssignedSubgraph* ep_subgraph,
+                    _Outptr_ const char** out) {
+  API_IMPL_BEGIN
 #if !defined(ORT_MINIMAL_BUILD)
-  return ep_subgraph->ep_name.c_str();
+  if (out == nullptr) {
+    return OrtApis::CreateStatus(ORT_INVALID_ARGUMENT,
+                                 "EpAssignedSubgraph_GetEpName requires a valid (non-null) `out` output parameter "
+                                 "into which to store the EP name string.");
+  }
+
+  *out = ep_subgraph->ep_name.c_str();
+  return nullptr;
 #else
   ORT_UNUSED_PARAMETER(ep_subgraph);
-  fprintf(stderr, "EP graph assignment information is not supported in this build\n");
-  return nullptr;
+  ORT_UNUSED_PARAMETER(out);
+  return OrtApis::CreateStatus(ORT_NOT_IMPLEMENTED, "EP graph assignment information is not supported in this build");
 #endif  // !defined(ORT_MINIMAL_BUILD)
+  API_IMPL_END
 }
 
 ORT_API_STATUS_IMPL(OrtApis::EpAssignedSubgraph_GetNodes, _In_ const OrtEpAssignedSubgraph* ep_subgraph,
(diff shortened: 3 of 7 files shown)


onnxruntime/core/session/inference_session.cc
```diff
@@ -1448,6 +1489,14 @@ common::Status InferenceSession::TransformGraph(onnxruntime::Graph& graph, bool
                                                        session_options_.config_options, *session_logger_,
                                                        mode, session_options_.GetEpContextGenerationOptions(), debug_graph_fn));
 
+#if !defined(ORT_MINIMAL_BUILD)
+  if (record_ep_graph_partitioning) {
+    for (std::unique_ptr<OrtEpAssignedSubgraph>& ep_subgraph : ep_graph_assignment_info_storage_) {
+      ep_graph_assignment_info_.push_back(ep_subgraph.get());
```
edgchen1: can this be done in `on_partition_assign_fn`? it'd be simpler to keep the related code together.

onnxruntime/core/session/inference_session.cc
```diff
@@ -1301,12 +1329,25 @@ common::Status InferenceSession::TransformGraph(onnxruntime::Graph& graph, bool
   // 8. Repeat steps 5 to 7 depending on the graph optimizations loop level.
   // 9. insert copy nodes (required transformer).
 
+  OnPartitionAssignmentFunction on_partition_assign_fn;
+#if !defined(ORT_MINIMAL_BUILD)
+  bool record_ep_graph_partitioning =
+      session_options_.config_options.GetConfigOrDefault(kOrtSessionOptionsRecordEpGraphAssignmentInfo, "0") == "1";
+  if (record_ep_graph_partitioning) {
+    on_partition_assign_fn = [this](const Graph& graph, const ComputeCapability& assigned_subgraph,
```
edgchen1: naming nit: `on_partition_assignment_fn`

onnxruntime/core/session/onnxruntime_c_api.cc
```diff
@@ -872,6 +873,88 @@ ORT_API_STATUS_IMPL(OrtApis::RunAsync, _Inout_ OrtSession* sess, _In_opt_ const
...
+#if !defined(ORT_MINIMAL_BUILD)
+  if (ep_subgraphs == nullptr) {
+    return OrtApis::CreateStatus(ORT_INVALID_ARGUMENT, "'ep_subgraphs' argument is null");
+  }
+
+  if (num_ep_subgraphs == nullptr) {
+    return OrtApis::CreateStatus(ORT_INVALID_ARGUMENT, "'num_ep_subgraphs' argument is null");
+  }
+
+  auto inference_session = reinterpret_cast<const onnxruntime::InferenceSession*>(session);
```
edgchen1: general: `auto*` for pointers

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -6912,6 +6914,69 @@ struct OrtApi {
...
+   * \return The execution provider name.
+   *
+   * \since Version 1.24.
+   */
+  const char*(ORT_API_CALL* EpAssignedSubgraph_GetEpName)(_In_ const OrtEpAssignedSubgraph* ep_subgraph);
+
+  /** \brief Get the list of nodes assigned to an execution provider.
+   *
+   * \param[in] ep_subgraph The OrtEpAssignedSubgraph instance to query.
+   * \param[out] ep_nodes Output parameter set to the list of OrtEpAssignedNode instances.
```
edgchen1: why do we introduce a new node type instead of reusing `OrtNode`?

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -6912,6 +6914,69 @@ struct OrtApi {
...
+  const char*(ORT_API_CALL* EpAssignedNode_GetName)(_In_ const OrtEpAssignedNode* ep_node);
+
+  /** \brief Get the operator type of the node assigned to an execution provider.
+   *
+   * \param[in] ep_node The OrtEpAssignedNode instance to query.
+   * \return The node's operator type.
+   *
+   * \since Version 1.24.
+   */
+  const char*(ORT_API_CALL* EpAssignedNode_GetOperatorType)(_In_ const OrtEpAssignedNode* ep_node);
```
edgchen1: maybe we should also provide a way to get the domain. domain + op type identifies the op

include/onnxruntime/core/session/onnxruntime_session_options_config_keys.h
```diff
@@ -415,3 +415,11 @@ static const char* const kOrtSessionOptionsFailOnSuboptimalCompiledModel =
 // "high_power_saver", "low_balanced", "extreme_power_saver", "low_power_saver", "power_saver",
 // "sustained_high_performance". Default to "default".
 static const char* const kOrtEpDynamicOptionsQnnHtpPerformanceMode = "ep.dynamic.qnn_htp_performance_mode";
+
+// Enables the session to record information about the subgraphs/nodes assigned to execution providers.
+// When enabled, an application may call Session_GetEpGraphAssignmentInfo() to retrieve the information.
```
edgchen1, line 420: why is it disabled by default? is it expected to typically be too expensive to collect or store this information?

onnxruntime/test/autoep/test_execution.cc
```diff
@@ -223,6 +224,37 @@ void RunPartiallySupportedModelWithPluginEp(const Ort::SessionOptions& session_o
   // This model has Add -> Mul -> Add. The example plugin EP supports Mul but not Add.
   Ort::Session session(*ort_env, ORT_TSTR("testdata/add_mul_add.onnx"), session_options);
 
+  // Check that the ep graph partitioning (Mul on plugin EP, others on CPU EP).
```
edgchen1: nice, thanks for adding this. good to be able to check that it is actually a partially supported model now.

onnxruntime/test/autoep/test_execution.cc
```diff
@@ -289,6 +321,8 @@ TEST(OrtEpLibrary, PluginEp_AppendV2_PartiallySupportedModelInference) {
   // Create session with example plugin EP
   Ort::SessionOptions session_options;
   std::unordered_map<std::string, std::string> ep_options;
+
+  session_options.AddConfigEntry(kOrtSessionOptionsRecordEpGraphAssignmentInfo, "1");
```
edgchen1, line 302: might be good to set this option closer to the code that depends on it. e.g., put it in `RunPartiallySupportedModelWithPluginEp` or move the EP assignment info checks to this scope.

include/onnxruntime/core/session/onnxruntime_cxx_api.h
```diff
@@ -1164,6 +1164,38 @@ OrtCompiledModelCompatibility GetModelCompatibilityForEpDevices(
     const std::vector<ConstEpDevice>& ep_devices,
     const char* compatibility_info);
 
+namespace detail {
+template <typename T>
+struct EpAssignedNodeImpl : Ort::detail::Base<T> {
+  using B = Ort::detail::Base<T>;
+  using B::B;
+
+  const char* GetName() const;
```
edgchen1: return `std::string` instead of `const char*`? that would be consistent with the `ConstNodeImpl` methods.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -6912,6 +6914,69 @@ struct OrtApi {
...
+  /** \brief Get information about the subgraphs assigned to each EP and the nodes within.
+   *
+   * Each returned OrtEpAssignedSubgraph instance contains details of the subgraph/nodes assigned to an execution provider,
+   * including the execution provider's name, and the name and operator type for every node. For compiling EPs,
+   * a subgraph contains one or more nodes. Alternatively, for EPs that use kernel registration (e.g., CPU EP), each
+   * registered kernel for a node is contained in its own subgraph (i.e., a subgraph contains one node).
+   *
+   * \note Application must enable the recording of graph partitioning information by enabling the session configuration
+   *       for the key "session.record_ep_graph_assignment_info". Refer to onnxruntime_session_options_config_keys.h.
+   *       If the session configuration is not enabled, this function returns an empty result.
```
edgchen1: would it be useful to error out with a message about how to enable it?

adrianlizarraga: Yep, updated.

include/onnxruntime/core/session/onnxruntime_c_api.h
adrianlizarraga: An `OrtNode` has to refer to an actual `onnxruntime::Node`. However, many of the nodes reported by this API will have been removed after session finalization. Specifically:
- For compiling EPs, the nodes in a ComputeCapability, which we report here, are replaced with a single fused node after partitioning.
- For kernel-based EPs, there's a chance that a L2+ optimization could remove or modify the nodes.

Additionally, an `OrtNode` provides access to other graph elements (inputs, initializers, etc) that we may not want to expose yet via this api.

include/onnxruntime/core/session/onnxruntime_c_api.h
adrianlizarraga: Good call, added.

include/onnxruntime/core/session/onnxruntime_cxx_api.h
adrianlizarraga: yep, updated.

include/onnxruntime/core/session/onnxruntime_session_options_config_keys.h
adrianlizarraga: That's right. I'm concerned about the storage size for applications that never use this. We essentially store ~200 bytes of information for every node in the model.

Also, I'm assuming that this API is mostly used for debugging/validation, so I wouldn't want to make non-users pay the cost too.

What do you think?

onnxruntime/core/session/inference_session.cc
adrianlizarraga: updated

onnxruntime/core/session/inference_session.cc
adrianlizarraga: Yes, good call. I moved it all together.

onnxruntime/core/session/onnxruntime_c_api.cc
adrianlizarraga: oops, i'm at it again :)

onnxruntime/test/autoep/test_execution.cc
adrianlizarraga: yep, updated. I moved the check here.

include/onnxruntime/core/session/onnxruntime_cxx_inline.h
```diff
@@ -747,6 +747,45 @@ inline EpDevice::EpDevice(OrtEpFactory& ep_factory, ConstHardwareDevice& hardwar
...
+      ep_nodes.emplace_back(ep_node_ptrs[i]);
+    }
+  }
+
+  return ep_nodes;
+}
+
+template <typename T>
+inline std::string EpAssignedNodeImpl<T>::GetName() const {
+  return std::string(GetApi().EpAssignedNode_GetName(this->p_));
```
edgchen1: I don't think std::string will handle being given a nullptr, as the C API will return in a minimal build. also, nullptr doesn't quite fit with the documented return value of "a UTF-8 null-terminated string"

adrianlizarraga: The `EpAssignedNode_GetName` api will not return a nullptr since the node name is stored directly as a std::string internally.

Would you like a check just in case?

adrianlizarraga: oh I see, a minimal build.
will add a check. good catch.

May be better to just make all of these APIs return a status.

onnxruntime/core/framework/graph_partitioner.cc
```diff
@@ -518,6 +521,12 @@ static Status PartitionOnnxFormatModelImpl(Graph& graph, FuncManager& func_mgr,
 
     Node* n = nullptr;
     if (sub_graph_available_for_assignment) {
+      if (on_partition_assignment_fn) {
+        // Call custom function provided by owner of GraphPartitioner whenever a subgraph is assigned to an EP.
+        // This can be used, for example, to collect partitioning information.
+        on_partition_assignment_fn(graph, *capability, type);
```
edgchen1, line 527: do we want to support recording this information for ORT format models too?

include/onnxruntime/core/session/onnxruntime_session_options_config_keys.h
edgchen1: ok, makes sense. thanks for the context.

onnxruntime/core/framework/graph_partitioner.cc
adrianlizarraga: Perhaps it would be a good idea. Would prefer to do it separately as I'm not super familiar with ORT format models and handling ONNX models is the intent for now.

edgchen1: could be done in a future PR if needed. we can get the APIs in first.

include/onnxruntime/core/session/onnxruntime_cxx_inline.h
adrianlizarraga: updated these getter apis to return a status. If they return successfully, then the `std::string` constructor will not be called a null `const char*`.

onnxruntime/test/shared_lib/test_inference.cc
```diff
@@ -4827,3 +4827,26 @@ TEST(CApiTest, ModelWithExternalDataOutsideModelDirectoryShouldFailToLoad) {
               exception_message.find("model") != std::string::npos)
       << "Exception message should indicate external data or security issue. Got: " << exception_message;
 }
+
+#if !defined(ORT_MINIMAL_BUILD)
```
edgchen1, line 4831: nit: this ifdef is not necessary since this whole file gets excluded for a minimal build.

https://github.com/microsoft/onnxruntime/blob/2d2ba6b4c537e4c39d2a2cde0d09164faaab0dd7/cmake/onnxruntime_unittests.cmake#L597-L598

edgchen1 (approved):

onnxruntime/test/shared_lib/test_inference.cc
adrianlizarraga: ah yeah

Issue #26780, opened 2025-12-12
[Build] Suppress RuntimeWarning about GIL enabling when importing onnxruntime on Free-threaded Python (3.13t)


nimdrak:
### Describe the issue

**[Issue]**
When importing onnxruntime built for Python 3.13t (Free-threaded Python), the following RuntimeWarning is displayed, which is unnecessary as ONNX Runtime is intended to be thread-safe and capable of running without the GIL.

```bash
root@dc97bf6c52fc:/opt# python3.13t -c "import onnxruntime"
<frozen importlib._bootstrap>:488: RuntimeWarning: The global interpreter lock (GIL) has been enabled to load module 'onnxruntime.capi.onnxruntime_pybind11_state', which has not declared that it can run safely without the GIL. To override this behavior and keep the GIL disabled (at your own risk), run with PYTHON_GIL=0 or -Xgil=0.
```

**[Root Cause]**
The onnxruntime_pybind11_state.so module is compiled using pybind11. Although pybind11 (v2.13+) supports injecting the necessary Py_MOD_GIL_NOT_USED flag, the code responsible for this is conditionally compiled:
```c++
if (gil_not_used_option(options...)) {
#if defined(Py_mod_gil) && defined(Py_GIL_DISABLED)
        mod_def_slots[next_slot++] = {Py_mod_gil, Py_MOD_GIL_NOT_USED}; // This line is skipped
#endif
    }
```
https://github.com/pybind/pybind11/blob/228f56361016ab9e27d5ef21853542dab3e37693/include/pybind11/pybind11.h#L1314-L1321

**[Proposed Solution]**
Pass the definition to the C++ compiler:
```bash
CXXFLAGS="-DPy_GIL_DISABLED=1" ./build.sh ...
```
This ensures that the conditional compilation block for Py_MOD_GIL_NOT_USED is activated, resolving the warning and allowing true Free-threaded operation.

### To reproduce

```bash
root@dc97bf6c52fc:/opt# python3.13t -c "import onnxruntime"
<frozen importlib._bootstrap>:488: RuntimeWarning: The global interpreter lock (GIL) has been enabled to load module 'onnxruntime.capi.onnxruntime_pybind11_state', which has not declared that it can run safely without the GIL. To override this behavior and keep the GIL disabled (at your own risk), run with PYTHON_GIL=0 or -Xgil=0.
```

### Urgency

Not urgent

### Platform

Linux

### OS Version

Ubuntu 22.04.4 LTS

[...]

PR #26786, opened 2025-12-13
Enable Free-threaded Python (PEP 703) support for Python 3.13t+

tianleiwu:
This PR resolves the `RuntimeWarning` encountered when importing `onnxruntime` in free-threaded Python environments (e.g., Python 3.13t, 3.14t).

Previously, the module did not explicitly declare that it could run safely without the GIL, causing the interpreter to re-enable the GIL at runtime.

### The Warning

```text
RuntimeWarning: The global interpreter lock (GIL) has been enabled to load module
'onnxruntime.capi.onnxruntime_pybind11_state', which has not declared that it can
run safely without the GIL.
````

### 1. Build System (`cmake/onnxruntime_python.cmake`)

* Added robust detection logic to check if the current Python interpreter is free-threaded.

**Detection strategy:**

* **Primary:** Check `sysconfig.get_config_var('Py_GIL_DISABLED')` (PEP 703 standard).

* **Fallback:** Inspect ABI flags (`ABIFLAGS` or `SOABI`) for the `t` suffix (e.g., `cp313t`), handling cases on Windows where config variables may return `None` or empty strings.

* **Caching:** Cache the result as `ORT_PYTHON_FREE_THREADED_DETECTED` (`BOOL`) to avoid re-running detection on every configure.

---

### 2. C++ Source (`onnxruntime/python/onnxruntime_pybind_module.cc`)

* Updated the `PYBIND11_MODULE` definition to conditionally enable the GIL-free slot.
* Uses `py::mod_gil_not_used()` (available in pybind11 ≥ 2.13) when `Py_GIL_DISABLED` is defined.

---

## Verification

**Environment:**

* Python 3.14t (Free-threaded)
* Windows & Linux

### Before Change

```bash
python3.14t -c "import onnxruntime"
```

Resulted in the `RuntimeWarning` shown above.

### After Change

1. **Build log confirms correct detection:**

```text
-- Python_EXECUTABLE: D:\py314t\Scripts\python.exe
-- Py_GIL_DISABLED=1 detected: Enabling free-threaded support for onnxruntime_pybind11_state
```

2. **Runtime behavior:**

```bash
python3.14t -c "import onnxruntime"
```

Runs silently with no warnings, confirming the module loads successfully with the GIL disabled.

## Related Issue

* [https://github.com/microsoft/onnxruntime/issues/26780](https://github.com/microsoft/onnxruntime/issues/26780)

### DIFF
--- a/cmake/onnxruntime_python.cmake
+++ b/cmake/onnxruntime_python.cmake
@@ -73,12 +73,12 @@ message(STATUS "Python_EXECUTABLE: ${Python_EXECUTABLE}")
 
 # Query Py_GIL_DISABLED (PEP 703)
 execute_process(
-    COMMAND "${Python_EXECUTABLE}" -c
-            "import sysconfig; print(sysconfig.get_config_var('Py_GIL_DISABLED') or '0')"
-    RESULT_VARIABLE _py_result
-    OUTPUT_VARIABLE _py_gil_disabled
-    OUTPUT_STRIP_TRAILING_WHITESPACE
-  )
+  COMMAND "${Python_EXECUTABLE}" -c
+          "import sysconfig; print(sysconfig.get_config_var('Py_GIL_DISABLED') or '0')"
+  RESULT_VARIABLE _py_result
+  OUTPUT_VARIABLE _py_gil_disabled
+  OUTPUT_STRIP_TRAILING_WHITESPACE
+)
 if (_py_result EQUAL 0 AND _py_gil_disabled STREQUAL "1")
   message(STATUS "Py_GIL_DISABLED=1 detected: Enabling free-threaded support for onnxruntime_pybind11_state")
 endif()


cmake/onnxruntime_python.cmake
```diff
@@ -69,6 +69,66 @@ endif()
...
+        set(Py_GIL_DISABLED_VALUE "0")
+    endif()
+
+    # Cache the result as BOOL so users can override it (e.g. -DORT_PYTHON_FREE_THREADED_DETECTED=OFF)
+    set(ORT_PYTHON_FREE_THREADED_DETECTED "${Py_GIL_DISABLED_VALUE}" CACHE BOOL "Detected free-threaded Python status" FORCE)
+  else()
+    message(STATUS "Using cached free-threaded Python detection: ORT_PYTHON_FREE_THREADED_DETECTED=${ORT_PYTHON_FREE_THREADED_DETECTED}")
+  endif()
+
+  if (ORT_PYTHON_FREE_THREADED_DETECTED)
```
edgchen1: I'm not seeing what `ORT_PYTHON_FREE_THREADED_DETECTED` is used for, other than printing the status message. is that intentional?

cmake/onnxruntime_python.cmake
```diff
@@ -69,6 +69,66 @@ endif()
...
+      execute_process(
+        COMMAND "${Python_EXECUTABLE}" -c "import sysconfig; print(sysconfig.get_config_var('ABIFLAGS') or sysconfig.get_config_var('SOABI') or '')"
+        RESULT_VARIABLE _abi_result
+        OUTPUT_VARIABLE Py_ABI_VALUE
+        OUTPUT_STRIP_TRAILING_WHITESPACE
+      )
+
+      if (_abi_result EQUAL 0)
+        # Matches "cp313t", "cp314t" etc., or "t" followed by hyphen/end-of-string (e.g., "cp313t-win_amd64")
+        if ("${Py_ABI_VALUE}" MATCHES "cp[0-9]+t" OR "${Py_ABI_VALUE}" MATCHES "t(-|$)")
```
edgchen1: what is the particular pattern we are looking for? given the example strings in the comment, I'm not sure why we need the second regex `t(-|$)`

cmake/onnxruntime_python.cmake
tianleiwu: Right. The cmake change is only to print the status message.

edgchen1: is it worth the additional complexity just to print a status message?

tianleiwu: Simplified for the use case of ci pipeline

cmake/onnxruntime_python.cmake
tianleiwu: removed.

edgchen1 (approved):

PR #26804, opened 2025-12-15
Proposed changes to model compilation telemetry

adrastogi:
### Description

This change adjusts the compilation telemetry to span the start and stop of the operation, with additional recording of various facets of the compilation session and information about the result. 

### Motivation and Context

These changes enable deeper understanding of compilation usage and health, which in turn can help prioritize more investments in making compilation better overall for users of the platform (e.g., more reliable, performant, better tuned options / defaults, etc.).

### DIFF
--- a/onnxruntime/core/platform/windows/telemetry.cc
+++ b/onnxruntime/core/platform/windows/telemetry.cc
@@ -334,20 +334,6 @@ void WindowsTelemetry::LogSessionCreation(uint32_t session_id, int64_t ir_versio
   }
 }
 
-void WindowsTelemetry::LogCompileModel(uint32_t session_id) const {
-  if (global_register_count_ == 0 || enabled_ == false)
-    return;
-
-  TraceLoggingWrite(telemetry_provider_handle,
-                    "CompileModel",
-                    TraceLoggingBool(true, "UTCReplace_AppSessionGuid"),
-                    TelemetryPrivacyDataTag(PDT_ProductAndServiceUsage),
-                    TraceLoggingKeyword(MICROSOFT_KEYWORD_MEASURES),
-                    // Telemetry info
-                    TraceLoggingUInt8(0, "schemaVersion"),
-                    TraceLoggingUInt32(session_id, "sessionId"));
-}
-
 void WindowsTelemetry::LogCompileModelStart(uint32_t session_id,
                                             const std::string& input_source,
                                             const std::string& output_target,
--- a/onnxruntime/core/session/utils.cc
+++ b/onnxruntime/core/session/utils.cc
@@ -430,9 +430,6 @@ Status CompileModel(const Environment& env, const ModelCompilationOptions& model
       model_compile_options.HasExternalInitializersFileForTelemetry(),
       session->GetRegisteredProviderTypes());
 
-  // Keep existing telemetry call for backward compatibility
-  telemetry_provider.LogCompileModel(session->GetCurrentSessionId());
-
   status = ToStatusAndRelease(InitializeSession(session_options, *session));
 
   telemetry_provider.LogCompileModelComplete(
--- a/onnxruntime/core/platform/telemetry.cc
+++ b/onnxruntime/core/platform/telemetry.cc
@@ -81,10 +81,6 @@ void Telemetry::LogSessionCreation(uint32_t session_id, int64_t ir_version, cons
   ORT_UNUSED_PARAMETER(captureState);
 }
 
-void Telemetry::LogCompileModel(uint32_t session_id) const {
-  ORT_UNUSED_PARAMETER(session_id);
-}
-
 void Telemetry::LogCompileModelStart(uint32_t session_id,
                                      const std::string& input_source,
                                      const std::string& output_target,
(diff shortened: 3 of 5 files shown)


onnxruntime/core/session/utils.cc
```diff
@@ -388,25 +390,59 @@ namespace onnxruntime {
...
+      session->GetCurrentSessionId(),
+      model_compile_options.GetInputSourceForTelemetry(),
+      model_compile_options.GetOutputTargetForTelemetry(),
+      model_compile_options.GetFlagsForTelemetry(),
+      model_compile_options.GetGraphOptimizationLevelForTelemetry(),
+      model_compile_options.GetEmbedEpContextForTelemetry(),
+      model_compile_options.HasExternalInitializersFileForTelemetry(),
+      session->GetRegisteredProviderTypes());
+
+  // Keep existing telemetry call for backward compatibility
```
adrastogi: @xieofxie I was wondering if, by splitting out the start / end into separate events (which I wanted for tracking overall wall-clock time), your needs for the profiling tool would still be covered. In which case, I could potentially remove the LogCompileModel event. What do you think?

xieofxie: yes, it is totally fine for me. Please ask onnxruntime team to see if OK to remove unreleased code https://github.com/microsoft/onnxruntime/commit/7845ea8442993def9138f7b53cd5228e0e42b11e

adrastogi: Thanks! @adrianlizarraga, would there be any concerns with removing the code from the aforementioned commit? I don't think it is part of any release, so I believe it should be safe, but wanted to cross-check.

onnxruntime/core/platform/windows/telemetry.cc
```diff
@@ -348,6 +348,69 @@ void WindowsTelemetry::LogCompileModel(uint32_t session_id) const {
...
+
+  std::string execution_provider_string;
+  bool first = true;
+  for (const auto& ep_id : execution_provider_ids) {
+    if (first) {
+      first = false;
+    } else {
+      execution_provider_string += ',';
+    }
+    execution_provider_string += ep_id;
```
ashrit-ms, line 356: It might be better to only log the EP we are compiling against.

adrastogi: Good point. I guess this ends up logging the registered EPs in the session which would include the CPU EP. Do you have a suggestion on how to resolve this? I'm looking through what data we have access to- the first thought I had was to get this from the session options, but the factories don't seem to have a way of getting the EP name directly.  cc: @adrianlizarraga for any ideas here as well.

onnxruntime/core/session/utils.cc
adrastogi: (I went ahead and removed this under the assumption that it was ok- but if we need to restore it for any reason, please let me know.)

onnxruntime/core/session/utils.cc
```diff
@@ -388,25 +390,56 @@ namespace onnxruntime {
...
+                                                              model_compile_options.GetInputModelData(),
+                                                              model_compile_options.GetInputModelDataSize(),
+                                                              session));
   }
 
-  Env::Default().GetTelemetryProvider().LogCompileModel(session->GetCurrentSessionId());
-  ORT_RETURN_IF_ERROR(ToStatusAndRelease(InitializeSession(session_options, *session)));
-  return Status::OK();
+  if (!status.IsOK()) {
+    telemetry_provider.LogCompileModelComplete(
```
adrianlizarraga, line 413: qq: do we need to check if the telemetry provider is enabled before calling its methods? Or does the is_enabled check occur internally within the method?

onnxruntime/core/platform/windows/telemetry.cc
adrianlizarraga: I don't think we currently have knowledge of which of the registered EPs actually created compiled nodes (EPContext nodes). The session assumes that any registered EP could potentially generate compiled nodes in the output model.

In most cases, the list of registered EPs would only include CPU EP + <the compiling ep>, so maybe this is ok as is? Otherwise, one could manually filter out the CPU EP here.

onnxruntime/core/session/utils.cc
adrianlizarraga: Thanks for checking. Yes, it should be fine to remove the old event given that it isn't included in any official ORT release.

onnxruntime/core/session/utils.cc
adrastogi: Good question- the enablement check is done in the method implementation. Every logging call first consults an internal `enabled_` state variable before proceeding with its work.

onnxruntime/core/platform/windows/telemetry.cc
adrastogi: Thanks! Your analysis matches what I'd expect here as well. What do you all think about keeping this as is for now and we can revisit in the future as needed? (I saw the related comment [here](https://github.com/microsoft/onnxruntime/pull/26804#discussion_r2633191217), but will keep the discussion in this comment thread.)

PR #26805, opened 2025-12-16
Fix ExpandDims axis handling bug.

edgchen1:
### Description

Fix some issues with axis handling of ExpandDims.
- It was improperly getting the rank with `TensorShape::Size()` which actually gets the number of elements.
- The range of valid axis values is actually `[-(rank+1), rank]`.

### Motivation and Context

Fix crash on out-of-range axis values. Now it will error out instead.

### DIFF
--- a/onnxruntime/test/contrib_ops/expand_dims_test.cc
+++ b/onnxruntime/test/contrib_ops/expand_dims_test.cc
@@ -5,17 +5,31 @@
 
 #include "gtest/gtest.h"
 
+#include "core/common/common.h"
+#include "core/common/narrow.h"
+#include "core/framework/tensor_shape.h"
 #include "test/providers/provider_test_utils.h"
 
 namespace onnxruntime {
 namespace test {
 
 namespace {
+size_t ShapeSize(const std::vector<int64_t>& shape) {
+  return narrow<size_t>(TensorShape(shape).Size());
+}
+
+std::vector<float> GenerateInput(size_t size) {
+  std::vector<float> result(size);
+  std::iota(result.begin(), result.end(), 1.0f);
+  return result;
+}
+
 void RunExpandDimsTest(const std::vector<int64_t>& input_shape, const int32_t expand_axis,
                        const std::vector<int64_t>& expected_output_shape) {
-  const auto input_size = std::accumulate(input_shape.begin(), input_shape.end(),
-                                          int64_t{1}, std::multiplies<>{});
-  const auto input_data = std::vector<float>(input_size, 2.0f);
+  SCOPED_TRACE(MakeString("input_shape: ", TensorShape(input_shape), ", expand_axis: ", expand_axis));
+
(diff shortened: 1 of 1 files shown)


onnxruntime/contrib_ops/cpu/expand_dims.h
```diff
@@ -31,9 +31,9 @@ class ExpandDims final : public OpKernel {
     const TensorShape& X_shape = X->Shape();
 
     TensorShapeVector expanded_shape(X_shape.AsShapeVector());
```
yuslepukhin, line 33: >  TensorShapeVecto [](http://example.com/codeflow?start=2&length=18)

Is there a CUDA impl to check?

edgchen1: `git grep` only shows this one.

yuslepukhin:
Consider adding all or some of the test coverage suggested below.

**Boundary axes (positive & negative):**

- axis = 0 on non‑scalar input
- axis = rank on non‑scalar input
- axis = -(rank+1) (maps to 0)
- Invalid: axis = rank+1 and axis = -(rank+2) → expect error


**Scalar input (rank = 0):**

- X is scalar; axis = 0 → output rank 1
- axis = -1 → same as above
- Invalid: any other axis → error

**Axis type mismatch (negative test):**

- Provide axis as int64 → expect type‑constraint failure (kernel requires int32)

**Different dtypes for X:**
**At least one non‑float type: int32, bool, double (the kernel accepts “AllTensorTypes”); confirm dtype plumbing doesn’t break. [github.com]**

**Shape correctness & data integrity:**
Verify output shape equals insert(X.shape, axis, 1) and that data content is unchanged (values preserved and layout consistent). The existing tests use uniform 1.0f, which hides layout errors; include a non‑uniform input to ensure element mapping isn’t corrupted.


**Error message quality:**
When axis is out of range, assert the returned Status message contains axis value and allowed range—useful for triage.

onnxruntime/test/contrib_ops/expand_dims_test.cc
```diff
@@ -1,26 +1,65 @@
...
 
 namespace onnxruntime {
 namespace test {
 
-TEST(ContribOpTest, ExpandDims_0) {
+namespace {
+void RunExpandDimsTest(const std::vector<int64_t>& input_shape, const int32_t expand_axis,
+                       const std::vector<int64_t>& expected_output_shape) {
+  const auto input_size = std::accumulate(input_shape.begin(), input_shape.end(),
+                                          int64_t{1}, std::multiplies<>{});
```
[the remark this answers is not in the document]
edgchen1: the indentation appears to be correct.

edgchen1:
> Consider adding all or some of the test coverage suggested below.

made some of the suggested updates.

yuslepukhin (approved):
:shipit:

PR #26806, opened 2025-12-16
[webgpu] Support `Split-K` in more situations

Jiawei-Shao:
### Description
This patch supports more `dim_inner` (up to 4096) for `Split-K` to optimize more models.
This patch also enables `Split-K` on `gen-12lp`.

### Motivation and Context
With this PR we can achieve about 30% improvement on `jina-clip-v1-text-fp16` and 20% improvement on `jina-embeddings-v2-base-code-fp16` on Lunar Lake iGPUs.


Jiawei-Shao:
@qjia7 @jchen10  PTAL, thanks!

onnxruntime/core/providers/webgpu/webgpu_utils.cc
```diff
@@ -30,21 +30,30 @@ SplitKConfig::SplitKConfig(const wgpu::AdapterInfo& adapter_info) {
     if (adapter_info.architecture == std::string_view{"xe-2lpg"} ||
         adapter_info.architecture == std::string_view{"xe-2hpg"} ||
         adapter_info.architecture == std::string_view{"xe-lpg"} ||
-        adapter_info.architecture == std::string_view{"gen-12hp"}) {
+        adapter_info.architecture == std::string_view{"gen-12hp"} ||
+        adapter_info.architecture == std::string_view{"xe-3lpg"}) {
```
qjia7: I hope the splitK optimization can applied to all intel devices. Can we just relax it to all intel devices since the listed devices here along with the theoretical benefits of splitK (improving GPU parallelism) both confirm it?

onnxruntime/core/providers/webgpu/webgpu_utils.cc
```diff
@@ -30,21 +30,30 @@ SplitKConfig::SplitKConfig(const wgpu::AdapterInfo& adapter_info) {
...
-      // atomic calls for each output value.
+      // proper value of `max_dim_inner_with_rate` may be reduced for a larger `dim_inner` because
+      // larger `dim_inner` will add more atomic calls for each output value.
       split_dim_inner_ = 256;
       min_dim_inner_with_split_k_ = split_dim_inner_ * 2;
-      max_dim_inner_with_split_k_ = split_dim_inner_ * 9;
-      max_dim_a_outer_multiplies_dim_b_outer_divides_dim_inner_ = 35.0f;
+
+      configs_per_dim_inner_range_.emplace_back(2304, 35.0f);
+      configs_per_dim_inner_range_.emplace_back(3072, 24.0f);
```
qjia7: Theoretically, I hope splitK is only used when K/split_dim_inner (the dispatched workgroups based on K) is much larger than M / 32 * K/32 (the dispatch workgroups based on M/N) to improve GPU parallelism. Otherwise, we should still dispatch based M*N instead of K. But it seems that it's not the case. Can you share what's the M/N's size when K=3072 or 2304?
Have you also checked that K is very large but M/N is very small case? For example, K = 4096, M=32/N=32. In current change, it will not go to the splitK path, right?

onnxruntime/core/providers/webgpu/webgpu_utils.cc
Jiawei-Shao: I'm afraid not because with these parameters Split-K doesn't work well on Gen9 and Gen12 GPUs. I will tune another value for these GPUs.

onnxruntime/core/providers/webgpu/webgpu_utils.cc
Jiawei-Shao: > Can you share what's the M/N's size when K=3072 or 2304?

Below are the shapes in the models:
M = 16, N = 768, K = 3072 (M * N / K = 4.0)
M = 52, N = 768, K = 3072 (M * N / K = 13.0)
M = 400, N = 64, K = 2304 (M * N / K = 11.11)
M = 16, N = 512, K = 2304 (M * N / K = 3.56)

Below are the shapes I've verified with Split-K:
K = 3072, M * N / K = 24.0

M = 2304 N = 32
M = 1152 N = 64
M = 768 N = 96
M = 576 N = 128
M = 384 N = 192
M = 288 N = 256
M = 256 N = 288
M = 192 N = 384
M = 128 N = 576
M = 96  N = 768
M = 64 N = 1152
M = 32 N = 2304

K = 2304, 34.0 < M * N / K < 35.0

M = 64 N = 1248
M = 96 N = 832
M = 832 N = 96
M = 192 N = 416
M = 416 N = 192
M = 224 N = 352
M = 352 N = 224

> Have you also checked that K is very large but M/N is very small case? For example, K = 4096, M=32/N=32. In current change, it will not go to the splitK path, right?

Yes. SplitK will always bring performance gain with very large K but small M/N. For example with K = 4096, a safe M * N / K can be 18. I've verified on below shapes:

M = 32 N = 2304
M = 64 N = 1152
M = 96 N = 768
M = 128	N = 576
M = 192	N = 384
M = 256	N = 288
M = 288	N = 256
M = 384	N = 192
M = 576	N = 128
M = 768	N = 96
M = 1152 N = 64
M = 2304 N = 32

In fact there can be a different rate for different K. I just don't want to write a very complicated rule for Split-K so I decide to simplify everything right now. Now I just add the K we need right now.

onnxruntime/core/providers/webgpu/webgpu_utils.cc
qjia7: It’s a bit unfortunate that we need such detailed, device‑specific fine‑tuning thresholds for different Intel GPUs—it adds maintenance overhead and makes the logic harder to sustain over time.

I’m wondering whether we could define a more conservative, “safe” set of thresholds that may not capture every corner case but would work across all Intel GPUs. At the same time, it would be helpful for us to dig deeper into the underlying theory to better understand what’s driving the optimal thresholds.
From current observations, the speedup from split‑K doesn’t seem to come solely from increased parallelism. It also improves GPU utilization even when the number of workgroups is small, possibly due to more efficient I/O access patterns. This is something we should investigate further to guide future tuning.

No additional comments on this PR—the changes look good to me as long as you’ve validated that they are effective on your models and target devices.

Jiawei-Shao: > I’m wondering whether we could define a more conservative, “safe” set of thresholds that may not capture every corner case but would work across all Intel GPUs. 

Hi Jiajia, now I choose the thresholds on Gen12 32EU as the default ones for all Intel GPUs. LNL has much better values so I keep them there.

PTAL, thanks!

qjia7 (dismissed):

guschmue (dismissed):

guschmue:
CI is nagging on linux and macos:
image

Jiawei-Shao:
> CI is nagging on linux and macos: image

Hi @guschmue, sorry about that! I've fixed this failure, PTAL, thanks!

guschmue (approved):

PR #26808, opened 2025-12-16
Update protobuf 4.25.1 to 4.25.8

eserscor:
### Description

Update protobuf from version 4.25.1 to 4.25.8.

### Motivation and Context

There is an underlying security vulnerability with 4.25.1. Updating to 4.25.8 will improve our security posture.

### DIFF
--- a/tools/ci_build/github/linux/docker/scripts/manylinux/requirements.txt
+++ b/tools/ci_build/github/linux/docker/scripts/manylinux/requirements.txt
@@ -4,7 +4,7 @@ mypy
 pytest
 setuptools>=68.2.2
 wheel
-protobuf==4.25.1; python_version < "3.14"
+protobuf==4.25.8; python_version < "3.14"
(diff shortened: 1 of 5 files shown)


tools/ci_build/github/linux/docker/scripts/manylinux/requirements.txt
```diff
@@ -4,7 +4,7 @@ mypy
 pytest
 setuptools>=68.2.2
 wheel
-protobuf==4.25.1; python_version < "3.14"
+protobuf==4.25.8; python_version < "3.14"
```
tianleiwu, line 7: I guess it is possible to upgrade to 6.33.0 like python 3.14.

eserscor: I'll give it a run through CICD and see if we hit problems.

PR #26815, opened 2025-12-17
[MLAS] Enable FP16 for Gelu

akote123:
Enabled fp16 Gelu for opset20.Gelu uses tanh and ERF functions depending on the approximation method used. Implemented tanh in sve and erf in sve and neon .
Gr3E results: with tanh and erf approximation:
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/arunak/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/arunak/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">

</head>

<body link="#467886" vlink="#96607D">

GELU(ms) | Tanh_SVE | ERF_SVE | Tanh_NEON | ERF_NEON
-- | -- | -- | -- | --
Shape | F32 | F16 | F32 | F16 | F32 | F16 | F32 | F16
100 | 0.007 | 0.007 | 0.007 | 0.007 | 0.007 | 0.007 | 0.007 | 0.007
1000 | 0.008 | 0.007 | 0.012 | 0.008 | 0.008 | 0.008 | 0.012 | 0.008
1000000 | 0.076 | 0.039 | 0.203 | 0.07 | 0.066 | 0.043 | 0.151 | 0.064

</body>

</html>

Gr4 results: with tanh and erf approximation:
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/arunak/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/arunak/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">

</head>
<body link="#467886" vlink="#96607D">

GELU(ms) | Tanh_SVE | ERF_SVE | Tanh_NEON | ERF_NEON
-- | -- | -- | -- | --
Shape | F32 | F16 | F32 | F16 | F32 | F16 | F32 | F16
100 | 0.005 | 0.005 | 0.005 | 0.005 | 0.005 | 0.005 | 0.005 | 0.005
1000 | 0.006 | 0.006 | 0.008 | 0.006 | 0.006 | 0.006 | 0.008 | 0.006
1000000 | 0.092 | 0.046 | 0.224 | 0.088 | 0.091 | 0.051 | 0.213 | 0.086
</body>
</html>

This PR is a joint contribution by:
Aruna K(@akote123)
Abhishek Jain(@abhijain1204fujitsu)
Sanket Kale(@sanketkaleoss )


cmake/onnxruntime_mlas.cmake
```diff
@@ -460,13 +462,17 @@ else()
...
+          ${MLAS_SRC_DIR}/erf_neon_fp16.cpp
         )
 
         # Conditionally add the SVE implementation if compiler supports it
         if (onnxruntime_USE_SVE)
           list(APPEND mlas_platform_srcs ${MLAS_SRC_DIR}/sve/mlasi_sve.h)
           list(APPEND mlas_platform_srcs ${MLAS_SRC_DIR}/sve/elementwise_sve.cpp)
+          list(APPEND mlas_platform_srcs ${MLAS_SRC_DIR}/sve/Elementwise_sve_fp16.cpp)
           set_source_files_properties(${MLAS_SRC_DIR}/sve/elementwise_sve.cpp PROPERTIES COMPILE_FLAGS " -march=armv8.2-a+sve+fp16 ")
+          set_source_files_properties(${MLAS_SRC_DIR}/sve/Elementwise_sve_fp16.cpp PROPERTIES COMPILE_FLAGS " -march=armv8.2-a+sve+fp16 ")
```
hariharans29: Nit: Rename to elementwise_sve_fp16.cpp ? (casing is inconsistent)

onnxruntime/core/providers/cpu/math/element_wise_ops.cc
```diff
@@ -11,6 +11,10 @@
 #include "core/util/math.h"
 #include "core/mlas/inc/mlas.h"
 
+#if defined(MLAS_NEON_INTRINSICS)
```
hariharans29: Isn't this macro defined in mlasi.h (which is not included here). 
There seems to be something wrong with the design here - I think there is usage of MLAS internal implementation macros leaking over to the CPU EP files. It may need a re-think.

onnxruntime/core/providers/cpu/tensor/gelu.cc
```diff
@@ -105,4 +119,160 @@ Status Gelu<T>::Compute(OpKernelContext* context) const {
...
+  const float16_t c2 = -5.0f;
+  const MLAS_FLOAT16X8 v_half = MlasBroadcastF16Float16x8(v_half1);
+  const MLAS_FLOAT16X8 v_one = MlasBroadcastF16Float16x8(v_one1);
+  const MLAS_FLOAT16X8 v_sqrt1_2 = MlasBroadcastF16Float16x8(v_sqrt1_21);
+  const MLAS_FLOAT16X8 v_B = MlasBroadcastF16Float16x8(v_B1);
+  const MLAS_FLOAT16X8 v_C = MlasBroadcastF16Float16x8(v_C1);
+
+  int64_t i = 0;
+
+  if (algo == "tanh") {
```
hariharans29: I think all this logic needs to live in MLAS and not in a CPU EP file. is there a limitation that necessitated doing things this way ?

onnxruntime/core/mlas/lib/sve/mlasi_sve.h
```diff
@@ -32,7 +32,18 @@ typedef svfloat32_t MLAS_SVFLOAT32;
...
 typedef svuint32_t MLAS_SVUINT32;
 typedef svbool_t MLAS_SVBOOL;
+typedef svfloat16_t MLAS_SVFLOAT16;
+typedef svuint16_t MLAS_SVUINT16;
 
+using _mlas_fp16_ = uint16_t;
+
+void
+MLASCALL
+MlasSveErfKernelFp16(
```
hariharans29: It would be nice to stick to one naming convention- some APIs have `MlasSve...` and some others have `..._SVE` which is confusing. Some have `Mlas....` and another one has `Compute....`.

cmake/onnxruntime_providers_cpu.cmake
```diff
@@ -172,7 +172,14 @@ if (onnxruntime_ENABLE_CPU_FP16_OPS)
   set_source_files_properties(${ORTTRAINING_SOURCE_DIR}/training_ops/cuda/collective/adasum_kernels.cc PROPERTIES COMPILE_FLAGS " -fassociative-math -ffast-math -ftree-vectorize -funsafe-math-optimizations -mf16c -mavx -mfma ")
 endif()
 
-target_include_directories(onnxruntime_providers PRIVATE ${ONNXRUNTIME_ROOT})
+if(onnxruntime_target_platform STREQUAL "aarch64" OR onnxruntime_target_platform STREQUAL "ARM64" OR onnxruntime_target_platform STREQUAL "arm64")
+set_source_files_properties("${ONNXRUNTIME_ROOT}/core/providers/cpu/tensor/gelu.cc" PROPERTIES COMPILE_FLAGS -march=armv8.2-a+fp16)
```
hariharans29: This may be a duplicated comment - I think this is coming about because some CPU EP files are now directly using intrinsics and I feel the hardware accelerated intrinsic using routines should live in MLAS and only be called from the CPU EP files

onnxruntime/core/providers/cpu/tensor/gelu.cc
akote123: Will update these logic in MLAS

onnxruntime/core/providers/cpu/math/element_wise_ops.cc
akote123: Will update

onnxruntime/core/mlas/lib/sve/mlasi_sve.h
```diff
@@ -32,7 +32,36 @@ typedef svfloat32_t MLAS_SVFLOAT32;
...
+);
+
+void 
+MLASCALL 
+MlasSveGeluF16Kernel(
+    const MLAS_FP16* input,
+    MLAS_FP16* output,
+    MLAS_FP16* temp,
+    int64_t count,
+    const std::string& algo
```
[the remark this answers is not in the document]
sanketkaleoss: While avoiding string comparisons in the hot path makes sense, the approximation algorithm is currently part of the shared `~/onnxruntime/onnxruntime/core/providers/cpu/tensor/gelu.h` interface as a `std::string`. Changing it to an enum would impact other GELU implementations right?

abhijain1204fujitsu:
@hariharans29 we have pushed the code to resolve all the above comments 
Kindly support for further review and merger of the PR

hariharans29:
> @hariharans29 we have pushed the code to resolve all the above comments Kindly support for further review and merger of the PR

Please manually "resolve" Copilot's comments and add comments if you think Copilot's suggestion is not applicable and you re not taking it in. Thanks.

abhijain1204fujitsu:
Hi @hariharans29 , 
Thank you for your previous review comments
We have pushed the changes as per review, kindly support us to check the CI Pipelines and further review the code for merger.

onnxruntime/core/mlas/lib/sve/mlasi_sve.h
```diff
@@ -32,8 +32,37 @@ typedef svfloat32_t MLAS_SVFLOAT32;
...
+typedef svuint16_t MLAS_SVUINT16;
 
-// function decarations
+using _mlas_fp16_ = uint16_t;
+
+void
+MLASCALL
+MlasSveErfF16Kernel(
+    const _mlas_fp16_* Input,
+    _mlas_fp16_* Output,
```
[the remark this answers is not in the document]
sanketkaleoss: > Inconsistent type usage for FP16 pointers. MlasSveErfF16Kernel uses `_mlas_fp16_*` (which is `uint16_t*`), while MlasSveTanhF16Kernel and MlasSveGeluF16Kernel use `MLAS_FP16*` (which is `onnxruntime::MLFloat16*`). While these types are binary compatible, it would be better to use a consistent type throughout for clarity and maintainability. Consider standardizing on `MLAS_FP16*` to match the public API in mlas.h.

Acknowledged. Skipping this intentionally for now.

hariharans29:
> Hi @hariharans29 , Thank you for your previous review comments We have pushed the changes as per review, kindly support us to check the CI Pipelines and further review the code for merger.

There are still some unaddressed Copilot comments. Please manually resolve them with a comment as to whether going with Copilot's recommendation or not.

sanketkaleoss:
Hi @hariharans29 ,
We have pushed the new changes resolving CI failures, kindly support us by restarting the CI pipeline.

hariharans29:
> Hi @hariharans29 , We have pushed the new changes resolving CI failures, kindly support us by restarting the CI pipeline.

As stated above, please resolve all Copilot comments and my old comments. Resolving comments is a gating check for merging a PR.

I ll start another round of CI.

sanketkaleoss:
> > Hi @hariharans29 , We have pushed the new changes resolving CI failures, kindly support us by restarting the CI pipeline.
> 
> As stated above, please resolve all Copilot comments and my old comments. Resolving comments is a gating check for merging a PR.
> 
> I ll start another round of CI.

Thanks, will resolve them.

sanketkaleoss:
Hi @hariharans29, we pushed the latest commit with the CI fixes. Could you please trigger the CI pipeline?
Thanks.

hariharans29:
> Hi @hariharans29, we pushed the latest commit with the CI fixes. Could you please trigger the CI pipeline? Thanks.

Thanks. Thanks for taking a look at most of Copilot's comments and manually resolving them too. There are a few more remaining. Please take a look. I llk take another look at this PR soon.

onnxruntime/core/mlas/lib/sve/mlas_sve_fp16.h
```diff
@@ -0,0 +1,182 @@
+/*++
+
+Copyright 2025 FUJITSU LIMITED
```
hariharans29, line 3: Can you also add the Microsoft copyright along with this in the new files please ?

onnxruntime/core/mlas/inc/mlas.h
```diff
@@ -2127,3 +2127,32 @@ MlasFlashAttention(
...
+
+#if defined(USE_KLEIDIAI) && !defined(_MSC_VER)
+/**
+ * @brief Function to override the packing mechanism decision if kleidi ai is included
+ * @param enable     enable kleidiai packing (allow or disallow depending on true/false)
+ * @return
+*/
+void
+MLASCALL
+MlasGemmBatchPackUseKleidi(bool enable);
```
hariharans29: Don't think this existys anymore. Merge issue ?

onnxruntime/core/mlas/lib/sve/mlas_sve_fp16.h
sanketkaleoss: Sure, will add it

onnxruntime/core/mlas/inc/mlas.h
sanketkaleoss: Yes, seems like that, will remove it

sanketkaleoss:
@hariharans29 incorporated the suggested changes

onnxruntime/core/mlas/inc/mlas.h
```diff
@@ -2127,3 +2127,21 @@ MlasFlashAttention(
...
+    size_t N
+);
+
+void
+MLASCALL 
+MlasComputeFP16Gelu(
+    const MLAS_FP16* input,
+    MLAS_FP16* output,
+    MLAS_FP16* temp,
+    int64_t count,
```
hariharans29: Can this be size_t for consistency ?

onnxruntime/core/mlas/inc/mlas.h
```diff
@@ -2127,3 +2127,21 @@ MlasFlashAttention(
...
+    const std::string& algo
```
hariharans29: I think originally this was meantr to be a C header - not sure if std::string usage is encouraged

onnxruntime/core/providers/cpu/tensor/gelu.cc
```diff
@@ -105,4 +136,54 @@ Status Gelu<T>::Compute(OpKernelContext* context) const {
...
+  int64_t elem_count = input->Shape().Size();
+  constexpr int64_t length_per_task = 4096;
+  int64_t task_count = (elem_count + length_per_task - 1) / length_per_task;
+
+  if (approximation_algorithm_ != "tanh" && approximation_algorithm_ != "none") {
+    return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT, "Unsupported approximation_algorithm: ", approximation_algorithm_);
+  }
+
+  // Alignment and buffer size for aligned_alloc
+  constexpr size_t alignment = 64;
```
hariharans29: Can you please add some comments as to why this is needed ?

cmake/onnxruntime_mlas.cmake
```diff
@@ -54,6 +54,7 @@ onnxruntime_add_static_library(onnxruntime_mlas
   ${MLAS_SRC_DIR}/rotary_embedding.cpp
   ${MLAS_SRC_DIR}/softmax.h
   ${MLAS_SRC_DIR}/saturation_check.cpp
+  ${MLAS_SRC_DIR}/gelu.cpp
```
hariharans29: Seems like some new files are missing here in this file ? gelu.h , nlas_sve_fp16.h ?

onnxruntime/core/mlas/lib/platform.cpp
```diff
@@ -635,6 +640,18 @@ Return Value:
         this->ComputeLogSoftmaxOutputF32Kernel = MlasComputeLogSoftmaxOutputF32Kernel;
         this->ComputeSoftmaxOutputF32Kernel = MlasComputeSoftmaxOutputF32Kernel;
     }
+
+    #if defined(__ARM_FEATURE_FP16_VECTOR_ARITHMETIC) && \
+        defined(MLAS_F16VEC_INTRINSICS_SUPPORTED)
+            if (MLAS_CPUIDINFO::GetCPUIDInfo().HasArmSve()) {
+                this->ErfF16KernelRoutine = MlasSveErfF16Kernel;
```
hariharans29: Should this be guarded by USE_SVE ?

onnxruntime/core/mlas/lib/gelu.cpp
```diff
@@ -0,0 +1,54 @@
...
+                    int64_t count,
+                    const std::string& algo)
+{
+#if defined(MLAS_USE_SVE) || defined(MLAS_NEON_INTRINSICS)
+    #if defined(MLAS_F16VEC_INTRINSICS_SUPPORTED) && defined(__ARM_FEATURE_FP16_VECTOR_ARITHMETIC)
+        GetMlasPlatform().GeluF16KernelRoutine(input, output, temp, count, algo);
+        return;
+    #endif
+#endif 
+    (void)temp;
```
hariharans29: temp ?

onnxruntime/core/mlas/inc/mlas.h
sanketkaleoss: Updated

onnxruntime/core/mlas/inc/mlas.h
sanketkaleoss: Noted, updated the parameter

onnxruntime/core/providers/cpu/tensor/gelu.cc
sanketkaleoss: Added comments

cmake/onnxruntime_mlas.cmake
sanketkaleoss: Added missing files

onnxruntime/core/mlas/lib/platform.cpp
sanketkaleoss: Updated, added outer flag for FP16 vector intrinsic availability and inner for USE_SVE

onnxruntime/core/mlas/lib/gelu.cpp
sanketkaleoss: temp is used in NEON/SVE paths and unused in scalar path.

onnxruntime/core/mlas/lib/platform.cpp
```diff
@@ -635,6 +640,18 @@ Return Value:
...
+                this->GeluF16KernelRoutine = MlasSveGeluF16Kernel;
+            }
+            else{
+                this->ErfF16KernelRoutine = MlasNeonErfF16Kernel;
+                this->GeluF16KernelRoutine = MlasNeonGeluF16Kernel; 
+            }
```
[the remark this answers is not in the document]
sanketkaleoss: Added

onnxruntime/core/mlas/lib/mlasi.h
```diff
@@ -610,6 +610,25 @@ void
...
 
+using _mlas_fp16_ = uint16_t;
+typedef
+void
+(MLASCALL MLAS_COMPUTE_ERF_FP16_KERNEL)(
+    const _mlas_fp16_* Input,
+    _mlas_fp16_* Output,
+    size_t N
+);
+
```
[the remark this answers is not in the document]
sanketkaleoss: Incorporated

onnxruntime/core/mlas/lib/erf_neon_fp16.h
```diff
@@ -0,0 +1,27 @@
...
+
+#pragma once
+
+#include <arm_neon.h>
+
+#include "mlasi.h"
+#include "fp16_common.h"
+#include "softmax_kernel_neon.h"
+
+using _mlas_fp16_ = uint16_t;
```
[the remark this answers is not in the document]
sanketkaleoss: Incorporated

onnxruntime/core/mlas/lib/sve/mlasi_sve.h
```diff
@@ -32,8 +32,37 @@ typedef svfloat32_t MLAS_SVFLOAT32;
...
+    size_t N
+);
```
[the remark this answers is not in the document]
sanketkaleoss: Incorporated

onnxruntime/core/mlas/inc/mlas.h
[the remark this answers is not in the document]
sanketkaleoss: Using enum instead

onnxruntime/core/mlas/lib/erf.cpp
```diff
@@ -266,3 +276,26 @@ Return Value:
...
+        return;    
+    #endif
+#endif
+    std::vector<float> input_fp32(N);
+    std::vector<float> output_fp32(N);
+
+    MlasConvertHalfToFloatBuffer(Input, input_fp32.data(), N);
+    MlasComputeErf(input_fp32.data(), output_fp32.data(), N);
+    MlasConvertFloatToHalfBuffer(output_fp32.data(), Output, N);
+}
```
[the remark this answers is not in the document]
sanketkaleoss: Noted, not taking the suggestion as of now

onnxruntime/core/providers/cpu/tensor/gelu.cc
```diff
@@ -105,4 +136,54 @@ Status Gelu<T>::Compute(OpKernelContext* context) const {
...
+
+  size_t buffer_size = static_cast<size_t>(elem_count) * sizeof(MLFloat16);
+  size_t aligned_size =
+      ((buffer_size + alignment - 1) / alignment) * alignment;
+
+  void* raw = AlignedAlloc(alignment, aligned_size);
+  if (!raw) {
+    return ORT_MAKE_STATUS(ONNXRUNTIME, FAIL,
+                           "Failed to allocate aligned temporary buffer.");
+  }
```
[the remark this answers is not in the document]
sanketkaleoss: Incorporated

sanketkaleoss:
Hi @hariharans29, resolved these recent comments in the latest commit

sanketkaleoss:
Hi @hariharans29 updated the code. Regarding the Windows GPU Doc Gen CI, I added `MLAS_F16_VECTOR_INTRINSICS_SUPPORTED` guard in `onnxruntime/core/providers/cpu/cpu_execution_provider.cc` as for Gelu we changed kernel class to `ONNX_OPERATOR_TYPED_KERNEL_CLASS_NAME` from `ONNX_OPERATOR_KERNEL_CLASS_NAME`. Will this suffice for CI failure or is there something else need to be changed?

hariharans29:
> Hi @hariharans29 updated the code. Regarding the Windows GPU Doc Gen CI, I added `MLAS_F16_VECTOR_INTRINSICS_SUPPORTED` guard in `onnxruntime/core/providers/cpu/cpu_execution_provider.cc` as for Gelu we changed kernel class to `ONNX_OPERATOR_TYPED_KERNEL_CLASS_NAME` from `ONNX_OPERATOR_KERNEL_CLASS_NAME`. Will this suffice for CI failure or is there something else need to be changed?

Let me take a look - I think it may need some changes to the the kernels doc. If it is cumbersome to setup the cuda build for you, when we are ready to merge the PR, I will provide you with the kernels doc from CI and you can push directly to your branch.

To generate the kernel docs, it needs this option - https://github.com/microsoft/onnxruntime/blob/0f938536517b549a24a9fe839dadf0d9816d4a7e/tools/ci_build/build_args.py#L262 with cuda and dml enabled in the build.

hariharans29:
Please also run the mlas tests (onnxruntime_mlas_test.exe) and onnxruntime tests (onnxruntime_test-all.exe) in your SVE environment

Also, the PR has had a few iterations - is the gain listed in the PR description still valid ?

cmake/onnxruntime_mlas.cmake
```diff
@@ -522,6 +533,8 @@ else()
             ${MLAS_SRC_DIR}/halfgemm_kernel_neon_fp16.cpp
             ${MLAS_SRC_DIR}/softmax_kernel_neon_fp16.cpp
             ${MLAS_SRC_DIR}/eltwise_kernel_neon_fp16.cpp
+            ${MLAS_SRC_DIR}/erf_neon_fp16.cpp
```
hariharans29, line 555: Is this inclusion necessary when we already have includes in the block 453 ? Is it not duplicated inclusion ?

onnxruntime/core/mlas/inc/mlas.h
```diff
@@ -2127,3 +2127,26 @@ MlasFlashAttention(
     MlasFlashAttentionThreadedArgs* args,
     MLAS_THREADPOOL* ThreadPool
 );
+
+typedef enum MLAS_GELU_ALGORITHM {
+    MlasGeluNone = 0,
+    MlasGeluTanh = 1
+} MLAS_GELU_ALGORITHM;
+
+void
```
hariharans29, line 2268: Please add a brief for each new API and document the arguments. Please refer to other APIs above for a sample.

onnxruntime/core/mlas/lib/sve/mlas_sve_fp16.h
```diff
@@ -0,0 +1,186 @@
...
+#pragma once
+#include <arm_fp16.h>
+#include <math.h>  // for isnan if needed
+#include <stddef.h>
+
+#include "mlasi_sve.h"
+
+#if defined(__ARM_FEATURE_FP16_VECTOR_ARITHMETIC) && \
+    defined(MLAS_F16VEC_INTRINSICS_SUPPORTED)
+
```
hariharans29, line 22: Should the above inclusions be moved down to below the condition on line 25 ?

onnxruntime/core/mlas/lib/sve/mlasi_sve.h
```diff
@@ -32,8 +32,35 @@ typedef svfloat32_t MLAS_SVFLOAT32;
 typedef svint32_t MLAS_SVINT32;
 typedef svuint32_t MLAS_SVUINT32;
 typedef svbool_t MLAS_SVBOOL;
+typedef svfloat16_t MLAS_SVFLOAT16;
+typedef svuint16_t MLAS_SVUINT16;
 
-// function decarations
+void
+MLASCALL
+MlasSveErfF16Kernel(
```
hariharans29: Nit: Please use FP16 instead of F16 to keep it consistent with the top level MLAS API ?

onnxruntime/core/mlas/lib/sve/mlasi_sve.h
```diff
@@ -32,8 +32,35 @@ typedef svfloat32_t MLAS_SVFLOAT32;
...
+    const _mlas_fp16_* Input,
```
hariharans29: Should MLAS_FP16 be used here instead of _mlas_fp16_ ? Or should the below ones be made _mlas_fp16_ ? There seems to be an inconsistency.

onnxruntime/core/mlas/lib/fp16_common.h
```diff
@@ -579,4 +579,56 @@ MlasShiftLeftInt16(MLAS_INT16X4 Vector)
     return vshl_n_s16(Vector, ShiftCount);
 }
 
-#endif  // fp16 vector intrinsic supported
+// NEON FP16 vector intrinsics
+#if defined(__ARM_FEATURE_FP16_VECTOR_ARITHMETIC)
```
hariharans29, line 583: I am curious about the usage of __ARM_FEATURE_FP16_VECTOR_ARITHMETIC in this PR. IIUC- it is a flag set by GCC/Clang. Is it really needed ? Or can we just rely on the MLAS macro only - MLAS_F16VEC_INTRINSICS_SUPPORTED to guard the fp16 source code ?

onnxruntime/core/mlas/lib/gelu.h
```diff
@@ -0,0 +1,39 @@
...
+#pragma once
+
+#include "fp16_common.h"
+#include "../inc/mlas.h"
+#if defined(MLAS_NEON_INTRINSICS) && defined(__ARM_FEATURE_FP16_VECTOR_ARITHMETIC) && defined(MLAS_F16VEC_INTRINSICS_SUPPORTED)
+#include "erf_neon_fp16.h"
+
+void
+MLASCALL
+MlasNeonGeluF16Kernel(
```
hariharans29: FP16 ?

onnxruntime/core/mlas/inc/mlas.h
```diff
@@ -2127,3 +2127,26 @@ MlasFlashAttention(
     MlasFlashAttentionThreadedArgs* args,
     MLAS_THREADPOOL* ThreadPool
 );
+
+typedef enum MLAS_GELU_ALGORITHM {
+    MlasGeluNone = 0,
```
hariharans29: Is using MlasGeluNone equivalent to using the Erf path ? If so, can it be renamed to make it explicit ?

onnxruntime/core/mlas/lib/mlasi.h
```diff
@@ -610,6 +610,25 @@ void
     size_t N
     );
 
+using _mlas_fp16_ = uint16_t;
+typedef
+void
+(MLASCALL MLAS_COMPUTE_ERF_FP16_KERNEL)(
+    const _mlas_fp16_* Input,
```
hariharans29: Inconsistency: _mlas_fp16_ vs MLAS_FP16

onnxruntime/core/mlas/lib/platform.cpp
```diff
@@ -654,9 +675,10 @@ Return Value:
     this->ArmNeonIsQuantActivationsUnsigned = HasI8MMInstructions ? false : true;
     this->QNBitGemmDispatch = &GetMlasQNBitGemmDispatchNeon(HasDotProductInstructions, HasI8MMInstructions);
 
-#if defined(MLAS_F16VEC_INTRINSICS_SUPPORTED)
-    this->CastF16ToF32Kernel = &MlasCastF16ToF32KernelNeon;
-    this->CastF32ToF16Kernel = &MlasCastF32ToF16KernelNeon;
+#if defined(__ARM_FEATURE_FP16_VECTOR_ARITHMETIC) && \
```
hariharans29: Why is this change needed ?

onnxruntime/test/providers/cpu/activation/activation_op_test.cc
```diff
@@ -752,6 +752,58 @@ TEST_F(ActivationOpTest, ONNX_Gelu) {
       {},
       {{"approximate", "tanh"}}, true, 20);
 }
+
+#if defined(MLAS_F16VEC_INTRINSICS_SUPPORTED)
```
hariharans29, line 756: Can these tests be enabled even if MLAS_F16VEC_INTRINSICS_SUPPORTED is not defined (i.e.) should there be a fallback path where the input buffer is casted to fp32, processed and casted back in that case ?

onnxruntime/core/mlas/lib/sve/mlas_sve_fp16.h
```diff
@@ -0,0 +1,186 @@
...
+  mlas_sve_fp16.h
+
+Abstract:
+
+    This module contains the procedure prototypes for the SVE FP16 intrinsics.
+
+--*/
+
+#pragma once
+#include <arm_fp16.h>
```
hariharans29: Are all these includes necessary ?

onnxruntime/core/mlas/lib/erf.cpp
```diff
@@ -266,3 +276,26 @@ Return Value:
...
+MLASCALL
+MlasComputeFP16Erf(
+    const MLAS_FP16* Input,
+    MLAS_FP16* Output,
+    size_t N
+    )
+{
+#if defined(MLAS_USE_SVE) || defined(MLAS_NEON_INTRINSICS)
+    #if defined(MLAS_F16VEC_INTRINSICS_SUPPORTED) && defined(__ARM_FEATURE_FP16_VECTOR_ARITHMETIC)
+        GetMlasPlatform().ErfF16KernelRoutine(reinterpret_cast<const _mlas_fp16_*>(Input), reinterpret_cast<_mlas_fp16_*>(Output), N);
```
hariharans29: Instead of all the #if #endif, can we just check if ErfF16KernelRoutine() is not null and invoke it if present ?

onnxruntime/core/mlas/lib/erf_neon_fp16.h
```diff
@@ -0,0 +1,26 @@
...
+
+#pragma once
+
+#include <arm_neon.h>
+
+#include "mlasi.h"
+#include "fp16_common.h"
+#include "softmax_kernel_neon.h"
+
+void MlasNeonErfF16Kernel(const _mlas_fp16_* Input, _mlas_fp16_* Output, size_t N);
```
hariharans29: FP16 ?

onnxruntime/core/mlas/lib/erf_neon_fp16.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+Abstract:
+
+    This module contains the procedure prototypes for the ERF NEON FP16 intrinsics.
+
+--*/
+
+#include "erf_neon_fp16.h"
+
+#if defined(__ARM_FEATURE_FP16_VECTOR_ARITHMETIC) && \
```
hariharans29: Can you please go over the PR and remove any __ARM_FEATURE_FP16_VECTOR_ARITHMETIC that isn't absolutely needed ? I am not sure if this works for MSVC (need to check). In any case, let us add the guard only when absolutely needed.

onnxruntime/core/mlas/lib/gelu.cpp
```diff
@@ -0,0 +1,54 @@
...
+MLASCALL
+MlasComputeFP16Gelu(const MLAS_FP16* input,
+                    MLAS_FP16* output,
+                    MLAS_FP16* temp,
+                    size_t count,
+                    MLAS_GELU_ALGORITHM algo)
+{
+#if defined(MLAS_USE_SVE) || defined(MLAS_NEON_INTRINSICS)
+    #if defined(MLAS_F16VEC_INTRINSICS_SUPPORTED) && defined(__ARM_FEATURE_FP16_VECTOR_ARITHMETIC)
+        GetMlasPlatform().GeluF16KernelRoutine(input, output, temp, count, algo);
```
hariharans29: Similar to Erf - can we just check if the GeluF16KernelRoutine function pointer is not null and invoke it (can remove the noisy #ifdef guards)

onnxruntime/core/mlas/lib/gelu.cpp
```diff
@@ -0,0 +1,54 @@
...
+        return;
+    #endif
+#endif 
+    MLAS_UNREFERENCED_PARAMETER(temp); // 'temp' is only used by vectorized kernel implementations and it is unused in the scalar fallback path.
+    for (size_t i = 0; i < count; ++i) {
+        float x = static_cast<float>(input[i]);
```
hariharans29, line 80: How does the scalar pathprocess float inputs instead of fp16 when the input to this routine is fp16 ?

onnxruntime/core/mlas/lib/gelu_neon_fp16.h
```diff
@@ -0,0 +1,39 @@
...
+
+#pragma once
+
+#include "fp16_common.h"
+#include "../inc/mlas.h"
+#if defined(MLAS_NEON_INTRINSICS) && defined(__ARM_FEATURE_FP16_VECTOR_ARITHMETIC) && defined(MLAS_F16VEC_INTRINSICS_SUPPORTED)
+#include "erf_neon_fp16.h"
+
+void
+MLASCALL
```
hariharans29, line 24: Should this header be named gelu_neon_fp16.h (consistent with erf_neon_fp16.h) ?

onnxruntime/core/mlas/lib/tanh.cpp
```diff
@@ -193,6 +195,13 @@ MlasComputeTanh<MLAS_FP16>(
     MLAS_FP16* Output,
     size_t N
 ) {
+#if defined(MLAS_USE_SVE)
+    if (MLAS_CPUIDINFO::GetCPUIDInfo().HasArmSve()) {
+        MlasSveTanhF16Kernel(Input, Output, N);
```
hariharans29: Can you add afunction pointer in the platfrom struct and invoke that if it is assigned. That will remove the sve header inclusion and these checks here.

onnxruntime/core/mlas/lib/gelu.cpp
sanketkaleoss: The scalar path processes the data as fp16 → fp32 → fp16.

Specifically, the fp16 inputs are first converted to fp32 for computation in the scalar path (to maintain numerical stability and match the vectorized path behavior), and the final results are converted back to fp16 before storing.

onnxruntime/test/providers/cpu/activation/activation_op_test.cc
sanketkaleoss: For FP32 there is a different path where MLAS_F16VEC_INTRINSICS_SUPPORTED is not present.

In the FP16 case, the test already covers all available implementations — SVE, NEON, and the scalar fallback. So even when MLAS_F16VEC_INTRINSICS_SUPPORTED is not defined, execution falls back to the scalar path (fp16 → fp32 → fp16), and that path is still included in the test coverage.

onnxruntime/core/mlas/lib/sve/mlas_sve_fp16.h
sanketkaleoss: Removing those unnecessary headers

onnxruntime/core/mlas/lib/fp16_common.h
sanketkaleoss: `__ARM_FEATURE_FP16_VECTOR_ARITHMETIC `was added to fix MSVC CI issues. It’s required in all compilable files that use FP16 vector intrinsics.

`MLAS_F16VEC_INTRINSICS_SUPPORTED `alone isn’t enough, since the compiler only exposes FP16 intrinsics when building with armv8.2-a+fp16, which defines `__ARM_FEATURE_FP16_VECTOR_ARITHMETIC`

sanketkaleoss:
> Please also run the mlas tests (onnxruntime_mlas_test.exe) and onnxruntime tests (onnxruntime_test-all.exe) in your SVE environment
> 
> Also, the PR has had a few iterations - is the gain listed in the PR description still valid ?

Hi @hariharans29, both of the mentioned tests are passing in my environment (Graviton3E). Regarding performance, it still achieves the results stated in the description. Please proceed with the code review and CI run. Thanks.

cmake/onnxruntime_mlas.cmake
sanketkaleoss: Not required here, removed from this block

onnxruntime/core/mlas/inc/mlas.h
sanketkaleoss: Incorporated in the latest commit

onnxruntime/core/mlas/lib/sve/mlasi_sve.h
sanketkaleoss: Incorporated

onnxruntime/core/mlas/lib/sve/mlasi_sve.h
sanketkaleoss: Updated for using `MLAS_FP16` in function signature and later reinterpreting to `_mlas_fp16_` inside kernel

onnxruntime/core/mlas/inc/mlas.h
sanketkaleoss: Updated to replace it with `MlasGeluErf`

onnxruntime/core/mlas/lib/mlasi.h
sanketkaleoss: Updated to `MLAS_FP16`

onnxruntime/core/mlas/lib/platform.cpp
sanketkaleoss: It's needed for adding guard for `MSVC`, otherwise it throws build errors in Windows QNN

onnxruntime/core/mlas/lib/sve/mlas_sve_fp16.h
sanketkaleoss: Removed unnecessary headers

onnxruntime/core/mlas/lib/erf.cpp
sanketkaleoss: Thanks for the suggestion, incorporated in the latest commit

onnxruntime/core/mlas/lib/erf_neon_fp16.cpp
sanketkaleoss: Removed unnecessary guards, but is required in some places

onnxruntime/core/mlas/lib/gelu.cpp
sanketkaleoss: Incorporated

onnxruntime/core/mlas/lib/gelu_neon_fp16.h
sanketkaleoss: Sure, updated

onnxruntime/core/mlas/lib/tanh.cpp
sanketkaleoss: Incorporated

cmake/onnxruntime_mlas.cmake
```diff
@@ -538,6 +549,11 @@ else()
           set_source_files_properties(${MLAS_SRC_DIR}/halfgemm_kernel_neon_fp16.cpp PROPERTIES COMPILE_FLAGS " -march=armv8.2-a+fp16 ")
           set_source_files_properties(${MLAS_SRC_DIR}/softmax_kernel_neon_fp16.cpp PROPERTIES COMPILE_FLAGS " -march=armv8.2-a+fp16 ")
           set_source_files_properties(${MLAS_SRC_DIR}/eltwise_kernel_neon_fp16.cpp PROPERTIES COMPILE_FLAGS " -march=armv8.2-a+fp16 ")
+          set_source_files_properties(${MLAS_SRC_DIR}/erf_neon_fp16.cpp PROPERTIES COMPILE_FLAGS " -march=armv8.2-a+fp16 ")
+          set_source_files_properties(${MLAS_SRC_DIR}/gelu_neon_fp16.cpp PROPERTIES COMPILE_FLAGS " -march=armv8.2-a+fp16 ")
+          set_source_files_properties(${MLAS_SRC_DIR}/gelu.cpp PROPERTIES COMPILE_FLAGS " -march=armv8.2-a+fp16 ")
```
hariharans29: Should the last 3 files be here ?

cmake/onnxruntime_mlas.cmake
```diff
@@ -54,6 +54,8 @@ onnxruntime_add_static_library(onnxruntime_mlas
   ${MLAS_SRC_DIR}/rotary_embedding.cpp
   ${MLAS_SRC_DIR}/softmax.h
   ${MLAS_SRC_DIR}/saturation_check.cpp
+  ${MLAS_SRC_DIR}/gelu.cpp
+  ${MLAS_SRC_DIR}/gelu_neon_fp16.h
```
hariharans29: Please move the header below for consistency (like erf) ?

onnxruntime/core/mlas/lib/erf.cpp
```diff
@@ -22,6 +22,15 @@ Module Name:
 --*/
 
 #include "mlasi.h"
+
+#ifdef MLAS_USE_SVE
+#include "sve/mlasi_sve.h"
+#endif
```
hariharans29: Are these includes necessary here ? If the platform function pointer set, we invoke them on line 288 - so I assume these redundants are not needed ?

onnxruntime/core/mlas/lib/gelu_neon_fp16.h
```diff
@@ -0,0 +1,36 @@
...
+MlasNeonGeluFP16Kernel(
+    const MLAS_FP16* input,
+    MLAS_FP16* output,
+    MLAS_FP16* temp,
+    size_t count,
+    MLAS_GELU_ALGORITHM algo
+);
+
+#ifdef MLAS_USE_SVE
+#include "sve/mlasi_sve.h"
```
hariharans29: Why do we need an SVE header in a NEOn implementation header file ?

onnxruntime/core/mlas/lib/tanh.cpp
```diff
@@ -22,7 +22,9 @@ Module Name:
 
 #include "mlasi.h"
 #include "softmax.h"
-
+// #ifdef MLAS_USE_SVE
+// #include "sve/mlasi_sve.h"
```
hariharans29: Remove ?

onnxruntime/core/mlas/lib/mlasi.h
```diff
@@ -610,6 +610,32 @@ void
     size_t N
     );
 
+using _mlas_fp16_ = uint16_t;
```
hariharans29: Is this using needed ?

onnxruntime/core/mlas/lib/erf_neon_fp16.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+Abstract:
+
+    This module contains the procedure prototypes for the ERF NEON FP16 intrinsics.
+
+--*/
+
+#include "erf_neon_fp16.h"
+
+#if defined(__ARM_FEATURE_FP16_VECTOR_ARITHMETIC) && defined(MLAS_F16VEC_INTRINSICS_SUPPORTED)
```
hariharans29: Should these also include MLAS_NEON_INTRINSICS ?

onnxruntime/core/mlas/lib/gelu.cpp
hariharans29: But here, they just seem to be casted statically to float ? Is that correct ?

onnxruntime/core/mlas/lib/platform.cpp
hariharans29: But I am confues why they are needed for this change ? What has changed in the casting logic that needs these now ?

hariharans29:
> > Please also run the mlas tests (onnxruntime_mlas_test.exe) and onnxruntime tests (onnxruntime_test-all.exe) in your SVE environment
> > Also, the PR has had a few iterations - is the gain listed in the PR description still valid ?
> 
> Hi @hariharans29, both of the mentioned tests are passing in my environment (Graviton3E). Regarding performance, it still achieves the results stated in the description. Please proceed with the code review and CI run. Thanks.

Kicked off another round. Please also resolve copilot comments when done incorporating the change

onnxruntime/core/mlas/lib/erf_neon_fp16.cpp
sanketkaleoss: `__ARM_FEATURE_FP16_VECTOR_ARITHMETIC ` already implies NEON intrinsic support

onnxruntime/core/mlas/lib/mlasi.h
sanketkaleoss: Yes, as it it is used in `elementwise_sve_fp16.cpp` and `erf_neon_fp16.cpp`

cmake/onnxruntime_mlas.cmake
sanketkaleoss: @hariharans29 can you suggest the place where these files should be moved? As, these include `__ARM_FEATURE_FP16_VECTOR_ARITHMETIC  `, these are not completely hardware agnostic. Also, it won't throw error if the compile flag `+fp16` not present. It would just ignore the flag. I'm thinking to move all these 5 files from `line 552-556` to `line 125`. Is that okay?

hariharans29: I get that we need to include erf_neon_fp16.cpp and gelu_neon_fp16.cpp, but why do we need to include the wrapper files - gelu.cpp, erf.cpp, and platform.cpp ?

onnxruntime/core/mlas/lib/mlasi.h
hariharans29: Can it be included locally in those files rather than including in a top level header ?

cmake/onnxruntime_mlas.cmake
sanketkaleoss: @hariharans29 you're right, we don't need fp16 header files in `erf.cpp` and `gelu.cpp`. It was required in the initial commits, can be safely removed now. But, in `platform.cpp`  as kernel assignment is guarded with `__ARM_FEATURE_FP16_VECTOR_ARITHMETIC ` for bypassing MSVC build errors, we need to include `platform.cpp` right?

hariharans29: Sorry I might have missed this - What is the error you see from MSVC without that ? I am guessing there are other similar fp16 kernels used in platform.cpp - What is different from the other fp16 kernels here to necessitate this now ? Can you please expand ?

sanketkaleoss: @hariharans29 The build encountered an error in `fp16_common.h ` related to the latest FP16 APIs introduced in this PR. For example, it reported 'float16_t': undeclared identifier.
I’m not certain why similar errors were not triggered for the older FP16 intrinsics, even though they are defined under the same compile flag. To work around this, I added a compile flag while assigning the kernel to avoid this code path.
Could you please suggest if there is a cleaner or more appropriate way to implement this?

onnxruntime/core/mlas/lib/platform.cpp
sanketkaleoss: This may have been an oversight, as the flag does not seem to be required for this assignment.

onnxruntime/core/mlas/lib/gelu.cpp
sanketkaleoss: `MLAS_FP16 `is `MLFloat16`, which defines operator` float(),` so `static_cast<float>(input[i])` invokes a correct  half-to-float conversion. This looks okay right?

sanketkaleoss:
Hi @hariharans29, added new commit for resolving some of the comments and CI. Please proceed with the review and CI run. Thanks.

hariharans29:
> Hi @hariharans29, added new commit for resolving some of the comments and CI. Please proceed with the review and CI run. Thanks.

Hi @sanketkaleoss - Thanks. I am heads down with some important work at the moment and I will get to reviewing this PR soon. Meanwhile, I kicked off CI again.

hariharans29:
Can you resolve comments for already addressed comments please ?

sanketkaleoss:
> Can you resolve comments for already addressed comments please ?

Sure

sanketkaleoss:
> > Hi @hariharans29, added new commit for resolving some of the comments and CI. Please proceed with the review and CI run. Thanks.
> 
> Hi @sanketkaleoss - Thanks. I am heads down with some important work at the moment and I will get to reviewing this PR soon. Meanwhile, I kicked off CI again.

Sure, thanks

sanketkaleoss:
Hi @hariharans29, it appears that the failing CI is unrelated to the code changes. Could you please confirm if my understanding is correct? Thanks.

hariharans29:
> Hi @hariharans29, it appears that the failing CI is unrelated to the code changes. Could you please confirm if my understanding is correct? Thanks.

@sanketkaleoss - Apologies. I have this PR on my list. Will take another pass at this sometime next week hopefully. 

Re the failing CI - you're right it is not something to do with your PR.

hariharans29:
Apoogies - I have been a little busy lately - couldn't get to this. Can you please resolve the conflicts ? I am taking another look meanwhile.

sanketkaleoss:
Hi @hariharans29, I have rebased the changes and incorporated some of the Copilot comments. I also attempted to remove the `+fp16` compilation tag from` platform.cpp`; however, this may lead to CI failures on MSVC machines.
We can decide whether to retain or remove `__ARM_FEATURE_FP16_VECTOR_ARITHMETIC ` based on the CI results. Thanks.

hariharans29:
As discussed already, **please** resolve copilot comments if you have addressed them - or add reasoning if you are choosing not to take copilot guidance.

onnxruntime/core/mlas/lib/erf.cpp
```diff
@@ -266,3 +267,24 @@ Return Value:
...
+    size_t N
+    )
+{
+    if(GetMlasPlatform().ErfFP16KernelRoutine){
+        GetMlasPlatform().ErfFP16KernelRoutine(Input, Output, N);
+        return;
+    }
+    std::vector<float> input_fp32(N);
+    std::vector<float> output_fp32(N);
+
```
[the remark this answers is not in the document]
sanketkaleoss: Incorporated

onnxruntime/core/mlas/lib/erf_neon_fp16.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+using _mlas_fp16_ = uint16_t;
+// Helpers to safely convert between float and FP16-bit representation
+static float
+fp16_to_float(uint16_t h)
+{
+    __fp16 tmp;
+    memcpy(&tmp, &h, sizeof(h));
+    return (float)tmp;
```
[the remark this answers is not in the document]
sanketkaleoss: Incorporated

cmake/onnxruntime_mlas.cmake
```diff
@@ -57,6 +57,7 @@ onnxruntime_add_static_library(onnxruntime_mlas
   ${MLAS_SRC_DIR}/rotary_embedding.cpp
   ${MLAS_SRC_DIR}/softmax.h
   ${MLAS_SRC_DIR}/saturation_check.cpp
+  ${MLAS_SRC_DIR}/gelu.cpp
```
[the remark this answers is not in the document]
sanketkaleoss: Updated

cmake/onnxruntime_mlas.cmake
```diff
@@ -565,6 +578,9 @@ else()
           set_source_files_properties(${MLAS_SRC_DIR}/halfgemm_kernel_neon_fp16.cpp PROPERTIES COMPILE_FLAGS " -march=armv8.2-a+fp16 ")
           set_source_files_properties(${MLAS_SRC_DIR}/softmax_kernel_neon_fp16.cpp PROPERTIES COMPILE_FLAGS " -march=armv8.2-a+fp16 ")
           set_source_files_properties(${MLAS_SRC_DIR}/eltwise_kernel_neon_fp16.cpp PROPERTIES COMPILE_FLAGS " -march=armv8.2-a+fp16 ")
+          set_source_files_properties(${MLAS_SRC_DIR}/erf_neon_fp16.cpp PROPERTIES COMPILE_FLAGS " -march=armv8.2-a+fp16 ")
+          set_source_files_properties(${MLAS_SRC_DIR}/gelu_neon_fp16.cpp PROPERTIES COMPILE_FLAGS " -march=armv8.2-a+fp16 ")
+          set_source_files_properties(${MLAS_SRC_DIR}/platform.cpp PROPERTIES COMPILE_FLAGS " -march=armv8.2-a+fp16 ")
```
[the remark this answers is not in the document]
sanketkaleoss: Updated

onnxruntime/core/mlas/lib/platform.cpp
```diff
@@ -656,6 +660,23 @@ Return Value:
...
+            this->TanhFP16KernelRoutine = MlasSveTanhFP16Kernel;
+        }
+        else{
+            this->ErfFP16KernelRoutine = MlasNeonErfFP16Kernel;
+            this->GeluFP16KernelRoutine = MlasNeonGeluFP16Kernel; 
+        }
+    #else
+        this->ErfFP16KernelRoutine = MlasNeonErfFP16Kernel;
+        this->GeluFP16KernelRoutine = MlasNeonGeluFP16Kernel;
+    #endif
```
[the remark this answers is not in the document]
sanketkaleoss: Ignored

onnxruntime/core/providers/cpu/tensor/gelu.cc
```diff
@@ -98,4 +143,68 @@ Status Gelu<T>::Compute(OpKernelContext* context) const {
...
+
+  auto deleter = [](MLFloat16* p) {
+    AlignedFree(p);
```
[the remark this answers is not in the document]
sanketkaleoss: Ignored, suggested change causes build errors

hariharans29:
> Hi @hariharans29, I have rebased the changes and incorporated some of the Copilot comments. I also attempted to remove the `+fp16` compilation tag from` platform.cpp`; however, this may lead to CI failures on MSVC machines. We can decide whether to retain or remove `__ARM_FEATURE_FP16_VECTOR_ARITHMETIC ` based on the CI results. Thanks.

I am not able to re-run CI again because of the conflicts I think. If you resolve them, I ll re-run them. In general, I am ot a fan of adding the compiler tag for the platform.cpp file but we can see how the CI goes as you suggest

sanketkaleoss:
> > Hi @hariharans29, I have rebased the changes and incorporated some of the Copilot comments. I also attempted to remove the `+fp16` compilation tag from` platform.cpp`; however, this may lead to CI failures on MSVC machines. We can decide whether to retain or remove `__ARM_FEATURE_FP16_VECTOR_ARITHMETIC ` based on the CI results. Thanks.
> 
> I am not able to re-run CI again because of the conflicts I think. If you resolve them, I ll re-run them. In general, I am ot a fan of adding the compiler tag for the platform.cpp file but we can see how the CI goes as you suggest

@hariharans29 now the conflicts have been resolved. You can proceed with the CI. I have also run `onnxruntime_test_all` and `onnxruntime_mlas_test` for validation

cmake/onnxruntime_mlas.cmake
```diff
@@ -504,13 +508,19 @@ else()
           ${MLAS_SRC_DIR}/eltwise_kernel_neon.cpp
           ${MLAS_SRC_DIR}/sqnbitgemm_kernel_neon_int8_i8mm.cpp
           ${MLAS_SRC_DIR}/sconv_nchw_depthwise_multiplier_1.cpp
+          ${MLAS_SRC_DIR}/erf_neon_fp16.h
+          ${MLAS_SRC_DIR}/erf_neon_fp16.cpp
+          ${MLAS_SRC_DIR}/gelu_neon_fp16.cpp
```
hariharans29, line 513: gelu_neon_fp16.h here ?

onnxruntime/core/mlas/lib/platform.cpp
```diff
@@ -676,8 +697,8 @@ Return Value:
     this->QNBitGemmDispatch = &GetMlasQNBitGemmDispatchNeon(HasDotProductInstructions, HasI8MMInstructions);
 
 #if defined(MLAS_F16VEC_INTRINSICS_SUPPORTED)
-    this->CastF16ToF32Kernel = &MlasCastF16ToF32KernelNeon;
-    this->CastF32ToF16Kernel = &MlasCastF32ToF16KernelNeon;
+        this->CastF16ToF32Kernel = &MlasCastF16ToF32KernelNeon;
```
hariharans29, line 700: Nit: Unnecessary spaces (formating issue)

onnxruntime/core/mlas/lib/erf.cpp
```diff
@@ -266,3 +267,24 @@ Return Value:
...
+    const MLAS_FP16* Input,
+    MLAS_FP16* Output,
+    size_t N
+    )
+{
+    if(GetMlasPlatform().ErfFP16KernelRoutine){
+        GetMlasPlatform().ErfFP16KernelRoutine(Input, Output, N);
+        return;
+    }
+    std::vector<float> input_fp32(N);
```
hariharans29, line 283: The issue with this is that previously this temporary buffers were allocated by the CPU EP allocator which could be backed by a memory pool but now the vector will make dynamic memory allocations (because of no backing memory pool). This may cause a regression in some small models. Any way we can address this ?

onnxruntime/core/mlas/lib/gelu_neon_fp16.h
```diff
@@ -0,0 +1,32 @@
...
+Abstract:
+
+    This module contains Gelu helper functions .
+
+--*/
+
+#pragma once
+
+#include "fp16_common.h"
+#include "../inc/mlas.h"
```
hariharans29, line 21: Is mlas.h inclusion necessary here and is erf_neon_fp.h inclusion needed here ?

onnxruntime/core/mlas/lib/erf_neon_fp16.cpp
```diff
@@ -0,0 +1,156 @@
...
+#include "erf_neon_fp16.h"
+
+#if defined(MLAS_F16VEC_INTRINSICS_SUPPORTED)
+
+using _mlas_fp16_ = uint16_t;
+// Helpers to safely convert between float and FP16-bit representation
+static float
+fp16_to_float(uint16_t h)
+{
+    __fp16 tmp;
```
hariharans29, line 27: Should MSVC use some other type since it does not seem to know definition for `__fp16` type (per CI failure) ?

hariharans29: Alternatively, if you have trouble getting it to work on Windows, I'd be okay to limit this to non-Windows platforms (GCC and CLang) and coverage can be expanded to Windows in a future PR.

sanketkaleoss: Thanks for the suggestion, updating to limit this to non-windows platforms for now

onnxruntime/core/mlas/lib/erf.cpp
sanketkaleoss: @hariharans29  `IAllocatorUniquePtr` would address this, but this file doesn’t have access to that abstraction. Given that `qgemm.h `already uses a `std::unique_ptr`-based approach for similar temporary buffers, would it make sense to follow that pattern here as well, or should we instead try to plumb the allocator down to preserve pooling?

hariharans29: Hmm- thinking out loud - can the memory pool backed scratch buffer allocation be done in the CPU EP and passed along to MLAS to be used ? Are there any limitations to that approach ? Of course, it will be passed in only when we have to do the compute in fp32 due to lack of the platform specific fp16 kernel which may require an mlas api to query if the fp16 fused erf kernel functionality is preset (I don't think this is a bad approach - but let me know if you find some obvious flaw)

I am not familiar with the setup in qgemm.h but if possible I'd like to not deviate the existing flow too much from how it is currently. We don't know when these subtle changes will result in regressions especially in tiny models where such behavioral changes could show up negatively.

onnxruntime/core/mlas/lib/erf_neon_fp16.cpp
hariharans29: Sounds good. Please go ahead.

onnxruntime/core/mlas/lib/erf.cpp
sanketkaleoss: @hariharans29 Incorporated memory pooled buffer using suggested method in the latest commit

onnxruntime/core/mlas/lib/gelu_neon_fp16.h
sanketkaleoss: updated

onnxruntime/core/mlas/lib/platform.cpp
sanketkaleoss: updated

cmake/onnxruntime_mlas.cmake
sanketkaleoss: updated

onnxruntime/core/providers/cpu/tensor/gelu.cc
```diff
@@ -12,18 +12,63 @@
...
+}
+
+inline void AlignedFree(void* p) {
+#if defined(_WIN32)
+  _aligned_free(p);
+#else
+  std::free(p);
+#endif
+}
+
```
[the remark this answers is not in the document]
sanketkaleoss: Not incorporated
In this case I intentionally used a local aligned allocation to guarantee 64-byte alignment for the vectorized FP16 path.

onnxruntime/core/mlas/lib/sve/mlasi_sve.h
```diff
@@ -32,8 +32,35 @@ typedef svfloat32_t MLAS_SVFLOAT32;
 typedef svint32_t MLAS_SVINT32;
 typedef svuint32_t MLAS_SVUINT32;
 typedef svbool_t MLAS_SVBOOL;
+typedef svfloat16_t MLAS_SVFLOAT16;
+typedef svuint16_t MLAS_SVUINT16;
 
-// function decarations
+void
```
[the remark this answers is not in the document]
sanketkaleoss: Not needed, comment was removed

onnxruntime/core/mlas/lib/erf_neon_fp16.cpp
```diff
@@ -0,0 +1,156 @@
...
+    std::memcpy(&tmp, &h, sizeof(h));
+    return (float)tmp;
+}
+
+static uint16_t
+float_to_fp16(float f)
+{
+    __fp16 tmp = (__fp16)f;
+    uint16_t h;
+    memcpy(&h, &tmp, sizeof(h));
```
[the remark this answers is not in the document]
sanketkaleoss: Incorporated

onnxruntime/core/mlas/lib/gelu_neon_fp16.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+    if (algo == MlasGeluTanh) {
+        // Preprocess input into temp[] for tanh
+        for (; i + 7 < count; i += 8) {
+            MLAS_FLOAT16X8 x = MlasLoadf16Float16x8(reinterpret_cast<const float16_t*>(input + i));
+            MLAS_FLOAT16X8 x2 = MlasMultiplyFloat16(x, x);
+            MLAS_FLOAT16X8 inner = MlasMultiplyAddFloat16(v_C, x2, v_B);  // B + C * x^2
+            MLAS_FLOAT16X8 tanh_arg = MlasMultiplyFloat16(x, inner);      // x * (B + C * x^2)
+            tanh_arg = MlasMaximumFloat16(MlasBroadcastF16Float16x8(c2), MlasMinimumFloat16(tanh_arg, MlasBroadcastF16Float16x8(c1)));
+            MlasStoref16Float16x8(reinterpret_cast<float16_t*>(temp + i), tanh_arg);
```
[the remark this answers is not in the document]
sanketkaleoss: In this context, the data is only being loaded/stored, and immediately consumed by NEON intrinsics. Thus, the reinterpret cast is effectively operating on the underlying bit representation and is equivalent to the existing MLAS pattern of loading raw 16-bit values and reinterpreting them as FP16 vectors. So, the current approach is functionally correct for the intended targets.

onnxruntime/core/mlas/lib/sve/mlas_sve_fp16.h
```diff
@@ -0,0 +1,181 @@
...
+{
+    return svst1_u16(pg, Buffer, Vector);
+}
+MLAS_SVE_TARGET
+MLAS_FORCEINLINE
+void
+MlasSveStoreF16(MLAS_SVBOOL pg, MLAS_FP16* Buffer, MLAS_SVFLOAT16 Vector)
+{
+    return svst1_f16(pg, reinterpret_cast<__fp16*>(Buffer), Vector);
+}
```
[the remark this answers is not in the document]
sanketkaleoss: Similar reason as the previous comment.

sanketkaleoss:
@hariharans29 Resolved pending comments and incorporated required suggestion.

onnxruntime/core/providers/cpu/tensor/gelu.cc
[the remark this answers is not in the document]
hariharans29: Quick note about this- how hard is it to use existing core infrastructure to allocate aligned memory as copilot suggests rather than do the manual way ?

It isn't a blocker but it would be nice to know if this is supported via existing core infrastructure ?

hariharans29:
> @hariharans29 Resolved pending comments and incorporated required suggestion.

Overall - LGTM. I just re-opened discussion on one comment (manual aligned alloc and aligned free) - nothing that is a blocker.

I ll seek Copilot's opinion once more and we can merge it soon.

Thanks.

onnxruntime/core/providers/cpu/tensor/gelu.cc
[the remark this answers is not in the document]
sanketkaleoss: Sure @hariharans29 , incorporated memory allocation using existing method in the latest commit. This should be fine right?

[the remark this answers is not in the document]
hariharans29: Hi @sanketkaleoss - I checked the change, the entire aligned allocation block has been removed an uses ORT's default allocator (with default memory alignment it uses). Is this in line with your expectation/needs ? If so, why was aligned allocation needed in the first place ?

hariharans29:
Overall looks good to me:

1) I had a question on the open comment (aligned allocator)
2) Copilot has a last set of comments - please check if it adds value
3) I think the pending CI failures have nothing to do with the PR. But can you please double check ?
4) Please resolve any pending old comments so that it is mergeable

Thanks !

onnxruntime/core/mlas/lib/gelu.cpp
```diff
@@ -63,3 +62,37 @@ MlasComputeGeluErf(
...
+            // GELU approximation (tanh)
+            const float B = 0.7978845608f;
+            const float C = 0.044715f * B;
+            float tanh_arg = x * (B + C * x * x);
+            float tanh_res = std::tanh(tanh_arg);
+            gelu_val = 0.5f * x * (1.0f + tanh_res);
+        } else {
+            // GELU exact (erf)
+            gelu_val = 0.5f * x *
+                (1.0f + std::erf(x * static_cast<float>(M_SQRT1_2)));
```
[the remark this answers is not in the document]
sanketkaleoss: `M_SQRT1_2 ` is consistently available in our supported configurations. Not necessary for now.

onnxruntime/core/mlas/lib/sve/elementwise_sve_fp16.cpp
```diff
@@ -0,0 +1,256 @@
...
+        i += svcntp_b16(svptrue_b16(), pg);
+    }
+}
+
+void MLASCALL
+MlasSveGeluFP16Kernel(const MLAS_FP16* input, MLAS_FP16* output, MLAS_FP16* temp, size_t count, MLAS_GELU_ALGORITHM algo)
+{
+    const __fp16 r1 = 0.5f;
+    const __fp16 r2 = 1.0f;
+    const __fp16 r3 = static_cast<float>(M_SQRT1_2);
```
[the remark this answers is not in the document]
sanketkaleoss: Using standard `M_SQRT1_2  ` throughout the PR, tested for portability. Not necessary.

onnxruntime/core/mlas/lib/gelu_neon_fp16.cpp
```diff
@@ -0,0 +1,96 @@
...
+    const float16_t v_one1 = 1.0f;
+    const float16_t v_sqrt1_21 = static_cast<float>(M_SQRT1_2);
+    const float16_t v_B1 = 0.7978845608028654f;
+    const float16_t v_C1 = 0.035677408136300125f;
+    const float16_t c1 = 5.0f;
+    const float16_t c2 = -5.0f;
+    const MLAS_FLOAT16X8 v_half = MlasBroadcastF16Float16x8(v_half1);
+    const MLAS_FLOAT16X8 v_one = MlasBroadcastF16Float16x8(v_one1);
+    const MLAS_FLOAT16X8 v_sqrt1_2 = MlasBroadcastF16Float16x8(v_sqrt1_21);
+    const MLAS_FLOAT16X8 v_B = MlasBroadcastF16Float16x8(v_B1);
```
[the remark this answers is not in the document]
sanketkaleoss: Same reason as before. Not necessary.

onnxruntime/core/providers/cpu/tensor/gelu.cc
```diff
@@ -98,4 +107,75 @@ Status Gelu<T>::Compute(OpKernelContext* context) const {
...
+                           "Failed to allocate temporary buffer.");
+  }
+
+  auto deleter = [allocator](MLFloat16* p) {
+    if (p) allocator->Free(p);
+  };
+
+  std::unique_ptr<MLFloat16, decltype(deleter)> temp_fp16(
+      static_cast<MLFloat16*>(raw), deleter);
+
```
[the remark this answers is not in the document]
sanketkaleoss: The temp buffer is intentional to avoid aliasing between input/output and intermediates, which simplifies the vectorized kernel and keeps it consistent with MLAS patterns. Reusing output would require stricter ordering and aliasing guarantees, so I preferred keeping this explicit.
Not necessary.

onnxruntime/core/providers/cpu/tensor/gelu.cc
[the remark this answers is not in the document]
sanketkaleoss: The aligned allocator was initially used to ensure SIMD-friendly alignment, but seems like ORT’s allocator already provides sufficient alignment. Since no performance difference was observed, the custom allocation was removed for consistency.

sanketkaleoss:
> Overall looks good to me:
> 
> 1. I had a question on the open comment (aligned allocator)
> 2. Copilot has a last set of comments - please check if it adds value
> 3. I think the pending CI failures have nothing to do with the PR. But can you please double check ?
> 4. Please resolve any pending old comments so that it is mergeable
> 
> Thanks !

Hi @hariharans29 , resolved the pending comments and incorporated one of the Copilot suggestions. The CI issues appear to be unrelated, and performance has been preserved across these PR iterations. Thanks.

hariharans29 (approved):

hariharans29:
> > Overall looks good to me:
> > 
> > 1. I had a question on the open comment (aligned allocator)
> > 2. Copilot has a last set of comments - please check if it adds value
> > 3. I think the pending CI failures have nothing to do with the PR. But can you please double check ?
> > 4. Please resolve any pending old comments so that it is mergeable
> > 
> > Thanks !
> 
> Hi @hariharans29 , resolved the pending comments and incorporated one of the Copilot suggestions. The CI issues appear to be unrelated, and performance has been preserved across these PR iterations. Thanks.

Thanks for the contribution. I merged it. I take it you would have tried `onnxruntime_mlas_test.exe` and `onnxruntime_test_all.exe` in your env right ?

sanketkaleoss:
> > > Overall looks good to me:
> > > 
> > > 1. I had a question on the open comment (aligned allocator)
> > > 2. Copilot has a last set of comments - please check if it adds value
> > > 3. I think the pending CI failures have nothing to do with the PR. But can you please double check ?
> > > 4. Please resolve any pending old comments so that it is mergeable
> > > 
> > > Thanks !
> > 
> > 
> > Hi @hariharans29 , resolved the pending comments and incorporated one of the Copilot suggestions. The CI issues appear to be unrelated, and performance has been preserved across these PR iterations. Thanks.
> 
> Thanks for the contribution. I merged it. I take it you would have tried `onnxruntime_mlas_test.exe` and `onnxruntime_test_all.exe` in your env right ?

I have executed both test suites in my environment, and all tests are passing. Thank you for merging this, and for your continued reviews and suggestions.

PR #26818, opened 2025-12-17
Base test TRT update

eserscor:

### DIFF
--- a/.github/workflows/windows_tensorrt.yml
+++ b/.github/workflows/windows_tensorrt.yml
@@ -52,8 +52,8 @@ jobs:
           dir
         shell: pwsh
 
-      - name: Download TensorRT-10.14.1.48.Windows.win10.x86_64.cuda-12.9
-        run: 'azcopy.exe cp --recursive "https://lotusscus.blob.core.windows.net/models/local/TensorRT-10.14.1.48.Windows.win10.x86_64.cuda-12.9" ${{ runner.temp }}'
+      - name: Download TensorRT-10.14.1.48.Windows.win10.cuda-12.9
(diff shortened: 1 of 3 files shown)


tools/ci_build/github/linux/docker/Dockerfile.manylinux2_28_cuda
```diff
@@ -5,7 +5,7 @@
 ARG BASEIMAGE=nvidia/cuda:12.8.1-cudnn-devel-ubi8
 
 FROM $BASEIMAGE
-ARG TRT_VERSION=10.9.0.34-1.cuda12.8
+ARG TRT_VERSION=10.14.1.48-1.cuda12.8
```
tianleiwu: 10.14.1.48-1.cuda12.9 ?

tools/ci_build/github/azure-pipelines/templates/common-variables.yml
```diff
@@ -2,7 +2,7 @@ variables:
   cuda12_trt_version: '10.9.0.34'
   cuda13_trt_version: '10.13.3.9'
```
tianleiwu: Update these as well. 

Please search these version numbers in files. Just in case you missed other files as well.

eserscor: Yup, thanks! I figured I'd update Linux first then then will update windows. Once both are finished, I will publish the PR. Separating out the 'variables' portion here was intentional so I could work on a single platform at a time.

eserscor: Hm, it looks like 12.8 is pretty pervasive. It might make more sense to update 12.8 -> 12.9 then do the TRT update.

PR #26822, opened 2025-12-17
Address DML crashes in WebNN QDQ subgraph tests 

### LINKED ISSUE #26531: [ORT GPU (DML)][WebNN] two WebNN QDQ subgraph tests get crash
### Describe the issue

Edge Canary browser crashed when running `quantized leaky relu` and `quantized softmax`  two WebNN QDQ subgraph tests by ORT default GPU DML EP with the error
>GpuProcessHost: The GPU process crashed! Exit code: STATUS_ACCESS_VIOLATION.

/cc @fdwr PTAL, thanks!

### To reproduce

1. Install Windows App SDK Stable [1.8.2 (1.8.251003001)](https://learn.microsoft.com/en-us/windows/apps/windows-app-sdk/stable-channel#version-18)
2. Install [lates Edge Canary browser](https://www.microsoft.com/en-us/edge/download/insider?form=MA13FJ) 
3. Launch Edge Canary browser, navigate to `about://flags` enabled "Enables WebNN API" and "ONNX Runtime backend for WebNN" flags, then relaunch browser
4. Navigate to below test url, crash happened

|Test| Test URL|
|-- | --|
quantized leaky relu |https://wpt.live/webnn/conformance_tests/qdq_subgraph.https.any.html?device=gpu&tc=quantized%20leaky%20relu
quantized softmax |https://wpt.live/webnn/conformance_tests/qdq_subgraph.https.any.html?device=gpu&tc=quantized%20softmax

### Urgency

_No response_

### ONNX Runtime Installation

Released Package Windows App SDK Stable [1.8.2 (1.8.251003001)](https://learn.microsoft.com/en-us/windows/apps/windows-app-sdk/stable-channel#version-18)

1.23.25.928

### Execution Provider

ORT default GPU EP

adrastogi:
### Description

This change proposes a fix to some DML crashes observed in WebNN QDQ subgraph tests. The root cause of the problem as analyzed in the issue linked below is that the DML EP does not have kernels implemented for the QLinear versions of LeakyRelu and Softmax. The QDQ transformer has logic to fuse a 3-node sequence in the graph to the associated QLinear custom op. There is an operator registry which contains the mapping between EPs and the ops that the transformer consults for computing the quantized variants. 

The registry currently has the CPU and DML EPs supporting the same sets of operators (thereby assuming that the fused node would also be supported by the EP), so what I did was split out a separate registration for the LeakyRelu and Softmax ops and associate them with the CPU EP, while removing those ops from the prior list. 

I was trying to also explore other options like seeing if I could somehow check to see if the EP has a supported kernel for the op before proceeding with the assignment- I could not figure out a way to do this, though. 

I validated this change by building a private version of ORT and running the affected tests in Edge Canary. The crash did not occur for those tests with the fix in place. 

### Motivation and Context

This bug causes crashes in the web browser's GPU sandbox process, which is a blocker for WebNN origin trials. The crashes need to be resolved. 

Fixes #26531.

### DIFF
--- a/onnxruntime/core/optimizer/qdq_transformer/selectors_actions/qdq_selector_action_transformer.cc
+++ b/onnxruntime/core/optimizer/qdq_transformer/selectors_actions/qdq_selector_action_transformer.cc
@@ -163,8 +163,8 @@ void UnaryOpQDQRules(SelectorActionRegistry& qdq_selector_action_registry) {
   // Operators without DML QLinear kernel support (CPU only).
   // DML EP does not have QLinearLeakyRelu or QLinearSoftmax kernels.
   // See https://github.com/microsoft/onnxruntime/issues/26531
-  std::vector<const char*> cpu_only_providers = {kCpuExecutionProvider};
-  std::unique_ptr<NodeSelector> cpu_selector = std::make_unique<QDQ::UnarySelector>(cpu_only_providers);
+  std::vector<const char*> cpu_only_provider = {kCpuExecutionProvider};
+  std::unique_ptr<NodeSelector> cpu_selector = std::make_unique<QDQ::UnarySelector>(cpu_only_provider);
   qdq_selector_action_registry.RegisterSelectorAndAction(cpu_action_name,
                                                          {{"LeakyRelu", {}},
                                                           {"Softmax", {}}},


onnxruntime/core/optimizer/qdq_transformer/selectors_actions/qdq_selector_action_transformer.cc
```diff
@@ -148,14 +148,24 @@ void UnaryOpQDQRules(SelectorActionRegistry& qdq_selector_action_registry) {
   std::unique_ptr<Action> action = std::make_unique<QDQ::UnaryReplaceWithQLinear>(kMSDomain);
 
 #if !defined(ORT_MINIMAL_BUILD)
+  // Operators without DML QLinear kernel support (CPU only).
+  // DML EP does not have QLinearLeakyRelu or QLinearSoftmax kernels.
+  // See https://github.com/microsoft/onnxruntime/issues/26531
+  std::vector<const char*> cpu_only_providers = {kCpuExecutionProvider};
+  std::unique_ptr<NodeSelector> cpu_selector = std::make_unique<QDQ::UnarySelector>(cpu_only_providers);
+  qdq_selector_action_registry.RegisterSelectorAndAction("1DQ_CPU",
```
fdwr: Is it possible from here (unsure if this class has enough information passed to it) to ask the EP itself whether it supports the target operator (e.g. `QLinearAveragePool`) via `TryFindKernel`? Then this logic would be more generic to other EP's and less brittle over time, rather than hard-coded enums of targets. If not, this works! TY.

fdwr (approved):
👍

onnxruntime/core/optimizer/qdq_transformer/selectors_actions/qdq_selector_action_transformer.cc
adrastogi: Yeah, good question- agree that this would be a better approach, but I couldn't figure out a way to do it. The registry is populated as part of constructing the transformer, where (best as I can tell) we don't yet have the kernel manager or individual EP instances / graphs (which is the challenge you hypothesized / identified). And then it looks like TryFindKernel needs a fully constructed node, but we don't have it yet as the transformer is in the midst of doing that.

fdwr: Weird, because previous graph transformers I looked at *did* have this level of `TryFindKernel` information, but this one seems different (alas that was years ago, and I don't remember which ones they were). Oh well.

@edgchen1 or @skottmckay might know.

edgchen1: I don't think the DML EP's kernel registry is easily accessible here. some transformers do make use of the CPU EP kernel registry, but that's a special case.

edgchen1: a call to `qdq_selector_action_registry.RegisterAction()` with the same action name needs to be made in the minimal build path, otherwise it won't be able to look up the corresponding action in a minimal build.

here's an example:
https://github.com/microsoft/onnxruntime/blob/07724d5c3277d2b2bdaad7687272a9beb6bd2ef7/onnxruntime/core/optimizer/qdq_transformer/selectors_actions/qdq_selector_action_transformer.cc#L61-L116

adrastogi: Thanks! I think this explains some pipeline failures I was seeing... will address.

adrastogi: That was indeed the culprit- PR builds pass now. Thank you!

onnxruntime/core/optimizer/qdq_transformer/selectors_actions/qdq_selector_action_transformer.cc
```diff
@@ -148,18 +148,30 @@ void UnaryOpQDQRules(SelectorActionRegistry& qdq_selector_action_registry) {
...
-                                                          {"LeakyRelu", {}},
                                                           {"GlobalAveragePool", {}},
-                                                          {"Sigmoid", {}},
-                                                          {"Softmax", {}}},
+                                                          {"Sigmoid", {}}},
                                                          std::move(selector),
                                                          std::move(action));
 #else
   qdq_selector_action_registry.RegisterAction(action_name, std::move(action));
+  qdq_selector_action_registry.RegisterAction("1DQ_CPU",
```
edgchen1: thanks for making the change.

nit: would you mind following the existing pattern using variables? e.g.,:

```c++

const std::string cpu_action_name{"1DQ_CPU"};
std::unique_ptr<Action> cpu_action = std::make_unique<QDQ::UnaryReplaceWithQLinear>(kMSDomain);


qdq_selector_action_registry.RegisterSelectorAndAction(cpu_action_name, ..., std::move(cpu_selector), std::move(cpu_action));

#else

qdq_selector_action_registry.RegisterAction(cpu_action_name, std::move(cpu_action));

#endif
```

edgchen1 (approved):

PR #26824, opened 2025-12-18
Add type definitions, registration, utilities for INT2/UINT2 support

vraspar:
### Description
This is first PR towards adding 2 bit support to ORT

I will create follow up PRs to 
- Add CPU operators (CAST, QDQ, non compute ops)

### Motivation and Context

- ONNX recently added 2 bit support https://github.com/onnx/onnx/pull/7446

### DIFF
--- a/include/onnxruntime/core/framework/int2.h
+++ b/include/onnxruntime/core/framework/int2.h
@@ -77,8 +77,8 @@ struct Int2x4Base {
     const uint8_t shift = 2 * static_cast<uint8_t>(index);
     const std::byte clear_mask = ~(std::byte{0x3} << shift);
 
-    bits_ &= clear_mask;                                     // Clear 2-bit element to 0
-    bits_ |= static_cast<std::byte>((val & 0x3) << shift);   // Set 2-bit element to val
+    bits_ &= clear_mask;                                    // Clear 2-bit element to 0
+    bits_ |= static_cast<std::byte>((val & 0x3) << shift);  // Set 2-bit element to val
   }
 
   inline std::byte ToBits() const {


include/onnxruntime/core/framework/data_types.h
```diff
@@ -285,7 +287,7 @@ template <typename T>
 struct IsTensorContainedType : public IsAnyOf<T, float, uint8_t, int8_t, uint16_t, int16_t,
                                               int32_t, int64_t, std::string, bool, MLFloat16,
                                               double, uint32_t, uint64_t, BFloat16,
-                                              Int4x2, UInt4x2
+                                              Int4x2, UInt4x2, Int2x4, UInt2x4
```
hariharans29, line 290: (Optional) Is there a need to have a cmake option to disabled int2 types in the framework (and int4 by the same reasoning) or is the binary size increase negligible (asking from the POV of minimal builds which are quite binary size sensitive) ?

onnxruntime/core/framework/data_types.cc
```diff
@@ -645,6 +645,9 @@ ORT_REGISTER_TENSOR_TYPE(Float4E2M1x2);
 ORT_REGISTER_TENSOR_TYPE(Int4x2);
 ORT_REGISTER_TENSOR_TYPE(UInt4x2);
 
+ORT_REGISTER_TENSOR_TYPE(Int2x4);
```
hariharans29, line 648: (Optional) Consider implementing one (simple) op (like Flatten or Transpose perhaps) that demonstrates the usage of the new type introduced in the ORT core type system ?

vraspar: Was planning to create second PR from this branch that implements QDQ, cast, transpose and other crucial ops: https://github.com/microsoft/onnxruntime/compare/jambayk/int2-qdq

include/onnxruntime/core/framework/data_types.h
vraspar: I did see `onnxruntime_DISABLE_FLOAT4_TYPES` and `onnxruntime_DISABLE_FLOAT8_TYPES` but there isn't one to disable any other types

Issue #26821, opened 2025-12-17
[Feature Request] D3D12 External Resource Interop API for Plugin EPs


nieubank:
### Describe the feature request

# ONNX Runtime: D3D12 External Resource Import

## 1. Problem statement
Windows graphics/video apps increasingly hold data in GPU-resident D3D12 resources (captured frames, render targets, compute buffers). Today, using those resources as ONNX Runtime inputs/outputs typically requires extra copies (GPU→CPU→GPU) or EP-specific private hooks.

We need a **generic, EP-agnostic ORT public API** that lets an application:
- provide D3D12-shared resources/heaps as **zero-copy tensor storage**, and
- perform **explicit GPU↔GPU synchronization** between the app’s D3D12 work and the EP’s compute stream,
without sharing command queues and without exposing GPU virtual addresses.

## 2. Requirements (MVP)
- **Zero-copy tensor views** over imported external GPU allocations.
- **D3D12 sharing** via Win32 shared handles:
  - committed `ID3D12Resource` (buffer) shared handles
  - `ID3D12Heap` shared handles
- **Explicit synchronization** using imported external semaphores:
  - D3D12 timeline fence shared handle + 64-bit value wait/signal
- **EP capability discovery** (`CanImport*`) with clear failure: `ORT_NOT_IMPLEMENTED`.
- **Device identity matching** so apps can select an EP device compatible with a given D3D12 adapter.
- **Fits existing ORT primitives**: `OrtEpDevice`, `OrtEpFactory`, `OrtSyncStream`, opaque handles, explicit `Release*`.
- **Output tensors supported** (write access) when EP supports it.

## 3. Non-goals (MVP)
- **No queue sharing requirement** — app and EP keep independent queues/streams.
- **No GPU virtual addresses** in the public contract.
- **No implicit fallback copies** — apps choose behavior; EPs return NOT_IMPLEMENTED when unsupported.
- **Texture/surface-native tensors** are **vNext** — MVP targets resources that can be treated as dense, linearly-addressable tensor storage (D3D12 buffers / heaps).

## 4. Design overview
The public API introduces:
- `OrtExternalResourceImporter`: A **capability object** for external resource interop (memory + semaphore operations). Named following the agent-noun pattern (`Allocator`, `Importer`).
- `OrtExternalMemoryHandle`: EP-imported view of a shared external allocation.
- `OrtExternalSemaphoreHandle`: EP-imported view of a shared external semaphore.

[...]

PR #26828, opened 2025-12-18
Add OrtExternalResourceImporter API for D3D12 shared resource import

nieubank:
### Description
Introduces the OrtExternalResourceImporter API enabling execution providers to import D3D12 shared resources and timeline fences for zero-copy GPU-to-GPU data sharing with ORT inference.

Public API additions:
- OrtExternalResourceImporter capability object
- OrtExternalMemoryHandle for imported D3D12 allocations
- OrtExternalSemaphoreHandle for imported D3D12 timeline fences
- SessionGetEpDeviceForOutputs to query output EP device placement

EP Plugin API:
- OrtExternalResourceImporterImpl interface for EP implementations
- OrtEpFactory::CreateExternalResourceImporterForDevice extension

Design:
- No GPU virtual addresses in public API
- EP-agnostic design allows any EP to implement import
- Capability discovery with explicit ORT_NOT_IMPLEMENTED
- Follows existing patterns (Allocator, DataTransfer, SyncStream)

Includes example_plugin_ep mock implementation and autoep tests.

### Motivation and Context
#26821


include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -6608,6 +6692,223 @@ struct OrtApi {
...
+  ORT_API2_STATUS(ExternalResourceImporter_CanImportMemory,
+                  _In_ const OrtExternalResourceImporter* importer,
+                  _In_ OrtExternalMemoryHandleType handle_type,
+                  _Out_ bool* out_supported);
+
+  /** \brief Import external memory into the execution provider.
+   *
+   * \param[in] importer The OrtExternalResourceImporter instance.
+   * \param[in] desc Descriptor containing the external memory handle and properties.
+   * \param[out] out_handle Output parameter set to the created OrtExternalMemoryHandle.
```
skottmckay: nit: document that the caller owns the out_handle and should call ReleaseExternalMemoryHandle.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -6608,6 +6692,223 @@ struct OrtApi {
...
+  /** \brief Create a tensor backed by imported external memory.
+   *
+   * The created tensor is a view over the imported memory and does not copy data.
+   * The OrtExternalMemoryHandle must remain valid for the lifetime of the tensor.
+   *
+   * \param[in] importer The OrtExternalResourceImporter instance.
+   * \param[in] mem_handle The imported external memory handle.
+   * \param[in] tensor_desc Descriptor specifying tensor element type, shape, and optional offset.
+   * \param[in] tensor_location Optional OrtMemoryInfo for the tensor location. May be nullptr.
+   * \param[out] out_tensor Output parameter set to the created OrtValue containing the tensor.
```
skottmckay: nit: document that caller owns and should call ReleaseValue

include/onnxruntime/core/framework/run_options.h
```diff
@@ -51,6 +51,11 @@ struct OrtRunOptions {
 
   onnxruntime::InlinedVector<const onnxruntime::lora::LoraAdapter*> active_adapters;
 
+  // Optional sync stream for external resource import.
+  // When set, the EP uses this stream for execution, enabling proper
+  // synchronization with imported external semaphores.
+  OrtSyncStream* sync_stream = nullptr;
```
skottmckay: We need to hook this up with the logic in the execution which currently gets streams from SessionState::AcquireDeviceStreamCollection. I can help out with the implementation details there as that code is somewhat convoluted to understand.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -1413,6 +1581,32 @@ struct OrtEpFactory {
...
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \note Implementation of this function is optional.
+   *       An EP factory should only implement this if it supports external resource import.
+   *       If not implemented or not supported, return ORT_NOT_IMPLEMENTED or set out_importer to nullptr.
+   *
+   * \since Version 1.24.
+   */
+  ORT_API2_STATUS(CreateExternalResourceImporterForDevice, _In_ OrtEpFactory* this_ptr,
+                  _In_ const OrtMemoryDevice* memory_device,
```
skottmckay: I'm 50/50 on whether we should pass in the OrtMemoryDevice or the OrtEpDevice. 

Leaning slightly towards the OrtEpDevice as the factory owns that (so we're not really decoupling much by passing in the OrtMemoryDevice) and in theory there may be other values at the OrtEpDevice level that the implementation might want to take advantage of. 

I don't have any concrete scenarios where OrtEpDevice would be required, however that is a place where EP options that the user can set would be available if they were ever relevant.

onnxruntime/core/session/onnxruntime_c_api.cc
```diff
@@ -3536,6 +3564,297 @@ ORT_API_STATUS_IMPL(OrtApis::GetModelCompatibilityForEpDevices,
...
+
+  ~ExternalResourceImporterWrapper() {
+    if (impl && impl->Release) {
+      impl->Release(impl);
+    }
+  }
+
+  // Non-copyable
+  ExternalResourceImporterWrapper(const ExternalResourceImporterWrapper&) = delete;
+  ExternalResourceImporterWrapper& operator=(const ExternalResourceImporterWrapper&) = delete;
```
skottmckay: nit: ORT_DISALLOW_COPY_AND_ASSIGNMENT. or maybe ORT_DISALLOW_COPY_ASSIGNMENT_AND_MOVE?

onnxruntime/core/session/onnxruntime_c_api.cc
```diff
@@ -3536,6 +3564,297 @@ ORT_API_STATUS_IMPL(OrtApis::GetModelCompatibilityForEpDevices,
...
+
+  *out_importer = nullptr;
+
+  const OrtDevice* device = ep_device->device_memory_info ? &ep_device->device_memory_info->device : nullptr;
+  if (device == nullptr || device->MemType() != OrtDevice::MemType::DEFAULT) {
+    return OrtApis::CreateStatus(ORT_INVALID_ARGUMENT, "ep_device does not use DEFAULT memory of a non-CPU device.");
+  }
+
+  const auto* factory = ep_device->ep_factory;
+  if (factory == nullptr || factory->CreateExternalResourceImporterForDevice == nullptr) {
```
skottmckay: check the ep_factory's version supported. would need to be >= 24 for these new functions to be available.

onnxruntime/core/session/plugin_ep/ep_factory_provider_bridge.h
```diff
@@ -65,6 +65,16 @@ class ProviderBridgeEpFactory : public EpFactoryInternalImpl {
     return ep_factory_.CreateSyncStreamForDevice(&ep_factory_, device, stream_options, stream);
   }
 
+  OrtStatus* CreateExternalResourceImporterForDevice(
+      const OrtMemoryDevice* device,
+      OrtExternalResourceImporterImpl** importer) noexcept override {
+    if (ep_factory_.CreateExternalResourceImporterForDevice == nullptr) {
```
skottmckay: Check version supported as well

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
nieubank: Good call - will updated to pass OrtEpDevice* instead of OrtMemoryDevice*. This simplifies the API and gives EP implementations direct access to the full device context (including EP options/metadata) without requiring additional lookups. The factory can still extract the OrtMemoryDevice internally if needed, but now has flexibility to access other OrtEpDevice properties in the future.

include/onnxruntime/core/framework/run_options.h
nieubank: Ah yes, I didn't get to that one. The currently wired up path is pretty close to ready I think with the next iteration I'm going to push. I think it is best to pull out the run option for the first pass and add that as a follow up to avoid making this more complex.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -24,6 +24,68 @@ ORT_RUNTIME_CLASS(DataTransferImpl);
...
+ *   CUexternalMemory ext_memory;
+ *   CUdeviceptr mapped_ptr;
+ *   bool is_dedicated;
+ * };
+ * \endcode
+ *
+ * \since Version 1.24.
+ */
+#define ORT_EXTERNAL_MEMORY_HANDLE_VERSION 1
+struct OrtExternalMemoryHandle {
```
skottmckay, line 45: Do we need a separate value or is the ORT version sufficient, as we use that everywhere else?

nieubank: ORT version is sufficient

onnxruntime/test/autoep/library/example_plugin_ep/ep_external_resource_importer.h
```diff
@@ -0,0 +1,137 @@
...
+    handle_type = ORT_EXTERNAL_MEMORY_HANDLE_TYPE_D3D12_RESOURCE;
+    size_bytes = 0;
+    offset_bytes = 0;
+    Release = ReleaseCallback;
+  }
+
+  ~ExampleExternalMemoryHandle() {
+    // Free the simulated pointer if allocated
+    if (simulated_ptr != nullptr) {
+      free(simulated_ptr);
```
yuslepukhin: > free(simulated_ptr); [](http://example.com/codeflow?start=5&length=21)

No frees

yuslepukhin (dismissed):
:shipit:

nieubank:
@skottmckay / @yuslepukhin - any concerns with getting this in this week prior to the 1.24 snap? 

I think this provides a sufficient baseline for achieving the core of the asks around shared resources and we continue iterating on specific EP implementations with this foundation.

yuslepukhin:
Looks great to me

skottmckay:
@gaugarg-nv and @gedoensmax can you please take a look and let us know if there are any gaps/issues with this approach?

onnxruntime/core/session/ep_interop_api.cc
```diff
@@ -0,0 +1,389 @@
...
+
+  *out_importer = nullptr;
+
+  // OrtEpFactory::CreateExternalResourceImporterForDevice was added in ORT 1.24.
+  const auto* factory = ep_device->ep_factory;
+  if (factory == nullptr ||
+      factory->ort_version_supported < 24 ||
+      factory->CreateExternalResourceImporterForDevice == nullptr) {
+    // EP doesn't support external resource import - not an error, just return nullptr
+    return nullptr;
```
yuslepukhin, line 59: Generally, this type of dual result reporting is not encouraged since the caller must check two things.
I would expect to get the importer on success.
We have NOT_IMPLEMENTED error code.

skottmckay: This is optional for the EP to provide and is clearly documented that it returns nullptr if the EP doesn't provide. To me that's not an error because there's no expectation that it should be implemented. 

If we return NOT_IMPLEMENTED that creates the need to handle both a 'real' error and a 'not implemented' which feels more complicated to me as I can't use a default "if status != nullptr bad stuff has happened" handler.

nieubank: Good debate here, I'm think it's best to go back to the nullptr pattern here. This is capability discovery - the EP may or may not support this optional feature. Returning nullptr on success lets callers use the simple "if status != nullptr, something went wrong" pattern, then check the output separately for capability.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -955,6 +961,84 @@ typedef void (*RunAsyncCallbackFn)(void* user_data, OrtValue** outputs, size_t n
...
  * \nosubgrouping
  */
+
+/** \addtogroup Global
+ * @{
+ */
+
+/** \brief External memory handle type for importing GPU resources.
+ *
+ * \since Version 1.24.
```
skottmckay, line 975: Can we add something like  `\todo Add Vulkan types` to make it clear this is not just for D3D12? #Closed

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -6667,6 +6751,45 @@ struct OrtApi {
...
+   */
+  const OrtInteropApi*(ORT_API_CALL* GetInteropApi)(void);
+
+  /** \brief Get the EP device assigned to each session output.
+   *
+   * Returns the OrtEpDevice assigned to each output of the session after graph partitioning.
+   * This allows validation that outputs are placed on the expected device for external resource sharing.
+   *
+   * The EP device for each output is determined by which execution provider claims that output
+   * during graph partitioning. This information is useful for:
```
skottmckay: User doesn't necessarily have any idea what 'partitioning' is so might be more easily understood this way

```suggestion
   * The EP device for each output is determined by which execution provider will produce that output
   * during inferencing. This information is useful for:
``` #Closed

File: include/onnxruntime/core/session/onnxruntime_c_api.h [hunk 13]
```diff
@@ -7464,6 +7587,219 @@ struct OrtCompileApi {
...
+ * between ORT inference and other GPU workloads (e.g., D3D12 applications, media pipelines).
+ *
+ * The API is designed to be EP-agnostic and can be extended to support various GPU interop mechanisms
+ * (D3D12 shared handles, CUDA external memory, Vulkan, etc.).
+ *
+ * Example usage (error handling not shown):
+ *   const OrtInteropApi* interop_api = ort_api->GetInteropApi();
+ *   OrtExternalResourceImporter* importer = NULL;
+ *
+ *   status = interop_api->CreateExternalResourceImporterForDevice(ep_device, &importer);
```
skottmckay, line 7599: Should we add the check if the returned `importer` is nullptr here with a comment saying this is optional for the EP to implement to really drive that point home? #Closed

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -7464,6 +7587,219 @@ struct OrtCompileApi {
...
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \since Version 1.24.
+   */
+  ORT_API2_STATUS(CreateTensorFromMemory,
+                  _In_ OrtExternalResourceImporter* importer,
+                  _In_ const OrtExternalMemoryHandle* mem_handle,
+                  _In_ const OrtExternalTensorDescriptor* tensor_desc,
+                  _In_opt_ const OrtMemoryInfo* tensor_location,
```
skottmckay: Not quite clear to me when tensor_location should be provided. 

We create the importer with an OrtEpDevice so internally it should know the applicable memory info. The only thing I can think of would be differentiating between device and host accessible memory, but I'm not sure there's any scenario where you'd want/need to use this with host accessible memory. We also don't seem to be using tensor_location in the currently implementation so maybe this argument can be removed. #Closed

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -7464,6 +7587,219 @@ struct OrtCompileApi {
...
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \since Version 1.24.
+   */
+  ORT_API2_STATUS(CanImportSemaphore,
+                  _In_ const OrtExternalResourceImporter* importer,
+                  _In_ OrtExternalSemaphoreType type,
+                  _Out_ bool* out_supported);
+
+  /** \brief Import an external semaphore into the execution provider.
```
skottmckay, line 7733: nit: might be helpful to mention usage of the OrtExternalSemaphoreHandle that is created with WaitSemaphore and an OrtSyncStream. what the 'import' does is a little vague without that connection. #Closed

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -955,6 +961,84 @@ typedef void (*RunAsyncCallbackFn)(void* user_data, OrtValue** outputs, size_t n
...
+ *       This ensures forward compatibility as fields may be added in future versions.
+ *
+ * \since Version 1.24.
+ */
+typedef struct OrtExternalTensorDescriptor {
+  uint32_t version;                       /**< Must be ORT_API_VERSION */
+  ONNXTensorElementDataType element_type; /**< Data type of tensor elements */
+  const int64_t* shape;                   /**< Array of dimension sizes */
+  size_t rank;                            /**< Number of dimensions */
+  size_t offset_bytes;                    /**< Optional offset within imported memory (default 0) */
```
skottmckay: Do we need to support an offset here? I can understand it being in OrtExternalMemoryDescriptor, but as this is tensor info when would the offset be set and/or not defined in OrtExternalMemoryDescriptor? #Closed

onnxruntime/test/autoep/library/example_plugin_ep/ep_external_resource_importer.cc
```diff
@@ -0,0 +1,272 @@
...
+      0,  // device ID
+      OrtMemTypeDefault,
+      &memory_info);
+
+  if (status != nullptr) {
+    return status;
+  }
+
+  // Calculate buffer size
+  size_t buffer_size = handle->size_bytes - handle->offset_bytes - tensor_desc->offset_bytes;
```
skottmckay, line 135: This feels a little confusing. The `tensor_desc` has `shape` which defines the size when combined with `element_type`, but then we also have `tensor_desc->offset_bytes` being used to reduce the size of the buffer. #Closed

onnxruntime/test/autoep/library/example_plugin_ep/ep_factory.cc
```diff
@@ -308,3 +310,22 @@ OrtStatus* ORT_API_CALL ExampleEpFactory::CreateSyncStreamForDeviceImpl(OrtEpFac
 
   return nullptr;
 }
+
+/*static*/
+OrtStatus* ORT_API_CALL ExampleEpFactory::CreateExternalResourceImporterForDeviceImpl(
+    OrtEpFactory* this_ptr,
+    const OrtEpDevice* /*ep_device*/,
```
skottmckay, line 317: should we capture the specific device in the returned importer so we know for sure the device involved? not required in the example code, but if a factory potentially creates multiple OrtEpDevice instances for different physical devices it might be. #Closed

onnxruntime/test/autoep/test_external_resource_importer.cc
```diff
@@ -0,0 +1,393 @@
...
+  OrtStatus* status = GetInteropApi().CreateExternalResourceImporterForDevice(ep_device_, &importer);
+  ASSERT_EQ(status, nullptr) << "CreateExternalResourceImporterForDevice should succeed";
+  if (importer == nullptr) {
+    GTEST_SKIP() << "External resource interop not supported";
+  }
+
+  const OrtExternalMemoryAccessMode access_modes[] = {
+      ORT_EXTERNAL_MEMORY_ACCESS_READ_WRITE,
+      ORT_EXTERNAL_MEMORY_ACCESS_READ_ONLY,
+      ORT_EXTERNAL_MEMORY_ACCESS_WRITE_ONLY};
```
skottmckay: How do these get used in practical terms? We create an OrtValue eventually with the imported memory, but the data pointer inside that is non-const, with constness being at the OrtValue level. If I create an OrtValue with read only external memory what ensures there's only a `const OrtValue` containing that? #Closed

include/onnxruntime/core/session/onnxruntime_c_api.h
nieubank: Added TODOs for other various types

include/onnxruntime/core/session/onnxruntime_c_api.h
nieubank: Yes, added example

include/onnxruntime/core/session/onnxruntime_c_api.h
nieubank: You're right, hold over from previous proof of concept. Removed the parameter entirely - the importer already knows the device and memory info from creation.

include/onnxruntime/core/session/onnxruntime_c_api.h
nieubank: Yes - enables multiple tensors from the same imported memory:
- `OrtExternalMemoryDescriptor::offset_bytes` - Base offset into the imported resource (e.g., start of usable region in a heap)
- `OrtExternalTensorDescriptor::offset_bytes` - Additional offset for each specific tensor

An example would be importing a 100MB D3D12 heap once, then create multiple tensors at different offsets within it. I added clarifying documentation to both structures.

onnxruntime/test/autoep/library/example_plugin_ep/ep_external_resource_importer.cc
nieubank: Added a comment in the example code noting this is a simplified calculation for testing.

onnxruntime/test/autoep/library/example_plugin_ep/ep_factory.cc
nieubank: Added a comment in the example code noting this consideration for production multi-GPU EPs

onnxruntime/test/autoep/test_external_resource_importer.cc
nieubank: You're right, more hold over from the poc - the access mode wasn't being used for anything in the implementation, and for D3D12, the resource properties can be queried after calling OpenSharedHandle() via GetDesc(). The resource already has its flags and capabilities baked in, making the field redundant.

Removed OrtExternalMemoryAccessMode enum and the access_mode field from OrtExternalMemoryDescriptor. EPs can query resource properties directly from the imported handle as needed.

gedoensmax:
@skottmckay thanks for tagging, this looks good to me. One question i would have is if we could handle e.g. overallocation of allocation with a callback could work. Usually one can simply bin a memory info and rely on ORT to handle the correct allocation size. With this approach it will require preallocation and importing the memory beforehand with the correct shape i assume. 
I think this is an edge case just wanted to hear your thoughts.

nieubank:
> @skottmckay thanks for tagging, this looks good to me. One question i would have is if we could handle e.g. overallocation of allocation with a callback could work. Usually one can simply bin a memory info and rely on ORT to handle the correct allocation size. With this approach it will require preallocation and importing the memory beforehand with the correct shape i assume. I think this is an edge case just wanted to hear your thoughts.

Right, good callout, the current design is intentionally kept simple and focused on the core import-existing-memory scenario for basic zero-copy interop.

The dual offset pattern OrtExternalMemoryDescriptor::offset_bytes + OrtExternalTensorDescriptor::offset_bytes does allow importing a larger buffer and carving out regions, but I agree it's not quite the same as ORT-driven binning strategies.

This is definitely something we can build on top of in the future, through allocator callbacks or other extensions.

skottmckay (approved):
:shipit:

PR #26829, opened 2025-12-18
[TRTRTX EP] Implement OrtExternalResourceImporter for D3D12-CUDA interop

nieubank:
### Description
This commit adds the OrtExternalResourceImporter implementation for the NvTensorRtRtx execution provider, enabling zero-copy D3D12 to CUDA memory sharing and GPU synchronization.

Implementation:
- NvTrtRtxExternalResourceImporterImpl: Full implementation of the OrtExternalResourceImporter interface using CUDA Driver APIs
- Memory import: cuImportExternalMemory for D3D12_RESOURCE and D3D12_HEAP
- Semaphore import: cuImportExternalSemaphore for D3D12_FENCE
- Tensor creation: CreateTensorFromMemory wraps imported CUDA device pointers
- Synchronization: WaitSemaphore/SignalSemaphore using cuWaitExternalSemaphoresAsync/cuSignalExternalSemaphoresAsync

Tests (nv_external_resource_importer_test.cc):
- CreateExternalResourceImporter: Basic importer creation
- CanImportMemoryCapabilities: D3D12 Resource/Heap capability queries
- CanImportSemaphoreCapabilities: D3D12 Fence capability queries
- ImportD3D12SharedResource: Memory import validation
- CreateTensorFromImportedMemory: Tensor creation with CUDA device ptr verification
- ImportD3D12Fence: Semaphore import validation
- WaitAndSignalSemaphore: Bidirectional D3D12-CUDA sync
- FullInferenceWithExternalMemory: E2E test with ReLU model verifying D3D12 upload -> CUDA inference -> D3D12 readback pipeline

### Motivation and Context
#26821

### DIFF
--- a/include/onnxruntime/core/session/onnxruntime_c_api.h
+++ b/include/onnxruntime/core/session/onnxruntime_c_api.h
@@ -881,6 +881,9 @@ typedef struct OrtModelEditorApi OrtModelEditorApi;
 struct OrtCompileApi;
 typedef struct OrtCompileApi OrtCompileApi;
 
+struct OrtInteropApi;
+typedef struct OrtInteropApi OrtInteropApi;
+
 struct OrtEpApi;
 typedef struct OrtEpApi OrtEpApi;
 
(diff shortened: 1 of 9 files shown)


onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc
```diff
@@ -516,32 +519,481 @@ struct NvTrtRtxSyncStreamImpl : OrtSyncStreamImpl {
...
+    OrtMemoryInfo* memory_info = nullptr;
+    OrtStatus* status = impl.ort_api.CreateMemoryInfo_V2(
+        "NvTensorRTRTX",
+        OrtMemoryInfoDeviceType_GPU,
+        OrtDevice::VendorIds::NVIDIA,
+        impl.device_id_,
+        OrtDeviceMemoryType_DEFAULT,
+        0,  // alignment
+        OrtDeviceAllocator,
+        &memory_info);
```
skottmckay: Ideally the OrtMemoryInfo info should come from the OrtEpDevice. That's available to an API user via EpDevice_MemoryInfo. Either the OrtMemoryInfo or OrtEpDevice could be passed into this call depending on whether anything else from OrtEpDevice is likely to be required to implement this for other IHVs. #Closed

onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc
```diff
@@ -516,32 +519,481 @@ struct NvTrtRtxSyncStreamImpl : OrtSyncStreamImpl {
...
+
+    // Validate offset does not exceed allocation size
+    if (desc->offset_bytes > desc->size_bytes) {
+      return impl.ort_api.CreateStatus(ORT_INVALID_ARGUMENT,
+                                       "offset_bytes exceeds size_bytes in OrtExternalMemoryDescriptor");
+    }
+
+    // Set CUDA device
+    CUresult cu_result = cuCtxSetCurrent(nullptr);  // Reset context
+    CUDA_RETURN_IF_ERROR(cudaSetDevice(impl.device_id_));
```
skottmckay: What happens if there's an import for a different device (e.g. you have 2 NV GPUs being used in different sessions)? Does memory previously imported stay valid if the current device changes?

onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc
```diff
@@ -516,32 +519,481 @@ struct NvTrtRtxSyncStreamImpl : OrtSyncStreamImpl {
...
+      case ORT_EXTERNAL_MEMORY_HANDLE_TYPE_D3D12_RESOURCE:
+        cu_handle_type = CU_EXTERNAL_MEMORY_HANDLE_TYPE_D3D12_RESOURCE;
+        is_dedicated = true;  // D3D12 committed resources are dedicated
+        break;
+      case ORT_EXTERNAL_MEMORY_HANDLE_TYPE_D3D12_HEAP:
+        cu_handle_type = CU_EXTERNAL_MEMORY_HANDLE_TYPE_D3D12_HEAP;
+        is_dedicated = false;  // D3D12 heaps are not dedicated
+        break;
+      default:
+        return impl.ort_api.CreateStatus(ORT_NOT_IMPLEMENTED, "Unknown external memory handle type");
```
skottmckay: Should this be an error as we'd only call this function if 'can import' returned true.

onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc
```diff
@@ -516,32 +519,481 @@ struct NvTrtRtxSyncStreamImpl : OrtSyncStreamImpl {
...
+
+    CUdeviceptr mapped_ptr = 0;
+    cu_result = cuExternalMemoryGetMappedBuffer(&mapped_ptr, ext_memory, &buffer_desc);
+    if (cu_result != CUDA_SUCCESS) {
+      cuDestroyExternalMemory(ext_memory);
+      const char* error_str = nullptr;
+      cuGetErrorString(cu_result, &error_str);
+      std::string error_msg = "cuExternalMemoryGetMappedBuffer failed: ";
+      error_msg += error_str ? error_str : "unknown error";
+      return impl.ort_api.CreateStatus(ORT_FAIL, error_msg.c_str());
```
skottmckay: For EP implementation code ORT_EP_FAIL is more correct than ORT_FAIL.

```suggestion
      return impl.ort_api.CreateStatus(ORT_EP_FAIL, error_msg.c_str());
``` #Closed

File: onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc [hunk 5]
```diff
@@ -516,32 +519,481 @@ struct NvTrtRtxSyncStreamImpl : OrtSyncStreamImpl {
...
+    }
+
+    // Create and return the handle wrapper
+    auto* handle = new (std::nothrow) NvTrtRtxExternalMemoryHandle();
```
skottmckay: nit: Use a std::unique_ptr and call `.release()` to return.  #Closed

onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc
```diff
@@ -516,32 +519,481 @@ struct NvTrtRtxSyncStreamImpl : OrtSyncStreamImpl {
...
+
+    if (mem_handle == nullptr || tensor_desc == nullptr || out_tensor == nullptr) {
+      return impl.ort_api.CreateStatus(ORT_INVALID_ARGUMENT, "Invalid arguments to CreateTensorFromMemory");
+    }
+
+    // Validate descriptor version
+    if (tensor_desc->version != ORT_EXTERNAL_TENSOR_DESCRIPTOR_VERSION) {
+      return impl.ort_api.CreateStatus(ORT_INVALID_ARGUMENT,
+                                       "Invalid OrtExternalTensorDescriptor version");
+    }
```
skottmckay: Typically this only needs to be used when changes are made to the struct, so we're not looking for equality but rather selectively enabling usage of new things if available. e.g. a new field is added in version 24 but if tensor_desc->version is 23 we can't use that new field.  #Closed

onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc
```diff
@@ -516,32 +519,481 @@ struct NvTrtRtxSyncStreamImpl : OrtSyncStreamImpl {
...
+        0,  // alignment
+        OrtDeviceAllocator,
+        &memory_info);
+
+    if (status != nullptr) {
+      return status;
+    }
+
+    // Create tensor with pre-allocated memory
+    status = impl.ort_api.CreateTensorWithDataAsOrtValue(
```
skottmckay: I'm assuming the interop layer is only used for inference inputs and outputs, and therefore the user manages the lifetime of the OrtValue and there's no need to ever use something like CreateTensorWithDataAndDeleterAsOrtValue. #Closed

onnxruntime/test/providers/nv_tensorrt_rtx/nv_external_resource_importer_test.cc
```diff
@@ -0,0 +1,848 @@
...
+  Ort::AllocatedStringPtr output_name = session.GetOutputNameAllocated(0, allocator);
+
+  io_binding.BindInput(input_name.get(), Ort::Value(input_tensor));
+  io_binding.BindOutput(output_name.get(), Ort::Value(output_tensor));
+  io_binding.SynchronizeInputs();
+
+  // Run inference with synchronization disabled (we handle it manually)
+  Ort::RunOptions run_options;
+  run_options.AddConfigEntry("disable_synchronize_execution_providers", "1");
+  session.Run(run_options, io_binding);
```
skottmckay: Would be good to note that disable_synchronize_execution_providers only affects the end of inferencing behavior (according to the comment). Not required if the output is not being hooked up to an external resource.

Also might be an area that needs validation around how the stream uses memory if involved in multiple inferences with this option set. We can't cleanup a bunch of things (don't know if all operations have completed so we don't call Stream::CleanUpOnRunEnd ) at the end of inferencing, and it's not clear to me if they should be manually cleaned up later (either via a new public function or by releasing the OrtSyncStream and creating a new one) or whether it's fine to leave them around across multiple inferences.

onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc
skottmckay: Actually the OrtExternalResourceImporterImpl should be associated with the OrtEpDevice so potentially we can get the OrtEpDevice from `this_ptr` and the memory info from that.

onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc
nieubank: yes, will add a comment clarifying this

onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc
nieubank: This is simplfied greatly when passing the OrtEpDevice directly.

onnxruntime/test/providers/nv_tensorrt_rtx/nv_external_resource_importer_test.cc
nieubank: I dropped the disable_synchronize_execution_providers override so the test now runs with the default synchronization path, avoiding the deferred-buffer cleanup issue you mentioned. We can revisit that perf optimization once RunOptions::sync_stream is wired into the execution pipeline.

onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc
nieubank: Added a comment to clarify here

onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc
```diff
@@ -735,6 +1215,43 @@ struct NvTensorRtRtxEpFactory : OrtEpFactory {
...
   }
 
+  /** @brief Create an external resource importer for D3D12 to CUDA import.
+   *
+   * This enables zero-copy import of D3D12 shared resources and timeline fences.
+   * The implementation uses CUDA Driver APIs (cuImportExternalMemory, cuImportExternalSemaphore).
+   *
+   * @param this_ptr The OrtEpFactory instance.
+   * @param ep_device The OrtEpDevice to create the importer for (must have "device_id" in ep_options).
+   * @param out_importer Output parameter set to the created OrtExternalResourceImporterImpl.
```
skottmckay, line 1225: The "device_id" in ep_options should have been processed when creating the OrtEpDevice, with that value ending up in the OrtMemoryDevice device_id (MemoryDevice_GetDeviceId)

nieubank: yes, outdated/redundant comment now, removed

onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc
```diff
@@ -516,32 +519,509 @@ struct NvTrtRtxSyncStreamImpl : OrtSyncStreamImpl {
...
+    const OrtMemoryInfo* memory_info = impl.ep_device_->device_memory_info;
+
+    // Create tensor that references the imported memory. The tensor does not own the memory -
+    // the user manages the lifetime of both the OrtValue and OrtExternalMemoryHandle.
+    // The user must keep the handle alive while the tensor is in use.
+    // No deleter is needed since this is for inference inputs/outputs where the user controls lifetime.
+    OrtStatus* status = impl.ort_api.CreateTensorWithDataAsOrtValue(
+        memory_info,
+        data_ptr,
+        handle->size_bytes - tensor_desc->offset_bytes,
```
yuslepukhin, line 798: >handle->size_bytes - tensor_desc->offset_bytes, [](http://example.com/codeflow?start=8&length=47)

Should we check for overlfow?

onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc
```diff
@@ -516,32 +519,509 @@ struct NvTrtRtxSyncStreamImpl : OrtSyncStreamImpl {
...
+
+    // Validate tensor offset does not exceed available buffer size
+    size_t available_size = handle->size_bytes - handle->offset_bytes;
+    if (tensor_desc->offset_bytes > available_size) {
+      return impl.ort_api.CreateStatus(ORT_INVALID_ARGUMENT,
+                                       "tensor offset_bytes exceeds available imported memory size");
+    }
+
+    // Calculate the data pointer with tensor offset
+    void* data_ptr = reinterpret_cast<void*>(handle->mapped_ptr + tensor_desc->offset_bytes);
```
yuslepukhin, line 786: >handle->mapped_ptr + tensor_desc->offset_byte [](http://example.com/codeflow?start=45&length=45)

Suggest a check that this does not bring us outside tensor memory.

praneshgo:
[ Repeating comment from https://github.com/microsoft/onnxruntime/issues/26821 ]
Setting up CUDA context in a certain way using CIG params, ID3D12Device and ID3D12CommandQueue has a lot of performance benefits (reference - https://github.com/microsoft/onnxruntime/pull/26543 there's a CUContext setup in [nv_provider_factory.cc:SetupCigContextImpl](https://github.com/microsoft/onnxruntime/pull/26543/files#diff-026edd29533f9cabc45e49a95906434460e3c5e4a78bcb88c8f0be65b895ced6)). On NV GPUs, this allows simultaneous execution of graphics and compute work on the GPU thereby maximizing its utilization and avoiding unnecessary context switch penalty.
We ideally would want to setup this context as not doing so would mean missing out on these critical performance improvements.

gedoensmax:
Additions in  #26948, otherwise the concept looks good to me.

PR #26833, opened 2025-12-19
[WebGPU] Fix Conv prepacking mismatch with Im2ColMatMul path detection

jchen10:
Fix a bug where Conv kernel prepacking could incorrectly happen when the Im2ColMatMul execution path would be taken at runtime. This caused a "Missing Input" error because:

1. In PrePackInternal, the is_fused template parameter was passed to CanApplyIm2ColMatMulProgram
2. In ComputeInternal, activation_.activation_kind_ != ActivationKind::None was passed instead

For fused Conv kernels (is_fused=true) without an actual activation set, this mismatch caused CanApplyIm2ColMatMulProgram to return false during prepacking (allowing prepacking to occur) but true at runtime (selecting the Im2ColMatMul path which needs context.Input(1)).

The fix uses the same activation check in both places to ensure consistent path detection between prepack time and compute time.


jchen10:
A quick fix. PTAL.
@qjia7 @fs-eire @guschmue

qjia7 (approved):

jchen10:
The CI errors seem irrelevant to this PR.

PR #26834, opened 2025-12-19
[MLAS] Add an NHWC implementation of convolution to avoid transposes

orlmon01:
* Modification to the CPU EP to specify channels_last when data format is NWHC
* Added a FusedNhwcConv kernel
* Implementation of the kernel in mlas
* Added compiler guards so it is inly used with KleidiAi (for now, can be removed if needed)
* Added unittests

### Description
Currently OnnxRT supports NCHW as a default datalayout. For optimisations and kernels that operate better in NHWC layout, or where the datalayout is NHWC in the first place Transposes are added around the layers. This patch seeks to eliminate them in cases of convolutions where it would cause a performance decrease. 

### Motivation and Context
KleidiAi specific implementation of this feature. Only supports convolutions, DepthWise to follow. Currently a little strict with the filters as a result.


orlmon01:

orlmon01:
Feedback appreciated as this PR makes quite a lot of changes to the codebase well outside of the normal KleidiAI scope.

Rohanjames1997:
Hi @orlmon01, I imagine that avoiding transposes also improves performance.
Do you have any performance results to share?
TIA!

orlmon01:
> Hi @orlmon01, I imagine that avoiding transposes also improves performance. Do you have any performance results to share? TIA!

Hiya,
Sorry, just back after the holidays. Yes, there is a performance increase. It depends on the model. Ones with multiple consecutive Convolutions where transposes can be eliminated will see a larger speedup. Even with the limited range of convolutions it's implemented for there should still be a performance increase in most cases. 

I have some numbers somewhere from a Mobilenet model I was using for testing that I'll add in a bit, once I find / regenerate them. :)

orlmon01:
mobilenet model without the current patch:

> Setting intra_op_num_threads to 1
> Overriding dimension with name, N, to 1
> Overriding dimension with name, T, to 1000
> Overriding dimension with name, cache_T_attn, to 32
> Overriding dimension with name, right_context, to 5
> Session creation time cost: 0.020627 s
> First inference time cost: 10 ms
> Total inference time cost: 1.33257 s
> Total inference requests: 200
> Average inference time cost total: 6.662851 ms
> Total inference run time: 1.33266 s
> Number of inferences per second: 150.075 
> Avg CPU usage: 16 %
> Peak working set size: 85429583872 bytes
> Avg CPU usage:16
> Peak working set size:85429583872
> Runs:200
> Min Latency: 0.006193 s
> Max Latency: 0.007625 s
> P50 Latency: 0.00666992 s
> P90 Latency: 0.00686983 s
> P95 Latency: 0.00694425 s
> P99 Latency: 0.00733196 s
> P999 Latency: 0.007625 s

Same model with changes:

> Setting intra_op_num_threads to 1
> Overriding dimension with name, N, to 1
> Overriding dimension with name, T, to 1000
> Overriding dimension with name, cache_T_attn, to 32
> Overriding dimension with name, right_context, to 5
> Session creation time cost: 0.0217724 s
> First inference time cost: 7 ms
> Total inference time cost: 1.12897 s
> Total inference requests: 200
> Average inference time cost total: 5.644857 ms
> Total inference run time: 1.12905 s
> Number of inferences per second: 177.14 
> Avg CPU usage: 16 %
> Peak working set size: 80362864640 bytes
> Avg CPU usage:16
> Peak working set size:80362864640
> Runs:200
> Min Latency: 0.00527438 s
> Max Latency: 0.006706 s
> P50 Latency: 0.00566529 s
> P90 Latency: 0.00579958 s
> P95 Latency: 0.00585429 s
> P99 Latency: 0.00639058 s
> P999 Latency: 0.006706 s

onnxruntime/core/framework/kernel_type_str_resolver.cc
```diff
@@ -36,6 +36,18 @@ static OpKernelTypeStrMap::const_iterator LookUpOpId(const OpIdentifier& op_id,
     }
   }
 
+#ifdef USE_KLEIDIAI
```
edgchen1: why is this special case required?

onnxruntime/test/framework/ort_model_only_test.cc
```diff
@@ -17,16 +17,42 @@
...
+    return path;
+  }
+
+  std::filesystem::path workspace_candidate = std::filesystem::current_path() / path;
+  if (std::filesystem::exists(workspace_candidate)) {
+    return workspace_candidate;
+  }
+
+  static const std::filesystem::path kSourceTestRoot =
+      std::filesystem::path{WIDEN(__FILE__)}.parent_path().parent_path();
```
edgchen1: does `WIDEN` here only work on Windows?

onnxruntime/test/framework/ort_model_only_test.cc
```diff
@@ -17,16 +17,42 @@
...
 #include "gtest/gtest.h"
+#define WIDEN2(x) L##x
+#define WIDEN(x) WIDEN2(x)
 
 using namespace ONNX_NAMESPACE;
 
 namespace onnxruntime {
 namespace test {
+namespace {
+std::filesystem::path ResolveTestPath(const std::filesystem::path& path) {
```
edgchen1: the existing setup is to use test data files with paths relative to the test binary. why are we doing something differently here?

orlmon01: We were running into issues with it on CI (Both yours and ours) and doing this to fully resolve the path was something I added initially as a way to debug where it was looking for the file. But it also fixed the problem so I left it in. I could look to change to relative similar to the other tests if needed.

onnxruntime/contrib_ops/cpu/fused_conv.cc
```diff
@@ -26,5 +26,13 @@ ONNX_CPU_OPERATOR_TYPED_MS_KERNEL(
         .TypeConstraint("T", DataTypeImpl::GetTensorType<float>()),
     FusedConvFloat);
 
+ONNX_CPU_OPERATOR_TYPED_MS_KERNEL(
+    NhwcFusedConv,
+    1,
+    float,
+    KernelDefBuilder()
```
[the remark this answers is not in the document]
hariharans29: @orlmon01 - I think this may be a worthy suggestion ?

onnxruntime/core/graph/contrib_ops/nhwc_schema_defs.cc
```diff
@@ -403,7 +403,7 @@ Only has fp16 implementation as of 2023/04/15.
                                 .Input(2, "B", "", "T", OpSchema::Optional)
                                 .Input(3, "Z", "Tensor to be added to the output, must be the same shape and format as the output tensor.", "T", OpSchema::Optional)
                                 .Output(0, "Y", "", "T")
-                                .TypeConstraint("T", {"tensor(float16)"}, "Constrain input and output types to float tensors")
+                                .TypeConstraint("T", {"tensor(float16)", "tensor(float)"}, "Constrain input and output types to float tensors")
```
hariharans29, line 420: Nit: The comment above days that it only supports float16 - which isn;t true anymore

hariharans29, line 420: As part of this PR, it would nice to update the descriptions for each of the op's inputs - what is the expected data layout for each of its inputs, what is the data layout of the output, etc.

onnxruntime/core/optimizer/nhwc_transformer.cc
```diff
@@ -21,6 +24,72 @@ namespace onnxruntime {
 
 using namespace layout_transformation;
 
+#ifdef USE_KLEIDIAI
+bool KleidiFp32NhwcFilter(const onnx_transpose_optimization::api::GraphRef& graph,
```
hariharans29: A brief description of what this routine is trying to achieve might help

onnxruntime/core/framework/kernel_type_str_resolver.cc
orlmon01: NhwcFusedConv may not exist in the saved kernel-type map, even though the runtime graph ends up containing it

KernelTypeStrResolver is serialized from op schemas seen at save/load time. Later, NHWC layout transformation (KleidiAI path) can rewrite Conv/FusedConv nodes to NhwcFusedConv. If the resolver map has no NhwcFusedConv entry, type-constraint resolution fails during kernel matching

This is a fallback to reuse theequivalent from FusedConv in its place

onnxruntime/core/optimizer/nhwc_transformer.cc
```diff
@@ -21,6 +24,72 @@ namespace onnxruntime {
...
+    const auto& pads = pads_attr.value();
+    if (pads.size() != 4 || pads[0] != pads[2] || pads[1] != pads[3]) {
+      return false;
+    }
+  }
+
+  const auto inputs = node.Inputs();
+  if (inputs.size() > 3 && !inputs[3].empty()) {
+    return false;
+  }
```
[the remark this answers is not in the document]
orlmon01: Loosening this restriction led to other errors and was not considered necessary for now. A future patch which implements Depthwise for this feature will make this restriction moot

onnxruntime/test/framework/ort_model_only_test.cc
orlmon01: It should work on all systems. It was implemented locally on a Mac and tested on linux via our CI. So it seems to be properly portable

onnxruntime/core/graph/contrib_ops/nhwc_schema_defs.cc
damdoo01-arm: @orlmon01 has applied as above

damdoo01-arm: Descriptions added

onnxruntime/core/optimizer/nhwc_transformer.cc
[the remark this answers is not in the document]
damdoo01-arm: I've made an attempt to remove this, no issues so far, will reinstate if needed

onnxruntime/core/providers/cpu/nn/conv.cc
```diff
@@ -160,11 +200,10 @@ Status Conv<float>::Compute(OpKernelContext* context) const {
   const Tensor* B = num_inputs >= 3 ? context->Input<Tensor>(2) : nullptr;
   const Tensor* Sum = num_inputs >= 4 ? context->Input<Tensor>(3) : nullptr;
   const int64_t N = X->Shape()[0];
-  const int64_t C = X->Shape()[1];
+  const int64_t C = X->Shape()[channels_last_ ? 3 : 1];
   const int64_t M = W->Shape()[0];
-  ORT_RETURN_IF_ERROR(conv_attrs_.ValidateInputShape(X, W));
+  ORT_RETURN_IF_ERROR(conv_attrs_.ValidateInputShape(X->Shape(), W->Shape(), channels_last_));
```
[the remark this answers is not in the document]
damdoo01-arm: Moved validation condition

onnxruntime/test/framework/ort_model_only_test.cc
damdoo01-arm: I've modified this for robustness, but if there's a need to remain consistent here, we can have a further discussion on the specific issue with path resolution and how to address.

onnxruntime/test/framework/ort_model_only_test.cc
damdoo01-arm: Replaced with ORT_TSTR_ON_MACRO

onnxruntime/test/framework/ort_model_only_test.cc
```diff
@@ -213,17 +242,36 @@ static void CompareSessionMetadata(const InferenceSessionWrapper& session_object
...
+    so.session_logid = "SerializeToOrtFormat";
+    so.optimized_model_filepath = ort_path.native();
+    so.graph_optimization_level = optimization_level;
+    ASSERT_STATUS_OK(so.config_options.AddConfigEntry(kOrtSessionOptionsConfigSaveModelFormat, "ORT"));
+    InferenceSessionWrapper session_object{so, GetEnvironment()};
+    ASSERT_STATUS_OK(session_object.Load(orig_path.native()));
+    ASSERT_STATUS_OK(session_object.Initialize());
+    return;
+  }
+
```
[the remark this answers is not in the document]
damdoo01-arm: Removed

onnxruntime/test/optimizer/nhwc_transformer_test.cc
```diff
@@ -224,6 +224,31 @@ TEST(NhwcTransformerTests, ConvGlobalAveragePool) {
                     TransformerLevel::Level3);
 }
 
+TEST(NhwcTransformerTests, ConvDepthwiseFloat) {
```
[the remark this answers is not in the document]
damdoo01-arm: Good point — this test is verifying that depthwise convolution is skipped by the NHWC transformer, not transformed. I renamed it to ConvDepthwiseFloat_SkipNhwc to make the intent explicit.

onnxruntime/core/optimizer/nhwc_transformer.cc
```diff
@@ -108,14 +145,61 @@ NhwcTransformer::NhwcTransformer(AllocatorPtr cpu_allocator,
...
+      };
+
       conv_table_.emplace(
           OpIdInfo("Conv", kOnnxDomain, api::DataType::FLOAT16),
-          OpTransformInfo{nhwc_conv_fp16.op_type_, nhwc_conv_fp16.domain_, nhwc_conv_fp16.version_, false});
+          OpTransformInfo{nhwc_conv_fp16.op_type_, nhwc_conv_fp16.domain_, nhwc_conv_fp16.version_, false, false, filter});
       conv_table_.emplace(
           OpIdInfo("FusedConv", kMSDomain, api::DataType::FLOAT16),
-          OpTransformInfo{nhwc_conv_fp16.op_type_, nhwc_conv_fp16.domain_, nhwc_conv_fp16.version_, false});
+          OpTransformInfo{nhwc_conv_fp16.op_type_, nhwc_conv_fp16.domain_, nhwc_conv_fp16.version_, false, false, filter});
```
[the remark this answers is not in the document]
orlmon01: Fix pushed

onnxruntime/core/framework/kernel_type_str_resolver_utils.cc
```diff
@@ -45,6 +46,18 @@ Status LoadKernelTypeStrResolverFromBuffer(KernelTypeStrResolver& kernel_type_st
...
 
 Status AddLayoutTransformationRequiredOpsToKernelTypeStrResolver(KernelTypeStrResolver& kernel_type_str_resolver) {
+#if !defined(ORT_MINIMAL_BUILD)
+  const auto required_op_ids = GetLayoutTransformationRequiredOpIdentifiers();
+  const auto schema_registry = SchemaRegistryManager{};
+  for (const auto& op_id : required_op_ids) {
+    const auto* op_schema = schema_registry.GetSchema(std::string{op_id.op_type}, op_id.since_version,
+                                                      std::string{op_id.domain});
+    ORT_RETURN_IF(op_schema == nullptr, "Failed to get op schema.");
+    ORT_RETURN_IF_ERROR(kernel_type_str_resolver.RegisterOpSchema(*op_schema));
```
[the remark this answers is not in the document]
orlmon01: Fixed

onnxruntime/core/mlas/lib/convolve.cpp
```diff
@@ -1245,6 +1245,85 @@ Return Value:
 // Chance of arithmetic overflow could be reduced
 #pragma warning(disable : 26451)
 #endif
+
+namespace {
+
+#if defined(USE_KLEIDIAI) && defined(__aarch64__)
```
hariharans29: Is the MlasBackendKernelSelector's use_kleidiai flag read anywere ? That was introduced sometime AFTER this PR was originally proposed. Can you please include logic to read and honor that flag ? That is a way for the user to opt-out of KleidiAI should they want to.

hariharans29: Is `MLAS_TARGET_ARM64` more inclusive to include MSVC/Windows platforms instead of `__aarch64__` ?

onnxruntime/test/mlas/bench/bench_transcendental.cpp
```diff
@@ -19,7 +19,9 @@ constexpr float kSiluMaxValue = 20.0f;
 constexpr float kGeluMinValue = -10.0f;
 constexpr float kGeluMaxValue = 10.0f;
 constexpr float kInvSqrt2 = 0.7071067811865475244f;
+#if defined(MLAS_TARGET_AMD64)
```
hariharans29, line 22: Why does this only need to be only AMD64 ?

onnxruntime/test/optimizer/conv_add_act_test.cc
```diff
@@ -30,9 +30,10 @@ void TestConvPath(const std::vector<int64_t>& input_shape, const std::vector<int
 
   auto check_graph = [&](InferenceSessionWrapper& session) {
     auto op_to_count = CountOpsInGraph(session.GetGraph());
-    EXPECT_EQ(op_to_count["com.microsoft.FusedConv"], 1);
+    const auto fused_count = op_to_count["com.microsoft.FusedConv"] + op_to_count["com.microsoft.NhwcFusedConv"];
```
hariharans29: Can the check be more specific in its check - check to see if we see NCHW FusedConv or the NHWC FusedConv as appropriate ?

onnxruntime/test/optimizer/transpose_optimizer_test.cc
```diff
@@ -4604,6 +4604,64 @@ TEST(TransposeOptimizerTests, QnnTransposeReshape) {
   }
 }
 
+TEST(TransposeOptimizerTests, LayoutTransformDoesNotRetargetNhwcFusedConv) {
```
hariharans29, line 4609: A brief description about what this test is checking would be nice

onnxruntime/core/providers/cpu/nn/conv.cc
```diff
@@ -159,12 +207,11 @@ Status Conv<float>::Compute(OpKernelContext* context) const {
   const Tensor* W = context->Input<Tensor>(1);
   const Tensor* B = num_inputs >= 3 ? context->Input<Tensor>(2) : nullptr;
   const Tensor* Sum = num_inputs >= 4 ? context->Input<Tensor>(3) : nullptr;
+  ORT_RETURN_IF_ERROR(conv_attrs_.ValidateInputShape(X->Shape(), W->Shape(), channels_last_));
   const int64_t N = X->Shape()[0];
-  const int64_t C = X->Shape()[1];
+  const int64_t C = X->Shape()[channels_last_ ? 3 : 1];
```
hariharans29: I think per-spec, input `X` can be N-dimensional - so assuming it is a 2-D conv seems risky (by assuming the channel dimension is 3 when channels_last_ is true) ?

onnxruntime/core/providers/cpu/nn/conv.cc
```diff
@@ -198,20 +247,69 @@ Status Conv<float>::Compute(OpKernelContext* context) const {
   auto Xdata = X->DataAsSpan<float>();
   const auto* Bdata = B != nullptr ? B->Data<float>() : nullptr;
   auto Ydata = Y->MutableDataAsSpan<float>();
-  // Check for the optional Conv/Sum fusion.
+  const size_t kernel_rank = kernel_shape.size();
+  concurrency::ThreadPool* thread_pool = context->GetOperatorThreadPool();
+
+  if (channels_last_) {
+    ORT_RETURN_IF_NOT(kernel_rank == 2, "Conv with channels_last layout currently supports 2D kernels.");
```
hariharans29: I see that we are returning if we aren't processing a 2-D conv and channels_last_ is true. Can we hoist this above to somewhere near the entry of the routine or is that tricky to do ?

cmake/CMakeLists.txt
```diff
@@ -635,6 +635,7 @@ else()
   check_cxx_compiler_flag(-Wcatch-value HAS_CATCH_VALUE)
   check_cxx_compiler_flag(-Wclass-memaccess HAS_CLASS_MEMACCESS)
   check_cxx_compiler_flag(-Wcharacter-conversion HAS_CHARACTER_CONVERSION)
+  check_cxx_compiler_flag(-Wno-error=character-conversion HAS_NO_ERROR_CHARACTER_CONVERSION)
```
hariharans29, line 641: Why are these compiler flag checks needed ?

onnxruntime/core/graph/contrib_ops/nhwc_schema_defs.cc
hariharans29: Was this change pushed ? Because I still see the comment as: Only has fp16 implementation as of 2023/04/15

hariharans29: Ditto - I don't see the descriptions yet.

onnxruntime/test/mlas/bench/bench_sconv.cpp
```diff
@@ -217,6 +218,7 @@ void SCONV_NCHW_THREADED(benchmark::State& state, const char* /*dummy*/) {
                   static_cast<size_t>(output_channels_per_group),
                   &activation,
                   &WorkingBufferSize,
+                  false,
```
hariharans29, line 221: Nit: Can come in another PR - should there be similar benchmarks to show the KleidiAI NHWC kernel does better than the default NCHW kernel ?

onnxruntime/core/providers/cpu/nn/conv.cc
```diff
@@ -230,22 +328,82 @@ Status Conv<float>::Compute(OpKernelContext* context) const {
...
+             input_compute,
              W->Data<float>(),
              Bdata,
-             static_cast<float*>(working_buffer.get()),
-             Ydata.data(),
+             working_data,
+             output_compute,
              thread_pool);
+
+    if (wants_channels_last && !nhwc_fastpath) {
```
hariharans29, line 439: So this is currently for the cases where we are dealing with an NHWC (fused) Conv - but KleidiAI is unable to handle the shapes and we drop down to the default MLAS path ?

onnxruntime/core/optimizer/nchwc_transformer.cc
```diff
@@ -325,6 +325,14 @@ void NchwcTransformerImpl::TransformConv(Node& node) {
   auto& input_defs = node.MutableInputDefs();
   auto& output_defs = node.MutableOutputDefs();
 
+  // The internal NCHWc Conv kernel can consume an optional fused Sum input, but it expects
+  // that tensor to already be in NCHWc layout. This transform only legalizes the main Conv
+  // input and static weights/bias, so a pre-existing FusedConv(X, W, B, Sum) would feed a
+  // plain NCHW tensor into the NCHWc kernel and produce incorrect results.
+  if (node.OpType() == "FusedConv" && input_defs.size() >= 4 && input_defs[3] != nullptr && input_defs[3]->Exists()) {
+    return;
```
hariharans29: Nice catch - could you also please add a test in the NCHWc optimizer suite for this ?

onnxruntime/core/providers/cpu/nn/conv.cc
```diff
@@ -24,6 +26,52 @@
 namespace onnxruntime {
 using ConvPadVector = ConvAttributes::ConvPadVector;
 
+namespace {
+
+template <typename T>
+void ConvertNHWCToNCHW(const T* src, T* dst,
+                       int64_t n, int64_t c, int64_t h, int64_t w) {
+  const size_t n_count = narrow<size_t>(n);
```
hariharans29, line 39: Optional (can be future work): Can the NCHW <-> NHWC work be parallelized at all if the threadpool is available for use ?

Optional item 2: Is there scope for vectorized reads/writes here and if so, can these be standalone MLAS APIs backed by NEON/SVE reads/writes on ARM platforms ?

onnxruntime/test/optimizer/fuse_initializers_transformer_test.cc
```diff
@@ -490,11 +493,11 @@ TEST(TransformerTest, FuseFp16InitializersWithGraphOutputs) {
   so.graph_optimization_level = TransformerLevel::MaxLevel;
   // Create session and check graph before / after initiation
   InferenceSessionWrapper session{so, GetEnvironment()};
-  // Disabling ConstantFolding optimizer as it will remove the Cast node
+  // Disabling ConstantFolding and NhwcTransformer optimizer as it will remove the Cast node
```
hariharans29: Does NhwcTransformer also have the same effect as ConstantFolding - the reason seems to indicate that ?

onnxruntime/test/optimizer/fuse_initializers_transformer_test.cc
```diff
@@ -363,6 +363,7 @@ TEST(TransformerTest, FuseFp16InitializersWithFp32Node_with_graph_optimizations_
 
   // Create session and check graph before / after initiation
   InferenceSessionWrapper session{so, GetEnvironment()};
+  ASSERT_STATUS_OK(session.FilterEnabledOptimizers({"NhwcTransformer"}));
```
hariharans29, line 368: Please document why we are filtering these optimizers - and should they be done universally like how it is being currently done ?

onnxruntime/test/optimizer/nchwc_optimizer_test.cc
```diff
@@ -202,6 +202,7 @@ void NchwcOptimizerTester(const std::function<void(NchwcTestHelper& helper)>& bu
     session_options.session_logid = "NchwcOptimizerTests";
     InferenceSessionWrapper session{session_options, GetEnvironment()};
     ASSERT_STATUS_OK(session.Load(model_data.data(), static_cast<int>(model_data.size())));
+    ASSERT_STATUS_OK(session.FilterEnabledOptimizers({"NhwcTransformer"}));
```
hariharans29, line 205: Ditto - Please document why

onnxruntime/core/optimizer/nchwc_transformer.cc
hariharans29: Wait - but what if the "sum" into the FusedConv is an NCHWc output from another `Conv` node or another NCHWc compliant node. Shouldn't we check for that first before blocking the data layout change ?

onnxruntime/core/providers/cpu/nn/conv.cc
```diff
@@ -24,6 +26,52 @@
 namespace onnxruntime {
 using ConvPadVector = ConvAttributes::ConvPadVector;
 
+namespace {
+
+template <typename T>
+void ConvertNHWCToNCHW(const T* src, T* dst,
```
hariharans29, line 36: (Optional) Aren't these just transposes ? Can we invoke an optimed MLAS transpose routine for these instead of re-implemting it again ?

onnxruntime/core/framework/kernel_type_str_resolver.cc
edgchen1: if this is a new op related to layout transformation, would it be better to include it in `kernel_type_str_resolver_utils::AddLayoutTransformationRequiredOpsToKernelTypeStrResolver` instead of modifying the op lookup with this special case?

onnxruntime/core/framework/kernel_type_str_resolver_utils.cc
```diff
@@ -45,6 +46,18 @@ Status LoadKernelTypeStrResolverFromBuffer(KernelTypeStrResolver& kernel_type_st
 }
 
 Status AddLayoutTransformationRequiredOpsToKernelTypeStrResolver(KernelTypeStrResolver& kernel_type_str_resolver) {
+#if !defined(ORT_MINIMAL_BUILD)
```
edgchen1: why do we need this new `!defined(ORT_MINIMAL_BUILD)` path? ideally it would work for minimal builds too, and the less ifdef'd code we have the easier it is to maintain. should we update the byte array instead? see the comment below for how to regenerate it.

https://github.com/microsoft/onnxruntime/blob/0cf32ad89c9cf515b0c6d1ddd2447f0f4f1d63db/onnxruntime/core/framework/kernel_type_str_resolver_utils.cc#L50-L51

onnxruntime/test/framework/ort_model_only_test.cc
edgchen1: consistency would be nice. can you elaborate on what issues you were encountering?

onnxruntime/core/mlas/lib/convolve.cpp
orlmon01: It is not, I will look to add it

onnxruntime/core/optimizer/conv_add_act_fusion.cc
```diff
@@ -211,7 +211,22 @@ class FuseConvAddActivationAction : public ReplaceWithNew {
...
+    if (channels_last) {
+      return "NhwcFusedConv";
+    }
+
+    // Without channels_last, convert Conv to FusedConv, and leave other op types unchanged.
+    if (op_type == "Conv") {
+      return "FusedConv";
+    }
+
+    return op_type;
```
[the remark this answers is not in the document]
orlmon01: This isn't a valid concern. As the possible options coming in here are: 
  - Conv becomes FusedConv when channels_last is false.
  - any channels_last target becomes NhwcFusedConv.
  - NhwcFusedConv stays NhwcFusedConv, which is already the correct fused op.

onnxruntime/test/mlas/bench/bench_transcendental.cpp
orlmon01: - kFusedBytesPerElement is only used inside the AMD64-only dispatch checks in onnxruntime/test/mlas/bench/bench_transcendental.cpp:33 and onnxruntime/test/mlas/bench/bench_transcendental.cpp:45.
  - On non-AMD64 builds, those #if defined(MLAS_TARGET_AMD64) blocks disappear, so the constant becomes unused.
  - With the warning settings used on macOS, that can trip an unused-const warning and fail the build under -Werror.

  So the guard was added to keep non-AMD64/macOS builds clean. It does not mean the benchmark is AMD64-only. It just means the avx512_fused byte-count constant is only relevant on AMD64 with the current MLAS implementation.

onnxruntime/test/optimizer/conv_add_act_test.cc
orlmon01: I have made the check more specific. I am not checking for “NCHW FusedConv” because NCHWc uses com.microsoft.nchwc.Conv, and that optimizer is disabled in this test

onnxruntime/core/providers/cpu/nn/conv.cc
orlmon01: Done

cmake/CMakeLists.txt
orlmon01: I was getting a lot of errors for warnings on Apple Clang and this was a fix for that specific error without changing behavior or code for everything when I did it.

onnxruntime/core/optimizer/nchwc_transformer.cc
orlmon01: This is a good point. The check here is too safe. I'm adding a narrower check and a test for it.

hariharans29:
Will take another pass at this PR early next week.

onnxruntime/test/framework/ort_model_only_test.cc
orlmon01: Changed it so everything uses the old relative method. The new method is more robust to testdata not always being where expected but consistency is better I think

onnxruntime/core/providers/cpu/nn/conv.cc
orlmon01: They are, they're only intended for fallback in cases where it has started down the NHWC pipeline but finds something (Usually around fused sum) that means it won't work. I'll add a patch to use the MLAS transpose for fallback instead as that should be more efficient.

onnxruntime/core/providers/cpu/nn/conv.cc
orlmon01: Yes, this is for falling back in cases where model expects NHWC behavior/layout but MLAS NHWC kernel cannot handle this case. In which case it does the following to drop down to the default path:
      - allocate a temporary NCHW input/output buffer at onnxruntime/core/providers/cpu/nn/conv.cc:333 and onnxruntime/core/providers/cpu/nn/conv.cc:344
      - transpose input NHWC -> NCHW at onnxruntime/core/providers/cpu/nn/conv.cc:354
      - run MlasConv in the existing NCHW path at onnxruntime/core/providers/cpu/nn/conv.cc:359
      - if needed, transpose Sum the same way and apply add/activation manually at onnxruntime/core/providers/cpu/nn/conv.cc:367
      - transpose output back NCHW -> NHWC at onnxruntime/core/providers/cpu/nn/conv.cc:390

onnxruntime/core/providers/cpu/nn/conv.cc
orlmon01: This has been changed to use the MLAS transpose routine for fallback instead. This should also utilise threadpool

onnxruntime/test/optimizer/fuse_initializers_transformer_test.cc
orlmon01: This looks to have been a holdover from an older version of this change when fused ops could remove the cast node. After doing some digging it seems this is no longer needed so I have removed it from the filter

onnxruntime/test/optimizer/fuse_initializers_transformer_test.cc
orlmon01: This optimizer is hardware / kernel specific and can modify the behaviour of this test on some devices. So a blanket filter in these test cases was the easiest solution. The tests are for FuseInitializersTransformer/NCHWC behavior and if this optimizer is applied after it can change the node counts that the asserts are looking for. Added comments to clarify within the code.

onnxruntime/core/framework/kernel_type_str_resolver_utils.cc
orlmon01: NhwcFusedConv is already in kLayoutTransformationPotentiallyAddedOps at onnxruntime/core/optimizer/layout_transformation/layout_transformation_potentially_added_ops.h:73.

The new special case in onnxruntime/core/framework/kernel_type_str_resolver.cc:39 is a fallback for when the resolver map does not contain com.microsoft.NhwcFusedConv explicitly, but does contain com.microsoft.FusedConv or onnx.Conv.

This also applies to the minimal build as adding it to the minimal build and regenerating the byte code means it has to be always available. Or perhaps more accurately it can always handle finding com.microsoft.NhwcFusedConv
I'm looking at creating a patch for this but I have a hunch I might break stuff so fingers crossed.

onnxruntime/core/mlas/lib/convolve.cpp
```diff
@@ -1324,6 +1324,85 @@ Return Value:
...
+
+namespace {
+
+#if defined(USE_KLEIDIAI) && defined(MLAS_TARGET_ARM64)
+static constexpr size_t ComputeChannelsLastDilatedKernelSize(size_t dilation, size_t kernel) {
+    return (dilation * kernel) - (dilation - 1);
+}
+
+static constexpr size_t ComputeChannelsLastConvOutSize(size_t input, size_t kernel, size_t padding, size_t stride) {
+    if (stride > 0 && (input + 2 * padding) >= kernel) {
```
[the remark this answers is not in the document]
orlmon01: These checks already happen earlier and are not needed at this stage.

onnxruntime/core/framework/kernel_type_str_resolver_utils.cc
orlmon01: Pushed a patch. It's all green locally so hopefully it stays green on the next CI run :)

onnxruntime/test/optimizer/nchwc_optimizer_test.cc
orlmon01: Same as above. Added comments to the code

onnxruntime/core/graph/contrib_ops/nhwc_schema_defs.cc
orlmon01: Done

orlmon01: Done

onnxruntime/core/mlas/lib/convolve.cpp
orlmon01: Done

orlmon01: Done

onnxruntime/test/optimizer/transpose_optimizer_test.cc
orlmon01: Done

onnxruntime/test/mlas/bench/bench_sconv.cpp
orlmon01: Yes, they would be good to add. I'll look to add them in the next patch with depthwise support.

onnxruntime/core/optimizer/nhwc_transformer.cc
```diff
@@ -21,6 +28,215 @@ namespace onnxruntime {
...
+    return false;
+  }
+
+  for (size_t i = 0; i < out.size(); ++i) {
+    if (values[i] < 0) {
+      return false;
+    }
+
+    out[i] = narrow<size_t>(values[i]);
+  }
```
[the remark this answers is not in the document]
orlmon01: A valid concern. Pushed a fix

onnxruntime/core/providers/cpu/nn/conv.cc
orlmon01: I can't put it all the way to the top as it relies on calculating the kernel_rank but I have moved it up a little higher in the code so it can return sooner

onnxruntime/core/framework/kernel_type_str_resolver_utils.cc
edgchen1: cool, thanks for updating the byte array. I'd expect that to be necessary after adding the new entry to `kLayoutTransformationPotentiallyAddedOps`.

so could we actually do without this newly added ifdef path?

onnxruntime/core/framework/kernel_type_str_resolver.cc
```diff
@@ -35,7 +35,6 @@ static OpKernelTypeStrMap::const_iterator LookUpOpId(const OpIdentifier& op_id,
       }
     }
   }
-
```
edgchen1, line 38: nit: please revert the whitespace-only change

onnxruntime/core/framework/kernel_type_str_resolver_utils.cc
orlmon01: I think so. I'm going to push a patch for it and see what happens :)

onnxruntime/core/mlas/lib/kleidiai/convolve_kleidiai.cpp
```diff
@@ -341,7 +344,11 @@ static std::shared_ptr<std::byte[]> RhsPackWeightsBiasSme(const size_t co, const
     // Cache of prepacked kai rhs weights and biases. thread_local to prevent interference from parallel sessions.
     thread_local std::unordered_map<RhsCacheKey, std::shared_ptr<std::byte[]>> rhs_cache;
 
-    RhsCacheKey key = { co, ci, kh, kw, dilationh, dilationw, HashWeights(weights) };
+    // The packed RHS buffer includes both weights and bias, so both must participate in the cache key.
+    const size_t weights_hash = HashTensorPrefix(weights, co * ci * kh * kw);
+    const bool has_bias = bias != nullptr;
+    const size_t bias_hash = has_bias ? HashTensorPrefix(bias, co) : 0;
+    RhsCacheKey key = { co, ci, kh, kw, dilationh, dilationw, has_bias, weights_hash, bias_hash };
```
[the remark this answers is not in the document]
orlmon01: This looks like a valid concern, I'll look to add a cleaner way of doing the hashes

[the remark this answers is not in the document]
orlmon01: Pushed a patch to fix this.

onnxruntime/core/framework/kernel_type_str_resolver_utils.cc
orlmon01: Patch has been pushed. It's all green locally so fingers crossed it's good on your end too :)

onnxruntime/core/optimizer/conv_add_act_fusion.cc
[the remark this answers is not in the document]
orlmon01: Same as the last one

onnxruntime/core/providers/cpu/nn/conv.cc
```diff
@@ -198,20 +239,64 @@ Status Conv<float>::Compute(OpKernelContext* context) const {
...
-    // If the output was not allocated inplace with the sum tensor, then copy here.
-    auto sum_data = Sum->DataAsSpan<float>();
-    if (Ydata.data() != sum_data.data()) {
-      gsl::copy(sum_data, Ydata);
+    if (manual_sum) {
+      auto sum_span = Sum->DataAsSpan<float>();
+      sum_manual_buffer.assign(sum_span.begin(), sum_span.end());
+      sum_manual_data = sum_manual_buffer.data();
+    } else {
+      auto sum_span = Sum->DataAsSpan<float>();
```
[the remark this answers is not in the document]
orlmon01: This is a valid issue, testing a fix locally.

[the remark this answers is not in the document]
orlmon01: Pushed a fix

hariharans29:
Can you please push the doc change too ? I ll kick off CI post that

orlmon01:
> Can you please push the doc change too ? I ll kick off CI post that

Done. I hadn't seen your message when I was pushing the Co-Pilot fix. :)

onnxruntime/core/optimizer/nchwc_transformer.cc
```diff
@@ -490,6 +502,11 @@ void NchwcTransformerImpl::TransformConv(Node& node) {
     nchwc_node.MutableInputDefs()[2] = nchwc_conv_B_arg;
   }
 
+  if (nchwc_sum_input != nullptr) {
```
hariharans29, line 505: Nit: nchwc_sum_input can't be nullptr at this point as we would have bailed out before had it been nullptr (on line 335 or so) ?

onnxruntime/core/session/inference_session.cc
```diff
@@ -1052,6 +1052,14 @@ common::Status InferenceSession::SaveToOrtFormat(const std::filesystem::path& fi
     ORT_RETURN_IF_ERROR(kernel_type_str_resolver.RegisterOpSchema(*op_schema));
   }
 
+#if !defined(ORT_MINIMAL_BUILD) || defined(ORT_EXTENDED_MINIMAL_BUILD)
+  // Persist resolver entries for ops that layout transformation may add so newly saved ORT models do not
+  // rely on load-time aliasing to resolve their kernel type strings.
+  ORT_RETURN_IF_ERROR(
+      kernel_type_str_resolver_utils::AddLayoutTransformationRequiredOpsToKernelTypeStrResolver(
+          kernel_type_str_resolver));
+#endif
```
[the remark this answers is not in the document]
orlmon01: This doesn't look to be a valid concern. All of the call sites around this have the same check (onnxruntime/core/session/inference_session.cc:1055, onnxruntime/core/
  framework/kernel_type_str_resolver_utils.h:5 and onnxruntime/core/framework/kernel_type_str_resolver_utils.cc:4)  So AddLayoutTransformationRequiredOpsToKernelTypeStrResolver(...) is
  available for extended minimal builds, and that specific mismatch/build break should not happen.

onnxruntime/core/mlas/lib/kleidiai/convolve_kleidiai.cpp
```diff
@@ -341,7 +350,11 @@ static std::shared_ptr<std::byte[]> RhsPackWeightsBiasSme(const size_t co, const
     // Cache of prepacked kai rhs weights and biases. thread_local to prevent interference from parallel sessions.
     thread_local std::unordered_map<RhsCacheKey, std::shared_ptr<std::byte[]>> rhs_cache;
 
-    RhsCacheKey key = { co, ci, kh, kw, dilationh, dilationw, HashWeights(weights) };
+    // The packed RHS buffer includes both weights and bias, so both must participate in the cache key.
+    const size_t weights_hash = HashTensorContents(weights, co * ci * kh * kw);
+    const bool has_bias = bias != nullptr;
+    const size_t bias_hash = has_bias ? HashTensorContents(bias, co) : 0;
+    RhsCacheKey key = { co, ci, kh, kw, dilationh, dilationw, has_bias, weights_hash, bias_hash };
```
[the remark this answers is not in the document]
orlmon01: This does seem like a performance concern. I'll do some digging and see if there's a more performant way of doing this.

onnxruntime/core/optimizer/nchwc_transformer.cc
orlmon01: Done

onnxruntime/core/mlas/lib/kleidiai/convolve_kleidiai.cpp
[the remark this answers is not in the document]
orlmon01: This can be doing a couple of ways, I have a patch locally but it's failing on our CI with a segfault. It's also not small. Pulling it out to persist outside is quite a bit change it turns out. I won't have a fix before Tuesday if you think it's worth merging as is and I'll add in a faster way of doing it alongside Depthwise support and the benchmark tests in the next PR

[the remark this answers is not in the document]
hariharans29: Sure, I ll merge this as is and performance improvements + benchmarks + depthwise support can come later. Thank you for trying this.

hariharans29 (dismissed):

onnxruntime/core/session/inference_session.cc
```diff
@@ -1052,6 +1052,14 @@ common::Status InferenceSession::SaveToOrtFormat(const std::filesystem::path& fi
     ORT_RETURN_IF_ERROR(kernel_type_str_resolver.RegisterOpSchema(*op_schema));
   }
 
+#if !defined(ORT_MINIMAL_BUILD) || defined(ORT_EXTENDED_MINIMAL_BUILD)
+  // Persist resolver entries for ops that layout transformation may add so newly saved ORT models do not
+  // rely on load-time aliasing to resolve their kernel type strings.
+  ORT_RETURN_IF_ERROR(
+      kernel_type_str_resolver_utils::AddLayoutTransformationRequiredOpsToKernelTypeStrResolver(
+          kernel_type_str_resolver));
```
edgchen1: why do we need to add them to the saved ORT format model? they are already added at load time:

https://github.com/microsoft/onnxruntime/blob/8dd4a06811179a7c04aea0a97c65f45a3dc8c6af/onnxruntime/core/session/inference_session.cc#L1896-L1900

onnxruntime/test/framework/kernel_type_str_resolver_utils_test.cc
```diff
@@ -49,6 +60,86 @@ TEST(KernelTypeStrResolverUtilsTest, VerifyLayoutTransformationRequiredOpsResolv
 #endif  // !defined(DISABLE_CONTRIB_OPS)
 }
 
+#if !defined(ORT_MINIMAL_BUILD) && !defined(DISABLE_CONTRIB_OPS) && defined(USE_KLEIDIAI)
```
edgchen1: is this specific to `USE_KLEIDIAI`? NhwcFusedConv is added to the layout transformation ops even if `USE_KLEIDIAI` is not defined.

onnxruntime/test/framework/kernel_type_str_resolver_utils_test.cc
```diff
@@ -49,6 +60,86 @@ TEST(KernelTypeStrResolverUtilsTest, VerifyLayoutTransformationRequiredOpsResolv
...
+  auto& nhwc_fused_conv = graph.AddNode(
+      "nhwc_fused_conv", "NhwcFusedConv", "test node", {&x, &w}, {&y}, nullptr, kMSDomain);
+  nhwc_fused_conv.SetSinceVersion(1);
+
+  gsl::span<const ArgTypeAndIndex> resolved_args;
+  ASSERT_STATUS_OK(resolver.ResolveKernelTypeStr(nhwc_fused_conv, "T", resolved_args));
+  ASSERT_FALSE(resolved_args.empty());
+}
+
+TEST(KernelTypeStrResolverUtilsTest, SavedOrtModelResolverContainsNhwcFusedConv) {
```
edgchen1: I think the ORT format model itself doesn't need to contain the layout transformation ops if we add them at load time. that decision was originally made to avoid unnecessarily increasing the model size.

onnxruntime/core/mlas/lib/kleidiai/convolve_kleidiai.cpp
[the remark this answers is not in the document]
orlmon01: Pushed a patch for this. Since I got working I felt like it might as well go in. :)

onnxruntime/core/optimizer/nchwc_transformer.cc
orlmon01: This seems to have been the cause of segfaults and failing CI so I reverted it. Not entirely sure why but it seems like it's actually quite important :D

onnxruntime/core/session/inference_session.cc
orlmon01: It's not needed, removed.

onnxruntime/test/framework/kernel_type_str_resolver_utils_test.cc
orlmon01: It is no longer specific to USE_KLEIDIAI, removing USE_KLEIDIAI from the guard

onnxruntime/test/framework/kernel_type_str_resolver_utils_test.cc
orlmon01: True. Changing this test to instead check that nhwc_fused_conv is present at load time.

onnxruntime/test/framework/kernel_type_str_resolver_utils_test.cc
```diff
@@ -86,7 +86,7 @@ TEST(KernelTypeStrResolverUtilsTest, ResolveNhwcFusedConvFromLayoutTransformatio
   ASSERT_FALSE(resolved_args.empty());
 }
 
-TEST(KernelTypeStrResolverUtilsTest, SavedOrtModelResolverContainsNhwcFusedConv) {
+TEST(KernelTypeStrResolverUtilsTest, LoadedOrtModelResolverCanResolveNhwcFusedConv) {
```
edgchen1: thanks for updating inference_session.cc to not save all of the layout transformation required ops to the ORT format model.

now I'm wondering how much value this test adds, given we also have the `KernelTypeStrResolverUtilsTest.ResolveNhwcFusedConvFromLayoutTransformationRequiredOps` test. do we still need it?

orlmon01: Probably not. Although I'm always hesitant to remove tests as you never know what they might catch. If you like I can delete it and do a rebase for this PR?

edgchen1: that'd be good. I think we can do with just the `KernelTypeStrResolverUtilsTest.ResolveNhwcFusedConvFromLayoutTransformationRequiredOps` test.

orlmon01: Done and rebased :)

onnxruntime/test/framework/kernel_type_str_resolver_utils_test.cc
```diff
@@ -177,4 +117,6 @@ TEST(KernelTypeStrResolverUtilsTest, DISABLED_PrintExpectedLayoutTransformationR
 #endif  // defined(DISABLE_CONTRIB_OPS)
 }
 
+#endif  // !defined(ORT_MINIMAL_BUILD) && !defined(DISABLE_CONTRIB_OPS)
```
edgchen1: nit: could this `#endif` be moved right below the newly added `KernelTypeStrResolverUtilsTest.ResolveNhwcFusedConvFromLayoutTransformationRequiredOps` test?

the `KernelTypeStrResolverUtilsTest.DISABLED_PrintExpectedLayoutTransformationRequiredOpsResolverByteArray` test does have some handling for `DISABLE_CONTRIB_OPS` to give an error message. IMO, it's slightly better than the test not being compiled at all.

orlmon01: Done :)

orlmon01:
Fixed the lintrunner error. Sorry about that, it was green locally

orlmon01:
Regenerated the docs and pushed it as well as rebased. I'm really not sure what the cause of the  Windows GPU CUDA CI Pipeline Test Job is, it's really not clear from the logs what is failing.

hariharans29:
> Regenerated the docs and pushed it as well as rebased. I'm really not sure what the cause of the Windows GPU CUDA CI Pipeline Test Job is, it's really not clear from the logs what is failing.

Highly likely it was a flaky failure - but I ll keep an eye

edgchen1 (approved):

hariharans29 (approved):

PR #26835, opened 2025-12-19
[QNN-EP] Update gather op input tensor cast logic.

quic-muchhsu:
### Description

Gather op was referring to onnx graph when deciding whether to insert `Cast->int32` on indices. But input tensor is created by QNN and it could already casted into int32. Which cause mismatch and resulting adding redundant Cast. This PR changes Gather Op builder to refer to QNN tenser before adding int64->int32 cast.

### Motivation and Context

It solve QNN-Gather op not to insert redundant Cast->int32.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/gather_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/gather_op_builder.cc
@@ -160,7 +160,7 @@ static Status ProcessIndicesInput(QnnModelWrapper& qnn_model_wrapper,
   }
 
   // Insert QNN Cast op to convert dynamic indices from int64 to int32.
-  auto& input_tensorwrapper = qnn_model_wrapper.GetQnnTensorWrapper(indices_tensor_name);
+  const auto& input_tensorwrapper = qnn_model_wrapper.GetQnnTensorWrapper(indices_tensor_name);
 
   std::string indices_casted_name{indices_tensor_name};
   // Check QNN Tensor data type.


onnxruntime/core/providers/qnn/builder/opbuilder/gather_op_builder.cc
```diff
@@ -160,8 +160,11 @@ static Status ProcessIndicesInput(QnnModelWrapper& qnn_model_wrapper,
   }
 
   // Insert QNN Cast op to convert dynamic indices from int64 to int32.
+  auto& input_tensorwrapper = qnn_model_wrapper.GetQnnTensorWrapper(indices_tensor_name);
+
   std::string indices_casted_name{indices_tensor_name};
-  if (indices_info.qnn_data_type == QNN_DATATYPE_INT_64) {
+  // Check QNN Tensor data type.
+  if (input_tensorwrapper.GetTensorDataType() == QNN_DATATYPE_INT_64) {
```
edgchen1, line 167: out of curiosity, when would this be different from `indices_info.qnn_data_type`?

onnxruntime/core/providers/qnn/builder/opbuilder/gather_op_builder.cc
```diff
@@ -160,8 +160,11 @@ static Status ProcessIndicesInput(QnnModelWrapper& qnn_model_wrapper,
   }
 
   // Insert QNN Cast op to convert dynamic indices from int64 to int32.
+  auto& input_tensorwrapper = qnn_model_wrapper.GetQnnTensorWrapper(indices_tensor_name);
```
edgchen1: nit: just to be consistent with the const& return type
```suggestion
  const auto& input_tensorwrapper = qnn_model_wrapper.GetQnnTensorWrapper(indices_tensor_name);
```

onnxruntime/core/providers/qnn/builder/opbuilder/gather_op_builder.cc
quic-muchhsu: indices_info is get by `qnn_model_wrapper.GetTensorInfo(indices_input, indices_info)` which refers to the dtype from ONNX model and cast to the corresponding QNN datatype, and `input_tensorwrapper.GetTensorDataType()` returns the dtype of the QNN tensor. However, QnnTensorWrapper auto casts int64 to int32 which cause dtype mismatch when the input tensor is int64 in ONNX. Op builder will see int64 when using `indices_info` but it's actually already casted to int32.

onnxruntime/core/providers/qnn/builder/opbuilder/gather_op_builder.cc
quic-muchhsu: Yeah, thanks.

onnxruntime/core/providers/qnn/builder/opbuilder/gather_op_builder.cc
edgchen1: thanks for the explanation

edgchen1: would it be worth adding a comment with this explanation?

edgchen1 (approved):

onnxruntime/core/providers/qnn/builder/opbuilder/gather_op_builder.cc
quic-muchhsu: I think it's more like a bug fix for QNN, also the goal is to migrate all dtype checking for int64 cast to use `tensorwrapper.GetTensorDataType()` instead of from `GetTensorInfo`. So I don't think we need a comment for it.
If any, we should add comment on `GetTensorInfo` for other developers to aware that this function gives you ONNX tensor info not the QNN tensor info.

PR #26837, opened 2025-12-19
Do not run auditwheel on QNN whls

eserscor:
### Description
[Change 26634](https://github.com/microsoft/onnxruntime/pull/26634) added the qnn variant to the audit wheel process, but the qnn libraries do not pass audit repair due to a dependency on libc++. Continue to not audit qnn and rely on the correct dependencies being available for the user.

### Motivation and Context
Fixes build packaging pipeline

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -299,7 +299,7 @@ try:
 
             # qnn links libc++ rather than libstdc++ for its x86_64 dependencies which we currently do not
             # support for many_linux. This is not the case for other platforms.
-            qnn_run_audit = os.environ.get("AUDITWHEEL_ARCH", "x86_64") != "x86_64"
+            qnn_run_audit = environ.get("AUDITWHEEL_ARCH", "x86_64") != "x86_64"
 
             _bdist_wheel.run(self)
             if is_manylinux and not disable_auditwheel_repair and not is_openvino and (not is_qnn or qnn_run_audit):


setup.py
edgchen1, line 315: given that the only usage of `qnn_dependencies` is within this `if` branch, and now it won't run when `is_qnn` is true, do we still need `qnn_dependencies`?

setup.py
```diff
@@ -298,7 +298,9 @@ def run(self):
                 pass
 
             _bdist_wheel.run(self)
-            if is_manylinux and not disable_auditwheel_repair and not is_openvino:
+            # qnn links libc++ rather than libstdc++ for it's x86_64 dependencies which we currently do not
+            # support for many_linux.
+            if is_manylinux and not disable_auditwheel_repair and not is_openvino and not is_qnn:
```
qti-jkilpatrick: Is it possible to gate this on x86_64? QNN for ARM64 does link against libstdc++ so it's possible to target `manylinux_2_34`.

eserscor: I can, but it's not obvious to me that auditwheel is running for ARM64 builds. Auditwheel is running on linux and I only see packaging/builds for x86_64. Do we also need to add a stage to package ARM64?

qti-jkilpatrick: I'm not aware of ARM64 QNN wheels being built in your CI, but we use this script in ours to do so.

From my perspective, the original change that kicked off this effort is all about making progress towards Qualcomm taking over distribution in the new year. To that end, we'd like to keep things working in both CIs and the sources as close as possible until the moment when we have to diverge.

setup.py
eserscor: It sounds like we want to run the audit, just not on x86_64

setup.py
eserscor: Thanks! That was the context I was missing. I've pushed a change to run audit for qnn on non-x86_64 platforms.

adrianlizarraga (approved):

PR #17031, opened 2023-08-07
[aarch64] Add Sbgemm kernel to accelerate fp32 tensor matmul with bfloat16

snadampal:
### Description

This PR adds SbgemmKernel for aarch64. This includes Sbegmm kernel to implement matrix multiplication with bfloat16 SIMD instructions (bfmmla) and MatMul operator changes to invoke the Sbgemm kernel. To enable Sbgemm kernel, set the following session option:
"kOrtSessionOptionsGemmFastMathMode"

The PR also adds new test cases for mlas and ort.

### Motivation and Context

This is to improve MatMul performance on aarch64 platform.
I have run the below benchmarking script (bert , roberta and gpt2 model inference) on AWS Graviton3 based c7g.4xl instance and observed 1.2x -1.76x performance improvement compared to sgemm (fp32) kernel performance.

```
cd onnxruntime/python/tools/transformers
python3 benchmark.py
```
And the unit test precision results are matching to sgemm kernel results.
`./build.sh --config RelWithDebInfo --build_shared_lib --parallel --compile_no_warning_as_error --skip_submodule_sync `

PR #26838, opened 2025-12-19
Introducing BF16 Pointwise NCHWc Convolution for Arm64

Rohanjames1997:
## Description

This PR adds a BF16 (bfloat16) pointwise convolution kernel for ARM64 NCHWc format, leveraging the existing SBGEMM infrastructure. When the `mlas.enable_gemm_fastmath_arm64_bfloat16` session option is enabled on supported ARM64 Linux hardware, Pointwise Conv is rerouted to use this BF16 implementation. This is an opt-in feature, similar to how BF16 matmul is opt-in.

Added a bool ZeroMode field to `MLAS_SBGEMM_DATA_PARAMS` (default `true` for backward compatibility) to enable per-batch control over output accumulation. This mirrors the beta parameter in FP32's `MlasGemmBatch` and is required for Pointwise convolutions with >128 input channels, where multiple GEMM calls must accumulate into the same output buffer.

## Motivation and Context

The existing `mlas.enable_gemm_fastmath_arm64_bfloat16` session option accelerates MatMul operations on ARM64 processors with BF16 support, but convolution operations did not benefit from this optimization. Pointwise convolutions (1x1 kernels) are essentially batched matrix multiplications.

This change extends the BF16 fastmath optimization to pointwise NCHWc convolutions, reusing the same session option. The implementation mirrors the FP32 pointwise kernel structure while delegating the actual computation to SBGEMM, ensuring correctness and maintainability.

## Performance improvement
Measured a 15-20% gain on Mobilenet inference on an AWS Graviton4 instance. 

Before (FP32)
```
/build/Linux/Release/onnxruntime_perf_test -C "mlas.enable_gemm_fastmath_arm64_bfloat16|0" -x 32 -I -m times -r 2000 ~/scripts/mobilenet.onnx

Number of inferences per second: 559.154
```

After (BF16)
```
./build/Linux/Release/onnxruntime_perf_test -C "mlas.enable_gemm_fastmath_arm64_bfloat16|1" -x 32 -I -m times -r 2000 ~/scripts/mobilenet.onnx

Number of inferences per second: 651.221

```


Rohanjames1997:
@hariharans29 another PR that's up your alley. 

Can you request a preliminary review from Copilot & run CI? 

Thanks!

Rohanjames1997:
Thanks @hariharans29 ! 
Looks like the failures are due to inconsistent `ifdefs`(?). I'm looking into it. 
Do let me know if you have ideas too, but I may need you to rerun CI a few times more after I push fixes.

Rohanjames1997:
Since SBGemm is compiled only on linux, I have enabled this BF16 Pointwise Conv kernel only on linux as well.

Can the CI be run again? 🤞

aviralagrawal:
Exciting stuff. Looking forward to seeing this merged.

Rohanjames1997:
Thanks @aviralagrawal!

@hariharans29 gentle reminder 🙂

hariharans29:
> Thanks @aviralagrawal!
> 
> @hariharans29 gentle reminder 🙂

Hi - Sorry , I was OOF. Kicked off CI and will review this PR soon. Thanks!

onnxruntime/core/mlas/lib/snchwc.cpp
```diff
@@ -881,6 +882,11 @@ struct MLAS_NCHWC_CONV_POINTWISE_ALGORITHM : MLAS_NCHWC_GROUPED_CONV_ALGORITHM
 
 #if defined(MLAS_TARGET_AMD64) || defined(MLAS_TARGET_LARCH64) || (defined(MLAS_TARGET_ARM64) && defined(MLAS_USE_ARM_NEON_NCHWC))
         MLAS_CONV_POINTWISE_FLOAT_KERNEL* Kernel = GetMlasPlatform().ConvPointwiseFloatKernel;
+#if defined(MLAS_TARGET_ARM64) && defined(__linux__)
```
hariharans29: For completeness sake, shall we also include MLAS_USE_ARM_NEON_NCHWC here as well ?

onnxruntime/core/mlas/lib/mlasi.h
```diff
@@ -1368,6 +1371,9 @@ struct MLAS_PLATFORM {
     MLAS_CONV_FLOAT_KERNEL* ConvNchwcFloatKernel;
     MLAS_CONV_DEPTHWISE_FLOAT_KERNEL* ConvDepthwiseFloatKernel;
     MLAS_CONV_POINTWISE_FLOAT_KERNEL* ConvPointwiseFloatKernel;
+#if defined(__linux__)
```
hariharans29: Can we include `MLAS_TARGET_ARM64` in the ifdef to keep it consistent ?

onnxruntime/core/mlas/lib/platform.cpp
```diff
@@ -570,6 +570,9 @@ Return Value:
     this->ConvNchwcFloatKernel = MlasConvNchwcFloatKernelNeon;
     this->ConvDepthwiseFloatKernel = MlasConvDepthwiseFloatKernelNeon;
     this->ConvPointwiseFloatKernel = MlasConvPointwiseFloatKernelNeon;
+#if defined(__linux__)
```
hariharans29: Ditto

onnxruntime/core/mlas/lib/mlasi.h
```diff
@@ -967,6 +967,9 @@ extern "C" {
     MLAS_CONV_FLOAT_KERNEL MlasConvNchwcFloatKernelNeon;
     MLAS_CONV_DEPTHWISE_FLOAT_KERNEL MlasConvDepthwiseFloatKernelNeon;
     MLAS_CONV_POINTWISE_FLOAT_KERNEL MlasConvPointwiseFloatKernelNeon;
+#if defined(__linux__)
```
hariharans29: MLAS_TARGET_ARM64 in the ifdef ?

onnxruntime/contrib_ops/cpu/nchwc_ops.h
```diff
@@ -43,6 +44,10 @@ class NchwcConv final : public OpKernel {
  public:
   NchwcConv(const OpKernelInfo& info) : OpKernel(info), conv_attrs_(info) {
     ORT_ENFORCE(GetFusedActivationAttr(info, activation_).IsOK());
+#if defined(__aarch64__) && defined(__linux__)
```
hariharans29, line 47: Can we use `MLAS_TARGET_ARM64` instead of __aarch64__ for consistency ?

onnxruntime/contrib_ops/cpu/nchwc_ops.h
```diff
@@ -51,6 +56,9 @@ class NchwcConv final : public OpKernel {
   ConvAttributes conv_attrs_;
 
   MLAS_ACTIVATION activation_;
+#if defined(__aarch64__) && defined(__linux__)
```
hariharans29, line 59: Nit: Maybe we can do without this ifdef as it is defaulting to false anyway ?

onnxruntime/contrib_ops/cpu/nchwc_ops.cc
```diff
@@ -202,6 +202,12 @@ Status NchwcConv::Compute(OpKernelContext* context) const {
     }
   }
 
+#if defined(__aarch64__) && defined(__linux__)
+  bool use_bf16 = use_fastmath_mode_;
```
hariharans29: Are there tests that exercise the fast math path for the NCHWC Conv on Linux ?

onnxruntime/core/mlas/lib/mlasi.h
Rohanjames1997: Initially, I included `MLAS_TARGET_ARM64` as well. But the CI failed, and I had to restrict it to `__linux__` to fix the CI: https://github.com/microsoft/onnxruntime/pull/26838/changes/de547060e331fc3cc046e8a8581b862627591359

This is because I am reusing the `SBGEMM` algorithm which is restricted to linux. More discussion about it here: https://github.com/microsoft/onnxruntime/pull/17031

Resolving the other similar comments, and we can use this thread if you have further comments.

onnxruntime/contrib_ops/cpu/nchwc_ops.cc
Rohanjames1997: Since I am reusing ORT's SBGEMM, I am relying on its tests. 

Do you suggest adding new tests?

onnxruntime/contrib_ops/cpu/nchwc_ops.h
Rohanjames1997: Hmm, that would mean that the ifdef would be true for `_M_ARM64` as well, which is not supported by SBGEMM (?)
Do let me know if I am missing something,

onnxruntime/contrib_ops/cpu/nchwc_ops.h
Rohanjames1997: Maybe I misunderstood this comment.

The variable `use_fastmath_mode_` does not exist if this condition is false: `__aarch64__ && __linux__`. So, I only want to assign a value to it if the condition is true.

onnxruntime/core/mlas/lib/snchwc.cpp
Rohanjames1997: The `MLAS_USE_ARM_NEON_NCHWC` flag is already a part of the ifdef two lines above it, so this line will be reached only if it was true, no?

onnxruntime/contrib_ops/cpu/nchwc_ops.h
```diff
@@ -43,6 +44,10 @@ class NchwcConv final : public OpKernel {
  public:
   NchwcConv(const OpKernelInfo& info) : OpKernel(info), conv_attrs_(info) {
     ORT_ENFORCE(GetFusedActivationAttr(info, activation_).IsOK());
+#if defined(__aarch64__) && defined(__linux__)
+    auto config_ops = info.GetConfigOptions().GetConfigEntry(kOrtSessionOptionsMlasGemmFastMathArm64Bfloat16);
```
yuslepukhin, line 48: >y(kOrtSessionOptionsMlasGemmFastMathArm64Bfloat16); [](http://example.com/codeflow?start=59&length=51)

Does this need a default value?

onnxruntime/contrib_ops/cpu/nchwc_ops.cc
yuslepukhin: > bool [](http://example.com/codeflow?start=1&length=5)

nit: const #Resolved

onnxruntime/contrib_ops/cpu/nchwc_ops.cc
```diff
@@ -202,6 +202,12 @@ Status NchwcConv::Compute(OpKernelContext* context) const {
...
+#else
+  bool use_bf16 = false;
```
yuslepukhin: >bool  [](http://example.com/codeflow?start=2&length=5)

nit: constexpr const #Resolved

onnxruntime/contrib_ops/cpu/nchwc_ops.cc
hariharans29: Thanks. It would be nice to have a set of separated Nchwc sbconv tests as well - do you think it is simple enough to do by expanding the current NCHWC CPU Conv tests ?

onnxruntime/contrib_ops/cpu/nchwc_ops.h
hariharans29: Ah.. I see. Thanks for that.

onnxruntime/core/mlas/lib/snchwc.cpp
hariharans29: Shouldn't this be __aarch64__  (instead of MLAS_TARGET_ARM64) ?

onnxruntime/core/mlas/lib/mlasi.h
hariharans29: Sorry, I am not following this - if sbgemm implementation is only available on Linux, it shouldn't matter if the function pointer struct contains ConvPointwiseBf16Kernel for __linux__ && __aarch64__ right ?

Rohanjames1997: Ah, I misunderstood.
Yes, it can be enabled for `linux && aarch64`

Since it is already under the ifdef for `MLAS_USE_ARM_NEON_NCHWC`, adding `aarch64` will be redundant, but I can enable it for the sake of consistency.

onnxruntime/contrib_ops/cpu/nchwc_ops.cc
Rohanjames1997: Hmm I can check it out. I am not sure at this point 🤔

onnxruntime/core/mlas/lib/snchwc.cpp
Rohanjames1997: It may have the same effect, but I will make it `__aarch64__` for consistency


Btw out of curiosity, is this true?

```
MLAS_TARGET_ARM64 = __aarch64__ || _M_ARM64

MLAS_TARGET_ARM64 && __linux__ = (__aarch64__ && __linux__) || (_M_ARM64 && __linux__) 

MLAS_TARGET_ARM64 && __linux__ = __aarch64__ && __linux__
```

onnxruntime/contrib_ops/cpu/nchwc_ops.h
Rohanjames1997: Probably not.

It is assigned a value inside [onnxruntime_session_options_config_keys.h](https://github.com/microsoft/onnxruntime/blob/main/include/onnxruntime/core/session/onnxruntime_session_options_config_keys.h)


`static const char* const kOrtSessionOptionsMlasGemmFastMathArm64Bfloat16 = "mlas.enable_gemm_fastmath_arm64_bfloat16";`

In turn,`enable_gemm_fastmath_arm64_bfloat16` is controlled via a runtime flag (and defaults to false).

onnxruntime/contrib_ops/cpu/nchwc_ops.cc
Rohanjames1997: Thanks! 
~What do you suggest between `const` and `constexpr` ?~

Going with const since it's assigned at runtime.

onnxruntime/core/mlas/lib/mlasi.h
Rohanjames1997: Made it `defined(__aarch64__) && defined(__linux__)`

onnxruntime/core/mlas/lib/platform.cpp
Rohanjames1997: Made it `defined(__aarch64__) && defined(__linux__)`

onnxruntime/core/mlas/lib/mlasi.h
Rohanjames1997: Made it `defined(__aarch64__) && defined(__linux__)`

onnxruntime/contrib_ops/cpu/nchwc_ops.cc
Rohanjames1997: Done

onnxruntime/core/mlas/lib/snchwc.cpp
Rohanjames1997: Made it `defined(__aarch64__) && defined(__linux__)`

Rohanjames1997:
Thanks for the reviews @hariharans29 & @yuslepukhin!

Could you run CI again please?

Meanwhile I will try to figure out how to add tests for SBConv specifically.

hariharans29:
> Thanks for the reviews @hariharans29 & @yuslepukhin!
> 
> Could you run CI again please?
> 
> Meanwhile I will try to figure out how to add tests for SBConv specifically.

Thanks @Rohanjames1997 ! Sorry - I will respond to your comments later today. Running CI now.

onnxruntime/contrib_ops/cpu/nchwc_ops.h
yuslepukhin: Unless I am missing something GetConfigEntry() returns std::optional which may be not there, and that needs to be accounted for. GetConfigOrDefault may be a better way of handling that. 

[https://github.com/microsoft/onnxruntime/blob/9659a858808654ffb6a34a77016fb735fdf5d44f/onnxruntime/core/framework/config_options.h?plain=1#L27C1-L27C91](https://github.com/microsoft/onnxruntime/blob/9659a858808654ffb6a34a77016fb735fdf5d44f/onnxruntime/core/framework/config_options.h?plain=1#L27C1-L27C91)

Rohanjames1997: I see.. I reused the logic from line 34 in [matmul.h](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/providers/cpu/math/matmul.h#L34). 

My doubt is - wouldn't this [line](https://github.com/microsoft/onnxruntime/blob/main/include/onnxruntime/core/session/onnxruntime_session_options_config_keys.h#L375) run without fail? (And hence, always have a value)

hariharans29:
You can ignore the failing CUDA checks. They just need https://github.com/microsoft/onnxruntime/pull/27020 to go in.

Rohanjames1997:
Phew... Thanks 😅

onnxruntime/contrib_ops/cpu/nchwc_ops.cc
Rohanjames1997: I have extended the NCHWc CPU Conv test suite to accommodate the new BF16 kernel as well.
I intentionally broke the kernel to verify that the tests fail, and they did.

Resolving this.

hariharans29:
Can you rebase as well ? You may need some changes in main for some checks to pass

Rohanjames1997:
Done. Ready for CI when you are. Feel free to merge if the CI passes 🤞

hariharans29:
> Done. Ready for CI when you are. Feel free to merge if the CI passes 🤞

Thank you ! I will take one final look this evening and merge should everything look fine. Thanks again for the contribution !

onnxruntime/test/mlas/unittest/test_conv2d_nchwc.h
```diff
@@ -224,3 +227,51 @@ class MlasNchwcConv2DTest : public MlasConv2DTest<Threaded> {
     }
   }
 };
+
+#if defined(__aarch64__) && defined(__linux__)
+template <bool Threaded>
+class MlasNchwcConv2DBf16Test : public MlasNchwcConv2DTest<Threaded> {
+ public:
+  MlasNchwcConv2DBf16Test() { this->UseBf16_ = true; }
```
hariharans29, line 235: This test needs to be skipped on platforms that don't support BF16 acceleration

onnxruntime/test/mlas/unittest/test_conv2d_nchwc.h
```diff
@@ -8,6 +8,8 @@
 template <bool Threaded>
 class MlasNchwcConv2DTest : public MlasConv2DTest<Threaded> {
  protected:
+  bool UseBf16_ = false;
```
hariharans29, line 11: Can we pass this in an input param to MlasConv2D(...) instead off having a member variable ?

hariharans29:
Left behind a couple of comments - can you please address them when you get a chance ? The rest of the code looks fine ! Thank you again !

onnxruntime/test/mlas/unittest/test_conv2d_nchwc.h
Rohanjames1997: This is inside `#if defined(__aarch64__) && defined(__linux__)` on line 231

Which platforms do you suggest to filter out?

Rohanjames1997: Currently, that is the condition for [sbgemm_kernel_neon.cpp](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/mlas/lib/sbgemm_kernel_neon.cpp#L18) to compile

hariharans29: Thanks Rohan. I guess my concern was what if this test runs on a platform where `MlasBf16AccelerationSupported()`returns false. I was thinking that when `MlasBf16AccelerationSupported()` evaluated to false, we would have a gtest skip.

onnxruntime/test/mlas/unittest/test_conv2d_nchwc.h
Rohanjames1997: Hmm, I tried it, but it increases the diff size & the complexity unnecessarily 🤔 

Adding it as a parameter would require modifying the base class virtual function signature in `test_conv2d.h`, even though the base class doesn't use it (it calls `MlasConv`, not `MlasNchwcConv`). This introduced an unused parameter and polluted the interface for all derived classes.

onnxruntime/test/mlas/unittest/test_conv2d_nchwc.h
Rohanjames1997: Ah, yes. In those cases, it would run the fp32 test case itself. 

I will register the tests only when `MlasBf16AccelerationSupported()` is true.

onnxruntime/test/mlas/unittest/test_conv2d_nchwc.h
hariharans29: I see, ok, thanks. Maybe we can keep it as is for this PR. In your other PR, we can try and see if we can refactor.

onnxruntime/test/mlas/unittest/test_conv2d_nchwc.h
Rohanjames1997: Done.

onnxruntime/test/mlas/unittest/test_conv2d_nchwc.h
Rohanjames1997: Sureo

PR #26840, opened 2025-12-20
Plugin EP API doc updates and misc fixes

edgchen1:
### Description

- Some documentation updates. Goal is to clarify and make consistent. Changes are definitely open to discussion.
- Fix possible leak of `OrtAllocator`.

### Motivation and Context

Documentation updates and clean up.

### DIFF
--- a/include/onnxruntime/core/session/onnxruntime_ep_c_api.h
+++ b/include/onnxruntime/core/session/onnxruntime_ep_c_api.h
@@ -1348,18 +1348,23 @@ struct OrtEpFactory {
 
   /** \brief Get information from the execution provider about OrtHardwareDevice support.
    *
+   * If this function succeeds, ORT will take ownership of the first `num_ep_devices` OrtEpDevice instances populated
+   * in the `ep_devices` array.
+   *
    * \param[in] this_ptr The OrtEpFactory instance.
    *                     Non-const as the factory is passed through to the CreateEp call via the OrtEpDevice.
    * \param[in] devices The OrtHardwareDevice instances that are available.
    * \param[in] num_devices The number of OrtHardwareDevice instances.
    * \param[out] ep_devices OrtEpDevice instances for each OrtHardwareDevice that the EP can use.
-   *                        The implementation should call OrtEpApi::CreateEpDevice to create, and add the OrtEpDevice
-   *                        instances to this pre-allocated array. ORT will take ownership of the values returned.
-   *                        i.e. usage is `ep_devices[0] = <ptr to OrtEpDevice created with OrtEpApi::CreateEpDevice>;`
+   *                        The implementation should call OrtEpApi::CreateEpDevice to create the OrtEpDevice
+   *                        instances and add them to this pre-allocated array.
+   *                        I.e., usage is something like this:
+   *                        `ep_devices[0] = <ptr to OrtEpDevice created with OrtEpApi::CreateEpDevice>;`
    * \param[in] max_ep_devices The maximum number of OrtEpDevices that can be added to ep_devices.
    *                           Current default is 8. This can be increased if needed.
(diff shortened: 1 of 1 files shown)


include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -204,14 +204,7 @@ typedef struct OrtNodeComputeInfo OrtNodeComputeInfo;
...
  */
 struct OrtNodeFusionOptions {
-  /** \brief The ONNX Runtime version the OrtNodeFusionOptions was compiled with.
-   *
-   * Implementation should set to ORT_API_VERSION.
-   * ORT will use this to ensure it does not use members that were not available when the EP library was compiled.
-   *
-   * \since Version 1.23.
-   */
-  uint32_t ort_version_supported;
```
skottmckay, line 215: Why delete this info?

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -332,6 +318,12 @@ typedef OrtStatus*(ORT_API_CALL* OrtKernelCreateFunc)(_In_ void* kernel_create_f
  */
 struct OrtEpApi {
   /** \brief Create an OrtEpDevice for the EP and an OrtHardwareDevice.
+   *
+   * The created instance must be released with OrtEpApi::ReleaseEpDevice.
+   *
+   * \note When called from OrtEpFactory::GetSupportedDevices to populate the ep_devices output, ORT will take
+   *       ownership of the OrtEpDevice instances.
```
skottmckay, line 421: When else would it be called? Wondering if the doco should be saying 'release if you don't provide to ORT in GetSupportedDevices'. Or maybe reverse the order of the two statements and remove the `\note` if we only expect usage in GetSupportedDevices.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -533,6 +525,8 @@ struct OrtEpApi {
             _In_ const OrtSyncStream* producer_stream, _In_ const OrtSyncStream* consumer_stream);
 
   /** \brief Create an OrtHardwareDevice.
+   *
+   * The created instance must be released with OrtEpApi::ReleaseHardwareDevice.
```
skottmckay, line 625: If the OrtEpDevice using this is provided to ORT do we take a copy, or the user needs to keep this valid and not release until EP unload?

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
edgchen1: was trying to make it consistent and I chose the more concise version. alternatively, I could add this info to the other `ort_version_supported` docs.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
edgchen1: AFAIK, we do not copy the `OrtHardwareDevice`. `OrtEpDevice` creation just copies the pointer. currently, the user would need to ensure that a created `OrtHardwareDevice` doesn't get released before `OrtEpDevices` are done using it.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
edgchen1: hm, the main point I'd like to convey is that `CreateEpDevice`'s output must be released with `ReleaseEpDevice`. 

it just so happens that `OrtEpFactory::GetSupportedDevices` is the primary place `CreateEpDevice` is expected to be called, and that ORT will take ownership of the `OrtEpDevice` on success.

PR #26846, opened 2025-12-22
Add enable_profiling in runoptions

xiaofeihan1:
### Description
Support run-level profiling

This PR adds support for profiling individual Run executions, similar to session-level profiling. Developers can enable run-level profiling by setting `enable_profiling` and `profile_file_prefix` in RunOptions. Once the run completes, a JSON profiling file will be saved using profile_file_prefix + timestamp.

png (2)

### Key Changes
1. Introduced a local variable `run_profiler` in `InferenceSession::Run`, which is destroyed after the run completes. Using a dedicated profiler per run ensures that profiling data is isolated and prevents interleaving or corruption across runs.
2. To maintain accurate execution time when both session-level and run-level profiling are enabled, overloaded `Start` and `EndTimeAndRecordEvent` functions have been added. These allow the caller to provide timestamps instead of relying on `std::chrono::high_resolution_clock::now()`, avoiding potential timing inaccuracies.
3. Added a TLS variable `tls_run_profiler_` to support run-level profiling with WebGPU Execution Provider (EP). This ensures that when multiple threads enable run-level profiling, each thread logs only to its own WebGPU profiler, keeping thread-specific data isolated.
4. Use `HH:MM:SS.mm` instead of `HH:MM:SS`in the JSON filename to prevent conflicts when profiling multiple consecutive runs.

### Motivation and Context
Previously, profiling only for session level. Sometimes developer want to profile for specfic run . so the PR comes.

### Some details

When profiling is enabled via RunOptions, it should ideally collect two types of events:
1. Profiler events
Used to calculate the CPU execution time of each operator.
2. Execution Provider (EP) profiler events
Used to measure GPU kernel execution time. 

Unlike session-level, we need to ensure the collecting events is correct for multiple thread scenario.

For 1, this can be supported easily(sequential_executor.cc). We use a thread-local storage (TLS) variable, RunLevelState (defined in profiler.h), to maintain run-level profiling state for each thread.

For 2, each Execution Provider (EP) has its own profiler implementation, and each EP must ensure correct behavior under run-level profiling. This PR ensures that the WebGPU profiler works correctly with run-level profiling.

# Test Cases

| Scenario | Example | Expected Result |
|---------|---------|-----------------|
| Concurrent runs on the same session with different run-level profiling settings| t1: `sess1.Run({ enable_profiling: true })`<br>t2: `sess1.Run({ enable_profiling: false })`<br>t3: `sess1.Run({ enable_profiling: true })` | Two trace JSON files are generated: one for `t1` and one for `t3`. |
| Run-level profiling enabled together with session-level profiling| `sess1 = OrtSession({ enable_profiling: true })`<br>`sess1.Run({ enable_profiling: true })` | Two trace JSON files are generated: one corresponding to session-level profiling and one corresponding to run-level profiling. |

### DIFF
--- a/cmake/external/abseil-cpp.cmake
+++ b/cmake/external/abseil-cpp.cmake
@@ -21,7 +21,8 @@ else()
 endif()
 
 if(Patch_FOUND AND WIN32)
-  set(ABSL_PATCH_COMMAND ${Patch_EXECUTABLE} --binary --ignore-whitespace -p1 < ${PROJECT_SOURCE_DIR}/patches/abseil/absl_windows.patch)
+  set(ABSL_PATCH_COMMAND ${Patch_EXECUTABLE} --binary --ignore-whitespace -p1 < ${PROJECT_SOURCE_DIR}/patches/abseil/absl_windows.patch &&
+                         ${Patch_EXECUTABLE} --binary --ignore-whitespace -p1 < ${PROJECT_SOURCE_DIR}/patches/abseil/absl_cuda_warnings.patch)
 else()
   set(ABSL_PATCH_COMMAND "")
 endif()
--- a/cmake/patches/abseil/absl_cuda_warnings.patch
+++ b/cmake/patches/abseil/absl_cuda_warnings.patch
@@ -1,0 +1,40 @@
+diff --git a/absl/hash/internal/hash.h b/absl/hash/internal/hash.h
+index 1234567..abcdefg 100644
+--- a/absl/hash/internal/hash.h
++++ b/absl/hash/internal/hash.h
+@@ -477,7 +477,7 @@ H AbslHashValue(H hash_state, T (&)[N]) {
+ template <typename H, typename T, size_t N>
+ H AbslHashValue(H hash_state, T (&)[N]) {
+   static_assert(
+-      sizeof(T) == -1,
++      sizeof(T) == size_t(-1),
+       "Hashing C arrays is not allowed. For string literals, wrap the literal "
+       "in absl::string_view(). To hash the array contents, use "
+       "absl::MakeSpan() or make the array an std::array. To hash the array "
+diff --git a/absl/hash/hash.h b/absl/hash/hash.h
+index 1234567..abcdefg 100644
+--- a/absl/hash/hash.h
++++ b/absl/hash/hash.h
+@@ -333,7 +333,8 @@ class HashState : public hash_internal::HashStateBase<HashState> {
+       absl::enable_if_t<
+           std::is_base_of<hash_internal::HashStateBase<T>, T>::value, int> = 0>
+   static HashState Create(T* state) {
+-    HashState s;
++    HashState s = {};
++    (void)s;
+     s.Init(state);
+     return s;
+   }
+diff --git a/absl/container/internal/raw_hash_set.h b/absl/container/internal/raw_hash_set.h
+index 1234567..abcdefg 100644
+--- a/absl/container/internal/raw_hash_set.h
++++ b/absl/container/internal/raw_hash_set.h
+@@ -464,7 +464,7 @@ inline uint16_t NextSeed() {
+ inline uint16_t NextSeed() {
+   static_assert(PerTableSeed::kBitCount == 16);
+   thread_local uint16_t seed =
+-      static_cast<uint16_t>(reinterpret_cast<uintptr_t>(&seed));
++      static_cast<uint16_t>(reinterpret_cast<uintptr_t>(&seed) & 0xFFFFu);
+   seed += uint16_t{0xad53};
+   return seed;
+ }
--- a/cmake/vcpkg-ports/abseil/absl_cuda_warnings.patch
+++ b/cmake/vcpkg-ports/abseil/absl_cuda_warnings.patch
@@ -1,0 +1,40 @@
+diff --git a/absl/hash/internal/hash.h b/absl/hash/internal/hash.h
+index 1234567..abcdefg 100644
+--- a/absl/hash/internal/hash.h
++++ b/absl/hash/internal/hash.h
+@@ -477,7 +477,7 @@ H AbslHashValue(H hash_state, T (&)[N]) {
+ template <typename H, typename T, size_t N>
+ H AbslHashValue(H hash_state, T (&)[N]) {
+   static_assert(
+-      sizeof(T) == -1,
++      sizeof(T) == size_t(-1),
+       "Hashing C arrays is not allowed. For string literals, wrap the literal "
+       "in absl::string_view(). To hash the array contents, use "
+       "absl::MakeSpan() or make the array an std::array. To hash the array "
+diff --git a/absl/hash/hash.h b/absl/hash/hash.h
+index 1234567..abcdefg 100644
+--- a/absl/hash/hash.h
++++ b/absl/hash/hash.h
+@@ -333,7 +333,8 @@ class HashState : public hash_internal::HashStateBase<HashState> {
+       absl::enable_if_t<
+           std::is_base_of<hash_internal::HashStateBase<T>, T>::value, int> = 0>
+   static HashState Create(T* state) {
+-    HashState s;
++    HashState s = {};
++    (void)s;
+     s.Init(state);
+     return s;
+   }
+diff --git a/absl/container/internal/raw_hash_set.h b/absl/container/internal/raw_hash_set.h
+index 1234567..abcdefg 100644
+--- a/absl/container/internal/raw_hash_set.h
++++ b/absl/container/internal/raw_hash_set.h
+@@ -464,7 +464,7 @@ inline uint16_t NextSeed() {
+ inline uint16_t NextSeed() {
+   static_assert(PerTableSeed::kBitCount == 16);
+   thread_local uint16_t seed =
+-      static_cast<uint16_t>(reinterpret_cast<uintptr_t>(&seed));
++      static_cast<uint16_t>(reinterpret_cast<uintptr_t>(&seed) & 0xFFFFu);
+   seed += uint16_t{0xad53};
+   return seed;
+ }
--- a/cmake/vcpkg-ports/abseil/portfile.cmake
+++ b/cmake/vcpkg-ports/abseil/portfile.cmake
@@ -9,6 +9,7 @@ vcpkg_from_github(
     SHA512 4ee1a217203933382e728d354a149253a517150eee7580a0abecc69584b2eb200d91933ef424487e3a3fe0e8ab5e77b0288485cac982171b3585314a4417e7d4
     HEAD_REF master
     PATCHES absl_windows.patch
+            absl_cuda_warnings.patch
 )
 
 
--- a/onnxruntime/core/platform/linux/device_discovery.cc
+++ b/onnxruntime/core/platform/linux/device_discovery.cc
@@ -7,6 +7,7 @@
 #include <fstream>
 #include <iterator>
 #include <optional>
+#include <regex>
 #include <string_view>
 
 #include "core/common/common.h"
@@ -114,6 +115,28 @@ std::optional<bool> IsGpuDiscrete(uint16_t vendor_id, uint16_t device_id) {
   return std::nullopt;
 }
 
+Status GetPciBusId(const std::filesystem::path& sysfs_path, std::optional<std::string>& pci_bus_id) {
+  constexpr const char* regex_pattern{R"([0-9a-f]+:[0-9a-f]+:[0-9a-f]+[.][0-9a-f]+)"};
+  static const std::regex pci_bus_id_regex(regex_pattern);
+
+  std::error_code error_code;
+  auto pci_bus_id_path = std::filesystem::canonical(sysfs_path / "device", error_code);  // resolves symlink to PCI bus id, e.g. 0000:65:00.0
+  ORT_RETURN_IF_ERROR(ErrorCodeToStatus(error_code));
+
+  auto pci_bus_id_filename = pci_bus_id_path.filename();
+  if (std::regex_match(pci_bus_id_filename.string(), pci_bus_id_regex)) {
+    pci_bus_id = pci_bus_id_filename.string();
+  } else {
+    pci_bus_id = {};
+    LOGS_DEFAULT(WARNING) << MakeString("Skipping pci_bus_id for PCI path at \"",
+                                        pci_bus_id_path.string(),
+                                        "\" because filename \"", pci_bus_id_filename, "\" dit not match expected pattern of ",
+                                        regex_pattern);
+  };
+
+  return Status::OK();
+}
+
 Status GetGpuDeviceFromSysfs(const GpuSysfsPathInfo& path_info, OrtHardwareDevice& gpu_device_out) {
   OrtHardwareDevice gpu_device{};
   const auto& sysfs_path = path_info.path;
@@ -138,6 +161,12 @@ Status GetGpuDeviceFromSysfs(const GpuSysfsPathInfo& path_info, OrtHardwareDevic
   if (const auto is_gpu_discrete = IsGpuDiscrete(vendor_id, device_id);
       is_gpu_discrete.has_value()) {
     gpu_device.metadata.Add("Discrete", (*is_gpu_discrete ? "1" : "0"));
+  }
+
+  std::optional<std::string> pci_bus_id;
+  ORT_RETURN_IF_ERROR(GetPciBusId(sysfs_path, pci_bus_id));
+  if (pci_bus_id) {
+    gpu_device.metadata.Add("pci_bus_id", std::move(*pci_bus_id));
   }
 
   gpu_device.type = OrtHardwareDeviceType_GPU;
--- a/onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc
+++ b/onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc
@@ -584,6 +584,7 @@ struct NvTensorRtRtxEpFactory : OrtEpFactory {
    * @return True if the device is a supported NVIDIA GPU, false otherwise.
    */
   bool IsOrtHardwareDeviceSupported(const OrtHardwareDevice& device) {
+#if _WIN32
     const auto& metadata_entries = device.metadata.Entries();
     const auto it = metadata_entries.find("LUID");
     if (it == metadata_entries.end()) {
@@ -625,6 +626,25 @@ struct NvTensorRtRtxEpFactory : OrtEpFactory {
     }
 
     return false;
+#else
+    const auto& metadata_entries = device.metadata.Entries();
+    const auto it = metadata_entries.find("pci_bus_id");
+    if (it == metadata_entries.end()) {
+      return false;
+    }
+    auto& target_id = it->second;
+    int cuda_device_idx = 0;
+    if (cudaDeviceGetByPCIBusId(&cuda_device_idx, target_id.c_str()) != cudaSuccess) {
+      return false;
+    }
+
+    cudaDeviceProp prop;
+    if (cudaGetDeviceProperties(&prop, cuda_device_idx) != cudaSuccess) {
+      return false;
+    }
+    // Ampere architecture or newer is required.
+    return prop.major >= 8;
+#endif
   }
 
   // Creates and returns OrtEpDevice instances for all OrtHardwareDevices that this factory supports.
--- a/onnxruntime/core/providers/nv_tensorrt_rtx/version_script.lds
+++ b/onnxruntime/core/providers/nv_tensorrt_rtx/version_script.lds
@@ -2,6 +2,8 @@
 VERS_1.0 {
   global:
     GetProvider;    
+    CreateEpFactories;
+    ReleaseEpFactory;
 
   # Hide everything else.
   local:
--- a/onnxruntime/test/providers/nv_tensorrt_rtx/nv_basic_test.cc
+++ b/onnxruntime/test/providers/nv_tensorrt_rtx/nv_basic_test.cc
@@ -278,7 +278,6 @@ INSTANTIATE_TEST_SUITE_P(NvExecutionProviderTest, TypeTests,
                                            ),
                          [](const testing::TestParamInfo<TypeTests::ParamType>& info) { return getTypeAsName(info.param); });
 
-#ifdef _WIN32
 static bool SessionHasEp(Ort::Session& session, const char* ep_name) {
   // Access the underlying InferenceSession.
   const OrtSession* ort_session = session;
@@ -295,7 +294,6 @@ static bool SessionHasEp(Ort::Session& session, const char* ep_name) {
 }
 
 // Tests autoEP feature to automatically select an EP that supports the GPU.
-// Currently only works on Windows.
 TEST(NvExecutionProviderTest, AutoEp_PreferGpu) {
   PathString model_name = ORT_TSTR("nv_execution_provider_auto_ep.onnx");
   std::string graph_name = "test";
@@ -305,7 +303,11 @@ TEST(NvExecutionProviderTest, AutoEp_PreferGpu) {
   CreateBaseModel(model_name, graph_name, dims);
 
   {
+#if _WIN32
     ort_env->RegisterExecutionProviderLibrary(kNvTensorRTRTXExecutionProvider, ORT_TSTR("onnxruntime_providers_nv_tensorrt_rtx.dll"));
+#else
+    ort_env->RegisterExecutionProviderLibrary(kNvTensorRTRTXExecutionProvider, ORT_TSTR("libonnxruntime_providers_nv_tensorrt_rtx.so"));
+#endif
 
     Ort::SessionOptions so;
     so.SetEpSelectionPolicy(OrtExecutionProviderDevicePolicy_PREFER_GPU);
@@ -601,8 +603,6 @@ TEST(NvExecutionProviderTest, FP4CustomOpModel) {
 
   LOGS_DEFAULT(INFO) << "[NvExecutionProviderTest] TRT FP4 dynamic quantize model run completed successfully";
 }
-
-#endif
 
 }  // namespace test
 }  // namespace onnxruntime
--- a/onnxruntime/test/providers/nv_tensorrt_rtx/nv_ep_context_test.cc
+++ b/onnxruntime/test/providers/nv_tensorrt_rtx/nv_ep_context_test.cc
@@ -14,7 +14,6 @@ namespace test {
 
 RegisteredEpDeviceUniquePtr AppendTrtEtxEP(Ort::SessionOptions& session_options, std::unordered_map<std::string, std::string>& option_map) {
   RegisteredEpDeviceUniquePtr nv_tensorrt_rtx_ep;
-#ifdef _WIN32
   /// Since this test runs after other tests that use registration interface this test has to use it as well
   /// windows as otherwise the kernel registry inside the EP will not be populated. The legacy APis ony call the initialize once.
   Utils::RegisterAndGetNvTensorRtRtxEp(*ort_env, nv_tensorrt_rtx_ep);
@@ -26,9 +25,6 @@ RegisteredEpDeviceUniquePtr AppendTrtEtxEP(Ort::SessionOptions& session_options,
     }
   }
   session_options.AppendExecutionProvider_V2(*ort_env, {selected_device}, option_map);
-#else
-  session_options.AppendExecutionProvider(onnxruntime::kNvTensorRTRTXExecutionProvider, option_map);
-#endif
   return nv_tensorrt_rtx_ep;
 }
 
--- a/onnxruntime/test/providers/nv_tensorrt_rtx/test_nv_trt_rtx_ep_util.cc
+++ b/onnxruntime/test/providers/nv_tensorrt_rtx/test_nv_trt_rtx_ep_util.cc
@@ -24,7 +24,6 @@
 
 namespace onnxruntime {
 namespace test {
-#ifdef _WIN32
 
 Utils::NvTensorRtRtxEpInfo Utils::nv_tensorrt_rtx_ep_info;
 
@@ -61,7 +60,6 @@ void Utils::RegisterAndGetNvTensorRtRtxEp(Ort::Env& env, RegisteredEpDeviceUniqu
     c_api.UnregisterExecutionProviderLibrary(env, nv_tensorrt_rtx_ep_info.registration_name.c_str());
   });
 }
-#endif  // _WIN32
 
 void CreateBaseModel(const PathString& model_name,
                      std::string graph_name,


onnxruntime/core/framework/utils.cc
```diff
@@ -745,17 +751,16 @@ common::Status ExecuteGraph(const SessionState& session_state,
...
+                            profiling::Profiler* run_profiler) {
+  return ExecuteGraph(session_state, feeds_fetches_manager, feeds, fetches,
+                      execution_mode, run_options.terminate, logger,
 #ifdef ORT_ENABLE_STREAM
                       device_stream_collection_holder,
 #endif
-                      run_options.only_execute_path_to_fetches);
+                      run_options.only_execute_path_to_fetches,
+                      nullptr,
+                      run_profiler);
```
yuslepukhin, line 763: We have a number of things being passed from RunOptions here. Can we modify the signature in a way that a reference to RunOptions is being passed?

Then we can instantiate the profiler higher in the stack, inside ExecuteGraph? 

I can see that RunOptions are being passed in one of the overloads, that seems sensible. #Resolved

onnxruntime/core/framework/sequential_executor.cc
```diff
@@ -363,26 +385,42 @@ class KernelScope {
...
-                                         {"activation_size", std::to_string(input_activation_sizes_)},
-                                         {"parameter_size", std::to_string(input_parameter_sizes_)},
-                                         {"output_size", std::to_string(total_output_sizes_)},
-                                         {"input_type_shape", input_type_shape_},
-                                         {"output_type_shape", output_type_shape_},
-                                         {"thread_scheduling_stats",
-                                          concurrency::ThreadPool::StopProfiling(session_state_.GetThreadPool())},
-                                     });
+
+      std::initializer_list<std::pair<std::string, std::string>> event_args = {
```
yuslepukhin: > std:: [](http://example.com/codeflow?start=5&length=6)

Should this be constexpr? Should this be outside the destructor? #Resolved

onnxruntime/core/framework/sequential_executor.cc
```diff
@@ -173,12 +173,18 @@ class SessionScope {
...
         ,
-        dump_context_{
-            session_state_.GetGraphExecutionCounter(), 0}
+        dump_context_{session_state_.GetGraphExecutionCounter(), 0}
 #endif
   {
-    if (session_state_.Profiler().IsEnabled()) {
-      session_start_ = session_state.Profiler().Start();
+    bool session_profiling_enabled = session_state_.Profiler().IsEnabled();
+    bool run_profiling_enabled = run_profiler_ && run_profiler_->IsEnabled();
```
yuslepukhin: > boo [](http://example.com/codeflow?start=3&length=4)

const  #Resolved

onnxruntime/core/framework/sequential_executor.cc
```diff
@@ -173,12 +173,18 @@ class SessionScope {
...
 #ifdef DEBUG_NODE_INPUTS_OUTPUTS
         ,
-        dump_context_{
-            session_state_.GetGraphExecutionCounter(), 0}
+        dump_context_{session_state_.GetGraphExecutionCounter(), 0}
 #endif
   {
-    if (session_state_.Profiler().IsEnabled()) {
-      session_start_ = session_state.Profiler().Start();
+    bool session_profiling_enabled = session_state_.Profiler().IsEnabled();
```
yuslepukhin: > boo [](http://example.com/codeflow?start=3&length=4)

const #Resolved

yuslepukhin: > session_state_.Profiler().IsEnabled(); [](http://example.com/codeflow?start=36&length=39)

I am still not convinced that we should allow both profilers to run in parallel. 

Do you have a use case for that? What would be the purpose to collect the same data?

If someone wants continuous profiling, would it not be the same thing as running it with RunOptons? #Resolved

onnxruntime/core/providers/webgpu/webgpu_context.cc
```diff
@@ -660,7 +661,16 @@ void WebGpuContext::CollectProfilingData(profiling::Events& events) {
                                      static_cast<int64_t>(std::round(start_time / 1000.0)),
                                      static_cast<int64_t>(std::round((end_time - start_time) / 1000.0)),
                                      event_args);
-        events.emplace_back(std::move(event));
+
+        // Distribute the event to all WebGPU EP profilers.
```
yuslepukhin: > Distribute the event to all WebGPU EP profilers. [](http://example.com/codeflow?start=10&length=49)

How many profilers are we expecting? #Resolved

onnxruntime/core/providers/webgpu/webgpu_execution_provider.cc
```diff
@@ -1023,8 +1030,20 @@ Status WebGpuExecutionProvider::OnRunEnd(bool /* sync_stream */, const onnxrunti
...
   }
 
-  if (profiler_->Enabled()) {
-    context_.CollectProfilingData(profiler_->Events());
+  std::vector<WebGpuProfiler*> profilers;
+  if (session_profiler_ && session_profiler_->Enabled()) {
+    profilers.push_back(session_profiler_);
+  }
+
+  if (run_options.enable_profiling && tls_run_profiler_) {
```
yuslepukhin: >tls_run_profiler_ [](http://example.com/codeflow?start=38&length=17)

Why do we need TLS if RunOptions profiler is thread local already? TLS and RunOptions are two different things.
Also who is going to clean up TLS if we no longer need profiling, it would be better to keep everything on the stack or at least Run() scoped. #Resolved

include/onnxruntime/core/framework/run_options.h
```diff
@@ -35,6 +35,14 @@ struct OrtRunOptions {
...
   bool only_execute_path_to_fetches = false;
 
+  // Set to 'true' to enable profiling for this run.
+  bool enable_profiling = false;
+
+  // File prefix for profiling result for this run.
+  // The actual filename will be: <profile_file_prefix>_<timestamp>.json
+  // Only used when enable_profiling is true.
+  std::string profile_file_prefix = "onnxruntime_run_profile";
+
```
yuslepukhin, line 45: We need C and C++ API and Python comes after that.

And we need tests #Resolved

onnxruntime/core/framework/sequential_executor.cc
xiaofeihan1: This depends on how we want to handle the case when both run-level and session-level profiling are enabled.

For example, when a user calls Session::Run with both run-level and session-level profiling enabled, there will be two profilers active: a local run_profiler and the session_profiler_ owned by InferenceSession. The current implementation guarantees that two JSON files are generated, and that the events recorded in the run-level profiling output are a strict subset of those in the session-level profiling output. 

In this scenario, each operator execution generates two identical profiling events: one is recorded by the session-level profiler, and the other is recorded by the run-level profiler.

onnxruntime/core/providers/webgpu/webgpu_context.cc
xiaofeihan1: Let me say the 

### case 1: session-level: ON, run-level: OFF
```
session1 = InferenceSession(enable_profiling = true /*session-level*/)
thread1: session1.Run(enable_profiling=false /*run-level*/)
```

There are one Profiler instance, one WebGpuProfiler instance(is owned by `Profiler`). So here the number of profilers is one. It always use session-level profiler, i.g. `session_profiler_` of InfereceSession.

### case 2: session-level: OFF, run-level: ON
```
session1 = InferenceSession(enable_profiling = false /*session-level*/)
session1.Run(enable_profiling=true /*run-level*/)
```

There are one Profiler instance, one WebGpuProfiler instance(is owned by `Profiler`). So here the number of profilers is one. It always use run-level profiler, i.g. local variable `profiler`.


### case 3: session-level: ON, run-level: ON
```
session1 = InferenceSession(enable_profiling = true /*session-level*/)
session1.Run(enable_profiling=true /*run-level*/)
```

There are two Profiler instance, two WebGpuProfiler instance(is owned by `Profiler`). So here the number of profilers is two. It always use run-level profiler.

### case 4: session-level: ON,  run-level: two threads ON, one thread OFF
```
session1 = InferenceSession(enable_profiling = true /*session-level*/)
thread1: session1.Run(enable_profiling=true /*run-level*/)
thread2: session1.Run(enable_profiling=true /*run-level*/)
thread3: session1.Run(enable_profiling=false /*run-level*/)
```

There are three Profiler instances and three WebGpuProfiler instances(are owned by `Profiler`). Because webgpu ep doesn't support concurrent run yet, so when the number of profilers in this `CollectProfilingData` is two: one is session-level profiler and another one is one of the two EP profilers(which one is used is determined by the current thread during Run)

onnxruntime/core/providers/webgpu/webgpu_execution_provider.cc
xiaofeihan1: There are two types of profiling events: CPU time(collected by the Profiler) and GPU time(collected by EP profilers, e.g., the WebGPU profiler). The Profiler own EP Profilers. Here I have no better way to obtain correct `WebGPU Profiler` when two threads run the same session with run-level profiling enabled for both, so I stored it in TLS.

onnxruntime/core/framework/utils.cc
xiaofeihan1: The `run_profiler` needs to be instantiated in Inference::Run because it records the `model_run`. Instantiating it inside ExecuteGraph would prevent the model’s execution time from being recorded.
```
session_profiler_.EndTimeAndRecordEvent(profiling::SESSION_EVENT, "model_run", tp, now);
```
image

onnxruntime/core/framework/sequential_executor.cc
xiaofeihan1: We shouldn’t use constexpr because this records the execution time of each operator, and the event arguments differ for each operator.
In the main branch, the kernel execution time is recorded in the kernel scope destructor, which signals the end of the kernel execution. Why should this be moved outside the scope?

onnxruntime/core/framework/sequential_executor.cc
xiaofeihan1: Thanks, done!

onnxruntime/core/framework/sequential_executor.cc
xiaofeihan1: Done!

include/onnxruntime/core/framework/run_options.h
xiaofeihan1: Thanks for the info! I’ve exposed the C/C++ API and added the corresponding test cases.

onnxruntime/core/framework/sequential_executor.cc
yuslepukhin: Still, it is not clear, what is the use case for such parallel profiling? If not, then it should be disabled. It is either one or another.

onnxruntime/core/providers/webgpu/webgpu_execution_provider.cc
yuslepukhin: Can we not store a reference to it in the profiler instance? The profiler instance is specific to the Run(), correct?

onnxruntime/core/framework/sequential_executor.cc
yuslepukhin: I do not understand. Initializer list is not a mutable container.

xiaofeihan1: Let me answer your question again:
1. “Should this be constexpr?”
Using `const std::initializer_list` here is fine, but `constexpr std::initializer_list` is not suitable because some of the values are only available at runtime.
2. “Should this be outside the destructor?”
Why do you think it should be moved outside the destructor? The previous code also constructed the event inside the destructor.

onnxruntime/core/providers/webgpu/webgpu_execution_provider.cc
xiaofeihan1: The profiler instance is specific to a single Run, but currently we have no way to access the Profiler instance in `WebGpuExecutionProvider::OnRunEnd(bool /* sync_stream */, const onnxruntime::RunOptions& run_options)`.

One possible approach is to change the virtual function signature to `OnRunEnd(bool, RunOptions, Profiler*)`. However, this would require modifying many files, and it only benefits the WebGPU EP profiler, so it seems unnecessary.

In short, supporting a multithreaded scenario for the `Profiler` is easy, but supporting it for `EP Profiler` is difficult, because each EP profiler has its own implementation. In the WebGPU EP, we use thread-local storage (TLS) to ensure that the WebGPU EP profiler works correctly in multithreaded scenarios. Therefore, I believe the current solution is simple and effective.

onnxruntime/core/framework/sequential_executor.cc
xiaofeihan1: Changed the code. When session-level and run-level profiling is enabled both, we will ignore run-level profiling with a warning. Because session-level profiling may have collected some logs from previous runs, it would be a pity to discard them. 

Consider the case:
A session (with session-level profiling enabled) runs three times — Run1 → Run2 → Run3(with run-level profiling enabled). It will output a session-level profiling JSON file.

onnxruntime/core/providers/webgpu/webgpu_execution_provider.cc
xiaofeihan1: Ignore the previous comments. I have now successfully removed the TLS by refactoring the WebGPU code. Thanks for you suggestion!

include/onnxruntime/core/framework/run_options.h
```diff
@@ -35,6 +35,14 @@ struct OrtRunOptions {
   // So it is possible that only some of the nodes are executed.
   bool only_execute_path_to_fetches = false;
 
+  // Set to 'true' to enable profiling for this run.
+  bool enable_profiling = false;
```
yuslepukhin, line 39: >bool enable_profiling = false; [](http://example.com/codeflow?start=2&length=30)

Should the default be false? #Resolved

onnxruntime/core/framework/sequential_executor.cc
```diff
@@ -340,16 +351,21 @@ class KernelScope {
...
       node_name_ = node.Name().empty() ? MakeString(node.OpType(), "_", node.Index()) : node.Name();
       concurrency::ThreadPool::StartProfiling(session_state_.GetThreadPool());
       VLOGS(session_state_.Logger(), 1) << "Computing kernel: " << node_name_;
-      kernel_begin_time_ = session_state_.Profiler().Start();
+
+      if (session_profiling_enabled) {
+        kernel_begin_time_ = session_state_.Profiler().Start();
+      } else {
+        kernel_begin_time_ = session_scope_.run_profiler_->Start();
+      }
```
yuslepukhin: Can we wrap this into a SessionScopt method? #Resolved

onnxruntime/core/framework/sequential_executor.cc
```diff
@@ -340,16 +351,21 @@ class KernelScope {
     utils::DumpNodeInputs(dump_context_, kernel_context_, kernel_.Node(), session_state_, session_scope_.dump_analysis_);
 #endif
 
-#ifdef ENABLE_NVTX_PROFILE
-    node_compute_range_.Begin();
-#endif
+    const bool session_profiling_enabled = session_state_.Profiler().IsEnabled();
+    const bool run_profiling_enabled = session_scope_.run_profiler_ && session_scope_.run_profiler_->IsEnabled();
```
yuslepukhin: This can be wrapped into a session scope method IsProfilingEnabled() const. #Resolved

onnxruntime/core/framework/sequential_executor.cc
```diff
@@ -363,26 +379,37 @@ class KernelScope {
     node_compute_range_.End();
 #endif
 
-    if (session_state_.Profiler().IsEnabled()) {
-      auto& profiler = session_state_.Profiler();
+    const bool session_profiling_enabled = session_state_.Profiler().IsEnabled();
+    const bool run_profiling_enabled = session_scope_.run_profiler_ && session_scope_.run_profiler_->IsEnabled();
```
yuslepukhin: IsProfilingEnabled() #Resolved

onnxruntime/core/framework/sequential_executor.cc
yuslepukhin: I have not seen the last member. Can this be a std::array to reduce memory allocations?

onnxruntime/core/framework/sequential_executor.cc
```diff
@@ -363,26 +379,37 @@ class KernelScope {
...
+      if (session_profiling_enabled) {
+        session_state_.Profiler().EndTimeAndRecordEvent(profiling::NODE_EVENT,
+                                                        node_name_ + "_kernel_time",
+                                                        kernel_begin_time_,
+                                                        event_args);
+      } else if (run_profiling_enabled) {
+        session_scope_.run_profiler_->EndTimeAndRecordEvent(profiling::NODE_EVENT,
+                                                            node_name_ + "_kernel_time",
+                                                            kernel_begin_time_,
+                                                            event_args);
```
yuslepukhin: Let's wrap this into a SessionScope method. #Resolved

onnxruntime/core/framework/sequential_executor.cc
```diff
@@ -417,6 +443,8 @@ class KernelScope {
   size_t total_output_sizes_{};
   std::string input_type_shape_;
 
+  TimePoint kernel_begin_time_;
```
yuslepukhin: Curious why the move? #Resolved

onnxruntime/core/session/inference_session.cc
```diff
@@ -128,7 +131,10 @@ inline std::basic_string<T> GetCurrentTimeString() {
 
   T time_str[32];
   OrtStrftime<T>(time_str, sizeof(time_str), GetDateFormatString<T>(), &local_tm);
-  return std::basic_string<T>(time_str);
+
+  std::basic_stringstream<T> ss;
+  ss << time_str << T('_') << std::setfill(T('0')) << std::setw(3) << ms.count();
```
yuslepukhin, line 137: can we do std::format? #Resolved

onnxruntime/core/session/inference_session.cc
```diff
@@ -2925,9 +2931,28 @@ Status InferenceSession::Run(const RunOptions& run_options,
...
+  std::optional<profiling::Profiler> run_profiler;
+  if (run_options.enable_profiling && session_profiler_.IsEnabled()) {
+    LOGS(*session_logger_, WARNING) << "RunOptions requests profiling but session-level profiling is already enabled. "
+                                       "Ignoring run-level profiling request.";
+  }
+  if (run_options.enable_profiling && !session_profiler_.IsEnabled()) {
+    run_profiler.emplace();
+    run_profiler->Initialize(session_logger_);
+    std::basic_ostringstream<ORTCHAR_T> oss;
+    oss << ToPathString(run_options.profile_file_prefix) << "_" << GetCurrentTimeString<ORTCHAR_T>() << ".json";
```
yuslepukhin: >oss << ToPathString(run_options.profile_file_prefix) << "_" << GetCurrentTimeString<ORTCHAR_T>() << ".json"; [](http://example.com/codeflow?start=4&length=108)

I think a simple concat of strings, via .append would be a simpler and faster option. #Resolved

onnxruntime/core/session/inference_session.cc
```diff
@@ -2925,9 +2931,28 @@ Status InferenceSession::Run(const RunOptions& run_options,
...
+    for (auto& ep : execution_providers_) {
+      run_profiler->AddEpProfilers(ep->GetProfiler());
+    }
+    run_profiler->StartProfiling(oss.str());
+  }
+
   TimePoint tp = std::chrono::high_resolution_clock::now();
   if (session_profiler_.IsEnabled()) {
     tp = session_profiler_.Start();
+  } else if (run_profiler && run_profiler->IsEnabled()) {
```
yuslepukhin: >&& run_profiler->IsEnabled() [](http://example.com/codeflow?start=26&length=28)

Is there a case when run_profiler is instantiated but not enabled? #Resolved

onnxruntime/core/session/inference_session.cc
```diff
@@ -2925,9 +2931,28 @@ Status InferenceSession::Run(const RunOptions& run_options,
...
                              const std::vector<OrtDevice>* p_fetches_device_info) {
+  // Ignore run-level profiling request if session-level profiling is already enabled.
+  std::optional<profiling::Profiler> run_profiler;
+  if (run_options.enable_profiling && session_profiler_.IsEnabled()) {
+    LOGS(*session_logger_, WARNING) << "RunOptions requests profiling but session-level profiling is already enabled. "
+                                       "Ignoring run-level profiling request.";
+  }
+  if (run_options.enable_profiling && !session_profiler_.IsEnabled()) {
+    run_profiler.emplace();
+    run_profiler->Initialize(session_logger_);
```
yuslepukhin, line 2994: >run_profiler->Initialize(session_logger_) [](http://example.com/codeflow?start=4&length=41)

Does this return an error code? #Resolved

onnxruntime/core/session/inference_session.cc
```diff
@@ -2925,9 +2934,31 @@ Status InferenceSession::Run(const RunOptions& run_options,
...
+    oss << ToPathString(run_options.profile_file_prefix) << "_" << GetCurrentTimeString<ORTCHAR_T>() << ".json";
+    run_profiler->StartProfiling(oss.str());
+
+    for (auto& ep : execution_providers_) {
+      auto p = ep->GetProfiler(run_options.enable_profiling);
+      if (p) {
+        run_profiler->AddEpProfilers(std::move(p));
+      }
+    }
+  }
```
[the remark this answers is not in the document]
yuslepukhin: I think this is related to checking return codes.

onnxruntime/core/session/inference_session.cc
```diff
@@ -3164,6 +3202,10 @@ Status InferenceSession::Run(const RunOptions& run_options,
     ORT_RETURN_IF_ERROR(Run(run_options, feed_names, feeds, output_names, p_fetches, p_fetches_device_info));
```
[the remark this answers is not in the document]
yuslepukhin: Please, respond to this comment

include/onnxruntime/core/framework/run_options.h
xiaofeihan1: Sure. We don't enable run-level profiling by default.

onnxruntime/core/framework/sequential_executor.cc
xiaofeihan1: Do you mean add  e.g. `StopEvent` in sessionScope and change here to 
```
  if (session_profiling_enabled) {
        session_state_.Profiler().EndTimeAndRecordEvent(profiling::NODE_EVENT,
                                                        node_name_ + "_kernel_time",
                                                        kernel_begin_time_,
                                                        event_args);
    } else if (run_profiling_enabled) {
        session_scope_.StopEvent(profiling::NODE_EVENT,
                                                        node_name_ + "_kernel_time",
                                                        kernel_begin_time_,
                                                        event_args);
```

yuslepukhin: Yes

onnxruntime/core/session/inference_session.cc
xiaofeihan1: No. `Initialize` in `Profiler` doesn't return an error code.
```
void Initialize(const logging::Logger* session_logger);
```

The session profiler doesn’t use error codes either, so here I think this is reasonable.
```
session_profiler_.Initialize(session_logger_);
```

onnxruntime/core/session/inference_session.cc
xiaofeihan1: ONNX Runtime does not yet fully support C++20 on all platforms.

onnxruntime/core/framework/sequential_executor.cc
xiaofeihan1: This change was unintentional. Revert

onnxruntime/core/session/inference_session.cc
xiaofeihan1: No. Change the condition and add ORT_ENFORCE to check.
```
 else if (run_profiler) {
    ORT_ENFORCE(run_profiler->IsEnabled(), "Run profiler should be enabled here.");
    tp = run_profiler->Start();
  }
```

onnxruntime/core/session/inference_session.cc
xiaofeihan1: Used `.append` API.

onnxruntime/core/framework/sequential_executor.cc
xiaofeihan1: Done.

onnxruntime/core/framework/sequential_executor.cc
xiaofeihan1: Thanks. Added `IsRunProfilingEnabled` in SessionScope.

onnxruntime/core/framework/sequential_executor.cc
xiaofeihan1: Thanks. Added `StopEvent` in SessionScope.

yuslepukhin:
Please, comment on all of the Copilot issues before resolving them. #Resolved

onnxruntime/core/common/profiler.h
```diff
@@ -77,7 +77,7 @@ class Profiler {
   void EndTimeAndRecordEvent(EventCategory category,
                              const std::string& event_name,
                              const TimePoint& start_time,
-                             const std::initializer_list<std::pair<std::string, std::string>>& event_args = {},
+                             std::unordered_map<std::string, std::string> event_args = {},
```
yuslepukhin: >td::unordered_map<std::string, std::string> event_args  [](http://example.com/codeflow?start=30&length=55)

InlinedHashMap #Resolved

onnxruntime/core/framework/sequential_executor.cc
```diff
@@ -155,8 +155,8 @@ std::string ComposeSeriesName(const GraphViewer& graph_viewer) {
 class SessionScope {
  public:
   friend class KernelScope;
-  SessionScope(const SessionState& session_state, const ExecutionFrame& frame)
-      : session_state_(session_state)
+  SessionScope(const SessionState& session_state, const ExecutionFrame& frame, profiling::Profiler* run_profiler = nullptr)
```
yuslepukhin: >profiling::Profiler* run_profiler = nullpt [](http://example.com/codeflow?start=79&length=42)

Let; get rid of the default value and pass it explicitly. #Resolved

onnxruntime/core/framework/sequential_executor.cc
```diff
@@ -225,9 +229,15 @@ class SessionScope {
...
 
-    if (session_state_.Profiler().IsEnabled()) {
+    const bool session_profiling_enabled = session_state_.Profiler().IsEnabled();
+    const bool run_profiling_enabled = run_profiler_ && run_profiler_->IsEnabled();
+
+    if (session_profiling_enabled) {
       session_state_.Profiler().EndTimeAndRecordEvent(profiling::SESSION_EVENT, "SequentialExecutor::Execute", session_start_);
+    } else if (run_profiling_enabled) {
+      StopEvent(profiling::SESSION_EVENT, "SequentialExecutor::Execute", session_start_);
     }
```
yuslepukhin: Can we wrap this into a function StopProfilingIfEnabled() #Resolved

onnxruntime/core/framework/sequential_executor.cc
yuslepukhin: InlinedHashMap per coding standards and reduce memory allocations.

onnxruntime/core/framework/sequential_executor.cc
yuslepukhin: I think both of this can be wrapped into StopProfilingIfEnabled(). Lines 417-422.
The same is true for StartProfilingIfEnabled()

onnxruntime/core/session/inference_session.cc
yuslepukhin: I would remove ENFORCE. It increases binary size. If necessary, we need to validate that the profiler is enabled at the point we discover that it needs to be enabled.

onnxruntime/core/session/inference_session.cc
```diff
@@ -3069,6 +3098,12 @@ Status InferenceSession::Run(const RunOptions& run_options,
         ORT_CHECK_AND_SET_RETVAL(status);
       }
 
+      if (run_profiler) {
+        ORT_ENFORCE(run_profiler->IsEnabled(), "Run profiler should be enabled here.");
```
yuslepukhin: >ORT_ENFORCE(run_profiler->IsEnabled( [](http://example.com/codeflow?start=8&length=36)

Let's remove ENFORCE #Resolved

onnxruntime/core/session/inference_session.cc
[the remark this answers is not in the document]
xiaofeihan1: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/microsoft/onnxruntime/pull/26846#discussion_r2667265014)

onnxruntime/core/framework/sequential_executor.cc
xiaofeihan1: Added `StopProfilingIfEnabled` and `StartProfilingIfEnabled` as suggested. Done!

onnxruntime/core/session/inference_session.cc
```diff
@@ -2968,9 +2974,30 @@ Status InferenceSession::Run(const RunOptions& run_options,
...
+    LOGS(*session_logger_, WARNING) << "RunOptions requests profiling but session-level profiling is already enabled. "
+                                       "Ignoring run-level profiling request.";
+  }
+  if (run_options.enable_profiling && !session_profiler_.IsEnabled()) {
+    run_profiler.emplace();
+    run_profiler->Initialize(session_logger_);
+    PathString profile_file = ToPathString(run_options.profile_file_prefix);
+    profile_file.append(ORT_TSTR("_"));
+    profile_file.append(GetCurrentTimeString<ORTCHAR_T>());
+    profile_file.append(ORT_TSTR(".json"));
```
yuslepukhin: Maybe you can make this a function? #Resolved

include/onnxruntime/core/framework/run_options.h
```diff
@@ -35,6 +35,14 @@ struct OrtRunOptions {
...
+
+  // File prefix for profiling result for this run.
+  // The actual filename will be: <profile_file_prefix>_<timestamp>.json
+  // Only used when enable_profiling is true.
+  std::basic_string<ORTCHAR_T> profile_file_prefix = ORT_TSTR("onnxruntime_run_profile");
+
```
yuslepukhin, line 45: Since the API would always set this prefix, we probably do not need this default value. #Resolved

onnxruntime/core/session/inference_session.cc
```diff
@@ -2969,9 +2983,27 @@ Status InferenceSession::Run(const RunOptions& run_options,
...
+  // Ignore run-level profiling request if session-level profiling is already enabled.
+  std::optional<profiling::Profiler> run_profiler;
+  if (run_options.enable_profiling && session_profiler_.IsEnabled()) {
+    LOGS(*session_logger_, WARNING) << "RunOptions requests profiling but session-level profiling is already enabled. "
+                                       "Ignoring run-level profiling request.";
+  }
+  if (run_options.enable_profiling && !session_profiler_.IsEnabled()) {
+    run_profiler.emplace();
+    run_profiler->Initialize(session_logger_);
+    std::basic_string<ORTCHAR_T> profile_file = GenerateProfileFilePath(run_options.profile_file_prefix);
```
yuslepukhin, line 2995: >std::basic_string<ORTCHAR_T> [](http://example.com/codeflow?start=4&length=28)

PathString? #Resolved

onnxruntime/core/framework/utils.cc
```diff
@@ -745,17 +751,16 @@ common::Status ExecuteGraph(const SessionState& session_state,
...
+                            const logging::Logger& logger,
+                            profiling::Profiler* run_profiler) {
+  return ExecuteGraph(session_state, feeds_fetches_manager, feeds, fetches,
+                      execution_mode, run_options.terminate, logger,
 #ifdef ORT_ENABLE_STREAM
                       device_stream_collection_holder,
 #endif
-                      run_options.only_execute_path_to_fetches);
+                      run_options.only_execute_path_to_fetches,
+                      nullptr,
```
yuslepukhin: > nullptr, [](http://example.com/codeflow?start=21&length=9)

Add a comment /* parent_stream */ #Resolved

include/onnxruntime/core/framework/run_options.h
xiaofeihan1: My implementation follows the design of `profile_file_prefix` in session_options. See https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/framework/session_options.h#L128

This seems to be required by the ORT Python API. For example, when a developer calls `run_options.enable_profiling = True` in Python, it only sets `enable_profiling` in `run_options.h`. If there is no default value for `profile_file_prefix`, the resulting profile file prefix will be empty.

onnxruntime/core/session/inference_session.cc
xiaofeihan1: Here I intentionally chose not to use `PathString` and instead used `std::basic_string<ORTCHAR_T>`.
The first reason is that the StartProfiling API already takes this type as its parameter.
The second reason is that, in this case, we are only dealing with a filename rather than a full path.

yuslepukhin (approved):
:shipit:

Issue #26739, opened 2025-12-06
[Feature Request] Support QNN HTA backend through better Op validation


MarwanNIbrahim:
### Describe the feature request

Currently, the [QNN Execution Provider](https://onnxruntime.ai/docs/execution-providers/QNN-ExecutionProvider.html#ep-provider-options), only exposes the `cpu`, `gpu`, `htp`, and `saver` backends through the `backend_type` config option. While other backends are supported through the `backend_path` option (e.g: Qnn DSP backend), I discovered during testing that the Qnn HTA backend does not seem to work, even when enabling the CPU EP fallback (testing done using onnxruntime 1.24.0, and qnn_sdk 2.38.0.250901 / 2.41.0.251128).

Digging a bit closer, the issue seemed to be caused by bad operation validation in the qnn `BaseOpBuilder`. By default, onnxruntime automatically inserts transpose nodes to compensate for the difference in preferred data layout (NCHW vs NHWC). The issue lies in that, currently, transpose nodes are unsupported for the [HTA backend](https://docs.qualcomm.com/doc/80-63442-10/topic/HtaOpDefSupplement.html). Normally, this wouldn't be an issue if the CPU EP fallback is enabled, as the unsupported nodes would just fallback to the CPU EP. Sadly, the `IsOpSupported` validation incorrectly marks the unsupported operations as supported, so that the fallback does not correctly take place

Simply fixing the `IsOpSupported` check to correctly mark unsupported operations as unsupported resolves the issue, allowing users to use the HTA backend

The feature request here is, therefore, to:
- Add better operation validation for the QNN EP (e.g: the QNN sdk offers a `backendGetSupportedOperations`, which can be used to verify if the current operation is supported on the loaded backend)
- Similarly to the `htp` backend, expose the `dsp` and `hta` backends through the `backend_type` config option

### Describe scenario use case

Anywhere where the user would like to use the QNN HTA backend (especially relevant for older chipsets where the hardware required for the HTP backend is not available)

PR #26848, opened 2025-12-22
[QNN EP] Add more detailed operation validation for QNN EP (#26739)

MarwanNIbrahim:
### Description

Provide better operation validation for the different QNN backends through `backendGetSupportedOperations`.

This resolves multiple issues with QNN EP, including:
- Session creation failing if an operation is unsupported on the QNN EP, even if the CPU EP fallback is enabled
- Operation validation not taking into account the specific backend used
- QNN Hta backend failing by default. This is equivalent to the first issue, only that the default inserted transpose operations to handle the different expected input shapes (Nchw vs Nhwc) are unsupported for the Hta backend

Public APIs (C/C++ APIs and corresponding C# native methods) are not touched.

Refer to the issue ticket https://github.com/microsoft/onnxruntime/issues/26739

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/base_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/base_op_builder.cc
@@ -23,6 +23,12 @@ std::string BaseOpBuilder::GetOpBuilderType() const {
 Status BaseOpBuilder::IsOpSupported(QnnModelWrapper& qnn_model_wrapper,
                                     const NodeUnit& node_unit,
                                     const logging::Logger& logger) const {
+  const auto qnn_op_name = GetOptionalQnnOpType(node_unit.OpType());
+  ORT_RETURN_IF_NOT(qnn_op_name.has_value(), "Failed to find equivalent QNN operation for ", node_unit.OpType());
+
+  const bool isOpSupported = qnn_model_wrapper.IsOpSupported(qnn_op_name);
+  ORT_RETURN_IF_NOT(isOpSupported, "Operation ", qnn_op_name, " is unsupported for the current backend");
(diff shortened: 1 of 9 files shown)


MarwanNIbrahim:

quic-ashigarg:
@MarwanNIbrahim we do not support the HTA backend through ORT QNN-EP. Can you share more on the use case and chipset you are trying to use the HTA on.

MarwanNIbrahim:
@quic-ashigarg Regarding my use case, I am using the QRB5165 SoC, which comes with a DSP and HTA, but no HTP (if I understand correctly, that is a separate physical component that the SoC does not have). Supporting the HTA backend allows for running mixed precision models (as DSP only supports 8-bit models) as well as faster inference (from a few simple tests I did with a simple convolutional model, I get up to 60% inference time reduction when using the HTA over the DSP when using the same model).

Note that the above fix, however, fixes a more fundamental problem with the QNN-EP; namely, that the CPU fallback does not behave as expected. Currently, even with the CPU fallback enabled, inference session creation fails for the QNN-EP if:
- An operation in the onnx graph is not in the hardcoded list under `BaseOpBuilder::GetQnnOpType`
- An operation is not implemented by the specific backend (currently assumed that all above operations are supported on all backends, which is simply not the case)

The fix for both issues is added under the QNN [`BaseOpBuilder::IsOpSupported`](https://github.com/microsoft/onnxruntime/pull/26848/files#diff-d21892d7e24907e8a5412c790f102d06b35f8d6e1047dfa622c81c81735f913fR26-R30). With this, I am able to run inference using the HTA backend for ORT QNN-EP (with the CPU fallback enabled to run any unsupported operations)

PR #26853, opened 2025-12-22
[QNN EP] Add RMSNorm Op support in QNN EP

tirupath-qti:
### Description
 - Add standalone RMSNorm op translation in QNN EP
 - Add unit tests

### Motivation and Context
 - This fixes the CPU fallback of ONNX RMSNormalization operator when running inferences using QNN EP

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/rms_norm_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/rms_norm_op_builder.cc
@@ -58,13 +58,10 @@ Status RMSNormOpBuilder::IsOpSupported(QnnModelWrapper& qnn_model_wrapper,
   // Additional constraints for NPU backend
   bool is_npu_backend = IsNpuBackend(qnn_model_wrapper.GetQnnBackendType());
   if (is_npu_backend) {
-    int32_t default_axis = -1;
+    int32_t axis = -1;
     Qnn_Scalar_t axis_qnn_scalar = QNN_SCALAR_INIT;
-    ORT_RETURN_IF_ERROR(ProcessAxisAttribute(qnn_model_wrapper, node_unit, axis_qnn_scalar, default_axis));
-
-    // Normalize negative axis
-    int32_t normalized_axis = default_axis < 0 ? static_cast<int32_t>(input_rank) + default_axis : default_axis;
-    ORT_RETURN_IF(static_cast<size_t>(normalized_axis) != input_rank - 1,
+    ORT_RETURN_IF_ERROR(ProcessAxisAttribute(qnn_model_wrapper, node_unit, axis_qnn_scalar, axis));
+    ORT_RETURN_IF(static_cast<size_t>(axis) != input_rank - 1,
                   "QNN RMSNorm for NPU backend only supports axis with last input dimension");
   }
 
@@ -108,13 +105,8 @@ Status RMSNormOpBuilder::ProcessInputs(QnnModelWrapper& qnn_model_wrapper,
       beta_quant_param = QnnQuantParamsWrapper(quant_scale, zero_point);
     }
 
-    size_t beta_num_elements = 1;
-    for (uint32_t dim : beta_shape) {
-      beta_num_elements *= dim;
-    }
-    const size_t element_size = utils::GetElementSizeByType(beta_data_type);
-    std::vector<uint8_t> beta_data(beta_num_elements * element_size, 0);
-
+    const size_t beta_size_in_bytes = utils::GetQnnTensorDataSizeInBytes(beta_shape, beta_data_type);
+    std::vector<uint8_t> beta_data(beta_size_in_bytes, 0);
     const std::string beta_tensor_name = node_unit.Name() + "_beta_dummy";
     QnnTensorWrapper beta_tensor_wrapper(beta_tensor_name,
                                          QNN_TENSOR_TYPE_STATIC,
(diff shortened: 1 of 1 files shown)


tirupath-qti:
> @tirupath-qti please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> ```
> 
> > Options:
> > 
> > 
> > ```
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > 
> > ```
> > ```
> 

tirupath-qti:

onnxruntime/core/optimizer/qdq_transformer/selectors_actions/shared/utils.cc
```diff
@@ -131,7 +129,8 @@ static const OpVersionsAndSelector::OpVersionsMap GetGemmOpVersionsMap() {
 }
 static const OpVersionsAndSelector::OpVersionsMap GetInstanceAndLayerNormalizationOpVersionsMap() {
   return {{"InstanceNormalization", {}},
-          {"LayerNormalization", {}}};
+          {"LayerNormalization", {}},
+          {"RMSNormalization", {}}};
```
edgchen1: can you provide some background on why the additional op type was added here?

and if a new op type is added here, the name `GetInstanceAndLayerNormalizationOpVersionsMap` is no longer accurate.

tirupath-qti: We need to add a selector for RMSNormalization for fusing QDQ nodes into RMSNormalization Node unit. Since it is of Normalization operator type, instead of adding a new selector function we are trying to leverage the existing one. 
We can add a new selector function for RMSNormalization but it would be very similar to this.

onnxruntime/core/optimizer/qdq_transformer/selectors_actions/qdq_selectors.cc
```diff
@@ -868,6 +833,21 @@ bool ScatterElementsNodeGroupSelector::Check(const GraphViewer& graph_viewer, co
...
+                                              const std::vector<const Node*>& q_nodes) const {
+  if (!CheckQDQNodes(graph_viewer, node, redundant_clip_node, dq_nodes, q_nodes)) {
+    return false;
+  }
+
+  int32_t dt_scale = dq_nodes[1]->InputDefs()[0]->TypeAsProto()->tensor_type().elem_type();
+  int32_t dt_output = q_nodes[0]->OutputDefs()[0]->TypeAsProto()->tensor_type().elem_type();
+
+  // Scale and output need to be the same type.
+  return (dt_scale == dt_output);
```
edgchen1: > Scale and output need to be the same type.

is this already guaranteed by the ONNX spec?

https://github.com/onnx/onnx/blob/69e39205729d6f58867c809423ba75ffef117c92/onnx/defs/nn/defs.cc#L2848-L2849

input "scale" and output "Y" both have the same type constraint "V".

on the other hand, input "X" may have a different type. should the types of "X" and "Y" be checked instead?

onnxruntime/core/providers/qnn/builder/opbuilder/rms_norm_op_builder.cc
```diff
@@ -0,0 +1,191 @@
...
+  // Validate scale input is present
+  constexpr size_t SCALE_IDX = 1;
+  const bool has_scale_input = inputs.size() > SCALE_IDX && inputs[SCALE_IDX].node_arg.Exists();
+  ORT_RETURN_IF_NOT(has_scale_input, "QNN EP requires scale input for RMSNorm operator");
+
+  // Validate input and output rank constraints
+  std::vector<uint32_t> input_shape;
+  ORT_RETURN_IF_NOT(qnn_model_wrapper.GetOnnxShape(inputs[0].node_arg, input_shape), "Cannot get shape of input 0");
+  const size_t input_rank = input_shape.size();
+  ORT_RETURN_IF(input_rank > 4, "QNN RMSNorm only supports input rank <= 4, got rank: %zu", input_rank);
```
edgchen1: > got rank: %zu

`ORT_RETURN_IF` does not take a format string. it concatenates the string representations of its trailing arguments.

onnxruntime/core/optimizer/qdq_transformer/selectors_actions/qdq_selectors.cc
tirupath-qti: I got the same doubt. Ideally Input and output types should be checked. Interestingly the ONNX spec is referring to input "scale: and output "Y" both have the same type constraint "V".
Initially i intended to check input "X" and output "Y".

tirupath-qti: i modified the check.

tirupath-qti (commented):
addressed the review comments in new commit added to this PR.

onnxruntime/core/providers/qnn/builder/opbuilder/rms_norm_op_builder.cc
```diff
@@ -0,0 +1,191 @@
...
+      int32_t zero_point = 0;
+      beta_quant_param = QnnQuantParamsWrapper(quant_scale, zero_point);
+    }
+
+    size_t beta_num_elements = 1;
+    for (uint32_t dim : beta_shape) {
+      beta_num_elements *= dim;
+    }
+    const size_t element_size = utils::GetElementSizeByType(beta_data_type);
+    std::vector<uint8_t> beta_data(beta_num_elements * element_size, 0);
```
edgchen1: nit: can use the helper which uses SafeInt to compute the shape size
```suggestion
    const size_t beta_size_in_bytes = utils::GetQnnTensorDataSizeInBytes(beta_shape, beta_data_type);
    std::vector<uint8_t> beta_data(beta_size_in_bytes, 0);
```

onnxruntime/core/providers/qnn/builder/opbuilder/rms_norm_op_builder.cc
```diff
@@ -0,0 +1,191 @@
...
+  // Process axis attribute and create axes parameter
+  std::vector<uint32_t> input_shape;
+  ORT_RETURN_IF_NOT(qnn_model_wrapper.GetOnnxShape(node_unit.Inputs()[0].node_arg, input_shape), "Cannot get shape of Input 0");
+  const size_t input_rank = input_shape.size();
+  int32_t default_axis = -1;
+  Qnn_Scalar_t axis_qnn_scalar = QNN_SCALAR_INIT;
+  ORT_RETURN_IF_ERROR(ProcessAxisAttribute(qnn_model_wrapper, node_unit, axis_qnn_scalar, default_axis));
+
+  // Normalize negative axis to prevent underflow
+  int32_t normalized_axis = default_axis < 0 ? static_cast<int32_t>(input_rank) + default_axis : default_axis;
```
edgchen1: do we need to validate the range of axis? that's done in this helper function:
https://github.com/microsoft/onnxruntime/blob/5d59608817819879db9721b1be9c349aa1f0d2d9/onnxruntime/core/providers/common.h#L30-L35

tirupath-qti: Seems ProcessAxisAttribute() is already normalizing the axis and always returns axis >=0. So we can avoid redundant normalization logic.

onnxruntime/core/providers/qnn/builder/opbuilder/rms_norm_op_builder.cc
tirupath-qti: Thanks for the suggestion. I will replace the logic with the helper function.

tirupath-qti: Addressed in the new commit added to this PR

onnxruntime/core/providers/qnn/builder/opbuilder/rms_norm_op_builder.cc
tirupath-qti: addressed in new commit added to this PR

tirupath-qti:
@edgchen1 
I addressed the comments. Could you please review and trigger bait.

onnxruntime/core/providers/qnn/builder/opbuilder/rms_norm_op_builder.cc
edgchen1: ah ok, good to know

PR #26855, opened 2025-12-23
[webgpu] refactor initialization of WebGPU Context params

fs-eire:
### Description

refactor initialization of WebGPU Context params

The refactor:
- makes all WebGPU options into ~~3~~ 2 classes:
  - `WebGpuExecutionProviderConfig`: configuration that passed to and stored inside the EP class.
  - ~~`WebGpuContextCreationParams`: configuration that passed to constructor of class `WebGpuContext`.~~
  - ~~`WebGpuContextInitializationParams`: configuration that passed to `WebGpuContext::Initialize()`.~~
  - `WebGpuContextConfig`: configuration that passed to construct and initialize `WebGpuContext`.
- ensure all instance of the classes are created with default value initialized.
- ensure all of the following happens only at one place:
  - setting default value
  - parse option
- add `WebGpuContextFactory::DefaultContext` to allow "get or create" the default context.

### Motivation and Context

- Make code more clean and consistent.

### DIFF
--- a/onnxruntime/core/providers/webgpu/webgpu_context.h
+++ b/onnxruntime/core/providers/webgpu/webgpu_context.h
@@ -34,10 +34,21 @@ struct CapturedCommandInfo {
   WGPUBuffer indirect_buffer;  // WGPUBuffer for indirect dispatch, nullptr if not using indirect dispatch
 };
 
+struct WebGpuBufferCacheConfig {
+  struct ConfigEntry {
+    BufferCacheMode mode;
+    std::string config_string;  // preserved for customized configuration, eg. bucket sizes
+  };
+  ConfigEntry storage{BufferCacheMode::Bucket, {}};
+  ConfigEntry uniform{BufferCacheMode::Simple, {}};
+  ConfigEntry query_resolve{BufferCacheMode::Disabled, {}};
+  ConfigEntry default_entry{BufferCacheMode::Disabled, {}};
+};
+
 /// <summary>
-/// Represents the configuration parameters for creating a WebGpuContext.
+/// Represents the configuration options for creating a WebGpuContext.
 /// </summary>
-struct WebGpuContextCreationParams {
+struct WebGpuContextConfig {
   int context_id{0};
   WGPUInstance instance{nullptr};
   WGPUDevice device{nullptr};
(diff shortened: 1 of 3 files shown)


onnxruntime/core/providers/webgpu/webgpu_context.h
```diff
@@ -34,26 +34,57 @@ struct CapturedCommandInfo {
...
+#endif
+#else
+      0
+#endif
   };
-  ConfigEntry storage;
-  ConfigEntry uniform;
-  ConfigEntry query_resolve;
-  ConfigEntry default_entry;
+  bool enable_pix_capture{false};
```
qjia7, line 81: `enable_pix_capture` is a profiling option. How do you think if we make it per-session instead of per-context?

fs-eire: `enable_pix_capture` controls whether to create an instance of `WebGpuPIXFrameGenerator`. Currently, this instance is per-context. Do you think it is better to create it as per-session? If so, I can make the change to put `enable_pix_capture` as session option.

fs-eire:
Updated the PR to make `WebGpuContextCreationParams` and `WebGpuContextInitializationParams` into a single class `WebGpuContextConfig`.

onnxruntime/core/providers/webgpu/webgpu_context.h
qjia7: Yes, I prefer creating it per‑session since that makes the profiling results easier to analyze.

onnxruntime/core/providers/webgpu/webgpu_context.h
```diff
@@ -34,26 +34,51 @@ struct CapturedCommandInfo {
...
+      webgpu::ValidationMode::Basic  // for release build, enable basic validation by default
+#endif  // !NDEBUG
   };
-  ConfigEntry storage;
-  ConfigEntry uniform;
-  ConfigEntry query_resolve;
-  ConfigEntry default_entry;
+  bool preserve_device{false};
+  uint64_t max_storage_buffer_binding_size{0};
+  WebGpuBufferCacheConfig buffer_cache_config{};
```
qjia7, line 65: `buffer_cache_config` isn’t required for creating a WebGPU context, but it is shared across sessions. From a functionality perspective, do you think it makes more sense to define `buffer_cache_config` as a session option, or should we keep it as a WebGPU‑context option since all sessions share it?

fs-eire: The current implementation of the buffer cache is per-WebGpuContext, so it makes sense to me that the corresponding config is per-WebGpuContext.

We can make some optimization to the buffer cache to further optimize GPU memory usage. But this is going to be a future work item and I am not including it in this PR.

onnxruntime/core/providers/webgpu/webgpu_context.h
fs-eire: Will change it to per-session.

PR #26856, opened 2025-12-23
Add enable_profiling API in RunOptions using Thread-Local Storage

xiaofeihan1:
### Description
Support profiling for specific Run.

Similar to session-level profiling, developers can specify `enable_profiling` and `profile_file_prefix` in RunOptions. After the run completes, the resulting JSON file will be saved with the given profile_file_prefix + timestamp.

When profiling is enabled via RunOptions, it should ideally collect two types of events:
1. Profiler events
Used to calculate the total execution time of each operator.
2. Execution Provider (EP) profiler events
Used to measure GPU kernel execution time. 

Unlike session-level, we need to ensure the collecting events is correct for multiple thread scenario.

For 1, this can be supported easily(sequential_executor.cc). We use a thread-local storage (TLS) variable, RunLevelState (defined in profiler.h), to maintain run-level profiling state for each thread.

For 2, each Execution Provider (EP) has its own profiler implementation, and each EP must ensure correct behavior under run-level profiling. This PR ensures that the WebGPU profiler works correctly with run-level profiling.

| Scenario | Example | Expected Result |
|---------|---------|-----------------|
| Concurrent runs on the same session with different run-level profiling settings| t1: `sess1.Run({ enable_profiling: true })`<br>t2: `sess1.Run({ enable_profiling: false })`<br>t3: `sess1.Run({ enable_profiling: true })` | Two trace JSON files are generated: one for `t1` and one for `t3`. |
| Run-level profiling enabled together with session-level profiling| `sess1 = OrtSession({ enable_profiling: true })`<br>`sess1.Run({ enable_profiling: true })` | Two trace JSON files are generated: one corresponding to session-level profiling and one corresponding to run-level profiling. |

### Motivation and Context
Previously, profiling only for session level. Sometimes developer want to profile for specfic run . so the PR comes.

### DIFF
--- a/include/onnxruntime/core/common/gpu_profiler_common.h
+++ b/include/onnxruntime/core/common/gpu_profiler_common.h
@@ -443,6 +443,8 @@ class GPUProfilerBase : public EpProfiler {
     return true;
   }
 
+  virtual void StartRunProfiling() override {}
+
   virtual void EndProfiling(TimePoint start_time, Events& events) override {
     auto& manager = TManager::GetInstance();
     std::map<uint64_t, Events> event_map;
--- a/include/onnxruntime/core/common/profiler_common.h
+++ b/include/onnxruntime/core/common/profiler_common.h
@@ -85,7 +85,7 @@ class EpProfiler {
   virtual void Stop(uint64_t) {}                                        // called after op stop, accept an id as argument to identify the op
 
   // Run-level profiling support
-  virtual void StartRunProfiling();
+  virtual void StartRunProfiling() {}
   virtual void EndRunProfiling(TimePoint start_time, Events& events) {
     ORT_UNUSED_PARAMETER(start_time);
     ORT_UNUSED_PARAMETER(events);
--- a/onnxruntime/core/common/profiler_common.cc
+++ b/onnxruntime/core/common/profiler_common.cc
@@ -3,8 +3,6 @@
 
 #include "core/common/profiler_common.h"
 
-#include "core/common/logging/logging.h"
-
 #ifndef _MSC_VER
 #include <cxxabi.h>
 #endif
@@ -14,9 +12,7 @@
 namespace onnxruntime {
 namespace profiling {
 
-void EpProfiler::StartRunProfiling() {
-  LOGS_DEFAULT(WARNING) << "Run-level profiling is not implemented for this EP profiler, so GPU kernel events for this EP cannot be captured.";
-}
+
 
 static constexpr int kMaxSymbolSize = 1024;
 
--- a/onnxruntime/core/providers/cuda/cuda_profiler.cc
+++ b/onnxruntime/core/providers/cuda/cuda_profiler.cc
@@ -6,6 +6,7 @@
 #include <iostream>
 
 #include "cuda_profiler.h"
+#include "core/common/logging/logging.h"
 
 namespace onnxruntime {
 namespace profiling {
@@ -20,6 +21,16 @@ CudaProfiler::CudaProfiler() {
 CudaProfiler::~CudaProfiler() {
   auto& manager = CUPTIManager::GetInstance();
   manager.DeregisterClient(client_handle_);
+}
+
+void CudaProfiler::StartRunProfiling() {
+  LOGS_DEFAULT(WARNING) << "Run-level profiling is not implemented for this EP profiler, so GPU kernel events for this EP cannot be captured.";
+}
+
+#else
+
+void CudaProfiler::StartRunProfiling() {
+  LOGS_DEFAULT(WARNING) << "Run-level profiling is not implemented for this EP profiler, so GPU kernel events for this EP cannot be captured.";
 }
 
 #endif /* #if defined(USE_CUDA) && defined(ENABLE_CUDA_PROFILING) */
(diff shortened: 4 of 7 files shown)


onnxruntime/core/common/profiler.cc
```diff
@@ -157,5 +169,79 @@ std::string Profiler::EndProfiling() {
...
+  for (const auto& ep_profiler : ep_profilers_) {
+    ep_profiler->EndRunProfiling(run_level_state_.start_time, run_level_state_.events);
+  }
+
+  run_level_state_.enabled = false;
+
+  if (file_path.empty()) {
+    run_level_state_.events.clear();
+    return std::string();
+  }
```
qjia7, line 204: Move L201-L204 before collecting EP profiler events?

onnxruntime/core/session/inference_session.cc
```diff
@@ -3138,6 +3143,14 @@ Status InferenceSession::Run(const RunOptions& run_options,
   if (session_profiler_.IsEnabled()) {
```
qjia7, line 3143: Use `if (session_profiler_.IsEnabled() || run_options.enable_profiling)`?

onnxruntime/core/providers/webgpu/webgpu_profiler.h
```diff
@@ -21,13 +21,33 @@ class WebGpuProfiler final : public onnxruntime::profiling::EpProfiler {
   }
   void Stop(uint64_t) override {
   }
+
+  // Run-level profiling support
+  void StartRunProfiling() override;
+  void EndRunProfiling(TimePoint start_time, onnxruntime::profiling::Events& events) override;
```
qjia7, line 27: Can we reuse `StartProfiling` and `EndProfiling` no matter it's session level or run level? If it's session-level, just pass session-level's events to EndProfiling. Otherwise, pass run-level events to EndProfiling. Can it satisfy the requirement?

include/onnxruntime/core/common/gpu_profiler_common.h
```diff
@@ -443,6 +443,8 @@ class GPUProfilerBase : public EpProfiler {
     return true;
   }
 
+  virtual void StartRunProfiling() override {}
```
yuslepukhin, line 446: >  virtual vo [](http://example.com/codeflow?start=0&length=12)

Is this necessary?

include/onnxruntime/core/common/profiler_common.h
```diff
@@ -83,6 +83,14 @@ class EpProfiler {
   virtual void EndProfiling(TimePoint start_time, Events& events) = 0;  // called when profiling ends, save all captures numbers to "events"
   virtual void Start(uint64_t) {}                                       // called before op start, accept an id as argument to identify the op
   virtual void Stop(uint64_t) {}                                        // called after op stop, accept an id as argument to identify the op
+
+  // Run-level profiling support
+  virtual void StartRunProfiling() {}
+  virtual void EndRunProfiling(TimePoint start_time, Events& events) {
+    ORT_UNUSED_PARAMETER(start_time);
+    ORT_UNUSED_PARAMETER(events);
```
yuslepukhin, line 91: Suggest simply commenting out param names

onnxruntime/core/common/profiler.cc
```diff
@@ -157,5 +169,79 @@ std::string Profiler::EndProfiling() {
...
+  for (size_t i = 0; i < run_level_state_.events.size(); ++i) {
+    auto& rec = run_level_state_.events[i];
+    out << R"({"cat" : ")" << event_category_names_[rec.cat] << "\",";
+    out << "\"pid\" :" << rec.pid << ",";
+    out << "\"tid\" :" << rec.tid << ",";
+    out << "\"dur\" :" << rec.dur << ",";
+    out << "\"ts\" :" << rec.ts << ",";
+    out << R"("ph" : "X",)";
+    out << R"("name" :")" << rec.name << "\",";
+    out << "\"args\" : {";
```
yuslepukhin, line 224: Could this be a separate operator<<(std::ostream&) for the record or function?

onnxruntime/core/common/profiler_common.cc
```diff
@@ -12,6 +12,8 @@
 namespace onnxruntime {
 namespace profiling {
 
+
+
 static constexpr int kMaxSymbolSize = 1024;
```
yuslepukhin, line 17: This file does not seem to have changes.

yuslepukhin:
Here is the big question. Can we have things pertaining to RunOptions profiling stored within RunOptions and not TLS?
It may be possible to instantiate a separate profiler instance within RunOptions when enabled, then add an API to save to a file OR even retrieve in memory?
The in memory thing is optional, but having everything in RunOptions would be swell. Not having to do many changes in the profiler.

xiaofeihan1:
> Here is the big question. Can we have things pertaining to RunOptions profiling stored within RunOptions and not TLS? It may be possible to instantiate a separate profiler instance within RunOptions when enabled, then add an API to save to a file OR even retrieve in memory? The in memory thing is optional, but having everything in RunOptions would be swell. Not having to do many changes in the profiler.

Thanks for the suggestion! I’m re-implementing this using your proposed solution. I’ll let you know once it’s done.

xiaofeihan1:
Close this PR because we will use another solution. See https://github.com/microsoft/onnxruntime/pull/26846

PR #26857, opened 2025-12-23
WebGPU: Use dedicated prepack allocator for kernel prepacking

jchen10:
Remove the CreateUnmappedGPUTensor workaround by using a proper prepack allocator that creates unmapped GPU buffers directly to avoid the need to manually unmap buffers after allocation.


jchen10:
@qjia7  @fs-eire @guschmue PTAL

fs-eire (approved):

PR #26872, opened 2025-12-29
[QNN EP] Fix the build error when the specified QNN library is missing

kuanyul-qti:
### Description
Use `file(GLOB ...)` to allow for the specified file to not exist.

### Motivation and Context
If the QNN library is missing, using `list(APPEND ...)` will cause errors during the build process.

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -891,11 +891,12 @@ if (onnxruntime_USE_QNN OR onnxruntime_USE_QNN_INTERFACE)
                                              "${onnxruntime_QNN_HOME}/lib/${QNN_ARCH_ABI}/libHtpPrepare.so"
                                              "${onnxruntime_QNN_HOME}/lib/${QNN_ARCH_ABI}/HtpPrepare.dll")
       if (${QNN_ARCH_ABI} STREQUAL "aarch64-windows-msvc" OR ${QNN_ARCH_ABI} STREQUAL "arm64x-windows-msvc")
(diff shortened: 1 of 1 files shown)


cmake/CMakeLists.txt
```diff
@@ -891,11 +891,12 @@ if (onnxruntime_USE_QNN OR onnxruntime_USE_QNN_INTERFACE)
                                              "${onnxruntime_QNN_HOME}/lib/${QNN_ARCH_ABI}/libHtpPrepare.so"
                                              "${onnxruntime_QNN_HOME}/lib/${QNN_ARCH_ABI}/HtpPrepare.dll")
       if (${QNN_ARCH_ABI} STREQUAL "aarch64-windows-msvc" OR ${QNN_ARCH_ABI} STREQUAL "arm64x-windows-msvc")
-        list(APPEND QNN_LIB_FILES "${onnxruntime_QNN_HOME}/lib/hexagon-v68/unsigned/libQnnHtpV68Skel.so"
+        file(GLOB EXTRA_HTP_LIB LIST_DIRECTORIES false "${onnxruntime_QNN_HOME}/lib/hexagon-v68/unsigned/libQnnHtpV68Skel.so"
```
edgchen1, line 894: which file is not being found, and is that expected?

kuanyul-qti: `libqnnhtpv73.cat` and `libqnnhtpv81.cat` may not exist at the same time. This situation occurs during internal testing when only one of the hexagon targets is being focused on, and it's expected.

edgchen1 (approved):

PR #26879, opened 2025-12-30
[EP API] extract common code for EP API adapter

fs-eire:
### Description

This PR generalizes the class `OpKernelInfo` and `OpKernelContext` using template so that the code is able to be used on a different type, which helps for future changes that supports migration of WebGPU EP and CUDA EP to the EP API implementation.

Currently only applies to base classes that is reused by WebGPU EP.


onnxruntime/core/providers/cpu/nn/pool_base.h
```diff
@@ -102,13 +102,15 @@ class LpPool {
...
+  static int GetStartVersion(const KernelInfoType& info) {
     return info.node().SinceVersion();
   }
 
  protected:
-  PoolBase(const OpKernelInfo& info)
-      : op_name_(info.GetKernelDef().OpName().rfind("QLinear", 0) != 0 ? info.GetKernelDef().OpName() : info.GetKernelDef().OpName().substr(7)),
+  template <typename KernelInfoType>
+  PoolBase(const KernelInfoType& info)
+      : op_name_(info.node().OpType().rfind("QLinear", 0) != 0 ? info.node().OpType() : info.node().OpType().substr(7)),
```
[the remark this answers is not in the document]
fs-eire: `info.GetKernelDef().OpName()` and `info.node().OpType()` suppose to return the same string representing the operator type.

onnxruntime/core/providers/cpu/tensor/padbase.h
```diff
@@ -148,19 +196,16 @@ class PadBase {
...
-    if (start_ver >= 11 || kernel_def.Domain() == kMSDomain) {
+    if (info.node().SinceVersion() >= 11 || kernel_def.Domain() == kMSDomain) {
       is_dynamic_ = true;
     }
 
     if (!is_dynamic_) {
-      gsl::span<const int64_t> pads_span;
-      if (!info.GetAttrsAsSpan("pads", pads_span).IsOK())
+      std::vector<int64_t> pads_attr;
+      if (!info.GetAttrs("pads", pads_attr).IsOK())
```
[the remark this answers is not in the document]
fs-eire: An equivalent implementation of `GetAttrsAsSpan` is not able to be done because ORT C API defines the behavior of function `KernelInfoGetAttributeArray_float` to **write to the buffer that user specified** instead of **get the memory address that stores the data**.

As a result, we need to change the code to use `GetAttrs` which writes to a vector. Not a big performance concern because the length of `pads` are usually small enough.

onnxruntime/contrib_ops/cpu/bert/attention_base.h
```diff
@@ -32,22 +32,22 @@ class AttentionBase {
                      int& past_sequence_length) const;
 
  protected:
-  AttentionBase(const OpKernelInfo& info, bool require_same_hidden_size) {
+  template <typename KernelInfoType>
```
adrianlizarraga, line 35: We should probably have some comment or documentation somewhere that explains the need for this templated type. Someone reading this code may find this confusing because normally a `OrtKernelInfo` can be `reinterpret_cast` to a `onnxruntime::OpKernelInfo`. However, if I'm not mistaken, this templated type is meant to be either a `OrtKernelInfo` or a `adapter::OpKernelInfo` from https://github.com/microsoft/onnxruntime/pull/26919.

I'm not sure where this documentation should exist but I think it should probably exist somewhere.

fs-eire: the `KernelInfoType` should be either a `onnxruntime::OpKernelInfo` or a `onnxruntime::ep::adapter::OpKernelInfo`. It's a little different from using `OrtKernelInfo`. `onnxruntime::ep::adapter::OpKernelInfo` is a wrapper class of `OrtKernelInfo` which implement identical facade of `onnxruntime::OpKernelInfo`

It can be documented in [include/onnxruntime/ep/README.md](https://github.com/microsoft/onnxruntime/pull/26919/changes#diff-a3712d038a49c0492ba3688f08edeb978f1ab4ae49fccddcb29cc13a4e896060).

edgchen1: I also found it a bit confusing.

would it be possible to use a distinct type that can be instantiated from both a `onnxruntime::OpKernelInfo` and a `onnxruntime::ep::adapter::OpKernelInfo` instead of using a template that's effectively limited to those specific types?

fs-eire: I cannot think of a very elegant solution. (the current change is all that I can think of)

I am open to suggestions.

fs-eire: EDIT: some potential approaches:

### classic OOP solutions:

- extract an interface (actually abstract class in C++, please ignore the naming) `onnxruntime::IKernelInfo`, and both `onnxruntime::OpKernelInfo` `onnxruntime::ep::adapter::OpKernelInfo` inherit from it. Use `const onnxruntime::IKernelInfo&` replacing `const onnxruntime::OpKernelInfo&`

- making `onnxruntime::OpKernelInfo` itself **the interface**. Move implementation to somewhere like `onnxruntime::OpKernelInfoImpl` that inherit from `onnxruntime::OpKernelInfo`.

**Both approaches above introduces extra runtime cost (vtable lookup) in hot path.**

### compile time polymorphism solutions:

- the current implementation - minimal changes.

- using library like [proxy](https://github.com/microsoft/proxy)
  may be IDE unfriendly

- using `concept`: but requires c++20

edgchen1: hm, I can't think of any good alternatives now. C++20 concepts would be nice. hopefully we can upgrade and use them soon.

edgchen1: let's document it in the readme.

PR #26895, opened 2026-01-02
[MLAS] Fix building on FreeBSD / powerpc64le

pkubaj:
### Description

1. platform.cpp missed inclusion of sys/auxv.h (for elf_aux_info)
and machine/cpu.h (for PPC_FEATURE2_ARCH_3_00). I missed that in my
previous commit.
2. Same as on AIX, __vector int32_t is not defined and __vector int
needs to be used.

### Motivation and Context

Fixes build on FreeBSD / powerpc64le platform.


pkubaj:

PR #26896, opened 2026-01-02
Fix TreeEnsemble LOGISTIC post_transform for single target regression

### LINKED ISSUE #26596: TreeEnsemble post_transform LOGISTIC does not work
### Describe the issue

According to https://onnx.ai/onnx/operators/onnx_aionnxml_TreeEnsemble.html the following `post_transform` options are available for the `TreeEnsemble` node:
- `NONE` (0)
- `SOFTMAX` (1)
- `LOGISTIC` (2)
- `SOFTMAX_ZERO` (3)
- `PROBIT` (4)

However, when I create a simple model with one target the `LOGISTIC` transform is not used.

### To reproduce

I've attached a minimal example of a `TreeEnsemble` with `post_transform` set to logistic.
[tree_ensemble.onnx.zip](https://github.com/user-attachments/files/23609597/tree_ensemble.onnx.zip)

Code to reproduce the issue:
```python
import onnxruntime as ort

session = ort.InferenceSession("tree_ensemble.onnx")
print(session.run(None, {"X": [[1, 0, 1, 0]]}))
```

This prints `[array([[10.3366785]], dtype=float32)]` while I would have expected a number between 0 and 1.

### Urgency

_No response_

### Platform

Mac

### OS Version

26.1

### ONNX Runtime Installation

Released Package

1.23.2

### ONNX Runtime API

Python

### Architecture

ARM64

### Execution Provider

Default CPU

_No response_

sreya-rc:
Fixes #26596

- Updated FinalizeScores1 methods in TreeAggregator, TreeAggregatorSum, and TreeAggregatorAverage to handle LOGISTIC transform
- Added test case TreeEnsembleSingleTargetLogistic to verify correct behavior

### Description
- The LOGISTIC post-transform was not being applied to single-target TreeEnsemble regression models. The `FinalizeScores1` methods only checked for PROBIT transform and ignored LOGISTIC, causing models to output raw scores instead of applying the logistic function.

### Motivation and Context
- According to the ONNX ML spec, TreeEnsemble operators should support LOGISTIC post-transform. However, for single-target regression models, only PROBIT was being applied. This caused models with `post_transform=LOGISTIC` to return raw prediction scores (10.0) instead of transformed values (0.9999546).

### DIFF
--- a/onnxruntime/core/providers/cpu/ml/tree_ensemble_aggregator.h
+++ b/onnxruntime/core/providers/cpu/ml/tree_ensemble_aggregator.h
@@ -223,9 +223,18 @@ class TreeAggregator {
 
   void FinalizeScores1(OutputType* Z, ScoreValue<ThresholdType>& prediction, int64_t* /*Y*/) const {
     prediction.score = prediction.has_score ? (prediction.score + origin_) : origin_;
-    *Z = this->post_transform_ == POST_EVAL_TRANSFORM::PROBIT
-             ? static_cast<OutputType>(ComputeProbit(static_cast<float>(prediction.score)))
-             : static_cast<OutputType>(prediction.score);
+    switch (this->post_transform_) {
+      case POST_EVAL_TRANSFORM::PROBIT:
+        *Z = static_cast<OutputType>(ComputeProbit(static_cast<float>(prediction.score)));
+        break;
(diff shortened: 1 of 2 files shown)


sreya-rc:

cbourjau:
It would be great to merge this. While it is easy to work around this when creating a new model, it is still a very annoying and dangerous paper cut.

PR #26897, opened 2026-01-02
Reduce GQA cpu test combinations

tianleiwu:
The current testing strategy for GQA on CPU attempts to run a Cartesian product of all configuration parameters (batch size, sequence length, rotary embeddings, packed KV, softcap, etc.), leading to over 2000 test combinations. This causes significant runtime overhead and potential timeouts.

This PR optimizes `test_gqa_cpu.py` by:
- Replacing the nested loop over all parameters with a round-robin selection strategy (`combo_index`).
- Significantly reducing the number of test cases (from ~2304 to ~32 in pipeline mode) while maintaining coverage of individual features (rotary, packed, softcap, etc.).

This ensures the test suite remains robust but much faster. 
It reduces test time from minutes to seconds, and saves lot of compute resource in CI pipeline.

### DIFF
--- a/onnxruntime/test/python/transformers/test_gqa_cpu.py
+++ b/onnxruntime/test/python/transformers/test_gqa_cpu.py
@@ -2491,90 +2491,105 @@ class TestGQA(unittest.TestCase):
             print(
                 f"\nRunning tests with precision: {'FLOAT16' if precision['ort_type'] == TensorProto.FLOAT16 else 'FLOAT32'}"
             )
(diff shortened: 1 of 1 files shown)


nenad1002:
Doesn't this reduce the total coverage?

nenad1002 (approved):

tianleiwu:
> Doesn't this reduce the total coverage?

It remains enough test coverage in CI pipeline. Developer could run with PIPELINE_MODE=0 to run more test cases if needed.

PR #26803, opened 2025-12-15
[EP ABI] Add OrtKernelInfo APIs for kernel-based plugin EPs

adrianlizarraga:
### Description
Adds new `OrtKernelInfo` APIs to enable plugin EP kernel implementations to retrieve operator metadata and access the parent EP instance.
- Added four new KernelInfo APIs: `KernelInfo_GetOperatorDomain()`, `KernelInfo_GetOperatorType()`, `KernelInfo_GetOperatorSinceVersion()`, and `KernelInfo_GetEp()`
- Extended example plugin EP to demonstrate accessing EP configuration from kernel implementations.

PR #26907, opened 2026-01-05
Make WebGPU EP compatible with EP API

fs-eire:
### Description

This PR makes it possible to build WebGPU EP as an EP API based plugin EP.

#### Requirements

The goal of this PR is to support both building WebGPU EP as a bundled EP and an EP API based plugin EP. This approach allows:
- enabling WebGPU EP as a standalone plugin EP package for WCR usage
- graceful transition for WebGPU EP as an native EP for language binding, from the bundled EP to an EP API based plugin EP
- keep the existing usage (static library) working (majorly for web)

#### Design & Implementation

Instead of **changing** WebGPU EP from a bundled EP to an EP API based plugin EP in one shot, this PR  **extend** WebGPU EP to support building as plugin EP.

- add a new folder `include/onnxruntime/ep` with a bunches of header files. Those files are not WebGPU specific. They are used for:
  - include common defines/functions/macros for plugin EP to use
  - include a few "adapter" classes that takes C-API objects to simulate ORT internal classes behaviors
  - include a few "override" classes that simulate ORT internal classes, but using implementations that only depend on C-API
  - include a special base class `onnxruntime::ep::Ep` to inherit from
  
  These header files allow a compile time "switch" to the different set of types to minimize changes to existing code. Specifically, `pch.h` is required to be included as PCH to make sure the "override" to take place correctly.

- add a new folder `onnxruntime/core/providers/webgpu/ep` for EP API implementation, specifically:
  - `api.cc`: implements `CreateEpFactories` and `ReleaseEpFactory`
  - `ep.cc` `ep.h`: implement class `onnxruntime::webgpu::ep::Ep`
  - `factory.cc` `factory.h`: implement class `onnxruntime::webgpu::ep::Factory`

#### Dependencies and Prerequisites

(unmerged changes are included as a part of current PR)

- https://github.com/microsoft/onnxruntime/pull/26855
- https://github.com/microsoft/onnxruntime/pull/26803
- https://github.com/microsoft/onnxruntime/pull/26859
- https://github.com/microsoft/onnxruntime/pull/26879
- https://github.com/microsoft/onnxruntime/pull/26919
- https://github.com/microsoft/onnxruntime/pull/27569
- https://github.com/microsoft/onnxruntime/pull/27587

#### Missing Parts

- Allow setting Global/Default EP options

### DIFF
--- a/onnxruntime/test/unittest_util/base_tester.cc
+++ b/onnxruntime/test/unittest_util/base_tester.cc
@@ -749,10 +749,8 @@ void BaseTester::RunWithConfig(size_t* number_of_pre_packed_weights_counter,
           execution_provider = DefaultXnnpackExecutionProvider();
         else if (provider_type == onnxruntime::kDmlExecutionProvider)
           execution_provider = DefaultDmlExecutionProvider();
-#if !defined(USE_WEBGPU) || !defined(ORT_USE_EP_API_ADAPTERS)
         else if (provider_type == onnxruntime::kWebGpuExecutionProvider)
           execution_provider = DefaultWebGpuExecutionProvider();
-#endif
         else if (provider_type == dynamic_plugin_ep_name) {
           execution_provider = dynamic_plugin_ep_infra::MakeEp();
         }


tianleiwu:
## AI Summary

The overall direction is good: the PR adds a real EP API surface for WebGPU instead of relying on the legacy internal factory path, and most of the capability logic is intentionally mirrored from the native WebGPU EP. The blocking issue is that the new plugin path is still hard-wired to the default WebGPU context/device, so non-default device selection and the shared environment plumbing do not actually work once you leave the device-0 happy path.

---

## Review

### 1. Device Routing And Allocator Metadata (`onnxruntime/core/providers/webgpu/ep/factory.cc`)

**Positive**:
- The factory cleanly exposes the expected EP API entry points and reuses the existing `WebGpuProviderFactoryCreator` instead of duplicating provider construction logic.

**Concern**:
- ⚠️ **Selected device is dropped and all EP devices are advertised as device 0**: the factory constructor bakes `config_` against `WebGpuContextFactory::DefaultContext()` and creates both exported `OrtMemoryInfo`s with device id `0` ([onnxruntime/core/providers/webgpu/ep/factory.cc](https://github.com/microsoft/onnxruntime/blob/5ceb23cf6af9efdcc51aa5f90f5e303adb9e1bde/onnxruntime/core/providers/webgpu/ep/factory.cc#L23), [onnxruntime/core/providers/webgpu/ep/factory.cc](https://github.com/microsoft/onnxruntime/blob/5ceb23cf6af9efdcc51aa5f90f5e303adb9e1bde/onnxruntime/core/providers/webgpu/ep/factory.cc#L29)). `GetSupportedDevicesImpl` then returns every GPU with those same fixed memory infos and no per-device metadata/options ([onnxruntime/core/providers/webgpu/ep/factory.cc](https://github.com/microsoft/onnxruntime/blob/5ceb23cf6af9efdcc51aa5f90f5e303adb9e1bde/onnxruntime/core/providers/webgpu/ep/factory.cc#L75)). Finally, `CreateEpImpl` ignores the `devices` / `ep_metadata` arguments completely and recreates the provider only from session config entries ([onnxruntime/core/providers/webgpu/ep/factory.cc](https://github.com/microsoft/onnxruntime/blob/5ceb23cf6af9efdcc51aa5f90f5e303adb9e1bde/onnxruntime/core/providers/webgpu/ep/factory.cc#L104)). ORT derives the plugin EP's `OrtDevice` from `OrtEpDevice.device_memory_info` ([onnxruntime/core/session/plugin_ep/ep_plugin_provider_interfaces.cc](https://github.com/microsoft/onnxruntime/blob/5ceb23cf6af9efdcc51aa5f90f5e303adb9e1bde/onnxruntime/core/session/plugin_ep/ep_plugin_provider_interfaces.cc#L119)), so selecting device 1 still produces a provider/allocator surface that claims to be device 0 and allocates against the default context. That breaks the core EP API contract for multi-device or externally supplied WebGPU devices.

### 2. Shared Data Transfer (`onnxruntime/core/providers/webgpu/webgpu_provider_factory.cc`)

**Positive**:
- Reusing `DataTransfer::CopyTensorImpl` avoids duplicating the actual copy logic, and the vendor-id filter is a good guard against accidentally claiming unrelated GPU tensors.

**Concern**:
- ⚠️ **Environment data transfer is still hard-coded to the default context**: the EP factory always creates transfers with `OrtWebGpuCreateDataTransfer()` and never passes a selected context/device id ([onnxruntime/core/providers/webgpu/ep/factory.cc](https://github.com/microsoft/onnxruntime/blob/5ceb23cf6af9efdcc51aa5f90f5e303adb9e1bde/onnxruntime/core/providers/webgpu/ep/factory.cc#L169)). The implementation then rejects GPU copies unless both memory devices match `context_id_` ([onnxruntime/core/providers/webgpu/webgpu_provider_factory.cc](https://github.com/microsoft/onnxruntime/blob/5ceb23cf6af9efdcc51aa5f90f5e303adb9e1bde/onnxruntime/core/providers/webgpu/webgpu_provider_factory.cc#L328)), and lazy initialization explicitly errors out for any `context_id_ != 0` before creating `DefaultContext()` ([onnxruntime/core/providers/webgpu/webgpu_provider_factory.cc](https://github.com/microsoft/onnxruntime/blob/5ceb23cf6af9efdcc51aa5f90f5e303adb9e1bde/onnxruntime/core/providers/webgpu/webgpu_provider_factory.cc#L354)). Even after fixing device selection, shared copies, env allocators, and `IOBinding`-style CPU<->GPU transfers for device 1+ will still fail because the only supported transfer object is pinned to device 0.

### 3. Capability Selection Parity (`onnxruntime/core/providers/webgpu/ep/ep.cc`)

**Positive**:
- The EP API capability implementation correctly carries over the WebGPU-specific unsupported `Attention` cases, which keeps most of the selection logic aligned with the native provider.

**Concern**:
- ⚠️ **Already-assigned WebGPU nodes can now be demoted back to CPU in the plugin build**: the native WebGPU EP keeps two lists, `candidates` and `tenative_candidates`, and only passes the tentative set to `GetCpuPreferredNodes`, specifically so layout-transformer-inserted WebGPU nodes are never reconsidered for CPU fallback ([onnxruntime/core/providers/webgpu/webgpu_execution_provider.cc](https://github.com/microsoft/onnxruntime/blob/5ceb23cf6af9efdcc51aa5f90f5e303adb9e1bde/onnxruntime/core/providers/webgpu/webgpu_execution_provider.cc#L963), [onnxruntime/core/providers/webgpu/webgpu_execution_provider.cc](https://github.com/microsoft/onnxruntime/blob/5ceb23cf6af9efdcc51aa5f90f5e303adb9e1bde/onnxruntime/core/providers/webgpu/webgpu_execution_provider.cc#L1026)). The new EP API version pushes nodes whose `ep_name` is already `WebGpuExecutionProvider` into `candidate_nodes` and then passes that whole vector into `GetCpuPreferredNodes` ([onnxruntime/core/providers/webgpu/ep/ep.cc](https://github.com/microsoft/onnxruntime/blob/5ceb23cf6af9efdcc51aa5f90f5e303adb9e1bde/onnxruntime/core/providers/webgpu/ep/ep.cc#L69), [onnxruntime/core/providers/webgpu/ep/ep.cc](https://github.com/microsoft/onnxruntime/blob/5ceb23cf6af9efdcc51aa5f90f5e303adb9e1bde/onnxruntime/core/providers/webgpu/ep/ep.cc#L138)). That creates plugin-only partitioning behavior where layout-transformed nodes can be excluded from `EpGraphSupportInfo_AddSingleNode`, leaving nodes that were already assigned to WebGPU stranded or re-routed differently from the non-plugin build.

---

## Summary of Concerns

| 1 | High | EP factory | Device selection is ignored and all allocators/memory infos are exported as device 0. |
| 2 | High | Data transfer | Shared WebGPU data transfer only works for context/device 0. |
| 3 | High | Capability selection | Preassigned WebGPU nodes are incorrectly included in CPU-preference fallback. |

---

## Verdict

REQUEST CHANGES — the EP API path is still functionally tied to the default WebGPU device, and it also regresses capability selection behavior versus the native WebGPU EP.

fs-eire:
Explanation of (1) and (2) from @tianleiwu's review:

ORT device discovery never worked with WebGPU, majorly because of 2 reasons:
- WebGPU EP already use its own sematic of "Device ID" (or Context ID, which are the same concept in the context of WebGPU). Specifically, "Device 0" represents the default device, and all other number other than 0 represents a custom device, which is expected to be created and passed in by user. This is before the Device discovery feature introduced into ORT.
- WebGPU has its own implementation for device discovery and selection, and it cannot work together with ORT's device discovery feature. the `OrtDevice` info is not helpful for creating a webgpu device.

Based on this fact, we create at most one WebGPU device inside WebGPU EP.

The real "missing" features are the webgpu options. They are planned to be supported in future changes.

onnxruntime/core/providers/webgpu/ep/factory.cc
```diff
@@ -0,0 +1,213 @@
...
+const char* ORT_API_CALL Factory::GetNameImpl(const OrtEpFactory* /*this_ptr*/) noexcept {
+  return kWebGpuExecutionProvider;
+}
+
+const char* ORT_API_CALL Factory::GetVendorImpl(const OrtEpFactory* /*this_ptr*/) noexcept {
+  return "Microsoft";
+}
+
+uint32_t ORT_API_CALL Factory::GetVendorIdImpl(const OrtEpFactory* /*this_ptr*/) noexcept {
+  return 0;
```
tianleiwu, line 65: 0 is None. 
Shall this return actual vendor Id here?

onnxruntime/core/providers/webgpu/webgpu_execution_provider.cc
```diff
@@ -837,6 +838,68 @@ std::unique_ptr<KernelRegistry> RegisterKernels(bool enable_graph_capture = fals
...
+std::shared_ptr<KernelRegistry> GetKernelRegistry(bool enable_graph_capture, bool enable_int64) {
+  // kernel registry variables are defined differently based on build configuration
+  //
+  // - When building as a static library, use static local variable. This is because
+  //   we don't have a reliable way to explicitly destroy the kernel registry after
+  //   use.
+  //
+  // - When building as a shared library, use global variables. The cleanup will be performed
+  //   when `ReleaseEpFactory` is called.
+  if (enable_graph_capture) {
```
tianleiwu, line 871: Please add some comments why graph capture need separate kernel registration.

onnxruntime/core/providers/webgpu/webgpu_execution_provider.cc
```diff
@@ -837,6 +838,68 @@ std::unique_ptr<KernelRegistry> RegisterKernels(bool enable_graph_capture = fals
...
+    static std::shared_ptr<KernelRegistry> registry = RegisterKernels(true, true);
+    return registry;
+#else
+    std::lock_guard<std::mutex> lock(g_kernel_registry_mutex);
+    if (g_graph_capture_kernel_registry == nullptr) {
+      g_graph_capture_kernel_registry = RegisterKernels(true, true);
+    }
+    return g_graph_capture_kernel_registry;
+#endif
+  } else if (enable_int64) {
```
tianleiwu, line 882: Can current design handle both enable_graph_capture and enable_int64 are true?

Edit: we shall add back the comments:
enable_int64_ is always true when enable_graph_capture_ is true

tianleiwu (dismissed):

.github/workflows/windows_webgpu.yml
```diff
@@ -155,6 +155,107 @@ jobs:
...
+    runs-on: [
+      "self-hosted",
+      "1ES.Pool=onnxruntime-github-Win2022-GPU-A10",
+      "JobId=webgpu_plugin_build_x64_RelWithDebInfo-${{ github.run_id }}-${{ github.run_number }}-${{ github.run_attempt }}"
+      ]
+    timeout-minutes: 300
+    env:
+      OrtPackageId: Microsoft.ML.OnnxRuntime
+      OnnxRuntimeBuildDirectory: ${{ github.workspace }}
+      DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true
```
edgchen1: I didn't see `--build_csharp` or `--build_nuget` in the `build.py` options. do we need these variables and the NuGet/.NET-related setup steps?

.github/workflows/windows_webgpu.yml
```diff
@@ -155,6 +155,107 @@ jobs:
...
+            --wgsl_template static `
+            --use_vcpkg --use_vcpkg_ms_internal_asset_cache `
+            --cmake_extra_defines onnxruntime_BUILD_UNIT_TESTS=ON `
+            --disable_rtti `
+            --enable_lto
+
+          if ($lastExitCode -ne 0) {
+            exit $lastExitCode
+          }
+          Remove-Item "${{ github.workspace }}\RelWithDebInfo" -Include "*.obj" -Recurse
```
edgchen1: why do we need to delete the .obj files here?

cmake/onnxruntime_providers_webgpu.cmake
```diff
@@ -119,6 +119,12 @@
     if (CMAKE_SYSTEM_NAME STREQUAL "Emscripten")
       message(FATAL_ERROR "WebGPU EP shared library build is not supported on Emscripten. Please use static library build.")
     endif()
+
+    # Configure precompiled headers for shared library build
+    # PCH ensures ep/adapters.h is included first and improves compilation speed
+    target_precompile_headers(onnxruntime_providers_webgpu PRIVATE
+      "${REPO_ROOT}/include/onnxruntime/ep/adapters.h"
+    )
```
edgchen1, line 127: this is the plugin EP path, right? should the target be created with `onnxruntime_add_shared_library_module()` instead?
https://github.com/fs-eire/onnxruntime/blob/5ceb23cf6af9efdcc51aa5f90f5e303adb9e1bde/cmake/onnxruntime_providers_webgpu.cmake#L59

the test plugin EP targets are modules.
https://github.com/microsoft/onnxruntime/blob/d6dcef197c2214c0e0e291768a5b79677f186d8e/cmake/onnxruntime_unittests.cmake#L2035

onnxruntime/core/providers/webgpu/ep/ep.h
```diff
@@ -0,0 +1,78 @@
...
+/// </summary>
+class Ep : public onnxruntime::ep::adapter::Ep {
+ public:
+  struct Config {
+    AllocatorPtr cpu_allocator;
+    AllocatorPtr device_allocator;
+    AllocatorPtr initializer_allocator;
+  };
+
+  // Do not use a std::unique_ptr for impl_ because this requires the actual type definition.
```
edgchen1: what is `impl_` referring to?

onnxruntime/core/providers/webgpu/ep/ep.cc
```diff
@@ -0,0 +1,263 @@
...
+
+using onnxruntime::ep::Api;
+
+// Constructor
+Ep::Ep(std::unique_ptr<IExecutionProvider> impl, Factory& factory, const OrtLogger& logger, const Config& config)
+    : onnxruntime::ep::adapter::Ep{std::move(impl), config.cpu_allocator, config.device_allocator},
+      factory_{factory},
+      logger_{logger},
+      config_{config} {
+  ort_version_supported = ORT_API_VERSION;
```
edgchen1: I wonder if the initialization of `ort_version_supported` should go into `onnxruntime::ep::adapter::Ep::Ep()` instead?

onnxruntime/core/providers/webgpu/ep/ep.cc
```diff
@@ -0,0 +1,263 @@
...
+        bool has_past_present_share_buffer = false;
+        for (const auto& attr : node.GetAttributes()) {
+          if (attr.GetName() == "past_present_share_buffer") {
+            int64_t val = 0;
+            RETURN_IF_ERROR(attr.GetValue(val));
+            if (val != 0) {
+              has_past_present_share_buffer = true;
+            }
+          }
+        }
```
edgchen1: nit: early exit
```suggestion
        for (const auto& attr : node.GetAttributes()) {
          if (attr.GetName() == "past_present_share_buffer") {
            int64_t val = 0;
            RETURN_IF_ERROR(attr.GetValue(val));
            if (val != 0) {
              has_past_present_share_buffer = true;
            }
            break;
          }
        }
```

onnxruntime/core/providers/webgpu/ep/ep.cc
```diff
@@ -0,0 +1,263 @@
...
+        RETURN_IF_ERROR(Api().ep.EpGraphSupportInfo_AddSingleNode(graph_support_info, node));
+      }
+    }
+  } catch (const Ort::Exception& ex) {
+    Ort::Status status(ex);
+    return status.release();
+  } catch (const std::exception& ex) {
+    Ort::Status status(ex.what(), ORT_EP_FAIL);
+    return status.release();
+  }
```
edgchen1: maybe it would be useful to have some helper macros to reduce the boilerplate. something like this:
https://github.com/microsoft/onnxruntime/blob/e5f5a6a0c4fdfc044131479c0cc873758931ae5d/onnxruntime/test/autoep/library/plugin_ep_utils.h#L79-L93

onnxruntime/core/providers/webgpu/ep/ep.cc
```diff
@@ -0,0 +1,263 @@
...
+    *should_convert = result.value() ? 1 : 0;
+  } else {
+    *should_convert = -1;
+  }
+  return nullptr;
+}
+
+OrtStatus* ORT_API_CALL Ep::OnRunStartImpl(_In_ OrtEp* this_ptr,
+                                           _In_ const OrtRunOptions* run_options) noexcept {
+  onnxruntime::RunOptions options{};
```
edgchen1, line 217: general: would suggest handling exceptions for all API entrypoints. even if they are trivial and unlikely to throw now, it would be future proof in case that changes later.

onnxruntime/core/providers/webgpu/ep/ep.cc
```diff
@@ -0,0 +1,263 @@
...
+
+    if (all_nodes.empty()) {
+      return nullptr;  // No nodes to process
+    }
+
+    std::vector<const OrtNode*> candidate_nodes;
+
+    // For each node, check if we have a registered kernel for it
+    for (const auto& node : all_nodes) {
+      if (node.GetEpName() == kWebGpuExecutionProvider) {
```
adrianlizarraga: Do we need to explicitly reject nodes that have been assigned to a different EP too?

onnxruntime/core/providers/webgpu/ep/ep.cc
```diff
@@ -0,0 +1,263 @@
...
+    *should_convert = -1;
+  }
+  return nullptr;
+}
+
+OrtStatus* ORT_API_CALL Ep::OnRunStartImpl(_In_ OrtEp* this_ptr,
+                                           _In_ const OrtRunOptions* run_options) noexcept {
+  onnxruntime::RunOptions options{};
+  // currently only option "gpu_graph_id" is used
+  auto graph_annotation_str = Api().ort.GetRunConfigEntry(run_options, kOrtRunOptionsConfigCudaGraphAnnotation);
```
adrianlizarraga, line 219: Is this only handling this specific run option because we're missing a public C API to get all config entries from a `OrtRunOptions`? If so, I suppose that would be a good thing to add in a separate PR. It would allow this function to pass along all configs.

onnxruntime/core/providers/webgpu/ep/factory.cc
```diff
@@ -0,0 +1,213 @@
...
+      ep_devices[num_ep_devices++] = ep_device;
+    }
+  }
+
+  return nullptr;
+}
+
+OrtStatus* ORT_API_CALL Factory::CreateEpImpl(
+    OrtEpFactory* this_ptr,
+    const OrtHardwareDevice* const* /*devices*/,
```
adrianlizarraga, line 105: Should this function return a status error if there is more than one hardware device specified here? For example, an application may try to create a session with two OrtEpDevice instances: a discrete gpu and an integrated gpu.

I'm assuming this is not supported by webgpu EP, so perhaps this should return an error to the user.

onnxruntime/core/providers/webgpu/ep/factory.cc
```diff
@@ -0,0 +1,213 @@
...
+    OrtEpFactory* /*this_ptr*/,
+    const OrtMemoryDevice* /*memory_device*/,
+    const OrtKeyValuePairs* /*stream_options*/,
+    OrtSyncStreamImpl** stream) noexcept {
+  *stream = nullptr;
+  return Api().ort.CreateStatus(ORT_NOT_IMPLEMENTED,
+                                "CreateSyncStreamForDevice is not implemented for this EP factory.");
+}
+
+OrtStatus* ORT_API_CALL Factory::ValidateCompiledModelCompatibilityInfoImpl(
```
adrianlizarraga: This function is optional to implement. Can probably be deleted.

onnxruntime/core/providers/webgpu/ep/factory.cc
```diff
@@ -0,0 +1,213 @@
...
+    OrtEpFactory* /*this_ptr*/,
+    const OrtHardwareDevice* const* /*devices*/,
+    size_t /*num_devices*/,
+    const char* /*compatibility_info*/,
+    OrtCompiledModelCompatibility* model_compatibility) noexcept {
+  *model_compatibility = OrtCompiledModelCompatibility_EP_NOT_APPLICABLE;
+  return nullptr;
+}
+
+OrtStatus* ORT_API_CALL Factory::SetEnvironmentOptionsImpl(
```
adrianlizarraga: This function is optional. Could also be deleted.

onnxruntime/core/providers/webgpu/data_transfer.cc
```diff
@@ -13,32 +13,45 @@ bool DataTransfer::CanCopy(const OrtDevice& src_device, const OrtDevice& dst_dev
...
+                               bytes);
       }
-    } else /* if (src_device.Type() == OrtDevice::GPU) */ {
+    } else {
       // copy from GPU to CPU
-      buffer_manager_.Download(static_cast<WGPUBuffer>(const_cast<void*>(src_data)), dst_data, bytes);
+      buffer_manager_.Download(static_cast<WGPUBuffer>(const_cast<void*>(src_data)),
+                               dst_data,
+                               bytes);
     }
```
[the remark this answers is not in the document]
fs-eire: The WebGPU data transfer does not deal with CPU-CPU data transfer. No need to add this code.

onnxruntime/core/providers/webgpu/ep/ep.cc
fs-eire: I think it is OK for WebGPU. The current design seems requires a deep copy to convert `OrtRunOptions` into a `onnxruntime::RunOptions`, so even if we can get the all entries we may still want to just look at the ones that we are interested in.

onnxruntime/core/providers/webgpu/ep/factory.cc
fs-eire: This is totally a valid use scenario, however WebGPU does not work in this way.

Please check my comment reply at https://github.com/microsoft/onnxruntime/pull/26907#issuecomment-4050627656. In short, the ORT device discovery does not work with WebGPU. Given a specific `OrtHardwareDevice`, I cannot pass any info from the OrtHardwareDevice to WebGPU device creation, and I cannot tell whether the WebGPU backend will use it or not.

This is the reason why I just ignore the param

onnxruntime/core/providers/webgpu/ep/ep.h
fs-eire: this seems to be out of dated. a previous implementation uses `impl_` in this class but that is already changed.

remove the comment.

onnxruntime/core/providers/webgpu/ep/factory.cc
fs-eire: For WebGPU, Vendor ID is explicitly defined to 0. See https://github.com/microsoft/onnxruntime/blob/b2808015829d08c92f6fe0395e6ae0211a8ee3be/include/onnxruntime/core/framework/ortdevice.h#L56-L58

onnxruntime/test/unittest_util/test_dynamic_plugin_ep.h
```diff
@@ -74,7 +75,7 @@ bool IsInitialized();
 void Shutdown();
 
 // Returns a dynamic plugin EP `IExecutionProvider` instance, or `nullptr` if uninitialized.
-std::unique_ptr<IExecutionProvider> MakeEp(const logging::Logger* logger = nullptr);
+std::unique_ptr<IExecutionProvider> MakeEp(const logging::Logger* logger = nullptr, const ConfigOptions* config_options = nullptr);
```
adrianlizarraga: nit: maybe rename `config_options` to `ep_options` and a quick comment for clarity.

onnxruntime/test/util/default_providers.cc
```diff
@@ -273,28 +278,53 @@ std::unique_ptr<IExecutionProvider> DefaultXnnpackExecutionProvider() {
...
                   .IsOK());
   if (!is_nhwc) {
     // Enable NCHW support
-    ORT_ENFORCE(config_options.AddConfigEntry(webgpu::options::kPreferredLayout,
+    ORT_ENFORCE(config_options.AddConfigEntry(strip_ep_prefix(webgpu::options::kPreferredLayout),
                                               webgpu::options::kPreferredLayout_NCHW)
                     .IsOK());
   }
+#if defined(ORT_USE_EP_API_ADAPTERS)
+  return dynamic_plugin_ep_infra::MakeEp(nullptr, &config_options);
```
adrianlizarraga: @edgchen1 Was this meant to be called directly like this? Would this fail if the infra is not initialized via a json config?

edgchen1: `MakeEp()` could be called directly. if the dynamic plugin EP infra hasn't been initialized, it will return nullptr.

however, the result of `MakeEp()` is dependent on whatever is in that JSON configuration, which is not guaranteed to be the WebGPU plugin EP. so I'm not sure if we want to be calling this from `DefaultWebGpuExecutionProvider()`.

right now, the place that "knows" about a dynamic plugin EP is in base_tester.cc:
https://github.com/microsoft/onnxruntime/blob/72d46059c11f7c5f63b6a2d0121426b3cb60c65e/onnxruntime/test/unittest_util/base_tester.cc#L756-L758

edgchen1: if the plugin EP has a known name, we could also look for it.

e.g., assuming WebGPU plugin EP has the same name,
`ORT_ENFORCE(dynamic_plugin_ep_infra::GetEpName() == kWebGpuExecutionProvider);`

include/onnxruntime/ep/adapter/allocator.h
```diff
@@ -18,23 +18,49 @@ namespace adapter {
...
     Alloc = AllocImpl;
     Free = FreeImpl;
     Info = InfoImpl;
   }
+  AllocatorPtr GetImpl() {
+    if (!impl_) {
+      std::call_once(init_flag_, [this]() {
+        impl_ = get_allocator_impl_(*memory_info_);
+      });
+    }
```
edgchen1, line 53: nit: we don't need to check `!impl_` here, right? `std::call_once` will do the initialization once and other threads should see its result.

```suggestion
    std::call_once(init_flag_, [this]() {
      impl_ = get_allocator_impl_(*memory_info_);
    });
```

onnxruntime/core/providers/webgpu/ep/api.cc
```diff
@@ -0,0 +1,84 @@
...
+
+extern "C" {
+//
+// Public symbols
+//
+EXPORT_SYMBOL OrtStatus* CreateEpFactories(const char* /*registration_name*/, const OrtApiBase* ort_api_base,
+                                           const OrtLogger* default_logger,
+                                           OrtEpFactory** factories, size_t max_factories, size_t* num_factories) {
+  // Manual init for the C++ API
+  onnxruntime::ep::ApiInit(ort_api_base);
```
edgchen1, line 42: this is outside of the try block. is it guaranteed to not throw? it could be made `noexcept`

onnxruntime/core/providers/webgpu/ep/api.cc
```diff
@@ -0,0 +1,84 @@
...
+namespace protobuf {
+void ShutdownProtobufLibrary();
+}  // namespace protobuf
+}  // namespace google
+
+extern "C" {
+//
+// Public symbols
+//
+EXPORT_SYMBOL OrtStatus* CreateEpFactories(const char* /*registration_name*/, const OrtApiBase* ort_api_base,
```
edgchen1, line 36: should `CreateEpFactories()` and `ReleaseEpFactory()` also be `noexcept`?

onnxruntime/core/providers/webgpu/data_transfer.h
```diff
@@ -20,6 +20,12 @@ class DataTransfer : public IDataTransfer {
 
   common::Status CopyTensor(const Tensor& src, Tensor& dst) const override;
 
+  common::Status CopyTensorImpl(void const* src_data,
```
edgchen1: does the `Impl` function need to be public? if so, maybe don't include "Impl" in the name.

onnxruntime/core/providers/webgpu/webgpu_execution_provider.h
```diff
@@ -44,13 +52,21 @@ class WebGpuExecutionProvider : public IExecutionProvider {
   WebGpuExecutionProvider(int context_id, webgpu::WebGpuContext& context, WebGpuExecutionProviderConfig&& config);
   ~WebGpuExecutionProvider() override;
 
+  inline auto GetKernelRegistryImpl() const {
+    return webgpu::GetKernelRegistry(enable_graph_capture_, enable_int64_);
+  }
+
+#if !defined(ORT_USE_EP_API_ADAPTERS)
```
edgchen1, line 59: when would `WebGpuExecutionProvider` be used in a plugin EP build?

onnxruntime/test/unittest_util/test_dynamic_plugin_ep.cc
```diff
@@ -167,7 +167,7 @@ void Shutdown() {
   g_plugin_ep_infrastructure_state.reset();
 }
 
-std::unique_ptr<IExecutionProvider> MakeEp(const logging::Logger* logger) {
+std::unique_ptr<IExecutionProvider> MakeEp(const logging::Logger* logger, const ConfigOptions* ep_options) {
```
edgchen1, line 171: nit: include the `config_options.h` header

onnxruntime/core/providers/webgpu/ep/api.cc
```diff
@@ -0,0 +1,84 @@
...
+  onnxruntime::webgpu::CleanupKernelRegistries();
+
+  // STEP.3 - Clean up WebGPU contexts
+  onnxruntime::webgpu::CleanupWebGpuContexts();
+
+  // STEP.4 - Destroy the global default logger wrapper
+  ::onnxruntime::ep::adapter::LoggingManager::DestroyDefaultLogger();
+
+  // STEP.5 - Shutdown protobuf library
+  google::protobuf::ShutdownProtobufLibrary();
```
tianleiwu, line 78: If the host process still uses protobuf (e.g., for another EP or direct model loading), this will corrupt protobuf's internal state. ShutdownProtobufLibrary call should only happen if the plugin exclusively owns the protobuf instance, or be gated behind a flag / reference count

onnxruntime/core/providers/webgpu/ep/ep.cc
```diff
@@ -0,0 +1,271 @@
...
+    return nullptr;  // No nodes to process
+  }
+
+  std::vector<const OrtNode*> candidate_nodes;
+
+  // For each node, check if we have a registered kernel for it
+  for (const auto& node : all_nodes) {
+    std::string ep_name = node.GetEpName();
+
+    if (ep_name == kWebGpuExecutionProvider) {
```
tianleiwu, line 72: **Layout-assigned nodes are included in CPU-preference evaluation**: In the [native WebGPU EP](https://github.com/microsoft/onnxruntime/blob/a25b7ee05189a286ae2424bba64e1ea8b5577a73/onnxruntime/core/providers/webgpu/webgpu_execution_provider.cc#L963-L1026), nodes already assigned to `kWebGpuExecutionProvider` (inserted by layout transformer) go into `candidates` but **not** into `tenative_candidates`, which is the vector passed to `GetCpuPreferredNodes`. This prevents layout-transformed nodes from being demoted back to CPU. In the [EP API version](https://github.com/microsoft/onnxruntime/blob/a25b7ee05189a286ae2424bba64e1ea8b5577a73/onnxruntime/core/providers/webgpu/ep/ep.cc#L69-L138), all nodes — including those already assigned to WebGPU — are pushed into a single `candidate_nodes` vector and then the full vector is given to `GetCpuPreferredNodes`. This creates a plugin-only behavioral divergence where layout-transformed nodes can be excluded from the WebGPU partition. Fix by splitting into two lists:
  ```cpp
  std::vector<const OrtNode*> candidate_nodes;
  std::vector<const OrtNode*> tentative_candidates;

  for (const auto& node : all_nodes) {
    std::string ep_name = node.GetEpName();
    if (ep_name == kWebGpuExecutionProvider) {
      candidate_nodes.push_back(node);
      continue;  // do NOT add to tentative_candidates
    }
    // ... existing checks ...
    candidate_nodes.push_back(node);
    tentative_candidates.push_back(node);
  }

  // Pass only tentative_candidates to GetCpuPreferredNodes
  RETURN_IF_ERROR(onnxruntime::ep::GetCpuPreferredNodes(
      *ort_graph, *graph_support_info, ..., tentative_candidates, cpu_preferred_nodes));
  ```

onnxruntime/test/util/default_providers.cc
```diff
@@ -273,28 +278,54 @@ std::unique_ptr<IExecutionProvider> DefaultXnnpackExecutionProvider() {
...
   return nullptr;
 #endif
 }
 
 std::unique_ptr<IExecutionProvider> WebGpuExecutionProviderWithOptions(const ConfigOptions& config_options) {
-#if defined(USE_WEBGPU) && !defined(ORT_USE_EP_API_ADAPTERS)
+#if defined(USE_WEBGPU)
+#if defined(ORT_USE_EP_API_ADAPTERS)
+  auto ep_name = dynamic_plugin_ep_infra::GetEpName();
+  ORT_ENFORCE(ep_name.has_value() && *ep_name == kWebGpuExecutionProvider,
```
edgchen1: nit: could be simplified - a useful property of `std::optional` is that an empty one will not compare equal to any value.
```suggestion
  ORT_ENFORCE(ep_name == kWebGpuExecutionProvider,
```

include/onnxruntime/ep/adapter/allocator.h
fs-eire: This code is designed to support both lazy init and immediate init scenarios. If the impl is passed in constructor, the init flag and get_allocator_impl_ is expected never be used.

onnxruntime/core/providers/webgpu/ep/factory.cc
```diff
@@ -0,0 +1,206 @@
...
+const char* ORT_API_CALL Factory::GetVendorImpl(const OrtEpFactory* /*this_ptr*/) noexcept {
+  return "Microsoft";
+}
+
+uint32_t ORT_API_CALL Factory::GetVendorIdImpl(const OrtEpFactory* /*this_ptr*/) noexcept {
+  return 0;
+}
+
+const char* ORT_API_CALL Factory::GetVersionImpl(const OrtEpFactory* /*this_ptr*/) noexcept {
+  return "0.1.0";
```
fs-eire, line 69: Do we have any recommended pattern for the version string?

onnxruntime/core/providers/webgpu/webgpu_execution_provider.h
fs-eire: it's used here: https://github.com/fs-eire/onnxruntime/blob/a25b7ee05189a286ae2424bba64e1ea8b5577a73/onnxruntime/core/providers/webgpu/ep/factory.cc#L132-L137

onnxruntime/core/providers/webgpu/ep/api.cc
fs-eire: As long as protobuf is static linked to both ORT and WebGPU DLL, it is safe to call `ShutdownProtobufLibrary` in both places since they are working on totally different objects.

include/onnxruntime/ep/adapter/allocator.h
edgchen1: I see. but now I notice that we don't check for `impl != nullptr` in `Allocator(const OrtMemoryInfo* memory_info, AllocatorPtr impl)`.

onnxruntime/core/providers/webgpu/ep/factory.cc
edgchen1: it just needs to be [semver](https://semver.org/spec/v2.0.0.html)

onnxruntime/test/util/default_providers.cc
edgchen1: maybe also consider updating `base_tester.cc` to so that the plugin WebGPU EP will use this same initialization function. possibly we could just remove the `USE_WEBGPU` ifdefs there

onnxruntime/test/unittest_util/test_dynamic_plugin_ep.h
```diff
@@ -17,8 +17,10 @@
 namespace onnxruntime {
 struct IExecutionProviderFactory;
 class IExecutionProvider;
+struct ConfigOptions;
 
 namespace logging {
+struct ConfigOptions;
```
edgchen1: why is there another forward declaration for `ConfigOptions`, but in `logging`?

fs-eire: my bad. I try to fix the previous comment for forward declaration of `ConfigOptions` and put it in the wrong place. let me revert this.

fs-eire: revert is pushed.

onnxruntime/test/unittest_util/test_dynamic_plugin_ep.cc
edgchen1: I still think that including the `config_options.h` header would be clearer. now, we rely on it being included by something else. we actually need the definition of the type at this point.

include/onnxruntime/ep/adapter/allocator.h
fs-eire: added check for it

onnxruntime/core/providers/webgpu/ep/factory.cc
fs-eire: thank you. I will modify this part in a following up PR and create a packaging pipeline for it.

PR #26910, opened 2026-01-05
Update protobuf references from 3.20.3 to 4.25.8

eserscor:
### Description
Version update for security fixes.

### Motivation and Context
Version update for security fixes.


tianleiwu:
Could you try run the notebook to see whether it is still working?

PR #26911, opened 2026-01-05
[java] Adds MemorySegment support to allow OrtValues greater than 2GB

Craigacp:
### Description
Adds a method to create `OnnxTensor` instances backed by Java 22's `MemorySegment`, and also a method to get a value out from a `OnnxTensor` in a `MemorySegment`. This is done via reflection over the JDK classes to determine if the `MemorySegment` class is available, and if not it throws `UnsupportedOperationException`. Consequently this code compiles fine on Java 8, but the `MemorySegment` based operations are only available when ORT is running on Java 22 or newer. It has a different shim for Android which unconditionally throws `UnsupportedOperationException` as Android doesn't support the Java 22 FFM API.

Due to the reflection the methods which interact with `MemorySegment` accept `Object` and do a runtime type check that the supplied argument is actually a `MemorySegment`. It means that the compile time type check that would ordinarily occur in user code is shifted to runtime, but it's still type safe and checked.

It would be good to update the CI to run the Java tests on Java 22 or newer (e.g. on Java 25, the LTS version which includes Java 22 features) as well as at least one test on Java 21 or older, but I'm not sure what the Azure CI environment is like so I might need some assistance to do that.

There are two related bug fixes in here as well, the first is that `getBufferRef` used to throw `NullPointerException` if it was called on a tensor that wasn't backed by a buffer, that was an oversight when I modified the duplicate call, and the second is that the `get<type>Buffer` calls can throw `OrtException` up from the JNI code and so should have been tagged `throws OrtException`. This latter issue is pretty unlikely as ORT doesn't error from that method very often otherwise we'd have seen issues.

### Motivation and Context
Java's byte buffers and arrays are limited to a maximum of 2^31 - 1 elements as they are indexed by a Java int. To load some models with external initializers we need to be able to create tensors with more elements (e.g. to hold the embedding matrix for some LLMs), which is impossible with the current code.

### DIFF
--- a/java/src/main/java/ai/onnxruntime/OnnxTensor.java
+++ b/java/src/main/java/ai/onnxruntime/OnnxTensor.java
@@ -475,12 +475,18 @@ public class OnnxTensor extends OnnxTensorLike {
    *     native code encountered an error.
    */
   private ByteBuffer getBuffer() throws OrtException {
-    try {
+    // Definitely can't allocate a byte buffer greater than Integer.MAX_VALUE, and
+    // it's typically recommended to make it a little smaller than that as the actual
+    // upper limit is somewhat JVM dependent.
+    int maxSize = (Integer.MAX_VALUE / info.type.size) - 4;
+    if (info.getNumElements() < maxSize) {
       return getBuffer(OnnxRuntime.ortApiHandle, nativeHandle).order(ByteOrder.nativeOrder());
-    } catch (IllegalArgumentException e) {
-      // thrown by the byte buffer constructor if the tensor is bigger than Integer.MAX_VALUE.
+    } else {
       throw new OrtException(
-          "Cannot construct a java.nio.Buffer of this size. Message: " + e.getMessage());
+          "Cannot construct a java.nio.Buffer of this size. This tensor has "
+              + info.getNumElements()
+              + ", and the maximum supported is "
+              + maxSize);
     }
   }
 
@@ -844,8 +850,11 @@ public class OnnxTensor extends OnnxTensorLike {
       long[] shape,
       OnnxJavaType type)
       throws OrtException {
-    if (!allocator.isClosed()) {
+    if (!allocator.isClosed() && env != null) {
       return createTensor(type, allocator, data, shape);
+    } else if (env == null) {
+      throw new IllegalStateException(
+          "Trying to create an OnnxTensor with an invalid OrtEnvironment.");
     } else {
       throw new IllegalStateException("Trying to create an OnnxTensor on a closed OrtAllocator.");
     }


java/src/main/jvm/ai/onnxruntime/MemorySegmentShim.java
```diff
@@ -0,0 +1,261 @@
...
+package ai.onnxruntime;
+
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodType;
+import java.util.logging.Logger;
+
+/**
+ * Wrapper for java.lang.foreign.MemorySegment instances which uses reflection to access the methods
+ * so it can be compiled on Java 21 and earlier. Requires Java 22 or newer to use MemorySegments,
```
yuslepukhin, line 15: The following is not clear from documentation for two versions of this class:
1) If the version of Java is 22, does it mean that `java.lang.foreign.MemorySegment` is guaranteed to be supported? If so, then why we need reflection?
2) If the answer to the above is no, then would it not make sense to create only one version of the class that would check both Java version and the availability of the feature?

java/src/main/jvm/ai/onnxruntime/MemorySegmentShim.java
```diff
@@ -0,0 +1,261 @@
...
+              .asType(
+                  MethodType.methodType(
+                      void.class, Object.class, Object.class, long.class, float.class));
+      tmpLayout =
+          lookup.findStaticGetter(valueLayoutClass, "JAVA_FLOAT", floatValueLayoutClass).invoke();
+    } catch (IllegalAccessException
+        | NoSuchMethodException
+        | ClassNotFoundException
+        | NoSuchFieldException e) {
+      logger.fine("Running on Java 21 or earlier, MemorySegment not available");
```
yuslepukhin, line 40: >Running on Java 21 or earlier [](http://example.com/codeflow?start=19&length=29)

Do you need so much code and throwing and catching exceptions to detect the version of Java?

java/src/main/jvm/ai/onnxruntime/MemorySegmentShim.java
```diff
@@ -0,0 +1,261 @@
...
+      tmpOfAddress = null;
+      tmpReinterpret = null;
+      tmpAddress = null;
+      tmpByteSize = null;
+      tmpIsNative = null;
+      tmpSet = null;
+      tmpLayout = null;
+    } catch (Throwable e) {
+      logger.severe(
+          "Failed to load float value layout, while other Java 22 features were available.");
```
yuslepukhin: >while other Java 22 features were available [](http://example.com/codeflow?start=46&length=43)

What other features of Java 22 are available?
Do we need to check for java version in other portions of the code?

java/src/main/jvm/ai/onnxruntime/MemorySegmentShim.java
```diff
@@ -0,0 +1,261 @@
...
+   * @throws UnsupportedOperationException If java.lang.foreign.MemorySegment is not available in
+   *     the running JDK.
+   */
+  MemorySegmentShim(Object segment) {
+    if (memorySegmentClass != null) {
+      if (memorySegmentClass.isInstance(segment)) {
+        this.segment = segment;
+      } else {
+        throw new IllegalArgumentException(
+            "Segment argument was not a java.lang.foreign.MemorySegment, found "
```
yuslepukhin, line 131: >Segment argument was not a java.lang.foreign.MemorySegment, found " [](http://example.com/codeflow?start=13&length=67)

My understanding that MemorySegment class is not functional if the feature is not available. If the constructor throws and prevents the instantiation, then code checks in other methods are redundant?

java/src/main/java/ai/onnxruntime/OnnxTensor.java
```diff
@@ -291,8 +331,10 @@ public synchronized void close() {
    * the OnnxTensor.
    *
    * @return A ByteBuffer copy of the OnnxTensor.
+   * @throws OrtException If the value could not be extracted as the Tensor is invalid, or if the
+   *     native code encountered an error.
    */
-  public ByteBuffer getByteBuffer() {
+  public ByteBuffer getByteBuffer() throws OrtException {
```
yuslepukhin, line 337: > throws OrtException [](http://example.com/codeflow?start=35&length=20)

This changes the signature and would cause a potentially cascading signature changes in the client code. Do we really need to do it?
To preserve the API, would it not be better to translate the exception to something that was thrown before?

java/src/main/java/ai/onnxruntime/OnnxTensor.java
```diff
@@ -310,8 +352,10 @@ public ByteBuffer getByteBuffer() {
    * into a float (i.e. it's a float, fp16 or bf16), otherwise it returns null.
    *
    * @return A FloatBuffer copy of the OnnxTensor.
+   * @throws OrtException If the value could not be extracted as the Tensor is invalid, or if the
+   *     native code encountered an error.
    */
-  public FloatBuffer getFloatBuffer() {
+  public FloatBuffer getFloatBuffer() throws OrtException {
```
yuslepukhin, line 358: > throws OrtException  [](http://example.com/codeflow?start=37&length=21)

Ditto here and below. #Resolved

java/src/main/java/ai/onnxruntime/OnnxTensor.java
```diff
@@ -419,9 +471,32 @@ public LongBuffer getLongBuffer() {
...
+   * @throws OrtException If the value could not be extracted as the Tensor is invalid, or if the
+   *     native code encountered an error.
    */
-  private ByteBuffer getBuffer() {
-    return getBuffer(OnnxRuntime.ortApiHandle, nativeHandle).order(ByteOrder.nativeOrder());
+  private ByteBuffer getBuffer() throws OrtException {
+    try {
+      return getBuffer(OnnxRuntime.ortApiHandle, nativeHandle).order(ByteOrder.nativeOrder());
+    } catch (IllegalArgumentException e) {
+      // thrown by the byte buffer constructor if the tensor is bigger than Integer.MAX_VALUE.
```
yuslepukhin: >if the tensor is bigger than Integer.MAX_VALUE. [](http://example.com/codeflow?start=47&length=47)

Is there a better and more deterministic way to detect a big buffer and throw OrtException to that effect? IllegalArgument is too vague and may be thrown by other parts of the code for other reasons?

java/src/main/java/ai/onnxruntime/OnnxTensor.java
```diff
@@ -419,9 +471,32 @@ public LongBuffer getLongBuffer() {
...
+   *
+   * <p>MemorySegments are only supported on Java 22 or newer, if called in an earlier version of
+   * Java this method throws {@link UnsupportedOperationException}.
+   *
+   * @return A MemorySegment wrapping the data.
+   * @throws OrtException If the native code encountered an error.
+   */
+  public Object getMemorySegment() throws OrtException {
+    long[] info = getSegmentPointer(OnnxRuntime.ortApiHandle, nativeHandle);
+    MemorySegmentShim shim = new MemorySegmentShim(info[0], info[1]);
```
yuslepukhin, line 504: >MemorySegmentShim shim = new MemorySegmentShim(info[0], info[1]) [](http://example.com/codeflow?start=4&length=64)

This will throw if MemorySegment is not available.
Is there a better way to discover the fact so the client code can choose a different path of execution? Handling exceptions to alter the flow of execution is less convenient than if/else.

java/src/main/java/ai/onnxruntime/OnnxTensor.java
```diff
@@ -720,6 +823,34 @@
...
+   * @param allocator The allocator to use.
+   * @param data The tensor data.
+   * @param shape The shape of tensor.
+   * @param type The type to use for the byte buffer elements.
+   * @return An OnnxTensor of the required shape.
+   * @throws IllegalArgumentException If the MemorySegment is not on the native heap.
+   * @throws OrtException Thrown if there is an onnx error or if the data and shape don't match.
+   */
+  static OnnxTensor createTensor(
+      OrtEnvironment env,
```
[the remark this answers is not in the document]
yuslepukhin: Can we comment out the name, or remove the parameter?

java/src/main/java/ai/onnxruntime/TensorInfo.java
```diff
@@ -332,7 +332,7 @@ public long getNumElements() {
    */
   public Object makeCarrier() throws OrtException {
     // Zero length tensors are allowed to be returned.
-    if (!validateShape() && numElements != 0) {
+    if ((!validateShape() && numElements != 0) || (numElements * type.size >= Integer.MAX_VALUE)) {
       throw new OrtException(
           "This tensor is not representable in Java, it's too big - shape = "
```
yuslepukhin: >his tensor is not representable in Java, it's too big - shape = [](http://example.com/codeflow?start=12&length=63)

Would it be helpful to recommend using Java 22 with MemorySegments available?

java/src/main/java/ai/onnxruntime/OnnxTensor.java
```diff
@@ -702,6 +786,25 @@ static OnnxTensor createTensor(
...
+   *
+   * @param env The current OrtEnvironment.
+   * @param data The tensor data in a native {@code java.lang.foreign.MemorySegment}.
+   * @param shape The shape of tensor.
+   * @param type The type to use for the byte buffer elements.
+   * @return An OnnxTensor of the required shape.
+   * @throws IllegalArgumentException If the MemorySegment is not on the native heap.
+   * @throws OrtException Thrown if there is an onnx error or if the data and shape don't match.
+   */
+  public static OnnxTensor createTensorFromMemorySegment(
```
yuslepukhin, line 808: > public static OnnxTens [](http://example.com/codeflow?start=1&length=23)

Here and in other places. Would it be possible to annotate public API in such a way that it does not show in the IDE when the version of Java is less than 22? #Resolved

java/src/main/native/ai_onnxruntime_OnnxTensor.c
```diff
@@ -170,6 +204,39 @@ JNIEXPORT jobject JNICALL Java_ai_onnxruntime_OnnxTensor_getBuffer
...
 }
 
+/*
+ * Class:     ai_onnxruntime_OnnxTensor
+ * Method:    getSegmentPointer
+ * Signature: (JJ)[J;
+ */
+JNIEXPORT jobject JNICALL Java_ai_onnxruntime_OnnxTensor_getSegmentPointer
+        (JNIEnv * jniEnv, jobject jobj, jlong apiHandle, jlong handle) {
+    (void) jobj;  // Required JNI parameter not needed by functions which don't need to access their host object.
```
yuslepukhin, line 214: >void) jobj; [](http://example.com/codeflow?start=5&length=11)

suggest simply commenting out the name of the function parameter. #Resolved

java/src/main/java/ai/onnxruntime/OnnxTensor.java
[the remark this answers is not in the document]
Craigacp: It enforces in the type system that the `OrtEnvironment` has been created before creating any tensors, and I've seen weird error messages if the user creates them the other way around. Java 8 doesn't have a way to specify this variable is unused (e.g. `_` as a variable name), that only came in in Java 21.

java/src/main/jvm/ai/onnxruntime/MemorySegmentShim.java
Craigacp: If it's on Java 22 or newer then it is guaranteed to be supported. However Java doesn't have conditional compilation, and the multi-release jar feature which allows different code to be loaded depending on the Java version has poor tooling support and requires that every version have the same public API which make it not suitable for this use case.

java/src/main/jvm/ai/onnxruntime/MemorySegmentShim.java
Craigacp: This exception is only thrown in the case where we failed to execute the method to create the float value layout, and I can't think of a case where that could actually happen. However the `invokeExact` call declares that it throws `Throwable` so I need to catch it here. I've refactored this check a bit now.

java/src/main/jvm/ai/onnxruntime/MemorySegmentShim.java
Craigacp: I've refactored the lookup logic here to make it smaller.

java/src/main/jvm/ai/onnxruntime/MemorySegmentShim.java
Craigacp: Yes, that's probably unnecessary safety as it shouldn't be possible to manufacture one of these without doing more nasty reflection tricks or using JNI to make an instance without going through the constructor.

java/src/main/java/ai/onnxruntime/OnnxTensor.java
Craigacp: Unfortunately this is a bug on my part. That method has always been able to throw `OrtException` as the exception can be created in JNI by [this line](https://github.com/microsoft/onnxruntime/blob/main/java/src/main/native/ai_onnxruntime_OnnxTensor.c#L164), but I forgot to add it to the Java signature [here](https://github.com/microsoft/onnxruntime/blob/main/java/src/main/java/ai/onnxruntime/OnnxTensor.java#L434) so it was never checked.

Now I think the underlying `GetTensorMutableData` call can't currently return anything other than `ORT_OK`, but I don't watch PRs that modify the implementation of the C API, so it could do in the future. If we can guarantee that it'll never return anything other than `ORT_OK` then I can take it back out and add a comment to that effect in the Java code.

java/src/main/java/ai/onnxruntime/OnnxTensor.java
Craigacp: The `IllegalArgumentException` that is thrown is thrown by the JDK itself from the JNI method that wraps a direct byte buffer around a pointer. I could modify the JNI logic to check how large the pointer is before calling the direct byte buffer constructor and throw an `OrtException`, but I'd have to construct the string error message down in C and I try to avoid doing string handling in C if I can.

java/src/main/java/ai/onnxruntime/OnnxTensor.java
Craigacp: My expectation is that all user code that wants to call this ORT code will have the concrete `MemorySegment` type in it somewhere so it'll have to compile and run with Java 22. Then all this will work and it won't throw exceptions. It's only if the user also can't control what JVM they are running on but wants to selectively run things through `MemorySegment` if they are running on a newer JVM. But if they did that they still wouldn't be able to make 2GB+ tensors on Java 21 or earlier and the user code logic would have to be even more contorted and full of reflection than this is.

java/src/main/java/ai/onnxruntime/TensorInfo.java
Craigacp: Yes, good point, I've updated the message.

java/src/main/java/ai/onnxruntime/OnnxTensor.java
Craigacp: Unfortunately not, there's no conditional compilation or method visibility. Using a multi-release jar which allows different implementations for different Java versions also requires that the public API of all the classes is the same, so even if I did this the more complicated way it would still be visible but unusable in earlier Java versions.

java/src/main/native/ai_onnxruntime_OnnxTensor.c
Craigacp: Oh. I thought I got compiler errors on some platforms when I did that. All the JNI methods are written like this with the same idioms, so I'd prefer to fix all of them at once if we did that.

yuslepukhin:
Please, address and/or comment on the Copilot comments.

java/src/main/java/ai/onnxruntime/OnnxTensor.java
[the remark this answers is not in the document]
yuslepukhin: I am afraid it does not enforce anything. I can simply pass null as a value. People will discover this code and will use the opportunity and then it would be hard to change.

java/src/main/jvm/ai/onnxruntime/MemorySegmentShim.java
yuslepukhin: My question was, can you call some API, find out the Java version, simply compare an integer and not proceed any further?

java/src/main/jvm/ai/onnxruntime/MemorySegmentShim.java
yuslepukhin: Conditional compilation is not the only way to address the issue. In fact, since we publish binaries, it would not be applicable anyway.
One way to address this is to have difference private implementation classes behind MemorySegment facade and instantiate one depending on which version and environment we run in.

However, my question remains. Suppose I am a developer who is writing code that is supposed to work under the following circumstances:
1) Android
2) Java version < 22
3) Java version >=22

How do I write it in  way that is convenient and operational and transparent? This is the key design question.

Here is one more specific: How do I avoid changing my client code and it would automatically take advantage of big tensors, so I do not have to do anything.
If, we are not in the environment that does not support big tensors, everything functions as before UNLESS the tensor is too big.

java/src/main/jvm/ai/onnxruntime/MemorySegmentShim.java
yuslepukhin: The message is confusing. There are no other Java 22 features that are supported in the code, otherwise we would have more checks for Java version.

java/src/main/java/ai/onnxruntime/OnnxTensor.java
yuslepukhin: I think you can rely on the ABI guarantees. However, if the ABI signature returns an OrtStatus, then one should be prepared to handle an error.

java/src/main/java/ai/onnxruntime/OnnxTensor.java
yuslepukhin: Perhaps, you can return a meaningful error code and deal with strings in Java code. This would also help to create a transparent API on the Java side that would automatically handle big buffers without need to introduce new public API classes.

java/src/main/java/ai/onnxruntime/OnnxTensor.java
yuslepukhin: I think what users want is to have their code continue as before, with an added benefit of large tensors behind the scenes if the environment supports it.

java/src/main/java/ai/onnxruntime/OnnxTensor.java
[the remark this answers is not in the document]
Craigacp: Ok, I'll put a null check in then.

java/src/main/jvm/ai/onnxruntime/MemorySegmentShim.java
Craigacp: In Java 8 I can parse the version string out of the `java.version` [system property](https://docs.oracle.com/en/java/javase/25/docs/api/java.base/java/lang/System.html#getProperties()), but that's more code than the `Class.forName` and I'll still need the try catch around the `Class.forName` and the method handle lookups as those throw checked exceptions. So it'll be longer.

java/src/main/java/ai/onnxruntime/OnnxTensor.java
Craigacp: It returns an `OrtStatus`, and everywhere else code that touches one of those is marked `throws OrtException`.

java/src/main/java/ai/onnxruntime/OnnxTensor.java
Craigacp: I'm not sure what you mean? It would need to be a different return type on the Java side to send back more than 2GB, and we can't bump the version to Java 25 which is the LTS version which supports this. Android doesn't have `MemorySegment` and I don't think it's going to get it any time soon.

java/src/main/java/ai/onnxruntime/OnnxTensor.java
Craigacp: Unfortunately they can't have that, because they can't get the value out of a big `OnnxTensor` into Java or construct a big tensor from a Java side value without being on a new Java version, and the big values have to use a different type because `MemorySegment` and `ByteBuffer` don't share any super type other than `Object`.

If we wanted to rev the Java version to 25 we could only use `MemorySegment` the way that [OpenJDK's ONNX code transformation experiment](https://github.com/openjdk/babylon/tree/code-reflection/cr-examples/onnx) does, but that would fork the Java & Android APIs, and also abandon everyone who hadn't moved to 25 yet (which only came out September last year).

java/src/main/jvm/ai/onnxruntime/MemorySegmentShim.java
Craigacp: I'm not sure what "take advantage of big tensors" means? If the user wants to interact with a big tensor in Java they need to get it into a Java side object rather than an ORT native object, and there are no Java side objects big enough until Java 22. If they want to use a big tensor output by one model and pass it into another model without looking at it, they can already do that and nothing changes.

Craigacp:
> Please, address and/or comment on the Copilot comments.

Will do.

java/src/test/java/ai/onnxruntime/MemorySegmentTest.java
```diff
@@ -0,0 +1,260 @@
...
+          return allocate.invokeExact(arena, size);
+        } catch (Throwable e) {
+          throw new AssertionError("Should not reach here", e);
+        }
+      } else {
+        throw new UnsupportedOperationException("java.lang.foreign.Arena is not available.");
+      }
+    }
+
+    public void close() {
```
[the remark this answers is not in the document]
Craigacp: Fixed.

java/src/test/java/ai/onnxruntime/MemorySegmentTest.java
```diff
@@ -0,0 +1,260 @@
...
+  }
+
+  @Test
+  public void testNotASegment() throws OrtException {
+    OrtEnvironment env = OrtEnvironment.getEnvironment();
+    long[] shape = new long[] {256 * 1024, 4 * 1024};
+    Object notASegment = new Object();
+    try {
+      OnnxTensor tensor =
+              OnnxTensor.createTensorFromMemorySegment(env, notASegment, shape, OnnxJavaType.FLOAT);
```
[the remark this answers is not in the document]
Craigacp: This method throws an exception when called with these arguments and the test is checking that it throws the right exception with the right message.

java/src/test/java/ai/onnxruntime/MemorySegmentTest.java
```diff
@@ -0,0 +1,260 @@
...
+        float floati = (float) i;
+        for (int j = 0; j < 4096; j++) {
+          segment.set(4L * i * 4096L + 4L * j, floati);
+        }
+      }
+      OnnxTensor bigTensor =
+          OnnxTensor.createTensorFromMemorySegment(env, segment.get(), shape, OnnxJavaType.FLOAT);
+
+      try {
+        FloatBuffer fb = bigTensor.getFloatBuffer();
```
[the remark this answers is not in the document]
Craigacp: This method throws an exception when called in this state and the test is checking that it throws the right exception with the right message.

java/src/test/java/ai/onnxruntime/MemorySegmentTest.java
```diff
@@ -0,0 +1,260 @@
...
+        Assertions.fail("Should have thrown an exception");
+      } catch (OrtException e) {
+        Assertions.assertTrue(
+            e.getMessage().contains("Cannot construct a java.nio.Buffer of this size."));
+      }
+
+      try {
+        float[][] arr = (float[][]) bigTensor.getValue();
```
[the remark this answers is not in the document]
Craigacp: This method throws an exception when called in this state and the test is checking that it throws the right exception with the right message.

java/src/main/java/ai/onnxruntime/OnnxTensor.java
yuslepukhin: You can create/call a native method to find out from ORT the size of the date and then bail early.

java/src/main/java/ai/onnxruntime/OnnxTensor.java
yuslepukhin: I think we have a problem with the API. It does not expose the right data types. For example, the native size of the tensor in ORT is expressed as 64-bit integer, but the public API exposed primitives are dealing with int as sizes. This is wrong.

I suggest we re-think the API at this point and create a new one, that would expose the right data types and then transparently deals with both small and big tensors w/o user being concerned about the environment it runs in and the size of the tensor.
Is this not the premise of Java anyway?

java/src/test/java/ai/onnxruntime/MemorySegmentTest.java
```diff
@@ -0,0 +1,264 @@
...
+import java.nio.FloatBuffer;
+import java.util.Collections;
+import java.util.logging.Logger;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.EnabledForJreRange;
+import org.junit.jupiter.api.condition.JRE;
+
+/** Tests for interop with Java 22's MemorySegments. */
+@EnabledForJreRange(min = JRE.JAVA_22)
```
yuslepukhin, line 20: Looks like a conditional compilation or something like that.
From the example of this test, the user would have to create different binaries for Android and different versions of Java. Can we avoid that?

java/src/main/jvm/ai/onnxruntime/MemorySegmentShim.java
yuslepukhin: I am talking about the right abstraction on top whatever Java provides and maybe does not, to expose the right types, right sizes and right buffers. I think it is possible to do given the AI capabilities.

yuslepukhin: And I still do not understand why we need a separate empty shell class for Android. Would not simply fail to load as in this class?

Craigacp: Some Android versions don't have support for `MethodHandle` and throw if you try to use them. I'm not sure if ORT targets a new enough version of Android to avoid this problem, but if it does then I can remove the Android side bits for this and also for the FP16 conversions which use similar tricks.

Craigacp: > I am talking about the right abstraction on top whatever Java provides and maybe does not, to expose the right types, right sizes and right buffers. I think it is possible to do given the AI capabilities.

I am wary of writing a Java tensor type that only lives in this library and is disconnected from the ORT native code. Other Java ML libraries also have those and they all have no interop between those libraries aside from JDK classes like ByteBuffer and MemorySegment because the JDK does not provide useful interfaces for tensor types to implement. The choices I've been making with ORT are to allow users to bring their data in whatever form they have it and it's their job to get it into a JDK class which we consume directly without copies (which is typically where their data already is because those JDK classes underlie all fast off heap memory in Java and so they are underneath whatever user tensor abstraction they already have).

java/src/main/jvm/ai/onnxruntime/MemorySegmentShim.java
Craigacp: Ok, this has been fixed in the refactor of the method handle construction.

java/src/main/java/ai/onnxruntime/OnnxTensor.java
Craigacp: Fixed.

PR #26919, opened 2026-01-06
[EP API] header-only adapter for EP API

fs-eire:
### Description

This PR adds a few headers for supporting building WebGPU EP and CUDA EP as plugin EPs.

See summary of https://github.com/microsoft/onnxruntime/pull/26907

### DIFF
--- a/include/onnxruntime/ep/adapter/op_kernel.h
+++ b/include/onnxruntime/ep/adapter/op_kernel.h
@@ -90,7 +90,7 @@ struct OpKernelContext {
       return nullptr;
     }
 
-    input_tensors_[index] = CreateTensorFromApiValue(input);
+    input_tensors_[index] = CreateTensorFromApiValue(const_cast<OrtValue*>(static_cast<const OrtValue*>(input)));
     return &input_tensors_[index];
   }
   Tensor* Output(int index, const TensorShape& shape) {
@@ -191,7 +191,7 @@ struct KernelImpl : OrtKernelImpl {
     Status status;
     ORT_TRY {
       auto* kernel_impl = static_cast<KernelImpl*>(this_ptr)->impl_.get();
-      const auto tensor = CreateTensorFromApiValue(weight);
+      const auto tensor = CreateTensorFromApiValue(const_cast<OrtValue*>(weight));
       status = kernel_impl->PrePack(tensor, input_index, AllocatorPtr{}, *is_packed, nullptr);
     }
     ORT_CATCH(const std::exception& ex) {
--- a/include/onnxruntime/ep/adapter/op_kernel_info.h
+++ b/include/onnxruntime/ep/adapter/op_kernel_info.h
@@ -47,7 +47,7 @@ struct OpKernelInfo {
         int is_constant = 0;
         Ort::ConstValue const_input = info.GetTensorConstantInput(i, &is_constant);
         if (is_constant && const_input != nullptr && const_input.IsTensor()) {
-          constant_input_tensors[i] = CreateTensorFromApiValue(const_input);
+          constant_input_tensors[i] = CreateTensorFromApiValue(const_cast<OrtValue*>(static_cast<const OrtValue*>(const_input)));
         }
       }
     }
--- a/include/onnxruntime/ep/adapter/tensor_helper.h
+++ b/include/onnxruntime/ep/adapter/tensor_helper.h
@@ -19,8 +19,8 @@ namespace adapter {
 /// <summary>
 /// Create an unowned onnxruntime::Tensor from a tensor OrtValue from C API.
 /// </summary>
-inline onnxruntime::Tensor CreateTensorFromApiValue(const OrtValue* ort_value) {
-  Ort::ConstValue value{ort_value};
+inline onnxruntime::Tensor CreateTensorFromApiValue(OrtValue* ort_value) {
+  Ort::UnownedValue value{ort_value};
   EP_ENFORCE(value.IsTensor(), "Only tensor OrtValue is supported.");
 
   ONNXTensorElementDataType element_type;
@@ -43,7 +43,7 @@ inline onnxruntime::Tensor CreateTensorFromApiValue(const OrtValue* ort_value) {
 
   return Tensor(data_type,
                 TensorShape{shape.shape, shape.shape_len},
-                const_cast<void*>(value.GetTensorRawData()),
+                value.GetTensorMutableRawData(),
                 tensor_memory_info);
 }
 


adrianlizarraga:
@fs-eire, thank you for adding these utilities. We currently have some plugin EP utilities in this directory: https://github.com/microsoft/onnxruntime/tree/main/include/onnxruntime/core/providers/utils

And there's another on the way here: https://github.com/microsoft/onnxruntime/pull/25753

Could we consolidate all of these utilities into one location?

include/onnxruntime/ep/adapters.h
```diff
@@ -0,0 +1,61 @@
...
+#pragma once
+
+#include "api.h"
+#include "common.h"
+
+// This header is only used when building WebGPU/CUDA EP as a shared library.
+//
+// This header file is used as a precompiled header so it is always included first.
+
+#pragma push_macro("ORT_EP_API_ADAPTER_HEADER_INCLUDED")
```
edgchen1, line 13: why do we need `push_macro` and `pop_macro` instead of just using `#define` and `#undef`?

include/onnxruntime/ep/README.md
```diff
@@ -0,0 +1,7 @@
+## EP adapter
+
+This folder contains a set of C++ header files. They are used specifically for allowing ONNX Runtime internal kernel-based EPs to use the plugin-style EP API while keep minimal changes to existing code.
+
+### Usage
+
+Make sure to include "ep/_pch.h" for all source code in the implementation. Using PCH is recommended.
```
edgchen1: > Using PCH is recommended.

does this mean something other than including "_pch.h"? it's not obvious to me.

perhaps I'm just not understanding the name. could you either explain it more here or give `_pch.h` a more descriptive name?

include/onnxruntime/ep/adapter/allocator.h
```diff
@@ -0,0 +1,47 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#pragma once
+
+#include "core/framework/allocator.h"
```
edgchen1, line 10: naive question - is this header intended to be used by a plugin EP which doesn't have access to internal ORT code? if so, how can we use internal ORT headers?

edgchen1:
> @fs-eire, thank you for adding these utilities. We currently have some plugin EP utilities in this directory: https://github.com/microsoft/onnxruntime/tree/main/include/onnxruntime/core/providers/utils
> 
> And there's another on the way here: #25753
> 
> Could we consolidate all of these utilities into one location?

it may be worth considering whether the utility is generally useful to any plugin EP vs. useful to plugin EPs within the ORT repo. some of these seem like the latter.

include/onnxruntime/ep/adapter/allocator.h
fs-eire: No. The whole purpose of the folder `include/onnxruntime/ep/adapter` is designed for a very specific goal: to support the existing EPs (majorly WebGPU EP and CUDA EP) to be able to use the EP API, with minimal code changes.

This means most of the existing code can be kept as-is. For some components, we need a "reverse bridge" to support an implementation of `OrtAllocator` which wraps an ORT internal allocator, like this.

include/onnxruntime/ep/README.md
fs-eire: `_pch.h` is the file, but we want to actually use compiler flags to specify the file as a PCH to enforce.

I probably should write `Using PCH compiler flag is recommended.`

fs-eire:
> > @fs-eire, thank you for adding these utilities. We currently have some plugin EP utilities in this directory: https://github.com/microsoft/onnxruntime/tree/main/include/onnxruntime/core/providers/utils
> > And there's another on the way here: #25753
> > Could we consolidate all of these utilities into one location?
> 
> it may be worth considering whether the utility is generally useful to any plugin EP vs. useful to plugin EPs within the ORT repo. some of these seem like the latter.

I agree. we should not have separated implementation in different places of the repo.

In general, I think there are 2 different types of headers:
- the headers that ONLY depends on ORT public headers. This can include some useful common utils/macros and helper functions.
- the headers also depending on ORT internal headers. Specifically, for the EP adapter implementation. (currently they are in `include/onnxruntime/ep/adapter`)

I would recommend to put them somewhere under `include/onnxruntime/`, because considering the usage header-only will be much more easier for users to use. And if it's header-only, it should be put inside `include/onnxruntime/`.

All other details are open to discussion.

include/onnxruntime/ep/adapters.h
fs-eire: it's a practice to allow to preserve the old value of the macro if exists (although in this case it's very unlikely that user may use "ORT_EP_API_ADAPTER_HEADER_INCLUDED" to do something else).

The only concern would be the compiler support, but since ORT uses modern compiler (with at least support of c++17) this is no longer a concern.

include/onnxruntime/ep/adapter/allocator.h
```diff
@@ -0,0 +1,47 @@
...
+  static void ORT_API_CALL FreeImpl(OrtAllocator* this_ptr, void* p) noexcept {
+    auto* allocator = static_cast<Allocator*>(this_ptr);
+    allocator->impl_->Free(p);
+  }
+
+  static const OrtMemoryInfo* ORT_API_CALL InfoImpl(const OrtAllocator* this_ptr) noexcept {
+    auto* allocator = static_cast<const Allocator*>(this_ptr);
+    return allocator->memory_info_;
+  }
+
```
adrianlizarraga, line 44: Just want to double check that `AllocOnStream` is not also needed for cuda or webgpu EPs.

include/onnxruntime/ep/adapter/ep.h
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+ public:
+  inline IExecutionProvider* EpImpl() const noexcept {
+    return impl_.get();
+  }
+  inline const DataTransferManager& GetDataTransferManager() const noexcept {
+    return data_transfer_manager_;
+  }
+  [[nodiscard]] Status GetTempSpaceCPUAllocator(AllocatorPtr* output) const {
```
adrianlizarraga: `GetTempSpaceCPUAllocator` and `GetTempSpaceAllocator` are internally implemented by `onnxruntime::OpKernelContext`. Would it be possible/appropriate to add public C APIs to `OrtKernelContext` for these functions instead of including them here?

include/onnxruntime/ep/adapter/kernel_registry.h
```diff
@@ -0,0 +1,96 @@
...
+  static OrtStatus* CreateKernel(void* kernel_create_func_state, const OrtKernelInfo* info, OrtKernelImpl** out) {
+    FuncManager func_mgr;  // not used
+    std::unique_ptr<OpKernel> kernel;
+    KernelCreatePtrFn create_func = reinterpret_cast<KernelCreatePtrFn>(kernel_create_func_state);
+    Status status = create_func(func_mgr, OpKernelInfo(info), kernel);
+    if (!status.IsOK()) {
+      return ToOrtStatus(status);
+    }
+    *out = nullptr;
+    status = kernel->CreateControlFlowKernelImpl(info, out);
```
adrianlizarraga: nit: maybe add a comment explaining why we try to create a control flow kernel first.

include/onnxruntime/ep/adapter/op_kernel_info.h
```diff
@@ -0,0 +1,164 @@
...
+namespace adapter {
+
+/// <summary>
+/// An adapter class partially implementing the facade of `onnxruntime::OpKernelInfo`.
+/// </summary>
+struct OpKernelInfo {
+  //
+  // A helper struct to cache kernel info data
+  //
+  // Because `KernelCreatePtrFn` is defined to use `const OrtKernelInfo&` as parameter type of the kernel creation function, `OpKernelInfo` has to be copyable.
```
adrianlizarraga:
```suggestion
  // Because `KernelCreatePtrFn` is defined to use `const OpKernelInfo&` as parameter type of the kernel creation function, `OpKernelInfo` has to be copyable.
```

include/onnxruntime/ep/api.h
```diff
@@ -0,0 +1,51 @@
...
+namespace ep {
+
+struct ApiPtrs {
+  const OrtApi& ort;
+  const OrtEpApi& ep;
+  const OrtModelEditorApi& model_editor;
+};
+
+namespace detail {
+inline std::unique_ptr<ApiPtrs> g_api_ptrs;
```
adrianlizarraga: nit: Could this be a `std::optional<ApiPtrs>` or just `ApiPtrs`?

include/onnxruntime/ep/adapter/kernel_registry.h
fs-eire: added, please review the comments.

include/onnxruntime/ep/adapter/ep.h
fs-eire: I would prefer to keep it in the current way.

There are 2 reasons:
- performance consideration. Here, `onnxruntime::ep::adapter::EP::GetTempSpaceCPUAllocator` accepts a `AllocatorPtr*`, which is a pointer to an internal type. We have access to the reference to the `WebGpuExecutionProvider` instance any way so it's not a problem of getting the actual allocator pointer returned by it.
  If we change that to use an allocator returned by C API, then we have to use `onnxruntime::ep::adapter::Allocator` to wrap it (which has some overhead).

- for usage inside `include/onnxruntime/ep/adapter`, there is no point to do this as long as WebGPU EP creates the allocator and use it both without the C API.

include/onnxruntime/ep/adapter/allocator.h
fs-eire: we can always extend those classes to support more function that may be missing for CUDA. The classes are generally designed to be generic enough for being reused in a future migration for CUDA EP.

include/onnxruntime/ep/README.md
fs-eire: updated comments to make it clear.

include/onnxruntime/ep/adapters.h
edgchen1: if the macro is already defined, wouldn't there be a warning or error when it is defined again on the next line?

include/onnxruntime/ep/adapter/allocator.h
```diff
@@ -0,0 +1,47 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#pragma once
+
```
edgchen1, line 5: why does this header not have the following code like the other ones?

```
#if !defined(ORT_EP_API_ADAPTER_HEADER_INCLUDED)
#error "This header should not be included directly. Include ep/_pch.h instead."
#endif
```

include/onnxruntime/ep/README.md
```diff
@@ -0,0 +1,7 @@
+## EP adapter
+
+This folder contains a set of C++ header files. They are used specifically for allowing ONNX Runtime internal kernel-based EPs to use the plugin-style EP API while keep minimal changes to existing code.
+
+### Usage
+
+Make sure to include "ep/_pch.h" for all source code in the implementation. Using PCH compiler flag is recommended.
```
edgchen1: I still don't understand the name `_pch.h`. maybe you can explain it to me. or could we name it something like `ep/adapters.h`?

include/onnxruntime/ep/README.md
```diff
@@ -0,0 +1,7 @@
+## EP adapter
+
+This folder contains a set of C++ header files. They are used specifically for allowing ONNX Runtime internal kernel-based EPs to use the plugin-style EP API while keep minimal changes to existing code.
```
edgchen1, line 3: this readme is not located in the `adapters` directory. maybe move it there, or explicitly name the `adapters` directory.

fs-eire: I added a section in the README to explain the folder structure. Since there is an ongoing discussion about unifying the shared headers, I expect the current folder structure to be refactored later.

include/onnxruntime/ep/README.md
fs-eire: we are using `EP_SPECIFIC_USING_DECLARATIONS` to "override" some ORT classes with our adapter classes. This requires the header file to be included before any header that defines the old ORT classes. The best way to achieve this is to use the PCH (pre-compiled header), which is guaranteed by the compiler to ensure the file is included first.

we do have some 'pch.h' in winml folder, a 'test_pch.h' in cmake folder, and a PCH for CUDA EP: onnxruntime\core\providers\cuda\cuda_pch.h. Using the filename containing 'pch' may be a good way to indicating that the file is used as PCH explicitly.

The reason why the file is not inside `adapter` folder is because it depends on the shared plugin EP headers. Before we refactored them we probably need to keep it because otherwise `#include "../common.h"` seems not a very good way.

include/onnxruntime/ep/README.md
fs-eire: in future, when we have a dedicated folder for the shared plugin EP headers, we can put everything in the `adapter` directory.

include/onnxruntime/ep/adapters.h
```diff
@@ -0,0 +1,62 @@
...
+#include "adapter/ep.h"
+#include "adapter/kernel_registry.h"
+
+#pragma pop_macro("ORT_EP_API_ADAPTER_HEADER_INCLUDED")
+
+//
+// EP specific using declarations
+//
+
+#define EP_SPECIFIC_USING_DECLARATIONS                                              \
```
edgchen1, line 28: can we add `#undef EP_SPECIFIC_USING_DECLARATIONS` after the macro is no longer needed?

include/onnxruntime/ep/api.h
```diff
@@ -0,0 +1,50 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#pragma once
+
+#include <optional>
+
+#pragma push_macro("ORT_API_MANUAL_INIT")
+#define ORT_API_MANUAL_INIT
```
edgchen1, line 10: similar comment for handling an existing definition of `ORT_API_MANUAL_INIT`

include/onnxruntime/ep/common.h
```diff
@@ -0,0 +1,38 @@
...
+      return (ort_api).CreateStatus(ORT_EP_FAIL, (msg)); \
+    }                                                    \
+  } while (0)
+
+// see ORT_ENFORCE for implementations that also capture a stack trace and work in builds with exceptions disabled
+// NOTE: In this simplistic implementation you must provide an argument, even it if's an empty string
+#define EP_ENFORCE(condition, ...)                       \
+  do {                                                   \
+    if (!(condition)) {                                  \
+      std::ostringstream oss;                            \
```
edgchen1, line 31: add necessary includes to make this header self-contained. e.g., `<sstream>`, `<stdexcept>`, `"api.h"`

include/onnxruntime/ep/get_capability_utils.h
```diff
@@ -0,0 +1,253 @@
...
+  } catch (const Ort::Exception& ex) {
+    Ort::Status status(ex);
+    return status.release();
+  } catch (const std::exception& ex) {
+    Ort::Status status(ex.what(), ORT_EP_FAIL);
+    return status.release();
+  } catch (...) {
+    Ort::Status status("Unknown exception", ORT_EP_FAIL);
+    return status.release();
+  }
```
edgchen1, line 35: the exception handling adds some boilerplate. do we need it in these helper functions or can we move it to a higher level?

include/onnxruntime/ep/get_capability_utils.h
```diff
@@ -0,0 +1,253 @@
...
+
+using NodeId = size_t;
+constexpr int64_t kSmallInitializerThreshold = 100;
+
+constexpr inline bool MemTypeOnCpuExplicitly(OrtMemType mem_type) {
+  return mem_type == OrtMemTypeCPUInput || mem_type == OrtMemTypeCPUOutput;
+}
+
+// Get all output nodes that consume an output from the given node.
+static OrtStatus* GetOutputNodes(gsl::span<Ort::ConstValueInfo const> node_outputs, std::vector<Ort::ConstNode>& result) {
```
edgchen1: a `static` function inside a header will have a duplicate, file-scoped definition for each translation unit that includes it. is that intended?

include/onnxruntime/ep/get_capability_utils.h
```diff
@@ -0,0 +1,253 @@
...
+    return status.release();
+  } catch (...) {
+    Ort::Status status("Unknown exception", ORT_EP_FAIL);
+    return status.release();
+  }
+}
+
+// Returns nodes that should be assigned to CPU EP instead of this example EP to avoid costly I/O copies.
+// Based on GetCpuPreferredNodes from onnxruntime/core/framework/fallback_cpu_capability.cc
+OrtStatus* GetCpuPreferredNodes(const OrtGraph& ort_graph, OrtEpGraphSupportInfo& graph_support_info,
```
edgchen1: should this function be `inline`?

include/onnxruntime/ep/adapter/ep.h
edgchen1: nit: a `Status` return type does not need to be marked `[[nodiscard]]` since the `Status` class is already marked with that attribute.

include/onnxruntime/ep/README.md
edgchen1: > we are using EP_SPECIFIC_USING_DECLARATIONS to "override" some ORT classes with our adapter classes. This requires the header file to be included before any header that defines the old ORT classes.

that "_pch.h" file is including the old ORT classes and then adding aliases for them in other namespaces, right? can you elaborate on why it must be included first?

> The reason why the file is not inside adapter folder is because it depends on the shared plugin EP headers.

I was suggesting that it just be renamed to `adapters.h`. it can stay in the same directory.

include/onnxruntime/ep/get_capability_utils.h
fs-eire: fixed to use 'inline'.

include/onnxruntime/ep/README.md
fs-eire: renamed

include/onnxruntime/ep/adapter/ep.h
```diff
@@ -0,0 +1,58 @@
...
+namespace onnxruntime {
+namespace ep {
+namespace adapter {
+
+/// <summary>
+/// Wrapper around IExecutionProvider to expose via OrtEp.
+/// </summary>
+class Ep : public OrtEp {
+ protected:
+  explicit Ep(IExecutionProvider* impl, AllocatorPtr temp_space_cpu_allocator, AllocatorPtr temp_space_allocator)
```
edgchen1: since `impl_` is a `unique_ptr`, should we make this parameter a `unique_ptr` as well so the ownership transfer is more obvious?

include/onnxruntime/ep/adapter/kernel_def.h
```diff
@@ -0,0 +1,37 @@
...
+#endif
+
+#include <memory>
+
+namespace onnxruntime {
+namespace ep {
+namespace adapter {
+
+/// <summary>
+/// An adapter class partially implementing the facade of `onnxruntime::KernelDef`.
```
edgchen1: nit: ["facade"](https://en.wikipedia.org/wiki/Facade_pattern) as I understand it implies a different and possibly simpler interface. here, are we pretty much implementing the same interface? would "interface" be a more appropriate name for it?

include/onnxruntime/ep/adapter/op_kernel.h
```diff
@@ -0,0 +1,212 @@
...
+    Compute = ComputeImpl;
+    Release = ReleaseImpl;
+    PrePackWeight = PrePackWeightImpl;
+  }
+
+ private:
+  static OrtStatus* ORT_API_CALL ComputeImpl(_In_ OrtKernelImpl* this_ptr,
+                                             _In_ OrtKernelContext* context) noexcept {
+    const auto* kernel_impl = static_cast<KernelImpl*>(this_ptr)->impl_.get();
+    OpKernelContext ctx{context, *kernel_impl};
```
edgchen1: general: for functions on the API boundary, we should generally put everything that could possibly throw into the `try` block.

include/onnxruntime/ep/adapter/logging.h
```diff
@@ -0,0 +1,111 @@
...
+    const CodeLocation& location) {
+  return detail::LoggerCapture(logger, severity, category, datatype, location);
+}
+
+}  // namespace detail
+}  // namespace adapter
+}  // namespace ep
+}  // namespace onnxruntime
+
+// Undefine and redefine LOGS_DEFAULT
```
edgchen1: nit: macro name doesn't match
```suggestion
// Undefine and redefine logging macros
```

include/onnxruntime/ep/adapter/node.h
```diff
@@ -0,0 +1,40 @@
...
+namespace ep {
+namespace adapter {
+
+/// <summary>
+/// An adapter class partially implementing the facade of `onnxruntime::Node`.
+/// </summary>
+struct Node {
+  explicit Node(const OrtKernelInfo* kernel_info) : kernel_info_{kernel_info} {}
+  /** Gets the Node's name. */
+  const std::string Name() const noexcept {
```
edgchen1: nit: not much point in returning a const type (that can be copied) by value. same comment for `OpType()`
```suggestion
  std::string Name() const noexcept {
```

include/onnxruntime/ep/adapter/op_kernel.h
```diff
@@ -0,0 +1,212 @@
...
+  ORT_DISALLOW_COPY_ASSIGNMENT_AND_MOVE(OpKernelContext);
+  Ort::KernelContext context_;
+  const OpKernel& op_kernel_;
+  const std::vector<Tensor>& constant_input_tensors_;
+  mutable std::vector<Tensor> input_tensors_;
+  std::vector<Tensor> output_tensors_;
+};
+
+/// <summary>
+/// A bridge class between `onnxruntime::ep::adapter::OpKernel` and `onnxruntime::OrtKernelImpl`.
```
edgchen1:
```suggestion
/// A bridge class between `onnxruntime::ep::adapter::OpKernel` and `::OrtKernelImpl`.
```

include/onnxruntime/ep/get_capability_utils.h
```diff
@@ -0,0 +1,235 @@
...
+
+  return output_nodes;
+}
+
+// Returns nodes that should be assigned to CPU EP instead of this example EP to avoid costly I/O copies.
+// Based on GetCpuPreferredNodes from onnxruntime/core/framework/fallback_cpu_capability.cc
+inline OrtStatus* GetCpuPreferredNodes(const OrtGraph& ort_graph, OrtEpGraphSupportInfo& graph_support_info,
+                                       const OrtLogger& logger, gsl::span<const OrtNode* const> tentative_nodes,
+                                       /*out*/ std::unordered_set<const OrtNode*>& cpu_preferred_nodes) {
+  try {
```
edgchen1, line 45: if handling exceptions in this function, maybe mark this function `noexcept` so it is obvious to the caller that they don't need to handle them.

include/onnxruntime/ep/README.md
```diff
@@ -0,0 +1,19 @@
...
+  - `include/onnxruntime/ep/` (#26919)
+  - `onnxruntime/test/autoep/library/plugin_ep_utils.h`
+  - `include/onnxruntime/core/providers/utils/` (#25753)
+
+- Header files specifically used for supporting WebGPU EP and CUDA EP to use EP APIs. These header files do not only depend on ONNX Runtime public API, but also depend on ONNX Runtime internal headers.
+  - `include/onnxruntime/ep/adapter/`
+
+### Usage
+
+Make sure to include "ep/adapters.h" for all source code in the implementation. Using PCH compiler flag is recommended.
```
edgchen1: after our discussion, I think I understand the setup better. thanks.

perhaps some more detail would be helpful here too.

maybe something along these lines:

```suggestion
Make sure to include "ep/adapters.h" to include all adapter implementation code. This file brings the adapter classes into the EP's namespace, so it should be included before other EP code that relies on the adapter classes. Using "ep/adapters.h" as a pre-compiled header is the recommended way to include it first.
```

include/onnxruntime/ep/README.md
```diff
@@ -0,0 +1,19 @@
...
+### Folder Structure
+
+There are 2 types of header files:
+
+- General header files for plugin EP. This may include utilities, macros and shared routines that depending on ONNX Runtime public API only. There are multiple places for header files of this category (which we are going to unify them to one place. There is an ongoing discussion about unifying shared headers for plugin EPs):
+  - `include/onnxruntime/ep/` (#26919)
+  - `onnxruntime/test/autoep/library/plugin_ep_utils.h`
+  - `include/onnxruntime/core/providers/utils/` (#25753)
+
+- Header files specifically used for supporting WebGPU EP and CUDA EP to use EP APIs. These header files do not only depend on ONNX Runtime public API, but also depend on ONNX Runtime internal headers.
```
edgchen1: perhaps add some description of the adapter classes too
```suggestion
- Header files specifically used for supporting WebGPU EP and CUDA EP to use EP APIs. These header files do not only depend on ONNX Runtime public API, but also depend on ONNX Runtime internal headers. They define adapter classes that replace their compatible, internal ONNX Runtime equivalents.
```

include/onnxruntime/ep/adapter/kernel_def.h
fs-eire: I choose the term "facade" (which is borrowed from the proxy library) to imply the difference between this approach and "interface" (usually means an abstract class in C++)

edgchen1: I think it could use some clarification. this is not a mainstream definition of "facade", right?

fs-eire: using "interface" may cause it misleading in a different way. do you have suggestions? If no better choice, I can change it to "interface" if you are fine with it

edgchen1: I am fine with "interface"

include/onnxruntime/ep/README.md
```diff
@@ -0,0 +1,19 @@
...
+  - `include/onnxruntime/ep/` (#26919)
+  - `onnxruntime/test/autoep/library/plugin_ep_utils.h`
+  - `include/onnxruntime/core/providers/utils/` (#25753)
+
+- Header files specifically used for supporting WebGPU EP and CUDA EP to use EP APIs. These header files do not only depend on ONNX Runtime public API, but also depend on ONNX Runtime internal headers. They define adapter classes that replace their compatible, internal ONNX Runtime equivalents.
+  - `include/onnxruntime/ep/adapter/`
+
+### Usage
+
+Make sure to include "ep/adapters.h" to include all adapter implementation code. This file brings the adapter classes into the EP's namespace, so it should be included before other EP code that relies on the adapter classes. Using "ep/adapters.h" as a pre-compiled header is the recommended way to include it first.
```
tianleiwu, line 19: Please add a comment that ep/adapters.h has conflicts with shared provider. So shared provider shall be disabled when using these adapters.

include/onnxruntime/ep/get_capability_utils.h
```diff
@@ -0,0 +1,235 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#pragma once
+
+#include <gsl/span>
+#include <unordered_set>
```
tianleiwu: Remove duplicated `#include <unordered_set>`

include/onnxruntime/ep/adapter/kernel_def_builder.h
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  KernelDefBuilder& OutputMemoryType(OrtMemType type, const std::vector<int>& output_indexes) {
+    for (int output_index : output_indexes) {
+      builder_.SetOutputMemType(output_index, type);
+    }
+    return *this;
+  }
+
+  KernelDefBuilder& ExecQueueId(int queue_id) { return *this; }
```
tianleiwu: `int /*queue_id*/` to avoid build error

include/onnxruntime/ep/adapter/op_kernel.h
```diff
@@ -0,0 +1,212 @@
...
+  virtual ~OpKernel() {}
+
+  Node Node() const {
+    return op_kernel_info_.node();
+  }
+  const OpKernelInfo& Info() const {
+    return op_kernel_info_;
+  }
+
+  virtual Status CreateControlFlowKernelImpl(const OrtKernelInfo* info, OrtKernelImpl** impl) {
```
tianleiwu: `const OrtKernelInfo* /*info*/, OrtKernelImpl** /*impl*/` to avoid build error

include/onnxruntime/ep/adapter/op_kernel.h
```diff
@@ -0,0 +1,212 @@
...
+    return Status::OK();
+  }
+
+  virtual Status Compute(OpKernelContext* p_op_kernel_context) const = 0;
+  virtual Status PrePack(const Tensor& tensor,
```
tianleiwu: To avoid build error:
```
  virtual Status PrePack(const Tensor& /*tensor*/,
                         int /*input_idx*/,
                         AllocatorPtr /*alloc*/,
                         /*out*/ bool& is_packed,
                         /*out*/ PrePackedWeights* /*prepacked_weights*/) {
```

include/onnxruntime/ep/adapter/op_kernel.h
```diff
@@ -0,0 +1,212 @@
...
+            typename = std::enable_if_t<std::is_same_v<T, Tensor>>>
+  const T* Input(int index) const {
+    if (index < 0 || static_cast<size_t>(index) >= input_tensors_.size()) {
+      return nullptr;
+    }
+    if (input_tensors_[index].DataRaw() != nullptr) {
+      return &input_tensors_[index];
+    }
+
+    if (index < constant_input_tensors_.size() && constant_input_tensors_[index].DataRaw() != nullptr) {
```
tianleiwu: `if (static_cast<size_t>(index) < constant_input_tensors_.size() && constant_input_tensors_[index].DataRaw() != nullptr) {`

include/onnxruntime/ep/adapter/op_kernel.h
```diff
@@ -0,0 +1,212 @@
...
+    return static_cast<int>(input_tensors_.size());
+  }
+  int OutputCount() const {
+    return static_cast<int>(output_tensors_.size());
+  }
+  bool GetUseDeterministicCompute() const {
+    // TODO(fs-eire): Implement GetUseDeterministicCompute().
+    return false;
+  }
+
```
tianleiwu, line 134: Miss a function:
 
```
void* GetGPUComputeStream() const {
    return context_.GetGPUComputeStream();
  }
```

include/onnxruntime/ep/adapter/tensor_helper.h
```diff
@@ -0,0 +1,52 @@
...
+                                       static_cast<OrtDevice::MemoryType>(memory_info.GetMemoryType()),
+                                       static_cast<OrtDevice::VendorId>(memory_info.GetVendorId()),
+                                       static_cast<OrtDevice::DeviceId>(memory_info.GetDeviceId()),
+
+                                   },
+                                   memory_info.GetMemoryType()};
+
+  return Tensor(data_type,
+                TensorShape{shape.shape, shape.shape_len},
+                const_cast<void*>(value.GetTensorRawData()),
```
edgchen1: hm, this seems dangerous. but it's hard to be const-correct with our existing `Tensor` class.
would it be slightly better to have this function take a non-const `OrtValue*`? then at least the caller will have to be explicit about getting rid of the `const` if they need to.

tianleiwu: Another possible fix is that we have a variant of Tensor constructor accepts const void* data. The tensor object will have a flag that whether it is mutable or not. If it is not mutable, tensor object will not allow calling mutable getter (For example, throw exception if MutableDataRaw is called).

PR #23549, opened 2025-01-31
Whisper Redesigned Solution

kunal-vaishnavi:
### Description

This PR re-designs how Whisper is created and supported in ONNX Runtime. The new solution leverages [previous optimization work](https://github.com/microsoft/onnxruntime/pull/15473), and it is designed to be used in conjunction with [this work](https://github.com/microsoft/onnxruntime-genai/pull/1229) in ONNX Runtime GenAI.

Some of the added changes include:
- Re-designed export that creates new ONNX models without needing a `WhisperBeamSearch` op
  - Creates one encoder model that also pre-computes the cross-attention KV caches (since they only need to be calculated once)
  - Creates one decoder model that can be used during pre-fill and token generation
  - Creates one jump-times model that can be used for word-level timestamps
  - Removes need for a `WhisperBeamSearch` op to chain the encoder and decoder subgraphs
  - Removes need to duplicate decoder's weights in memory
    - Previous solution with the `WhisperBeamSearch` op created an encoder-decoder-init model and decoder-with-past model. The decoder was duplicated twice, one in each.
  - Removes need for separate logic to export the PyTorch model coming from OpenAI vs. the PyTorch model coming from Hugging Face
- Re-factors common parameters and logic used in CPU and CUDA attention kernels
  - Adds `DUMP_STRING` to enable easy logging of intermediate information when running in debug mode to debug a problem. This info is not printed in release mode so it will not impact performance.
  - Integrates `DecoderMaskedMultiHeadAttention` into `MultiHeadAttention`
  - Enables past-present buffer sharing in the `MultiHeadAttention` op for improved performance
  - Adds `cache_indirection` and `past_sequence_length` as new optional inputs to `MultiHeadAttention`
  - Adds `output_qk` as new optional output to `MultiHeadAttention`
  - Enables calculating `output_qk` tensor with FP16 or FP32 precision, regardless of the model's precision
- CI tests that run end-to-end across various flag combinations that are used by many customers internally and externally

The existing solutions are still available if desired.

### Known Issues

- The FP32 CPU model with the `WhisperBeamSearch` op and output QK is currently disabled. This is because ONNX Runtime doesn't currently support output QK kernels on CPU, only on CUDA.
- The `DecoderMaskedMultiHeadAttention` CPU kernel has a parity mismatch with the `DecoderMaskedMultiHeadAttention` CUDA kernel.
[...]

PR #26920, opened 2026-01-06
[CUDA] GQA CUDA Kernel Fusion and Performance Optimization

tianleiwu:
## Summary

This PR significantly improves GroupQueryAttention (GQA) performance on CUDA by fusing multiple kernel launches, improving memory access patterns, and cleaning up sequence length semantics.

### 1. Fused Kernels for Reduced Launch Overhead

| New Kernel | Operations Fused | Kernels Saved |
|------------|------------------|---------------|
| `UnpackQKVWithRoPEAndAppendKV` | Unpack packed QKV + RoPE Q/K + KV cache append | 4-5 |
| `ConcatNewToPastKVFused` | K append + V append (separate buffer mode) | 1 |
| `ConcatKVInPlaceFused` | K append + V append (shared buffer mode) | 1 |

### 2. New `RotaryDispatcher` Template (`rotary_common.cuh`)

Reusable RoPE implementation for fused kernels supporting:
- `float`, `half`, `BFloat16` element types
- `float2`, `float4` vector types
- Interleaved and half-split rotation modes

### 3. Sequence Length Semantics Cleanup

**Before:** Confusing `seqlens_k` / `seqlens_k_buff` with overloaded meanings.

**After:** Clear separation:
- `past_seq_lens` - offset where new tokens are appended
- `total_seq_lens` - total valid tokens after append
- `padded_seq_lens` - padded length for first prompt masking

### 4. FlashAttention Fast Decode Path

New optimized path for token generation (`sequence_length == 1`, shared buffer):
- Bypasses `GetSequenceLengths` kernel
- Passes `past_seq_lens` directly to Flash Attention
- Controlled by `ORT_DISABLE_FLASH_DECODE` env var

### 5. Integer Overflow Prevention

All KV cache index calculations use `int64_t` to handle large `batch * heads * seq * head_size` products.

### 6. BFloat16 Vectorization

Added `float4` (8 elements) vectorized path for BFloat16 in `ConcatTensorToTensor`.

## Environment Variables

| Variable | Default | Description |
|----------|---------|-------------|
| `ORT_DISABLE_FLASH_DECODE` | `false` | Disable fast decode optimization |
| `ORT_DISABLE_FUSED_KV` | `false` | Use unfused K/V append kernels |

### Improved Test Coverage Strategy

Restructured `gqa_cuda_prompt_test_cases()` and `gqa_cuda_past_test_cases()` to explicitly iterate over kernel code path parameters:

```python
# NEW: Primary iteration over kernel code paths
for h in h_sizes_to_test:
    for packed in packed_opts:
        for rotary, rotary_interleaved in rotary_opts:
            for share_buffer in share_buffer_opts:
                # Secondary params (batch, seq, heads) rotate via modulo
```

| Mode | Before | After |
|------|--------|-------|
| Pipeline | 16 tests, 4/12 combos | 42 tests, 8/12 combos |
| Comprehensive | 81 tests, 4/12 combos | 178 tests, 12/12 combos |

### New Test Parameters

- Added `seqs = [(1, 1)]` for edge case testing
- Added `heads = [(3, 1)]` for non-standard GQA ratios
- Added `h_sizes = [40]` for non-power-of-2 head sizes (tests rotary skip logic)

### New Test Configurations

- `share_buffer` config option (tests both buffer modes)
- `has_position_ids` testing on CUDA
- Padding prompt parity test
- Fused vs unfused kernel parity tests (`TestFusedKernelParity`)
- Decoding from empty cache test case `(1, 1)`

## Files Changed

**Core:**
- `group_query_attention_impl.cu` - Main implementation refactoring
- `attention_kv_cache.cu` - Fused append kernels
- `flash_api.cc` - Packed QKV stride handling

**New:**
- `rotary_common.cuh` - Reusable RoPE dispatcher

**Tests:**
- `test_gqa.py` - Extended test coverage

## Performance

For decoding or subsequent prompt, we still use original flash attention kernel, so the performance is almost same as baseline. Here we only show the results of first prompt.

Below are results of benchmark_gqa.py on H200 GPU. Note that the latency is measured from onnx model of a GQA node, so the latency includes extra cost. The kernel speed up can be larger (See profiling results below).

### prompt-sm90-Llama3-8B-b1-h32_8x128-float16
**Configuration**: `batch=1, prompt (past_seq=0), num_heads=32, kv_heads=8, head_size=128, dtype=float16, gpu=H200`

Dense mean Q, K and V are separated inputs. Packed means Q, K and V are packed into one input.

| Sequence Length | Dense Base (ms) | Dense Treat (ms) | **Dense Speedup** | Packed Base (ms) | Packed Treat (ms) | **Packed Speedup** |
| --------------: | --------------: | ---------------: | :---------------- | ---------------: | ----------------: | :----------------- |
|            1024 |           0.470 |            0.277 | **1.70x**         |            0.468 |             0.320 | **1.46x**          |
|            2048 |           1.001 |            0.517 | **1.94x**         |            0.990 |             0.590 | **1.68x**          |
|            4096 |           2.691 |            1.174 | **2.29x**         |            1.504 |             1.242 | **1.21x**          |
|            8192 |           7.780 |            2.292 | **3.39x**         |            7.933 |             4.004 | **1.98x**          |

### prompt-sm90-Llama3-8B-b1-h32_8x128-bfloat16
**Configuration**: `batch=1, prompt (past_seq=0), num_heads=32, kv_heads=8, head_size=128, dtype=bfloat16, gpu=H200`

| Sequence Length | Dense Base (ms) | Dense Treat (ms) | **Dense Speedup** | Packed Base (ms) | Packed Treat (ms) | **Packed Speedup** |
| --------------: | --------------: | ---------------: | :---------------- | ---------------: | ----------------: | :----------------- |
|            1024 |           0.477 |            0.274 | **1.74x**         |            0.486 |             0.332 | **1.46x**          |
|            2048 |           1.078 |            0.500 | **2.16x**         |            1.087 |             0.601 | **1.81x**          |
|            4096 |           2.633 |            1.144 | **2.30x**         |            3.017 |             1.282 | **2.35x**          |
|            8192 |           7.933 |            2.712 | **2.93x**         |            7.933 |             4.003 | **1.98x**          |

# Profiling Comparison (Prompt Phase)

**Summary**:
Switching from `flash_fwd_splitkv_kernel` to standard `flash_fwd_kernel` for the prompt phase (SeqLen=2048) results in a **~3x reduction in attention kernel latency** and a **~2x improvement in total operator latency**.

## 1. Packed QKV
**Configuration**: `batch=1, seq_len=2048, past_seq=0, num_heads=32, kv_heads=8, head_size=128`

| Metric | Baseline | Treatment | Delta |
| :--- | :--- | :--- | :--- |
| **Total Latency** | **639.3 us** | **287.0 us** | **2.23x Speedup** |
| **Attention Kernel** | `flash_fwd_splitkv_kernel`<br>567.10 us | `flash_fwd_kernel`<br>187.70 us | **3.08x Speedup** |
| **Helper Kernels** | `ConcatNewToPastKV`: 4.71 us | `UnpackQKVWithRoPEAndAppendKV`:  32.44 us<br>`GetSequenceLengths`: 1.63 us | *Fused ops added* |

> **Note**: The Treatment implementation introduces a fused `UnpackQKVWithRoPEAndAppendKV` kernel which performs necessary pre-processing. Despite this added cost (~29 us), the massive gain from using the efficient `flash_fwd_kernel` instead of `flash_fwd_splitkv_kernel` yields a significant net speedup.

## 2. Dense (Separated QKV)
**Configuration**: `batch=1, seq_len=2048, past_seq=0, num_heads=32, kv_heads=8, head_size=128`

| Metric | Baseline | Treatment | Delta |
| :--- | :--- | :--- | :--- |
| **Total Latency** | **0.6468 ms** | **0.3226 ms** | **2.00x Speedup** |
| **Attention Kernel** | `flash_fwd_splitkv_kernel`<br>567.25 us | `flash_fwd_kernel`<br> 184.29 us | **3.08x Speedup** |
| **Helper Kernels** | `ConcatNewToPastKV`: 4.68 us | `RotaryEmbeddingBSNH`: 48.94 us<br>`ConcatNewToPastKVFused`: 13.04 us<br>`GetSequenceLengths`: 1.52 us | *See below* |

> **Note**: Similar to the Packed case, the switch to the standard Flash Attention forward kernel drives the performance improvement. The pre-processing is handled by `RotaryEmbeddingBSNH` and `ConcatNewToPastKVFused` in the treatment.

### DIFF
--- a/onnxruntime/contrib_ops/cuda/bert/group_query_attention.cc
+++ b/onnxruntime/contrib_ops/cuda/bert/group_query_attention.cc
@@ -162,6 +162,7 @@ Status GroupQueryAttention<T>::ComputeInternal(OpKernelContext* context) const {
   IAllocatorUniquePtr<void> k_buffer;
   IAllocatorUniquePtr<void> v_buffer;
   IAllocatorUniquePtr<void> rotary_buffer;
+  IAllocatorUniquePtr<void> position_ids_buffer;
   IAllocatorUniquePtr<void> fmha_buffer;
   IAllocatorUniquePtr<void> unpacked_qkv_buffer;
   IAllocatorUniquePtr<int> seq_lens_buffer;
@@ -249,24 +250,18 @@ Status GroupQueryAttention<T>::ComputeInternal(OpKernelContext* context) const {
         !disable_memory_efficient_attention_ &&
         has_memory_efficient_attention(sm, std::is_same<T, MLFloat16>::value, std::is_same<T, BFloat16>::value, parameters.head_size, parameters.head_size);
 
+    // KV buffer for head expansion (when num_heads != kv_num_heads)
     size_t kv_buffer_bytes = (use_memory_efficient_attention && (parameters.num_heads != parameters.kv_num_heads))
                                  ? (sizeof(T) * parameters.batch_size * parameters.num_heads * parameters.seqlen_present_kv_cache * parameters.head_size)
                                  : 0;
-    size_t rotary_buffer_bytes = use_memory_efficient_attention && do_rotary_
-                                     ? (sizeof(T) * 2 * parameters.batch_size * parameters.num_heads * parameters.sequence_length * parameters.head_size +
-                                        sizeof(int64_t) * parameters.batch_size * parameters.sequence_length)
-                                     : 0;
+    // FMHA workspace
     size_t fmha_buffer_bytes = (use_memory_efficient_attention && MemoryEfficientAttentionParams::need_workspace(parameters.head_size, sizeof(T) == sizeof(float)))
                                    ? (sizeof(float) * parameters.batch_size * parameters.sequence_length * parameters.num_heads * parameters.head_size)
                                    : 0;
-    size_t unpacked_qkv_bytes = use_memory_efficient_attention && parameters.is_packed_qkv
-                                    ? (sizeof(T) * parameters.batch_size * parameters.sequence_length * (parameters.num_heads + 2 * parameters.kv_num_heads) * parameters.head_size)
-                                    : 0;
+
     k_buffer = GetScratchBuffer<void>(kv_buffer_bytes, context->GetComputeStream());
     v_buffer = GetScratchBuffer<void>(kv_buffer_bytes, context->GetComputeStream());
-    rotary_buffer = GetScratchBuffer<void>(rotary_buffer_bytes, context->GetComputeStream());
     fmha_buffer = GetScratchBuffer<void>(fmha_buffer_bytes, context->GetComputeStream());
-    unpacked_qkv_buffer = GetScratchBuffer<void>(unpacked_qkv_bytes, context->GetComputeStream());
 #else
     constexpr bool use_memory_efficient_attention = false;
 #endif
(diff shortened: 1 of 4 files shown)


onnxruntime/contrib_ops/cuda/bert/attention_impl.cu
```diff
@@ -918,7 +918,7 @@ Status PastPresentBufferShare(int batch_size, int num_heads, int qk_head_size, i
     constexpr bool is_new_kv_bnsh_format = true;
     ORT_RETURN_IF_ERROR(LaunchConcatKVInPlace(
         batch_size, num_heads, qk_head_size, parameters.max_sequence_length,
-        data.seqlens_k_total, nullptr, parameters.sequence_length, data.k, data.v, data.present_key, data.present_value,
+        nullptr, data.seqlens_k_total, parameters.sequence_length, data.k, data.v, data.present_key, data.present_value,
```
apsonawane, line 921: So here past_seq_lens is nullptr and not total_seq_lens? Was that a bug earlier?

tianleiwu: @kunal-vaishnavi, looks like a bug introduced in https://github.com/microsoft/onnxruntime/pull/23549. It will cause past_seq_len be actual past length +1 and cause appending position shifts one.

onnxruntime/test/python/transformers/test_gqa.py
```diff
@@ -1071,12 +1085,164 @@ def parity_check_gqa_past(
     present_k_np = present_k.to(torch.float32).detach().cpu().numpy()
     present_v_np = present_v.to(torch.float32).detach().cpu().numpy()
 
+    if not config.share_buffer:
+        total_len = config.past_kv_sequence_length + config.q_sequence_length
+        k_cache_ref_np = k_cache_ref_np[:, :, :total_len, :]
+        v_cache_ref_np = v_cache_ref_np[:, :, :total_len, :]
```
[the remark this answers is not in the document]
tianleiwu: See line 965 has a comment "Random past sequence lengths. This tests paddings in decoding". The function is used to test non-prompt cases where past_seq_len > 0, so both subsequent prompt and token generation has random padding.  In summary, padding is adequately tested.

onnxruntime/contrib_ops/cuda/bert/attention_impl.cu
kunal-vaishnavi: Thanks for catching this. Looks like the parameter order was mixed when invoking `LaunchConcatKVInPlace`.

apsonawane (dismissed):

kunal-vaishnavi (dismissed):

onnxruntime/contrib_ops/cuda/bert/attention_kv_cache.cu
```diff
@@ -772,6 +959,223 @@ template Status LaunchConcatKVInPlace<float>(int batch_size,
...
+//   is_past_kv_bnsh_format - True if KV buffer is BNSH, false for BSNH
+//   is_new_kv_bnsh_format  - True if new K/V is BNSH, false for BSNH
+//
+// OUTPUTS:
+//   k_buff - Updated K cache
+//   v_buff - Updated V cache
+//
+// NOTE:
+//   RoPE should be applied BEFORE calling this kernel.
+//   For fused RoPE+append, use ConcatNewToPastKVFused instead.
```
kunal-vaishnavi, line 983: Should we rename the kernels to make it clear in the names then?

onnxruntime/contrib_ops/cuda/bert/group_query_attention.cc
```diff
@@ -35,6 +38,9 @@ namespace cuda {
 REGISTER_KERNEL_TYPED(MLFloat16)
 REGISTER_KERNEL_TYPED(BFloat16)
 
+constexpr const char* kDisableFlashDecode = "ORT_DISABLE_FLASH_DECODE";
```
kunal-vaishnavi, line 41: Can flash decode be used in other ops such as Attention and MHA? If so, should this be in a different file to enable disabling flash decode for those ops?

kunal-vaishnavi:
Should we consider fusing RMSNorm inside RoPE as well? Some models such as [Qwen-3](https://github.com/huggingface/transformers/blob/37974267efefe020168ff27081fbab8bbce04720/src/transformers/models/qwen3/modeling_qwen3.py#L266-L271) apply RMSNorm to Q and K after MatMul and before RoPE.

onnxruntime/contrib_ops/cuda/bert/group_query_attention.cc
tianleiwu: Currently, the input parameters and data is GQA specified. 

In the future, we can do some refactoring to make it easily works with other ops like MHA etc.

onnxruntime/contrib_ops/cuda/bert/attention_kv_cache.cu
tianleiwu: I could remove "Fused" and use ConcatKVInPlace for this kernel name.

Let me address it in next GQA PR.

tianleiwu:
> Should we consider fusing RMSNorm inside RoPE as well? Some models such as [Qwen-3](https://github.com/huggingface/transformers/blob/37974267efefe020168ff27081fbab8bbce04720/src/transformers/models/qwen3/modeling_qwen3.py#L266-L271) apply RMSNorm to Q and K after MatMul and before RoPE.

It is feasible. We can support it in the future.

kunal-vaishnavi (approved):

PR #26922, opened 2026-01-06
Add EP Device Compatibility APIs

xhan65:
### Description

This PR adds new C API for querying hardware device and execution provider (EP) compatibility/incompatibility reasons. The API allows applications to determine why a specific EP may not be compatible with a given hardware device before attempting to use it. The EP implementor decides how device compatibility is defined for their stack and communicates this via the possible states in the OrtDeviceEpIncompatibilityDetails struct.

New API Functions
- GetHardwareDeviceEPIncompatibilityReasons - Query incompatibility reasons between an EP and a hardware device
- DeviceEpIncompatibilityDetails_GetReasonsBitmask - Get a bitmask of incompatibility reasons
- DeviceEpIncompatibilityDetails_GetErrorCode - Get an EP-specific error code
- DeviceEpIncompatibilityDetails_GetNotes - Get optional human-readable notes about incompatibility
- ReleaseDeviceEpIncompatibilityDetails - Release the details object

New Types
- OrtDeviceEpIncompatibilityDetails - Opaque type holding incompatibility information
- OrtDeviceEpIncompatibilityReason - Enum defining standard incompatibility reason flags

Testing
- Added unit tests in hardware_device_compatibility_test.cc for the new C API functions
- Added plugin EP compatibility tests in test_execution.cc
. Tests verify both compatible (CPU device with CPU-supporting EP) and incompatible (GPU device with CPU-only EP) scenarios
- Updated example plugin EP to implement the compatibility checking interface

### Motivation and Context

Applications using ONNX Runtime need a way to understand why an execution provider cannot run on a particular hardware device. This is especially important for:

- User feedback - Providing meaningful error messages when hardware is incompatible
- EP selection - Choosing the best available EP based on compatibility
- Diagnostics - Understanding driver/dependency requirements

Previously, there was no standardized way to query this information. EPs would simply fail at session creation/model load/inference time without providing actionable feedback.

### DIFF
--- a/include/onnxruntime/core/session/onnxruntime_c_api.h
+++ b/include/onnxruntime/core/session/onnxruntime_c_api.h
@@ -6912,43 +6912,6 @@ struct OrtApi {
   ORT_CLASS_RELEASE(DeviceEpIncompatibilityDetails);
 
   /// @}
-
-  /** \brief Get the EP Interop API instance.
-   *
-   * Get the Interop API instance to work with external resources. This API provides functions
-   * for importing external GPU memory and semaphores for zero-copy sharing between ORT inference
-   * and other GPU workloads.
-   *
-   * \return Interop API struct instance.
-   *
-   * \since Version 1.24.
-   */
-  const OrtInteropApi*(ORT_API_CALL* GetInteropApi)(void);
-
-  /** \brief Get the EP device assigned to each session output.
-   *
-   * Returns the OrtEpDevice assigned to each output of the session after graph partitioning.
-   * This allows validation that outputs are placed on the expected device for external resource sharing.
-   *
-   * The EP device for each output is determined by which execution provider will produce that output
-   * during inferencing. This information is useful for:
-   * - Validating that outputs will be placed on the expected device for external resource sharing
-   * - Deciding whether to use external memory handles for outputs
-   *
-   * \param[in] session The OrtSession instance to query.
-   * \param[out] outputs_ep_devices An array to be filled with the EP device for each output.
-   *                                The array must be allocated by the caller with space for
-   *                                OrtEpDevice* values for each output.
-   *                                The order is the same as returned by SessionGetOutputName.
-   * \param[in] num_outputs The number of outputs in the session. Must match SessionGetOutputCount.
-   *
-   * \snippet{doc} snippets.dox OrtStatus Return Value
-   *
-   * \since Version 1.24
-   */
-  ORT_API2_STATUS(SessionGetEpDeviceForOutputs, _In_ const OrtSession* session,
-                  _Out_writes_(num_outputs) const OrtEpDevice** outputs_ep_devices,
-                  _In_ size_t num_outputs);
 };
 
 /*


include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -6667,6 +6678,104 @@ struct OrtApi {
...
+   * \param[in] env The OrtEnv instance where device discovery results are stored.
+   * \param[out] devices The OrtHardwareDevice instances that are available.
+   * \param[out] num_devices The number of OrtHardwareDevice instances returned.
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \since Version 1.24.
+   */
+  ORT_API2_STATUS(GetOrtHardwareDevices, _In_ const OrtEnv* env,
+                  _Outptr_ const OrtHardwareDevice* const** devices,
```
adrianlizarraga: If I'm not mistaken, this returns a pointer to the entire array owned by the environment (not just the elements). If ORT ever reallocates the array, then the user's pointer would be silently invalidated.

I would recommend having the user allocate the array, passing their array to this function, and then having ORT copy the elements (of type `const OrtHardwareDevice*`) into the user's array.  IMO, this is less error prone and more consistent with other existing APIs that return elements.

There's at least two different ways to accomplish the above:
1. Use two API functions: `GetNumHardwareDevices()`, `GetHardwareDevices()`. The user gets the number of elements with the `GetNumHardwareDevices()`, pre-allocates their own array (e.g., `std::vector<> devices(num_devices)`), and calls `GetHardwareDevices()` to fill out their array.
2. Use a single `GetHardwareDevices()` API function that operates in two different modes: 
  - if the `devices` argument is null, it sets `num_devices` equal to the number of elements and returns a success status.
  - if the `devices` argument is not null, it copies the devices into the user's array.

I personally prefer the first approach but either one works.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -983,6 +983,25 @@ struct OrtEpApi {
...
+   * \param[in] error_code EP-specific error code (0 = no error).
+   * \param[in] notes Optional human-readable notes. Can be null.
+   * \param[out] details Output parameter set to the new OrtDeviceEpIncompatibilityDetails instance.
+   *                     Must be released by ORT using ReleaseDeviceEpIncompatibilityDetails.
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \since Version 1.24.
+   */
+  ORT_API2_STATUS(CreateDeviceEpIncompatibilityDetails, _In_ uint32_t reasons_bitmask,
```
adrianlizarraga: There's a bit of an asymmetry here where the object is created with ep api, but deleted with the ort api. I wonder if it would be simpler (and more consistent) if the EP didn't create the object, but instead the ep initialized an existing object created by ORT.

The workflow could be:
1. Application calls `GetHardwareDeviceEPIncompatibilityReasons()`.
2. ORT internally allocates an `OrtDeviceEpIncompatibilityDetails` object.
3. ORT calls `OrtEpFactory::GetHardwareDeviceIncompatibilityReasons()` and passes along the existing object.
4. The ep factory initializes the object with some api like `DeviceEpIncompatibilityDetails_Initialize()`.
5. ORT returns the initialized object to the application.
6. The app can call `OrtApi::Release*` to release the object. In this case, both object creation and release happen in the ORT API.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -507,6 +508,16 @@ typedef enum OrtExecutionProviderDevicePolicy {
...
 } OrtExecutionProviderDevicePolicy;
 
+/** \brief Reasons why an execution provider might not be compatible with a device
+ */
+typedef enum OrtDeviceEpIncompatibilityReason {
+  OrtDeviceEpIncompatibility_NONE = 0,
+  OrtDeviceEpIncompatibility_DRIVER_INCOMPATIBLE = 1 << 0,
+  OrtDeviceEpIncompatibility_DEVICE_INCOMPATIBLE = 1 << 1,
+  OrtDeviceEpIncompatibility_MISSING_DEPENDENCY = 1 << 2,
+  OrtDeviceEpIncompatibility_UNKNOWN = 1 << 30
```
adrianlizarraga: Is this supposed to be shifted by `31` (instead of 30)?

xhan65: I had it 31, but the build failed on some mac os pipeline due to integer too large, so I made it smaller

include/onnxruntime/core/session/onnxruntime_c_api.h
xhan65: The reason I made it return the pointer to the array is because GetEPDevices is using the same pattern. I'm happy to fix it if our recommendation has shifted.

include/onnxruntime/core/session/environment.h
```diff
@@ -146,6 +146,17 @@ class Environment {
     return execution_devices_;
   }
 
+  /// Get hardware device incompatibility reasons for a specific EP.
+  /// @param ep_name The name of the execution provider to check.
+  /// @param hw The hardware device to check for incompatibility.
+  /// @param details Output: Incompatibility details including reasons for incompatibility if any.
+  /// @returns Status indicating success or failure.
+  Status GetHardwareDeviceEPIncompatibilityReasons(const std::string& ep_name,
```
edgchen1: general naming: use "Ep" instead of "EP" in function and type names

include/onnxruntime/core/session/onnxruntime_c_api.h
edgchen1: hm, I think `GetEpDevices` has the same issue. and it's more serious there because the underlying vector gets modified as EP libraries are registered/unregistered.

include/onnxruntime/core/session/environment.h
```diff
@@ -242,6 +253,8 @@ class Environment {
 
   DataTransferManager data_transfer_mgr_;  // plugin EP IDataTransfer instances
 
+  mutable std::vector<const OrtHardwareDevice*> ort_hardware_devices_;
```
edgchen1: there's already a static instance here:
https://github.com/microsoft/onnxruntime/blob/39d8520b9bfe16217d56be38293fc82556d67b5b/onnxruntime/core/session/environment.cc#L790-L792

can we reuse a single instance? this isn't expected to change so a static instance should be ok.

e.g., with a helper function:
```c++
const std::vector<const OrtHardwareDevice*>& GetOrtHardwareDevices() {
  static const auto hardware_devices = SortDevicesByType();
  return hardware_devices;
}
```

then we can also avoid a mutable data member.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -6667,6 +6678,104 @@ struct OrtApi {
...
+   * \param[in] ep_name The name of the execution provider to check. Required and cannot be null or empty.
+   * \param[in] hw The hardware device to check for incompatibility.
+   * \param[out] details Compatibility details including reasons for incompatibility if any.
+   *                     Must be freed with OrtApi::ReleaseDeviceEpIncompatibilityDetails.
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \since Version 1.24.
+   */
+  ORT_API2_STATUS(GetHardwareDeviceEPIncompatibilityReasons, _In_ const OrtEnv* env,
```
edgchen1: would a better name for this function be `GetHardwareDeviceEpIncompatibilityDetails`? "reasons" is used in the enum name, but this function is actually returning a "details" instance.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -6667,6 +6678,104 @@ struct OrtApi {
...
+                  _Outptr_ const OrtHardwareDevice* const** devices,
+                  _Out_ size_t* num_devices);
+
+  /** \brief Check for known incompatibility reasons between hardware device and a specific execution provider.
+   *
+   * This function checks for known incompatibility reasons between the specified hardware device
+   * and a specific execution provider.
+   * If incompatibility reasons are returned, it indicates the device is not compatible.
+   * However, if no reasons are returned, it doesn't guarantee 100% compatibility for all models,
+   * as models may have specific requirements.
```
edgchen1: sounds like it is possible to not return "reasons". does that imply that the `*details` may be set to null?

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -1564,6 +1583,29 @@ struct OrtEpFactory {
...
+   *                     ORT will take ownership and call ReleaseDeviceEpIncompatibilityDetails.
+   *
+   * \note Implementation of this function is optional.
+   *       If not implemented, ORT will assume the device is compatible with this EP.
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \since Version 1.24.
+   */
+  ORT_API2_STATUS(GetHardwareDeviceIncompatibilityReasons, _In_ OrtEpFactory* this_ptr,
```
edgchen1: similarly, would `GetHardwareDeviceIncompatibilityDetails` be a better name?

onnxruntime/core/session/environment.cc
```diff
@@ -635,6 +635,112 @@ Status Environment::UnregisterExecutionProviderLibrary(const std::string& ep_nam
...
+      return ToStatusAndRelease(status);
+    }
+
+    if (factory_details != nullptr) {
+      details.reset(factory_details);
+      return Status::OK();
+    }
+  }
+
+  // Factory doesn't implement the hook or returned nullptr - create empty details (compatible)
```
edgchen1: what's the difference between nullptr vs an empty details instance? could the "compatible" case be represented more consistently?

onnxruntime/test/autoep/test_execution.cc
```diff
@@ -352,5 +352,100 @@ TEST(OrtEpLibrary, KernelPluginEp_Inference) {
...
+
+  // Check compatibility - ExampleEP supports CPU, so should return no incompatibility reasons
+  OrtDeviceEpIncompatibilityDetails* details = nullptr;
+  ASSERT_ORTSTATUS_OK(api->GetHardwareDeviceEPIncompatibilityReasons(env, Utils::example_ep_info.registration_name.c_str(),
+                                                                     cpu_device, &details));
+  ASSERT_NE(details, nullptr);
+
+  // Verify compatible (no incompatibility reasons)
+  uint32_t reasons_bitmask = 0xFFFFFFFF;
+  ASSERT_ORTSTATUS_OK(api->DeviceEpIncompatibilityDetails_GetReasonsBitmask(details, &reasons_bitmask));
```
edgchen1, line 592: can we add C++ APIs? those will be easier to use in test code too.

xhan65: Yes I will add c++/c#/python API in another PR

include/onnxruntime/core/session/onnxruntime_c_api.h
xhan65: No the returned detail is always not null. When no incompatibility is found, the returned details should have reason == 0. I updated the comment to meke it clearer.

onnxruntime/core/session/environment.cc
xhan65: Modified with latest update - since details are initialized by ORT, EP can leave it unchanged if it's compatible.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
xhan65: Fixed

include/onnxruntime/core/session/onnxruntime_c_api.h
xhan65: Resolved.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -1990,6 +2010,30 @@ struct OrtEpFactory {
...
+   *
+   * \param[in] this_ptr The OrtEpFactory instance.
+   * \param[in] hw The hardware device to check for incompatibility.
+   * \param[in,out] details Pre-allocated incompatibility details object created by ORT.
+   *                        The EP should initialize this using OrtEpApi::DeviceEpIncompatibilityDetails_Initialize
+   *                        to set any incompatibility information. If the device is compatible, the EP can
+   *                        leave the object uninitialized (it defaults to no incompatibility).
+   *
+   * \note Implementation of this function is optional.
+   *       If not implemented, ORT will assume the device is compatible with this EP.
```
javier-intel, line 2028: Why is compatibility assumed? I'd expect the opposite, a hw_device is incompatible unless the EP creates an ep_device for it. Unless, the EP supports this API and returns compatible.

onnxruntime/test/autoep/test_execution.cc
```diff
@@ -461,6 +462,102 @@ TEST(OrtEpLibrary, KernelPluginEp_Inference) {
...
+  // Check compatibility - ExampleEP only supports CPU, so GPU should return incompatibility reasons
+  OrtDeviceEpIncompatibilityDetails* details = nullptr;
+  ASSERT_ORTSTATUS_OK(api->GetHardwareDeviceEpIncompatibilityDetails(env, Utils::example_ep_info.registration_name.c_str(),
+                                                                     gpu_device, &details));
+  ASSERT_NE(details, nullptr);
+
+  // Verify incompatible (should have incompatibility reasons)
+  uint32_t reasons_bitmask = 0;
+  ASSERT_ORTSTATUS_OK(api->DeviceEpIncompatibilityDetails_GetReasonsBitmask(details, &reasons_bitmask));
+  EXPECT_NE(reasons_bitmask, 0u) << "GPU device should be incompatible with example_plugin_ep (CPU-only)";
```
javier-intel, line 645: What is the expected error in this case, `OrtDeviceEpIncompatibility_DEVICE_INCOMPATIBLE`?

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
xhan65: We're not assuming compatibility. We specifically mentionedin the onnxruntime_c_api.h (developer facing one) that "If returned incompatibility details have non-zero reasons, it indicates the device is not compatible. However, if returned detail have reason == 0, it doesn't guarantee 100% compatibility for all models, as models may have specific requirements." This API gives developer an opportunity to know that a device might not work ahead of time, and what action they can take to make it working (i.e. remind user to update driver if it returned incompatibility reason driver not compatible).

In our previous discussions with IHVs, almost all of them were hesitant in doing device filtering in GetEpDevices. Right now I believe most IHV returns all devices from their vendor in that API, regardless of compatibility state, so we can't assume that devices returned from that API are all compatible.

onnxruntime/test/autoep/test_execution.cc
xhan65: Correct. OrtDeviceEpIncompatibility_DEVICE_INCOMPATIBLE should be returned if the device is incompatible, so if the EP doesn't support this device type, the device family, or the device vendor

include/onnxruntime/core/session/onnxruntime_c_api.h
edgchen1: what's the error? would `1u << 31` work?

include/onnxruntime/core/session/environment.h
```diff
@@ -146,6 +146,17 @@ class Environment {
     return execution_devices_;
   }
 
+  /// Get hardware device incompatibility reasons for a specific EP.
```
edgchen1:
```suggestion
  /// Get hardware device incompatibility details for a specific EP.
```

onnxruntime/core/session/onnxruntime_c_api.cc
```diff
@@ -4368,3 +4374,132 @@ DEFINE_RELEASE_ORT_OBJECT_FUNCTION(Value, OrtValue)
...
+  if (env == nullptr) {
+    return OrtApis::CreateStatus(ORT_INVALID_ARGUMENT, "env must not be null");
+  }
+  if (devices == nullptr) {
+    return OrtApis::CreateStatus(ORT_INVALID_ARGUMENT, "devices must not be null");
+  }
+
+  const auto& device_vector = env->GetEnvironment().GetSortedOrtHardwareDevices();
+  size_t available_devices = device_vector.size();
+  size_t copy_count = std::min(num_devices, available_devices);
```
edgchen1: should we ensure that `num_devices` is large enough?

onnxruntime/core/session/environment.cc
```diff
@@ -637,6 +637,110 @@ Status Environment::UnregisterExecutionProviderLibrary(const std::string& ep_nam
...
+  if (matched_factory == nullptr) {
+    return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT,
+                           "No valid factory found for execution provider '", ep_name, "'.");
+  }
+
+  // ORT creates the details object with default values (compatible)
+  details = std::make_unique<OrtDeviceEpIncompatibilityDetails>();
+  details->reasons_bitmask = 0;
+  details->error_code = 0;
+  details->notes = "";
```
edgchen1: nit: we don't need to repeat the default values here - it should already be handled by `OrtDeviceEpIncompatibilityDetails`'s default member initializers.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -1308,6 +1308,26 @@ struct OrtEpApi {
    */
   ORT_API2_STATUS(KernelInfo_GetEp, _In_ const OrtKernelInfo* info, _Outptr_ const OrtEp** ep);
 
+  /** \brief Initialize an OrtDeviceEpIncompatibilityDetails instance.
```
edgchen1: naming nit: regarding "Initialize", the instance passed to the EP is already initialized, right? users are not required to do any further initialization to make it usable. maybe something like "SetDetails" would be more accurate.

include/onnxruntime/core/session/onnxruntime_c_api.h
xhan65: Good point. We should fix that with a V2 API in a future release.

include/onnxruntime/core/session/onnxruntime_c_api.h
xhan65: Fixed. I think my original code 1u << 31 caused issue, and 1 << 31 seems fine.

adrianlizarraga (dismissed):

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
javier-intel: Ahh I missed that bit about not being a 100% guarantee; that makes sense and thanks for pointing it out.

javier-intel (approved):

onnxruntime/test/autoep/test_execution.cc
edgchen1: at that point, please update the tests to use the C++ API. it would make for better example code in the unit tests. C API usage like this has the potential to leak resources (e.g., not calling `ReleaseDeviceEpIncompatibilityDetails`) with all the early returns.

edgchen1 (dismissed):

adrastogi (dismissed):

adrianlizarraga (approved):

rwinterton:
Determining the optimal EP for execution requires inputs from the application, the OS, and the EP driver. To establish priority, the application must provide the OS with a deadline. While the OS retains exclusive control over thread scheduling, the hardware manages frequency and thermal constraints. Ultimately, the API should be simple yet provide enough data for each component to make the correct decision.

PR #26923, opened 2026-01-07
Fix ClipQuantFusion crash when Clip has multiple input edges

qti-yuduo:
## Motivation

`ClipQuantFusion in clip_quantizelinear.cc` calls `graph_utils::RemoveNode()` without first checking `graph_utils::CanRemoveNode()`. When a Clip node has min/max inputs from DequantizeLinear nodes (instead of initializers), it has multiple input edges. `RemoveNode()` throws exception:

```
 [ONNXRuntimeError] : 6 : RUNTIME_EXCEPTION : Exception during initialization: graph_utils.cc:650 bool onnxruntime::graph_utils::RemoveNode(onnxruntime::Graph&, onnxruntime::Node&) Should be unreachable if CanRemoveNodeAndMergeEdges is in sync with the logic here.
```

## Fix:
Added `CanRemoveNode()` check to `ClipQuantFusion::SatisfyCondition()` to skip nodes that cannot be safely removed. 

## Test:
Added `ClipQuantFusion_MultipleInputEdges` test that creates a Clip node with min from a DQ node (2 input edges) and verifies the optimizer doesn't crash.

### DIFF
--- a/onnxruntime/core/optimizer/qdq_transformer/clip_quantizelinear.cc
+++ b/onnxruntime/core/optimizer/qdq_transformer/clip_quantizelinear.cc
@@ -81,7 +81,7 @@ static bool GetQConstantLowerUpper(const Graph& graph, const Node& node, float& 
   return true;
 }
 
-bool ClipQuantFusion::SatisfyCondition(const Graph& graph, const Node& node, const logging::Logger& /*logger*/) const {
+bool ClipQuantFusion::SatisfyCondition(const Graph& graph, const Node& node, const logging::Logger& logger) const {
   if (!graph_utils::IsSupportedOptypeVersionAndDomain(node, "Clip", {1, 6, 11, 12, 13}) ||
       !graph_utils::IsSupportedProvider(node, {kCpuExecutionProvider}) ||
       !optimizer_utils::CheckOutputEdges(graph, node, 1)) {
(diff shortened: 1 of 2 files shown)


tianleiwu:
From AI
--------
## Summary
This PR fixes a runtime exception in the `ClipQuantFusion` optimizer. The crash occurred because the code attempted to remove a `Clip` node without verifying if it was safe to do so, specifically when the node had multiple input edges (e.g., dynamic min/max inputs from DequantizeLinear nodes).

## Key Changes
-   **Fix**: Added a call to `graph_utils::CanRemoveNode()` within `ClipQuantFusion::SatisfyCondition()`.
-   **Test**: Added a new test case `ClipQuantFusion_MultipleInputEdges` to reproduce the scenario and verify the fix.

## Review Analysis
### Correctness
-   **Pattern**: It is a standard and necessary practice in ONNX Runtime graph transformations to check `CanRemoveNode()` before calling `RemoveNode()`. The fix correctly implements this missing check.
-   **Root Cause**: The description clearly explains why the crash happened (multiple input edges causing `RemoveNode` to fail), and the fix directly addresses it.

### Performance
-   **Impact**: Negligible. The added check is lightweight and prevents a crash, which is infinitely better than failing.

## Conclusion
This is a straightforward and correct bug fix. The inclusion of a regression test gives high confidence in the solution. **LGTM**.

edgchen1:
hm, there appears to be an invisible unresolved comment that's blocking the merge

edgchen1:
can't figure out how to merge this one. replacing with #27016.

PR #26927, opened 2026-01-07
[EP ABI] Add APIs to create control flow kernels for plugin EPs

adrianlizarraga:
### Description
Adds APIs to enable plugin EPs to create and register kernels for control flow operators (If, Loop, and Scan). The implementation provides ORT-managed kernel implementations that handle subgraph execution while allowing EPs to provide device-specific helper functions for operations like tensor concatenation and transposition.

Key changes:

- Adds four EP API functions: `CreateIfKernel`, `CreateLoopKernel`, `CreateScanKernel`, and `ReleaseKernelImpl` for creating control flow kernel implementations
- Introduces public helper structures (`OrtLoopKernelHelper`, `OrtScanKernelHelper`) that EPs implement to provide device-specific operations
- Updates the example kernel-based EP with kernel registrations for all control flow operators and adds corresponding test models.

### DIFF
--- a/onnxruntime/core/session/plugin_ep/ep_api.cc
+++ b/onnxruntime/core/session/plugin_ep/ep_api.cc
@@ -748,7 +748,7 @@ ORT_API_STATUS_IMPL(CreateScanKernel, _In_ const OrtKernelInfo* kernel_info, _In
     // Note: CPU EP always uses Scan<9> for all opsets >= 9.
     auto kernel_unique_ptr = std::make_unique<PluginEpScanKernelImpl>(*op_kernel_info, helper);
     *kernel_out = kernel_unique_ptr.release();
-  } else /*if (opset < 8)*/ {
+  } else {
     return OrtApis::CreateStatus(ORT_FAIL,
                                  "Kernel implementations for Scan older than opset version 9 are not supported");
   }
--- a/onnxruntime/core/session/plugin_ep/ep_kernel_registration.cc
+++ b/onnxruntime/core/session/plugin_ep/ep_kernel_registration.cc
@@ -218,11 +218,9 @@ Status PluginEpOpKernel::Create(FuncManager& /*fn_manager*/, const OpKernelInfo&
 
   const auto& op_type = info.node().OpType();
   const auto& node_name = info.node().Name();
-  std::string_view ep_name = "(unknown ep)";
-
-  if (const auto* ep = info.GetExecutionProvider(); ep != nullptr) {
-    ep_name = ep->Type();
-  }
+  const auto* ep = info.GetExecutionProvider();
+  ORT_ENFORCE(ep != nullptr, "IExecutionProvider* retrieved from OpKernelInfo should never be nullptr");
+  const auto& ep_name = ep->Type();
 
   // Do some basic checks for the OrtKernelImpl provided by the EP. Other checks for missing function implementations
   // that are only required in certain situations (e.g., pre-packing) happen later as soon as we know they are required.


onnxruntime/core/session/plugin_ep/ep_api.cc
```diff
@@ -711,6 +814,10 @@ static constexpr OrtEpApi ort_ep_api = {
     &OrtExecutionProviderApi::EpGraphSupportInfo_LookUpKernel,
     &OrtExecutionProviderApi::SharedPrePackedWeightCache_StoreWeightData,
     &OrtExecutionProviderApi::KernelInfo_GetEp,
+    &OrtExecutionProviderApi::CreateIfKernel,
+    &OrtExecutionProviderApi::CreateLoopKernel,
+    &OrtExecutionProviderApi::CreateScanKernel,
+    &OrtExecutionProviderApi::ReleaseKernelImpl,
```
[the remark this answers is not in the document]
adrianlizarraga: Not yet.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -290,6 +300,7 @@ typedef struct OrtKernelImpl OrtKernelImpl;
  */
 struct OrtKernelImpl {
   uint32_t ort_version_supported;  ///< Must be initialized to ORT_API_VERSION
+  uint32_t creator;                ///< EP must set to ORT_KERNEL_IMPL_CREATOR_EP (0) if it created the OrtKernelImpl
```
skottmckay: Does this need to be added at the end of the struct? if someone passes in an old version we know it won't have a `creator` field based on `ort_version_supported`, but it will also have the original ordering so I'm not sure that would match.

One other thought would be whether a `flags` field would work better with something like a 'use ORT control flow kernel' as an initial flag. feels slightly more natural to me to set that to 0 as the default and it would allow other use cases.  #Closed

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -421,6 +434,82 @@ typedef OrtStatus*(ORT_API_CALL* OrtKernelCreateFunc)(_In_ void* kernel_create_f
...
+
+  /** \brief Called by ORT to release the OrtScanKernelHelper instance and its resources.
+   *
+   * \param[in] this_ptr The OrtScanKernelHelper instance.
+   *
+   * \since Version 1.24.
+   */
+  ORT_API_T(void, Release, _In_ OrtScanKernelHelper* this_ptr);
+
+  /** \brief Helper function that transposes an OrtValue instance during execution of a Scan kernel.
```
skottmckay, line 726: Maybe note it's required to support Scan(9) and later when scan_input_axes  or scan_output_axes contain non-zero values. #Closed

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -976,13 +1065,74 @@ struct OrtEpApi {
...
    * \snippet{doc} snippets.dox OrtStatus Return Value
    * \since Version 1.24
    */
   ORT_API2_STATUS(KernelInfo_GetEp, _In_ const OrtKernelInfo* info, _Outptr_ const OrtEp** ep);
+
+  /** \brief Creates an OrtKernelImpl instance for an If operator.
+   *
+   * \note Control flow operators normally require access to ORT session internals to orchestrate subgraph operations,
+   *       This function allows an EP to create a properly configured kernel for an If operator that the EP can then add
+   *       to its kernel registry.
```
skottmckay: Might be worth mentioning that the data inputs/outputs of the control flow node will use the device of the EP registering the kernel. i.e. we avoid copying to/from CPU where applicable. 

exception is the `cond` and `M` inputs of Loop as those values are used by the CPU logic that implements the subgraph execution loop so we need those on CPU. the kernel registration handles setting that up. #Closed

onnxruntime/core/framework/session_state.cc
```diff
@@ -1183,11 +1183,15 @@ const InlinedHashSet<NodeIndex>* SessionState::GetToBeExecutedRange(
...
   for (auto& node : graph_.Nodes()) {
     for (auto& entry : node.GetAttributeNameToMutableSubgraphMap()) {
-      const auto& ep = node.GetExecutionProviderType();
-      if (!ep.empty() &&
-          ep != kCpuExecutionProvider && ep != kCudaExecutionProvider &&
-          ep != kDmlExecutionProvider &&
-          ep != kJsExecutionProvider && ep != kWebGpuExecutionProvider) {
+      const auto& ep_type = node.GetExecutionProviderType();
+      const IExecutionProvider* ep = execution_providers_.Get(ep_type);
+      const bool is_plugin_ep = ep != nullptr && ep->GetOrtEp() != nullptr;
```
skottmckay, line 1188: IIRC this was added to avoid the expense of creating the SessionState for a subgraph if TRT was taking all of it. I assume TRT-RTX could do the same thing, so wondering how we differentiate between plugin EPs that do and don't need the session state. 

I believe this is called before we lookup kernels so that info isn't available yet, although it may be possible to change that and create the kernels before the subgraph session state. #Closed

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
adrianlizarraga: OrtKernelImpl hasn't been released in an official ORT release. ORT 1.24 will be the first version that has it so it should be ok?

I agree with changing it to a flag. I'll do that.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -655,6 +658,87 @@ typedef OrtStatus*(ORT_API_CALL* OrtKernelCreateFunc)(_In_ void* kernel_create_f
...
+  /** \brief Helper function that concatenates OrtValue instances from each loop iteration into a single
+   *         pre-allocated output buffer.
+   *
+   * \note Implementing this function is required for all Loop opset versions.
+   *
+   * \param[in] this_ptr The OrtLoopKernelHelper instance.
+   * \param[in] stream_handle Optional native stream handle that enables asynchronous operations. May be NULL.
+   * \param[in] per_iteration_output Array of OrtValue instances from each iteration. All OrtValue elements have the
+   *                                 same shape.
+   * \param[in] num_iteration_outputs The number of OrtValue* elements in the `per_iteration_output` array.
```
edgchen1: how about the names `num_per_iteration_outputs`?

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -655,6 +658,87 @@ typedef OrtStatus*(ORT_API_CALL* OrtKernelCreateFunc)(_In_ void* kernel_create_f
...
+  ORT_API_T(void, Release, _In_ OrtLoopKernelHelper* this_ptr);
+
+  /** \brief Helper function that concatenates OrtValue instances from each loop iteration into a single
+   *         pre-allocated output buffer.
+   *
+   * \note Implementing this function is required for all Loop opset versions.
+   *
+   * \param[in] this_ptr The OrtLoopKernelHelper instance.
+   * \param[in] stream_handle Optional native stream handle that enables asynchronous operations. May be NULL.
+   * \param[in] per_iteration_output Array of OrtValue instances from each iteration. All OrtValue elements have the
```
edgchen1: how about the name `per_iteration_outputs` (plural)?

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -655,6 +658,87 @@ typedef OrtStatus*(ORT_API_CALL* OrtKernelCreateFunc)(_In_ void* kernel_create_f
...
+   *                                 same shape.
+   * \param[in] num_iteration_outputs The number of OrtValue* elements in the `per_iteration_output` array.
+   * \param[out] output The pre-allocated output buffer. Memory is allocated on the device for the EP running the
+   *                    Loop node.
+   * \param[in] output_size_in_bytes The size in bytes of the `output` buffer.
```
edgchen1: maybe specify any guarantees for this size value. e.g., I'm assuming that it is large enough to hold the concatenated data of each per_iteration_output element.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -655,6 +658,87 @@ typedef OrtStatus*(ORT_API_CALL* OrtKernelCreateFunc)(_In_ void* kernel_create_f
...
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \since Version 1.24.
+   */
+  ORT_API2_STATUS(ConcatOutput, _In_ OrtLoopKernelHelper* this_ptr, _In_opt_ void* stream_handle,
+                  _In_ OrtValue* const* per_iteration_output, _In_ size_t num_iteration_outputs,
```
edgchen1: can it be a pointer to const OrtValue?
`_In_ OrtValue* const* per_iteration_output` -> `_In_ const OrtValue* const* per_iteration_output`

onnxruntime/core/session/plugin_ep/ep_control_flow_kernel_impls.cc
```diff
@@ -0,0 +1,136 @@
...
+
+PluginEpIfKernelImpl::PluginEpIfKernelImpl(const OpKernelInfo& info) : kernel_(info) {
+  Compute = ComputeImpl;
+  Release = ReleaseImpl;
+}
+
+/*static*/
+OrtStatus* ORT_API_CALL PluginEpIfKernelImpl::ComputeImpl(OrtKernelImpl* this_ptr,
+                                                          OrtKernelContext* kernel_ctx) noexcept {
+  auto plugin_ep_kernel = static_cast<PluginEpIfKernelImpl*>(this_ptr);
```
edgchen1: general: `auto*` for pointers
```suggestion
  auto* plugin_ep_kernel = static_cast<PluginEpIfKernelImpl*>(this_ptr);
```

onnxruntime/core/session/plugin_ep/ep_control_flow_kernel_impls.cc
```diff
@@ -0,0 +1,136 @@
...
+  ORT_API_RETURN_IF_STATUS_NOT_OK(plugin_ep_kernel->kernel_.Compute(reinterpret_cast<OpKernelContext*>(kernel_ctx)));
```
edgchen1, line 42: general: API entrypoints should handle and propagate errors from exceptions

onnxruntime/core/framework/session_state.cc
skottmckay: This is probably fine for now. If we find a plugin EP that doesn't want session state in the future we can think about how best to handle it at that point. May never happen.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -524,6 +524,7 @@ typedef struct OrtKernelImpl OrtKernelImpl;
  */
 struct OrtKernelImpl {
   uint32_t ort_version_supported;  ///< Must be initialized to ORT_API_VERSION
+  uint32_t flags;                  ///< EP must initialize to 0. Used internally by ORT.
```
skottmckay, line 528: optional: when an EP registers a kernel should we have a paranoid check that the value in flags in within the expected range (`flags` <= <latest flag>) to catch an EP not initializing the value to 0. not perfect, but might prevent unexpected behavior if that was the case.

adrianlizarraga: Added this check (and other basic checks that were missing):
https://github.com/microsoft/onnxruntime/blob/af265888e5f6fb821de119361f8063c50cf25c11/onnxruntime/core/session/plugin_ep/ep_kernel_registration.cc#L231-L237

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
adrianlizarraga: I believe so, but the internal function passes these as non-const:
https://github.com/microsoft/onnxruntime/blob/39d8520b9bfe16217d56be38293fc82556d67b5b/onnxruntime/core/providers/cpu/controlflow/loop.h#L43-L47

All existing implementations of this function seem to treat it as const, so I'll go ahead and do that. I don't see why we would ever want to modify these.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
adrianlizarraga: yep, changed

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
adrianlizarraga: yep, changed

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
adrianlizarraga: added a note about the guaranteed size.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
adrianlizarraga: Made these fully const.

onnxruntime/core/session/plugin_ep/ep_control_flow_kernel_impls.cc
adrianlizarraga: updated all relevant use to `auto*` in this file. I forget to do this often in my PRs, whoops :)

onnxruntime/core/session/plugin_ep/ep_control_flow_kernel_impls.cc
adrianlizarraga: Good catch. Added macros that catch exceptions and return them as statuses.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -683,20 +687,22 @@ struct OrtLoopKernelHelper {
...
-   *                                 same shape.
-   * \param[in] num_iteration_outputs The number of OrtValue* elements in the `per_iteration_output` array.
+   * \param[in] per_iteration_outputs Array of OrtValue instances from each iteration. All OrtValue elements have the
+   *                                  same shape.
+   * \param[in] num_per_iteration_outputs The number of OrtValue* elements in the `per_iteration_outputs` array.
    * \param[out] output The pre-allocated output buffer. Memory is allocated on the device for the EP running the
    *                    Loop node.
-   * \param[in] output_size_in_bytes The size in bytes of the `output` buffer.
+   * \param[in] output_size_in_bytes The size in bytes of the `output` buffer. It is guaranteed to be large enough
+   *                                 to hold the concatenated data of each element in `per_iteration_outputs`.
```
edgchen1, line 696: nit: I interpret "large enough" to allow for the buffer to be larger than necessary, but the annotation `_Out_writes_bytes_all_(output_size_in_bytes) void* output` means that the whole buffer should be written to

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -655,6 +662,89 @@ typedef OrtStatus*(ORT_API_CALL* OrtKernelCreateFunc)(_In_ void* kernel_create_f
...
+  /** \brief Called by ORT to release the OrtLoopKernelHelper instance and its resources.
+   *
+   * \param[in] this_ptr The OrtLoopKernelHelper instance.
+   *
+   * \since Version 1.24.
+   */
+  ORT_API_T(void, Release, _In_ OrtLoopKernelHelper* this_ptr);
+
+  /** \brief Helper function that concatenates OrtValue instances from each loop iteration into a single
+   *         pre-allocated output buffer.
```
edgchen1, line 684: just a general question - the helper functions seem to involve copying data around. naively, couldn't that be done by a data transfer implementation?

adrianlizarraga: Yeah, it can but the src and dst buffers are known to be on the same device, so maybe only a stream is potentially needed for synchronization. I suppose a data transfer implementation may be a better tool when the src and dst could be on different devices. In any case, there's a comment in the helper implementations that says that an actual implementation could use a data transfer impl for the copy if desired.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
adrianlizarraga: Ah yeah, I can see that interpretation. I'll adjust it in my next PR coming up if that's ok.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
edgchen1: I'm wondering if it's necessary for the EP implementers to provide these helpers or if ORT can just make use of a data transfer implementation to do the same thing.

PR #26929, opened 2026-01-07
Cuda Graph on by default

umangb-09:
### Description
Support for Cuda Graph on by default

### Motivation and Context
We wanted to have Cuda Graph support on by default for NV TRT-RTX EP. Added the use of RTX graph capture and removed the access external checks for the same


onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider_info.h
```diff
@@ -46,7 +46,7 @@ struct NvExecutionProviderInfo {
   std::string profile_min_shapes{""};
   std::string profile_max_shapes{""};
   std::string profile_opt_shapes{""};
-  bool cuda_graph_enable{false};
+  bool cuda_graph_enable{true};
```
ishwar-raut1, line 49: There are 2 cases 

- if EP is compiled with older TRT RTX, in this case, do we need to maintain the older behaviour? In this case, if the user enables the cuda graph from the provider options. He will not see the benefit. We could show the warning to use the latest trt rtx. 
- With latest TRT RTX , we will not see any issue.

umangb-09: ACK. Done

umangb-09:
@chilo-ms @jywu-msft check this for merging

ishwar-raut1 (approved):

chilo-ms:
Please run `lintrunner -a` to fix the format

chilo-ms (approved):

PR #26937, opened 2026-01-08
[CUDA] Update Flash Attention Implementation and APIs

tianleiwu:
## Summary
This PR updates the Flash Attention implementation in ONNX Runtime, syncing with newer kernel sources in https://github.com/Dao-AILab/flash-attention, and extending the internal API to support additional features required for advanced caching scenarios. It also aligns specific kernels with the official implementation.

## Changes
- **Flash Attention Kernels**: Updated/Added Flash Attention forward kernels and headers in `onnxruntime/contrib_ops/cuda/bert/flash_attention/`.
- **API Extension**: Updated `mha_fwd` and `mha_fwd_kvcache` in `flash_api.h` and `flash_api.cc` to accept two new optional parameters:
    - `cache_batch_idx`: Indices to index into the KV cache (support for non-contiguous batch indices).
    - `leftpad_k`: Support for left-padding in the key sequence.
- **Alignment & Fixes**:
    - **Cleanup**: Removed redundant `kInfinity` definition in `flash_fwd_kernel.h`.
    - **Includes**: Added missing `<core/providers/cuda/shared_inc/cuda_call.h>` in `flash_fwd_launch_template.h`.
- **Integration**: Updated `group_query_attention_impl.cu` to align with the new `mha_fwd_kvcache` signature.
- **Build Configuration**: Adjusted `onnxruntime_providers_cpu.cmake` to update the exclusion list for Flash Attention kernels in quick build mode.

## Implementation Details
- The `run_mha_fwd` helper now checks if `cache_batch_idx` is provided alongside `k_new` to determine if the split kernel should be forced.
- New parameters are propagated through the call stack to the underlying Flash Attention kernels.


onnxruntime/contrib_ops/cuda/bert/flash_attention/update_kernels.py
```diff
@@ -0,0 +1,77 @@
...
+
+template_split = (
+    copyright_header
+    + """
+#include "contrib_ops/cuda/bert/flash_attention/namespace_config.h"
+#include "contrib_ops/cuda/bert/flash_attention/flash_fwd_launch_template.h"
+
+namespace FLASH_NAMESPACE {{
+
+template void run_mha_fwd_splitkv_dispatch<{cpp_type}, {dim}, {is_causal}>(Flash_fwd_params &params, cudaStream_t stream);
```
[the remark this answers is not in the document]
tianleiwu: It does not matter since the generated code will be formatted by "lintrunner"

onnxruntime/contrib_ops/cuda/bert/flash_attention/softmax.h
```diff
@@ -76,10 +77,17 @@ __forceinline__ __device__ void scale_apply_exp2(Tensor<Engine0, Layout0>& tenso
...
-      // max * log_2(e)) This allows the compiler to use the ffma
-      // instruction instead of fadd and fmul separately.
+// Instead of computing exp(x - max), we compute exp2(x * log_2(e) -
+// max * log_2(e)) This allows the compiler to use the ffma
+// instruction instead of fadd and fmul separately.
+// The following macro will disable the use of fma.
+// See: https://github.com/pytorch/pytorch/issues/121558 for more details
+// This macro is set in PyTorch and not FlashAttention
+#ifdef UNFUSE_FMA
+      tensor(mi, ni) = exp2f(__fmul_rn(tensor(mi, ni), scale) - max_scaled);
```
kunal-vaishnavi, line 87: Will this help reduce NaNs in the GQA op?

onnxruntime/contrib_ops/cuda/bert/flash_attention/softmax.h
```diff
@@ -177,16 +183,15 @@ struct Softmax {
                     ? (Split ? -kInfinity : kInfinity)
                     : max_unscaled * softmax_scale + __logf(sum);
 
-      float inv_sum = (sum == 0.f || !isfinite(sum)) ? 1.f : 1.f / sum;
+      float inv_sum = (sum == 0.f || sum != sum) ? 1.f : 1.f / sum;
```
kunal-vaishnavi, line 186: Why check `sum != sum` instead of `!isfinite(sum)`? If [`-ffast-math`](https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html) is used during compilation, this is reduced to `false`.

onnxruntime/contrib_ops/cuda/bert/flash_attention/update_kernels.py
```diff
@@ -0,0 +1,77 @@
...
+base_dir = os.path.dirname(os.path.realpath(__file__))
+dims = [32, 64, 96, 128, 192, 256]
+types = ["fp16", "bf16"]
+
+copyright_header = """/******************************************************************************
+ * Copyright (c) 2024, Tri Dao.
+ ******************************************************************************/
+"""
+
+template_standard = (
```
kunal-vaishnavi, line 12: Should we consider using this approach to auto-generate the `.cu` files as well during compilation?

tianleiwu: To auto-generate, we need a place (like build.py) to call python to generate the cu files.

Current approach is more straightforward, and I do not want to add more complexity to build.

onnxruntime/contrib_ops/cuda/bert/flash_attention/softmax.h
tianleiwu: sum != sum is false for inf and true for NaN, so inf will get inv_sum=0, nan will get inv_sum=1.0.

The old code will get inv_sum=1 for both inf and nan.

So new code looks better on dealing with inf (although chance of inf shall be tiny)

1/inf = 0 from pure math point of view. 

For example, I have a row of (max_float, 2). For ideal normalization, the normalized value shall be close to (1.0, 0.0). 
The sum is inf since max_float+2 is out of limit of float.
For new code, scale (inv_sum)=0.0 is used, the normalized value is (max_float, 2) * 0.0 = (0.0, 0.0)
For old code, scale=1.0 is used, the normalized value is (max_float, 2) * 1.0 = (max_float, 2)

Although both are not ideal. The old code might cause more downstream error (like NaN) since it has not normalized values, so future computation might out of float limit. The new code could avoid that.

onnxruntime/contrib_ops/cuda/bert/flash_attention/softmax.h
tianleiwu: That need the max value to be very large to cause NaN. For example, the pytorch issue uses 133120 to reproduce. For LLM, the value likely be much smaller (max fp16 is 65504). 

Since official flash attention did not enable this, I think the chance is rare.

If we do find such case in real model, we can add a build flag to enable this.

PR #26938, opened 2026-01-08
fix lint errors in js pipeline

prathikr:
- suppresses errors related to uncessary type assertions only where it is creating noise for the pipelines
- resolves `[baseline-browser-mapping] The data in this module is over two months old.  To ensure accurate Baseline data, please update: npm i baseline-browser-mapping@latest -D` by running the recommended update

### DIFF
--- a/js/web/lib/wasm/jsep/backend-webgpu.ts
+++ b/js/web/lib/wasm/jsep/backend-webgpu.ts
@@ -247,7 +247,7 @@ export class WebGpuBackend {
     const requireFeatureIfAvailable = (feature: GPUFeatureName) =>
       adapter.features.has(feature) && requiredFeatures.push(feature) && true;
     // Try chromium-experimental-timestamp-query-inside-passes and fallback to timestamp-query
-    if (!requireFeatureIfAvailable('chromium-experimental-timestamp-query-inside-passes' as GPUFeatureName)) {
+    if (!requireFeatureIfAvailable('chromium-experimental-timestamp-query-inside-passes')) {
       requireFeatureIfAvailable('timestamp-query');
     }
     requireFeatureIfAvailable('shader-f16');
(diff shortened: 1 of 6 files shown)


js/eslint.config.mjs
```diff
@@ -139,7 +139,9 @@ export default [
         },
       ],
 
-      '@typescript-eslint/no-unnecessary-type-assertion': 'error',
+      // extra type assertions are sometimes needed for complex scenarios
+      // https://eslint.org/docs/latest/use/configure/rules#rule-severities
+      '@typescript-eslint/no-unnecessary-type-assertion': 'warn',
```
edgchen1: should we just suppress the rule locally at the place(s) where it is deemed ok instead of changing it to a warning in general?

fs-eire:
I think this is incorrectly setup. without any of the change in this PR, the `npm run lint` under folder `/js/` does not report any problem. (this is performed in the web CI pipeline)

I think the lint CI pipeline does not respect the typescript project file so that it generates the wrong result.

prathikr:
> I think this is incorrectly setup. without any of the change in this PR, the `npm run lint` under folder `/js/` does not report any problem. (this is performed in the web CI pipeline)
> 
> I think the lint CI pipeline does not respect the typescript project file so that it generates the wrong result.

I can repro the issue locally...

image

fs-eire:
do you still see the error when doing this:

- run `npm ci` in `/js/`
- run `npm ci` in `/js/common/`
- run `npm ci` in `/js/web/`
- run `npm run lint` in `/js/`

prathikr:
> do you still see the error when doing this:
> 
> * run `npm ci` in `/js/`
> * run `npm ci` in `/js/common/`
> * run `npm ci` in `/js/web/`
> * run `npm run lint` in `/js/`

No, now there is no error. Is this what should be done in the pipeline?

fs-eire:
I think we can probably ignore the /js/ folder in the Lint, since it is already covered in the Web CI pipeline

prathikr:
Looks like the lint javascript job only operates on the `/js/` folder, should we just remove this stage entirely? [link](https://github.com/microsoft/onnxruntime/blob/ce4a7c58cecac3627e030ca5d38598f4777084e0/.github/workflows/lint.yml#L119)

prathikr:
Continued in https://github.com/microsoft/onnxruntime/pull/26984

PR #26948, opened 2026-01-08
[TRT RTX EP] Add support for D3D12 external resource import

gedoensmax:
This PR adds a test for CiG inference to demonstrate how usage for it should look like. 
It is important to not call `cudaSetDevice` in that flow since it will create a new context. @nieubank I am not sure why there was a `cudaSetDevice` on each import call 🤔 Is this done to enable importing semaphores of e.g. GPU:1 to a session running on GPU:0 ? Context management is unreliable with the current CUDA runtime and CUDA driver API mixing.

### DIFF
--- a/cmake/onnxruntime_providers_nv.cmake
+++ b/cmake/onnxruntime_providers_nv.cmake
@@ -2,7 +2,7 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # Licensed under the MIT License.
   if(onnxruntime_CUDA_HOME)
-    file(TO_CMAKE_PATH CUDAToolkit_ROOT ${onnxruntime_CUDA_HOME})
+    file(TO_CMAKE_PATH ${onnxruntime_CUDA_HOME} CUDAToolkit_ROOT)
   endif()
   find_package(CUDAToolkit REQUIRED)
   enable_language(CUDA)
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -1441,7 +1441,7 @@ get_property(onnxruntime_GENERATOR_IS_MULTI_CONFIG GLOBAL PROPERTY GENERATOR_IS_
 if (onnxruntime_USE_CUDA)
   set(CMAKE_CUDA_STANDARD 17)
   if(onnxruntime_CUDA_HOME)
-    file(TO_CMAKE_PATH CUDAToolkit_ROOT ${onnxruntime_CUDA_HOME})
+    file(TO_CMAKE_PATH ${onnxruntime_CUDA_HOME} CUDAToolkit_ROOT)
   endif()
   find_package(CUDAToolkit REQUIRED)
 


nieubank:
> This PR adds a test for CiG inference to demonstrate how usage for it should look like. It is important to not call `cudaSetDevice` in that flow since it will create a new context. @nieubank I am not sure why there was a `cudaSetDevice` on each import call 🤔 Is this done to enable importing semaphores of e.g. GPU:1 to a session running on GPU:0 ? Context management is unreliable with the current CUDA runtime and CUDA driver API mixing.

Awesome, thanks for this! You're seeing my inexperience with the CUDA API here :), I have another branch I was working on to fix some of the context stuff, but I figure this implementation will be a longer-term collaboration/hand-off at some point. Just wanted to validate the API with some real code.

gedoensmax:
Yes sure, we (or in other words @praneshgo ) will probably take it over. I made these changes for the exact same reason to experiment with it. And i already identified a TRT RTX optimization opportunity that we will fix internally. 

To better test the correct async behaviour by the way it might be better to submit multiple inferences and wait on the last result to ensure that we are not synchronous due to CPU overhead.

praneshgo:
Functionally, the change looks good to me.

praneshgo:
@gaugarg-nv @ankan-ban can you please review this as well? Thanks.

ankan-ban:
Thanks, Max, for writing the test. It looks good. It's nice to see most of the interop functionality nicely abstracted out behind the new ORT interop APIs. 

There are just a couple of things that are still Nvidia specific in the test (maybe consider abstracting out these too in the future with more additions to ORT APIs):

1. Use of CUDA APIs by the app to create the context before invoking ORT (this requires a command queue handle that would share same TSG with the cuda context. The same TSG is a requirement to enable CiG on our hardware).

2. Use of nv-specific session options (kUserComputeStream, kHasUserComputeStream, kMaxSharedMemSize). Maybe having a mechanism for passing the generic ORT-stream object to session.run() makes even more sense. The shared memory size limit is again required for running in CiG mode - and hopefully if we have a generic way of doing "1" above - it can allow the EP to automatically set the correct value depending on the GPU.

I think resolving the above 2 would allow app developers to write truly IHV agonistic code that runs everywhere (e.g, using DX12 APIs for allocating resources, doing any pre/post processing and generic ORT APIs to run the model).

gedoensmax:
Thanks @ankan-ban for the review. 

1. DO you consider this as a big blocker ? I though of this small driver API usage as OK for an ISV to integrate, let me know if you think different. 
2. Fully agree ! I would love to set the kMaxSharedMemSize implicit but i did not found a way to check for the max supported shared mem size based on the currently pushed context. Having a stream as input on `Ort::Session::Run` would be great and @skottmckay has mentioned that on another thread, can you let us know what the status on this is ? 

@nieubank can you help tag this for 1.24 since we would like to make sure that this goes in with the newly added Interop API. I can take care of rebasing to main and accepting some of the copilot comments if that's all.

skottmckay:
#26988 added the stream in RunOptions and is now merged.

ankan-ban:
_>DO you consider this as a big blocker ? I though of this small driver API usage as OK for an ISV to integrate, let me know if you think different_
Agree that it's not a big blocker - but after https://github.com/microsoft/onnxruntime/pull/26988 it seems the only NV specific thing that the app needs to do. Agree that it's likely not too much for ISVs to integrate.

gedoensmax:
I rebased on the refined structs and started providing the stream as run option. There are missing changes to support this for CiG but we are tracking this internally.

onnxruntime/test/providers/nv_tensorrt_rtx/nv_external_resource_importer_test.cc
```diff
@@ -0,0 +1,1190 @@
...
+  Ort::AllocatedStringPtr input_name = session.GetInputNameAllocated(0, allocator);
+  Ort::AllocatedStringPtr output_name = session.GetOutputNameAllocated(0, allocator);
+
+  io_binding.BindInput(input_name.get(), Ort::Value(input_tensor));
+  io_binding.BindOutput(output_name.get(), Ort::Value(output_tensor));
+  io_binding.SynchronizeInputs();
+  // Run inference. ORT submits all work to the stream before returning, so we signal the async semaphore below.
+  Ort::RunOptions run_options;
+  run_options.SetSyncStream(ort_stream);
+  run_options.AddConfigEntry(kOrtRunOptionsConfigDisableSynchronizeExecutionProviders, "1");
```
skottmckay, line 837: Is this setting required? Not sure that everything is fully thought through if someone sets this flag.

e.g. if set, we do not call `Stream::CleanupOnRunEnd` [here](https://github.com/microsoft/onnxruntime/blob/1817f4a81072c49dea2984853edca412674d66fb/onnxruntime/core/framework/device_stream_collection.cc#L53-L67), and that (afaics) is the only place `CudaStream` frees `deferred_cpu_buffers_`. `~CudaStream` ignores them and we're storing them a `void*` not a unique_ptr that could auto-delete so potentially that's a leak.

We also don't call `Stream::Flush` if set which has a `cudaStreamSynchronize` in `CudaStream::Flush`

gedoensmax: It is required to get all the benefits of DX interop. We are synchronizing with a semaphore to not wait on CPU, but have an async synchronization between CUDA and DX on the GPU. 

If this setting can not be used there's not really a point in using semaphores.

skottmckay: This is all in the CudaStream implementation so you control it. Was mainly pointing out that as it currently stands I think that implementation has issues like a memory leak of deferred_cpu_buffers_. 

Are there scenarios where the inference output is doing to be externally consumed on GPU and a cudaStreamSynchronize needs to be avoided?

gedoensmax: Yes for sure. Any stable diffusion inference or video processing pipeline should always be synchronized externally. We wrote a devblog on this a few years back: https://developer.nvidia.com/blog/end-to-end-ai-for-nvidia-based-pcs-cuda-and-tensorrt-execution-providers-in-onnx-runtime/#sample_application

onnxruntime/test/providers/nv_tensorrt_rtx/nv_external_resource_importer_test.cc
```diff
@@ -0,0 +1,1190 @@
...
+// 2. Memory import capability check (D3D12 Resource & Heap)
+// 3. Semaphore import capability check (D3D12 Fence)
+// 4. D3D12 shared resource import to CUDA
+// 5. Tensor creation from imported external memory
+// 6. D3D12 timeline fence import for GPU synchronization
+// 7. Wait/Signal semaphore operations
+// 8. Full inference pipeline with zero-copy external memory
+// 9. Full inference pipeline with zero-copy external memory and a CIG context
+
+#include "../../../../include/onnxruntime/core/session/onnxruntime_cxx_api.h"
```
yuslepukhin: >include "../../../../include/onnxruntime/core/session/onnxruntime_cxx_api.h" [](http://example.com/codeflow?start=1&length=76)

Appears to be a duplicate include

onnxruntime/test/providers/nv_tensorrt_rtx/nv_external_resource_importer_test.cc
```diff
@@ -0,0 +1,1190 @@
...
+// Include CUDA headers for pointer attribute verification
+#include <cuda.h>
+#include <cuda_runtime.h>
+#include <core/providers/nv_tensorrt_rtx/nv_provider_options.h>
+
+using namespace std;
+using namespace ONNX_NAMESPACE;
+using namespace ::onnxruntime::logging;
+
+extern std::unique_ptr<Ort::Env> ort_env;
```
yuslepukhin, line 50: This can be obtained via provider_api.h

onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc
```diff
@@ -516,6 +523,465 @@ struct NvTrtRtxSyncStreamImpl : OrtSyncStreamImpl {
...
+  }
+
+  static void ORT_API_CALL ReleaseCallback(_In_ OrtExternalMemoryHandle* handle) noexcept {
+    if (handle == nullptr) return;
+    auto* derived = static_cast<NvTrtRtxExternalMemoryHandle*>(handle);
+    // Destroy the external memory object (also releases mapped buffer)
+    if (derived->ext_memory != nullptr) {
+      cuDestroyExternalMemory(derived->ext_memory);
+    }
+    delete derived;
```
yuslepukhin: >delete derived; [](http://example.com/codeflow?start=4&length=15)

unique_ptr? #Resolved

onnxruntime/test/providers/nv_tensorrt_rtx/nv_external_resource_importer_test.cc
gedoensmax: Accidentally added by my IDE.

onnxruntime/test/providers/nv_tensorrt_rtx/nv_external_resource_importer_test.cc
gedoensmax: Why would i prefer the provider API over the normal `Ort::Env()` ?

onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc
gedoensmax: changed in a few places.

onnxruntime/test/providers/nv_tensorrt_rtx/nv_external_resource_importer_test.cc
yuslepukhin: Sorry, I forgot, it is an interface.

yuslepukhin:
@gedoensmax Please, review, address, comment on, resolve Copilot comments. They are often useful.

onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc
```diff
@@ -516,6 +523,465 @@ struct NvTrtRtxSyncStreamImpl : OrtSyncStreamImpl {
...
+    // This method is called from OrtExternalResourceImporterImpl::ReleaseMemory
+    // The Release callback in the handle will call the static ReleaseCallback
+    auto* mem_handle = static_cast<NvTrtRtxExternalMemoryHandle*>(handle);
+
+    // Destroy the external memory object (also releases mapped buffer)
+    if (mem_handle->ext_memory != nullptr) {
+      cuDestroyExternalMemory(mem_handle->ext_memory);
+    }
+
+    delete mem_handle;
```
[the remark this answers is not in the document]
gedoensmax: I would stick to the current handling - if i understand correctly a change would mean that we need to delete inside the API implementation in `core/session/interop_api.cc`. If you want me to change this just let me know.

onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider_utils.h
```diff
@@ -697,4 +700,26 @@ static bool checkTrtTensorIsDynamic(nvinfer1::ITensor* tensor) {
...
+  explicit ScopedContext(int device_id) {
+    CUcontext cu_context = 0;
+    CU_CALL_THROW(cuCtxGetCurrent(&cu_context));
+    if (!cu_context) {
+      // cuCtxGetCurrent succeeded but returned nullptr, which indicates that no CUDA context
+      // is currently set for this thread. This implicates that there is not user created context.
+      // We use runtime API to initialize a context for the specified device.
+      CUDA_CALL_THROW(cudaSetDevice(device_id));
+      CU_CALL_THROW(cuCtxGetCurrent(&cu_context));
+    }
```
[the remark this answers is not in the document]
gedoensmax: I am aware of this but we will handle this in another PR. We have other `cudaSetDevice` calls in the EP that we need to remove as well.

onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc
```diff
@@ -516,6 +523,465 @@ struct NvTrtRtxSyncStreamImpl : OrtSyncStreamImpl {
...
+    // No deleter is needed since this is for inference inputs/outputs where the user controls lifetime.
+    OrtStatus* status = impl.ort_api.CreateTensorWithDataAsOrtValue(
+        memory_info,
+        data_ptr,
+        handle->descriptor.size_bytes - tensor_desc->offset_bytes,
+        tensor_desc->shape,
+        tensor_desc->rank,
+        tensor_desc->element_type,
+        out_tensor);
+
```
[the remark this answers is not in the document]
gedoensmax: Changed according to the comment.

gedoensmax:
@yuslepukhin all the comments i resolved i already implemented from copilot. I missed the one on raw tensor size handling and made the according change.

yuslepukhin:
It looks good. You will need to resolve conflicts. It will also require to mark all the comments as resolved.

cmake/onnxruntime_providers_nv.cmake
```diff
@@ -1,7 +1,10 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # Licensed under the MIT License.
-  find_package(CUDAToolkit REQUIRED 12.8)
+  if(onnxruntime_CUDA_HOME)
+    file(TO_CMAKE_PATH CUDAToolkit_ROOT ${onnxruntime_CUDA_HOME})
```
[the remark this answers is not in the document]
gedoensmax: Copied from the main CMakeLists, but the AI is right.

gedoensmax:
@nieubank @yuslepukhin Is anything else missing here ?

yuslepukhin (approved):
:shipit:

PR #26952, opened 2026-01-09
[QNN-EP] Implement file mapped weights feature

quic-calvnguy:
Description
Enables the file mapping of weights as well as the overall context bin. This feature is currently only enabled for ARM64 WIN devices

Motivation and Context
Currently, when reading the context bin, ORT allocates a large buffer on the heap. Assuming the same model is used, each ORT session will allocate a buffer for the context bin. This is incredibly wasteful when large models are used. Instead, WIN file mapping can be leveraged to map the context bin, then every time a context needs to be created with the context bin, the pointer to the context bin can be retrieved and used instead of some pre-allocated buffer, thus making QNN EP more memory-efficient. In the case of multiple ORT sessions, the context bin will only be loaded once for all sessions, increasing memory efficiency and overall initialization performance. This is very useful regarding the use of LLMs going forward.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
@@ -814,7 +814,6 @@ Qnn_ErrorHandle_t QnnBackendManager::MapDmaData(Qnn_ContextBinaryDataRequest_t r
                                                 Qnn_ContextBinaryDmaDataResponse_t* response,
                                                 void* const mapped_base_ptr,
                                                 const size_t file_size) {
-  std::lock_guard<std::mutex> lock(file_mapping_mutex_);
   if (!file_mapped_weights_enabled_) {
     LOGS(*logger_, WARNING) << "Attempting to map DMA data but file mapping has been disabled, "
                             << "possibly due to an error in a previous request.";
@@ -976,21 +975,17 @@ Status QnnBackendManager::GetFileSizeIfValid(const std::string& filepath,
 }
 
 Status QnnBackendManager::ReadContextBinIfValid(const std::string& context_bin_filepath,
-                                                BufferInfo_t& buffer_info) {
+                                                std::vector<char>& buffer) {
   size_t buffer_size;
   ORT_RETURN_IF_ERROR(GetFileSizeIfValid(context_bin_filepath, buffer_size));
 
-  std::vector<char> buffer;
-  buffer.reserve(buffer_size);
+  buffer.resize(buffer_size);
 
   std::ifstream cache_file(context_bin_filepath.c_str(), std::ifstream::binary);
   ORT_RETURN_IF(!cache_file || !cache_file.good(), "Failed to read context binary from: ", context_bin_filepath);
 
   const auto& read_result = cache_file.read(buffer.data(), buffer_size);
   ORT_RETURN_IF(!read_result, "Failed to read contents from cached context file.");
-
-  buffer_info.data = std::move(buffer);
-  buffer_info.size = buffer_size;
 
   return Status::OK();
 }
@@ -1032,14 +1027,11 @@ Status QnnBackendManager::CreateContextVtcmBackupBufferSharingEnabled(std::unord
                                           nullptr};
 
 #ifdef QNN_FILE_MAPPED_WEIGHTS_AVAILABLE
-  if (FileMappingIsEnabled() && file_mapper_) {
+  if (file_mapped_weights_enabled_ && file_mapper_) {
     // Retry logic -- if context creation failed with file mapped weights, then retry with feature disabled
     auto res = CreateContextFromListAsyncWithCallback(configs, context_bin_map);
     if (!res.IsOK()) {
-      DisableFileMapping();
       LOGS(*logger_, WARNING) << res.ErrorMessage() << ". Retrying with feature disabled.";
-
-      file_mapped_weights_enabled_ = false;
     } else {
       return Status::OK();
     }
@@ -1062,11 +1054,11 @@ Status QnnBackendManager::CreateContextFromListAsync(const QnnContext_Config_t**
   for (auto& it : context_bin_map) {
     auto context_bin_filepath = it.first;
 
-    BufferInfo_t buffer_info;
-    ORT_RETURN_IF_ERROR(ReadContextBinIfValid(context_bin_filepath, buffer_info));
+    std::vector<char> buffer;
+    ORT_RETURN_IF_ERROR(ReadContextBinIfValid(context_bin_filepath, buffer));
 
-    buffer_list.push_back(std::move(buffer_info.data));
-    size_t buffer_size = buffer_info.size;
+    size_t buffer_size = buffer.size();
+    buffer_list.push_back(std::move(buffer));
 
     QnnContext_ParamsV1_t context_params_v1 = {nullptr,
                                                buffer_list.back().data(),
@@ -1367,7 +1359,7 @@ Status QnnBackendManager::LoadCachedQnnContextFromBuffer(char* buffer, uint64_t 
 
   void* bin_buffer = nullptr;
 #ifdef QNN_FILE_MAPPED_WEIGHTS_AVAILABLE
-  if (FileMappingIsEnabled()) {
+  if (file_mapped_weights_enabled_) {
     ORT_RETURN_IF(!file_mapper_, "Attemping to use File Mapping feature but file_mapper_ is uninitialized");
 
     ORT_RETURN_IF_ERROR(GetFileSizeIfValid(context_bin_filepath, buffer_length));
@@ -1467,7 +1459,7 @@ Status QnnBackendManager::LoadCachedQnnContextFromBuffer(char* buffer, uint64_t 
 
 #ifdef QNN_FILE_MAPPED_WEIGHTS_AVAILABLE
     Qnn_ContextBinaryCallback_t callbacks;
-    if (FileMappingIsEnabled() && file_mapper_) {
+    if (file_mapped_weights_enabled_ && file_mapper_) {
       ORT_RETURN_IF(nullptr == qnn_interface_.contextCreateFromBinaryWithCallback,
                     "Invalid function pointer for contextCreateFromBinaryWithCallback.");
 
@@ -1494,7 +1486,7 @@ Status QnnBackendManager::LoadCachedQnnContextFromBuffer(char* buffer, uint64_t 
 
 #ifdef QNN_FILE_MAPPED_WEIGHTS_AVAILABLE
     std::vector<char> backup_buffer;
-    if (FileMappingIsEnabled() && file_mapper_) {
+    if (file_mapped_weights_enabled_ && file_mapper_) {
       rt = qnn_interface_.contextCreateFromBinaryWithCallback(backend_handle_,
                                                               device_handle_,
                                                               context_configs,
@@ -1506,22 +1498,19 @@ Status QnnBackendManager::LoadCachedQnnContextFromBuffer(char* buffer, uint64_t 
                                                               NULL);
 
       if (rt != QNN_SUCCESS) {
-        DisableFileMapping();
         LOGS(*logger_, WARNING) << "Failed to create context with file mapping enabled. Error: "
                                 << QnnErrorHandleToString(rt) << ", Code : " << rt
                                 << ". Retrying with feature disabled.";
 
         // Read context bin from file since file mapping has failed
-        BufferInfo_t buffer_info;
-        ORT_RETURN_IF_ERROR(ReadContextBinIfValid(context_bin_filepath, buffer_info));
-        backup_buffer = std::move(buffer_info.data);
+        ORT_RETURN_IF_ERROR(ReadContextBinIfValid(context_bin_filepath, backup_buffer));
 
         bin_buffer = static_cast<void*>(backup_buffer.data());
       }
     }
 #endif  // QNN_FILE_MAPPED_WEIGHTS_AVAILABLE
 
-    if (!FileMappingIsEnabled() || rt != QNN_SUCCESS) {
+    if (!file_mapped_weights_enabled_ || rt != QNN_SUCCESS) {
       rt = qnn_interface_.contextCreateFromBinary(backend_handle_,
                                                   device_handle_,
                                                   context_configs,
--- a/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
+++ b/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
@@ -257,7 +257,6 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
 #endif
 
   bool FileMappingIsEnabled() {
-    std::lock_guard<std::mutex> lock(file_mapping_mutex_);
     return file_mapped_weights_enabled_;
   }
 
@@ -286,11 +285,6 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
 #endif
 
  private:
-  typedef struct BufferInfo {
-    std::vector<char> data;
-    size_t size;
-  } BufferInfo_t;
-
   Status LoadBackend();
 
   Status InitializeBackend();
@@ -310,7 +304,7 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
   Status GetFileSizeIfValid(const std::string& filepath, size_t& file_size);
 
   Status ReadContextBinIfValid(const std::string& context_bin_filepath,
-                               BufferInfo_t& buffer_info);
+                               std::vector<char>& buffer);
 
   Status CreateContextVtcmBackupBufferSharingEnabled(std::unordered_map<std::string,
                                                                         std::unique_ptr<std::vector<std::string>>>& context_bin_map);
@@ -467,11 +461,6 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
     return Status::OK();
   }
 
-  void DisableFileMapping() {
-    std::lock_guard<std::mutex> lock(file_mapping_mutex_);
-    file_mapped_weights_enabled_ = false;
-  }
-
  private:
   const std::string backend_path_;
   std::recursive_mutex logger_recursive_mutex_;
@@ -514,8 +503,6 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
   bool context_created_ = false;
   bool backend_setup_completed_ = false;
   bool vtcm_backup_buffer_sharing_enabled_ = false;
-
-  std::mutex file_mapping_mutex_;
   bool file_mapped_weights_enabled_ = false;
 
 #ifdef QNN_FILE_MAPPED_WEIGHTS_AVAILABLE


yuslepukhin:
The observation is not entirely true. ORT memory map external weights. You have an ability to request a weight as an ORT Value from the EP. If the weight is external it will be memory mapped.

See `Graph::LoadExternalInitializerAsOrtValue`. One can add this to the provider_wrapped_types.h to `Graph` and expose to DLL based EPs. Then you will get mapping for free.

yuslepukhin:
I suggest not to implement QNN specific mapping, but re-use code in ORT.

yuslepukhin:
Discussed offline, the EP maps initializers form the binary context, not from the external weights files.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -475,6 +475,21 @@ QNNExecutionProvider::QNNExecutionProvider(const ProviderOptions& provider_optio
 #endif
   }
 
+  static const std::string DISABLE_FILE_MAPPED_WEIGHTS = "disable_file_mapped_weights";
```
yuslepukhin, line 478: > static const std::string DISABLE_FILE_MAPPED_WEIGHTS  [](http://example.com/codeflow?start=1&length=54)

Is there a separate place where all constants are collected together? #Resolved

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -475,6 +475,21 @@ QNNExecutionProvider::QNNExecutionProvider(const ProviderOptions& provider_optio
...
+  auto disable_file_mapped_weights_pos = provider_options_map.find(DISABLE_FILE_MAPPED_WEIGHTS);
+  if (disable_file_mapped_weights_pos != provider_options_map.end()) {
+    if ("1" == disable_file_mapped_weights_pos->second) {
+      enable_file_mapped_weights_ = false;
+    }
+    LOGS_DEFAULT(VERBOSE) << "User specified disable_file_mapped_weights: " << enable_file_mapped_weights_;
```
yuslepukhin, line 484: >LOGS_DEFAULT(VERBOSE) [](http://example.com/codeflow?start=4&length=21)

I believe we pass a logger to use to every EP so we can see which session emits which message. Pls, check if it is available. #Resolved

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -770,22 +774,54 @@ Status SetQnnContextConfig(ContextPriority context_priority, QnnContext_Config_t
...
+// callback required for allocating file mapping resources
+static Qnn_ErrorHandle_t DmaDataProvider(Qnn_ContextBinaryDataRequest_t request,
+                                         Qnn_ContextBinaryDmaDataResponse_t* response, void* notify_param) {
+  if (notify_param == nullptr) {
+    LOGS_DEFAULT(ERROR) << "DmaProvider: notify_param is null";
+    return QNN_CONTEXT_ERROR_INVALID_ARGUMENT;
+  }
+  auto pair = reinterpret_cast<std::pair<FileMappingCallbackInterface*, void*>*>(notify_param);
+
+  if (pair->first == nullptr) {
```
yuslepukhin: >if (pair->first == nullptr) { [](http://example.com/codeflow?start=2&length=29)

Do we want some validation for the second? #Resolved

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -809,6 +845,32 @@ void QnnBackendManager::ProcessContextFromBinListAsync(Qnn_ContextHandle_t conte
...
 }
 
+Status QnnBackendManager::ReadContextBinIfValid(const std::string& context_bin_filepath,
+                                                BufferInfo_t& buffer_info,
+                                                bool read_file_contents) {
+  std::ifstream cache_file(context_bin_filepath.c_str(), std::ifstream::binary);
+  ORT_RETURN_IF(!cache_file || !cache_file.good(), "Failed to retrieve context binary from: ", context_bin_filepath);
+
+  cache_file.seekg(0, cache_file.end);
+  size_t buffer_size = static_cast<size_t>(cache_file.tellg());
```
yuslepukhin: >static_cast<size_t> [](http://example.com/codeflow?start=23&length=19)

Would this require a safe cast? It may fail on 32-bit systems if that matters. #Resolved

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -901,6 +975,88 @@ Status QnnBackendManager::CreateContextVtcmBackupBufferSharingEnabled(std::unord
...
+  file_mapping_notify_param_list.reserve(context_bin_map.size());
+
+  for (auto& it : context_bin_map) {
+    auto context_bin_filepath = it.first;
+
+    BufferInfo_t buffer_info;
+    ORT_RETURN_IF_ERROR(ReadContextBinIfValid(context_bin_filepath, buffer_info, false));
+
+    size_t buffer_size = buffer_info.size;
+    ORT_RETURN_IF(buffer_size == 0, "Context bin has a size of 0 bytes: ", context_bin_filepath);
```
yuslepukhin: >ORT_RETURN_IF(buffer [](http://example.com/codeflow?start=4&length=20)

The size has already been validated inside the function and the error code was checked. #Resolved

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -901,6 +975,88 @@ Status QnnBackendManager::CreateContextVtcmBackupBufferSharingEnabled(std::unord
...
+  auto result = qnn_interface_.contextCreateFromBinaryListAsync(backend_handle_,
+                                                                device_handle_,
+                                                                context_params_ptr_list.data(),
+                                                                configs,
+                                                                nullptr);
+  file_mapping_notify_param_list.clear();
+  context_params_ptr_list.clear();
+  context_callbacks_list.clear();
+  context_paramsv2_list.clear();
+  context_params_list.clear();
```
yuslepukhin: Is this necessary? If it is, then the code is not exception safe. #Resolved

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1106,14 +1263,36 @@ Status QnnBackendManager::LoadCachedQnnContextFromBuffer(char* buffer, uint64_t
                 nullptr == qnn_sys_interface_.systemContextFree;
   ORT_RETURN_IF(result, "Failed to get valid function pointer.");
 
+  void* bin_buffer = nullptr;
+#ifdef QNN_FILE_MAPPED_WEIGHTS_ENABLED
+  void* file_mapping_handle;
+  if (buffer == nullptr) {
+    ORT_RETURN_IF(!file_mapped_weights_enabled_, "Attempting to load QNN context from buffer but buffer is null");
```
yuslepukhin: >file_mapped_weights_enabled_ [](http://example.com/codeflow?start=19&length=28)

This line does not check buffer for nullptr #Resolved

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1106,14 +1263,36 @@ Status QnnBackendManager::LoadCachedQnnContextFromBuffer(char* buffer, uint64_t
...
+    ORT_RETURN_IF(!file_mapper_, "Attemping to use File Mapping feature but file_mapper_ is uninitialized");
+
+    BufferInfo_t buffer_info;
+    ORT_RETURN_IF_ERROR(ReadContextBinIfValid(context_bin_filepath, buffer_info, false));
+    buffer_length = buffer_info.size;
+
```
yuslepukhin, line 1366: This is not clear. This is #ifdef mapping enabled and yet we allocate lots of memory and read the entire file into it. This is going to create memory spikes and may prevent model running in a constrained env. #Resolved

onnxruntime/core/providers/qnn/builder/qnn_windows_file_mapper.cc
```diff
@@ -0,0 +1,379 @@
...
+
+WindowsFileMapper::WindowsFileMapper(const logging::Logger& logger) : logger_(&logger) {
+}
+
+// Close all handles and registered buffers
+// Use LOGS_DEFAULT here as this function will be called during destruction of QnnBackendManager
+// At time of destruction. Usage of logger_ will not be available and will result in a seg fault
+WindowsFileMapper::~WindowsFileMapper() {
+  std::lock_guard<std::mutex> lock(map_mutex_);
+
```
yuslepukhin: Why do we need tis lock? Are we expecting multiple threads destroying the same object? Then we have bigger problems. #Resolved

onnxruntime/core/providers/qnn/builder/qnn_windows_file_mapper.cc
```diff
@@ -0,0 +1,379 @@
...
+  HANDLE file_handle = CreateFileA(bin_filepath.c_str(),
+                                   GENERIC_READ,
+                                   FILE_SHARE_READ,
+                                   NULL,
+                                   OPEN_EXISTING,
+                                   FILE_ATTRIBUTE_NORMAL,
+                                   NULL);
+  ORT_RETURN_IF(file_handle == INVALID_HANDLE_VALUE,
```
yuslepukhin: >file_handle [](http://example.com/codeflow?start=16&length=11)

This handle is leaking on line 72. Check the wil library that has windows oriented smart pointers.
Also, you can close the file handle as soon as mapping handle is obtained, not need to cache it.
 #Resolved

onnxruntime/core/providers/qnn/builder/qnn_windows_file_mapper.cc
```diff
@@ -0,0 +1,379 @@
...
+  return Status::OK();
+}
+
+Status WindowsFileMapper::ReleaseContextBin(const std::string& bin_filepath) {
+  LOGS(*logger_, INFO) << "Removing context bin file mapping for "
+                       << bin_filepath;
+  std::lock_guard<std::mutex> lock(map_mutex_);
+  auto status = Status::OK();
+
+  auto bin_map_it = std::find_if(context_bin_to_mapping_handle_map_.begin(),
```
yuslepukhin: >context_bin_to_mapping_handle_map_ [](http://example.com/codeflow?start=33&length=34)

Here and below. If this is a map, would not make sense to avoid sequential scan and use map find() method? #Resolved

yuslepukhin:
Please, avoid force pushes.

onnxruntime/core/providers/qnn/builder/qnn_windows_file_mapper.cc
```diff
@@ -0,0 +1,379 @@
...
+                                 });
+
+  if (bin_map_it == context_bin_to_mapping_handle_map_.end()) {
+    LOGS(*logger_, VERBOSE) << "File handle does not exist for " << bin_filepath;
+    return status;
+  }
+
+  HANDLE file_mapping_handle = bin_map_it->second;
+  auto mapping_it = std::find_if(mapping_handle_to_info_map_.begin(),
+                                 mapping_handle_to_info_map_.end(),
```
yuslepukhin: Hash tables exist to query elements in constant time, not O(n) #Resolved

onnxruntime/core/providers/qnn/builder/qnn_windows_file_mapper.cc
```diff
@@ -0,0 +1,379 @@
...
+  ORT_UNUSED_PARAMETER(context_bin_to_mapping_handle_map_.erase(bin_filepath));
+
+  return Status::OK();
+}
+
+Status WindowsFileMapper::GetContextBinMappingPointer(const std::string& bin_filepath, void** mapping_ptr) {
+  LOGS(*logger_, INFO) << "Creating mapping pointer for " << bin_filepath;
+
+  std::lock_guard<std::mutex> lock(map_mutex_);
+  auto it = std::find_if(context_bin_to_mapping_handle_map_.begin(),
```
yuslepukhin: > std::find_if [](http://example.com/codeflow?start=11&length=13)

Ditto #Resolved

onnxruntime/core/providers/qnn/builder/qnn_windows_file_mapper.cc
```diff
@@ -0,0 +1,379 @@
...
+
+  HANDLE& file_mapping_handle = it->second;
+
+  LPVOID mapview_ptr = MapViewOfFile(file_mapping_handle,
+                                     FILE_MAP_READ,
+                                     0, 0, 0);
+
+  ORT_RETURN_IF(mapview_ptr == nullptr, "Failed to create mapping pointer for ", bin_filepath);
+
+  if (!context_bin_map_view_pointers_.insert(mapview_ptr).second) {
```
yuslepukhin: >(!context_bin_map_view_pointers_.insert(mapview_ptr).second)  [](http://example.com/codeflow?start=5&length=61)

How is this even possible? You just created a new mapping that is guaranteed to have a unique pointer returned. #Resolved

yuslepukhin:
Please, comment on all Copilot review issues before resolving them.

onnxruntime/core/providers/qnn/builder/qnn_windows_file_mapper.cc
```diff
@@ -0,0 +1,379 @@
...
+
+  HANDLE file_mapping_handle = CreateFileMappingA(file_handle, NULL, PAGE_READONLY, 0x00, 0x00, NULL);
+  ORT_RETURN_IF(file_mapping_handle == INVALID_HANDLE_VALUE,
+                "Failed to create file mapping for context bin ", bin_filepath);
+
+  LOGS(*logger_, INFO) << "Created file mapping with handle (" << file_mapping_handle << ") for context bin:"
+                       << bin_filepath;
+
+  auto inserted = context_bin_to_mapping_handle_map_[bin_filepath] = file_mapping_handle;
+  if (!inserted) {
```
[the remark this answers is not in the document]
quic-calvnguy: Done

onnxruntime/core/providers/qnn/builder/qnn_windows_file_mapper.cc
```diff
@@ -0,0 +1,379 @@
...
+  ORT_UNUSED_PARAMETER(request);
+  ORT_UNUSED_PARAMETER(response);
+  ORT_UNUSED_PARAMETER(notify_param);
+
+  LOGS(*logger_, ERROR) << "File mapping for raw binary data is unsupported on Windows";
+  return QNN_CONTEXT_ERROR_UNSUPPORTED_FEATURE;
+}
+
+// Use LOGS_DEFAULT for all clean up functions below as they will be called during destruction of
+// QnnBackendManagerAt time of destruction. Usage of logger_ will not be available and will result
```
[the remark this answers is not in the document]
quic-calvnguy: Done

onnxruntime/core/providers/qnn/builder/qnn_file_mapping_callback_interface.h
```diff
@@ -0,0 +1,45 @@
...
+  virtual Status FreeContextBinMappingPointer(LPVOID bin_mapping_pointer) = 0;
+
+  virtual Qnn_ErrorHandle_t MapDmaData(Qnn_ContextBinaryDataRequest_t request,
+                                       Qnn_ContextBinaryDmaDataResponse_t* _response,
+                                       void* notify_param) = 0;
+  virtual Qnn_ErrorHandle_t ReleaseDmaData(Qnn_ContextBinaryDmaDataMem_t data_mem,
+                                           void* notify_param) = 0;
+
+  virtual Qnn_ErrorHandle_t MapRawData(Qnn_ContextBinaryDataRequest_t request,
+                                       Qnn_ContextBinaryRawDataResponse_t* _response,
```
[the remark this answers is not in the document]
quic-calvnguy: Done

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
quic-calvnguy: The second parameter will be checked for null in WindowsFileMapper::MapDmaData prior to any operations

onnxruntime/core/providers/qnn/builder/qnn_windows_file_mapper.cc
```diff
@@ -0,0 +1,379 @@
...
+    status = ORT_MAKE_STATUS(ONNXRUNTIME, FAIL,
+                             "File mapping information does not exist for file mapping handle: ",
+                             file_mapping_handle, ", context bin: ", bin_filepath);
+  } else {
+    MappingInfo_t& mapping_info = it->second;
+    file_handle = mapping_info.file_handle;
+    auto mapped_data = mapping_info.mapped_data;
+
+    if (!mapped_data.empty()) {
+      LOGS(*logger_, WARNING) << "Attemping to remove context bin: " << bin_filepath
```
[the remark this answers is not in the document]
quic-calvnguy: Done

onnxruntime/core/providers/qnn/builder/qnn_file_mapping_callback_interface.h
```diff
@@ -0,0 +1,45 @@
...
+class FileMappingCallbackInterface {
+ public:
+  virtual ~FileMappingCallbackInterface() = default;
+  virtual Status MapContextBin(const std::string& bin_filepath,
+                               void** notify_param) = 0;
+  virtual Status ReleaseContextBin(const std::string& model_name) = 0;
+
+  virtual Status GetContextBinMappingPointer(const std::string& bin_filepath, void** mapping_ptr) = 0;
+
+  virtual Status FreeContextBinMappingPointer(LPVOID bin_mapping_pointer) = 0;
```
edgchen1: avoid Windows-specific typedef in the interface? is this feature intended to be enabled on other platforms too?

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
```diff
@@ -451,6 +478,15 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
   bool context_created_ = false;
   bool backend_setup_completed_ = false;
   bool vtcm_backup_buffer_sharing_enabled_ = false;
+  bool file_mapped_weights_enabled_ = false;
+
+#ifdef QNN_FILE_MAPPED_WEIGHTS_ENABLED
+  std::shared_ptr<FileMappingCallbackInterface> file_mapper_ = nullptr;
```
edgchen1: why is shared ownership needed?

onnxruntime/core/providers/qnn/builder/qnn_windows_file_mapper.cc
edgchen1: general: it would be good to output more information about errors from Windows APIs, e.g., with `::GetLastError()`

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
quic-calvnguy: This code is in the provider constructor, and the QnnExecutionProvider class only uses the ctor without the logger in its signature. The logger does not seem to be available at the time this code is executed

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
quic-calvnguy: Changed method of retrieving file size from using ifstream to filesystem::file_size, which should avoid the memory spikes when reading file contents is not required

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
quic-calvnguy: Updated logic flow so checks/error logs are more direct

onnxruntime/core/providers/qnn/builder/qnn_file_mapping_callback_interface.h
quic-calvnguy: This feature is intended to be enabled for other platforms in the future. I wrapped this in a WIN-specific typedef so this won't be included for the time-being. Will remove the typedef though.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -475,6 +475,21 @@ QNNExecutionProvider::QNNExecutionProvider(const ProviderOptions& provider_optio
...
+  }
+
+#ifndef QNN_FILE_MAPPED_WEIGHTS_ENABLED
+  enable_file_mapped_weights_ = false;
+  LOGS_DEFAULT(WARNING) << "File mapped weights feature is only available on Windows arm64 devices for QNN API versions >= 2.32. "
+                        << "Feature will be disabled by default";
```
edgchen1, line 490: is there sufficient test coverage for this feature? looks like this feature will be enabled by default now if supported, but should we also ensure that the path where it is disabled (on the same platform) still works?

onnxruntime/core/providers/qnn/builder/qnn_windows_file_mapper.cc
quic-calvnguy: Done

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
quic-calvnguy: Added two new unit tests (one for each context creation pathway)

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
quic-calvnguy: Changed to unique_ptr

onnxruntime/core/providers/qnn/builder/qnn_windows_file_mapper.cc
quic-calvnguy: Removed the if statement

onnxruntime/core/providers/qnn/builder/qnn_windows_file_mapper.cc
quic-calvnguy: Replaced with unordered_map::find()

onnxruntime/core/providers/qnn/builder/qnn_windows_file_mapper.cc
quic-calvnguy: Replaced with unordered_map::find()

onnxruntime/core/providers/qnn/builder/qnn_windows_file_mapper.cc
quic-calvnguy: Replaced with unordered_map::find()

onnxruntime/core/providers/qnn/builder/qnn_windows_file_mapper.cc
quic-calvnguy: Added code to clean up the file mapping handle if the mapview handle creation fails

onnxruntime/core/providers/qnn/builder/qnn_windows_file_mapper.cc
quic-calvnguy: Removed the lock

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
quic-calvnguy: Removed these clear statements

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
quic-calvnguy: Removed

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
quic-calvnguy: Replaced the code to retrieve file size with std::filesystem::file_size. Additionally, there is now code that handles 32-bit systems--if the file size is greater than the max supported size_t value, then error out.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
quic-calvnguy: All the QNN EP option strings can be centralized in a new file. However, is it possible to do this in a clean-up task of some sorts given the scope of this PR?

edgchen1:
@quic-calvnguy would you mind merging from main? github is claiming there are conflicts and as a result we can't start the CI builds.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -845,6 +916,26 @@ Status QnnBackendManager::CreateContextVtcmBackupBufferSharingEnabled(std::unord
 #endif
                                           nullptr};
 
+#ifdef QNN_FILE_MAPPED_WEIGHTS_ENABLED
+  if (file_mapped_weights_enabled_ && file_mapper_) {
+    // Retry logic -- if context creation failed with file mapped weights, then retry with feature disabled
+    if (CreateContextFromListAsyncV2(configs, context_bin_map) != Status::OK()) {
+      LOGS(*logger_, WARNING) << "Failed to create context with file mapping enabled. Retrying with feature disabled.";
```
edgchen1: should the failed status details also be logged?

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
```diff
@@ -263,9 +276,23 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
...
+
   Status CreateContextVtcmBackupBufferSharingEnabled(std::unordered_map<std::string,
                                                                         std::unique_ptr<std::vector<std::string>>>& context_bin_map);
 
+  Status CreateContextFromListAsyncV1(const QnnContext_Config_t** configs,
+                                      std::unordered_map<std::string,
+                                                         std::unique_ptr<std::vector<std::string>>>& context_bin_map);
+
+#ifdef QNN_FILE_MAPPED_WEIGHTS_ENABLED
+  Status CreateContextFromListAsyncV2(const QnnContext_Config_t** configs,
```
edgchen1: maybe document the difference between V1 and V2, or name them more descriptively. e.g., `CreateContextFromListWithFileMappingAsync`

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -892,14 +975,82 @@ Status QnnBackendManager::CreateContextVtcmBackupBufferSharingEnabled(std::unord
...
+  }
+  context_params_ptr_list.push_back(nullptr);
+  auto result = qnn_interface_.contextCreateFromBinaryListAsync(backend_handle_,
+                                                                device_handle_,
+                                                                context_params_ptr_list.data(),
+                                                                configs,
+                                                                nullptr);
+
+  for (auto& buffer : buffer_list) {
+    ORT_RETURN_IF_ERROR(file_mapper_->FreeContextBinMappingPointer(buffer));
```
edgchen1: there are potential leaks here. can this mapping pointer be managed by some RAII type like `unique_ptr`? there's also [`gsl::finally`](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#e19-use-a-final_action-object-to-express-cleanup-if-no-suitable-resource-handle-is-available) that can be used for ad-hoc cleanup when leaving a scope. #Resolved

onnxruntime/core/providers/qnn/builder/qnn_windows_file_mapper.cc
```diff
@@ -0,0 +1,372 @@
...
+  return Status::OK();
+}
+
+Status WindowsFileMapper::ReleaseContextBin(const std::string& bin_filepath) {
+  LOGS(*logger_, INFO) << "Removing context bin file mapping for "
+                       << bin_filepath;
+  std::lock_guard<std::mutex> lock(map_mutex_);
+  auto status = Status::OK();
+
+  auto bin_map_it = context_bin_to_mapping_handle_map_.find(bin_filepath);
```
yuslepukhin: >context_bin_to_mapping_handle_map_ [](http://example.com/codeflow?start=20&length=34)

What is the purpose of storing file_mapping_handles?

please, refer to ORT implementation of file mapping.

[https://github.com/microsoft/onnxruntime/blob/4013dc193f66e8e1d8b51b5fd693ca285b251226/onnxruntime/core/platform/windows/env.cc?plain=1#L403C1-L403C72](https://github.com/microsoft/onnxruntime/blob/4013dc193f66e8e1d8b51b5fd693ca285b251226/onnxruntime/core/platform/windows/env.cc?plain=1#L403C1-L403C72) #Resolved

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -892,14 +975,82 @@ Status QnnBackendManager::CreateContextVtcmBackupBufferSharingEnabled(std::unord
...
+
+    auto& notify_param = file_mapping_notify_params_.emplace_back(file_mapper_.get(),
+                                                                  file_mapping_handle);
+
+    Qnn_ContextBinaryCallback_t context_file_map_callbacks;
+    context_file_map_callbacks.type = QNN_CONTEXT_CALLBACK_DMA_BUFFER;
+    context_file_map_callbacks.dmaBufferCallback.version = QNN_CONTEXT_CALLBACK_DMA_BUFFER_VERSION_1;
+    context_file_map_callbacks.dmaBufferCallback.v1.dataProvide = DmaDataProvider;
+    context_file_map_callbacks.dmaBufferCallback.v1.dataRelease = DmaDataRelease;
+    context_file_map_callbacks.dmaBufferCallback.v1.notifyParam = reinterpret_cast<void*>(&notify_param);
```
edgchen1: storing a pointer to a vector element is risky. the vector's storage can be reallocated which would invalidate the pointer.

an alternative is to use a vector of unique_ptrs instead and store the address returned by `unique_ptr::get()`.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1106,14 +1258,36 @@ Status QnnBackendManager::LoadCachedQnnContextFromBuffer(char* buffer, uint64_t
...
+#ifdef QNN_FILE_MAPPED_WEIGHTS_ENABLED
+  void* file_mapping_handle;
+  if (file_mapped_weights_enabled_) {
+    ORT_RETURN_IF(!file_mapper_, "Attemping to use File Mapping feature but file_mapper_ is uninitialized");
+
+    BufferInfo_t buffer_info;
+    ORT_RETURN_IF_ERROR(ReadContextBinIfValid(context_bin_filepath, buffer_info, false));
+    buffer_length = buffer_info.size;
+
+    ORT_RETURN_IF(buffer_length == 0, "Context bin has a size of 0 bytes", context_bin_filepath);
```
edgchen1:
```suggestion
    ORT_RETURN_IF(buffer_length == 0, "Context bin has a size of 0 bytes: ", context_bin_filepath);
```

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1188,20 +1362,73 @@ Status QnnBackendManager::LoadCachedQnnContextFromBuffer(char* buffer, uint64_t
...
+                                                              bin_buffer,
+                                                              buffer_length,
+                                                              &context,
+                                                              profile_backend_handle_,
+                                                              NULL);
+
+      ORT_RETURN_IF_ERROR(file_mapper_->FreeContextBinMappingPointer(bin_buffer));
+
+      if (rt != QNN_SUCCESS) {
+        LOGS(*logger_, WARNING) << "Failed to create context with file mapping enabled. Retrying with feature disabled.";
```
edgchen1: can we log the details of the failure from `rt`?

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1188,20 +1362,73 @@ Status QnnBackendManager::LoadCachedQnnContextFromBuffer(char* buffer, uint64_t
...
+                                                              device_handle_,
+                                                              context_configs,
+                                                              &callbacks,
+                                                              bin_buffer,
+                                                              buffer_length,
+                                                              &context,
+                                                              profile_backend_handle_,
+                                                              NULL);
+
+      ORT_RETURN_IF_ERROR(file_mapper_->FreeContextBinMappingPointer(bin_buffer));
```
edgchen1: there's a leak if `FreeContextBinMappingPointer()` is not called, e.g., with an early return/exception

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1335,6 +1564,12 @@ Status QnnBackendManager::SetupBackend(const logging::Logger& logger,
     LOGS(logger, VERBOSE) << "LoadOpPackage succeed.";
   }
 
+#ifdef QNN_FILE_MAPPED_WEIGHTS_ENABLED
+  if (file_mapped_weights_enabled_ && !file_mapper_ && GetQnnBackendType() == QnnBackendType::HTP) {
```
edgchen1: the names `QNN_FILE_MAPPED_WEIGHTS_ENABLED` and `file_mapped_weights_enabled_` mean different things. perhaps "available" would be better than "enabled" for the former.

onnxruntime/core/providers/qnn/builder/qnn_file_mapping_callback_interface.h
edgchen1:
```suggestion
  virtual Status FreeContextBinMappingPointer(void* bin_mapping_pointer) = 0;
```

onnxruntime/core/providers/qnn/builder/qnn_windows_file_mapper.h
```diff
@@ -0,0 +1,75 @@
...
+
+  Qnn_ErrorHandle_t MapRawData(Qnn_ContextBinaryDataRequest_t request,
+                               Qnn_ContextBinaryRawDataResponse_t* response,
+                               void* notify_param) override;
+  Qnn_ErrorHandle_t ReleaseRawData(Qnn_ContextBinaryRawDataMem_t data_mem,
+                                   void* notify_param) override;
+
+ private:
+  typedef struct MappedDataInfo {
+    LPVOID aligned_data_ptr = nullptr;
```
edgchen1: include the appropriate Windows header (Windows.h?) if we are using Windows typedefs and functions.

onnxruntime/core/providers/qnn/builder/qnn_windows_file_mapper.h
```diff
@@ -0,0 +1,75 @@
...
+  void CloseHandles(HANDLE file_handle, HANDLE file_mapping_handle);
+
+  std::mutex map_mutex_;  // Applies to both unordered maps
+  std::unordered_map<std::string, HANDLE> context_bin_to_mapping_handle_map_;
+  std::unordered_map<HANDLE, MappingInfo_t> mapping_handle_to_info_map_;
+  std::unordered_set<LPVOID> context_bin_map_view_pointers_;
+
+  const logging::Logger* logger_;
+
+  RpcMemLibrary rpcmem_lib_;
```
edgchen1: this may be considered a lower priority enhancement, but it would be nice to reuse an existing `RpcMemLibrary` instance if possible. there's already a `shared_ptr<RpcMemLibrary>` that gets passed to `HtpSharedMemoryAllocator`.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -809,6 +845,41 @@ void QnnBackendManager::ProcessContextFromBinListAsync(Qnn_ContextHandle_t conte
...
+  ORT_RETURN_IF(ec, "Failed to retrieve size of context binary: ", context_bin_filepath,
+                ", error: ", ec.message());
+  ORT_RETURN_IF(file_size == 0, "Context binary is empty: ", context_bin_filepath);
+  ORT_RETURN_IF(file_size > SIZE_MAX, "Context binary (", context_bin_filepath, ") file size (", file_size,
+                " bytes) exceeds maximum value of size_t for this platform (", SIZE_MAX, " bytes).");
+
+  size_t buffer_size = static_cast<size_t>(file_size);
+
+  std::unique_ptr<char[]> buffer;
+  if (read_file_contents) {
```
yuslepukhin: >read_file_contents [](http://example.com/codeflow?start=6&length=18)

What happens otherwise?

Why don't we exit early? #Resolved

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -809,6 +845,41 @@ void QnnBackendManager::ProcessContextFromBinListAsync(Qnn_ContextHandle_t conte
...
+
+    buffer = std::make_unique<char[]>(buffer_size);
+    ORT_RETURN_IF(nullptr == buffer, "Failed to allocate memory for cache file.");
+
+    const auto& read_result = cache_file.read(buffer.get(), buffer_size);
+    ORT_RETURN_IF(!read_result, "Failed to read contents from cached context file.");
+  }
+
+  buffer_info.data = std::move(buffer);
+  buffer_info.size = buffer_size;
```
yuslepukhin: So it is possible that the buffer_size is non zero, but the buffer is nullptr?

Would it not be better simply to create a function GetFileSize() which is already a part of the std library? #Resolved

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -770,22 +774,54 @@ Status SetQnnContextConfig(ContextPriority context_priority, QnnContext_Config_t
   return Status::OK();
 }
 
+#ifdef QNN_FILE_MAPPED_WEIGHTS_ENABLED
+// callback required for allocating file mapping resources
+static Qnn_ErrorHandle_t DmaDataProvider(Qnn_ContextBinaryDataRequest_t request,
```
yuslepukhin: >DmaDataProvider [](http://example.com/codeflow?start=25&length=15)

Is this a noun or a verb? It is not clear from the name what it does. #Resolved

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -770,22 +774,54 @@ Status SetQnnContextConfig(ContextPriority context_priority, QnnContext_Config_t
...
+
+  if (pair->first == nullptr) {
+    LOGS_DEFAULT(ERROR) << "DmaProvider: file mapper is null";
+    return QNN_CONTEXT_ERROR_INVALID_ARGUMENT;
+  }
+
+  return pair->first->MapDmaData(request, response, pair->second);
+}
+
+// callback required for releasing file mapping resources
```
yuslepukhin: >file mapping [](http://example.com/codeflow?start=35&length=12)

What does file mapping has to do with DMA? #Resolved

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
yuslepukhin: This leaks if the function exists on error from a loop or throws an exception

onnxruntime/core/providers/qnn/builder/qnn_windows_file_mapper.h
```diff
@@ -0,0 +1,75 @@
...
+
+  void CleanUpDataMapping(LPVOID unaligned_data_ptr, LPVOID aligned_data_ptr,
+                          size_t buffer_size);
+  void CleanUpDataMappings(const std::unordered_map<LPVOID, MappedDataInfo_t>& mapped_data);
+  void CloseHandles(HANDLE file_handle, HANDLE file_mapping_handle);
+
+  std::mutex map_mutex_;  // Applies to both unordered maps
+  std::unordered_map<std::string, HANDLE> context_bin_to_mapping_handle_map_;
+  std::unordered_map<HANDLE, MappingInfo_t> mapping_handle_to_info_map_;
+  std::unordered_set<LPVOID> context_bin_map_view_pointers_;
```
yuslepukhin: No documentation as to what maps to what and how it is used. #Resolved

onnxruntime/core/providers/qnn/builder/qnn_windows_file_mapper.cc
edgchen1: wil looks cool. they have a `unique_hfile` which seems well-suited for this.

https://github.com/microsoft/wil/wiki/RAII-resource-wrappers#available-unique_any-handle-types

johnpaultaken:
> Discussed offline, the EP maps initializers form the binary context, not from the external weights files.

It will be useful to have binary context memory mapping supported for all EPs at an infrastructure level instead of all EPs having to implement it. Afaik other EPs like OpenVino also use binary context. I wonder if/how they do it.

onnxruntime/core/providers/qnn/builder/qnn_file_mapping_callback_interface.h
quic-calvnguy: No longer relevant

onnxruntime/core/providers/qnn/builder/qnn_windows_file_mapper.cc
quic-calvnguy: Synced offline -- no longer need to store file_mapping_handles

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
quic-calvnguy: Done

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
quic-calvnguy: Updated to use unique_ptrs

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
quic-calvnguy: Done

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
quic-calvnguy: Updated to QNN_FILE_MAPPED_WEIGHTS_AVAILABLE

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
quic-calvnguy: Added separate function (platform agnostic) to retrieve file size and removed this branch case

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
quic-calvnguy: Updated to MapDmaData

onnxruntime/core/providers/qnn/builder/qnn_windows_file_mapper.h
quic-calvnguy: Synced offline & maps are no longer relevant

onnxruntime/core/providers/qnn/builder/qnn_windows_file_mapper.h
quic-calvnguy: No longer relevant

onnxruntime/core/providers/qnn/builder/qnn_windows_file_mapper.h
quic-calvnguy: Done

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -892,15 +981,74 @@ Status QnnBackendManager::CreateContextVtcmBackupBufferSharingEnabled(std::unord
...
+                                                                 std::unordered_map<std::string,
+                                                                                    std::unique_ptr<std::vector<std::string>>>& context_bin_map) {
+  std::vector<QnnContext_Params_t> context_params_list;
+  std::vector<QnnContext_ParamsV2_t> context_paramsv2_list;
+  std::vector<Qnn_ContextBinaryCallback_t> context_callbacks_list;
+  std::vector<const QnnContext_Params_t*> context_params_ptr_list;
+
+  context_params_list.reserve(context_bin_map.size());
+  context_callbacks_list.reserve(context_bin_map.size());
+  context_params_ptr_list.reserve(context_bin_map.size() + 1);
```
edgchen1, line 1100: also reserve space for `context_paramsv2_list`?

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -892,15 +981,74 @@ Status QnnBackendManager::CreateContextVtcmBackupBufferSharingEnabled(std::unord
...
+                                               buffer,
+                                               buffer_size,
+                                               nullptr,
+                                               ContextCreateAsyncCallback,
+                                               it.second.get(),
+                                               &context_file_map_callbacks};
+
+    QnnContext_Params_t context_params = {QnnContext_ParamsVersion_t::QNN_CONTEXT_PARAMS_VERSION_2,
+                                          {}};
+    context_params.v2 = &context_params_v2;
```
edgchen1: this local `context_params_v2` will go out of scope. do we want to be taking its address, or the address of the `context_paramsv2_list` element instead?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -552,7 +567,7 @@ QNNExecutionProvider::QNNExecutionProvider(const ProviderOptions& provider_optio
   }
 
   static const std::string QNN_HTP_SHARED_MEMORY_ALLOCATOR_ENABLED = "enable_htp_shared_memory_allocator";
-  if (ParseBoolOption(QNN_HTP_SHARED_MEMORY_ALLOCATOR_ENABLED, false, provider_options_map)) {
+  if (ParseBoolOption(QNN_HTP_SHARED_MEMORY_ALLOCATOR_ENABLED, false, provider_options_map) || enable_file_mapped_weights_) {
     // Initialize rpcmem_library_.
     // This is necessary for HtpSharedMemoryAllocator to function and also indicates that the allocator is available.
```
edgchen1, line 573: the original assumption was that `rpcmem_library_` was only used by the HTP shared memory allocator. now that it may also be used by the file mapper, code relying on the original assumption should be updated. e.g., `QNNExecutionProvider::IsHtpSharedMemoryAllocatorAvailable()` and the comments for `QNNExecutionProvider::rpcmem_library_` in the header. possibly another bool can be introduced to indicate whether the HTP shared memory allocator is enabled.

onnxruntime/core/providers/qnn/builder/qnn_windows_file_mapper.h
```diff
@@ -0,0 +1,54 @@
...
+
+#include "core/providers/qnn/builder/qnn_file_mapping_callback_interface.h"
+#ifdef QNN_FILE_MAPPED_WEIGHTS_AVAILABLE
+
+#include <string>
+#include <unordered_map>
+
+#include <QnnContext.h>
+
+#include "core/platform/env.h"
```
edgchen1: this should already be covered by `#include "core/providers/qnn/ort_api.h"`. ideally, inclusion of ORT code from QNN EP should be limited to that header so that both the provider-bridge and static library builds work.

onnxruntime/core/providers/qnn/builder/qnn_windows_file_mapper.cc
```diff
@@ -0,0 +1,224 @@
...
+
+WindowsFileMapper::WindowsFileMapper(const logging::Logger& logger,
+                                     std::shared_ptr<qnn::RpcMemLibrary> rpcmem_lib)
+    : logger_(&logger),
+      rpcmem_lib_(rpcmem_lib) {
+}
+
+// Close all handles and registered buffers
+// Use LOGS_DEFAULT here as this function will be called during destruction of QnnBackendManager
+// At time of destruction. Usage of logger_ will not be available and will result in a seg fault
```
edgchen1: nice that the dtor is now empty. this comment can be updated.

onnxruntime/core/providers/qnn/builder/qnn_windows_file_mapper.cc
```diff
@@ -0,0 +1,224 @@
...
+                                                        UnmapFile(mapped_base_ptr);
+                                                      }};
+
+  *mapped_data_ptr = mapped_memory_ptr.get();
+  mapped_memory_ptrs.push_back(std::move(mapped_memory_ptr));
+
+  return Status::OK();
+}
+
+void WindowsFileMapper::UnmapFile(void* addr) noexcept {
```
edgchen1: nit: this doesn't need to be a public static member function. it could be a file-scoped free function (e.g., `static` or in the unnamed namespace).

onnxruntime/core/providers/qnn/builder/qnn_windows_file_mapper.cc
```diff
@@ -0,0 +1,224 @@
...
+  SIZE_T delta = request.offset - aligned_offset;
+
+  void* aligned_data_ptr = static_cast<char*>(mapped_data_ptr) + aligned_offset;
+
+  void* unaligned_data_ptr = static_cast<char*>(aligned_data_ptr) + delta;
+  LOGS(*logger_, INFO) << "Created DMA data mapping with: address(" << aligned_data_ptr
+                       << "), aligned offset(" << aligned_offset << "), delta(" << delta
+                       << "), unaligned address(" << unaligned_data_ptr << ")";
+
+  if (!rpcmem_lib_) {
```
edgchen1: perhaps `rpcmem_lib_ != nullptr` can be checked in the constructor

onnxruntime/core/providers/qnn/builder/qnn_windows_file_mapper.h
```diff
@@ -0,0 +1,54 @@
...
+
+  Qnn_ErrorHandle_t MapRawData(Qnn_ContextBinaryDataRequest_t request,
+                               Qnn_ContextBinaryRawDataResponse_t* response,
+                               void* mapped_data_ptr) override;
+  Qnn_ErrorHandle_t ReleaseRawData(Qnn_ContextBinaryRawDataMem_t data_mem,
+                                   void* mapped_data_ptr) override;
+
+ private:
+  // A container of smart pointers of mapview memory pointers to mapped context bins
+  std::vector<onnxruntime::Env::MappedMemoryPtr> mapped_memory_ptrs;
```
edgchen1: data member name should end with "_", `mapped_memory_ptrs_`

onnxruntime/core/providers/qnn/builder/qnn_windows_file_mapper.cc
```diff
@@ -0,0 +1,224 @@
...
+    return QNN_COMMON_ERROR_RESOURCE_UNAVAILABLE;
+  }
+
+  rpcmem_lib_->Api().register_buf(unaligned_data_ptr, buffer_size, NULL,
+                                  rpcmem::RPCMEM_ATTR_IMPORT_BUFFER | rpcmem::RPCMEM_ATTR_READ_ONLY);
+
+  auto fd = rpcmem_lib_->Api().to_fd(unaligned_data_ptr);
+  if (fd == -1) {
+    LOGS(*logger_, ERROR) << "Failed to register DMA data mapping to RPCMEM";
+    if (!UnmapViewOfFile(aligned_data_ptr)) {
```
edgchen1: where is the corresponding call to `MapViewOfFile()`?

onnxruntime/core/providers/qnn/builder/qnn_windows_file_mapper.cc
```diff
@@ -0,0 +1,224 @@
...
+  LOGS(*logger_, INFO) << "Created mapview pointer with address " << mapped_base_ptr
+                       << " for context bin " << bin_filepath;
+
+  onnxruntime::Env::MappedMemoryPtr mapped_memory_ptr{reinterpret_cast<char*>(mapped_base_ptr),
+                                                      [mapped_base_ptr](void*) {
+                                                        UnmapFile(mapped_base_ptr);
+                                                      }};
+
+  *mapped_data_ptr = mapped_memory_ptr.get();
+  mapped_memory_ptrs.push_back(std::move(mapped_memory_ptr));
```
edgchen1: if this function gets called multiple times with the same `bin_filepath`, it will keep creating new mappings. should it attempt to reuse an existing mapping for the same `bin_filepath` value?

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
quic-calvnguy: Updated in yesterday's patch -- fail status & error code now logged

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
quic-calvnguy: Outdated -- previous patch removed this code entirely, and all file handle destruction is now handled by smart pointers

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
quic-calvnguy: Done

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
quic-calvnguy: Outdated -- no longer applicable as of yesterday's patch

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
quic-calvnguy: Updated such that file size retrieval and reading of context bin have been separated

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
quic-calvnguy: Discussed offline: DMA mapping is required to map weights from context bin to NPU memory space. The core feature is still Windows file mapping of context bin/weights

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
quic-calvnguy: Done

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
quic-calvnguy: Updated to set context_params.v2 to the context_paramsv2_list element after adding to the list

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
quic-calvnguy: Updated such that rpcmem_library_ is initialized here only if QnnHTPSharedMemoryAllocator is required. IsHtpSharedMemoryAllocatorAvailable() now also checks if the memory allocator is enabled as well.

There is another check right after specifically for file mapping such that if rpcmem_library_ is not already initialized. If rpcmem_library_ fails to initialize, then file mapping will be turned off entirely, and normal operation will resume.

onnxruntime/core/providers/qnn/builder/qnn_windows_file_mapper.cc
quic-calvnguy: Comments have been removed

onnxruntime/core/providers/qnn/builder/qnn_windows_file_mapper.cc
quic-calvnguy: Done

onnxruntime/core/providers/qnn/builder/qnn_windows_file_mapper.h
quic-calvnguy: Updated to be an unordered_map alongside a mutex. The mutex lock only occurs during actual mapping of the context binary, and there is now a check to see if a mapview memory pointer already exists for some context binary.

onnxruntime/core/providers/qnn/builder/qnn_windows_file_mapper.cc
quic-calvnguy: Code to unmap pointer removed as it is no longer needed

onnxruntime/core/providers/qnn/builder/qnn_windows_file_mapper.cc
quic-calvnguy: Updated to be an unordered_map alongside a mutex. The mutex lock only occurs during actual mapping of the context binary, and there is now a check to see if a mapview memory pointer already exists for some context binary.

onnxruntime/core/providers/qnn/builder/qnn_windows_file_mapper.h
quic-calvnguy: Removed include line

onnxruntime/core/providers/qnn/builder/qnn_windows_file_mapper.cc
quic-calvnguy: Removed this check as ctor now includes an ORT_ENFORCE on the rpcmem_lib_ pointer

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1030,11 +1031,13 @@ Status QnnBackendManager::CreateContextFromListAsyncWithCallback(const QnnContex
...
     QnnContext_Params_t context_params = {QnnContext_ParamsVersion_t::QNN_CONTEXT_PARAMS_VERSION_2,
                                           {}};
-    context_params.v2 = &context_params_v2;
+
     context_params_list.push_back(std::move(context_params));
     context_callbacks_list.push_back(std::move(context_file_map_callbacks));
     context_paramsv2_list.push_back(std::move(context_params_v2));
     context_params_ptr_list.push_back(&(context_params_list.back()));
+
+    context_params.v2 = &context_paramsv2_list.back();
```
edgchen1, line 1137: `context_params` has already been moved from into `context_params_list` at this point.

onnxruntime/test/providers/qnn/qnn_test_utils.cc
```diff
@@ -408,12 +408,28 @@ static BackendSupport GetHTPSupport(const onnxruntime::logging::Logger& logger)
...
+  std::vector<std::unique_ptr<ComputeCapability>> result;
+  std::unique_ptr<onnxruntime::IExecutionProvider> qnn_ep;
+  try {
+    qnn_ep = QnnExecutionProviderWithOptions(
+        {{"backend_type", "htp"}, {"offload_graph_io_quantization", "0"}, {"enable_htp_shared_memory_allocator", "1"}});
+    GraphOptimizerRegistry graph_optimizer_registry(nullptr, nullptr, nullptr);  // as a placeholder to feed into GetCapability
+
+    qnn_ep->SetLogger(&logger);
+    result = qnn_ep->GetCapability(graph_viewer, kernel_lookup, graph_optimizer_registry, nullptr);
+  } catch (const std::exception& e) {
```
quic-calvnguy, line 421: This change is based on existing code here:
https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/test/shared_lib/test_inference.cc#L1987

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
quic-calvnguy: Updated

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -552,11 +567,24 @@ QNNExecutionProvider::QNNExecutionProvider(const ProviderOptions& provider_optio
...
     rpcmem_library_ = std::make_shared<qnn::RpcMemLibrary>();
-    model_settings_.htp_shared_memory = true;
+    model_settings_.htp_shared_memory = enable_htp_shared_mem_allocator_;
+  }
+
+  if (enable_file_mapped_weights_ && !rpcmem_library_) {
+    // Attempt to init rpcmem_library_ if needed. If this fails, then
+    // disable file mapped weights and proceed with normal operation
+    try {
+      rpcmem_library_ = std::make_shared<qnn::RpcMemLibrary>();
```
edgchen1, line 582: nit: I think that the initialization of `rpcmem_library_` can be unified. something like:

```c++
if (enable_file_mapped_weights_ || enable_htp_shared_mem_allocator_) {
    // initialize rpcmem_library_.
    // if that fails, disable the features that depend on it.
}
```

it can be cleaned up in a later PR.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -909,6 +937,11 @@ QNNExecutionProvider::GetCapability(const onnxruntime::GraphViewer& graph_viewer
 
   const auto& logger = *GetLogger();
 
+  if (enable_file_mapped_weights_ && !rpcmem_library_) {
```
edgchen1: when would this condition be true? wouldn't the constructor logic ensure otherwise?

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1188,20 +1358,75 @@ Status QnnBackendManager::LoadCachedQnnContextFromBuffer(char* buffer, uint64_t
...
+                                                              NULL);
+
+      if (rt != QNN_SUCCESS) {
+        LOGS(*logger_, WARNING) << "Failed to create context with file mapping enabled. Error: "
+                                << QnnErrorHandleToString(result) << ", Code : " << result
+                                << ". Retrying with feature disabled.";
+
+        file_mapped_weights_enabled_ = false;
+        // Destruction of file_mapper_ to prevent resource leaks
+        file_mapper_.reset();
```
edgchen1: I'm not very familiar with the overall workflow, but just a basic question - is it safe to destroy the file mapper here? if there was a previous call to this function where `contextCreateFromBinaryWithCallback()` succeeded, could the entity that invokes the callbacks still have a pointer to the destroyed file mapper instance?

onnxruntime/core/providers/qnn/builder/qnn_windows_file_mapper.cc
```diff
@@ -0,0 +1,222 @@
...
+
+  *mapped_data_ptr = mapped_memory_ptr.get();
+  mapped_memory_ptrs_.emplace(bin_filepath, std::move(mapped_memory_ptr));
+
+  return Status::OK();
+}
+
+Qnn_ErrorHandle_t WindowsFileMapper::MapDmaData(Qnn_ContextBinaryDataRequest_t request,
+                                                Qnn_ContextBinaryDmaDataResponse_t* response,
+                                                void* mapped_data_ptr) {
```
yuslepukhin: What is confusing about this code is that DmaData mapping is implemented in the windows_file_mapping class and code. I feel these two must be separated. Then both would become re-usable and not tightly married to each other.
That would also allow creation a platform independent interface for file_mapping and have separate implementation for Posix platforms. That would enable the code that makes use of it generic.
 #Resolved

onnxruntime/test/providers/qnn/qnn_test_utils.cc
edgchen1: is this still necessary? now that there's also exception handling added to the QNNExecutionProvider constructor

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
quic-calvnguy: The main reason I split it up is because a lot of the unit tests check for NPU availability the thrown exception to as an indicator that the NPU backend is not available. I agree though, this should be unified and cleaned up in a later PR.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -786,22 +790,136 @@ Status SetQnnContextConfig(ContextPriority context_priority, QnnContext_Config_t
...
+    return QNN_CONTEXT_ERROR_INVALID_ARGUMENT;
+  }
+
+  return pair->first->MapDmaData(request, response, pair->second);
+}
+
+Qnn_ErrorHandle_t QnnBackendManager::MapDmaData(Qnn_ContextBinaryDataRequest_t request,
+                                                Qnn_ContextBinaryDmaDataResponse_t* response,
+                                                void* mapped_base_ptr) {
+  if (!file_mapped_weights_enabled_) {
```
edgchen1, line 817: is a callback able to run from a separate thread while other methods (e.g., the code that resets `file_mapped_weights_enabled_` and `file_mapper_`) may be running? should we take some precautions for thread-safety?

onnxruntime/core/providers/qnn/builder/qnn_windows_file_mapper.cc
```diff
@@ -0,0 +1,135 @@
...
+
+FileMappingInterface::MappedWeightInfo_t
+WindowsFileMapper::GetMappedWeightMemoryPtr(void* mapped_base_ptr,
+                                            const size_t offset) {
+  MappedWeightInfo_t mapped_weight_info;
+
+  // Align to nearest granularity boundary
+  SYSTEM_INFO sys_info;
+  GetSystemInfo(&sys_info);
+  Qnn_ContextBinarySize_t granularity = sys_info.dwAllocationGranularity;
```
edgchen1: `Qnn_ContextBinarySize_t` seems like an unusual choice. would `size_t` be more consistent with the other values?

onnxruntime/core/providers/qnn/builder/qnn_file_mapping_interface.h
```diff
@@ -0,0 +1,35 @@
...
+
+namespace onnxruntime {
+namespace qnn {
+
+class FileMappingInterface {
+ public:
+  typedef struct MappedWeightInfo {
+    size_t aligned_offset = 0;
+    size_t delta = 0;
+    void* aligned_data_ptr = nullptr;
```
edgchen1: perhaps I'm missing something. I can't seem to find where `aligned_offset`, `delta`, or `aligned_data_ptr` are used outside of being logged. what are they for?

onnxruntime/core/providers/qnn/builder/qnn_windows_file_mapper.cc
```diff
@@ -0,0 +1,135 @@
...
+
+  *mapped_data_ptr = mapped_memory_ptr.get();
+  mapped_memory_ptrs_.emplace(bin_filepath, std::move(mapped_memory_ptr));
+
+  return Status::OK();
+}
+
+FileMappingInterface::MappedWeightInfo_t
+WindowsFileMapper::GetMappedWeightMemoryPtr(void* mapped_base_ptr,
+                                            const size_t offset) {
```
yuslepukhin: > const size_t offset) [](http://example.com/codeflow?start=43&length=21)

I would actually add a check that offset does not go beyond the mapping so the client code does not deref an invalid ptr. #Resolved

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
```diff
@@ -248,10 +256,24 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
...
+
   QnnLog_Level_t MapOrtSeverityToQNNLogLevel(logging::Severity ort_log_level);
   static logging::Severity MapQNNLogLevelToOrtSeverity(QnnLog_Level_t qnn_log_level);
 
  private:
+  typedef struct BufferInfo {
+    std::unique_ptr<char[]> data;
+    size_t size;
+  } BufferInfo_t;
+
```
yuslepukhin, line 283: I would simply use an std::vector for this. #Resolved

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -861,33 +1031,45 @@ Status QnnBackendManager::CreateContextVtcmBackupBufferSharingEnabled(std::unord
...
 
+#ifdef QNN_FILE_MAPPED_WEIGHTS_AVAILABLE
+  if (FileMappingIsEnabled() && file_mapper_) {
+    // Retry logic -- if context creation failed with file mapped weights, then retry with feature disabled
+    auto res = CreateContextFromListAsyncWithCallback(configs, context_bin_map);
+    if (!res.IsOK()) {
+      DisableFileMapping();
+      LOGS(*logger_, WARNING) << res.ErrorMessage() << ". Retrying with feature disabled.";
+
+      file_mapped_weights_enabled_ = false;
```
edgchen1: I see there are new helper functions `FileMappingIsEnabled()` and `DisableFileMapping()` that lock a mutex.

however, assuming that the mutex is intended to synchronize access to `file_mapped_weights_enabled_` (and possibly other state), the locking here seems too fine-grained to prevent other threads from reading/writing the shared state.

effectively, this is:
```c++
file_mapping_mutex_.lock();
bool enabled = file_mapping_weights_enabled_;
file_mapping_mutex_.unlock();
if (enabled && ...) {

    //...

    file_mapping_mutex_.lock();
    file_mapping_weights_enabled_ = false;
    file_mapping_mutex_.unlock();
}
```

the mutex should probably be locked for the entirety of the reading and updating of the shared state.

this is a general comment that applies to other places too.

edgchen1: also, it would be simpler if the shared state was immutable by the time it gets accessed by multiple threads. is the capability of turning off this feature useful enough to warrant the additional complexity needed for shared mutable state?

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -825,6 +956,45 @@ void QnnBackendManager::ProcessContextFromBinListAsync(Qnn_ContextHandle_t conte
...
+  return Status::OK();
+}
+
+Status QnnBackendManager::ReadContextBinIfValid(const std::string& context_bin_filepath,
+                                                BufferInfo_t& buffer_info) {
+  size_t buffer_size;
+  ORT_RETURN_IF_ERROR(GetFileSizeIfValid(context_bin_filepath, buffer_size));
+
+  std::vector<char> buffer;
+  buffer.reserve(buffer_size);
```
edgchen1: [`vector::reserve()`](https://en.cppreference.com/w/cpp/container/vector/reserve.html) does not actually change the vector size. writing past the end of a vector is undefined behavior. as the vector will be filled later, you probably want to use [`vector::resize()`](https://en.cppreference.com/w/cpp/container/vector/resize).

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -825,6 +956,45 @@ void QnnBackendManager::ProcessContextFromBinListAsync(Qnn_ContextHandle_t conte
...
+
+  std::ifstream cache_file(context_bin_filepath.c_str(), std::ifstream::binary);
+  ORT_RETURN_IF(!cache_file || !cache_file.good(), "Failed to read context binary from: ", context_bin_filepath);
+
+  const auto& read_result = cache_file.read(buffer.data(), buffer_size);
+  ORT_RETURN_IF(!read_result, "Failed to read contents from cached context file.");
+
+  buffer_info.data = std::move(buffer);
+  buffer_info.size = buffer_size;
```
edgchen1: `BufferInfo::data` is a vector which has a size. can we just replace `BufferInfo` with `std::vector`?

PR #26966, opened 2026-01-10
OpenVINO: replace std::format with string concatenation for C++20 com…

### LINKED ISSUE #26936: [Build] OpenVINO Provider use C++20 feature that unavailable in supported linux distro
### Describe the issue

OpenVINO Provider use ``std::format`` that implemented in GCC 13. For Example supported Debian 12 distro ships with GCC 12, CentOS Stream 9 ships with GCC 11.2.  In result there are build failure.

Is minimum GCC version 13 a planned change?

### Urgency

_No response_

### Target platform

Linux/Debian 12

### Build script

````
./build.sh --config RelWithDebInfo  --use_openvino --build_shared_lib
````

### Error / output

```
/~/onnxruntime/onnxruntime/core/providers/openvino/ov_factory.cc:10:10: fatal error: format: No such file or directory
   10 | #include <format>
      |          ^~~~~~~~
```

### Visual Studio Version

_No response_

### GCC / Compiler Version

11.4.0

### PR DESCRIPTION
**Pull request by ingyukoh**
…patibility (#26936)

### Description
Fixes GCC build failure on Linux distributions where GCC < 13 is used (e.g., Ubuntu 22.04, Debian 12, CentOS Stream 9) by removing the usage of C++20 `std::format`.

### Motivation and Context
Resolves #26936.
The OpenVINO provider recently introduced a dependency on `<format>` which makes the build fail on systems with older GCC versions (missing `std::format` support).

### DIFF
--- a/onnxruntime/core/providers/openvino/backend_manager.cc
+++ b/onnxruntime/core/providers/openvino/backend_manager.cc
@@ -4,6 +4,7 @@
 #include <array>
 #include <algorithm>
 #include <cassert>
+#include <cstdlib>
 #include <fstream>
 #include <regex>
 #include <sstream>
@@ -434,8 +435,9 @@ static void ReadExternalDataFields(const ONNX_NAMESPACE::TensorProto* src_init, 
     if (pb_key == "location") {
       location = pb_value;
     } else if (pb_key == "offset") {
-      const auto res = std::from_chars(pb_value.data(), pb_value.data() + pb_value.size(), offset);
-      if (res.ec != std::errc()) {
+      char* end_ptr = nullptr;
+      offset = std::strtoull(pb_value.c_str(), &end_ptr, 10);
+      if (end_ptr == pb_value.c_str() || (end_ptr != nullptr && *end_ptr != '\0')) {
         std::ostringstream err_msg;
         err_msg << "External data in memory has invalid offset field: "
                 << src_init->name() << "], location: " << location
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by egortech** [comment]
There are also build error around: 

```
/home/egortech/onnxruntime/onnxruntime/core/providers/openvino/qdq_transformations/qdq_scales_fix.cpp:540:74:   required from here
/opt/x86_64-linux-gnu/x86_64-linux-gnu/include/c++/11.4.0/ranges:3451:5: error: template constraint failure for 'template<class _Range>  requires  viewable_range<_Range> using all_t = decltype (std::ranges::views::all(declval<_Range>()))'
/opt/x86_64-linux-gnu/x86_64-linux-gnu/include/c++/11.4.0/ranges:3451:5: note: constraints not satisfied
```

I have local fixed that polyfills std::format with fmt library. I believe this is better fixes)

ingyukoh:

ingyukoh:
@egortech After deeper investigation, I found the scope is larger than initially expected:

  Remaining C++20 issues:
  - qdq_scales_fix.cc: std::ranges (2 usages), std::format (1 usage)
  - ov_interface.cc: std::format_string template, std::format
  - ov_factory.cc: #include <ranges>

  My PR addresses:
  - ov_factory.cc (std::format → string concat)
  - exceptions.h (std::from_chars → strtoul)
  - backend_manager.cc (std::from_chars → strtoull)

  The std::format_string<Args...> in ov_interface.cc is complex to replace without fmt library.

  I think combining efforts would be most effective:
  - My std::from_chars fixes are still needed
  - Your fmt approach is better for the std::format/std::ranges issues

  Would maintainers prefer:
  1. Merge both PRs (mine for from_chars, yours for format/ranges)?
  2. I add my from_chars fixes to your PR #26970?

  Happy to collaborate either way.

egortech:
If OKs for maintainers to add dependencies on fmt library, it's is best options. Because it's not limited to use new library feature, and when C++20 in the future begin mandatory is use to migrate to std::format will be more convenient. Also there are options using it side by side, depending on compiler capability.

``std::from_chars`` problems is surprised me because in the theories GCC 11 in C+20 mode should support it

huber-mvtec:
+1 from my side for this PR.

I would really like to see this merged, as this not only breaks the build on said systems but also limits the (easy) usage on OSes that aren't new enough. E.g. Ubuntu 22.04 has EOL in 2027 (or even 2032 if you are on Ubuntu Pro/ESM), Debian 12 has EOL 2028.

Also I think the C++20 dependency is not in line with the Centos 7 compatibility stated e.g. here https://onnxruntime.ai/docs/reference/compatibility.html

PR #26971, opened 2026-01-12
Add APIs `OrtApi::CreateEnvWithOptions()` and `OrtEpApi::GetEnvConfigEntries()`

adrianlizarraga:
### Description
- Adds API `CreateEnvWithOptions` that allows the user to pass in a `OrtEnvCreationOptions` struct that aggregates all env creation options (log id, logging level, custom logger, thread pool options, etc.).
- The new `OrtEnvCreationOptions` struct also allows the application to provide an `OrtKeyValuePairs` containing general configuration entries.
- Adds EP API `GetEnvConfigEntries` that returns the environment's configuration entries as a `OrtKeyValuePairs` instance.
- Remove `OrtEpFactory::SetEnvironmentOptions` and update example EP to pull the necessary configuration entry via the new api `GetEnvConfigEntries`.

### DIFF
--- a/include/onnxruntime/core/session/onnxruntime_c_api.h
+++ b/include/onnxruntime/core/session/onnxruntime_c_api.h
@@ -6998,6 +6998,8 @@ struct OrtApi {
    */
   ORT_CLASS_RELEASE(DeviceEpIncompatibilityDetails);
 
+  /// @}
+
   /** \brief Create an OrtEnv instance with the given options.
    *
    * \note Invoking this function will return the same instance of the environment as that returned by a previous call


onnxruntime/test/autoep/library/plugin_ep_utils.h
```diff
@@ -103,6 +103,15 @@ struct FloatInitializer {
...
 };
 
+// Returns a lower case version of the input string.
+inline std::string GetLowercaseString(std::string str) {
+  // https://en.cppreference.com/w/cpp/string/byte/tolower
+  std::transform(str.begin(), str.end(), str.begin(), [](unsigned char c) {
+    return static_cast<char>(std::tolower(c));
+  });
+  return str;
+}
```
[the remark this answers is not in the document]
adrianlizarraga: plugin_ep_utils.h is not supposed to depend on ORT utils

onnxruntime/test/autoep/test_registration.cc
```diff
@@ -166,6 +169,49 @@ TEST(OrtEpLibrary, LoadUnloadPluginVirtGpuLibraryCxxApi) {
 
     ort_env->UnregisterExecutionProviderLibrary(registration_name_for_virtual_devices.c_str());
   }
+
+#if !defined(TEST_MAIN_ENABLE_DYNAMIC_PLUGIN_EP_USAGE)
```
edgchen1: what is this ifdef supposed to enable/disable?

https://github.com/microsoft/onnxruntime/blob/de398c41259ca8a6c68759d3ec81ec54d5e9fa3e/onnxruntime/test/unittest_main/test_main.cc#L34-L36

currently, `TEST_MAIN_ENABLE_DYNAMIC_PLUGIN_EP_USAGE` is only defined for onnxruntime_provider_test in a non-minimal build and it is only defined in test_main.cc. the tests in this file would be in onnxruntime_autoep_test.

adrianlizarraga: Aren't most of these unit tests sharing the same test_main.cc? This specific test calls `ortenv_teardown()` and `ortenv_setup()`, which internally modify the dynamic plugin ep state. I added this guard to avoid running this test if someone enables this feature since I don't know what the effect of repeated teardown and setup would be.

Maybe I'm worrying about something that would never happen.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -1045,6 +1037,96 @@ typedef enum OrtCompiledModelCompatibility {
...
+  /** \brief Optional environment configuration entries represented as string key-value pairs. May be set to NULL.
+   *
+   * \note The OrtKeyValuePairs instance is copied by ORT.
+   *
+   * \note Refer to onnxruntime_env_config_keys.h for common config entry keys and their supported values.
+   *
+   * \note An application may provide configuration options for execution provider libraries by using keys with the
+   *       prefix 'ep.<ep_name>.'. For example, the key 'ep.my_ep.some_ep_key' represents a key named 'some_ep_key'
+   *       that is meant to be consumed by an execution provider named 'my_ep'. Refer to the specific execution
+   *       provider's documentation for valid keys and values.
```
edgchen1: maybe explain the difference between env-level EP-specific options and EP-specific options that can be added to session options. are they distinct? if an option appears in both contexts, which takes precedence?

include/onnxruntime/core/session/onnxruntime_cxx_inline.h
```diff
@@ -784,6 +784,15 @@ inline Env::Env(const OrtThreadingOptions* tp_options, OrtLoggingFunction loggin
   }
 }
 
+inline Env::Env(const OrtEnvCreationOptions* options) {
+  ThrowOnError(GetApi().CreateEnvWithOptions(options, &p_));
+  if (strcmp(options->log_id, "onnxruntime-node") == 0) {
+    ThrowOnError(GetApi().SetLanguageProjection(p_, OrtLanguageProjection::ORT_PROJECTION_NODEJS));
```
edgchen1, line 790: if the env has already been created, the env won't be updated, but this could have the side effect of updating the language projection.

looks like this is an existing issue/"feature" in the other ctors though. could potentially address it by having the new API indicate whether the env was actually created.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -1408,6 +1408,22 @@ struct OrtEpApi {
...
+  /** \brief Gets a new OrtKeyValuePairs instance containing all configuration entries set on the environment.
+   *
+   * \note An application may provide configuration options for EP libraries by using keys with the
+   *       prefix 'ep.<ep_name>.'. For example, the key 'ep.my_ep.some_ep_key' represents a key named 'some_ep_key'
+   *       that is meant to be consumed by an EP named 'my_ep'.
+   *
+   * \note Refer to onnxruntime_env_config_keys.h for common configuration entry keys and their supported values.
+   *
+   * \param[out] out Output parameter set to the OrtKeyValuePairs instance containing all configuration entries.
+   *                 Must be released via OrtApi::ReleaseKeyValuePairs.
```
edgchen1, line 1443: as I understand it, the config entries are const (specified in at env creation). should we return a const pointer to it instead of a copy that needs to be released?

include/onnxruntime/core/session/onnxruntime_cxx_inline.h
adrianlizarraga: I suppose this would be potentially encountered in a program that mixed usage of C and C++ apis. For example, if the user used the C API to create env_1, the projection would be set to C (the default). Then, if the user used this C++ API to create env_2, then the projection would change to CPLUSPLUS.

There's a case to be made that this could be the desired outcome since the user may continue to use the C++ API in the future and the purpose of capturing the language projection is for telemetry.

Also, the projection information is not actually stored in the OrtEnv; it is stored in the platform env (different). I think we just put the projection stuff here for convenience even though it doesn't actually modify the OrtEnv state.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
adrianlizarraga: The env does actually modify its entries when a plugin EP library is registered and unregistered. Specifically, it adds/removes an entry if the plugin EP library's registration name ends in ".virtual".

Also, given that OrtEnv may be accessed by multiple threads, it feels safer to just give the EP a copy/snapshot of the config entries. What's your take?

onnxruntime/test/autoep/test_registration.cc
edgchen1: I see. the macro to check would be `ORT_UNIT_TEST_ENABLE_DYNAMIC_PLUGIN_EP_USAGE`. as is, it will not be enabled for this test because it is only enabled in onnxruntime_provider_test [here](https://github.com/microsoft/onnxruntime/blob/912f652321bae5d3ed4c5eae3aea3ed28d6c14fc/cmake/onnxruntime_unittests.cmake#L1251).

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
adrianlizarraga: I pushed a commit that added a mutex to guard access to the env's config entries map. I'm still returning a copy from this this API for now.

include/onnxruntime/core/session/onnxruntime_c_api.h
skottmckay: I would avoid trying to merge env level options with EP level options as that's going to be super messy. 

Could we use `ep_factory` as the prefix to differentiate?

include/onnxruntime/core/session/onnxruntime_env_config_keys.h
```diff
@@ -0,0 +1,23 @@
...
+// The full key has the form: "allow_virtual_devices.<EP_LIBRARY_REGISTRATION_NAME>" all in lower case.
+//
+// Note: A virtual OrtHardwareDevice does not represent actual hardware on the device, and is identified via the
+// metadata entry "is_virtual" with a value of "1".
+//
+// Allowed values:
+//  - "0": Default. Creation of virtual devices is not allowed.
+//         This is the assumed default value if this key is not present in the environment's configuration entries.
+//  - "1": Creation of virtual devices is allowed.
+static const char* const kOrtEnv_AllowVirtualDevicesPrefix = "allow_virtual_devices.";
```
skottmckay:
```suggestion
static const char* const kOrtEnv_AllowVirtualDevicesPrefix = "allow_virtual_devices";
``` #Closed

File: onnxruntime/test/autoep/test_registration.cc (hunk 2 above)
Comment (adrianlizarraga) [reply onnxruntime/test/autoep/test_registration.cc]: Ah I see. I was using the wrong macro. In any case, I removed the macro check since, as you said, it is only enabled in onnxruntime_provider_test.

File: include/onnxruntime/core/session/onnxruntime_ep_c_api.h (hunk 5 above)
Comment (edgchen1) [reply include/onnxruntime/core/session/onnxruntime_ep_c_api.h:1443]: oh, I see. sorry, I missed that it gets modified.

I do think it would be simpler if we could make the config entries immutable. if that's not possible, then returning a copy is safer.

File: include/onnxruntime/core/session/onnxruntime_c_api.h (hunk 3 above)
Comment (adrianlizarraga) [reply include/onnxruntime/core/session/onnxruntime_c_api.h]: Perhaps we have two kinds of env-level prefixes for EPs:
- `ep_lib.<registration_name>.`: configuration for an EP library. We need to be able to configure an ep library before there are any factories created (i.e., see the `allow_virtual_devices` use case in this PR).
- `ep_factory.<ep_name>.`: This is for configuring a specific EP factory. Note that an EP factory and its EP instances have the same name.

Separately, we have one session-level prefix provided in session options.
- `ep.<ep_name>.`

File: include/onnxruntime/core/session/onnxruntime_ep_c_api.h (hunk 5 above)
Comment (skottmckay) [reply include/onnxruntime/core/session/onnxruntime_ep_c_api.h:1443]: If there's a copy and the EP saves the returned `OrtKeyValuePairs*` they won't see the latest values if they change in the OrtEnv later on. Might also be confusing if they set a value in the returned OrtKeyValuePairs given it's non-const as that won't affect the env.
But a copy protects against an EP reading a value when there's a concurrent update to OrtKeyValuePairs that causes the key/value data structure to be reallocated at the exact same time. Odds of that are probably extremely small.

Even though it's slightly less safe, not taking a copy might give more predictable behavior. But it's a 50/50 to me as to which is better.

File: onnxruntime/core/session/ort_env.h [hunk 7]
```diff
@@ -48,6 +65,9 @@ struct OrtEnv {
   onnxruntime::logging::LoggingManager* GetLoggingManager() const;
   void SetLoggingManager(std::unique_ptr<onnxruntime::logging::LoggingManager> logging_manager);
 
+  // Returns a copy of the environment's config entries.
+  OrtKeyValuePairs GetConfigEntries() const;
```
skottmckay: Does this need to be at this level or can it come from ort_env.GetEnvironment().GetConfigEntries()?  #Closed

onnxruntime/core/session/environment.cc
```diff
@@ -576,27 +607,47 @@ Status Environment::RegisterExecutionProviderLibrary(const std::string& registra
   std::vector<EpFactoryInternal*> internal_factories = {};
   std::unique_ptr<EpLibrary> ep_library;
 
+  // An application can allow an EP library to create virtual devices by using an EP library registration name that
+  // ends in the suffix ".virtual". If so, ORT automatically sets the config key
+  // "allow_virtual_devices.<EP_LIB_REGISTRATION_NAME>" to "1" in the environment.
```
skottmckay: What is the current implementation logic when virtual devices are enabled via the registration name? Does the registration with ".virtual" only create virtual devices (so you need one non-virtual registration and one virtual), or it creates both? If the former this potentially makes things confusing. If the latter, the ".virtual" is potentially confusing.

Depending on the answer having a key defined in onnxruntime_env_config_keys.h may or may not be confusing. :P #Closed

onnxruntime/core/session/environment.cc
```diff
@@ -242,11 +245,16 @@ Status Environment::CreateAndRegisterAllocator(const OrtMemoryInfo& mem_info, co
...
                                const OrtThreadingOptions* tp_options,
-                               bool create_global_thread_pools) {
+                               bool create_global_thread_pools,
+                               const OrtKeyValuePairs* config_entries) {
   auto status = Status::OK();
 
   logging_manager_ = std::move(logging_manager);
 
+  if (config_entries != nullptr) {
+    config_entries_.CopyFromMap(config_entries->Entries());
```
edgchen1: nit: can we just copy it?
```suggestion
    config_entries_ = *config_entries;
```

onnxruntime/test/autoep/test_registration.cc
```diff
@@ -166,6 +169,46 @@ TEST(OrtEpLibrary, LoadUnloadPluginVirtGpuLibraryCxxApi) {
 
     ort_env->UnregisterExecutionProviderLibrary(registration_name_for_virtual_devices.c_str());
   }
+
+  // Test using OrtApi::CreateEnvWithOptions to explicitly set a config that enables virtual devices.
+  // The EP should return a OrtEpDevice for a virtual GPU.
+
+  ortenv_teardown();  // Release current OrtEnv as we need to recreate it.
```
skottmckay, line 176: nit: In theory tests should be able to run in parallel and this could break another test. In practice I don't think they do currently and I don't have a better option for testing this unless we split out the implementation of OrtEnv into something like OrtEnvImpl with OrtEnv just becoming a holder for an instance of that (and the test being against OrtEnvImpl). Doesn't seem worth the time and effort at this point though.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
adrianlizarraga: >If there's a copy and the EP saves the returned OrtKeyValuePairs* they won't see the **latest values if they change** in the OrtEnv later on

I prefer the copy for the following reason. If the env configs _can_ change, then that change could reasonably happen on another thread because the OrtEnv is global. 

In this case, having a const reference to the OrtKeyValuePairs could give way to race conditions even now. This would become more likely if in the future someone adds an API to add/remove configs outside of env creation. I just have the feeling someone is going to want to do that at some point :). 

The only way to make this safe is to guarantee that the env configs are immutable after creation. But that doesn't support the "virtual" device feature needed for cross-compilation. It may also limit future use of the configs.

onnxruntime/core/session/ort_env.h
```diff
@@ -31,9 +31,26 @@ struct OrtEnv {
...
+                                     const OrtThreadingOptions* tp_options = nullptr,
+                                     const OrtKeyValuePairs* config_entries = nullptr);
+
+  using UniquePtr = std::unique_ptr<OrtEnv, void (*)(OrtEnv*)>;
+
+  /// <summary>
+  /// Gets the global OrtEnv instance. Returns nullptr if the instance has not yet been created.
+  /// </summary>
+  /// <returns>The OrtEnv instance or nullptr.</returns>
+  static UniquePtr TryGetInstance();
```
edgchen1: nit: it seems inconsistent that `GetOrCreateInstance()` returns a raw pointer and `TryGetInstance()` returns a `UniquePtr`. also, the name "unique ptr" has the conventional meaning of "single owner", which doesn't quite fit for this reference counted instance.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
adrianlizarraga: A way to potentially mitigate the downsides of copying is to introduce a similar API function (e.g., `GetEnvConfigEntry()`) that retrieves an individual config entry. If an EP knows the specific key it wants, it can call that API on demand and always get the latest value.

If instead, the ep wants all key/value pairs at once, then they have to be ok with getting a full copy.

What do you think?

skottmckay: I'd vote for either but not both.

Maybe we could just be more explicit about it being a copy in the doco (e.g. 'containing a copy of all configuration entries'). As it's an opaque type the user might not know what internally a new instance of OrtKeyValuePairs equates to.

onnxruntime/test/autoep/test_registration.cc
adrianlizarraga: I see. I was surprised to see that we didn't have any unit tests to test all the current variations of `CreateEnvWith*`. This may be one of the reasons why. Hmmm, yeah I may have to punt on this for now.

orttraining/orttraining/python/orttraining_python_module.cc
```diff
@@ -173,7 +173,8 @@ static Status CreateOrtEnv() {
   Env::Default().GetTelemetryProvider().SetLanguageProjection(OrtLanguageProjection::ORT_PROJECTION_PYTHON);
   OrtEnv::LoggingManagerConstructionInfo lm_info{nullptr, nullptr, ORT_LOGGING_LEVEL_WARNING, "Default"};
   Status status;
-  std::unique_ptr<OrtEnv> ort_env(OrtEnv::GetInstance(lm_info, status, use_global_tp ? &global_tp_options : nullptr));
+  std::unique_ptr<OrtEnv> ort_env(OrtEnv::GetOrCreateInstance(lm_info, status,
+                                                              use_global_tp ? &global_tp_options : nullptr));
```
edgchen1: not related to this PR, but is this correct? the pointer should be freed with `OrtEnv::Release` instead, right?

onnxruntime/core/session/ort_env.h
adrianlizarraga: Yeah, it is not super consistent. `GetOrCreateInstance()` (formerly GetInstance) returns a raw pointer because it relies on the user calling the corresponding `ReleaseEnv` api, which calls into `OrtEnv::Release()`.

On the other hand, `TryGetInstance()` is used internally within ORT and it's nice to have a way to automatically call `OrtEnv::Release()` via `std::unique_ptr`. Would renaming `OrtEnv::UniquePtr` to `OrtEnvPtr` be a bit more understandable?

Edit; or maybe have both functions return `OrtEnvPtr` consistently.

orttraining/orttraining/python/orttraining_python_module.cc
adrianlizarraga: Yeah, that seems wrong. We probably don't have many (or any) orttraining CIs that check memleaks.

onnxruntime/core/session/ort_env.h
edgchen1: sure, I like `OrtEnvPtr`. and consistently returning it would be nice too :)

onnxruntime/core/session/environment.cc
adrianlizarraga: I updated this key to just be "allow_virtual_devices". It is automatically set to "1" if any EP library is registered with "*.virtual". In practice, we expect only one such library to be registered and it would only create virtual devices for cross-compilation. However, this is not enforced.

onnxruntime/core/session/environment.cc
adrianlizarraga: updated

onnxruntime/core/session/ort_env.h
adrianlizarraga: updated.

orttraining/orttraining/python/orttraining_python_module.cc
adrianlizarraga: updated.

onnxruntime/core/session/ort_env.h
adrianlizarraga: removed.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
adrianlizarraga: I left `GetEnvConfigEntries()` as is. I updated the docs to more explicitly state that it returns a copy of all entries.

include/onnxruntime/core/session/onnxruntime_c_api.h
adrianlizarraga: I updated this to `ep_factory.`.

onnxruntime/core/session/environment.cc
```diff
@@ -474,6 +482,21 @@ Status Environment::GetSharedAllocator(const OrtMemoryInfo& mem_info, OrtAllocat
   return Status::OK();
 }
 
+OrtKeyValuePairs Environment::GetConfigEntries() const {
+  std::lock_guard<std::mutex> lock{config_entries_mutex_};
```
edgchen1: could we use a RW lock (std::shared_mutex)? that would allow multiple readers at a time.

onnxruntime/core/session/ort_env.h
```diff
@@ -14,6 +14,9 @@ namespace onnxruntime {
 class Environment;
 }
 
+// Managed pointer type for OrtEnv that calls OrtEnv::Release to release memory.
```
edgchen1: nit: it's not exactly the same as releasing memory
```suggestion
// Managed pointer type for OrtEnv that calls OrtEnv::Release as its deleter.
```

include/onnxruntime/core/session/environment.h
```diff
@@ -243,6 +265,15 @@ class Environment {
...
 #endif  // !defined(ORT_MINIMAL_BUILD)
+
+  // Application-specified environment configuration entries
+  // The environment may add or remove an entry on EP library registration and unregistration, respectively.
+  OrtKeyValuePairs config_entries_;
+  mutable std::mutex config_entries_mutex_;  // Should be locked when accessing config_entries_
+
+  // Tracks the number of registered EP libraries that can create virtual devices.
+  // Only modified during library registration/unregistration.
+  int32_t num_virtual_ep_libraries_{};
```
edgchen1: if it shouldn't be negative, how about `size_t` instead?

onnxruntime/test/shared_lib/test_env_creation.cc
```diff
@@ -0,0 +1,90 @@
...
+    options.logging_severity_level = OrtLoggingLevel::ORT_LOGGING_LEVEL_VERBOSE;
+    options.log_id = "EnvCreation.CreateEnvWithOptions_2";
+    options.config_entries = env_configs.GetConst();
+
+    Ort::Env tmp_env(&options);
+
+    // Use EP API to retrieve environment configs and check contents
+    OrtKeyValuePairs* c_kvps = nullptr;
+    ASSERT_ORTSTATUS_OK(Ort::GetEpApi().GetEnvConfigEntries(&c_kvps));
+    Ort::KeyValuePairs env_configs_2(c_kvps);
```
edgchen1: should we have a C++ API that directly returns an `Ort::KeyValuePairs` instance for the env config entries?

include/onnxruntime/core/session/onnxruntime_cxx_inline.h
edgchen1: I suppose it's fine to use the same approach for now

orttraining/orttraining/python/orttraining_python_module.cc
edgchen1: thanks for fixing it

onnxruntime/core/session/environment.cc
adrianlizarraga: good call. updated

include/onnxruntime/core/session/environment.h
adrianlizarraga: yep updated

onnxruntime/test/shared_lib/test_env_creation.cc
adrianlizarraga: added

onnxruntime/core/session/environment.cc
```diff
@@ -576,27 +607,51 @@ Status Environment::RegisterExecutionProviderLibrary(const std::string& registra
...
   ORT_TRY {
-    auto ep_info = std::move(ep_libraries_[ep_name]);
+    auto ep_info = std::move(ep_libraries_[registration_name]);
+
+    // Clean up environment config entry that may have been added to enable virtual devices.
+    if (AreVirtualDevicesAllowed(registration_name)) {
+      num_virtual_ep_libraries_ -= 1;
+
+      if (num_virtual_ep_libraries_ == 0) {
+        RemoveConfigEntry(kOrtEnv_AllowVirtualDevices);
```
skottmckay: nit: if the user sets the config value when constructing the OrtEnv AND registered a factory with ".virtual" this would undo the initial value. I don't think it matters much and about the only reasonable thing I can think of is mentioning in the doco for kOrtEnv_AllowVirtualDevices that if set manually they shouldn't use the ".virtual" path. Optional to do that.

include/onnxruntime/core/session/onnxruntime_env_config_keys.h
```diff
@@ -0,0 +1,24 @@
...
+// name that ends in the suffix ".virtual". See OrtApi::RegisterExecutionProviderLibrary().
+//
+// Note: A virtual OrtHardwareDevice does not represent actual hardware on the device, and is identified via the
+// metadata entry "is_virtual" with a value of "1".
+//
+// Allowed values:
+//  - "0": Default. Creation of virtual devices is not allowed.
+//         This is the assumed default value if this key is not present in the environment's configuration entries.
+//  - "1": Creation of virtual devices is allowed.
+static const char* const kOrtEnv_AllowVirtualDevices = "allow_virtual_devices";
```
skottmckay: nit: the session and run config keys don't have a '_' in the middle of the option names so this is slightly inconsistent.

skottmckay (dismissed):

onnxruntime/core/session/environment.cc
```diff
@@ -637,12 +640,13 @@ Status Environment::UnregisterExecutionProviderLibrary(const std::string& regist
...
-    // Clean up environment config entry that may have been added for this EP library.
+    // Clean up environment config entry that may have been added to enable virtual devices.
     if (AreVirtualDevicesAllowed(registration_name)) {
-      std::string key = kOrtEnv_AllowVirtualDevicesPrefix;
-      key += utils::GetLowercaseString(registration_name);
+      num_virtual_ep_libraries_ -= 1;
 
-      RemoveConfigEntry(key);
+      if (num_virtual_ep_libraries_ == 0) {
+        RemoveConfigEntry(kOrtEnv_AllowVirtualDevices);
```
edgchen1: do we want to allow the user to explicitly specify `kOrtEnv_AllowVirtualDevices` in the env config? if so, I think this would remove it even in that case, which may be unexpected. perhaps we could use something like a reference count for "allow virtual device usages" in place of `num_virtual_ep_libraries_` and initialize it to 1 if `kOrtEnv_AllowVirtualDevices` is present in the env config at initialization time.

edgchen1 (dismissed):

include/onnxruntime/core/session/onnxruntime_env_config_keys.h
adrianlizarraga: updated.

onnxruntime/core/session/environment.cc
adrianlizarraga: Updated as per edgchen1's recommendation (i.e., count starts at 1 in this case to prevent removal of the key).

onnxruntime/core/session/environment.cc
adrianlizarraga: updated. thanks

skottmckay (approved):

edgchen1 (dismissed):

edgchen1 (approved):

PR #26982, opened 2026-01-12
Enhance stale issue management with new label checks

justinchuby:
- Updated stale issue policy to include 'no stale' label checks and modified reply message.
- Do not reopen closed stale issues automatically, as it is just noisy. Instead, we could encourage issue owners to reopen manually or create new issues.


justinchuby:
@hariharans29

hariharans29 (approved):

PR #26984, opened 2026-01-12
remove lint for /js/ folder

prathikr:
### Description

`js/` linting is already done in the Web CI pipeline so removing redundant check as its signaling false-positives


guschmue (changes requested):
lots of code left in /js - we don't want to remove lint for /js

fs-eire (approved):

PR #26988, opened 2026-01-13
[CXX] Enable users to specify custom OrtSyncStream via RunOptions

yuslepukhin:
### Description
Enable device stream override using RunOptions for a particular run.
The stream is restored after Run() completes.

### Motivation and Context
Gpu InterOp requirements.
When enabled, the inference would run with the specified stream with proper synchronization with imported external synchronization facilities.

### DIFF
--- a/include/onnxruntime/core/session/onnxruntime_c_api.h
+++ b/include/onnxruntime/core/session/onnxruntime_c_api.h
@@ -7113,8 +7113,8 @@ struct OrtApi {
    * \since Version 1.24.
    */
   ORT_API2_STATUS(EpAssignedNode_GetOperatorType, _In_ const OrtEpAssignedNode* ep_node, _Outptr_ const char** out);
-  
-/** \brief Sets OrtSyncStream for the run options
+
+  /** \brief Sets OrtSyncStream for the run options
    *
    * OrtSyncStream is used to synchronize the execution of the model run for the device
    * of the stream. It overrides the existing stream for the duration of the Run().
@@ -7126,7 +7126,6 @@ struct OrtApi {
    * \since 1.24
    */
   ORT_API_T(void, RunOptionsSetSyncStream, _Inout_ OrtRunOptions* options, _In_ OrtSyncStream* sync_stream);
- 
 };
 
 /*
--- a/onnxruntime/test/shared_lib/test_inference.cc
+++ b/onnxruntime/test/shared_lib/test_inference.cc
@@ -4921,8 +4921,6 @@ TEST(CApiTest, TestSyncStreamOverride) {
 #endif
 #endif
 
-
-
 #if !defined(ORT_MINIMAL_BUILD)
 TEST(CApiTest, GetEpGraphAssignmentInfo_NotEnabledError) {
   // Test that calling OrtApi::Session_GetEpGraphAssignmentInfo() without enabling the appropriate
@@ -4944,4 +4942,4 @@ TEST(CApiTest, GetEpGraphAssignmentInfo_NotEnabledError) {
     ASSERT_THAT(ex.what(), testing::HasSubstr(oss.str()));
   }
 }
-
+#endif


onnxruntime/core/session/inference_session.cc
```diff
@@ -3050,6 +3051,14 @@ Status InferenceSession::Run(const RunOptions& run_options,
 
 #ifdef ORT_ENABLE_STREAM
       DeviceStreamCollectionHolder device_stream_collection_holder(session_state_.get());
+      if (run_options.sync_stream != nullptr) {
+        if (session_options_.execution_mode != ExecutionMode::ORT_SEQUENTIAL) {
+          LOGS(*session_logger_, WARNING) << "Setting sync stream is not supported in parallel execution mode.";
+        } else {
+          ORT_RETURN_IF_ERROR_SESSIONID_(
+              device_stream_collection_holder.p_->SetStreamOverride(run_options.sync_stream));
+        }
```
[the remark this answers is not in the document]
yuslepukhin: False alaram.

include/onnxruntime/core/session/onnxruntime_cxx_api.h
```diff
@@ -1304,6 +1304,15 @@ struct RunOptions : detail::Base<OrtRunOptions> {
    * \param adapter The LoraAdapter to be used as the active adapter
    */
   RunOptions& AddActiveLoraAdapter(const LoraAdapter& adapter);
+
+  /** \brief Associate a sync stream with the run options.
+   *
+   * When set, the EP uses this stream for execution, enabling proper
+   * synchronization with imported external semaphores. Wraps OrtApi::RunOptionsSetSyncStream.
+   *
+   * \param stream The OrtSyncStream to associate with these run options. May be nullptr to clear.
```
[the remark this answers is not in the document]
yuslepukhin: False alarm

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -6785,6 +6785,21 @@ struct OrtApi {
...
                   _In_ size_t num_outputs);
 
+  /** \brief Sets OrtSyncStream for the run options
+   *
+   * OrtSyncStream is used to synchronize the execution of the model run for the device
+   * of the stream. It overrides the existing stream for the duration of the Run().
+   * The stream instance must be alive for the duration of the Run() call.
+   *
+   * \param[in] options
+   * \param[in] sync_stream The synchronization stream.
```
skottmckay: nit: Should this also say pass nullptr to clear? #Closed

onnxruntime/core/framework/device_stream_collection.cc
```diff
@@ -76,11 +78,39 @@ class DeviceStreamCollectionImpl {
...
     device_streams_[idx] = stream;
   }
 
+  Status SetStreamOverride(Stream* stream) {
+    ORT_ENFORCE(stream != nullptr);
+    for (size_t i = 0, lim = device_streams_.size(); i < lim; ++i) {
+      if (device_streams_[i] != nullptr &&
+          // Exact match
+          device_streams_[i]->GetDevice() == stream->GetDevice()) {
+        stream_override_.emplace(i, stream);
```
skottmckay, line 100: why not replace device_streams_[i] (and remove from owned_streams if applicable)?  #Closed

onnxruntime/core/session/inference_session.cc
```diff
@@ -3050,6 +3051,14 @@ Status InferenceSession::Run(const RunOptions& run_options,
 
 #ifdef ORT_ENABLE_STREAM
       DeviceStreamCollectionHolder device_stream_collection_holder(session_state_.get());
+      if (run_options.sync_stream != nullptr) {
+        if (session_options_.execution_mode != ExecutionMode::ORT_SEQUENTIAL) {
+          LOGS(*session_logger_, WARNING) << "Setting sync stream is not supported in parallel execution mode.";
```
skottmckay, line 3100: nit: add a comment to the code explaining why so the next developer a) knows it's possible, and b) has an idea of the code changes required to support in the future if we want to.

onnxruntime/test/shared_lib/test_inference.cc
```diff
@@ -4827,3 +4827,54 @@ TEST(CApiTest, ModelWithExternalDataOutsideModelDirectoryShouldFailToLoad) {
               exception_message.find("model") != std::string::npos)
       << "Exception message should indicate external data or security issue. Got: " << exception_message;
 }
+
+#ifdef ORT_ENABLE_STREAM
+#if USE_CUDA
+TEST(CApiTest, TestSyncStreamOverride) {
```
skottmckay, line 4860: Can we leverage anything from onnxruntime\test\autoep\library\example_plugin_ep\ep_external_resource_importer.cc and onnxruntime\test\autoep\test_external_resource_importer.cc to test in more depth?

One simple thing to validate would be that the stream gets signaled when the inference completes.

onnxruntime/core/framework/device_stream_collection.cc
yuslepukhin: Because we need the original to cache after inference is complete.

skottmckay: Would be great to add a comment capturing that info where `stream_override_` is declared.

onnxruntime/test/shared_lib/test_inference.cc
skottmckay: Might be able to cut-and-paste a lot from NvExecutionProviderExternalResourceImporterTest, FullInferenceWithExternalMemoryCIG as a base. Doesn't need the CIG context side of things. Difference would be that instead of specifying the user stream using EP options (onnxruntime::nv::provider_option_names::kUserComputeStream) we would use the new code. 

[https://github.com/microsoft/onnxruntime/pull/26948](https://github.com/microsoft/onnxruntime/pull/26948)

yuslepukhin: Discussed offline.

onnxruntime/test/shared_lib/test_inference.cc
```diff
@@ -4827,3 +4829,93 @@ TEST(CApiTest, ModelWithExternalDataOutsideModelDirectoryShouldFailToLoad) {
...
+
+  // Create a stream on CUDA Device
+  const auto sync_stream = cuda_device.CreateSyncStream();
+  TestCudaStreamOverrideUsed cuda_override_stream(sync_stream);
+
+  // Create session with CUDA EP using C++ public API in Ort:: namespace
+  Ort::SessionOptions session_options;
+  Ort::CUDAProviderOptions cuda_options;
+  session_options.AppendExecutionProvider_CUDA_V2(*cuda_options);
+
```
skottmckay, line 4893: This is a bit confusing as a test setup. You're using the old-style EP selection (i.e. not AppendExecutionProvider_V2 with an OrtEpDevice), which by-passes everything to do with OrtSyncStream (which is 'v2' along with using OrtEpFactory to construct an OrtSyncStream etc. etc.), with the exception of getting it from RunOptions. 

Can we instead use the 'v2' world and register the CUDA EP (RegisterExecutionProviderLibrary), check the results from GetEpDevices() for a CUDA device, call CreateSyncStreamForEpDevice if found to create the stream, use the V2 registration to enable the CUDA EP, and go from there? I think that will make it much easier for the next developer to understand how everything fits together.

You can get the cudaStream from the OrtSyncStream using SyncStream_GetHandle in order to do the assert at the end.

yuslepukhin: Changed to `session_options.AppendExecutionProvider_V2`. I am not seeing why assert needs to be changed.

onnxruntime/test/shared_lib/test_inference.cc
```diff
@@ -4827,3 +4827,54 @@ TEST(CApiTest, ModelWithExternalDataOutsideModelDirectoryShouldFailToLoad) {
...
+      Ort::MemoryInfo::CreateCpu(OrtDeviceAllocator, OrtMemTypeCPU),
+      x_value, std::size(x_value), input_shape.data(), input_shape.size());
+  Ort::Value ort_inputs[] = {std::move(input_value)};
+
+  Ort::RunOptions run_options;
+  run_options.SetSyncStream(sync_stream);
+
+  auto output_values = session.Run(run_options,
+                                   input_names.data(), ort_inputs, std::size(ort_inputs),
+                                   output_names.data(), output_names.size());
```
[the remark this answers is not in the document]
yuslepukhin: This is now added in test_inference.cc

skottmckay (approved):
:shipit:

nieubank (approved):

PR #27004, opened 2026-01-14
Add CreateTensor with byte offset support for WebGPU sub-buffer views

qjia7:
## Problem

The existing `CreateTensorWithDataAsOrtValue` API supports sub-buffer views and zero-copy transfers by accepting a pre-advanced pointer — the caller performs `(char*)p_data + byte_offset` before passing it in. This works for CPU and CUDA, where `p_data` is a real, addressable memory pointer.

However, for backends such as WebGPU, `p_data` is an **opaque, non-addressable buffer handle** (`WGPUBuffer`). Applying pointer arithmetic to such a handle produces a corrupted value that is no longer a valid buffer handle, making it impossible to express sub-buffer tensor views or perform zero-copy transfers over regions of a larger GPU allocation.

## Solution

This PR introduces a new C API entry point:

```c
OrtApi::CreateTensorWithDataAsOrtValueWithByteOffset(
    const OrtMemoryInfo* info,
    void* p_data,
    size_t p_data_byte_count,
    size_t p_data_byte_offset,
    const int64_t* shape, size_t shape_len,
    ONNXTensorElementDataType type,
    OrtValue** out);
```
Instead of advancing the pointer before creating the tensor, the byte offset is stored directly in the tensor's internal byte_offset_ field. For addressable backends (CPU, CUDA), DataRaw() / [MutableData<T>()](vscode-file://vscode-app/c:/Users/jiajiaqin/AppData/Local/Programs/Microsoft%20VS%20Code/072586267e/resources/app/out/vs/code/electron-browser/workbench/workbench.html) already apply the offset via [(char*)p_data_ + byte_offset_](vscode-file://vscode-app/c:/Users/jiajiaqin/AppData/Local/Programs/Microsoft%20VS%20Code/072586267e/resources/app/out/vs/code/electron-browser/workbench/workbench.html), so behavior is identical to the old approach. For non-addressable backends (WebGPU), the raw handle remains intact in p_data_, and the WebGPU DataTransfer and kernel dispatch code recover the correct buffer region at runtime using the stored ByteOffset().

### DIFF
--- a/include/onnxruntime/core/session/onnxruntime_c_api.h
+++ b/include/onnxruntime/core/session/onnxruntime_c_api.h
@@ -7243,33 +7243,6 @@ struct OrtApi {
    */
   ORT_API2_STATUS(RunOptionsDisableProfiling, _Inout_ OrtRunOptions* options);
 
-  /** \brief Copy OrtValue instances containing Tensors between devices with offset and size control.
-   *
-   * Extended version of CopyTensors that supports copying with source/destination offsets and custom sizes.
-   * All offsets and sizes are in bytes.
-   *
-   * \param[in] env The OrtEnv instance to use.
-   * \param[in] src_tensors Array of OrtValue instances containing the source tensors to copy.
-   * \param[in] dst_tensors Array of OrtValue instances to copy the source tensors to.
-   * \param[in] source_offsets Optional array of source offsets in bytes. May be nullptr for all zeros.
-   * \param[in] destination_offsets Optional array of destination offsets in bytes. May be nullptr for all zeros.
-   * \param[in] sizes Optional array of sizes in bytes to copy. May be nullptr to copy entire tensors.
-   * \param[in] stream Optional OrtSyncStream that can be used to perform the copy asynchronously. May be nullptr.
-   * \param[in] num_tensors The number of tensors to copy.
-   *
-   * \snippet{doc} snippets.dox OrtStatus Return Value
-   *
-   * \since Version 1.25
-   */
-  ORT_API2_STATUS(CopyTensorsEx, _In_ const OrtEnv* env,
-                  _In_reads_(num_tensors) const OrtValue* const* src_tensors,
-                  _In_reads_(num_tensors) OrtValue* const* dst_tensors,
-                  _In_reads_opt_(num_tensors) const size_t* source_offsets,
-                  _In_reads_opt_(num_tensors) const size_t* destination_offsets,
-                  _In_reads_opt_(num_tensors) const size_t* sizes,
-                  _In_opt_ OrtSyncStream* stream,
-                  _In_ size_t num_tensors);
-
   /** \brief Create a tensor backed by a user supplied buffer, starting at a byte offset into that buffer
    *
    * Equivalent to CreateTensorWithDataAsOrtValue but the tensor's data pointer is advanced by
--- a/include/onnxruntime/core/session/onnxruntime_ep_c_api.h
+++ b/include/onnxruntime/core/session/onnxruntime_ep_c_api.h
@@ -127,9 +127,6 @@ struct OrtDataTransferImpl {
    * \param[in] this_ptr Pointer to the OrtDataTransferImpl instance.
    * \param[in] src_tensors Array of source OrtValue pointers to copy from.
    * \param[in] dst_tensors Array of destination OrtValue pointers to copy to.
-   * \param[in] source_offsets Optional array of source offsets in bytes. May be nullptr for all zeros.
-   * \param[in] destination_offsets Optional array of destination offsets in bytes. May be nullptr for all zeros.
-   * \param[in] sizes Optional array of sizes in bytes to copy. May be nullptr to copy entire tensors.
    * \param[in] streams Array of OrtSyncStream pointers for the copy operations, if the execution provider is stream
    *                    aware. nullptr if it is not.
    * \param[in] num_tensors Number of tensors to copy.
@@ -141,9 +138,6 @@ struct OrtDataTransferImpl {
   ORT_API2_STATUS(CopyTensors, _In_ OrtDataTransferImpl* this_ptr,
                   _In_reads_(num_tensors) const OrtValue** src_tensors,
                   _In_reads_(num_tensors) OrtValue** dst_tensors,
-                  _In_reads_opt_(num_tensors) const size_t* source_offsets,
-                  _In_reads_opt_(num_tensors) const size_t* destination_offsets,
-                  _In_reads_opt_(num_tensors) const size_t* sizes,
                   _In_reads_(num_tensors) OrtSyncStream** streams,
                   _In_ size_t num_tensors);
 };
--- a/onnxruntime/core/framework/data_transfer.cc
+++ b/onnxruntime/core/framework/data_transfer.cc
@@ -21,30 +21,14 @@ common::Status IDataTransfer::CopyTensor(const Tensor& /*src*/, Tensor& /*dst*/)
   ORT_NOT_IMPLEMENTED(__FUNCTION__, " is not implemented");
 }
 
-common::Status IDataTransfer::CopyTensor(const Tensor& /*src*/, Tensor& /*dst*/, size_t /*src_offset*/, size_t /*dst_offset*/, size_t /*size*/) const {
-  ORT_NOT_IMPLEMENTED(__FUNCTION__, " with offsets/size is not implemented");
-}
-
 common::Status IDataTransfer::CopyTensors(const std::vector<IDataTransfer::SrcDstPair>& src_dst_pairs) const {
-  // All offset and size information is provided via SrcDstPair fields.
-  // This ensures CopyTensors is fully decoupled from any external offset/size arrays.
   for (const auto& pair : src_dst_pairs) {
-    if (pair.source_offset != 0 || pair.destination_offset != 0 || pair.size != 0) {
-      // Use offset-aware copy methods
-      if (pair.src_stream) {
-        ORT_RETURN_IF_ERROR(CopyTensorAsync(pair.src, pair.dst, pair.source_offset, pair.destination_offset, pair.size, *pair.src_stream));
-      } else {
-        ORT_RETURN_IF_ERROR(CopyTensor(pair.src, pair.dst, pair.source_offset, pair.destination_offset, pair.size));
-      }
-    } else {
-      // Use default copy methods (no offsets)
-      if (pair.src_stream) {
-        ORT_RETURN_IF_ERROR(CopyTensorAsync(pair.src, pair.dst, *pair.src_stream));
-      } else {
-        ORT_RETURN_IF_ERROR(CopyTensor(pair.src, pair.dst));
-      }
-    }
+    if (pair.src_stream)
+      ORT_RETURN_IF_ERROR(CopyTensorAsync(pair.src, pair.dst, *pair.src_stream));
+    else
+      ORT_RETURN_IF_ERROR(CopyTensor(pair.src, pair.dst));
   }
+
   return Status::OK();
 }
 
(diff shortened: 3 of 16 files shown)


fs-eire:
@tianleiwu @skottmckay

skottmckay:
> This pull request introduces enhanced support for copying tensor data with fine-grained control over source/destination offsets and copy sizes, both in the ONNX Runtime C API and the WebGPU provider. The changes add new API methods and extend existing interfaces to allow partial or offset-based tensor copies, with corresponding updates to the CPU and WebGPU data transfer implementations and buffer management logic.

Can you please explain 'why' this is required? Changing `OrtDataTransferImpl::CopyTensors` will impact every non-CPU EP. 

A user can create an OrtValue using existing data, so could they not take that approach to create an OrtValue for the subset of data to copy and use the existing interfaces as-is?

tianleiwu:
> A user can create an OrtValue using existing data, so could they not take that approach to create an OrtValue for the subset of data to copy and use the existing interfaces as-is?

I agree. From API point of review, we only need an API to copy from source location to target location for specified number of bytes.

User can add a helper function for sub-tensor copy, but not needed to expose it as API. The helper function can be shared by some EPs if it is used internally.

fs-eire:
> A user can create an OrtValue using existing data, so could they not take that approach to create an OrtValue for the subset of data to copy and use the existing interfaces as-is?

For webgpu, a buffer is just a handle. Unlike CUDA which uses memory model that allows to add offset to pointers, there is no way to represent that in webgpu.

skottmckay:
> > A user can create an OrtValue using existing data, so could they not take that approach to create an OrtValue for the subset of data to copy and use the existing interfaces as-is?
> 
> For webgpu, a buffer is just a handle. Unlike CUDA which uses memory model that allows to add offset to pointers, there is no way to represent that in webgpu.

It's up to the EP (and its data transfer implementation) to interpret what the `void*` for data in the tensor represents. e.g. DML EP that is struct with metadata and not just a raw pointer to memory. WebGPU EP could do something similar and have a struct with the handle and offset info. User would need to know how to create that struct but could use CreateTensorWithDataAsOrtValue and pass it in as the `void* p_data`.

Adding a whole new API to ORT this sort of thing just for the WebGPU EP feels like the wrong place to be doing it. The ORT API in general deals in a granularity of tensors not small chunks of data within a tensor. What's the use-case where you need to copy a subset of data from an OrtValue?

qjia7:
> Adding a whole new API to ORT this sort of thing just for the WebGPU EP feels like the wrong place to be doing it. The ORT API in general deals in a granularity of tensors not small chunks of data within a tensor. What's the use-case where you need to copy a subset of data from an OrtValue?

Two use cases in onnxruntime-genai:
1. Support `CopyFrom(size_t begin_dest, DeviceBuffer& source, size_t begin_source, size_t size_in_bytes)` in gpu. I ever commented there https://github.com/microsoft/onnxruntime-genai/pull/1848/files#r2513245293.
2. Support `UpdateAttentionMask` in gpu.  When graph capture is enabled, the attention mask is in gpu with shape [batch_size,max_sequence_length], we can use copyTensors with offset to update the content of `attention_mask` at the specific position. For example, just copy 1 to of `attention_mask[total_sequence_length-1]`.

skottmckay:
What's the motivation for other EP authors to handle offset based copy in their IDataTransfer implementation? it would be a bit of a smell that it doesn't belong in the ORT API if only the CPU and WebGPU EPs support it.

FWIW it feels a little loose to be taking arbitrary offsets and size given the source and destination are Tensor instances with specific shapes. could we use something like an axis and index for the source and the target locations to ensure the copy makes sense? 

If this is purely to enable some copies in genai is another option to do that via a model by either augmenting the original model with the model editor API, or having a small helper model that is used? e.g. something like ScatterElements or ScatterND might be applicable. if the model input and output for the copy was the same buffer it should in-place the writes.

Is another alternative having a CreateTensor variant that takes an offset to support the `void*` not necessarily being an addressable pointer to memory? Or a CreateSubTensor that takes an OrtValue with Tensor, axis and index and returns a Tensor for that slice (ownership would obviously stay with the original OrtValue, might need const and non-const versions).

qjia7:
@skottmckay

>What's the motivation for other EP authors to handle offset based copy in their IDataTransfer implementation? it would be a bit of a smell that it doesn't belong in the ORT API if only the CPU and WebGPU EPs support it.

You're absolutely right. From an implementation perspective, CPU/CUDA already support creating tensors with offsets implicitly (via pointer arithmetic), while handle-based EPs like WebGPU cannot. However, from a user's perspective, we should provide a unified API that works consistently across all EPs. Your suggestions of either a CreateTensor variant that takes an offset or a CreateSubTensor that takes an OrtValue would achieve this nicely.

>FWIW it feels a little loose to be taking arbitrary offsets and size given the source and destination are Tensor instances with specific shapes. could we use something like an axis and index for the source and the target locations to ensure the copy makes sense?

I agree that using axis+index would provide better type safety. If we align on the overall direction, I'm happy to work out the specific API design with you—whether that's axis/index based or offset based with proper validation.

>If this is purely to enable some copies in genai is another option to do that via a model by either augmenting the original model with the model editor API, or having a small helper model that is used? e.g. something like ScatterElements or ScatterND might be applicable. if the model input and output for the copy was the same buffer it should in-place the writes.

I've prototyped the model-based approach (in fact, I'm using it for Cast operations in [https://github.com/microsoft/onnxruntime-genai/pull/1895](vscode-file://vscode-app/c:/Users/jiajiaqin/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)). However, for frequent small updates like CopyFrom and UpdateAttentionMask, my benchmarks show that direct copying significantly outperforms running a dedicated ONNX model, even a single-op one.

>Is another alternative having a CreateTensor variant that takes an offset to support the void* not necessarily being an addressable pointer to memory? Or a CreateSubTensor that takes an OrtValue with Tensor, axis and index and returns a Tensor for that slice (ownership would obviously stay with the original OrtValue, might need const and non-const versions).

Both approaches would work well for the WebGPU use case. From the EP's perspective, they achieve the same goal: enabling partial tensor updates for both pointer-based EPs (CPU, CUDA) and handle-based EPs (WebGPU, and potentially Vulkan/Metal).

To confirm the path forward: Do I understand correctly that you're supportive of adding new API functionality to enable partial tensor updates across all EPs, and we need to finalize whether to use:


1. CreateTensor with offset + existing CopyTensors

2. CreateSubTensor (with axis/index) + existing CopyTensors

3. Enhanced CopyTensors with axis/index parameters

Happy to prototype whichever approach you think fits best with ORT's API design principles.

skottmckay:
> To confirm the path forward: Do I understand correctly that you're supportive of adding new API functionality to enable partial tensor updates across all EPs, and we need to finalize whether to use:
> 
> 1. CreateTensor with offset + existing CopyTensors
> 2. CreateSubTensor (with axis/index) + existing CopyTensors
> 3. Enhanced CopyTensors with axis/index parameters

One of the challenges with creating a new Tensor instance pointing to a subset of an existing tensor is the path to EP-specific logic to interpret the handle. Typically that sort of logic is in the IAllocator for the OrtDevice, but as we're not allocating in this scenario that doesn't quite fit. Doesn't quite fit to add to IDataTransfer either given we're not doing a data transfer at that point.

One option might be the new external resource importer where you can import memory and create an OrtValue from it. Added recently in #26828. That already supports a `void*` + `offset` and provides an EP specific implementation. Whilst we're technically not importing 'external' memory (e.g. D3D12 or Vulkan), it does seem to have the features required. It's also optional for an EP to implement so zero cost for an EP implementer if not needed. In terms of API changes it would possibly only require adding a value to OrtExternalMemoryHandleType. You could actually import the entire Tensor once, and call CreateTensorFromMemory for each slice as that supports an offset in OrtExternalTensorDescriptor.

Maybe a choice between that and #3. Would like to know what others on the ORT team think about the options. @adrianlizarraga @edgchen1 @yuslepukhin

qjia7:
Ping @adrianlizarraga @edgchen1 @yuslepukhin. Any suggestions on Scott's [above comment](https://github.com/microsoft/onnxruntime/pull/27004#issuecomment-3766993510)?

qjia7:
@skottmckay, I’ve added a new CreateTensor interface that supports byte offsets to replace the previous one CopyTensors with byte offset. This change should have no side effects for the CPU and CUDA EPs. At the same time, it enables the WebGPU EP to support sub-buffer views.
For WebGPU specifically, there is a small overhead when retrieving the raw WGPUBuffer. Previously, we could directly obtain the raw WGPUBuffer via `tensor.MutableDataRaw()` due to byte_offset_ is always 0 although it returns [`static_cast<char*>(p_data_) + byte_offset_`](https://github.com/microsoft/onnxruntime/blob/41f9056ccad76c5dbe52861e81e0b357a324b72f/include/onnxruntime/core/framework/tensor.h#L234-L242), now we need to adjust the logic slightly. The new approach is:
```
reinterpret_cast<WGPUBuffer>(static_cast<char*>(tensor.MutableDataRaw()) - tensor.ByteOffset());
```
This allows us to recover the original WGPUBuffer correctly.
Are you okay with using this new interface CreateTensor with offset to move forward?

yuslepukhin:
> > To confirm the path forward: Do I understand correctly that you're supportive of adding new API functionality to enable partial tensor updates across all EPs, and we need to finalize whether to use:
> > 
> > 1. CreateTensor with offset + existing CopyTensors
> > 2. CreateSubTensor (with axis/index) + existing CopyTensors
> > 3. Enhanced CopyTensors with axis/index parameters
> 
> ...
> Maybe a choice between that and #3. Would like to know what others on the ORT team think about the options. @adrianlizarraga @edgchen1 @yuslepukhin

My preferred approach is to build this work on top of the external resource importer and extend the CopyTensors() implementation to support these tensor types, without requiring any changes to the existing API signature.

skottmckay:
My concern would be that creates a hidden requirement that an EP with an opaque handle in `Tensor::p_data_` needs to always be checking for an offset before accessing data in a Tensor. All the Tensor data access methods automatically apply the offset, which silently invalidates the pointer returned until the EP undoes that.

e.g. would this WebGPU kernel break if it received a tensor instance with an offset?

https://github.com/microsoft/onnxruntime/blob/1acddf2cb9fa356f77bc8ec08e927359ce681e77/onnxruntime/contrib_ops/webgpu/quantization/gather_block_quantized.cc#L160

That doesn't feel overly safe. If an offset can't be directly applied to the data pointer because it's opaque, would it be better to always store a tuple of <WGPUBuffer, offset> in the data pointer so the Tensor class is not aware of that?

fs-eire:
> My concern would be that creates a hidden requirement that an EP with an opaque handle in `Tensor::p_data_` needs to always be checking for an offset before accessing data in a Tensor. All the Tensor data access methods automatically apply the offset, which silently invalidates the pointer returned until the EP undoes that.
> 
> e.g. would this WebGPU kernel break if it received a tensor instance with an offset?
> 
> https://github.com/microsoft/onnxruntime/blob/1acddf2cb9fa356f77bc8ec08e927359ce681e77/onnxruntime/contrib_ops/webgpu/quantization/gather_block_quantized.cc#L160
> 
> That doesn't feel overly safe. If an offset can't be directly applied to the data pointer because it's opaque, would it be better to always store a tuple of <WGPUBuffer, offset> in the data pointer so the Tensor class is not aware of that?

Using a `std::tuple<WGPUBuffer, size_t> *` in `Tensor::p_data_` sounds not a good idea:
- there need extra allocation and memory management for the tuple
- unless `Tensor::byte_offset_` is removed, there is no point to introduce the complexity. Both value can be retrieved from Tensor using existing methods.

Can we add a new method like `Tensor::DataRawWithoutOffset()` so that handle based EPs can use? Anyway WebGPU EP need to make changes to support offset.

EDIT: another proposal is to modify the definition of `DataRaw()` and add a new method `DataRawPointer()` or something similar, which may be less confusing ("DataRaw" sounds like to get the original data instead of implicitly adding an offset) but that requires code changes in the code base (which is easy to do because that basically involves changing everywhere)

EDIT2: seems that currently only training code uses offset so either way is unlikely to break existing inference usage.

skottmckay:
> EDIT2: seems that currently only training code uses offset so either way is unlikely to break existing inference usage.

It's more future usage I'm concerned about. You're adding a new function to the public API to create a Tensor that has an offset. Anyone can call that to create an OrtValue that is used during inferencing. IIUC, any EP that has an opaque value in `Tensor::p_data_` now needs to ensure that it does not use the return value of any of the `Tensor` data access methods directly if there's an offset. A Tensor instance that returns invalid data if `byte_offset_` is set feels pretty fragile. Is that not something we should avoid?

PR #27011, opened 2026-01-14
Fix GPU JAR testing on Linux

eserscor:
### Description
Fix GPU JAR testing

### Motivation and Context
Testing JAR for GPU was missing libcustom_library.so on Linux.


tools/ci_build/github/azure-pipelines/templates/final-jar-testing-linux.yml
```diff
@@ -64,7 +64,10 @@ stages:
 
     - task: Bash@3
       displayName: 'Run Java Tests on Linux/macOS'
```
tianleiwu: nit: change name since macOS is removed.

hariharans29 (approved):

PR #27012, opened 2026-01-14
Publish both Cuda 13 and Cuda 12 packages in packaging pipeline

eserscor:


tools/ci_build/github/azure-pipelines/templates/c-api-cpu.yml
```diff
@@ -383,7 +383,7 @@ stages:
   - Setup
   - Windows_Nodejs_Packaging_x64
   - Windows_Nodejs_Packaging_arm64
-  - Linux_Nodejs_Packaging_x64
+  - Linux_Nodejs_Packaging_x64_cuda_12
```
eserscor, line 387: Should publish both CUDA 13 and 12 NodeJS packages?

PR #27013, opened 2026-01-15
[test] refactor common test target settings

fs-eire:
### Description
- factor duplicated test target settings into helper functions
- reuse helpers for onnxruntime_test_all and onnxruntime_provider_test
- keep target-specific settings intact

### Motivation and Context

There are some duplicated codes in the onnxruntime_unittests. Originally there is only one unit test `onnxruntime_test_all` and later it is split into two: `onnxruntime_test_all` and `onnxruntime_provider_test`. Some lines for setting up build flags are simply copied. This causes potential risk for inconsistency in future.

### DIFF
--- a/cmake/onnxruntime_unittests.cmake
+++ b/cmake/onnxruntime_unittests.cmake
@@ -1023,14 +1023,8 @@ endif()
 
 partition_provider_test_srcs(all_tests onnxruntime_provider_test_srcs onnxruntime_test_all_srcs)
 
-# Shared settings for onnxruntime test targets.
-function(onnxruntime_apply_common_test_target_settings target)
-  if (UNIX AND (onnxruntime_USE_TENSORRT OR onnxruntime_USE_NV))
-    # The test_main.cc includes NvInfer.h where it has many deprecated declarations
-    # simply ignore them for TensorRT EP build
-    set_property(TARGET ${target} APPEND_STRING PROPERTY COMPILE_FLAGS "-Wno-deprecated-declarations")
-  endif()
-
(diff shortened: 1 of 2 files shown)


cmake/onnxruntime_unittests.cmake
```diff
@@ -1023,6 +1023,49 @@ endif()
 
 partition_provider_test_srcs(all_tests onnxruntime_provider_test_srcs onnxruntime_test_all_srcs)
 
+# Shared settings for onnxruntime test targets.
+function(onnxruntime_apply_common_test_target_settings target)
+  if (UNIX AND (onnxruntime_USE_TENSORRT OR onnxruntime_USE_NV))
+    # The test_main.cc includes NvInfer.h where it has many deprecated declarations
+    # simply ignore them for TensorRT EP build
+    set_property(TARGET ${target} APPEND_STRING PROPERTY COMPILE_FLAGS "-Wno-deprecated-declarations")
```
edgchen1: can we limit this to test_main.cc? maybe even around the problematic include?

cmake/onnxruntime_unittests.cmake
```diff
@@ -1023,6 +1023,49 @@ endif()
 
 partition_provider_test_srcs(all_tests onnxruntime_provider_test_srcs onnxruntime_test_all_srcs)
 
+# Shared settings for onnxruntime test targets.
+function(onnxruntime_apply_common_test_target_settings target)
```
edgchen1: nit: it's not so "common" if it is only for two targets. maybe `onnxruntime_apply_test_target_workarounds` is a more specific name.

PR #27015, opened 2026-01-15
Add dedicated API to support extracting compatibility string from model metadata

adrastogi:
### Description
This change proposes a new helper ORT API for callers that need to extract the model compatibility string from a precompiled model. 

### Motivation and Context

See #25749 for more background on the model compatibility concept and infrastructure; #25841 provides a related helper API for an application to call to do a validation check using the compatibility info string. However, there is no direct way to get to the model metadata without creating a session (which some callers may prefer to avoid) or by taking a dependency on a separate library to parse the model's protobuf (which again callers may prefer to avoid). 

This change proposes a separate helper API which can be used to retrieve the compatibility info string, thereby avoiding session creation or an external dependency. This does incur some amount of redundant work in that the model protobuf will be parsed again during session creation- but for some callers, this tradeoff may be acceptable.


include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -6912,6 +6912,77 @@ struct OrtApi {
...
+   *                                Returns nullptr if no compatibility info exists for the specified EP.
+   *                                Caller must free with OrtApi::AllocatorFree when non-null.
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \since Version 1.24.
+   */
+  ORT_API2_STATUS(GetCompatibilityInfoFromModel,
+                  _In_ const ORTCHAR_T* model_path,
+                  _In_ const char* ep_type,
```
skottmckay, line 7048: nit: given the compatibility string is opaque data, is there a need/reason to return it to the user or could this instead be a 'check compatibility' function that took an OrtEpDevice and directly returns OrtCompiledModelCompatibility. if there's nothing meaningful they can do with the string are we forcing them to make 2 calls when 1 would be fine?

adrastogi: Yeah, great observation, I was grappling with this trade-off (i.e., should we have separate functions or a consolidated function).  

I kept them separate because I was perhaps indexing on less common scenarios where an app may want to explicitly track / log the compatibility string (e.g., maybe they are interested in the distribution of precompiled model versions across their user base and would like to report this in their analytics), or for diagnostic purposes (e.g., logging as diagnostic info to perhaps aid in troubleshooting / bug reports). And while the strings are described as opaque, I would imagine there would still be some structural elements that scripts / tools could reason over (e.g., as part of building system inventories of models). 

All of these scenarios are less common, probably, than just doing a basic check which requires two separate calls as you note. The alternative I could think of was to have both, i.e., keep what is proposed in the PR and introduce a parallel pair of functions (one for a model path, one for a buffer) that operates in the way you describe using the OrtEpDevice and returning the status (here, the compatibility string remains internal to the implementation). But then I am not sure if there are desires to try and limit the proliferation of various APIs across ORT. 

What do you think? Maybe adding the extra functions isn't that onerous.

skottmckay: We definitely want to minimize the ORT API surface areas as it's rather large already.

Maybe it's better to have this separate as a) it supports a user wanting to do something with the compat info (in case that ever happens no new API is needed), and b) provides a direct signal if there's no compat info available (I don't believe there's an explicit value in OrtCompiledModelCompatibility for that).

PR #27022, opened 2026-01-15
Add INT2 and UINT2 support for QDQ, transpose and cast ops

vraspar:
### Description
Add INT2 and UINT2 support for QDQ, transpose and cast ops

### Motivation and Context

Thanks to @jambayk for implementing most of the ops

### DIFF
--- a/onnxruntime/test/onnx/TestCase.cc
+++ b/onnxruntime/test/onnx/TestCase.cc
@@ -1462,41 +1462,6 @@ std::unique_ptr<std::set<BrokenTest>> GetBrokenTests(const std::string& provider
     broken_tests->insert({"attention_4d_with_past_and_present_qk_matmul_bias_3d_mask_causal_expanded", "unknown version"});
     broken_tests->insert({"attention_4d_with_past_and_present_qk_matmul_bias_4d_mask_causal_expanded", "unknown version"});
     broken_tests->insert({"convinteger_with_padding", "unknown version"});
-    // crashes on QNN EP, needs investigation
-    broken_tests->insert({"dequantizelinear_int2", "unknown version"});
-    broken_tests->insert({"dequantizelinear_uint2", "unknown version"});
-    broken_tests->insert({"quantizelinear_int2", "unknown version"});
-    broken_tests->insert({"quantizelinear_uint2", "unknown version"});
-    broken_tests->insert({"cast_FLOAT16_to_INT2", "unknown version"});
-    broken_tests->insert({"cast_FLOAT16_to_UINT2", "unknown version"});
-    broken_tests->insert({"cast_FLOAT_to_INT2", "unknown version"});
-    broken_tests->insert({"cast_FLOAT_to_UINT2", "unknown version"});
-    broken_tests->insert({"cast_INT2_to_FLOAT", "unknown version"});
-    broken_tests->insert({"cast_INT2_to_FLOAT16", "unknown version"});
-    broken_tests->insert({"cast_INT2_to_INT8", "unknown version"});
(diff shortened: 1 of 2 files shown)


onnxruntime/core/providers/cpu/tensor/transpose.cc
```diff
@@ -8,6 +8,8 @@
 #include "core/framework/utils.h"
 #include "core/framework/transpose_helper.h"
 #include "core/framework/op_kernel_type_control_utils.h"
+#include "core/framework/int2.h"
+#include "core/framework/int4.h"
```
hariharans29: Nit: Is int4 inclusion necessary here ?

jambayk: both int2 and int4 get indirectly exported through element_type_lists.h so we could keep/remove both to be consistent.

jambayk: removed!

onnxruntime/test/onnx/tensorprotoutils.cc
```diff
@@ -648,6 +648,8 @@ ONNXTensorElementDataType CApiElementTypeFromProtoType(int type) {
     CASE_TYPE(FLOAT4E2M1)
     CASE_TYPE(UINT4)
     CASE_TYPE(INT4)
+    CASE_TYPE(UINT2)
+    CASE_TYPE(INT2)
```
adrianlizarraga, line 652: I found what is making the onnx "test_quantizelinear_int2" tests to crash. Handling of INT2/UINT2 is missing from `utils::CApiElementTypeFromProtoType()` in : https://github.com/microsoft/onnxruntime/blob/d8f0318694c8898bc5e23ff6bfd1167e11f70ab7/onnxruntime/core/framework/tensorprotoutils.cc#L1590-L1591

Because this is not implemented, the following statement (line 628) crashes while running the test.
https://github.com/microsoft/onnxruntime/blob/d8f0318694c8898bc5e23ff6bfd1167e11f70ab7/onnxruntime/test/util/compare_ortvalue.cc#L625-L629

This will affect any onnx test model that outputs a INT2 or UINT2 (e.g., test_quantizelinear_int2, test_cast_FLOAT_to_INT2, etc.). It is not specific to QNN. I replicate on CPU EP.

onnxruntime/test/onnx/TestCase.cc
```diff
@@ -1462,6 +1462,41 @@ std::unique_ptr<std::set<BrokenTest>> GetBrokenTests(const std::string& provider
     broken_tests->insert({"attention_4d_with_past_and_present_qk_matmul_bias_3d_mask_causal_expanded", "unknown version"});
     broken_tests->insert({"attention_4d_with_past_and_present_qk_matmul_bias_4d_mask_causal_expanded", "unknown version"});
     broken_tests->insert({"convinteger_with_padding", "unknown version"});
+    // crashes on QNN EP, needs investigation
```
adrianlizarraga: Should be able to re-enable this if fix https://github.com/microsoft/onnxruntime/pull/27022/files#r2697185885

onnxruntime/test/onnx/tensorprotoutils.cc
jambayk: done! thanks for catching this.

onnxruntime/test/onnx/TestCase.cc
jambayk: done!

hariharans29 (approved):

adrianlizarraga (approved):

PR #27037, opened 2026-01-16
Move model compatibility checks ahead of session initialization

adrastogi:
### Description

The current infrastructure for validating compatibility of a precompiled model does the check after session initialization occurs, which turns out to be quite costly. The check should ideally happen beforehand, to short-circuit those expensive operations.

### Motivation and Context

This change will make it more tractable for applications to rely on the existing session machinery to check compatibility of any of their models.


adrianlizarraga (approved):

skottmckay (approved):
:shipit:

PR #27039, opened 2026-01-16
Fix incorrect warning suppression in projects using precompiled headers on Windows

skottmckay:
### Description

Disable cmake prologue in pch file which results in warnings being unexpectedly suppressed in the unit test projects that use precompiled headers.

e.g. C4834 was suppressed so there was no warning for the nodiscard Status return value not being checked in a Windows build. Generates a warning on other platforms.

Update a few tests to resolve warnings that now get triggered.

### Motivation and Context

Warnings causing build failures for non-Windows CIs that should have been generated for Windows builds.

### DIFF
--- a/onnxruntime/contrib_ops/cpu/cdist.cc
+++ b/onnxruntime/contrib_ops/cpu/cdist.cc
@@ -72,6 +72,7 @@ static void CalculateSqeuclidean(const Tensor& a, const Tensor& b, Tensor& c, co
   // if you're on x86 and care about performance try MKL first. if there's a good enough argument for optimizing this
   // we can look into it in the future.
   ORT_UNUSED_PARAMETER(threadpool);
+  ORT_UNUSED_PARAMETER(mlas_backend_kernel_selector_config);
 
   // https://eigen.tuxfamily.org/dox/TopicWritingEfficientProductExpression.html
   auto out_map = EigenMatrixMapRowMajor<T>(c_data, SafeInt<size_t>(m), SafeInt<size_t>(n));


onnxruntime/test/providers/cpu/math/gemm_test.cc
```diff
@@ -27,7 +27,7 @@ auto initialize_matrix = [](int64_t rows, int64_t cols) {
   std::vector<float> data;
   data.reserve(rows * cols);
   for (int64_t i = 0; i < rows * cols; ++i) {
-    data.push_back(((i % 7) + 1));
+    data.push_back(float((i % 7) + 1));
```
yuslepukhin: >float [](http://example.com/codeflow?start=19&length=5)

should be static_cast

onnxruntime/core/providers/cuda/tensor/pad.cc
```diff
@@ -94,7 +94,7 @@ Status Pad<T>::ComputeInternal(OpKernelContext* ctx) const {
   typedef typename ToCudaType<T>::MappedType CudaT;
   const auto& input_tensor = *ctx->Input<Tensor>(0);
   auto const& input_shape = input_tensor.Shape();
-  const size_t dimension_count = input_shape.NumDimensions();
+  const int32_t dimension_count = narrow<int32_t>(input_shape.NumDimensions());
```
yuslepukhin, line 97: Why is this needed? `size_t` is the right type to use.

onnxruntime/core/providers/cuda/cuda_execution_provider.cc
```diff
@@ -3043,11 +3045,6 @@ CUDAExecutionProvider::GetCapability(const onnxruntime::GraphViewer& graph,
         result.push_back(ComputeCapability::Create(std::move(sub_graph)));
       } else {
         // We break here so we do not have patches of CUDA assigned nodes.
-        auto* node = graph.GetNode(node_index);
-        if (node != nullptr) {
```
yuslepukhin, line 3069: You do not need a check, but I would like a warning to be there.

onnxruntime/test/framework/hardware_device_compatibility_test.cc
```diff
@@ -186,7 +186,7 @@ TEST(GetHardwareDeviceEpIncompatibilityDetailsCapiTest, CpuEp_ReturnsEmptyDetail
   ASSERT_ORTSTATUS_OK(api->DeviceEpIncompatibilityDetails_GetErrorCode(details, &error_code));
   EXPECT_EQ(error_code, 0);
 
-  const char* notes = reinterpret_cast<const char*>(0xDEADBEEF);  // Initialize to non-null
+  const char* notes = "...";  // Initialize to non-null to verify it gets set
```
yuslepukhin, line 189: deadbeef is always very telling in the debugger.

skottmckay: Does that matter here given it's a very local test that the initial value is overridden.

onnxruntime/core/providers/cuda/tensor/pad.cc
skottmckay: The loop on line 142 uses int32_t for `i` as well as dimension_count. Can change that to use size_t instead.

skottmckay: Using int32_t is the cleanest way to fix the warnings. TArray indicies are int32_t and there are lots of places in this function using TArray variables with dimension_count.

onnxruntime/contrib_ops/cuda/bert/xqa/xqa_impl_gen.cuh
```diff
@@ -14,8 +14,34 @@ namespace NAMESPACE_NAME {
...
+#define XQA_HAS_SM80_TARGET 1
+#endif
+#elif defined(__CUDA_ARCH_LIST__)
+#if __CUDA_ARCH_LIST__ >= 800
+#define XQA_HAS_SM80_TARGET 1
+#endif
+#else
+// Non-nvcc fallback: assume supported (IDE parsers, etc.)
+#define XQA_HAS_SM80_TARGET 1
+#endif
```
skottmckay, line 41: @tianleiwu does this change look fine to fix build warnings for older architectures?

onnxruntime/core/providers/cuda/llm/attention.cc
```diff
@@ -125,364 +125,365 @@ Status Attention<T>::ComputeInternal(OpKernelContext* context) const {
   if (is_gqa) {
     // Use GQA path with Flash Attention or Memory Efficient Attention
     // GQA only supports float16 and bfloat16 types
-    if (std::is_same<T, float>::value) {
+    if constexpr (std::is_same<T, float>::value) {
```
skottmckay, line 129: Note: Turn off whitespace diffs for this file. 

`QkvToContext<float, float>` is not defined. Using `constexpr` avoids that error from line 484.

onnxruntime/contrib_ops/cuda/bert/xqa/xqa_impl_gen.cuh
tianleiwu: Yes, it looks good to me.

PR #27040, opened 2026-01-16
Tweak external resource importer handle structs

skottmckay:
### Description

Use the descriptor struct in the external resource handle. 

We were copying most fields, but the setup is a little more intuitive when the descriptor is used directly.


nieubank (approved):

edgchen1:
looks like this updates the EP API so I'll add the release:1.24.0 label.

PR #27045, opened 2026-01-16
Raise the timeout for the ios simulator job

hariharans29:
### Description
As title - it looks like the duration of the job is very close to the timeout

### Motivation and Context
Reduce retrry attempts for the ios sim job

My own PR - https://github.com/microsoft/onnxruntime/pull/26688 keep timing out this job leg

### DIFF
--- a/.github/workflows/mac.yml
+++ b/.github/workflows/mac.yml
@@ -72,7 +72,7 @@ jobs:
       matrix:
         target_arch: [x86_64, arm64]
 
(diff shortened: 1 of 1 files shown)


vraspar:
I might be mistaken, but I vaguely remember someone having similar pr last year relating to increase build time in iOS or Mac related ci but it was closed because of some issue.

eserscor (approved):

PR #27048, opened 2026-01-17
Bump version to 1.25.0

tianleiwu:
### Description
Increase version number to 1.25.0.

This may need update after we cherry-pick API changes to rel-1.24.0 so that we know the exact boundary of APIs.


hariharans29 (dismissed):

edgchen1:
there's also `OrtEpApi`. do we need to add a similar static_assert there?

https://github.com/microsoft/onnxruntime/blob/4013dc193f66e8e1d8b51b5fd693ca285b251226/onnxruntime/core/session/plugin_ep/ep_api.cc#L870-L877

onnxruntime/core/session/onnxruntime_c_api.cc
```diff
@@ -4547,18 +4548,19 @@ static_assert(offsetof(OrtApi, SetEpDynamicOptions) / sizeof(void*) == 284, "Siz
...
 
 // So that nobody forgets to finish an API version, this check will serve as a reminder:
-static_assert(std::string_view(ORT_VERSION) == "1.24.0",
+static_assert(std::string_view(ORT_VERSION) == "1.25.0",
               "ORT_Version change detected, please follow below steps to ensure OrtApi is updated properly");
 // 1. Update the hardcoded version string in above static_assert to silence it
-// 2. If there were any APIs added to ort_api_1_to_24 above:
+// 2. If there were any APIs added to ort_api_1_to_25 above:
 //    a. Add the 'End of version #' markers (pattern above should be obvious)
 //    b. Add a static_assert in the directly above list of version sizes to ensure nobody adds any more functions to the just shipped API version
```
edgchen1, line 4733: thanks for updating `OrtEpApi`.
there are actually a couple such structs to check. I think the only other one that needed updating this time was `OrtEpApi`. however, maybe it would be useful to also mention these structs in these instructions, so folks know where to check.

https://github.com/microsoft/onnxruntime/blob/50a03e4179809c0299bc143e521dcfc7817ace75/onnxruntime/core/session/model_editor_c_api.cc#L366
https://github.com/microsoft/onnxruntime/blob/50a03e4179809c0299bc143e521dcfc7817ace75/onnxruntime/core/session/compile_api.cc#L324
https://github.com/microsoft/onnxruntime/blob/50a03e4179809c0299bc143e521dcfc7817ace75/onnxruntime/core/session/plugin_ep/ep_api.cc#L809

edgchen1: please add a reply if not making a code change.

tianleiwu: ep_api.cc change is included in this PR.
The other two api files has no new API for version 24 so they are not updated.

edgchen1: I meant adding some more instructions to these comments. we can do that in another PR.

PR #27049, opened 2026-01-17
Fix some issues when building with the latest CUDA and cuDNN versions

skottmckay:
### Description

Fix some issues when building with the latest CUDA and cuDNN versions on Windows.

* Latest cuDNN install has the CUDA toolkit version in the path. 
  * Adjust cmake files to support that.
* CUDA 13.x drops support for compute capability 6.0 and 7.0. 
  * Remove from CMAKE_CUDA_ARCHITECTURES.
* Remove a LINK_LANGUAGE:CUDA flag for CETCOMPAT
  * Syntax doesn't seem to be supported with MSVC. Build is successful without this (CUDA 13.1, cuDNN 9.17).
    * `LINK : warning LNK4044: unrecognized option '/Xlinker=/CETCOMPAT'; ignored [D:\src\github\ort.cuda\build\Windows.CUDA\Debug\onnxruntime_providers_cuda_ut.vcxproj]`
* Memory leak checker fixes
  * onnxruntime_providers_cuda_ut was incorrectly linking against ORT common causing a duplicate symbol when the debug leak checker is enabled (multiple overrides of `new` and `delete`. 
    * As the CUDA EP is built as a separate library it shouldn't need to link against `common`.
  * Use the debug alloc/free for provider bridge when leak checker is enabled
  * Ignore EtwEventWriteNoRegistration in leak checker output as we don't control that.

### DIFF
--- a/cmake/onnxruntime_python.cmake
+++ b/cmake/onnxruntime_python.cmake
@@ -297,12 +297,23 @@ if (WIN32)
   if (onnxruntime_USE_CUDA)
     file(WRITE "${VERSION_INFO_FILE}" "use_cuda = True\n")
     if(onnxruntime_CUDNN_HOME)
-      file(GLOB CUDNN_DLL_PATH "${onnxruntime_CUDNN_HOME}/bin/cudnn64_*.dll")
-      if (NOT CUDNN_DLL_PATH)
-        file(GLOB CUDNN_DLL_PATH "${onnxruntime_CUDNN_HOME}/bin/${onnxruntime_CUDA_VERSION}/cudnn64_*.dll")
-        if (NOT CUDNN_DLL_PATH)
-          message(FATAL_ERROR "cuDNN not found in ${onnxruntime_CUDNN_HOME}")
+      # may have x64 in the path
+      # may have a path with CUDA toolkit version if multiple installed on the machine
+      set(CUDNN_SEARCH_PATHS
+        "${onnxruntime_CUDNN_HOME}/bin/cudnn64_*.dll"
+        "${onnxruntime_CUDNN_HOME}/bin/x64/cudnn64_*.dll"
+        "${onnxruntime_CUDNN_HOME}/bin/${onnxruntime_CUDA_VERSION}/cudnn64_*.dll"
+        "${onnxruntime_CUDNN_HOME}/bin/${onnxruntime_CUDA_VERSION}/x64/cudnn64_*.dll"
+      )
+      set(CUDNN_DLL_PATH "")
+      foreach(search_path ${CUDNN_SEARCH_PATHS})
+        file(GLOB CUDNN_DLL_PATH "${search_path}")
+        if(CUDNN_DLL_PATH)
+          break()
         endif()
+      endforeach()
+      if(NOT CUDNN_DLL_PATH)
+        message(FATAL_ERROR "cuDNN not found in ${onnxruntime_CUDNN_HOME}")
       endif()
     else()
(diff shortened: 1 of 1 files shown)


cmake/external/cuDNN.cmake
```diff
@@ -3,7 +3,7 @@ add_library(CUDNN::cudnn_all INTERFACE IMPORTED)
 find_path(
     CUDNN_INCLUDE_DIR cudnn.h
     HINTS $ENV{CUDNN_PATH} ${CUDNN_PATH} ${Python_SITEARCH}/nvidia/cudnn ${CUDAToolkit_INCLUDE_DIRS}
-    PATH_SUFFIXES include
+    PATH_SUFFIXES include include/${onnxruntime_CUDA_VERSION}
```
tianleiwu, line 6: Is it 9.17 specified? 9.18 is back to include

cmake/external/cuDNN.cmake
```diff
@@ -15,7 +15,7 @@ function(find_cudnn_library NAME)
     find_library(
         ${NAME}_LIBRARY ${NAME} "lib${NAME}.so.${CUDNN_MAJOR_VERSION}"
         HINTS $ENV{CUDNN_PATH} ${CUDNN_PATH} ${Python_SITEARCH}/nvidia/cudnn ${CUDAToolkit_LIBRARY_DIR}
-        PATH_SUFFIXES lib64 lib/x64 lib
+        PATH_SUFFIXES lib64 lib/x64 lib lib/${onnxruntime_CUDA_VERSION}/x64
```
tianleiwu, line 18: Is it 9.17 specified? 9.18 is back to lib/x64

cmake/onnxruntime_python.cmake
```diff
@@ -299,7 +299,10 @@ if (WIN32)
     if(onnxruntime_CUDNN_HOME)
       file(GLOB CUDNN_DLL_PATH "${onnxruntime_CUDNN_HOME}/bin/cudnn64_*.dll")
       if (NOT CUDNN_DLL_PATH)
-        message(FATAL_ERROR "cuDNN not found in ${onnxruntime_CUDNN_HOME}")
+        file(GLOB CUDNN_DLL_PATH "${onnxruntime_CUDNN_HOME}/bin/${onnxruntime_CUDA_VERSION}/cudnn64_*.dll")
```
tianleiwu: Need add similar handling of cudnn 9.18, its dll is in a new path: bin\x64\cudnn64_9.dll

tianleiwu (approved):

cmake/external/cuDNN.cmake
skottmckay: Might depend on what's on the system. I just tried installing 9.18 and I get different directories for the 2 CUDA toolkit versions I have installed. This is the case for `bin`, `include` and `lib` directories. It also adds 'x64' to the `bin` and `lib` paths as well.

image

cmake/onnxruntime_python.cmake
skottmckay: Updated to support 'x64' and/or the CUDA toolkit version potentially being in the path.

cmake/external/cuDNN.cmake
skottmckay: Seems to be conditional on whether multiple CUDA toolkits are installed. I get separate directories for 12.9 and 13.1 with 9.17 and 9.18.

cmake/external/cuDNN.cmake
tianleiwu: If you use exe installer, it will have cuda version. If you download the zip file (choose Tarball), it does not have cuda version sub-directory since the zip filename contains cuda version.

PR #27050, opened 2026-01-18
[EP ABI] Add CreateCustomOpDomains() API for plugin EP to register custom ops

chilo-ms:
### Description

The newly added two APIs, `CreateCustomOpDomains()` and `GetNumCustomOpDomains`,  are used when running inference on a model that contains EP-specific custom operations.
  
 Workflow:
 1. The EP implements these functions to supply a list of `OrtCustomOpDomain` instances.
 2. The application either 1) calls `SessionOptionsAppendExecutionProvider_V2()` with an `OrtEpDevice` containing
     the plugin EP's factory or 2) enables auto ep selection.
 3. Then ORT either 1) `SessionOptionsAppendExecutionProvider_V2()` appends the provided OrtCustomOpDomains to the
     session options or 2) registers the OrtCustomOpDomains from the selected EP devices.
  
 As a result, any session created from these session options will have these custom op domains registered
 in ORT, ensuring that the custom ops are properly recognized and validated when the model is loaded.

Plugin EPs can provide two types of custom ops:
   1. A full OrtCustomOp with a concrete kernel implementation
      - This Example EP demonstrates this approach.
      - In GetCapability(), it calls EpGraphSupportInfo_AddSingleNode() to inform ORT
        that the custom node should NOT be fused or compiled. Instead, ORT should invoke
        the custom node's Compute() function at runtime.

   2. A "placeholder" OrtCustomOp with an empty kernel implementation
      - A compile-based Plugin EP can supply an OrtCustomOp whose CustomKernel::Compute()
        does nothing. The purpose is to satisfy model validation during model loading by
        registering the custom op as a valid operator in the session.
      - In GetCapability(), the EP should call EpGraphSupportInfo_AddNodesToFuse() to
        notify ORT that this custom node should be fused and compiled by the EP.
      - In Compile(), the EP executes its compiled bits to perform inference for
      the fused custom node.

### Motivation and Context

Currently, the provider-bridge TRT RTX EP and TRT EP supports registering custom op domain list in session option so
that it can run model contains TRT specific custom ops.

This PR adds the same feature for plugin EP.

### DIFF
--- a/onnxruntime/test/testdata/custom_mul.py
+++ b/onnxruntime/test/testdata/custom_mul.py
@@ -3,15 +3,15 @@ import onnx
 
 def create_custom_mul_model():
     # === Inputs ===
-    x = onnx.helper.make_tensor_value_info("X", TensorProto.FLOAT, [3, 2])
-    w = helper.make_tensor_value_info("W", TensorProto.FLOAT, [3, 2])
+    x = onnx.helper.make_tensor_value_info("X", onnx.TensorProto.FLOAT, [3, 2])
+    w = onnx.helper.make_tensor_value_info("W", onnx.TensorProto.FLOAT, [3, 2])
 
     # === Output ===
-    y = helper.make_tensor_value_info("Y", TensorProto.FLOAT, [3, 2])
+    y = onnx.helper.make_tensor_value_info("Y", onnx.TensorProto.FLOAT, [3, 2])
 
     # === Custom Node: Custom_Mul ===
     # Replace "Mul" with your custom op name and domain
-    custom_node = helper.make_node(
+    custom_node = onnx.helper.make_node(
         op_type="Custom_Mul",  # <-- custom op name
         inputs=["X", "W"],
         outputs=["Y"],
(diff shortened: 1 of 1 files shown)


chilo-ms:
This PR is identical to https://github.com/microsoft/onnxruntime/pull/26759, as for some reasons i can't find the unresolved comment in the conversation and resolve it, it's blocking that PR to be merged.

onnxruntime/core/session/utils.cc
```diff
@@ -195,6 +247,31 @@ static OrtStatus* CreateSessionAndLoadModelImpl(_In_ const OrtSessionOptions* op
...
+      options->provider_factories.empty() &&
+      options->value.ep_selection_policy.enable) {
+    InlinedVector<OrtCustomOpDomain*> all_ep_custom_op_domains;
+
+    for (const OrtEpDevice* ep_device : env.GetOrtEpDevices()) {
+      InlinedVector<OrtCustomOpDomain*> domains;
+      ORT_API_RETURN_IF_STATUS_NOT_OK(GetCustomOpDomainsFromEpDevice(*ep_device, domains));
+
+      for (auto domain : domains) {
+        if (ShouldAddDomain(domain, options->custom_op_domains_)) {
```
edgchen1, line 263: do we want to check the domains in `all_ep_custom_op_domains` here too?

PR #27055, opened 2026-01-18
[test] move IO binding test to onnxruntime_provider_tests

fs-eire:
### Description

move IO binding test to onnxruntime_provider_tests from onnxruntime_test_all

### Motivation and Context

Currently, a few unit test cases are included in `onnxruntime_test_all`. This works well until we want to support WebGPU EP as a plugin EP.

The main differences between `onnxruntime_test_all` and `onnxruntime_provider_tests` is that the latter respects plugin EP configuration. The IO Binding test cases involve EP configuration (eg. "enableGraphCapture" for WebGPU) so it no longer works with WebGPU EP.

### DIFF
--- a/onnxruntime/test/providers/io_binding_test.cc
+++ b/onnxruntime/test/providers/io_binding_test.cc
@@ -4,16 +4,7 @@
 #include "core/graph/onnx_protobuf.h"
 #include "core/session/inference_session.h"
 
-#include <algorithm>
-#include <cfloat>
-#include <functional>
-#include <future>
-#include <iterator>
-#include <thread>
 #include <sstream>
-#include <iostream>
-#include <fstream>
-#include <random>
 
 #include "core/graph/model.h"
 #include "core/framework/tensorprotoutils.h"
@@ -64,7 +55,7 @@ static void CreateMatMulModel(std::unique_ptr<onnxruntime::Model>& p_model, Prov
 #endif
   }
   Status status = graph.Resolve();
-  ASSERT_TRUE(status.IsOK()) << status.ErrorMessage();
+  ASSERT_STATUS_OK(status);
 }
 
 void RunModelWithBindingMatMul(InferenceSession& session_object,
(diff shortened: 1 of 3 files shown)


onnxruntime/test/providers/io_binding_test.cc
```diff
@@ -0,0 +1,404 @@
...
+  auto& node = graph.AddNode("node1", "MatMul", "MatMul", input_defs, output_defs, nullptr, onnxruntime::kOnnxDomain);
+  if (provider_type == kCpuExecutionProvider) {
+    node.SetExecutionProviderType(provider_type);
+  } else {
+#if defined(USE_CUDA) || defined(USE_WEBGPU)
+    node.SetExecutionProviderType(provider_type);
+#endif
+  }
+  Status status = graph.Resolve();
+  ASSERT_TRUE(status.IsOK()) << status.ErrorMessage();
```
edgchen1: general comment: `ASSERT_TRUE(<status>.IsOK())` -> `ASSERT_STATUS_OK(<status>)`

onnxruntime/test/providers/io_binding_test.cc
```diff
@@ -0,0 +1,404 @@
...
+}
+
+#if defined(USE_CUDA) || defined(USE_WEBGPU)
+#if USE_CUDA
+constexpr const char* kGpuExecutionProvider = kCudaExecutionProvider;
+#elif USE_WEBGPU
+constexpr const char* kGpuExecutionProvider = kWebGpuExecutionProvider;
+#endif
+
+TEST(InferenceSessionTests, TestBindCuda) {
```
edgchen1: general: if it is more generic than CUDA (this test appears to also support webgpu), can we give the test a more generic name? e.g., `TestBindGpu`.

onnxruntime/test/unittest_util/framework_test_utils.h
```diff
@@ -128,5 +129,22 @@ inline int OpCount(const OpCountMap& op_count_map, const std::string& op_type) {
...
+void VerifyOutputs(const Tensor& tensor, const std::vector<int64_t>& expected_dims,
+                   const std::vector<T>& expected_values) {
+  TensorShape expected_shape(expected_dims);
+  ASSERT_EQ(expected_shape, tensor.Shape());
+  const std::vector<T> found(tensor.Data<T>(),
+                             tensor.Data<T>() + expected_values.size());
+  ASSERT_EQ(expected_values, found);
+}
+
+inline void VerifyOutputs(const std::vector<OrtValue>& fetches, const std::vector<int64_t>& expected_dims,
```
edgchen1: maybe a better name would be `VerifySingleOutput()` as this function expects just one value in `fetches`.

edgchen1 (commented):
looks good in general. some minor comments. copilot comments also look valid.

PR #27059, opened 2026-01-19
Drop QDQ around MaxPool at opset 22

mcollinswisc:
### Description

Includes MaxPool at opset 22 in the DropQDQNodesSelector, so that this gets optimized to a full-integer graph. Previously this was only done for MaxPool at opset 12.

### Motivation and Context

Right now, at the most recent opsets, this optimization is no longer being applied.

I'd guess MaxPool was limited to only being selected by the DropQDQ action at opset 12 because at earlier opsets MaxPool did not support any integer types. For example at opset 11:

https://github.com/onnx/onnx/blob/main/docs/Changelog.md#MaxPool-11

but opset 12 added support int8/uint8:

https://github.com/onnx/onnx/blob/main/docs/Changelog.md#MaxPool-12

Now a new version has been introduced at opset 22, that still supports int8/uint8:

https://github.com/onnx/onnx/blob/main/docs/Changelog.md#MaxPool-22


mcollinswisc:
Also made an issue in case that helps resolve this: https://github.com/microsoft/onnxruntime/issues/28770

mcollinswisc:
The MaxPool opsets were updated for QDQ propagation in https://github.com/microsoft/onnxruntime/commit/a70ac2f69eb15e82afabf40fab373d2c8582d057. Just not dropping QDQ

PR #27038, opened 2026-01-16
[webgpu] support arbitrary input_channel size for im2col

wenqinI:
### Description

This PR supports `vec1/vec2` for arbitrary `input_channel` in `im2col` kernel, which could bring performance gain to more models.

Like for `yolov8n_pose` model, there is about **~7%** gain for whole model, and **~50%** for those `conv2d` op which `input_size` are not multiple of 4.

PR #27069, opened 2026-01-20
[webgpu] Fix a bug for im2col

wenqinI:
### Description

This PR fix a bug in `im2col` for `pads` in some dimension.

### DIFF
--- a/onnxruntime/core/providers/webgpu/nn/im2col_matmul.wgsl.template
+++ b/onnxruntime/core/providers/webgpu/nn/im2col_matmul.wgsl.template
@@ -33,7 +33,7 @@ fn load_src(batch : u32, m : u32, k_packed_idx : u32) -> src_value_t {
 
   // 4. Calculate the coordinate in the original input tensor
   let src_h_coord : i32 = i32(src_h_coord_padded) - i32(uniforms.pads.x);
-  let src_w_coord : i32 = i32(src_w_coord_padded) - i32(uniforms.pads.z);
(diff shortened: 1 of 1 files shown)


wenqinI:

qjia7:
Could we also include a test to help detect this type of error in the future?

wenqinI:
> Could we also include a test to help detect this type of error in the future?

There are already some test cases for that, but their `input_channel` size is not multiple of `4`, so after this [PR](https://github.com/microsoft/onnxruntime/pull/27038), there will be some test cover for this issue, do you think we need add some specific test case for `input_channel` as multiple for `4`?

qjia7:
> > Could we also include a test to help detect this type of error in the future?
> 
> There are already some test cases for that, but their `input_channel` size is not multiple of `4`, so after this [PR](https://github.com/microsoft/onnxruntime/pull/27038), there will be some test cover for this issue, do you think we need add some specific test case for `input_channel` as multiple for `4`?

Yes, if it's easy to add.

guschmue:
nice catch!

wenqinI:
> > > Could we also include a test to help detect this type of error in the future?
> > 
> > 
> > There are already some test cases for that, but their `input_channel` size is not multiple of `4`, so after this [PR](https://github.com/microsoft/onnxruntime/pull/27038), there will be some test cover for this issue, do you think we need add some specific test case for `input_channel` as multiple for `4`?
> 
> Yes, if it's easy to add.

Hi @qjia7,

Given this PR is merged, I will add a test case in this [PR](https://github.com/microsoft/onnxruntime/pull/27038) later.

PR #27070, opened 2026-01-20
[CPU] Fix arithmetic overflow and legacy TODO in Det operator

the0cp:
### Description
This PR fixes the legacy `TODO: fix the warnings` in the `Det` operator.
The arithmetic overflow warning (C26451) is addressed by using `int64_t` for tensor dimension and batch size calculations, ensuring safe pointer arithmetic.

### Motivation and Context
- Removes unused warning suppression pragma.
- Prevents potential overflow when handling large batches of matrices.

### DIFF
--- a/onnxruntime/core/providers/cpu/math/det.cc
+++ b/onnxruntime/core/providers/cpu/math/det.cc
@@ -3,11 +3,7 @@
 
 #include "core/providers/cpu/math/det.h"
 #include "core/util/math_cpuonly.h"
-// TODO: fix the warnings
-#if defined(_MSC_VER) && !defined(__clang__)
-// Chance of arithmetic overflow could be reduced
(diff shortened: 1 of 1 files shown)


hariharans29 (approved):

the0cp:

the0cp:
Hi @hariharans29, I've updated the code to address the feedback:
- Used `onnxruntime::narrow<Eigen::Index>` for safe type conversion.
- Switched tensor rank and loop indices to `size_t` to resolve the WASM `-Wshorten-64-to-32` errors while keeping `int64_t` for batch size/offset calculations to prevent overflow.
- I've also squashed the commits to keep the history clean and rebased onto the latest main.

Thanks for your help with the typo and the guidance on ORT conventions!

Issue #27068, opened 2026-01-20
[Web] Report `Buffer (unlabled) used in submit while destoryed` when enabling SDXL Turbo demo


xhcao:
### Describe the issue

After debugging and investigating the issue, it may be related with Pre-Pack feature, the input tensor of Pre-Pack may be already deleted before the Pre-Pack kernels being submitted.
WebGPU need to prepare 16 kernels for one command buffer submitting. But there are only 11 nodes needing Pre-Pack in `unet` model, so it will not submit command buffer until next model uploads its weights. But the inputs' tensors of Pre-Pack are deleted immediately https://github.com/microsoft/onnxruntime/blob/1817f4a81072c49dea2984853edca412674d66fb/onnxruntime/core/framework/session_state.cc#L644, and the buffer manager for initializers is `BufferCacheMode::Disabled` https://github.com/microsoft/onnxruntime/blob/1817f4a81072c49dea2984853edca412674d66fb/onnxruntime/core/providers/webgpu/webgpu_context.cc#L142,  so error happens.

### To reproduce

Use webgpu (not JS) EP, run SDXL turbo demo https://github.com/Honry/webnn-developer-preview/blob/sdxl-turbo-pr/demos/sdxl-turbo/index.js, throws an error `Buffer (unlabled) used in submit while destoryed` in `loadModels` function when creating sessions.

### ONNX Runtime Installation

Built from Source

### ONNX Runtime Version or Commit ID

7e1d818ba7923c1e44187284a6ca77bbe13aa1eb

### Execution Provider

'webgpu' (WebGPU)

xhcao:
@fs-eire @qjia7 I do not know the good place to submit/flush command buffer after Pre-Pack, please help fix the issue, thanks.
@jchen10

xhcao:
I had written a C++ native case, but could not reproduce this issue. When releasing the buffer in the `BufferCacheMode::Disabled`, would only dec the ref_count in dawn, and the ref_count was not 0, so the buffer was not destoryed. May `compute_pipeline` reference the buffer?
I confuse the different behaviors between the web and native.

qjia7:
@xhcao, please check whether #27077 can fix this issue. Thanks.

fs-eire:
This issue should be fixed by #27077.

RReverser:
Did that fix make it into 1.24.1? I'm seeing issue with very similar symptoms (at least same error), not sure whether to bother investigating further if the fix has already landed.

[...]

PR #27077, opened 2026-01-20
[webgpu] Use LazyRelease for prepack allocator

qjia7:
BUG #27068

### DIFF
--- a/onnxruntime/core/providers/webgpu/webgpu_kernel.cc
+++ b/onnxruntime/core/providers/webgpu/webgpu_kernel.cc
@@ -49,9 +49,11 @@ Status WebGpuKernel::PrePack(const Tensor& tensor, int input_idx, AllocatorPtr /
 
   Status s = PrePackInternal(context, tensor, input_idx, ep_.PrepackAllocator(), is_packed);
 
-  // Flush pending commands to ensure GPU buffer creations are completed.
-  // This allows the initializer buffer manager to release temporary buffers and reduce memory usage.
-  webgpu_context_.Flush(webgpu_context_.InitializerBufferManager());
+  if (is_packed) {
+    // Flush pending commands to ensure GPU buffer creations are completed.
+    // This allows the initializer buffer manager to release temporary buffers and reduce memory usage.
+    webgpu_context_.Flush(webgpu_context_.InitializerBufferManager());
+  }
 
   if (webgpu_context_.ValidationMode() >= ValidationMode::Full) {
     ORT_RETURN_IF_ERROR(webgpu_context_.PopErrorScope());


fs-eire:
thank you for the fix. I have a question: should we call `context.Flush()` after each prepack (or after the initialization?)

qjia7:
> thank you for the fix. I have a question: should we call `context.Flush()` after each prepack (or after the initialization?)

Good point. I definitely don't want to call `context.Flush()`  after the initialization which is not good for memory. To not impact the original initializer, I add a new one specific for prepack which will follow the normal buffer manager to flush when pending patches  reach to max_num_pending_dispatches_. What do you think?

qjia7:
> > thank you for the fix. I have a question: should we call `context.Flush()` after each prepack (or after the initialization?)
> 
> Good point. I definitely don't want to call `context.Flush()` after the initialization which is not good for memory. To not impact the original initializer, I add a new one specific for prepack which will follow the normal buffer manager to flush when pending patches reach to max_num_pending_dispatches_. What do you think?

Sorry—this approach may still have issues. With this change, I can’t properly control the release of the original weights. I’ll think through this more carefully and send an updated proposal. Thanks.

qjia7:
> > > thank you for the fix. I have a question: should we call `context.Flush()` after each prepack (or after the initialization?)
> > 
> > 
> > Good point. I definitely don't want to call `context.Flush()` after the initialization which is not good for memory. To not impact the original initializer, I add a new one specific for prepack which will follow the normal buffer manager to flush when pending patches reach to max_num_pending_dispatches_. What do you think?
> 
> Sorry—this approach may still have issues. With this change, I can’t properly control the release of the original weights. I’ll think through this more carefully and send an updated proposal. Thanks.

Restore to the original one with LazyRelease plus flush after each prepack.

xhcao:
With this PR, the issue https://github.com/microsoft/onnxruntime/issues/27068 cannot been reproduced again. Thanks.

fs-eire (dismissed):

onnxruntime/core/providers/webgpu/webgpu_kernel.cc
```diff
@@ -49,6 +49,10 @@ Status WebGpuKernel::PrePack(const Tensor& tensor, int input_idx, AllocatorPtr /
 
   Status s = PrePackInternal(context, tensor, input_idx, ep_.PrepackAllocator(), is_packed);
 
+  // Flush pending commands to ensure GPU buffer creations are completed.
+  // This allows the initializer buffer manager to release temporary buffers and reduce memory usage.
+  webgpu_context_.Flush(webgpu_context_.InitializerBufferManager());
```
fs-eire:
```suggestion
  if (is_packed) {
    // Flush pending commands to ensure GPU buffer creations are completed.
    // This allows the initializer buffer manager to release temporary buffers and reduce memory usage.
    webgpu_context_.Flush(webgpu_context_.InitializerBufferManager());
  }
```

qjia7: We need to flush no matter is_packed is true or false since the cache mode becomes `LazyRelease` instead of `Disabled` if we want to release the gpu buffer aggressively.

qjia7: @xhcao reminds me that uploading already has context_.Flush and it happens before prepack phase so we don't need to concern it. So I will change it as suggested.

BTW, for is_packed = true path, in fact, the call stack prepack->transpose->run->flush can ensure that prepack will be flushed with batch=16. The left ones(<16) at the end can be flushed in OnSessionInitializationEnd if we want to support flush with a batch size.

PR #27078, opened 2026-01-20
cmake: auto-detect vendored ONNX when onnx_SOURCE_DIR is unset

aabhinavg1:
### Description
This change makes CMake automatically use the vendored ONNX source when onnx_SOURCE_DIR is not set.

### Motivation and Context

The build currently fails when `onnx_SOURCE_DIR` is not set, even though ONNX is vendored in the repository. 
This change fixes that.

https://github.com/onnx/onnx/issues/7587


aabhinavg1:

Issue #27003, opened 2026-01-14
[WebGPU EP] Current Split-K implementation may cause non-deterministic results


Jiawei-Shao:
In current WebGPU EP we use Split-K to accelerate `GEMM` and `Conv2D|MatMul` operators when `K` is large enough compared with `M` and `N`. This Split-K implementation is referenced from many existing open-sourced machine learning implementations (for example, [`BitBLAS`](https://github.com/microsoft/BitBLAS/blob/main/bitblas/ops/general_matmul/tilelang/dequantize/matmul_dequantize_mma.py#L469), [`tilelang`](https://github.com/tile-ai/tilelang/blob/main/examples/gemm_splitk/example_tilelang_gemm_splitk.py) and [`intel-xpu-backend-for-triton`](https://github.com/intel/intel-xpu-backend-for-triton/blob/b4acf89537714278a59b8dd40e084588350c0fae/benchmarks/triton_kernels_benchmark/gemm_splitk_benchmark.py#L23)). However, due to **the random order of summation of partial results from different threads** and **non-associative nature of floating point arithmetic**, the results may vary slightly between runs, which may lead to non-deterministic outputs for models that use these operators with same inputs.

Here is an example on the **non-associative nature of floating-point arithmetic**. 
```
Let x = 1e30, y = -1e30, z = 1

1. (x + y) + z = 0 + 1 = 1
2.  x + (y + z) = 1e30 - 1e30 = 0
```
2 is because in floating-point arithmetic `(y + z)` is rounded to `-1e30` because `1` is too small relative to `1e30` to affect the stored precision.

**The random order of summation of partial results from different threads** comes from the current implementation of `Split-K`. In current implementation of Split-K, we first split a large `K` dimension into smaller chunks, then each thread computes a partial result for its assigned chunk, and finally we sum up all partial results with `atomicAdd` operations (actually it is a pair of `atomicLoad` and `atomicCompareExchangeWeak` because latest WGSL doesn't support `atomicAdd` on floating-point numbers). As we are still doing the summation of floating-point numbers, theoretically the precision should be the same as the original non-Split-K implementation, but as the order of summation may vary between runs due to the concurrent nature of GPU threads, the final results may become slightly different after every run.

For example, when `K` is `1024` and `Split-K` is enabled (`256` elements per chunk):
1. Initialize the output tensor with `bias` or `0`.
2. The `1024` elements are split into `4` chunks.
[...]

PR #27086, opened 2026-01-21
[WebGPU EP] Disable `Split-K` when `use_deterministic_compute` is true

Jiawei-Shao:
### Description
This patch disables `Split-K` when the onnxruntime session is created with `use_deterministic_compute` set to true in `SessionOptions`, which is set to false by default right now.

In `onnxruntime_provider_tests`, we can add `so.use_deterministic_compute`  to true in the `BaseTester::Run()` to require the result must be deterministic with same input.

### Motivation and Context
Current implementation of `Split-K` only relies on atomic operations which will cause non-deterministic answers, so we shouldn't use `Split-K` when we always want deterministic output with same input.

See #27003 for more details.

### DIFF
--- a/js/web/test/e2e/exports/testcases/nextjs-default/package-lock.json
+++ b/js/web/test/e2e/exports/testcases/nextjs-default/package-lock.json
@@ -8,25 +8,35 @@
       "name": "nextjs-default",
       "version": "0.1.0",
       "dependencies": {
-        "next": "15.4.10",
+        "next": "16.1.5",
         "react": "^19.0.0",
         "react-dom": "^19.0.0"
       }
     },
     "node_modules/@emnapi/runtime": {
-      "version": "1.5.0",
-      "resolved": "https://registry.npmjs.org/@emnapi/runtime/-/runtime-1.5.0.tgz",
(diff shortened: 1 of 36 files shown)


Jiawei-Shao:
@fs-eire @guschmue  PTAL, thanks!

onnxruntime/core/providers/webgpu/math/matmul.cc
```diff
@@ -246,7 +246,7 @@ Status ComputeMatMul(ComputeContext* context,
   uint32_t split_dim_inner = 1;
 
   const SplitKConfig& split_k_config = context->GetSplitKConfig();
```
fs-eire: I would prefer to move the check of `if (context->KernelContext().GetUseDeterministicCompute())` here instead of inside `UseSplitK()`. Technically the function should not depend on the compute context

Jiawei-Shao: Done

fs-eire:
change looks good but need to wait for the CI fix (https://github.com/microsoft/onnxruntime/pull/27235). May need sync to latest main once it's merged.

guschmue (approved):

PR #27088, opened 2026-01-21
Implement compiled model compatibility APIs in example plugin EP and add tests 

adrastogi:
### Description

Add tests for the compiled model compatibility APIs (`GetCompiledModelCompatibilityInfo` and `ValidateCompiledModelCompatibilityInfo`) using the example plugin EP.

Changes:
- Implement the compatibility methods in example_plugin_ep to enable testing
- Update `GetCapabilityImpl` and `CompileImpl` to handle EPContext nodes (required for loading compiled models in tests)
- Add `PluginEp_CompatibilityInfo_WrittenToMetadata` test - verifies compatibility info is written to compiled model metadata
- Add `PluginEp_CompatibilityInfo_ValidatedOnLoad` test - verifies the round-trip: compile → load → validate compatibility

### Motivation and Context

The compiled model compatibility APIs need test coverage to ensure they work correctly end-to-end. This exercises the full flow of writing compatibility metadata during compilation and validating it when loading a pre-compiled model. Additionally, by adding this support into the plugin EP, it enables us to add scenario tests that would exercise this in the context of Windows ML.

### DIFF
--- a/onnxruntime/test/autoep/library/example_plugin_ep/ep.cc
+++ b/onnxruntime/test/autoep/library/example_plugin_ep/ep.cc
@@ -430,9 +430,9 @@ OrtStatus* ORT_API_CALL ExampleEp::CompileImpl(_In_ OrtEp* this_ptr, _In_ const 
       // Create MulKernel for Mul nodes
       ep->mul_kernels_.emplace(fused_node_name,
                                std::make_unique<MulKernel>(ep->ort_api, ep->logger_,
-                                                          ep->float_initializers_,
-                                                          node_inputs[0].GetName(),
-                                                          node_inputs[1].GetName()));
+                                                           ep->float_initializers_,
+                                                           node_inputs[0].GetName(),
+                                                           node_inputs[1].GetName()));
 
       // Use ExampleNodeComputeInfo for Mul nodes
       auto node_compute_info = std::make_unique<ExampleNodeComputeInfo>(*ep);


onnxruntime/test/autoep/library/example_plugin_ep/ep.cc
```diff
@@ -207,11 +208,29 @@ OrtStatus* ORT_API_CALL ExampleEp::GetCapabilityImpl(OrtEp* this_ptr, const OrtG
     }
 
     std::vector<Ort::ConstNode> supported_nodes;
+    std::vector<Ort::ConstNode> ep_context_nodes;
```
adrianlizarraga: I'm not sure if this suggestion will make things more understandable, but I wonder if it would be simpler to just have a `supported_nodes` array (not a separate one for ep context nodes).

This example EP is a bit strange since it only supports one node at a time (a problem to be addressed at a later time). So, perhaps the first loop over the nodes just adds the one node into `supported_nodes`. It could either be a Mul, EPContext, or CustomMul node.

Then, depending on which node it is, the EP calls either `EpGraphSupportInfo_AddNodesToFuse` or `EpGraphSupportInfo_AddSingleNode`

onnxruntime/test/autoep/library/example_plugin_ep/ep.cc
```diff
@@ -305,29 +341,60 @@ OrtStatus* ORT_API_CALL ExampleEp::CompileImpl(_In_ OrtEp* this_ptr, _In_ const
...
+      // `ep_cache_context` attribute and deserialize it here. This example EP does not do that.
+      //
+      // As a result:
+      // - Session creation with a compiled model will succeed (for metadata access, compatibility testing)
+      // - Inference may fail at runtime if MulKernel::Compute cannot find expected inputs/initializers
+      //
+      // To fully support EPContext execution, deserialize `ep_cache_context` and restore initializers.
+      for (size_t i = 0; i < node_inputs.size() && i < 2; ++i) {
+        node_input_names[i] = node_inputs[i].GetName();
+      }
```
adrianlizarraga: This implementation is using a "MulKernel" to represent both a Mul and an EPContext node, which may not have the same number of inputs. I wonder if this compile function would be easier to understand if we introduced a dummy "EpContextKernel" that does no computation (could be a passthrough).

Aside: This example EP has a few shortcomings that tend to complicate the implementations of GetCapability() and Compile. I'm hoping that this suggestion could make things easier for someone using this EP as a reference.

onnxruntime/test/autoep/library/example_plugin_ep/ep.cc
```diff
@@ -340,7 +407,8 @@ OrtStatus* ORT_API_CALL ExampleEp::CompileImpl(_In_ OrtEp* this_ptr, _In_ const
     node_compute_infos[0] = node_compute_info.release();
 
     // Create EpContext nodes for the fused nodes we compiled.
-    if (ep->config_.enable_ep_context) {
+    // Don't create new EPContext nodes if we're already processing an EPContext node!
+    if (ep->config_.enable_ep_context && !is_ep_context_node) {
```
adrianlizarraga: Would it make sense to make the condition `enable_ep_context && is_ep_context_node` an error?

onnxruntime/test/autoep/library/example_plugin_ep/ep.cc
adrastogi: Hm yeah, fair point here. I tried out this suggestion, which had a few ripple effects as I was iterating.

adrianlizarraga (approved):
I think there are some linter issues remaining. Thank you for implementing the feedback.

PR #27099, opened 2026-01-21
mlas/arm64: add NEON conv asm kernels and tune NCHWC kernel selection

milpuz01:
##  Overview

This PR adds ARM64 NEON assembly micro‑kernels for NCHW, depthwise, and pointwise convolution, wires them into the MLAS build, and adds shape‑based selection heuristics for NCHWC depthwise/pointwise to favor the asm kernels in safe cases (stride‑1 pointwise; wider depthwise outputs). The BF16 path is unchanged.

## Key changes

  - cmake/onnxruntime_mlas.cmake
      - Add new AArch64 assembly sources for NCHW, depthwise, and pointwise conv to the MLAS build.
  - onnxruntime/core/mlas/lib/aarch64/SconvKernelNeon.S
      - New vectorised NCHW convolution micro‑kernel.
  - onnxruntime/core/mlas/lib/aarch64/SconvDepthwiseKernelNeon.S
      - New vectorised depthwise micro‑kernel (fast path for in‑bounds loads, slow path for padding).
  - onnxruntime/core/mlas/lib/aarch64/SconvPointwiseKernelNeon.S
      - New vectorised pointwise micro‑kernel (multi‑output reuse).
  - onnxruntime/core/mlas/lib/mlasi.h, onnxruntime/core/mlas/lib/platform.cpp
      - Declare/register new asm kernels and prefer them on ARM64.
  - onnxruntime/core/mlas/lib/snchwc.cpp
      - Heuristics: use pointwise asm when StrideHeight == 1 && StrideWidth == 1 and OutputThisIteration >= 4; use depthwise asm when OutputWidth >= 4.
  - onnxruntime/core/mlas/lib/sbconv_kernel_neon.cpp
      - Include fix for the conv kernel flags header.

 ## Performance 

Numbers below are expressed as multipliers vs the non‑NCHWC baseline (same model and perf_test settings):

Baseline (no `--enable_arm_neon_nchwc`)

  - 8 cores: 1.00×
  - 16 cores: 1.00×

With `--enable_arm_neon_nchwc` (no asm additions/heuristics)

  - 8 cores: 1.18×
  - 16 cores: 1.24×

With this PR (asm kernels + heuristics)

  - 8 cores: 1.77×
  - 16 cores: 2.54×

## Testing

  - `./build.sh --config Release --build_shared_lib --parallel --compile_no_warning_as_error --skip_submodule_sync --skip_tests --enable_pybind --build_wheel --enable_arm_neon_nchwc`
  - `OMP_NUM_THREADS=8 ./build/Linux/Release/onnxruntime_perf_test -I -m times -r 1000 --x 8 ~/mobilenetv2-7.onnx`

### DIFF
--- a/onnxruntime/test/perftest/ort_test_session.cc
+++ b/onnxruntime/test/perftest/ort_test_session.cc
@@ -258,12 +258,8 @@ OnnxRuntimeTestSession::OnnxRuntimeTestSession(Ort::Env& env, std::random_device
                          "qnn_saver_path", "htp_graph_finalization_optimization_mode", "qnn_context_priority",
                          "htp_arch", "enable_htp_fp16_precision", "offload_graph_io_quantization",
                          "enable_htp_spill_fill_buffer", "enable_htp_shared_memory_allocator", "dump_json_qnn_graph",
-<<<<<<< HEAD
                          "json_qnn_graph_dir", "disable_file_mapped_weights", "htp_bf16_enable", "enable_vtcm_backup_buffer_sharing", "extended_udma"});
 
-=======
-                         "json_qnn_graph_dir", "disable_file_mapped_weights", "htp_bf16_enable", "enable_vtcm_backup_buffer_sharing"});
->>>>>>> 21cfa8d39 ([QNN-EP] Implement file mapped weights feature (#26952))
     for (const auto& provider_option : provider_options) {
       const std::string& key = provider_option.first;
       const std::string& value = provider_option.second;
@@ -328,10 +324,7 @@ OnnxRuntimeTestSession::OnnxRuntimeTestSession(Ort::Env& env, std::random_device
                  key == "enable_htp_spill_fill_buffer" ||
                  key == "enable_htp_shared_memory_allocator" ||
                  key == "dump_json_qnn_graph" ||
-<<<<<<< HEAD
                  key == "extended_udma" ||
-=======
->>>>>>> 21cfa8d39 ([QNN-EP] Implement file mapped weights feature (#26952))
                  key == "disable_file_mapped_weights" ||
                  key == "enable_vtcm_backup_buffer_sharing") {
         std::set<std::string> supported_options = {"0", "1"};


aviralagrawal:
Interesting contribution - thank you!

A few questions -
1. Pointwise convolution is currently implemented via direct GEMM - which I assume is optimized. How does this kernel beat the performance of GEMM?
2. Can you share the link to the mobilenet model that you used for performance benchmarking?
3. How does it perform on single threaded experiments? Afaik, the original nchwc kernels in #25580 suffered in the single threaded setting but outperformed the default in thread counts>8.

milpuz01:
Hi @aviralagrawal, thank you vey much for your prompt feedback.

> 1. Pointwise convolution is currently implemented via direct GEMM - which I assume is optimized. How does this kernel beat the performance of GEMM?

Compared to direct GEMM implementation of pointwise convolution asm kernel computes 1x1 conv directly:
- it explicitly tiles 4 outputs: computes up to 4 output positions in parallel and reuses filter loads across those outputs so with single load we are able to accumulate 4 outputs while direct GEMM doesn't tile multiple outputs together
- fuses accumulate/bias/ReLU into store path instead of separate passes with direct GEMM
- unrolls the block size explicitly with 16 invocations to keep accumulators in registers and minimise loop overheads thus reducing dispatch/param overhead and output read-modify-write passes compared to direct GEMM

As usual there are trade-offs so direct GEMM would be faster when output count is small because then asm kernel drops to single-output path which has less ILP and won't be able to reuse filter loads, non-unit stride and non-contigious output regions hence why we have heuristics checking for stride width and height and very large K/M when GEMM blocking can make better use of caches then a fixed 4-output tile.

This is best illustrated if we extract pointwise convolutions from mobilnet that we ran and we can see that on average asm  implementation is 1.07x faster, and significant speed ups are when number of channels is high and K/M are small (in the image those are H and W dimensions). In convolution heavy networks the convolutions that are dominant are ones with large number of channels and low height and width so we see visible performance improvements as optimisations from this PR are weighted in that direction.

image


> 2. Can you share the link to the mobilenet model that you used for performance benchmarking?

For benchmarking we used the model from:   https://github.com/onnx/models/blob/main/validated/vision/classification/mobilenet/model/mobilenetv2-7.onnx

> 3. How does it perform on single threaded experiments? Afaik, the original nchwc kernels in [NEON kernels for NCHWc Convolution and Pooling #25580](https://github.com/microsoft/onnxruntime/pull/25580) suffered in the single threaded setting but outperformed the default in thread counts>8.

Running `OMP_NUM_THREADS=1 ./build/Linux/Release/onnxruntime_perf_test -I -m times -r 1000 --x 1 ~/mobilenetv2-7.onnx` on Graviton 4 with binary that was built with `--enable_arm_neon_nchwc` it is slower by 0.89x then building without that flag, while with this PR it is actually 1.25x faster than the baseline.

Rohanjames1997:
Thanks @milpuz01 for the detailed description & comment!

A couple questions from my side:

1. Is there a reason why [ConvNchwcFloatKernel](https://github.com/microsoft/onnxruntime/pull/27099/changes#diff-2f1ef5fcaaa83277145acadea37a3d9da65e9605664f03efd4ca6208fa191e3bR587) was not optimized? Afaik, It is not very different from `ConvNchwFloatKernel`. The x86 asm implementations of these two kernels [differ very slightly](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/mlas/lib/x86_64/SconvKernelCommon.h#L204C26-L223) too. It is a much heavier kernel than Pointwise and Depthwise, and many larger Conv models stress this kernel. An example for this type of model is in this comment:   https://github.com/microsoft/onnxruntime/pull/25580#issuecomment-3321304864. 

2. Can we switch the default path of Fp32 Conv on Arm64 to use these new kernels? (effectively voiding `--enable_arm_neon_nchwc ` like it was before) Asking because this PR improves upon the single-threaded performance as well. I'd love to hear your thoughts, but also would be wise to hear from @hariharans29 before implementing.

onnxruntime/core/mlas/lib/platform.cpp
```diff
@@ -571,7 +571,19 @@ Return Value:
...
+    this->ConvNchwFloatKernel = MlasConvNchwFloatKernelNeonAsm;
+#if defined(_WIN32)
+     this->ConvDepthwiseFloatKernel = MlasConvDepthwiseFloatKernelNeon;
+#else
+     this->ConvDepthwiseFloatKernel = MlasConvDepthwiseFloatKernelNeonAsm;
+#endif
+    // Prefer the hand written AArch64 micro-kernel for pointwise convolution
+    // as it computes multiple output positions at once and significantly
+    // reduces memory traffic
+    this->ConvPointwiseFloatKernel = MlasConvPointwiseFloatKernelNeonAsm;
```
Rohanjames1997: These lines are overridden by the lines below, and the asm kernels are referenced correctly in `snchwc.cpp`. Are they required here?

onnxruntime/core/mlas/lib/sbconv_kernel_neon.cpp
```diff
@@ -17,7 +17,7 @@ Module Name:
 #if defined(MLAS_USE_ARM_NEON_NCHWC) && defined(__linux__)
 
 #include "mlasi.h"
-#include "sconv.h"
+#include "sconv_nchwc_kernel_neon.h"
```
Rohanjames1997, line 20: Thank you for this! 
The CI did not catch this, so I had to submit this change alone in https://github.com/microsoft/onnxruntime/pull/27101 for the 1.24 release.

You might need to rebase this PR before CI runs.

onnxruntime/core/mlas/lib/platform.cpp
```diff
@@ -571,7 +571,19 @@ Return Value:
     this->EltwiseDispatch = &MlasEltwiseDispatchNeon;
 
 #if defined(MLAS_USE_ARM_NEON_NCHWC)
-    this->ConvNchwFloatKernel = MlasConvNchwFloatKernelNeon;
```
Rohanjames1997, line 574: Since `MlasConvNchwFloatKernelNeon` is no longer used anywhere, should we also remove its implementation & its references in other files?

onnxruntime/core/mlas/lib/mlasi.h
```diff
@@ -963,9 +963,15 @@ extern "C" {
     MLAS_SBGEMM_FLOAT_KERNEL MlasSbgemmKernelAdd;
 #endif
 #if defined(MLAS_TARGET_ARM64) && defined(MLAS_USE_ARM_NEON_NCHWC)
+    // AArch64 assembly micro-kernel for direct NCHW convolution
+    MLAS_CONV_FLOAT_KERNEL MlasConvNchwFloatKernelNeonAsm;
     MLAS_CONV_FLOAT_KERNEL MlasConvNchwFloatKernelNeon;
     MLAS_CONV_FLOAT_KERNEL MlasConvNchwcFloatKernelNeon;
+    // AArch64 assembly micro-kernel for depthwise NCHW convolution
```
Rohanjames1997: Should this and the below comment say `NCHWc` instead of `NCHW` ?

cmake/onnxruntime_mlas.cmake
```diff
@@ -460,6 +464,8 @@ else()
           ${MLAS_SRC_DIR}/aarch64/QgemmS8S8KernelSdot.S
           ${MLAS_SRC_DIR}/aarch64/SgemmKernelNeon.S
           ${MLAS_SRC_DIR}/aarch64/SgemvKernelNeon.S
+          ${MLAS_SRC_DIR}/aarch64/SconvDepthwiseKernelNeon.S
+          ${MLAS_SRC_DIR}/aarch64/SconvPointwiseKernelNeon.S
```
Rohanjames1997: Is there a reason these files were added outside of the `setup_arm_neon_nchwc` function?

onnxruntime/core/mlas/lib/aarch64/SconvPointwiseKernelNeon.S
```diff
@@ -0,0 +1,441 @@
...
+
+// Stack layout for arguments passed on the stack.  The first eight arguments
+// are in x0-x7, the remaining four are placed on the stack by the caller.
+
+        .equ    .LPW_OutputStride, 0
+        .equ    .LPW_OutputCount,  8
+        .equ    .LPW_Bias,         16
+        .equ    .LPW_Flags,        24
+
+// Kernel flag bits.  Keep these in sync with sconv.h.
```
Rohanjames1997: Nit: `sconv_nchwc_kernel_neon.h`

milpuz01:
Hi @Rohanjames1997, thank you very much for your comments.


> 1. Is there a reason why [ConvNchwcFloatKernel](https://github.com/microsoft/onnxruntime/pull/27099/changes#diff-2f1ef5fcaaa83277145acadea37a3d9da65e9605664f03efd4ca6208fa191e3bR587) was not optimized? 

No, particular reason. Mostly because the focus for this PR was on MobileNet model and lack of bandwidth. Thank you for sharing the model where `ConvNchwcFloatKernel` is invoked. We will take a look at optimising it to, but I would suggest that we add optimisation as a follow up PR so that we do not overload this PR with too many changes to review.

> 2. Can we switch the default path of Fp32 Conv on Arm64 to use these new kernels? (effectively voiding `--enable_arm_neon_nchwc ` like it was before) Asking because this PR improves upon the single-threaded performance as well. I'd love to hear your thoughts, but also would be wise to hear from @hariharans29 before implementing.

Yes, I think that is great idea and would be interesting to hear from @hariharans29 too what other testing we should make to try to make these kernels default. As you can see above this change is not going to accelerate all possible pointwise convolutions for example but on average it will show the improvements so if we could agree on a set of performance targets we can use that to drive the decision.

Also thank you for your code review I will address them in a separate commit.

hariharans29:
> Hi @Rohanjames1997, thank you very much for your comments.
> 
> > 1. Is there a reason why [ConvNchwcFloatKernel](https://github.com/microsoft/onnxruntime/pull/27099/changes#diff-2f1ef5fcaaa83277145acadea37a3d9da65e9605664f03efd4ca6208fa191e3bR587) was not optimized?
> 
> No, particular reason. Mostly because the focus for this PR was on MobileNet model and lack of bandwidth. Thank you for sharing the model where `ConvNchwcFloatKernel` is invoked. We will take a look at optimising it to, but I would suggest that we add optimisation as a follow up PR so that we do not overload this PR with too many changes to review.
> 
> > 2. Can we switch the default path of Fp32 Conv on Arm64 to use these new kernels? (effectively voiding `--enable_arm_neon_nchwc ` like it was before) Asking because this PR improves upon the single-threaded performance as well. I'd love to hear your thoughts, but also would be wise to hear from @hariharans29 before implementing.
> 
> Yes, I think that is great idea and would be interesting to hear from @hariharans29 too what other testing we should make to try to make these kernels default. As you can see above this change is not going to accelerate all possible pointwise convolutions for example but on average it will show the improvements so if we could agree on a set of performance targets we can use that to drive the decision.
> 
> Also thank you for your code review I will address them in a separate commit.

Unfortunately, I don't have a comprehensive list of performance targets to be met to make the feature default. Since, the performance testing may not include all possible Conv shapes, I would like to err on the side of caution and atleast provide one release timeline heads-up to the users before considering making the feature default. I would also encourage you to open a discussion to solicit feedback from other ORT users on ARM if they see speed-up for their models with this feature. It would provide greater confidence and a strong data point to turn it on by default.

Thanks for this contribution, we will review it shortly !

onnxruntime/core/mlas/lib/aarch64/SconvKernelNeon.S
```diff
@@ -0,0 +1,835 @@
+/*++
+SPDX-FileCopyrightText: Copyright 2026 Arm Limited and/or its affiliates <open-source-office@arm.com>
+SPDX-License-Identifier: MIT
+
+Module Name:
+    SconvKernelNeon.S
+
+Abstract:
+    Hand written AArch64 vectorised kernel used by the NCHW convolution path.  The
```
hariharans29: Is this for NCHW or NCHWc ?

EDIT: I think it is for NCHW activations and NCHWc weights ?

cmake/onnxruntime_mlas.cmake
hariharans29: +1. Not able to understand the deviation in file inclusions. Can you please clarify ?

onnxruntime/core/mlas/lib/aarch64/SconvKernelNeon.S
```diff
@@ -0,0 +1,835 @@
...
+        b.ne    2b
+        ldr     x12,[sp,#.KO_DilatedInputWidth]
+        add     x25,x25,x12
+        add     x26,x26,x12
+        subs    x9,x9,#1
+        b.ne    1b
+        .endm
+
+// -----------------------------------------------------------------------------
+// void MlasConvNchwFloatKernelNeonAsm(...)
```
hariharans29, line 680: Just to clarify: So this is for the case where the input is NCHW and the weights are NCHWc. Is that correct ?

onnxruntime/core/mlas/lib/platform.cpp
```diff
@@ -571,7 +571,19 @@ Return Value:
...
+    // Prefer the hand written micro-kernel for the NCHW convolution path.  It
+    // offers a tighter schedule and a specialised two-output inner loop that
+    // reduces pressure on the memory system compared
+    this->ConvNchwFloatKernel = MlasConvNchwFloatKernelNeonAsm;
+#if defined(_WIN32)
+     this->ConvDepthwiseFloatKernel = MlasConvDepthwiseFloatKernelNeon;
```
hariharans29: This seems to have Windows platform based special-casing. The others don't have it. Is that intentional ?

onnxruntime/core/mlas/lib/snchwc.cpp
```diff
@@ -1024,6 +1034,9 @@ struct MLAS_NCHWC_CONV_DEPTHWISE_ALGORITHM : MLAS_NCHWC_CONV_ALGORITHM
 
 #if defined(MLAS_TARGET_AMD64) || defined(MLAS_TARGET_LARCH64) || (defined(MLAS_TARGET_ARM64) && defined(MLAS_USE_ARM_NEON_NCHWC))
         MLAS_CONV_DEPTHWISE_FLOAT_KERNEL* Kernel = GetMlasPlatform().ConvDepthwiseFloatKernel;
+#if defined(MLAS_TARGET_ARM64) && defined(MLAS_USE_ARM_NEON_NCHWC)
+        MLAS_CONV_DEPTHWISE_FLOAT_KERNEL* const KernelFast = MlasConvDepthwiseFloatKernelNeonAsm;
```
hariharans29, line 1041: Is this available on all platforms or non-Windows ? The platform.cc changes seem to suggest non-Windows only ?

onnxruntime/core/mlas/lib/platform.cpp
milpuz01: No, they are not. Fixed in https://github.com/microsoft/onnxruntime/pull/27099/commits/6c980e30d58d7f60ef394f5e56ec1b15d4aa414d

onnxruntime/core/mlas/lib/mlasi.h
milpuz01: Yes, Fixed in https://github.com/microsoft/onnxruntime/pull/27099/commits/6c980e30d58d7f60ef394f5e56ec1b15d4aa414d

onnxruntime/core/mlas/lib/platform.cpp
milpuz01: Removed the implementation in https://github.com/microsoft/onnxruntime/pull/27099/commits/6c980e30d58d7f60ef394f5e56ec1b15d4aa414d

cmake/onnxruntime_mlas.cmake
milpuz01: The files should in `setup_aerm_neon_nchwc`. Moved them in https://github.com/microsoft/onnxruntime/pull/27099/commits/6c980e30d58d7f60ef394f5e56ec1b15d4aa414d

onnxruntime/core/mlas/lib/aarch64/SconvPointwiseKernelNeon.S
milpuz01: Fixed in https://github.com/microsoft/onnxruntime/pull/27099/commits/6c980e30d58d7f60ef394f5e56ec1b15d4aa414d

onnxruntime/core/mlas/lib/aarch64/SconvKernelNeon.S
milpuz01: As per discussion in https://github.com/microsoft/onnxruntime/pull/27099#issuecomment-3792485400 it is for NCHW and comment update in https://github.com/microsoft/onnxruntime/pull/27099/commits/6c980e30d58d7f60ef394f5e56ec1b15d4aa414d to reflect that.

onnxruntime/core/mlas/lib/aarch64/SconvKernelNeon.S
milpuz01: As above.

onnxruntime/core/mlas/lib/platform.cpp
milpuz01: That was a mistake. Removed in https://github.com/microsoft/onnxruntime/pull/27099/commits/6c980e30d58d7f60ef394f5e56ec1b15d4aa414d

onnxruntime/core/mlas/lib/snchwc.cpp
milpuz01: Same as above.

milpuz01:
> Unfortunately, I don't have a comprehensive list of performance targets to be met to make the feature default. Since, the performance testing may not include all possible Conv shapes, I would like to err on the side of caution and atleast provide one release timeline heads-up to the users before considering making the feature default. I would also encourage you to open a discussion to solicit feedback from other ORT users on ARM if they see speed-up for their models with this feature. It would provide greater confidence and a strong data point to turn it on by default.

Thanks @hariharans29. I agree with erring on the side of caution. If this PR goes through and it is in the main release is it possible to add a note that we would like to make `--enable_arm_neon_nchwc` as a default in the future releases so that we can try to get some feedback on that via that route too? Thanks again.

hariharans29:
> > Unfortunately, I don't have a comprehensive list of performance targets to be met to make the feature default. Since, the performance testing may not include all possible Conv shapes, I would like to err on the side of caution and atleast provide one release timeline heads-up to the users before considering making the feature default. I would also encourage you to open a discussion to solicit feedback from other ORT users on ARM if they see speed-up for their models with this feature. It would provide greater confidence and a strong data point to turn it on by default.
> 
> Thanks @hariharans29. I agree with erring on the side of caution. If this PR goes through and it is in the main release is it possible to add a note that we would like to make `--enable_arm_neon_nchwc` as a default in the future releases so that we can try to get some feedback on that via that route too? Thanks again.

Thanks @milpuz01. The PR should go through in main eventually but I don't think it will go in 1.24.0 unfortunately as the release branch is cut and the bar to take in new code at this point is critical bug fixes and urgent customer asks only. I will try to take this in for 1.24.1 when it happens and sure I will add a note about considering making it default in one of the future releases, but ultimately, as discussed in the comment https://github.com/microsoft/onnxruntime/pull/27099#issuecomment-3791673372, I expect the NchwcFloatKernel needs optimizations before considering that.

onnxruntime/core/mlas/lib/aarch64/SconvKernelNeon.S
```diff
@@ -0,0 +1,836 @@
...
+// Implements the convolution micro-kernel for the NCHW input format.
+// -----------------------------------------------------------------------------
+
+        FUNCTION_ENTRY MlasConvNchwFloatKernelNeonAsm
+
+        // prologue -------------------------------------------------------------
+        stp     x19,x20,[sp,#-.LFrame_SavedRegs]!
+        stp     x21,x22,[sp,#.LFrame_x21_x22]
+        stp     x23,x24,[sp,#.LFrame_x23_x24]
+        stp     x25,x26,[sp,#.LFrame_x25_x26]
```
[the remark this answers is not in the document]
milpuz01: This was addressed in https://github.com/microsoft/onnxruntime/pull/27099/commits/5cba128f0f6e497bf5654f858ea808826dab6281

onnxruntime/core/mlas/lib/platform.cpp
```diff
@@ -571,7 +571,15 @@ Return Value:
     this->EltwiseDispatch = &MlasEltwiseDispatchNeon;
 
 #if defined(MLAS_USE_ARM_NEON_NCHWC)
-    this->ConvNchwFloatKernel = MlasConvNchwFloatKernelNeon;
+    // Prefer the hand written micro-kernel for the NCHW convolution path.  It
+    // offers a tighter schedule and a specialised two-output inner loop that
+    // reduces pressure on the memory system compared
```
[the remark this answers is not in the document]
milpuz01: This was addressed in https://github.com/microsoft/onnxruntime/pull/27099/commits/5cba128f0f6e497bf5654f858ea808826dab6281

onnxruntime/core/mlas/lib/platform.cpp
```diff
@@ -571,7 +571,15 @@ Return Value:
...
+    this->ConvNchwFloatKernel = MlasConvNchwFloatKernelNeonAsm;
+    // Prefer the hand written AArch64 micro-kernel for pointwise convolution
+    // as it computes multiple output positions at once and significantly
+    // reduces memory traffic. The AArch64 assembly kernel is picked up by
+    // heuristics in platform.cpp to avoid regressions on small convolutions.
```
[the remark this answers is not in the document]
milpuz01: This was addressed in https://github.com/microsoft/onnxruntime/pull/27099/commits/5cba128f0f6e497bf5654f858ea808826dab6281

onnxruntime/core/mlas/lib/aarch64/SconvPointwiseKernelNeon.S
```diff
@@ -0,0 +1,441 @@
...
+        stp     x28,x19,[sp,#-16]!
+
+        mov     x14,#0                     // current filter set
+        cbz     x5,.Lpw_exit               // nothing to do
+
+.Lpw_filter_loop:
+        // Compute the base pointers for this filter block.
+        madd    x16,x14,x8,x2              // output pointer for this filter
+        madd    x17,x14,x7,x1              // filter pointer for this filter
+        add     x18,x10,x14,lsl #6         // bias pointer (if used)
```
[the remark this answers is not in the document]
milpuz01: This was addressed in https://github.com/microsoft/onnxruntime/pull/27099/commits/5cba128f0f6e497bf5654f858ea808826dab6281

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
```diff
@@ -304,6 +304,11 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
 
   Status GetFileSizeIfValid(const std::string& filepath, size_t& file_size);
 
+  Status ReadContextBinIfValid(const std::string& context_bin_filepath,
```
Rohanjames1997: Nit: Need to remove these lines as they are already present. Probably were added during rebase & conflict resolution.

milpuz01: Thanks for spotting this. Fixed in https://github.com/microsoft/onnxruntime/pull/27099/commits/c54266b87fc6cf996cce213bd4650ba852c39c16

cmake/onnxruntime_mlas.cmake
```diff
@@ -460,6 +466,8 @@ else()
           ${MLAS_SRC_DIR}/aarch64/QgemmS8S8KernelSdot.S
           ${MLAS_SRC_DIR}/aarch64/SgemmKernelNeon.S
           ${MLAS_SRC_DIR}/aarch64/SgemvKernelNeon.S
+          ${MLAS_SRC_DIR}/aarch64/SconvDepthwiseKernelNeon.S
```
hariharans29: Is the inclusion of the files here an accidental regression ? I saw that they were removed in some commit but they show up here again ?

cmake/onnxruntime_mlas.cmake
```diff
@@ -327,6 +327,12 @@ function (setup_arm_neon_nchwc)
    ${MLAS_SRC_DIR}/sconv_nchwc_kernel_neon.h
    ${MLAS_SRC_DIR}/sconv_nchwc_kernel_neon.cpp
    ${MLAS_SRC_DIR}/spool_nchwc_kernel_neon.cpp
+   # Hand written AArch64 micro-kernel for NCHW convolution.  Using a
+   # separate assembly file allows tighter control over register allocation
+   # and avoids the overhead of C++/intrinsics based code generation.
+   ${MLAS_SRC_DIR}/aarch64/SconvKernelNeon.S
```
hariharans29: Wait - Can MSVC handle .S assembly files ? Should this be enabled on non-Windows platforms only (or come in with equivalent .asm files for MSVC/Windows) ?

hariharans29:
This change looks good to me. Thanks. Can you please address remaining comments (Copilot + mine) so that it can be merged ?

FYI - I have called out the NCHWc layout support on ARM in the 1.24 release notes, so that the community can give a try and share feedback/issues if any - https://github.com/microsoft/onnxruntime/releases. CC: @Rohanjames1997

aviralagrawal:
Nice 🚀 
Thanks @hariharans29!
I checked the contributor list in the 1.24.1 release and I think it's wrong. I don't see @Rohanjames1997 and a few other names in the list. And I see a few names that did not contribute to this release. What can explain this? 🙂

hariharans29:
> Nice 🚀 Thanks @hariharans29! I checked the contributor list in the 1.24.1 release and I think it's wrong. I don't see @Rohanjames1997 and a few other names in the list. And I see a few names that did not contribute to this release. What can explain this? 🙂

Thanks for bringing this to my attention. I am not sure how the contributors list is generated myself. I ll pass along the information for folks to take a look. Meanwhile, I have added Rohan manually to the list and apologies. 

EDIT: Filed as issue for tracking: https://github.com/microsoft/onnxruntime/issues/27274

hariharans29:
> > Nice 🚀 Thanks @hariharans29! I checked the contributor list in the 1.24.1 release and I think it's wrong. I don't see @Rohanjames1997 and a few other names in the list. And I see a few names that did not contribute to this release. What can explain this? 🙂
> 
> Thanks for bringing this to my attention. I am not sure how the contributors list is generated myself. I ll pass along the information for folks to take a look. Meanwhile, I have added Rohan manually to the list and apologies.
> 
> EDIT: Filed as issue for tracking: #27274



> > Nice 🚀 Thanks @hariharans29! I checked the contributor list in the 1.24.1 release and I think it's wrong. I don't see @Rohanjames1997 and a few other names in the list. And I see a few names that did not contribute to this release. What can explain this? 🙂
> 
> Thanks for bringing this to my attention. I am not sure how the contributors list is generated myself. I ll pass along the information for folks to take a look. Meanwhile, I have added Rohan manually to the list and apologies.
> 
> EDIT: Filed as issue for tracking: #27274

FYI - https://github.com/microsoft/onnxruntime/issues/27274#issuecomment-3865091814

cmake/onnxruntime_mlas.cmake
```diff
@@ -460,6 +466,8 @@ else()
           ${MLAS_SRC_DIR}/aarch64/QgemmS8S8KernelSdot.S
           ${MLAS_SRC_DIR}/aarch64/SgemmKernelNeon.S
           ${MLAS_SRC_DIR}/aarch64/SgemvKernelNeon.S
+          ${MLAS_SRC_DIR}/aarch64/SconvDepthwiseKernelNeon.S
+          ${MLAS_SRC_DIR}/aarch64/SconvPointwiseKernelNeon.S
```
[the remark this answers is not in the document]
milpuz01: Fixed in https://github.com/microsoft/onnxruntime/pull/27099/commits/040fe49444809e08561c5e53afced054d4625476

onnxruntime/core/mlas/lib/aarch64/SconvDepthwiseKernelNeon.S
```diff
@@ -0,0 +1,239 @@
...
+SPDX-License-Identifier: MIT
+
+Module Name:
+
+    SconvDepthwiseFloatKernelNeon.S
+
+Abstract:
+
+    Optimized AArch64 assembly implementation of the depthwise convolution
+    micro-kernel used by the NCHWc single precision path.
```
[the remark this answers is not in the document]
milpuz01: Fixed in https://github.com/microsoft/onnxruntime/pull/27099/commits/040fe49444809e08561c5e53afced054d4625476

onnxruntime/core/mlas/lib/aarch64/SconvDepthwiseKernelNeon.S
```diff
@@ -0,0 +1,239 @@
...
+
+    This kernel performs the following optimisations:
+      * Produce a fast path for interior output positions where all input
+        accesses are guaranteed to be in-bounds and can be loaded with a pair
+        of 128-bit loads.
+      * When an output position touches padding, only the affected 4-wide
+        lanes are checked individually and loaded; others are zeroed.  This
+        mirrors the behavior of the C++ helper LoadInputVectorWithBounds.
+        mirrors the behaviour of the C++ helper LoadInputVectorWithBounds.
```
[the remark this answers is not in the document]
milpuz01: Removed extra line in https://github.com/microsoft/onnxruntime/pull/27099/commits/040fe49444809e08561c5e53afced054d4625476

cmake/onnxruntime_mlas.cmake
milpuz01: No, MSVC cannot handle .S assembly files. I have put in https://github.com/microsoft/onnxruntime/pull/27099/commits/040fe49444809e08561c5e53afced054d4625476 to only enable this on non-Windows platforms.

cmake/onnxruntime_mlas.cmake
milpuz01: Yes, it is accidental merge conflict that was also pointed out by Copilot. It was removed in https://github.com/microsoft/onnxruntime/pull/27099/commits/040fe49444809e08561c5e53afced054d4625476

milpuz01:
> Can you please address remaining comments (Copilot + mine) so that it can be merged ?

Thanks @hariharans29. I have address all the remaining comments in: https://github.com/microsoft/onnxruntime/pull/27099/commits/040fe49444809e08561c5e53afced054d4625476.

> FYI - I have called out the NCHWc layout support on ARM in the 1.24 release notes, so that the community can give a try and share feedback/issues if any - https://github.com/microsoft/onnxruntime/releases. CC: @Rohanjames1997

That is awesome. Thanks for doing it!

onnxruntime/core/mlas/lib/platform.cpp
```diff
@@ -571,7 +571,20 @@ Return Value:
...
+    // Use the AArch64 assembly implementation on non-Windows platforms.
+#if !defined(_WIN32)
+    // Prefer the hand written micro-kernel for the NCHW convolution path. It
+    // offers a tighter schedule and a specialised two-output inner loop that
+    // reduces pressure on the memory system compared to the generic kernel.
+    this->ConvNchwFloatKernel = MlasConvNchwFloatKernelNeonAsm;
+#else
     this->ConvNchwFloatKernel = MlasConvNchwFloatKernelNeon;
+#endif
+    // Prefer the hand written AArch64 micro-kernel for pointwise convolution
```
hariharans29: Comment out of place ?

onnxruntime/core/mlas/lib/snchwc.cpp
```diff
@@ -1047,7 +1060,13 @@ struct MLAS_NCHWC_CONV_DEPTHWISE_ALGORITHM : MLAS_NCHWC_CONV_ALGORITHM
...
             //
 
-            Kernel(Input + BlockSize * (ih * InputWidth - PaddingLeftX), filter,
+            MLAS_CONV_DEPTHWISE_FLOAT_KERNEL* KernelToUse = Kernel;
+#if defined(MLAS_TARGET_ARM64) && defined(MLAS_USE_ARM_NEON_NCHWC) && !defined(_WIN32)
+            if (OutputWidth >= 4) {
+                KernelToUse = KernelFast;
+            }
+#endif
+            KernelToUse(Input + BlockSize * (ih * InputWidth - PaddingLeftX), filter,
```
hariharans29, line 1072: Nit: Formating ?

onnxruntime/core/mlas/lib/platform.cpp
milpuz01: Yes, it is. I moved it out to snchwc.cpp in https://github.com/microsoft/onnxruntime/pull/27099/commits/2d0585378a9041c1aa9a2ea065ebf5d971f3abc0 where the reference to assembler is used and also where heuristic is implemented.

onnxruntime/core/mlas/lib/snchwc.cpp
milpuz01: I haven't changed anything here in terms of formatting apart from using `KernelToUse` instead of `Kernel`. What is the formatting that needs correcting?

hariharans29: My bad - for some reason when I viewed it previously, the formating seemed off.

hariharans29:
Can you please rebase with main ?

milpuz01:
> Can you please rebase with main ?

Just rebased.

hariharans29 (approved):

Issue #27079, opened 2026-01-20
[CUDA] Qwen3 Model quality regression


jixiongdeng:
### Describe the issue

Qwen3 0.6B models generate gibberish outputs on `CUDA` backend on main branch.
After investigation, the issue was introduced from an earlier commit:  `39d8520b9bfe16217d56be38293fc82556d67b5b`
It is reproducible on both unquantized model(fp16) and rtn 4bit model.
```
#fp16
Prompt: 1+1=?

1+1+1+1+10+11+100

systemsystemsystem


#u4 rtn
Prompt: 1+1=?

1+1+1+1+1+1+1+1+1+1+1+1=?

Let? 1+1+11+1+1+1+1+1. Let 1+1+1+1+11+1+11+11+1+11+1+1+1+1+1+1+1+1+1+1+1+1+
```

The last work commit is: `5b88e4ed6778615113ad4831b160ec52e0d5ac71` 

### To reproduce

Models are generated from onnxruntime-genai's model builder. fp16 and int4.

I built onnxruntime-gpu with following command:
```
./build.sh --config Release --build_wheel --use_cuda --cuda_home /usr/local/cuda --cudnn_home /usr/local/cuda --parallel --skip_tests

pip install onnxruntime/build/Linux/Release/dist/onnxruntime_gpu-*.whl --force-reinstall
```

Prompt&Answers are generated with og.generator. 

### Urgency

This commit is included in ORT1.24 release. Qwen3 are popular open source models. 

### Platform

Linux

### OS Version

6.6.117.1-1.azl3

### ONNX Runtime Installation

Built from Source

### ONNX Runtime Version or Commit ID

1817f4a81072c49dea2984853edca412674d66fb

### ONNX Runtime API

Python

### Architecture

X64

### Execution Provider

CUDA

### Execution Provider Library Version

CUDA 12.8

apsonawane:
Hello @jixiongdeng, what version of onnxruntime-genai are you using?

titaiwangms:
I can repro from main. Not the exact output, but it does output something weird, while 1.23.2 ort output fine and perfectly.

I confirm the commit 5b88e4ed6778615113ad4831b160ec52e0d5ac71 working. But the commit 36017adcf6d25cc4d4f08ab4ef30ad04a4f06c7b before 1817f4a81072c49dea2984853edca412674d66fb does not work already, so I am not sure if the PR is the cause though (also I don't see it's included in rel-1.24.0).

[...]

PR #27108, opened 2026-01-22
Fix GQA Parity

tianleiwu:
Fix [#27079](https://github.com/microsoft/onnxruntime/issues/27079) - Qwen3 model quality regression on CUDA backend.
### Root Cause Analysis
The parity issue was caused by **buffer pointer misconfiguration** in the GQA (Group Query Attention) QKV preprocessing pipeline. The original implementation used multiple separate kernels for:
1. Unpacking packed QKV tensor
2. Applying RoPE (Rotary Position Embedding) to Q and K 
3. Appending K/V to cache
This multi-kernel approach created opportunities for misconfiguration:
- Buffers were allocated but not properly used
- Pointers could reference memory that was not yet allocated or initialized
- Buffer sharing logic was fragmented across different code paths
### Solution
Consolidate QKV preprocessing into a **single fused kernel** (`UnpackRoPEAppend`) that performs all operations in one pass:
1. **Unified kernel design**: A single kernel handles unpacking, RoPE application, and cache append operations
2. **Simplified buffer management**: The new `PrepareQKV` function clearly manages buffer allocation and ensures proper initialization
3. **Explicit past-to-present cache copy**: When `past_present_share_buffer` is false, explicitly copy past KV cache to present buffer before appending new tokens
4. **Zero-initialization for non-shared buffers**: Clear present KV buffers when not sharing with past to ensure deterministic output
### Changes Summary
| File | Changes |
|------|---------|
| [group_query_attention_qkv.cuh](cci:7://file:///home/tlwu/onnxruntime/onnxruntime/contrib_ops/cuda/bert/group_query_attention_qkv.cuh:0:0-0:0) | New fused `UnpackRoPEAppend` kernel with shared memory optimization for non-interleaved RoPE |
| `group_query_attention_impl.cu` | New `PrepareQKV` helper function that orchestrates buffer setup and kernel launch |
| `group_query_attention.cc` | Simplified operator logic by delegating QKV prep to unified helper |
| `test_gqa.py` | Enhanced test coverage for various QKV configurations |
### Key Improvements
- **Reduced kernel launches**: From 4-5 separate kernel calls to a single fused kernel
- **Better memory safety**: All buffer pointers are validated in a single location
- **Improved RoPE handling**: Uses shared memory for efficient non-interleaved RoPE computation
- **Deterministic output**: Explicit buffer initialization ensures consistent results across runs
- **Compatible with quantized KV cache**: The new preprocessing kernel design supports future quantization work
### Testing
- All existing GQA unit tests pass
- Verified Qwen3 model no longer produces gibberish output
- Tested both fp16/bf16 and various head configurations

### DIFF
--- a/onnxruntime/contrib_ops/cuda/bert/group_query_attention_qkv.cuh
+++ b/onnxruntime/contrib_ops/cuda/bert/group_query_attention_qkv.cuh
@@ -2,9 +2,6 @@
 // Licensed under the MIT License.
 #pragma once
 
-// Enable quantized KV cache support for INT8/INT4
-#define KV_QUANT_SUPPORTED 0
-
 #include <cuda_fp16.h>
 
 #include "contrib_ops/cuda/bert/group_query_attention_impl.h"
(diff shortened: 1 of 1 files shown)


onnxruntime/contrib_ops/cuda/bert/attention_strided_copy.cu
```diff
@@ -180,6 +180,18 @@ template Status LaunchStridedCopy<half>(
     half* out, LongLong4 out_strides,
     int max_threads_per_block);
 
+template Status LaunchStridedCopy<__nv_bfloat16>(
+    cudaStream_t stream,
+    const __nv_bfloat16* in, int4 in_shape, LongLong4 in_strides,
+    __nv_bfloat16* out, LongLong4 out_strides,
+    int max_threads_per_block);
+
```
kunal-vaishnavi: Why do we need both the hardware bf16 representation (`__nv_bfloat16`) and higher-level bf16 representation (`BFloat16`)?

Also, I only see `LaunchStridedCopy` referenced in the `attention_impl.h` header file and in this file. Where is `LaunchStridedCopy` called?

onnxruntime/contrib_ops/cuda/bert/group_query_attention.cc
```diff
@@ -39,7 +43,6 @@ REGISTER_KERNEL_TYPED(MLFloat16)
 REGISTER_KERNEL_TYPED(BFloat16)
 
 constexpr const char* kDisableFlashDecode = "ORT_DISABLE_FLASH_DECODE";
-constexpr const char* kDisableFusedKv = "ORT_DISABLE_FUSED_KV";
```
kunal-vaishnavi, line 42: Why remove?

onnxruntime/contrib_ops/cuda/bert/group_query_attention_qkv.cuh
```diff
@@ -0,0 +1,283 @@
...
+  constexpr int vector_bytes = sizeof(float4);
+  constexpr int element_bytes = sizeof(T);
+  constexpr int elements_per_vector = vector_bytes / element_bytes;
+
+  if (head_size % elements_per_vector != 0) {
+    return ORT_MAKE_STATUS(ONNXRUNTIME, FAIL, "Head size must be divisible by ", elements_per_vector, " for vectorized GQA kernel.");
+  }
+
+  if (sequence_length > 65535) {
+    return ORT_MAKE_STATUS(ONNXRUNTIME, FAIL, "Sequence length ", sequence_length, " exceeds CUDA grid limit (65535).");
```
kunal-vaishnavi: For large GPUs where such a large sequence length can fit into memory, what is the fallback so that the runtime still continues?

onnxruntime/contrib_ops/cuda/bert/group_query_attention.cc
tianleiwu: We only have one fused kernel right now. So remove this option (we cannot disable the fused kernel)

onnxruntime/contrib_ops/cuda/bert/group_query_attention_qkv.cuh
tianleiwu: Updated kernel to remove the limit.

onnxruntime/contrib_ops/cuda/bert/attention_strided_copy.cu
tianleiwu: This is not needed.

onnxruntime/contrib_ops/cuda/bert/group_query_attention_qkv.cuh
```diff
@@ -0,0 +1,248 @@
...
+  const bool valid = (h < head_size);
+
+  const int q_hidden = num_heads * head_size;
+  const int k_hidden = kv_num_heads * head_size;
+  const int sequence_length = gridDim.x;
+
+  __shared__ T shared_head[MAX_HEAD_SIZE];
+
+  // Determine Head Type and Offset within hidden dimension
+  enum HeadType { QUERY,
```
kunal-vaishnavi, line 72: Should we define this enum outside of the kernel?

tianleiwu: It's fine. There shall not be difference on compiled code.

PR #27121, opened 2026-01-23
perftest: support plugin eps for compile_ep_context

jnagi-intel:
* Extend compile_ep_context to also support plugin eps
* Adds compile_only option to skip execution, can be used when compiling for virtual devices

compile_ep_context (physical device)
image

compile_ep_context + compile_only (virtual device)
image

### DIFF
--- a/onnxruntime/test/perftest/common_utils.cc
+++ b/onnxruntime/test/perftest/common_utils.cc
@@ -90,6 +90,116 @@ std::vector<char*> CStringsFromStrings(std::vector<std::string>& utf8_args) {
   return utf8_argv;
 }
 
+void AppendPluginExecutionProviders(Ort::Env& env,
+                                    Ort::SessionOptions& session_options,
+                                    const PerformanceTestConfig& test_config) {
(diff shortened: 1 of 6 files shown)


onnxruntime/test/perftest/common_utils.cc
```diff
@@ -90,6 +90,116 @@ std::vector<char*> CStringsFromStrings(std::vector<std::string>& utf8_args) {
   return utf8_argv;
 }
 
+void AppendPluginExecutionProviders(Ort::Env& env,
+                                    Ort::SessionOptions& session_options,
+                                    const PerformanceTestConfig& test_config) {
+  if (test_config.registered_plugin_eps.empty()) {
```
adrianlizarraga, line 96: qq: Could you please confirm that this code was moved from its original location without any major alterations? thanks

jnagi-intel: Yes, there is no alteration to this code. it was simply moved so it can be reused

jnagi-intel:
@adrianlizarraga CI failure seems unrelated to the change in the PR - 

image

PR #27129, opened 2026-01-24
Support boolean attention mask in Attention(23) CUDA

titaiwangms:
Follow up https://github.com/microsoft/onnxruntime/pull/27030

This PR adds boolean to float conversion to help bool `attn_mask` to become `attn_bias`. `mask_filter_value (-10000.0f)` is referenced in the value conversion: https://github.com/microsoft/onnxruntime/blob/main/docs/ContribOperators.md#com.microsoft.MultiHeadAttention

---

This pull request adds support for boolean attention masks in the CUDA Attention operator, allowing models to use boolean masks directly on GPU. The main change is the introduction of a CUDA kernel to convert boolean masks to the required float bias format, enabling previously unsupported test cases and improving compatibility with other attention implementations.

### CUDA Attention operator: Boolean mask support

* Added a new CUDA kernel (`attention_mask_convert.cu`, `attention_mask_convert.h`) that converts boolean attention masks to float bias values, mapping `true` to `0.0f` and `false` to `-10000.0f` by default. [[1]](diffhunk://#diff-be265a398893445e439e3cf0c1517e5220bd34eedcb95b60690c12c0063ef9d6R1-R53) [[2]](diffhunk://#diff-09494a2c2711553915360102d935a8ef0d61b49811dab4bef3692df93c533657R1-R23)
* Modified `attention.cc` to use the new kernel: if a boolean mask is detected, it is converted to a float bias buffer and used in the attention computation. [[1]](diffhunk://#diff-0701e4cc6d4951894ae1a60f35c1e6c0f69ba7595f896a23c8f5ed7265eab4ffR131-R151) [[2]](diffhunk://#diff-0701e4cc6d4951894ae1a60f35c1e6c0f69ba7595f896a23c8f5ed7265eab4ffR237-R243) [[3]](diffhunk://#diff-0701e4cc6d4951894ae1a60f35c1e6c0f69ba7595f896a23c8f5ed7265eab4ffR7)

### Test and filter updates

* Enabled CUDA tests for boolean attention masks by removing their exclusion from the test series filter file (`onnx_backend_test_series_filters.jsonc`).
* Updated unit tests to cover edge cases for boolean masks, including cases where all tokens are blocked, and ensured CUDA tests are enabled for these scenarios. [[1]](diffhunk://#diff-3ff6dfa2ce407ae0073009174c37d1756509e8bbc434dee7c44cd55a996bb777R451) [[2]](diffhunk://#diff-3ff6dfa2ce407ae0073009174c37d1756509e8bbc434dee7c44cd55a996bb777L620-R621)

### DIFF
--- a/onnxruntime/core/providers/cuda/llm/attention.cc
+++ b/onnxruntime/core/providers/cuda/llm/attention.cc
@@ -208,7 +208,6 @@ Status Attention<T>::ComputeInternal(OpKernelContext* context) const {
     ORT_THROW("softmax_precision is not supported yet in Attention op (CUDA).");
   }
 
-  // TODO(titaiwang): Continue on these parameters
   // Construct AttentionData to pass to QkvToContext
   typedef typename ToCudaType<T>::MappedType CudaT;
   onnxruntime::contrib::cuda::AttentionData<CudaT> data;
@@ -247,12 +246,12 @@ Status Attention<T>::ComputeInternal(OpKernelContext* context) const {
   }
   data.qkv_format = contribop_parameters.qkv_format;
 
-  // TODO: Determine which kernel to use (Flash Attention, Memory Efficient Attention, etc.)
   // For now, set flags to false and let QkvToContext use the unfused path
   data.use_flash_attention = false;
   data.use_memory_efficient_attention = false;
   data.fused_runner = nullptr;
   data.fused_cross_attention_kernel = nullptr;
+  data.kernel_type = onnxruntime::contrib::AttentionKernelType::AttentionKernel_Unfused;
 
   // Allocate workspace for Q, K, V processing and scratch buffer
   const bool no_qkv_workspace = onnxruntime::contrib::cuda::NoQkvWorkspace(contribop_parameters, data);
--- a/onnxruntime/contrib_ops/cuda/bert/attention_prepare_qkv.cu
+++ b/onnxruntime/contrib_ops/cuda/bert/attention_prepare_qkv.cu
@@ -258,7 +258,9 @@ Status PrepareQkv_MHA_NoPast(contrib::AttentionParameters& parameters,
   assert(data.past_value == nullptr);
   assert(data.present_key == nullptr);
   assert(data.present_value == nullptr);
-  assert(!parameters.is_unidirectional);
+  // Note: is_unidirectional (causal) is supported by flash attention, memory efficient attention,
+  // cuDNN flash attention, and unfused kernel. TRT fused runner is only used when !is_unidirectional
+  // (enforced in MultiHeadAttention::ComputeInternal).
   assert(data.has_qkv_workspace == !NoQkvWorkspace_MHA_NoPast(data));
 
   if (parameters.qkv_format == AttentionQkvFormat::Q_K_V_BSNH) {
--- a/onnxruntime/contrib_ops/webgpu/moe/moe.h
+++ b/onnxruntime/contrib_ops/webgpu/moe/moe.h
@@ -3,6 +3,8 @@
 
 #pragma once
 
+#include <limits>
+
 #include "core/providers/webgpu/program.h"
 #include "core/providers/webgpu/webgpu_kernel.h"
 
@@ -31,7 +33,7 @@ class MoE : public WebGpuKernel {
     activation_alpha_ = static_cast<float>(info.GetAttrOrDefault<float>("activation_alpha", 1.0));
     activation_beta_ = static_cast<float>(info.GetAttrOrDefault<float>("activation_beta", 1.0));
     swiglu_fusion_ = static_cast<int>(info.GetAttrOrDefault<int64_t>("swiglu_fusion", 0));
-    swiglu_limit_ = info.GetAttrOrDefault<float>("swiglu_limit", 0);
+    swiglu_limit_ = info.GetAttrOrDefault<float>("swiglu_limit", std::numeric_limits<float>::infinity());
     k_ = static_cast<int>(info.GetAttrOrDefault<int64_t>("k", 4));
     normalize_routing_weights_ = info.GetAttrOrDefault<int64_t>("normalize_routing_weights", 0) == 1;
     use_sparse_mixer_ = info.GetAttrOrDefault<int64_t>("use_sparse_mixer", 0) == 1;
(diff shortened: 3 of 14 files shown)


onnxruntime/test/providers/cpu/llm/attention_op_test.cc
```diff
@@ -448,6 +448,7 @@ TEST(AttentionTest, Attention4DDefault) {
   );
 }
 
+// Edge case where attention mask blocks all tokens (all false)
```
titaiwangms, line 451: @xadupre I can't pass this. IIUC, this means attn_mask is blocking every token?

onnxruntime/core/providers/cuda/llm/attention.cc
```diff
@@ -127,10 +128,27 @@ Status Attention<T>::ComputeInternal(OpKernelContext* context) const {
...
     if (attn_mask->IsDataType<bool>()) {
-      ORT_THROW("Boolean attn_mask is not supported yet in Attention op (CUDA).");
+      // Allocate space for converted mask
+      size_t mask_size = SafeInt<size_t>(attn_mask->Shape().Size());
+      converted_bool_mask_buffer = GetScratchBuffer<T>(mask_size, context->GetComputeStream());
+
+      // Launch CUDA kernel to convert: true->0.0f, false->mask_filter_value
+      // Note: mask_filter_value default is -10000.0f, same as MultiHeadAttention op
+      typedef typename ToCudaType<T>::MappedType CudaT;
+      const float mask_filter_value = -10000.0f;
```
tianleiwu: `const T mask_filter_value  = std::numeric_limits<T>::lowest();`
You can find the definition in include/onnxruntime/core/framework/float16.h

onnxruntime/core/providers/cuda/llm/attention_mask_convert.cu
```diff
@@ -0,0 +1,53 @@
...
+}
+
+// Explicit template instantiations
+template Status LaunchConvertBoolMaskToFloatBias<float>(
+    cudaStream_t stream, float* output, const bool* input, int64_t size, float mask_filter_value);
+
+template Status LaunchConvertBoolMaskToFloatBias<half>(
+    cudaStream_t stream, half* output, const bool* input, int64_t size, half mask_filter_value);
+
+#if !defined(DISABLE_FLOAT8_TYPES)
```
tianleiwu: No need #if !defined(DISABLE_FLOAT8_TYPES)
BFloat16 is not fp8 type

onnxruntime/core/providers/cuda/llm/attention.cc
```diff
@@ -216,7 +234,12 @@ Status Attention<T>::ComputeInternal(OpKernelContext* context) const {
   // Set additional fields
   data.bias = nullptr;  // New Attention op doesn't have bias
   if (nullptr != attn_mask) {
-    data.attention_bias = reinterpret_cast<const CudaT*>(attn_mask->Data<T>());
+    // Use converted buffer if boolean mask was converted, otherwise use original mask
+    if (converted_bool_mask_buffer) {
+      data.attention_bias = reinterpret_cast<const CudaT*>(converted_bool_mask_buffer.get());
+    } else {
+      data.attention_bias = reinterpret_cast<const CudaT*>(attn_mask->Data<T>());
```
tianleiwu, line 244: `attn_mask->Data<T>()` will raise exception if attn_mask is bool type. Did you add test case for this?

tianleiwu:
Instead of converting mask to bias, it might be better to change some kernel to template mask/bias.
For example,
```
template <typename T>
Status ComputeSoftmax(cudaStream_t stream, const int all_sequence_length, const int sequence_length,
const int batch_size, const int num_heads, const T* attn_bias,
const bool broadcast_attn_bias_dim_0, const bool broadcast_attn_bias_dim_1,
T* input, T* output, bool causal);
```

We can change it to 
```
template <typename T, typename M>
Status ComputeSoftmax(cudaStream_t stream, const int all_sequence_length, const int sequence_length,
const int batch_size, const int num_heads, const M* attn_mask_or_bias,
const bool broadcast_attn_mask_or_bias_dim_0, const bool broadcast_attn_mask_or_biass_dim_1,
T* input, T* output, bool causal);
```

If M is bool, it is mask. When M=T, it is attention bias.

titaiwangms:
> Instead of converting mask to bias, it might be better to change some kernel to template mask/bias. For example,
> 
> ```
> template <typename T>
> Status ComputeSoftmax(cudaStream_t stream, const int all_sequence_length, const int sequence_length,
> const int batch_size, const int num_heads, const T* attn_bias,
> const bool broadcast_attn_bias_dim_0, const bool broadcast_attn_bias_dim_1,
> T* input, T* output, bool causal);
> ```
> 
> We can change it to
> 
> ```
> template <typename T, typename M>
> Status ComputeSoftmax(cudaStream_t stream, const int all_sequence_length, const int sequence_length,
> const int batch_size, const int num_heads, const M* attn_mask_or_bias,
> const bool broadcast_attn_mask_or_bias_dim_0, const bool broadcast_attn_mask_or_biass_dim_1,
> T* input, T* output, bool causal);
> ```
> 
> If M is bool, it is mask. When M=T, it is attention bias.

Thanks for the suggestion. That will need to change the existing softmax kernels though. Are you suggesting because the performance will be better in this way?

onnxruntime/core/providers/cuda/llm/attention.cc
titaiwangms: Added comment and exception.

onnxruntime/core/providers/cuda/llm/attention.cc
titaiwangms: Done. I am using fp32 only though. I think the kernel should be able to handle it? Or I should use T (following the dtype of the op inputs)?

titaiwangms:
> Instead of converting mask to bias, it might be better to change some kernel to template mask/bias. For example,
> 
> ```
> template <typename T>
> Status ComputeSoftmax(cudaStream_t stream, const int all_sequence_length, const int sequence_length,
> const int batch_size, const int num_heads, const T* attn_bias,
> const bool broadcast_attn_bias_dim_0, const bool broadcast_attn_bias_dim_1,
> T* input, T* output, bool causal);
> ```
> 
> We can change it to
> 
> ```
> template <typename T, typename M>
> Status ComputeSoftmax(cudaStream_t stream, const int all_sequence_length, const int sequence_length,
> const int batch_size, const int num_heads, const M* attn_mask_or_bias,
> const bool broadcast_attn_mask_or_bias_dim_0, const bool broadcast_attn_mask_or_biass_dim_1,
> T* input, T* output, bool causal);
> ```
> 
> If M is bool, it is mask. When M=T, it is attention bias.

I think a better idea would be we reuse this kernel (https://github.com/microsoft/onnxruntime/pull/27082/changes#diff-8aa9a15a92d7dc138346dce5de055911895d940ba2183b4ba45bd95ac0e5bfc9R42) to include `key_padding_mask` conversion as well. It only has 1 difference from `seqlens_k` if I understand correctly.

titaiwangms:
Convert this to draft to avoid the huge conflict with https://github.com/microsoft/onnxruntime/pull/27082. We should land that one first.

titaiwangms:
Replaced by #27428

Issue #26889, opened 2026-01-01
could coloredlogs be replaced per something not using pyreadline3 ?


stonebig:
coloredlogs is a problem, on Windows:
- it makes  python3.13 console crash on windows, because it embarks pyreadline3
- pyreadline3 is no more friendly with python-3.13+

Can it be removed/replaced ?

 
<img width="1706" height="536" alt="Image" src="https://github.com/user-attachments/assets/688e7dd8-1fdb-4e7b-84c1-aca78d626da3" />

tianleiwu:
Sure. We can make it optional.

stonebig:
well, the patch to do in pyreadline3 is so simple, I hope someone from pyreadline3 will publish it
https://github.com/pyreadline3/pyreadline3/pull/44/files

stonebig:
the person who founds the bug is pretty negative on pyreadline3 readline.py... as it will be increasingly disruptive as cpython evolves.

stonebig:
it's solved now

PR #27135, opened 2026-01-24
remove coloredlogs

tianleiwu:
See related issues: https://github.com/microsoft/onnxruntime/issues/26889


onnxruntime/python/tools/transformers/optimizer.py
```diff
@@ -562,12 +561,11 @@ def _parse_arguments():
...
     if verbose:
-        coloredlogs.install(
-            level="DEBUG",
-            fmt="[%(filename)s:%(lineno)s - %(funcName)20s()] %(message)s",
+        logging.basicConfig(
+            format="[%(filename)s:%(lineno)s - %(funcName)20s()] %(message)s", level=logging.DEBUG, force=True
         )
     else:
-        coloredlogs.install(fmt="%(funcName)20s: %(message)s")
+        logging.basicConfig(format="%(funcName)20s: %(message)s", level=logging.INFO, force=True)
```
kunal-vaishnavi, line 568: One of the nice subtle features about the ORT transformer optimizer is the yellow text that says whether a model has been fully optimized or not. Will that behavior still remain after these changes?

kunal-vaishnavi (approved):

onnxruntime/python/tools/transformers/optimizer.py
tianleiwu: I think currently it use logger.info for that so no yellow text: https://github.com/microsoft/onnxruntime/blob/68615264f884a2a560fcdd2f2b6403852245fe8b/onnxruntime/python/tools/transformers/optimizer.py#L608-L610

PR #27151, opened 2026-01-26
[WebGPU EP] Reduce duplicated code in `MatMulReadFnSource()`

Jiawei-Shao:
### Description
Previously in `MatMulReadFnSource()` we use duplicated code to read data from two inputs `a` and `b`. This patch implements another overload of `MatMulReadFnSource()` to only read data from one input to reduce duplicated code and get ready for further use.

### DIFF
--- a/onnxruntime/core/providers/webgpu/math/gemm_utils.cc
+++ b/onnxruntime/core/providers/webgpu/math/gemm_utils.cc
@@ -129,8 +129,8 @@ void MatMulReadFnSource(ShaderHelper& shader,
                         const ShaderIndicesHelper* batch_dims,
                         std::string_view rows,
                         std::string_view components_per_row,
-                        bool transpose,
-                        int components) {
+                        bool transpose) {
+  const int components = input.NumComponents();
   const std::string data_type = "output_element_t";
   const std::string type_string = MakeScalarOrVectorType(components, data_type);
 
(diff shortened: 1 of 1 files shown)


Jiawei-Shao:
@qjia7 @fs-eire @guschmue PTAL, thanks!

onnxruntime/core/providers/webgpu/math/gemm_utils.cc
```diff
@@ -123,67 +123,58 @@ void InitializeLogicalWorkgroupIDAndGlobalID(ShaderHelper& shader) {
...
                                     << "}\n\n";
 }
 
+void MatMulReadFnSource(ShaderHelper& shader,
+                        const ShaderVariableHelper& a,
+                        const ShaderVariableHelper& b,
+                        const ShaderIndicesHelper* batch_dims,
+                        bool transA,
+                        bool transB,
+                        bool is_vec4) {
```
qjia7: nit:
```
int components = is_vec4 ? 4 : 1;
MatMulReadFnSource(shader, "mm_readA", a, "a", batch_dims, "uniforms.dim_a_outer", "uniforms.dim_inner", transA, components );
MatMulReadFnSource(shader, "mm_readB", b, "b", batch_dims, "uniforms.dim_inner", "uniforms.dim_b_outer", transB, components );

Jiawei-Shao: Done

PR #27174, opened 2026-01-27
[MLAS] Fix Flaky LuT GEMM Tests by Replacing Gather with Shuffle

tianleiwu:
## Problem Description
The `MatMulNBitsLutGemm` test suite, specifically `Float32_2Bits_Symmetric_256x256_BlkLen64`, was observing intermittent failures (flakiness).
The failure manifested as numerical mismatches exceeding the tolerance, suggesting non-deterministic behavior in the kernel execution.

## Root Cause Analysis
The issue was traced to the usage of `_mm256_i32gather_ps` in sqnbitgemm_lut_kernel_avx2.cpp
While the gather indices were technically calculating addresses within the bounds of the allocated buffer, gather instructions on certain AVX2 hardware implementations can exhibit non-deterministic behavior or subtle performance/prefetching artifacts when operating on specific stride patterns (in this case, gathering with a stride of 4 floats).

## Solution
This PR replaces the `_mm256_i32gather_ps` instruction with a sequence of **contiguous loads (`_mm256_loadu_ps`) followed by deterministic shuffles**.

### How it works:
1.  **Contiguous Load**: We load 4 contiguous vectors of 8 floats elements using `_mm256_loadu_ps`. This is always memory-safe and deterministic.
2.  **Deterministic Shuffle**: We apply a verified sequence of `unpack` and `permutevar8x32` instructions to rearrange these 32 linearly loaded elements into the exact same stride-4 layout that the gather instruction produced.

### Benefits:
*   **Stability**: Eliminates the hardware-dependent non-determinism of gather.
*   **Safety**: Usage of `loadu` guarantees we only touch memory within the explicit range of the 32 elements we intend to load.
*   **Correctness**: The shuffle logic was verified against the reference gather behavior using a C++ reproduction script to ensure bit-exact layout equivalence.

### Performance

Micro-benchmark on MatMulNBitsLutGemm (256x256, BlkLen=64).
Original (Gather): ~55.55 us
Fixed (Load+Shuffle): ~57.79 us
Delta: +2.24 us (~4% slower)

The slight performance regression is expected because replacing a single hardware gather instruction with a sequence of loadu, unpack, and permute instructions adds instruction count overhead. However, this is a necessary tradeoff to ensure deterministic behavior and memory safety across all AVX2 implementations.

## Verification
*   **Tests**: All 10 tests in `MatMulNBitsLutGemm` passed successfully (including the previously flaky `BlkLen64` case).

### DIFF
--- a/onnxruntime/core/mlas/lib/sqnbitgemm_lut_kernel_avx2.cpp
+++ b/onnxruntime/core/mlas/lib/sqnbitgemm_lut_kernel_avx2.cpp
@@ -188,7 +188,7 @@ get_bias_scale()
 }
 
 static inline void
-_mm256_loadu_deinterleave_32ps(const float* src, __m256& v0, __m256& v1, __m256& v2, __m256& v3)
+MlasAvx2LoaduDeinterleave32Ps(const float* src, __m256& v0, __m256& v1, __m256& v2, __m256& v3)
 {
     // Process 32 activations contiguously using loadu + shuffle.
     // This allows us to mix neighbors (src[4i], src[4i+1], src[4i+2], src[4i+3]) across lanes,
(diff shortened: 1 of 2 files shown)


vraspar:
I tested changes on 2 bit llama model, and didn't see any noticeable throughput difference

PR #27175, opened 2026-01-27
Add API GetTensorElementTypeAndShapeDataReference

adrianlizarraga:
### Description
Adds C/C++ API named `GetTensorElementTypeAndShapeDataReference` that returns an OrtValue tensor's shape and type without allocating a new buffer for the shape data.

### Motivation and Context
This new API function can be used instead of `OrtApi::GetTypeInfo()` or `OrtApi::GetTensorTypeAndShape` to decrease the number of heap allocations and thus improve inference latency for plugin EPs kernels that frequently retrieve tensor shapes during inference. (e.g., WebGPU plugin EP)

### DIFF
--- a/onnxruntime/core/framework/tensor_type_and_shape.cc
+++ b/onnxruntime/core/framework/tensor_type_and_shape.cc
@@ -362,7 +362,7 @@ ORT_API_STATUS_IMPL(OrtApis::GetTensorElementTypeAndShapeDataReference, _In_ con
   }
 
   *elem_type = type;
-  *shape_data = shape_span.data();
+  *shape_data = shape_span.empty() ? nullptr : shape_span.data();
   *shape_data_count = shape_span.size();
   return nullptr;
   API_IMPL_END


include/onnxruntime/core/session/onnxruntime_cxx_inline.h
```diff
@@ -2387,6 +2387,14 @@ inline const R* ConstValueImpl<T>::GetSparseTensorValues() const {
 
 #endif
 
+template <typename T>
+Ort::Status ConstValueImpl<T>::GetTensorElementTypeAndShape(ONNXTensorElementDataType& elem_type,
+                                                            const int64_t*& shape_data,
+                                                            size_t& shape_data_count) const {
+  Ort::Status status{GetApi().Value_GetTensorElementTypeAndShape(this->p_, &elem_type, &shape_data, &shape_data_count)};
```
fs-eire: use `ThrowOnError` to keep consistent with other implementation.

onnxruntime/core/framework/tensor_type_and_shape.cc
```diff
@@ -310,6 +310,64 @@ std::unique_ptr<OrtTensorTypeAndShapeInfo> OrtTensorTypeAndShapeInfo::GetTensorS
...
 }
 
+ORT_API_STATUS_IMPL(OrtApis::Value_GetTensorElementTypeAndShape, _In_ const OrtValue* value,
+                    _Out_ ONNXTensorElementDataType* elem_type,
+                    _Outptr_ const int64_t** shape_data,
+                    _Out_ size_t* shape_data_count) {
+  API_IMPL_BEGIN
+  if (!value->IsAllocated() || (!value->IsTensor() && !value->IsSparseTensor())) {
+    return OrtApis::CreateStatus(ORT_INVALID_ARGUMENT,
+                                 "the ort_value must contain a constructed tensor or sparse tensor");
```
edgchen1:
```suggestion
                                 "Input parameter `value` must contain a constructed tensor or sparse tensor");
```

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -7217,6 +7217,29 @@ struct OrtApi {
...
+   * \param[in] value The OrtValue instance.
+   * \param[out] elem_type Output parameter set to the tensor element data type.
+   * \param[out] shape_data Output parameter set to the OrtValue instance's internal shape data array.
+   *                        Must not be released as it is owned by the OrtValue instance. This data becomes invalid
+   *                        when the OrtValue is released.
+   * \param[out] shape_data_count Output parameter set to the number of elements in `shape_data`.
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \since Version 1.25.
```
adrianlizarraga: 1.24 and move up to the correct spot

onnxruntime/core/session/onnxruntime_c_api.cc
```diff
@@ -4806,6 +4806,7 @@ static constexpr OrtApi ort_api_1_to_25 = {
 
     &OrtApis::RunOptionsEnableProfiling,
     &OrtApis::RunOptionsDisableProfiling,
+    &OrtApis::Value_GetTensorElementTypeAndShape,
```
adrianlizarraga: Move above to the 1.24 apis and update static_assert for the 1.24 APIs

include/onnxruntime/core/session/onnxruntime_c_api.h
edgchen1: is this going into 1.24?

include/onnxruntime/core/session/onnxruntime_cxx_inline.h
adrianlizarraga: done

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -7196,6 +7196,29 @@ struct OrtApi {
...
+   * \note This function is an alternative to ::GetTensorTypeAndShape() that does not allocate a new array for
+   *       the shape data. The OrtValue instance's internal shape data is returned directly.
+   *
+   * \note Returns an error if the underlying OrtValue is not a Tensor.
+   *
+   * \param[in] value The OrtValue instance.
+   * \param[out] elem_type Output parameter set to the tensor element data type.
+   * \param[out] shape_data Output parameter set to the OrtValue instance's internal shape data array.
+   *                        Must not be released as it is owned by the OrtValue instance. This data becomes invalid
+   *                        when the OrtValue is released.
```
edgchen1: is there any other case where it may be invalidated? e.g., if the underlying shape is updated? not sure if it is possible to do this with the current public APIs but if it could happen, it may be worth mentioning.

adrianlizarraga: I looked at the APIs and I don't see a way to update or reallocate the underlying shape array via public APIs.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -7196,6 +7196,29 @@ struct OrtApi {
...
+   * \param[out] shape_data_count Output parameter set to the number of elements in `shape_data`.
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \since Version 1.24.
+   */
+  ORT_API2_STATUS(GetTensorElementTypeAndShapeDataReference, _In_ const OrtValue* value,
+                  _Out_ ONNXTensorElementDataType* elem_type,
+                  _Outptr_ const int64_t** shape_data,
```
fs-eire: need to mention what the `shape_data` will be when the tensor is a scalar (`shape_data_count == 0`)

include/onnxruntime/core/session/onnxruntime_c_api.h
edgchen1: ok, good to know that it is not directly modifiable with the public APIs. I do see an `onnxruntime::Tensor::Reshape()` member function.

https://github.com/microsoft/onnxruntime/blob/30ad3505c6961c2aa8d19aa3e18a8ae7bd3153ff/include/onnxruntime/core/framework/tensor.h#L256-L266

but I'm not sure if this affects `OrtValue`s that users would deal with.

edgchen1: btw this is just something to think about. I'm happy with refining the docs later if we want to.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -7196,6 +7196,31 @@ struct OrtApi {
...
+   *
+   * \note This function is an alternative to ::GetTensorTypeAndShape() that does not allocate a new array for
+   *       the shape data. The OrtValue instance's internal shape data is returned directly.
+   *
+   * \note Returns an error if the underlying OrtValue is not a Tensor.
+   *
+   * \param[in] value The OrtValue instance.
+   * \param[out] elem_type Output parameter set to the tensor element data type.
+   * \param[out] shape_data Output parameter set to the OrtValue instance's internal shape data array.
+   *                        For a scalar, `shape_data` is NULL and `shape_data_count` is 0.
```
edgchen1, line 7209: > `shape_data` is NULL

where is it set to null? is `gsl::span::data()` for an empty span guaranteed to return nullptr?

include/onnxruntime/core/session/onnxruntime_c_api.h
adrianlizarraga: It looks like Reshape is currently only called by optimizers or CPU EP. There isn't an equivalent public API at the moment. But I went ahead and added the possibility of invalidation through shape updates to the docs.

include/onnxruntime/core/session/onnxruntime_c_api.h
adrianlizarraga: This is what it is returning. Should we set it explicitly just in case?

adrianlizarraga: I set shape_data to null explicitly if shape_data_count is 0. Let me know if you prefer anything else

include/onnxruntime/core/session/onnxruntime_c_api.h
adrianlizarraga: Good point. Updated. Also made the API function implementation set `shape_data` to null if `shape_data_count` is 0.

PR #27176, opened 2026-01-27
BaseTester: support plugin EPs with compiled nodes and registered kernels

adrianlizarraga:
### Description
Updates the `BaseTester` class used by the `onnxruntime_provider_test` tool to support plugin EPs that use a kernel registry but compile other nodes. For example, TRT EP only uses registered kernels for Memcpy* nodes, but compiles every other node.

Without this change, plugin EPs that use a mix of compiled nodes and registered kernels cannot be tested with `onnxruntime_provider_test`.


onnxruntime/test/unittest_util/base_tester.cc
```diff
@@ -830,12 +830,12 @@ void BaseTester::ExecuteModelForEps(
 
   ASSERT_TRUE(!execution_providers.empty()) << "Empty execution providers vector.";
   if (try_assign_ep_for_nodes) {
-    auto ep_uses_kernel_registry = [](const IExecutionProvider& ep) {
+    auto ep_only_uses_kernel_registry = [](const IExecutionProvider& ep) {
       const auto& provider_type = ep.Type();
 
-      constexpr std::array kEpsThatDoNotUseKernelRegistry{
+      constexpr std::array kEpsThatDoNotUseKernelRegistryForAllNodes{
```
edgchen1: maybe call them "EPs that compile nodes" instead of "EPs that do not use kernel registry for all nodes"

onnxruntime/test/unittest_util/base_tester.cc
```diff
@@ -830,12 +830,12 @@ void BaseTester::ExecuteModelForEps(
...
           kOpenVINOExecutionProvider,
-          kTensorrtExecutionProvider,
+          kTensorrtExecutionProvider,  // uses kernel registry for Memcpy* nodes, but compiles all others.
```
edgchen1: just to confirm, the plugin EP name will be different, right? that assumption is made here:
https://github.com/microsoft/onnxruntime/blob/f83d4d06e4283d53a10c54ce84da3455cfb4e21d/onnxruntime/test/unittest_util/base_tester.cc#L689-L691

adrianlizarraga: Good question. @chilo-ms do you know if the plugin EP name is different?

chilo-ms: I assume so, but let me confirm with Nvidia for their plugin TRT RTX EP.

Also, do we want to include provider-bridge TRT RTX EP, kNvTensorRTRTXExecutionProvider, in this list as well?

adrianlizarraga: we should only add kNvTensorRTRTXExecutionProvider to this list if its name differs from the plugin EP version, right?

chilo-ms: > we should only add kNvTensorRTRTXExecutionProvider to this list if its name differs from the plugin EP version, right?

Yes, waiting for Nvidia's reply for the name of their plugin TRT RTX EP, will update.

chilo-ms: @ishwar-raut1 as you mentioned offline, the name to register EP DLL is also set for plugin TRT RTX EP name. Could you help confirm that the EP name doesn't conflict with “TensorRTExecutionProvier” and "NvTensorRTRTXExecutionProvier” (will be added) in the `kEpsThatCompileNodes` list?

@keshavv27 @gedoensmax

ishwar-raut1: The name to register the EP DLL is also set for the plugin TRT RTX EP name. if the same is used to register the dll then there could be a conflict.

chilo-ms: To run `onnxruntime_provider_test` with plugin TRT RTX EP, the plugin EP name must not conflict with `NvTensorRTRTXExecutionProvider`; otherwise, the test will fail at this check that Ed mentioned:
https://github.com/microsoft/onnxruntime/blob/f83d4d06e4283d53a10c54ce84da3455cfb4e21d/onnxruntime/test/unittest_util/base_tester.cc#L689-L691

adrianlizarraga: I think we can conditionally add `NvTensorRTRTXExecutionProvider` to `all_provider_types` **only if** `USE_NV` is defined, which occurs only when the provider-bridge NV TRT RTX EP is built with ORT. 

The result would be that this error would only occur if the user tried to use both the provider-bridge and plugin EPs at the same time (with a build that included the provide-bridge EP).

PR #27178, opened 2026-01-28
Update to C++20

edgchen1:
### Description

Update all builds to C++20. Previously, C++20 was only enabled for MacOS builds.

### Motivation and Context

Update to C++20 standard to enable use of new language and standard library features.


.github/workflows/linux_minimal_build.yml
```diff
@@ -530,93 +530,43 @@ jobs:
             --cmake_extra_defines onnxruntime_BUILD_UNIT_TESTS=OFF
 
   # Job 7: Extended minimal build with NNAPI EP for Android(arm64-v8a) and skip tests.
-  # NOTE: Keeping this as direct docker run due to custom volume mounts needed for Android SDK/NDK
```
edgchen1, line 533: there's an error with CMake not finding `clang-scan-deps` from the NDK for the particular configuration of:
- Linux build with Docker container
- C++20
- Ninja generator

it works fine when running outside of Docker. this is what the other Android CI builds do, e.g., here:
https://github.com/microsoft/onnxruntime/blob/a5dc0f95d8598f13d326765976ef39bd8253735d/.github/workflows/android.yml#L194

I just simplified this to make it consistent instead of debugging further for now.

tools/ci_build/github/android/default_qnn_aar_build_settings.json
```diff
@@ -12,6 +12,7 @@
         "--build_java",
         "--build_shared_lib",
         "--use_qnn=static_lib",
+        "--cmake_extra_defines=CMAKE_CXX_SCAN_FOR_MODULES=OFF",
```
edgchen1: also related to the issue in this comment:
https://github.com/microsoft/onnxruntime/pull/27178/changes#r2756173440

this setting has already been added here:
https://github.com/microsoft/onnxruntime/blob/735e69a396afe7e1bdfbc6b38c1d43eeb68b8484/tools/ci_build/github/android/default_full_aar_build_settings.json#L20

cmake/CMakeLists.txt
```diff
@@ -259,6 +245,37 @@ option(onnxruntime_USE_OPENVINO_INTERFACE "Build ONNXRuntime shared lib which is
...
+    #     `std::vector<onnxruntime::llm::cutlass_extensions::CutlassGemmConfig>`:
+    #       /opt/rh/gcc-toolset-14/root/usr/include/c++/14/bits/stl_construct.h:97:14: error: writing 1 byte into a
+    #       region of size 0 [-Werror=stringop-overflow=]
+    #   - Possibly a spurious warning
+    #
+    # When the CUDA EP becomes an independent plugin EP, hopefully we can update all of onnxruntime to C++20.
+    # We can keep building the CUDA plugin EP with C++17 if needed.
+    set(CMAKE_CXX_STANDARD 17)
+  else()
+    set(CMAKE_CXX_STANDARD 20)
```
fs-eire: can we set CMAKE_CXX_STANDARD 20 here, and only set it 17 for CUDA?

```
  target_compile_features(onnxruntime_providers_cuda PRIVATE cxx_std_17)
```

edgchen1: I think that is a riskier change. If we try it, I'd prefer to do so in a separate PR.

Abseil has a warning message that explicitly says that all code using it should use the same C++ standard version.
https://github.com/abseil/abseil-cpp/blob/4a9a571e98f8302dce7aa402fd93790bf5694586/CMake/AbseilDll.cmake#L752

It might work, but we should test things out more thoroughly if we try mixing C++ standard versions across targets.

cmake/CMakeLists.txt
```diff
@@ -259,6 +245,37 @@ option(onnxruntime_USE_OPENVINO_INTERFACE "Build ONNXRuntime shared lib which is
 option(onnxruntime_USE_VITISAI_INTERFACE "Build ONNXRuntime shared lib which is compatible with Vitis-AI EP interface" OFF)
 option(onnxruntime_USE_QNN_INTERFACE "Build ONNXRuntime shared lib which is compatible with QNN EP interface" OFF)
 
+# Set C/C++ standard versions
+if (NOT CMAKE_C_STANDARD)
```
skottmckay: AI says CMAKE_CXX_STANDARD should be set before the `include(...)` lines as the `Check...` includes do test compiles and may use that value. In the suggested setup it had it was even before the `project(...)` statement.

Previously we set CMAKE_CXX_STANDARD after those, so maybe moving it to prior would be an improvement. I'm also wondering if there was there a reason to move this down in the file? I think it's harder to find if we're setting something I would consider important like this after the many many lines of options.

fs-eire: I think it's simply because it depends on `onnxruntime_USE_CUDA` so moved after option sections.

edgchen1: yes, it depends on `onnxruntime_USE_CUDA` now. I moved some code around so that the options and setting of CMAKE_CXX_STANDARD are before most other `include()`s now.

cmake/CMakeLists.txt
edgchen1: will defer this for now, possibly for another PR.

fs-eire (dismissed):
LGTM.

Regarding the comments from Copilot:
- Timestamp: I think it is fine since this is not a part of the public API.
- u8string: not 100% sure about this part, maybe need vitis owner to take a look.

include/onnxruntime/core/common/logging/logging.h
```diff
@@ -56,43 +55,30 @@ struct OrtLogger;  // opaque API type. is always an instance of Logger
...
-#elif TARGET_OS_MAC
-// Xcode added support for C++20's std::chrono::operator<< in SDK version 14.4,
-// but the target macOS version must also be >= 13.3 for it to be used.
-#if (defined(__MAC_OS_X_VERSION_MAX_ALLOWED) && __MAC_OS_X_VERSION_MAX_ALLOWED < 140400) || \
-    (defined(__MAC_OS_X_VERSION_MIN_REQUIRED) && __MAC_OS_X_VERSION_MIN_REQUIRED < 130300)
-#undef ORT_USE_CXX20_STD_CHRONO
-#endif
+  friend std::ostream& operator<<(std::ostream& os, const Timestamp& time_stamp) {
+    return time_stamp.WriteToStream(os);
+  }
```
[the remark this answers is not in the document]
edgchen1: `Timestamp` is an internal type and shouldn't be used by consumers across shared library boundaries.

include/onnxruntime/core/common/logging/logging.h
```diff
@@ -56,43 +55,30 @@ struct OrtLogger;  // opaque API type. is always an instance of Logger
...
-#endif
+// This class wraps `std::chrono::system_clock::time_point` and provides `operator<<`.
+// It is a workaround for the inconsistent availability of `std::chrono::operator<<` for
+// `std::chrono::system_clock::time_point`.
+// When all builds support `std::chrono::operator<<`, we can simplify to this:
+//   `using Timestamp = std::chrono::system_clock::time_point;`
+class Timestamp {
+ public:
+  using TimePoint = std::chrono::system_clock::time_point;
+  Timestamp(const TimePoint& time_point) noexcept : time_point_{time_point} {}
```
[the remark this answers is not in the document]
edgchen1: `Timestamp` is an internal type.

fs-eire (dismissed):

onnxruntime/core/providers/vitisai/imp/global_api.cc
edgchen1, line 1: @wcy123 please take a look at these changes and let me know if there are any issues

include/onnxruntime/core/common/gpu_profiler_common.h
```diff
@@ -379,8 +379,8 @@ class GPUProfilerBase : public EpProfiler {
   void MergeEvents(std::map<uint64_t, Events>& events_to_merge, Events& events) {
     Events merged_events;
 
-    auto event_iter = std::make_move_iterator(events.begin());
-    auto event_end = std::make_move_iterator(events.end());
+    auto event_iter = events.begin();
```
edgchen1, line 382: note: `std::move_iterator::operator->()` is deprecated in C++20

cmake/CMakeLists.txt
```diff
@@ -1476,7 +1472,7 @@ configure_file(onnxruntime_config.h.in ${CMAKE_CURRENT_BINARY_DIR}/onnxruntime_c
 get_property(onnxruntime_GENERATOR_IS_MULTI_CONFIG GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
 
 if (onnxruntime_USE_CUDA)
-  set(CMAKE_CUDA_STANDARD 17)
+  set(CMAKE_CUDA_STANDARD ${CMAKE_CXX_STANDARD})
```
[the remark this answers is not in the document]
edgchen1: minimum CUDA version is now 12

tools/ci_build/github/android/default_qnn_aar_build_settings.json
edgchen1: update: now setting CMAKE_CXX_SCAN_FOR_MODULES in CMakeLists.txt

cmake/CMakeLists.txt
```diff
@@ -34,19 +47,14 @@ include(CheckFunctionExists)
...
-    set(CMAKE_CXX_STANDARD 17)
-  endif()
-endif()
-
 if (MSVC)
   #  Make sure Visual Studio sets __cplusplus macro correctly: https://learn.microsoft.com/en-us/cpp/build/reference/zc-cplusplus
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /Zc:__cplusplus")
 
+  # Enable the standards-conforming preprocessor.
+  add_compile_options("$<$<COMPILE_LANGUAGE:C,CXX>:/Zc:preprocessor>"
```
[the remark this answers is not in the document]
edgchen1: CMake docs say that specifying multiple languages is valid since CMake 3.15:
https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#genex:COMPILE_LANGUAGE:languages

.github/workflows/linux_minimal_build.yml
edgchen1: update: reverted this change. disabling CMAKE_CXX_SCAN_FOR_MODULES should avoid this error.

.github/workflows/linux_minimal_build.yml
```diff
@@ -530,93 +530,43 @@ jobs:
...
+              --build_dir ./build.extended_minimal.nnapi \
               --cmake_generator Ninja \
               --config MinSizeRel \
               --skip_submodule_sync \
               --parallel --use_binskim_compliant_compile_flags \
               --android \
-              --android_sdk_path /android_home \
-              --android_ndk_path /ndk_home \
+              --android_sdk_path "$ANDROID_HOME" \
+              --android_ndk_path "$ANDROID_NDK_HOME" \
```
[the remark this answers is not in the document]
edgchen1: reverted

.github/workflows/linux_minimal_build.yml
```diff
@@ -530,93 +530,43 @@ jobs:
...
               --android_abi=arm64-v8a \
               --android_api=29 \
               --use_nnapi \
               --minimal_build extended \
               --build_shared_lib \
               --disable_ml_ops \
               --disable_exceptions \
-              --skip_tests"
+              --skip_tests
```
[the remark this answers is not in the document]
edgchen1: reverted

cmake/external/cutlass.cmake
```diff
@@ -4,7 +4,7 @@ onnxruntime_fetchcontent_declare(
   URL ${DEP_URL_cutlass}
   URL_HASH SHA1=${DEP_SHA1_cutlass}
```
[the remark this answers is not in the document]
edgchen1: this is a valid point. we have a bunch of other SHA1 hashes in deps.txt though. I think it's beyond the scope of this PR.

onnxruntime/core/providers/cuda/cu_inc/cub.cuh
```diff
@@ -0,0 +1,20 @@
...
+// As a workaround, undefine `__out` before including cub/cub.cuh.
+#if defined(_MSC_VER)
+#pragma push_macro("__out")
+#undef __out
+#endif
+
+#include <cub/cub.cuh>
+
+#if defined(_MSC_VER)
+#pragma pop_macro("__out")
```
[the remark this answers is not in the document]
edgchen1: Windows CUDA build is ok so it seems like we don't need additional warning suppression.

tools/ci_build/github/azure-pipelines/stages/py-win-gpu-stage.yml
```diff
@@ -122,6 +122,7 @@ stages:
             --enable_pybind
             --enable_onnx_tests
             --parallel
+            --nvcc_threads 1
```
edgchen1, line 125: note: to avoid running out of memory during compilation

cmake/CMakeLists.txt
edgchen1: managed to update everything to C++20

cmake/onnxruntime_providers_cuda.cmake
```diff
@@ -149,6 +149,17 @@
     onnxruntime_add_shared_library_module(onnxruntime_providers_cuda ${onnxruntime_providers_cuda_all_srcs})
   endif()
 
+  if (MSVC)
+    # Use /permissive to work around compilation error from CUTLASS header cute/tensor.hpp:
+    #   cutlass-src\include\cute\stride.hpp(299,46): error C3545: 'Ints': parameter pack expects a non-type
+    #     template argument
+    # See https://github.com/NVIDIA/cutlass/issues/3065
```
tianleiwu, line 156: Using `/permissive` may mask conformance issues in ORT's own code compiled in this target. Ideally this should be narrowed to only CUTLASS-consuming TUs if possible. Add a comment here for future awareness.

include/onnxruntime/core/common/logging/logging.h
```diff
@@ -56,43 +55,30 @@ struct OrtLogger;  // opaque API type. is always an instance of Logger
...
```
tianleiwu, line 70: explicit

tianleiwu (dismissed):

include/onnxruntime/core/common/logging/logging.h
edgchen1: the non-explicit ctor was intentional to allow implicit conversions. added some comments.

tianleiwu (approved):

cmake/onnxruntime_providers_cuda.cmake
edgchen1: it would be good to narrow the scope of `/permissive`, but I think that would require tracking which source files use the CUTLASS headers. will just set `/permissive` for the entire target for simplicity, for now.

onnxruntime/core/providers/vitisai/imp/global_api.cc
edgchen1: if there are issues, we can fix them in a later change. resolving this.

kevinmariano684-web (approved):
👍🏻

alg21990w-ai (approved):

PR #27195, opened 2026-01-28
Fix API doc comment for OrtApi::RunOptionsEnableProfiling

adrianlizarraga:
### Description
Fixes C++ documentation generation by replacing `<` and `>` with `[` and `]`. Angle brackets are mistaken as html tags.

Successful run: https://github.com/microsoft/onnxruntime/actions/runs/21456738258

### Motivation and Context
Allow C++ document generation to succeed.

### DIFF
--- a/include/onnxruntime/core/session/onnxruntime_c_api.h
+++ b/include/onnxruntime/core/session/onnxruntime_c_api.h
@@ -7200,7 +7200,7 @@ struct OrtApi {
    *
    * \param[in] options
    * \param[in] profile_file_prefix The prefix for the profile file. The actual filename will be:
-   *                                <profile_file_prefix>_<timestamp>.json
+   *                                [profile_file_prefix]_[timestamp].json
    *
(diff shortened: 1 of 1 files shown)


include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -7200,7 +7200,7 @@ struct OrtApi {
    *
    * \param[in] options
    * \param[in] profile_file_prefix The prefix for the profile file. The actual filename will be:
-   *                                <profile_file_prefix>_<timestamp>.json
+   *                                [profile_file_prefix]_[timestamp].json
```
adrianlizarraga, line 7203: Note: could also escape with `\\<` and `\\>`, but I think using `[` and `]` is easier to read here.

yuslepukhin (approved):
:shipit:

xiaofeihan1 (approved):

PR #27197, opened 2026-01-28
Session config to create weightless EPContext models during compilation

adrianlizarraga:
### Description
Adds a session configuration entry named `""ep.enable_weightless_ep_context_nodes"` that an application enables to explicitly request that EPs create compiled models (i.e., EPContext models) with EPContext nodes that do not store model weights internally. Instead, the weights should be provided by ONNX Runtime as explicit inputs to the EPContext nodes.

### Motivation and Context
If the weights are originally stored in an external file, this allows multiple models to share the same external weights file.


javier-intel:
Looks good to me

javier-intel:
Hi @adrianlizarraga, is this still planned for the next ORT release?

PR #27211, opened 2026-01-30
[WebGPU EP] Remove unused `has_bias` in `MatMul::ComputeInternal`

Jiawei-Shao:
### Description
This patch removes all the logic that handles `bias` in `MatMul::ComputeInternal()`.

### Motivation and Context
As `MatMul` operator doesn't support `bias` as its input, we don't need to handle `bias` in `MatMul::ComputeInternal()`.

### DIFF
--- a/onnxruntime/core/providers/webgpu/math/matmul.cc
+++ b/onnxruntime/core/providers/webgpu/math/matmul.cc
@@ -115,7 +115,6 @@ Status MatMul::ComputeInternal(ComputeContext& context) const {
     // If the output tensor is empty, we can return early.
     return Status::OK();
   }
-  bool has_bias = context.InputCount() > 2;
 
   if (helper.N() < 8 && helper.K() < 8) {  // call MatMulNaiveProgram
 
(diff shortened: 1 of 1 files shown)


onnxruntime/core/providers/webgpu/math/matmul.cc
qjia7, line 156: I believe it should be `return ComputeMatMul(&context, Activation(), inputs, output_tensor, true);` instead of `return ComputeMatMul(&context, Activation(), inputs, output_tensor, false);`. The previous one will prevent you go to the splitK path.

Please consider directly return ComputeMatMul(&context, Activation(), inputs, output_tensor, true) in line 118 and move all necessary optimizations in ComputeMatMul to reduce duplicated code.

daijh:
Is this bias reserved for fusing `MatMul` + `Add` sequence.

fs-eire:
> Is this bias reserved for fusing `MatMul` + `Add` sequence.

I think it is because when porting from TypeScript, matmal is called by conv implementation so that bias is needed. It finally turns out to be `ComputeMatMul()` is the function that called by Conv and there is no need to support bias for `MatMul::ComputeInternal` now.

I don't see any existing fuse for MatMul uses bias. MatMul and GEMM are quite similar, and whether to support optional `bias` input is one of the differences.

Jiawei-Shao:
I will close this PR as we have already been working on https://github.com/microsoft/onnxruntime/pull/27674.

PR #27221, opened 2026-02-01
Fix WebGPU MoE swiglu_limit (default to infinity)

### LINKED ISSUE #27220: [WebGPU] com.microsoft.QMoE produces invalid results for certain attribute combinations
### Describe the issue

The com.microsoft.QMoE operation fails with large correctness issue for the following attributes:

Image

cc @guschmue 

### To reproduce

```py
import onnxruntime as ort
from huggingface_hub import hf_hub_download

path = hf_hub_download(repo_id="Xenova/onnxruntime-webgpu-test-data", filename="fail_673_standalone.onnx", repo_type="dataset")
cpu_session = ort.InferenceSession(path, providers=['CPUExecutionProvider'])
cpu_result = cpu_session.run(None, {})

webgpu_session = ort.InferenceSession(path, providers=['WebGpuExecutionProvider'])
webgpu_result = webgpu_session.run(None, {})

diff = cpu_result[0] - webgpu_result[0]
print('Diff stats: ', f'min={diff.min()}, max={diff.max()}, mean={diff.mean()}, std={diff.std()}')
```

produces this error:
```
Diff stats:  min=-7.402378082275391, max=8.158369064331055, mean=-0.0036832974292337894, std=0.3670535087585449
```

### Urgency

medium-high -- blocks certain models which use this operation.

### ONNX Runtime Installation

Built from Source

main

### Execution Provider

'webgpu' (WebGPU)

xenova:
### Description

According to https://github.com/microsoft/onnxruntime/blob/main/docs/ContribOperators.md#commicrosoftmoe, 

> swiglu_limit : float
The limit used to clamp in SwiGLU. No clamp when limit is not provided.

However, currently, the default is set to 0, meaning we clamp to 0 if no limit is provided.

### Motivation and Context

Fixes #27220. See there for bug description and reproduction.

Hoping to get this in before 1.24.0 releases. cc @guschmue

### DIFF
--- a/onnxruntime/contrib_ops/webgpu/moe/moe.h
+++ b/onnxruntime/contrib_ops/webgpu/moe/moe.h
@@ -3,6 +3,8 @@
 
 #pragma once
 
+#include <limits>
+
 #include "core/providers/webgpu/program.h"
 #include "core/providers/webgpu/webgpu_kernel.h"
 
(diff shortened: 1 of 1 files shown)


guschmue:
nice catch!

ekzhang:
Yes this is really nice haha :D

PR #27222, opened 2026-02-02
Only reduce the range when in force symmetric case

yiliangbetter:
### Description
- Align C++ quantization range logic with Python implementation in `quant_utils.py`
- Only reduce quantization range for `int8_t` when `force_symmetric=true` is specified
- For non-symmetric quantization, use full range `[-128, 127]` for `int8_t` to improve precision
- For symmetric quantization, use reduced range `[-127, 127]` for `int8_t` to ensure zero point is exactly 0
- Update `QuantizeLinear` function to conditionally apply range reduction based on `force_symmetric` parameter
- Add comprehensive unit tests including:
  - Test for non-symmetric quantization with full range
  - Test for symmetric quantization with reduced range
  - Test verifying zero point is always 0 for symmetric `int8_t` quantization across various input distributions

### Motivation and Context
- Previously, the C++ implementation always used reduced range `[-127, 127]` for `int8_t` quantization, regardless of whether symmetric quantization was requested
- This unnecessarily reduced precision for non-symmetric quantization cases, where the full range `[-128, 127]` should be used
- The Python implementation in `quant_utils.py` correctly uses:
  - `ONNX_INT_TYPE_RANGE`: Full range `[-128, 127]` for non-symmetric `int8_t`
  - `ONNX_INT_TYPE_SYMMETRIC_RANGE`: Reduced range `[-127, 127]` for symmetric `int8_t`
- This change aligns the C++ implementation with the Python reference implementation, ensuring consistency across the codebase
- The improved precision for non-symmetric quantization can lead to better model accuracy in certain scenarios
- The symmetric quantization path remains unchanged, still guaranteeing zero point = 0 as required

### DIFF
--- a/onnxruntime/core/quantization/quantization.h
+++ b/onnxruntime/core/quantization/quantization.h
@@ -94,16 +94,20 @@ Params<T> QuantizeLinear(const float* data,
 
   T T_min = std::numeric_limits<T>::min();
   T T_max = std::numeric_limits<T>::max();
-  // NOTE: ORT currently clamps signed quantization values to -127,127 instead
-  //       of -128/127. This is done to ensure that forced symmetric
-  //       quantization results in a zero point of exactly 0 for signed 8 bit
-  //       ints.
-  // TODO(kreeger): Consider adjusting this clamping to enable more precision
-  //                for signed 8 bit ints.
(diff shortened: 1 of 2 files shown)


yiliangbetter:

PR #27223, opened 2026-02-02
[webgpu] Optimize string stream used in WebGPU EP

fs-eire:
### Description

Optimize the string stream used in WebGPU EP.

### Motivation and Context

The current implementation uses a `absl::OStringStream`, which is faster than `std::ostringstream`. However, it is still slow in the usage of generating the program cache key.

From the profiling data, `CalculateProgramCacheKey()` is extremely time consuming. It can consume up to 1/3 of all CPU time inside `WebGpuContext::Run()`:

image

The basic analyze shows that most time spent in the `std::basic_ostream operator <<()` implementation, and this is way slower than expected.

To optimize, this PR uses a simplified implementation `FastOStringStream`, which does not inherit from `std::basic_ostream`. Instead, the class implementation only includes necessary overrides for the minimum requirements of generating cache key and shader code, to reduce the unnecessary overhead as much as possible.

image

As a result, the CPU sampling of `CalculateProgramCacheKey()` in the same test dropped from 2555 to 176. Generation TPS of E2E model benchmark on Qwen3-0.6B increased from ~90 to ~130 on Windows11/13900k/RTX4070.

### DIFF
--- a/include/onnxruntime/core/framework/allocator.h
+++ b/include/onnxruntime/core/framework/allocator.h
@@ -85,8 +85,8 @@ constexpr const char* OpenVINO_GPU = "OpenVINO_GPU";
 constexpr const char* OpenVINO_RT = "OpenVINO_RT";
 constexpr const char* OpenVINO_RT_NPU = "OpenVINO_RT_NPU";
 constexpr const char* QNN_HTP_SHARED = "QnnHtpShared";
-constexpr const char* WEBGPU_BUFFER = "WebGPU_Buffer";
-constexpr const char* WEBNN_TENSOR = "WebNN_Tensor";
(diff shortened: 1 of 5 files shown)


guschmue:
I think I see some issue when using in my local repo. Debugging ...

guschmue:
python binding looks ok.
But genai main has issues with ort main + this PR: 
RuntimeError: Specified device is not supported. Try CreateMemoryInfo_V2

guschmue:
void - all good.

xenova:
> Generation TPS of E2E model benchmark on Qwen3-0.6B increased from ~90 to ~130 on Windows11/13900k/RTX4070.

Oh wow, that's a huge difference. Great work @fs-eire!

Issue #18654, opened 2023-12-01
einsum can't handle equation 'ijk->ki' using onnxruntime


xiangzho72:
### Describe the issue

I have simple einsum equation 'ijk->kj' with input shape [2,3,4] and output shape as [4,2].  However, onnx runtime can't get it computed with error ('onnxruntime::common::Status onnxruntime::EinsumOp::DeviceHelpers::CpuDeviceHelpers::DataCopy(const onnxruntime::Tensor&, onnxruntime::Tensor&, void*) output.SizeInBytes() == input.SizeInBytes() was false. Einsum op: The candidate output does not match the actual output's shape). 

Looking through code,  I would think: 
EinsumTypedComputeProcessor<T>::FinalizeOutput(const Tensor& candidate_output,
                                                    const gsl::span<const int64_t>& ordered_subscript_indices_in_candidate) 
would implicitly require candidate_output has the same rank as ordered_subscript_indices_in_candidate. But in this case, rank of candidate_output would still be 3 with shape {2,1,4} while ordered_subscript_indices_in_candidate would be 2 with shape{0, 2} , this will generate candidate_output_shape_without_reduced_dims as [2,1], which is incorrect and should be [2,4].  

### To reproduce

(py11) sjt@sjt:~/piano1$ cat test.py
from onnx import helper,checker
from onnx import TensorProto
import numpy as np
import onnxruntime as rt
import onnx

def generate_onnx(node_name, op_set):
    Eqn = "ijk->ki"
    test_node = helper.make_node(
        node_name,
        inputs=["X"],
        outputs=["Y"],
        equation = Eqn
    )

    inputs = [
        helper.make_tensor_value_info("X", TensorProto.DOUBLE, [2, 3, 4]),
    ]

    outputs = [helper.make_tensor_value_info("Y", TensorProto.DOUBLE, [4, 2])]

    graph = helper.make_graph(
        [test_node],
        "test",
        inputs,
        outputs,
    )

    model = helper.make_model(graph)
    model.opset_import[0].version = op_set
    checker.check_model(model)
    onnx.save(model, node_name + ".onnx")
    print(f"model is generated: {node_name}.onnx")

generate_onnx('Einsum', op_set=12)

input = np.random.randn(2,3,4)
sess = rt.InferenceSession("./Einsum.onnx")
output = sess.run(None, {'X': input})

(py11) sjt@sjt:~/piano1$ python test.py
model is generated: Einsum.onnx
[...]

PR #27225, opened 2026-02-03
[Kernel] Fix bug in Einsum implementation when a lone operand had a reduction operation

hariharans29:
### Description
When the lone operand had a reduction operation in Einsum, there was a mis-management in the shape which led to a mis-match in terms of the subscript indices in the output as deciphered from the Einsum equation and the rank of the candidate output being produced after the reduction operation which caused issues while finalizing the output of the Einsum op.

This change fixes the bug and adds a test

### Motivation and Context
Resolve https://github.com/microsoft/onnxruntime/issues/18654

### DIFF
--- a/onnxruntime/core/providers/cpu/math/einsum_utils/einsum_typed_compute_processor.cc
+++ b/onnxruntime/core/providers/cpu/math/einsum_utils/einsum_typed_compute_processor.cc
@@ -367,7 +367,7 @@ Status EinsumTypedComputeProcessor<T>::Run() {
   std::unique_ptr<const Tensor> result;
 
   {
-    TensorShapeVector reduced_dims;                // All dims of the input that are reduced using the `ReduceSum` op
+    TensorShapeVector reduced_dims;              // All dims of the input that are reduced using the `ReduceSum` op
     reduced_dims.reserve(num_subscript_labels);  // num_subscript_labels is the upper bound. No harm in over-reserving
 
     TensorShapeVector all_dims;              // All dimension indices from 0 to num_subscript_labels - 1


onnxruntime/core/providers/cpu/math/einsum_utils/einsum_typed_compute_processor.cc
```diff
@@ -361,17 +367,19 @@ Status EinsumTypedComputeProcessor<T>::Run() {
...
   {
-    TensorShapeVector reduced_dims;
-    TensorShapeVector preserved_dims;                                           // dims which were not reduced
-    reduced_dims.reserve(onnxruntime::narrow<size_t>(num_subscript_labels));    // num_subscript_labels is the upper bound. No harm in over-reserving.
-    preserved_dims.reserve(onnxruntime::narrow<size_t>(num_subscript_labels));  // num_subscript_labels is the upper bound. No harm in over-reserving.
+    TensorShapeVector reduced_dims;                                           // All dims of the input that are reduced using the `ReduceSum` op
+    reduced_dims.reserve(onnxruntime::narrow<size_t>(num_subscript_labels));  // num_subscript_labels is the upper bound. No harm in over-reserving
+
+    TensorShapeVector all_dims;                                           // All dimension indices from 0 to num_subscript_labels - 1
+    all_dims.reserve(onnxruntime::narrow<size_t>(num_subscript_labels));  // num_subscript_labels is the number of elements
```
edgchen1: nit: could we make `num_subscript_labels` a `size_t` to reduce the number of calls to `narrow`?

PR #27235, opened 2026-02-04
Fix regression in agents for Windows ARM64

eserscor:
### Description
A recent push to the agent infrastructure broke python on Windows ARM64. This patch will unblock CICD until an infrastructure patch can get rolled out.

### DIFF
--- a/tools/ci_build/github/azure-pipelines/templates/py-win-arm64-qnn.yml
+++ b/tools/ci_build/github/azure-pipelines/templates/py-win-arm64-qnn.yml
@@ -33,7 +33,8 @@ jobs:
     os: windows
     hostArchitecture: Arm64
     demands:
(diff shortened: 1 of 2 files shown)


tools/ci_build/github/azure-pipelines/templates/py-win-arm64-qnn.yml
```diff
@@ -32,7 +32,8 @@ jobs:
     name: ${{ parameters.MACHINE_POOL }}
     os: windows
     hostArchitecture: Arm64
-  templateContext:
+    demands:
+      - Agent.Version -equals 4.264.2  templateContext:
```
edgchen1: "templateContext:" looks out of place here

eserscor: Wups! I also forgot a file!

PR #27239, opened 2026-02-04
Fix LSTM clipping before h activation (Issue #27224)

### LINKED ISSUE #27224: No clipping before activation in LSTM
Here is the line of code, where activation function h is applied:

https://github.com/microsoft/onnxruntime/blob/a5dc0f95d8598f13d326765976ef39bd8253735d/onnxruntime/core/providers/cpu/rnn/uni_directional_lstm.cc#L581

ONNX documentation for LSTM says that input to every activation function must be clipped if clip parameter is specified:

https://onnx.ai/onnx/operators/onnx__LSTM.html#attributes

chaya2350:
This PR fixes Issue #27224 where LSTM cell state (Ct) was not being clipped before applying the h() activation function, which violates the ONNX specification.

## Changes
1. Modified the LSTM forward pass to clip Ct before applying the h() activation function
2. Fixed SEGFAULT caused by using clip_with_bias_ptr_ with nullptr bias by switching to clip_ignore_bias for Ct clipping

## Testing
- All 10 test suites pass (100%)
- Specifically verified DynamicQuantLSTMTest.SmallSize which was failing with SEGFAULT before the fix

## Related Issue
Fixes #27224

### DIFF
--- a/onnxruntime/core/providers/cpu/rnn/uni_directional_lstm.cc
+++ b/onnxruntime/core/providers/cpu/rnn/uni_directional_lstm.cc
@@ -583,7 +583,8 @@ void UniDirectionalLstm<T>::GateComputations(
     for (int i = 0; i < hidden_size_; ++i) {
       pC_prev_clipped[i] = pC_cur[i];
     }
-    clip_with_bias_ptr_(clip_, nullptr, pC_prev_clipped, hidden_size_);
+    // Ct clipping should not add bias; always use clip_ignore_bias.
+    deepcpu::clip_ignore_bias(clip_, nullptr, pC_prev_clipped, hidden_size_);
 
     activation_h_.func(pC_prev_clipped, pC_prev_clipped, po, pH, hidden_size_, activation_h_.alpha, activation_h_.beta);
 


chaya2350:
@bac-cab Could you please review this PR? It fixes the issue you reported. Thanks! 🙂

chaya2350:
@fs-eire can you trigger the CI?
Thanks!

hariharans29:
Thanks for this contribution. Can you please consider adding a test for this ?

bac-cab:
@chaya2350 I have a test, could you please give me the permission to commit it?

Also a general question, could this be a flaw in ONNX specification, not in the code? Now that I think about it, the elements of Ct never get too large (in fact they are bounded by seq_len for default activation functions), so clipping them is not so important. The [attn_lstm](https://github.com/microsoft/onnxruntime/blob/b9035574f48cd4f892edbf22894664592a0d586c/onnxruntime/contrib_ops/cpu/attnlstm/uni_dir_attn_lstm.cc#L463) and [python](https://github.com/microsoft/onnxruntime/blob/5e66b6138676f633189402c6287fc8a42df263e7/onnxruntime/test/providers/cpu/rnn/LSTM.py#L225) implementations don't clip it either. So maybe the documentation just fails to mention it? @hariharans29

Issue #26992, opened 2026-01-13
[Performance] ORT_ENABLE_ALL graph optimization causes significant slowdowns on Intel CPUs (vs AMD)


marchello2000:
### Describe the issue

We are running inference with ONNX at reasonably large scale on CPU. We have a small model ~1M parameters that we converted from Torch to ONNX. 

The following table shows run speed of the inference (the model processes audio so we use x Real Time as an indicator of speed)

| CPU | Optimization Level | x Real Time Speed |
|-----|--------------------|-------------------|
|AMD  | ORT_ENABLE_ALL     | 200x              |
|AMD  | ORT_ENABLE_EXTENDED| 177x              |
|Intel| ORT_ENABLE_ALL     | 7x                |
|Intel| ORT_ENABLE_EXTENDED| 50x               |

Intel CPU = AWS m7i class 8 core (ex Intel(R) Xeon(R) Platinum, but also observed behavior on other Intel CPUs, such as Core Ultra 7)
AMD CPU   = AWS m7a class 8 core (ex AMD EPYC 9R14)

We tried playing with other session parameters to no avail.

Also tried reverting to older versions of the ONNX Runtime (down to 1.19 with same results)

Any suggestions how we can get Intel CPUs to perform similar to AMD?

### To reproduce

Run inference on Intel vs AMD CPU

### Platform

Linux

### OS Version

Jammy

### ONNX Runtime Installation

Released Package

### ONNX Runtime Version or Commit ID

1.23.2

### ONNX Runtime API

Python

### Architecture

X64

### Execution Provider

Default CPU

### Model File

Model/code is not quite open source, but can be provided if needed

### Is this a quantized model?

No

hariharans29:
Do you know which op(s) the latency delta originates from ?

What happens if you disable all optimizations ? Does the perf look comparabale between the Intel and AMD ? 
AFAIK - we do not have separate codepaths for Intel and AMD - my guess is it may be down to vectorization instruction usage quirk perhaps.

A minimal repro to demonstrate the issue would help

marchello2000:
@hariharans29 Thank you for your response!
Below are some numbers to answer your question about disabling optimizations, but short of it is that optimization levels (`ORT_DISABLE_ALL`, `ORT_ENABLE_BASIC` and `ORT_ENABLE_EXTENDED`) don't seem to make much difference. But `ORT_ENABLE_ALL` does make a difference for AMD it's faster, but Intel much slower

[...]

Issue #23587, opened 2025-02-05
[Performance] Distinct 2x inference time difference between AMD and Intel processors


dsakharuk:
### Describe the issue

I am using ONNX to convert a sequential PyTorch model into the ONNX format to be able to run with c++ scripts. 
Inferences are all done on a CPU using a single core. On an Intel(R) Xeon(R) Gold 6154 CPU each inference takes approximately 20 ms while on an AMD EPYC 7713 64-Core Processor each inference takes approximately 12 ms. 
I tried using the built in ONNX optimizer on the model and setting the compiler flag to use AVX512 instructions for the Intel chip and there have not been any improvements. 

I'm wondering if anyone has encountered similar performance discrepancies and has been able to improve Intel cpu benchmarks. 

### To reproduce

Using Clang compiler

The inference is run using: 
`session_->Run`

The model uses float32 data.
Inferences are conducted hundreds of times in a row.

The following session options are set:
`session_options_.SetIntraOpNumThreads(1); session_options_.DisableCpuMemArena(); 
session_options_.SetGraphOptimizationLevel(GraphOptimizationLevel::ORT_ENABLE_ALL);
session_options_.AddConfigEntry("session.set_denormal_as_zero", "1");`

The memory info is set as:
`Ort::MemoryInfo memory_info = Ort::MemoryInfo::CreateCpu(OrtAllocatorType::OrtArenaAllocator, OrtMemType::OrtMemTypeDefault);`

### Platform

Linux

### OS Version

Centos 7

### ONNX Runtime Installation

Built from Source

### ONNX Runtime Version or Commit ID

1.21.0

### ONNX Runtime API

C++

### Architecture

X86

### Execution Provider

Default CPU

### Is this a quantized model?

No

yuslepukhin:
What is the outcome with denormal = 0?

snnn:
AMD EPYC 7713 doesn't support AVX512.  Intel(R) Xeon(R) Gold 6154 CPU does. So the later one should be faster?

dsakharuk:
> AMD EPYC 7713 doesn't support AVX512. Intel(R) Xeon(R) Gold 6154 CPU does. So the later one should be faster?

I agree which is what puzzles me. When running the build on the Intel CPU I made sure the build info at the beginning had the avx flag set to true:  `use_avx512=True`. So I am assuming the cpu used those instructions although I don't know if there is a way to verify that.

dsakharuk:
[...]

PR #27248, opened 2026-02-05
[Core] Expose a session option to disable NCHWc layout optimizations

hariharans29:
### Description

Sometimes, a model could have an "outlier" convolution operation with large kernel sizes (eg) 11x11 and a good amout of filters (eg) 64 filters. Since, NCHWc layout transformations are turned on by default, data layout for the `Conv` operation is changed to NCHWc (if other `Conv` node parameters fit the bill). The NCHWc `Conv` implementations in MLAS use a "direct" convolution implementation and large kernel sizes + a good amount of filters make the entire operation a heavily  memory badwidth bound operation. With this background, it makes NCHWc data layouts for such models not a good choice for certain platforms whereas on some other platforms (possibly platforms with superior memory bandwidth characteristics and/or other memory characteristics like cache sizes), it still works well. The work-around today is for users to drop down to a lower graph optimization level that does not include the NCHWc transformer on platforms where perf is poor for such models. This seems like a bad precedence to set because it would mean users will miss out on other L3 and L4 optimizers. 

**Short term:**
1) Introduce a session option for users to disable NCHWc layout transformation
2) Log a warning prompting users to consider disabling NCHWc if while ORT is performing the NCHWc layout transofmation, we encounter a `Conv` node that is an outlier in terms of the kernel size

Ultimately, the short term fix gives users the data to help them pick the right data layout for their target model on their target hardware to run it most optimally with ORT.

**Longer term:**
Add complementary convolution implementations in the NCHWc convolution suite in MLAS (like Im2Col + SGemm) to go with the direct convolution implementations and add heuristics to pick an implementation

**More longer term:**
Enable online benchmarking infrastructure to help pick the best algo (direct vs Im2Col + SGemm) for a given data layout and Conv parameters (filter sizes, dilations, strides, etc.) based on the data from warm-up runs and use that for algo selection for future runs

### Motivation and Context
Takeaway from https://github.com/microsoft/onnxruntime/issues/26992 and possibly https://github.com/microsoft/onnxruntime/issues/23587 (although it needs a repro model to be sure)

### DIFF
--- a/onnxruntime/test/optimizer/nchwc_optimizer_test.cc
+++ b/onnxruntime/test/optimizer/nchwc_optimizer_test.cc
@@ -255,8 +255,8 @@ TEST(NchwcOptimizerTests, DisableNchwcLayoutTransformationSessionOption) {
   SessionOptions session_options;
   session_options.graph_optimization_level = TransformerLevel::Level3;
   session_options.session_logid = "NchwcOptimizerDisableTests";
-  //ASSERT_STATUS_OK(session_options.config_options.AddConfigEntry(
-  //    kOrtSessionOptionsDisableNchwcLayoutTransformation, "1"));
+  ASSERT_STATUS_OK(session_options.config_options.AddConfigEntry(
+      kOrtSessionOptionsDisableNchwcLayoutTransformation, "1"));
 
   InferenceSessionWrapper session{session_options, GetEnvironment()};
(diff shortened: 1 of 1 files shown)


hariharans29:
Contents of this PR merged into https://github.com/microsoft/onnxruntime/pull/27368

PR #27249, opened 2026-02-05
Fix WebGPU Conv auto_pad=SAME_UPPER padding calculation

### LINKED ISSUE #26734: [Bug] [WebGPU] [WebGL] Conv+auto_pad produces significantly wrong values
### Describe the issue

As the title says, when Conv node is used on WebGPU with strides>1 and auto_pad e.g. "SAME_UPPER", it produces values that are very much off the expected ones. I'm guessing padding gone wrong?

### To reproduce

Here's the absolutely minimal onnxscript-based repro.

[repro.py](https://github.com/user-attachments/files/23945731/repro.py)
[repro.html](https://github.com/user-attachments/files/23945733/repro.html)
[repro.js](https://github.com/user-attachments/files/23945732/repro.js)

To run,  just use repro.py to generate two .onnx models and start a static server with repro.html and repro.js.

You'll see a table like this (I've included WebGL just for completeness, I'm aware it's deprecated):

  | wasm | webgl | webgpu | webnn (cpu) | webnn (gpu)
-- | -- | -- | -- | -- | --
Stride 1 | (reference)<br>Range: [8.925, 51.957] | MAD: 0.000<br>Range: [8.925, 51.957] | MAD: 0.000<br>Range: [8.925, 51.957] | MAD: 0.000<br>Range: [8.925, 51.957] | MAD: 0.024<br>Range: [8.919, 51.927]
Stride 2 | (reference)<br>Range: [9.601, 54.654] | MAD: 1.658<br>Range: [10.279, 53.070] | MAD: 1.658<br>Range: [10.279, 53.070] | MAD: 0.000<br>Range: [9.601, 54.654] | MAD: 0.024<br>Range: [9.595, 54.622]

plus the dumps of raw data in the console.

P.S. This one took a long time to narrow down for silly reasons, especially since the original symptom was just the images processed by a unet model looking slightly but visibly wrong.

Initially I suspected the Resize nodes, which proved a waste of time.

In the end I ended up writing throwaway scripts to bisect the ONNX graph and dump all intermediate outputs. This helped immensely and allowed to narrow down on the specific node where values diverge between backends very quickly, not counting the time of writing those scripts.

It would be great if ONNXRuntime had similar built-in bisect & comparison tools, like Git or C-Reduce, as it would allow to save days of debugging and help report detailed issues a lot quicker.

### Urgency

_No response_

### Platform

Web Browser

### OS Version

Chromium 142.0.7444.220

### ONNX Runtime Installation

Released Package

1.23.2; tried the latest @dev available on npm too

### ONNX Runtime API

JavaScript

### Architecture

Other / Unknown

### Execution Provider

Other / Unknown

_No response_

ingyukoh:
The WebGPU Conv and ConvTranspose operators were producing incorrect results when using auto_pad=SAME_UPPER with strides > 1.

Root cause: The head padding values were being unnecessarily recalculated after InferPadsAndOutputShape() had already computed the correct values. The recalculation formula could produce incorrect results.

Fix: Simply use pads[0] and pads[1] directly, which already contain the correct head padding values computed upstream. This matches the behavior of the TypeScript implementation.

Fixes #26734

### DIFF
--- a/onnxruntime/test/providers/cpu/nn/conv_op_test.cc
+++ b/onnxruntime/test/providers/cpu/nn/conv_op_test.cc
@@ -502,6 +502,49 @@ TEST(ConvTest, Conv2D_AutoPad1) {
   TestConvOp(attrs, {X, W}, {X_shape, W_shape}, expected_vals, Y_shape, true);
 }
 
+// Regression test for issue #26734: SAME_UPPER with stride > 1
+// Tests asymmetric padding calculation that was incorrect in WebGPU EP
+TEST(ConvTest, Conv2D_AutoPad_SAME_UPPER_Stride2) {
+  ConvOpAndTestAttributes attrs = {
+      "SAME_UPPER",           // auto_pad
+      vector<int64_t>{1, 1},  // dilations
+      1,                      // group
+      vector<int64_t>{3, 3},  // kernel_shape
+      {},                     // pads
+      vector<int64_t>{2, 2},  // strides > 1 triggers asymmetric padding
+      {}                      // excluded EPs
+  };
+
+  // 1x1x4x4 input
+  vector<float> X = {1.0f, 2.0f, 3.0f, 4.0f,
+                     5.0f, 6.0f, 7.0f, 8.0f,
+                     9.0f, 10.0f, 11.0f, 12.0f,
+                     13.0f, 14.0f, 15.0f, 16.0f};
+  vector<int64_t> X_shape = {1, 1, 4, 4};
+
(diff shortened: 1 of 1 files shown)


fs-eire:
The change itself looks good to me. Could you please add a test case in onnxruntime\test\providers\cpu\nn\conv_op_test.cc ?

ingyukoh:
@fs-eire Added regression test `Conv2D_AutoPad_SAME_UPPER_Stride2` as requested.

ingyukoh:
Thanks for the review and guidance on adding the test!

PR #27250, opened 2026-02-05
[Shape Inference] Fix GQA shape inference for present outputs

Honry:
### Description
When using pre-allocated KV cache with `freeDimensionOverrides`, the shape inference for `present_key` and `present_value` outputs failed silently. This caused downstream graph operations to receive tensors with unknown dynamic shapes, leading to unexpected fallback in execution providers like WebNN. (WebNN currently doesn't support dynamic shape)

### Motivation and Context
**Root cause**:
In `BaseGroupQueryAttentionTypeAndShapeInference()`, the shape inference logic for `use_max_past_present_buffer == -1` only propagated shapes when BOTH conditions were met:
  1. `total_sequence_length_value` was a concrete value (> 0)
  2. `past_dims[2]` had a concrete dimension value

When either condition failed (e.g., using `freeDimensionOverrides` which results in dynamic `past_sequence_length`), present output shapes were left uninitialized.

Additionally, when `past_key/past_value` is not provided (prefill/first-token mode), no shape inference was performed for present outputs at all.

**Fix**:
1. For `use_max_past_present_buffer == -1`:
   - Always construct and propagate `present_shape`
   - Compute `present_sequence_length = max(past_sequence_length, total_sequence_length)` when both values are concrete
   - Fall back to copying `past_key`'s sequence dimension when exact value cannot be computed

2. Add new else-if branch to handle prefill mode (no past_key/past_value input):
   - Infer `head_size` from query shape and `num_heads/kv_num_heads` attrs
   - Handle both separate Q/K/V and packed QKV input formats
   - Construct present shape from query dims, `kv_num_heads`, and `total_sequence_length` or `kv_sequence_length`

### DIFF
--- a/onnxruntime/core/graph/contrib_ops/bert_defs.cc
+++ b/onnxruntime/core/graph/contrib_ops/bert_defs.cc
@@ -322,23 +322,24 @@ void BaseGroupQueryAttentionTypeAndShapeInference(ONNX_NAMESPACE::InferenceConte
           updateOutputShape(ctx, 2, present_shape);
         }
       } else if (use_max_past_present_buffer == -1) {
+        // shape of present key/value is (batch_size, kv_num_heads, present_sequence_length, head_size)
+        ONNX_NAMESPACE::TensorShapeProto present_shape;
+        *present_shape.add_dim() = past_dims[0];  // batch_size
+        *present_shape.add_dim() = past_dims[1];  // kv_num_heads
         if (total_sequence_length_value > 0 && past_dims[2].has_dim_value()) {
           // present_sequence_length = max(past_sequence_length, total_sequence_length)
           const int64_t present_sequence_length = total_sequence_length_value > past_dims[2].dim_value()
                                                       ? total_sequence_length_value
                                                       : past_dims[2].dim_value();
-
-          ONNX_NAMESPACE::TensorShapeProto present_shape;
-          for (auto& dim : past_dims) {
-            *present_shape.add_dim() = dim;
-          }
-
-          // shape of present key/value is (batch_size, kv_num_heads, present_sequence_length, head_size)
-          present_shape.mutable_dim(2)->set_dim_value(present_sequence_length);
-
-          updateOutputShape(ctx, 1, present_shape);
-          updateOutputShape(ctx, 2, present_shape);
+          present_shape.add_dim()->set_dim_value(present_sequence_length);
+        } else {
+          // Cannot compute exact present_sequence_length, copy from past_key (may be dynamic)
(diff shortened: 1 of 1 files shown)


Honry:
@guschmue, @tianleiwu, PTAL, thanks!

onnxruntime/core/graph/contrib_ops/bert_defs.cc
```diff
@@ -370,6 +371,52 @@ void BaseGroupQueryAttentionTypeAndShapeInference(ONNX_NAMESPACE::InferenceConte
...
+        }
+
+        if (head_size > 0) {
+          // Determine present_sequence_length from total_sequence_length or kv_sequence_length
+          int64_t present_sequence_length = 0;
+          if (total_sequence_length_value > 0) {
+            present_sequence_length = total_sequence_length_value;
+          } else if (kv_sequence_length > 0) {
+            present_sequence_length = kv_sequence_length;
+          }
```
tianleiwu, line 402: The spec says "When past_key uses same tensor as present_key, it is of length max_sequence_length... otherwise of length past_sequence_length +kv_sequence_length.". That means the shape depends on whether past and present buffer sharing is used. If so, it is max_seq_length, otherwise total_seq_length.

Usually, cuda/cpu EP will allocate a buffer with max_seq_length for kv cache. So the shape is like [batch_size, kv_num_heads, max_seq_len, head_size] and use I/O binding to bind to past_key and present_key tensors.

Is this use case for webgpu that only allocate total_seq_len buffer for prefill?

Honry: Thanks @tianleiwu, this is a use case for WebNN Z-Image demo, which uses QWen3 as text encoder, and kv caches can be ommited (see the snapshot). Because the GQA's present_kv outputs are required outputs, with current code the shape of present kv caches can be shape inference with `freeDimensionOverrides`. That's the reason I propose this PR.

image

tianleiwu: Could you verify this change using onnxruntime-genai on some models (like gpt-oss-20b) for CUDA or CPU ep. 
I think this change might cause issue on those models.

But I do not have good idea since there is no attribute to indicate whether past and present buffer is shared. So it is little hard to do shape inference without that information. One thing might help is that those models for CPU/CUDA always have both past_key and present_key. If your model has present_key but no past_key, you can change the logic for that condition.

Honry: I tried gpt-oss-20b, the result looks good.

image

guschmue (approved):

PR #27252, opened 2026-02-05
Record service in telemetry events

bmehta001:
### Description
This change records the service name(s), if any, as part of the SessionCreation/ProcessInfo events.
We cache the service names after the first time we calculate them in order to avoid unnecessary overhead.

### Motivation and Context
These changes enable deeper understanding of ORT usage, since multiple services can run inside an application in svchost, which currently obscures our understanding of which services/use cases are most popular. Understanding which services are actually being used can help prioritize more investments in making ORT better targeted to end users.

### Testing

Have tested that the logic in GetServiceNamesForCurrentProcess can accurately return service name for a given process

### DIFF
--- a/onnxruntime/core/platform/windows/telemetry.cc
+++ b/onnxruntime/core/platform/windows/telemetry.cc
@@ -56,16 +56,20 @@ TRACELOGGING_DEFINE_PROVIDER(telemetry_provider_handle, "Microsoft.ML.ONNXRuntim
                              (0x3a26b1ff, 0x7484, 0x7484, 0x74, 0x84, 0x15, 0x26, 0x1f, 0x42, 0x61, 0x4d),
                              TraceLoggingOptionMicrosoftTelemetry());
 
-std::string WideToUtf8(const std::wstring& wide) {
+std::string ConvertWideStringToUtf8(const std::wstring& wide) {
   if (wide.empty())
     return {};
 
-  int utf8_length = ::WideCharToMultiByte(CP_UTF8, 0, wide.data(), static_cast<int>(wide.size()), nullptr, 0, nullptr, nullptr);
+  const UINT code_page = CP_UTF8;
+  const DWORD flags = 0;
+  LPCWCH const src = wide.data();
+  const int src_len = static_cast<int>(wide.size());
+  int utf8_length = ::WideCharToMultiByte(code_page, flags, src, src_len, nullptr, 0, nullptr, nullptr);
   if (utf8_length == 0)
     return {};
 
   std::string utf8(utf8_length, '\0');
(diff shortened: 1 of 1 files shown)


onnxruntime/core/platform/windows/telemetry.cc
```diff
@@ -51,6 +55,75 @@ TRACELOGGING_DEFINE_PROVIDER(telemetry_provider_handle, "Microsoft.ML.ONNXRuntim
...
+    return {};
+
+  std::string utf8(utf8_length, '\0');
+  if (::WideCharToMultiByte(CP_UTF8, 0, wide.data(), static_cast<int>(wide.size()), utf8.data(), utf8_length, nullptr, nullptr) == 0)
+    return {};
+
+  return utf8;
+}
+
+std::string GetServiceNamesForCurrentProcess() {
```
ashrit-ms, line 78: So, this would return a list of all the services currently running, correct? Any way to limit this just to the calling service?

onnxruntime/core/platform/windows/telemetry.cc
```diff
@@ -51,6 +55,75 @@ TRACELOGGING_DEFINE_PROVIDER(telemetry_provider_handle, "Microsoft.ML.ONNXRuntim
...
+
+    if (bytes_needed == 0) {
+      ::CloseServiceHandle(service_manager);
+      return;
+    }
+
+    std::vector<uint8_t> buffer(bytes_needed);
+    auto* services = reinterpret_cast<ENUM_SERVICE_STATUS_PROCESSW*>(buffer.data());
+    services_returned = 0;
+    resume_handle = 0;
```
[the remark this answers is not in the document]
bmehta001: It may be set to a different value in between by `EnumServicesStatusExW`

onnxruntime/core/platform/windows/telemetry.cc
bmehta001: That's correct. The Service Control Manager APIs that allow you to filter for a specific service only take a display name and return a PID; they have no option to query by PID. Without querying using COM/WMI, I don't think there is any way to get a service directly by PID. From what I could find, enumerating all Windows services and matching the PID seems to be the canonical way of getting the service name.

To decrease the cost, I am caching the service name(s) after checking for it. Now, I have further filtered EnumServicesStatusExW to look only for services that are active to decrease the number of services we iterate over. This will save us from looking through inactive services, which we would not be running within in anyway, and will likely have a PID of 0.

ashrit-ms: Sweet, looks good to me.

ashrit-ms (approved):

PR #27253, opened 2026-02-05
 Add per-session thread pool work callbacks API

sdotpeng:
## Description

Adds per-session thread pool work callbacks, allowing callers to hook into the enqueue/start/stop/abandon lifecycle of thread pool work items. The feature is gated behind a build flag (`--enable_session_threadpool_callbacks`) with zero overhead when disabled.

## API additions

- C API: `OrtApi::SetPerSessionThreadPoolCallbacks` — stores an `OrtThreadPoolCallbacksConfig` on the `OrtEnv`, applied to per-session thread pools
- C++ wrapper: `Ort::Env::SetPerSessionThreadPoolCallbacks`
- Versioned C config struct `OrtThreadPoolCallbacksConfig` with fields: `on_enqueue`, `on_start_work`, `on_stop_work`, `on_abandon`, `user_context`
- Four callback typedefs: `OrtThreadPoolWorkEnqueueFn`, `OrtThreadPoolWorkStartFn`, `OrtThreadPoolWorkStopFn`, `OrtThreadPoolWorkAbandonFn`

## Implementation

- `EigenNonBlockingThreadPool.h`: Introduced a policy-based design with two compile-time callback policies:
  - `WorkNoCallbackPolicy`: `Work = std::function<void()>`, all callback methods are trivial inlines eliminated by the compiler. Zero overhead for non-callback builds.
  - `WorkWithCallbackPolicy`: `Work = WorkItem` bundling tasks with callback data; invokes user callbacks around task execution via `MakeWork`/`Execute`/`OnEnqueue`/`OnAbandon` methods.
  - `ThreadPoolTempl<Environment, CallbackPolicy>` uses the policy for all callback-related operations.
  - `RunQueue::RevokeWithTag` calls `policy_->OnAbandon(e.w)` on successful revocation; the policy implementation decides whether to invoke user callbacks.
- `threadpool.h`: `extended_eigen_threadpool_` changed to `unique_ptr<ExtendedThreadPoolInterface>` for type erasure across policy instantiations. `EnableSpinning`/`DisableSpinning` added to the virtual interface.
- `threadpool.cc`: Single `#ifdef` selects policy at `ThreadPoolTempl` instantiation.
- `environment.h/.cc`: Added `SetPerSessionWorkCallbacks`/`GetPerSessionWorkCallbacks` on `Environment`.
- `inference_session.cc`: Propagates callbacks from `Environment` to per-session thread pool options.
- `thread_utils.h/.cc`: Added callback fields to `OrtThreadPoolParams` and wiring in `CreateThreadPoolHelper`.
- `env.h`: `OrtThreadPoolCallbacksConfig*` pointer in `ThreadOptions`.

## Build

- CMake option `onnxruntime_ENABLE_SESSION_THREADPOOL_CALLBACKS`; `build.py` argument `--enable_session_threadpool_callbacks`

## Tests

- 8 callback-specific tests: Schedule, OnEnqueueOnly, NoCallbacks, ParallelFor, ParallelSection, Abandon, EnqueueReturnsNull, NoEnqueueWithStartStop
- End-to-end C API test (`SetPerSessionThreadPoolCallbacks` via ModelBuilder with 1M-element Mul)
- All 73 existing ThreadPool tests pass unchanged with both callback-enabled and callback-disabled builds (81/81 and 73/73 respectively)

## Motivation and Context

Thread pool work callbacks enable telemetry, tracing, and resource management by providing visibility into when work is enqueued, executed, and abandoned in per-session thread pools. This is needed for production diagnostics and performance instrumentation scenarios.

### DIFF
--- a/include/onnxruntime/core/platform/EigenNonBlockingThreadPool.h
+++ b/include/onnxruntime/core/platform/EigenNonBlockingThreadPool.h
@@ -423,9 +423,6 @@ struct WorkNoCallbackPolicy {
   void Execute(const Work& w) const { w(); }
 
   void OnAbandon(const Work&) const noexcept {}
-
- private:
-  void* OnEnqueue() const noexcept { return nullptr; }
 };
 
 #ifdef ORT_ENABLE_SESSION_THREADPOOL_CALLBACKS


sdotpeng:

include/onnxruntime/core/platform/EigenNonBlockingThreadPool.h
```diff
@@ -764,7 +771,23 @@ class ThreadPoolTempl : public onnxruntime::concurrency::ExtendedThreadPoolInter
   };
 
   typedef std::function<void()> Task;
-  typedef RunQueue<Task, Tag, 1024> Queue;
+
+  // WorkItem bundles a task with optional callback data.
+  struct WorkItem {
+    Task task;
+    void* cb_data = nullptr;
```
chwarr:
```suggestion
    void* cb_data_ = nullptr;
```

Minor: it looks like this codebase uses an underscore suffix for members.

Similar comment for `task_`.

include/onnxruntime/core/platform/EigenNonBlockingThreadPool.h
```diff
@@ -764,7 +771,23 @@ class ThreadPoolTempl : public onnxruntime::concurrency::ExtendedThreadPoolInter
...
+
+    WorkItem() = default;
+    explicit WorkItem(Task t, void* data = nullptr) : task(std::move(t)), cb_data(data) {}
+
+    // Allow implicit conversion from Task for compatibility
+    WorkItem(Task t) : task(std::move(t)), cb_data(nullptr) {}
```
chwarr: Minor: this overload can be folded into the previous one by removing the `explicit` keyword from it. Conversion needs a ctor to be invokable with a single argument, and the defaulted `void* data = nullptr` means that the other ctor is invokable with a single argument.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -959,6 +959,33 @@ typedef OrtCustomThreadHandle (*OrtCustomCreateThreadFn)(void* ort_custom_thread
  */
 typedef void (*OrtCustomJoinThreadFn)(OrtCustomThreadHandle ort_custom_thread_handle);
 
+/** \brief Thread pool work enqueue callback
+ *
+ * Called when work is about to be enqueued to a thread pool worker thread.
+ * This runs on the thread that is submitting the work.
+ * \param[in] user_context User-provided context passed when configuring callbacks
+ * \return A callback-specific data pointer that will be passed to OnStartWork and OnStopWork
```
chwarr:
```suggestion
 * \return Callback-specific data that will be passed to OnStartWork and OnStopWork. May return NULL.
```

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -7196,52 +7223,34 @@ struct OrtApi {
...
-   *
-   * \since Version 1.25.
-   */
-  ORT_API2_STATUS(RunOptionsEnableProfiling, _Inout_ OrtRunOptions* options, _In_ const ORTCHAR_T* profile_file_prefix);
-
-  /** \brief Disable profiling for this run
+   * The callbacks follow this pattern:
+   *   1. on_enqueue: Called on the submitting thread when work is enqueued. Returns callback-specific data.
+   *   2. on_start: Called on the worker thread before executing the work.
+   *   3. on_stop: Called on the worker thread after executing the work (guaranteed even on exception).
```
chwarr:
```suggestion
   *   3. on_stop: Called on the worker thread after executing the work. Will be called even if the work fails.
```

Minor: since this is a C API, it seems odd to talk about exceptions.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -959,6 +959,33 @@ typedef OrtCustomThreadHandle (*OrtCustomCreateThreadFn)(void* ort_custom_thread
...
+typedef void* (*OrtThreadPoolWorkEnqueueFn)(void* user_context);
+
+/** \brief Thread pool work start callback
+ *
+ * Called when a worker thread is about to start executing work.
+ * This runs on the worker thread.
+ * \param[in] user_context User-provided context passed when configuring callbacks
+ * \param[in] cb_data Data returned by the corresponding OnEnqueue call
+ */
+typedef void (*OrtThreadPoolWorkStartFn)(void* user_context, void* cb_data);
```
chwarr:
```suggestion
typedef void (*OrtThreadPoolWorkStartFn)(void* user_context, void* enqueue_data);
```

Minor: consider calling this enqueue_data to link it to the return of `OrtThreadPoolWorkEnqueueFn`. Similar comment before for the stop function.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -959,6 +959,33 @@ typedef OrtCustomThreadHandle (*OrtCustomCreateThreadFn)(void* ort_custom_thread
...
 typedef void (*OrtCustomJoinThreadFn)(OrtCustomThreadHandle ort_custom_thread_handle);
 
+/** \brief Thread pool work enqueue callback
+ *
+ * Called when work is about to be enqueued to a thread pool worker thread.
+ * This runs on the thread that is submitting the work.
+ * \param[in] user_context User-provided context passed when configuring callbacks
+ * \return A callback-specific data pointer that will be passed to OnStartWork and OnStopWork
+ */
+typedef void* (*OrtThreadPoolWorkEnqueueFn)(void* user_context);
```
chwarr:
```suggestion
typedef _Ret_maybenull_ void* (*OrtThreadPoolWorkEnqueueFn)(_In_opt_ void* user_context);
```

Minor: it looks like annotations are used in this codebase. Consider putting them on the function pointer typedefs too. Similar comment for the ones below.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -959,6 +959,33 @@ typedef OrtCustomThreadHandle (*OrtCustomCreateThreadFn)(void* ort_custom_thread
  */
 typedef void (*OrtCustomJoinThreadFn)(OrtCustomThreadHandle ort_custom_thread_handle);
 
+/** \brief Thread pool work enqueue callback
+ *
+ * Called when work is about to be enqueued to a thread pool worker thread.
+ * This runs on the thread that is submitting the work.
+ * \param[in] user_context User-provided context passed when configuring callbacks
```
chwarr:
```suggestion
 * \param[in] user_context The user-provided context passed when configuring callbacks
```

Minor: "The" emphasizes that the same value that was passed when configuring will be passed in here. Similar comment below.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -959,6 +959,33 @@ typedef OrtCustomThreadHandle (*OrtCustomCreateThreadFn)(void* ort_custom_thread
...
+ * This runs on the worker thread.
+ * \param[in] user_context User-provided context passed when configuring callbacks
+ * \param[in] cb_data Data returned by the corresponding OnEnqueue call
+ */
+typedef void (*OrtThreadPoolWorkStartFn)(void* user_context, void* cb_data);
+
+/** \brief Thread pool work stop callback
+ *
+ * Called when a worker thread has finished executing work.
+ * This runs on the worker thread, and is guaranteed to be called even if the work throws an exception.
```
chwarr:
```suggestion
 * This runs on the worker thread, and is guaranteed to be called regardless of whether the work finishes successfully or not.
```

Minor: see comment below about "exception" and this being a C API.

include/onnxruntime/core/session/onnxruntime_cxx_api.h
```diff
@@ -848,6 +848,17 @@ struct ThreadingOptions : detail::Base<OrtThreadingOptions> {
...
   /// \brief Wraps OrtApi::SetGlobalCustomJoinThreadFn
   ThreadingOptions& SetGlobalCustomJoinThreadFn(OrtCustomJoinThreadFn ort_custom_join_thread_fn);
+
+  /// \brief Wraps OrtApi::SetGlobalWorkCallbacks
+  /// Configures callbacks invoked around thread pool work execution.
+  /// \param on_enqueue Callback invoked when work is enqueued (on submitting thread). May be nullptr.
+  /// \param on_start Callback invoked when work starts (on worker thread). May be nullptr.
+  /// \param on_stop Callback invoked when work completes (on worker thread). May be nullptr.
+  /// \param user_context User-provided context passed to all callbacks. May be nullptr.
+  ThreadingOptions& SetGlobalWorkCallbacks(OrtThreadPoolWorkEnqueueFn on_enqueue,
```
chwarr:
```suggestion
  ThreadingOptions& SetGlobalWorkCallbacks(_In_opt OrtThreadPoolWorkEnqueueFn on_enqueue,
```

Minor: looks like annotations are used in this code base. Similar comment applies to the other parameters.

include/onnxruntime/core/platform/EigenNonBlockingThreadPool.h
```diff
@@ -816,17 +841,24 @@ class ThreadPoolTempl : public onnxruntime::concurrency::ExtendedThreadPoolInter
...
+    }
+    WorkItem work_item(std::move(fn), cb_data);
+
     PerThread* pt = GetPerThread();
     int q_idx = Rand(&pt->rand) % num_threads_;
     WorkerData& td = worker_data_[q_idx];
     Queue& q = td.queue;
-    fn = q.PushBack(std::move(fn));
-    if (!fn) {
+    work_item = q.PushBack(std::move(work_item));
```
chwarr, line 925:
```suggestion
    WorkItem work_item = q.PushBack(WorkItem {std::move(fn), cb_data});
```

Does this work? It reduces the span over which work_item is alive.

include/onnxruntime/core/platform/EigenNonBlockingThreadPool.h
chwarr: Minor
```suggestion
    void* enqueue_data_ = nullptr;
```

Minor: see other comment about calling this "enqueue data" to indicate its source. Similar comment elsewhere in this file.

include/onnxruntime/core/platform/EigenNonBlockingThreadPool.h
```diff
@@ -1179,6 +1219,12 @@ class ThreadPoolTempl : public onnxruntime::concurrency::ExtendedThreadPoolInter
       if (dispatch_async && extra_needed > 1) {
         assert(current_dop == 1);
 
+        // Create callback data for dispatch task if callbacks are configured
+        void* dispatch_cb_data = nullptr;
```
chwarr:
```suggestion
        void* enqueue_data = nullptr;
```

Minor: see other comment. Also, I'm not sure if prefixing "dispatch" is needed here.

onnxruntime/test/platform/threadpool_test.cc
```diff
@@ -672,4 +672,151 @@ TEST(ThreadPoolTest, TestDefaultAffinity) {
...
+    enqueue_count = 0;
+    start_count = 0;
+    stop_count = 0;
+  }
+};
+
+void* TestOnEnqueue(void* user_context) {
+  auto* ctx = static_cast<WorkCallbackTestContext*>(user_context);
+  ctx->enqueue_count++;
+  return reinterpret_cast<void*>(static_cast<uintptr_t>(0xDEADBEEF));
```
chwarr: Minor: I  wouldn't return 0xDEADBEEF. That value is sometimes used by checker programs to mark memory that has been freeded to detect use-after-free.

0xCB00CB00 could work instead.

onnxruntime/test/platform/threadpool_test.cc
```diff
@@ -672,4 +672,151 @@ TEST(ThreadPoolTest, TestDefaultAffinity) {
...
+// Test for ThreadPoolWorkCallbacks - validates that callbacks are invoked
+// when work is scheduled to the thread pool.
+namespace {
+
+struct WorkCallbackTestContext {
+  std::atomic<int> enqueue_count{0};
+  std::atomic<int> start_count{0};
+  std::atomic<int> stop_count{0};
+
+  void Reset() {
```
chwarr: Minor: `Reset` appears unused and can be removed.

onnxruntime/test/platform/threadpool_test.cc
```diff
@@ -672,4 +672,151 @@ TEST(ThreadPoolTest, TestDefaultAffinity) {
...
+  constexpr int num_tasks = 50;
+  std::atomic<int> tasks_completed{0};
+
+  CreateThreadPoolAndTest("NoCallbacks", 2, [&](ThreadPool* tp) {
+    for (int i = 0; i < num_tasks; i++) {
+      ThreadPool::Schedule(tp, [&]() { tasks_completed++; });
+    }
+  });
+
+  ASSERT_EQ(tasks_completed.load(), num_tasks);
```
chwarr, line 782: Minor: missing assertions that the enqueue, start, and stop counts are zero.

onnxruntime/test/platform/threadpool_test.cc
```diff
@@ -672,4 +672,151 @@ TEST(ThreadPoolTest, TestDefaultAffinity) {
...
+void CreateThreadPoolWithCallbacksAndTest(
+    int num_threads,
+    WorkCallbackTestContext& ctx,
+    bool enable_start_stop,
+    const std::function<void(ThreadPool*)>& test_body) {
+  onnxruntime::ThreadPoolWorkCallbacks callbacks;
+  callbacks.on_enqueue = TestOnEnqueue;
+  callbacks.on_start_work = enable_start_stop ? TestOnStart : nullptr;
+  callbacks.on_stop_work = enable_start_stop ? TestOnStop : nullptr;
+  callbacks.user_context = &ctx;
```
chwarr:
```suggestion
  onnxruntime::ThreadPoolWorkCallbacks callbacks {
    .on_enqueue = TestOnEnqueue,
    .on_start_work = enable_start_stop ? TestOnStart : nullptr,
    .on_stop_work = enable_start_stop ? TestOnStop : nullptr,
    .user_context = &ctx,
```

Minor: designated initializers reduces the duplication.

onnxruntime/test/platform/threadpool_test.cc
```diff
@@ -672,4 +672,151 @@ TEST(ThreadPoolTest, TestDefaultAffinity) {
...
+  auto tp = std::make_unique<ThreadPool>(&onnxruntime::Env::Default(),
+                                         thread_options,
+                                         nullptr,
+                                         num_threads,
+                                         true);
+  test_body(tp.get());
+}
+
+}  // namespace
+
```
chwarr, line 736: Missing tests:
* When the enqueue callback returns nullptr
* When there is no enqueue callback, but start/stop exist(s)

include/onnxruntime/core/platform/EigenNonBlockingThreadPool.h
```diff
@@ -1348,6 +1397,32 @@ class ThreadPoolTempl : public onnxruntime::concurrency::ExtendedThreadPoolInter
     }
   }
 
+  // Invoke a WorkItem, executing callbacks around the task if configured.
+  // This is called from WorkerLoop and Schedule (for rejected work).
+  void InvokeWorkItem(WorkItem& work_item) {
+    void* cb_data = work_item.cb_data;
+    bool has_callbacks = (cb_data != nullptr) && work_callbacks_.has_value();
```
chwarr: The enqueue function is allowed to return nullptr, so that doesn't determine whether the callbacks should be invoked. Just whether there's a start or a stop callback.

include/onnxruntime/core/platform/EigenNonBlockingThreadPool.h
```diff
@@ -1496,6 +1571,7 @@ class ThreadPoolTempl : public onnxruntime::concurrency::ExtendedThreadPoolInter
   const unsigned num_threads_;
   const bool allow_spinning_;
   const bool set_denormal_as_zero_;
+  std::optional<ThreadPoolWorkCallbacks> work_callbacks_;  // Optional callbacks for work scheduling context preservation
```
chwarr: The `<optional>` here doesn't looks like it adds anything. If this were just a ThreadPoolWorkCallbacks struct, then the rest of the logic would be simplier. Each place that a callback might be invoked, check whether one is set. If so, invoke it. Otherwise, skip it.

include/onnxruntime/core/platform/EigenNonBlockingThreadPool.h
```diff
@@ -816,17 +841,24 @@ class ThreadPoolTempl : public onnxruntime::concurrency::ExtendedThreadPoolInter
   // reject work if the queue of pending work is full.
 
   void Schedule(std::function<void()> fn) override {
+    // Create WorkItem with callback data if callbacks are configured
+    void* cb_data = nullptr;
+    if (work_callbacks_.has_value() && work_callbacks_->on_enqueue) {
+      cb_data = work_callbacks_->on_enqueue(work_callbacks_->user_context);
+    }
+    WorkItem work_item(std::move(fn), cb_data);
```
chwarr: Consider adding a helper function that handles checking for the enqueue function and invoking it.

```
void* InvokeOnEnqueue() {
  if (work_callbacks_ && work_callbacks_->on_enqueue) {
    return work_callbacks_->on_enqueue(work_callbacks->user_context);
  } else {
    return nullptr;
  }
}
```

Then every place where on_enqueue is invokes can be simplified:

```suggestion
    // Create WorkItem with callback data if callbacks are configured
    WorkItem work_item(std::move(fn), InvokeOnEnqueue());
```

Since the enqueue logic is done in 3-4 places, this will simplify things.
I don't think similar functions for on_start and on_stop are needed, since InvokeWorkItem is the only place those are called.

chwarr (changes requested):
Looks pretty good. I see some issues around on_enqueue returning NULL.

include/onnxruntime/core/session/onnxruntime_c_api.h
chwarr:
```suggestion
typedef void* (*OrtThreadPoolWorkEnqueueFn)(void* user_context) NO_EXCEPTION;
```

Consider marking these callbacks with `noexcept` when compiling under C++. This will indicate to the code implementing the callback that they MUST NOT allow exceptions to leave the callback.

include/onnxruntime/core/platform/EigenNonBlockingThreadPool.h
```diff
@@ -816,17 +841,24 @@ class ThreadPoolTempl : public onnxruntime::concurrency::ExtendedThreadPoolInter
...
-    fn = q.PushBack(std::move(fn));
-    if (!fn) {
+    work_item = q.PushBack(std::move(work_item));
+    if (!work_item) {
       // The queue accepted the work; ensure that the thread will pick it up
       td.EnsureAwake();
     } else {
       // Run the work directly if the queue rejected the work
-      fn();
+      InvokeWorkItem(work_item);
```
chwarr: If the queue does not accept the work, then the function is going to be run on the current thread. Should on_start and on_stop be called in that case?

So on_enqueue only be called if the queue accepts to work? That is, should the WorkItem be mutated after being stored in the queue? (Does that work with the locking model?)

If the work item is going to be run on the current thread and on_start and on_stop _are not_ run, then there needs to be some way for potential state allocated in on_enqueue to be cleaned up. Does there need to be an on_abandoned?

include/onnxruntime/core/session/onnxruntime_cxx_api.h
sdotpeng: I don't see many functions with SAL annotations in this file. I did add them in C api files tho.

include/onnxruntime/core/platform/EigenNonBlockingThreadPool.h
sdotpeng: I believe mutating the work_item after it is accepted to the queue would break the lock-free queue's memory model (items become `kReady` after push and may be popped by another thread immediately).

Is there disadvantage of running on_start and on_stop on the current thread if the work_item is rejected by the queue?

onnxruntime/test/platform/threadpool_test.cc
sdotpeng: Looks like designated initializers are C++ 20 only. This uses C++ 17.

include/onnxruntime/core/platform/EigenNonBlockingThreadPool.h
chwarr: Yeah, looks like mutating the work item after enqueuing won't work.

Invoking on_start and on_stop when the work item is running on the current thread is fine. It will just need to be clearly documented what the lifecycle is and which callbacks are invoked in which situations.

include/onnxruntime/core/session/onnxruntime_cxx_api.h
```diff
@@ -1290,6 +1290,15 @@ struct Env : detail::Base<OrtEnv> {
...
                      OrtSyncStream* stream) const;  ///< Wraps OrtApi::CopyTensors
+
+  /// \\brief Wraps OrtApi::SetDefaultThreadPoolCallbacks
+  /// Stores default work callbacks on the Env for per-session thread pools.
+  /// Requires ORT built with --session_threadpool_callbacks.
+  Env& SetDefaultThreadPoolCallbacks(OrtThreadPoolWorkEnqueueFn on_enqueue,
+                                     OrtThreadPoolWorkStartFn on_start,
+                                     OrtThreadPoolWorkStopFn on_stop,
+                                     OrtThreadPoolWorkAbandonFn on_abandon,
+                                     void* user_context);
```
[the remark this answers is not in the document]
sdotpeng: Updated PR description.

include/onnxruntime/core/platform/EigenNonBlockingThreadPool.h
sdotpeng: Added `on_abandon` to address this.

include/onnxruntime/core/platform/EigenNonBlockingThreadPool.h
sdotpeng: No longer applicable as we move to compiler flag approach.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -7242,6 +7290,33 @@ struct OrtApi {
...
   ORT_API2_STATUS(RunOptionsDisableProfiling, _Inout_ OrtRunOptions* options);
+
+  /// @}
+  /// \name OrtEnv
+  /// @{
+
+  /** \brief Set default thread pool callbacks for per-session thread pools.
+   *
+   * Stores callbacks on the Env that will be applied to per-session thread pools.
+   * Requires ORT built with --session_threadpool_callbacks.
```
skottmckay: nit: The 'default' is a little off-putting here as there's no way to change these afaics. Would `SetPerSessionThreadPoolCallbacks` be clearer? 

Is my understanding correct that this only affects sessions created after this call? If so it would be good to call that out.

Similarly, should we call out that this does not affect the global threadpools if someone uses CreateEnvWithGlobalThreadPools?

https://github.com/microsoft/onnxruntime/blob/5c3f5449f9c66f51c9c25cd188fec98ad37d010c/onnxruntime/core/session/inference_session.h#L174-L184

Is it a problem if the WinML usage passes in external threadpools which would not have these callbacks added? #Closed

include/onnxruntime/core/platform/EigenNonBlockingThreadPool.h
```diff
@@ -1179,6 +1258,10 @@ class ThreadPoolTempl : public onnxruntime::concurrency::ExtendedThreadPoolInter
       if (dispatch_async && extra_needed > 1) {
         assert(current_dop == 1);
 
+#ifdef ORT_SESSION_THREADPOOL_CALLBACKS
+        void* enqueue_data = InvokeOnEnqueue();
+#endif
```
skottmckay: nit: Does this need to be called before we create `dispatch_task` or could it move down and be inside the following `#ifdef ORT_SESSION_THREADPOOL_CALLBACKS`? #Closed

include/onnxruntime/core/platform/EigenNonBlockingThreadPool.h
```diff
@@ -1348,6 +1441,49 @@ class ThreadPoolTempl : public onnxruntime::concurrency::ExtendedThreadPoolInter
...
+  void* InvokeOnEnqueue() {
+    if (work_callbacks_.on_enqueue) {
+      return work_callbacks_.on_enqueue(work_callbacks_.user_context);
+    }
+    return nullptr;
+  }
+
+  // Invoke a WorkItem, executing callbacks around the task if configured.
+  // This is called from WorkerLoop and Schedule (for rejected work).
+  void InvokeWorkItem(WorkItem& work_item) {
```
skottmckay: `const WorkItem&`? #Closed

onnxruntime/test/platform/threadpool_test.cc
```diff
@@ -672,4 +672,250 @@ TEST(ThreadPoolTest, TestDefaultAffinity) {
...
+  });
+
+  ASSERT_EQ(tasks_completed.load(), num_tasks);
+  // Worker threads get callbacks; main thread's fn(0) does not.
+  // Some enqueued tasks may be revoked before execution (work completed by other threads),
+  // so enqueue_count >= start_count. Start/stop must always be balanced.
+  // Every enqueued item must end with either start+stop or abandon.
+  ASSERT_GE(ctx.enqueue_count.load(), ctx.start_count.load());
+  ASSERT_EQ(ctx.start_count.load(), ctx.stop_count.load());
+  ASSERT_EQ(ctx.enqueue_count.load(), ctx.start_count.load() + ctx.abandon_count.load());
```
skottmckay, line 806: Are any tests guaranteed to have a non-zero abandoned count so we know the path calling `on_abandon` is hit? #Closed

onnxruntime/core/session/inference_session.cc
```diff
@@ -466,6 +466,19 @@ void InferenceSession::ConstructorCommon(const SessionOptions& session_options,
...
 
+#ifdef ORT_SESSION_THREADPOOL_CALLBACKS
+        {
+          const auto* env_cbs = session_env.GetDefaultSessionWorkCallbacks();
+          if (env_cbs != nullptr) {
+            to.work_enqueue_fn = env_cbs->on_enqueue;
+            to.work_start_fn = env_cbs->on_start_work;
+            to.work_stop_fn = env_cbs->on_stop_work;
+            to.work_abandon_fn = env_cbs->on_abandon;
+            to.work_callbacks_user_context = env_cbs->user_context;
```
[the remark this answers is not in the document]
sdotpeng: After consideration, I am removing the mutex and keeping the raw pointer. This follows the same pattern as `SetGlobalThreadingOptions` and `SetLoggingManager` in this class - both are one-time setup functions that are expected to be called before any sessions are created, and neither uses locking. The callbacks are set once during environment setup via `SetDefaultThreadPoolCallbacks` and read during `InferenceSession::ConstructorCommon`; there's no realistic concurrent access.

include/onnxruntime/core/platform/EigenNonBlockingThreadPool.h
sdotpeng: This doesn't. Moved it to the following compiler guard.

yuslepukhin:
The PR makes TP perf dependent on the nature of callbacks. Is there any perf numbers that could characterize the perf under the supposed usage?

onnxruntime/test/platform/threadpool_test.cc
sdotpeng: I added a new test that guarantees to have a non-zero abandoned count.

include/onnxruntime/core/session/onnxruntime_c_api.h
sdotpeng: Yes, I updated the new name to be `SetPerSessionThreadPoolCallbacks`, and added two comments about no edits after session creation, and does not configure global thread pool. 

On the WinML side, that shouldn't be an issue.

onnxruntime/core/session/onnxruntime_c_api.cc
```diff
@@ -4260,6 +4260,30 @@ ORT_API_STATUS_IMPL(OrtApis::SessionGetMemoryInfoForOutputs, _In_ const OrtSessi
...
+  ORT_UNUSED_PARAMETER(on_enqueue);
+  ORT_UNUSED_PARAMETER(on_start);
+  ORT_UNUSED_PARAMETER(on_stop);
+  ORT_UNUSED_PARAMETER(on_abandon);
+  ORT_UNUSED_PARAMETER(user_context);
+  return OrtApis::CreateStatus(ORT_NOT_IMPLEMENTED,
+      "SetDefaultThreadPoolCallbacks requires ORT built with --session_threadpool_callbacks");
+#endif
+  API_IMPL_END
+}
```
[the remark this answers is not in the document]
sdotpeng: Added a new C API test to test end-to-end.

sdotpeng:
> The PR makes TP perf dependent on the nature of callbacks. Is there any perf numbers that could characterize the perf under the supposed usage?

Yes, the overhead does depend on the callback implementation. We benchmarked the three relevant configurations to characterize this:
1. **Flag OFF**: baseline, identical to ORT main
2. **Flag ON, no callbacks registered**: measures structural overhead (extra pointer in work item, null-check branches)
3. **Flag ON, WinML callbacks registered**: the intended usage, where each callback performs a lightweight NT kernel call on thread-local state

**Methodology**: Built ARM64 Release with `--build_micro_benchmarks` for each configuration. Ran `onnxruntime_benchmark.exe` threadpool microbenchmarks (`BM_ThreadPoolParallelFor`) with `--benchmark_repetitions=3` on Snapdragon X Elite (12 cores, 2976 MHz). The benchmark parameters are iteration count (work volume) and cost (per-element cost that controls work partitioning across threads).

**Results** (mean real\_time, percentages relative to Flag OFF):

| Benchmark | Flag OFF (ns) | Flag ON, no callbacks (ns) | Flag ON, WinML callbacks (ns) |
|---|---|---|---|
| ParallelFor 100/1 | 236 | 260 (+10%) | 399 (+69%) |
| ParallelFor 100/400 | 225 | 253 (+12%) | 328 (+46%) |
| ParallelFor 1K/1 | 2,114 | 2,229 (+5%) | 2,853 (+35%) |
| ParallelFor 1K/200 | 2,046 | 2,237 (+9%) | 2,527 (+24%) |
| ParallelFor 10K/1 | 19,516 | 22,026 (+13%) | 22,624 (+16%) |
| ParallelFor 10K/200 | 30,423 | 30,031 (-1%) | 32,546 (+7%) |
| ParallelFor 20K/200 | 57,804 | 58,457 (+1%) | 61,884 (+7%) |
| ParallelFor 40K/200 | 114,663 | 119,569 (+4%) | 123,416 (+8%) |
| ParallelFor 80K/200 | 225,654 | 233,793 (+4%) | 248,085 (+10%) |
| ParallelFor 160K/200 | 454,959 | 475,879 (+5%) | 463,962 (+2%) |

Flag OFF and Flag ON numbers were collected in separate runs, so +/-5% variation is expected run-to-run noise.

**Summary**: The callback overhead is a fixed 100-500ns per dispatch from three kernel calls per work item. On very short loops (100 iterations, total time 230ns), this dominates. On realistic workloads (10K+ iterations), the overhead is 2-10%. In practice, ORT inference kernels run in the hundreds-of-microseconds to milliseconds range, making the per-dispatch callback cost negligible. Builds without `--session_threadpool_callbacks` have exactly zero overhead — the feature is entirely compiled out.

yuslepukhin:
As a general comment, I do not like the number of #ifdefs in the code.
Would it be possible to create an abstraction such a policy building block that can be substituted?
The perf impact is measurable.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -959,6 +959,54 @@ typedef OrtCustomThreadHandle (*OrtCustomCreateThreadFn)(void* ort_custom_thread
  */
 typedef void (*OrtCustomJoinThreadFn)(OrtCustomThreadHandle ort_custom_thread_handle);
 
+#ifdef ORT_SESSION_THREADPOOL_CALLBACKS
```
yuslepukhin: This macro is valid for internal builds. The client code is not going to define it.
Thus, you will have a build w/o callbacks, but the header will still define the structures.
You do not want these macros in the client facing headers.
The entry points should exist regardless, but return unsupported for builds where callbacks are not present. #Resolved

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -959,6 +959,54 @@ typedef OrtCustomThreadHandle (*OrtCustomCreateThreadFn)(void* ort_custom_thread
...
+ *
+ * Called when enqueued work is abandoned (revoked or rejected) without execution.
+ * This allows the caller to free any resources associated with enqueue_data.
+ * \param[in] user_context The user-provided context passed when configuring callbacks
+ * \param[in] enqueue_data Data returned by the corresponding OnEnqueue call
+ */
+typedef void (*OrtThreadPoolWorkAbandonFn)(_In_opt_ void* user_context, _In_opt_ void* enqueue_data) NO_EXCEPTION;
+#else
+// Typedefs must exist unconditionally because the OrtApi struct references them.
+// The implementation returns ORT_NOT_IMPLEMENTED when the flag is not defined.
```
yuslepukhin: This is a strange #else if they should always be defined.
See my other comment regarding user facing headers. #Resolved

onnxruntime/core/platform/env.h
```diff
@@ -90,6 +113,12 @@ struct ThreadOptions {
   void* custom_thread_creation_options = nullptr;
   OrtCustomJoinThreadFn custom_join_thread_fn = nullptr;
   int dynamic_block_base_ = 0;
+
+#ifdef ORT_SESSION_THREADPOOL_CALLBACKS
+  // Optional callbacks for thread pool work scheduling.
+  // The pointed-to struct must outlive the ThreadPool.
+  const ThreadPoolWorkCallbacks* work_callbacks = nullptr;
+#endif
```
yuslepukhin, line 99: So this is owned by Environment member and thus outlives the TP lifespan.

Curious, what happens if I:
1) Set the callbacks
2) Create a session with pools that are going to use the callbacks
3) Set other callbacks.

My understanding the original ThreadPoolWorkCallbacks structure is going to be re-created? Is the existing session going to point to callbacks OR to a ThreadPoolWorkCallbacks  instance? #Resolved

onnxruntime/core/common/threadpool.cc
```diff
@@ -391,11 +391,15 @@ ThreadPool::ThreadPool(Env* env,
     }
 
     extended_eigen_threadpool_ =
-        std::make_unique<ThreadPoolTempl<Env> >(name,
-                                                threads_to_create,
-                                                low_latency_hint,
-                                                *env,
-                                                thread_options_);
+#ifdef ORT_SESSION_THREADPOOL_CALLBACKS
+        std::make_unique<ThreadPoolTempl<Env, WithWorkCallbackPolicy> >(name,
```
yuslepukhin: >       std::make_unique<ThreadPoolTempl<Env, WithWorkCallbackPolicy [](http://example.com/codeflow?start=1&length=67)

Suggest to #ifdef the type of the TP only and make a single no-broken statement of make_unque for readability. #Resolved

onnxruntime/core/session/inference_session.cc
```diff
@@ -502,6 +515,20 @@ void InferenceSession::ConstructorCommon(const SessionOptions& session_options,
...
+          const auto* env_cbs = session_env.GetPerSessionWorkCallbacks();
+          if (env_cbs != nullptr) {
+            to.work_enqueue_fn = env_cbs->on_enqueue;
+            to.work_start_fn = env_cbs->on_start_work;
+            to.work_stop_fn = env_cbs->on_stop_work;
+            to.work_abandon_fn = env_cbs->on_abandon;
+            to.work_callbacks_user_context = env_cbs->user_context;
+          }
+        }
+#endif
```
yuslepukhin, line 522: Can we avoid code duplication and making this constructor longer? #Resolved

include/onnxruntime/core/platform/EigenNonBlockingThreadPool.h
```diff
@@ -404,6 +412,96 @@ class ThreadPoolLoop {
...
 };
 
+// Type trait to detect Work types that carry enqueue_data_ (used by RunQueue).
+template <typename T, typename = void>
+struct HasEnqueueData : std::false_type {};
+
+template <typename T>
+struct HasEnqueueData<T, std::void_t<decltype(std::declval<T>().enqueue_data_)>> : std::true_type {};
+
+// Callback policy that does nothing. Work = std::function<void()>.
```
yuslepukhin: > nothing [](http://example.com/codeflow?start=28&length=8)

does not invoke callbacks #Resolved

include/onnxruntime/core/platform/EigenNonBlockingThreadPool.h
```diff
@@ -404,6 +412,96 @@ class ThreadPoolLoop {
   ORT_DISALLOW_COPY_ASSIGNMENT_AND_MOVE(ThreadPoolLoop);
 };
 
+// Type trait to detect Work types that carry enqueue_data_ (used by RunQueue).
+template <typename T, typename = void>
+struct HasEnqueueData : std::false_type {};
+
+template <typename T>
+struct HasEnqueueData<T, std::void_t<decltype(std::declval<T>().enqueue_data_)>> : std::true_type {};
```
yuslepukhin: >HasEnqueueData [](http://example.com/codeflow?start=7&length=14)

Why not to specialize on the policy itself? #Resolved

include/onnxruntime/core/platform/EigenNonBlockingThreadPool.h
```diff
@@ -549,7 +647,7 @@ class RunQueue {
   // the caller must assume that it may still execute, for instance because it
   // has been pop'd from the queue concurrent with the revocation request.
 
-  bool RevokeWithTag(Tag tag, unsigned w_idx) {
+  bool RevokeWithTag(Tag tag, unsigned w_idx, void** out_enqueue_data = nullptr) {
```
yuslepukhin: >out_enqueue_data [](http://example.com/codeflow?start=53&length=16)

Must be documented #Resolved

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -959,6 +959,54 @@ typedef OrtCustomThreadHandle (*OrtCustomCreateThreadFn)(void* ort_custom_thread
...
 
+#ifdef ORT_SESSION_THREADPOOL_CALLBACKS
+/** \brief Thread pool work enqueue callback
+ *
+ * Called when work is about to be enqueued to a thread pool worker thread.
+ * This runs on the thread that is submitting the work.
+ * \param[in] user_context The user-provided context passed when configuring callbacks
+ * \return Callback-specific data that will be passed to OnStartWork and OnStopWork. May return NULL.
+ */
+typedef _Ret_maybenull_ void* (*OrtThreadPoolWorkEnqueueFn)(_In_opt_ void* user_context)NO_EXCEPTION;
```
yuslepukhin, line 971: Consider wrapping these callbacks into a versioned struct for extensibility. Then API may take a struct as well. #Resolved

yuslepukhin, line 971: >NO_EXCEPTION [](http://example.com/codeflow?start=88&length=12)

Should we use ORT_NO_EXCEPTION?
I understand this is contrary to the style of the header, but better in a long run. #Resolved

onnxruntime/core/platform/env.h
sdotpeng: Yes, calling SetPerSessionThreadPoolCallbacks again replaces the ThreadPoolWorkCallbacks in the Environment. But existing sessions are unaffected — the callbacks flow by value copy at each step: Environment → OrtThreadPoolParams fields → stack-local ThreadPoolWorkCallbacks in CreateThreadPoolHelper → WithWorkCallbackPolicy::callbacks_ member (value copy in Init). Each thread pool owns its own copy of the function pointers. It does not hold a pointer back to the Environment's struct.

onnxruntime/core/session/inference_session.cc
sdotpeng: Added a local lambda function and used it for intra-op and inter-op.

onnxruntime/core/common/threadpool.cc
sdotpeng: Good idea. Thanks!

include/onnxruntime/core/session/onnxruntime_c_api.h
sdotpeng: Done - the callback typedefs, OrtThreadPoolCallbacksConfig, and the SetPerSessionThreadPoolCallbacks entry in OrtApi are all unconditionally defined in the header (no #ifdef). The compile flag only gates the internal implementation in onnxruntime_c_api.cc, which returns ORT_NOT_IMPLEMENTED when callbacks aren't enabled.

include/onnxruntime/core/session/onnxruntime_c_api.h
sdotpeng: Should be good for macro collision. But only changing few instances related to my changes would make the header inconsistent. I'd prefer another PR to bulk edit the NO_EXCEPTION.

include/onnxruntime/core/platform/EigenNonBlockingThreadPool.h
sdotpeng: No longer applicable since the new commit. RevokeWithTag now does not take an output ptr for enqueue data.

include/onnxruntime/core/platform/EigenNonBlockingThreadPool.h
sdotpeng: Discussed offline - removing this and templating RunQueue on CallbackPolicy for better readability for various abandon paths.

onnxruntime/core/platform/env.h
```diff
@@ -61,6 +61,29 @@ class EnvThread {
 /// Type that holds a collection of logical processors IDs used for setting affinities.
 using LogicalProcessors = std::vector<int>;
 
+#ifdef ORT_SESSION_THREADPOOL_CALLBACKS
+/// Callbacks for thread pool work scheduling.
+struct ThreadPoolWorkCallbacks {
```
yuslepukhin: >struct ThreadPoolWorkCallbacks  [](http://example.com/codeflow?start=0&length=31)

Would it be possible to re-use the struct from the header? #Resolved

include/onnxruntime/core/platform/EigenNonBlockingThreadPool.h
```diff
@@ -404,9 +410,96 @@ class ThreadPoolLoop {
   ORT_DISALLOW_COPY_ASSIGNMENT_AND_MOVE(ThreadPoolLoop);
 };
 
-template <typename Work, typename Tag, unsigned kSize>
+// Callback policy that does not invoke callbacks. Work = std::function<void()>.
+// All methods are trivial no-ops; the compiler eliminates them entirely.
+struct NoWorkCallbackPolicy {
```
yuslepukhin: >NoWorkCallbackPolicy  [](http://example.com/codeflow?start=7&length=21)

Suggest WorkNoCallbackPolicy #Resolved

include/onnxruntime/core/platform/EigenNonBlockingThreadPool.h
```diff
@@ -404,9 +410,96 @@ class ThreadPoolLoop {
...
+
+  void* OnEnqueue() const { return nullptr; }
+
+  void OnAbandon(const Work&) const {}
+};
+
+#ifdef ORT_SESSION_THREADPOOL_CALLBACKS
+// Callback-aware policy that invokes user-supplied callbacks around work items.
+// Work = WorkItem, a struct bundling a task with opaque callback data.
+struct WithWorkCallbackPolicy {
```
yuslepukhin: >WithWorkCallbackPolicy [](http://example.com/codeflow?start=7&length=22)

WorkWithCallbackPolicy #Resolved

include/onnxruntime/core/platform/EigenNonBlockingThreadPool.h
```diff
@@ -404,9 +410,96 @@ class ThreadPoolLoop {
...
+  using Task = std::function<void()>;
+
+  struct WorkItem {
+    Task task_;
+    void* enqueue_data_ = nullptr;
+
+    WorkItem() = default;
```
yuslepukhin: >WorkItem() = default; [](http://example.com/codeflow?start=4&length=21)

Do we need default __ctor? #Resolved

include/onnxruntime/core/platform/EigenNonBlockingThreadPool.h
```diff
@@ -404,9 +410,96 @@ class ThreadPoolLoop {
...
+    explicit WorkItem(Task t, void* data = nullptr) : task_(std::move(t)), enqueue_data_(data) {}
+
+    explicit operator bool() const { return static_cast<bool>(task_); }
```
yuslepukhin, line 440: >explicit [](http://example.com/codeflow?start=4&length=8)

What is the reasoning behind explicit? #Resolved

include/onnxruntime/core/platform/EigenNonBlockingThreadPool.h
```diff
@@ -338,6 +342,8 @@ class ExtendedThreadPoolInterface : public Eigen::ThreadPoolInterface {
                              unsigned n, std::ptrdiff_t block_size) = 0;
   virtual void StartProfiling() = 0;
   virtual std::string StopProfiling() = 0;
+  virtual void EnableSpinning() = 0;
+  virtual void DisableSpinning() = 0;
```
yuslepukhin, line 346: Where these two are used? #Resolved

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -959,6 +959,63 @@ typedef OrtCustomThreadHandle (*OrtCustomCreateThreadFn)(void* ort_custom_thread
...
+ *
+ * \since Version 1.25.
+ */
+typedef struct OrtThreadPoolCallbacksConfig {
+  uint32_t version;                          /**< Must be ORT_API_VERSION */
+  OrtThreadPoolWorkEnqueueFn on_enqueue;     /**< Called when work is enqueued. May be NULL. */
+  OrtThreadPoolWorkStartFn on_start;         /**< Called when work starts. May be NULL. */
+  OrtThreadPoolWorkStopFn on_stop;           /**< Called when work completes. May be NULL. */
+  OrtThreadPoolWorkAbandonFn on_abandon;     /**< Called when work is abandoned. May be NULL. */
+  void* user_context;                        /**< User-provided context passed to all callbacks. May be NULL. */
```
yuslepukhin: >/**< User-provided context passed to all callbacks. May be NULL. */ [](http://example.com/codeflow?start=45&length=67)

Should we document lifetime requirements for user_context?

Also document that user_context is a subject to multiple threads concurrent invocations and is expected to be thread safe. This may also affect inference performance.

For the struct as a whole. Document that any new members in the future MUST be added at the end only. #Resolved

include/onnxruntime/core/platform/EigenNonBlockingThreadPool.h
sdotpeng: To prevent silent implicit conversions to bool/int in arithmetic or function arguments while still working in contextual boolean conversions (if, !, etc). All our usage sites (if (!w), if (w)) are contextual conversions and should work fine with it.

include/onnxruntime/core/platform/EigenNonBlockingThreadPool.h
sdotpeng: They are not used in my PR. I moved there to ExtendedThreadPoolInterface as part of the policy-based refactor: threadpool.cc now holds the pool via the interface pointer (since ThreadPoolTempl gained a second template parameter for CallbackPolicy), so these needed to be virtual to remain callable.

include/onnxruntime/core/platform/EigenNonBlockingThreadPool.h
sdotpeng: Yes, RunQueue's `Elem` array and sentinel returns `(Work(), e.w = Work())` require a default constructor. I am consolidating both ctors into one by giving `Task t` a default value of `{}`.

include/onnxruntime/core/session/onnxruntime_c_api.h
sdotpeng: Added comments.

onnxruntime/core/platform/env.h
sdotpeng: Yes. Replaced this struct with the struct from C API header.

include/onnxruntime/core/platform/EigenNonBlockingThreadPool.h
```diff
@@ -404,9 +410,95 @@ class ThreadPoolLoop {
...
+
+    explicit WorkItem(Task t = {}, void* data = nullptr) : task_(std::move(t)), enqueue_data_(data) {}
+
+    explicit operator bool() const { return static_cast<bool>(task_); }
+  };
+
+  using Work = WorkItem;
+
+  // Init is called from the ThreadPoolTempl constructor.
+  // Only instantiated when ORT_SESSION_THREADPOOL_CALLBACKS is defined.
```
chwarr: Not sure this comment "only instantiated when..." adds much. This is all already inside the `#idfef`

include/onnxruntime/core/platform/EigenNonBlockingThreadPool.h
```diff
@@ -404,9 +410,95 @@ class ThreadPoolLoop {
...
+        callbacks_.on_stop_work(callbacks_.user_context, data);
+      }
+      ORT_RETHROW;
+    }
+    if (callbacks_.on_stop_work) {
+      callbacks_.on_stop_work(callbacks_.user_context, data);
+    }
+  }
+
+  void* OnEnqueue() const {
```
chwarr: Minor: mark this function and OnAbandon as noexcept (since the callbacks will never throw).

chwarr: Minor: OnEnqueue looks like it can be private.

include/onnxruntime/core/platform/EigenNonBlockingThreadPool.h
```diff
@@ -404,9 +410,95 @@ class ThreadPoolLoop {
...
+    explicit WorkItem(Task t = {}, void* data = nullptr) : task_(std::move(t)), enqueue_data_(data) {}
+
+    explicit operator bool() const { return static_cast<bool>(task_); }
+  };
+
+  using Work = WorkItem;
+
+  // Init is called from the ThreadPoolTempl constructor.
+  // Only instantiated when ORT_SESSION_THREADPOOL_CALLBACKS is defined.
+  void Init(const onnxruntime::ThreadOptions& opts) {
```
chwarr: It looks like the Init function can instead be the ctor.

In ThreadPoolTempl, instead of calling `callback_policy_.Init(thread_options)`, add a line like this to the initialization list:

```cpp
callback_policy_(thread_options)
```

include/onnxruntime/core/session/environment.h
```diff
@@ -185,6 +186,29 @@ class Environment {
...
+  /**
+   * Returns the per-session thread pool work callbacks, or nullptr if not set.
+   *
+   * Not safe to call concurrently with SetPerSessionWorkCallbacks.
+   * Must be called after callbacks have been set and before they are modified.
+   */
+  const OrtThreadPoolCallbacksConfig* GetPerSessionWorkCallbacks() const {
+    return per_session_work_callbacks_.has_value()
+               ? &per_session_work_callbacks_.value()
+               : nullptr;
```
chwarr, line 198:
```suggestion
    return per_session_work_callbacks_.value_or(nullptr);
```

include/onnxruntime/core/session/environment.h
```diff
@@ -185,6 +186,29 @@ class Environment {
   /// <returns></returns>
   OrtKeyValuePairs GetConfigEntries() const;
 
+#ifdef ORT_SESSION_THREADPOOL_CALLBACKS
+  /**
+   * Returns the per-session thread pool work callbacks, or nullptr if not set.
+   *
+   * Not safe to call concurrently with SetPerSessionWorkCallbacks.
+   * Must be called after callbacks have been set and before they are modified.
```
chwarr: I don't understand what this "Must be...modified" comment is trying to say.
It's safe to call this before callbacks have been set. It will return nullptr...

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -959,6 +959,67 @@ typedef OrtCustomThreadHandle (*OrtCustomCreateThreadFn)(void* ort_custom_thread
  */
 typedef void (*OrtCustomJoinThreadFn)(OrtCustomThreadHandle ort_custom_thread_handle);
 
+/** \brief Thread pool work enqueue callback
+ *
+ * Called when work is about to be enqueued to a thread pool worker thread.
+ * This runs on the thread that is submitting the work.
+ * \param[in] user_context The user-provided context passed when configuring callbacks
+ * \return Callback-specific data that will be passed to OnStartWork and OnStopWork. May return NULL.
```
chwarr:
```suggestion
 * \return Callback-specific data that will be passed to the configured OrtThreadPoolCallbacksConfig.on_start_work and OrtThreadPoolCallbacksConfig.on_stop_work. May return NULL.
```

These need to be documented in terms of public APIs, not the implementation in EigenNonBlockingThreadpool. Similar comment for the other doc comments.

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -959,6 +959,67 @@ typedef OrtCustomThreadHandle (*OrtCustomCreateThreadFn)(void* ort_custom_thread
...
+ */
+typedef _Ret_maybenull_ void* (*OrtThreadPoolWorkEnqueueFn)(_In_opt_ void* user_context)NO_EXCEPTION;
```
chwarr, line 971:
```suggestion
typedef _Ret_maybenull_ void* (*OrtThreadPoolWorkEnqueueFn)(_In_opt_ void* user_context) NO_EXCEPTION;
```

Super minor: missing a space

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -7327,6 +7388,30 @@ struct OrtApi {
...
+   *
+   * \param[in] env OrtEnv instance.
+   * \param[in] config Versioned callbacks configuration. ORT copies the struct contents.
+   *
+   * \snippet{doc} snippets.dox OrtStatus Return Value
+   *
+   * \since Version 1.25.
+   */
+  ORT_API2_STATUS(SetPerSessionThreadPoolCallbacks, _Inout_ OrtEnv* env,
+                  _In_ const OrtThreadPoolCallbacksConfig* config);
```
chwarr, line 7417: Check what other functions doc comment do: might want to say that the config is copied and doesn't need to live for the length of the session. (But the things that config points to _do_ need to live for the duration of the session.)

include/onnxruntime/core/session/onnxruntime_cxx_api.h
```diff
@@ -1290,6 +1290,12 @@ struct Env : detail::Base<OrtEnv> {
   Status CopyTensors(const std::vector<Value>& src_tensors,
                      const std::vector<Value>& dst_tensors,
                      OrtSyncStream* stream) const;  ///< Wraps OrtApi::CopyTensors
+
+  /// \brief Wraps OrtApi::SetPerSessionThreadPoolCallbacks
+  /// Stores work callbacks on the Env for per-session thread pools.
+  /// Only affects sessions created after this call. Does not affect global thread pools.
+  /// Requires ORT built with --session_threadpool_callbacks.
+  Env& SetPerSessionThreadPoolCallbacks(const OrtThreadPoolCallbacksConfig& config);
```
chwarr, line 1298: Check how other functions are documented: does this need a `/since`?

onnxruntime/core/session/onnxruntime_c_api.cc
```diff
@@ -4260,6 +4260,28 @@ ORT_API_STATUS_IMPL(OrtApis::SessionGetMemoryInfoForOutputs, _In_ const OrtSessi
...
+    return OrtApis::CreateStatus(ORT_INVALID_ARGUMENT, "config must not be null");
+  }
+  if (config->version == 0 || config->version > ORT_API_VERSION) {
+    return OrtApis::CreateStatus(ORT_INVALID_ARGUMENT,
+                                 "OrtThreadPoolCallbacksConfig requires version set to ORT_API_VERSION");
+  }
+#ifdef ORT_SESSION_THREADPOOL_CALLBACKS
+  auto& env = ort_env->GetEnvironment();
+  return onnxruntime::ToOrtStatus(
+      env.SetPerSessionWorkCallbacks(*config));
```
chwarr:
```suggestion
  return onnxruntime::ToOrtStatus(ort_env->GetEnvironment().SetPerSessionWorkCallbacks(*config));
```

This avoids any temporary lifetime issues. I don't know the semantics of that `GetEnvironment()` returns, but code like this makes it moot.

onnxruntime/test/platform/threadpool_test.cc
```diff
@@ -672,4 +672,302 @@ TEST(ThreadPoolTest, TestDefaultAffinity) {
...
+    int num_threads,
+    WorkCallbackTestContext& ctx,
+    bool enable_start_stop,
+    const std::function<void(ThreadPool*)>& test_body) {
+  OrtThreadPoolCallbacksConfig callbacks{};
+  callbacks.on_enqueue = TestOnEnqueue;
+  callbacks.on_start_work = enable_start_stop ? TestOnStart : nullptr;
+  callbacks.on_stop_work = enable_start_stop ? TestOnStop : nullptr;
+  callbacks.on_abandon = enable_start_stop ? TestOnAbandon : nullptr;
+  callbacks.user_context = &ctx;
```
chwarr, line 722:
```suggestion
  OrtThreadPoolCallbacksConfig callbacks{
    .on_enqueue = TestOnEnqueue,
    .on_start_work = enable_start_stop ? TestOnStart : nullptr,
    .on_stop_work = enable_start_stop ? TestOnStop : nullptr,
    .on_abandon = enable_start_stop ? TestOnAbandon : nullptr,
    .user_context = &ctx };
```

Minor: designated initializers mean the variable name doesn't need to be repeated.

onnxruntime/test/platform/threadpool_test.cc
```diff
@@ -672,4 +672,302 @@ TEST(ThreadPoolTest, TestDefaultAffinity) {
...
+
+  onnxruntime::ThreadOptions thread_options;
+  thread_options.work_callbacks = &callbacks;
```
chwarr, line 725:
```suggestion
  onnxruntime::ThreadOptions thread_options { .work_callbacks = &callbacks };
```

Minor: designated initializer here too.

onnxruntime/test/platform/threadpool_test.cc
```diff
@@ -672,4 +672,302 @@ TEST(ThreadPoolTest, TestDefaultAffinity) {
...
+
+    // Reset counters so we only measure the parallel loop below.
+    ctx.enqueue_count = 0;
+    ctx.start_count = 0;
+    ctx.stop_count = 0;
+    ctx.abandon_count = 0;
+
+    // The parallel loop enqueues a dispatch task, but no worker can pick it up.
+    // The main thread completes all iterations, then EndParallelSection revokes
+    // the dispatch task, triggering on_abandon.
```
chwarr, line 869: abandon_count is asserted to be 0 below, so nothing is being abandoned. I'm not following this test... Can you clarify?

onnxruntime/test/shared_lib/test_env_creation.cc
```diff
@@ -86,3 +87,103 @@ TEST(EnvCreation, CreateEnvWithOptions) {
...
+    static_cast<CallbackState*>(ctx)->abandon_count++;
+  };
+
+  ortenv_teardown();
+
+  auto run_test = [&]() {
+    CallbackState state;
+
+    Ort::Env env(ORT_LOGGING_LEVEL_WARNING, "ThreadPoolCallbacksTest");
+    OrtThreadPoolCallbacksConfig cb_config = {0};
```
chwarr, line 121: Minor: designated initializer could be used here.

cmake/CMakeLists.txt
```diff
@@ -169,6 +169,7 @@ option(onnxruntime_DISABLE_FLOAT4_TYPES "Disable float 4 types" OFF)
 option(onnxruntime_DISABLE_STRING_TYPE "Disable string type" OFF)
 option(onnxruntime_MINIMAL_BUILD "Exclude as much as possible from the build. Support ORT format models. No support for ONNX format models." OFF)
 option(onnxruntime_CLIENT_PACKAGE_BUILD "Enables default settings that are more appropriate for client/on-device workloads." OFF)
+option(onnxruntime_SESSION_THREADPOOL_CALLBACKS "Enable per-session thread pool work callbacks." OFF)
```
chwarr:
```suggestion
option(onnxruntime_ENABLE_SESSION_THREADPOOL_CALLBACKS "Enable per-session thread pool work callbacks." OFF)
```

Minor: looks like the naming scheme here is ENABLE_whatever or USE_whatever. ENABLE sounds better to me here, since it allows these callbacks to be set, but doesn't require/set them by default.

include/onnxruntime/core/platform/EigenNonBlockingThreadPool.h
```diff
@@ -404,9 +410,95 @@ class ThreadPoolLoop {
...
+template <typename CallbackPolicy, typename Tag, unsigned kSize>
 class RunQueue {
  public:
+  using Work = typename CallbackPolicy::Work;
+
+  void SetPolicy(CallbackPolicy* p) { policy_ = p; }
+
   RunQueue() : front_(0), back_(0) {
     // require power-of-two for fast masking
     assert((kSize & (kSize - 1)) == 0);
```
chwarr, line 501: Unrelated to this change, but this could use static_assert to fail at compile time.

include/onnxruntime/core/session/environment.h
sdotpeng: Can't use `value_or` here - it returns by value, not a pointer, and the fallback type `nullptr` doesn't match `OrtThreadPoolCallbacksConfig`.

include/onnxruntime/core/session/onnxruntime_c_api.h
sdotpeng: Fixed. Updated all callback typedef doc comments to reference the public `OrtThreadPoolCallbacksConfig` field names instead of internal implementation names.

include/onnxruntime/core/session/onnxruntime_c_api.h
sdotpeng: This was forced by the lintrunner. Copilot suggested it might be a clang-format artifact. The `_Ret_maybenull_` SAL annotation causes it to remove the space befor `NO_EXCEPTION`.

include/onnxruntime/core/session/onnxruntime_cxx_api.h
sdotpeng: Looks like C++ wrapper methods don't use `\since`.

onnxruntime/test/platform/threadpool_test.cc
sdotpeng: Designated initializers require C++20; ORT targets C++17 so we can't use them here.

Edit: ORT defaults to C++17 (C++20 only on macOS), so designated initializers aren't portable across all targets.

onnxruntime/test/platform/threadpool_test.cc
sdotpeng: Same as above.

onnxruntime/test/platform/threadpool_test.cc
sdotpeng: `ASSERT_GT(ctx.abandon_count.load(), 0)` on line 879 asserts abandon_count > 0, not == 0. The test blocks all workers, forces the main thread to do all work, then verifies that the unexecuted dispatch tasks are properly abandoned.

onnxruntime/test/shared_lib/test_env_creation.cc
sdotpeng: Same as the comment in `threadpool_test.cc`.

onnxruntime/test/platform/threadpool_test.cc
yuslepukhin: We are now on C++20.

cmake/CMakeLists.txt
sdotpeng: Edited the compiler flag to be --enable-session-threadpool-callbacks.

include/onnxruntime/core/session/onnxruntime_c_api.h
sdotpeng: Updated comments.

include/onnxruntime/core/platform/EigenNonBlockingThreadPool.h
```diff
@@ -416,18 +416,19 @@ struct WorkNoCallbackPolicy {
...
   Work MakeWork(Task fn) const { return fn; }
 
   void Execute(const Work& w) const { w(); }
 
-  void* OnEnqueue() const { return nullptr; }
+  void OnAbandon(const Work&) const noexcept {}
 
-  void OnAbandon(const Work&) const {}
+ private:
+  void* OnEnqueue() const noexcept { return nullptr; }
```
chwarr: Minor: nothing appears to call `WorkNoCallbackPolicy::OnEnqueue`, so it can be deleted.

onnxruntime/test/platform/threadpool_test.cc
chwarr: Ah, I missed the GT vs EQ. Thanks.

include/onnxruntime/core/platform/EigenNonBlockingThreadPool.h
sdotpeng: Deleted `WorkNoCallbackPolicy::OnEnqueue`.

PR #27256, opened 2026-02-05
Log Framework name to more Windows ML relevant events

angelser:
This PR adds the frameworkName field to critical Windows ML telemetry events to ensure proper event attribution and prevent data loss.

## Reason
The frameworkName field is added to ensure that Windows ML events are not lost and do not require joins with events that might have been emitted outside the scope of the time span the processing scripts check for long-running apps/processes. This allows each event to be self-contained with framework identification.

## Events Modified
The following telemetry events now include the frameworkName field:

1. **SessionCreationStart** - Logs when session creation begins
2. **SessionCreation** - Logs session creation details including model metadata
3. **RuntimeError** - Logs runtime errors (both DEBUG and release builds)
4. **RuntimePerf** - Logs runtime performance metrics including total runs and duration
5. **AutoEpSelection** - Logs automatic execution provider selection policy and results
6. **ProviderOptions** - Logs execution provider configuration options

All events now include TraceLoggingString(ORT_CALLER_FRAMEWORK, "frameworkName") to maintain consistent framework identification across the telemetry pipeline.

### DIFF
--- a/onnxruntime/core/platform/windows/telemetry.cc
+++ b/onnxruntime/core/platform/windows/telemetry.cc
@@ -415,7 +415,8 @@ void WindowsTelemetry::LogSessionCreation(uint32_t session_id, int64_t ir_versio
                       TraceLoggingString(model_metadata_string.c_str(), "modelMetaData"),
                       TraceLoggingString(loaded_from.c_str(), "loadedFrom"),
                       TraceLoggingString(execution_provider_string.c_str(), "executionProviderIds"),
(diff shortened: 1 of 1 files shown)


bmehta001 (approved):

onnxruntime/core/platform/windows/telemetry.cc
ashrit-ms: You will need to add the frameworkName in the else condition to keep them in sync.

PR #27273, opened 2026-02-06
[CI] Build NCHWc code on Windows/Linux ARM64 CIs

hariharans29:
### Description

Initially the NCHWc code was built only on Mac CIs to keep the build path regression-free.

There were some Linux-specific paths introduced in https://github.com/microsoft/onnxruntime/pull/26838 and there is more community interest in contributing to these code paths. See https://github.com/microsoft/onnxruntime/pull/27099. Hence, it makes sense to keep these code paths built and tested on Linux and Windows too.

### Motivation and Context
Improve CI quality with regards to ARM64 NCHWc builds

CC: @RohanJames1997

### DIFF
--- a/tools/ci_build/github/azure-pipelines/win-qnn-arm64-ci-pipeline.yml
+++ b/tools/ci_build/github/azure-pipelines/win-qnn-arm64-ci-pipeline.yml
@@ -53,6 +53,10 @@ jobs:
     matrix:
       SHARED_LIB:
         QnnLibKind: 'shared_lib'
+        # We currently do not have a standalone Windows ARM64 CI pipeline for the CPU EP,
+        # so we enable --enable_arm_neon_nchwc for additional coverage of the NCHWc code paths in the CPU EP.
+        # Once we have a standalone Windows ARM64 CI pipeline for the CPU EP, we can consider adding --enable_arm_neon_nchwc
+        # to that pipeline and removing it from this pipeline since this pipeline is primarily for testing the QNN EP.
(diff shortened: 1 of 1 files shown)


tools/ci_build/github/azure-pipelines/win-qnn-arm64-ci-pipeline.yml
```diff
@@ -53,7 +53,7 @@ jobs:
     matrix:
       SHARED_LIB:
         QnnLibKind: 'shared_lib'
-        ExtraQnnBuildArgs: '--client_package_build'
+        ExtraQnnBuildArgs: '--client_package_build --enable_arm_neon_nchwc'
```
hariharans29, line 60: We only have Win QNN ARM64 and not a native Win ARM64 (CPU only) CI. Hence, piggy-backing on this CI to test the CPU EP path (i.e.) NCHWc code in MLAS.

edgchen1: should we consider having a standalone win arm64 build? QNN EP may move out of this repo eventually. that could be done later.

this information may be good to include in a code comment though. easier to see than tracking down this PR.

hariharans29: Good idea. Not sure if having a Windows CPU ARM CI is worth it just to keep this feature covered. But I suspect once we eventually start shipping Windows CPU ARM packages ( in addition to the QNN packages currently being shipped), we will start having a CI. I ll add a comment here and once we have a dedicated WIndows ARM CI, I will add it there.

PR #27279, opened 2026-02-07
Speed up iOS/macOS CocoaPods packaging pipeline

edgchen1:
### Description

- Split out sysroot/arch framework builds into separate jobs. Previously, they were all being built sequentially.
- Update build agent image to macOS-15 and update Xcode version accordingly. This also seems to speed up the framework builds.

### Motivation and Context

Speed up the packaging pipeline build. The average duration of recent pipeline runs in main is more than 4 hours. Here's a test run that took 1.5 hours: https://aiinfra.visualstudio.com/Lotus/_build/results?buildId=1078804&view=results

### DIFF
--- a/tools/ci_build/github/apple/build_settings_utils.py
+++ b/tools/ci_build/github/apple/build_settings_utils.py
@@ -19,17 +19,22 @@ def parse_build_settings_file(build_settings_file: pathlib.Path) -> dict[str, ty
     :return: The build settings dict.
     :rtype: dict[str, Any]
     """
-    with open(build_settings_file) as f:
-        build_settings_data = json.load(f)
+
+    def check(condition: bool, message: str):
+        if not condition:
+            raise ValueError(message)
 
     # validate that `input` is a dict[str, list[str]]
     def validate_str_to_str_list_dict(input: dict[str, list[str]]):
-        assert isinstance(input, dict), f"input is not a dict: {input}"
+        check(isinstance(input, dict), f"input is not a dict: {input}")
         for key, value in input.items():
-            assert isinstance(key, str), f"key is not a string: {key}"
-            assert isinstance(value, list), f"value is not a list: {value}"
+            check(isinstance(key, str), f"key is not a string: {key}")
+            check(isinstance(value, list), f"value is not a list: {value}")
             for value_element in value:
-                assert isinstance(value_element, str), f"list element is not a string: {value_element}"
+                check(isinstance(value_element, str), f"list element is not a string: {value_element}")
+
+    with open(build_settings_file) as f:
+        build_settings_data = json.load(f)
 
     build_settings = {}
 


tools/ci_build/github/apple/build_apple_framework.py
```diff
@@ -2,144 +2,202 @@
...
-            build_config + "-" + sysroot,
-            (
-                "onnxruntime.framework"
-                if build_dynamic_framework
-                else os.path.join("static_framework", "onnxruntime.framework")
-            ),
-        )
-        ort_libs.append(os.path.join(framework_dir, "onnxruntime"))
+        # build framework for specified sysroot/arch
+        subprocess.run(build_command, shell=False, check=True)
```
vraspar, line 76: I noticed that we removed `cwd=REPO_DIR` option. Was that intentional ?

tools/ci_build/github/apple/build_settings_utils.py
```diff
@@ -0,0 +1,75 @@
...
+            for value_element in value:
+                assert isinstance(value_element, str), f"list element is not a string: {value_element}"
+
+    build_settings = {}
+
+    build_osx_archs = build_settings_data.get("build_osx_archs", _DEFAULT_BUILD_SYSROOT_ARCHS)
+    validate_str_to_str_list_dict(build_osx_archs)
+    build_settings["build_osx_archs"] = build_osx_archs
+
+    build_params = build_settings_data.get("build_params", {})
```
vraspar, line 45: should we throw error if build params are empty. 

Previously, 

```
 if "build_params" in build_settings_data:
        build_settings["build_params"] = build_settings_data["build_params"]
    else:
        raise ValueError("build_params is required in the build config file")
```

tools/ci_build/github/apple/build_settings_utils.py
```diff
@@ -0,0 +1,75 @@
...
+import json
+import pathlib
+
+_DEFAULT_BUILD_SYSROOT_ARCHS = {
+    "iphoneos": ["arm64"],
+    "iphonesimulator": ["arm64", "x86_64"],
+}
+
+
+def parse_build_settings_file(build_settings_file: pathlib.Path) -> dict:
```
vraspar: nit/optional:  we should set return type to `dict[str, Any] ` or  `dict[str, dict[str, list[str]]]` instead of `dict` for consistency

vraspar (approved):

tools/ci_build/github/apple/build_apple_framework.py
edgchen1: yeah, there shouldn't be any dependency on the working directory being `REPO_DIR`

tools/ci_build/github/apple/build_settings_utils.py
edgchen1: updated to `dict[str, Any]`

tools/ci_build/github/apple/build_settings_utils.py
edgchen1: it seemed ok to allow an empty value. I'm open to more specific validation too, but just checking for something being provided didn't seem to do much.

tools/ci_build/github/apple/build_settings_utils.py
```diff
@@ -0,0 +1,76 @@
...
+                assert isinstance(value_element, str), f"list element is not a string: {value_element}"
+
+    build_settings = {}
+
+    build_osx_archs = build_settings_data.get("build_osx_archs", _DEFAULT_BUILD_SYSROOT_ARCHS)
+    validate_str_to_str_list_dict(build_osx_archs)
+    build_settings["build_osx_archs"] = build_osx_archs
+
+    build_params = build_settings_data.get("build_params", {})
+    validate_str_to_str_list_dict(build_params)
```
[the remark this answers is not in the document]
edgchen1: it's ok to return an empty list. "base" is just meant to be a convenience category to avoid repeating options

tools/ci_build/github/apple/build_apple_framework.py
```diff
@@ -275,7 +331,92 @@ def parse_args():
...
+
+        built_sysroot_arch_framework_infos: list[SysrootArchFrameworkInfo] = []
+
+        for sysroot, arch in sysroot_arch_pairs_to_build:
+            infos_for_sysroot = sysroot_to_sysroot_arch_framework_infos.setdefault(sysroot, [])
+            base_build_command = [
+                sys.executable,
+                BUILD_PY,
+                *get_build_params(build_settings, "base"),
+                *get_build_params(build_settings, sysroot),
```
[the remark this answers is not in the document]
edgchen1: I don't think an empty list is necessarily invalid. we can refine the validation later as needed.

PR #27280, opened 2026-02-07
Update ORT_API_VERSION

edgchen1:
### Description

Update ORT_API_VERSION to 25. The version has already been updated to 1.25.0.

Update instructions in Versioning.md.

Add a unit test to check that ORT_API_VERSION is equal to the second component of the version.

### Motivation and Context

Update ORT_API_VERSION following version update.

### DIFF
--- a/onnxruntime/test/shared_lib/test_version.cc
+++ b/onnxruntime/test/shared_lib/test_version.cc
@@ -19,8 +19,8 @@ TEST(CApiTest, VersionConsistencyWithApiVersion) {
 
   auto to_uint32_t = [](const std::string& s) -> std::optional<uint32_t> {
     uint32_t result{};
-    if(std::from_chars(s.data(), s.data() + s.size(), result).ec == std::errc{}) {
-        return result;
+    if (std::from_chars(s.data(), s.data() + s.size(), result).ec == std::errc{}) {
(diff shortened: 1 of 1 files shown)


onnxruntime/test/shared_lib/test_version.cc
```diff
@@ -0,0 +1,31 @@
...
+  auto to_uint32_t = [](const std::string& s) -> std::optional<uint32_t> {
+    uint32_t result{};
+    if(std::from_chars(s.data(), s.data() + s.size(), result).ec == std::errc{}) {
+        return result;
+    }
+    return std::nullopt;
+  };
+
+  ASSERT_NE(to_uint32_t(version_string_components[0]), std::nullopt);
+  ASSERT_EQ(to_uint32_t(version_string_components[1]), uint32_t{ORT_API_VERSION});
```
edgchen1, line 29: is `ORT_API_VERSION == <second component of version>` a valid assumption to make?

tianleiwu: It might not be true in the future. For example, we might have 2.0.0 release.

We can change this test later if needed.

tianleiwu (approved):

PR #27285, opened 2026-02-09
Improve WebGPU MatMulNBits to support zero pointer for 2bits

HectorSVC:
### Description
The existing WebGPU MatMulNBits op does not support zero pointer for 2bits. So it blocks some models. This PR enables the zero pointer support for 2bits support. UT tests are included for coverage.

### DIFF
--- a/js/web/lib/wasm/jsep/webgpu/ops/matmulnbits.ts
+++ b/js/web/lib/wasm/jsep/webgpu/ops/matmulnbits.ts
@@ -146,8 +146,9 @@ export const createMatMulNBitsProgramInfo = (
         for (let c = 0; c < components * outputNumber; c++) {
           calcStr += `
             b_value = ${bComponents === 1 ? `b${c}_data` : `b${c}_data[i]`};
-            ${attributes.bits === 2
-              ? `{
+            ${
+              attributes.bits === 2
+                ? `{
               let half_word = b_value >> ${pass * 16}u;
               let byte_lo = half_word & 0xFFu;
               let byte_hi = (half_word >> 8u) & 0xFFu;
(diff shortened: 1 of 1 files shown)


onnxruntime/contrib_ops/webgpu/quantization/matmul_nbits.cc
```diff
@@ -219,7 +221,9 @@ Status ApplyMatMulNBits(const Tensor* a, const Tensor* b, const Tensor* scales,
 #endif
 
   // On FP32 only GPUs, integer math is faster than FP32 therefore always use DP4A independent of length of M.
+  // DP4A Q2 path uses a hardcoded LUT with zero_point=2, so skip DP4A for Q2 with custom zero points.
   if ((M >= kMinMForTileOptimization || y->DataType() == DataTypeImpl::GetType<float>() || context.AdapterInfo().vendor == std::string_view{"qualcomm"}) &&
+      !(has_zero_points && nbits == 2) &&
```
sushraja-msft, line 226: it would be good to add support for dp4a, which is key to get good performance on broad market intel consumer devices. I think all you need is to modify 

onnxruntime/contrib_ops/webgpu/quantization/dp4a_matmul_common.wgsl.template
DequantizedFrom2BitsTo8Bits to accept a zeros parameter. 

This method DequantizedFrom2BitsTo8Bits simply handles bit unpacking going from a packed u32 which contains 16 weights when nbits is 2 to vec4(u32) -> 16 weights. Expanding the LUT to 1024 entries when there are zero points from 256 may be the easiest way to do this.

HectorSVC: Thanks for the comments, will try that.

PR #27289, opened 2026-02-09
[CORE]: Improve filesystem error messages during Linux device discovery

theHamsta:
### Description

This is a follow-up to https://github.com/microsoft/onnxruntime/pull/26210
to address https://github.com/microsoft/onnxruntime/pull/26210#discussion_r2693897142 and review dog lints. 

ErrorCodeToStatus currently does not include the filesystem path that
caused the error. This could it make difficult to know the root cause
of a reported filesystem error.

Review dog lints: https://github.com/microsoft/onnxruntime/pull/26210/changes
Plus a typo: `dit` -> `did`

### Motivation and Context

Clean up discussed issues and lints of #26210

### DIFF
--- a/onnxruntime/core/platform/linux/device_discovery.cc
+++ b/onnxruntime/core/platform/linux/device_discovery.cc
@@ -23,14 +23,14 @@ namespace onnxruntime {
 
 namespace {
 
-Status ErrorCodeToStatus(const std::error_code& ec) {
+Status ErrorCodeToStatus(const std::error_code& ec, const std::filesystem::path& path, const std::string_view context) {
   if (!ec) {
     return Status::OK();
   }
 
   return Status{common::StatusCategory::ONNXRUNTIME, common::StatusCode::FAIL,
                 MakeString("Error: std::error_code with category name: ", ec.category().name(),
-                           ", value: ", ec.value(), ", message: ", ec.message())};
+                           ", value: ", ec.value(), ", message: ", ec.message(), ", filesystem path: ", path, ", context: ", context)};
 }
 
 struct GpuSysfsPathInfo {
@@ -42,7 +42,7 @@ Status DetectGpuSysfsPaths(std::vector<GpuSysfsPathInfo>& gpu_sysfs_paths_out) {
   std::error_code error_code{};
   const fs::path sysfs_class_drm_path = "/sys/class/drm";
   const bool sysfs_class_drm_path_exists = fs::exists(sysfs_class_drm_path, error_code);
-  ORT_RETURN_IF_ERROR(ErrorCodeToStatus(error_code));
+  ORT_RETURN_IF_ERROR(ErrorCodeToStatus(error_code, sysfs_class_drm_path, "Checking existence of DRM sysfs path"));
 
   if (!sysfs_class_drm_path_exists) {
     gpu_sysfs_paths_out = std::vector<GpuSysfsPathInfo>{};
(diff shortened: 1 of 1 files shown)


onnxruntime/core/platform/linux/device_discovery.cc
```diff
@@ -21,14 +22,17 @@ namespace onnxruntime {
 
 namespace {
 
-Status ErrorCodeToStatus(const std::error_code& ec) {
+Status ErrorCodeToStatus(const std::error_code& ec, const fs::path& path, std::string_view context) {
```
theHamsta: the string "context" could be overkill. I thought it would provide useful context, why we accessed the file. But I would at least include the filesystem path that caused the error code. Otherwise, debugging the error code could be difficult

theHamsta:
@edgchen1 on one of our test systems the device discovery fails for the iGPU (there are permission problems for this one file). This has the effect that even card1 which would be the Nvidia GPU we actually want to use does not get discovered.

Could we change failures for individual sysfs paths to warnings while still allowing the discovery of other devices? 
```
0;93m2026-04-07 21:25:38.450380746 [W:onnxruntime:Default, device_discovery.cc:283 GetGpuDevices] Failed to detect devices under "/sys/class/drm/card0": device_discovery.cc:93 ReadFileContents Failed to open file: "/sys/class/drm/card0/device/vendor"
````
Before this change you would just get a failure without even knowing that a file path for the iGPU is at fault

**Comment by gedoensmax** [comment]
@chilo-ms can you trigger the CI and see if we can get this merged ?

File: onnxruntime/core/platform/linux/device_discovery.cc [hunk 2]
```diff
@@ -131,9 +132,9 @@ Status GetPciBusId(const std::filesystem::path& sysfs_path, std::optional<std::s
     pci_bus_id = {};
     LOGS_DEFAULT(WARNING) << MakeString("Skipping pci_bus_id for PCI path at \"",
                                         pci_bus_id_path.string(),
-                                        "\" because filename \"", pci_bus_id_filename, "\" dit not match expected pattern of ",
+                                        "\" because filename \"", pci_bus_id_filename, "\" did not match expected pattern of ",
```
edgchen1: [std::filesystem::path::operator<<](https://en.cppreference.com/w/cpp/filesystem/path/operator_ltltgtgt.html) already adds quotes so we don't need additional ones.
```suggestion
                                        "\" because filename ", pci_bus_id_filename, " did not match expected pattern of ",
```

edgchen1:
> Could we change failures for individual sysfs paths to warnings while still allowing the discovery of other devices?

sure, I think that's fine.

onnxruntime/core/platform/linux/device_discovery.cc
```diff
@@ -22,14 +23,14 @@ namespace onnxruntime {
...
-Status ErrorCodeToStatus(const std::error_code& ec) {
+Status ErrorCodeToStatus(const std::error_code& ec, const std::filesystem::path& path, const std::string& context) {
   if (!ec) {
     return Status::OK();
   }
 
   return Status{common::StatusCategory::ONNXRUNTIME, common::StatusCode::FAIL,
                 MakeString("Error: std::error_code with category name: ", ec.category().name(),
-                           ", value: ", ec.value(), ", message: ", ec.message())};
+                           ", value: ", ec.value(), ", message: ", ec.message(), ", filesystem path: ", path, ", context: ", context)};
```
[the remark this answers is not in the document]
theHamsta: I had doubts with usage of string_view before because not much used in the code base but now changed it to `string_view`

chilo-ms:
Could you merge main? We have a Web CI fix in the main.

anskumar01:
Can this be merged now or is there anything blocking. We wanted to take this PR in ORT 1.25 as it is blocking TRT RTX EP on Linux

chilo-ms:
> Can this be merged now or is there anything blocking. We wanted to take this PR in ORT 1.25 as it is blocking TRT RTX EP on Linux

The Web CI is failing, and the fix is to merge main.

PR #27292, opened 2026-02-09
Improve encapsulation of weight_data_type_freq_

ashrit-ms:
### Description
Move weight_data_type_freq_ to private section and add public getter method GetWeightDataTypeFrequency() to provide controlled read-only access. Update inference_session.cc to use the new getter instead of direct member access.

### Motivation and Context
To ensure we follow best practices when accessing class variables.

### DIFF
--- a/include/onnxruntime/core/graph/graph.h
+++ b/include/onnxruntime/core/graph/graph.h
@@ -779,10 +779,12 @@ class Graph {  // NOLINT(clang-analyzer-optin.performance.Padding): preserve exi
   bool IsSparseInitializer(const std::string& name) const;
 #endif
 
+#if !defined(ORT_MINIMAL_BUILD)
   /** Gets the frequency count of weight data types in this graph. */
   gsl::span<const int32_t> GetWeightDataTypeFrequency() const noexcept {
     return weight_data_type_freq_;
   }
+#endif
 
   /** Gets an initializer tensor with the provided name.
   @param[out] value Set to the TensorProto* if the initializer is found, or nullptr if not.


include/onnxruntime/core/graph/graph.h
```diff
@@ -779,6 +779,11 @@ class Graph {  // NOLINT(clang-analyzer-optin.performance.Padding): preserve exi
   bool IsSparseInitializer(const std::string& name) const;
 #endif
 
+  /** Gets the frequency count of weight data types in this graph. */
+  const int32_t* GetWeightDataTypeFrequency() const noexcept {
```
yuslepukhin: Return `gsl::span<const int32_t>`

onnxruntime/core/session/inference_session.cc
```diff
@@ -1818,15 +1818,16 @@ static bool ModelHasFP16Inputs(const Graph& graph) {
 #if !defined(ORT_MINIMAL_BUILD)
 [[maybe_unused]] static std::string ModelWeightDataType(const Graph& graph) {
   std::string data_type_list;
+  const int32_t* weight_freq = graph.GetWeightDataTypeFrequency();
```
yuslepukhin: Ditto

include/onnxruntime/core/graph/graph.h
ashrit-ms: Fixed.

onnxruntime/core/session/inference_session.cc
ashrit-ms: Fixed.

PR #27295, opened 2026-02-10
Remove s_kernel_registry_vitisaiep.reset() in deinitialize_vitisai_ep()

zz002:
Remove unnecessary s_kernel_registry_vitisaiep.reset() call in deinitialize_vitisai_ep() function. The kernel registry will be repopulated on next initialization, making this reset redundant.

### DIFF
--- a/onnxruntime/core/providers/vitisai/imp/global_api.cc
+++ b/onnxruntime/core/providers/vitisai/imp/global_api.cc
@@ -386,7 +386,6 @@ void deinitialize_vitisai_ep() {
   s_domains_vitisaiep.clear();
 
   s_library_vitisaiep.Clear();
-  s_kernel_registry_vitisaiep.reset();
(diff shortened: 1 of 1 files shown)


tianleiwu:
@zz002, could you review Copilot's feedback and see whether this change is expected?

amd-genmingz:
@tianleiwu  deinitialize_vitisai_ep is a Shutdown function, it means that the entire provider_vitisai.dll will be released via UnloadDynamicLibrary. At that time, the corresponding shared_ptr will also be deleted, which has the same effect as calling reset(). Therefore, we believe this modification is consistent with the expected behavior.
During our debugging process, we found that calling this function during the process detach phase would cause an error. The reason is that its control block had already been destroyed at that point for some unknown reason. Therefore, we removed this line to prevent the crash.

PR #27297, opened 2026-02-10
Check NodeArg has name in ORT formal model load

skottmckay:
### Description

Check for invalid node arg when loading ORT format model. A node arg would always have a valid string in an ORT created model so this requires manual editing of the model to make it invalid.


yuslepukhin (approved):
:shipit:

PR #25881, opened 2025-08-28
[Fix] illegal memory access in GetInputIndices with optional inputs

mingyueliuh:
### Description
Fix illegal memory access in GetInputIndices with optional inputs

### Motivation and Context
When an input is optional, its ValueInfo may be nullptr. 
The current implementation directly calls InputValueInfo->GetName(), leading to illegal memory access. 

Update logic to skip optional inputs when valueInfo is nullptr to avoid the crash.

PR #27301, opened 2026-02-10
[Fix] Illegal memory access in GetOutputIndex with optional outputs

qti-mattsinc:
### Description
* Fix an access violation in `GetOutputIndex`.
* Add related EpGraphTest UT.

### Motivation and Context
* When a non-trailing output is optional, its ValueInfo in `GetOutputIndex` may be null. The current implementation attempts to dereference the nullptr due to a missing check.
* This bug is exposed if an EP calls `ValueInfo_GetValueProducer` on a value info whose producer has non-trailing optional outputs.
* https://github.com/microsoft/onnxruntime/pull/25881 previously fixed the analogous issue for `GetInputIndices` but not for `GetOutputIndex`

### DIFF
--- a/onnxruntime/test/testdata/skip_simplified_layer_normalization.py
+++ b/onnxruntime/test/testdata/skip_simplified_layer_normalization.py
@@ -28,12 +28,7 @@ input_skip_bias_sum_vi = helper.make_tensor_value_info(
     shape=[batch_size, seq_len, hidden_size],
 )
 
-gamma_init = helper.make_tensor(
-    name="gamma",
-    data_type=TensorProto.FLOAT,
-    dims=[hidden_size],
-    vals=[1] * hidden_size
-)
+gamma_init = helper.make_tensor(name="gamma", data_type=TensorProto.FLOAT, dims=[hidden_size], vals=[1] * hidden_size)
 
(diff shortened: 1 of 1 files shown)


onnxruntime/core/graph/ep_api_types.cc
```diff
@@ -351,6 +351,10 @@ static Status GetOutputIndex(const EpNode& producer_node,
   gsl::span<const EpValueInfo* const> outputs = producer_node.GetOutputsSpan();
 
   for (size_t i = 0; i < outputs.size(); i++) {
+    if (outputs[i] == nullptr) {  // outputs == nullptr means the output is optional
```
edgchen1:
```suggestion
    if (outputs[i] == nullptr) {  // outputs[i] == nullptr means the output is a missing optional output
```

edgchen1 (approved):

qti-mattsinc:
Force pushing to fix author/committer metadata

qti-mattsinc:
Looks like this was fixed with https://github.com/microsoft/onnxruntime/pull/27644.

edgchen1:
sorry, I lost track of this PR. thanks anyway for contributing the fix.

PR #27306, opened 2026-02-11
[web] enable native EH for JSEP/Asyncify build

fs-eire:
### Description

Enables Wasm native exception handling for JSEP and asyncify build.


fs-eire:
I expect that using `-fwasm-exceptions` with `ASYNCIFY` should be working since emsdk v4.0.21 (see https://github.com/WebAssembly/binaryen/pull/5475). But somehow wasm-opt is not happy with it :(

fs-eire:
It looks like native EH is still not supported for asyncify build in latest emsdk.

PR #27312, opened 2026-02-11
Fix misaligned addresses while reading tensor attributes from raw data buffers

### LINKED ISSUE #27311: [Mobile] Android armeabi-v7a 32 bit bus error due to misaligned address
### Describe the issue

With some (rare) models ONNXRuntime built for Android armeabi-v7a 32 bit architecture crashes with bus error. After re-compiling  ORT with enabled sanitizers
`-fsanitize=alignment -fno-omit-frame-pointer`
I was able to find exact place:
```
/src/onnxruntime/core/optimizer/common_subexpression_elimination.cc:254:20: runtime error: reference binding to misaligned address 0xe022a091 for type 'const int64_t' (aka 'const long long'), which requires 8 byte alignment
0xe022a091: note: pointer points here
 00 00 00  10 01 00 00 00 00 00 00  00 00 00 00 00 00 00 00  01 81 00 00 00 00 4e b5  00 00 00 00 00
              ^ 
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /src/onnxruntime/core/optimizer/common_subexpression_elimination.cc:254:20 
```
Raw data buffers for tensor attributes of type `char*` has no proper alignment guarantees while reading values from them.

### Platform

Android

### OS Version

14

### ONNX Runtime Installation

Built from Source

### Compiler Version (if 'Built from Source')

NDK r28c

### Package Name (if 'Released Package')

None

4295524cd868938ea9ffc0408199b441bf8f3f22

### ONNX Runtime API

C++/C

### Architecture

ARM32

### Execution Provider

Default CPU

_No response_

n-v-k:
### Description
Explicitly copy tensor attribute values from raw data buffers instead of directly using pointers without checking proper memory alignment.

### Motivation and Context
Fixes #27311

With some (rare) models ONNXRuntime built for Android armeabi-v7a 32 bit architecture crashes with bus error. Raw data buffers for tensor attributes of type `char*` has no proper alignment guarantees while reading values from them.
Actually, it is in general (not only Android armeabi-v7a) UB to access objects at a misaligned address in C++, though many modern platforms allow it.

### DIFF
--- a/onnxruntime/core/optimizer/common_subexpression_elimination.cc
+++ b/onnxruntime/core/optimizer/common_subexpression_elimination.cc
@@ -172,12 +172,9 @@ bool AreRangesEqual(const Range& lhs, const Range& rhs) {
 }
 
 // Check if two tensor attributes are equal scalar tensors, mainly to support ConstantOfShape Op.
-// Currently support float, float16 and int64 data types.
 bool AreScalarTensorAttributeEqual(const ONNX_NAMESPACE::TensorProto& lhs_t, const ONNX_NAMESPACE::TensorProto& rhs_t) {
   if (!(utils::HasDataType(lhs_t) && utils::HasDataType(rhs_t) && lhs_t.data_type() == rhs_t.data_type() &&
-        (lhs_t.data_type() == onnx::TensorProto_DataType_FLOAT ||
-         lhs_t.data_type() == onnx::TensorProto_DataType_FLOAT16 ||
-         lhs_t.data_type() == onnx::TensorProto_DataType_INT64) &&
+        lhs_t.data_type() != onnx::TensorProto_DataType_STRING &&
         lhs_t.dims_size() == 1 && rhs_t.dims_size() == 1 && lhs_t.dims()[0] == 1 && rhs_t.dims()[0] == 1)) {
     return false;
   }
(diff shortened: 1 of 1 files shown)


yuslepukhin:
Thank you for reporting it and taking initiative to fix.

onnxruntime/core/optimizer/common_subexpression_elimination.cc
```diff
@@ -185,12 +185,24 @@ bool AreScalarTensorAttributeEqual(const ONNX_NAMESPACE::TensorProto& lhs_t, con
   const void* lhs_value = lhs_t.raw_data().data();
   const void* rhs_value = rhs_t.raw_data().data();
   switch (lhs_t.data_type()) {
-    case onnx::TensorProto_DataType_FLOAT:
-      return *reinterpret_cast<const float*>(lhs_value) == *reinterpret_cast<const float*>(rhs_value);
-    case onnx::TensorProto_DataType_FLOAT16:
-      return *reinterpret_cast<const MLFloat16*>(lhs_value) == *reinterpret_cast<const MLFloat16*>(rhs_value);
-    case onnx::TensorProto_DataType_INT64:
-      return *reinterpret_cast<const int64_t*>(lhs_value) == *reinterpret_cast<const int64_t*>(rhs_value);
+    case onnx::TensorProto_DataType_FLOAT: {
```
yuslepukhin: There is a number of problems with this code and the below. Let's try to make this fix complete.

To make sure the alignment and the endianess are respected on all platforms, as well as we cover more potential types, we want to use `utils::UnpackTensor<T>()` for individual types.

Another potentially better alternative for both locations here and for the hash computation would be `utils::UnpackInitializerData()` that returns `std::vector<uint8_t>`.

This way we also avoid the requirement of having raw data and cover more ground.

The functions are declared in core/framework/tensorprotoutils.h.

onnxruntime/core/optimizer/common_subexpression_elimination.cc
```diff
@@ -235,24 +227,22 @@ bool AreEqual(const ONNX_NAMESPACE::AttributeProto& lhs, const ONNX_NAMESPACE::A
...
 }
 
-// Support scalar float/int64/fp16 tensor attribute only for now, and requires data is raw data in TensorProto.
+// Support scalar float/int64/fp16 tensor attribute only for now.
 std::size_t GetTensorAttributeHash(const ONNX_NAMESPACE::TensorProto& attr_t) {
   std::size_t hash = 0;
-  if (utils::HasDataType(attr_t) && attr_t.dims_size() == 1 && attr_t.dims()[0] == 1 && utils::HasRawData(attr_t)) {
+  if (utils::HasDataType(attr_t) && attr_t.dims_size() == 1 && attr_t.dims()[0] == 1) {
     int data_type = attr_t.data_type();
     switch (data_type) {
```
yuslepukhin: This is much better. Here is an idea, let's make this function type independent.
We will check if the attribute tensor is not of string type using `ORT_ENFORCE(data_type != TensorDataType_STRING, "Unexpected tensor string type");` at the start of the function.
We can then remove the switch, Unpack the data, compute and return the hash regardless of the type.

n-v-k:
@yuslepukhin, I made it data type agnostic.
Copilot comments are from outdated review about raw_data usage with `memcpy`.

yuslepukhin (approved):
Please, review and address all of the Copilot comments.

PR #27318, opened 2026-02-11
[js/web] Use embedded WASM module in Blob URL workers when wasmBinary is provided

### LINKED ISSUE #27317: onnxruntime-web: Bundle variant fails in Blob URL Web Workers
`ort.wasm.bundle.min.mjs` fails in Blob URL Web Workers due to two issues:

1. **Eager URL resolution** — `new URL("ort-wasm-simd-threaded.wasm", import.meta.url)` throws because `blob:` URLs can't be a base for relative resolution. This runs even when `wasmBinary` is already provided.

2. **Same-origin check** — `importWasmModule` checks if the script URL is same-origin to decide between the embedded Emscripten module and a dynamic `import('./ort-wasm-simd-threaded.mjs')`. Blob URLs can fail this, triggering the import which doesn't exist.

Blob URL workers are the only option in environments where `proxy = true` fails (browser extensions, Electron apps with no static file serving).

## Workaround

```js
let code = readFileSync("ort.wasm.bundle.min.mjs", "utf-8");
code = code.replace(/import\.meta\.url/g, "self.location.href");
code = code.replace(/new URL\("ort-wasm-simd-threaded\.wasm",self\.location\.href\)\.href/g, '""');
code = code.replace(/export\{(.+?)\}/, (_, e) =>
  `self.ort={${e.replace(/(\w+) as (\w+)/g, "$2:$1")}}`,
);
```

## Suggestion

When `wasmBinary` is provided + `numThreads = 1`: skip eager URL resolution and always use the embedded Emscripten module. A UMD/IIFE build or `ort.env.wasm.skipUrlResolution` flag would also help.

**Environment:** onnxruntime-web 1.24.1, Electron/Chromium, `numThreads=1`, `wasmBinary` from IndexedDB

0-don:
Fixes #27317

When running inside a Blob URL Web Worker with `wasmBinary` provided and `numThreads=1`, `isSameOrigin(scriptSrc)` can fail because blob: URLs have opaque origins. This causes a fallback to dynamic `import('./ort-wasm-simd-threaded.mjs')` which doesn't exist in that context.

Since `wasmBinary` is already provided and no worker spawning is needed (single-threaded), the embedded Emscripten module can be used directly — no URL resolution or same-origin check is needed.

**Change:** One line in `wasm-utils-import.ts` line 275:
```typescript
// Before:
useEmbeddedModule = isSameOrigin(scriptSrc);

// After:
useEmbeddedModule = isSameOrigin(scriptSrc) || (isWasmOverridden && !isMultiThreaded);
```

This extends the existing pattern from the `!scriptSrc` case (line 268) to also apply when `scriptSrc` is available but fails same-origin checks. The condition (`wasmBinary` provided + single-threaded) guarantees no file resolution or worker spawning is needed.

### DIFF
--- a/onnxruntime/core/mlas/lib/kai_ukernel_interface.cpp
+++ b/onnxruntime/core/mlas/lib/kai_ukernel_interface.cpp
@@ -19,8 +19,12 @@
 #include "kai/ukernels/matmul/matmul_clamp_f32_f32_f32p/kai_matmul_clamp_f32_f32_f32p8x1biasf32_6x8x4_neon_mla.h"
 #include "kai/ukernels/matmul/matmul_clamp_f32_f32p_f32p/kai_matmul_clamp_f32_f32p2vlx1_f32p2vlx1biasf32_sme2_mopa.h"
 #include "kai/ukernels/matmul/matmul_clamp_f32_f32p_f32p/kai_matmul_clamp_f32_f32p2vlx1_f32p2vlx1b_2vlx2vl_sme_mopa.h"
+
+#include "kai/ukernels/matmul/matmul_clamp_f32_qai8dxp_qsi8cxp/kai_matmul_clamp_f32_qai8dxp1vlx4_qsi8cxp4vlx4_1vlx4vl_sme_mopa.h"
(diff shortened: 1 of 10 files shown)


fs-eire:
thank you for the fix!

fs-eire:
Please sync to latest main branch to pick up the CUDA build break fix.

0-don:
> Please sync to latest main branch to pick up the CUDA build break fix.

done

0-don:
seems quite difficult to make it work :laughing:

PR #27324, opened 2026-02-12
Fix CoreML reshape to support dynamic shapes

### LINKED ISSUE #26328: CoreML reshape doesn't work with dynamic shape
### Describe the issue

Trying to reshape a tensor to `[-1, 2048]` resulted in:
```
[V:onnxruntime:, shape_utils.cc:32 GetShapeImpl] NodeArg [/value/conv/relu] has shape with dynamic dimension
[V:onnxruntime:, base_op_builder.cc:46 IsOpSupported] Operator [Reshape] is not supported by the impl
[V:onnxruntime:, helper.cc:103 GetSupportedNodes] Operator type: [Reshape] index: [109] name: [/value/reshape] supported: [0]
```
Maybe [line 99](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/providers/coreml/builders/impl/reshape_op_builder.cc#L99C1-L100C18) in `reshape_op_builder.cc` should be
```c++
  if (!GetShape(*input_defs[0], input_shape, logger))
    return false;
```
instead of `GetStaticShape(...)`.

Something similar was also mentioned in https://github.com/microsoft/onnxruntime/issues/21271#issuecomment-2264859467 but without details.

### To reproduce

As far as I can tell, any reshape of a tensor with a dynamic shape falls back to the cpu execution provider.

### Urgency

_No response_

### Platform

Mac

### OS Version

15.3.2

### ONNX Runtime Installation

Released Package

1.23.1

### ONNX Runtime API

C++

### Architecture

ARM64

### Execution Provider

CoreML

_No response_

ingyukoh:
CoreML MIL supports -1 in shape dimensions and can infer at runtime. This fix allows dynamic 
  input shapes for reshape operations.                                                         
                                                                                               
  Fixes #26328                                                                                 
                                                                                               
  ### Description                                                                              
  - Change `GetStaticShape` to `GetShape` to allow dynamic dimensions                          
  - Only apply `ReshapeHelper` when input shape is fully static                                
  - CoreML MIL handles -1 dimensions at runtime                                                
                                                                                               
  ### Motivation and Context                                                                   
  Users reshaping tensors to `[-1, N]` were getting fallback to CPU with warnings about        
  unsupported reshape operations on dynamic shapes.


ingyukoh:
@skottmckay Could you take a look at this? The fix changes `GetStaticShape` to `GetShape` in 
  the reshape op builder and guards `ReshapeHelper` behind `IsStaticShape()`, so CoreML MIL can
   handle `-1` dimensions at runtime for dynamic input shapes. Added a test case and test model
   in the second commit.

yuslepukhin:
1. Missing validation for 0 in new_shape with dynamic input (ONNX/CoreML spec mismatch)
ONNX spec (allowzero=0, the default): A 0 in new_shape means "copy the corresponding dimension from the input tensor."
CoreML MIL: Does NOT support 0 meaning "copy from input." It only supports -1 for inference.

After this PR, when input is dynamic:

ReshapeHelper is skipped (it would have resolved 0 → actual dimension)
A raw 0 in new_shape is passed directly to CoreML, which will misinterpret it
Fix needed in IsOpSupportedImpl: When input shape is dynamic (!IsStaticShape(input_shape)), reject new_shape containing 0:

```.cpp
if (!IsStaticShape(input_shape)) {
  if (std::find(new_shape.begin(), new_shape.end(), int64_t{0}) != new_shape.end()) {
    LOGS(logger, VERBOSE) << "Reshape with 0 in shape requires static input to resolve. "
                          << "Input shape: " << Shape2String(input_shape)
                          << ", new shape: " << Shape2String(new_shape);
    return false;
  }
}
```

2. No validation that at most one -1 exists when ReshapeHelper is skipped
When IsStaticShape is false, ReshapeHelper is skipped entirely. ReshapeHelper enforces the ONNX rule: "at most one dimension of the new shape can be -1." Without it, an invalid model with multiple -1 values could pass through to CoreML without early detection.

Fix needed in IsOpSupportedImpl or AddToModelBuilderImpl:
```.cpp
if (!IsStaticShape(input_shape)) {
  if (std::count(new_shape.begin(), new_shape.end(), int64_t{-1}) > 1) {
    LOGS(logger, VERBOSE) << "Reshape cannot have more than one -1 in shape";
    return false;
  }
}
```

3. ReshapeHelper result is used but only conditionally applied
In AddToModelBuilderImpl, ReshapeHelper helper(TensorShape(input_shape), new_shape) modifies new_shape in-place (resolves -1 and 0). When static, the resolved shape is passed to CoreML. When dynamic, the raw shape (with -1) is passed. This is actually correct behavior for CoreML MIL which natively handles -1, but the code should add a comment clarifying this intentional difference.

4. NeuralNetwork path (reshapestatic) with -1
The else branch uses layer->mutable_reshapestatic(). Apple's NeuralNetwork spec ReshapeStaticLayerParams does support -1 in targetShape, so this is fine. However, it would be prudent to confirm NeuralNetwork behavior is tested or restrict dynamic shapes to ML Program only.

Test coverage Gaps:

- No test for new_shape containing 0 with dynamic input (should be rejected)
- No test for multiple -1 values in new_shape with dynamic input
- No negative test (e.g., incompatible reshape that should fail)
- No test with allowzero=1 attribute and dynamic input
- Only tests rank-2 reshape; no test for higher-rank (up to 5) dynamic reshapes

PR #16752, opened 2023-07-18
Align AvgPool ceil_mode on last value to torch

titaiwangms:
Fix #16203

Previous to this PR, if `ceil_mode` is on, the calculation of a value would divide the kernel size, even if remaining pixels is less than the kernel size, which causes the difference in this operator between ORT and torch.

However, this fix only applies to the change in #15597, which only supports AvgPool since 19. The older opset version is remain the same, as it's using mlas files.

Also, the PR fixes the shape mismatch caused by sliding window starting from padding. More detail: https://github.com/onnx/onnx/pull/6650 (And this PR is also validated with the tests added in https://github.com/onnx/onnx/pull/6650)

PR #27331, opened 2026-02-12
Fix for AveragePool performance regression by guarding ceil_mode clamp

### LINKED ISSUE #27190: [Performance] Performance regression in AveragePool operator
## Description

We observed a significant performance regression in the **AveragePool** operator between ONNXRuntime v1.20.0 and v1.21.0. The regression affects multiple AveragePool configurations with **+17% to +28% kernel slowdown**.

### AveragePool
- **Opset Version**: 11, 18
- **Regression**: **+17% to +30% kernel slowdown**

#### regression commit : 
https://github.com/microsoft/onnxruntime/pull/16752/files#diff-4cd5cf24d279e5bd52c1ee3888739b856b62be11d86e3f8271504d631b62af6c

### To reproduce

   ```bash
   python script_profiling.py averagepool_pool_11_18_average_pool_3d_valid_input 1.20.0 1.21.0
   ```

[Archive.zip](https://github.com/user-attachments/files/24907924/Archive.zip)

### Urgency

_No response_

### Platform

Linux

### OS Version

Ubuntu 24.04.3 LTS

### ONNX Runtime Installation

Released Package

1.20

### ONNX Runtime API

Python

### Architecture

X64

### Execution Provider

Default CPU

_No response_

### Model File

_No response_

Yes

tengli-alaska:
### Description

Fixes #27190

Guards the `std::min` bounds clamping in `AveragePool1DTask`, `AveragePool2DTask`, and `AveragePool3DTask` behind a `ceil_mode` check to restore performance for the default `ceil_mode=0` path.

### Root Cause

PR #16752 added `std::min` clamping on `hend`/`wend`/`dend` inside the hot pooling loops to fix correctness when `ceil_mode=1`. However, these clamps run unconditionally, including when `ceil_mode=0` (the default for the vast majority of models). This introduces a data-dependent loop bound that can prevent compiler auto-vectorization of the inner accumulation loop, causing a +17% to +28% kernel slowdown as reported in #27190.

### Fix

When `ceil_mode=0`, the output dimensions are computed with floor division, which guarantees that `hstart + kernel_size <= height + pad` — the clamp is a no-op. This PR guards each clamp behind `if (ceil_mode)`, restoring the original fast path while preserving the correctness fix for `ceil_mode=1`.

### Changes

- `pool_functors.h`: Added `int64_t ceil_mode` field to `AveragePool1DTask`, `AveragePool2DTask`, `AveragePool3DTask` and wrapped each `std::min` clamp in `if (ceil_mode)`
- `pool.cc`: Pass `pool_attrs_.ceil_mode` to the three AveragePool task initializers

### Benchmark Results

Regression confirmed on AMD EPYC 7543 (x86_64, Linux) using ORT's built-in profiling on `averagepool_pool_11_18_average_pool_3d_valid_input`:

| Version | Kernel Time (ms) | Change |
|---------|------------------|--------|
| 1.20.0  | 0.1655           | baseline |
| 1.21.0  | 0.1710           | +3.3%  |

The original issue observed +17% to +28% regression. Unable to build ORT from source locally to benchmark the patched version, would appreciate if  @junghyunpark2001 could validate the fix on their setup.

### Notes

- Zero-cost for `ceil_mode=0` (default path)
- `ceil_mode=1` correctness fix from #16752 is fully preserved
- Existing AveragePool tests cover both `ceil_mode=0` and `ceil_mode=1`


tengli-alaska:

.vscode/settings.json
```diff
@@ -1,36 +0,0 @@
-{
```
titaiwangms, line 1: You should not delete this file.

titaiwangms (commented):
I also only discovered non-significant changes.

```
{
  "model": "/home/titaiwang/onnxruntime/profiling_avgpool/averagepool_pool_11_18_average_pool_3d_valid_input/averagepool_pool_11_18_average_pool_3d_valid_input.onnx",
  "operator": "AveragePool",
  "old_version": "1.20.0",
  "new_version": "1.21.0",
  "summary": {
    "kernel_old_ms": 0.191,
    "kernel_new_ms": 0.198,
    "kernel_change_ms": 0.007000000000000006,
    "kernel_change_pct": 3.6649214659685896,
    "total_old_ms": 0.201,
    "total_new_ms": 0.204,
    "is_regression": false
  }
}

```

.vscode/settings.json
tengli-alaska: Fixed - restored .vscode/settings.json.

Issue #26750, opened 2025-12-08
[Feature Request] Enable interop between model editor APIs and new ORT compile APIs


adrastogi:
### Describe the feature request

There were some new APIs introduced earlier in the year to streamline the process of compiling ORT models, i.e., https://onnxruntime.ai/docs/api/c/struct_ort_compile_api.html. 

These APIs can accept the model as a file path or as a buffer. This feature request proposes extending the API to support models generated via the model editor APIs. https://onnxruntime.ai/docs/api/c/struct_ort_model_editor_api.html#details

### Describe scenario use case

This would be useful for WebNN scenarios where the browser crafts the model graph using the editor APIs. There is currently no way to pass the composed graph API to the compile API. Ideally this could be done in a way that performs the compilation in-memory (the browser's GPU sandbox process is heavily restricted and cannot write data to the file system).

PR #27332, opened 2026-02-12
Add OrtModel input support for Compile API

adrastogi:
### Description
This change adds a feature to the Compile API, allowing an in-memory OrtModel created via the Model Editor API to be compiled directly without first serializing to a file or buffer.

### Motivation and Context

The Model Editor API and Compile API are both public C APIs in the ONNX Runtime, but until now there was no way to pass a programmatically constructed model directly to compilation. This change attempts to closes that gap (see #26750) and ensures the new code path behaves identically to the established file and buffer paths.

### DIFF
--- a/onnxruntime/test/shared_lib/test_model_builder_api.cc
+++ b/onnxruntime/test/shared_lib/test_model_builder_api.cc
@@ -781,6 +781,41 @@ Ort::Model CreateSimpleGemmModel(std::vector<std::unique_ptr<std::vector<float>>
 
   return model;
 }
+
+// Helper to run inference on the simple Gemm model and verify all output values.
+// Model is Z = 2.0 * X * Y where X is 3x4 (all ones) and Y is 4x8 (iota 1..32).
+// Expected output: each row is 2 * column_sums_of_Y = {104, 112, 120, 128, 136, 144, 152, 160}.
+void RunAndVerifySimpleGemmModel(const Ort::Model& model) {
+  Ort::SessionOptions session_options;
+  Ort::Session session(*ort_env, model, session_options);
+  ASSERT_EQ(session.GetInputCount(), 1u);
+  ASSERT_EQ(session.GetOutputCount(), 1u);
+
+  std::vector<float> input_data(3 * 4, 1.0f);
+  std::vector<int64_t> input_dims = {3, 4};
+  auto memory_info = Ort::MemoryInfo::CreateCpu(OrtDeviceAllocator, OrtMemTypeDefault);
+  auto input_tensor = Ort::Value::CreateTensor<float>(memory_info, input_data.data(), input_data.size(),
+                                                      input_dims.data(), input_dims.size());
+
+  const char* input_names[] = {"X"};
+  const char* output_names[] = {"Z"};
+  auto outputs = session.Run(Ort::RunOptions{}, input_names, &input_tensor, 1, output_names, 1);
+  ASSERT_EQ(outputs.size(), 1u);
+  ASSERT_TRUE(outputs[0].IsTensor());
+
+  auto output_shape = outputs[0].GetTensorTypeAndShapeInfo().GetShape();
+  ASSERT_EQ(output_shape, (std::vector<int64_t>{3, 8}));
+
+  const float* output_data = outputs[0].GetTensorData<float>();
+  // alpha=2.0, X is all ones, so each output row = 2 * sum of each column of Y (iota 1..32 in 4x8)
+  const std::vector<float> expected_row = {104.0f, 112.0f, 120.0f, 128.0f, 136.0f, 144.0f, 152.0f, 160.0f};
+  for (int row = 0; row < 3; ++row) {
+    for (int col = 0; col < 8; ++col) {
+      EXPECT_FLOAT_EQ(output_data[row * 8 + col], expected_row[col])
+          << "Mismatch at row=" << row << " col=" << col;
+    }
+  }
+}
 }  // namespace
 
 // Test basic compilation from OrtModel
@@ -813,9 +848,10 @@ TEST(ModelEditorCompileAPITest, BasicCompileFromOrtModel) {
   if (output_buffer != nullptr) {
     allocator->Free(output_buffer);
   }
-}
 
-// Test validation: null model pointer
+  // Verify the model still produces correct inference results after compilation
+  RunAndVerifySimpleGemmModel(model);
+}
 TEST(ModelEditorCompileAPITest, CompileFromNullModel_Fails) {
   Ort::SessionOptions session_options;
   Ort::ModelCompilationOptions compile_options(*ort_env, session_options);
@@ -876,7 +912,9 @@ TEST(ModelEditorCompileAPITest, CompileFromModelWithEmptyInputsOutputs_Fails) {
   EXPECT_THAT(status.GetErrorMessage(), ::testing::HasSubstr("input"));
 }
 
-// Test: model can be reused after compilation
+// Test: model can be reused after compilation.
+// NOTE: This is not an explicit API guarantee. It documents current behavior so that if a future change
+// breaks model reuse, the regression is surfaced and can be evaluated.
 TEST(ModelEditorCompileAPITest, ModelCanBeReusedAfterCompilation) {
   std::vector<std::unique_ptr<std::vector<float>>> weights;
   auto model = CreateSimpleGemmModel(weights);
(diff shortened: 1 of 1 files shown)


csharp/src/Microsoft.ML.OnnxRuntime/NativeCompileApiMethods.shared.cs
```diff
@@ -25,6 +25,7 @@ public struct OrtCompileApi
         public IntPtr ModelCompilationOptions_SetGraphOptimizationLevel;
         public IntPtr ModelCompilationOptions_SetOutputModelWriteFunc;
         public IntPtr ModelCompilationOptions_SetOutputModelGetInitializerLocationFunc;
+        public IntPtr ModelCompilationOptions_SetInputModel;
```
adrastogi, line 28: Debating whether or not we need this, ultimately... I realized that there aren't managed versions of the model editor APIs. But since I am touching the API table, it seemed mandatory(?).

onnxruntime/core/session/model_compilation_options.cc
```diff
@@ -249,12 +270,77 @@ Status ModelCompilationOptions::Check() const {
...
+  if (input_from_model) {
+    const Env& os_env = Env::Default();
+    if (os_env.GetEnvironmentVar(inference_session_utils::kOrtLoadConfigFromModelEnvVar) == "1") {
+      return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT,
+                             "The environment variable ORT_LOAD_CONFIG_FROM_MODEL=1 is set, but loading "
+                             "config from model is not supported for in-memory OrtModel input. "
+                             "OrtModel is programmatically constructed and has no embedded ORT config. "
+                             "Unset ORT_LOAD_CONFIG_FROM_MODEL or use file/buffer input instead.");
+    }
+  }
```
skottmckay: nit: not sure this is worth the extra code here given (afaik) reading the config from the model is a very niche usage. #Closed

onnxruntime/core/session/model_compilation_options.cc
```diff
@@ -249,12 +270,77 @@ Status ModelCompilationOptions::Check() const {
...
   const epctx::ModelGenOptions& ep_context_gen_options = session_options_.value.ep_context_gen_options;
   bool has_no_output_model_location = std::holds_alternative<std::monostate>(
       ep_context_gen_options.output_model_location);
 
-  if (has_no_output_model_location && input_from_file) {
+  // Also treat an empty output file path as "no location" since it's not usable.
+  const auto* output_path = ep_context_gen_options.TryGetOutputModelPath();
+  if (!has_no_output_model_location && output_path != nullptr && output_path->empty()) {
+    has_no_output_model_location = true;
+  }
```
skottmckay: should we instead prevent an empty path being set? #Closed

onnxruntime/core/session/model_compilation_options.cc
```diff
@@ -249,12 +270,77 @@ Status ModelCompilationOptions::Check() const {
...
+  // Fast-fail: If OrtModel has no model_path and user hasn't specified output location or embed mode,
+  // EPs that need to write context binaries will fail later. Fail early with a clear error.
+  if (input_from_model && !model_has_path && has_no_output_model_location &&
+      !ep_context_gen_options.embed_ep_context_in_model) {
+    return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT,
+                           "OrtModel has no model_path set and no output location was specified. "
+                           "Please either: (1) set the model_path on the OrtGraph before adding to OrtModel, "
+                           "(2) call SetOutputModelPath/SetOutputModelBuffer to specify an output location, or "
+                           "(3) call SetEpContextEmbedMode(true) to embed EP context in the model.");
+  }
```
skottmckay: Is it worth duplicating code for this? Not sure we need a 'fast fail' path if the issue is a usage error as that should be caught and fixed during development. #Closed

onnxruntime/core/session/utils.cc
```diff
@@ -288,6 +289,88 @@ static OrtStatus* CreateSessionAndLoadModelImpl(_In_ const OrtSessionOptions* op
...
+  sess = std::make_unique<onnxruntime::InferenceSession>(
+      options == nullptr ? onnxruntime::SessionOptions() : options->value,
+      env);
+
+#if !defined(ORT_MINIMAL_BUILD) || defined(ORT_MINIMAL_BUILD_CUSTOM_OPS)
+  // Add custom domains
+  if (options && !options->custom_op_domains_.empty()) {
+    ORT_API_RETURN_IF_STATUS_NOT_OK(sess->AddCustomOpDomains(options->custom_op_domains_));
+  }
+#endif
```
skottmckay, line 341: Should this `#if` also cover the code to add custom domains from OrtEpDevice instances? #Closed

onnxruntime/test/shared_lib/test_model_builder_api.cc
```diff
@@ -725,3 +726,369 @@ TEST(ModelEditorAPITest, CreateTypeInfo) {
...
+    compile_options.SetInputModel(static_cast<const OrtModel*>(model));
+    compile_options.SetEpContextEmbedMode(true);
+
+    std::unique_ptr<MockedOrtAllocator> allocator = std::make_unique<MockedOrtAllocator>();
+    void* output_buffer = nullptr;
+    size_t output_size = 0;
+    compile_options.SetOutputModelBuffer(allocator.get(), &output_buffer, &output_size);
+
+    Ort::Status status = Ort::CompileModel(*ort_env, compile_options);
+    EXPECT_TRUE(status.IsOK()) << "First CompileModel failed: " << status.GetErrorMessage();
```
skottmckay: FWIW there are helpers like `EXPECT_ORTSTATUS_OK` in onnxruntime\test\util\include\api_asserts.h that can potentially simplify some of the checks #Closed

onnxruntime/test/shared_lib/test_model_builder_api.cc
```diff
@@ -725,3 +726,369 @@ TEST(ModelEditorAPITest, CreateTypeInfo) {
...
+  size_t output_size = 0;
+  compile_options.SetOutputModelBuffer(allocator.get(), &output_buffer, &output_size);
+
+  Ort::Status status = Ort::CompileModel(*ort_env, compile_options);
+  EXPECT_FALSE(status.IsOK()) << "Expected CompileModel to fail for model with empty inputs/outputs";
+  EXPECT_THAT(status.GetErrorMessage(), ::testing::HasSubstr("input"));
+}
+
+// Test: model can be reused after compilation
+TEST(ModelEditorCompileAPITest, ModelCanBeReusedAfterCompilation) {
```
skottmckay: Is this something we care about? It's nice if it does it, but I'm wondering if this is a guarantee we need/want to provide or not.  #Closed

onnxruntime/test/shared_lib/test_model_builder_api.cc
```diff
@@ -725,3 +726,369 @@ TEST(ModelEditorAPITest, CreateTypeInfo) {
...
+    if (output_buffer != nullptr) {
+      allocator->Free(output_buffer);
+    }
+  }
+
+  // Model should still be usable for creating a session
+  Ort::SessionOptions session_options;
+  Ort::Session session(*ort_env, model, session_options);
+  EXPECT_EQ(session.GetInputCount(), 1u);
+  EXPECT_EQ(session.GetOutputCount(), 1u);
```
skottmckay: Would be nice to execute the model to confirm it does all work as expected. e.g. say some memory has been prematurely freed. in theory we could create a session but might not notice an actual issue until we run the model as that might be the point we attempt to read from the memory.  #Closed

onnxruntime/core/session/model_compilation_options.cc
adrastogi: Yeah, fair enough- I didn't have a good sense of how prevalent that might be, so it was more of a defensive measure. I've removed it.

onnxruntime/core/session/utils.cc
adrastogi: _Think_ I got this addressed? Let me know if it looks correct to you.

onnxruntime/test/shared_lib/test_model_builder_api.cc
adrastogi: If nothing else, it codifies the behavior in a test, so if we deviate from this in the future, at least we will know... so IMO it is ok to keep. What do you think?

csharp/src/Microsoft.ML.OnnxRuntime/NativeCompileApiMethods.shared.cs
adrastogi: Will keep as-is (please let me know if it should be removed otherwise and I'll happily do so).

onnxruntime/test/shared_lib/test_model_builder_api.cc
skottmckay: It codifies the behaviour but in doing so might implicitly suggest we are guaranteeing this and that a developer making changes needs to take that into account. Doing so could potentially lead to unnecessary complications in the design/implementation of those changes. 

I'd suggest adding a comment that we have not provided any explicit guarantees around this currently, but we should check the latest usage if we're considering changing this aspect.

onnxruntime/test/shared_lib/test_model_builder_api.cc
skottmckay: nit: Only checking the first output value feels a little minimal.

Given we have a helper to create the model that is compiled, can we have a helper to execute that model and check the output. Call that every time post-compile. 

If the expected output is painful to create I assume the specific operator used doesn't matter too much and we could change the model to something simple like an Add that has one dynamic input and one initializer based input.

onnxruntime/test/shared_lib/test_model_builder_api.cc
adrastogi: That sounds reasonable to me. I edited the comment to reflect this.

skottmckay:
Is the branch sync'd with the latest `main`? Wondering if a diff with that is causing the RN CI failure.

adrastogi:
> Is the branch sync'd with the latest `main`? Wondering if a diff with that is causing the RN CI failure.

Thanks, that seems to have fixed it. Could you help me get this merged? Seems like I cannot do this directly (not sure if there was a policy change here). 

[edit] Oh, looks like a required task was cancelled. Trying to rerun now...

PR #27351, opened 2026-02-14
Refactor props and targets of Microsoft.ML.OnnxRuntime and DllImport

tianleiwu:
This PR refactors the MSBuild assets and the C# `DllImportResolver` to simplify native library deployment and improve cross-platform support (Windows, Linux, macOS). It also adds build-time verification to ensure native binaries are automatically copied to the output directory.

### 1. Refactored MSBuild Assets (`props.xml` and `targets.xml`)
- **Simplified `props.xml`**: Removed complex conditional logic for linking native libraries in Windows. The logic now focuses on providing the correct include and import library paths based on the target architecture (x64, ARM64).
- **Enhanced `targets.xml`**:
    - Added a platform compatibility check to explicitly block unsupported x86 builds, suggesting x64 or ARM64 instead.
    - Improved the legacy .NET Framework (non-SDK) fallback logic to automatically copy native binaries from the NuGet package to the output directory using a cleaner, glob-based approach.

### 2. Streamlined `DllImportResolver`
- **Simplified Resolution Logic**: Removed the manual runtime probing of `runtimes/` subfolders in `NativeMethods.shared.cs`.
- **Platform-Specific Mapping**: Instead of searching multiple paths, the resolver now maps the generic `onnxruntime` library name to its platform-specific filename:
    - **Windows**: `onnxruntime.dll`
    - **Linux**: `libonnxruntime.so`
    - **macOS**: `libonnxruntime.dylib`
- This relies on the standard .NET library loading mechanism to find the files in the output directory or standard search paths, ensuring consistency with how other .NET libraries behave.

### 3. End-to-End Build Verification
- Updated `runtest.bat` and `runtest.sh` to perform a `dotnet build` followed by a check for the existence of the native library (`onnxruntime.dll` or `libonnxruntime.so/dylib`) in the `bin/` directory.
- This ensures that any changes to the packaging or build logic that might break "automatic copying" are caught during CI.

## Benefits Compared to Original Files

| Feature | Original Implementation | New Implementation |
| :--- | :--- | :--- |
| **Complexity** | Heavily nested MSBuild logic for specific file links. | Consolidated, architecture-aware logic using standard paths. |
| **Cross-Platform** | Resolver had hardcoded RID probing logic which was hard to maintain. | Resolver uses explicit mapping to platform-native names, leveraging .NET's built-in loader. |
| **Reliability** | Native library loading might fail if RIDs didn't match exactly. | Simplified mapping reduces edge cases and follows OS conventions (`lib` prefix). |
| **Verification** | Tests relied on `dotnet test` which might obscure build-time copy failures. | Explicit build and file check ensures native assets are correctly deployed. |
| **Support** | Implicit x86 failures could be confusing. | Explicit error message blocking x86 and guiding users to supported architectures. |

## Tests
Verified in Nuget Test and Foundry Local Package Pipelines


csharp/src/Microsoft.ML.OnnxRuntime/targets/netstandard/targets.xml
```diff
@@ -1,16 +1,44 @@
...
-                      '$(SuppressOnnxRuntimePlatformCompatibilityError)' != 'true'"
-           Text="Microsoft.ML.OnnxRuntime only supports the AnyCPU, x64, arm32, arm64 and x86 platforms at this time."/>
+  <!-- ========================================================= -->
+  <!-- 1. Block unsupported x86 -->
+  <!-- ========================================================= -->
+<Target Name="OnnxRuntime_BlockUnsupportedPlatform"
+        BeforeTargets="Build"
+        Condition="
+          '$(SuppressOnnxRuntimePlatformCompatibilityError)' != 'true' AND
+          $([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) != 'ios' AND
```
skottmckay: AFAIK there are no 32-bit iOS builds.

csharp/src/Microsoft.ML.OnnxRuntime/targets/netstandard/targets.xml
```diff
@@ -1,16 +1,44 @@
...
+          $([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) != 'android' AND
+          '$(PlatformTarget)' == 'x86'">
```
skottmckay: Do we need a check for an AnyCPU build (based on the previous implementation I'm guessing PlatformTarget may not have a value in that case) with Prefer32Bit set to true?

csharp/src/Microsoft.ML.OnnxRuntime/targets/native/props.xml
```diff
@@ -0,0 +1,37 @@
...
+        $(MSBuildThisFileDirectory)include;
+        %(AdditionalIncludeDirectories)
+      </AdditionalIncludeDirectories>
+    </ClCompile>
+  </ItemDefinitionGroup>
+
+  <!-- ========================================================= -->
+  <!-- Windows native linking -->
+  <!-- ========================================================= -->
+  <ItemDefinitionGroup Condition="'$(PlatformTarget)' == 'x64'">
```
skottmckay: Should this condition also check the target is Windows or do we not have x64/arm64 as platform targets on Linux?

csharp/src/Microsoft.ML.OnnxRuntime/NativeMethods.shared.cs
```diff
@@ -915,89 +915,18 @@ private static IntPtr DllImportResolver(string libraryName, Assembly assembly, D
...
-                    // 1. Try default loading (name only)
                     if (NativeLibrary.TryLoad(mappedName, assembly, searchPath, out IntPtr handle))
                     {
                         return handle;
                     }
-
-                    // 2. Try relative to assembly location (look into runtimes subfolders)
-                    string assemblyLocation = null;
-                    try { assemblyLocation = assembly.Location; } catch { }
-                    if (!string.IsNullOrEmpty(assemblyLocation))
```
skottmckay: Do we still need this for development scenarios using an AnyCPU build? If that is the case dotnet might not copy from the runtimes directory to the build output as I don't think there's a specific RID at that point. It would do the copy when publishing though.

csharp/src/Microsoft.ML.OnnxRuntime/targets/netstandard/targets.xml
tianleiwu: Addressed in 83cbb347b9c3e3101e1a3a98fc5a847a9bd42706. Removed the iOS exemption from the unsupported-platform target, so an iOS x86 build is now rejected instead of implying a supported 32-bit iOS configuration. Focused MSBuild evaluation confirmed net9.0-ios + x86 fails with the 32-bit-target error, while the intentional net9.0-android + x86 case still succeeds.

csharp/src/Microsoft.ML.OnnxRuntime/targets/netstandard/targets.xml
tianleiwu: Addressed in 83cbb347b9c3e3101e1a3a98fc5a847a9bd42706. The compatibility condition now rejects Prefer32Bit=true independently of PlatformTarget, covering AnyCPU as well as an empty/default PlatformTarget. Focused MSBuild evaluation confirmed AnyCPU + Prefer32Bit=true fails and AnyCPU + Prefer32Bit=false succeeds.

csharp/src/Microsoft.ML.OnnxRuntime/targets/native/props.xml
tianleiwu: Addressed in 83cbb347b9c3e3101e1a3a98fc5a847a9bd42706. Windows import-library metadata is now guarded by OS == Windows_NT in both the native and netstandard props surfaces. Focused MSBuild metadata tests verified x64 emits onnxruntime.lib on Windows and emits no Link dependency on Unix for both props files; the generated nupkg contains the guarded XML in build/native and build/netstandard2.0/2.1.

csharp/src/Microsoft.ML.OnnxRuntime/NativeMethods.shared.cs
tianleiwu: Addressed in 83cbb347b9c3e3101e1a3a98fc5a847a9bd42706. The merged resolver keeps OS-specific library naming and now probes the assembly-adjacent runtimes/<rid>/native layout (plus current-OS RID fallbacks and AppContext.BaseDirectory), covering AnyCPU development builds where NuGet does not flatten a RID asset into the output root. The net8.0 managed project builds with 0 warnings/errors, and the end-to-end package script retains an explicit post-build native-library presence check before running tests.

PR #27354, opened 2026-02-15
Update python_requires

XXXXRT666:
### Description
The requires-python field in the [metadata](https://pypi.org/pypi/onnxruntime/json) is currently set to `python >= 3.10`. However, since ONNX Runtime has dropped support for `Python 3.10`, this constraint is now inaccurate and may lead to dependency resolution issues in `uv lock`.

Related Commits: #26397

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -874,7 +874,7 @@ setup(
     data_files=data_files,
     install_requires=install_requires,
     extras_require=extras_require,
-    python_requires=">=3.10",
+    python_requires=">=3.11",
     keywords="onnx machine learning",
     entry_points={
         "console_scripts": [


XXXXRT666:
```toml
[[package]]
name = "onnxruntime"
version = "1.20.1"
source = { registry = "https://pypi.org/simple" }
resolution-markers = [
    "python_full_version < '3.10'",
]

[[package]]
name = "onnxruntime"
version = "1.24.1"
source = { registry = "https://pypi.org/simple" }
resolution-markers = [
    "python_full_version >= '3.11'",
    "python_full_version == '3.10.*'",
]
```

Actually we don't have 1.24.1 wheels for `python 3.10`

tianleiwu:
onnxruntime does support build from source with python 3.10. python 3.10 end of life is October 2026.

It is also right that ORT 1.24 removed python 3.10 packages, as shown in 
https://github.com/microsoft/onnxruntime/blob/0df5dbc7c4d37c6384691504acdd25daeface454/setup.py#L629-L632

PR #27356, opened 2026-02-16
Add/Update telemetry events

dabhattimsft:
### Description
ModelLoadStart/End - InferenceSession::LoadWithLoader, InferenceSession::LoadOrtModelWithLoader
SessionCreationEnd - InferenceSession::Initialize
RegisterEpLibraryWithLibPath,  RegisterEpLibraryStart/End - Environment::RegisterExecutionProviderLibrary

Update: RuntimePerf event now logs status as well and is triggered more frequently with exponential backoff. 
It is also now triggered from ~InferenceSession() to log data in the tail.

### Motivation and Context
To better measure health


onnxruntime/core/session/inference_session.cc
```diff
@@ -976,6 +976,9 @@ common::Status InferenceSession::LoadWithLoader(std::function<common::Status(std
     tp = session_profiler_.Start();
   }
   ORT_TRY {
+    const Env& env = Env::Default();
+    env.GetTelemetryProvider().LogModelLoadStart(session_id_);
```
adrastogi, line 999: Non-blocking: this is more general than the changes here, but I guess a byproduct of the Start / Stop telemetry pattern is that the Stop call is not universally guaranteed to fire due to e.g., early returns from some of those ORT_RETURN macros. We should have scope guards (or whatever the equivalent pattern is in the code base). I can file an issue and try to fix this later.

dabhattimsft: Yep, I thought about that. My thinking was to have those instances end up as timeout in MC state-machine.

onnxruntime/core/platform/windows/telemetry.cc
```diff
@@ -457,7 +457,8 @@ void WindowsTelemetry::LogCompileModelStart(uint32_t session_id,
                     TraceLoggingInt32(graph_optimization_level, "graphOptimizationLevel"),
                     TraceLoggingBool(embed_ep_context, "embedEpContext"),
                     TraceLoggingBool(has_external_initializers_file, "hasExternalInitializersFile"),
-                    TraceLoggingString(execution_provider_string.c_str(), "executionProviderIds"));
+                    TraceLoggingString(execution_provider_string.c_str(), "executionProviderIds"),
+                    TraceLoggingString(ORT_CALLER_FRAMEWORK, "frameworkName"));
```
ashrit-ms, line 461: I wouldn't duplicate frameworkName by adding it to multiple events unless it is absolutely needed. In this you should be able to filter this out using AppSessionGuid from ProcessInfo.

onnxruntime/core/platform/windows/telemetry.cc
```diff
@@ -480,7 +481,8 @@ void WindowsTelemetry::LogCompileModelComplete(uint32_t session_id,
                     TraceLoggingBool(success, "success"),
                     TraceLoggingUInt32(error_code, "errorCode"),
                     TraceLoggingUInt32(error_category, "errorCategory"),
-                    TraceLoggingString(error_message.c_str(), "errorMessage"));
+                    TraceLoggingString(error_message.c_str(), "errorMessage"),
+                    TraceLoggingString(ORT_CALLER_FRAMEWORK, "frameworkName"));
```
ashrit-ms, line 485: Same as above.

onnxruntime/core/platform/windows/telemetry.cc
```diff
@@ -668,4 +670,132 @@ void WindowsTelemetry::LogProviderOptions(const std::string& provider_id, const
...
+  TraceLoggingWrite(telemetry_provider_handle,
+                    "RegisterEpLibraryWithLibPath",
+                    TraceLoggingBool(true, "UTCReplace_AppSessionGuid"),
+                    TelemetryPrivacyDataTag(PDT_ProductAndServiceUsage),
+                    TraceLoggingKeyword(MICROSOFT_KEYWORD_MEASURES),
+                    TraceLoggingLevel(WINEVENT_LEVEL_INFO),
+                    // Telemetry info
+                    TraceLoggingUInt8(0, "schemaVersion"),
+                    TraceLoggingString(registration_name.c_str(), "registrationName"),
+                    TraceLoggingString(lib_path.c_str(), "libPath"),
```
ashrit-ms, line 744: Would the library path in this case reveal user information, for example: username? Microsoft's telemetry guidelines prevent us from collecting such information.

onnxruntime/core/platform/windows/telemetry.cc
```diff
@@ -668,4 +670,132 @@ void WindowsTelemetry::LogProviderOptions(const std::string& provider_id, const
...
+                    TraceLoggingUInt8(0, "schemaVersion"),
+                    TraceLoggingUInt32(session_id, "sessionId"),
+                    TraceLoggingBool(status.IsOK(), "isSuccess"),
+                    TraceLoggingUInt32(status.Code(), "errorCode"),
+                    TraceLoggingUInt32(status.Category(), "errorCategory"),
+                    TraceLoggingString(status.IsOK() ? "" : status.ErrorMessage().c_str(), "errorMessage"),
+                    TraceLoggingString(ORT_CALLER_FRAMEWORK, "frameworkName"));
+}
+
+void WindowsTelemetry::LogRunStart(uint32_t session_id) const {
```
ashrit-ms: I feel some of these can be limited to a single event, instead of the current Start Stop implementation. Less is more when it comes to telemetry.

Is the goal to use these Start - Stop events to measure performance? Wouldn't it be better to time it in code and then share the timing data via a single event?

onnxruntime/core/session/inference_session.cc
```diff
@@ -3220,6 +3252,10 @@ Status InferenceSession::Run(const RunOptions& run_options,
       telemetry_.total_runs_since_last_ = 0;
       telemetry_.total_run_duration_since_last_ = 0;
       telemetry_.duration_per_batch_size_.clear();
+
+      // Double the interval, capping at kRuntimePerfMaxInterval
+      telemetry_.runtime_perf_interval_ = std::min(telemetry_.runtime_perf_interval_ * 2,
```
ashrit-ms: Can you please share the design decision behind this? Every event we fire will add a performance overhead which we want to reduce especially in the inferencing path.

adrastogi:
>It is also now triggered from ~InferenceSession() to log data in the tail.

Where does that happen? Having trouble finding it.

onnxruntime/core/session/inference_session.cc
```diff
@@ -3210,16 +3241,22 @@ Status InferenceSession::Run(const RunOptions& run_options,
...
+                                                telemetry_.duration_per_batch_size_,
+                                                retval);
       // reset counters
       telemetry_.time_sent_last_ = std::chrono::high_resolution_clock::now();
       telemetry_.total_runs_since_last_ = 0;
       telemetry_.total_run_duration_since_last_ = 0;
       telemetry_.duration_per_batch_size_.clear();
+
+      // Double the interval, capping at kRuntimePerfMaxInterval
+      telemetry_.runtime_perf_interval_ = std::min(telemetry_.runtime_perf_interval_ * 2,
```
adrastogi: I am thinking about this policy. If we always increase the backoff when there's a failure, I am wondering if there may be scenarios where the failure data actually ends up getting underreported relative to stable operation (i.e., because the backoff keeps increasing, or maybe because there are failures which are more temporally bounded that wouldn't show up after the interval update). The flip side of this is that we would then need some other control for failures, which would increase complexity.

onnxruntime/core/session/inference_session.cc
```diff
@@ -736,6 +736,18 @@ InferenceSession::InferenceSession(const SessionOptions& session_options, const
...
 InferenceSession::~InferenceSession() {
+  // Flush any remaining RuntimePerf counters
+  {
+    std::lock_guard<std::mutex> telemetry_lock(telemetry_mutex_);
+    if (telemetry_.total_runs_since_last_ > 0) {
+      Env::Default().GetTelemetryProvider().LogRuntimePerf(session_id_,
+                                                           telemetry_.total_runs_since_last_,
+                                                           telemetry_.total_run_duration_since_last_,
+                                                           telemetry_.duration_per_batch_size_,
+                                                           Status::OK());
```
adrastogi: How is this going to be used in the metric definition? The status for the tail events might not be OK, but that's what's always reported here. Maybe it's not a problem but just wanted to double-check.

onnxruntime/core/session/inference_session.cc
skottmckay: telemetry_ is per-session I believe. You can't re-use a session by loading a different model into it once the first model has successfully loaded. Given that I expect this code is only hit by someone repeatedly trying to load a model into an inference session instance and it failing every time.

adrastogi: Good point- ok, so I guess the concern I have is largely theoretical and we would not expect to see this in practice. I'll mark this resolved.

onnxruntime/core/session/inference_session.cc
dabhattimsft: We trigger telemetry on every failure from Run() so this batch shouldn't have any failures

adrastogi (approved):

onnxruntime/core/session/inference_session.h
```diff
@@ -976,8 +976,10 @@ class InferenceSession {
     std::unordered_map<int64_t, long long> duration_per_batch_size_;  // the duration (us) of Run() calls per batch size since the last report
 
     TimePoint time_sent_last_;  // the TimePoint of the last report
-    // Event Rate per provider < 20 peak events per second
-    constexpr static long long kDurationBetweenSending = 1000 * 1000 * 60 * 10;  // duration in (us).  send a report every 10 mins
+    // RuntimePerf backoff: starts at 2s, doubles each emission, caps at 10 min
+    constexpr static long long kRuntimePerfInitialInterval = 2 * 1000 * 1000;    // 2 seconds in (us)
+    constexpr static long long kRuntimePerfMaxInterval = 1000 * 1000 * 60 * 10;  // 10 minutes in (us)
+    long long runtime_perf_interval_ = kRuntimePerfInitialInterval;
```
skottmckay: nit: I don't think we use `long long` anywhere else in the codebase. could we use int64_t?  #Closed

onnxruntime/core/session/inference_session.cc
```diff
@@ -736,6 +736,18 @@ InferenceSession::InferenceSession(const SessionOptions& session_options, const
 #endif  // !defined(ORT_MINIMAL_BUILD)
 
 InferenceSession::~InferenceSession() {
+  // Flush any remaining RuntimePerf counters
+  {
+    std::lock_guard<std::mutex> telemetry_lock(telemetry_mutex_);
+    if (telemetry_.total_runs_since_last_ > 0) {
+      Env::Default().GetTelemetryProvider().LogRuntimePerf(session_id_,
```
skottmckay, line 743: Given this is in a dtor should there be a try/catch? #Closed

onnxruntime/core/session/inference_session.cc
```diff
@@ -976,6 +988,9 @@ common::Status InferenceSession::LoadWithLoader(std::function<common::Status(std
     tp = session_profiler_.Start();
   }
   ORT_TRY {
+    const Env& env = Env::Default();
```
skottmckay: nit: this could be called before the ORT_TRY so we don't need another Env::Default() call on line 1030

onnxruntime/core/session/inference_session.cc
skottmckay: Should this get halved on success until back to the initial interval? #Closed

onnxruntime/core/session/inference_session.h
```diff
@@ -976,8 +976,10 @@ class InferenceSession {
     std::unordered_map<int64_t, long long> duration_per_batch_size_;  // the duration (us) of Run() calls per batch size since the last report
 
     TimePoint time_sent_last_;  // the TimePoint of the last report
-    // Event Rate per provider < 20 peak events per second
-    constexpr static long long kDurationBetweenSending = 1000 * 1000 * 60 * 10;  // duration in (us).  send a report every 10 mins
+    // RuntimePerf backoff: starts at 2s, doubles each emission, caps at 10 min
+    constexpr static long long kRuntimePerfInitialInterval = 2 * 1000 * 1000;    // 2 seconds in (us)
+    constexpr static long long kRuntimePerfMaxInterval = 1000 * 1000 * 60 * 10;  // 10 minutes in (us)
```
skottmckay: Going from 10 minutes to 2 seconds seems like a significant change. How was the new value determined?

Can you also elaborate on the problem you're trying to address by making this change? Are we missing telemetry because the session is killed inside 10 minutes and we don't send anything? Or are we getting too much reporting from failed inferences? #Closed

onnxruntime/core/session/inference_session.cc
skottmckay: Or a misleading point. I missed that this was in Run not Load in the GH display of changes, so my point wasn't really applicable to this usage.

onnxruntime/core/session/inference_session.cc
```diff
@@ -3210,16 +3241,22 @@ Status InferenceSession::Run(const RunOptions& run_options,
     telemetry_.total_run_duration_since_last_ += TimeDiffMicroSeconds(tp);
     telemetry_.duration_per_batch_size_[batch_size] += TimeDiffMicroSeconds(tp);
 
-    if (TimeDiffMicroSeconds(telemetry_.time_sent_last_) > Telemetry::kDurationBetweenSending) {
-      // send the telemetry
+    // Emit RuntimePerf on scheduled interval or on error
+    if ((TimeDiffMicroSeconds(telemetry_.time_sent_last_) > telemetry_.runtime_perf_interval_) ||
+        !retval.IsOK()) {
```
skottmckay: Does increasing the interval need to be conflated with an inference failing?

Are there any potential performance issues from reporting every inference failure?

onnxruntime/core/session/inference_session.cc
dabhattimsft: I was seeing Run() trigger a lot (almost every token) on LLM models that's why this backoff.

onnxruntime/core/session/inference_session.h
dabhattimsft: Idea is to create a failure rate measure to track things across releases. Currently, Run() doesn't have a success event which makes it harder to do that. With 10 min interval, I suspect we might be missing instances which exit before that triggers. 2s is arbitrary - I was thinking to send telemetry frequently. But not so much that we end up with garrulous behavior - so exponential backoff strategy.

onnxruntime/core/session/inference_session.cc
dabhattimsft: I'd expect perf impact to be minimal. My thought process is to log these failures to track quality across releases.

onnxruntime/core/session/inference_session.cc
```diff
@@ -3210,16 +3241,22 @@ Status InferenceSession::Run(const RunOptions& run_options,
...
       env.GetTelemetryProvider().LogRuntimePerf(session_id_, telemetry_.total_runs_since_last_,
                                                 telemetry_.total_run_duration_since_last_,
-                                                telemetry_.duration_per_batch_size_);
+                                                telemetry_.duration_per_batch_size_,
```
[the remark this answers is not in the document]
dabhattimsft: Logging should be reasonably fast enough for this to be not a concern.

onnxruntime/core/session/environment.cc
```diff
@@ -611,6 +617,8 @@ Status Environment::CreateAndRegisterInternalEps() {
 Status Environment::RegisterExecutionProviderLibrary(const std::string& registration_name, const ORTCHAR_T* lib_path) {
   std::lock_guard<std::mutex> lock{mutex_};
 
+  Env::Default().GetTelemetryProvider().LogRegisterEpLibraryWithLibPath(registration_name, ToUTF8String(lib_path));
```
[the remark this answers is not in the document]
dabhattimsft: These libs are usually in system dirs and full path contains package name which would be helpful in telemetry.

onnxruntime/core/session/inference_session.cc
skottmckay: Wasn't questioning the backoff, but if something changes and Run calls start succeeding do we want to treat the backoff as a high water mark or to move back towards the initial reporting interval?

skottmckay: I guess it doesn't matter given the previous implementation always used a 10 minute interval.

onnxruntime/core/session/inference_session.cc
skottmckay: Regarding the conflation, if there's a failure should there be 2 calls to LogRuntimePerf? One with the stats for the previous successful requests and a separate one for the failed request? Feels a little off to call LogRuntimePerf with data from multiple requests and a status that only applies to one of them.

dabhattimsft: Since we always trigger it on failures, assumption is that all previous must have succeeded.

skottmckay: How do process the telemetry output when that happens? e.g. I have a number of successful requests that haven't been written yet. A failure causes that data to be written, but that output will have a failed status even though the failure is a single request.

ashrit-ms: Agree with Scott here, we should be using LogRuntimeError() telemetry event to log errors.

onnxruntime/core/platform/windows/telemetry.cc
dabhattimsft: Idea with these events is to create scenario specific RQVs to track quality across releases and builds. Having framework name in the event would reduce complexity on that.

onnxruntime/core/platform/windows/telemetry.cc
dabhattimsft: I can remove frameworkName from this event and keep it in CompileModelStart. I am not sure if I understand the concern with this though.

onnxruntime/core/platform/windows/telemetry.cc
```diff
@@ -668,4 +675,116 @@ void WindowsTelemetry::LogProviderOptions(const std::string& provider_id, const
...
+                    TraceLoggingString(ORT_CALLER_FRAMEWORK, "frameworkName"));
+}
+
+void WindowsTelemetry::LogRegisterEpLibraryEnd(const std::string& registration_name,
+                                               const common::Status& status) const {
+  if (global_register_count_ == 0 || enabled_ == false)
+    return;
+
+  TraceLoggingWrite(telemetry_provider_handle,
+                    "RegisterEpLibraryEnd",
```
angelser, line 770: Question, do you have in mind how you intend to use the start and end events? Could it be enough with one event that reports the result and if needed measures any perf (time between start and end) as needed?

onnxruntime/core/platform/windows/telemetry.cc
```diff
@@ -668,4 +675,116 @@ void WindowsTelemetry::LogProviderOptions(const std::string& provider_id, const
...
+                    TraceLoggingString(lib_path.c_str(), "libPath"),
+                    TraceLoggingString(ORT_CALLER_FRAMEWORK, "frameworkName"));
+}
+
+void WindowsTelemetry::LogRegisterEpLibraryStart(const std::string& registration_name) const {
+  if (global_register_count_ == 0 || enabled_ == false)
+    return;
+
+  TraceLoggingWrite(telemetry_provider_handle,
+                    "RegisterEpLibraryStart",
```
angelser, line 753: Same comment I had in the Load model. It would be good to consider how the event will be used and if possible condense to one that reports result and any overall info needed. Reduces data volume and also effort in any data processing pipelines.

onnxruntime/core/platform/windows/telemetry.cc
dabhattimsft: Idea is to have MC state machine with and start and stop event to calculate reliability percentage. With just end event, we'd miss early return scenarios. Also, if anything goes wrong after start, we should end up with timeout state in MC for those.

onnxruntime/core/platform/windows/telemetry.cc
dabhattimsft: Yeah with start and end pairs, we should be able to catch early returns/anything else going wrong after start but before end

onnxruntime/core/session/inference_session.cc
dabhattimsft: Idea is to calculate inference failure rate. This event has total runs in the batch so we can emit weighted failure count for that. And for periodic triggers of this event (ones which should have all success), we emit success count for that and we can calculate failure/success rate based on that.

onnxruntime/core/session/inference_session.cc
dabhattimsft: Yeah, idea is to calculate inference failure rate. While we have RuntimeError, we dont have info on how many of these succeeded. With exponential backoff, we should get ~10 events in 10 minutes and then 1 event per ~10 minutes or on every failure.

onnxruntime/core/session/inference_session.cc
skottmckay: To me that's unnecessarily complicating the logic to process the telemetry output as now I need to know "if there is more than one event in the batch and the status is failed the only failed event is the last one" to do that correctly. that's a lot of implementation detail required to process output in a completely different world vs ensuring that the status applies to all items in the batch.

You're also potentially skewing data in the batch. Say there are 2 successful requests that ran for 500ms each. Next request fails. If it fails 10ms in I now have 3 requests with a total time of 1010 ms in the batch, giving ~300ms latency which is badly off. If I split out the successful vs failed I don't have that issue.

onnxruntime/core/platform/windows/telemetry.cc
dabhattimsft: Idea is to have MC state machine with and start and stop event to calculate reliability percentage. With just end event, we'd miss early return scenarios. Also, if anything goes wrong after start, we should end up with timeout state in MC for those.

onnxruntime/core/platform/windows/telemetry.cc
```diff
@@ -528,7 +530,7 @@ void WindowsTelemetry::LogRuntimeError(uint32_t session_id, const common::Status
 }
 
 void WindowsTelemetry::LogRuntimePerf(uint32_t session_id, uint32_t total_runs_since_last, int64_t total_run_duration_since_last,
-                                      std::unordered_map<int64_t, long long> duration_per_batch_size) const {
+                                      const std::unordered_map<int64_t, long long>& duration_per_batch_size) const {
```
yuslepukhin, line 533: We use InlinedHashMap to reduce the number of memory allocations.

onnxruntime/core/platform/windows/telemetry.cc
yuslepukhin: Please, respond to this comment.

dabhattimsft: I am now only logging filename instead of full path. https://github.com/microsoft/onnxruntime/pull/27356/changes#diff-9dab491475a7715152e2349929066ccddaf334d8609f2fa941670e8c6e5035dcR620

yuslepukhin (dismissed):

yuslepukhin:
Looks like the build artifact does not have dependencies set up. You can probably revert this commit.

> In file included from /Users/runner/work/onnxruntime/onnxruntime/include/onnxruntime/core/common/inlined_containers.h:8:
/Users/runner/work/onnxruntime/onnxruntime/include/onnxruntime/core/common/inlined_containers_fwd.h:27:10: fatal error: 'absl/container/inlined_vector.h' file not found
   27 | #include <absl/container/inlined_vector.h>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

tianleiwu:
## Positives
1. **Performance optimization:** The signature of `LogRuntimePerf` has been thoughtfully updated to pass the `duration_per_batch_size` map by `const &`, avoiding unnecessary copies.
2. The exponential backoff approach for `RuntimePerf` logging interval is a good approach to capture early load volatility without spamming events downstream.
3. Catching exceptions ([ORT_CATCH(const std::exception& e)](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/session/inference_session.cc#L999-L1004) and [ORT_CATCH(...)](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/session/inference_session.cc#L999-L1004)) in [~InferenceSession()](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/session/inference_session.cc#L738-L757) across the telemetry flush protects the destructor from crashing the process if an error occurs.

## Bugs & Issues

### 1. [End](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/session/inference_session.cc#L3666-L3678) Telemetry Events bypassed on Early Returns
The newly added "End" telemetry events (e.g. `LogModelLoadEnd`, `LogSessionCreationEnd`, `LogRegisterEpLibraryEnd`) are placed at the very end of their respective functions. However, these functions use `ORT_RETURN_IF_ERROR_SESSIONID_` or `ORT_RETURN_IF_ERROR` macros which return early if the enclosed operation fails.

**Files & lines affected:**
- `Environment::RegisterExecutionProviderLibrary`: Early returns like `ORT_RETURN_IF_ERROR(EpInfo::Create(...))` completely bypass `env.GetTelemetryProvider().LogRegisterEpLibraryEnd(registration_name, status);`.
- `InferenceSession::LoadWithLoader`: `ORT_RETURN_IF_ERROR_SESSIONID_` macros expand to a function-level `return` even when inside `ORT_TRY { }`. Because `LogModelLoadEnd` is placed *after* the catch block, these early returns bypass it entirely.
- `InferenceSession::LoadOrtModelWithLoader`: `LogModelLoadStart` is called unconditionally at the very top, before the mutex is acquired and before the `is_model_loaded_` / `is_inited_` guard checks. Those guards do a bare `return status`, so a re-entrant or already-initialized call logs Start but never logs End. All subsequent `ORT_RETURN_IF_ERROR` paths inside the function have the same problem.
- `InferenceSession::Initialize`: Use of early return macros inside [Initialize](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/session/inference_session.cc#L2087-L2584) means `LogSessionCreationEnd` might not be logged for most failures inside the function.

**Recommendation:**
Use RAII (a scoped guard object) to trigger "Start" in the constructor and "End" in the destructor (capturing a reference to the `status` variable so the destructor can log the correct status upon exit). Alternatively, ensure all exit paths invoke the End telemetry method before returning.

### 2. Contradiction between PR Description and Code Implementation
The PR description states:
> *Update: RuntimePerf event now logs status as well and is triggered more frequently with exponential backoff and any time !retval.IsOK().*

However, the code diff shows:
- **Status is not logged:** `LogRuntimePerf`'s signature was not updated to accept a `status` argument, nor does it log one in the implementation.
- **Not triggered on error:** In `InferenceSession::Run`, the telemetry collection is explicitly gated behind `if (retval.IsOK()) { ... }`. Failed inferences are excluded and do *not* trigger the telemetry event.

If the intention was to flush `RuntimePerf` upon an error (to promptly report context when an inference finishes with failure), the `if (retval.IsOK())` gate inadvertently prevents exactly this. If excluding failed inferences from the metrics batch was the true intent, the PR description should be updated to reflect that `RuntimePerf` ignores failed inferences rather than triggering on them.

### 3. [RegisterExecutionProviderLibrary](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/session/environment.cc#L539-L596) — Same Early-Return Problem
The private overload of [RegisterExecutionProviderLibrary](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/session/environment.cc#L539-L588) (which does the real work) already has an `ORT_TRY`/[ORT_CATCH](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/session/inference_session.cc#L999-L1004) block. However, the calls inside that block — e.g. `ORT_RETURN_IF_ERROR(EpInfo::Create(...))`, `ORT_RETURN_IF_ERROR(CreateDataTransferForFactory(...))` — all cause early function returns that bypass `LogRegisterEpLibraryEnd` at the bottom. The PR only handles the one upfront duplicate-check early exit; all other failure paths are still uncovered.

### 4. `LogModelLoadEnd` Hardcodes `Status::OK()` in [LoadOrtModelWithLoader](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/session/inference_session.cc#L1649-L1772)
The call at the bottom of [LoadOrtModelWithLoader](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/session/inference_session.cc#L1649-L1772) is:
```cpp
env.GetTelemetryProvider().LogModelLoadEnd(session_id_, Status::OK());
return Status::OK();
```
Unlike [LoadWithLoader](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/session/inference_session.cc#L971-L999), which passes a `status` variable, this hardcodes success. The call is only reachable on the success path today, so it is functionally correct, but it is fragile: any future code that inserts a new conditional return above it will log the wrong (always-OK) status. It would be more robust to use a `status` variable throughout, consistent with the pattern used in [LoadWithLoader](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/session/inference_session.cc#L971-L999).

## Summary
The telemetry hooks are structurally unsound: `ORT_RETURN_IF_ERROR` / `ORT_RETURN_IF_ERROR_SESSIONID_` macros cause function-level returns that bypass all three End events across multiple functions. An RAII scope guard that captures a `status` reference and calls the End event in its destructor would fix this robustly. Additionally, the PR description contradicts the code on whether `RuntimePerf` fires on failed inferences.

adrastogi (approved):

PR #27359, opened 2026-02-16
[CPU] GridSample operator performance improvement on bilinear interpolation…

melkap01-Arm:
### Description
This change optimises the GridSample operator in onnxrt.
1- For the GridSample nodes having similar characteristics the camera based 3D object detection model in MLPerf Automotive space, transforming the input to output coordinates with : 2D interpolation mode = linear with padding mode = zeros, align_corners = 0, fast path added.

Linear interpolation: For each (x, y), the code locates the four surrounding integer pixel centers:

(x1, y1) = (floor(x), floor(y)) (top-left)
(x2, y2) = (x1 + 1, y1 + 1) (bottom-right)
The interpolation weights reflect the fractional positions:

dx1 = x - x1, dx2 = x2 - x
dy1 = y - y1, dy2 = y2 - y 
The resulting value is the bilinear blend dy2 * (dx2 * p11 + dx1 * p12) + dy1 * (dx2 * p21 + dx1 * p22) where p11…p22 are the input pixels at those four neighbor coordinates.

Padding mode = zeros: Any neighbor index that falls outside [0, W_in-1] × [0, H_in-1] contributes 0 to the interpolation.
Each output pixel (oy, ox) carries normalized coordinates (nx, ny) in [-1, 1]. With align_corners=0, nx = -1 corresponds to a location half a pixel before the leftmost input column (i.e., x = -0.5), and nx = 1 corresponds to half a pixel beyond the rightmost column (x = W_in - 0.5). Same idea vertically for ny.

Fast Path Optimisation : The implementation can precompute all neighbor indices/weights for each output pixel once (they depend only on the grid), then reuse them for every channel. Previously indices and weights were calculated inside the loops which can be as much as (H_out*W_out like 20,000 per batch element in one case) x 32 Channels.

2-optional ARM NEON vectorization added :
- vld1_f32(ptr): loads two contiguous float values into a float32x2_t. Used to read the top and bottom neighbor
    pairs ([p11, p12], [p21, p22]).
  - vcombine_f32(low, high): concatenates two float32x2_t values into one float32x4_t, giving [p11, p12, p21, p22].
  - vdup_n_f32(val): duplicates a scalar float into both lanes of a float32x2_t.
  - vset_lane_f32(val, vec, lane): writes val into the specified lane of a float32x2_t, letting us form [w11, w12] and
    [w21, w22].
  - vmulq_f32(a, b): multiplies two float32x4_t vectors element-wise (neighbor pixels × weights).
  - vget_low_f32(vec) / vget_high_f32(vec): extract the lower or upper 2 lanes from a float32x4_t as float32x2_t.
  - vadd_f32(a, b): adds two float32x2_t vectors element-wise (forming partial sums).
  - vpadd_f32(a, b): performs pairwise adds within and across two float32x2_t vectors, collapsing four elements down
    to two.
  - vget_lane_f32(vec, lane): reads a scalar from a specific lane, giving the final interpolated value.
Most of the performance uplift coming from the 1st optimisation. 2nd optimisation using NEON intrinsics still contributes but not as much as the 1st one.
Overall performance improvement :

1 thread :                                                                 
image

 2 threads:
image

### Motivation and Context
The fast path handles denormalisation of the linear coordinates and can handle the derivation of the indices by precomputing a separate plan entry per output pixel. In PrecomputeBilinearSamplePlan2D, the loop runs across all H_out * W_out points, using the right nx/ny for each (oy, ox) and storing that point’s four indices, four weights, and mask in plans[idx].
During evaluation, EvaluatePlanForChannel iterates through the same point_count(H_out*W_out) and uses the matching plan entry for each (oy, ox). So we are not reusing one plan across different spatial positions; we precompute one plan per output location and reuse it only across channels (which share the same grid).

### DIFF
--- a/onnxruntime/core/providers/cpu/tensor/grid_sample.cc
+++ b/onnxruntime/core/providers/cpu/tensor/grid_sample.cc
@@ -280,7 +280,7 @@ void TryRunBilinearZerosFastPath2D(const Tensor& input,
                                    int64_t W_out,
                                    concurrency::ThreadPool* tp,
                                    std::vector<BilinearSamplePlan2D<T>>& sampling_plan) {
-  const int64_t plane_in = H_in * W_in;
+  const size_t plane_in = static_cast<size_t>(H_in) * static_cast<size_t>(W_in);
   const size_t plane_out = static_cast<size_t>(H_out) * static_cast<size_t>(W_out);
   sampling_plan.resize(plane_out);
 


onnxruntime/core/providers/cpu/tensor/grid_sample.cc
```diff
@@ -1,8 +1,14 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-#include "core/providers/cpu/tensor/grid_sample.h"
+#include <type_traits>
+#include <vector>
+
+#if defined(MLAS_NEON_INTRINSICS)
```
hariharans29: Just curious - how will this macro get defined for this translation unit ? IIUC- It is defined in mlasi.h (not mlas.h),  I don't know of any indirect means of inclusion here in the CPU EP for mlasi.h ?

onnxruntime/core/providers/cpu/tensor/grid_sample.cc
```diff
@@ -148,6 +154,182 @@ T GridSample<T>::PixelAtGrid3D(const T* image, int64_t d, int64_t h, int64_t w,
...
+void PrecomputeBilinearSamplePlan2D(const T* grid_data,
+                                    int64_t H_out,
+                                    int64_t W_out,
+                                    int64_t H_in,
+                                    int64_t W_in,
+                                    std::vector<BilinearSamplePlan2D<T>>& plans) {
+  const int64_t point_count = H_out * W_out;
+
+  for (int64_t idx = 0; idx < point_count; ++idx) {
+    auto& plan = plans[onnxruntime::narrow<size_t>(idx)];
```
hariharans29: Nit: Just make idx size_t instead of narrowing conversion here ?

onnxruntime/core/providers/cpu/tensor/grid_sample.cc
```diff
@@ -148,6 +154,182 @@ T GridSample<T>::PixelAtGrid3D(const T* image, int64_t d, int64_t h, int64_t w,
...
+  T w11;
+  T w12;
+  T w21;
+  T w22;
+  uint8_t mask = 0;
+};
+// PrecomputeBilinearSamplePlan2D, the loop runs across all H_out * W_out points, using the right nx/ny for each (oy, ox) and
+// storing that point’s four indices, four weights, and mask in plans[idx]
+template <typename T>
+void PrecomputeBilinearSamplePlan2D(const T* grid_data,
```
hariharans29, line 176: Could you also document that it is not only simple bilinear but bilinear + zero padding + no align_corners ?

onnxruntime/core/providers/cpu/tensor/grid_sample.cc
```diff
@@ -148,6 +154,182 @@ T GridSample<T>::PixelAtGrid3D(const T* image, int64_t d, int64_t h, int64_t w,
...
+    plan.w22 = dy1 * dx1;
+    plan.mask = mask;
+  }
+}
+
+template <typename T>
+void EvaluatePlanForChannel(const T* input_data,
+                            T* output_data,
+                            int64_t W_in,
+                            const BilinearSamplePlan2D<T>* plan_data,
```
hariharans29, line 231: Nit: Can we just pass in a const reference instead of const pointer to make it explicit that we don't expect any nullptr ?

onnxruntime/core/providers/cpu/tensor/grid_sample.cc
```diff
@@ -148,6 +154,182 @@ T GridSample<T>::PixelAtGrid3D(const T* image, int64_t d, int64_t h, int64_t w,
...
+                                   int64_t C,
+                                   int64_t H_in,
+                                   int64_t W_in,
+                                   int64_t H_out,
+                                   int64_t W_out,
+                                   concurrency::ThreadPool* tp,
+                                   std::vector<BilinearSamplePlan2D<T>>& sampling_plan) {
+  const int64_t plane_in = H_in * W_in;
+  const int64_t plane_out = H_out * W_out;
+  sampling_plan.resize(onnxruntime::narrow<size_t>(plane_out));
```
hariharans29: Nit: plane_out can be size_t ?

onnxruntime/core/providers/cpu/tensor/grid_sample.cc
```diff
@@ -265,6 +448,21 @@ Status GridSample<T>::Compute(OpKernelContext* context) const {
               }
             }
           });
+    };
+
+    const bool can_use_fast_path = (mode_ == Linear && padding_mode_ == Zeros && !align_corners_);
```
hariharans29: Not sure why we need the changes in the generic path ? Can it just be kept as is (i.e.) move the fast path logic above the generic path and bail out of Compute() if the fast path is taken. This means we can avoid the change to make the generic path a lambda and call it in another loop in the else part. 

This is just a nit. I am just trying to keep code change minimal...

onnxruntime/core/providers/cpu/tensor/grid_sample.cc
```diff
@@ -1,8 +1,14 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-#include "core/providers/cpu/tensor/grid_sample.h"
+#include <type_traits>
```
hariharans29: Do we need type_traits for this change ?

melkap01-Arm: Yes, std::is_same_v is used for the neon intrinsics part.

onnxruntime/core/providers/cpu/tensor/grid_sample.cc
melkap01-Arm: done. First all the checks agains the constraints defined done and fast path is selected, else generic path used.

onnxruntime/core/providers/cpu/tensor/grid_sample.cc
```diff
@@ -234,14 +235,15 @@ void EvaluatePlanForChannel(const T* input_data,
                             int64_t W_in,
                             const BilinearSamplePlan2D<T>* plan_data,
                             int64_t point_count) {
+  ORT_ENFORCE(input_data != nullptr, "EvaluatePlanForChannel requires non-null input_data.");
```
hariharans29: Can the input just be a reference ? Looking at the call site, we don't expect nullable data here - so using a ref might simplify this function body and look cleaner (avoid such enforce, etc.)

onnxruntime/core/providers/cpu/tensor/grid_sample.cc
```diff
@@ -148,6 +154,184 @@ T GridSample<T>::PixelAtGrid3D(const T* image, int64_t d, int64_t h, int64_t w,
...
+      continue;
+    }
+
+#if defined(__ARM_NEON) || defined(__ARM_NEON__) || defined(__aarch64__) || defined(_M_ARM64)
+    if constexpr (std::is_same_v<T, float>) {
+      if (plan.mask == kAllNeighborsMask) {
+        const float* row1_ptr = input_data + plan.y1 * W_in + plan.x1;
+        const float* row2_ptr = input_data + plan.y2 * W_in + plan.x1;
+
+        float32x2_t row1 = vld1_f32(row1_ptr);  // [p11, p12]
```
hariharans29: Spoke offline with @edgchen1 - We both feel that the usage of intrinsics and platform specific code is better off in MLAS.

We have 2 possible approaches forward:

1) Add an MLAS API to perform the "specialized" dot product operation performed here (i.e.) read 2 values from one array + 2 values from another array + 4 weights and perform the dot product and move all this code to its implementation. We could do this in the same PR.

2) Remove this code for this PR and add the MLAS change in a separate PR later.

Which way forward depends on the perf gains from this vectorization logic. Can you please share that for your target workload ? if it isn't much gains, can we take approach 2 and think about adding this as a future enhancement when there is an appropriate workload to take advantage of this ?

onnxruntime/test/providers/cpu/tensor/grid_sample_test.cc
```diff
@@ -727,6 +727,38 @@ TYPED_TEST(GridSampleTest, test_grid_sample_20_5D_bilinear_zeros_no_align_corner
...
+      TypeParam(1.44f)};  // both right and bottom neighbors out-of-bounds
+  test.AddInput<TypeParam>("X", X_shape, X_data);
+  test.AddInput<TypeParam>("Grid", Grid_shape, Grid_data);
+  test.AddAttribute("mode", mode);
+  test.AddAttribute("padding_mode", padding_mode);
+  test.AddAttribute("align_corners", align_corners);
+  test.AddOutput<TypeParam>("Y", Y_shape, Y_data);
+  RunTests(test, GetExecutionProviders(20));
+}
+
```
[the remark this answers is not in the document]
melkap01-Arm: implemented in grid_sample_test_custom.inc

onnxruntime/core/providers/cpu/tensor/grid_sample.cc
```diff
@@ -148,6 +154,181 @@ T GridSample<T>::PixelAtGrid3D(const T* image, int64_t d, int64_t h, int64_t w,
...
+  int64_t x2;
+  int64_t y1;
+  int64_t y2;
+  T w11;
+  T w12;
+  T w21;
+  T w22;
+  uint8_t mask = 0;
+};
+// PrecomputeBilinearSamplePlan2D, the loop runs across all H_out * W_out points, using the right nx/ny for each (oy, ox) and storing that point’s four indices, four weights, and mask in plans[idx]
```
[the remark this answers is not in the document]
melkap01-Arm: done

onnxruntime/core/providers/cpu/tensor/grid_sample.cc
```diff
@@ -148,6 +149,162 @@ T GridSample<T>::PixelAtGrid3D(const T* image, int64_t d, int64_t h, int64_t w,
...
+  }
+}
+
+template <typename T>
+void EvaluatePlanForChannel(const T& input_data,
+                            T* output_data,
+                            int64_t W_in,
+                            const BilinearSamplePlan2D<T>* plan_data,
+                            int64_t point_count) {
+  const T* input_ptr = &input_data;
```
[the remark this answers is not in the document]
hariharans29: @melkap01-Arm : Apologies. I think I misunderstood the input_data concpet yesterday. I think we can go back to the way you originally had it (pointer based). It keeps the intent clearer. Copilot also points out the same.

onnxruntime/core/providers/cpu/tensor/grid_sample.cc
```diff
@@ -179,12 +179,13 @@ void PrecomputeBilinearSamplePlan2D(const T* grid_data,
                                     int64_t H_in,
                                     int64_t W_in,
                                     std::vector<BilinearSamplePlan2D<T>>& plans) {
-  const size_t point_count = H_out * W_out;
+  const int64_t point_count = H_out * W_out;
```
hariharans29: I think even the size_t change got reverted (hence onnxruntime::narrow came back)

onnxruntime/core/providers/cpu/tensor/grid_sample.cc
```diff
@@ -179,12 +179,13 @@ void PrecomputeBilinearSamplePlan2D(const T* grid_data,
...
 
-  for (size_t idx = 0; idx < point_count; ++idx) {
-    auto& plan = plans[idx];
-    const T nx = grid_data[idx * 2];
-    const T ny = grid_data[idx * 2 + 1];
+  for (int64_t idx = 0; idx < point_count; ++idx) {
```
hariharans29: Ditto

onnxruntime/core/providers/cpu/tensor/grid_sample.cc
```diff
@@ -282,8 +283,8 @@ void TryRunBilinearZerosFastPath2D(const Tensor& input,
                                    concurrency::ThreadPool* tp,
                                    std::vector<BilinearSamplePlan2D<T>>& sampling_plan) {
   const int64_t plane_in = H_in * W_in;
-  const size_t plane_out = H_out * W_out;
-  sampling_plan.resize(plane_out);
+  const int64_t plane_out = H_out * W_out;
```
hariharans29: Ditto

melkap01-Arm: same

onnxruntime/core/providers/cpu/tensor/grid_sample.cc
```diff
@@ -148,6 +149,164 @@ T GridSample<T>::PixelAtGrid3D(const T* image, int64_t d, int64_t h, int64_t w,
...
+                                    int64_t H_in,
+                                    int64_t W_in,
+                                    std::vector<BilinearSamplePlan2D<T>>& plans) {
+  const int64_t point_count = H_out * W_out;
+
+  for (int64_t idx = 0; idx < point_count; ++idx) {
+    const auto sidx = onnxruntime::narrow<size_t>(idx);
+    auto& plan = plans[sidx];
+    const T nx = grid_data[sidx * 2];
+    const T ny = grid_data[sidx * 2 + 1];
```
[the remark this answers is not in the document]
melkap01-Arm: yes, this looks like a better option. implemented.

onnxruntime/core/providers/cpu/tensor/grid_sample.cc
hariharans29: I think this is not resolved yet ?

hariharans29:
LGTM - Please resolve the last size_t / narrowing comment. Thanks.

onnxruntime/core/providers/cpu/tensor/grid_sample.cc
```diff
@@ -148,6 +149,163 @@ T GridSample<T>::PixelAtGrid3D(const T* image, int64_t d, int64_t h, int64_t w,
...
+                                   int64_t n,
+                                   int64_t C,
+                                   int64_t H_in,
+                                   int64_t W_in,
+                                   int64_t H_out,
+                                   int64_t W_out,
+                                   concurrency::ThreadPool* tp,
+                                   std::vector<BilinearSamplePlan2D<T>>& sampling_plan) {
+  const int64_t plane_in = H_in * W_in;
+  const int64_t plane_out = H_out * W_out;
```
hariharans29:
```suggestion
  const size_t plane_out = H_out * W_out;
```

onnxruntime/core/providers/cpu/tensor/grid_sample.cc
```diff
@@ -148,6 +149,163 @@ T GridSample<T>::PixelAtGrid3D(const T* image, int64_t d, int64_t h, int64_t w,
...
+  const size_t plane_out_size = onnxruntime::narrow<size_t>(plane_out);
```
hariharans29:
```suggestion
```

onnxruntime/core/providers/cpu/tensor/grid_sample.cc
```diff
@@ -148,6 +149,163 @@ T GridSample<T>::PixelAtGrid3D(const T* image, int64_t d, int64_t h, int64_t w,
...
+  sampling_plan.resize(plane_out_size);
```
hariharans29:
```suggestion
  sampling_plan.resize(plane_out);
```

hariharans29 (approved):

onnxruntime/core/providers/cpu/tensor/grid_sample.cc
```diff
@@ -148,6 +149,162 @@ T GridSample<T>::PixelAtGrid3D(const T* image, int64_t d, int64_t h, int64_t w,
...
+    plan.mask = mask;
+  }
+}
+
+template <typename T>
+void EvaluatePlanForChannel(const T* input_data,
+                            T* output_data,
+                            int64_t W_in,
+                            const BilinearSamplePlan2D<T>* plan_data,
+                            int64_t point_count) {
```
hariharans29:
```suggestion
                            size_t point_count) {
```

onnxruntime/core/providers/cpu/tensor/grid_sample.cc
```diff
@@ -148,6 +149,162 @@ T GridSample<T>::PixelAtGrid3D(const T* image, int64_t d, int64_t h, int64_t w,
...
+  for (size_t idx = 0; idx < static_cast<size_t>(point_count); ++idx) {
```
hariharans29:
```suggestion
  for (size_t idx = 0; idx < point_count; ++idx) {
```

onnxruntime/core/providers/cpu/tensor/grid_sample.cc
```diff
@@ -148,6 +149,162 @@ T GridSample<T>::PixelAtGrid3D(const T* image, int64_t d, int64_t h, int64_t w,
...
+                                   int64_t W_in,
+                                   int64_t H_out,
+                                   int64_t W_out,
+                                   concurrency::ThreadPool* tp,
+                                   std::vector<BilinearSamplePlan2D<T>>& sampling_plan) {
+  const int64_t plane_in = H_in * W_in;
+  const size_t plane_out = H_out * W_out;
+  sampling_plan.resize(plane_out);
+
+  const T* grid_data = grid.Data<T>() + n * plane_out * 2;
```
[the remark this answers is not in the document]
hariharans29: Incorporated

onnxruntime/core/providers/cpu/tensor/grid_sample.cc
```diff
@@ -148,6 +149,162 @@ T GridSample<T>::PixelAtGrid3D(const T* image, int64_t d, int64_t h, int64_t w,
...
+                                   Tensor& output,
+                                   int64_t n,
+                                   int64_t C,
+                                   int64_t H_in,
+                                   int64_t W_in,
+                                   int64_t H_out,
+                                   int64_t W_out,
+                                   concurrency::ThreadPool* tp,
+                                   std::vector<BilinearSamplePlan2D<T>>& sampling_plan) {
+  const int64_t plane_in = H_in * W_in;
```
hariharans29:
```suggestion
  const size_t plane_in = static_cast<size_t>(H_in) * static_cast<size_t>(W_in);
```

onnxruntime/core/providers/cpu/tensor/grid_sample.cc
```diff
@@ -148,6 +149,162 @@ T GridSample<T>::PixelAtGrid3D(const T* image, int64_t d, int64_t h, int64_t w,
...
+  const size_t plane_out = H_out * W_out;
```
hariharans29:
```suggestion
  const size_t plane_out = static_cast<size_t>(H_out) * static_cast<size_t>(W_out);
```

hariharans29 (approved):

tianleiwu (approved):
Issues & Suggestions
1. Floating-point evaluation order difference (Low severity)
The original generic path computes:

dy2 * (dx2 * p11 + dx1 * p12) + dy1 * (dx2 * p21 + dx1 * p22)
The fast path computes:

w11 * p11 + w12 * p12 + w21 * p21 + w22 * p22
// where w11 = dy2*dx2, w12 = dy2*dx1, etc.
These are algebraically equivalent but differ in floating-point evaluation order. Due to floating-point non-associativity, this may produce slightly different results (within ULP tolerance). For float this is unlikely to matter in practice and the tests pass, but it's worth being aware of if exact bitwise reproducibility with the generic path is a requirement.

Recommendation: Acceptable as-is. If strict bitwise match is needed, change the fast path to:

output_data[idx] = plan.w_dy2 * (plan.w_dx2 * p11 + plan.w_dx1 * p12) +
                   plan.w_dy1 * (plan.w_dx2 * p21 + plan.w_dx1 * p22);
with separate dx1/dx2/dy1/dy2 fields instead of pre-multiplied w11..w22.

2. plane_out == 0 check ordering (Nit)
In TryRunBilinearZerosFastPath2D:

sampling_plan.resize(plane_out);
const T* grid_data = grid.Data<T>() + n * plane_out * 2;
PrecomputeBilinearSamplePlan2D(grid_data, H_out, W_out, H_in, W_in, sampling_plan);
// ...
if (plane_out == 0) {
  return;
}
The plane_out == 0 check comes after resize and PrecomputeBilinearSamplePlan2D. While resize(0) is valid and the precompute loop would be a no-op (0 iterations), moving the check before these calls would be slightly cleaner:

if (plane_out == 0) {
  return;
}
sampling_plan.resize(plane_out);
// ...
However, the caller (Compute) already returns early if Y.Shape().Size() == 0, so plane_out == 0 may be unreachable here. Consider whether this guard is needed at all.

3. Missing #include <vector> placement (Nit)
The diff shows #include <vector> is added above the project header:

#include <vector>

#include "core/providers/cpu/tensor/grid_sample.h"
Many projects (including onnxruntime) follow the convention of placing the corresponding header first, then standard library headers. Consider:

#include "core/providers/cpu/tensor/grid_sample.h"

#include <vector>

#include "core/framework/element_type_lists.h"
This is a style nit — the code compiles fine either way.

4. Grid data pointer arithmetic with mixed types
In TryRunBilinearZerosFastPath2D:

const T* grid_data = grid.Data<T>() + n * plane_out * 2;
Here n is int64_t and plane_out is size_t. The multiplication n * plane_out * 2 involves mixed signed/unsigned arithmetic. While safe given that n >= 0 is ensured by the loop, explicitly casting n to size_t (like static_cast<size_t>(n) * plane_out * 2) would silence potential compiler warnings and make intent clearer.

5. Consider documenting the scope limitation
The fast path only applies to 2D grids. The 3D trilinear case (data_dims == 3) is not optimized. A brief comment in the code or PR description noting this as future work would be helpful for maintainability.

6. Memory overhead consideration
For very large output grids, the sampling_plan vector holds H_out × W_out entries, each ~42 bytes (4×int64_t + 4×float + uint8_t + padding). For the cited use case of 20,000 output pixels this is ~840 KB — very reasonable. For unusually large grids (e.g., 4K resolution = ~8M pixels), this would be ~336 MB. A size guard or documented expectation could be useful if the PR targets general use beyond the MLPerf workload.

7. Test verification approach
The manually computed expected values in the custom tests (e.g., Y_data{2.5f, 1.8f, 2.1f, 1.44f}) appear correct based on manual calculation of the bilinear interpolation with the 2×2 input {1, 2, 3, 4}. It would be helpful to include the manual calculation steps in comments for future readers (some are partially there).

Verdict
Approve with minor suggestions. The optimization is well-designed, correctly scoped, and has been thoroughly vetted across multiple review rounds by a project maintainer. The algorithmic approach of precomputing the sampling plan is sound and provides meaningful speedups for the targeted workload. The code is clean, well-tested, and thread-safe.

The issues identified above are minor (floating-point ordering awareness, style nits, defensive checks). None are blocking. The maintainer (hariharans29) has already approved the PR.

PR #27368, opened 2026-02-17
[Core] Expose a couple of new session options to provide users finer control over NCHWc data layout

hariharans29:
### Description
This PR introduces 2 new session options that allow users finer control over NCHWc data layout usage for their models and prompts users to consider using these 2 options when suitable candidate nodes are encountered in their models.

**Background:**
Sometimes, a model could have an "outlier" convolution operation with large kernel sizes (eg) 11x11 and a good amout of filters (eg) 64 filters. Since, NCHWc layout transformations are turned on by default, data layout for the Conv operation is changed to NCHWc (if other Conv node parameters fit the bill). The NCHWc Conv implementations in MLAS use a "direct" convolution implementation and large kernel sizes + a good amount of filters make the entire operation a heavily memory badwidth bound operation. With this background, it makes NCHWc data layouts for such models not a good choice for certain platforms whereas on some other platforms (possibly platforms with superior memory bandwidth characteristics and/or other memory characteristics like cache sizes), it still works well. The work-around today is for users to drop down to a lower graph optimization level that does not include the NCHWc transformer on platforms where perf is poor for such models. This seems like a bad precedence to set because it would mean users will miss out on other L3 and L4 optimizers. Also, may users may not even know that dropping down to a lower non-default optimization level suits their model better on their target platform. Hence, this PR introduces control knobs and relevant prompting for users to consider these options by appropriate warnings.

Introduced Session Options:

**1) session.use_nchw_layout_for_large_conv**
Only drops down "large conv" operations to NCHW when set. 
Rest of the model may still use NCHWc. Hence, re-order nodes are introduced at the boundary of NCHWc and NCHW nodes. 
This option is useful when there is one large Conv in the model and the rest of the model are pretty vanilla Conv operations and the re-ordering nodes don't adversely affect performance.

**1) session.disable_nchwc_layout_transformation**
Completely disables NCHWc layout transformation and keeps the entire model in NCHW (default data layout). 
No re-ordering overhead.
This option is useful when re-ordering nodes introduced by session option (1) dominates performance (i.e.) there is mix of large conv and vanilla small conv operations in the model

### Motivation and Context
Takeaway from https://github.com/microsoft/onnxruntime/issues/26992 and possibly https://github.com/microsoft/onnxruntime/issues/23587 (although it needs a repro model to be sure)

### DIFF
--- a/include/onnxruntime/core/session/onnxruntime_session_options_config_keys.h
+++ b/include/onnxruntime/core/session/onnxruntime_session_options_config_keys.h
@@ -267,9 +267,9 @@ static const char* const kDebugLayoutTransformation = "session.debug_layout_tran
 // Uses NCHW layout for large convolution operations when a large convolution operation is detected
 // in the model as the NCHWc layout implementation performance is worse on some platforms than the NCHW implementation.
 // This is only relevant when operating on an NCHWc supported platform.
-// When this option is used - the large convolution operations in a model that would originally used the NCHWc format
-// will now use NCHW for better performance.
-// The rest of the model will continue to use NCHWc format and benefit from the performance advantages of NCHWc where applicable.
+// When this option is used, large convolution operations in a model that would originally use the NCHWc format
+// will instead use NCHW for better performance.
+// The rest of the model will continue to use the NCHWc format and benefit from its performance advantages where applicable.
 // So, this option might introduce some overhead of re-ordering data between NCHW and NCHWc formats, but it can improve the
 // overall performance of models with large convolution operations on certain hardware platforms.
 // The default value is "0", which means the NCHWc layout transformation is not disabled and will be used even


onnxruntime/core/optimizer/graph_transformer_utils.cc
```diff
@@ -412,9 +412,13 @@ InlinedVector<std::unique_ptr<GraphTransformer>> GenerateTransformers(
...
 #ifndef DISABLE_CONTRIB_OPS
-      // Register the NCHWc layout transformer if supported by the platform.
-      if (MlasNchwcGetBlockSize() > 1) {
-        transformers.emplace_back(std::make_unique<NchwcTransformer>());
+      // Register the NCHWc layout transformer if supported by the platform and if user didn't explicitly disable it.
+      const bool disable_nchwc = session_options.config_options.GetConfigOrDefault(kOrtSessionOptionsDisableNchwcLayoutTransformation, "0") == "1";
+
+      if (MlasNchwcGetBlockSize() > 1 && !disable_nchwc) {
+        const bool disable_nchwc_layout_for_large_conv =
+            session_options.config_options.GetConfigOrDefault(kOrtSessionOptionsUseNchwLayoutForLargeConv, "0") == "1";
```
tianleiwu, line 420: The naming kOrtSessionOptionsUseNchwLayoutForLargeConv indicates  use_nchw instead of disable_nchwc. Do we have only two options (NCHW, NCHWc) instead of NHWC etc in MLAS?

onnxruntime/core/optimizer/nchwc_transformer.cc
```diff
@@ -335,6 +346,39 @@ void NchwcTransformerImpl::TransformConv(Node& node) {
...
+
+  // For large kernels with a large number of input and output channels, some platforms may perform better with the NCHWc transformation disabled.
+  // For example, on some AVX platforms, the direct convolution implementation used for NCHWc tensors makes the kernels extremely memory bound and
+  // perform poorly on these platforms and that coupled with potential under-clocking on Intel's AVX512 results in extremely poor performance
+  // (sometimes to the order of 10x slower than the NCHW implementation).
+  // In these cases, it's better to skip the NCHWc transformation and allow the convolution to be optimized as a standard NCHW convolution which may
+  // use a more efficient algorithm for large kernels. The thresholds here are based on empirical testing and may need to be adjusted based
+  // on the specific model and hardware characteristics.
+  // Although the re-ordering overhead for running the conv as NCHW exists, given the perf difference for large kernels, it is worth trying the option.
+  if (kernel_height >= 7 && kernel_width >= 7 && input_channels >= 64 && output_channels >= 32) {
```
tianleiwu, line 360: How do you get these thresholds? 

For example, I have kernel_height = 5, kernel_width = 5,  input_channels = 256, output_channels = 128. 
It could still be very expensive from cost estimation. Can we use cost estimation (FLOP) as threshold?

PR #27369, opened 2026-02-18
add option to disable generation contrib ops

fs-eire:
### Description
This pull request introduces a new build option to selectively disable generation-related contrib operators (such as BeamSearch, WhisperBeamSearch, GreedySearch, and Sampling) in ONNX Runtime.

The implementation of those operators contributes non-trivial binary size to the final artifact. This offers a way to reduce the binary size in some use scenarios.


tianleiwu:
Please fix build errors.

PR #27371, opened 2026-02-18
[wasm] add flags for WebGPU build to reduce binary size

fs-eire:
### Description

This change uses a combination of flags to reduce binary size for WebGPU wasm build:
- exclude a few data types: uint16/int16, float8, float4, float64, sparse tensor, string
- disable_ml_ops
- disable_generation_ops

### Motivation and Context

reduce binary size for ORT-web.

### DIFF
--- a/.github/actions/macos-ci-setup/action.yml
+++ b/.github/actions/macos-ci-setup/action.yml
@@ -8,7 +8,7 @@ inputs:
   python_version:
     required: false
     type: string
-    default: "3.11"
+    default: "3.14"
   node_version:
     required: false
     type: string
--- a/.github/workflows/linux_ci.yml
+++ b/.github/workflows/linux_ci.yml
@@ -68,6 +68,21 @@ jobs:
     secrets:
       GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+  build-linux-x64-release-py314:
+    name: Build Linux x64 Release (Python 3.14)
+    uses: ./.github/workflows/reusable_linux_build.yml
+    with:
+      pool_name: "onnxruntime-github-Ubuntu2204-AMD-CPU"
+      build_config: Release
+      architecture: x64
+      dockerfile_path: tools/ci_build/github/linux/docker/Dockerfile.manylinux2_28_cpu
+      docker_image_repo: onnxruntimecpubuildpythonx64
+      extra_build_flags: '--use_binskim_compliant_compile_flags --build_wheel --build_nuget --enable_transformers_tool_test --cmake_extra_defines onnxruntime_BUILD_BENCHMARKS=ON'
+      python_path_prefix: 'PATH=/opt/python/cp314-cp314/bin:$PATH' # $ needs escaping in single quotes
+      job_identifier: build-linux-x64-release-py314
+    secrets:
+      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
   orttraining-linux-ci-pipeline:
     name: Build Linux x64 Release with training
     uses: ./.github/workflows/reusable_linux_build.yml
@@ -109,7 +124,7 @@ jobs:
       dockerfile_path: tools/ci_build/github/linux/docker/inference/aarch64/python/cpu/Dockerfile
       docker_image_repo: onnxruntimecpubuildpythonaarch64
       extra_build_flags: '--use_binskim_compliant_compile_flags --build_wheel --cmake_extra_defines onnxruntime_BUILD_BENCHMARKS=ON'
-      python_path_prefix: 'PATH=/opt/python/cp310-cp310/bin:$PATH' # $ needs escaping in single quotes
+      python_path_prefix: 'PATH=/opt/python/cp314-cp314/bin:$PATH' # $ needs escaping in single quotes
       job_identifier: build-linux-arm64-release
     secrets:
       GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/mac.yml
+++ b/.github/workflows/mac.yml
@@ -16,7 +16,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  python_version: 3.11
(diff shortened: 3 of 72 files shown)


onnxruntime/core/providers/op_kernel_type_control_overrides.inc
```diff
@@ -42,6 +42,36 @@ namespace op_kernel_type_control {
     using types = ::onnxruntime::TypeList<__VA_ARGS__>;             \
   };
 
+//
+// The following section is the default overrides for a reduced-size WebAssembly build.
+//
+#if defined(__wasm__) && defined(ORT_WASM_REDUCE_DATATYPE)
```
skottmckay: Would it be cleaner to checkin a config file for a wasm build instead of altering the source?

cmake/CMakeLists.txt
```diff
@@ -209,6 +209,7 @@ option(onnxruntime_ENABLE_WEBASSEMBLY_DEBUG_INFO "Enable this option to turn on
 option(onnxruntime_ENABLE_WEBASSEMBLY_PROFILING "Enable this option to turn on WebAssembly profiling and preserve function names" OFF)
 option(onnxruntime_ENABLE_WEBASSEMBLY_OUTPUT_OPTIMIZED_MODEL "Enable this option to allow WebAssembly to output optimized model" OFF)
 option(onnxruntime_ENABLE_WEBASSEMBLY_RELAXED_SIMD "Enable WebAssembly Relaxed SIMD" OFF)
+option(onnxruntime_WEBASSEMBLY_REDUCE_DATATYPE "Enable reduced datatype macro in WebAssembly builds" OFF)
```
skottmckay: Maybe something more generic like 'reduce wasm build binary size' would be more flexible. For example disabling RTTI will also save you a good amount of size for virtually no loss of functionality so there may be a range of different settings this could alter.

fs-eire: we have multiple ways to reduce binary size and we apply them slightly differently for multiple builds(JSPI/Asyncify/CPU), and reducing the datatype registration is only one of them.

onnxruntime/core/providers/op_kernel_type_control_overrides.inc
fs-eire: sure, that makes sense. let me find out how to use the config file.

fs-eire:
Updated the change to use config file and `--enable_reduced_operator_type_support`.

Pending more local test, now switching to "WIP" and will switch back once it's complete.

cmake/CMakeLists.txt
fs-eire: anyway the options are removed.

onnxruntime/wasm/reduced_types.config
```diff
@@ -0,0 +1,14 @@
...
+# Usage: pass to build.py via:
+#   --include_ops_by_config onnxruntime/wasm/reduced_types.config --enable_reduced_operator_type_support
+#
+# Excluded types: float4, float8, float64 (double), int16, uint16
+
+# Keep all operators - only apply type reduction
+!no_ops_specified_means_all_ops_are_required
+
+# Globally allowed types for all operators
+!globally_allowed_types;bool,int8_t,uint8_t,int32_t,uint32_t,int64_t,uint64_t,float,MLFloat16
```
skottmckay, line 14: Are uint32_t or uint64_t needed? Don't recall much usage of those in real models.

.github/workflows/linux-wasm-ci-build-and-test-workflow.yml
```diff
@@ -50,6 +50,12 @@ jobs:
         --enable_wasm_simd
         ${{ inputs.enable_wasm_threads == true && '--enable_wasm_threads' || '' }}
         ${{ inputs.extra_build_args }}
+      reduced_size_build_args: >-
+        --disable_ml_ops
+        --disable_generation_ops
+        --disable_types string float4 float8 optional sparsetensor
+        --include_ops_by_config onnxruntime/wasm/reduced_types.config
+        --enable_reduced_operator_type_support
```
skottmckay, line 58: Do you need RTTI support or could that also be disabled?

fs-eire: --disable_rtti is already specified in the different places for release build

onnxruntime/wasm/reduced_types.config
fs-eire: I think it is OK to preserve these 2 types - these are very small binary size contribution.

xenova:
Very aligned with this fix, but we do get the following error for https://huggingface.co/onnx-community/BEN2-ONNX/tree/main/onnx.
```
ERROR_MESSAGE: /mnt/vss/_work/1/s/onnxruntime/core/optimizer/transformer_memcpy.cc:254 bool onnxruntime::TransformerMemcpyImpl::IsNodeCompatibleWithProvider(const onnxruntime::Node &) const !node_provider_type.empty() was false. Provider type for Pow node with name '/dec_blk3/norm1/Pow' is not set.
```

which uses float64 ops (first Pow).

This is VERY rare though... so probably not worth reverting/updating. We should update the model instead imo.

fs-eire:
> Very aligned with this fix, but we do get the following error for https://huggingface.co/onnx-community/BEN2-ONNX/tree/main/onnx.
> 
> ```
> ERROR_MESSAGE: /mnt/vss/_work/1/s/onnxruntime/core/optimizer/transformer_memcpy.cc:254 bool onnxruntime::TransformerMemcpyImpl::IsNodeCompatibleWithProvider(const onnxruntime::Node &) const !node_provider_type.empty() was false. Provider type for Pow node with name '/dec_blk3/norm1/Pow' is not set.
> ```
> 
> which uses float64 ops (first Pow).
> 
> This is VERY rare though... so probably not worth reverting/updating. We should update the model instead imo.

Yes usually fp32 works for most of the scenarios.

PR #27374, opened 2026-02-18
Enable Robust Symlink Support for External Data (HF Cache Support)

tianleiwu:
## Summary
This PR enables ONNX Runtime to correctly load models with external data when they are stored in symlinked directory structures. This is a common scenario for models cached by the Hugging Face Hub, where both the model file and data files are symlinks pointing into a flat `blobs/` directory.

## The Problem
Previously, ONNX Runtime's external data path validation would only check if the resolved data path was under the logical directory of the model. In symlinked structures (like Hugging Face's `snapshots/` and `blobs/` layout), the resolved data path often sits in a different physical directory than the logical model path, leading to a "path escapes model directory" error even when the data is safely associated with the model.

## The Fix
I have updated `ValidateExternalDataPath` to implement a dual-check mechanism:
1.  **Logical Check:** Verify if the resolved data path is under the provided `base_dir` (the directory where the model was loaded from).
2.  **Physical Check:** If the logical check fails, verify if the resolved data path is under the parent directory of the **real/canonical** model path.

This approach ensures that models can load external data from both their logical siblings and their physical siblings while maintaining security constraints.

## Changes
- **Core Logic:**
    - `onnxruntime/core/framework/tensorprotoutils.cc/h`: Updated `ValidateExternalDataPath` to accept `model_path` and perform the dual-check.
    - `onnxruntime/core/graph/graph.cc`: Updated call site in `ConvertInitializersIntoOrtValues` to pass the `model_path`.
- **Cleanup:**
    - Removed unused `ValidateExternalDataPath` from `provider_api.h`, `provider_interfaces.h`, and `provider_bridge_ort.cc`. These were redundant as path validation is handled by core during session initialization. Provider has no need to validate it.
- **Testing:**
    - Added `onnxruntime/test/python/onnxruntime_test_python_symlink_data.py` to simulate the Hugging Face Hub symlink structure and verify the fix.

### DIFF
--- a/tools/ci_build/build.py
+++ b/tools/ci_build/build.py
@@ -1769,9 +1769,6 @@ def run_onnxruntime_tests(args, source_dir, ctest_path, build_dir, configs):
                 [sys.executable, "onnxruntime_test_python.py"], cwd=cwd, dll_path=dll_path, python_path=python_path
             )
 
-            log.info("Testing Symlink Data feature")
-            run_subprocess([sys.executable, "onnxruntime_test_python_symlink_data.py"], cwd=cwd, dll_path=dll_path)
-
             log.info("Testing Global Thread Pool feature")
             run_subprocess([sys.executable, "onnxruntime_test_python_global_threadpool.py"], cwd=cwd, dll_path=dll_path)
 
@@ -1819,6 +1816,9 @@ def run_onnxruntime_tests(args, source_dir, ctest_path, build_dir, configs):
                 onnx_test = False
 
             if onnx_test:
+                log.info("Testing Symlink ONNX Model and External Data")
+                run_subprocess([sys.executable, "onnxruntime_test_python_symlink_data.py"], cwd=cwd, dll_path=dll_path)
+
                 # Disable python onnx tests for TensorRT and CANN EP, because many tests are
                 # not supported yet.
                 if args.use_tensorrt or args.use_cann:


onnxruntime/test/python/onnxruntime_test_python_symlink_data.py
```diff
@@ -0,0 +1,250 @@
...
+        # temp_dir/
+        #   model/
+        #     guid1
+        #   data/
+        #     guid2
+        #   snapshots/version/
+        #       model.onnx -> ../../model/guid1
+        #       data.bin   -> ../../data/guid2
+
+        self.temp_dir = tempfile.mkdtemp()
```
tianleiwu, line 182: This file can be refactored to avoid duplicated code. Keep it for now to meet release schedule. I can create another PR for refactoring after release.

onnxruntime/core/framework/tensorprotoutils.cc
```diff
@@ -329,22 +329,48 @@ Status TensorProtoWithExternalDataToTensorProto(
...
         resolved.begin(), resolved.end());
-    ORT_RETURN_IF(base_end != base_canonical.end(),
-                  "External data path: ", location, " escapes model directory: ", base_dir);
+
+    if (base_end != base_canonical.end()) {
+      // If validation against logical base_dir fails, we check against the
+      // real (canonical) path of the model file to support symlinked models
+      // (e.g. models in Hugging Face Hub local cache).
+      if (!model_path.empty()) {
+        auto real_model_dir = std::filesystem::weakly_canonical(model_path).parent_path();
```
edgchen1, line 354: why do we use `weakly_canonical` instead of `canonical`? is there a valid case where the full path may not actually exist?

tianleiwu: The purpose is to be consistent with existing code.

edgchen1: can clean it up later.

edgchen1 (approved):

PR #27376, opened 2026-02-18
chore: increase parallelism for win CI

sanaa-hamel-microsoft:
### Description

Remove parallelism limits for C++ translation. Primarily affects Win CI jobs.

### Motivation and Context

Win CI builds were limited to 4 cores. This is a complete waste of time when we're using 32 core machines.
This saves ~2.25h for a windows build. See: https://aiinfra.visualstudio.com/Lotus/_build/results?buildId=1089604

### DIFF
--- a/tools/ci_build/build.py
+++ b/tools/ci_build/build.py
@@ -1350,15 +1350,9 @@ def build_targets(args, cmake_path, build_dir, configs, num_parallel_jobs, targe
         build_tool_args = []
         if num_parallel_jobs != 0:
             if is_windows() and args.cmake_generator != "Ninja" and not args.build_wasm:
-                # https://github.com/Microsoft/checkedc-clang/wiki/Parallel-builds-of-clang-on-Windows suggests
-                # not maxing out CL_MPCount
-                # Start by having one less than num_parallel_jobs (default is num logical cores),
-                # limited to a range of 1..15
-                # that gives maxcpucount projects building using up to 15 cl.exe instances each
                 build_tool_args += [
                     f"/maxcpucount:{num_parallel_jobs}",
-                    # one less than num_parallel_jobs, at least 1, up to 15
-                    f"/p:CL_MPCount={min(max(num_parallel_jobs - 1, 1), 15)}",
+                    f"/p:CL_MPCount={num_parallel_jobs}",
                     # if nodeReuse is true, msbuild processes will stay around for a bit after the build completes
                     "/nodeReuse:False",
                 ]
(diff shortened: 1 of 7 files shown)


tools/ci_build/build.py
```diff
@@ -1350,15 +1350,9 @@ def build_targets(args, cmake_path, build_dir, configs, num_parallel_jobs, targe
         build_tool_args = []
         if num_parallel_jobs != 0:
             if is_windows() and args.cmake_generator != "Ninja" and not args.build_wasm:
-                # https://github.com/Microsoft/checkedc-clang/wiki/Parallel-builds-of-clang-on-Windows suggests
```
eserscor, line 1353: Is this still relevant? Should we pass in nproc - 1 or something from CICD?

tools/ci_build/github/azure-pipelines/stages/nuget-win-cuda-packaging-stage.yml
```diff
@@ -93,9 +93,9 @@ stages:
     CudaVersion: ${{ parameters.CudaVersion }}
     packageName: x64-tensorrt
     ${{ if ne(parameters.win_cudnn_home, '') }}:
-      buildparameter: --use_tensorrt --tensorrt_home=${{ parameters.win_trt_home }} --cuda_home=${{ parameters.win_cuda_home }} --cudnn_home=${{ parameters.win_cudnn_home }} --enable_onnx_tests --enable_wcos --parallel 4 --nvcc_threads 1 --cmake_extra_defines "CMAKE_CUDA_ARCHITECTURES=${{ parameters.CudaArchs }}"
+      buildparameter: --use_tensorrt --tensorrt_home=${{ parameters.win_trt_home }} --cuda_home=${{ parameters.win_cuda_home }} --enable_onnx_tests --enable_wcos --nvcc_threads 1 --cmake_extra_defines "CMAKE_CUDA_ARCHITECTURES=${{ parameters.CudaArchs }}" --cudnn_home=${{ parameters.win_cudnn_home }}
```
tianleiwu, line 96: add --parallel. Also double check other places in this file.

sanaa-hamel-microsoft: `win-ci.yml` automatically provides parallel. TBH there is no sane reason not to default to parallel-full-cores so even that shouldn't be required. Specifying `--parallel` here is what caused the issue in the first place.

eserscor (approved):

tools/ci_build/build.py
sanaa-hamel-microsoft: If we do need to limit the # of cores then it should be done by CICD rather than second guess the user.
I believe it was cargo-culted from the clang repo linked and it doesn't seem to have caused issues with CI.
I'd say leave it out until we see evidence we need it, and then add it to the pipeline.

PR #27391, opened 2026-02-19
Fix O(n²) model load time for TreeEnsemble with categorical feature chains

vektah:
### Description
Profiling shows that CheckIfSubtreesAreEqual is invoked recursively for many node pairs for LightGBM models with categorical features. A significant portion of this work consists of self-comparisons (left_id == right_id), leading to effectively O(n²) comparing trees to themselves during model loading.

This change adds a fast-path for trivial equality, avoiding unnecessary recursive comparisons.

Example results:
 - model with 7K BRANCH_EQ nodes: 527 ms → 47 ms (~11× faster)
 - model with 106K BRANCH_EQ nodes: 141 s → 80 ms (~1760× faster)

### Motivation and Context
We have some LightGBM exported models that make heavy use of categorical features and exhibit extremely slow load times (minutes for a single 2.5mb model).

Heres a diagram to illustrate the issue:
image

the 106K model has much longer "member of" chains, with chains that lead into more chains:

<details>
  <summary>"trees"</summary>
  
image

  
</details>

Interestingly we did also try using the new onnx.ml opset 5 node that has MEMBER, but it seems even slower as it recreates these branch EQ chains.

### DIFF
--- a/onnxruntime/core/providers/cpu/ml/tree_ensemble_common.h
+++ b/onnxruntime/core/providers/cpu/ml/tree_ensemble_common.h
@@ -384,6 +384,9 @@ bool TreeEnsembleCommon<InputType, ThresholdType, OutputType>::CheckIfSubtreesAr
     gsl::span<const ThresholdType> nodes_values_as_tensor, gsl::span<const float> node_values,
     gsl::span<const float> target_class_weights, gsl::span<const ThresholdType> target_class_weights_as_tensor,
     const InlinedVector<TreeNodeElementId>& node_tree_ids, InlinedVector<std::pair<TreeNodeElementId, uint32_t>> indices) {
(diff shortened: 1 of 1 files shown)


vektah:
@xadupre could you take a look 🙏🏻

cbourjau:
> Interestingly we did also try using the new onnx.ml opset 5 node that has MEMBER, but it seems even slower as it recreates these branch EQ chains.

Thanks so much for looking into this! IIRC, the current v5 reuses large parts of the v3 implementation. I think one could make a good case for eventually rewriting this operator from scratch at some point 🙈.

cbourjau:
Can this bei merged and possibly be included in 1.24.3 via https://github.com/microsoft/onnxruntime/pull/27501 ?

PR #27398, opened 2026-02-20
[build] Fix building with torch nightly

titaiwangms:
It was blocking ort build with torch-nightly


justinchuby (approved):

justinchuby:
What was the error?

titaiwangms:
> What was the error?

nn is not a module of torch

PR #27411, opened 2026-02-21
[web] fix usage of wasmBinary together with a blob URL for .mjs

fs-eire:
### Description

Fixes the issue 1 described in #27317


fs-eire:
@0-don

0-don:
thank you

guschmue (approved):

PR #27423, opened 2026-02-23
Fix build issues in winml_adapter

jnagi-intel:
Fixes the build when --use_winml --enable_wcos flags are added

### DIFF
--- a/winml/adapter/winml_adapter_model.cpp
+++ b/winml/adapter/winml_adapter_model.cpp
@@ -611,7 +611,7 @@ ORT_API_STATUS_IMPL(
   input.set_name(input_name);
 
   if (info->type == ONNXType::ONNX_TYPE_TENSOR) {
-    auto num_dims = info->tensor_type_info->GetShape()->GetDims().size();
+    auto num_dims = info->tensor_type_info->HasShape() ? info->tensor_type_info->GetShape()->GetDims().size() : 0;
     CreateTypeProto_Tensor(
       input.mutable_type()->mutable_tensor_type(),
       input_name,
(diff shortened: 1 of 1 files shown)


winml/adapter/winml_adapter_model.cpp
```diff
@@ -611,13 +611,13 @@ ORT_API_STATUS_IMPL(
   input.set_name(input_name);
 
   if (info->type == ONNXType::ONNX_TYPE_TENSOR) {
-    auto num_dims = info->tensor_type_info->shape.NumDimensions();
+    auto num_dims = info->tensor_type_info->GetShape()->GetDims().size();
     CreateTypeProto_Tensor(
       input.mutable_type()->mutable_tensor_type(),
       input_name,
-      (num_dims == 0) ? nullptr : &info->tensor_type_info->shape[0],
+      (num_dims == 0) ? nullptr : info->tensor_type_info->GetShape()->GetDims().data(),
```
adrianlizarraga, line 618: `OrtTensorTypeAndShapeInfo::GetShape()` may return `nullptr` [if there is no shape](https://github.com/microsoft/onnxruntime/blob/f501e1d0c7ed4c454bb5b3c3edee791068ec3472/onnxruntime/core/framework/tensor_type_and_shape.h#L43). 

Perhaps this API function should return an error status if `info->tensor_type_info->HasShape()` returns false.

```c++
if (info->type == ONNXType::ONNX_TYPE_TENSOR) {
  if (!info->tensor_type_info->HasShape()) {
    return OrtApis::CreateStatus(ORT_INVALID_ARGUMENT, "Model input tensor must have a shape");
  }
}
```

jnagi-intel: I have added HasShape check on initializing num_dims which should avoid calls to GetShape() if there is no shape.

winml/adapter/winml_adapter_model.cpp
```diff
@@ -611,13 +611,13 @@ ORT_API_STATUS_IMPL(
   input.set_name(input_name);
 
   if (info->type == ONNXType::ONNX_TYPE_TENSOR) {
-    auto num_dims = info->tensor_type_info->shape.NumDimensions();
+    auto num_dims = info->tensor_type_info->HasShape() ? info->tensor_type_info->GetShape()->GetDims().size() : 0;
```
adrianlizarraga, line 614: Looks like this doesn't properly differentiate between the following two cases:
- `HasShape()` is true but `GetDims().size()` is 0: This would occur for a scalar tensor.
- `HasShape()` is false. The shape info is missing completely. This is probably an error when specifying a model input.

That said, the original winml/adapter code seems to have never made this distinction, so probably doesn't need to be fixed here...

adrianlizarraga (approved):

Issue #26479, opened 2025-11-03
[Build] build onnxruntime for android with nnapi and qnn error


WangHHY19931001:
### Describe the issue

build a83fc4d58cb48eb68890dd689f94f28288cf2278 for android arm64 with nnapi and qnn ep error

### Target platform

android arm64

### Build script

.\build.bat --config Release --android --android_sdk_path "C:\Users\dingli\AppData\Local\Android\Sdk" --android_ndk_path "C:\Users\dingli\AppData\Local\Android\Sdk\ndk\27.3.13750724" --android_abi arm64-v8a --android_api 29 --cmake_generator Ninja --use_nnapi --skip_tests --cmake_extra_defines "CMAKE_CXX_FLAGS=-flto -funroll-loops -march=armv8.2-a+fp16" "CMAKE_C_FLAGS=-flto -funroll-loops -march=armv8.2-a+fp16" "CMAKE_INTERPROCEDURAL_OPTIMIZATION=ON"  --build_shared_lib --use_qnn static_lib --qnn_home "C:\Qualcomm\AIStack\QAIRT\2.31.0.250130" --parallel

### Error / output

Synchronizing submodule url for 'cmake/external/emsdk'
Synchronizing submodule url for 'cmake/external/libprotobuf-mutator'
Synchronizing submodule url for 'cmake/external/onnx'
Synchronizing submodule url for 'cmake/external/onnx/third_party/pybind11'
-- Loading Dependencies URLs ...
-- Loading Dependencies ...
-- Abseil source dir:F:/onnxruntime/build/Windows/Release/_deps/abseil_cpp-src
-- CMAKE_HOST_SYSTEM_NAME: Windows
-- Use prebuilt protoc
-- 
-- 3.21.12.0
-- Using the multi-header code from F:/onnxruntime/build/Windows/Release/_deps/nlohmann_json-src/include/
-- Proceeding with version: 23.5.26.0
-- CMAKE_CXX_FLAGS: -flto -funroll-loops -march=armv8.2-a+fp16 -ffunction-sections -fdata-sections -DCPUINFO_SUPPORTED
-- Using custom protoc executable
-- ONNX_PROTOC_EXECUTABLE: F:/onnxruntime/build/Windows/Release/_deps/protoc_binary-src/bin/protoc.exe
-- Protobuf_VERSION: 
-- 
-- ******** Summary ********
--   CMake version                     : 3.28.6
[...]

Issue #25501, opened 2025-07-23
[Build] build android aar with nnapi and qnn error


WangHHY19931001:
### Describe the issue

build error return not 0, due to qnn build error

build from git branch v1.22.1

### Target platform

android arm64

### Build script

.\build.bat --config RelWithDebInfo --android --android_sdk_path "C:\Users\dingli\AppData\Local\Android\Sdk" --android_ndk_path "C:\Users\dingli\AppData\Local\Android\Sdk\ndk\27.2.12479018" --android_abi arm64-v8a --android_api 29 --cmake_generator Ninja --use_nnapi --build_shared_lib --skip_tests --use_qnn static_lib --qnn_home "C:\Qualcomm\AIStack\QAIRT\2.31.0.250130" --build_java

### Error / output

[cmd_out.txt](https://github.com/user-attachments/files/21378933/cmd_out.txt)

PR #27440, opened 2026-02-24
Enhance QNN EP static build to reuse NodeAttrHelper from shared utils

XDcobra:
### Description
Fixes linker duplicate-symbol errors when building the Android AAR with both NNAPI and QNN enabled. The QNN provider defined its own NodeAttrHelper in `ort_api.cc`, which duplicated the implementation already built into the NNAPI provider (`utils.cc`). The linker then reported multiple definitions of `NodeAttrHelper` (constructors and all Get*/HasAttr methods).

Result: Android builds with --use_nnapi and --use_qnn static_lib link successfully and produce a single shared library.

### Motivation and Context
Building for Android with both NNAPI and QNN (e.g. for NNAPI fallback and QNN acceleration) currently fails at link time with many duplicate symbol: onnxruntime::NodeAttrHelper::... errors (ld.lld), so this configuration is unusable.
* Issues addressed:
* Fixes microsoft/onnxruntime#25501 (build android aar with nnapi and qnn error).
* Fixes microsoft/onnxruntime#26479 (build onnxruntime for android with nnapi and qnn error – includes the full duplicate-symbol log).


XDcobra:

PR #27442, opened 2026-02-24
Use `_tpause` also for clang

JanSellner:
### Description
[`_tpause` is also supported](https://clang.llvm.org/doxygen/waitpkgintrin_8h.html#a371e36b9996fa91a80c52c182ae6f0d4) when using the clang compiler so that it can be used in this location in the code. The alternative code path which uses `__builtin_ia32_tpause` lets it fail when using the clang compiler.

### Motivation and Context
I read somewhere that you are only supporting gcc but besides that change I was able to compile with clang so maybe you are interested in this simple fix for widening compiler support :-)


tianleiwu:
see https://github.com/microsoft/onnxruntime/pull/27607

PR #27448, opened 2026-02-25
webgpu: Split `MatMulWriteFnSource()` into smaller functions

Jiawei-Shao:
### Description
This patch splits `MatMulWriteFnSource()` into three smaller functions with fewer parameters and simpler logic as a preparation to implement `Split-K` on the `MatMul` operator.

### DIFF
--- a/onnxruntime/core/providers/webgpu/vendor/intel/math/gemm.cc
+++ b/onnxruntime/core/providers/webgpu/vendor/intel/math/gemm.cc
@@ -90,8 +90,7 @@ Status ApplyGemmIntel(const Tensor* a,
   const uint32_t dispatch_y = narrow<uint32_t>((M + kSubgroupLogicalWorkGroupSizeY * elements_per_thread[1] - 1) /
                                                (kSubgroupLogicalWorkGroupSizeY * elements_per_thread[1]));
 
-  GemmSubgroupProgram program{transA, transB, alpha, need_handle_bias, need_handle_matmul, c_components, c_is_scalar,
-                              is_vec4, elements_per_thread};
+  GemmSubgroupProgram program{transA, transB, alpha, need_handle_bias, need_handle_matmul, c_is_scalar, is_vec4, elements_per_thread};
 
   if (need_handle_matmul) {
     program.AddInputs({{a, ProgramTensorMetadataDependency::TypeAndRank, a_components},
--- a/onnxruntime/core/providers/webgpu/vendor/intel/math/gemm.h
+++ b/onnxruntime/core/providers/webgpu/vendor/intel/math/gemm.h
@@ -14,15 +14,13 @@ namespace intel {
 class GemmSubgroupProgram final : public Program<GemmSubgroupProgram> {
  public:
   GemmSubgroupProgram(bool transA, bool transB, float alpha, bool need_handle_bias, bool need_handle_matmul,
-                      int c_components, bool c_is_scalar, bool is_vec4,
-                      const gsl::span<int64_t>& elements_per_thread)
+                      bool c_is_scalar, bool is_vec4, const gsl::span<int64_t>& elements_per_thread)
       : Program{"GemmSubgroup"},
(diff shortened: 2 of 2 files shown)


Jiawei-Shao:
@qjia7  PTAL, thanks!

onnxruntime/core/providers/webgpu/math/gemm_utils.cc
```diff
@@ -173,45 +190,33 @@ void MatMulReadFnSource(ShaderHelper& shader,
...
+  HandleMaybeHaveBiasForGEMM(shader, output, bias, c_components, output.NumComponents(), c_is_scalar);
+  EmitMatMulWriteFnFooter(shader);
+}
 
-  shader.AdditionalImplementation()
-      << "  }\n"
-      << "}\n\n";
+void MatMulWriteFnSourceWithSplitK(ShaderHelper& shader,
+                                   const ShaderVariableHelper& output,
+                                   bool is_gemm,
```
qjia7, line 213: If `is_gemm` is only used to distinguish whether coords is vec2 or vec3, is it better to make gemm coords as vec3 (0, row, col). It means reshape [M, N] to [1, M, N] for gemm?

onnxruntime/core/providers/webgpu/math/gemm_utils.cc
```diff
@@ -173,45 +190,33 @@ void MatMulReadFnSource(ShaderHelper& shader,
...
-    HandleMatMulWithSplitK(shader, is_gemm, output, output_variable_type);
-  } else if (is_gemm) {
-    HandleMaybeHaveBiasForGEMM(shader, output, bias, c_components, output_components, c_is_scalar);
-  } else {
-    HandleMaybeBiasForMatMul(shader, output, bias, activation_snippet, is_channels_last);
-  }
+void MatMulWriteFnSourceForGemm(ShaderHelper& shader,
+                                const ShaderVariableHelper& output,
+                                const ShaderVariableHelper* bias,
+                                int c_components,
```
qjia7: nit: `c_components` can be obtained from `bias.NumComponents()`.

Jiawei-Shao: Oh I cannot do that because when `c_last_dim == N && output_components == 4` we will always set `c_components` to `4` whatever the real components of `bias` is:

https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/providers/webgpu/math/gemm_packed.cc#L97

qjia7: Still confusing: If bias is added as `program.AddInput({c, ProgramTensorMetadataDependency::TypeAndRank, c_components});` in api side, bias.NumComponents() must be c_components.

And for your case if c_last_dim == N && output_components == 4 is true, it means c_components is 4. And you will get bias.NumComponents() as 4. Are you getting the wrong number?

Jiawei-Shao: Sorry that I misunderstood your comment. Now I realize how to do it properly.

onnxruntime/core/providers/webgpu/math/gemm_utils.cc
Jiawei-Shao: I just tried to use `vec3` in `gemm` and realized we need more changes in many other places. Can we do this change in another follow-up PR? I'd prefer to only refactor `MatMulWriteFnSource()` in this PR.

onnxruntime/core/providers/webgpu/math/gemm_utils.cc
Jiawei-Shao: Done

Jiawei-Shao:
Thanks @qjia7.

Hi @guschmue @fs-eire, PTAL, thanks!

Jiawei-Shao:
I've fixed the build failure. PTAL, thanks!

PR #27454, opened 2026-02-25
Suppress spurious Array Out of Bounds warnings produced by GCC 14.2 compiler on Linux builds

vthaniel:
### Description
Suppress spurious Array Out of Bounds warnings produced by GCC 14.2 compiler on Linux builds

### Motivation and Context
Linux build fails when compiled with GCC 14.2 due to spurious Array Out of Bounds warnings (Warnings Treated as Errors)

### DIFF
--- a/include/onnxruntime/core/framework/int2.h
+++ b/include/onnxruntime/core/framework/int2.h
@@ -140,14 +140,14 @@ struct Int2x4Base {
     // Process complete groups of 4 elements
 
     for (; dst_i < full_quads; dst_i++) {
-      #if defined (__GNUC__) && defined(HAS_ARRAY_BOUNDS)
-      #pragma GCC diagnostic push
-      #pragma GCC diagnostic ignored "-Warray-bounds"
-      #endif
+#if defined(__GNUC__) && defined(HAS_ARRAY_BOUNDS)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Warray-bounds"
+#endif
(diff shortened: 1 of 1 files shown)


include/onnxruntime/core/framework/int2.h
```diff
@@ -138,7 +138,10 @@ struct Int2x4Base {
 
     // Process complete groups of 4 elements
     for (; dst_i < full_quads; dst_i++) {
+      #pragma GCC diagnostic push
```
edgchen1: please also check for `defined(__GNUC__)` and `defined(HAS_ARRAY_BOUNDS)` so that it works with other compilers where that warning may not be available.

for example:
https://github.com/microsoft/onnxruntime/blob/f3a628458238047974eac7dd2114c7bb34326841/include/onnxruntime/core/framework/tensor_shape.h#L18-L23

`HAS_ARRAY_BOUNDS` would be defined in "onnxruntime_config.h".
https://github.com/microsoft/onnxruntime/blob/f3a628458238047974eac7dd2114c7bb34326841/cmake/CMakeLists.txt#L613
https://github.com/microsoft/onnxruntime/blob/f3a628458238047974eac7dd2114c7bb34326841/cmake/onnxruntime_config.h.in#L6

xadupre (dismissed):

include/onnxruntime/core/framework/int2.h
vthaniel: Warning suppression directive is now moved under
 #if defined (__GNUC__) && defined(HAS_ARRAY_BOUNDS)

Issue #27286, opened 2026-02-09
[Feature Request] WSL device discovery


theHamsta:
### Describe the feature request

At the moment, GPU discovery on Linux worked by iterating over DRM nodes.
This approach won't work on WSL since the DRM kernel side isn't present.
There are different approaches to resolve this

- iterate over PCI devices for WSL instead of DRM devices. PCI devices with dxcore driver are paravirtualized GPUs
      - 👍 super-fast
      - 👍 mostly matches current Linux logic, could unify with Linux logic
      - 👎 can't get much information from paravirtualized sys fs
      - 👎 unclear how to get info to match virtual PCI bus ID with real CUDA device for CUDA EP / TRT NV EP
 -  use dxcore library, dynamically load it when `/dev/dxgi` is present
      - 👍 dxcore API gives rich info on devices (vendor, device type, VRAM, etc)
      - 👍 works cross-vendor
      - 👎 getting adapters seems to take longer than just iterating over sysfs files
      - 👎 to match with CUDA device, LUID doesn't work since it is the LUID of the paravirtualized device. One can get the real PCI bus id via KMT2/KMT3 API which allows matching with CUDA devices.
-  ioctls on `/dev/dxgi`
     - same tradeoffs as with dxcore lib, but we wouldn't need to dynload a library if we only need very few info
- nvml
     - 👍 solution much shorter in terms of lines of code than dxgi
     - 👍 nvml is available on Linux, in containers, VMs and WSL whenever a Nvidia driver is installed. It gives a strong indication that a Nvidia device is present
     - 👍 get real PCI bus id directly 
     - 👎 `nvmlInit` is slower than iterating sysfs

I have a draft with the nvml solution and parts of dxcore outcommented (I have a working dxcore solution as a standalone project) here https://github.com/microsoft/onnxruntime/pull/27287

@edgchen1 @gedoensmax 

### Describe scenario use case

Device discovery in WSL should behave like regular Linux, so WSL could serve as a developer platform for desktop and cloud Linux platforms

edgchen1:
thanks for looking into this. I don't have enough familiarity with the options that you have proposed to have much of a preference at this point.

regarding speed, I think this is not as important unless something is egregiously slow, as this is a one-time initialization cost.

perhaps we can start with the simplest (in terms of code changes, new dependencies) approach that gives us all the information we currently need?

PR #27456, opened 2026-02-25
[CORE][NVEP]: add support for Vulkan interop

theHamsta:
### Description

The Vulkan interop works in a similar way as D3D12 interop.

The shared handles from Vulkan work for CUDA the same way as D3D12 handles. For Linux, we can use file descriptors.

As a sync primitive we use Vulkan timeline semaphores. They are widely supported since Vulkan 1.2 and work in a similar way as the existing `ID3D12Fence`s.

### Motivation and Context

This change allows to use graphics interop also on Vulkan and on Linux. It addresses a TODO in the external memory API.

### DIFF
--- a/include/onnxruntime/core/session/onnxruntime_c_api.h
+++ b/include/onnxruntime/core/session/onnxruntime_c_api.h
@@ -972,15 +972,15 @@ typedef void (*RunAsyncCallbackFn)(void* user_data, OrtValue** outputs, size_t n
 
 /** \brief External memory handle type for importing GPU resources.
  *
- * \todo Add OPAQUE_WIN32 for Windows Vulkan-specific memory handles
- * \todo Add POSIX file descriptor (OPAQUE_FD) for Linux Vulkan/CUDA/OpenCL interop
  * \todo Add Linux DMA-BUF file descriptor for embedded GPU memory sharing
  *
  * \since Version 1.24.
  */
 typedef enum OrtExternalMemoryHandleType {
-  ORT_EXTERNAL_MEMORY_HANDLE_TYPE_D3D12_RESOURCE = 0, /**< Shared HANDLE from ID3D12Device::CreateSharedHandle(resource) */
-  ORT_EXTERNAL_MEMORY_HANDLE_TYPE_D3D12_HEAP = 1,     /**< Shared HANDLE from ID3D12Device::CreateSharedHandle(heap) */
+  ORT_EXTERNAL_MEMORY_HANDLE_TYPE_D3D12_RESOURCE = 0,      /**< Shared HANDLE from ID3D12Device::CreateSharedHandle(resource) */
+  ORT_EXTERNAL_MEMORY_HANDLE_TYPE_D3D12_HEAP = 1,          /**< Shared HANDLE from ID3D12Device::CreateSharedHandle(heap) */
+  ORT_EXTERNAL_MEMORY_HANDLE_TYPE_VK_MEMORY_WIN32 = 2,     /**< Shared HANDLE from vkGetMemoryWin32HandleKHR, non-dedicated allocation */
+  ORT_EXTERNAL_MEMORY_HANDLE_TYPE_VK_MEMORY_OPAQUE_FD = 3, /**< File descriptor from vkGetMemoryOpaqueFdKHR, non-dedicated allocation */
 } OrtExternalMemoryHandleType;
 
 /** \brief Descriptor for importing external memory.
@@ -1004,7 +1004,9 @@ typedef struct OrtExternalMemoryDescriptor {
  * \since Version 1.24.
  */
 typedef enum OrtExternalSemaphoreType {
-  ORT_EXTERNAL_SEMAPHORE_D3D12_FENCE = 0, /**< Shared HANDLE from ID3D12Device::CreateSharedHandle(fence) */
+  ORT_EXTERNAL_SEMAPHORE_D3D12_FENCE = 0,                     /**< Shared HANDLE from ID3D12Device::CreateSharedHandle(fence) */
+  ORT_EXTERNAL_SEMAPHORE_VK_TIMELINE_SEMAPHORE_WIN32 = 1,     /**< Shared HANDLE from vkGetSemaphoreWin32HandleKHR of a VkSemaphore created as VK_SEMAPHORE_TYPE_TIMELINE */
+  ORT_EXTERNAL_SEMAPHORE_VK_TIMELINE_SEMAPHORE_OPAQUE_FD = 2, /**< File descriptor from vkGetSemaphoreFdKHR of a VkSemaphore created as VK_SEMAPHORE_TYPE_TIMELINE */
 } OrtExternalSemaphoreType;
 
 /** \brief Descriptor for importing external semaphores.
@@ -1071,14 +1073,16 @@ typedef struct OrtGraphicsInteropConfig {
    * works; streams use the default context.
    *
    * For D3D12: ID3D12CommandQueue*
-   * For Vulkan: VkQueue (cast to void*)
+   * For Vulkan: pass NULL
    */
   void* command_queue;
 
   /** \brief Additional API-specific options (optional).
    *
    * Can be used for future extensibility without changing the struct layout.
-   * For example, Vulkan-specific queue family index, or D3D12 fence sharing flags.
+   * For example, D3D12 fence sharing flags or provider-specific options like
+   * onnxruntime::nv::provider_option_names::kExternalComputeQueueDataParamNV_data
+   * for Vulkan interop for the NvTensorRTRTX provider.
    */
   const OrtKeyValuePairs* additional_options;
 } OrtGraphicsInteropConfig;
--- a/onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc
+++ b/onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc
@@ -9,6 +9,7 @@
 #include <cstdlib>
 #include <mutex>
 #include <unordered_map>
+#include <cuda.h>
 
 #include "core/providers/shared_library/provider_api.h"
 #include "core/session/onnxruntime_c_api.h"
@@ -539,8 +540,6 @@ struct NvTrtRtxSyncStreamImpl : OrtSyncStreamImpl {
   const OrtApi& ort_api;
 };
 
-#if defined(_WIN32)
-
 // External Resource Import Implementation (D3D12 to CUDA)
 /**
  * @brief Derived handle for imported external memory from D3D12 to CUDA.
@@ -642,8 +641,13 @@ struct NvTrtRtxExternalResourceImporterImpl : OrtExternalResourceImporterImpl {
       _In_ OrtExternalMemoryHandleType handle_type) noexcept {
     (void)this_ptr;
     // CUDA supports both D3D12 resource and heap handles
+#if defined(_WIN32)
     return handle_type == ORT_EXTERNAL_MEMORY_HANDLE_TYPE_D3D12_RESOURCE ||
-           handle_type == ORT_EXTERNAL_MEMORY_HANDLE_TYPE_D3D12_HEAP;
+           handle_type == ORT_EXTERNAL_MEMORY_HANDLE_TYPE_D3D12_HEAP ||
+           handle_type == ORT_EXTERNAL_MEMORY_HANDLE_TYPE_VK_MEMORY_WIN32;
+#elif __linux__
+    return handle_type == ORT_EXTERNAL_MEMORY_HANDLE_TYPE_VK_MEMORY_OPAQUE_FD;
+#endif
   }
 
   static OrtStatus* ORT_API_CALL ImportMemoryImpl(
@@ -694,6 +698,14 @@ struct NvTrtRtxExternalResourceImporterImpl : OrtExternalResourceImporterImpl {
         cu_handle_type = CU_EXTERNAL_MEMORY_HANDLE_TYPE_D3D12_HEAP;
         is_dedicated = false;  // D3D12 heaps are not dedicated
         break;
+      case ORT_EXTERNAL_MEMORY_HANDLE_TYPE_VK_MEMORY_WIN32:
+        cu_handle_type = CU_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_WIN32;
+        is_dedicated = false;  // API header currently, documents that this handle currently is non-dedicated
+        break;
+      case ORT_EXTERNAL_MEMORY_HANDLE_TYPE_VK_MEMORY_OPAQUE_FD:
+        cu_handle_type = CU_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_FD;
+        is_dedicated = false;  // API header currently, documents that this handle currently is non-dedicated
+        break;
       default:
         // Should not reach here - CanImportMemory already validated handle type
         return impl.ort_api.CreateStatus(ORT_EP_FAIL, "Unexpected external memory handle type");
@@ -702,7 +714,11 @@ struct NvTrtRtxExternalResourceImporterImpl : OrtExternalResourceImporterImpl {
     // Setup external memory handle descriptor
     CUDA_EXTERNAL_MEMORY_HANDLE_DESC ext_mem_desc = {};
     ext_mem_desc.type = cu_handle_type;
+#if defined(_WIN32)
     ext_mem_desc.handle.win32.handle = desc->native_handle;
+#else
+    ext_mem_desc.handle.fd = static_cast<int>(reinterpret_cast<intptr_t>((desc->native_handle)));
+#endif
     ext_mem_desc.size = desc->size_bytes;
     ext_mem_desc.flags = is_dedicated ? CUDA_EXTERNAL_MEMORY_DEDICATED : 0;
 
@@ -825,8 +841,11 @@ struct NvTrtRtxExternalResourceImporterImpl : OrtExternalResourceImporterImpl {
       _In_ const OrtExternalResourceImporterImpl* this_ptr,
       _In_ OrtExternalSemaphoreType type) noexcept {
     (void)this_ptr;
-    // CUDA supports D3D12 timeline fences
-    return type == ORT_EXTERNAL_SEMAPHORE_D3D12_FENCE;
+#if defined(_WIN32)
+    return type == ORT_EXTERNAL_SEMAPHORE_D3D12_FENCE || type == ORT_EXTERNAL_SEMAPHORE_VK_TIMELINE_SEMAPHORE_WIN32;
+#else
+    return type == ORT_EXTERNAL_SEMAPHORE_VK_TIMELINE_SEMAPHORE_OPAQUE_FD;
+#endif
   }
 
   static OrtStatus* ORT_API_CALL ImportSemaphoreImpl(
@@ -857,8 +876,25 @@ struct NvTrtRtxExternalResourceImporterImpl : OrtExternalResourceImporterImpl {
 
     // Setup external semaphore handle descriptor for D3D12 fence
     CUDA_EXTERNAL_SEMAPHORE_HANDLE_DESC ext_sem_desc = {};
-    ext_sem_desc.type = CU_EXTERNAL_SEMAPHORE_HANDLE_TYPE_D3D12_FENCE;
+    switch (desc->type) {
+      case ORT_EXTERNAL_SEMAPHORE_D3D12_FENCE:
+        ext_sem_desc.type = CU_EXTERNAL_SEMAPHORE_HANDLE_TYPE_D3D12_FENCE;
+        break;
+      case ORT_EXTERNAL_SEMAPHORE_VK_TIMELINE_SEMAPHORE_WIN32:
+        ext_sem_desc.type = CU_EXTERNAL_SEMAPHORE_HANDLE_TYPE_TIMELINE_SEMAPHORE_WIN32;
+        break;
+      case ORT_EXTERNAL_SEMAPHORE_VK_TIMELINE_SEMAPHORE_OPAQUE_FD:
+        ext_sem_desc.type = CU_EXTERNAL_SEMAPHORE_HANDLE_TYPE_TIMELINE_SEMAPHORE_FD;
+        break;
+      default:
+        // Should not reach here - ImportSemaphoreImpl already validated handle type
+        return impl.ort_api.CreateStatus(ORT_EP_FAIL, "Unexpected external memory handle type");
+    }
+#if defined(_WIN32)
     ext_sem_desc.handle.win32.handle = desc->native_handle;
+#else
+    ext_sem_desc.handle.fd = static_cast<int>(reinterpret_cast<intptr_t>(desc->native_handle));
+#endif
     ext_sem_desc.flags = 0;
 
     // Import the external semaphore
@@ -956,7 +992,7 @@ struct NvTrtRtxExternalResourceImporterImpl : OrtExternalResourceImporterImpl {
     // Get the CUDA stream from OrtSyncStream
     cudaStream_t cuda_stream = static_cast<cudaStream_t>(impl.ort_api.SyncStream_GetHandle(stream));
 
-    // Setup signal parameters for D3D12 fence (timeline semaphore)
+    // Setup signal parameters for D3D12 fence / VK timeline semaphore
     CUDA_EXTERNAL_SEMAPHORE_SIGNAL_PARAMS signal_params = {};
     signal_params.params.fence.value = value;
     signal_params.flags = 0;
@@ -993,8 +1029,6 @@ struct NvTrtRtxExternalResourceImporterImpl : OrtExternalResourceImporterImpl {
   const OrtApi& ort_api;
   const OrtEpApi& ep_api;
 };
-
-#endif  // defined(_WIN32)
 
 // OrtEpApi infrastructure to be able to use the NvTensorRTRTX EP as an OrtEpFactory for auto EP selection.
 struct NvTensorRtRtxEpFactory : OrtEpFactory {
@@ -1280,17 +1314,11 @@ struct NvTensorRtRtxEpFactory : OrtEpFactory {
 
     *out_importer = nullptr;
 
-#if defined(_WIN32)
     // Create the external resource importer
     auto importer = std::make_unique<NvTrtRtxExternalResourceImporterImpl>(ep_device, factory.ort_api);
     *out_importer = importer.release();
 
     return nullptr;
-#else
-    ORT_UNUSED_PARAMETER(ep_device);
-    return factory.ort_api.CreateStatus(ORT_NOT_IMPLEMENTED,
-                                        "External resource import is only available on Windows builds.");
-#endif
   }
 
   /**
@@ -1529,9 +1557,40 @@ struct NvTensorRtRtxEpFactory : OrtEpFactory {
                                        "[NvTensorRTRTX EP] D3D12 CIG context creation not supported on this platform");
 #endif
     } else if (config->graphics_api == ORT_GRAPHICS_API_VULKAN) {
-      // TODO: Add Vulkan CIG context support if needed
-      return onnxruntime::CreateStatus(ORT_NOT_IMPLEMENTED,
-                                       "[NvTensorRTRTX EP] Vulkan CIG context not yet implemented");
+      int cig_supported{false};
+      if (cudaSuccess != cudaDeviceGetAttribute(&cig_supported, cudaDevAttrVulkanCigSupported, device_id)) {
+        return onnxruntime::CreateStatus(ORT_EP_FAIL,
+                                         "[NvTensorRTRTX EP] Could not determine CiG support for CUDA device");
+      }
+      if (!cig_supported) {
+        LOGS_DEFAULT(INFO) << "[NvTensorRTRTX EP] InitGraphicsInterop: CiG for Vulkan is not supported on the given device. Will use the default CUDA context";
+        return nullptr;
+      }
+      const char* nv_blob_ptr_str = factory.ort_api.GetKeyValue(config->additional_options, onnxruntime::nv::provider_option_names::kExternalComputeQueueDataParamNV_data);
+      if (!nv_blob_ptr_str) {
+        LOGS_DEFAULT(WARNING) << "[NvTensorRTRTX EP] InitGraphicsInterop: Can't enable CUDA in Graphics (CiG) for Vulkan without onnxruntime::nv::provider_option_names::kExternalComputeQueueDataParamNV_data";
+        return nullptr;
+      }
+      uint64_t nv_blob_ptr = std::stoull(nv_blob_ptr_str);
+      if (nv_blob_ptr == 0) {
+        return onnxruntime::CreateStatus(ORT_EP_FAIL,
+                                         "[NvTensorRTRTX EP] Could not parse provided values for onnxruntime::nv::provider_option_names::kExternalComputeQueueDataParamNV_data or onnxruntime::nv::provider_option_names::kExternalComputeQueueDataParamNV_data_len");
+      }
+
+      CUctxCigParam cig_params{};
+      cig_params.sharedDataType = CIG_DATA_TYPE_NV_BLOB;
+      cig_params.sharedData = reinterpret_cast<void*>(nv_blob_ptr);
+      CUctxCreateParams params{};
+      params.cigParams = &cig_params;
+
+      cu_result = cuCtxCreate_v4(&cig_context, &params, 0, device_id);
+      if (cu_result != CUDA_SUCCESS) {
+        const char* error_str = nullptr;
+        cuGetErrorString(cu_result, &error_str);
+        std::string error_msg = "[NvTensorRTRTX EP] Failed to create CIG context for Vulkan: ";
+        error_msg += error_str ? error_str : "unknown error";
+        return onnxruntime::CreateStatus(ORT_FAIL, error_msg.c_str());
+      }
     } else {
       return onnxruntime::CreateStatus(ORT_INVALID_ARGUMENT,
                                        "[NvTensorRTRTX EP] Unsupported graphics API for CIG context");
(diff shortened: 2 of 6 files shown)


gedoensmax:
@nieubank and @skottmckay could you review the vulkan interop. Especially the enum changes are required to use our next EP ABI drop.

nieubank:
> @nieubank and @skottmckay could you review the vulkan interop. Especially the enum changes are required to use our next EP ABI drop.

This looks like what I would expect. No concerned from me, @skottmckay?

include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -972,15 +972,15 @@ typedef void (*RunAsyncCallbackFn)(void* user_data, OrtValue** outputs, size_t n
...
  *
  * \since Version 1.24.
  */
 typedef enum OrtExternalMemoryHandleType {
-  ORT_EXTERNAL_MEMORY_HANDLE_TYPE_D3D12_RESOURCE = 0, /**< Shared HANDLE from ID3D12Device::CreateSharedHandle(resource) */
-  ORT_EXTERNAL_MEMORY_HANDLE_TYPE_D3D12_HEAP = 1,     /**< Shared HANDLE from ID3D12Device::CreateSharedHandle(heap) */
+  ORT_EXTERNAL_MEMORY_HANDLE_TYPE_D3D12_RESOURCE = 0,      /**< Shared HANDLE from ID3D12Device::CreateSharedHandle(resource) */
+  ORT_EXTERNAL_MEMORY_HANDLE_TYPE_D3D12_HEAP = 1,          /**< Shared HANDLE from ID3D12Device::CreateSharedHandle(heap) */
+  ORT_EXTERNAL_MEMORY_HANDLE_TYPE_VK_BUFFER_WIN32 = 2,     /**< Shared HANDLE from vkGetMemoryWin32HandleKHR */
+  ORT_EXTERNAL_MEMORY_HANDLE_TYPE_VK_BUFFER_OPAQUE_FD = 3, /**< File descriptor from vkGetMemoryWin32HandleKHR */
```
theHamsta: I'm undecided whether we should name this specific to VK_BUFFER or just VK_MEMORY generically. If we were to add images later they would also be created from exported memory. Naming this specific to VK_BUFFER would allow the EP to advertise only specific usages of vk_memory.

tianleiwu:
Below are AI analysis. Some might be wrong (like assumptions that it will be compiled for other OS other than Windows/Linux):

---------
## Summary

This PR adds Vulkan interop support for the NvTensorRtRtx execution provider, enabling graphics interop on Vulkan (and Linux). It mirrors the existing D3D12 interop approach:

- Adds new enum values for Vulkan memory handles (`VK_BUFFER_WIN32`, `VK_BUFFER_OPAQUE_FD`) and Vulkan timeline semaphores (`VK_TIMELINE_SEMAPHORE_WIN32`, `VK_TIMELINE_SEMAPHORE_OPAQUE_FD`) in the public C API.
- Extends `nv_provider_factory.cc` to handle Vulkan handle types for memory and semaphore import on both Windows and Linux (removing the Windows-only `#if defined(_WIN32)` guard).
- Adds a Vulkan-Headers dependency in `cmake/deps.txt`.
- Adds a comprehensive Vulkan interop test (`nv_vulkan_test.cc`, ~837 lines).

---

## Critical Issues (Must Fix)

### 1. **BUG: Missing comparison operator in `CanImportSemaphoreImpl` (Windows path)**

**File**: `onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc`

```cpp
// Current (buggy):
return type == ORT_EXTERNAL_SEMAPHORE_D3D12_FENCE || ORT_EXTERNAL_SEMAPHORE_VK_TIMELINE_SEMAPHORE_WIN32;

// Correct:
return type == ORT_EXTERNAL_SEMAPHORE_D3D12_FENCE || type == ORT_EXTERNAL_SEMAPHORE_VK_TIMELINE_SEMAPHORE_WIN32;
```

`ORT_EXTERNAL_SEMAPHORE_VK_TIMELINE_SEMAPHORE_WIN32` is enum value `1`, which is truthy. This means `CanImportSemaphoreImpl` will **always return `true`** on Windows regardless of `type`, bypassing the validation check entirely. Any invalid semaphore type would be accepted, potentially causing silent failures or crashes downstream.

---

## Major Issues

### 2. **Inconsistent preprocessor guard style: `#if _WIN32` vs `#if defined(_WIN32)`**

In the production code (`nv_provider_factory.cc`), the new code uses bare `#if _WIN32` and `#elif __linux__` in several places (lines around `CanImportMemoryImpl`, `CanImportSemaphoreImpl`, `ImportMemoryImpl`, `ImportSemaphoreImpl`), while the test file and existing code consistently use `#if defined(_WIN32)`.

- `#if _WIN32` will trigger `-Wundef` warnings when `_WIN32` is not defined (e.g., on Linux/macOS). The standard safe pattern is `#if defined(_WIN32)`.
- Similarly `#elif __linux__` should be `#elif defined(__linux__)`.

**Affected locations in `nv_provider_factory.cc`** (new code):
- `CanImportMemoryImpl`: `#if _WIN32` / `#elif __linux__`
- `ImportMemoryImpl`: `#if _WIN32` / `#else`
- `CanImportSemaphoreImpl`: `#if _WIN32` / `#elif __linux__`
- `ImportSemaphoreImpl`: `#if _WIN32` / `#else`

**Recommendation**: Use `#if defined(_WIN32)` and `#elif defined(__linux__)` consistently to match existing code style and avoid compiler warnings.

### 3. **No `default` case in semaphore type `switch` in `ImportSemaphoreImpl`**

```cpp
switch (desc->type) {
  case ORT_EXTERNAL_SEMAPHORE_D3D12_FENCE:
    ext_sem_desc.type = CU_EXTERNAL_SEMAPHORE_HANDLE_TYPE_D3D12_FENCE;
    break;
  case ORT_EXTERNAL_SEMAPHORE_VK_TIMELINE_SEMAPHORE_WIN32:
    ext_sem_desc.type = CU_EXTERNAL_SEMAPHORE_HANDLE_TYPE_TIMELINE_SEMAPHORE_WIN32;
    break;
  case ORT_EXTERNAL_SEMAPHORE_VK_TIMELINE_SEMAPHORE_OPAQUE_FD:
    ext_sem_desc.type = CU_EXTERNAL_SEMAPHORE_HANDLE_TYPE_TIMELINE_SEMAPHORE_FD;
    break;
  // Missing: default case
}
```

The equivalent `switch` in `ImportMemoryImpl` correctly has a `default:` case that returns an error. The semaphore switch should be consistent, especially since `CanImportSemaphoreImpl` is buggy (issue #1) and cannot be relied upon to filter invalid types.

### 4. **Missing return value on non-Win32/non-Linux platforms in `CanImportMemoryImpl` and `CanImportSemaphoreImpl`**

Both `CanImportMemoryImpl` and `CanImportSemaphoreImpl` use:
```cpp
#if _WIN32
    return ...;
#elif __linux__
    return ...;
#endif
```

If compiled on a platform that is neither Windows nor Linux (e.g., macOS), neither branch is taken and the function has **no return statement** — this is undefined behavior. Add an `#else return false;` fallback.

---

## Minor Issues / Suggestions

### 5. **File descriptor cast chain**

```cpp
ext_sem_desc.handle.fd = (int)(size_t)desc->native_handle;
ext_mem_desc.handle.fd = (int)(size_t)(desc->native_handle);
```

This double cast (`void* → size_t → int`) works in practice but is somewhat fragile. Consider using `static_cast<int>(reinterpret_cast<intptr_t>(desc->native_handle))` for clarity — which is actually the pattern used in the test code for the reverse direction.

### 6. **Comment updates**

The comment `// CUDA supports D3D12 resource and heap handles` in `CanImportMemoryImpl` is not updated to mention Vulkan buffer handles. Similarly, the comment above `CanImportSemaphoreImpl` still says `// CUDA supports D3D12 timeline fences` but should now mention Vulkan timeline semaphores.

### 7. **`is_dedicated = false` assumption for Vulkan**

```cpp
case ORT_EXTERNAL_MEMORY_HANDLE_TYPE_VK_BUFFER_WIN32:
    cu_handle_type = CU_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_WIN32;
    is_dedicated = false;  // assuming no for non-images
    break;
```

The comment "assuming no for non-images" is technically inaccurate — dedicated allocations are about whether `VkMemoryDedicatedAllocateInfo` was used during `vkAllocateMemory`, not about images vs buffers. With Vulkan, buffers can also have dedicated allocations. Consider either:
- Documenting this limitation more clearly, or
- Adding a field to `OrtExternalMemoryDescriptor` to specify dedicated vs. non-dedicated (future work).

### 8. **Test file: `VK_EXT_EXTERNAL_MEMORY_DMA_BUF_EXTENSION_NAME` requested but not used**

The test requests `VK_EXT_EXTERNAL_MEMORY_DMA_BUF_EXTENSION_NAME` on Linux but then asserts `EXPECT_FALSE(test_params.use_dmabuf)` — the extension is loaded but never exercised. This is not harmful but may confuse readers. Consider adding a comment explaining this is for future use, or removing it until DMA-BUF support is implemented.

### 9. **Test: CiG test only runs on Windows**

```cpp
#if defined(_WIN32)
TEST(NvExecutionProviderVulkanTest, VkForceCig) {
```

The CiG (CUDA in Graphics) test is Windows-only. Should consider adding a Linux CiG test as well if CiG is supported on Linux, or adding a comment explaining why it's Windows-only.

### 10. **Public C API: remaining TODO for DMA-BUF**

The remaining TODO in `onnxruntime_c_api.h`:
```cpp
 * \todo Add Linux DMA-BUF file descriptor for embedded GPU memory sharing
```
is appropriate to keep as it's not addressed by this PR.

---

## API Design Observations

- The enum extension pattern (adding new values at the end) is correct and ABI-compatible.
- The use of Vulkan timeline semaphores is a good match for the existing `ID3D12Fence` pattern since both are timeline-based synchronization primitives.
- The Linux path using opaque file descriptors is the standard approach for CUDA ↔ Vulkan interop.

---

## Verdict

The overall approach is sound and the test coverage is thorough. However, **Issue #1 (missing `type ==` comparison)** is a critical logic bug that must be fixed before merge. Issues #2–#4 should also be addressed as they affect code correctness/portability.

include/onnxruntime/core/session/onnxruntime_c_api.h
theHamsta: Renamed this to `XXX_TYPE_VK_MEMORY_XXX` since the API is about the memory being exported/imported, no specifics about VkBuffer. If extending this API in the future for images, then the specifics about the image could be specified during an alternative CreateTensorFromMemory for images

theHamsta:
I addressed most of the AI comments and renamed the handles from `ORT_EXTERNAL_MEMORY_HANDLE_TYPE_VK_BUFFER_WIN32`/`ORT_EXTERNAL_MEMORY_HANDLE_TYPE_VK_BUFFER_OPAQUE` to ORT_EXTERNAL_MEMORY_HANDLE_TYPE_VK_MEMORY_WIN32`/`ORT_EXTERNAL_MEMORY_HANDLE_TYPE_VK_MEMORY_OPAQUE`.

This PR does not consider https://github.com/microsoft/onnxruntime/commit/ae609bc67e03826100c229f0a5c827b647cc8892 yet. I might add support for https://github.com/microsoft/onnxruntime/commit/ae609bc67e03826100c229f0a5c827b647cc8892 as a separate commit if the required changes are small

I saw the Linux CI was failing for this PR, but updated this PR before checking the exact output. It might be that the Linux CI is running on a server GPU, VM or container that does not have graphics enabled and that we need to skip the test if device creation or instance creation fails. I'm skipping the test in that case now. We might not test the Linux path with that though and should check the reason why.

theHamsta:
Support for InitGraphicsInterop would probably mean https://github.com/microsoft/onnxruntime/commit/ce2a871438e448b6151eff8e7529d98546a7b439 . The API doesn't seem to be well suited for Vulkan: e.g. VkQueue is not very useful if one does not also have VkInstance/VkDevice. For the case of CUDA, one could require passing `onnxruntime::nv::provider_option_names::kExternalComputeQueueDataParamNV_data`

Please let me know if I should include it in this PR!

@tianleiwu could you re-trigger the CI?

skottmckay:
> Support for InitGraphicsInterop would probably mean [ce2a871](https://github.com/microsoft/onnxruntime/commit/ce2a871438e448b6151eff8e7529d98546a7b439) . The API doesn't seem to be well suited for Vulkan: e.g. VkQueue is not very useful if one does not also have VkInstance/VkDevice. 
> 
> Please let me know if I should include it in this PR!

Can the VkInstance/VkDevice not be obtained via the info in the `OrtEpDevice` that is passed to `InitGraphicsInteropForEpDeviceInitGraphicsInteropForEpDevice`?

If we need to adjust what is passed in we can as this is all new in the current release. we did want to avoid potential inconsistencies if possible. e.g. if VkDevice can be obtained from the OrtEpDevice that's ideal as there's no chance of a mismatch between a manually provided VkDevice and the device specified by the OrtEpDevice.

theHamsta:
> Can the VkInstance/VkDevice not be obtained via the info in the `OrtEpDevice` that is passed to `InitGraphicsInteropForEpDeviceInitGraphicsInteropForEpDevice`?
> 
> If we need to adjust what is passed in we can as this is all new in the current release. we did want to avoid potential inconsistencies if possible. e.g. if VkDevice can be obtained from the OrtEpDevice that's ideal as there's no chance of a mismatch between a manually provided VkDevice and the device specified by the OrtEpDevice.

Not really, the user can create an arbitrary amount of Vulkan instances and Vulkan device that are derived from that instances (they are all different). VkDevice is a logical device. The corresponding VkPhysicalDevice could maybe be derived from the info in EpDevice, although even that is challenging at the moment:
- LUID can be used for matching on Windows, which is ok
- non-Windows platforms use a UUID to match devices which is currently not exposed for EpDevice.

Also, the EP can not really do Vulkan calls without a https://docs.vulkan.org/refpages/latest/refpages/source/vkGetDeviceProcAddr.html and https://docs.vulkan.org/refpages/latest/refpages/source/vkGetInstanceProcAddr.html and of course VkDevice, VkInstance and VkPhysicalDevice.

It would be interesting what other EPs would require for Vulkan-interop. For CUDA, interop works without any special preparation and is more efficient with the data returned from https://docs.vulkan.org/refpages/latest/refpages/source/vkGetExternalComputeQueueDataNV.html (and that data is all that is needed)

skottmckay:
> Not really, the user can create an arbitrary amount of Vulkan instances and Vulkan device that are derived from that instances (they are all different). VkDevice is a logical device. The corresponding VkPhysicalDevice could maybe be derived from the info in EpDevice, although even that is challenging at the moment:
> 
> * LUID can be used for matching on Windows, which is ok
> * non-Windows platforms use a UUID to match devices which is currently not exposed for EpDevice.
> 
> Also, the EP can not really do Vulkan calls without a https://docs.vulkan.org/refpages/latest/refpages/source/vkGetDeviceProcAddr.html and https://docs.vulkan.org/refpages/latest/refpages/source/vkGetInstanceProcAddr.html and of course VkDevice, VkInstance and VkPhysicalDevice.
> 
> It would be interesting what other EPs would require for Vulkan-interop. For CUDA, interop works without any special preparation and is more efficient with the data returned from https://docs.vulkan.org/refpages/latest/refpages/source/vkGetExternalComputeQueueDataNV.html (and that data is all that is needed)

Is my understanding correct that we would need to pass in these for VK interop in general, but you don't need them as using `vkGetExternalComputeQueueDataNV` via EP options is a preferred approach?
* vkDevice - logical device, should match physical device represented by OrtEpDevice but not 1:1
* vkInstance - user could have multiple 
* vkQueue 

If the EP implementer is making Vulkan calls I'm assuming they would have a dependency on the Vulkan library and can call vkGetInstanceProcAddr and vkGetDeviceProcAddr using the provided vkInstance and vkDevice respectively. Is that assumption reasonable?

Is the VkPhysicalDevice required for the interop? I could see it _maybe_ being used to optionally validate that the vkDevice matches, but not clear to me if it's needed otherwise as I would have assumed using the vkDevice is preferred. 

If there's additional metadata that is meaningful to add to the OrtHardwareDevice on non-Windows platforms like the UUID we could add that during device discovery.

theHamsta:
> Is my understanding correct that we would need to pass in these for VK interop in general, but you don't need them as using vkGetExternalComputeQueueDataNV via EP options is a preferred approach?

yes, that's correct. It's difficult to say what is needed in general. I'm only aware of CUDA path.

> If the EP implementer is making Vulkan calls I'm assuming they would have a dependency on the Vulkan library and can call vkGetInstanceProcAddr and vkGetDeviceProcAddr using the provided vkInstance and vkDevice respectively. Is that assumption reasonable?

Vulkan can be consumed via linking, but also via dynamic loading, also the Vulkan call can be layered via Vulkan layers that get resolved by the Vulkan loader. A portable approach for Vulkan middleware are function pointers to https://docs.vulkan.org/refpages/latest/refpages/source/vkGetInstanceProcAddr.html and https://docs.vulkan.org/refpages/latest/refpages/source/vkGetDeviceProcAddr.html . This way you can get the Vulkan calls the way the application prefers (this PR uses dynamic loading via Vulkan-HPP). I'm not sure whether we want to perform Vulkan calls (e.g. perform the external compute queue calls on behalf of the user) or whether it is better to leave all the control to the application

I assume that VkInstance/VkDevice could potentially be useful to prepare for Vulkan, but I can provide concrete examples. I'm not an expert for other vendors. AMD RocM/Vulkan interop seems would work similar to the current CUDA interop that is happening without the InitForGraphics call. When I look for QNN documentation, it seem to me as if no special preparation for a specific Vulkan device is necessary. For OpenVino, I only seem to document some preparation steps for OpenCL interop where they would need either the cl_context or cl_queue for preparation https://docs.openvino.ai/2025/openvino-workflow/running-inference/inference-devices-and-modes/gpu-device/remote-tensor-api-gpu-plugin.html#examples which would fit into the current API design

> If there's additional metadata that is meaningful to add to the OrtHardwareDevice on non-Windows platforms like the UUID we could add that during device discovery.

Yes, UUID would indeed be useful. However, UUID can only be inferred from GPU APIs like CUDA, Vulkan, OpenCL or nvml. So it is likely something that the providers could add to the EPs if this is possible. LUID also does not work for CUDA<->dxcore matching on WSL https://github.com/microsoft/onnxruntime/issues/27286

For Linux, we currently use the PCI bus id to match sysfs files with CUDA. The Vulkan extension https://docs.vulkan.org/refpages/latest/refpages/source/VK_EXT_pci_bus_info.html seems to be mostly implemented only by AMD devices

If the user is willing to use CUDA in their application a matching of Vulkan and EpDevice is possible via the detour

Vulkan <- UUID -> CUDA <- CUDA device index -> ort_api.EpDevice_EpOptions(ep_device)

one could add UUID here https://github.com/theHamsta/onnxruntime/blob/ce2a871438e448b6151eff8e7529d98546a7b439/onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc#L1200 additionally to `device_id`, but it would be difficult to discover that UUID is in the EpOptions and not the general metadata.

theHamsta:
> > Is my understanding correct that we would need to pass in these for VK interop in general, but you don't need them as using vkGetExternalComputeQueueDataNV via EP options is a preferred approach?
> 
> yes, that's correct. It's difficult to say what is needed in general. I'm only aware of CUDA path.
> 
> > If the EP implementer is making Vulkan calls I'm assuming they would have a dependency on the Vulkan library and can call vkGetInstanceProcAddr and vkGetDeviceProcAddr using the provided vkInstance and vkDevice respectively. Is that assumption reasonable?
> 
> Vulkan can be consumed via linking, but also via dynamic loading, also the Vulkan call can be layered via Vulkan layers that get resolved by the Vulkan loader. A portable approach for Vulkan middleware are function pointers to https://docs.vulkan.org/refpages/latest/refpages/source/vkGetInstanceProcAddr.html and https://docs.vulkan.org/refpages/latest/refpages/source/vkGetDeviceProcAddr.html . This way you can get the Vulkan calls the way the application prefers (this PR uses dynamic loading via Vulkan-HPP). I'm not sure whether we want to perform Vulkan calls (e.g. perform the external compute queue calls on behalf of the user) or whether it is better to leave all the control to the application
> 
> I assume that VkInstance/VkDevice could potentially be useful to prepare for Vulkan, but I can provide concrete examples. I'm not an expert for other vendors. AMD RocM/Vulkan interop seems would work similar to the current CUDA interop that is happening without the InitForGraphics call. When I look for QNN documentation, it seem to me as if no special preparation for a specific Vulkan device is necessary. For OpenVino, I only seem to document some preparation steps for OpenCL interop where they would need either the cl_context or cl_queue for preparation https://docs.openvino.ai/2025/openvino-workflow/running-inference/inference-devices-and-modes/gpu-device/remote-tensor-api-gpu-plugin.html#examples which would fit into the current API design
> 
> > If there's additional metadata that is meaningful to add to the OrtHardwareDevice on non-Windows platforms like the UUID we could add that during device discovery.
> 
> Yes, UUID would indeed be useful. However, UUID can only be inferred from GPU APIs like CUDA, Vulkan, OpenCL or nvml. So it is likely something that the providers could add to the EPs if this is possible. LUID also does not work for CUDA<->dxcore matching on WSL #27286
> 
> For Linux, we currently use the PCI bus id to match sysfs files with CUDA. The Vulkan extension https://docs.vulkan.org/refpages/latest/refpages/source/VK_EXT_pci_bus_info.html seems to be mostly implemented only by AMD devices
> 
> If the user is willing to use CUDA in their application a matching of Vulkan and EpDevice is possible via the detour
> 
> Vulkan <- UUID -> CUDA <- CUDA device index -> ort_api.EpDevice_EpOptions(ep_device)
> 
> one could add UUID here https://github.com/theHamsta/onnxruntime/blob/ce2a871438e448b6151eff8e7529d98546a7b439/onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc#L1200 additionally to `device_id`, but it would be difficult to discover that UUID is in the EpOptions and not the general metadata.

Small correction, Nvidia seems to support `VK_EXT_pci_bus_info`. It could be used for Linux device matching. The EpDevices already have the PCI bus id in their metadata.

```cpp
      for (const auto& d : ep_devices) {
        if (d.Device().VendorId() == props.properties.vendorID && d.Device().DeviceId() == props.properties.deviceID) {
#if defined(_WIN32)
          // verify the real device with LUID, but on Linux we only have UUID which we don't know for EpDevice
          auto luid = d.Device().Metadata().GetValue("LUID");
          if (id_props.deviceLUIDValid && luid) {
            LUID vk_luid;
            std::memcpy(&vk_luid, dev.id_props.deviceLUID, sizeof(LUID));
            uint64_t ep_luid = std::stoull(luid);
            uint64_t vk = (uint64_t(vk_luid.HighPart) << 32) | uint64_t(vk_luid.LowPart);
            if (ep_luid != vk) {
              continue;
            }
          }
#else
          auto pci_bus_id = d.Device().Metadata().GetValue("pci_bus_id");
          if (pci_bus_id) {
            std::cmatch matches;
            if (std::regex_match(pci_bus_id, matches, pci_bus_id_pattern)) {
              auto domain = std::stoull(matches[1].str(), nullptr, 16);
              auto bus = std::stoull(matches[2].str(), nullptr, 16);
              auto device = std::stoull(matches[3].str(), nullptr, 16);
              auto function = std::stoull(matches[4].str(), nullptr, 16);
              if (domain != pci_props.pciDomain || bus != pci_props.pciBus || device != pci_props.pciDevice || function != pci_props.pciFunction) {
                continue;
              }
            }
          }
#endif
          dev.ep_device_candidates.push_back(d);
        }
```

tianleiwu:
@theHamsta,

Latest AI analysis:

---
### 1. Vulkan Test Resource Lifetime (`onnxruntime/test/providers/nv_tensorrt_rtx/nv_vulkan_test.cc`)

**Positive**:
- Consolidating the Vulkan objects, imported ORT handles, and cleanup into a single `VkResources` RAII holder is the right structure for a large interop test. It keeps the teardown ordering explicit and makes the test much easier to follow.

**Concern**:
- ⚠️ **Uninitialized handles in skip/failure paths**: [`VkResources`](https://github.com/microsoft/onnxruntime/blob/0ccb08722c52a362420ff2ffc26cc6d49ecef82c/onnxruntime/test/providers/nv_tensorrt_rtx/nv_vulkan_test.cc#L119-L123) leaves `instance` and `device` uninitialized, but the destructor branches on them in the common cleanup path. If `init_vulkan_interop()` hits an early `GTEST_SKIP()` or a failed Vulkan call before both handles are assigned, the destructor will read indeterminate values and may try to destroy garbage handles. These members should be value-initialized to `VK_NULL_HANDLE` up front.

### 2. Linux Opaque-FD Ownership (`onnxruntime/test/providers/nv_tensorrt_rtx/nv_vulkan_test.cc`)

**Positive**:
- The test covers the same handle types that the importer exposes in the public C API, which is exactly the right level of integration coverage for this feature.

**Concern**:
- ⚠️ **The test closes FDs after CUDA already owns them**: on Linux, the test imports Vulkan memory/semaphore FDs into CUDA in [`create_timeline_semaphore`](https://github.com/microsoft/onnxruntime/blob/0ccb08722c52a362420ff2ffc26cc6d49ecef82c/onnxruntime/test/providers/nv_tensorrt_rtx/nv_vulkan_test.cc#L389-L400) and [`allocate_buffer`](https://github.com/microsoft/onnxruntime/blob/0ccb08722c52a362420ff2ffc26cc6d49ecef82c/onnxruntime/test/providers/nv_tensorrt_rtx/nv_vulkan_test.cc#L483-L505), then later closes those same FDs in [`VkResources::~VkResources`](https://github.com/microsoft/onnxruntime/blob/0ccb08722c52a362420ff2ffc26cc6d49ecef82c/onnxruntime/test/providers/nv_tensorrt_rtx/nv_vulkan_test.cc#L137-L168). CUDA's `OPAQUE_FD` import takes ownership on success, so touching the FD afterward is undefined behavior. At best this becomes a spurious `EBADF`; at worst it closes an unrelated resource if the descriptor number has been reused.

### 3. Non-CiG Coverage (`onnxruntime/test/providers/nv_tensorrt_rtx/nv_vulkan_test.cc`)

**Positive**:
- Splitting the test matrix into `VkCigDisabled` and `VkForceCig` is a good design. It makes the intended capability split visible instead of burying it behind runtime branching.

**Concern**:
- ⚠️ **`VkCigDisabled` still hard-requires CiG plumbing**: `init_vulkan_interop()` always enables `VK_NV_external_compute_queue`, reserves an external queue, creates `VkExternalComputeQueueNV`, and later `test_vulkan_interop()` always fetches its queue data ([device setup](https://github.com/microsoft/onnxruntime/blob/0ccb08722c52a362420ff2ffc26cc6d49ecef82c/onnxruntime/test/providers/nv_tensorrt_rtx/nv_vulkan_test.cc#L306-L353), [queue-data fetch](https://github.com/microsoft/onnxruntime/blob/0ccb08722c52a362420ff2ffc26cc6d49ecef82c/onnxruntime/test/providers/nv_tensorrt_rtx/nv_vulkan_test.cc#L615-L619)). That means the supposedly non-CiG test cannot run unless the CiG extension stack is also present. Given the comment that Vulkan CiG is currently Windows-only, this blocks the new Linux path from being exercised on the configurations it is meant to cover.

### 4. Vulkan Version Requirement (`onnxruntime/test/providers/nv_tensorrt_rtx/nv_vulkan_test.cc`)

**Positive**:
- Using timeline semaphores and the Vulkan 1.1/1.2 property/feature structs is consistent with the feature set the test actually needs.

**Concern**:
- ⚠️ **The test asks for Vulkan 1.4 even though the exercised feature set is older**: [`app_info.apiVersion = VK_API_VERSION_1_4`](https://github.com/microsoft/onnxruntime/blob/0ccb08722c52a362420ff2ffc26cc6d49ecef82c/onnxruntime/test/providers/nv_tensorrt_rtx/nv_vulkan_test.cc#L227-L243) unnecessarily narrows the set of drivers that can run this test. The test only uses Vulkan 1.1/1.2-era capabilities, so this should either clamp to the loader's advertised version or request the minimum version the test actually requires.

---

## Summary of Concerns

| 1 | High | Vulkan Test Resource Lifetime | `VkResources` destructor can read/destroy uninitialized Vulkan handles after early skip/failure paths. |
| 2 | High | Linux Opaque-FD Ownership | The Linux test closes imported FDs after CUDA ownership has already transferred. |
| 3 | High | Non-CiG Coverage | `VkCigDisabled` still requires the CiG extension stack and external queue creation. |
| 4 | Suggestion | Vulkan Version Requirement | The test requires Vulkan 1.4 even though it only uses older capabilities. |

---

## Verdict

REQUEST CHANGES — the core importer wiring looks reasonable, but the new Vulkan test has a few correctness and portability issues that should be fixed before it can be relied on for this feature.

theHamsta:
I did the following changes and with testing on more platforms

From AI review
- initialized the Vulkan handles with zero, add some more checks around destruction
- removed closing handles/Fds. I confirmed that the Vulkan driver would close them on `vkFreeMemory` (ownership is not transferred to CUDA)
- I change the CiG testing, the three cases are now tested on all platforms (I check now for presence of CiG extension):
     - force CiG off (should always work)
     - use InitForGraphics (auto-behavior, the implementation can use CiG, but it would also work on platform where CiG does not work)
     - force CiG on, this test is skipped when CUDA says that Vulkan-CiG is not supported
- on Vulkan 1.4: did not change that. I'm assuming a recent Nvidia baseline in this test

Other changes:
- added InitForGraphics support
- guarded Vulkan::Headers: when you enable the WebGPU EP and Dawn with Vulkan, then Dawn will bring it's own Vulkan headers that are not controlled by ORT, we then use whatever Dawn uses, instead of creating a new conflicting version
- I removed the DEVICE_LOCAL flag for host allocations: this is not required

onnxruntime/test/providers/nv_tensorrt_rtx/nv_vulkan_test.cc
```diff
@@ -0,0 +1,895 @@
...
+    pci_props.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_PCI_BUS_INFO_PROPERTIES_EXT;
+    VkPhysicalDeviceVulkan11Properties id_props{};
+    id_props.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_VULKAN_1_1_PROPERTIES;
+    id_props.pNext = &pci_props;
+    VkPhysicalDeviceProperties2 props{};
+    props.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_PROPERTIES_2;
+    props.pNext = &id_props;
+    resources.loader.vkGetPhysicalDeviceProperties2(p, &props);
+    if (props.properties.vendorID == 0x10DE) {
+      std::memcpy(&dev.props, &props, sizeof(props));
```
tianleiwu, line 280: ⚠️ **`memcpy` of `VkPhysicalDeviceProperties2` into `VkPhysicalDeviceProperties`**: 
  ```cpp
  std::memcpy(&dev.props, &props, sizeof(props));
  ```
  Here `dev.props` is `VkPhysicalDeviceProperties` (no `2` suffix) but `props` is `VkPhysicalDeviceProperties2`. The `sizeof(props)` is the size of the `2` variant, which is larger. This **overflows** `dev.props` and corrupts adjacent memory (`dev.id_props`). Should be:
  ```cpp
  dev.props = props.properties;
  ```
  Similarly:
  ```cpp
  std::memcpy(&dev.id_props, &id_props, sizeof(id_props));
  ```
  This is fine since both sides are `VkPhysicalDeviceVulkan11Properties`, but the struct contains a `pNext` pointer — copying it means `dev.id_props.pNext` still points to the stack-local `pci_props`. This pointer becomes dangling once the loop iteration ends. If `id_props.pNext` is never dereferenced after the copy, this is benign, but it's fragile. Consider clearing `pNext` after copy.

onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc
```diff
@@ -1529,9 +1557,40 @@ struct NvTensorRtRtxEpFactory : OrtEpFactory {
...
+      if (!cig_supported) {
+        LOGS_DEFAULT(INFO) << "[NvTensorRTRTX EP] InitGraphicsInterop: CiG for Vulkan is not supported on the given device. Will use the default CUDA context";
+        return nullptr;
+      }
+      const char* nv_blob_ptr_str = factory.ort_api.GetKeyValue(config->additional_options, onnxruntime::nv::provider_option_names::kExternalComputeQueueDataParamNV_data);
+      if (!nv_blob_ptr_str) {
+        LOGS_DEFAULT(WARNING) << "[NvTensorRTRTX EP] InitGraphicsInterop: Can't enable CUDA in Graphics (CiG) for Vulkan without onnxruntime::nv::provider_option_names::kExternalComputeQueueDataParamNV_data";
+        return nullptr;
+      }
+      uint64_t nv_blob_ptr = std::stoull(nv_blob_ptr_str);
```
tianleiwu, line 1574: ⚠️ **`std::stoull` can throw on malformed input** (High-Priority): `std::stoull(nv_blob_ptr_str)` will throw `std::invalid_argument` or `std::out_of_range` if the string is not a valid unsigned integer. This is inside a `noexcept`-equivalent C API implementation — an uncaught exception will call `std::terminate`. Wrap in a try-catch or use a safe parser.
  ```cpp
  uint64_t nv_blob_ptr = 0;
  try {
    nv_blob_ptr = std::stoull(nv_blob_ptr_str);
  } catch (...) {
    return onnxruntime::CreateStatus(ORT_INVALID_ARGUMENT,
        "[NvTensorRTRTX EP] Invalid value for kExternalComputeQueueDataParamNV_data");
  }
  ```

onnxruntime/test/providers/nv_tensorrt_rtx/nv_vulkan_test.cc
tianleiwu: The fix can be in another PR.

onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc
tianleiwu: The fix can be in another PR.

onnxruntime/test/providers/nv_tensorrt_rtx/nv_vulkan_test.cc
```diff
@@ -0,0 +1,895 @@
+#include <cstddef>
```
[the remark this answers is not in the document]
tianleiwu: resolved

PR #27457, opened 2026-02-25
softplus support for webgpu

guschmue:
needed by onnx-community/Falcon-H1-Tiny-90M-Instruct-ONNX

### DIFF
--- a/onnxruntime/core/providers/webgpu/math/unary_elementwise_ops.cc
+++ b/onnxruntime/core/providers/webgpu/math/unary_elementwise_ops.cc
@@ -291,7 +291,29 @@ WEBGPU_ELEMENTWISE_KERNEL(LeakyRelu, 16, WebGpuSupportedFloatTypes())
 WEBGPU_LU_IMPL(ThresholdedRelu, "select(vec4<x_element_t>(0), a, a > vec4<x_element_t>(uniforms.attr))", "", 1.0f)
 WEBGPU_ELEMENTWISE_KERNEL(ThresholdedRelu, 10, WebGpuSupportedFloatTypes())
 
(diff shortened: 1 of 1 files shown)


onnxruntime/core/providers/webgpu/math/unary_elementwise_ops.cc
```diff
@@ -291,5 +291,9 @@ WEBGPU_ELEMENTWISE_KERNEL(LeakyRelu, 16, WebGpuSupportedFloatTypes())
 WEBGPU_LU_IMPL(ThresholdedRelu, "select(vec4<x_element_t>(0), a, a > vec4<x_element_t>(uniforms.attr))", "", 1.0f)
 WEBGPU_ELEMENTWISE_KERNEL(ThresholdedRelu, 10, WebGpuSupportedFloatTypes())
 
+WEBGPU_ELEMENTWISE_IMPL(Softplus, "select(log(1.0 + exp(a)), a + log(1.0 + exp(-a)), a > x_value_t(0))", "", ShaderUsage::UseValueTypeAlias)
```
xenova: for `a > 20` (or some threshold), pytorch typically just approximates it to `a` for numerical stability. Similarly for a < -20.

https://docs.pytorch.org/docs/stable/generated/torch.nn.Softplus.html

guschmue: onnx spec doesn't say anything. Let me check the cpu implementation - should do the same for webgpu, else people will notice that it is different.

fs-eire: vote for adding the threshold - there are some existing webgpu implementation using approximation like Erf and Tanh and I believe the reason is majorly for numerical stability - we don't want `Nan` or `Inf` in the result.

But since ONNX spec does not define the threshold, we probably need to figure out an appropriate threshold, especially for fp16.

guschmue: cpu ep does the same as this PR. Hm.
In the absence in the onnx spec we should follow pytorch if we want to add it.

PR #27464, opened 2026-02-26
Fix 3d attention mask broadcasting in MHA

titaiwangms:
(1) Fix 3d attention mask broadcasting in MHA
(2) Refactor attention python tests of LLM (add MHA)

--- 

This pull request includes a minor fix to the attention mask broadcasting logic in the CUDA attention kernel, as well as the addition of a missing license header in a test file.

Improvements to attention mask broadcasting logic:

* Updated the logic in `attention.cc` to clarify and correct how broadcasting is determined for 3D attention masks, ensuring the batch dimension always broadcasts and the heads dimension broadcasts only if its size is 1. This improves correctness and clarity for different mask shapes.

Documentation and compliance:

* Added the standard Microsoft MIT license header to the `test_onnx_attention/__init__.py` file to ensure proper licensing information is included.

### DIFF
--- a/onnxruntime/test/python/transformers/test_onnx_attention/test_mha.py
+++ b/onnxruntime/test/python/transformers/test_onnx_attention/test_mha.py
@@ -18,7 +18,6 @@ It uses the unfused attention kernel and supports:
   - 2D, 3D, 4D additive masks with broadcasting
 """
 
-import os
 import unittest
 
 import numpy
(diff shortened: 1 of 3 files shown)


onnxruntime/test/python/transformers/test_onnx_attention/test_mha.py
```diff
@@ -0,0 +1,730 @@
+# --------------------------------------------------------------------------
```
xadupre, line 1: Do we need to split the files?

titaiwangms: I will add tensorscatter test cases as follow up, so I don't want the file to be too long.

xadupre (approved):

PR #27466, opened 2026-02-26
DQ→MatMulNBits fusion transformer for NvTensorRtRtx ep 

anujj:
## Summary

Generalize the WebNN-specific DequantizeLinear → MatMulNBits graph fusion transformer so it can be
reused by other execution providers (e.g. NvTensorRTRTX), and add defensive shape/size validation
to prevent crashes on malformed tensors.

### Fusion patterns

**Pattern 1:** `DequantizeLinear → Reshape → Transpose → [Cast] → MatMul/Gemm`  →  **MatMulNBits**

**Pattern 2:** `DequantizeLinear (axis=0) → MatMul/Gemm`  →  **MatMulNBits**

### DIFF
--- a/onnxruntime/core/optimizer/dq_matmulnbits_fusion.cc
+++ b/onnxruntime/core/optimizer/dq_matmulnbits_fusion.cc
@@ -3,7 +3,10 @@
 
 #include "core/optimizer/dq_matmulnbits_fusion.h"
 
+#if !defined(ORT_MINIMAL_BUILD)
+
 #include "core/common/common.h"
+#include "core/common/safeint.h"
 #include "core/framework/tensorprotoutils.h"
 #include "core/graph/constants.h"
 #include "core/graph/graph_utils.h"
@@ -51,7 +54,9 @@ bool HasRank2Shape(const ONNX_NAMESPACE::TensorProto& tp, int64_t dim0, int64_t 
   return tp.dims_size() == 2 && tp.dims(0) == dim0 && tp.dims(1) == dim1;
 }
 
-uint8_t GetPackedUint4Element(const uint8_t* packed, size_t index) {
+uint8_t GetPackedUint4Element(const uint8_t* packed, size_t index, size_t num_elements) {
+  ORT_ENFORCE(index < num_elements, "GetPackedUint4Element: index ", index,
+              " out of bounds (num_elements=", num_elements, ")");
   const uint8_t packed_byte = packed[index / 2];
   return (index % 2 == 0) ? static_cast<uint8_t>(packed_byte & 0x0F)
                           : static_cast<uint8_t>((packed_byte >> 4) & 0x0F);
@@ -59,15 +64,17 @@ uint8_t GetPackedUint4Element(const uint8_t* packed, size_t index) {
 
 void PackUint4Rows(const Initializer& src, int64_t rows, int64_t cols, uint8_t* dst) {
   const int64_t row_bytes = (cols + 1) / 2;
-  memset(dst, 0, static_cast<size_t>(rows * row_bytes));
+  const size_t dst_bytes = SafeInt<size_t>(rows) * row_bytes;
+  const size_t total_elements = SafeInt<size_t>(rows) * cols;
+  memset(dst, 0, dst_bytes);
 
   const auto src_packed = src.DataAsByteSpan();
   for (int64_t r = 0; r < rows; ++r) {
     for (int64_t c = 0; c < cols; ++c) {
-      const size_t src_index = static_cast<size_t>(r * cols + c);
-      const uint8_t value = GetPackedUint4Element(src_packed.data(), src_index);
+      const size_t src_index = SafeInt<size_t>(r) * cols + c;
+      const uint8_t value = GetPackedUint4Element(src_packed.data(), src_index, total_elements);
 
-      const size_t dst_index = static_cast<size_t>(r * row_bytes + c / 2);
+      const size_t dst_index = SafeInt<size_t>(r) * row_bytes + c / 2;
       if ((c & 1) == 0) {
         dst[dst_index] = value;
       } else {
@@ -79,21 +86,26 @@ void PackUint4Rows(const Initializer& src, int64_t rows, int64_t cols, uint8_t* 
 
 // Transpose and pack UINT4 weights from DQ axis=0 layout [K, N] to MatMulNBits layout [N, k_blocks, blob_size].
 // Source: row-major UINT4 with quantization along K (axis=0), shape [K, N].
+// The nibble ordering follows ONNX UINT4 convention: even indices in the low nibble,
+// odd indices in the high nibble of each byte.
 // Dest: UINT8 [N, k_blocks, block_size/2] where each byte packs two 4-bit weights.
 void TransposePackWeightsAxis0(
     const uint8_t* src_packed, int64_t K, int64_t N, int64_t block_size,
     uint8_t* dst) {
   const int64_t k_blocks = (K + block_size - 1) / block_size;
   const int64_t blob_size = block_size / 2;
-  memset(dst, 0, static_cast<size_t>(N * k_blocks * blob_size));
+  const size_t dst_bytes = SafeInt<size_t>(N) * k_blocks * blob_size;
+  const size_t total_elements = SafeInt<size_t>(K) * N;
+  memset(dst, 0, dst_bytes);
 
   for (int64_t n = 0; n < N; ++n) {
     for (int64_t k = 0; k < K; ++k) {
-      const uint8_t val = GetPackedUint4Element(src_packed, static_cast<size_t>(k * N + n));
+      const size_t src_index = SafeInt<size_t>(k) * N + n;
+      const uint8_t val = GetPackedUint4Element(src_packed, src_index, total_elements);
 
       const int64_t kb = k / block_size;
       const int64_t off = k % block_size;
-      const size_t dst_byte = static_cast<size_t>(n * k_blocks * blob_size + kb * blob_size + off / 2);
+      const size_t dst_byte = SafeInt<size_t>(n) * k_blocks * blob_size + kb * blob_size + off / 2;
       if (off % 2 == 0) {
         dst[dst_byte] = static_cast<uint8_t>((dst[dst_byte] & 0xF0) | val);
       } else {
@@ -109,13 +121,16 @@ void TransposePackZPAxis0(
     const uint8_t* src_packed, int64_t k_blocks, int64_t N,
     uint8_t* dst) {
   const int64_t zp_bytes_per_n = (k_blocks + 1) / 2;
-  memset(dst, 0, static_cast<size_t>(N * zp_bytes_per_n));
+  const size_t dst_bytes = SafeInt<size_t>(N) * zp_bytes_per_n;
+  const size_t total_elements = SafeInt<size_t>(k_blocks) * N;
+  memset(dst, 0, dst_bytes);
 
   for (int64_t n = 0; n < N; ++n) {
     for (int64_t kb = 0; kb < k_blocks; ++kb) {
-      const uint8_t val = GetPackedUint4Element(src_packed, static_cast<size_t>(kb * N + n));
+      const size_t src_index = SafeInt<size_t>(kb) * N + n;
+      const uint8_t val = GetPackedUint4Element(src_packed, src_index, total_elements);
 
-      const size_t dst_byte = static_cast<size_t>(n * zp_bytes_per_n + kb / 2);
+      const size_t dst_byte = SafeInt<size_t>(n) * zp_bytes_per_n + kb / 2;
       if (kb % 2 == 0) {
         dst[dst_byte] = static_cast<uint8_t>((dst[dst_byte] & 0xF0) | val);
       } else {
@@ -239,7 +254,7 @@ std::vector<FusionMatch> CollectReshapeTransposeMatches(
     const int64_t bs_dim = weight_const_tp->dims(2);
     if (N <= 0 || blocks <= 0 || bs_dim <= 0) continue;
     if (bs_dim != block_size) continue;
-    const int64_t K = blocks * bs_dim;
+    const int64_t K = SafeInt<int64_t>(blocks) * bs_dim;
 
     const auto* scale_arg = dq_node->InputDefs()[1];
     if (!scale_arg || !scale_arg->Exists()) continue;
@@ -297,8 +312,8 @@ std::vector<FusionMatch> CollectReshapeTransposeMatches(
     }
 
     if (const auto* b_shape = mm_inputs[1]->Shape(); b_shape && b_shape->dim_size() == 2 &&
-        utils::HasDimValue(b_shape->dim(0)) && utils::HasDimValue(b_shape->dim(1)) &&
-        (b_shape->dim(0).dim_value() != K || b_shape->dim(1).dim_value() != N)) {
+                                                     utils::HasDimValue(b_shape->dim(0)) && utils::HasDimValue(b_shape->dim(1)) &&
+                                                     (b_shape->dim(0).dim_value() != K || b_shape->dim(1).dim_value() != N)) {
       continue;
     }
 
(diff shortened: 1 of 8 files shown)


anskumar01:
Adding @skottmckay , @nieubank , @devang-ml for review

onnxruntime/core/optimizer/graph_transformer_utils.cc
```diff
@@ -274,6 +276,17 @@ InlinedVector<std::unique_ptr<GraphTransformer>> GenerateTransformers(
         transformers.emplace_back(std::make_unique<WhereDummyDq>());
       }
 
+#if !defined(DISABLE_CONTRIB_OPS)
+      if (enable_dq_matmulnbits_fusion && !disable_quant_qdq) {
```
xadupre: Should we assert when enable_dq_matmulnbits_fusion  is true and contribs ops are disabled?

onnxruntime/core/optimizer/dq_matmulnbits_fusion.cc
```diff
@@ -0,0 +1,816 @@
...
+#include <cmath>
+#include <cstring>
+#include <optional>
+#include <unordered_set>
+
+namespace onnxruntime {
+
+namespace {
+
+bool IsUniformPackedUint4Value(const Initializer& init, uint8_t expected_nibble) {
```
xadupre, line 30: It would be great to add a unit test.

onnxruntime/core/optimizer/dq_matmulnbits_fusion.cc
```diff
@@ -0,0 +1,816 @@
...
+  ORT_ENFORCE(accuracy_level_ >= 0 && accuracy_level_ <= 4,
+              "MatMulNBits accuracy level must be between 0 and 4");
+}
+
+Status DQMatMulNBitsFusion::ApplyImpl(Graph& graph, bool& modified, int graph_level,
+                                      const logging::Logger& logger) const {
+  GraphViewer graph_viewer(graph);
+  const auto& node_topology_list = graph_viewer.GetNodesInTopologicalOrder();
+
+  // Collect nodes to remove after iteration (cannot mutate graph during iteration)
```
xadupre: This is a very long function. Maybe it would be better to split it.

onnxruntime/core/optimizer/graph_transformer_utils.cc
anujj: done

onnxruntime/core/optimizer/dq_matmulnbits_fusion.cc
anujj: Addressed this in the latest commit

onnxruntime/core/optimizer/dq_matmulnbits_fusion.cc
anujj: Do we run the tests with the NvTensorRTRTX EP? These changes are guarded behind the EP check.

fdwr:
> Generalize the WebNN-specific DequantizeLinear → MatMulNBits graph fusion transformer

FYI @honry.

tianleiwu:
Some feedbacks from AI:
-----
### 3. Change to Public Header API
**File:** `include/onnxruntime/core/optimizer/graph_transformer_utils.h`

`GenerateTransformers()` gains a new `bool enable_dq_matmulnbits_fusion = false` parameter. While a default value preserves backward compatibility for callers, this file is in the `include/` directory — it's part of the **public API**. Adding EP-specific boolean flags to a general-purpose public function doesn't scale well. Every new EP-specific fusion would need another boolean parameter.

**Recommendation:** Instead of adding a boolean to `GenerateTransformers()`, consider either:
- Passing the `ExecutionProviders` container (or a capability set) so the function can check EP presence itself.
- Checking the EP internally in `inference_session.cc` and using the existing `rules_and_transformers_to_disable` mechanism to skip it when not needed (register always, disable by name when EP is absent).

---

## Design Issues

### 4. Transformer Registered at Level1 — Ordering Concern
The commit message says "registered at Level1 QDQ stage" and the code confirms it's added inside the `Level1` switch case. However, the existing `QDQSelectorActionTransformer` (which handles the standard `DQ→MatMul→MatMulNBits` fusion) runs at **Level2**. 

Running this new fusion at Level1 means it will execute **before** EP partitioning/assignment. While the commit message notes "the transformation is targeted before EP capability partitioning" as intentional, it also means the `MatMulNBits` node created here won't have an EP assigned via the normal flow — it only copies `mm_node->GetExecutionProviderType()`, which at Level1 may be empty/unset.

**Question:** Is it intentional and safe that the fused node inherits an EP type that may not yet be assigned at Level1? How does the `NvTensorRTRTXExecutionProvider` claim the resulting `MatMulNBits` node during partitioning?

### 5. Naming: "WebNN" in Code Comments vs. Generalized Use
The class comments in `dq_matmulnbits_fusion.h` and node names (`WebNNFusedMatMulNBits`) reference WebNN throughout, yet the PR description says it's "generalized" for other EPs. The `DQMatMulNBitsFusion` class name is correctly generalized, but:
- Node names like `"WebNNFusedMatMulNBits"` embed WebNN in the graph.
- The header doc comment says "Fuses the WebNN-specific pattern."
- Log messages reference "WebNN" patterns.

**Recommendation:** Update names and comments to be EP-agnostic since this is now shared infrastructure.

### 6. `intra_op_thread_pool_` Stored but Never Used
The constructor accepts and stores `concurrency::ThreadPool* intra_op_thread_pool` as `intra_op_thread_pool_`, but it's never referenced in `ApplyImpl` or any helper function. This is dead code.

**Recommendation:** Remove the parameter and member variable, or add a comment explaining planned future usage.

---

## Code Quality Issues

### 7. Large Monolithic Implementation (~830 lines new code)
While the refactored code is split into four helper functions (`CollectReshapeTransposeMatches`, `CollectDirectDQMatches`, `ApplyReshapeTransposeFusions`, `ApplyDirectDQFusions`), the matching and rewriting functions are still very large (each 100-200+ lines) with deeply nested validation logic. The two `Apply*` functions share significant duplicated logic for:
- Building `MatMulNBits` attributes
- Handling zero-point elision logic
- Gemm bias preservation
- Node removal

**Recommendation:** Extract shared rewriting logic into a common helper (e.g., `CreateMatMulNBitsNode()`, `HandleZeroPointSemantics()`, `RemoveMatchedNodes()`).

### 8. No Unit Tests
There are no test files in the PR. The existing `qdq_matmulnbits_transformer_test.cc` tests the standard QDQ→MatMulNBits path but doesn't cover the new DQ→Reshape→Transpose→MatMul pattern or the direct DQ(axis=0)→MatMul pattern. The fusion logic includes complex weight transposition, zero-point packing, and Gemm bias handling that all need numeric verification.

**This is the biggest gap in the PR.** Test coverage should include:
- Pattern 1: DQ(3D)+Reshape+Transpose+MatMul → MatMulNBits (with/without Cast, with/without zero points)
- Pattern 2: DQ(2D, axis=0)+MatMul → MatMulNBits
- Negative tests: incorrect axis, non-power-of-2 block_size, wrong reshape dims, invalid transpose perm, non-constant weights
- Gemm bias preservation (1-D bias [N])
- Zero-point semantics: default zp=8 elision, explicit zeros when DQ omits zp
- Numeric correctness: fused vs. unfused output comparison

### 9. Potential Integer Overflow in Size Calculations
Several size calculations use `int64_t` multiplication without overflow checks:
- `K = quant_num * bs_dim` (line ~applying fusion for pattern 1)
- `N * k_blocks * blob_bytes` in `TransposePackWeightsAxis0`
- `N * quant_num` for zero-point size

While these are typically bounded by model sizes, defensive overflow checks (or use of `SafeInt`) would be consistent with ORT's coding patterns.

### 10. `const_cast` Usage
Multiple `const_cast<NodeArg*>` calls appear when building `mnb_inputs` and `mnb_outputs`. While this pattern exists in other ORT transformers, it's worth noting as a code smell. The ORT graph API could perhaps be improved to avoid this.

### 11. Missing `#if !defined(ORT_MINIMAL_BUILD)` Guard in New File
The new `dq_matmulnbits_fusion.cc` doesn't wrap its implementation in `#if !defined(ORT_MINIMAL_BUILD)`. Other optimizer files (e.g., `matmul_nbits_fusion.cc`) use this guard. Since this transformer uses `graph_utils`, `GetConstantInitializer`, etc., which are not available in minimal builds, the file will fail to compile in `ORT_MINIMAL_BUILD` configurations.

---

## Minor Issues

### 13. Typo in Branch Name
Branch name is `MatMulNBits-trasformation-nv-trt-rtx` — "trasformation" should be "transformation". (Cosmetic, no code impact.)

### 14. `GetPackedUint4Element` Could Use Bounds Checking
The utility function extracts a 4-bit nibble from a packed byte array by index but performs no bounds checking. In debug builds, an assertion would help catch invalid indices.

### 15. `HasRank2Shape` Validates Only for Pattern 2
`HasRank2Shape` is used for direct DQ matches but not for Pattern 1 scale/zp validation (which uses inline checks). Consider reusing the helper for consistency.

### 16. Comment in `TransposePackWeightsAxis0` References "row-major UINT4"
The comment says "row-major UINT4 with quantization along K (axis=0), shape [K, N]" but UINT4 packing in ONNX stores two values per byte in little-endian nibble order. The comment should clarify the nibble ordering convention being assumed.

----
BTW, the recommended changes from AI: https://github.com/microsoft/onnxruntime/commit/22dce1aab5a854bc3fbb7776c3cc9c5d59c70604

anujj:
Thanks for the thorough review @tianleiwu and @xadupre!

Addressed most of the feedback in the latest commit (`40207a9`). Adopted the recommended SafeInt, `ORT_MINIMAL_BUILD`, naming, and config-key changes from the [suggested commit](https://github.com/microsoft/onnxruntime/commit/22dce1aab5a854bc3fbb7776c3cc9c5d59c70604) with one fix: the `AddConfigEntry` call is now guarded to avoid failing when the user has already set the config key explicitly.

## What we addressed

| # | Concern | What we did |
|---|---------|-------------|
| 3 | `bool` param on public `GenerateTransformers()` API doesn't scale | Replaced with session config key `kOrtSessionOptionsEnableDQMatMulNBitsFusion`. Auto-set by `InferenceSession` when NvTensorRTRTX EP is registered. Adopted from suggested commit, with added guard so `AddConfigEntry` doesn't fail if user already set the key. |
| 5 | WebNN naming in node names, comments, log strings | Renamed `"WebNNFusedMatMulNBits"` → `"DQFusedMatMulNBits"`, cleaned up header comments and log strings. Adopted from suggested commit. |
| 6 | `intra_op_thread_pool_` stored but never used | Removed parameter and member from constructor, header, and all call sites. Adopted from suggested commit. |
| 8 | No unit tests | Added `dq_matmulnbits_fusion_test.cc` with 12 tests: Pattern 1 (no ZP, default ZP8, non-default ZP, Cast, Gemm+bias, Gemm), Pattern 2 (no ZP, default ZP8, Gemm+bias), negative tests (wrong axis, non-const weight), and config gating test. All pass. Based on test structure from suggested commit, with `MakePackedUint4` fixed to return `std::vector<UInt4x2>`. |
| 9 | Integer overflow in size calculations | Replaced all `static_cast<size_t>(X * Y)` with `SafeInt<size_t>(X) * Y` throughout packing and matching functions. Adopted from suggested commit. |
| 11 | Missing `#if !defined(ORT_MINIMAL_BUILD)` guard | Added to both `.h` and `.cc`, and moved `#include` inside the guard in `graph_transformer_utils.cc`. Adopted from suggested commit. |
| 14 | `GetPackedUint4Element` has no bounds check | Added `num_elements` parameter with `ORT_ENFORCE` bounds check. All callers updated. Adopted from suggested commit. |
| 16 | UINT4 nibble ordering comment unclear | Added clarification: "even indices in the low nibble, odd indices in the high nibble". Adopted from suggested commit. |
| — | `ORT_ENFORCE` when fusion enabled but `DISABLE_CONTRIB_OPS` defined | Done (now checks the config key in the `#else` branch). |
| — | Split long `ApplyImpl` into focused helpers | Done in previous commit (4 helper functions + shared `ValidateGemmForFusion`). |
| — | clang-format | Ran `lintrunner` on all 8 changed files — zero lint issues. |

## What we are NOT addressing (and why)

| # | Concern | Why we're not changing it |
|---|---------|--------------------------|
| 4 | Fusion at Level1 — fused node may have empty EP type | This is intentional. The fusion runs at Level1 *before* EP partitioning so the resulting `MatMulNBits` node can be claimed by NvTensorRTRTX EP during `GetCapability()`. The node inherits an empty EP type at Level1, which is correct — it gets assigned during the normal partitioning flow after Level1 transformers complete. |
| 7 | Further extraction of shared rewriting logic between `Apply*` functions | The two `Apply*` functions share some boilerplate (attrs, bias, removal) but their core packing logic differs significantly (3D memcpy vs 2D transpose-pack). Further splitting risks over-abstraction. Will consider extracting `CreateMatMulNBitsNode()` / `HandleZeroPointSemantics()` in a follow-up PR. |


## Notes

Most of the changes in this commit were adopted from the [AI-suggested commit `22dce1a`](https://github.com/microsoft/onnxruntime/commit/22dce1aab5a854bc3fbb7776c3cc9c5d59c70604) by @tianleiwu, with the following modifications:
- **`AddConfigEntry` safety**: Added a guard in `inference_session.cc` to check if the config key is already set before calling `AddConfigEntry`, preventing `Initialize()` failure when a user explicitly overrides the config.
- **Test fix**: Changed `MakePackedUint4` to return `std::vector<UInt4x2>` (instead of `std::vector<uint8_t>`) to match the `MakeInitializer<UInt4x2>` API.
- Ran lintrunner to address the formatting issues

tianleiwu (approved):
## Remaining Issues

### 2. Pattern 1: Scale Shape Not Explicitly Validated During Matching (P2 — minor)

In `CollectReshapeTransposeMatches`, the scale initializer's *shape* is never explicitly checked during matching (lines 259–265). The code verifies the scale is a constant initializer with float/float16 data type, but doesn't validate that its shape matches `[N, blocks, 1]` or `[N, blocks]`. Contrast with Pattern 2 which has `if (!HasRank2Shape(*scale_const_tp, k_blocks, N)) continue;`.

The shape is implicitly validated later in `ApplyReshapeTransposeFusions` when `scale_src.size()` is compared to `N * quant_num` (line 521). However, a malformed scale tensor that happens to have the right element count but wrong rank/dimensions would pass matching and could produce incorrect results.

**Recommendation:** Add an explicit shape check on the scale tensor in Pattern 1 matching (e.g., verify `dims_size == 3` and dims match `[N, blocks, 1]`).

### 3. Pattern 1: Cast Node Only Accepted If Same Input/Output Type (P2 — suspicious logic)

Lines 349–354 in `CollectReshapeTransposeMatches`:
```cpp
if (cast_in->TypeAsProto()->tensor_type().elem_type() !=
    cast_out->TypeAsProto()->tensor_type().elem_type()) {
  continue;
}
```

A Cast node that casts to the same type is a no-op. The test (`Pattern1_MatMul_WithCast`) creates a Cast with `to=1` (FLOAT→FLOAT), which is also a no-op. This means the fusion only accepts no-op Casts — which is fine as a safety measure, but a real ONNX model is unlikely to have no-op Cast nodes, so this pattern arm may never match in practice.

**Question:** Is this intentional (accepting only identity Casts) or should the fusion also handle real Cast nodes (e.g., float16→float32) with appropriate type handling?

### 4. Test Coverage Gaps (P2 — missing test coverage)

The test suite is a significant improvement (11 tests), but has some gaps:

- **No numeric verification:** Tests only check graph structure (node counts, attribute values, input counts). No test verifies that the fused `MatMulNBits` output matches the unfused `DQ→MatMul` output. The weight transposition, scale repacking, and zero-point packing logic could have subtle bugs that wouldn't be caught by structural checks alone.
- **No Pattern 2 non-default ZP test:** There's a test for Pattern 2 with default ZP8, but no test for Pattern 2 with non-default ZP values (to exercise `TransposePackZPAxis0`).

**Recommendation:** Add at least one numeric correctness test that runs the graph before and after fusion and compares outputs. Also add a Pattern 2 non-default ZP test.

### 5. `AddConfigEntry` Empty-String Edge Case (P3 — minor)

In `inference_session.cc`:
```cpp
if (session_options_.config_options.GetConfigOrDefault(
        kOrtSessionOptionsEnableDQMatMulNBitsFusion, "") == "") {
```

This checks for empty string to detect "not set". If the user explicitly sets the key to "0", it's respected (won't override). However, if the user explicitly sets the key to `""` (empty string), the code would override it with "1". This is unlikely in practice but worth a note.

---
**Approve with minor suggestions.** The code is well-structured, properly guarded, and has good test coverage for a first iteration. The config-driven gating and SafeInt changes are solid improvements. The remaining issues are non-blocking and can be addressed in this PR or a follow-up.

PR #27469, opened 2026-02-26
Fix Cast node naming collisions and opset 10 Resize in float16 conversion

### LINKED ISSUE #14827: working model with Resize node becomes invalid after using convert_float_to_float16
### Describe the issue

After converting a model to FP16 internally using `onnxruntime.transformers.float16.convert_float_to_float16`, the model can be loaded with `onnx.load_model` but fails to load with `onnxruntime.InferenceSession`. The model was valid before conversion and both ONNX and ORT could load it:

```shell
> python3 fp16-repro.py 
ONNX load succeeded for original model
ORT load succeeded for original model
saved optimized model
ONNX load succeeded for optimized model
ORT load failed for optimized model
Traceback (most recent call last):
  File "/home/ssube/onnx-repro/fp16-repro.py", line 42, in <module>
    sess = InferenceSession(opt_file, providers=["CPUExecutionProvider"])
  File "/tmp/ort_env/lib/python3.10/site-packages/onnxruntime/capi/onnxruntime_inference_collection.py", line 360, in __init__
    self._create_inference_session(providers, provider_options, disabled_optimizers)
  File "/tmp/ort_env/lib/python3.10/site-packages/onnxruntime/capi/onnxruntime_inference_collection.py", line 397, in _create_inference_session
    sess = C.InferenceSession(session_options, self._model_path, True, self._read_config_from_model)
onnxruntime.capi.onnxruntime_pybind11_state.InvalidGraph: [ONNXRuntimeError] : 10 : INVALID_GRAPH : Load model from ./optimized.onnx failed:This is an invalid model. Type Error: Type 'tensor(float16)' of input parameter (onnx::Resize_9300) of operator (Resize) in node (Resize_4928) is invalid.
```

This appears to be related to https://github.com/microsoft/onnxruntime/issues/8327 and https://github.com/microsoft/onnxruntime/issues/2848, both of which were closed as errors in third-party models, but this model _was_ valid before calling `convert_float_to_float16`.

The error seems to come from the FP16 type incorrectly being applied to `Resize` nodes: `This is an invalid model. Type Error: Type 'tensor(float16)' of input parameter (onnx::Resize_9300) of operator (Resize) in node (Resize_4928) is invalid.`

I've tested with the `CPUExecutionProvider`, which should be the most compatible?, as well as the `CUDAExecutionProvider` and they both report the same error.

https://github.com/onnx/onnxmltools/issues/361#issuecomment-927773054 suggests there might be an overly-aggressive conversion happening and suggests raising an issue here:

>  It assumes all operators in graph have a runtime which supports float16. That's not the case for onnxruntime. The model may be valid but cannot be run unless onnxruntime supports float16 for all operators in your graph. You need to use a different runtime or raise an issue on onnxruntime repository with the list of operators you need to support float16.

I don't see any way to exclude the `Resize` nodes during conversion.

### To reproduce

Download `https://huggingface.co/runwayml/stable-diffusion-v1-5/resolve/onnx/vae_decoder/model.onnx` ahead-of-time or within the notebook/script:

```python
!wget https://huggingface.co/runwayml/stable-diffusion-v1-5/resolve/onnx/vae_decoder/model.onnx

from onnx import load_model, save_model
from onnxruntime import InferenceSession
from onnxruntime.transformers.float16 import convert_float_to_float16
from traceback import print_exception

init_file = "./model.onnx"
opt_file = "./optimized.onnx"

# verify original model
try:
    model = load_model(init_file)
    print("ONNX load succeeded for original model")
except Exception as err:
    print("ONNX load failed for original model")
    print_exception(type(err), err, err.__traceback__)

# reload with ORT
try:
    sess = InferenceSession(init_file, providers=["CPUExecutionProvider"])
    print("ORT load succeeded for original model")
except Exception as err:
    print("ORT load failed for original model")
    print_exception(type(err), err, err.__traceback__)

# load and convert
model = load_model(init_file)
optimized = convert_float_to_float16(model, keep_io_types=True, force_fp16_initializers=False, disable_shape_infer=True) # TODO: test with initializers=false
save_model(optimized, opt_file, save_as_external_data=True, all_tensors_to_one_file=True)
print("saved optimized model")

# reload with ONNX
try:
    model = load_model(opt_file)
    print("ONNX load succeeded for optimized model")
except Exception as err:
    print("ONNX load failed for optimized model")
    print_exception(type(err), err, err.__traceback__)

# reload with ORT
try:
    sess = InferenceSession(opt_file, providers=["CPUExecutionProvider"])
    print("ORT load succeeded for optimized model")
except Exception as err:
    print("ORT load failed for optimized model")
    print_exception(type(err), err, err.__traceback__)
```

Using what appear to be the latest packages (fresh venv created for this repro):

```
pip3 list --local | grep -e onnx -e torch
onnx                     1.13.1
onnxruntime-gpu          1.14.0
torch                    1.13.1
torchaudio               0.13.1
torchvision              0.14.1
```

Forcing FP16 initializers, `force_fp16_initializers=True`, does not seem to make a difference.

Converting a model that does not use the `Resize` node, like the corresponding encoder https://huggingface.co/runwayml/stable-diffusion-v1-5/resolve/onnx/vae_encoder/model.onnx, does work correctly and can be loaded by both ONNX and ORT.

### Urgency

Not urgent, but limiting usage on cards with low VRAM (https://github.com/ssube/onnx-web/issues/121#issuecomment-1445141334). 

I'm happy to start a PR if this is the correct place and not user error.

### Platform

Linux

### OS Version

Linux compute-infer-1 5.15.0-60-generic #66-Ubuntu SMP Fri Jan 20 14:29:49 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux
Ubuntu 22.04.1 LTS

### ONNX Runtime Installation

Released Package

1.14.0

### ONNX Runtime API

Python

### Architecture

X64

### Execution Provider

Default CPU

ORT 1.14.0, PyTorch 1.13.1

Rishi-Dave:
## Summary
- Fix Cast node naming collisions in `convert_float_to_float16` when nodes have empty names (common in PyTorch exports)
- Fix `ALWAYS_FLOAT_INPUTS` for opset 10 Resize where scales at index 1 was unprotected
- Add dedicated test suite for float16 conversion (`test_float16.py`, 8 tests)

## Motivation
Fixes #14827

When `convert_float_to_float16` processes models with unnamed nodes (empty `node.name`, very common in PyTorch/TensorFlow-exported ONNX models), the generated Cast node names collide. For example, multiple Resize nodes all produce Cast nodes named `"_input_cast_2"` and output tensors named `"_input_cast_2"`, corrupting the graph with duplicate names.

Additionally, the `ALWAYS_FLOAT_INPUTS` dict only protected Resize scales at index 2 (opset 11+ layout: `[X, roi, scales, sizes]`), but opset 10 Resize has scales at index 1 (`[X, scales]`), leaving it unprotected.

## Changes
**`onnxruntime/python/tools/transformers/float16.py`** (11 lines changed):
- Use unique tensor names (`input_name`/`output`) as the base for generated Cast node and output names, instead of potentially-empty `node.name`
- Add index 1 to `ALWAYS_FLOAT_INPUTS["Resize"]` to protect opset 10 scales
- Fix misleading comment ("change current node's input name" → "output name")

**`onnxruntime/test/python/transformers/test_float16.py`** (new file, 8 tests):
- `test_resize_opset11_cast_naming_unique` — multiple unnamed Resize nodes produce unique Cast names
- `test_resize_opset11_scales_initializer_stays_fp32` — scales initializer preserved as float32
- `test_resize_opset10_scales_initializer_stays_fp32` — opset 10 scales protected at index 1
- `test_resize_opset10_multiple_unnamed_unique_names` — opset 10 naming uniqueness
- `test_blocked_node_cast_naming_unique` — blocked op nodes (Upsample) also get unique Cast names
- `test_resize_with_op_block_list` — Resize in op_block_list still produces unique names
- `test_data_input_converted_to_fp16` — data tensor correctly converts to fp16
- `test_force_fp16_initializers` — force flag overrides protection

## Test Plan
- All 8 new tests pass locally (`python -m unittest test_float16.TestFloat16Conversion -v`)
- Existing `test_gpt2_past_fp16` test passes (no regression in existing float16 behavior)
- `ruff check` passes on both files

### DIFF
--- a/onnxruntime/python/tools/transformers/float16.py
+++ b/onnxruntime/python/tools/transformers/float16.py
@@ -146,7 +146,7 @@ DEFAULT_OP_BLOCK_LIST = [
 
 # Some operators has data type fixed as float for some inputs. Key is op_type, value is list of input indices
 # Note that DirectML allows float16 gamma and beta in GroupNorm. Use force_fp16_inputs parameter could overwrite this.
-ALWAYS_FLOAT_INPUTS = {"Resize": [1, 2], "GroupNorm": [1, 2], "SkipGroupNorm": [1, 2]}
+ALWAYS_FLOAT_INPUTS = {"Resize": [2], "GroupNorm": [1, 2], "SkipGroupNorm": [1, 2]}
 
 
 class InitializerTracker:
@@ -237,6 +237,14 @@ def convert_float_to_float16(
         node_block_list = []
     op_block_list = set(op_block_list)
     node_block_list = set(node_block_list)
+
+    # Build opset-aware always_float_inputs: Resize input layout differs between opset 10 and 11+.
+    # Opset 10: [X, scales] — scales at index 1 must stay float32.
+    # Opset 11+: [X, roi, scales, sizes] — scales at index 2 must stay float32; roi (index 1) allows fp16.
+    onnx_opset = max((o.version for o in model.opset_import if o.domain in ("", "ai.onnx")), default=11)
+    always_float_inputs = dict(ALWAYS_FLOAT_INPUTS)
+    if onnx_opset <= 10:
+        always_float_inputs["Resize"] = [1]
 
     logger.debug(
         f"fp16 parameters: min_positive_val={min_positive_val} max_finite_val={max_finite_val} keep_io_types={keep_io_types} disable_shape_infer={disable_shape_infer} op_block_list={op_block_list} node_block_list={node_block_list} force_fp16_initializers={force_fp16_initializers}"
@@ -334,7 +342,7 @@ def convert_float_to_float16(
                         if input_name in fp32_initializers:
                             # For Resize/GroupNorm, only the first input can be float16
                             use_fp32_weight = is_node_blocked or (
-                                i in ALWAYS_FLOAT_INPUTS.get(n.op_type, [])
+                                i in always_float_inputs.get(n.op_type, [])
                                 and i not in force_fp16_inputs_dict.get(n.op_type, [])
                             )
                             fp32_initializers[input_name].add_node(n, use_fp32_weight)
@@ -371,7 +379,7 @@ def convert_float_to_float16(
                                 n.attribute.extend([helper.make_attribute("dtype", TensorProto.FLOAT16)])
 
                         # For Resize/GroupNorm, attribute data type cannot be changed
-                        if n.op_type not in ALWAYS_FLOAT_INPUTS or n.op_type in force_fp16_inputs_dict:
+                        if n.op_type not in always_float_inputs or n.op_type in force_fp16_inputs_dict:
                             for attr in n.attribute:
                                 next_level.append(attr)  # noqa: PERF402
                         else:
@@ -417,7 +425,7 @@ def convert_float_to_float16(
     # Some operators have data type fixed as float for some input. Add a float16 to float cast for those inputs.
     for node in mixed_float_type_node_list:
         for i, input_name in enumerate(node.input):
-            if i not in ALWAYS_FLOAT_INPUTS[node.op_type] or i in force_fp16_inputs_dict.get(node.op_type, []):
+            if i not in always_float_inputs[node.op_type] or i in force_fp16_inputs_dict.get(node.op_type, []):
                 continue
             for value_info in value_info_list:
                 if input_name == value_info.name:
--- a/onnxruntime/test/python/transformers/test_float16.py
+++ b/onnxruntime/test/python/transformers/test_float16.py
@@ -24,7 +24,7 @@ def _make_resize_model_opset11(num_resize_nodes=2, use_empty_names=True):
     """Create a minimal ONNX model with multiple Resize nodes (opset 11+).
 
     Resize opset 11+: inputs are [X, roi, scales, sizes].
-    Scales (index 2) and roi (index 1) must stay float32 per ALWAYS_FLOAT_INPUTS.
+    Scales (index 2) must stay float32 per ALWAYS_FLOAT_INPUTS; roi (index 1) allows fp16.
     """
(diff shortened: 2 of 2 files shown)


Rishi-Dave:

tianleiwu:
The core fix (using tensor names instead of potentially-empty node.name for Cast node naming) is correct and well-tested.

One concern:

Opset 11+ roi over-protection: Adding index 1 to ALWAYS_FLOAT_INPUTS["Resize"] correctly protects opset 10 scales, but also forces opset 11+ roi to stay fp32 — even though the ONNX spec allows roi to be fp16 (T2 constraint). Practically harmless since roi is usually empty, but technically imprecise. Suggest at minimum adding a comment, or ideally making protection opset-aware.

Other minor notes: shared-input edge case not fully addressed (not a regression), and the naming convention change could theoretically affect downstream tooling that pattern-matches Cast node names.

Rishi-Dave:
Thanks for the thorough review @tianleiwu! Great catch on the opset 11+ roi over-protection.

I've pushed a fix that makes the protection opset-aware:
- Reverted `ALWAYS_FLOAT_INPUTS["Resize"]` back to `[2]` (opset 11+ default)
- Added logic inside `convert_float_to_float16` that detects the ONNX opset version from `model.opset_import` and adjusts at runtime:
  - **Opset ≤ 10** (`[X, scales]`): protects index 1 (scales)
  - **Opset 11+** (`[X, roi, scales, sizes]`): protects index 2 (scales) only; roi at index 1 is left to fp16 per the ONNX spec

Updated the test to verify that opset 11+ roi is not over-protected.

Re: the shared-input edge case and naming convention — agreed these are pre-existing behaviors, not regressions from this PR. Happy to address them in a follow-up if you'd like.

tianleiwu (approved):

Issue #27329, opened 2026-02-12
TensorRT-RTX EP: Enabling CUDA Graph seems ineffective when running a session from a precompiled engine


matpoudret:
### Describe the issue

I noticed that CUDA Graph launch does not occur when running a TensorRT-RTX EP session from a precompiled (AOT build) TensorRT-RTX engine. This issue may be related to #26929.

In the following snippet from `NvExecutionProvider::CreateNodeComputeInfoFromGraph`, the CUDA Graph strategy is explicitly set to `kWHOLE_GRAPH_CAPTURE`. Indeed, according to the [TensorRT-RTX API documentation](https://docs.nvidia.com/deeplearning/tensorrt-rtx/latest/_static/cpp-api/classnvinfer1_1_1_i_runtime_config.html#a0785a3e75ae5b2e0e5ffa59f518aed57), the default strategy is `kDISABLED`.

https://github.com/microsoft/onnxruntime/blob/a3749f13536bd25e89ba8f8a5ae21e88b9337057/onnxruntime/core/providers/nv_tensorrt_rtx/nv_execution_provider.cc#L2951-L2962

It seems that this CUDA Graph strategy override does not apply when the session is run from a precompiled engine.

### To reproduce

1. Unzip [repro.tar.gz](https://github.com/user-attachments/files/25264345/repro.tar.gz)

```bash
tar -xzf repro.tar.gz
cd repro/
```

2. _(Expected behavior)_ When running `repro.py` without relying on precompiled TensorRT-RTX engine (script default), `--num-runs` calls to `cudaGraphLaunch`  occur:

```
python repro.py --num-runs 5
```

You may profile these calls by using NVIDIA Nsight Systems:

<img width="1084" height="717" alt="Image" src="https://github.com/user-attachments/assets/96f01226-32c4-49fb-b751-822723de8794" /><br>

3. _(Issue)_ When running `repro.py` with the `--use-precompiled-engine` option, CUDA kernels are launched directly without relying on `cudaGraphLaunch`. This behavior may lead to CPU overhead associated with launching CUDA kernels sequentially:

```
python repro.py --use-precompiled-engine --num-runs 5
```

Here is the corresponding NVIDIA Nsight Systems profile:

<img width="1084" height="717" alt="Image" src="https://github.com/user-attachments/assets/d4004686-c2d2-4039-bac4-8aa88093adff" />

### Platform

Linux

### OS Version

Rocky Linux 8.10

### ONNX Runtime Installation

[...]

PR #27477, opened 2026-02-27
Fix for Cuda Graph in pre-compiled path

umangb-09:
### Description
setCudaGraphStrategy(kWHOLE_GRAPH_CAPTURE) was present in the dynamic engine build path (CreateNodeComputeInfoFromGraph) but missing from the precompiled/AOT engine path (CreateNodeComputeInfoFromEPContext). Since TRT RTX defaults the CUDA Graph strategy to kDISABLED, CUDA Graph capture never occurred when loading precompiled engines. Applied the same setCudaGraphStrategy call (guarded by the existing TRT_MAJOR_RTX >= 1.3 version check) to the precompiled path to match the dynamic path behavior

### Motivation and Context
Fixes [#27329](https://github.com/microsoft/onnxruntime/issues/27329) — users reported that cudaGraphLaunch was not occurring when using precompiled (AOT-built) TensorRT-RTX engines, causing individual kernel launches and unnecessary CPU overhead instead of batched graph execution.


umangb-09:
@chilo-ms help review this. Thanks!

chilo-ms (approved):

PR #27479, opened 2026-02-27
[QNN EP] Enable offline x64 compilation with memhandle IO type

derdeljan-msft:
### Description

Enable offline compilation for QNN EP with MEMHADNLE IO type. It was previously enabled only on ARM because the QNN EP was loading rpcmem library (only available on ARM drivers), which is not actually used for the compilation (it is required only for inference to allocate the shared memory). 

Ensured that the MEMHANDLE IO type is correctly set regardless of the way how `QnnTensorWrapper` is created (either through factory function or by creating it directly). This ensures that mem type will be correctly configured regardless of the op builder implementation.

### DIFF
--- a/onnxruntime/test/providers/qnn/qnn_model_wrapper_test.cc
+++ b/onnxruntime/test/providers/qnn/qnn_model_wrapper_test.cc
@@ -13,6 +13,7 @@
 #include "core/providers/qnn/builder/qnn_model_wrapper.h"
 #include "core/providers/qnn/builder/qnn_def.h"
 #include "test/util/include/default_providers.h"
+#include "test/util/include/test_environment.h"
 
 using namespace onnxruntime;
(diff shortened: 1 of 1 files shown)


onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.cc
```diff
@@ -105,6 +101,12 @@ bool QnnModelWrapper::AddTensorWrapper(QnnTensorWrapper&& tensor_wrapper) {
     return true;
   }
 
+  Qnn_TensorMemType_t mem_type = QNN_TENSORMEMTYPE_RAW;
```
yuslepukhin: Should this be a function?

yuslepukhin (changes requested):
Need unit test.

PR #27482, opened 2026-02-27
Merge ConstantOfShape and Cast in clean_graph optimization

### LINKED ISSUE #27481: ConstantOfShape → Cast not merged in clean_graph optimization
### Description

In `BertOnnxModel.clean_graph()` (onnx_model_bert.py), the Concat path simplification for attention mask shapes produces:

```
input_ids → Shape → ConstantOfShape → Cast → EmbedLayerNormalization/ReduceSum
```

The `ConstantOfShape → Cast` sequence is not merged — the ConstantOfShape produces float values and the Cast converts them to the target type (typically int64). This is an unnecessary extra node that can be eliminated by updating the ConstantOfShape's `value` attribute to use the target dtype directly.

There is an existing TODO in the code at line 271:
```python
# TODO: merge ConstantOfShape -->Cast to ConstantOfShape (need update the data type of value)
```

### Expected behavior

After clean_graph, the Cast node should be removed and ConstantOfShape should directly produce the target type:

```
input_ids → Shape → ConstantOfShape(value=int64(1)) → EmbedLayerNormalization/ReduceSum
```

### Impact

This is a graph cleanup optimization — removes unnecessary Cast nodes from attention mask paths in BERT-family models. The optimization is cosmetic (no performance impact for most runtimes) but produces cleaner optimized graphs.

Rishi-Dave:
## Summary
- Resolve the TODO in `BertOnnxModel.clean_graph()` by folding the `Cast` node into `ConstantOfShape`
- Update the `value` attribute dtype so `ConstantOfShape` directly produces the target type
- Add first dedicated test suite for `clean_graph` with 3 tests

## Motivation
Fixes #27481

In the attention mask shape optimization path, `clean_graph()` simplifies a `Concat → Unsqueeze → Gather` chain to a direct `Shape → ConstantOfShape` connection, but leaves a redundant `ConstantOfShape(float) → Cast(to=int64)` sequence. The existing TODO (line 271) noted this should be merged.

## Changes
- **onnx_model_bert.py**: After the Concat path simplification, update `ConstantOfShape`'s `value` attribute to the Cast target dtype using `helper.tensor_dtype_to_np_dtype`, then redirect consumers from the Cast output to the ConstantOfShape output. The Cast node is collected for removal. (+2 imports: `numpy`, `numpy_helper`)
- **test_clean_graph.py** (new): 3 tests covering the merge (Cast removed, ConstantOfShape produces int64), fill value preservation (0 → int64(0)), and Concat path simplification (Concat/Gather pruned).

## Test Plan
- [x] 3 new tests pass: `python -m unittest test_clean_graph.TestCleanGraph -v`
- [x] Existing `test_attention_fusion` tests pass
- [x] `ruff check` clean on both files

### DIFF
--- a/onnxruntime/python/tools/transformers/onnx_model_bert.py
+++ b/onnxruntime/python/tools/transformers/onnx_model_bert.py
@@ -267,7 +267,7 @@ class BertOnnxModel(OnnxModel):
             #          |                                                     |
             #          |                                                     v
             #          +----> Shape --> Gather(indices=1) --> Unsqueeze--->  Concat --> ConstantOfShape -->Cast --> EmbedLayerNormaliation/ReduceSum
-            # After:
+            # After (Concat path simplified, Cast merged into ConstantOfShape):
             #  input_ids --> Shape --> ConstantOfShape --> EmbedLayerNormalization/ReduceSum
             op_input_id = {"EmbedLayerNormalization": 1, "ReduceSum": 0, "Attention": 3}
             if node.op_type in op_input_id:
@@ -300,20 +300,23 @@ class BertOnnxModel(OnnxModel):
                         # Merge ConstantOfShape → Cast: update the value attribute dtype
                         # so ConstantOfShape directly produces the target type.
                         cast_to_type = OnnxModel.get_node_attribute(cast, "to")
-                        if cast_to_type is not None:
-                            cos_value = constant_of_shape.attribute[0].t
-                            fill_val = numpy_helper.to_array(cos_value).flat[0]
+                        cos_tensor = OnnxModel.get_node_attribute(constant_of_shape, "value")
+                        if cast_to_type is not None and cos_tensor is not None:
+                            fill_val = numpy_helper.to_array(cos_tensor).flat[0]
                             np_dtype = helper.tensor_dtype_to_np_dtype(cast_to_type)
                             new_val = numpy_helper.from_array(np.array([fill_val], dtype=np_dtype))
-                            constant_of_shape.attribute[0].t.CopyFrom(new_val)
+                            for i, attr in enumerate(constant_of_shape.attribute):
+                                if attr.name == "value":
+                                    constant_of_shape.attribute[i].CopyFrom(helper.make_attribute("value", new_val))
+                                    break
                             self.replace_input_of_all_nodes(cast.output[0], constant_of_shape.output[0])
                             nodes_to_remove.append(cast)
 
                         output_name_to_node = self.output_name_to_node()
 
             if node.op_type == "Attention":
-                # Before:
-                #   input_ids --> Shape -->ConstantOfShape -->Cast --> ReduceSum --> Attention
+                # Before (Cast present or already merged into ConstantOfShape):
+                #   input_ids --> Shape --> ConstantOfShape [--> Cast] --> ReduceSum --> Attention
                 # After:
                 #   remove this path, and remove the optional mask_index input of Attention node.
                 parent_nodes = self.match_parent_path(
@@ -322,6 +325,14 @@ class BertOnnxModel(OnnxModel):
                     [3, 0, 0, 0],
                     output_name_to_node,
                 )
+                if parent_nodes is None:
+                    # Also try merged pattern (Cast already folded into ConstantOfShape).
+                    parent_nodes = self.match_parent_path(
+                        node,
+                        ["ReduceSum", "ConstantOfShape", "Shape"],
+                        [3, 0, 0],
+                        output_name_to_node,
+                    )
                 if parent_nodes is not None:
                     if parent_nodes[-1].input[0] == self.graph().input[0].name:
                         attention_node = helper.make_node(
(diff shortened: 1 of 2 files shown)


tianleiwu:
Please run [lintrunner -a](https://github.com/microsoft/onnxruntime/blob/12814f716de8b06b21d51e82617416e6182b5ffa/docs/Coding_Conventions_and_Standards.md?plain=1#L155) to format code.

## Issues

### 1. [High] Regression: Attention cleanup pattern broken by Cast merge

**File**: `onnxruntime/python/tools/transformers/onnx_model_bert.py` (lines ~314–340)

The Attention cleanup block (immediately below the Cast-merge code) matches the parent path:

```python
["ReduceSum", "Cast", "ConstantOfShape", "Shape"]
```

from `Attention.input[3]`. However, when a `ReduceSum` node is processed **earlier** in the same loop, the new merge logic calls:

```python
self.replace_input_of_all_nodes(cast.output[0], constant_of_shape.output[0])
```

This redirects `ReduceSum.input[0]` from `cast.output[0]` to `constant_of_shape.output[0]`. When the loop later reaches the `Attention` node:

- `Attention.input[3]` → `ReduceSum` ✓
- `ReduceSum.input[0]` → now `ConstantOfShape` (not `Cast`) ✗

The 4-node pattern **fails**, so the Attention node's `mask_index` input is **not** removed. In the original code (before this PR), the Cast node was never removed, so this pattern always matched.

**Impact**: The `Attention` mask optimization is silently skipped. The graph is still correct but retains an unnecessary `ReduceSum → ConstantOfShape → Shape` subgraph feeding the mask input.

**Suggestion**: Add a second match attempt for the post-merge case:

```python
if node.op_type == "Attention":
    # Try original pattern (Cast still present)
    parent_nodes = self.match_parent_path(
        node,
        ["ReduceSum", "Cast", "ConstantOfShape", "Shape"],
        [3, 0, 0, 0],
        output_name_to_node,
    )
    # Also try merged pattern (Cast already folded into ConstantOfShape)
    if parent_nodes is None:
        parent_nodes = self.match_parent_path(
            node,
            ["ReduceSum", "ConstantOfShape", "Shape"],
            [3, 0, 0],
            output_name_to_node,
        )
    if parent_nodes is not None:
        if parent_nodes[-1].input[0] == self.graph().input[0].name:
            # ... create new Attention node without mask_index ...
```

### 2. [Low] Unsafe `attribute[0]` access on `ConstantOfShape`

**File**: `onnxruntime/python/tools/transformers/onnx_model_bert.py` (the new merge block)

```python
cos_value = constant_of_shape.attribute[0].t
```

The ONNX spec defines `value` as an **optional** attribute of `ConstantOfShape` (defaults to `float32(0)`). If the attribute list is empty (no explicit `value` set), `attribute[0]` raises an `IndexError`.

In practice, the matched pattern always has an explicit value, so this is safe today. But using `OnnxModel.get_node_attribute(constant_of_shape, "value")` would be more robust and consistent with how `cast_to_type` is retrieved on the very next line.

**Suggestion**:
```python
cos_tensor = OnnxModel.get_node_attribute(constant_of_shape, "value")
if cos_tensor is not None:
    fill_val = numpy_helper.to_array(cos_tensor).flat[0]
    ...
```

### 3. [Low] No test covering the Attention cleanup interaction

**File**: `onnxruntime/test/python/transformers/test_clean_graph.py`

All three tests use `ReduceSum` as the terminal consumer. None test a graph with `ReduceSum → Attention` where the Attention pattern should also fire. A fourth test that includes an `Attention` node consuming the `ReduceSum` output would catch the regression described in Issue #1.

### 4. [Low] Stale comment in Attention cleanup block

**File**: `onnxruntime/python/tools/transformers/onnx_model_bert.py` (lines ~315–317)

```python
# Before:
#   input_ids --> Shape -->ConstantOfShape -->Cast --> ReduceSum --> Attention
```

After the Cast merge (which runs earlier in the same loop), the actual graph state when this block executes is:

```
input_ids --> Shape --> ConstantOfShape --> ReduceSum --> Attention
```

The comment should be updated to reflect both possible states (pre-merge and post-merge).

Rishi-Dave:
Thanks for the thorough review — the regression catch was important. All four items addressed:

**Issue 1 [High] — Attention cleanup regression:** Added a fallback `match_parent_path` for the post-merge 3-node pattern `["ReduceSum", "ConstantOfShape", "Shape"]`. After the Cast merge fires, the Attention cleanup now correctly matches the shorter pattern and removes the mask_index input. Also fixed a latent bug where `get_graph_by_node(attention_node)` was called on the newly-created node (not yet in the graph) — changed to `get_graph_by_node(node)` to reference the original Attention node.

**Issue 2 [Low] — Unsafe `attribute[0]` access:** Replaced with `OnnxModel.get_node_attribute(constant_of_shape, "value")` and proper attribute replacement via `helper.make_attribute("value", new_val)` + `CopyFrom`. Guarded with a `None` check.

**Issue 3 [Low] — Missing Attention test:** Added `test_attention_mask_cleanup_after_cast_merge` which creates a graph with `Shape → ConstantOfShape → Cast → ReduceSum → Attention(input[3])` plus the Concat path. Verifies both Cast removal (merge) and mask_index removal (cleanup) work together.

**Issue 4 [Low] — Stale comments:** Updated both the top comment (noting "Cast merged into ConstantOfShape" in the After diagram) and the Attention cleanup comment (showing `[--> Cast]` as optional).

Also ran `lintrunner -a` for formatting. All 4 tests pass.

tianleiwu:
LGTM

PR #27508, opened 2026-03-02
[WebNN] Support DepthToSpace op

miaobin:
This op is used in VSR models and can be decomposed into Reshape and Transpose for WebNN.

### DIFF
--- a/onnxruntime/core/providers/webnn/builders/impl/depthToSpace_op_builder.cc
+++ b/onnxruntime/core/providers/webnn/builders/impl/depthToSpace_op_builder.cc
@@ -54,70 +54,52 @@ Status DepthToSpaceOpBuilder::AddToModelBuilderImpl(ModelBuilder& model_builder,
   const int64_t new_width = width * blocksize;
 
   emscripten::val input = model_builder.GetOperand(input_defs[0]->Name());
-  emscripten::val output;
   emscripten::val options = emscripten::val::object();
 
+  // Define mode-specific parameters
+  std::vector<uint32_t> shape1;
+  std::vector<uint32_t> perm;
   if (mode == "DCR") {
-    // Step 1: reshape to [b, blocksize, blocksize, new_channels, h, w]
-    std::vector<uint32_t> shape1{
+    shape1 = {
         SafeInt<uint32_t>(batch),
         SafeInt<uint32_t>(blocksize),
         SafeInt<uint32_t>(blocksize),
         SafeInt<uint32_t>(new_channels),
         SafeInt<uint32_t>(height),
         SafeInt<uint32_t>(width)};
-    options.set("label", node.Name() + "_reshape1");
-    emscripten::val tmp = model_builder.GetBuilder().call<emscripten::val>(
-        "reshape", input, emscripten::val::array(shape1), options);
-
-    // Step 2: transpose to [b, new_channels, h, blocksize, w, blocksize]
-    std::vector<uint32_t> perm{0, 3, 4, 1, 5, 2};
-    options.set("label", node.Name() + "_transpose");
-    options.set("permutation", emscripten::val::array(perm));
-    tmp = model_builder.GetBuilder().call<emscripten::val>("transpose", tmp, options);
-
-    // Step 3: reshape to [b, new_channels, new_height, new_width]
-    std::vector<uint32_t> shape2{
-        SafeInt<uint32_t>(batch),
-        SafeInt<uint32_t>(new_channels),
-        SafeInt<uint32_t>(new_height),
-        SafeInt<uint32_t>(new_width)};
-    options = emscripten::val::object();
-    options.set("label", node.Name());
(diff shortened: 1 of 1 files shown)


miaobin:
@fdwr, @Honry PTAL, thanks.

onnxruntime/core/providers/webnn/builders/impl/depthToSpace_op_builder.cc
```diff
@@ -0,0 +1,211 @@
...
+  const int64_t new_channels = channels / (blocksize * blocksize);
+  const int64_t new_height = height * blocksize;
+  const int64_t new_width = width * blocksize;
+
+  emscripten::val input = model_builder.GetOperand(input_defs[0]->Name());
+  emscripten::val output;
+  emscripten::val options = emscripten::val::object();
+
+  if (mode == "DCR") {
+    // Step 1: reshape to [b, blocksize, blocksize, new_channels, h, w]
```
Honry: Both "DCR" and "CDR" mode need 3 steps, we can simplify the code by moving reshape, transpose and reshape operators out of the if/else branch.

onnxruntime/core/providers/webnn/builders/impl/depthToSpace_op_builder.cc
```diff
@@ -0,0 +1,211 @@
...
+  NodeAttrHelper helper(node);
+  const int64_t blocksize = *helper.GetInt64("blocksize");
+  const std::string mode = helper.Get("mode", "DCR");
+
+  const int64_t batch = input_shape[0];
+  const int64_t channels = input_shape[1];
+  const int64_t height = input_shape[2];
+  const int64_t width = input_shape[3];
+
+  const int64_t new_channels = channels / (blocksize * blocksize);
```
fdwr, line 52: Add sanity check that blockSize is > 0 (defense against division by zero for malformed bad values on the web), unless you can demonstrate that ONNX opset validation already verifies this.

onnxruntime/core/providers/webnn/builders/impl/depthToSpace_op_builder.cc
```diff
@@ -0,0 +1,211 @@
...
+  ORT_RETURN_IF_NOT(GetShape(*input_defs[0], input_shape, logger), "Cannot get input shape");
+
+  NodeAttrHelper helper(node);
+  const int64_t blocksize = *helper.GetInt64("blocksize");
+  const std::string mode = helper.Get("mode", "DCR");
+
+  const int64_t batch = input_shape[0];
+  const int64_t channels = input_shape[1];
+  const int64_t height = input_shape[2];
+  const int64_t width = input_shape[3];
```
fdwr, line 50: Assert `input_shape` >= 4D before dereferencing potentially invalid memory (presumably ONNX opset validation should ensure this already).

fdwr: Never mind - `IsOpSupportedImpl` already covers that.

onnxruntime/core/providers/webnn/builders/impl/depthToSpace_op_builder.cc
fdwr: Never mind - `IsOpSupportedImpl` already covers that.

onnxruntime/core/providers/webnn/builders/impl/depthToSpace_op_builder.cc
miaobin: Done and please take another look, thanks a lot!

miaobin:
> For symmetry, what do you say to adding `SpaceToDepth` too.

That's a good suggestion an I'll send another PR to add `SpaceToDepth`.

onnxruntime/core/providers/webnn/builders/impl/depthToSpace_op_builder.cc
```diff
@@ -54,71 +54,53 @@ Status DepthToSpaceOpBuilder::AddToModelBuilderImpl(ModelBuilder& model_builder,
   const int64_t new_width = width * blocksize;
 
   emscripten::val input = model_builder.GetOperand(input_defs[0]->Name());
-  emscripten::val output;
   emscripten::val options = emscripten::val::object();
 
+  // Define mode-specific parameters
+  std::vector<uint32_t> shape1;
+  std::vector<uint32_t> perm;
```
fdwr, line 61: I'm really glad that WebNN naming chooses clearer words, rather than unclear chopped off fragments like "perm" (some kind of hair do? 😉).

onnxruntime/core/providers/webnn/builders/impl/depthToSpace_op_builder.cc
fdwr: Yeah, I like that consolidation.

PR #27512, opened 2026-03-02
Fix allocator lifetime in EP library plugin

jtuyls:
### Description

This PR changes `ep_info` from `unique_ptr<EpInfo>` to `shared_ptr<EpInfo>` and captures `ep_info` by value in the `OrtAllocatorUniquePtr` deleter lambda inside `CreateSharedAllocatorImpl` to fix a lifetime issue that can cause a non-deterministic segfault: https://github.com/microsoft/onnxruntime/blob/46126133442d08cfcf98267702e602434b4f3578/onnxruntime/core/session/environment.cc#L841

### Motivation and Context

When a session or tensor holds a `shared_ptr<IAllocator>` past `Environment` destruction, the allocator's deleter may fire after `~EpInfo()` has called `EpLibrary::Unload()`, releasing the ep factory and dlclosing the plugin `.so`. At that point `ep_factory` captured in the `OrtAllocatorUniquePtr` lambda points into freed memory, making the `ReleaseAllocator` call in the deleter a use-after-free.

As `Tensor` stores the allocator as `buffer_deleter_` (a `shared_ptr<IAllocator>`) to free its buffer on destruction, this could happen: e.g. in Python, `OrtValue` could outlive the `Environment` and the GC could trigger this behaviour.

### DIFF
--- a/include/onnxruntime/core/session/environment.h
+++ b/include/onnxruntime/core/session/environment.h
@@ -197,10 +197,6 @@ class Environment {
 
   Status RegisterAllocatorImpl(AllocatorPtr allocator);
   Status UnregisterAllocatorImpl(const OrtMemoryInfo& mem_info, bool error_if_not_found = true);
-  Status CreateSharedAllocatorImpl(const OrtEpDevice& ep_device,
-                                   const OrtMemoryInfo& memory_info, OrtAllocatorType allocator_type,
-                                   const OrtKeyValuePairs* allocator_options, OrtAllocator** allocator,
-                                   bool replace_existing);
 
   // Inserts (or assigns) a config entry into `config_entries_`. Locks `config_entries_mutex_`.
   void InsertOrAssignConfigEntry(std::string key, std::string value);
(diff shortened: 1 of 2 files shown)


jtuyls:
Could someone help review this PR? Tagging a few folks that have reviewed recent changes in these files @skottmckay @edgchen1 @adrianlizarraga

jtuyls:
Hi @skottmckay @edgchen1 @adrianlizarraga, could you help review this PR or redirect to the right people?

skottmckay:
The environment should be the first thing created and last thing destroyed. Users should manage memory accordingly. Using shared_ptr makes the ownership semantics less clear.

jtuyls:
> The environment should be the first thing created and last thing destroyed. Users should manage memory accordingly. Using shared_ptr makes the ownership semantics less clear.

Can't user-owned output tensors from Run() outlive the Environment, so it isn't necessarily the last thing destroyed?

Those tensors already hold shared_ptr<IAllocator> by design: https://github.com/microsoft/onnxruntime/blob/46126133442d08cfcf98267702e602434b4f3578/include/onnxruntime/core/framework/tensor.h#L341. Requiring users to free all tensors before Environment destruction contradicts the lifetime guarantees this architecture provides and can't be enforced, resulting in crashes when violated.

The `shared_ptr<EpInfo>` in the allocator deleter isn't introducing new shared ownership — it's making the EP library lifetime consistent with the existing shared ownership in the tensor/allocator design. The alternative would be changing the tensor API away from `shared_ptr<IAllocator>`, which seems a lot more invasive?

Happy to explore alternative approaches if you have a preferred direction.

skottmckay:
It's not safe in general to use the output tensor after the environment is destroyed. For example the default logger goes away at that point so any code that uses it will segfault. e.g. the allocator that created the data may attempt to use the default logger if there's an error. There are also other dependencies on being able to use the allocator that may have been freed, rendering it invalid. e.g. the WebGPU EP requires a WebGPU context and that would have gone away.

So whilst internally an OrtValue uses a shared_ptr I would consider that an implementation detail and not a guarantee about the lifetime.

PR #27539, opened 2026-03-03
Cleanup external data path validation

adrianlizarraga:
### Description
- Updates the `ValidateExternalData()` function:
  - Resolves symlinks when validating external data path for **models loaded from memory**.
    - Previously only did a lexical check that did not resolve symlinks. Now, we resolve symlinks.
  - **Now requires external data path to exist.**
  - Replace the `(base_dir, model_dir)` function parameters with just `model_dir`. `base_dir` was always derived from `model_dir`.
  - Skip validation for WASM builds that do not have a filesystem.
  - Return `Status` instead of throwing exceptions when `std::filesystem` functions fail.
- Updates `Graph::ConvertInitializersIntoOrtValues()`:
  - Prevents unnecessary calls to `ValidateExternalData()` for external data paths that have already been validated.

### DIFF
--- a/onnxruntime/core/providers/cuda/tensor/grid_sample.cc
+++ b/onnxruntime/core/providers/cuda/tensor/grid_sample.cc
@@ -51,7 +51,6 @@ template <typename T, bool IsNHWC>
 GridSample<T, IsNHWC>::GridSample(const OpKernelInfo& info) : CudaKernel(info) {
   opset_start_version_ = info.node().SinceVersion();
 
-  std::string mode_str = info.GetAttrOrDefault<std::string>("mode", "bilinear");
   std::string padding_mode_str = info.GetAttrOrDefault<std::string>("padding_mode", "zeros");
   align_corners_ = static_cast<bool>(info.GetAttrOrDefault<int64_t>("align_corners", 0));
 
--- a/onnxruntime/python/tools/transformers/fusion_skiplayernorm.py
+++ b/onnxruntime/python/tools/transformers/fusion_skiplayernorm.py
@@ -110,9 +110,8 @@ class FusionSkipLayerNormalization(Fusion):
                     )
                     return
             else:
-                # Shape inference failed. Use default skip_index=1 (no broadcasting) since both
-                # Add inputs have already been verified as non-initializer dynamic tensors above.
-                logger.debug("symbolic shape inference failed, using default skip_index for SkipLayerNormalization")
+                logger.debug("skip SkipLayerNormalization fusion since symbolic shape inference failed")
+                return
 
         gather_path = self.model.match_parent_path(add, ["Gather"], [None])
         if gather_path is not None and self.model.find_graph_input(gather_path[0].input[1]) is None:
--- a/onnxruntime/test/providers/cpu/tensor/grid_sample_test_custom.inc
+++ b/onnxruntime/test/providers/cpu/tensor/grid_sample_test_custom.inc
@@ -37,7 +37,7 @@ TYPED_TEST(GridSampleCustomTest, test_grid_sample_20_4D_linear_zeros_mixed_bound
   test.AddAttribute("padding_mode", padding_mode);
   test.AddAttribute("align_corners", align_corners);
   test.AddOutput<TypeParam>("Y", Y_shape, Y_data);
-  RunTests(test, GetExecutionProviders(20));
+  RunTests(test, GetExecutionProviders());
 }
 
 TYPED_TEST(GridSampleCustomTest, test_grid_sample_20_4D_linear_zeros_mixed_bounds_left_top) {
@@ -69,6 +69,5 @@ TYPED_TEST(GridSampleCustomTest, test_grid_sample_20_4D_linear_zeros_mixed_bound
   test.AddAttribute("padding_mode", padding_mode);
   test.AddAttribute("align_corners", align_corners);
   test.AddOutput<TypeParam>("Y", Y_shape, Y_data);
-  RunTests(test, GetExecutionProviders(20));
+  RunTests(test, GetExecutionProviders());
 }
-
--- a/onnxruntime/test/python/transformers/test_attention_fusion.py
+++ b/onnxruntime/test/python/transformers/test_attention_fusion.py
@@ -395,17 +395,17 @@ class TestFusion(unittest.TestCase):
         ssln_count = sum(1 for n in nodes if n.op_type == "SkipSimplifiedLayerNormalization")
 
         # 4 RMSNorm patterns: pre-attn, Q-norm, K-norm, post-attn.
-        # Post-attn RMSNorm has an Add parent (residual) → fused as SkipSimplifiedLayerNormalization.
-        # Remaining 3 stay as SimplifiedLayerNormalization.
+        # Fallback for SkipLayerNormalization is disabled, so post-attn RMSNorm does not fuse.
+        # All 4 stay as SimplifiedLayerNormalization.
         self.assertEqual(
             sln_count,
-            3,
-            f"Expected 3 SimplifiedLayerNormalization (pre-attn + Q-norm + K-norm), got {sln_count}",
+            4,
+            f"Expected 4 SimplifiedLayerNormalization (pre-attn + Q-norm + K-norm + post-attn), got {sln_count}",
         )
         self.assertEqual(
             ssln_count,
-            1,
-            f"Expected 1 SkipSimplifiedLayerNormalization (residual + post-attn RMSNorm), got {ssln_count}",
+            0,
+            f"Expected 0 SkipSimplifiedLayerNormalization (residual + post-attn RMSNorm failed to fuse), got {ssln_count}",
         )
 
 


onnxruntime/core/framework/tensorprotoutils.cc
```diff
@@ -349,66 +349,84 @@ Status TensorProtoWithExternalDataToTensorProto(
...
+  if (IsUnderDirectory(anchor_canonical, resolved)) {
+    return Status::OK();
+  }
+
+  // Fall back to checking against the canonical model directory.
+  // This supports symlinked models (e.g., Hugging Face Hub local cache) where the canonical
+  // parent of the model file differs from the logical parent directory.
+  if (!model_path.empty()) {
+    std::filesystem::path real_model_path;
+    ORT_RETURN_IF_ERROR(WeaklyCanonicalPath(model_path, real_model_path));
```
yuslepukhin: See this comment in my DRAFT PR about weakly_canonical() on Windows.

https://github.com/microsoft/onnxruntime/blob/5402cf026a2f338349185d3ae92eaff7760c7fcb/onnxruntime/core/framework/tensorprotoutils.cc#L355

yuslepukhin (commented):
I am having a hard time reviewing since my tools are not working.

See the tests that I introduced in my DRAFT PR in connection with my comment.

onnxruntime/core/framework/tensorprotoutils.cc
adrianlizarraga: Thanks for link.

If I'm not mistaken, your comment correctly states that comparing the results of `canonical(path)` to `weakly_canonical(path)` does not properly detect symlinks. However, this is not the approach currently used in this function (i.e., we're not comparing the outputs of these two functions).

Instead, the current approach is to resolve the path to the file (resolving symlinks) and then checking if the path to the file is contained by the model's directory.

Please correct me if I'm misinterpreting your comment or if you see issues with the current approach.

onnxruntime/core/framework/tensorprotoutils.cc
```diff
@@ -349,66 +349,84 @@ Status TensorProtoWithExternalDataToTensorProto(
...
+  std::filesystem::path anchor_dir = model_path.parent_path();
+
+  if (anchor_dir.empty()) {  // Happens if either model_path.empty() or model_path.parent_path().empty()
+    anchor_dir = ORT_TSTR(".");
+  }
+
+  // Resolve the anchor directory and the full location path to their canonical forms.
+  // This resolves symlinks so that a link pointing outside the anchor is detected.
+  std::filesystem::path anchor_canonical;
+  ORT_RETURN_IF_ERROR(WeaklyCanonicalPath(anchor_dir, anchor_canonical));
```
edgchen1: why do we use `weakly_canonical` instead of `canonical`? we expect the anchor directory to exist, right?

I'm also wondering about other usages in this function.

onnxruntime/core/framework/tensorprotoutils.cc
```diff
@@ -349,66 +349,84 @@ Status TensorProtoWithExternalDataToTensorProto(
...
+  ORT_RETURN_IF(ec, "Failed to resolve path: ", path, " - ", ec.message());
+  return Status::OK();
+}
 
-        return ORT_MAKE_STATUS(ONNXRUNTIME, FAIL,
-                               "External data path: ", location, " (resolved path: ", resolved,
-                               ") escapes both model directory: ", base_dir,
-                               " and real model directory: ", real_model_dir);
-      }
+// Checks whether `child` is under `base` by comparing path components.
```
edgchen1: nit: with the name `IsUnderDirectory`, there's an assumption that `child` won't have `..` components or go through a symlink pointing outside of `base`. perhaps a name like `HasPathComponentPrefix` would be more general.

edgchen1: also this would return true if `base == child` which may be unexpected.

onnxruntime/core/framework/tensorprotoutils.cc
```diff
@@ -349,66 +349,84 @@ Status TensorProtoWithExternalDataToTensorProto(
...
+
+  // Fall back to checking against the canonical model directory.
+  // This supports symlinked models (e.g., Hugging Face Hub local cache) where the canonical
+  // parent of the model file differs from the logical parent directory.
+  if (!model_path.empty()) {
+    std::filesystem::path real_model_path;
+    ORT_RETURN_IF_ERROR(WeaklyCanonicalPath(model_path, real_model_path));
+    auto real_model_dir = real_model_path.parent_path();
+
+    if (!real_model_dir.empty() && IsUnderDirectory(real_model_dir, resolved)) {
```
edgchen1: is it possible for `real_model_dir` to be empty, if it is originating from absolute path `real_model_path`?

onnxruntime/core/framework/tensorprotoutils.h
```diff
@@ -540,18 +540,19 @@ Status TensorProtoWithExternalDataToTensorProto(
 
 /// <summary>
 /// Validates if the external data path is under the model directory.
-/// If the model is a symlink, it checks against both the logical model directory (base_dir)
+/// The model directory is derived from model_path.parent_path().
+/// If the model is a symlink, it checks against both the logical model directory
 /// and the real/canonical directory of the model.
```
edgchen1: would "directory containing the symlink" be clearer than "logical model directory"? IMO, the meaning of "logical" is not obvious.

onnxruntime/test/framework/tensorutils_test.cc
```diff
@@ -603,9 +636,63 @@ TEST_F(PathValidationTest, ValidateExternalDataPathWithSymlinkOutside) {
...
+TEST_F(PathValidationTest, ValidateExternalDataPathEmptyBasedirWithSymlinkOutside) {
+  // Symbolic link within the current working directory pointing to a file outside CWD.
+  std::filesystem::path cwd = std::filesystem::current_path();
+  std::filesystem::path sub_dir = cwd / "symlink_test_subdir2";
+  std::filesystem::create_directories(sub_dir);
+  AddDirToCleanUp(sub_dir);
+
+  // Check if we can actually make a file outside of the current working directory (i.e., in a temp dir).
+  // This is only possible if the current working directory is NOT the same as the temp directory.
+  // Otherwise, we need to skip this test. This happens in Android CI.
```
edgchen1: we could change the working directory to a test directory and restore the original working directory as part of test teardown

onnxruntime/core/framework/tensorprotoutils.h
```diff
@@ -540,18 +540,19 @@ Status TensorProtoWithExternalDataToTensorProto(
...
-/// If the `base_dir` is empty, the function only ensures that `location` is not an absolute path.
+/// If model_path is empty (model loaded from bytes), the function ensures that `location` is not
+/// an absolute path, does not contain ".." components that escape the current working directory, and
+/// resolves symlinks to verify the target stays within the current working directory.
```
edgchen1: nit: `location` is checked to be absolute regardless of whether `model_path` is empty

also if we need the special handling for `__wasm__`, maybe it should be documented here too

onnxruntime/core/framework/tensorprotoutils.cc
```diff
@@ -349,66 +349,84 @@ Status TensorProtoWithExternalDataToTensorProto(
...
+                                const std::filesystem::path& location) {
+  // Reject absolute paths
+  ORT_RETURN_IF(location.is_absolute(), "Absolute paths not allowed for external data location");
 
-    // We conservatively check that the normalized relative path does not contain ".." path components that would allow
-    // access to arbitrary files outside of the current working directory. Based on ONNX checker validation.
+#if defined(__wasm__)
+  if (model_path.empty()) {
+    // On WASM, filesystem utilities are not expected to always be fully supported.
+    // So, we only do a lexical check: reject ".." components that would escape the working directory.
```
edgchen1: why do we need this special handling for `__wasm__`? seems like the limitation of not checking symlinks is still present here.

onnxruntime/core/framework/tensorprotoutils.h
```diff
@@ -540,18 +540,19 @@ Status TensorProtoWithExternalDataToTensorProto(
...
+/// <param name="model_path">Path to the model file. If empty, the model was loaded from bytes.</param>
 /// <param name="location">Location string retrieved from TensorProto external data</param>
-/// <param name="model_path">Optional path to the model file, used for canonical path validation if base_dir check fails</param>
 /// <returns>The function will fail if the resolved full path is not under the logical model directory
 ///          nor the real directory of the model path</returns>
-Status ValidateExternalDataPath(const std::filesystem::path& base_dir,
-                                const std::filesystem::path& location,
-                                const std::filesystem::path& model_path = {});
+Status ValidateExternalDataPath(const std::filesystem::path& model_path,
+                                const std::filesystem::path& location);
```
edgchen1: nice, removing `base_dir` is simpler

onnxruntime/core/framework/tensorprotoutils.cc
adrianlizarraga: I assumed it was to not incur the cost of ensuring that all initializer files in the model actually exist. That check can be deferred to when the actual external data is accessed.

That said, I'm not 100% sure. Here are the PRs that initially created this function and used `weakly_canonical`:
- https://github.com/microsoft/onnxruntime/pull/26776
- https://github.com/microsoft/onnxruntime/pull/27374

onnxruntime/core/framework/tensorprotoutils.cc
adrianlizarraga: Yes, it is possible if the original model path is a bare file like `"model.onnx"`. In this case, `real_model_path` becomes `"model.onnx"` (unchanged) and `real_model_path.parent_path()` returns an empty path.

onnxruntime/core/framework/tensorprotoutils.cc
adrianlizarraga: A filesystem is not guaranteed on wasm. If I remove this special handling, then some onnxruntime web CI runs throw exceptions on calls to `std::filesystem::current_path`.

I'll temporarily remove this special handling so we can examine the errors.

onnxruntime/core/framework/tensorprotoutils.cc
edgchen1: I had a similar comment here:
https://github.com/microsoft/onnxruntime/pull/27374#discussion_r2823466770

at the very least, I'd think the model file path should exist if one is provided.

onnxruntime/core/framework/tensorprotoutils.cc
edgchen1: > In this case, real_model_path becomes "model.onnx" (unchanged)

oh, this is not what I expected. isn't `real_model_path` supposed to be an absolute path?

onnxruntime/core/framework/tensorprotoutils.cc
adrianlizarraga: I temporarily removed this special handling and a WASM test (browser) fails on a call to `weakly_canonical(".")`:
https://github.com/microsoft/onnxruntime/actions/runs/22875305891/job/66371261534?pr=27539#step:25:1589

```
  ✖ Browser E2E testing - WebGPU backend with external data (slow: 0.828 secs)
    Chrome 145.0.0.0 (Windows 10)
  Error: Can't create a session. ERROR_CODE: 1, ERROR_MESSAGE: External data path validation failed for initializer: model_with_orig_ext_data. Error: tensorprotoutils.cc:356 WeaklyCanonicalPath Failed to resolve path: "." - No such file or directory
      at E (node_modules/onnxruntime-web/dist/ort.webgpu.min.js:6:26225)
      at _e (node_modules/onnxruntime-web/dist/ort.webgpu.min.js:7:23188)
      at async Ke.loadModel (node_modules/onnxruntime-web/dist/ort.webgpu.min.js:7:35888)
      at async Qe.createInferenceSessionHandler (node_modules/onnxruntime-web/dist/ort.webgpu.min.js:7:37598)
      at async t.create (node_modules/onnxruntime-web/dist/ort.webgpu.min.js:6:19885)
      at async Context.<anonymous> (browser-test-webgpu-external-data.js:7:19)
```
Based on a prior experiment, this same test would also fail on a call to `std::filesystem::current_path()`.

Given that a wasm environment without a filesystem will load the model from bytes, I think we need a special wasm handling if the model is loaded from bytes.

Perhaps in a wasm build we can first check if `std::filesystem::current_path` returns an error. If it returns an error and the model path is empty, we just do a lexical check to ensure there are no ".." path components. If there is a model_path, we fall through to the general code that uses `std::filesystem::weakly_canonical`.

onnxruntime/core/framework/tensorprotoutils.cc
adrianlizarraga: > isn't real_model_path supposed to be an absolute path?

`weakly_canonical` does not always return an absolute path. In this case `weakly_canonical("model.onnx")` returns `"model.onnx"`.

edgchen1: would that only be the case if `model.onnx` does not actually exist?

onnxruntime/core/framework/tensorprotoutils.cc
edgchen1: taking a step back - on WASM, if there is no filesystem, would it even be possible to open an external data file with this relative path at all?

adrianlizarraga: Yes, with an ExternalDataLoader: 
https://github.com/microsoft/onnxruntime/blob/9597f8a7b10270e447ccb5fe2ecad77f61eaa45c/onnxruntime/core/framework/external_data_loader.cc#L24

https://github.com/microsoft/onnxruntime/blob/9597f8a7b10270e447ccb5fe2ecad77f61eaa45c/onnxruntime/core/providers/js/external_data_loader.cc#L38

onnxruntime/core/framework/tensorprotoutils.cc
adrianlizarraga: Actually, we can't always expect the model file path to exist.

When the model is loaded from bytes and the user specifies a folder that contains external data files, we set a fake model path of `external_data_folder_path + "/virtual_model.onnx"`:
https://github.com/microsoft/onnxruntime/blob/9597f8a7b10270e447ccb5fe2ecad77f61eaa45c/onnxruntime/core/session/inference_session.cc#L1146

However, perhaps we should always expect `model_path.parent_path()` to exist.

onnxruntime/core/framework/tensorprotoutils.cc
adrianlizarraga: > would that only be the case if model.onnx does not actually exist?

That's right, but we can't always expect the model path to exist. refer to https://github.com/microsoft/onnxruntime/pull/27539#discussion_r2919532576

onnxruntime/core/framework/tensorprotoutils.cc
```diff
@@ -349,66 +349,68 @@ Status TensorProtoWithExternalDataToTensorProto(
...
 }
 
-Status ValidateExternalDataPath(const std::filesystem::path& base_dir,
-                                const std::filesystem::path& location,
-                                const std::filesystem::path& model_path) {
+// Wraps std::filesystem::weakly_canonical with error_code handling.
+static Status WeaklyCanonicalPath(const std::filesystem::path& path, std::filesystem::path& result) {
+  std::error_code ec;
+  result = std::filesystem::weakly_canonical(path, ec);
+  ORT_RETURN_IF(ec, "Failed to resolve path: ", path, " - ", ec.message());
```
edgchen1: nit: maybe explicitly mention `std::filesystem::weakly_canonical` in the error message as it has its own special behavior. to me, a "resolved path" implies that it is an absolute one.

onnxruntime/core/framework/tensorprotoutils.cc
```diff
@@ -349,66 +349,68 @@ Status TensorProtoWithExternalDataToTensorProto(
...
+
+  // The model file itself may be a symlink. Therefore, check against the real/canonical directory of the model
+  // after resolving all symlinks.
+  //
+  // This supports symlinked models (e.g., Hugging Face Hub local cache) where the canonical
+  // parent of the model file differs from the parent directory of the symlinked model file.
+  if (!model_path.empty()) {
+    std::filesystem::path real_model_path;
+    ORT_RETURN_IF_ERROR(WeaklyCanonicalPath(model_path, real_model_path));
+    auto real_model_dir = real_model_path.parent_path();  // empty if model_path is a bare file (e.g., model.onnx)
```
edgchen1:
```suggestion
    auto real_model_dir = real_model_path.parent_path();  // empty if model_path is a bare file that doesn't exist
                                                          // (e.g., virtual_model.onnx)
```

onnxruntime/core/framework/tensorprotoutils.cc
edgchen1: > When the model is loaded from bytes and the user specifies a folder that contains external data files, we set a fake model path of external_data_folder_path + "/virtual_model.onnx"

oof, this special case is unfortunate. ok, thanks for pointing it out.

onnxruntime/core/framework/tensorprotoutils.h
```diff
@@ -539,19 +539,22 @@ Status TensorProtoWithExternalDataToTensorProto(
...
-/// If the `base_dir` is empty, the function only ensures that `location` is not an absolute path.
+/// Validates that the external data path given by `location` is not an absolute path and is under the model directory.
+/// The model directory is derived from model_path.parent_path().
+///
+/// If the model itself is a symlink, this function checks against both the directory containing the symlink
+/// and the real/canonical directory of the model after resolving all symlinks.
+///
+/// The model_path can be empty if the model is loaded from bytes and the application did not explicitly
+/// specify a directory for external data via the session config entry
+/// `kOrtSessionOptionsModelExternalInitializersFileFolderPath`. In this case (empty model_path), the model directory
```
edgchen1: maybe also mention the case where the model path might not actually exist (i.e., "virtual_model.onnx")

onnxruntime/core/framework/tensorprotoutils.cc
edgchen1: I see, I didn't know about this. thanks for pointing this out too.

I see that the external data loader does a lookup here:
https://github.com/microsoft/onnxruntime/blob/9597f8a7b10270e447ccb5fe2ecad77f61eaa45c/onnxruntime/core/framework/external_data_loader.cc#L39-L42

do we need to do any additional validation on the WASM path here (is the ".." check meaningful if there is no filesystem?) or can we just defer the validation to the external data loader?

yuslepukhin:
This is growing to be complicated. We will need a comprehensive documentation for this function detailing all what and why.

onnxruntime/test/framework/tensorutils_test.cc
adrianlizarraga: This may be a moot point now but changing the working directory will cause (more) problems if we ever decide to run tests concurrently.

That said, we now have tests that modify the OrtEnv, so this is already a problem now.

onnxruntime/core/framework/tensorprotoutils.cc
adrianlizarraga: updated

onnxruntime/core/framework/tensorprotoutils.cc
adrianlizarraga: True, I removed the ".." check for wasm.

onnxruntime/core/framework/tensorprotoutils.cc
adrianlizarraga: updated

onnxruntime/core/framework/tensorprotoutils.h
adrianlizarraga: updated

onnxruntime/core/framework/tensorprotoutils.cc
```diff
@@ -349,66 +349,118 @@ Status TensorProtoWithExternalDataToTensorProto(
...
-        return ORT_MAKE_STATUS(ONNXRUNTIME, FAIL,
-                               "External data path: ", location, " (resolved path: ", resolved,
-                               ") escapes both model directory: ", base_dir,
-                               " and real model directory: ", real_model_dir);
-      }
+// Wraps std::filesystem::absolute with error_code handling.
+static Status AbsolutePath(const std::filesystem::path& path, std::filesystem::path& result) {
+  std::error_code ec;
+  result = std::filesystem::absolute(path, ec);
+  ORT_RETURN_IF(ec, "Failed to get absolute path for: ", path, " - ", ec.message());
```
edgchen1: nit: slight inconsistency in the error message with the other path helpers
```suggestion
  ORT_RETURN_IF(ec, "Failed to get absolute path: ", path, " - ", ec.message());
```

onnxruntime/test/framework/tensorutils_test.cc
```diff
@@ -519,43 +520,108 @@ class PathValidationTest : public ::testing::Test {
...
+  }
 
   // Windows vs Unix path separators.
-  ASSERT_STATUS_OK(utils::ValidateExternalDataPath(base_dir_, "sub/data.bin"));
-  ASSERT_STATUS_OK(utils::ValidateExternalDataPath(base_dir_, "sub\\data.bin"));
+#ifdef _WIN32
+  ASSERT_STATUS_OK(utils::ValidateExternalDataPath(model_path, "sub\\data.bin"));
+#else
+  ASSERT_STATUS_OK(utils::ValidateExternalDataPath(model_path, "sub/data.bin"));
+#endif
```
edgchen1, line 600: should we test both types of separator? maybe '\' does not make sense on Linux, but both should be valid on Windows.
```suggestion
#ifdef _WIN32
  ASSERT_STATUS_OK(utils::ValidateExternalDataPath(model_path, "sub\\data.bin"));
#endif
  ASSERT_STATUS_OK(utils::ValidateExternalDataPath(model_path, "sub/data.bin"));
```

onnxruntime/core/framework/tensorprotoutils.cc
```diff
@@ -349,66 +349,118 @@ Status TensorProtoWithExternalDataToTensorProto(
...
+static bool HasPathComponentPrefix(const std::filesystem::path& prefix, const std::filesystem::path& path) {
+  auto [prefix_end, path_it] = std::mismatch(prefix.begin(), prefix.end(), path.begin(), path.end());
+  return prefix_end == prefix.end();
+}
+
+Status ValidateExternalDataPath(const std::filesystem::path& model_path,
+                                const std::filesystem::path& external_data_path) {
+  // Reject empty or absolute external data paths
+  ORT_RETURN_IF(external_data_path.empty(), "Empty external data path not allowed");
+  ORT_RETURN_IF(external_data_path.is_absolute(), "Absolute path not allowed for external data location");
```
edgchen1: something else to consider - on Windows, `/some/path` is not considered absolute (no drive letter, so it doesn't uniquely identify a location). however, we probably still don't want to allow this. maybe we should reject if `!external_data_path.root_path().empty()`.

some experimentation:
https://godbolt.org/z/1jYc63KGP

I know this is outside the scope of this PR's changes, but I just noticed it now.

onnxruntime/core/framework/tensorprotoutils.cc
```diff
@@ -349,66 +349,118 @@ Status TensorProtoWithExternalDataToTensorProto(
...
-    // We conservatively check that the normalized relative path does not contain ".." path components that would allow
-    // access to arbitrary files outside of the current working directory. Based on ONNX checker validation.
-    auto norm_location = location.lexically_normal();
-
-    for (const auto& path_component : norm_location) {
-      if (path_component == ORT_TSTR("..")) {
-        return ORT_MAKE_STATUS(ONNXRUNTIME, FAIL, "External data path: ", location,
-                               " (model loaded from bytes) escapes working directory");
-      }
+    ORT_RETURN_IF_ERROR(AbsolutePath(model_path, model_path_abs));
```
edgchen1: why do we need to call `AbsolutePath()`? would something like `model_dir = model_path.parent_path().empty() ? path{"."} : model_path.parent_path()` work if it will be converted to a canonical path later?

adrianlizarraga: Having the absolute model path is useful for the second half of this function where we consider if the model is itself a symlink. Having an absolute path for the model helps us avoid checking if `real_model_dir` is empty, as discussed in this comment: https://github.com/microsoft/onnxruntime/pull/27539#discussion_r2898606200

For clarity, I'm referring to these lines:
```c++
  if (!model_path_abs.empty()) {
    std::filesystem::path real_model_path;
    ORT_RETURN_IF_ERROR(WeaklyCanonicalPath(model_path_abs, real_model_path));
    auto real_model_dir = real_model_path.parent_path();

    // Check that the external data path is contained by the real model directory.
    // If it is, check if the external data file actually exists.
    if (HasPathComponentPrefix(real_model_dir, external_data_path_canonical)) {

```

adrianlizarraga: I'll try to remove/simplify.

onnxruntime/core/framework/tensorprotoutils.cc
adrianlizarraga: Thanks, yeah good idea. made the change.

onnxruntime/core/framework/tensorprotoutils.cc
```diff
@@ -349,66 +349,101 @@ Status TensorProtoWithExternalDataToTensorProto(
...
-        return ORT_MAKE_STATUS(ONNXRUNTIME, FAIL,
-                               "External data path: ", location, " (resolved path: ", resolved,
-                               ") escapes both model directory: ", base_dir,
-                               " and real model directory: ", real_model_dir);
+  // The model file itself may be a symlink. Therefore, check against the real/canonical directory of the model
+  // after resolving all symlinks.
+  //
+  // This supports symlinked models (e.g., Hugging Face Hub local cache) where the canonical
+  // parent of the model file differs from the parent directory of the symlinked model file.
+  if (!model_path.empty()) {
```
edgchen1: I think we could potentially avoid a redundant check in the non-symlink case. something like:

```suggestion
  if (!model_path.empty() && std::filesystem::is_symlink(model_path)) {
    // perhaps additionally check that std::filesystem::read_symlink(model_path) exists for good measure
```

but I think it's debatable whether the additional complexity is worth it. this is just an idea.

edgchen1 (approved):

onnxruntime/test/framework/tensorutils_test.cc
edgchen1: ah, that's a fair point

onnxruntime/core/framework/tensorprotoutils.cc
adrianlizarraga: Updated.

PR #27540, opened 2026-03-03
[EP Plugin] use ORT_EP_API_ADAPTER instead of BUILD_WEBGPU_EP_STATIC_LIB

fs-eire:
### Description

use a general macro that can be used later in CUDA as well.


include/onnxruntime/ep/adapters.h
```diff
@@ -3,13 +3,15 @@
...
 
-#include "api.h"
-#include "common.h"
-
-// This header is only used when building WebGPU/CUDA EP as a shared library.
+// This header is only used when building WebGPU/CUDA EP as a plugin EP library.
 //
 // This header file is used as a precompiled header so it is always included first.
 
+#define ORT_EP_API_ADAPTER
```
edgchen1: the meaning of the name wasn't so obvious to me, initially. consider a descriptive name like `ORT_USE_EP_API_ADAPTERS` or document the macro.

cmake/CMakeLists.txt
```diff
@@ -401,6 +397,11 @@ if (onnxruntime_ENABLE_TRITON)
   endif()
 endif()
 
+# Building plugin EPs that depend on EP API adapters.
+if(onnxruntime_USE_EP_API_ADAPTERS)
```
edgchen1, line 401: will `onnxruntime_USE_EP_API_ADAPTERS` also control whether `ep/adapters.h` is used as a precompiled header?

fs-eire: No. whether to use the prepcompiled header is per-target based.

PR #27542, opened 2026-03-03
Support Attention(24)-CUDA and disjoint from contrib op

### LINKED ISSUE #27485: [Feature Request] Support nonpad_kv_seqlen and external kv cache in-place update in Attention-24 (CUDA)
## Motivation
(1) Follow up https://github.com/microsoft/onnxruntime/pull/27384 supporting Attention-24 (CPU) - The current implementation of Attention-24 (CUDA) does not support nonpad_kv_seqlen yet: https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/providers/cuda/llm/attention.cc. We should follow the ONNX spec: https://github.com/onnx/onnx/blob/main/docs/Operators.md#Attention and reference implementation: https://github.com/onnx/onnx/blob/main/onnx/reference/ops/op_attention.py to cpmplete Attention op support.
(2) Follow up https://github.com/microsoft/onnxruntime/pull/27446. Demonstrate TensorScatter(24) + Attention(24) to achieve kv cache update external to Attention, and kv cache is updated in-place in TensorScatter

## Description
To implement the support of nonpad_kv_seqlen, We should know that the current ONNX attention CUDA kernel is using both Contrib op MHA and GQA kernels, so there could be differences because of the spec distinctions:
MHA: https://github.com/microsoft/onnxruntime/blob/main/docs/ContribOperators.md#com.microsoft.MultiHeadAttention and https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/contrib_ops/cuda/bert/attention_impl.cu and 
GQA: https://github.com/microsoft/onnxruntime/blob/main/docs/ContribOperators.md#com.microsoft.GroupQueryAttention and https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/contrib_ops/cuda/bert/multihead_attention.cchttps://github.com/microsoft/onnxruntime/blob/main/onnxruntime/contrib_ops/cuda/bert/group_query_attention_impl.cu and https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/contrib_ops/cuda/bert/group_query_attention.cc
In terms of GQA, seqlens_k looks very close to the idea of nonpad_kv_seqlen, and regarding MHA, key_padding_mask itself or extending attn_bias shoud be worth to looking into.

To demonstrate TensorScatter(24) + Attention(24) to achieve kv cache update external to Attention, and kv cache is updated in-place in TensorScatter, we should either extend https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/test/python/transformers/test_onnx_attention/test_gqa.py and https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/test/python/transformers/test_onnx_attention/test_mha.py or create a new test file under the same folder. We should test them both in CUDA and CPU.

## Reference
Reference of logic and ort conventions:
CPU kernel: https://github.com/microsoft/onnxruntime/tree/main/onnxruntime/core/providers/cpu/llm
Test files: https://github.com/microsoft/onnxruntime/tree/main/onnxruntime/test/providers/cpu/llm and https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/test/python/transformers/test_onnx_attention/test_gqa.py and https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/test/python/transformers/test_onnx_attention/test_mha.py
Contrib Op Attention-related CUDA Kernels: https://github.com/microsoft/onnxruntime/tree/main/onnxruntime/contrib_ops/cuda/bert
Contrib Op tests: https://github.com/microsoft/onnxruntime/tree/main/onnxruntime/test/python/transformers and https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/test/contrib_ops/multihead_attention_op_test.cc

titaiwangms:
Fix #27485
Reference https://github.com/microsoft/onnxruntime/pull/27486

## Summary

Implements CUDA kernel support for the **ONNX Attention op (opset 23 and 24)** using a thin-dispatcher architecture that directly calls Flash Attention, Memory Efficient Attention (cutlass FMHA), and Unfused (GEMM+softmax+GEMM) kernels — bypassing the contrib MHA/GQA dispatch layers entirely. This covers both Multi-Head Attention (MHA) and Grouped-Query Attention (GQA) variants, with support for internal KV cache (past/present), external KV cache (`nonpad_kv_seqlen`), boolean and float attention masks, combined `nonpad_kv_seqlen + attn_mask` composition, and 3D (BSNH) / 4D (BNSH) input formats.

The kernel respects `ORT_DISABLE_FLASH_ATTENTION` and `ORT_DISABLE_MEMORY_EFFICIENT_ATTENTION` environment variables, consistent with contrib MHA/GQA behavior.

---

## Architecture: Thin-Dispatcher Design

The core kernel (`attention.cc`) implements a **3-path dispatch cascade** with clear priority:

```
ComputeInternal
  │
  ├─ Early reject: softcap, softmax_precision (not supported by any CUDA kernel)
  │
  ├─ Flash Attention    (fp16/bf16, Ampere+, head_size==v_head_size, no output_qk)
  │   ├─ Path 1: nonpad_kv_seqlen (opset 24 external cache) → mha_fwd_kvcache
  │   ├─ Path 2: past_key + past_value (internal cache decode) → mha_fwd_kvcache
  │   │           Supports: bool mask → seqlens_k, no mask → fill past_seq_len
  │   └─ Path 3: no past, no mask (prompt) → mha_fwd
  │   Excluded: nonpad + mask combo (no bias param), float masks on prompt
  │
  ├─ Memory Efficient   (fp16/bf16, SM75+, no past_key, bias stride aligned)
  │   ├─ Path 1: nonpad_kv_seqlen (±mask) → custom_right_padding + additive bias
  │   ├─ Path 2: with mask (prompt) → standard MEA with additive bias
  │   └─ Path 3: no mask (prompt) → standard MEA
  │
  └─ Unfused Fallback   (all dtypes, all mask types, MHA only)
      ├─ nonpad_kv_seqlen (±mask) → attention_bias with additive composition
      └─ Bridges to contrib MHA's QkvToContext (GEMM+softmax+GEMM)
```

GQA head expansion is handled via `LaunchUngroup` in the MEA path. Flash handles GQA natively via `kv_num_heads` parameter.

---

## What DOESN'T Work ❌

| Combination | Reason | TODO? |
| :--- | :--- | :--- |
| **GQA + fp32** (any) | `MEA LaunchUngroup` is fp16/bf16 only; unfused rejects GQA | ✅ TODO added |
| **GQA + float mask + decode** | Flash rejects float masks; MEA rejects decode; unfused rejects GQA | Known limitation |
| **Softcap on CUDA** | All 3 kernels reject | ✅ TODO added |
| **softmax_precision on CUDA** | All 3 kernels reject | ✅ TODO added |
| **output_qk** beyond `kNone`/`kQK` | Only unfused supports, and only `kNone`/`kQK` | ✅ TODO added |

\* 3 kernels mean Flash/MEA/Unfused

### Additional Known Limitations

| Limitation | Details |
| :--- | :--- |
| **attn_mask shorter than total_seq** | Spec allows padding with -inf; validation enforces exact match. Tracked separately. |
| **No past_present_share_buffer** | Decode always copies past→present (memset + strided copy). Contrib GQA can share buffers for zero-copy. |
| **No XQA kernel** | Contrib GQA's specialized decode kernel requires shared buffers. ONNX Attention's internal-cache decode is ~15-30% slower for GQA workloads. This does not apply when using TensorScatter + `nonpad_kv_seqlen` (opset 24), which avoids the copy overhead entirely. |

---

## Bug Fixes

- **2D mask shape interpretation (MEA path):** Fixed incorrect interpretation of 2D `attn_mask` as `[B, kv_seq]` — the ONNX spec defines it as `[q_seq, total_seq]`. The old code could read out-of-bounds when `batch_size > q_seq` (e.g., batched decode). Fix uses MEA's native broadcast flags (`strideB=0, strideH=0`) instead of an expansion kernel, which is both correct and more efficient.

- **4D BNSH present_key/value (prompt path):** Added D2D `cudaMemcpyAsync` for the `!is_bsnh` case in both Flash and MEA prompt paths. Previously, `present_key`/`present_value` outputs were silently left uninitialized for 4D BNSH inputs. Decode path already worked via `mha_fwd_kvcache`; unfused already worked via `QkvToContext`.

- **Cutlass FMHA bias alignment:** Added alignment checks for attention bias strides in the FMHA launch template, mirroring checks in the kernel to prevent runtime enforcement errors on misaligned strides.

- **SEGFAULT fix** (`group_query_attention_impl.cu`): Added missing explicit template instantiations for `LaunchUngroup<__half>` and `LaunchUngroup<__nv_bfloat16>`. The new `attention.cc` calls `LaunchUngroup` cross-TU, but without explicit instantiations the symbol was undefined at runtime → `dlopen` crash → all CUDA test binaries failed.

- **`val >= 0` assert fix** (`attention_mask_impl.cu`): Fixed overly strict `val > 0` asserts in Flash kernel and AttentionBias kernel — `val=0` is valid for count semantics. Kept `val > 0` in GQA kernel with comment explaining last-valid-index convention.

- **Negative seqlens_k from all-false mask** (`attention_mask_impl.cu`): When mask is all-false (`seq_len=0`) and decode offset is negative, `seq_len + seqlen_offset` produced hugely negative `seqlens_k` passed to flash. Fixed with `max(0, seq_len + seqlen_offset)` clamp.

- **BFloat16 Flash gate fix:** Fixed `disable_flash_attention_` constructor gate that used `!std::is_same<T, MLFloat16>::value`, accidentally excluding BFloat16 from Flash. Changed to `sizeof(T) != 2` to match contrib GQA/PackedMHA/PagedAttention pattern.

---

### Attention Operator and Kernel Registration
* Added support for the Attention operator in opset 24 for CUDA, including kernel registration for `float`, `MLFloat16`, and `BFloat16` types, and refactored opset 23 kernel registration to use versioned kernel macros for clarity and future extensibility.

### Env Var and Kernel Options Support
* Added `disable_flash_attention_` and `disable_memory_efficient_attention_` members read from `AttentionKernelOptions` via `GetAttentionKernelOptions()`. The ONNX Attention kernel now respects `ORT_DISABLE_FLASH_ATTENTION` and `ORT_DISABLE_MEMORY_EFFICIENT_ATTENTION` environment variables, matching the pattern used by all contrib CUDA attention ops.

### Tensor Transposition Utilities
* Replaced 9 copy-paste `if constexpr` type-switch blocks with 2 templated helpers (`TransposeBNSHtoBSNH<T>`, `TransposeBSNHtoBNSH<T>`) for efficiently transposing tensors between BxNxSxH and BxSxNxH formats for `float`, `half`, and `BFloat16` types. Net -117 lines.

### Attention Mask and Sequence Length Handling
* Enhanced the mask-to-sequence length conversion kernel to support both GQA and Flash conventions with a configurable offset, improved handling for all-false masks, and refactored the launch function to accept the offset parameter.

### nonpad_kv_seqlen + attn_mask Composition
* MEA path: supports both via `has_custom_right_padding` (seqlens_k) + additive attn_bias simultaneously.
* Unfused path: converts nonpad to attention_bias, then composes mask additively via `LaunchAddBiasInPlace`. Supports 2D `[q, t]` and 4D `[B, 1, q, t]` mask shapes.
* Flash path: excluded (no bias parameter when seqlens_k is used). Warning logged on fallback.

### Alignment Checks for Attention Bias
* Added stricter alignment checks for attention bias strides in the CUTLASS FMHA launch template, mirroring checks in the kernel to prevent runtime enforcement errors.

### Helper Functions and API Improvements
* Added a new helper function for ungrouping in group query attention (`LaunchUngroup`) and improved the output shape computation for attention to optionally skip nonpad data validation for GPU paths.
* Refactored the CUDA Attention kernel class to split logic into dedicated methods for FlashAttention, MemoryEfficientAttention, and UnfusedAttention, each with ASCII dispatch diagrams documenting supported parameter combinations.

---

## Testing

**39 C++ tests** in `attention_op_test.cc` covering opset 23/24 correctness, nonpad+mask composition (bool and float), BFloat16 Flash guard, all-false mask regression, batch≠q_seq broadcast guard, and nonpad edge cases.

New test files under `onnxruntime/test/python/transformers/test_onnx_attention/`:

- **`test_mha.py`** (14 test classes, 16 test functions) — MHA path: flash, MEA, unfused (fp32), BF16, 3D/4D inputs, bool/float masks, past/present KV cache, causal mode, broadcast masks. Unfused tests explicitly disable flash+MEA via env vars.
- **`test_gqa.py`** (10 test classes, 13 test functions) — GQA path: flash, flash BF16, MEA, 3D/4D inputs, head ratios (32:8, 8:2), bool/float masks, past/present KV cache, 4D BNSH present_kv correctness, nonpad_kv_seqlen (Flash + MEA + CPU).
- **`test_tensorscatter_attention.py`** (5 test classes) — External KV cache with `nonpad_kv_seqlen` (opset 24), TensorScatter-based cache management, fp16 + fp32, with-mask variants (bool + float).
- **`common.py`** — Shared test infrastructure: ONNX model builders, reference implementations, config classes.


onnxruntime/contrib_ops/cuda/bert/group_query_attention_impl.h
```diff
@@ -122,6 +122,15 @@ struct GQABufferRequirements {
   }
 };
 
+template <typename T>
+Status LaunchUngroup(const GroupQueryAttentionParameters& parameters,
```
titaiwangms, line 127: Add a comment saying this is also used by ONNX Attention from providers/cuda/llm

onnxruntime/core/providers/cuda/llm/attention.cc
```diff
@@ -53,588 +68,1048 @@ Attention<T>::Attention(const OpKernelInfo& info) : CudaKernel(info) {
...
+        present_key->MutableData<T>(), present_k_row_bytes,
+        past_key->Data<T>(), past_k_row_bytes,
+        past_k_row_bytes, num_kv_rows,
+        cudaMemcpyDeviceToDevice, cuda_stream));
+
+    const size_t past_v_row_bytes = static_cast<size_t>(parameters.past_sequence_length) *
+                                    parameters.v_head_size * sizeof(T);
+    const size_t present_v_row_bytes = static_cast<size_t>(parameters.total_sequence_length) *
+                                       parameters.v_head_size * sizeof(T);
+    CUDA_RETURN_IF_ERROR(cudaMemcpy2DAsync(
```
titaiwangms, line 343: Wouldn't this cause graph break? Should we use CUDA_KERNEL_ASSSERT? check this: https://github.com/microsoft/onnxruntime/pull/27484

onnxruntime/core/providers/cuda/llm/attention.cc
```diff
@@ -53,588 +68,1048 @@ Attention<T>::Attention(const OpKernelInfo& info) : CudaKernel(info) {
...
+  return onnxruntime::contrib::cuda::QkvToContext<CudaT, CudaT>(
+      device_prop, cublas, cudnn, context->GetComputeStream(), contribop_parameters, data);
+}
+
+// ============================================================================
+// ComputeInternal: Dispatch to appropriate attention kernel
+// ============================================================================
+// MHA path (q_num_heads == kv_num_heads): uses direct kernel dispatch cascade
+//   flash → memory efficient → unfused
+// GQA path (q_num_heads != kv_num_heads): routes through GQA dispatch (kept for now)
```
titaiwangms: GQA isn't now using our own dispatch? This should be outdated.

onnxruntime/core/providers/cuda/llm/attention.cc
```diff
@@ -53,588 +68,1048 @@ Attention<T>::Attention(const OpKernelInfo& info) : CudaKernel(info) {
...
+                                            parameters.q_num_heads, parameters.kv_num_heads) &&
+        parameters.head_size == parameters.v_head_size &&
+        !has_output_qk &&
+        parameters.softcap == 0.0f &&
+        parameters.softmax_precision == 0 &&
+        // Bool masks without past_key (prompt) can't use flash because mha_fwd_kvcache's
+        // causal semantics are decode-oriented (window offset by seqlens_k). For causal
+        // prompt with padding, MEA handles it correctly via attention bias conversion.
+        // Flash handles: no mask, decode with past (±mask), nonpad_kv_seqlen.
+        (attn_mask == nullptr || (attn_mask->IsDataType<bool>() && past_key != nullptr));
```
titaiwangms: So in contrib_op MHA, whenever there is attn_bias, it uses MEA?

onnxruntime/core/providers/cuda/llm/attention.cc
```diff
@@ -53,588 +68,1048 @@ Attention<T>::Attention(const OpKernelInfo& info) : CudaKernel(info) {
...
+      qk_matmul_output_mode_ != attention_helper::QKMatMulOutputMode::kQK) {
+    return ORT_MAKE_STATUS(ONNXRUNTIME, NOT_IMPLEMENTED,
+                           "qk_matmul_output_mode other than kNone and kQK is not supported yet "
+                           "in Attention op (CUDA).");
+  }
+
+  if (is_gqa) {
+    // TODO: Support GQA in unfused attention path for fp32/old-GPU fallback.
+    // Requires ~160 lines: ExpandKVHeads kernel to replicate KV heads, wiring in unfused dispatch.
+    // See issue #27516 for tracking.
```
titaiwangms: Explain why GQA can't just directly use Unfused kernel into comment

onnxruntime/core/providers/cuda/llm/attention.cc
```diff
@@ -53,588 +68,1048 @@ Attention<T>::Attention(const OpKernelInfo& info) : CudaKernel(info) {
...
-  // The default scale depends on the input dimensions. It is set to nan to indicate that it should be computed.
   scale_ = info.GetAttrOrDefault<float>("scale", std::numeric_limits<T>::quiet_NaN());
   softcap_ = info.GetAttrOrDefault<float>("softcap", 0.0f);
   softmax_precision_ = static_cast<int>(info.GetAttrOrDefault<int64_t>("softmax_precision", 0));
   ORT_ENFORCE(scale_ > 0 || std::isnan(scale_), "scale must be greater than 0 if specified");
 }
 
+// ============================================================================
+// RunFlashAttention: Direct flash attention kernel call
+// ============================================================================
```
titaiwangms, line 163: Can you add more information (could be easy diagram/pipeline) of what combinations of parameters we support here, since there are IFs/ELSEs within the functions. It will be easier to check what we have and what we miss for future reference. Same for MEA and Unfused kernels.

onnxruntime/core/providers/cuda/llm/attention.cc
```diff
@@ -53,588 +68,1048 @@ Attention<T>::Attention(const OpKernelInfo& info) : CudaKernel(info) {
...
+Status Attention<T>::ComputeInternal(OpKernelContext* context) const {
+  const Tensor* Q = context->Input<Tensor>(0);
+  const Tensor* K = context->Input<Tensor>(1);
+  const Tensor* V = context->Input<Tensor>(2);
+  const Tensor* attn_mask = context->Input<Tensor>(3);
+  const Tensor* past_key = context->Input<Tensor>(4);
+  const Tensor* past_value = context->Input<Tensor>(5);
+  const Tensor* nonpad_kv_seqlen = context->Input<Tensor>(6);  // optional, Opset 24
+
+  ORT_ENFORCE(nonpad_kv_seqlen == nullptr || attn_mask == nullptr,
```
titaiwangms: Add a TODO on this as it's valid according to spec.

onnxruntime/test/python/transformers/test_onnx_attention/test_mha.py
```diff
@@ -8,16 +8,22 @@
 Tests for ONNX Attention op (opset 23) — MHA path (kv_num_heads == q_num_heads).
```
titaiwangms: we don't need to specify the tests are for opset23 anymore. It is supporting all version because we support nonpad_kv_seqlen?

onnxruntime/test/python/transformers/test_onnx_attention/test_gqa.py
```diff
@@ -152,7 +153,11 @@ def parity_check_gqa_prompt(
                 )
             torch.testing.assert_close(out, first_out, rtol=0, atol=0, msg="Output mismatch between two runs")
 
-    out = torch.reshape(out, (config.batch_size, config.q_sequence_length, config.q_num_heads, config.head_size))
+    if config.use_4d_bnsh:
+        # 4D BNSH output → BSNH for comparison
+        out = out.transpose(1, 2).contiguous()
+    else:
+        out = torch.reshape(out, (config.batch_size, config.q_sequence_length, config.q_num_heads, config.head_size))
```
titaiwangms, line 163: Is this because torch output shape different from ORT? I thought torch uses 3d. bsnh does not seem like torch would use? Or it's just our set up?

onnxruntime/core/providers/cuda/llm/attention.cc
```diff
@@ -53,588 +68,1048 @@ Attention<T>::Attention(const OpKernelInfo& info) : CudaKernel(info) {
...
+          parameters.batch_size, parameters.kv_sequence_length,
+          parameters.kv_num_heads, parameters.v_head_size,
+          V->Data<BFloat16>(), present_value->MutableData<BFloat16>(),
+          cuda_stream, device_prop.maxThreadsPerBlock));
+    } else {
+      ORT_RETURN_IF_ERROR(onnxruntime::contrib::cuda::Transpose_BSNH_to_BNSH(
+          parameters.batch_size, parameters.kv_sequence_length,
+          parameters.kv_num_heads, parameters.v_head_size,
+          V->Data<float>(), present_value->MutableData<float>(),
+          cuda_stream, device_prop.maxThreadsPerBlock));
```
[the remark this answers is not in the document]
titaiwangms: Added D2D memcpy for prompt path in both flash and MEA. Decode already worked via mha_fwd_kvcache. Removed test skip.

onnxruntime/core/providers/cuda/llm/attention_mask_impl.cu
```diff
@@ -173,5 +183,222 @@ template Status LaunchConvertBoolMaskToAttentionBias<__half>(
...
+    const int q_seq_len,
+    const int total_seq_len,
+    const float mask_filter_value) {
+  int64_t idx = static_cast<int64_t>(blockIdx.x) * blockDim.x + threadIdx.x;
+  int64_t total = static_cast<int64_t>(batch_size) * q_seq_len * total_seq_len;
+  for (; idx < total; idx += static_cast<int64_t>(gridDim.x) * blockDim.x) {
+    int b = static_cast<int>(idx / (static_cast<int64_t>(q_seq_len) * total_seq_len));
+    int t = static_cast<int>(idx % total_seq_len);
+    int64_t valid_len = nonpad_kv_seqlen[b];
+    CUDA_KERNEL_ASSERT(valid_len > 0 && valid_len <= static_cast<int64_t>(total_seq_len));
```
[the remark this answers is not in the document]
titaiwangms: ~~Current > 0 is correct. val=0 gives flash seqlens_k=-1 (undefined). No real LLM sends empty sequences. Relaxing to ≥ 0 would silently produce wrong results.~~

onnxruntime/core/providers/cuda/llm/attention_mask_impl.cu
```diff
@@ -72,32 +73,39 @@ __global__ void ConvertMaskToSeqlensKernel(
...
-    if (!found_first_false && !current) {
-      // Found first False - this is where padding starts
-      seq_len = i;
-      found_first_false = true;
-    } else if (found_first_false && current) {
-      // Found True after False - mask is not contiguous (invalid)
-      CUDA_KERNEL_ASSERT(false);  // mask must be contiguous (no True after False)
+  int seq_len;
+  if (!mask_row[0]) {
+    // Entire row is padding (all-false mask)
```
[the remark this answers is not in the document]
titaiwangms: O(seq_len) validation inside GPU kernel is a performance anti-pattern. Not a realistic LLM scenario — causal decode always attends to at least itself. Invalid inputs are caller's responsibility.

onnxruntime/core/providers/cuda/llm/attention.cc
```diff
@@ -53,588 +68,1048 @@ Attention<T>::Attention(const OpKernelInfo& info) : CudaKernel(info) {
...
+                              ? reinterpret_cast<const NativeCudaT*>(converted_mask_buffer.get())
+                              : reinterpret_cast<const NativeCudaT*>(attn_mask->Data<T>());
+        auto* dst = reinterpret_cast<NativeCudaT*>(expanded_buffer.get());
+        ORT_RETURN_IF_ERROR(LaunchBroadcastBias2DToQSeq<NativeCudaT>(
+            src, dst, parameters.batch_size, q_len, kv_len,
+            cuda_stream, device_prop.maxThreadsPerBlock));
+        attn_bias_data = expanded_buffer.get();
+        converted_mask_buffer = std::move(expanded_buffer);
+        // Expanded shape is [B, 1, q_seq, kv_seq]
+        broadcast_bias_dim_0 = false;
```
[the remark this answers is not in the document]
titaiwangms: Fixed MEA path: changed from wrong [B, kv_seq] to correct [q_seq, total_seq] per spec. Used broadcast flags instead of expansion kernel — simpler and faster.

onnxruntime/core/providers/cuda/llm/attention.cc
```diff
@@ -53,588 +68,1048 @@ Attention<T>::Attention(const OpKernelInfo& info) : CudaKernel(info) {
...
-        k_buffer = GetScratchBuffer<void>(kv_buffer_bytes, context->GetComputeStream());
-        v_buffer = GetScratchBuffer<void>(kv_buffer_bytes, context->GetComputeStream());
-        fmha_buffer = GetScratchBuffer<void>(fmha_buffer_bytes, context->GetComputeStream());
-
-        gqa_data.k = reinterpret_cast<CudaT*>(k_buffer.get());
-        gqa_data.v = reinterpret_cast<CudaT*>(v_buffer.get());
-        gqa_data.fmha_buffer = reinterpret_cast<CudaT*>(fmha_buffer.get());
+    }
+    if (present_value != nullptr && is_bsnh) {
+      if constexpr (std::is_same_v<T, MLFloat16>) {
```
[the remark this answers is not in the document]
titaiwangms: Added D2D memcpy for prompt path in both flash and MEA. Decode already worked via mha_fwd_kvcache. Removed test skip.

onnxruntime/core/providers/cuda/llm/attention.cc
titaiwangms: Investigated — all ORT_ENFORCE calls in attention.cc are already host-side (before kernel launch). Mask kernels already use CUDA_KERNEL_ASSERT. No changes needed.

onnxruntime/contrib_ops/cuda/bert/cutlass_fmha/fmha_launch_template.h
```diff
@@ -258,6 +258,22 @@ void DispatchIsAligned(const MemoryEfficientAttentionParams& params) {
...
+    int bias_strideH = params.broadcast_attn_bias_dim_1 ? 0 : num_queries * num_keys;
+    int bias_strideB = params.broadcast_attn_bias_dim_0
+                           ? 0
+                           : ((params.broadcast_attn_bias_dim_1 ? 1 : params.num_heads) * num_queries * num_keys);
+    is_aligned = is_aligned &&
+                 bias_strideM % AlignedAK::kAlignmentQ == 0 &&
+                 (params.num_heads <= 1 || bias_strideH % AlignedAK::kAlignmentQ == 0) &&
+                 (params.batch_size <= 1 || bias_strideB % AlignedAK::kAlignmentQ == 0);
+  }
+
```
titaiwangms, line 287: @tianleiwu Can you help me make sure this is a bug fixed as Claude suggested?

onnxruntime/core/providers/cuda/llm/attention.cc
```diff
@@ -53,588 +68,1074 @@ Attention<T>::Attention(const OpKernelInfo& info) : CudaKernel(info) {
...
-        k_buffer = GetScratchBuffer<void>(kv_buffer_bytes, context->GetComputeStream());
-        v_buffer = GetScratchBuffer<void>(kv_buffer_bytes, context->GetComputeStream());
-        fmha_buffer = GetScratchBuffer<void>(fmha_buffer_bytes, context->GetComputeStream());
-
-        gqa_data.k = reinterpret_cast<CudaT*>(k_buffer.get());
-        gqa_data.v = reinterpret_cast<CudaT*>(v_buffer.get());
-        gqa_data.fmha_buffer = reinterpret_cast<CudaT*>(fmha_buffer.get());
+    } else if (present_key != nullptr && !is_bsnh) {
+      // 4D BNSH prompt: K is already BNSH, just D2D copy to present
+      CUDA_RETURN_IF_ERROR(cudaMemcpyAsync(
```
titaiwangms, line 468: @hariharans29 Claude says this does not break graph. Can you confirm that? I am a bit suspicious (ref: https://github.com/microsoft/onnxruntime/pull/27484)

onnxruntime/core/providers/cuda/llm/attention_mask_impl.cu
```diff
@@ -173,5 +183,222 @@ template Status LaunchConvertBoolMaskToAttentionBias<__half>(
...
+    const int kv_seq_len) {
+  int64_t total = static_cast<int64_t>(batch_size) * q_seq_len * kv_seq_len;
+  for (int64_t idx = static_cast<int64_t>(blockIdx.x) * blockDim.x + threadIdx.x;
+       idx < total;
+       idx += static_cast<int64_t>(gridDim.x) * blockDim.x) {
+    // Map flat index → (batch, q_pos, kv_pos)
+    int kv_pos = static_cast<int>(idx % kv_seq_len);
+    int b = static_cast<int>(idx / (static_cast<int64_t>(q_seq_len) * kv_seq_len));
+    // All q positions read from the same src row for this batch
+    dst[idx] = src[static_cast<int64_t>(b) * kv_seq_len + kv_pos];
```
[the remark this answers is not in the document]
titaiwangms: Fixed MEA path: changed from wrong [B, kv_seq] to correct [q_seq, total_seq] per spec. Used broadcast flags instead of expansion kernel — simpler and faster.

tianleiwu:
## Must-Fix Issues

### 1. ⚠️ Massive code duplication in transpose/present_kv population

The same `if constexpr` type-switch pattern for `Transpose_BNSH_to_BSNH` / `Transpose_BSNH_to_BNSH` is repeated **at least 8 times** across `RunFlashAttention` and `RunMemoryEfficientAttention` (for Q transpose, output transpose, present_key, present_value, K/V in decode, etc.). Each instance is 12+ lines of nearly identical code.

**Recommendation:** Create a templated helper function:
```cpp
template <typename T>
Status TransposeHelper(TransposeDirection dir, int batch, int seq, int heads, int head_size,
                       const void* input, void* output, cudaStream_t stream, int max_threads);
```
This would reduce ~150 lines of boilerplate and eliminate the risk of copy-paste errors (e.g., using `head_size` when `v_head_size` is needed).

**Severity:** High (maintainability). While not a correctness issue today, this is a maintenance hazard. Should be addressed before merge or tracked as a follow-up.

### 2. ⚠️ `ConvertNonpadKvSeqlenToFlashSeqlensKKernel` asserts `val > 0` but `val == 0` should be valid

In `attention_mask_impl.cu`, line ~581:
```cpp
CUDA_KERNEL_ASSERT(val > 0);
```
But the PR description and test code explicitly handle `nonpad_kv_seqlen=0` (zeroing out NaN batches). The `ConvertNonpadKvSeqlenToSeqlensKKernel` (GQA variant) also has this assert but with a comment explaining the garbage-at-pos-0 issue.

For the Flash variant, if `val == 0`, the kernel should produce `seqlens_k=0` which tells `mha_fwd_kvcache` there are 0 valid tokens. The assert prevents this from working in debug builds.

**The CPU test `gqa_nonpad_kv_seqlen_cpu_test_cases` explicitly tests `zero_seqlen`** with comment "triggers CUDA_KERNEL_ASSERT in debug builds" — this confirms the assert is too strict for the CUDA path.

**Recommendation:** Change to `CUDA_KERNEL_ASSERT(val >= 0)` for the Flash variant, or validate at the host level and allow 0 through.

---

## Suggestions

### 3. Present KV population logic duplicated across Flash and MEA paths

Both `RunFlashAttention` and `RunMemoryEfficientAttention` have ~50 lines of identical present_key/present_value population code (BSNH→BNSH transpose for 3D inputs, D2D copy for 4D inputs). This should be extracted into a shared `PopulatePresentKV()` helper.

### 4. MEA `workspace_buffer` scope issue

In `RunMemoryEfficientAttention`, for the standard MEA path (not nonpad_kv_seqlen):
```cpp
auto workspace_buffer = GetScratchBuffer<void>(workspace_bytes, ...);
p.workspace = workspace_buffer.get();
onnxruntime::contrib::cuda::run_memory_efficient_attention(p);
```
The `workspace_buffer` is declared inside the `if` block, but `run_memory_efficient_attention` is called in both `if` and `else` branches. In the `else` branch (`p.workspace = nullptr`), this is fine. But the pattern is fragile — if someone adds code after the if/else, `workspace_buffer` may go out of scope while the kernel is still running.

**Recommendation:** Move `workspace_buffer` declaration to function scope (before the if/else) for clarity.

### 5. `Transpose_BNSH_to_BSNH` reuses `LaunchTransCtx` — verify parameter mapping

The new `Transpose_BNSH_to_BSNH` calls `LaunchTransCtx(stream, sequence_length, batch_size, head_size, num_heads, ...)`. This reuses the kernel designed for the inverse operation. The swapping of sequence_length/batch_size position in the argument list is how the transpose direction is reversed — this is correct but subtle.

**Recommendation:** Add a brief comment in the implementation noting that the arg-swap is intentional, or better yet, add a unit test for `Transpose_BNSH_to_BSNH` roundtrip correctness.

### 6. Bias alignment check in `fmha_launch_template.h` may have edge case

The alignment check uses `AlignedAK::kAlignmentQ` for bias stride alignment:
```cpp
bias_strideM % AlignedAK::kAlignmentQ == 0
```
This mirrors what the kernel enforces. However, when `broadcast_attn_bias_dim_1 = true`, `bias_strideH = 0`, and the check `bias_strideH % kAlignmentQ == 0` passes vacuously (0 is divisible by anything). This is correct but should have a comment explaining it.

### 7. `LaunchFillInt32` for seqlens_k initialization — good for CUDA graph safety

The old code used `cudaMemcpyAsync` from a `std::vector<int>` (host memory) to fill `seqlens_k`. The new `LaunchFillInt32` kernel is CUDA-graph-capturable.

### 8. `skip_nonpad_data_validation` parameter name

The parameter name `skip_nonpad_data_validation` in `ComputeOutputShapeForAttention` is clear but the comment says "data is on GPU". Consider a more descriptive name like `nonpad_data_on_gpu` since that's the actual reason for skipping.

onnxruntime/contrib_ops/cuda/bert/cutlass_fmha/fmha_launch_template.h
tianleiwu: No, the added alignment check is not really needed. Here's why:

The key line is line 244 in fmha_launch_template.h:
```
ORT_ENFORCE(Attention::check_supported(p));
```
And looking at check_supported in kernel_forward.h
 (line 619–663), it already enforces exactly the same bias stride alignment checks:
```
cpp
if (kSupportsBias) {
    CHECK_ALIGNED_PTR(p.attn_bias_ptr, kAlignmentQ);
    XFORMERS_CHECK(p.num_batches <= 1 || p.bias_strideB % kAlignmentQ == 0, ...);
    XFORMERS_CHECK(p.num_heads <= 1 || p.bias_strideH % kAlignmentQ == 0, ...);
    XFORMERS_CHECK(p.bias_strideM % kAlignmentQ == 0, ...);
}
```
The critical thing is: 

check_supported is called regardless of whether is_aligned is true or false. The flow is:


1. DispatchIsAligned determines is_aligned (true/false)
2. It dispatches to LaunchCutlassFmha<..., is_aligned=true> or LaunchCutlassFmha<..., is_aligned=false>
3. Both paths call ORT_ENFORCE(Attention::check_supported(p)) at line 244
4. check_supported enforces the same bias alignment checks with XFORMERS_CHECK
So the PR's new code in 

`DispatchIsAligned` is trying to force is_aligned = false when bias strides are misaligned, hoping the "unaligned" kernel path won't enforce alignment. But the unaligned path still calls 

`check_supported`  which crashes with the same XFORMERS_CHECK. The kAlignmentQ/kAlignmentK/kAlignmentV constants in 

`check_supported`  are the same regardless of is_aligned — the "unaligned" template specialization only affects Q/K/V data pointer alignment, not stride alignment.

In other words:
- is_aligned=true: Uses aligned memory loads for Q/K/V data (faster). Requires both pointer alignment AND stride alignment.
- is_aligned=false: Uses unaligned memory loads for Q/K/V data. But `check_supported`  still enforces stride alignment for bias regardless.

So setting is_aligned = false due to bias misalignment doesn't help — the ORT_ENFORCE(Attention::check_supported(p)) will still crash. The real fix would need to be either:

Skip MEA entirely when bias strides are misaligned (which the caller in attention.cc already does with mea_eligible checks) Or pad/reshape the bias to satisfy alignment

Since attention.cc already has alignment-based MEA eligibility gating (the min_bias_align check around line 348–358 of the new code), this `DispatchIsAligned`  change is redundant defense-in-depth at best, and misleading at worst — it suggests that routing to the unaligned kernel would fix the problem, when it won't.

onnxruntime/core/providers/cuda/llm/attention_mask_impl.h
```diff
@@ -8,37 +8,30 @@
 namespace onnxruntime {
 namespace cuda {
 
-// Convert a boolean attention mask to sequence lengths for use with GQA kernels.
+// Convert a boolean attention mask to sequence lengths with a configurable offset.
 //
 // The mask is expected to have the following properties:
 // 1. It represents right-padding only (valid tokens first, padding at the end)
 // 2. Each batch's mask should start with True (valid) values
 // 3. True values should be contiguous, followed by contiguous False (padding) values
```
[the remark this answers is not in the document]
titaiwangms: Fixed

onnxruntime/core/providers/cuda/llm/attention.cc
```diff
@@ -53,588 +68,1110 @@ Attention<T>::Attention(const OpKernelInfo& info) : CudaKernel(info) {
...
-    data.output = reinterpret_cast<CudaT*>(Y->MutableData<T>());
-    data.present_key = (present_key == nullptr) ? nullptr : reinterpret_cast<CudaT*>(present_key->MutableData<T>());
-    data.present_value = (present_value == nullptr) ? nullptr : reinterpret_cast<CudaT*>(present_value->MutableData<T>());
-    if (nullptr != output_qk) {
-      data.output_qk = reinterpret_cast<CudaT*>(output_qk->MutableData<T>());
-    }
+  // Handle attention mask → attention_bias conversion
+  IAllocatorUniquePtr<void> converted_mask_buffer;
+  IAllocatorUniquePtr<void> nonpad_bias_buffer;
+  const void* attn_bias_data = nullptr;
```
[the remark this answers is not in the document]
titaiwangms: Fixed

onnxruntime/test/python/transformers/test_onnx_attention/test_tensorscatter_attention.py
```diff
@@ -0,0 +1,589 @@
...
+            ep="CPUExecutionProvider",
+            device="cpu",
+            torch_type=torch.float32,
+            ort_type=TensorProto.FLOAT,
+            is_causal=is_causal,
+        )
+        numpy.testing.assert_allclose(output, ref_output, rtol=cpu_fp32_rtol, atol=cpu_fp32_atol)
+
+
+@unittest.skipIf(not has_flash_attention(), "Flash Attention (SM80+) is not available, skipping tests.")
```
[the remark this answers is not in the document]
titaiwangms: CI is 90

onnxruntime/core/providers/cuda/llm/attention_mask_impl.cu
```diff
@@ -72,32 +74,39 @@ __global__ void ConvertMaskToSeqlensKernel(
...
+        CUDA_KERNEL_ASSERT(false);  // mask must be contiguous (no True after False)
+      }
     }
   }
 
-  // seqlens_k is total_sequence_length - 1 for GQA convention
-  seqlens_k[batch_idx] = seq_len - 1;
+  // seqlens_k output: seq_len + seqlen_offset
+  // GQA convention (seqlen_offset=-1): stores last valid index (count - 1)
+  // Flash convention (seqlen_offset=0): stores actual count
```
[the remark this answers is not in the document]
titaiwangms: Fixed. clamp to max(0, seq_len + seqlen_offset)

onnxruntime/core/providers/cuda/llm/attention_mask_impl.cu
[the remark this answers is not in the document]
titaiwangms: Sorry, Fixed

titaiwangms:
# PR Summary — Changes from Review Response Session

Thanks @tianleiwu for the thorough review. Here's a summary of all changes made to address review feedback.

---

## Bug Fixes

**SEGFAULT fix** (`group_query_attention_impl.cu`)
Added missing explicit template instantiations for `LaunchUngroup<__half>` and `LaunchUngroup<__nv_bfloat16>`. Root cause of all CI failures: the new `attention.cc` calls `LaunchUngroup` cross-TU, but without explicit instantiations the symbol was undefined at runtime → `dlopen` crash → every CUDA test binary failed.

**`val > 0` → `val >= 0`** (`attention_mask_impl.cu`)
Fixed overly strict asserts in Flash kernel (line 243) and AttentionBias kernel (line 288) — `val=0` is valid for count semantics. Kept `val > 0` in GQA kernel (line 202) with comment: GQA uses last-valid-index convention where `val=0` yields `seqlens_k=-1` (undefined).

**Negative seqlens_k from all-false mask** (`attention_mask_impl.cu`)
When mask is all-false (`seq_len=0`) and decode offset is negative, `seq_len + seqlen_offset` produced hugely negative seqlens_k passed to flash. Fixed with `max(0, seq_len + seqlen_offset)` clamp (line 106).

## Code Quality

**Transpose refactoring** (`attention.cc`)
Replaced 9 copy-paste `if constexpr` type-switch blocks with 2 templated helpers: `TransposeBNSHtoBSNH<T>` and `TransposeBSNHtoBNSH<T>`. Net **-117 lines**.

**Dead code removal** (`attention_mask_impl.cu`, `attention_mask_impl.h`)
Deleted unused `LaunchConvertNonpadKvSeqlenToSeqlensK` function and its kernel. **-61 lines**.

**Unused variable** (`attention.cc`)
Removed unreferenced `nonpad_bias_buffer` declaration (would trigger `-Werror=unused-variable`).

**workspace_buffer scope** (`attention.cc`)
Moved `workspace_buffer` from inner if-block to outer scope, ensuring lifetime covers async kernel launch. Eliminated duplicated `run_memory_efficient_attention` call.

## Documentation

- **14 comment fixes** across `attention.cc`, `attention_mask_impl.cu/.h`, `fmha_launch_template.h`
- Updated Flash/MEA/Unfused eligibility docstrings with accurate dispatch criteria
- Fixed misleading docstring: "mask must start with True" → "all-false masks are valid"
- Added **11-line clarifying comment** in `DispatchIsAligned` explaining why the bias stride alignment check is needed: `kAlignmentQ` is template-dependent (8 for aligned, 4 for unaligned on SM75+ fp16), the ONNX Attention outer gate uses threshold 4 (allowing strides like 12), and contrib MHA's stricter gate at 8 makes the check redundant there

## Tests

- **2 new regression tests** in `test_mha.py`: `AllMasked` on CUDA EP (+ fp16 GQA variant), decode with all-false mask and `past_sequence_length > 0`
- **All 196 tests pass**: 36 C++ AttentionTest + 55 GQA + 47 MHA + 58 TensorScatter

## Review Items Status

| # | Item | Status |
|---|---|---|
| MF-1 | Transpose code duplication | ✅ Refactored into 2 helpers (-117 lines) |
| MF-2 | `val > 0` assert too strict | ✅ Fixed to `val >= 0` (Flash + Bias kernels) |
| S-3 | Present KV dedup | ✅ Covered by transpose refactoring |
| S-4 | workspace_buffer scope | ✅ Moved to outer scope |
| S-5 | Arg-swap comment | ✅ Already documented in `attention_transpose.cu:396` |
| S-6 | Bias alignment edge case comment | ✅ Added broadcast stride=0 note |
| S-7 | FillInt32 CUDA graph safety | N/A (positive feedback) |
| S-8 | `skip_nonpad_data_validation` rename | Deferred — behavior-descriptive name (`skip_*`) is more future-proof; call site comment explains the "why" |
| T-18 | Revert alignment check | **Keeping** — see comment in code for rationale |
| T-21 | Remove unused `nonpad_bias_buffer` | ✅ Removed |
| T-23 | Negative seqlens_k clamp | ✅ Fixed with `max(0, ...)` |

### Note on the alignment check (Thread 18)

The bias stride check in `DispatchIsAligned` is needed for the ONNX Attention path. The key insight: `kAlignmentQ` differs between aligned (8) and unaligned (4) kernel instantiations (`kernel_forward.h:414`). The ONNX Attention outer gate checks `total_seq % 4 == 0`, so strides like 12 reach `DispatchIsAligned` and need routing to the unaligned kernel. Contrib MHA gates at `seq % 8 == 0`, making this check redundant there. Added a detailed comment in the code explaining this.

onnxruntime/contrib_ops/cuda/bert/cutlass_fmha/fmha_launch_template.h
titaiwangms: Thanks, I have checked this gain. It's a requirement to ONNX Attention looser gate on `min_bias_align=4`, but in contrib op MHA is more strict.

Detail:

The new ONNX Attention kernel path (core/providers/cuda/llm/attention.cc lines 1003-1013) intentionally uses a looser outer gate — min_bias_align=4 (GemmType::kMinimumAlignment) instead of 8 — to maximize MEA coverage for broader input shapes. This creates a gap: inputs where total_seq_len % 4 == 0 but total_seq_len % 8 != 0 (e.g., 12, 20, 28) pass the outer gate and need DispatchIsAligned to route them to the unaligned kernel.

Concrete example with kv_seq_len=12, fp16, SM80+:
1. Outer gate: 12 % 4 == 0 → passes, MEA called
2. Without bias check: head_size=64 % 8 == 0 → is_aligned=true → aligned kernel → check_supported: 12 % 8 ≠ 0 → ORT_ENFORCE crash
3. With bias check: bias_strideM=12 % 8 ≠ 0 → is_aligned=false → unaligned kernel → check_supported: 12 % 4 == 0 → passes ✅

onnxruntime/core/providers/cuda/llm/attention.cc
```diff
@@ -52,589 +67,995 @@ Attention<T>::Attention(const OpKernelInfo& info) : CudaKernel(info) {
...
+  const Tensor* nonpad_kv_seqlen = context->Input<Tensor>(6);  // optional, Opset 24
+
+  // TODO: Support nonpad_kv_seqlen + attn_mask together. The ONNX spec allows both
+  // (additive composition: attn_bias += attn_mask then attn_bias += padding_mask).
+  // CPU implementation supports it. CUDA blocks it because flash has no bias parameter.
+  // To support: route to MEA (set both seqlen_k_ptr and attn_bias) or unfused
+  // (compose two additive biases). ~20 lines. Implement when a real model needs it.
+  ORT_ENFORCE(nonpad_kv_seqlen == nullptr || attn_mask == nullptr,
+              "nonpad_kv_seqlen and attn_mask cannot both be provided.");
+
```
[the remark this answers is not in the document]
titaiwangms: Fallback to MEA and Unfused now.

onnxruntime/test/python/transformers/test_onnx_attention/test_tensorscatter_attention.py
```diff
@@ -0,0 +1,589 @@
...
+            head_size=head_size,
+            nonpad_seqlens=seqlens,
+            scatter_positions=scatter_pos,
+            ep="CUDAExecutionProvider",
+            device="cuda",
+            torch_type=torch.float16,
+            ort_type=TensorProto.FLOAT16,
+            is_causal=is_causal,
+        )
+        numpy.testing.assert_allclose(output, ref_output, rtol=rtol["fp16"], atol=atol["fp16"])
```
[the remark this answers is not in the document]
titaiwangms: Done

onnxruntime/test/python/transformers/test_onnx_attention/test_mha.py
```diff
@@ -920,5 +926,580 @@ def test_mha_bool_mask_fp16(self, name, config):
...
+                    device="cuda",
+                    torch_type=torch.float16,
+                    ort_type=TensorProto.FLOAT16,
+                    causal=config.is_causal == 1,
+                    rtol=rtol["fp16"],
+                    atol=atol["fp16"],
+                )
+        finally:
+            os.environ.pop("ORT_DISABLE_FLASH_ATTENTION", None)
+            os.environ.pop("ORT_DISABLE_MEMORY_EFFICIENT_ATTENTION", None)
```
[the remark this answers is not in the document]
titaiwangms: Supported

onnxruntime/test/providers/cpu/llm/attention_op_test.cc
```diff
@@ -1783,5 +1783,102 @@ TEST(AttentionTest, Attention_NonPadKVSeqLen_ExceedsTotalSeqLen) {
            {}, nullptr, &execution_providers);
 }
 
+// Test combined nonpad_kv_seqlen + bool attn_mask (T26).
```
titaiwangms: T26 is not an effective information here.

onnxruntime/test/providers/cpu/llm/attention_op_test.cc
```diff
@@ -1783,5 +1783,102 @@ TEST(AttentionTest, Attention_NonPadKVSeqLen_ExceedsTotalSeqLen) {
...
+// Both masks should compose additively: nonpad_kv_seqlen masks positions >= valid_len,
+// and attn_mask further masks positions within the valid range.
+// Previously this combination crashed with ORT_ENFORCE; now it falls back to MEA gracefully.
+TEST(AttentionTest, Attention_NonPadKVSeqLen_WithBoolAttnMask) {
```
titaiwangms, line 1847: Consider having nonpad_kv_seqlen + attn_mask tests into the three test file @onnxruntime/test/python/transformers/test_onnx_attention/ as well?

onnxruntime/core/providers/cuda/llm/attention.cc
```diff
@@ -567,6 +571,39 @@ Status Attention<T>::RunMemoryEfficientAttention(
         cuda_stream,
         device_prop.maxThreadsPerBlock));
 
+    // When attn_mask is also provided, convert it to additive attn_bias so MEA
+    // applies both custom right padding (seqlens_k) and the attention mask (attn_bias).
+    if (attn_mask != nullptr) {
```
titaiwangms, line 635: Is there a shared function in MEA already to convert attn_mask to additive attn_bias or I am missing something?

onnxruntime/core/providers/cpu/llm/attention_helper.h
```diff
@@ -115,18 +116,26 @@ inline Status ComputeOutputShapeForAttention(
...
     parameters.has_nonpad_kv_seqlen = false;
     parameters.nonpad_kv_seqlen_data = nullptr;
   }
 
   ORT_ENFORCE(parameters.q_num_heads % parameters.kv_num_heads == 0, "q_num_heads must be a multiple of kv_num_heads. This is required for grouped/multi-query and multi-headed attention.");
+  // TODO: The ONNX spec allows attn_mask last dim to be shorter than total_sequence_length,
+  // with positions beyond the mask padded with -inf. Currently we enforce exact match.
+  // To support: change == to <=, allocate padded buffer, fill remainder with -inf.
+  // See ONNX spec: 'The last dimension can also be shorter than total_sequence_length
+  // and will be padded to total_sequence_length with negative infinity.'
```
titaiwangms, line 145: With nonpad_kv_seqllen + attn_mask supported, do we need to update this?

onnxruntime/core/providers/cuda/llm/attention.cc
```diff
@@ -52,589 +67,1044 @@ Attention<T>::Attention(const OpKernelInfo& info) : CudaKernel(info) {
...
+// Flash Attention dispatch paths:
+//   Path 1: nonpad_kv_seqlen (opset 24 external cache) -> mha_fwd_kvcache
+//   Path 2: past_key + past_value (internal cache decode) -> mha_fwd_kvcache
+//           - Supports: bool mask -> seqlens_k, no mask -> fill past_seq_len
+//           - 4D BNSH: transposes Q/K/V to BSNH before kernel
+//   Path 3: no past, no mask (prompt) -> mha_fwd
+//   Eligibility: fp16/bf16, head_size==v_head_size, no output_qk, no softcap,
+//                no softmax_precision, (no mask OR bool mask + past OR nonpad_kv_seqlen)
+//
+// PERFORMANCE NOTE: ONNX Attention's decode path is inherently ~15-30% slower than
```
titaiwangms: If you consider TensorScatter(kv cache inplace) + Attention(nonpad_kv_seqlen), is it still 15~30% slower? If not, we should clarify here.

onnxruntime/core/providers/cuda/llm/attention.cc
```diff
@@ -52,589 +67,1044 @@ Attention<T>::Attention(const OpKernelInfo& info) : CudaKernel(info) {
...
-      // Check Memory Efficient Attention support (fallback if flash attention not available)
+// ============================================================================
+// RunMemoryEfficientAttention: Direct memory-efficient attention kernel call
+// ============================================================================
+//
+// Memory Efficient Attention (cutlass FMHA) dispatch paths:
+//   Path 1: nonpad_kv_seqlen (opset 24 external cache) -> has_custom_right_padding mode
+//   Path 2: no past, with mask (prompt) -> standard MEA with additive bias
+//   Path 3: no past, no mask (prompt) -> standard MEA
+//   Eligibility: see has_memory_efficient_attention() (SM50+/53+/80+ by dtype,
```
titaiwangms, line 513: Which path does nonpad_kv_seqlen + mask take?

onnxruntime/core/providers/cuda/llm/attention.cc
```diff
@@ -52,589 +67,1044 @@ Attention<T>::Attention(const OpKernelInfo& info) : CudaKernel(info) {
...
-                                             sizeof(int) * parameters.batch_size,
-                                             cudaMemcpyHostToDevice, cuda_stream));
-      }
+  // --- Transpose Q from BNSH to BSNH if 4D input ---
+  IAllocatorUniquePtr<void> q_bsnh_buffer;
+  if (!is_bsnh) {
+    size_t q_bytes = sizeof(T) * parameters.batch_size * parameters.q_sequence_length *
+                     parameters.q_num_heads * parameters.head_size;
+    q_bsnh_buffer = GetScratchBuffer<void>(q_bytes, context->GetComputeStream());
+    ORT_RETURN_IF_ERROR(TransposeBNSHtoBSNH<T>(
```
titaiwangms, line 545: Does all three kernels (flash/MEA/Unfused) transpose 4d to 3d? I thought only flash kernel strictly supports 3d, so we need transpose first?

onnxruntime/core/providers/cuda/llm/attention.cc
```diff
@@ -52,589 +67,1044 @@ Attention<T>::Attention(const OpKernelInfo& info) : CudaKernel(info) {
...
+  if (output_qk != nullptr) {
+    data.output_qk = reinterpret_cast<CudaT*>(output_qk->MutableData<T>());
+  }
+  data.bias = nullptr;
+
+  // Handle attention mask / nonpad_kv_seqlen → attention_bias
+  IAllocatorUniquePtr<void> converted_mask_buffer;
+  if (nonpad_kv_seqlen != nullptr) {
+    // Convert nonpad_kv_seqlen to additive attention bias: [B, q_seq, total_seq]
+    // TODO: Support nonpad_kv_seqlen + attn_mask composition in unfused path.
```
titaiwangms: So only MEA supports nonpad_kv_seqlen + mask?

onnxruntime/core/providers/cuda/llm/attention.cc
```diff
@@ -52,589 +67,1044 @@ Attention<T>::Attention(const OpKernelInfo& info) : CudaKernel(info) {
...
+      return RunMemoryEfficientAttention(context, Q, K, V, attn_mask, past_key, past_value,
+                                         nonpad_kv_seqlen, Y, present_key, present_value, parameters);
+    }
+  }
+#endif
 
-    return onnxruntime::contrib::cuda::QkvToContext<CudaT, CudaT>(
-        device_prop, cublas, cudnn, context->GetComputeStream(), contribop_parameters, data);
+  // Fallback: unfused attention
+  // TODO: Support softcap and softmax_precision on CUDA kernels.
```
titaiwangms: Can you put all TODO with my name - TODO(titaiwang)

onnxruntime/core/providers/cuda/llm/attention.cc
```diff
@@ -52,589 +67,1044 @@ Attention<T>::Attention(const OpKernelInfo& info) : CudaKernel(info) {
...
   }
+  if (parameters.softmax_precision != 0) {
+    return ORT_MAKE_STATUS(ONNXRUNTIME, NOT_IMPLEMENTED,
+                           "softmax_precision is not supported yet in Attention op (CUDA).");
+  }
+  // TODO: Support additional output_qk modes beyond kNone and kQK.
+  // Currently only unfused handles output_qk, and only kNone/kQK modes.
+  if (qk_matmul_output_mode_ != attention_helper::QKMatMulOutputMode::kNone &&
+      qk_matmul_output_mode_ != attention_helper::QKMatMulOutputMode::kQK) {
+    return ORT_MAKE_STATUS(ONNXRUNTIME, NOT_IMPLEMENTED,
```
titaiwangms, line 1145: I wonder if we sholud raise earlier (before flash and MEA) when softmax_precision, softcap , unsupported and qk_matmul_output_mode are met? Is there any considerations here?

onnxruntime/core/providers/cuda/llm/attention.cc
```diff
@@ -52,589 +67,1044 @@ Attention<T>::Attention(const OpKernelInfo& info) : CudaKernel(info) {
...
+        parameters.total_sequence_length,
+        contribop_parameters.mask_filter_value,
+        cuda_stream,
+        device_prop.maxThreadsPerBlock));
+    data.attention_bias = reinterpret_cast<const CudaT*>(converted_mask_buffer.get());
+    // nonpad bias is [B, q_seq, total_seq] → broadcasts over heads but not batch
+    contribop_parameters.broadcast_attn_bias_dim_0 = false;
+    contribop_parameters.broadcast_attn_bias_dim_1 = true;
+  } else if (attn_mask != nullptr) {
+    if (attn_mask->IsDataType<bool>()) {
```
titaiwangms, line 949: I somehow feel like there might be duplication to mask conversion code. Am I missing something?

onnxruntime/core/providers/cuda/llm/attention.cc
```diff
@@ -52,589 +67,1044 @@ Attention<T>::Attention(const OpKernelInfo& info) : CudaKernel(info) {
...
-              "qk_matmul_output_mode must be 0, 1, 2, or 3.");
-  // The default scale depends on the input dimensions. It is set to nan to indicate that it should be computed.
+              "qk_matmul_output_mode must be one of: kNone(-1), kQK(0), kQKMask(1), kQKSoftCap(2), kQKSoftMax(3).");
   scale_ = info.GetAttrOrDefault<float>("scale", std::numeric_limits<T>::quiet_NaN());
   softcap_ = info.GetAttrOrDefault<float>("softcap", 0.0f);
   softmax_precision_ = static_cast<int>(info.GetAttrOrDefault<int64_t>("softmax_precision", 0));
   ORT_ENFORCE(scale_ > 0 || std::isnan(scale_), "scale must be greater than 0 if specified");
+
+  const auto* kernel_options = this->GetAttentionKernelOptions();
+  disable_flash_attention_ = !std::is_same<T, MLFloat16>::value || !kernel_options->UseFlashAttention();
```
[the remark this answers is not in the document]
titaiwangms: Fixed — changed the MLFloat16-only gate to sizeof(T) != 2, which correctly allows both fp16 and bf16 to use Flash. Added a BF16-specific regression test with MEA disabled to guard against future regressions.

onnxruntime/core/providers/cuda/llm/attention.cc
```diff
@@ -52,589 +67,1044 @@ Attention<T>::Attention(const OpKernelInfo& info) : CudaKernel(info) {
...
+        parameters.q_sequence_length,
+        parameters.total_sequence_length,
+        contribop_parameters.mask_filter_value,
+        cuda_stream,
+        device_prop.maxThreadsPerBlock));
+    data.attention_bias = reinterpret_cast<const CudaT*>(converted_mask_buffer.get());
+    // nonpad bias is [B, q_seq, total_seq] → broadcasts over heads but not batch
+    contribop_parameters.broadcast_attn_bias_dim_0 = false;
+    contribop_parameters.broadcast_attn_bias_dim_1 = true;
+  } else if (attn_mask != nullptr) {
```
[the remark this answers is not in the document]
titaiwangms: Fixed — wired up nonpad_kv_seqlen + attn_mask composition in the unfused path using a new LaunchAddBiasInPlace kernel. The unfused path now serves as the true final fallback. Added ORT_ENFORCE guard limiting composition to 2D and 4D head_dim=1 masks (3D per-head masks can't be broadcast over the head-agnostic nonpad bias).

onnxruntime/test/python/transformers/test_onnx_attention/common.py
```diff
@@ -217,15 +245,19 @@ def create_attention_node_and_io(
             elif config.attn_mask_dims == 3:
```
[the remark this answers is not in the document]
titaiwangms: Fixed — updated test mask creation from [batch, total_seq] to [q_seq, total_seq] per ONNX spec. Added a batch_size=4, q_seq=1 guard case that would have caught this. Also added 3D mask dim[-2] == q_seq validation in attention_helper.h.

onnxruntime/test/providers/cpu/llm/attention_op_test.cc
titaiwangms: Done — added 5 test cases in test_tensorscatter_attention.py covering MHA/GQA, batch sizes 1-4, bool/float masks. Both CPU and CUDA test classes.

onnxruntime/core/providers/cuda/llm/attention.cc
titaiwangms: Extracted ConvertAttnMaskToBias helper — reduced 3 duplicate blocks to 1 shared function, saving ~42 lines.

onnxruntime/core/providers/cpu/llm/attention_helper.h
titaiwangms: Audited — no changes needed. The helper validates nonpad and mask shapes independently and never blocks their co-existence.

onnxruntime/core/providers/cuda/llm/attention.cc
titaiwangms: Updated comment — clarified: 15-30% with internal cache (memset+copy overhead), 5-15% with external cache via TensorScatter (only XQA kernel gap remains).

onnxruntime/core/providers/cuda/llm/attention.cc
titaiwangms: Added comment — MEA path (Flash excluded due to no bias param), unfused now also supports it as fallback.

onnxruntime/core/providers/cuda/llm/attention.cc
titaiwangms: Added 10-line comment explaining per-kernel behavior: Flash strictly requires BSNH (transpose needed), MEA and Unfused handle both BSNH and BNSH natively.

onnxruntime/core/providers/cuda/llm/attention.cc
titaiwangms: Updated — both MEA and Unfused now support it. Flash still cannot (no bias parameter in mha_fwd_kvcache).

onnxruntime/core/providers/cuda/llm/attention.cc
titaiwangms: Implemented — moved early-reject before the dispatch cascade. Fail fast with clear error messages instead of falling through all 3 paths.

onnxruntime/core/providers/cuda/llm/attention.cc
titaiwangms: Addressed via T32's shared helper extraction. Remaining structural duplication across paths deferred to contrib reorg issue.

onnxruntime/core/providers/cuda/llm/attention.cc
```diff
@@ -52,589 +67,1100 @@ Attention<T>::Attention(const OpKernelInfo& info) : CudaKernel(info) {
...
+  const Tensor* past_value = context->Input<Tensor>(5);
+  const Tensor* nonpad_kv_seqlen = context->Input<Tensor>(6);  // optional, Opset 24
+
+  // When both nonpad_kv_seqlen and attn_mask are provided, Flash Attention cannot handle
+  // the combination (no bias parameter). Route to MEA or Unfused which support composition.
+  if (nonpad_kv_seqlen != nullptr && attn_mask != nullptr) {
+    LOGS_DEFAULT(WARNING) << "Both nonpad_kv_seqlen and attn_mask provided. "
+                          << "Flash Attention does not support this combination; "
+                          << "falling back to Memory Efficient Attention or Unfused path.";
+  }
```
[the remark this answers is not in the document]
titaiwangms: Agreed — changed to LOGS_DEFAULT(VERBOSE). This is expected dispatch behavior for spec-legal models, not an error condition.

onnxruntime/test/python/transformers/test_onnx_attention/test_tensorscatter_attention.py
```diff
@@ -0,0 +1,965 @@
...
+            head_size=head_size,
+            nonpad_seqlens=seqlens,
+            scatter_positions=scatter_pos,
+            ep="CPUExecutionProvider",
+            device="cpu",
+            torch_type=torch.float32,
+            ort_type=TensorProto.FLOAT,
+            is_causal=is_causal,
+        )
+        numpy.testing.assert_allclose(output, ref_output, rtol=cpu_fp32_rtol, atol=cpu_fp32_atol)
```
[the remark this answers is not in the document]
titaiwangms: Good catch — added present_key/present_value validation to all tensorscatter test methods. Expected values computed as BSNH→BNSH transpose of the updated cache reference.

onnxruntime/test/python/transformers/test_onnx_attention/test_tensorscatter_attention.py
```diff
@@ -0,0 +1,965 @@
...
+    head_size,
+    nonpad_seqlens,
+    scatter_positions,
+    ep,
+    device,
+    torch_type,
+    ort_type,
+    is_causal=0,
+    std=0.2,
+):
```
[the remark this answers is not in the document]
titaiwangms: Removed from both function signatures and all call sites. Audited all other test files — no similar issues found.

onnxruntime/test/python/transformers/test_onnx_attention/test_gqa.py
```diff
@@ -937,29 +937,414 @@ def test_gqa_prompt_padding_mea(self, name, config):
...
+            atol=atol["fp16"],
+        )
+
+    @parameterized.expand(gqa_4d_bnsh_past_test_cases())
+    def test_gqa_4d_bnsh_decode(self, name, config):
+        os.environ["ORT_DISABLE_FLASH_ATTENTION"] = "0"
+        parity_check_gqa_past(
             config=config,
-            past_seqlens=past_seqlens,
             ep="CUDAExecutionProvider",
```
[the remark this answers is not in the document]
titaiwangms: Done

onnxruntime/test/python/transformers/test_onnx_attention/test_gqa.py
```diff
@@ -937,29 +937,414 @@ def test_gqa_prompt_padding_mea(self, name, config):
...
+    The C++ attention op detects 4D inputs and sets transpose_output=false.
+    Flash/MEA always expect BSNH, so the dispatcher transposes Q internally.
+    """
+
+    @parameterized.expand(gqa_4d_bnsh_test_cases())
+    def test_gqa_4d_bnsh_prompt(self, name, config):
+        os.environ["ORT_DISABLE_FLASH_ATTENTION"] = "0"
+        parity_check_gqa_prompt(
+            config=config,
+            ep="CUDAExecutionProvider",
```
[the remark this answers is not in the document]
titaiwangms: Done

onnxruntime/test/python/transformers/test_onnx_attention/test_gqa.py
```diff
@@ -937,29 +937,414 @@ def test_gqa_prompt_padding_mea(self, name, config):
...
+
+
+@unittest.skipIf(not has_cuda_device(53), "CUDA device not available, skipping tests.")
+class TestONNXAttentionGQANonpadKVSeqlenMEA(unittest.TestCase):
+    """Test ONNX Attention op (opset 24) GQA path with nonpad_kv_seqlen (Memory Efficient Attention)."""
+
+    @parameterized.expand(gqa_nonpad_kv_seqlen_test_cases())
+    def test_gqa_nonpad_kv_seqlen_mea(self, name, config, seqlens):
         os.environ["ORT_DISABLE_FLASH_ATTENTION"] = "1"
+        nonpad_seqlens = torch.tensor(seqlens, dtype=torch.int64, device="cuda")
```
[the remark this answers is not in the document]
titaiwangms: Done

onnxruntime/test/python/transformers/test_onnx_attention/test_gqa.py
```diff
@@ -937,29 +937,414 @@ def test_gqa_prompt_padding_mea(self, name, config):
...
+    """Test ONNX Attention op (opset 24) GQA path with nonpad_kv_seqlen (Flash Attention).
+
+    Requires SM80+: tests explicitly force Flash via ORT_DISABLE_FLASH_ATTENTION=0.
+    """
+
+    @parameterized.expand(gqa_nonpad_kv_seqlen_test_cases())
+    def test_gqa_nonpad_kv_seqlen_flash(self, name, config, seqlens):
+        os.environ["ORT_DISABLE_FLASH_ATTENTION"] = "0"
+        nonpad_seqlens = torch.tensor(seqlens, dtype=torch.int64, device="cuda")
+
```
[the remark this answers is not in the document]
titaiwangms: Done

onnxruntime/contrib_ops/cuda/bert/cutlass_fmha/fmha_launch_template.h
titaiwangms: Feel free to comment more. I will make follow-ups if needed.

onnxruntime/core/providers/cuda/llm/attention.cc
titaiwangms: Feel free to comment more. I will make follow-ups if needed.

onnxruntime/core/providers/cuda/llm/attention_mask_impl.cu
```diff
@@ -72,32 +74,41 @@ __global__ void ConvertMaskToSeqlensKernel(
...
+    for (int i = 1; i < total_seq_len; ++i) {
+      bool current = mask_row[i];
+
+      if (!found_first_false && !current) {
+        // Found first False - this is where padding starts
+        seq_len = i;
+        found_first_false = true;
+      } else if (found_first_false && current) {
+        // Found True after False - mask is not contiguous (invalid)
+        CUDA_KERNEL_ASSERT(false);  // mask must be contiguous (no True after False)
```
tianleiwu, line 96: CUDA_KERNEL_ASSERT is noop in release build.

titaiwangms: Thanks, I will create a follow-up.

titaiwangms: https://github.com/microsoft/onnxruntime/issues/27612

PR #27543, opened 2026-03-03
Fix RoiAlign heap out-of-bounds read via unchecked batch_indices

tianleiwu:
# Fix RoiAlign heap out-of-bounds read via unchecked batch_indices

## Description

Add value-range validation for `batch_indices` in the RoiAlign operator to prevent out-of-bounds heap reads from maliciously crafted ONNX models.

`CheckROIAlignValidInput()` previously validated tensor shapes but never checked that the **values** in `batch_indices` fall within `[0, batch_size)`. An attacker could supply `batch_indices` containing values exceeding the batch dimension of the input tensor `X`, causing the kernel to read arbitrary heap memory at:

- **CPU:** `roialign.cc:212` — `roi_batch_ind` used as unchecked index into `bottom_data`
- **CUDA:** `roialign_impl.cu:109` — `batch_indices_ptr[n]` used as unchecked index into `bottom_data` on GPU

## Impact

- **Vulnerability type:** Heap out-of-bounds read
- **Impact:** Arbitrary heap memory read, potential information disclosure, program crash
- **Trigger:** Construct `batch_indices` with values ≥ `batch_size` or < 0
- **Affected providers:** CPU and CUDA (both call `CheckROIAlignValidInput()`)

### `onnxruntime/core/providers/cpu/object_detection/roialign.cc`
- Added per-element bounds check in `CheckROIAlignValidInput()`: each `batch_indices[i]` must satisfy `0 <= value < X.shape[0]`
- Returns `INVALID_ARGUMENT` with a descriptive error message on violation
- Guarded by `batch_indices_ptr->Location().device.Type() == OrtDevice::CPU` so it only runs when the tensor data is host-accessible (CPU EP and CropAndResize). For the CUDA EP, `batch_indices` lives in GPU memory and cannot be safely dereferenced on the host.

### `onnxruntime/test/providers/cpu/object_detection/roialign_test.cc`
- Added `BatchIndicesOutOfRange` test: `batch_indices={1}` with `batch_size=1` (exercises `>= batch_size` path)
- Added `BatchIndicesNegative` test: `batch_indices={-1}` (exercises `< 0` path)

## Known Limitation

The CUDA execution path is **not** protected by this bounds check because `batch_indices` is a GPU tensor and cannot be read on the host. Adding a device-side bounds check would require passing `batch_size` into the CUDA kernel — this is tracked as a follow-up.

Note: Using `.InputMemoryType(OrtMemTypeCPUInput, 2)` was considered but rejected because it would force a GPU→CPU transfer of `batch_indices`, breaking CUDA graph capture for models like Masked R-CNN where `batch_indices` is produced by upstream GPU ops.

## Validation

- Full `RoiAlignTest.*` suite passes (12/12 tests) on CPU build
- Full `RoiAlignTest.*` suite passes (12/12 tests) on CUDA build
- No regressions in existing positive or negative tests

### DIFF
--- a/onnxruntime/test/providers/cpu/object_detection/roialign_test.cc
+++ b/onnxruntime/test/providers/cpu/object_detection/roialign_test.cc
@@ -829,7 +829,9 @@ TEST(RoiAlignTest, BatchIndicesOutOfRange) {
   test.AddInput<int64_t>("batch_indices", {1}, {1});  // <-- failure condition
   test.AddOutput<float>("Y", {1, 1, 2, 2}, {0.f, 0.f, 0.f, 0.f});
 
-  test.Run(OpTester::ExpectResult::kExpectFailure, "batch_indices value 1 at index 0 is out of range [0, 1)");
+  std::vector<std::unique_ptr<IExecutionProvider>> execution_providers;
+  execution_providers.push_back(DefaultCpuExecutionProvider());
+  test.Run(OpTester::ExpectResult::kExpectFailure, "batch_indices value 1 at index 0 is out of range [0, 1)", {}, nullptr, &execution_providers);
 }
 
 TEST(RoiAlignTest, BatchIndicesNegative) {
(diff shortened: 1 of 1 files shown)


yuslepukhin (approved):
LGTM

onnxruntime/core/providers/cpu/object_detection/roialign.cc
```diff
@@ -294,6 +294,41 @@ Status CheckROIAlignValidInput(const Tensor* X_ptr, const Tensor* rois_ptr, cons
     return Status(common::ONNXRUNTIME, common::INVALID_ARGUMENT,
                   "First dimension (num_rois) of batch_indices and rois don't match");
   }
+
+  // Validate batch_indices values are within [0, batch_size).
+  // Only check when the tensor data is accessible from the host (CPU).
+  // For GPU tensors (e.g. CUDA EP), Data<T>() returns a device pointer
+  // that cannot be safely dereferenced on the host. A device-side bounds
+  // check for the CUDA path would require passing batch_size into the
```
hariharans29, line 302: Additionally, maybe add a TODO stating that we need a CUDA_ASSERT(...) in the kernel here once we have batch size data in the kernel: https://github.com/microsoft/onnxruntime/blob/df8f4a7d0ad47155bb9e48437c788e107d5717e4/onnxruntime/core/providers/cuda/object_detection/roialign_impl.cu#L108C16-L108C30

?

hariharans29 (approved):

onnxruntime/core/providers/cpu/object_detection/roialign.cc
tianleiwu: I will address it in another PR.

PR #27553, opened 2026-03-04
Lazy initialization of per-session thread pools on first Run/RunAsync

### LINKED ISSUE #27150: [Improvement] Unnecessary Thread Pool Creation even when using only GPU in model graph nodes
#### Problem Description

When using execution providers that don't require CPU thread pools (e.g., TensorRT EP), thread pools are still created during session construction, even though they won't be used during graph execution.

#### Current Behavior && Root Cause
- Thread pools (intra-op and inter-op) are created in [ConstructorCommon()](https://github.com/microsoft/onnxruntime/blob/84e657a11d088152526a74d05258cb141f2ecb4f/onnxruntime/core/session/inference_session.cc#L367) at session construction time
- This happens **before** execution provider assignment, which occurs later in [Initialize()](https://github.com/microsoft/onnxruntime/blob/727db0d3dc9f7dc5958891d80c1073ef7190f316/onnxruntime/core/session/inference_session.cc#L2086) → `TransformGraph()` → `GraphPartitioner::Partition()`
- For GPU-only EPs with `ORT_SEQUENTIAL` mode, thread pools are never actually used during execution

#### Impact
- Unnecessary resource allocation (thread creation, memory)
- Thread pools remain idle for the entire session lifetime (although they are blocked)

#### Propose Solution
Implement **lazy initialization** for thread pools:
1. **Defer thread pool creation** until after EP assignment in `Initialize()`
2. **Determine necessity** based on, during `Initialize()`:
   - Execution mode (`ORT_PARALLEL` requires inter-op thread pool)
   - Assigned execution providers (CPU EP nodes require thread pools)
   - Graph structure (presence of CPU nodes)
3. **Create thread pools only when needed**

### To reproduce

None

### Urgency

Not urgent

### Platform

Linux

### OS Version

Ubuntu 24.04

### ONNX Runtime Installation

Released Package

1.22.0

### ONNX Runtime API

Python

### Architecture

X64

### Execution Provider

TensorRT

_No response_

TsofnatMaman:
### Description

This change introduces lazy initialization of per-session thread pools.

Previously, when `use_per_session_threads` was enabled, thread pools were created during session initialization. This could result in unnecessary allocations in scenarios where the session does not execute CPU kernels.

With this change:
- Per-session thread pools are no longer created during `Initialize()`.
- Instead, they are created on-demand during the first execution path (`Run` / `RunAsync`).
- Once created, thread pools are propagated to `SessionState` (including subgraphs) and reused for subsequent runs.

Additional updates:
- `SessionState` now supports late initialization of thread pools and propagates them to subgraph session states.
- Thread-safe one-time initialization is ensured using a guarded lazy-init mechanism.

Tests are added to cover:
- lazy initialization behavior before and after first run
- `Run` and `RunAsync` paths
- subgraph propagation
- concurrent first-run scenarios
- external thread pool usage

### Motivation and Context

Fixes #27150

This change avoids unnecessary per-session thread pool allocation in cases where they are not needed, such as GPU-only execution scenarios.

Previously, thread pools were created during session initialization, before execution begins. By deferring creation until the first execution, the session only allocates thread pools when they are actually required.

### DIFF
--- a/onnxruntime/core/framework/session_state.cc
+++ b/onnxruntime/core/framework/session_state.cc
@@ -1152,8 +1152,9 @@ void SessionState::BindThreadPoolsOnce(concurrency::ThreadPool* thread_pool,
       (current_inter_op_thread_pool == nullptr || current_inter_op_thread_pool == inter_op_thread_pool),
       "SessionState thread pools must be bound once with stable pointers.");
 
-  // Release stores publish both pointers to lock-free readers via GetThreadPool/
-  // GetInterOpThreadPool acquire loads.
+  // Release stores publish each pointer independently to lock-free readers via
+  // GetThreadPool()/GetInterOpThreadPool() acquire loads. This does not provide
+  // an atomic pair snapshot.
   thread_pool_.store(thread_pool, std::memory_order_release);
   inter_op_thread_pool_.store(inter_op_thread_pool, std::memory_order_release);
 
--- a/onnxruntime/core/framework/session_state.h
+++ b/onnxruntime/core/framework/session_state.h
@@ -296,8 +296,9 @@ class SessionState {
   /// Return SessionState for the given Node index and attribute name if found.
   const SessionState* GetSubgraphSessionState(NodeIndex index, const std::string& attribute_name) const;
 
-  // Acquire semantics pair with BindThreadPoolsOnce release stores so readers
-  // running without external synchronization observe fully-published pool pointers.
+  // Each getter uses an acquire load that pairs with the corresponding
+  // BindThreadPoolsOnce release store. Pointers are published independently, so
+  // this does not provide an atomic pair snapshot.
   concurrency::ThreadPool* GetThreadPool() const noexcept { return thread_pool_.load(std::memory_order_acquire); }
   concurrency::ThreadPool* GetInterOpThreadPool() const noexcept { return inter_op_thread_pool_.load(std::memory_order_acquire); }
 
@@ -400,11 +401,14 @@ class SessionState {
   friend class InferenceSession;
   friend class SessionStateThreadPoolBindingTestAccessor;
 
-  // One-shot late binding for execution thread pools.
+  // One-shot publish for execution thread pools. Must be called before any
+  // execution starts.
   // - No ownership transfer: SessionState stores raw pointers only.
   // - Allows nullptr -> final pointer transition once.
   // - Allows idempotent rebinding to the same pointers.
   // - Rejects rebinding to different non-null pointers.
+  // - Publishes each pointer with release semantics; readers use acquire loads
+  //   via GetThreadPool()/GetInterOpThreadPool().
   // - Recursively publishes the same pointers to all subgraph SessionStates.
   void BindThreadPoolsOnce(concurrency::ThreadPool* thread_pool,
                            concurrency::ThreadPool* inter_op_thread_pool);
@@ -553,7 +557,9 @@ class SessionState {
 
   // Thread-pool pointers are set either in the constructor or via one-shot
   // BindThreadPoolsOnce publication on first Run/RunAsync when per-session
-  // thread pools are lazily created. Either pointer can be nullptr.
+  // thread pools are lazily created. Either pointer can be nullptr. Pointers are
+  // published independently; a coherent required pair is gated by
+  // the session-level readiness flag in InferenceSession.
   std::atomic<concurrency::ThreadPool*> thread_pool_{nullptr};
   std::atomic<concurrency::ThreadPool*> inter_op_thread_pool_{nullptr};
 
--- a/onnxruntime/core/session/inference_session.cc
+++ b/onnxruntime/core/session/inference_session.cc
@@ -594,13 +594,13 @@ void InferenceSession::CreateIntraOpThreadPoolIfNeeded() {
   thread_pool_ = concurrency::CreateThreadPool(&Env::Default(), to, concurrency::ThreadPoolType::INTRA_OP);
 }
 
-void InferenceSession::CreateInterOpThreadPoolIfNeeded() {
+Status InferenceSession::CreateInterOpThreadPoolIfNeeded() {
   if (external_inter_op_thread_pool_ || inter_op_thread_pool_) {
-    return;
+    return Status::OK();
   }
 
   if (session_options_.execution_mode != ExecutionMode::ORT_PARALLEL) {
-    return;
+    return Status::OK();
   }
 
   bool allow_inter_op_spinning =
@@ -613,6 +613,8 @@ void InferenceSession::CreateInterOpThreadPoolIfNeeded() {
 #endif
 
   OrtThreadPoolParams to = session_options_.inter_op_param;
+  // Preserve historical assignment logic. In this lazy inter-op path we are in
+  // ORT_PARALLEL mode, so this remains intentionally false.
   to.auto_set_affinity = to.thread_pool_size == 0 && session_options_.execution_mode == ExecutionMode::ORT_SEQUENTIAL;
   inter_thread_pool_name_ = BuildPerSessionThreadPoolName(to.name, ORT_TSTR("inter-op"));
   to.name = inter_thread_pool_name_.c_str();
@@ -620,21 +622,34 @@ void InferenceSession::CreateInterOpThreadPoolIfNeeded() {
 
   PopulateThreadPoolParamsFromSessionOptions(to, "inter op");
 
-  inter_op_thread_pool_ = concurrency::CreateThreadPool(&Env::Default(), to, concurrency::ThreadPoolType::INTER_OP);
-  if (inter_op_thread_pool_ == nullptr) {
-    LOGS(*session_logger_, INFO) << "Failed to create the inter-op thread pool for the parallel executor, setting ExecutionMode to SEQUENTIAL";
-    session_options_.execution_mode = ExecutionMode::ORT_SEQUENTIAL;
+  constexpr const char* kInterOpThreadPoolCreateFailureMessage =
+      "Failed to create inter-op thread pool while execution_mode=ORT_PARALLEL";
+  ORT_TRY {
+    inter_op_thread_pool_ =
+        concurrency::CreateThreadPool(&Env::Default(), to, concurrency::ThreadPoolType::INTER_OP);
   }
+  ORT_CATCH(const std::exception& ex) {
+    return ORT_MAKE_STATUS(ONNXRUNTIME, FAIL, kInterOpThreadPoolCreateFailureMessage, ". ", ex.what());
+  }
+  ORT_CATCH(...) {
+    return ORT_MAKE_STATUS(ONNXRUNTIME, FAIL, kInterOpThreadPoolCreateFailureMessage);
+  }
+
+  if (inter_op_thread_pool_ == nullptr) {
+    return ORT_MAKE_STATUS(ONNXRUNTIME, FAIL, kInterOpThreadPoolCreateFailureMessage);
+  }
+
+  return Status::OK();
 }
 
-void InferenceSession::EnsureThreadPoolsCreatedForRun() {
+Status InferenceSession::EnsureThreadPoolsCreatedForRun() {
   if (!use_per_session_threads_) {
-    return;
+    return Status::OK();
   }
 
   // Fast path for Run/RunAsync after first-use publication succeeds.
   if (thread_pools_ready_for_run_.load(std::memory_order_acquire)) {
-    return;
+    return Status::OK();
   }
 
   // Slow path: first-use creation/publication is serialized under session_mutex_.
@@ -642,14 +657,14 @@ void InferenceSession::EnsureThreadPoolsCreatedForRun() {
   // pointers cannot race with session setup transitions.
   std::lock_guard<std::mutex> l(session_mutex_);
   if (thread_pools_ready_for_run_.load(std::memory_order_relaxed)) {
-    return;
+    return Status::OK();
   }
 
   // Run and RunAsync both come through this gate so lazy creation semantics are
   // identical for synchronous and asynchronous execution.
   CreateIntraOpThreadPoolIfNeeded();
   if (session_options_.execution_mode == ExecutionMode::ORT_PARALLEL) {
-    CreateInterOpThreadPoolIfNeeded();
+    ORT_RETURN_IF_ERROR(CreateInterOpThreadPoolIfNeeded());
   }
 
   auto* intra_op_tp = GetIntraOpThreadPoolToUse();
@@ -659,8 +674,9 @@ void InferenceSession::EnsureThreadPoolsCreatedForRun() {
               "This is an internal ORT bug.");
 
   if (session_state_) {
-    // One-shot publication into SessionState (and all subgraphs). After this call,
-    // SessionState getters are safe for lock-free readers due release/acquire order.
+    // One-shot publication into SessionState (and all subgraphs). Thread-pool
+    // pointers are published independently with release stores, and readers use
+    // acquire loads.
     session_state_->BindThreadPoolsOnce(intra_op_tp, inter_op_tp);
   }
 
@@ -669,6 +685,8 @@ void InferenceSession::EnsureThreadPoolsCreatedForRun() {
   if (AreRequiredPerSessionThreadPoolsAvailable(intra_op_tp, inter_op_tp)) {
     thread_pools_ready_for_run_.store(true, std::memory_order_release);
   }
+
+  return Status::OK();
 }
 
 void InferenceSession::TraceSessionOptions(const SessionOptions& session_options, bool captureState, const logging::Logger& logger) {
@@ -3102,7 +3120,7 @@ Status InferenceSession::Run(const RunOptions& run_options,
   }
 
   // Run and RunAsync intentionally share the same lazy thread-pool gate.
-  EnsureThreadPoolsCreatedForRun();
+  ORT_RETURN_IF_ERROR_SESSIONID_(EnsureThreadPoolsCreatedForRun());
 
   // Increment/decrement concurrent_num_runs_ and control
   // session threads spinning as configured. Do nothing for graph replay except the counter.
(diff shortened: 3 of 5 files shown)


TsofnatMaman:
@tianleiwu Could you please trigger the CI for this PR? Thanks!

TsofnatMaman:
@tianleiwu @skottmckay — could you please take a look and help trigger CI when you have a chance? Thanks!

TsofnatMaman:
Thanks for the review @tianleiwu. I addressed the feedback in this update:

- Fixed the compile issues in `InitializeThreadPoolsIfNeeded`:
  - recomputed `set_denormal_as_zero` in the new scope
  - replaced the remaining `session_options` references with `session_options_`

- Preserved `SessionState` invariants without adding broad friend access:
  - instead of using `friend class InferenceSession`, which would grant access to all private members,
    I introduced a single initialization-only API:
    `SessionState::InitializeThreadPools(intra, inter)`
  - this keeps the thread pool members encapsulated while still allowing late binding
  - the API uses one-shot enforcement so the thread pool pointers can only be initialized once
    (or rebound to the same values), which preserves the original invariant

- Made the optimization configurable (enabled by default):
  - added a new session config option:
    `session.create_threadpools_only_for_cpu_nodes`
  - when enabled, per-session thread pools are created only if CPU-assigned nodes are detected
  - when disabled, existing behavior remains unchanged
  - I chose to gate this behind a config option because the detection is based on finalized EP assignment
    and may not capture every case of internal CPU fallback inside non-CPU EPs
  - this allows users who expect frequent CPU fallback to keep the current conservative behavior and avoid
    any risk from the optimization

- Updated CPU-node detection to recurse into subgraphs:
  - `GraphHasCpuNodesInternal` now walks nested subgraphs recursively

- Documented the limitation explicitly:
  - the check is based on finalized EP assignment and may not detect internal CPU fallback
    inside non-CPU execution providers

- Added a warning if thread pools are created while `session_state_` is still null

- Added test coverage for:
  - unchanged default behavior without the new config
  - config enabled + no CPU nodes
  - config enabled + CPU nodes in the main graph
  - config enabled + control-flow/subgraph cases, including explicit recursive detection

onnxruntime/core/framework/session_state.h
```diff
@@ -296,6 +296,10 @@ class SessionState {
 
   concurrency::ThreadPool* GetThreadPool() const noexcept { return thread_pool_; }
   concurrency::ThreadPool* GetInterOpThreadPool() const noexcept { return inter_op_thread_pool_; }
+  void SetThreadPool(concurrency::ThreadPool* thread_pool) noexcept { thread_pool_ = thread_pool; }
+  void SetInterOpThreadPool(concurrency::ThreadPool* inter_op_thread_pool) noexcept {
+    inter_op_thread_pool_ = inter_op_thread_pool;
+  }
```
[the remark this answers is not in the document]
TsofnatMaman: Addressed via a one-shot initialization-only API.

Instead of leaving public mutable setters or adding broad friend access, I replaced
them with `SessionState::InitializeThreadPools(intra, inter)`, which enforces
one-time binding (or idempotent re-binding to the same values). This keeps the
thread pool pointers from being mutated arbitrarily after initialization.

onnxruntime/core/framework/session_state.h
```diff
@@ -297,6 +297,8 @@ class SessionState {
   concurrency::ThreadPool* GetThreadPool() const noexcept { return thread_pool_; }
   concurrency::ThreadPool* GetInterOpThreadPool() const noexcept { return inter_op_thread_pool_; }
 
+  void InitializeThreadPools(concurrency::ThreadPool* intra, concurrency::ThreadPool* inter);
```
yuslepukhin: Need good documentation, what it does, why it is needed, when it is called. #Resolved

onnxruntime/core/framework/session_state.cc
```diff
@@ -1133,6 +1133,28 @@ const SessionState* SessionState::GetSubgraphSessionState(onnxruntime::NodeIndex
   return const_cast<SessionState*>(this)->GetMutableSubgraphSessionState(index, attribute_name);
 }
 
+void SessionState::InitializeThreadPools(concurrency::ThreadPool* intra, concurrency::ThreadPool* inter) {
+  // NOTE: This enforce also runs recursively for each subgraph SessionState
+  // via the loop below. All subgraph SessionStates are expected to hold nullptr
+  // or the same pool pointers as the parent - never a different non-null pool.
+  ORT_ENFORCE((thread_pool_ == nullptr || thread_pool_ == intra) &&
+              (inter_op_thread_pool_ == nullptr || inter_op_thread_pool_ == inter));
```
yuslepukhin: The ORT_ENFORCE "one-shot" assertion in session_state.cc:1136 is a runtime check, not compile-time. 

Code that was previously guaranteed immutability now relies on discipline.  #Resolved

include/onnxruntime/core/session/onnxruntime_session_options_config_keys.h
```diff
@@ -41,6 +41,15 @@ static const char* const kOrtSessionOptionsConfigSaveModelFormat = "session.save
 // and that's recommended because turning this option on may hurt model accuracy.
 static const char* const kOrtSessionOptionsConfigSetDenormalAsZero = "session.set_denormal_as_zero";
 
+// Controls lazy creation of per-session thread pools.
+// Default is "1":
```
yuslepukhin: Default behavior change — silent optimization change for all users (HIGH)
The new config defaults to "1", meaning every existing user gets different behavior without opting in. Previously, use_per_session_threads=true guaranteed thread pools were created. Now they may not be.

While this is the purpose of the PR, defaulting a new optimization to enabled is risky for a first release. Any EP that internally falls back to CPU would silently lose thread pool parallelism. Recommendation: Consider defaulting to "0" (legacy) initially and letting users opt in, or at least make it prominent in release notes. #Resolved

onnxruntime/core/session/inference_session.cc
```diff
@@ -3343,6 +3464,7 @@ common::Status InferenceSession::RunAsync(const RunOptions* run_options,
                                           RunAsyncCallbackFn callback,
                                           void* user_data) {
   size_t num_fetches = fetch_names.size();
+  EnsureIntraOpThreadPoolInitialized();
```
yuslepukhin: > EnsureIntraOpThreadPoolInitialized(); [](http://example.com/codeflow?start=1&length=38)

Run() vs RunAsync() asymmetry
RunAsync() calls EnsureIntraOpThreadPoolInitialized() to lazily create the intra-op pool if it wasn't created during Initialize(). However, Run() does not have equivalent logic.

This creates an inconsistency: calling RunAsync() on a GPU-only model will create thread pools; calling Run() will not. While Run() on a GPU-only model shouldn't need thread pools, the behavior divergence between the two codepaths is a maintenance hazard and could cause confusion.

If RunAsync lazily creates the intra-op pool, it also calls session_state_->InitializeThreadPools() with GetInterOpThreadPoolToUse() — which returns nullptr since no inter-op pool was created. This is fine for RunAsync's own purposes, but if the Run() call inside RunAsync later needs the inter-op pool (e.g., ORT_PARALLEL mode), it won't be there.
 #Resolved

onnxruntime/core/session/inference_session.cc
```diff
@@ -3030,8 +3146,13 @@ Status InferenceSession::Run(const RunOptions& run_options,
...
                                 !cached_execution_provider_for_graph_replay_.IsGraphCaptured(graph_annotation_id);
-  auto* intra_tp = (control_spinning) ? thread_pool_.get() : nullptr;
-  auto* inter_tp = (control_spinning) ? inter_op_thread_pool_.get() : nullptr;
+  concurrency::ThreadPool* intra_tp = nullptr;
+  concurrency::ThreadPool* inter_tp = nullptr;
+  if (control_spinning) {
+    std::lock_guard<std::mutex> lock(session_mutex_);
+    intra_tp = thread_pool_.get();
+    inter_tp = inter_op_thread_pool_.get();
+  }
```
yuslepukhin, line 3136: This correctly uses session_mutex_ to read thread_pool_. However, GetIntraOpThreadPoolToUse() and GetInterOpThreadPoolToUse() (called elsewhere without locks, e.g., inference_session.cc:4065 in AddPredefinedTransformers) perform unsynchronized reads of thread_pool_.get(). 

This is only safe because those calls happen during single-threaded Initialize(). The code should document this assumption clearly. #Resolved

onnxruntime/test/framework/inference_session_test.cc
```diff
@@ -2379,6 +2418,466 @@ TEST(InferenceSessionTests, InvalidSessionEnvCombination) {
...
+  ASSERT_STATUS_OK(session_object.RegisterExecutionProvider(std::make_unique<MainGraphIfOnlyExecutionProvider>()));
+
+  ASSERT_STATUS_OK(session_object.Load(model_file_name));
+  const auto init_status = session_object.Initialize();
+  // Initialize() is expected to fail here due to a known limitation of the
+  // artificial MainGraphIfOnlyExecutionProvider (missing subgraph kernel info).
+  // This failure is intentional and unrelated to the thread-pool behavior under test.
+  // The assertions below verify that thread pools were still correctly created
+  // before the failure occurred.
+  ASSERT_STATUS_NOT_OK(init_status);
```
yuslepukhin: Test 6b expects Initialize() failure — fragile 
The subgraph test DefaultLazyInitPerSessionThreadPoolsWithCpuNodesInSubgraph inference_session_test.cc:2828 that intentionally causes Initialize() to fail, then asserts thread pools were still created. This is testing an intermediate state of a failed initialization — a fragile pattern. If the implementation changes the order of operations, this test breaks for the wrong reasons.

yuslepukhin:
>   bool set_denormal_as_zero =

set_denormal_as_zero recomputed: The original code computed this once in the constructor; the new code recomputes it in CreateIntraOpThreadPoolIfNeeded() and InitializeThreadPoolsIfNeeded(). While still correct at present (config is immutable), but slightly less clean and fragile. #Resolved

---
Refers to: onnxruntime/core/session/inference_session.cc:439 in d46d902. [](commit_id = d46d9021bc6998696e25a6d4b615b75b1f263d5a, deletion_comment = False)

onnxruntime/core/session/inference_session.cc
```diff
@@ -606,6 +524,171 @@ void InferenceSession::ConstructorCommon(const SessionOptions& session_options,
...
+}
+
+bool InferenceSession::AreRequiredPerSessionThreadPoolsAvailable(concurrency::ThreadPool* intra_op_tp,
+                                                                 concurrency::ThreadPool* inter_op_tp) const {
+  const bool intra_op_ready = intra_op_tp != nullptr;
+  const bool inter_op_ready = session_options_.execution_mode != ExecutionMode::ORT_PARALLEL || inter_op_tp != nullptr;
+  return intra_op_ready && inter_op_ready;
+}
+
+void InferenceSession::CreateIntraOpThreadPoolIfNeeded() {
```
yuslepukhin, line 563: >CreateIntraOpThreadPoolIfNeeded [](http://example.com/codeflow?start=23&length=31)

It seems to me that this function breaks for thread_pool_size == 1
When thread_pool_size == 1, concurrency::CreateThreadPool legitimately returns nullptr (ORT runs kernels single-threaded inline). The current logic in AreRequiredPerSessionThreadPoolsAvailable:

This returns false for a legitimately-null thread pool.

Consequence:

thread_pools_ready_for_run_ never becomes true
Every Run() takes the slow path through session_mutex_
CreateIntraOpThreadPoolIfNeeded() re-executes the full parameter setup + CreateThreadPool call each time (because thread_pool_ unique_ptr is still null), only to get nullptr again
This is a performance regression for single-threaded configurations. The mutex acquisition and repeated creation attempts on every Run() add overhead that didn't exist before.

include/onnxruntime/core/optimizer/graph_transformer_utils.h
```diff
@@ -50,7 +50,9 @@ std::unique_ptr<RuleBasedGraphTransformer> GenerateRuleBasedGraphTransformer(
     const bool enable_cast_chain_elimination = false);
 
 /** Generates all predefined (both rule-based and non-rule-based) transformers for this level.
-    Any transformers or rewrite rules named in rules_and_transformers_to_disable will be excluded. */
+    Any transformers or rewrite rules named in rules_and_transformers_to_disable will be excluded.
+    `intra_op_thread_pool` is optional and may be nullptr when per-session
+    thread pools are lazily initialized. */
```
yuslepukhin, line 55: With this change graph optimizers run single-threaded during Initialize().

During Initialize(), AddPredefinedTransformers and ApplyOrtFormatModelRuntimeOptimizations pass GetIntraOpThreadPoolToUse() which is nullptr at that point. This means constant folding and other optimizations that use the thread pool run single-threaded.

Previously, the thread pool existed during Initialize(), so optimizers could use it. This is a minor performance regression during model loading/optimization. This is already documented this in graph_transformer_utils.h.

This is acceptable for most workloads (model loading is typically one-time), but worth noting for large models with expensive constant folding.

However, some big models loading time exceeds minutes and for this reason we need to explore supplying the TP for optimization purposes.

onnxruntime/core/session/inference_session.cc
```diff
@@ -3025,13 +3119,21 @@ Status InferenceSession::Run(const RunOptions& run_options,
     }
   }
 
+  // Run and RunAsync intentionally share the same lazy thread-pool gate.
+  ORT_RETURN_IF_ERROR_SESSIONID_(EnsureThreadPoolsCreatedForRun());
```
yuslepukhin, line 3123: There is no longer a  config key to revert to eager thread pool creation. If lazy init causes issues in production, users have no escape hatch without code changes. Consider whether a config entry for "eager" mode should be retained.

onnxruntime/core/session/inference_session.cc
```diff
@@ -3025,13 +3119,21 @@ Status InferenceSession::Run(const RunOptions& run_options,
...
   // session threads spinning as configured. Do nothing for graph replay except the counter.
   const bool control_spinning = use_per_session_threads_ &&
                                 force_spinning_stop_between_runs_ &&
                                 !cached_execution_provider_for_graph_replay_.IsGraphCaptured(graph_annotation_id);
-  auto* intra_tp = (control_spinning) ? thread_pool_.get() : nullptr;
-  auto* inter_tp = (control_spinning) ? inter_op_thread_pool_.get() : nullptr;
+  concurrency::ThreadPool* intra_tp = nullptr;
+  concurrency::ThreadPool* inter_tp = nullptr;
+  if (control_spinning) {
+    std::lock_guard<std::mutex> lock(session_mutex_);
```
yuslepukhin, line 3133: > std::lock_guard<std::mutex> lock(session_mutex_); [](http://example.com/codeflow?start=3&length=50)

ORT has immutable design with every kernel having no mutable state. This and many other things are done for a single purpose: avoid locking and support parallel Run().
Even though this is a brief mutex acquisition, consider way to avoid an extra Run() serialziation.

After EnsureThreadPoolsCreatedForRun() returns with thread_pools_ready_for_run_ == true, the unique_ptr values are immutable. The mutex is technically unnecessary at that point — the acquire-release pair on thread_pools_ready_for_run_ already guarantees visibility. 

The unique_ptr::get() calls don't need mutex protection here. After EnsureThreadPoolsCreatedForRun() returns, the thread_pools_ready_for_run_ atomic has been set to true via a release-store. The acquire-load at inference_session.cc:651 in the fast path already provides a happens-before relationship that guarantees visibility of the unique_ptr values written under the mutex in the slow path. The unique_ptrs are never reassigned after publication, making the subsequent mutex-protected read at line 3133 redundant and harmful.

Additionally, session_mutex_ is the same mutex used by the sequential-run lock at inference_session.cc:3198

Even when is_concurrent_run_supported_ == true, the spinning-control lock now creates the same serialization effect.

I strongly suggest we move this phase back to Session Initialize(). I am not seeing a good reason not to perform everything that is necessary at the end of Initialize() rather than Run().

onnxruntime/test/framework/session_state_test.cc
```diff
@@ -708,6 +718,231 @@ static void PlaceAllNodesToCPUEP(Graph& graph) {
   }
 }
 
+#if !defined(ORT_MINIMAL_BUILD)
```
yuslepukhin, line 721: Missing test: thread_pool_size == 1 — verifying that single-threaded configurations still work correctly

onnxruntime/core/session/inference_session.cc
```diff
@@ -2211,6 +2294,17 @@ common::Status InferenceSession::Initialize() {
...
         prepacked_weights_container_);
 
+    if (use_per_session_threads_) {
+      // Some per-session configurations (for example externally supplied pools)
+      // have all required pools available before first Run/RunAsync.
+      // SessionState was constructed with these pointers, so publication is already done.
+      auto* intra_op_tp = GetIntraOpThreadPoolToUse();
+      auto* inter_op_tp = GetInterOpThreadPoolToUse();
+      if (AreRequiredPerSessionThreadPoolsAvailable(intra_op_tp, inter_op_tp)) {
+        thread_pools_ready_for_run_.store(true, std::memory_order_release);
```
yuslepukhin, line 2304: >std::memory_order_release [](http://example.com/codeflow?start=48&length=25)

Correct, but relaxed would be OK too here

onnxruntime/core/framework/session_state.cc
```diff
@@ -1203,7 +1232,8 @@ Status SessionState::CreateSubgraphSessionState() {
 
       auto subgraph_session_state =
           std::make_unique<SessionState>(*subgraph, execution_providers_,
-                                         thread_pool_, inter_op_thread_pool_, data_transfer_mgr_,
+                                         thread_pool_.load(std::memory_order_acquire),
+                                         inter_op_thread_pool_.load(std::memory_order_acquire), data_transfer_mgr_,
```
yuslepukhin, line 1236: Correct, but relaxed would be OK too. No concurrent Runs() exist yet.

onnxruntime/core/framework/session_state.cc
```diff
@@ -1138,6 +1138,35 @@ const NodeIndexInfo& SessionState::GetNodeIndexInfo() const {
   return *node_index_info_;
 }
 
+void SessionState::BindThreadPoolsOnce(concurrency::ThreadPool* thread_pool,
+                                       concurrency::ThreadPool* inter_op_thread_pool) {
+  const auto* current_thread_pool = thread_pool_.load(std::memory_order_acquire);
+  const auto* current_inter_op_thread_pool = inter_op_thread_pool_.load(std::memory_order_acquire);
```
yuslepukhin, line 1144: This is called from under mutex. Correct, but stronger than needed.

onnxruntime/core/framework/session_state.cc
```diff
@@ -1138,6 +1138,35 @@ const NodeIndexInfo& SessionState::GetNodeIndexInfo() const {
...
+
+  // This enforces one-shot semantics for both the root SessionState and all
+  // subgraph SessionStates reached by recursive propagation below.
+  // Current pointer states are allowed to stay at nullptr or to match exactly
+  // the incoming pointers; rebinding to different non-null pointers is rejected.
+  ORT_ENFORCE(
```
yuslepukhin, line 1150: > ORT_ENFORCE( [](http://example.com/codeflow?start=1&length=13)

CAS enforcement would be somewhat better if this is ever called w/o a mutex.

PR #27561, opened 2026-03-05
webgpu: support bool for Expand, Flatten, Gather and Unsqueeze

xhcao:


xhcao:
The PR makes all nodes of `detr-resnet-50` and `detr-resnet-50-fp16` running on WebGPU EP, and we could get ~8% performance improvement if using graph capture.
@jchen10

PR #27565, opened 2026-03-05
Add C# and Python language bindings for GetCompatibilityInfoFromModel / GetCompatibilityInfoFromModelBytes

adrastogi:
### Description

This change adds C# and Python language bindings and tests for the recently-introduced GetCompatibilityInfoFromModel / GetCompatibilityInfoFromModelBytes API. 

### Motivation and Context

#27015 introduced a new API to facilitate getting the model compatibility information from the metadata of a model (either a file or the model bytes). For convenience, we should ideally have some other language bindings included to make consumption a little easier.

### DIFF
--- a/csharp/src/Microsoft.ML.OnnxRuntime/OrtEnv.shared.cs
+++ b/csharp/src/Microsoft.ML.OnnxRuntime/OrtEnv.shared.cs
@@ -535,7 +535,7 @@ namespace Microsoft.ML.OnnxRuntime
         /// check if a precompiled model is compatible with the current system.
         /// </remarks>
         /// <param name="modelPath">Path to the ONNX model file.</param>
-        /// <param name="epType">The execution provider type string (e.g. "CPUExecutionProvider").</param>
+        /// <param name="epType">The execution provider type string. Use <see cref="OrtEpDevice.EpName"/> to get this value.</param>
         /// <returns>The compatibility info string, or null if no compatibility info exists for the specified EP.</returns>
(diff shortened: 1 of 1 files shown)


adrastogi:
@adrianlizarraga / @skottmckay, mind taking a look at this one whenever you can? Fills in some missing language bindings for that recently added API to get compat info from the model metadata. Thanks!

csharp/src/Microsoft.ML.OnnxRuntime/OrtEnv.shared.cs
```diff
@@ -524,6 +524,75 @@ public OrtCompiledModelCompatibility GetModelCompatibilityForEpDevices(
...
+        /// </summary>
+        /// <remarks>
+        /// Parses the model file to extract the compatibility info string for a specific execution provider
+        /// from the model's metadata properties. This is only applicable to models that have been precompiled
+        /// for an EP. Standard ONNX models do not contain this information.
+        /// The compatibility info can then be passed to <see cref="GetModelCompatibilityForEpDevices"/> to
+        /// check if a precompiled model is compatible with the current system.
+        /// </remarks>
+        /// <param name="modelPath">Path to the ONNX model file.</param>
+        /// <param name="epType">The execution provider type string (e.g. "CPUExecutionProvider").</param>
```
adrianlizarraga: nit: `CPUExecutionProvider` doesn't generate precompiled models so perhaps this could be a bit misleading. Similar to the documentation on the C APIs, perhaps could say to use the value from OrtEpDevice::EpName.

PR #27573, opened 2026-03-05
Add validation for negative position_ids in EmbedLayerNorm

vraspar:
### Description
Adds a missing lower-bound validation check for `position_col_index` (derived from `position_ids` input) in the `EmbedLayerNormalization` contrib operator. A negative value in `position_ids` could bypass the existing upper-bound-only check and cause an out-of-bounds heap read before the `position_embedding` buffer.


yuslepukhin (dismissed):

yuslepukhin:
Make sure to comment on the Copilot comments and resolve them

PR #27577, opened 2026-03-06
[NvTensorRTRTX EP] Implement GetHardwareDeviceIncompatibilityDetails with driver and compute capability checks

umangb-09:
### Description
Implements `GetHardwareDeviceIncompatibilityDetailsImpl` for the NvTensorRTRTX EP factory, wiring it into ORT's `GetHardwareDeviceIncompatibilityDetails` EP API.

### Motivation and Context
The GetHardwareDeviceEpIncompatibilityDetails API (introduced in #[26922](https://github.com/microsoft/onnxruntime/pull/26922/)) allows EPs to return structured diagnostic information when a device is not compatible, rather than silently failing.

This PR implements that API for the NvTensorRTRTX EP so that users get actionable error messages when their GPU architecture or NVIDIA driver version does not meet the EP's requirements, instead of an opaque failure at session creation time.

### DIFF
--- a/onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc
+++ b/onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc
@@ -10,9 +10,7 @@
 #include <cstdlib>
 #include <mutex>
 #include <unordered_map>
-#include <sstream>
 #include <string_view>
-#include <algorithm>
 #include <cuda.h>
 
 #include "core/providers/shared_library/provider_api.h"


umangb-09:
@chilo-ms @adrianlizarraga review this

umangb-09:
@tianleiwu help review and close this.

chilo-ms:
might need to merge the main to fix CI build issues.

cmake/onnxruntime_providers_nv.cmake
```diff
@@ -148,10 +149,15 @@ endif ()
...
   target_link_libraries(onnxruntime_providers_nv_tensorrt_rtx PRIVATE Eigen3::Eigen  onnx flatbuffers::flatbuffers Boost::mp11 safeint_interface Eigen3::Eigen)
   add_dependencies(onnxruntime_providers_nv_tensorrt_rtx onnxruntime_providers_shared ${onnxruntime_EXTERNAL_DEPENDENCIES})
+  # Make NVML linkage optional: only link CUDA::nvml if the target exists.
+  set(_nvml_lib "")
+  if (TARGET CUDA::nvml)
+    set(_nvml_lib CUDA::nvml)
+  endif()
   if (onnxruntime_USE_TENSORRT_BUILTIN_PARSER)
-    target_link_libraries(onnxruntime_providers_nv_tensorrt_rtx PRIVATE ${trt_link_libs} ${ONNXRUNTIME_PROVIDERS_SHARED} ${PROTOBUF_LIB} flatbuffers::flatbuffers Boost::mp11 safeint_interface ${ABSEIL_LIBS} PUBLIC CUDA::cudart CUDA::cuda_driver)
+    target_link_libraries(onnxruntime_providers_nv_tensorrt_rtx PRIVATE ${trt_link_libs} ${ONNXRUNTIME_PROVIDERS_SHARED} ${PROTOBUF_LIB} flatbuffers::flatbuffers Boost::mp11 safeint_interface ${ABSEIL_LIBS} PUBLIC CUDA::cudart CUDA::cuda_driver ${_nvml_lib})
```
tianleiwu: **[High]** `${_nvml_lib}` is placed after the `PUBLIC` keyword, making NVML a public (transitive) dependency of the EP shared library. NVML is an implementation detail for driver version checking and should be `PRIVATE`. Leaking it as PUBLIC means any downstream target linking against `onnxruntime_providers_nv_tensorrt_rtx` transitively links NVML.

Suggested fix — move `${_nvml_lib}` to the PRIVATE section (both here and the non-builtin parser path below):
```cmake
target_link_libraries(onnxruntime_providers_nv_tensorrt_rtx PRIVATE ${trt_link_libs} ... ${ABSEIL_LIBS} ${_nvml_lib} PUBLIC CUDA::cudart CUDA::cuda_driver)
```

onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc
```diff
@@ -1128,11 +1144,194 @@ struct NvTensorRtRtxEpFactory : OrtEpFactory {
...
+   * @brief Parses NVML driver version string and compares with minimum required version.
+   * 
+   * NVML always returns driver version in standard format: "major.minor" (e.g., "581.80", "555.85", "570.00")
+   * The minimum version is also provided in standard format.
+   * 
+   * @param driver_version_str NVML driver version string (e.g., "581.80")
+   * @param min_version_str Minimum required version string (e.g., "570.00")
+   * @return true if driver_version >= min_version, false otherwise
+   */
+  bool CompareNVMLDriverVersion(const std::string& driver_version_str, const std::string& min_version_str) {
```
tianleiwu: **[High] Linux three-part driver version format**: On Linux, `nvmlSystemGetDriverVersion()` returns a three-part version string like `"555.42.06"`, while on Windows it returns a two-part string like `"555.85"`. The hard-coded minimum versions (`"555.85"`, `"570.00"`) are Windows-format version numbers. On Linux, the R555 branch has minor version `42`, not `85`, so a valid Linux driver would be incorrectly flagged as too old.

The `parseVersion` lambda technically handles three-part strings (since `std::stoi("42.06")` returns `42`), but the minor version number is different between platforms for the same driver release. This means the comparison is semantically incorrect on Linux.

Suggested fix — use platform-specific minimums or compare only the major branch number:
```cpp
#if _WIN32
  min_driver_version = "555.85";
#else
  min_driver_version = "555.42";  // Linux R555 equivalent
#endif
```
Alternatively, compare only the major branch number (e.g., `driver_major >= 555`) which is consistent across platforms.

---

**[Suggestion]** This function does not access `this` or any member state. Consider making it `static` (or a free function) to clarify intent and avoid requiring a factory instance.

onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc
```diff
@@ -1128,11 +1144,194 @@ struct NvTensorRtRtxEpFactory : OrtEpFactory {
...
+   *
+   * @param this_ptr The OrtEpFactory instance.
+   * @param hw The hardware device to check for incompatibility.
+   * @param details Pre-allocated incompatibility details object initialized by ORT.
+   * @return nullptr on success (compatible or details set), OrtStatus on error.
+   */
+  static OrtStatus* ORT_API_CALL GetHardwareDeviceIncompatibilityDetailsImpl(
+      OrtEpFactory* this_ptr,
+      const OrtHardwareDevice* hw,
+      OrtDeviceEpIncompatibilityDetails* details) noexcept {
```
tianleiwu: **[High] `noexcept` with throwing operations**: This function is `noexcept` (required by `ORT_API_CALL` convention) but performs multiple `std::string` allocations (via constructor, `std::to_string`, `operator+`) that can throw `std::bad_alloc`. If memory allocation fails, the program terminates via `std::terminate()` rather than returning an error status.

While this matches the existing pattern in the file (no other callbacks use try/catch), this callback does significantly more string work than the simpler release/version callbacks.

Suggested fix:
```cpp
static OrtStatus* ORT_API_CALL GetHardwareDeviceIncompatibilityDetailsImpl(
    OrtEpFactory* this_ptr,
    const OrtHardwareDevice* hw,
    OrtDeviceEpIncompatibilityDetails* details) noexcept {
  try {
    // ... existing implementation ...
  } catch (const std::exception& e) {
    auto* factory = static_cast<NvTensorRtRtxEpFactory*>(this_ptr);
    return factory->ort_api.CreateStatus(ORT_RUNTIME_EXCEPTION,
        "[NvTensorRTRTX EP] Exception in GetHardwareDeviceIncompatibilityDetails");
  }
}
```

onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc
```diff
@@ -1128,11 +1144,194 @@ struct NvTensorRtRtxEpFactory : OrtEpFactory {
...
+          0,  // error_code
+          msg.c_str());
+    }
+
+#if defined(HAVE_NVML)
+    // Determine minimum driver version based on GPU architecture
+    // Note: NVML returns driver version in standard format (e.g., "581.80"), not "R570_00" format
+    // So we use standard format for minimum versions
+    std::string min_driver_version;
+    if (compute_capability_major >= 12) {
```
tianleiwu, line 1330: **[Suggestion] Compute capability gap for CC 9.x–11.x**: The architecture-to-driver mapping jumps from `>= 8` (requiring 555.85) to `>= 12` (requiring 570.00). Compute capabilities 9.x (Hopper), 10.x, and 11.x would all fall into the Ampere minimum driver bucket. If this is intentional, adding a brief comment explaining the mapping rationale would help future maintainability.

onnxruntime/core/providers/nv_tensorrt_rtx/nv_provider_factory.cc
```diff
@@ -1128,11 +1144,194 @@ struct NvTensorRtRtxEpFactory : OrtEpFactory {
...
+    } else if (compute_capability_major >= 8) {
+      // Ampere and Ada architectures (CC 8.x) - requires driver 555.85 or higher
+      min_driver_version = "555.85";
+    } else {
+      // Should not reach here (already checked compute capability above)
+      min_driver_version = "555.85";
+    }
+
+    // Initialize NVML and get driver version
+    nvmlReturn_t nvml_result = nvmlInit_v2();
```
tianleiwu, line 1350: **[Suggestion] NVML init/shutdown per invocation**: `nvmlInit_v2()` and `nvmlShutdown()` are called on every invocation. While NVML uses internal reference counting, repeated init/shutdown cycles add overhead and create a potential race if this callback is invoked concurrently from multiple threads (interleaved init/shutdown from different threads could cause the ref count to reach zero while another thread is still querying). Consider a once-init pattern (e.g., `std::call_once`) or caching the driver version string.

tianleiwu:
@umangb-09, I added some recommended changes from AI. Please review the changes If you do not like the changes, you can replace them with a force push.

tianleiwu:
Code review feedback threads have been reviewed. Unused headers (e.g. sstream, algorithm) were successfully removed from nv_provider_factory.cc. The previous concern regarding NVML init failures and conditional linkage is addressed in the current code. Resolving remaining concerns.

PR #27578, opened 2026-03-06
CUDA Resize: add optimized 3D nearest resize kernel for 5D up/down sa…

### LINKED ISSUE #14596: [Performance] Pytorch is faster than ONNX when running inference multiple times
### Describe the issue

I have a model that I'd like to run using onnxruntime to decrease the install size. Here's a table comparing the total time of ONNX and pytorch when running different amount of inference (time in seconds)

|   |1|5|10|50|100|
|---|---|---|---|---|---|
|pytorch|0.45|0.75|1.11|3.85|7.17|
|onnxruntime|0.26|0.79|1.51|6.98|13.87|

To me it seems like while the first inference of pytorch is slow, sequential once are much faster compared to onnx. I believe this has to do with cudnn plans not being cached. When I digged into the pytorch code I found that the plan for convultion layers is being cached. In the `run_single_conv` function in [Conv_v8.cpp](https://github.com/pytorch/pytorch/blob/523d4f2562580a6cd9888cfbc9b9ae8ed2a61ed1/aten/src/ATen/native/cudnn/Conv_v8.cpp), a function `benchmark_cache.find(key);` is executed which will search the cache to see if a convoltion plan with the same parameters has already been calculated. If not, it generates the plan and adds it to the cache. I was wondering if onnxruntime has something similar in place since I couldn't find this.

### To reproduce

I convert the model like this: 
```py
import torch

input_data = torch.randn((1, 1, 96, 96, 96)).cuda().half()
torch.onnx.export(
    model,  # model being run
    input_data,  # model input (or a tuple for multiple inputs)
    "model.onnx",  # where to save the model (can be a file or file-like object)
    export_params=True,  # store the trained parameter weights inside the model file
    opset_version=11,  # the ONNX version to export the model to
    do_constant_folding=True,  # whether to execute constant folding for optimization
    input_names=["input"],  # the model's input names
    output_names=["output"],  # the model's output names
    keep_initializers_as_inputs=False,
)
```
And run inference as followes:
```py
import onnxruntime
import numpy as np
import time

NUMBER: int = 1

opts = onnxruntime.SessionOptions()
# opts.enable_profiling = True
opts.graph_optimization_level = onnxruntime.GraphOptimizationLevel.ORT_ENABLE_EXTENDED
# opts.intra_op_num_threads = 1

ort_session = onnxruntime.InferenceSession(
    "outline.onnx",
    opts,
    providers=[
        (
            "CUDAExecutionProvider",
            {
                "device_id": 0,
                # "arena_extend_strategy": "kNextPowerOfTwo",
                # "do_copy_in_default_stream": False,
                "cudnn_conv_use_max_workspace": "1",
                # "gpu_mem_limit": 4 * 1024 * 1024 * 1024,
                # "cudnn_conv1d_pad_to_nc1d": "1",
                "cudnn_conv_algo_search": "HEURISTIC",
            },
        ),
        "CPUExecutionProvider",
    ],
)
def get_data():
    return np.array(np.random.rand(1, 1, 96, 96, 96) * 20, dtype=np.float16)

# compute ONNX Runtime output prediction
start = time.time()
io_binding = ort_session.io_binding()
for i in range(NUMBER):
    io_binding.bind_cpu_input("input", get_data())
    io_binding.bind_output("output")
    ort_session.run_with_iobinding(io_binding, runOptions)
    io_binding.copy_outputs_to_cpu()[0]
print(f"ONNX: {time.time() - start}")
```

Sadly I'm not able to share the model but I'm using a 3D Unet model similar to [this](https://github.com/wolny/pytorch-3dunet) model.

### Urgency

_No response_

### Platform

Linux

### OS Version

Ubuntu 18.04

### ONNX Runtime Installation

Released Package

1.13.1

### ONNX Runtime API

Python

### Architecture

X64

### Execution Provider

CUDA

_No response_

### Model File

_No response_

Yes

johannes-rehm-snkeos:
## Summary
This PR adds CUDA support for optimized **nearest-neighbor 3D resize mapping/execution** in the Resize operator path, and adds targeted regression coverage.

The implementation introduces a dedicated 3D fast path for nearest resize to handle the last three spatial dimensions (`D/H/W`) efficiently when outer dimensions are unchanged.

### CUDA Resize implementation
`onnxruntime/core/providers/cuda/tensor/resize_impl.cu`

- Added 3D nearest mapping kernel:
  - `_ResizeNearestMappingKernel3D`
- Added 3D nearest compute kernel:
  - `_ResizeNearestKernel3D`
- Added optimized 3D dispatch path in `ResizeNearestImpl`:
  - Enabled when:
    - `rank >= 3`
    - `coordinate_transformation_mode != tf_crop_and_resize`
    - all outer scales (except last 3 dims) are `1.0`

This keeps existing behavior unchanged for other cases while using the optimized path for true 3D nearest resize workloads.

### Regression tests
`onnxruntime/test/providers/cpu/tensor/resize_op_test.cc`

Added CUDA-targeted regression tests:
- `ResizeOpNearestUpSampleTest_5D_CudaRegression_Optimized3DMapping`
- `ResizeOpNearestDownSampleTest_5D_CudaRegression_Optimized3DMapping`

## Why
The previous nearest implementation relied on the generic path for these 3D scenarios. This change introduces a dedicated CUDA 3D path to improve performance for 5D nearest resize workloads.

Fixes #14596

### DIFF
--- a/onnxruntime/core/providers/cuda/tensor/resize_impl.cu
+++ b/onnxruntime/core/providers/cuda/tensor/resize_impl.cu
@@ -154,6 +154,70 @@ __global__ void _ResizeNearestMappingKernel2D(
 }
 
 template <typename T, typename CudaFunctionOriginalCoordinate, typename CudaFunctionNearestPixel>
+__global__ void _ResizeNearestMappingKernel3D(
+    const int input_depth, const int input_height, const int input_width,
+    const int output_depth, const int output_height, const int output_width,
+    const float scales_depth, const float scales_height, const float scales_width,
+    const float roi_start_depth, const float roi_end_depth,
+    const float roi_start_height, const float roi_end_height,
+    const float roi_start_width, const float roi_end_width,
+    const bool extrapolation_enabled,
+    const CudaFunctionOriginalCoordinate& transform_coordinate,
+    const CudaFunctionNearestPixel& calc_nearest_pixel,
+    NearestMappingInfo* dims_mapping) {
+  CALCULATE_ELEMENTWISE_INDEX_OR_EXIT(id, output_depth + output_height + output_width);
+  if (id < output_depth) {  // for Depth
+    int dim = id;
+    if (scales_depth == 1.0f) {
+      dims_mapping[id].extrapolate_ = 0;
+    } else {
+      float orig_coord = transform_coordinate(static_cast<float>(dim), scales_depth,
+                                              static_cast<float>(output_depth),
+                                              static_cast<float>(input_depth),
+                                              roi_start_depth, roi_end_depth);
+      dims_mapping[id].extrapolate_ = static_cast<int>(
+          extrapolation_enabled && (orig_coord < 0.f || orig_coord > static_cast<float>(input_depth - 1)));
+      dim = calc_nearest_pixel(orig_coord, scales_depth < 1);
+      if (dim >= input_depth) dim = input_depth - 1;
+      if (dim < 0) dim = 0;
+    }
+    dims_mapping[id].origin_ = dim;
+  } else if (id < output_depth + output_height) {  // for Height
+    int dim = id - output_depth;
+    if (scales_height == 1.0f) {
+      dims_mapping[id].extrapolate_ = 0;
+    } else {
+      float orig_coord = transform_coordinate(static_cast<float>(dim), scales_height,
+                                              static_cast<float>(output_height),
+                                              static_cast<float>(input_height),
+                                              roi_start_height, roi_end_height);
+      dims_mapping[id].extrapolate_ = static_cast<int>(
+          extrapolation_enabled && (orig_coord < 0.f || orig_coord > static_cast<float>(input_height - 1)));
+      dim = calc_nearest_pixel(orig_coord, scales_height < 1);
+      if (dim >= input_height) dim = input_height - 1;
+      if (dim < 0) dim = 0;
+    }
+    dims_mapping[id].origin_ = dim;
+  } else {  // for Width
+    int dim = id - output_depth - output_height;
+    if (scales_width == 1.0f) {
+      dims_mapping[id].extrapolate_ = 0;
+    } else {
+      float orig_coord = transform_coordinate(static_cast<float>(dim), scales_width,
+                                              static_cast<float>(output_width),
+                                              static_cast<float>(input_width),
+                                              roi_start_width, roi_end_width);
+      dims_mapping[id].extrapolate_ = static_cast<int>(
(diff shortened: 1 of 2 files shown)


johannes-rehm-snkeos:

tianleiwu:
@johannes-rehm-snkeos, could you share benchmark/profiling results that show that the new kernel is better?

johannes-rehm-snkeos:
@tianleiwu I used your script from here: https://github.com/microsoft/onnxruntime/issues/14596#issuecomment-1469508841 and got the following results:

Profiling of Torch:
```
           Type  Time(%)      Time     Calls       Avg       Min       Max  Name
 GPU activities:   54.73%  4.98039s       503  9.9014ms  5.6153ms  39.858ms  void at::native::_GLOBAL__N__b9911c4e_20_UpSampleNearest3d_cu_2b4cf812::upsample_nearest3d_out_frame<float, __operator_&__(at::native::nearest_neighbor_compute_source_index(float, int, int))>(float const *, __int64, __int64, __int64, __int64, __int64, __int64, __int64, __int64, at::native::nearest_neighbor_compute_source_index*, float, float, float)
```
Profiling of onnxruntime-gpu==1.24.3:
```
           Type  Time(%)      Time     Calls       Avg       Min       Max  Name
 GPU activities:   90.56%  40.4292s       503  80.376ms  76.639ms  98.755ms  void onnxruntime::cuda::_ResizeNearestKernel<float>(int, onnxruntime::cuda::TArray<__int64, int=8>, onnxruntime::cuda::_ResizeNearestKernel<float, onnxruntime::cuda::DivMod<int>, int=8>, float const *, onnxruntime::cuda::_ResizeNearestKernel<float, onnxruntime::cuda::DivMod<int>, int=8>*, __int64, onnxruntime::cuda::_ResizeNearestKernel<float, onnxruntime::cuda::DivMod<int>, int=8>, __int64 const *, onnxruntime::cuda::NearestMappingInfo const *)
```

Profiling of johannes-rehm-snkeos:cuda-resize-nearest-3d-kernel:
```
           Type  Time(%)      Time     Calls       Avg       Min       Max  Name
 GPU activities:   43.42%  3.75516s       503  7.4655ms  1.9272ms  37.947ms  void onnxruntime::cuda::_ResizeNearestKernel3D<float, bool=0>(__int64, __int64, __int64, __int64, __int64, int, onnxruntime::cuda::DivMod<int>, onnxruntime::cuda::DivMod, onnxruntime::cuda::DivMod, float const *, onnxruntime::cuda::DivMod<int>*, __int64, onnxruntime::cuda::DivMod<int>, onnxruntime::cuda::NearestMappingInfo const *)
```

PR #27582, opened 2026-03-06
Validate g_idx values in MatMulNBits to prevent OOB read

vraspar:
### Description
In `Dequantize4BitsKernelReOrder` (CPU and CUDA EP), values from the `g_idx` tensor are used directly as array indices into the `scales` and `zero_points` buffers without bounds checking. This PR adds value-range validation and tests for the `g_idx`  input tensor in the `MatMulNBits` operator.

### DIFF
--- a/onnxruntime/contrib_ops/cuda/quantization/dequantize_blockwise_4bits.cu
+++ b/onnxruntime/contrib_ops/cuda/quantization/dequantize_blockwise_4bits.cu
@@ -113,6 +113,7 @@ __global__ void Dequantize4BitsKernelReOrder(
   for (int i = 0; i < element_per_thread; i++) {
     int32_t rid = reorder_idx_with_off[i];
     CUDA_KERNEL_ASSERT(rid >= 0 && rid < groups_per_K);
+    rid = max(0, min(rid, groups_per_K - 1));  // Clamp for release safety
     T scale = *(scale_data + n_idx * scales_shape_x + rid);
     uint8_t zp = 8;  // Default zero point is 1 << (bits - 1)
     if (zero_points) {
--- a/onnxruntime/test/contrib_ops/matmul_4bits_test.cc
+++ b/onnxruntime/test/contrib_ops/matmul_4bits_test.cc
@@ -881,8 +881,7 @@ TEST(MatMulNBits, Basic_M10_N128_K512) {
 #endif
 
 // Test that out-of-range g_idx values are rejected with INVALID_ARGUMENT.
-// Skip in debug builds to avoid hitting CUDA_KERNEL_ASSERT which corrupts the CUDA device context.
-#ifdef NDEBUG
+// CUDA EP is excluded from these tests, so no risk of hitting CUDA_KERNEL_ASSERT.
 TEST(MatMulNBits, InvalidGIdx_OutOfRange) {
   constexpr int64_t M = 2, N = 4, K = 32, block_size = 16;
   constexpr int64_t k_blocks = (K + block_size - 1) / block_size;  // 2
@@ -969,7 +968,6 @@ TEST(MatMulNBits, InvalidGIdx_Negative) {
            {kCudaExecutionProvider, kCudaNHWCExecutionProvider, kDmlExecutionProvider, kWebGpuExecutionProvider,
             kOpenVINOExecutionProvider});
 }
-#endif  // NDEBUG
 
 }  // namespace test
 }  // namespace onnxruntime


onnxruntime/contrib_ops/cpu/quantization/matmul_nbits_helper.h
```diff
@@ -67,6 +67,19 @@ Status CheckInputs(const T* /*activation*/,
   // Group_index shall be 1D of K, or K padded to multiple of block_size
   ASSERT_TENSOR_SHAPE_2(group_index, make_shape(k), make_shape(k_blocks * block_size));
 
+  // Validate group_index values are within valid range [0, k_blocks)
+  if (group_index != nullptr) {
```
tianleiwu: You may want to check that the group_index is in CPU memory. If it is in GPU memory, it will failed to read it.
```
if (group_index ->Location().device.Type() == OrtDevice::CPU) {
 ...
}
```

onnxruntime/contrib_ops/cpu/quantization/matmul_nbits_helper.h
```diff
@@ -67,6 +67,19 @@ Status CheckInputs(const T* /*activation*/,
...
+    auto g_idx_data = group_index->template Data<int32_t>();
+    auto g_idx_size = group_index->Shape().Size();
+    for (int64_t i = 0; i < g_idx_size; ++i) {
```
yuslepukhin: int64_t should not be used as index vars as it fails on 32-bit systems. It should be converted to either `size_t` or `ptrdiff_t`.

vraspar: Thanks! converted to size_t

onnxruntime/test/contrib_ops/matmul_4bits_test.cc
```diff
@@ -880,6 +880,92 @@ TEST(MatMulNBits, Basic_M10_N128_K512) {
...
+  test.AddInput<int32_t>("g_idx", {K}, g_idx, true);
+
+  // bias: optional (skip)
+  test.AddOptionalInputEdge<float>();
+
+  // Output placeholder (won't actually be compared since we expect failure)
+  std::vector<float> y_data(M * N, 0.0f);
+  test.AddOutput<float>("Y", {M, N}, y_data);
+
+  test.Run(OpTester::ExpectResult::kExpectFailure, "group_index value",
```
hariharans29, line 926: For these tests on CUDA EP, once the kernel assert is triggered, will it pollute the entire EP Context without some sort of a reset ? Is it safe to use as is ?

tianleiwu: Need to skip these new tests in debug build to avoid hitting CUDA_KERNEL_ASSERT.

vraspar: Thanks! I made the fix

onnxruntime/contrib_ops/cpu/quantization/matmul_nbits_helper.h
```diff
@@ -67,6 +67,19 @@ Status CheckInputs(const T* /*activation*/,
...
+    auto g_idx_data = group_index->template Data<int32_t>();
+    auto g_idx_size = static_cast<size_t>(group_index->Shape().Size());
+    for (size_t i = 0; i < g_idx_size; ++i) {
+      if (g_idx_data[i] < 0 || g_idx_data[i] >= k_blocks) {
+        return ORT_MAKE_STATUS(ONNXRUNTIME, INVALID_ARGUMENT,
+                               "group_index value at index ", i, " is ", g_idx_data[i],
+                               ", which is out of valid range [0, ", k_blocks, ")");
+      }
+    }
+  }
```
tianleiwu, line 81: **CUDA EP validation gap**: When the CUDA EP calls `CheckInputs`, `group_index` resides on the GPU, so this CPU-side validation is silently skipped (the `OrtDevice::CPU` guard correctly prevents a host dereference of a device pointer). This means the only defense for the CUDA EP is the `CUDA_KERNEL_ASSERT` in the `.cu` file, which is **debug-only** — it compiles to a no-op in release builds.

In release/production builds with CUDA EP, invalid `g_idx` values will still cause unchecked OOB GPU memory accesses — the exact vulnerability this PR aims to fix.

Options to close the gap:
1. **GPU-side min/max reduction** before launching `Dequantize4BitsKernelReOrder` (e.g., `cub::DeviceReduce::Min`/`Max`, or a small dedicated bounds-check kernel). Return an error if out-of-range.
2. **Copy `g_idx` to CPU for validation** — adds a sync point but `g_idx` is typically small (K elements, constant per model).
3. **Release-safe clamping in the kernel** (see my other inline comment) — doesn't report the error but prevents memory corruption.

onnxruntime/contrib_ops/cuda/quantization/dequantize_blockwise_4bits.cu
```diff
@@ -112,6 +112,7 @@ __global__ void Dequantize4BitsKernelReOrder(
   const int32_t* reorder_idx_with_off = reorder_idx + kb_idx * block_size + ((threadIdx.x * element_per_thread) & (block_size - 1));
   for (int i = 0; i < element_per_thread; i++) {
     int32_t rid = reorder_idx_with_off[i];
+    CUDA_KERNEL_ASSERT(rid >= 0 && rid < groups_per_K);
```
tianleiwu, line 115: ⚠️ **Debug-only protection**: `CUDA_KERNEL_ASSERT` expands to `assert()` in debug builds and to **nothing** in release builds (also a no-op on Apple/HIP — see `common.cuh` L706-708). In release builds, invalid `rid` values will silently index OOB into `scale_data` and `zero_points`.

The codebase already has a pattern for this in `tensorscatter_impl.cu` — add a runtime clamp after the assert for release safety:

```cuda
int32_t rid = reorder_idx_with_off[i];
CUDA_KERNEL_ASSERT(rid >= 0 && rid < groups_per_K);
rid = max(0, min(rid, groups_per_K - 1));  // Clamp for release safety
```

This prevents OOB access in release while still flagging the bug in debug via the assert.

onnxruntime/test/contrib_ops/matmul_4bits_test.cc
```diff
@@ -880,6 +880,97 @@ TEST(MatMulNBits, Basic_M10_N128_K512) {
...
+           {kCudaExecutionProvider, kCudaNHWCExecutionProvider, kDmlExecutionProvider, kWebGpuExecutionProvider,
+            kOpenVINOExecutionProvider});
+}
+
```
tianleiwu, line 930: Nit: Since the CUDA EP is already in the exclusion list, the `#ifdef NDEBUG` guard at line 886 appears redundant for this specific concern. The comment says "skip in debug builds to avoid hitting CUDA_KERNEL_ASSERT which corrupts the CUDA device context" — but the CUDA EP is excluded regardless. Consider clarifying whether the guard protects against a different EP's debug assert, or removing it if it's solely for CUDA.

Also: these tests only validate the CPU EP path. Once the GPU-side gap is addressed with a proper runtime check, a corresponding CUDA EP test should be added.

tianleiwu (changes requested):
Thanks for addressing this OOB read vulnerability — the CPU-side validation logic is well-structured with a clear error message. However, the CUDA EP path still has a gap in release builds.

See inline comments for details.

onnxruntime/contrib_ops/cpu/quantization/matmul_nbits_helper.h
vraspar: Thanks for the detailed analysis. Added a release-safe clamp (`max(0, min(rid, groups_per_K - 1))`) in the CUDA kernel after the assert, following the existing pattern in `tensorscatter_impl.cu`. This prevents OOB memory access in release builds. A full GPU-side validation (min/max reduction or D2H copy) for proper error reporting could be done as a follow-up if needed.

onnxruntime/contrib_ops/cuda/quantization/dequantize_blockwise_4bits.cu
vraspar: Done — added the clamp as suggested. Thanks for pointing to the `tensorscatter_impl.cu` pattern.

onnxruntime/test/contrib_ops/matmul_4bits_test.cc
vraspar: Good catch — removed the `#ifdef NDEBUG` guard since CUDA EP is already excluded in the `Run()` call. The tests now compile and run in both debug and release builds. Will add a CUDA EP-specific test once we have a proper GPU-side validation path that returns an error.

PR #27596, opened 2026-03-09
[WebGPU] Allow reusing an external GPUDevice via env.webgpu.device

fdb:
## Summary

When a web application already owns a `GPUDevice` (e.g. for rendering), ONNX Runtime's WebGPU backend currently creates a **second** device internally. Because two `GPUDevice` instances cannot share GPU textures, every inference pass that feeds render output into a model requires a full GPU → CPU → GPU readback — a synchronous stall that causes severe frame-time jank.

This PR wires up the **already-documented** `env.webgpu.device` property so the backend actually honors it:

- **`wasm-core-impl.ts`** — reads `env.webgpu.device` at session-creation time, validates that `env.webgpu.adapter` is also provided, and forwards both to the JSEP init path.
- **`jsep/init.ts`** — passes the optional `gpuDevice` parameter through to the backend.
- **`jsep/backend-webgpu.ts`** — when an external device is supplied, validates it against ORT's required limits/features, skips `adapter.requestDevice()`, chains onto the existing `onuncapturederror` handler instead of overwriting it, and uses a safe `setFrozenProperty` helper that tolerates re-initialization with the same object.
- **`env.ts`** — expands the JSDoc for the `device` property to describe shared-device mode and the requirement to also set `adapter`.

### Usage

```js
const adapter = await navigator.gpu.requestAdapter();
const device = await adapter.requestDevice({ /* app limits */ });

ort.env.webgpu.adapter = adapter;
ort.env.webgpu.device = device;

const session = await ort.InferenceSession.create(modelUrl, {
  executionProviders: ['webgpu'],
});
// ORT now uses the same device — GPU textures can be shared zero-copy
```

### Validation

Tested in a real-time creative-coding app ([Figment](https://github.com/fdb/figment)) that runs ONNX image models (depth estimation, style transfer) on every frame alongside WebGPU rendering. With this patch:

- Max frame-callback time dropped from **~180 ms** (readback stall) to **~1.5 ms**
- Mean frame time reduced from ~12 ms to ~8 ms
- Chrome DevTools traces confirm zero `GPUBuffer.mapAsync` stalls in the render path

The `device` property already exists in the TypeScript type definitions and JSDoc but was previously ignored by the backend initialization code.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


fdb:

js/common/lib/env.ts
```diff
@@ -228,7 +228,10 @@ export declare namespace Env {
      *
      * There are 3 valid scenarios of accessing this property:
      * - Set a value before the first WebGPU inference session is created. The value will be used by the WebGPU backend
-     * to perform calculations. If the value is not a `GPUDevice` object, an error will be thrown.
+     * to perform calculations. If the value is not a `GPUDevice` object, an error will be thrown. When providing your
+     * own device, you must also set {@link adapter} to the `GPUAdapter` that created it. This enables shared-device
+     * mode, where ORT reuses the external device instead of creating its own — allowing zero-copy GPU tensor sharing
+     * between ORT and the host application.
```
fs-eire, line 234:
```
When providing your own device, you must also set {@link adapter} to the `GPUAdapter` that created it.
```

I am thinking about whether we need to enforce it or not.

PR #27599, opened 2026-03-09
Add KernelInfo string-array attribute APIs to the ORT C and C++ APIs

tianleiwu:
## Summary

This PR adds a new `OrtApi` entry point for reading repeated string attributes from `OrtKernelInfo`:

- `KernelInfoGetAttributeArray_string`

It also wires that support through the C++ wrapper so callers can use:

- `Ort::ConstKernelInfo::GetAttributes<std::string>(...)`

## Problem

The existing kernel info APIs already support scalar and array attribute access for numeric types, but there was no C API for reading string-array attributes from `OrtKernelInfo`.

That created a gap for code paths that rely on repeated string attributes in kernel metadata, such as:

- custom op / kernel consumers using the public C API
- C++ wrapper callers expecting `GetAttributes<std::string>` to work end-to-end
- plugin EP scenarios that need to compile existing kernels against the adapter/C API surface

One concrete case is CUDA plugin EP RNN support, where the RNN kernels read the `activations` attribute via `GetAttrs<std::string>("activations", ...)`. The adapter path needed a corresponding ORT C API to expose that data.

### C API

Added `OrtApi::KernelInfoGetAttributeArray_string` to fetch repeated string attributes from `OrtKernelInfo`.

Behavior:

- If `out == nullptr`, the API returns the attribute count in `size`.
- Otherwise, the API allocates the pointer array and each UTF-8 string with the provided `OrtAllocator`.
- For empty attributes, `*out` is set to `nullptr` and `*size` is set to `0`.
- The caller frees each string and the pointer array with the same allocator.

### Implementation

Added the implementation in the ORT session/custom-op API layer by:

- reading the underlying attribute with `OpKernelInfo::GetAttrs<std::string>`
- copying the result into allocator-owned C-style string storage for the public API

### C++ wrapper

Completed C++ wrapper support so `Ort::ConstKernelInfo::GetAttributes<std::string>(name)` works through the new C API.

The wrapper follows the standard two-call pattern:

1. query the number of strings
2. allocate and fetch the returned string array
3. copy into `std::vector<std::string>` and release allocator-owned memory

### Tests

Added framework tests covering:

- non-empty string-array attributes
- empty string-array attributes
- missing attribute failure path
- C++ wrapper access through `Ort::ConstKernelInfo`

## Files Changed

- `include/onnxruntime/core/session/onnxruntime_c_api.h`
- `include/onnxruntime/core/session/onnxruntime_cxx_api.h`
- `include/onnxruntime/core/session/onnxruntime_cxx_inline.h`
- `onnxruntime/core/session/custom_ops.cc`
- `onnxruntime/core/session/onnxruntime_c_api.cc`
- `onnxruntime/core/session/ort_apis.h`
- `onnxruntime/test/framework/kernel_info_test.cc`

## Why This Change

This closes a real API gap in kernel attribute access and makes the public API surface more consistent with the existing numeric attribute helpers.

It also unblocks plugin/adapter-based kernel code that depends on repeated string attributes without requiring those kernels to special-case plugin builds.

For example, porting rnn operator to cuda plugin EP will need this API.

## Validation

Validated with new unit coverage in `kernel_info_test.cc` for:

- `KernelInfoGetAttributeArray_string` with populated attributes
- `KernelInfoGetAttributeArray_string` with empty attributes
- missing-attribute error handling
- `Ort::ConstKernelInfo::GetAttributes<std::string>` parity with the C API

### DIFF
--- a/include/onnxruntime/core/session/onnxruntime_c_api.h
+++ b/include/onnxruntime/core/session/onnxruntime_c_api.h
@@ -7243,11 +7243,6 @@ struct OrtApi {
    */
   ORT_API2_STATUS(RunOptionsDisableProfiling, _Inout_ OrtRunOptions* options);
 
-  /// @}
-  /// \name OrtKernelInfo
-  /// Custom operator APIs.
-  /// @{
-
   /** \brief Fetch an array of strings stored as an attribute in the graph node
    *
    * If `out` is nullptr, the value of `size` is set to the true size of the attribute
--- a/onnxruntime/core/session/custom_ops.cc
+++ b/onnxruntime/core/session/custom_ops.cc
@@ -547,6 +547,20 @@ static Status CopyDataFromVectorToMemory(const std::vector<T>& values, T* out, s
   return Status::OK();
 }
 
+static char* DuplicateStringToAllocatorMemory(const std::string& value, OrtAllocator* allocator) {
+  SafeInt<size_t> allocation_size(value.size());
+  allocation_size += 1;
+
+  char* duplicated_value = static_cast<char*>(allocator->Alloc(allocator, allocation_size));
+  if (duplicated_value == nullptr) {
+    return nullptr;
+  }
+
+  std::memcpy(duplicated_value, value.data(), value.size());
+  duplicated_value[value.size()] = '\0';
+  return duplicated_value;
+}
+
 static Status CopyStringDataFromVectorToMemory(const std::vector<std::string>& values, OrtAllocator* allocator, char*** out, size_t* size) {
   *size = values.size();
 
@@ -555,9 +569,9 @@ static Status CopyStringDataFromVectorToMemory(const std::vector<std::string>& v
   }
 
   ORT_RETURN_IF_NOT(allocator != nullptr, "allocator must not be null when out is provided");
+  *out = nullptr;
 
   if (values.empty()) {
-    *out = nullptr;
     return Status::OK();
   }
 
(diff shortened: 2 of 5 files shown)


include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -7305,6 +7242,28 @@ struct OrtApi {
...
+   * If `out` is nullptr, the value of `size` is set to the true size of the attribute
+   * array and a success status is returned.
+   *
+   * Otherwise, the strings and pointer array are allocated using `allocator`.
+   * The caller must free each string and the pointer array with `allocator`.
+   * If the attribute array is empty, `*out` is set to nullptr and `*size` is set to 0.
+   *
+   * \param[in] info instance
+   * \param[in] name name of the attribute to be parsed
+   * \param[in] allocator allocator used to allocate the returned string array and strings
```
adrianlizarraga, line 7257: nit: perhaps mention that allocator can be null if only querying the size

onnxruntime/core/session/custom_ops.cc
```diff
@@ -545,6 +547,65 @@ static Status CopyDataFromVectorToMemory(const std::vector<T>& values, T* out, s
...
+    char* duplicated_value = DuplicateStringToAllocatorMemory(values[i], allocator);
+    if (duplicated_value == nullptr) {
+      for (size_t j = 0; j < allocated_string_count; ++j) {
+        if (array[j] != nullptr) {
+          allocator->Free(allocator, array[j]);
+        }
+      }
+
+      return Status(onnxruntime::common::ONNXRUNTIME, onnxruntime::common::FAIL,
+                    "Failed to allocate string attribute array");
```
adrianlizarraga, line 597:
```suggestion
                    "Failed to allocate string element in string attribute array");
```

include/onnxruntime/core/session/onnxruntime_c_api.h
tianleiwu: Good suggestion. I will update it later

onnxruntime/core/session/custom_ops.cc
tianleiwu: Good suggestion. I will update it later

PR #27603, opened 2026-03-09
[CUDA] Add Validation of batch_indices in RoiAlign

tianleiwu:
## Description
This PR implements a device-side bounds check for `batch_indices` in the RoiAlign CUDA operator. This is a follow-up to https://github.com/microsoft/onnxruntime/pull/27543, which fixed the same vulnerability in the CPU implementation.

Previously, CheckROIAlignValidInput() only validated `batch_indices` when they were accessible on the host (CPU). For the CUDA EP, `batch_indices` reside in GPU memory, so host-side validation would require an expensive GPU-to-CPU copy, which could also break CUDA graph capture.

This change:
1.  Passes `batch_size` from the host to the CUDA kernel.
2.  Adds a check within the `RoIAlignForward` kernel to ensure `0 <= batch_index < batch_size`.
3.  If an invalid `batch_index` is encountered, the kernel sets the output value for that specific RoI element to 0 and returns early for that thread.

## Impact
-   **Vulnerability fixed:** Heap out-of-bounds read on GPU.
-   **Performance:** Negligible impact as it's a simple range check within the existing kernel.
-   **Compatibility:** No changes to ONNX models or public APIs.

## Validation
-   Existing `RoiAlignTest` suite.
-   Added two new test cases: `BatchIndicesOutOfRange_CUDA` and `BatchIndicesNegative_CUDA` to verify that the CUDA provider correctly handles out-of-range batch indices.
-   Verified that the CUDA provider handles opset 10 without falling back to the CPU EP for these tests.


onnxruntime/core/providers/cuda/object_detection/roialign_impl.cu
```diff
@@ -106,6 +107,12 @@ __global__ void RoIAlignForward(
     // RoI could have 4 or 5 columns
     const T* offset_bottom_rois = bottom_rois + n * roi_cols;
     const auto roi_batch_ind = batch_indices_ptr[n];
+    // Validate batch_indices values are within [0, batch_size).
+    // If the index is out of range, we set the output to 0 for this RoI element.
+    if (roi_batch_ind < 0 || roi_batch_ind >= batch_size) {
+      top_data[index] = 0;
```
hariharans29, line 114: Could we add a cuda kernel assert here to asynchronously report issues in debug builds ?

tianleiwu: yes

onnxruntime/test/providers/cpu/object_detection/roialign_test.cc
```diff
@@ -854,5 +854,49 @@ TEST(RoiAlignTest, BatchIndicesNegative) {
...
+  test.AddAttribute<float>("spatial_scale", 1.0f);
+
+  test.AddInput<float>("X", {1, 1, 4, 4}, {1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16});
+  test.AddInput<float>("rois", {1, 4}, {0, 0, 3, 3});
+  test.AddInput<int64_t>("batch_indices", {1}, {1});  // batch_size is 1, so 1 is out of range
+  test.AddOutput<float>("Y", {1, 1, 2, 2}, {0.f, 0.f, 0.f, 0.f});
+
+  std::vector<std::unique_ptr<IExecutionProvider>> execution_providers;
+  execution_providers.push_back(std::move(cuda_ep));
+  test.Run(OpTester::ExpectResult::kExpectSuccess, "", {}, nullptr, &execution_providers);
```
hariharans29, line 880: Shouldn't these tests actually "fail" on debug builds due to the reported asynchronous error which will be reported during stream sync at the end of the run ?

yuslepukhin: I think that check is commented out in Compute() override?

tianleiwu: It is right that test will fail in debug build, so I skip the test in debug build.

A clean in-process recovery for cuda assert is not realistic. Once a CUDA kernel hits assert(), the cuda context is effectively poisoned until reset, and resetting it inside the provider test process would invalidate the EP state.

PR #27608, opened 2026-03-10
[Plugin EP] Make compiled kernel names "more" unique

apwojcik:
### Description
Plugin EPs have no control over the compiled kernel names, unlike Host API EPs, which use ComputeCapability and MetaDef directly. With this change, I want to add more uniqueness to the compiled kernel names.

### Motivation and Context
Currently, the name is chosen based on the graph view, which represents the model graph. For subgraphs of the same graph, the names are identical, and the Kernel Registry refuses to register the subgraph kernel, citing that the same-name kernels must have different op sets.


onnxruntime/core/session/plugin_ep/ep_plugin_provider_interfaces.cc
```diff
@@ -101,10 +101,10 @@ struct PluginEpMetaDefNameFunctor {
                                       const std::string& prefix)
       : generator_(generator), graph_viewer_(graph_viewer), prefix_(prefix) {}
 
-  std::string operator()() {
+  std::string operator()() const {
     uint64_t model_hash = 0;
     int id = generator_.GenerateId(graph_viewer_, model_hash);
-    return MakeString(prefix_, "_", model_hash, "_", id);
+    return MakeString(prefix_, "_", graph_viewer_.Name(), "_", model_hash, "_", id);
```
adrianlizarraga, line 107: Alternatively, could this be fixed by making the `ModelMetadefIdGenerator` local variable that is currently declared in [PluginExecutionProvider::GetCapability()](https://github.com/microsoft/onnxruntime/blob/ed3f8bf1fce77dd2458b7b37a26afbc086c0d8fc/onnxruntime/core/session/plugin_ep/ep_plugin_provider_interfaces.cc#L230) into a member of `PluginExecutionProvider`? I suspect that this would properly make every call to `generator_.GenerateId()` return a monotonically increasing number. It would also prevent redundant regenerations of the `model_hash` as a nice bonus.

I can try to experiment on my side.

Aside: for a future release of ORT, perhaps we can allow EPs to provide a custom metadef name via [OrtNodeFusionOptions](https://github.com/microsoft/onnxruntime/blob/ed3f8bf1fce77dd2458b7b37a26afbc086c0d8fc/include/onnxruntime/core/session/onnxruntime_ep_c_api.h#L439).

adrianlizarraga:
Hi @apwojcik,
Thank you very much for bringing this bug to our attention. I created an alternative PR with the fix described in this comment: https://github.com/microsoft/onnxruntime/pull/27608#discussion_r2933971715

The PR also adds testing. Could you please verify on your end.

@tianleiwu, I'm removing the 1.24.4 release tag from this PR.

onnxruntime/core/session/plugin_ep/ep_plugin_provider_interfaces.cc
adrianlizarraga: Also, I'm worried about two things with this approach:
 - Two different subgraphs (e.g., branches of an If node) may have the same name. This approach would still lead to duplicate metadef names.
 - External tools may be depending on the current format of this metadef name.

apwojcik:
Closing this PR in favour of https://github.com/microsoft/onnxruntime/pull/27666

PR #27609, opened 2026-03-10
[MLAS][KleidiAI]Catlaw01/sgemm epilogue neon opt

Laan33:
### Description

This change updates the KleidiAI SGEMM post-processing path in onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp with two parts:
  - Correctness fix: in the alpha == 0 || K == 0 fast path, beta handling is now applied for every batch entry (not just batch 0), so batched SGEMM behaviour is correct.
  - NEON SGEMM epilogue optimisation: adds a vectorised alpha/beta post-processing path for contiguous outputs, with guarded fallback to scalar for non-contiguous or small cases. The 2D epilogue path also routes contiguous tiles through the contiguous 1D epilogue path to enable vectorisation.

  ### Motivation and Context

This change addresses correctness and performance in the SGEMM post-processing stage:
  - The batched alpha == 0 || K == 0 path previously used only Data[0], which could produce incorrect results for BatchSize > 1.
  - The post-processing loop (C = alpha * (A*B) + beta * C) is a known latency contributor when memcpy fast paths are not applicable. The NEON epilogue changes are intended to reduce this cost on supported ARM platforms while preserving existing fallback behaviour.

### DIFF
--- a/onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
+++ b/onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
@@ -8,9 +8,7 @@
 #include <algorithm>
 #include <cstring>
 #include <cstddef>
-#if defined(__ARM_NEON)
 #include <arm_neon.h>
-#endif
 
 #include "mlas.h"
 
@@ -62,7 +60,6 @@ static inline void ApplyAlphaBetaStrided(const float* src, size_t num_elements, 
         return;
     }
 
-#if defined(__ARM_NEON)
     // Contiguous-only vectorized path with strict correctness guards.
     if (dst_stride == 1 && num_elements >= kAlphaBetaNeonMinElements) {
         size_t i = 0;
@@ -108,7 +105,6 @@ static inline void ApplyAlphaBetaStrided(const float* src, size_t num_elements, 
         }
         return;
     }
-#endif
 
     for (size_t i = 0; i < num_elements; ++i) {
         const float ab = src[i];


Laan33:

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
```diff
@@ -413,20 +491,17 @@ Return Value:
...
-            }
-        } else if (Data->beta != 1.0f) {
-            for (size_t i = 0; i < M; ++i) {
-                for (size_t j = 0; j < N; ++j) {
-                    Data->C[i * Data->ldc + j] *= Data->beta;
-                }
+        if (BatchSize == 1) {
+            ApplyBetaToC(Data->C, Data->ldc, M, N, Data->beta);
+        } else {
+            // This reduction must be applied to every batch entry.
```
hariharans29: As discussed, please consider adding a test to over this change

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
```diff
@@ -5,8 +5,12 @@
 //
 
 #include <vector>
+#include <algorithm>
 #include <cstring>
 #include <cstddef>
+#if defined(__ARM_NEON)
```
hariharans29: Would MLAS_TARGET_ARM64 be a more inclusive guard (MSVC) ?

Also since these files get compiled only for ARM64 anyway and I think building MLAS requires NEON support to begin with - is this guard even necessary - can we just assume NEON is available on the building ARM platform ?

hariharans29:
Please includ ethis - https://github.com/microsoft/onnxruntime/pull/27618/ when it goes through eventually

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
Laan33: Fair point, changed to use `MLAS_TARGET_ARM64`, thanks!

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
Laan33: Added [here](https://github.com/microsoft/onnxruntime/pull/27609/changes#diff-8a415f628c6fc926dcdecfab5d5ce38abe4ab62a2f4aca059e44bc3c1928c11dR87), thanks!

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
```diff
@@ -90,14 +146,36 @@ Routine Description:
...
-    if (alpha == 1.0f && beta == 0.0f && ldc == cols && rows != 0 && cols != 0) {
-        std::memcpy(dst, src, rows * cols * sizeof(float));
+    if (rows == 0 || cols == 0) {
+        return;
+    }
+
+    if (ldc == cols) {
+        // Contiguous destination: flatten so we can hit the contiguous SIMD path.
+        ApplyAlphaBetaStrided(src, rows * cols, alpha, beta, dst, 1, /*allow_memcpy*/ true);
         return;
```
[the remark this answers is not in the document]
Laan33: Fixed, updated the ApplyAlphaBeta2D comment to match current behaviour

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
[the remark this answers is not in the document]
Laan33: Fixed, I changed the early-exit logic so it no longer keys off Data->alpha (batch 0 only). It now:
  - handles K == 0 as a per-batch beta-only path, and
  - only takes the alpha-zero fast path when all batches have alpha == 0.0f.

I also added short-test coverage for BatchSize > 1, including a batched K == 0 case, to guard this path.

onnxruntime/core/mlas/lib/kleidiai/sgemm_kleidiai.cpp
hariharans29: It seems like, there are other places in the file that still use __ARM_NEON.

As suggested above, please consider removing it altogether if this file only compiles for ARM platforms anyway and we require NEON capability to build MLAS for ARM to begin with (making the guard itself redundant)

Laan33: Removed both guards of MLAS_TARGET_ARM64 and __ARM_NEON there, thanks

hariharans29:
Can you please fix the CI issues ?

Laan33:
Pushed new commits there to fix issues caused by when I merged main into the branch

hariharans29 (approved):

Issue #27167, opened 2026-01-27
[Feature Request] Support querying the device type to which the subgraph is assigned


mingmingtasd:
### Describe the feature request

The ORT PR [Add API to get ep graph partitioning info](https://github.com/microsoft/onnxruntime/pull/26781) already supports querying the ep name to which the subgraph is assigned, is it possible to also support querying the device type?

### Describe scenario use case

The requirement is from https://github.com/webmachinelearning/webnn/issues/836 by MLGraph.devices.

adrastogi:
@mingmingtasd, will also use this to track the request to surface when fallback has happened (via the example you shared here: https://github.com/microsoft/onnxruntime/blob/a3749f13536bd25e89ba8f8a5ae21e88b9337057/onnxruntime/core/providers/openvino/backends/basic_backend.cc#L123)

mingmingtasd:
@adrastogi Thanks for your support!

PR #27610, opened 2026-03-10
Add support to fetch device type from EP subgraph assignment

adrastogi:
### Description

#26781 added support for retrieving subgraph metadata for the assigned EPs in the session. There was a request via #27167 to add device type support, so this change attempts to implement that suggestion.

### Motivation and Context

See #27167 for details.


include/onnxruntime/core/session/onnxruntime_c_api.h
```diff
@@ -7305,6 +7305,25 @@ struct OrtApi {
    * \since Version 1.25.
    */
   ORT_API2_STATUS(RunOptionsDisableProfiling, _Inout_ OrtRunOptions* options);
+
+  /** \brief Get the device type of the execution provider to which the subgraph was assigned.
+   *
+   * Returns the default device type of the execution provider that claimed this subgraph.
+   * This is the device type the EP registered with (e.g., CPU, GPU, or NPU).
+   *
+   * \note For execution providers that internally manage multiple device types (e.g., OpenVINO in
```
adrastogi: This is the part of the design I am somewhat stuck on- for multi-device EPs (like OpenVINO), I don't _think_ it is sufficient to pick the default registered type on the EP. But then I couldn't figure out where the mapping happens between the subgraph and the device.

1. ComputeCapability - has an IndexedSubGraph (node indices + optional MetaDef for fusion), but no device field.
2. IndexedSubGraph - has nodes, MetaDef, and resource accounting, but no device field.
3. OrtEpGraphSupportInfo (plugin EP API) - the newer plugin API lets EPs declare supported node groupings (single nodes or fused groups). Each NodeGrouping has a kind + nodes + fusion options. No device field.
4. IExecutionProvider::GetCapability() - Returns vector<ComputeCapability>. The EP says "I can handle these subgraphs" but doesn't say "this subgraph goes to GPU, that one to NPU".

onnxruntime/core/framework/graph_partitioner.cc
```diff
@@ -524,7 +524,7 @@ static Status PartitionOnnxFormatModelImpl(Graph& graph, FuncManager& func_mgr,
       if (on_partition_assignment_fn) {
         // Call custom function provided by owner of GraphPartitioner whenever a subgraph is assigned to an EP.
         // This can be used, for example, to collect partitioning information.
-        on_partition_assignment_fn(graph, *capability, type);
+        on_partition_assignment_fn(graph, *capability, type, current_ep.GetDevice());
```
adrianlizarraga: Unfortunately, I don't think using `OrtDevice` will work as we want. For example, QNN and the other npu-based EPs use the default "cpu" `OrtDevice` to ensure that their inputs are placed on CPU.

I think we may be able to do the following:
- Add a new virtual function `IExecutionProvider::GetEpDevices()` that returns an empty result by default.
- Update the [PluginExecutionProvider](https://github.com/microsoft/onnxruntime/blob/50f3cc3c667985a0947fd870e1c6a56feeb664ec/onnxruntime/core/session/plugin_ep/ep_plugin_provider_interfaces.h#L83) class to define a `PluginExecutionProvider::GetEpDevices()` implementation that returns all of its `OrtEpDevices` (which are already stored in a member field).
- We pass the `OrtEpDevices` to `on_partition_assignment_fn`.
- Add one (or more!) `OrtHardwareDevices` to `OrtEpAssignedSubgraph`.

Hopefully the above is somewhat right :)

include/onnxruntime/core/session/onnxruntime_c_api.h
adrianlizarraga: As you alluded, we could add a `OrtHardwareDevice` field to the `OrtNodeFusionOptions`. A plugin EP can optionally set that to indicate the exact hw device that executes the subgraph (for observability purposes). I don't think we can force EPs to set the device since that would break existing plugin EPs.

If the plugin EP does not set the `OrtHardwareDevice` on the `OrtNodeFusionOptions`, then maybe we fallback to my idea in this comment: https://github.com/microsoft/onnxruntime/pull/27610#discussion_r2934908780 . Perhaps we just take the first `ep_device[0].hardware_device`, which should be correct 99% of the time.

skottmckay:
I'd be reticent to add `fused_node_hardware_device` at this point as it's a value that potentially adds confusion when it's not the EPs default device. User has no way to directly use that device and it won't match the device in the OrtEpDevice default allocator.

This possibly requires a much broader design to best support EPs which have multiple devices internally. The current setup of an EP having a single default device type worked fine years ago where it was CPU or discrete GPU but limits what's possible. Unfortunately handling this well will require changes into many places.

One potential option would be to allow the EP to explicitly specify the device for inputs/outputs in the ComputeCapability during partitioning. We _could_ set the OrtDevice in the NodeInfo values at that point (if not specified use current logic to set) instead of later downstream as we do currently. That's not a clean change though as for some usages you want OrtMemoryInfo and not just OrtDevice. 

At least these and probably other areas need updates
- Should all devices be explicitly provided during EP registration for clarity on what's possible
- EP must provide CopyTensors implementations to handle any devices it uses.
  - this enables ORT users to be able copy/read data to/from the correct input/output for the inference setup if they want to do that directly
- Logic in the memcpy transformer needs to be able to handle these additional devices.
  - that's potentially quite a complicated update
- allocation planning and any code that uses the EP's default device needs review and potential update
- need to consider things like SessionGetMemoryInfoForInputs/SessionGetMemoryInfoForOutputs.
  - I think it would be much better for that to be able to return the 'real' device, but that could be considered to be a breaking change. 
  - e.g. currently for OV or QNN it's always going to say CPU. those EPs provide a HOST_ACCESSIBLE allocator, and memory _can_ be provided using that, however you don't actually know whether that's required/beneficial as the input may be used on CPU by the EP internally. if the EP can directly say where it will consume the data during partitioning there's no confusion.

PR #27623, opened 2026-03-11
[CI] feat: use ccache & vcpkg cache for linux workflows

sanaa-hamel-microsoft:
### Description
Enable ccache and vcpkg caching for Linux workflows that use `reusable_linux_build.yml`. Saves about ~15-20 min on a 100% cache hit.
Also parallelises tests. Saves ~6 minutes.

Additionally, enable vcpkg and ccache for other Linux workflows. No numbers avail for comparison.

### Motivation and Context

This change reduces wasted CO2 and time.

### Known Issues

Benign  - Android workflow doesn't seem to be populating its ccache.

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -29,6 +29,8 @@ jobs:
         "1ES.Pool=onnxruntime-github-Ubuntu2204-AMD-CPU",
         "JobId=AndroidBinarySizeCheckJob_MinimalBaseline-${{ github.run_id }}-${{ github.run_number }}-${{ github.run_attempt }}"
         ]
+    env:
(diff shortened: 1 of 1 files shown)


.github/workflows/android.yml
```diff
@@ -144,6 +150,23 @@ jobs:
         with:
           ndk-version: 28.0.13004108
 
+      - name: Setup CCache
```
eserscor, line 153: Can setup ccache be made something similar to a template in ADO?

sanaa-hamel-microsoft: Yes. I'll leave this to a future PR. Windows/mac will have similar issues.

PR #27625, opened 2026-03-11
Use SafeInt in implementation of Col2Im

fs-eire:
### Description

Fix integer overflow in Col2Im shape calculation by using SafeInt for attacker-controlled dimension arithmetic


tianleiwu (approved):
LGTM

PR #27626, opened 2026-03-11
Add PE version info to onnxruntime_providers_vitisai.dll

xhan65:
### Description
Add a Windows VERSIONINFO resource (.rc file) for the Vitis AI provider DLL, following the same pattern used for CUDA, TensorRT, and QNN EPs (added in #24606). This embeds the ORT version into the DLL's PE header so it shows up in file properties.

### Motivation and Context
Need version in onnxruntime_providers_vitisai.dll to track changes.


adrianlizarraga:
Hi @mingyueliuh and @zz002, can you please take a look when you have a chance?

PR #27627, opened 2026-03-12
[QNN-EP] Disable file mapping for embedded cache

quic-calvnguy:
### Description
File mapping is enabled by default, but when the model contains an embedded EP context in a node, QNN EP attempts to map the EP context as a file that doesn't exist. This change is to disable file mapping entirely if econtext_embed_mode = 1.

### DIFF
--- a/onnxruntime/core/providers/qnn/qnn_execution_provider.cc
+++ b/onnxruntime/core/providers/qnn/qnn_execution_provider.cc
@@ -488,12 +488,12 @@ QNNExecutionProvider::QNNExecutionProvider(const ProviderOptions& provider_optio
   enable_file_mapped_weights_ = false;
   LOGS_DEFAULT(WARNING) << "File mapped weights feature is only available on Windows arm64 devices for QNN API versions >= 2.32. "
                         << "Feature will be disabled by default";
-#endif
-
+#else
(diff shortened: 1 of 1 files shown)


onnxruntime/core/providers/qnn/qnn_execution_provider.cc
yuslepukhin, line 488: Is this member present when file mapping is not available and if so why?

I would think that entire pieces of code should not be there as well as this logging message.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -490,6 +490,11 @@ QNNExecutionProvider::QNNExecutionProvider(const ProviderOptions& provider_optio
                         << "Feature will be disabled by default";
 #endif
 
+  if (qnn_context_embed_mode_ && enable_file_mapped_weights_) {
```
yuslepukhin: Would it not make sense to place it also under the macro?

quic-calvnguy: Moved so code is only executed when QNN_FILE_MAPPED_WEIGHTS_AVAILABLE is defined

PR #27630, opened 2026-03-12
[VitisAI] Fix unused warnings

amd-genmingz:
### Description
The code contains some unreferenced variables or functions, which will generate a warning.

### Motivation and Context
The change only removes two warnings, but it will cause compilation failure when a Werror enabled.


amd-genmingz:
@microsoft-github-policy-service agree company="AMD"

PR #27638, opened 2026-03-12
Improve checks in concat_slice_elimination against optional attrs and unsqueeze_elimination against invalid model

yuslepukhin:
This pull request introduces improvements to the `concat_slice_elimination` and `unsqueeze_elimination` optimizer passes, focusing on correctness, robustness, and code clarity. The changes include enhanced handling of optional Slice operator attributes, stricter validation of axes and steps, and improved error handling for invalid model inputs.

Improvements to `concat_slice_elimination`:

* Materialized default values for optional `axes` and `steps` in the Slice operator, ensuring safe indexing and alignment with ONNX defaults. (`onnxruntime/core/optimizer/concat_slice_elimination.cc`)
* Refined the fusion pattern to only allow `starts.size() == 1`, clarifying the scope of the optimization and preventing incorrect fusions. (`onnxruntime/core/optimizer/concat_slice_elimination.cc`)
* Added `<numeric>` include to support new code using `std::iota`. (`onnxruntime/core/optimizer/concat_slice_elimination.cc`)

Improvements to `unsqueeze_elimination`:

* Added validation for axes values, including range checks and detection of duplicate axes, returning errors for invalid models instead of silently proceeding. (`onnxruntime/core/optimizer/unsqueeze_elimination.cc`)
* Added `core/providers/common.h` include for utility functions used in validation. (`onnxruntime/core/optimizer/unsqueeze_elimination.cc`)


onnxruntime/core/optimizer/concat_slice_elimination.cc
```diff
@@ -219,7 +234,12 @@ bool ConcatSliceElimination::FuseConcatSliceSubgraph(Node& concat, Graph& graph,
   for (auto slice : concat_outputs) {
     InlinedVector<int64_t> starts, ends, axes, steps;
     if (!GetSliceInfo(graph, *slice, logger, starts, ends, axes, steps)) return false;
-    if (starts.size() > 1) return false;
+    // This check must come before any axes/steps indexing
+    // The code already enforces starts.size() == ends.size() (opset >=10)
```
edgchen1: "The code" refers to `GetSliceInfo()` right? it looks like `starts.size() == ends.size()` is also checked in that function for opset 1.

consider an `assert()` instead of (or in addition to) a comment.

onnxruntime/test/optimizer/graph_transform_test.cc
```diff
@@ -2534,6 +2534,408 @@ TEST_F(GraphTransformationTests, NegativeFuseConvAddNoBias) {
...
+  ASSERT_STATUS_OK(rule_transformer->Register(std::make_unique<UnsqueezeElimination>()));
+  ASSERT_STATUS_OK(TestGraphTransformer(build_test_case, 12, *logger_, std::move(rule_transformer),
+                                        TransformerLevel::Level1, 1, pre_graph_checker, post_graph_checker));
+}
+
+// NOTE: Duplicate-axis and out-of-range axis error paths in UnsqueezeElimination::Apply
+// are defense-in-depth guards. They cannot be exercised through ModelTestBuilder because
+// ONNX schema validation during graph.Resolve() rejects such invalid models before the
+// optimizer runs. The guards protect against malformed models that bypass schema validation
+// (e.g., hand-crafted protobufs or future opset changes).
```
edgchen1: how can we add some regression tests for the issues being fixed?

yuslepukhin: Regression tests have been added for `concat_slice_elimination`. Also added unit tests for `unsqueeze_elimination ` as none existed. However, it is difficult to create a regression test for the latter since we call Resolve() prior to invoking the optimizers (and in between) and ONNX shape inferencing catches both issues before.

edgchen1: "The guards protect against malformed models that bypass schema validation (e.g., hand-crafted protobufs or future opset changes)."

is it possible to bypass schema validation like this? wouldn't that happen during `Graph::Resolve()`?

yuslepukhin: I am not aware of such way to bypass it. I will remove the wording.

PR #27644, opened 2026-03-13
[Plugin EP] Check for nullptr before dereferencing

apwojcik:
It seems some outputs on the spanned list may be nullptrs. By checking for nullptr and skipping them if found, it does not seem to disturb proper execution on models.

That check was not required for legacy EP using the Host API.


adrianlizarraga:
Hi @apwojcik, can you please sync with the latest main branch to get CI checks to pass? Thanks.

onnxruntime/core/graph/ep_api_types.cc
```diff
@@ -351,7 +351,8 @@ static Status GetOutputIndex(const EpNode& producer_node,
   gsl::span<const EpValueInfo* const> outputs = producer_node.GetOutputsSpan();
 
   for (size_t i = 0; i < outputs.size(); i++) {
-    if (outputs[i]->GetName() == value_info_name) {
+    const auto output = outputs[i];
+    if (output != nullptr && output->GetName() == value_info_name) {
```
adrianlizarraga, line 355: Thanks for this fix. It looks like we check for null inputs but neglected to check for null outputs. I suppose this would occur for operators with optional outputs, which would be good to test. The ONNX [Unique](https://onnx.ai/onnx/operators/onnx__Unique.html) operator would probably the easies to test with.

adrianlizarraga (approved):

PR #27646, opened 2026-03-13
[CUDA] RoiAlign for opset versions 16 and 22

tianleiwu:
Support RoiAlign for opset versions 16 and 22


onnxruntime/test/providers/cpu/object_detection/roialign_test.cc
```diff
@@ -906,5 +907,78 @@ TEST(RoiAlignTest, BatchIndicesNegative_CUDA) {
...
+  if (cuda_ep.get() == nullptr) {
+    GTEST_SKIP() << "Skipping because there is no CUDA execution provider available.";
+  }
+
+  OpTester test("RoiAlign", 16);
+  test.AddAttribute<int64_t>("output_height", 1);
+  test.AddAttribute<int64_t>("output_width", 1);
+  test.AddAttribute<int64_t>("sampling_ratio", 1);
+  test.AddAttribute<float>("spatial_scale", 1.0f);
+  test.AddAttribute<std::string>("coordinate_transformation_mode", "v0_6");
```
yuslepukhin: >"v0_6" [](http://example.com/codeflow?start=67&length=6)

What kind of mode is this? #Resolved

onnxruntime/core/providers/cuda/object_detection/roialign.cc
```diff
@@ -67,13 +93,19 @@ Status RoiAlign<T>::ComputeInternal(OpKernelContext* context) const {
...
+#define SPECIALIZED_COMPUTE(T)          \
+  ADD_VERSIONED_TYPED_ROIALIGN_OP_10(T) \
+  ADD_VERSIONED_TYPED_ROIALIGN_OP_16(T) \
+  ADD_TYPED_ROIALIGN_OP_22(T)           \
   template Status RoiAlign<T>::ComputeInternal(OpKernelContext* ctx) const;
 
 SPECIALIZED_COMPUTE(float)
 SPECIALIZED_COMPUTE(double)
-// SPECIALIZED_COMPUTE(MLFloat16)
+SPECIALIZED_COMPUTE(MLFloat16)
```
yuslepukhin: This registers for version 15, but cuda_execution_provider does not register MLFloat16 for version 15. So this creates dead code? #Resolved

onnxruntime/core/providers/cuda/cuda_execution_provider.cc
```diff
@@ -1424,6 +1427,7 @@ class ONNX_OPERATOR_TYPED_KERNEL_CLASS_NAME(kCudaExecutionProvider, kOnnxDomain,
 class ONNX_OPERATOR_VERSIONED_KERNEL_CLASS_NAME(kCudaExecutionProvider, kOnnxDomain, 16, 17, ScatterElements);
 class ONNX_OPERATOR_VERSIONED_KERNEL_CLASS_NAME(kCudaExecutionProvider, kOnnxDomain, 16, 17, ScatterND);
 class ONNX_OPERATOR_VERSIONED_TYPED_KERNEL_CLASS_NAME(kCudaExecutionProvider, kOnnxDomain, 16, 19, float, GridSample);
+class ONNX_OPERATOR_VERSIONED_TYPED_KERNEL_CLASS_NAME(kCudaExecutionProvider, kOnnxDomain, 16, 21, MLFloat16, RoiAlign);
```
yuslepukhin: This is a duplicate with line 951 #Resolved

onnxruntime/core/providers/cuda/object_detection/roialign_impl.cu
```diff
@@ -17,64 +17,75 @@
...
-    T x,
+    AccumulationType_t<T> y,
+    AccumulationType_t<T> x,
     const bool is_mode_avg,
     const int index /* index for debug only*/) {
+  using TAcc = AccumulationType_t<T>;
+
+  TAcc y_acc = y;
+  TAcc x_acc = x;
+
```
yuslepukhin: the function parameters y and x are already AccumulationType_t<T>, but they're immediately copied to y_acc and x_acc of the same type. This is purely cosmetic/redundant. #Resolved

onnxruntime/test/providers/cpu/object_detection/roialign_test.cc
```diff
@@ -906,5 +907,78 @@ TEST(RoiAlignTest, BatchIndicesNegative_CUDA) {
   test.Run(OpTester::ExpectResult::kExpectSuccess, "", {}, nullptr, &execution_providers);
 #endif
 }
+
+TEST(RoiAlignTest, Float16_Opset16) {
```
yuslepukhin, line 911: The three new tests all use the same trivial 2x2 input with sampling_ratio=1. Consider adding:

- A test with "half_pixel" mode (the default for opset 16+) to verify the -0.5 offset path works correctly with half/bfloat16 types
- A test with mode="max" for the new types
- A test with sampling_ratio=0 (adaptive) to exercise the _Ceil() code path
- Larger spatial dimensions where precision differences between float16/bfloat16 and float32 could surface #Resolved

onnxruntime/core/providers/cuda/object_detection/roialign_impl.cu
```diff
@@ -17,64 +17,75 @@
...
+  TAcc w2 = hy * lx;
+  TAcc w3 = ly * hx;
+  TAcc w4 = ly * lx;
 
-  T val = is_mode_avg
-              ? (w1 * v1 + w2 * v2 + w3 * v3 + w4 * v4)             // mode Avg
-              : max(max(max(w1 * v1, w2 * v2), w3 * v3), w4 * v4);  // mode Max
+  TAcc val = is_mode_avg
+                 ? (w1 * v1 + w2 * v2 + w3 * v3 + w4 * v4)             // mode Avg
+                 : max(max(max(w1 * v1, w2 * v2), w3 * v3), w4 * v4);  // mode Max
```
yuslepukhin, line 85: >max(max(max(w1 * v1, w2 * v2), w3 * v3), w4 * v4); [](http://example.com/codeflow?start=19&length=50)

Weighted maximum of the four corner values is computed, rather than the standard bilinear interpolation w1*v1 + w2*v2 + w3*v3 + w4*v4. 

The reference implementations (e.g. Detectron2) always use bilinear interpolation at the sub-pixel sampling level and only apply max pooling when aggregating across the sampling grid within each bin. 

The weighting by w1..w4 in the max means the result is influenced by proximity to corners, not just the values themselves. #Resolved

yuslepukhin:
There is also an issue in the header which is a pre-existing issue but might as well be fixed now.

Default [coordinate_transformation_mode](vscode-file://vscode-app/c:/VSCode/07ff9d6178/resources/app/out/vs/code/electron-browser/workbench/workbench.html) for opset 16+:

[roialign.h:129-131](vscode-file://vscode-app/c:/VSCode/07ff9d6178/resources/app/out/vs/code/electron-browser/workbench/workbench.html) — The attribute is only read inside the if (GetAttr(...).IsOK()) guard. If the attribute is absent (which is valid per the spec), the if body is skipped and [half_pixel_](vscode-file://vscode-app/c:/VSCode/07ff9d6178/resources/app/out/vs/code/electron-browser/workbench/workbench.html) stays at its [roialign.h:143](vscode-file://vscode-app/c:/VSCode/07ff9d6178/resources/app/out/vs/code/electron-browser/workbench/workbench.html). But the ONNX spec for opset 16+ says the default for [coordinate_transformation_mode](vscode-file://vscode-app/c:/VSCode/07ff9d6178/resources/app/out/vs/code/electron-browser/workbench/workbench.html) is "half_pixel", which should set [half_pixel_ = true](vscode-file://vscode-app/c:/VSCode/07ff9d6178/resources/app/out/vs/code/electron-browser/workbench/workbench.html). So when the attribute is omitted, the kernel silently behaves as "output_half_pixel" instead of the spec-mandated "half_pixel".

Note: for opset 10 (which has no [coordinate_transformation_mode](vscode-file://vscode-app/c:/VSCode/07ff9d6178/resources/app/out/vs/code/electron-browser/workbench/workbench.html) attribute), the false default is correct — it matches the original opset 10 behavior. #Resolved

yuslepukhin (changes requested):
:clock1:

onnxruntime/core/providers/cuda/object_detection/roialign_impl.cu
tianleiwu: Unfortunately, this is the onnx reference: https://github.com/onnx/onnx/blob/e30c6935d67cc3eca2fa284e37248e7c0036c46b/onnx/reference/ops/op_roi_align.py#L235-L244

I have a draft for Detectron2 like definition: https://github.com/microsoft/onnxruntime/pull/27554. That is not in the scope of this PR.

yuslepukhin:
[roialign.h:129-131](vscode-file://vscode-app/c:/VSCode/07ff9d6178/resources/app/out/vs/code/electron-browser/workbench/workbench.html) — The attribute is only read inside the if (GetAttr(...).IsOK()) guard.

If the attribute is absent (which is valid per the spec), the if body is skipped and [half_pixel_](vscode-file://vscode-app/c:/VSCode/07ff9d6178/resources/app/out/vs/code/electron-browser/workbench/workbench.html) stays at its [roialign.h:143](vscode-file://vscode-app/c:/VSCode/07ff9d6178/resources/app/out/vs/code/electron-browser/workbench/workbench.html). 

But the ONNX spec for opset 16+ says the default for [coordinate_transformation_mode](vscode-file://vscode-app/c:/VSCode/07ff9d6178/resources/app/out/vs/code/electron-browser/workbench/workbench.html) is "half_pixel", which should set [half_pixel_ = true](vscode-file://vscode-app/c:/VSCode/07ff9d6178/resources/app/out/vs/code/electron-browser/workbench/workbench.html). So when the attribute is omitted, the kernel silently behaves as "output_half_pixel" instead of the spec-mandated "half_pixel".

tianleiwu:
> roialign.h:129-131 — The attribute is only read inside the if (GetAttr(...).IsOK()) guard.
> 
> If the attribute is absent (which is valid per the spec), the if body is skipped and half_pixel_ stays at its roialign.h:143.
> 
> But the ONNX spec for opset 16+ says the default for coordinate_transformation_mode is "half_pixel", which should set half_pixel_ = true. So when the attribute is omitted, the kernel silently behaves as "output_half_pixel" instead of the spec-mandated "half_pixel".

Current logic: 
```
If there is coordinate_transformation_mode attribute
    set half_pixel_ according to whether the attribute value is "half_pixel". 
else
    use the default value according to spec. The default is "half_pixel" for opset 16+, and "output_half_pixel" for opset 10.
``` #Resolved

PR #13706, opened 2022-11-19
Session-aware and thread-safe CUDA profiler

abudup:
### Description
The existing CUDA profiler is neither session-aware, nor thread-safe. This PR ensures both.

### Motivation and Context
[PR 13549](https://github.com/microsoft/onnxruntime/pull/13549) brought thread-safety and session-awareness to the ROCm profiler. This PR brings the same goodness to the CUDA profiler as well.

Sample outputs of a profiling run from the StableDiffusion model (this model was chosen because it requires orchestration of multiple sessions, and verifies that the profilers are now indeed session-aware) on both CUDA and ROCm EPs are attached, along with a script that checks that the trace files generated by the profile are well-formed.

Update 11/29: Updated the profile outputs. The older profile outputs exhibited an issue where some timestamps were wildly out of range, leading to problems visualizing the traces. The bug has been fixed and the profile outputs have been updated, along with an update to the check script to ensure that timestamps are monotonically increasing.

[sd_profile_outputs_cuda.tar.gz](https://github.com/microsoft/onnxruntime/files/10118088/sd_profile_outputs_cuda.tar.gz)
[sd_profile_outputs_rocm.tar.gz](https://github.com/microsoft/onnxruntime/files/10118089/sd_profile_outputs_rocm.tar.gz)
[check_profile_output_well_formedness.zip](https://github.com/microsoft/onnxruntime/files/10118090/check_profile_output_well_formedness.zip)

PR #27649, opened 2026-03-13
Plugin EP event profiling APIs

adrianlizarraga:
#### TLDR

This PR ports the existing C++ [EpProfiler](https://github.com/microsoft/onnxruntime/blob/faad20f9d3264c7f3b6d4e4398990e13ee864512/include/onnxruntime/core/framework/execution_provider.h#L359) interfaces used by provider-bridge EPs to the binary-stable C APIs for plugin EPs. It introduces C/C++ APIs for creating/querying profiling events, a container for appending EP events, and callback hooks (`StartEvent`/`StopEvent`) that give EPs access to ORT event metadata in real-time.

#### Changes to the original C++ API

The original `EpProfiler` C++ interface was adapted for the C API with the following intentional changes:

1. **`StartProfiling`** now receives an offset indicating the elapsed time since profiling started, as opposed to receiving an absolute/epoch-dependent profiling start time. This prevents EPs from having to do epoch conversions. Credit to @edgchen1 for the idea.
2. **`StartEvent`/`StopEvent` receive an absolute, epoch-based correlation ID (`ort_event_correlation_id`)** instead of a relative ORT event ID. The `PluginEpProfiler` bridge layer automatically converts the C++ `relative_ort_event_id` (microseconds since profiling start) to an absolute `ort_event_correlation_id` by adding the epoch-based profiling start time. This means plugin EPs can use the correlation ID directly with profiling utilities like CUPTI or ROCTracer without computing the conversion themselves.
3. **`StopEvent` now receives the completed ORT event as a parameter.** This allows EPs to optionally inspect ORT event metadata (e.g., `op_name`, `event_name`) at the time the event ends, facilitating annotation of correlated EP events.
4. **`EndProfiling` only allows EPs to *append* events (via `OrtProfilingEventsContainer`), not read or modify the full events array.** This is motivated by:
   - Prevent any one EP from modifying events generated by ORT or another EP.
   - Certain EPs (VitisAI and WebGPU) already only append events without reading the entire events array.
   - The CUDA EP reads the entire events array solely to merge/sort its own EP events next to correlated ORT events and add `parent_name`/`op_name` metadata. However:
     - Merging/sorting is mostly unnecessary since trace viewers that load these files do their own event sorting.
     - This merging/sorting step was previously required to augment CUDA EP events with metadata from the correlated ORT event. However, that can now be obtained more simply via the new `StopEvent` parameter that provides the EP with the full correlated ORT event.
     - The [merge algorithm used by CUDA EP](https://github.com/microsoft/onnxruntime/blob/faad20f9d3264c7f3b6d4e4398990e13ee864512/include/onnxruntime/core/common/gpu_profiler_common.h#L391-L397) **incorrectly** assumes ORT events are sorted by non-decreasing *start* time, but they are actually sorted by [non-decreasing *end* time](https://github.com/microsoft/onnxruntime/blob/faad20f9d3264c7f3b6d4e4398990e13ee864512/onnxruntime/core/common/profiler.cc#L91) (also see https://github.com/microsoft/onnxruntime/pull/13706#discussion_r1042750808). Fixing this would require sorting the entire Events array before asking a provider-bridge EP to merge in its events into the global events array. Not sure this is worth the runtime cost.

#### Naming conventions for ORT event IDs

- **C++ `EpProfiler` interface** (existing): Uses `relative_ort_event_id` — a timestamp offset in microseconds relative to profiling start.
- **C API `OrtEpProfilerImpl`** (new in this PR): Uses `ort_event_correlation_id` — an absolute, epoch-based timestamp in microseconds computed from `std::chrono::high_resolution_clock` (platform-defined epoch). Unique across concurrent profiling sessions within the same process.
- **Conversion**: The `PluginEpProfiler` bridge class (in `ep_event_profiling.cc`) performs `ort_event_correlation_id = relative_ort_event_id + profiling_start_time_epoch_us_`, mirroring the pattern in `GPUTracerManager::PushCorrelation`.

### New C APIs

| API | Description |
|-----|-------------|
| `CreateProfilingEvent` | Create a profiling event with category, process/thread IDs, name, timestamp, duration, and key-value args |
| `ReleaseProfilingEvent` | Release a profiling event |
| `ProfilingEvent_GetCategory` | Get event category (`SESSION`, `NODE`, `KERNEL`, `API`) |
| `ProfilingEvent_GetName` | Get event name |
| `ProfilingEvent_GetTimestampUs` | Get event start timestamp (µs) |
| `ProfilingEvent_GetDurationUs` | Get event duration (µs) |
| `ProfilingEvent_GetArgValue` | Get an event argument value by key |
| `ProfilingEventsContainer_AddEvents` | Append an array of EP events to the output container |
| `OrtEp::CreateProfiler` | Returns an instance of the EP's profiler implementation |
| `OrtEpProfilerImpl::StartProfiling` | Called by ORT to start a profiling session. Receives elapsed time offset (ns) since ORT profiling started |
| `OrtEpProfilerImpl::StartEvent` | Called by ORT to notify that an ORT event has started. Receives an absolute `ort_event_correlation_id` |
| `OrtEpProfilerImpl::StopEvent` | Called by ORT to notify that an ORT event has ended. Receives the same `ort_event_correlation_id` and ORT event metadata |
| `OrtEpProfilerImpl::EndProfiling` | Called by ORT to end the profiling session and collect EP events into the output container |
| `OrtEpProfilerImpl::Release` | Release the profiler instance |

### New C++ wrapper classes

| Class | Description |
|-------|-------------|
| `Ort::ConstProfilingEvent` | Non-owning const wrapper for reading fields from an `OrtProfilingEvent` (e.g., in `StopEvent`) |
| `Ort::ProfilingEvent` | Owning wrapper that creates and manages an `OrtProfilingEvent` (e.g., for `EndProfiling`) |
| `Ort::UnownedProfilingEventsContainer` | Non-owning wrapper for adding events to an `OrtProfilingEventsContainer` during `EndProfiling` |

### Example EP profiling implementation
This PR updates an example plugin EP to use the new profiling APIs:
- Plugin EP code: [test/autoep/library/example_plugin_ep_kernel_registry](https://github.com/microsoft/onnxruntime/tree/adrianl/PluginEp_ProfilingApis/onnxruntime/test/autoep/library/example_plugin_ep_kernel_registry)
- `OrtEpProfilerImpl` implementation: [ep_profiling.h](https://github.com/microsoft/onnxruntime/blob/adrianl/PluginEp_ProfilingApis/onnxruntime/test/autoep/library/example_plugin_ep_kernel_registry/ep_profiling.h) / [ep_profiling.cc](https://github.com/microsoft/onnxruntime/blob/adrianl/PluginEp_ProfilingApis/onnxruntime/test/autoep/library/example_plugin_ep_kernel_registry/ep_profiling.cc)
- `OrtEp::CreateProfiler()` implementation: [ep.cc](https://github.com/microsoft/onnxruntime/blob/adrianl/PluginEp_ProfilingApis/onnxruntime/test/autoep/library/example_plugin_ep_kernel_registry/ep.cc)

### Existing bugs found
Not fixed in this PR.

 - The [merge algorithm used by CUDA EP](https://github.com/microsoft/onnxruntime/blob/faad20f9d3264c7f3b6d4e4398990e13ee864512/include/onnxruntime/core/common/gpu_profiler_common.h#L391-L397) **incorrectly** assumes ORT events are sorted by non-decreasing *start* time, but they are actually sorted by [non-decreasing *end* time](https://github.com/microsoft/onnxruntime/blob/faad20f9d3264c7f3b6d4e4398990e13ee864512/onnxruntime/core/common/profiler.cc#L91) (also see https://github.com/microsoft/onnxruntime/pull/13706#discussion_r1042750808).
 - Run profilers do not handle subgraphs (e.g., subgraph of a control-flow operator). Has been the case since run profilers were [introduced](https://github.com/microsoft/onnxruntime/pull/26846).

### Motivation and Context
Allows plugin EPs to generate profiling events, further closing the functionality gap between provider-bridge EPs and plugin EPs.


include/onnxruntime/core/common/profiler_common.h
```diff
@@ -11,6 +11,8 @@
 namespace onnxruntime {
 namespace profiling {
 
+// Profiling event categories.
+// Note: Keep in sync with OrtEpProfilingEventCategory in onnxruntime_ep_c_api.h.
```
adrianlizarraga: Note:  just added comments to this file. documentation was missing for provider-bridge EP profilers.

onnxruntime/test/autoep/library/example_plugin_ep_kernel_registry/ep_profiling.cc
```diff
@@ -0,0 +1,234 @@
...
+void EpEventManager::AddEpEvent(uint64_t client_id, Event event) {
+  std::lock_guard<std::mutex> lock(mutex_);
+  auto iter = client_state_.find(client_id);
+
+  if (iter == client_state_.end()) {
+    return;
+  }
+
+  iter->second.events.push_back(std::move(event));
+}
```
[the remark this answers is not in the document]
adrianlizarraga: updated

onnxruntime/core/session/plugin_ep/ep_api.cc
```diff
@@ -806,6 +808,122 @@ ORT_API_STATUS_IMPL(GetEnvConfigEntries, _Outptr_ OrtKeyValuePairs** config_entr
...
+      args[key] = value;
+    }
+  }
+
+  auto record = std::make_unique<profiling::EventRecord>(
+      static_cast<onnxruntime::profiling::EventCategory>(category_value),
+      static_cast<int>(process_id),
+      static_cast<int>(thread_id),
+      std::string(event_name ? event_name : ""),
+      static_cast<long long>(timestamp_us),
```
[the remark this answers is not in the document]
adrianlizarraga: Updated to return an error if the `event_name` is null

onnxruntime/test/autoep/library/example_plugin_ep_kernel_registry/ep_profiling.cc
```diff
@@ -0,0 +1,224 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+#include "ep_profiling.h"
```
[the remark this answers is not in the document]
adrianlizarraga: updated

onnxruntime/test/autoep/library/example_plugin_ep_kernel_registry/kernels/binary_op.cc
```diff
@@ -72,6 +75,15 @@ void ORT_API_CALL BinaryOp::ReleaseImpl(OrtKernelImpl* this_ptr) noexcept {
...
   BinaryOp* binary_op_kernel = static_cast<BinaryOp*>(this_ptr);
+  const ExampleKernelEp* ep = static_cast<const ExampleKernelEp*>(binary_op_kernel->info_.GetEp());
+
+  std::optional<uint64_t> profiler_client_id = ep != nullptr ? ep->GetProfilerClientId() : std::nullopt;
+  std::chrono::high_resolution_clock::time_point kernel_start_ts;
+  std::chrono::high_resolution_clock::time_point kernel_end_ts;
+
+  if (profiler_client_id.has_value()) {
+    kernel_start_ts = std::chrono::high_resolution_clock::now();
+  }
```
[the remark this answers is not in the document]
adrianlizarraga: updated to support both:
1. a single session profiler (one profiler id) that is used across different concurrent runs.
2. multiple run profilers (each with a different profiler id) that are used concurrently.

onnxruntime/core/session/plugin_ep/ep_api.cc
```diff
@@ -878,7 +1005,6 @@ static_assert(offsetof(OrtEpApi, GetSyncIdForLastWaitOnSyncStream) / sizeof(void
               "Size of version 23 API cannot change");
 static_assert(offsetof(OrtEpApi, GetEnvConfigEntries) / sizeof(void*) == 49,
               "Size of version 24 API cannot change");
```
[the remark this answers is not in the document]
adrianlizarraga: We typically only do this when we cut a release.

include/onnxruntime/core/session/onnxruntime_cxx_inline.h
```diff
@@ -802,6 +802,75 @@ inline std::string EpAssignedNodeImpl<T>::GetOperatorType() const {
...
+inline Status ProfilingEventsContainerImpl<T>::AddEvents(const OrtProfilingEvent* const* events, size_t num_events) {
+  return Status{GetEpApi().ProfilingEventsContainer_AddEvents(this->p_, events, num_events)};
+}
+
+template <typename T>
+inline Status ProfilingEventsContainerImpl<T>::AddEvents(const std::vector<ProfilingEvent>& events) {
+  static_assert(sizeof(ProfilingEvent) == sizeof(const OrtProfilingEvent*),
+                "ProfilingEvent must have the same size as a raw pointer for reinterpret_cast");
+  const auto* event_ptrs = reinterpret_cast<const OrtProfilingEvent* const*>(events.data());
+  return AddEvents(event_ptrs, events.size());
```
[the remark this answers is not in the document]
adrianlizarraga: This is a good thing to note. Existing practice within the onnxruntime_cxx_inline.h file has been to include a `static_cast` and allow the `reinterpret_cast`.

onnxruntime/core/session/plugin_ep/ep_api.cc
```diff
@@ -869,6 +925,8 @@ static constexpr OrtEpApi ort_ep_api = {
     &OrtExecutionProviderApi::ReleaseKernelImpl,
     &OrtExecutionProviderApi::GetEnvConfigEntries,
     // End of Version 24 - DO NOT MODIFY ABOVE
+
+    &OrtExecutionProviderApi::EpProfilingEventsContainer_AddEvents,
 };
```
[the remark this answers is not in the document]
yuslepukhin: This needs to be addressed.

onnxruntime/core/session/plugin_ep/ep_event_profiling.cc
```diff
@@ -0,0 +1,93 @@
...
+PluginEpProfiler::~PluginEpProfiler() {
+  profiler_impl_.Release(&profiler_impl_);
+}
+
+bool PluginEpProfiler::StartProfiling(TimePoint profiling_start_time) {
+  int64_t ns = std::chrono::duration_cast<std::chrono::nanoseconds>(
+                   profiling_start_time.time_since_epoch())
+                   .count();
+  return profiler_impl_.StartProfiling(&profiler_impl_, ns);
+}
```
yuslepukhin: Let's add a comment about time_sinc_epoch()

This uses .time_since_epoch() which works with whatever epoch high_resolution_clock uses. The documentation says "nanoseconds since epoch (from std::chrono::high_resolution_clock)" which is technically correct but could be clearer that this may not be Unix epoch on all platforms. Since both ORT and the plugin use the same clock, the relative offsets are consistent — this is fine in practice.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -1444,6 +1607,120 @@ struct OrtEpApi {
...
+   */
+  ORT_API2_STATUS(CreateProfilingEvent,
+                  _In_ OrtProfilingEventCategory category,
+                  _In_ int32_t process_id,
+                  _In_ int32_t thread_id,
+                  _In_ const char* event_name,
+                  _In_ int64_t timestamp_us,
+                  _In_ int64_t duration_us,
+                  _In_reads_(num_args) const char* const* arg_keys,
+                  _In_reads_(num_args) const char* const* arg_values,
```
yuslepukhin, line 1918: Can this be OrtKeyValuePair?

onnxruntime/core/session/plugin_ep/ep_api.cc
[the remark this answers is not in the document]
adrianlizarraga: If I'm not mistaken, I thought we typically only added the static_assert at the time we cut the release.
I suppose it wouldn't be much of an issue to add it as we develop.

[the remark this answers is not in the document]
yuslepukhin: I think we update it everytime we add an entry.

onnxruntime/test/autoep/test_execution.cc
```diff
@@ -935,6 +936,133 @@ TEST(OrtEpLibrary, KernelPluginEp_ControlFlow_Scan) {
...
+  profile.close();
+
+  // Verify Mul kernel timing events generated by EP are present.
+  EXPECT_TRUE(content.find("Kernel") != std::string::npos &&
+              content.find("ExampleKernelEp_Mul") != std::string::npos)
+      << "Expected to find Mul kernel op timing events in profile output. Profile content:\n"
+      << content;
+
+  // Clean up profile file.
+  std::filesystem::remove(profile_path);
```
yuslepukhin: >std::filesystem::remove(profile_path); [](http://example.com/codeflow?start=2&length=38)

I feel this needs RAII

yuslepukhin:
Overall pretty good.

onnxruntime/test/autoep/test_execution.cc
adrianlizarraga: This now handles cleanup with a `gsl::finally`.

include/onnxruntime/core/common/profiler_common.h
```diff
@@ -79,10 +81,61 @@ using Events = std::vector<EventRecord>;
...
+
+  /// <summary>
+  /// Called when profiling ends to collect the EP's new profiling events since the last call to StartProfiling.
+  /// </summary>
+  /// <param name="start_time">Timepoint denoting the start of profiling. Same value passed to StartProfiling.</param>
+  /// <param name="events">Modifiable events container to which the EP profiler appends its events.</param>
+  virtual void EndProfiling(TimePoint start_time, Events& events) = 0;
+
+  /// <summary>
+  /// Optionally called when an ORT event (e.g., session initialization, node kernel execution, etc.) starts.
```
edgchen1: could you elaborate on what "optionally" means? is it when profiling is enabled?

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -1741,6 +2018,31 @@ struct OrtEp {
...
+
+  /** \brief Gets the execution provider's profiler, if any.
+   *
+   * If the EP supports profiling, it should create and return an OrtEpProfilerImpl instance.
+   * ORT takes ownership of each non-NULL instance returned and will call OrtEpProfilerImpl::Release when
+   * it is no longer needed.
+   *
+   * ORT may call this function multiple times over the lifetime of a single OrtEp instance, for example
+   * during EP registration and again when run-level profiling is enabled. Each call is independent and
+   * the EP must treat it as a request for a new profiler instance (or NULL if profiling is not supported
```
edgchen1: > EP must treat it as a request for a new profiler instance

does this mean that the EP must return a new profiler instance, or NULL? if so, would `CreateProfiler()` be a better name?

onnxruntime/core/session/plugin_ep/ep_api.cc
```diff
@@ -806,6 +808,122 @@ ORT_API_STATUS_IMPL(GetEnvConfigEntries, _Outptr_ OrtKeyValuePairs** config_entr
...
+  const auto* record = FromOpaqueProfilingEvent(event);
+  return static_cast<int64_t>(record->dur);
+}
+
+ORT_API_STATUS_IMPL(ProfilingEvent_GetArgValue,
+                    _In_ const OrtProfilingEvent* event,
+                    _In_ const char* key,
+                    _Outptr_result_maybenull_ const char** out) {
+  API_IMPL_BEGIN
+  ORT_API_RETURN_IF(event == nullptr, ORT_INVALID_ARGUMENT, "OrtProfilingEvent is NULL");
```
edgchen1, line 1164: why do we check `event != nullptr` here but not in the other accessors?

onnxruntime/core/session/plugin_ep/ep_event_profiling.cc
```diff
@@ -0,0 +1,93 @@
...
+#include "core/framework/error_code_helper.h"
+
+namespace onnxruntime {
+
+/*status*/
+Status PluginEpProfiler::Create(OrtEpProfilerImpl& profiler_impl, const logging::Logger& logger,
+                                const std::string& ep_name, std::unique_ptr<PluginEpProfiler>& profiler_out) {
+  // plugin EP profiling APIs were introduced in ORT 1.25
+  if (auto profiler_version = profiler_impl.ort_version_supported; profiler_version < 25) {
+    if (profiler_impl.Release != nullptr) {
```
edgchen1: `Release` is available starting from 1.25. should we be calling it in this branch?

onnxruntime/core/session/plugin_ep/ep_event_profiling.cc
```diff
@@ -0,0 +1,93 @@
...
+
+  // Append EP events to the overall events list.
+  events.reserve(events.size() + ep_events_container.events.size());
+  for (auto& record : ep_events_container.events) {
+    events.emplace_back(std::move(record));
+  }
+}
+
+void PluginEpProfiler::Start(uint64_t ort_event_id) {
+  if (profiler_impl_.ort_version_supported >= 25 && profiler_impl_.StartEvent != nullptr) {
```
edgchen1: isn't version >= 25 already verified by `PluginEpProfiler::Create()`?

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -423,6 +426,166 @@ struct OrtExternalResourceImporterImpl {
...
+   * \since Version 1.25.
+   */
+  ORT_API_T(void, Release, _In_ OrtEpProfilerImpl* this_ptr);
+
+  /** \brief Called when profiling starts.
+   *
+   * Allows EP profiler to initialize profiling utilities and record the profiling start time.
+   *
+   * \param[in] this_ptr Pointer to the OrtEpProfilerImpl instance.
+   * \param[in] profiling_start_time_ns The profiling start time in nanoseconds since epoch
```
edgchen1: is there a good reason for using nanoseconds here and microseconds elsewhere, like in the profiling event?

onnxruntime/test/autoep/library/example_plugin_ep_kernel_registry/ep_profiling.cc
```diff
@@ -0,0 +1,206 @@
...
+  }
+
+  size_t start_index = tls_profiling_state_.ort_event_start_indices.back();
+  tls_profiling_state_.ort_event_start_indices.pop_back();
+
+  // Annotate this thread's EP events (added since StartEvent) with metadata from the correlated ORT event.
+  auto current_thread_id = std::this_thread::get_id();
+  for (size_t i = start_index; i < iter->second.events.size(); ++i) {
+    if (iter->second.events[i].thread_id == current_thread_id) {
+      iter->second.events[i].ort_event_name = ort_event_name;
```
edgchen1: I believe this could overwrite an existing annotation (e.g., if a more inner event was already popped)

onnxruntime/test/autoep/library/example_plugin_ep_kernel_registry/ep_profiling.h
```diff
@@ -0,0 +1,99 @@
...
+/// </summary>
+class EpEventManager {
+ public:
+  struct Event {
+    Event(std::string event_name, int64_t timestamp_ns, int64_t duration_ns)
+        : name(std::move(event_name)), ts_ns(timestamp_ns), dur_ns(duration_ns), thread_id(std::this_thread::get_id()) {}
+
+    std::string name;
+    int64_t ts_ns;
+    int64_t dur_ns;
```
edgchen1: we're in C++ code already - would it be easier to use the `std::chrono` types?

onnxruntime/test/autoep/library/example_plugin_ep_kernel_registry/ep_profiling.cc
```diff
@@ -0,0 +1,206 @@
...
+  if (raw_ep_events.empty()) {
+    return nullptr;
+  }
+
+  std::vector<Ort::ProfilingEvent> events;
+  events.reserve(raw_ep_events.size());
+
+  for (EpEventManager::Event& raw_ep_event : raw_ep_events) {
+    // ORT requires timestamps relative to the profiling start time. This example EP uses timestamps with the same
+    // epoch as ORT (i.e., Unix epoch), so we can just subtract off `profiling_start_time_ns` from `raw_ep_event.ts_ns`.
```
edgchen1: > (i.e., Unix epoch)

as @yuslepukhin mentioned, UNIX epoch is not guaranteed

edgchen1: for an EP which may not know what epoch the ORT profiling timestamp is using, how would it go about producing relative timestamps? would it be useful for ORT to pass in the time since profiling started?

onnxruntime/test/autoep/test_execution.cc
```diff
@@ -935,6 +936,105 @@ TEST(OrtEpLibrary, KernelPluginEp_ControlFlow_Scan) {
...
+    Ort::AllocatorWithDefaultOptions allocator;
+    Ort::AllocatedStringPtr profile_file = session.EndProfilingAllocated(allocator);
+    profile_file_path = profile_file.get();
+  } else {
+    // The APIs for run-profiling do not support retrieving the name of the actual profile file
+    // generated by ORT, so find it by prefix in the current directory.
+    for (const auto& entry : std::filesystem::directory_iterator(ORT_TSTR("."))) {
+      auto filename = entry.path().filename().native();
+      if (filename.find(profile_prefix) == 0 &&
+          filename.find(ORT_TSTR(".json")) != std::basic_string<ORTCHAR_T>::npos) {
```
edgchen1: string has `starts_with` and `ends_with` in c++20

onnxruntime/test/autoep/library/example_plugin_ep_kernel_registry/ep_profiling.cc
```diff
@@ -0,0 +1,206 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+#include "ep_profiling.h"
+#include <array>
+#include <chrono>
+#include <optional>
+
+// Thread-local profiling state. Tracks the active profiler ID and a per-thread stack of ORT event
+// boundaries. Per-thread state is necessary because a single session profiler (single profiler_id)
+// can have StartEvent/StopEvent called from multiple threads (e.g., via inter-op parallelism).
```
edgchen1, line 10: perhaps we should clarify the behavior of how StartEvent/StopEvent may be called from different threads and how profilers are supposed to interpret that. would it make sense for ORT to also pass in a thread id?

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
adrianlizarraga: We use nanoseconds as much as we can for precision. We only use microseconds when we have to save the event data into an `EventRecord`. Note that `EventRecord` is based on the [Chrome Trace Event format](https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU/preview?tab=t.0#heading=h.yr4qxyxotyw), which requires microseconds.

Generally, this is also consistent with the existing provider-bridge apis. I'll update API docs with a mention of the Trace Event Format.

onnxruntime/test/autoep/library/example_plugin_ep_kernel_registry/ep_profiling.cc
adrianlizarraga: Yep, good catch

onnxruntime/test/autoep/library/example_plugin_ep_kernel_registry/ep_profiling.h
adrianlizarraga: I don't necessarily think it would be easier. We're going to be converting this back to `int64_t` to give it back to ORT. The only thing that would change is where that happens (either in the EP kernel or in the EP profiler).

include/onnxruntime/core/common/profiler_common.h
adrianlizarraga: Bad wording. I only meant that this virtual function does not need to overridden as it has a default implementation that does nothing. I'll update

onnxruntime/core/session/plugin_ep/ep_event_profiling.cc
adrianlizarraga: hmmm maybe? If the user unintentionally set the wrong version but provided a Release function, we could call it to free memory. I originally did not call Release here, but every copilot review flagged it, and I can see it both ways.

What is your preference?

onnxruntime/core/session/plugin_ep/ep_event_profiling.cc
adrianlizarraga: Yes, this is redundant in this case because `StartEvent` happened to be introduced in the same version that introduced the entire `OrtEpProfilerImpl` struct. However, I'm leaning on keeping it here to serve as an example for when/if a new function is added. It may help another developer remember to check for the correct version that introduced that function.

onnxruntime/test/autoep/library/example_plugin_ep_kernel_registry/ep_profiling.cc
adrianlizarraga: The time since profiling started is already passed to the EPs via `StartProfiling` and `EndProfiling`. However, those times use ORT's clock, which may have a different epoch than whatever clock the EP is using.

That said, I have added a new way for an EP to compute the offset between the ORT epoch and EP epoch. It is based on this existing approach: https://github.com/microsoft/onnxruntime/blob/44ce4f5f13fc332285a9cd865f5e729dcca084c3/include/onnxruntime/core/common/gpu_profiler_common.h#L222-L236

It requires adding a new C API for the EP to get a timestamp using ORT's clock. I'll post a reply when that is pushed up.

onnxruntime/core/session/plugin_ep/ep_api.cc
adrianlizarraga: I've been meaning to bring this up. If we always check that pointer parameters annotated with `_In_` are not null, then we can never really have any C API function (for an opaque type) that does not return a status.

I think we've been inconsistent here (e.g., see `HardwareDevice_Vendor` and other accessors).

I'll update all of these to return a status (a bit of a shame).

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
adrianlizarraga: Yes, a new instance or NULL. I'll update to `CreateProfiler()`.

onnxruntime/test/autoep/library/example_plugin_ep_kernel_registry/ep_profiling.cc
adrianlizarraga: I'll try to update docs to summarize the different concurrency scenarios for session and run profilers. This was not really documented in the original C++ APIs , so I'll give it my best shot.

Regarding passing in a thread id to `StartEvent`:
We can't really do that because the value returned by `std::this_thread::get_id()` is dependent on the C++ std library implementation. An EP dll could be using a different lib (or could be a pure C implementation with posix threads, etc.).

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
adrianlizarraga: I don't think this would be an improvement. Using a `OrtKeyValuePair` here would add more friction for the caller as they would have to convert their keys/values from whatever they are originally using (e.g., a `unordered_map` or plain C arrays) into an `OrtKeyValuePair` object, which ORT will just have to deconstruct back into a `InlinedHashMap`. It would just add extra allocations and copies for two less parameter.

onnxruntime/core/session/plugin_ep/ep_event_profiling.cc
edgchen1: I interpret `version < 25` to not guarantee the existence of a Release function. in this case, the `OrtEpProfilerImpl` struct is not available at all, so who knows what is being passed in. my preference would be to not attempt to access anything else.

onnxruntime/test/autoep/library/example_plugin_ep_kernel_registry/ep_profiling.h
edgchen1: IMO, I would prefer to use the time-related types as much as we can over raw ints. we can convert at the API boundary. maybe "harder to misuse" would be a better way to put it.

onnxruntime/test/autoep/library/example_plugin_ep_kernel_registry/ep_profiling.cc
edgchen1: > The time since profiling started is already passed to the EPs via StartProfiling and EndProfiling.

it's not exactly what I meant. instead of a timestamp of when profiling started (based off of some ORT epoch), we could provide a duration (how much time has elapsed since profiling started).

since it appears that ORT profiling may start much earlier than the call to the EP's `StartProfiling`, an EP could record its own profiling start time too and use that to get a relative timestamp for EP events.

something like this (offset means duration since ort profiling started, time_points are measured by EP):
`ep_event_offset = ep_profile_start_offset + (ep_event_time_point - ep_profiling_start_time_point)`

adrianlizarraga: This is a great idea. Thanks. It should hopefully simplify things. I'll try it.

onnxruntime/core/session/plugin_ep/ep_plugin_provider_interfaces.cc
```diff
@@ -777,4 +782,33 @@ const OrtEp* PluginExecutionProvider::GetOrtEp() const {
...
+  if (profiler_impl == nullptr) {
+    return {};  // plugin EP doesn't have a profiler
+  }
+
+  std::unique_ptr<PluginEpProfiler> ep_profiler;
+  status = PluginEpProfiler::Create(*profiler_impl, logger, Type(), /*out*/ ep_profiler);
+  if (!status.IsOK()) {
+    LOGS(logger, ERROR) << status.ErrorMessage();
+    return {};
+  }
```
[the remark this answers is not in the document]
adrianlizarraga: PluginEpProfiler::Create() will release `profiler_impl` before it returns with an error. There is only one exception (which we're ok with): an `ort_version_supported` version that does not support `Release()` is set on the profiler_impl.

onnxruntime/core/session/plugin_ep/ep_event_profiling.cc
```diff
@@ -0,0 +1,122 @@
...
+  if (auto profiler_version = profiler_impl.ort_version_supported; profiler_version < 25) {
+    // Note: it is not clear whether ORT should try to release the EP's profiler if the version is incorrect
+    // since Release() was introduced in version 25 (along with the all profiling APIs).
+    // if (profiler_impl.Release != nullptr) {
+    //   profiler_impl.Release(&profiler_impl);
+    // }
+
+    return ORT_MAKE_STATUS(ONNXRUNTIME, FAIL, "OrtEpProfilerImpl::ort_version_supported (", profiler_version, ") for ",
+                           ep_name, " expected to be >= 25");
+  }
```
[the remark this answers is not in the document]
adrianlizarraga: We've agreed to let this leak since `Release()` is not supported before version 25: https://github.com/microsoft/onnxruntime/pull/27649#discussion_r2991546059

onnxruntime/test/autoep/test_execution.cc
```diff
@@ -935,6 +937,158 @@ TEST(OrtEpLibrary, KernelPluginEp_ControlFlow_Scan) {
...
+    for (const auto& entry : std::filesystem::directory_iterator(ORT_TSTR("."))) {
+      auto filename = entry.path().filename().native();
+      if (filename.starts_with(profile_prefix) && filename.ends_with(ORT_TSTR(".json"))) {
+        profile_file_path = entry.path();
+        break;
+      }
+    }
+
+    ASSERT_FALSE(profile_file_path.empty()) << "Could not find run profile with prefix '"
+                                            << profile_prefix << "' in current directory";
```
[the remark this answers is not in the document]
adrianlizarraga: updated.

onnxruntime/test/autoep/test_execution.cc
```diff
@@ -935,6 +937,158 @@ TEST(OrtEpLibrary, KernelPluginEp_ControlFlow_Scan) {
...
+  }
+
+  auto cleanup_profile_file = gsl::finally([&profile_file_path] { std::filesystem::remove(profile_file_path); });
+  std::ifstream profile(profile_file_path);
+  ASSERT_TRUE(profile.is_open()) << "Could not open profile file: " << profile_file_path;
```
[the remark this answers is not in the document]
adrianlizarraga: updated

tianleiwu (commented):
## Review

### 1. Run-Level Profiling Propagation ([onnxruntime/core/framework/utils.cc](https://github.com/microsoft/onnxruntime/blob/a7563a345ad96ebdd4dc2c806ee565700ad026ed/onnxruntime/core/framework/utils.cc#L881-L903))

**Positive**:
- The main execution path is wired correctly: top-level run profiling now creates EP profilers and feeds them through `ExecuteGraph`, so plugin EPs can participate in both session-level and run-level profiling without special-case code in `InferenceSession`.

**Concern**:
- ⚠️ **Run profilers stop at control-flow boundaries**: `ExecuteSubgraph` still has no `run_profiler` parameter and therefore never forwards the active run profiler into nested `If`/`Loop`/`Scan` subgraphs. The new test suite already documents the resulting mis-correlation: in run-profiling mode it knowingly accepts `"if_kernel"` as the parent instead of the nested `Mul` node ([onnxruntime/test/autoep/test_execution.cc](https://github.com/microsoft/onnxruntime/blob/a7563a345ad96ebdd4dc2c806ee565700ad026ed/onnxruntime/test/autoep/test_execution.cc#L1060-L1072)). For a CUDA plugin EP, that means GPU events from kernels launched inside control-flow subgraphs cannot be attributed back to the real ORT node during run profiling, which is exactly the scenario these new correlation hooks are meant to support.
  ```cpp
  common::Status ExecuteSubgraph(...,
                                 Stream* parent_stream,
                                 bool sync_subgraph_fetches,
                                 profiling::Profiler* run_profiler);

  auto retval = ExecuteGraphImpl(..., parent_stream, run_profiler);
  ```

### 2. Public Metadata Contract ([include/onnxruntime/core/session/onnxruntime_ep_c_api.h](https://github.com/microsoft/onnxruntime/blob/a7563a345ad96ebdd4dc2c806ee565700ad026ed/include/onnxruntime/core/session/onnxruntime_ep_c_api.h#L515-L619), [include/onnxruntime/core/session/onnxruntime_ep_c_api.h](https://github.com/microsoft/onnxruntime/blob/a7563a345ad96ebdd4dc2c806ee565700ad026ed/include/onnxruntime/core/session/onnxruntime_ep_c_api.h#L1739-L1753), [include/onnxruntime/core/common/gpu_profiler_common.h](https://github.com/microsoft/onnxruntime/blob/a7563a345ad96ebdd4dc2c806ee565700ad026ed/include/onnxruntime/core/common/gpu_profiler_common.h#L402-L421))

**Positive**:
- Using opaque `OrtProfilingEvent`/`OrtProfilingEventsContainer` types keeps the C ABI stable and avoids exposing ORT's internal `profiling::Events` representation to plugin libraries.

**Concern**:
- ⚠️ **The API does not clearly stabilize the ORT metadata a CUDA-style profiler needs for parity**: the built-in GPU profiler currently copies both `op_name` and `parent_name` from the correlated ORT event when merging GPU activities. The new plugin API exposes only `GetName()` plus string-key lookup into opaque event args. That means a CUDA plugin EP can implement the correlation flow, but only by depending on implicit arg names such as `"op_name"` that are not actually specified as a stable public contract. If we want plugin CUDA profiling to match the bundled CUDA EP reliably across ORT releases, we should either document the guaranteed fields on `NODE_EVENT`s or add explicit accessors/helper APIs for the common node metadata.

---

## Summary of Concerns

| 1 | High | Runtime integration | Run-level profiling is not propagated into control-flow subgraphs, so plugin EP events under `If`/`Loop`/`Scan` are mis-correlated or skipped. |
| 2 | Suggestion | Public API | The metadata contract for correlated ORT node events is underspecified for CUDA-profiler parity (`op_name`/related fields). |

---

## Verdict

REQUEST CHANGES — the public API is close to what a CUDA plugin EP needs for GPU event reporting, but the current implementation still breaks run-level profiling in control-flow subgraphs, and that should be fixed before relying on these hooks for real CUDA plugin profiling.

adrianlizarraga:
Hi @tianleiwu

>Concern: Run profilers stop at control-flow boundaries: 

This is an existing bug I uncovered while working on this PR. It has existed since [run profilers were introduced](https://github.com/microsoft/onnxruntime/pull/26846). I would not want to fix that here. I would prefer to fix that separately as it is orthogonal to this PR.

Update: here is a separate PR with the fix: https://github.com/microsoft/onnxruntime/pull/27870

>Concern:
>⚠️ The API does not clearly stabilize the ORT metadata a CUDA-style profiler needs for parity: the built-in GPU profiler currently copies both op_name and parent_name from the correlated ORT event when merging GPU activities. The new plugin API exposes only GetName() plus string-key lookup into opaque event args. That means a CUDA plugin EP can implement the correlation flow, but only by depending on implicit arg names such as "op_name" that are not actually specified as a stable public contract. If we want plugin CUDA profiling to match the bundled CUDA EP reliably across ORT releases, we should either document the guaranteed fields on NODE_EVENTs or add explicit accessors/helper APIs for the common node metadata.

We could specify that "op_name" is expected to be in the ORT event args. That said, I'm not sure it is super important to tag an EP event with the "op_name". That association can be made by setting a "parent_name" arg on the EP's event, and notably, there is a C API to get the name of an ORT event (`ProfilingEvent_GetName()`).

onnxruntime/test/autoep/library/example_plugin_ep_kernel_registry/ep_profiling.cc
adrianlizarraga: updated docs

onnxruntime/test/autoep/test_execution.cc
adrianlizarraga: updated. thanks

onnxruntime/test/autoep/library/example_plugin_ep_kernel_registry/ep_profiling.cc
adrianlizarraga: updated.

onnxruntime/test/autoep/library/example_plugin_ep_kernel_registry/ep_profiling.h
adrianlizarraga: updated

onnxruntime/test/autoep/library/example_plugin_ep_kernel_registry/ep_profiling.cc
adrianlizarraga: updated

onnxruntime/core/session/plugin_ep/ep_event_profiling.cc
adrianlizarraga: updated.

onnxruntime/core/session/plugin_ep/ep_api.cc
adrianlizarraga: updated all of these APIs to return OrtStatus. I also updated all functions in `OrtEpProfilerImpl` to return a status to allow the EP to propagate such API errors to ORT. ORT will log these errors, but will not end execution.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
adrianlizarraga: updated

include/onnxruntime/core/session/onnxruntime_cxx_api.h
```diff
@@ -1230,6 +1231,84 @@ struct EpAssignedSubgraphImpl : Ort::detail::Base<T> {
...
+      : ConstProfilingEventImpl<OrtProfilingEvent>{p} {}  ///< Take ownership
+
+  /// \brief Wraps OrtEpApi::CreateProfilingEvent
+  ProfilingEvent(OrtProfilingEventCategory category,
+                 int32_t process_id,
+                 int32_t thread_id,
+                 const char* event_name,
+                 int64_t timestamp_us,
+                 int64_t duration_us,
+                 const std::unordered_map<std::string, std::string>& args = {});
```
edgchen1, line 1288: would it be worth having an overload that takes `const char* const*` arg keys and values and a count to avoid copying from the map? is this potentially called in a hot path?

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -1741,6 +2066,31 @@ struct OrtEp {
...
+  /** \brief Return a new profiler for the execution provider.
+   *
+   * If the EP supports profiling, it should create and return an OrtEpProfilerImpl instance.
+   * ORT takes ownership of each non-NULL instance returned and will call OrtEpProfilerImpl::Release when
+   * it is no longer needed.
+   *
+   * ORT may call this function multiple times over the lifetime of a single OrtEp instance, for example
+   * during EP registration and again when run-level profiling is enabled. Each call is independent and
+   * the EP must treat it as a request for a new profiler instance (or NULL if profiling is not supported
+   * or disabled). ORT does not require that the same profiler instance be returned on each call.
```
edgchen1: suggested wording based on my understanding
```suggestion
   * ORT may call this function multiple times over the lifetime of a single OrtEp instance, for example
   * during EP registration and again per run if run-level profiling is enabled. Each call is independent and
   * the EP must return a new profiler instance (or NULL if profiling is not supported).
```

onnxruntime/test/autoep/library/example_plugin_ep_kernel_registry/ep_profiling.cc
```diff
@@ -0,0 +1,230 @@
...
+    int64_t ep_elapsed_ns = std::chrono::duration_cast<std::chrono::nanoseconds>(
+                                raw_ep_event.start_time - self->ep_profiling_start_time_point_)
+                                .count();
+    int64_t rel_ts_us = (self->ep_profiling_start_offset_ns_ + ep_elapsed_ns) / 1000;
+
+    int64_t dur_us = std::chrono::duration_cast<std::chrono::microseconds>(
+                         raw_ep_event.end_time - raw_ep_event.start_time)
+                         .count();
+
+    // Set parent_name an event arg. The parent_name is just the name of the correlated ORT event.
```
edgchen1:
```suggestion
    // Set parent_name as an event arg. The parent_name is just the name of the correlated ORT event.
```

onnxruntime/test/autoep/library/example_plugin_ep_kernel_registry/ep_profiling.cc
edgchen1: regarding thread_id, there is a thread_id field in the profiling event structure. a related question is, how should an EP set that?

onnxruntime/core/session/plugin_ep/ep_event_profiling.cc
adrianlizarraga: Good point. I added a note that the epoch is platform dependent.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -423,6 +426,197 @@ struct OrtExternalResourceImporterImpl {
...
+   *
+   * An EP profiler converts its events to OrtProfilingEvent instances and adds them into the provided
+   * OrtProfilingEventsContainer container. Call OrtEpApi::CreateProfilingEvent to create a new
+   * OrtProfilingEvent instance. Then call OrtEpApi::ProfilingEventsContainer_AddEvents to add one or more
+   * events to the container.
+   *
+   * After this function returns, ORT appends the EP's events to the profiling timeline.
+   *
+   * \param[in] this_ptr The OrtEpProfilerImpl instance.
+   * \param[in] ep_profiling_end_offset_ns The elapsed time in nanoseconds (using ORT's profiling clock) between
```
edgchen1: so this is different from `EpProfiler::EndProfiling()` in that it now gets the end time instead of the start time. admittedly, I'm not sure why the start time was useful to `EndProfiling()` in the first place (although the CUDA EP profiler seems to use it to match the corresponding `StartProfiling()` call).

so I guess my question is, why is a time parameter useful here, and why is the profiling end time preferred over the profiling start time?

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -423,6 +426,197 @@ struct OrtExternalResourceImporterImpl {
...
+   * elapsed time since this call. The formula is:
+   *
+   *   event_timestamp_us = (ep_profiling_start_offset_ns + (ep_event_time_ns - ep_profiling_start_time_ns)) / 1000
+   *
+   * where `ep_event_time_ns` and `ep_profiling_start_time_ns` are measured using the EP's own clock.
+   *
+   * \param[in] this_ptr Pointer to the OrtEpProfilerImpl instance.
+   * \param[in] ep_profiling_start_offset_ns The elapsed time in nanoseconds (using ORT's profiling clock) between
+   *                                         ORT's profiling start and this call to StartProfiling.
+   * \param[out] success_out Output parameter set to true if profiling was successfully started, or false otherwise.
```
edgchen1: why do we need `success_out` if we already return a status?

onnxruntime/test/autoep/library/example_plugin_ep_kernel_registry/ep_profiling.cc
adrianlizarraga: EPs like CUDA and ROCM (deprecated) use libraries that actually do the tracking of thread IDs themselves. This example EP does it manually.

For example, CUDA ep uses CUPTI to track which "external correlation ids" (ORT events) are correlated with cuda events. ﻿ From what I understand, CUPTI's external correlation is entirely thread-local on the CPU side:

 1. `cuptiActivityPushExternalCorrelationId(ort_event_correlation_id)` pushes onto the current CPU thread's correlation stack
 2. Any CUDA API call (e.g., cudaLaunchKernel) made on that thread automatically gets tagged with the top-of-stack correlation ID
 3. When the GPU kernel completes, CUPTI's activity record contains that correlation ID
 4. ORT's GPUTracerManager matches it back to the original ORT event

So, cupti automatically tags everything launched on that CPU thread while the ID is on the stack. The implicit assumption is that between PushCorrelation and PopCorrelation, the CUDA calls on that thread correspond to that ORT event.

include/onnxruntime/core/session/onnxruntime_cxx_api.h
adrianlizarraga: Yeah, this is potentially in a hot path. I'll add.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
adrianlizarraga: I was a bit torn about this too. This approach allows saying that an EP profiler can opt out of profiling without necessarily returning an error. However, it looks like [ORT actually ignores](https://github.com/microsoft/onnxruntime/blob/1c05b13ddef3f40fcca7781d437029848d3b2fdb/onnxruntime/core/common/profiler.cc#L52) the return boolean return value when calling `StartProfiling` anyway (we may need to fix that separately?).

I'll remove the `success_out` param for simplicity.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
adrianlizarraga: I don't know why the start time was passed to `EpProfiler::EndProfiling()` originally. From looking at all the existing ep profiler implementations, I am pretty sure it is not strictly necessary since it could have been saved from the first call to `EpProfiler::StartProfiler()`.

This new `ep_profiling_end_offset_ns` is also not strictly necessary, but at least it is not "duplicate" information that the ep profiler already received before. Perhaps this could be useful so that the EP profiler knows how long ORT _thinks_ the profiling lasted. If any of the EP's events end times exceed this value, then an EP could potentially know about it ... it's a bit of a stretch.

Would you prefer to remove it? I don't think it necessarily does any harm.

edgchen1: if it's the case that `EpProfiler::StartProfiling()` will have a corresponding call to `EpProfiler::EndProfiling()` without any intermediate (i.e. nested or interleaved) call to `StartProfiling()`, then I think we don't need provide the start timestamp to `EndProfiling()`. this seems to be the case, unless I'm missing something.

as far as the end timestamp for `EndProfiling()`, I'm not seeing a compelling reason to provide it since it could be derived by the EP profiler based on the start time it received in `StartProfiling()` (if it's just `now - start time`). so I'd slightly prefer to remove it.

include/onnxruntime/core/session/onnxruntime_cxx_api.h
adrianlizarraga: added

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
adrianlizarraga: removed parameter

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
adrianlizarraga: removed parameter

include/onnxruntime/core/common/profiler_common.h
```diff
@@ -79,10 +81,70 @@ using Events = std::vector<EventRecord>;
...
+  /// EP profiler implementations may use the calls to Start and Stop to maintain a stack of ORT event IDs
+  /// that can be correlated with EP events (e.g., GPU kernel events).
+  ///
+  /// A relative ORT event ID is computed as a timestamp offset relative to the profiling start time:
+  ///     relative_ort_event_id =
+  ///         std::chrono::duration_cast<std::chrono::microseconds>(event_start_time - profiling_start_time).count();
+  ///
+  /// Because relative ORT event IDs are relative to profiling start, different profiling sessions may reuse the same
+  /// values. If the EP's profiling utilities (e.g., CUPTI or ROCTracer) require correlation IDs that are unique
+  /// across concurrent profiling sessions, then the EP profiler should compute an absolute correlation ID:
```
edgchen1: just to confirm my understanding - is it only possible to have concurrent profiling sessions with multiple `EpProfiler` instances?

onnxruntime/core/session/plugin_ep/ep_api.cc
```diff
@@ -806,6 +808,146 @@ ORT_API_STATUS_IMPL(GetEnvConfigEntries, _Outptr_ OrtKeyValuePairs** config_entr
...
+  ORT_API_RETURN_IF(events == nullptr || num_events == 0, ORT_INVALID_ARGUMENT,
+                    "Must provide at least one event to add to OrtProfilingEventsContainer.");
+
+  auto& all_events = events_container->events;
+  all_events.reserve(all_events.size() + num_events);
+
+  for (size_t i = 0; i < num_events; ++i) {
+    ORT_API_RETURN_IF(events[i] == nullptr, ORT_INVALID_ARGUMENT,
+                      "OrtProfilingEvent instance at index ", i, " is NULL");
+    all_events.push_back(*FromOpaqueProfilingEvent(events[i]));
```
edgchen1, line 1194: nit: if we do error out because an event is `nullptr`, it would leave `events_container` partially modified. perhaps either ignore `nullptr`s or check all of the events first before adding them.

onnxruntime/core/session/plugin_ep/ep_api.cc
```diff
@@ -806,6 +808,146 @@ ORT_API_STATUS_IMPL(GetEnvConfigEntries, _Outptr_ OrtKeyValuePairs** config_entr
...
+                      "`arg_keys` and `arg_values` must be non-null when `num_args` > 0");
+
+    args.reserve(num_args);
+
+    for (size_t i = 0; i < num_args; ++i) {
+      const char* key = arg_keys[i];
+      const char* value = arg_values[i];
+      ORT_API_RETURN_IF(key == nullptr, ORT_INVALID_ARGUMENT, "Arg key at index ", i, " is NULL");
+      ORT_API_RETURN_IF(value == nullptr, ORT_INVALID_ARGUMENT, "Arg value at index ", i, " is NULL");
+      args[key] = value;
```
edgchen1: minor nit: we can avoid the mapped type default construction from `map::operator[]`

```suggestion
      args.emplace(key, value);
```

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -423,6 +426,192 @@ struct OrtExternalResourceImporterImpl {
...
+   *   OrtEpProfilerImpl::StopEvent(x) ->  EP ort event stack: [ ]    <- top of stack
+   *
+   * Tagging EP events with the ORT event correlation ID enables the EP to annotate its own events
+   * with metadata from the parent ORT event (e.g., operator name).
+   *
+   * \note The ORT event correlation ID is an absolute, epoch-based timestamp in microseconds. It is computed
+   *       from the ORT event's start time using std::chrono::high_resolution_clock (platform-defined epoch).
+   *       Because it is absolute rather than relative to profiling start, it is unique across concurrent profiling
+   *       sessions within the same process and can be used directly as a correlation ID for EP profiling
+   *       utilities (e.g., CUPTI or ROCTracer).
```
edgchen1: perhaps this is a somewhat pedantic concern - I think an absolute timestamp does not guarantee uniqueness in the presence of parallelism (or the ability to change the system clock, which `std::chrono::high_resolution_clock` could be using). if we want a uniqueness guarantee, we probably need to generate IDs some other way. otherwise, practically speaking, they may be unique enough.

onnxruntime/test/autoep/library/example_plugin_ep_kernel_registry/ep_profiling.cc
edgchen1: good to know. perhaps it's a good time to document any implicit assumptions.

so, based on my understanding, can we assume that
- one event's StartEvent/StopEvent calls will be made on the same thread?
- for a kernel event, this thread will be the same one that calls the kernel compute function?

include/onnxruntime/core/common/profiler_common.h
adrianlizarraga: The term "session" is a bit overloaded here. However, if we define "profiling session" to be everything that gets captured between a call to `EpProfiler::StartProfiling()` and `EpProfiler::EndProfiling()` for a single `EpProfiler` instance, then yes.

With "run profiling" there will be one `EpProfiler` instance **per** call of `session.run()` (potentially many profilers)

With "session profiling", there is a single `EpProfiler` instance that handles one more calls to `session.run()`.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
adrianlizarraga: > or the ability to change the system clock, which std::chrono::high_resolution_clock could be using

Right, it is a bit of a shame imo that we've been using `std::chrono::high_resolution_clock`, which could be aliased to the system clock. It would be preferable to use a steady monotonic clock. Though, `std::chrono::high_resolution_clock` does seem to alias to a steady clock on most platforms (but not all) ...

Beside the system clock issue, it is practically unique within the same process (collision if parallel events happen in the same microsecond). I'm not so sure what to do about this now given that this has been the approach for years: https://github.com/microsoft/onnxruntime/blob/358628a8ce10bebdc52c236607a69ff741f67cb9/include/onnxruntime/core/common/gpu_profiler_common.h#L173-L193

Perhaps, can adjust doc to state that collisions could happen?

edgchen1: yeah, I think clarifying in the docs that it is "practically unique" instead of calling it "unique" would be good.

include/onnxruntime/core/common/profiler_common.h
edgchen1: > However, if we define "profiling session" to be everything that gets captured between a call to EpProfiler::StartProfiling() and EpProfiler::EndProfiling() for a single EpProfiler instance, then yes.

that's what I wanted to confirm. thanks.

onnxruntime/test/autoep/library/example_plugin_ep_kernel_registry/ep_profiling.cc
adrianlizarraga: Yep, I added a doc note to `StartEvent` and `StopEvent`:

>\note **Threading:** For a given ORT event, StartEvent, the kernel's Compute() entry point, and StopEvent
      are all invoked on the same CPU thread (Compute() may asynchronously launch device work).
      Different ORT events may run on different threads (e.g., inter-op parallelism), so correlation
      stacks (e.g., for CUPTI) should use thread-local storage.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
adrianlizarraga: I updated the doc note to state the collisions can happen:

>\note The ORT event correlation ID is an absolute, epoch-based timestamp in microseconds. It is computed
    from the ORT event's start time using std::chrono::high_resolution_clock (platform-defined epoch).
    Because it is absolute rather than relative to profiling start, it is practically unique across
    concurrent profiling sessions within the same process (collisions require sub-microsecond event
    concurrency) and can be used directly as a correlation ID for EP profiling utilities
    (e.g., CUPTI or ROCTracer).

onnxruntime/core/session/plugin_ep/ep_api.cc
adrianlizarraga: updated. thanks

onnxruntime/core/session/plugin_ep/ep_api.cc
adrianlizarraga: add an initial check loop. thanks

edgchen1 (approved):

tianleiwu (approved):
Approved. The overall API shape looks solid.

PR #27657, opened 2026-03-14
Fix Python lint issues in onnxruntime_test_python.py

edenfunf:
## What

Fix two ruff lint errors in `onnxruntime/test/python/onnxruntime_test_python.py` introduced alongside new tests for the `ORT_LOGGING_LEVEL` environment variable (see #27645):

- Move `import subprocess` from inside a test method body to the module-level stdlib import block
- Add explicit `check=False` to `subprocess.run()`

Also includes the two new tests themselves:

- `test_get_default_logger_severity` – verifies the new `get_default_logger_severity()` API (added in #27645) returns a value in `[0, 4]`
- `test_ort_logging_level_env_var` – spawns a subprocess for each valid level (0–4) with `ORT_LOGGING_LEVEL` set and confirms the reported severity matches

## Why

`PLC0415` (`import` inside function body) and `PLW1510` (`subprocess.run` without explicit `check` argument) are enforced by the project's ruff configuration and would fail CI lint checks.

## How

- `import subprocess` relocated to the module-level stdlib imports section (alphabetically correct position)
- `check=False` added to `subprocess.run()` call to satisfy `PLW1510`; the return code is already checked manually via `self.assertEqual(result.returncode, 0, ...)`

## Files Modified

- `onnxruntime/test/python/onnxruntime_test_python.py`

## Lint Status

| Check | Before | After |
|-------|--------|-------|
| ruff (this file) | 2 errors | 2 errors (pre-existing, unrelated to this PR) |
| `PLC0415` | present | fixed |
| `PLW1510` | present | fixed |

The 2 remaining errors (`I001` import ordering, `FURB171` single-item membership test) exist in `main` and are out of scope.

## Note

The new tests depend on `ort.get_default_logger_severity()` introduced in #27645. They are included here as the tests are the reason this file was touched at all, and keeping the lint fixes alongside the tests avoids a dangling import on `main`.

### DIFF
--- a/onnxruntime/test/python/onnxruntime_test_python.py
+++ b/onnxruntime/test/python/onnxruntime_test_python.py
@@ -9,6 +9,7 @@ import os
 import pathlib
 import platform
 import queue
+import subprocess
 import sys
 import threading
 import unittest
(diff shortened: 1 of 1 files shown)


tianleiwu:
You shall commit this change to the branch of https://github.com/microsoft/onnxruntime/pull/27645.

Issue #27487, opened 2026-02-27
[Feature Request] Universal cache invalidation


henryruhs:
### Describe the feature request

From my research only TensorRT has proper cache invalidation, once onnxruntime has been updated the outdated caches are causing onnxruntime to crash.

| EP | Auto-invalidation | Version Detection |
|---|---|---|
| TensorRT (ORT 1.20+) | Yes | Runtime |
| MIGraphX | Partial | Compile-time |
| CoreML | No | None |

For now I use this workaround.

```python
import os
import onnxruntime

cache_path = os.path.join('.caches', onnxruntime.get_version_string())

session = onnxruntime.InferenceSession('model.onnx', providers = [('CoreMLExecutionProvider',
{
	'ModelCacheDirectory': cache_path
})])
```

### Describe scenario use case

I recommend a universal solution to handle caches, it should not be EP specific.

yiliangbetter:
I created this PR trying to address it https://github.com/microsoft/onnxruntime/pull/27660, please take a look at it

henryruhs:
@yiliangbetter thanks so much, I will comment on the PR

PR #27660, opened 2026-03-15
Add EP cache versioning to avoid crashes from outdated caches after ORT updates

yiliangbetter:
### Description

This PR adds support for **execution provider (EP) cache versioning** so that cache directories are tied to the ONNX Runtime version. When enabled, known EP cache paths are automatically suffixed with the ORT version (e.g. `.caches` → `.caches/1.20.0`). That invalidates old caches when ORT is upgraded and reduces crashes or bad behavior from loading incompatible cached artifacts.

**Changes:**

- **Session option:** `session.ep_cache_use_ort_version` (C++: `kOrtSessionOptionsEpCacheUseOrtVersion`). When set to `"1"`, known EP cache directory options are suffixed with the ORT version. Set this **before** appending execution providers.
- **Supported cache options:** CoreML `ModelCacheDirectory`; TensorRT `trt_engine_cache_path` and `trt_timing_cache_path`; MIGraphX `migraphx_model_cache_dir`; NvTensorRtRtx `nv_runtime_cache_path`.
- **Python helper:** `onnxruntime.get_versioned_ep_cache_path(base_path)` for building versioned cache paths manually (e.g. for provider options).
- **Core logic:** New `ep_cache_versioning.cc` / `.h` apply version suffixes using the `ORT_VERSION` macro; wired into `SessionOptionsAppendExecutionProvider` and `InitializeSession` so both config options and provider options get versioned paths when the flag is on.
- **Docs:** `InferenceSession` docstring updated with a short “EP cache versioning” section and usage notes.
- **Tests:** Unit tests in `test_session_options.cc` for config and provider path versioning (on/off, multiple EPs, empty paths, unknown providers/options, case insensitivity, no mutation of inputs).

### Motivation and Context

Many execution providers (CoreML, TensorRT, MIGraphX, etc.) compile and cache artifacts to speed up session creation. Those caches are tied to the ORT build; after an upgrade, loading them can cause crashes, wrong results, or ABI issues in detail please see (https://github.com/Microsoft/onnxruntime/issues/27487)

This change:

1. **Reduces crashes** by storing caches in versioned directories so old caches are not loaded after an upgrade.
2. **Makes upgrades safer** by keeping cache layout under ORT’s control when the option is enabled.
3. **Keeps flexibility** by offering both a global session option and a Python helper for manual path construction.

### How to use

**Automatic (session option):**

```python
sess_options = ort.SessionOptions()
sess_options.add_session_config_entry("session.ep_cache_use_ort_version", "1")
session = ort.InferenceSession("model.onnx", sess_options, providers=[...])

### DIFF
--- a/onnxruntime/core/session/ep_cache_versioning.cc
+++ b/onnxruntime/core/session/ep_cache_versioning.cc
@@ -24,8 +24,6 @@ namespace onnxruntime {
 namespace {
 
 // Registry of EP cache path option keys, keyed by provider name (lowercased).
-// This allows EP-specific code to register which options are cache directories,
-// so the generic versioning logic does not need hardcoded knowledge of each EP.
 std::unordered_map<std::string, EpCachePathOptionKeys>& EpCachePathOptionsRegistry() {
   // Function-local static to avoid static initialization order issues across translation units.
   static std::unordered_map<std::string, EpCachePathOptionKeys> registry;
@@ -89,7 +87,9 @@ static const EpCachePathOptionKeys* TryGetEpCachePathOptionKeys(const std::strin
 }
 
 void ApplyEpCacheVersionToConfigOptions(ConfigOptions& config_options) {
-  if (config_options.GetConfigOrDefault(kOrtSessionOptionsEpCacheUseOrtVersion, "0") != "1") {
+  const std::string mode = config_options.GetConfigOrDefault(kOrtSessionOptionsEpCacheUseOrtVersion, "0");
+  const bool ep_cache_versioning_enabled = (mode == "1");
+  if (!ep_cache_versioning_enabled) {
     return;
   }
 #ifdef ORT_VERSION
@@ -100,10 +100,12 @@ void ApplyEpCacheVersionToConfigOptions(ConfigOptions& config_options) {
     if (key.size() <= prefix.size() || key.compare(0, prefix.size(), prefix) != 0) {
       continue;
     }
-    size_t dot = key.find('.', prefix.size());
-    if (dot == std::string::npos) continue;
-    std::string provider_lower = GetLowercaseString(key.substr(prefix.size(), dot - prefix.size()));
-    std::string option_key = key.substr(dot + 1);
+    const size_t dot = key.find('.', prefix.size());
+    if (dot == std::string::npos) {
+      continue;
+    }
+    const std::string provider_lower = GetLowercaseString(key.substr(prefix.size(), dot - prefix.size()));
+    const std::string option_key = key.substr(dot + 1);
     const EpCachePathOptionKeys* cache_keys = TryGetEpCachePathOptionKeys(provider_lower);
     if (cache_keys == nullptr) {
       continue;
@@ -112,7 +114,9 @@ void ApplyEpCacheVersionToConfigOptions(ConfigOptions& config_options) {
       continue;
     }
     const std::string& value = kv.second;
-    if (value.empty()) continue;
+    if (value.empty()) {
+      continue;
+    }
     to_update.push_back({key, GetVersionedCachePath(value)});
   }
   for (const auto& p : to_update) {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by henryruhs** [comment]
@yiliangbetter Disclaimer: I never wrote a single line of C++ so my review is eventual bias.

1. Did you consolidate the existing caching logic for TensorRT - from what I know there is already something in the codebase?

2. How I read the code changes, the caching module has knowledge about the EPs to deal with. In my opinion every EP related logic has to live in the EP code themselve and paths for caching or even the need for caching should be exposed via an hook. That means the caching module should be "dumb" and iterates over loaded EPs rather than having hard coded values like this.

```
const std::array<std::pair<std::string, std::string>, 5> kEpCachePathOptions = {
    std::pair<std::string, std::string>{"coreml", "ModelCacheDirectory"},
    std::pair<std::string, std::string>{"tensorrt", "trt_engine_cache_path"},
    std::pair<std::string, std::string>{"tensorrt", "trt_timing_cache_path"},
    std::pair<std::string, std::string>{"migraphx", "migraphx_model_cache_dir"},
    std::pair<std::string, std::string>{"nvtensorrtrtx", "nv_runtime_cache_path"},
};
```

yiliangbetter:
> @yiliangbetter Disclaimer: I never wrote a single line of C++ so my review is eventual bias.
> 
> 1. Did you consolidate the existing caching logic for TensorRT - from what I know there is already something in the codebase?
> 2. How I read the code changes, the caching module has knowledge about the EPs to deal with. In my opinion every EP related logic has to live in the EP code themselve and paths for caching or even the need for caching should be exposed via an hook. That means the caching module should be "dumb" and iterates over loaded EPs rather than having hard coded values like this.
> 
> ```
> const std::array<std::pair<std::string, std::string>, 5> kEpCachePathOptions = {
>     std::pair<std::string, std::string>{"coreml", "ModelCacheDirectory"},
>     std::pair<std::string, std::string>{"tensorrt", "trt_engine_cache_path"},
>     std::pair<std::string, std::string>{"tensorrt", "trt_timing_cache_path"},
>     std::pair<std::string, std::string>{"migraphx", "migraphx_model_cache_dir"},
>     std::pair<std::string, std::string>{"nvtensorrtrtx", "nv_runtime_cache_path"},
> };
> ```

Thanks for the feedback — I made two tweaks based on your comments:

TensorRT scope / “consolidation” TensorRT’s cache behavior (option parsing + validation + semantics) still lives in the TensorRT EP. This PR doesn’t change the TensorRT V2 C API (OrtTensorRTProviderOptionsV2) path. session.ep_cache_use_ort_version is only applied when EPs are added via the generic SessionOptionsAppendExecutionProvider(..., ProviderOptions) flow or via config options. For callers using the TensorRT V2 API, the options are: use the generic provider-options map, or pass a pre-versioned cache dir (e.g. Python onnxruntime.get_versioned_ep_cache_path()).

Keeping core “dumb” without spreading registration everywhere The versioning code no longer hardcodes EP names/keys

Why not just put registration inside each EP (TensorRT/MIGraphX/etc.)? For EPs that are built/loaded as shared libraries, having them “self-register” would either:

register into a different copy of global state (if the registry lives in the main binary), or
force a new host callback / ABI so the EP can register back into the main runtime at load time.
That’s a bigger cross-cutting change than this PR. Keeping registration near the EP but executed in the main binary (e.g. bridge/creator code) avoids that ABI work while still keeping the registry logic generic.

henryruhs:
Good job, I hope the onnxruntime team can take over from here.

I found a couple of general coding style issues were you can improve - reach out if you want to know.

yiliangbetter:
> Good job, I hope the onnxruntime team can take over from here.
> 
> I found a couple of general coding style issues were you can improve - reach out if you want to know.

sure, I would very much like to know it.

henryruhs:
1. Avoid unnecessary else branches when a method already returns
1. Group const declarations at the top of methods for cleaner flow
1. Prefer positive conditions first, guard clauses last
1. Avoid != in conditionals or else where an exact match is possible
1. Make sure method names accurately reflect what they do
1. Reduce comments - good naming make comments redundant

PR #27665, opened 2026-03-15
Fix tf.function retracing in TensorFlow benchmark

### LINKED ISSUE #14953: Some probable performance issues in onnxruntime/python/tools/transformers/benchmark.py
Hello! Our static bug checker has found a performance issue in onnxruntime/python/tools/transformers/benchmark.py:  `run_with_tf_optimizations` [(1)](https://github.com/microsoft/onnxruntime/blob/2ef25a22007f894e811553ab954b588b93897d56/onnxruntime/python/tools/transformers/benchmark.py#L506)[(2)](https://github.com/microsoft/onnxruntime/blob/2ef25a22007f894e811553ab954b588b93897d56/onnxruntime/python/tools/transformers/benchmark.py#L510)[(3)](https://github.com/microsoft/onnxruntime/blob/2ef25a22007f894e811553ab954b588b93897d56/onnxruntime/python/tools/transformers/benchmark.py#L514) is repeatedly called in a for loop, but there are tf.function decorated functions [`run_in_eager_mode`](https://github.com/microsoft/onnxruntime/blob/2ef25a22007f894e811553ab954b588b93897d56/onnxruntime/python/tools/transformers/benchmark.py#L411) and [`run_in_graph_mode`](https://github.com/microsoft/onnxruntime/blob/2ef25a22007f894e811553ab954b588b93897d56/onnxruntime/python/tools/transformers/benchmark.py#L416) defined and called in  [`run_with_tf_optimizations`](https://github.com/microsoft/onnxruntime/blob/2ef25a22007f894e811553ab954b588b93897d56/onnxruntime/python/tools/transformers/benchmark.py#L404). 

In that case, when `run_with_optimizations` is called in a loop, the functions `run_in_eager_mode` and `run_in_graph_mode` will create a new graph every time, and that can trigger tf.function retracing warning.

Here is [the tensorflow document](https://tensorflow.google.cn/guide/function#tracing) to support it.

Briefly, for better efficiency, it's better to use:
```python
@tf.function
def inner():
    pass

def outer():
    inner()  
```
than:
```python
def outer():
    @tf.function
    def inner():
        pass
    inner()
```

Looking forward to your reply.

Rishi-Dave:
## Summary

- Move `tf.function`-decorated forward functions out of the inner benchmark loop to prevent unnecessary graph retracing on every `(batch_size, sequence_length)` iteration
- Update deprecated `experimental_compile` to `jit_compile` (available since TF 2.4)
- Hoist `import random` out of the inner loop

Fixes #14953

## Motivation

When `run_with_tf_optimizations` is used as a decorator inside the innermost `(batch_size, sequence_length)` loop, each iteration creates a new Python function object. Since `tf.function` keys its trace cache on function identity, a new object means a forced retrace every iteration — the cached graph is never reused. This defeats the purpose of `tf.function` and adds significant overhead from repeated graph construction and optimization passes.

The [TensorFlow documentation on tracing](https://www.tensorflow.org/guide/function#rules_of_tracing) explicitly warns against defining `tf.function`-decorated functions inside loops.

## Changes

**`onnxruntime/python/tools/transformers/benchmark.py`** (1 file, ~35 insertions / ~31 deletions):

1. **Hoisted forward function definitions** (`encoder_forward`, `encoder_decoder_forward`, `lxmert_forward`) from the inner `batch_size × sequence_length` loop to the per-model scope. They are now defined once per model, and the `@run_with_tf_optimizations` decorator (which applies `@tf.function`) is only invoked once per model.

2. **Changed forward functions to accept `input_ids` as a parameter** instead of closing over the loop variable. This lets `tf.function` trace based on the tensor's `(dtype, shape)` spec and reuse cached concrete functions when shapes repeat across iterations.

3. **Updated `experimental_compile=use_xla`** to **`jit_compile=use_xla`**. The `experimental_compile` parameter was deprecated in TF 2.4 (Dec 2020) and removed in TF 2.12.

4. **Moved `import random`** from the innermost loop body to before the outer model loop — the module only needs to be imported once.

5. **Moved inference function selection** (`if config.is_encoder_decoder ... elif isinstance(config, LxmertConfig) ...`) outside the batch/sequence loops since it depends only on the model config, not on batch size or sequence length. The original priority order (`is_encoder_decoder` checked before `LxmertConfig`) is preserved.

## Test Plan

- [x] `lintrunner -a` passes cleanly (no RUFF or RUFF-FORMAT violations)
- [x] `python -m py_compile benchmark.py` — syntax verified
- [x] Change is purely structural — function behavior (inputs, outputs, control flow) is identical
- [ ] Manual verification with TensorFlow installed (TF is an optional dependency not present in the standard CI matrix; this code path is exercised via `python benchmark.py -e tensorflow`)

### DIFF
--- a/onnxruntime/python/tools/transformers/benchmark.py
+++ b/onnxruntime/python/tools/transformers/benchmark.py
@@ -43,6 +43,7 @@ It is recommended to use run_benchmark.sh to launch benchmark.
 import argparse
 import logging
 import os
+import random
 import timeit
 from datetime import datetime
 
@@ -483,8 +484,6 @@ def run_tensorflow(
 
     if precision == Precision.FLOAT16 or precision == Precision.INT8:
         raise NotImplementedError("Mixed precision is currently not supported.")
-
-    import random  # noqa: PLC0415
 
     for model_name in model_names:
         config = AutoConfig.from_pretrained(model_name, cache_dir=cache_dir)


onnxruntime/python/tools/transformers/benchmark.py
```diff
@@ -484,6 +484,8 @@ def run_tensorflow(
     if precision == Precision.FLOAT16 or precision == Precision.INT8:
         raise NotImplementedError("Mixed precision is currently not supported.")
 
+    import random  # noqa: PLC0415
```
tianleiwu: nit: this can be moved to the beginning of this file.

tianleiwu (approved):
## Review

### 1. Forward-Function Hoisting (`benchmark.py` — new lines 505-531)

**Positive**:
- Hoisting `encoder_forward`, `encoder_decoder_forward`, and `lxmert_forward` out of the inner loop is the right fix. The original code created new Python function objects every iteration, invalidating `tf.function`'s trace cache — this completely defeated graph caching. The new placement (once per model) is the correct scope.
- Accepting `input_ids` as a parameter instead of closing over the loop variable is the key enabler: `tf.function` now caches by `(dtype, shape)` spec, so repeated calls with the same shape (e.g., within `timeit.repeat`) reuse the cached concrete function.
- Moving the `if config.is_encoder_decoder … elif isinstance(config, LxmertConfig)` dispatch outside the batch/sequence loops is correct — the model config is invariant within those loops.
- The `# noqa: B023` suppressions on `model` and `config` closures are appropriate: these variables are set once per `model_name` iteration and are consumed within the same iteration.

**Concern**:
- ⚠️ **`lxmert_forward` hard-codes batch dimension to 1**: The random visual features are created as `tf.random.normal([1, 1, config.visual_feat_dim])`. This is not a regression — the original code had the identical shape — but it means Lxmert benchmarking ignores the `batch_size` variable for visual inputs. Consider filing a follow-up issue if Lxmert benchmarking fidelity matters.

### 2. Deprecation Fix (`benchmark.py` — line 434)

**Positive**:
- Replacing `experimental_compile=use_xla` with `jit_compile=use_xla` is correct. `experimental_compile` was deprecated in TF 2.4 (Dec 2020) and removed in TF 2.12. This unblocks use with modern TF versions.

### 3. `import random` Hoist (`benchmark.py` — line 487)

**Positive**:
- Moving `import random` from the innermost loop to the function body is a minor but correct cleanup. Python caches module imports, so the performance gain is negligible, but it's cleaner and eliminates a linter-flagged pattern.

### 4. `timeit.repeat` Lambda (`benchmark.py` — line 548)

**Positive**:
- `lambda: inference(input_ids)` correctly passes the tensor explicitly, consistent with the new function signatures.

**Concern**:
- ⚠️ **Lambda still closes over `input_ids`**: The `lambda` in `timeit.repeat(lambda: inference(input_ids), …)` captures `input_ids` via closure. This is fine because `input_ids` doesn't change during the `timeit.repeat` calls. However, the existing `# noqa: B023` suppression remains necessary for this reason. No action needed — just noting for reviewer awareness.

---

## Summary of Concerns

| 1 | Nitpick | lxmert_forward | Visual feature batch dimension hard-coded to 1 (pre-existing, not a regression) |
| 2 | Nitpick | timeit lambda | Closure over `input_ids` is safe but warrants the existing B023 suppression |

---

## Verdict

**APPROVE** — Correct, well-scoped fix for a real performance issue. No regressions, excellent PR description, and the code is cleaner after the change.

Rishi-Dave:
Thanks for the thorough review @tianleiwu!

Addressed your nit — moved `import random` to the top-level imports at the beginning of the file.

Re: the Lxmert visual feature batch dimension being hard-coded to 1 — agreed that's pre-existing and orthogonal to this change. Happy to file a follow-up issue for it if you think it's worth fixing.

Rishi-Dave:
Rebased on latest upstream/main. All changes intact — the `import random` hoist to top-level imports and the tf.function retracing fix are both in place. Ready for merge whenever CI is green.

Rishi-Dave:
Rebased on latest upstream/main — no conflicts. Both commits intact:

1. **tf.function retracing fix** — forward functions hoisted outside inner loops, `input_ids` passed as argument, `experimental_compile` → `jit_compile`
2. **`import random` moved to top-level imports** (per review nit)

Ruff check + format passing locally. Ready for re-approval + merge.

PR #27666, opened 2026-03-15
Plugin EP: Fix bug that incorrectly assigned duplicate MetDef IDs to fused nodes in different GraphViews

adrianlizarraga:
### Description
 Fixes a bug where `PluginExecutionProvider::GetCapability()` incorrectly assigned duplicate MetaDef IDs to fused nodes that live in different GraphViewer instances (e.g., the then/else branches of an If node).

The root cause was that `GetCapability()` created a new `ModelMetadefIdGenerator` on every invocation. Since the graph partitioner calls `GetCapability()` once per subgraph, the generator's monotonic counter reset each time, producing colliding IDs across subgraphs. This caused session creation to fail with:

> Failed to add kernel for example_ep_9433721956998717990_0 example_ep example_ep: Conflicting with a registered kernel with op versions. the since version is: 1

#### Fix
   - Promoted `ModelMetadefIdGenerator` to an instance member of `PluginExecutionProvider` so the same generator is reused across all `GetCapability()` calls, ensuring unique MetaDef IDs.
     - This is also consistent with how existing provider-bridge EPs create and use a single generator instance.
     - **Bonus perf improvement**: No longer recomputes the entire model's hash on every call to `GetCapability()`.

#### Testing

Example EP changes:
   - Refactored `SaveConstantInitializers()` → `TrySaveConstantInitializer()` to save initializers per-node-input instead of via `graph.GetInitializers()`, which doesn't return initializers defined in parent or sibling subgraphs.
   - Extracted `CopiesConstantInitializers()` helper to deduplicate the condition for drop_constant_initializers.

Unit testing:
- Added unit test called `CompilingPluginEp_MultiSubgraphs_DuplicateMetaDefIdBug` — runs an If model with Mul nodes in both branches, verifying that both fused nodes receive unique MetaDef IDs and the session creates/runs successfully.

Credit to @apwojcik for [finding the bug.](https://github.com/microsoft/onnxruntime/pull/27608)

### DIFF
--- a/onnxruntime/test/autoep/library/example_plugin_ep/ep.cc
+++ b/onnxruntime/test/autoep/library/example_plugin_ep/ep.cc
@@ -196,44 +196,33 @@ const char* ORT_API_CALL ExampleEp ::GetNameImpl(const OrtEp* this_ptr) noexcept
   return ep->name_.c_str();
 }
 
-OrtStatus* ExampleEp::SaveConstantInitializers(const OrtGraph* ort_graph) {
-  Ort::ConstGraph graph{ort_graph};
+bool ExampleEp::CopiesConstantInitializers() const {
+  return !(config_.enable_ep_context && config_.enable_weightless_ep_context_nodes);
+}
 
-  try {
-    std::vector<Ort::ConstValueInfo> initializers = graph.GetInitializers();
+OrtStatus* ExampleEp::TrySaveConstantInitializer(Ort::ConstValueInfo maybe_initializer) {
+  EXCEPTION_TO_RETURNED_STATUS_BEGIN
+  const bool is_constant = maybe_initializer.IsConstantInitializer();
 
-    for (const auto& initializer : initializers) {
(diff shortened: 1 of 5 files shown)


onnxruntime/test/autoep/library/example_plugin_ep/ep.cc
```diff
@@ -377,11 +365,6 @@ OrtStatus* ORT_API_CALL ExampleEp::CompileImpl(_In_ OrtEp* this_ptr, _In_ const
 
     Ort::ConstGraph graph{ort_graphs[0]};
 
-    // In GetCapability(), this EP specified that it doesn't need ORT to provide constant initializers during inference.
```
tianleiwu, line 380: The `SaveConstantInitializers` removal from here (before the node count check) to after the node inputs check (L423-431) is a subtle correctness fix: the old code called `graph.GetInitializers()` which couldn't see initializers in parent/sibling subgraphs, silently failing for the multi-subgraph case.

PR #27668, opened 2026-03-15
Java: avoid provider resource extraction when library already exists in onnxruntime.native.path

### LINKED ISSUE #27655: [Java] Redundant library extraction when `onnxruntime.native.path` is set
### Description

In `OnnxRuntime.java`, `extractProviderLibrary` currently prioritizes `extractFromResources` regardless of whether `onnxruntime.native.path` is set.

If a user provides a native path, the main runtime DLL is loaded from that location. The runtime then locates its provider DLLs via standard OS lookups. Extracting these same DLLs from the JAR to a temporary directory is redundant, as the runtime will already find them in the `onnxruntime.native.path`.

### Inconsistency

The `load(String library)` method correctly skips extraction if `libraryDirPathProperty` is present. `extractProviderLibrary` should do the same to avoid unnecessary I/O and leftover temp files.

### Proposed Fix

Skip resource extraction if the library is already present in the user-specified native path:

```java
static synchronized boolean extractProviderLibrary(String libraryName) {
    if (isAndroid()) return false;
    if (extractedSharedProviders.contains(libraryName)) return true;

    // If path is specified, check there first and skip extraction if found
    if (libraryDirPathProperty != null) {
        File libraryFile = Paths.get(libraryDirPathProperty, mapLibraryName(libraryName)).toFile();
        if (libraryFile.exists()) {
            extractedSharedProviders.add(libraryName);
            return true; 
        }
    }

    // Only extract if not found in the specified native path
    Optional<File> file = extractFromResources(libraryName);
    if (file.isPresent()) {
        extractedSharedProviders.add(libraryName);
        return true;
    }

    return false;
}

```

TsofnatMaman:
## Summary

Fixes the Java native provider loading flow so that `extractProviderLibrary` checks
`onnxruntime.native.path` before attempting extraction from classpath resources.

Previously, when a provider library was already present in `onnxruntime.native.path`,
the Java loader could still attempt `extractFromResources(...)` first, and only fall
back to the configured native path afterwards. This caused an unnecessary extraction
attempt even though the library was already available on disk.

This change updates the lookup order to:

1. Return immediately if the provider was already marked as ready
2. Check whether the provider library already exists in `onnxruntime.native.path`
3. Only if not found there, attempt extraction from classpath resources

## Tests

Added a regression test covering the case where the requested provider library already
exists in `onnxruntime.native.path`.

The test verifies that:

- `extractProviderLibrary(...)` returns `true`
- extraction from resources is not attempted in that case

A small test-only hook was added to observe calls to `extractFromResources(...)` so the
regression can be validated directly and deterministically.

## Issue

Fixes #27655


TsofnatMaman:
Thanks for the review, @skottmckay
Some required checks are stuck in “Expected — Waiting for status to be reported”.
Could you please trigger them?
Thanks!

PR #27669, opened 2026-03-15
[Build] Fix clang build issues for CPU and CUDA builds

tianleiwu:
## Description

This PR fixes clang-specific build failures that show up in both the standalone clang build and the CUDA clang build. It keeps the build-system changes targeted, prefers source fixes where the warnings indicate real type or declaration issues, and avoids broader warning suppression than necessary for the CUDA provider target.

### Build System

| File | Change |
|------|--------|
| `cmake/CMakeLists.txt` | Stop forwarding `-Wshorten-64-to-32` through CUDA host compilation where the GNU host compiler does not recognize it. |
| `cmake/onnxruntime_providers_cuda.cmake` | Add targeted clang `-Wno-error` handling for warning classes that are currently triggered by CUDA provider code and third-party CUDA headers under clang. |

### CPU / Common clang fixes

| File | Change |
|------|--------|
| `onnxruntime/core/common/cpuid_info.cc` | Replace the clang-incompatible `__builtin_cpu_supports("waitpkg")` path with the CPUID-bit check for TPAUSE detection. |
| `onnxruntime/test/framework/allocation_planner_test.cc` | Refactor `typeid` assertions to avoid clang's potentially-evaluated-expression warning while keeping test coverage unchanged. |

### CUDA provider and contrib fixes

| File | Change |
|------|--------|
| `onnxruntime/contrib_ops/cuda/utils/dump_cuda_tensor.h` | Mark the `IConsoleDumper` overrides explicitly while leaving CUDA-only overloads unchanged. |
| `onnxruntime/contrib_ops/cuda/bert/group_query_attention.cc` | Use `template` on the dependent `GetAttrOrDefault` call so clang parses it correctly. |
| `onnxruntime/contrib_ops/cuda/bert/flash_attention/flash_api.cc` | Make narrowing conversions to flash-attention parameter fields explicit. |
| `onnxruntime/contrib_ops/cuda/quantization/matmul_nbits.cc` | Make the `nbits_` conversion explicit when calling the CUDA helper. |
| `onnxruntime/contrib_ops/cuda/quantization/moe_quantization.cc` | Restrict the GCC-only warning pragma so clang does not treat it as an unknown warning option. |
| `onnxruntime/contrib_ops/cuda/transformers/generation_device_helper.cc` | Fix explicit state-field assignments to use the actual `int` field type. |
| `onnxruntime/core/providers/cuda/cuda_mempool_arena.h` | Remove an unused private field that clang flagged in the CUDA provider build. |

## Testing

Tested CPU and CUDA 12.8 builds in Azure Linux with
- clang 18.1.8
- gcc 13.2
- cmake 4.2.3

Example for CPU build:
```
export CC=clang
export CXX=clang++
bash build.sh --config RelWithDebInfo --parallel --cmake_extra_defines onnxruntime_BUILD_UNIT_TESTS=ON
```
## Motivation and Context

Clang is stricter than GCC/MSVC in a few areas that affect this tree: CUDA host flag forwarding, explicit narrowing, dependent template parsing, warnings emitted from third-party CUDA headers, and RTTI/typeid expressions in tests. The goal here is to keep the staged fix minimal and maintainable by correcting real source issues where practical and confining warning downgrades to the CUDA provider target where third-party header noise is currently unavoidable.

### DIFF
--- a/onnxruntime/test/framework/allocation_planner_test.cc
+++ b/onnxruntime/test/framework/allocation_planner_test.cc
@@ -103,21 +103,6 @@ class AllocationPlanTestUtility {
   // TODO: add the tests for new release plan.
 };
 
-void ExpectExecutionStepTypeContains(const SessionState& state,
-                                     size_t stream_idx,
-                                     size_t step_idx,
-                                     const char* expected_type_name,
-                                     const char* message) {
-  const auto* execution_plan = state.GetExecutionPlan();
-  ASSERT_NE(execution_plan, nullptr) << message;
-  ASSERT_LT(stream_idx, execution_plan->execution_plan.size()) << message;
-  const auto& steps = execution_plan->execution_plan[stream_idx]->steps_;
-  ASSERT_LT(step_idx, steps.size()) << message;
-  const auto* step = steps[step_idx].get();
-  ASSERT_NE(step, nullptr) << message;
-  EXPECT_NE(strstr(typeid(*step).name(), expected_type_name), nullptr) << message;
-}
-
 typedef std::unordered_map<const onnxruntime::NodeArg*, TensorShapeProto*> ShapeMap;
 
 class SequentialPlannerTestContext : public ISequentialPlannerContext {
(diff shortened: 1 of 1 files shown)


onnxruntime/test/framework/allocation_planner_test.cc
```diff
@@ -1261,6 +1261,21 @@ TEST_F(PlannerTest, LocationPlanningForImplicitInputsWithoutExplicitConsumersInM
   // EXPECT_EQ(para_graph_plan->allocation_plan[input_data_index].location.device.Type(), OrtDevice::GPU);
```
kunal-vaishnavi, line 1261:
```suggestion
```

cmake/onnxruntime_providers_cuda.cmake
```diff
@@ -222,6 +224,38 @@
...
+          defaulted-function-deleted
+          inconsistent-missing-override
+          instantiation-after-specialization
+          logical-op-parentheses
+          mismatched-tags
+          shorten-64-to-32
+          unneeded-internal-declaration
+          unknown-warning-option
+          unused-private-field
+          unused-variable)
```
edgchen1, line 239: could we define a variable to contain these warnings so they only need to be maintained in one place?

onnxruntime/contrib_ops/cuda/quantization/moe_quantization.cc
```diff
@@ -162,7 +162,7 @@ Status QMoE<T>::ComputeInternal(OpKernelContext* context) const {
       activation_type_ == ort_fastertransformer::ActivationType::SwiGLU,
       block_size_));
 
-#if defined(__GNUC__)
+#if defined(__GNUC__) && !defined(__clang__)
```
edgchen1, line 165: can use `defined(HAS_MAYBE_UNINITIALIZED)` from onnxruntime_config.h instead of checking specifically for clang.

https://github.com/microsoft/onnxruntime/blob/535161a87bcb6182899c85197e1601a27658ffa7/cmake/onnxruntime_config.h.in#L17

tianleiwu: Good suggestion. I'll update it later.

onnxruntime/test/framework/allocation_planner_test.cc
tianleiwu: That's existed code. I think it's fine to leave it there.

cmake/onnxruntime_providers_cuda.cmake
tianleiwu: Good suggestion. Let's refactor it later.

PR #27674, opened 2026-03-16
[webgpu] Set `is_channels_last` to true by default in `ComputeMatMul`

Jiawei-Shao:
This patch sets `is_channels_last` to true by default in the parameter
of `ComputeMatMul` and ignores it in `UseSplitK` when there is no
`bias`.


Jiawei-Shao:
@qjia7 PTAL, thanks!

onnxruntime/core/providers/webgpu/math/matmul.h
```diff
@@ -14,7 +14,8 @@
 namespace onnxruntime {
 namespace webgpu {
 
-Status ComputeMatMul(ComputeContext* context, const Activation& activation, std::vector<const Tensor*>& inputs, Tensor* output, bool is_channels_last,
+Status ComputeMatMul(ComputeContext* context, const Activation& activation, std::vector<const Tensor*>& inputs, Tensor* output,
+                     std::optional<bool> is_channels_last = {},
```
qjia7: How about setting is_channels_last to true by default. Then in many places, we even don't need to pass this value since it's only for conv. For example, in attention.cc, directly call `onnxruntime::webgpu::ComputeMatMul(&context, Activation(), matmul_inputs, &packed_qkv)`.
```suggestion
                     bool is_channels_last = true,
```

Jiawei-Shao: Done

onnxruntime/core/providers/webgpu/webgpu_utils.cc
```diff
@@ -89,9 +90,14 @@ bool SplitKConfig::UseSplitK(
...
   use_split_k &= batch_size == 1;
-  // Now `is_channels_last` is only supported because we only generate vec4 shaders in
-  // `MatMulFillBiasOrZeroBeforeSplitKProgram` when `is_gemm` is false.
-  use_split_k &= (is_channels_last || is_gemm);
+
+  // Now we only need `is_channels_last` in `Conv|MatMul` with `bias`. We don't need to care about
+  // it in other places (`GEMM`, `MatMul` and `Conv|MatMul` without `bias`).
+  // When `is_channels_last` has valid value `is_channels_last` is required to be true because
+  // we only generate `vec4` shaders in  `MatMulFillBiasOrZeroBeforeSplitKProgram`.
+  if (has_bias && !is_gemm) {
```
qjia7: Why we need `!is_gemm` check? Are below checking enough?
```
if (has_bias)
  use_split_k &= is_channels_last;
```

Jiawei-Shao: Done

onnxruntime/core/providers/webgpu/webgpu_utils.h
```diff
@@ -106,9 +106,8 @@ class SplitKConfig {
   explicit SplitKConfig(const wgpu::AdapterInfo& adapter_info);
 
   bool UseSplitK(
-      bool is_vec4, ActivationKind activation_kind, uint64_t batch_size, bool is_gemm,
-      bool is_channels_last, uint32_t dim_a_outer,
-      uint32_t dim_b_outer, uint32_t dim_inner) const;
+      bool is_vec4, ActivationKind activation_kind, uint64_t batch_size, bool has_bias,
+      bool is_channels_last, uint32_t dim_a_outer, uint32_t dim_b_outer, uint32_t dim_inner) const;
```
qjia7: How about adjusting `is_channels_last` as the last parameter and set to true by default like ComputeMatMul did?

qjia7: In this case, we even don't need to pass `has_bias`. In the function, just directly use `use_split_k &= is_channels_last;`.

Jiawei-Shao: Done

Jiawei-Shao:
Hi @guschmue, could you take a look at this PR?

guschmue (approved):

Jiawei-Shao:
The errors are not related to this PR:

The self-hosted runner lost communication with the server. Verify the machine is running and has a healthy network connection. Anything in your workflow that terminates the runner process, starves it for CPU/Memory, or blocks its network access can cause this error.

Jiawei-Shao:
The failures on DirectML CI are not related to this PR.

PR #27681, opened 2026-03-16
[CoreML EP] Add Expand and Where op builders for BERT model support

### LINKED ISSUE #25421: [Feature Request] Support BERT models In CoreML execution provider

We want to run on-device a fine-tuned BERT model for token classification.

Currently, the partial support of the models' operations does not utilize ANE/GPU fully, and some operations are assigned to the CPU, causing the model to be too slow for production purposes.

Converting the model to `.mlpackage` and running it directly with CoreML works with ANE/GPU and runs fast.

It seems there are 2 parts to enable BERT models support:
1. Support `dim > 16384` - base BERT default vocab size is `30522` ([source](https://huggingface.co/docs/transformers/en/model_doc/bert)), which requires retraining the model with a smaller, more limited vocab size
2. Support the relevant operations. When loading the model, this are the operations assigned to the CPU:
```
2025-07-16 13:16:26.883836 [V:onnxruntime:, session_state.cc:1266 VerifyEachNodeIsAssignedToAnEp]  Node(s) placed on [CPUExecutionProvider]. Number of nodes: 12
2025-07-16 13:16:26.883841 [V:onnxruntime:, session_state.cc:1268 VerifyEachNodeIsAssignedToAnEp]   Unsqueeze (/bert/Unsqueeze)
2025-07-16 13:16:26.883845 [V:onnxruntime:, session_state.cc:1268 VerifyEachNodeIsAssignedToAnEp]   Unsqueeze (/bert/Unsqueeze_1)
2025-07-16 13:16:26.883850 [V:onnxruntime:, session_state.cc:1268 VerifyEachNodeIsAssignedToAnEp]   Expand (/bert/Expand)
2025-07-16 13:16:26.883854 [V:onnxruntime:, session_state.cc:1268 VerifyEachNodeIsAssignedToAnEp]   Cast (/bert/Cast)
2025-07-16 13:16:26.883858 [V:onnxruntime:, session_state.cc:1268 VerifyEachNodeIsAssignedToAnEp]   Cast (/bert/Cast_1)
2025-07-16 13:16:26.883863 [V:onnxruntime:, session_state.cc:1268 VerifyEachNodeIsAssignedToAnEp]   Where (/bert/Where_1)
2025-07-16 13:16:26.883867 [V:onnxruntime:, session_state.cc:1268 VerifyEachNodeIsAssignedToAnEp]   MatMul (/bert/encoder/layer.0/attention/self/MatMul)
2025-07-16 13:16:26.883872 [V:onnxruntime:, session_state.cc:1268 VerifyEachNodeIsAssignedToAnEp]   MatMul (/bert/encoder/layer.0/attention/self/MatMul_1)
2025-07-16 13:16:26.883876 [V:onnxruntime:, session_state.cc:1268 VerifyEachNodeIsAssignedToAnEp]   MatMul (/bert/encoder/layer.1/attention/self/MatMul)
2025-07-16 13:16:26.883880 [V:onnxruntime:, session_state.cc:1268 VerifyEachNodeIsAssignedToAnEp]   MatMul (/bert/encoder/layer.1/attention/self/MatMul_1)
2025-07-16 13:16:26.883885 [V:onnxruntime:, session_state.cc:1268 VerifyEachNodeIsAssignedToAnEp]   MatMul (/bert/encoder/layer.2/attention/self/MatMul)
2025-07-16 13:16:26.883892 [V:onnxruntime:, session_state.cc:1268 VerifyEachNodeIsAssignedToAnEp]   MatMul (/bert/encoder/layer.2/attention/self/MatMul_1)
```

BERT is a very common model family used for natural text tasks.
While GPT-like models dominate a lot of text tasks, specifically classification tasks on-device will probably be left for the BERT family and it's evolutions (modern-BERT) - they are much smaller in size, and better suited for classification as an encoder.

Supporting them in ONNX will allow also efficient cross-platform hardware utilization.
They are already supported in OpenVino and Cuda Execution Providers.

Rishi-Dave:
## Summary

- Add `Expand` op builder mapping ONNX Expand to CoreML ML Program `broadcast_to`
- Add `Where` op builder mapping ONNX Where to CoreML ML Program `select`
- Fix `Cast` op builder to support graph inputs in ML Program mode (no preceding nodes required)
- Add BOOL type support to Cast's input/output type validation for ML Program

## Motivation

Fixes #25421

BERT-family models running on the CoreML execution provider currently fall back to CPU for several critical operations: Expand, Where, Cast (on graph inputs), Unsqueeze, and attention-layer MatMul. This causes significant performance degradation compared to running the model natively through CoreML.

This PR addresses the most impactful gaps — the two completely missing op builders (Expand and Where) and the Cast limitation that blocked type casting on graph inputs like `attention_mask`.

With these changes, the BERT attention masking pipeline (Unsqueeze → Expand → Cast → Where) can execute entirely on CoreML's ML Program path, enabling ANE/GPU acceleration.

## Changes

**New files:**
- `expand_op_builder.cc` — Maps ONNX `Expand` to CoreML `broadcast_to`. Handles both constant initializer shapes (consumed directly as const ops) and dynamic runtime shapes. ML Program only, opset 8+.
- `where_op_builder.cc` — Maps ONNX `Where` to CoreML `select`. Validates BOOL condition input and FLOAT/FLOAT16/INT64 value inputs. ML Program only, opset 9+.

**Modified files:**
- `cast_op_builder.cc` — Moved the ML Program early return before the `GetInputEdgesCount() == 0` check, so Cast on graph inputs (e.g., casting `attention_mask` from INT64 to FLOAT) is supported in ML Program mode. Also added BOOL to the supported input/output type list for ML Program Cast.
- `op_builder_factory.h` / `op_builder_factory.cc` — Registered the new Expand and Where op builders.

## Test Plan

- [ ] Verify with a BERT token classification model (e.g., `bert-base-uncased` exported to ONNX) that Expand, Where, and Cast nodes are now assigned to CoreMLExecutionProvider instead of CPUExecutionProvider
- [ ] Run existing CoreML EP tests to confirm no regressions
- [ ] Validate on macOS with ML Program mode enabled

### DIFF
--- a/onnxruntime/core/providers/coreml/builders/impl/cast_op_builder.cc
+++ b/onnxruntime/core/providers/coreml/builders/impl/cast_op_builder.cc
@@ -150,8 +150,8 @@ bool CastOpBuilder::HasSupportedInputsImpl(const Node& node, [[maybe_unused]] co
     } else {
       LOGS(logger, VERBOSE) << "[" << node.OpType()
                             << "] Input type: [" << input_type
-                            << "] or output type: [" << output_type
-                            << "] is not supported.";
+                            << "] and/or output type: [" << output_type
+                            << "] is not supported for ML Program Cast.";
       return false;
     }
   }
--- a/onnxruntime/core/providers/coreml/builders/impl/cast_op_builder.cc
+++ b/onnxruntime/core/providers/coreml/builders/impl/cast_op_builder.cc
@@ -77,13 +77,16 @@ Status CastOpBuilder::AddToModelBuilderImpl([[maybe_unused]] ModelBuilder& model
 
 bool CastOpBuilder::IsOpSupportedImpl(const Node& node, const OpBuilderInputParams& input_params,
                                       const logging::Logger& logger) const {
+  // ML Program supports Cast on all inputs including graph inputs with no preceding nodes.
+  // This is important for models like BERT where Cast is applied directly to graph inputs
+  // (e.g., casting attention_mask from INT64 to FLOAT).
+  if (input_params.create_mlprogram) {
+    return true;
+  }
+
   if (node.GetInputEdgesCount() == 0) {
     LOGS(logger, VERBOSE) << "Cast has no preceding nodes.";
     return false;
-  }
-
-  if (input_params.create_mlprogram) {
-    return true;
   }
 
   const auto& prec_node = node.InputEdgesBegin()->GetNode();
@@ -132,18 +135,22 @@ bool CastOpBuilder::HasSupportedInputsImpl(const Node& node, [[maybe_unused]] co
   }
 
   if (input_params.create_mlprogram) {
-    if ((input_type == ONNX_NAMESPACE::TensorProto_DataType_INT32 ||
-         input_type == ONNX_NAMESPACE::TensorProto_DataType_INT64 ||
-         input_type == ONNX_NAMESPACE::TensorProto_DataType_FLOAT ||
-         input_type == ONNX_NAMESPACE::TensorProto_DataType_FLOAT16) &&
-        (output_type == ONNX_NAMESPACE::TensorProto_DataType_INT32 ||
-         output_type == ONNX_NAMESPACE::TensorProto_DataType_INT64 ||
-         output_type == ONNX_NAMESPACE::TensorProto_DataType_FLOAT ||
-         output_type == ONNX_NAMESPACE::TensorProto_DataType_FLOAT16)) {
+    // ML Program Cast supports INT32, INT64, FLOAT, FLOAT16, and BOOL types.
+    // BOOL support is important for models like BERT that cast boolean attention masks.
+    const auto is_supported_type = [](int32_t type) {
+      return type == ONNX_NAMESPACE::TensorProto_DataType_INT32 ||
+             type == ONNX_NAMESPACE::TensorProto_DataType_INT64 ||
+             type == ONNX_NAMESPACE::TensorProto_DataType_FLOAT ||
+             type == ONNX_NAMESPACE::TensorProto_DataType_FLOAT16 ||
+             type == ONNX_NAMESPACE::TensorProto_DataType_BOOL;
+    };
+
+    if (is_supported_type(input_type) && is_supported_type(output_type)) {
       return true;
     } else {
       LOGS(logger, VERBOSE) << "[" << node.OpType()
                             << "] Input type: [" << input_type
+                            << "] or output type: [" << output_type
                             << "] is not supported.";
       return false;
     }
--- a/onnxruntime/core/providers/coreml/builders/impl/expand_op_builder.cc
+++ b/onnxruntime/core/providers/coreml/builders/impl/expand_op_builder.cc
@@ -1,0 +1,111 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include "core/optimizer/initializer.h"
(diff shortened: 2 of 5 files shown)


Rishi-Dave:
Thanks for the review — all three points addressed:

1. **Expand `broadcast_to` shape dtype** — For the constant-initializer path, `AddConstant` with `span<int64_t>` already produces an INT32 CoreML constant internally (via `CreateTensorValue<int64_t, int32_t>`), so that was correct. For the **dynamic path**, I've inserted an explicit MIL `cast` op that converts the int64 shape input to int32 before wiring it to `broadcast_to.shape`. This follows the same cast-insertion pattern used elsewhere in the EP (e.g., `cast_op_builder.cc`).

2. **Where X/Y type mismatch** — Added an explicit check in `HasSupportedInputsImpl` that fetches both X and Y dtypes and rejects the op if they don't match. This runs after the individual `IsInputDtypeSupport` checks, so we still validate each input is in the supported set before comparing.

3. **Cast log message** — Reworded to `"and/or output type ... is not supported for ML Program Cast"` to clearly indicate that either (or both) types may be the unsupported one.

yuslepukhin:
1. Where builder allows INT64 but CoreML select does not support it
The HasSupportedInputsImpl validates X/Y via IsInputDtypeSupport, and the PR description says it accepts FLOAT/FLOAT16/INT64. However, per the CoreML MIL spec, select supports T: fp16, fp32, i32, bool — not int64. This will produce an invalid MIL program at runtime.

Where builder should either:

Reject INT64 inputs, or
Insert a cast op (int64 → int32) before feeding to select

2. Expand constant-path int64→int32 truncation risk
In expand_op_builder.cc line 60-62, the constant initializer path passes span<int64_t> directly to AddConstant. The author's follow-up comment claims AddConstant internally does CreateTensorValue<int64_t, int32_t>, which would handle conversion. However:

There is no explicit range check before the narrowing conversion.
If a shape dimension exceeds INT32_MAX (unlikely but possible in theory), silent truncation would occur.
The code comment on lines 58 still says "Shape is a constant initializer — read it and add as a const operation" which doesn't document the implicit narrowing.

3. Expand HasSupportedInputsImpl is missing
Unlike the Where builder, the Expand builder does not override HasSupportedInputsImpl. This means it relies on the base class default validation. The ONNX Expand op supports all tensor types, but CoreML broadcast_to likely only supports fp16, fp32, i32, bool. Without explicit type validation, unsupported types (e.g., int64, uint8, string) could slip through and cause CoreML compilation failures.

4. Cast builder early return bypasses legitimate NeuralNetwork-mode checks
In cast_op_builder.cc lines 80-85, the ML Program early return (return true) skips all further validation for ML Program mode in IsOpSupportedImpl. While the type checking in HasSupportedInputsImpl covers types, there's no node-level structural validation for ML Program anymore.

5. ORT_UNUSED_PARAMETER(logger) is unnecessary
In both expand_op_builder.cc line 86 and where_op_builder.cc line 54, logger is marked unused even though the parameter is [[maybe_unused]] in the signature or could be used for diagnostics. This is a minor code quality issue.

6. Test Coverage Assessment
The PR has NO unit tests. The "Test Plan" only describes manual validation with a BERT model. This is a significant gap:

Missing positive tests: No test for Expand with constant shape, Expand with dynamic shape, Where with float inputs, Where with float16 inputs, Cast on graph inputs with BOOL type.
Missing negative tests: No test for unsupported rank (>5), unsupported types, Where with mismatched X/Y types.
Missing edge cases: Expand with scalar input, Where with all-true/all-false condition, broadcasting behavior verification.

Issue #27177, opened 2026-01-27
[Feature Request] Reliable runtime detection of usable Execution Providers (e.g. CUDA)


MaizeShark:
### Describe the feature request

ONNX Runtime exposes `onnxruntime.get_available_providers()` to query available execution providers. However, this API reports **compile-time–enabled providers**, not providers that are **actually usable at runtime**.

As a result, `CUDAExecutionProvider` (and `TensorrtExecutionProvider`) may be reported as available even when required runtime dependencies (e.g. CUDA, cuDNN, or specific shared libraries like `libcublasLt.so`) are missing or incompatible. Attempting to use these providers then fails during session creation or silently falls back to CPU.

This behavior makes it difficult for applications, especially consumer-facing tools, to reliably select the best execution provider or present accurate diagnostics to users.

### Proposed feature

Add a new API (or extend existing APIs) that allows applications to **reliably detect whether an execution provider is usable at runtime**, not just compiled in.

Examples of possible APIs:

```python
onnxruntime.is_provider_usable("CUDAExecutionProvider")
```

or

```python
onnxruntime.get_loadable_providers()
```

Such an API would:

* Attempt to load the provider’s shared library
* Validate required runtime dependencies (e.g. CUDA / cuDNN)
* Return a deterministic result without requiring model loading or exception-based control flow

### Describe scenario use case

This feature would benefit applications that need to **automatically select the optimal execution provider** at startup, including:

* Consumer-facing ML inference applications
* GUI tools that want to display accurate GPU availability status
* Cross-platform software distributed to users with varying CUDA setups
* Applications that want to avoid silent CPU fallbacks or confusing runtime errors

Currently, developers must attempt to create an `InferenceSession` with `CUDAExecutionProvider` and catch exceptions to determine usability. This approach:

* Requires a valid model
* Relies on exceptions for normal control flow
* Produces poor UX when CUDA appears “available” but fails at runtime

A first-class, runtime-accurate provider detection API would simplify application logic, reduce user confusion, and improve robustness.

sagarbhure-msft:
Raised a PR with an implementation for this: #27697 adds get_usable_providers() and is_provider_usable() APIs that check runtime dependencies instead of compile-time flags. This would eliminate a common pain point where apps silently fall back to CPU without any indication, which is especially impactful for consumer-facing tools where users don't understand why their GPU isn't being used.

Could I be assigned this issue? Would love any feedback on the PR

PR #27697, opened 2026-03-17
Add runtime detection of usable execution providers

sagarbhure-msft:
## Overview
Adds runtime-aware APIs to detect **usable execution providers**.

Unlike `get_available_providers()` (compile-time), these APIs verify that required runtime dependencies (CUDA, cuDNN, shared libs, etc.) are actually present.

This PR adds two new APIs that check whether provider shared libraries actually exist on disk at runtime, without loading them.

---

## New APIs

Python:
```python
 import onnxruntime as ort
 
 # Returns only providers whose shared libraries are present
 ort.get_usable_providers()
 
 # Check a specific provider
 ort.is_provider_usable("CUDAExecutionProvider")
```

C++:
 
```cpp
 onnxruntime::GetUsableExecutionProviderNames();
 onnxruntime::IsExecutionProviderUsable("CUDAExecutionProvider");
``` 

---

## Key Changes

 - Extended existing ProviderInfo struct in kProvidersInPriorityOrder with library_base_name field — single source of truth, no 
duplicate EP lists
 - Runtime check uses std::filesystem::exists(GetRuntimePath() / lib_filename) — no DLL loading, no hardware init, no error logs, 
no memory footprint increase
 - Static providers (CPU, DML, CoreML, XNNPACK, etc.) have library_base_name = nullptr → always usable if compiled in
 - QNN static build (BUILD_QNN_EP_STATIC_LIB) and WebGPU plugin EP adapters handled
 - ORT_MINIMAL_BUILD falls back to compile-time availability (documented in API docstrings)
 - No changes to provider_bridge_ort.cc or .h

---

## Motivation

`get_available_providers()` can report providers as available even when runtime dependencies are missing.

Current workaround (bad):

```python
try:
    ort.InferenceSession(model, providers=["CUDAExecutionProvider"])
    has_cuda = True
except:
    has_cuda = False
```

New APIs remove the need for exception-based checks.

---

## Tested on Real Hardware

Machine: AMD Ryzen AI MAX+ 395 / Radeon 8060S, Windows

**CPU-only build  (onnxruntime 1.24.3)**

```
get_available_providers: ['AzureExecutionProvider', 'CPUExecutionProvider']
get_usable_providers:    ['CPUExecutionProvider']
```

**DirectML build  (onnxruntime-directml 1.24.3)**

```
get_available_providers: ['DmlExecutionProvider', 'CPUExecutionProvider']
get_usable_providers:    ['DmlExecutionProvider', 'CPUExecutionProvider']
```
Both providers correctly reported as usable, verified with real inference
---

Who this affects

   - Consumer-facing ML inference applications
   - Cross-platform software distributed to users with varying hardware setups
   - Applications that want to avoid silent CPU fallbacks or confusing runtime errors

Fixes https://github.com/microsoft/onnxruntime/issues/27177

### DIFF
--- a/onnxruntime/core/providers/get_execution_providers.cc
+++ b/onnxruntime/core/providers/get_execution_providers.cc
@@ -6,7 +6,13 @@
 #include "core/graph/constants.h"
 #include "core/common/common.h"
 
+#include <algorithm>
 #include <string_view>
+
+#if !defined(ORT_MINIMAL_BUILD)
+#include "core/platform/env.h"
+#include <filesystem>
+#endif
 
 namespace onnxruntime {
 
@@ -14,10 +20,19 @@ namespace {
 struct ProviderInfo {
   std::string_view name;
   bool available;
+  // Shared library base name (without prefix/extension), or nullptr for
+  // statically linked providers. This is the single source of truth for
+  // mapping provider names to their library filenames.
+  const char* library_base_name;
 };
 
 // all providers ordered by default priority from highest to lowest
 // kCpuExecutionProvider should always be last
+//
+// library_base_name: the base name of the shared library for this provider,
+// e.g. "onnxruntime_providers_cuda" -> onnxruntime_providers_cuda.dll (Win)
+//      or libonnxruntime_providers_cuda.so (Linux).
+// nullptr means the provider is statically linked (no shared library to check).
 constexpr ProviderInfo kProvidersInPriorityOrder[] =
     {
         {
@@ -27,6 +42,7 @@ constexpr ProviderInfo kProvidersInPriorityOrder[] =
 #else
             false,
 #endif
+            "onnxruntime_providers_nv_tensorrt_rtx",
         },
         {
             kTensorrtExecutionProvider,
@@ -35,6 +51,7 @@ constexpr ProviderInfo kProvidersInPriorityOrder[] =
 #else
             false,
 #endif
+            "onnxruntime_providers_tensorrt",
         },
         {
             kCudaExecutionProvider,
@@ -43,6 +60,7 @@ constexpr ProviderInfo kProvidersInPriorityOrder[] =
 #else
             false,
 #endif
+            "onnxruntime_providers_cuda",
         },
         {
             kMIGraphXExecutionProvider,
@@ -51,6 +69,7 @@ constexpr ProviderInfo kProvidersInPriorityOrder[] =
 #else
             false,
 #endif
+            "onnxruntime_providers_migraphx",
         },
         {
             kOpenVINOExecutionProvider,
@@ -59,6 +78,7 @@ constexpr ProviderInfo kProvidersInPriorityOrder[] =
 #else
             false,
 #endif
+            "onnxruntime_providers_openvino",
         },
         {
             kDnnlExecutionProvider,
@@ -67,6 +87,7 @@ constexpr ProviderInfo kProvidersInPriorityOrder[] =
 #else
             false,
 #endif
+            "onnxruntime_providers_dnnl",
         },
         {
             kVitisAIExecutionProvider,
@@ -75,6 +96,7 @@ constexpr ProviderInfo kProvidersInPriorityOrder[] =
 #else
             false,
 #endif
+            "onnxruntime_providers_vitisai",
         },
         {
             kQnnExecutionProvider,
@@ -82,6 +104,11 @@ constexpr ProviderInfo kProvidersInPriorityOrder[] =
             true,
 #else
             false,
+#endif
+#if defined(BUILD_QNN_EP_STATIC_LIB) && BUILD_QNN_EP_STATIC_LIB
+            nullptr,  // QNN is statically linked in this build
+#else
+            "onnxruntime_providers_qnn",
 #endif
         },
         {
@@ -91,6 +118,7 @@ constexpr ProviderInfo kProvidersInPriorityOrder[] =
 #else
             false,
 #endif
+            nullptr,
         },
         {
             kVSINPUExecutionProvider,
@@ -99,6 +127,7 @@ constexpr ProviderInfo kProvidersInPriorityOrder[] =
 #else
             false,
 #endif
+            nullptr,
         },
         {
             kJsExecutionProvider,
@@ -107,6 +136,7 @@ constexpr ProviderInfo kProvidersInPriorityOrder[] =
 #else
             false,
 #endif
+            nullptr,
         },
         {
             kCoreMLExecutionProvider,
@@ -115,6 +145,7 @@ constexpr ProviderInfo kProvidersInPriorityOrder[] =
 #else
             false,
 #endif
+            nullptr,
         },
         {
             kAclExecutionProvider,
@@ -123,6 +154,7 @@ constexpr ProviderInfo kProvidersInPriorityOrder[] =
 #else
             false,
 #endif
+            nullptr,
         },
         {
             kDmlExecutionProvider,
@@ -131,6 +163,7 @@ constexpr ProviderInfo kProvidersInPriorityOrder[] =
 #else
             false,
 #endif
+            nullptr,
         },
         {
             kRknpuExecutionProvider,
@@ -139,6 +172,7 @@ constexpr ProviderInfo kProvidersInPriorityOrder[] =
 #else
             false,
 #endif
+            nullptr,
         },
         {
             kWebNNExecutionProvider,
@@ -147,6 +181,7 @@ constexpr ProviderInfo kProvidersInPriorityOrder[] =
 #else
             false,
 #endif
+            nullptr,
         },
         {
             kWebGpuExecutionProvider,
@@ -155,6 +190,11 @@ constexpr ProviderInfo kProvidersInPriorityOrder[] =
 #else
             false,
 #endif
+            // When USE_WEBGPU is defined with ORT_USE_EP_API_ADAPTERS, WebGPU is
+            // loaded via the plugin EP system, not via the provider bridge. We mark
+            // it as nullptr (statically linked) here; the plugin EP adapter handles
+            // its own loading. If WebGPU is built without adapters, it's static.
+            nullptr,
         },
         {
             kXnnpackExecutionProvider,
@@ -163,6 +203,7 @@ constexpr ProviderInfo kProvidersInPriorityOrder[] =
 #else
             false,
 #endif
+            nullptr,
         },
         {
             kCannExecutionProvider,
@@ -171,6 +212,7 @@ constexpr ProviderInfo kProvidersInPriorityOrder[] =
 #else
             false,
 #endif
+            "onnxruntime_providers_cann",
         },
         {
             kAzureExecutionProvider,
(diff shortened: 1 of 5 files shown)


sagarbhure-msft:

sagarbhure-msft:
Friendly nudge — would love eyes on this when someone gets a chance. cc @microsoft/onnxruntime-api 

cc: @yuslepukhin

onnxruntime/test/providers/get_execution_providers_test.cc
```diff
@@ -48,5 +48,48 @@ TEST(GetExecutionProvidersTest, NoDuplicates) {
   check(GetAvailableExecutionProviderNames());
 }
 
+TEST(GetExecutionProvidersTest, CpuAlwaysUsable) {
```
yuslepukhin, line 51: Would all of these test run OK in the minimal build?

onnxruntime/core/session/provider_bridge_ort.cc
```diff
@@ -2069,6 +2070,84 @@ void UnloadSharedProviders() {
   s_library_nv.Unload();
 }
 
+bool IsProviderLibraryLoadable(const std::string& provider_name) {
```
yuslepukhin: In this function the TryGetProviderInfo_*() functions call ProviderLibrary::Get(), which calls Load() and Initialize(). Once called, the shared library (.dll/.so) stays loaded for the process lifetime because ProviderLibrary::Unload() is only called at shutdown. Likewise s_library_vitisai.Load() and s_library_qnn.Load() also permanently load the library.

This means calling get_usable_providers() has the side effect of loading every shared-library provider into the process, consuming memory and potentially initializing hardware contexts. The docstring says "verifies that...dependencies are actually present and loadable," but the implementation goes beyond verification — it actually loads and initializes. A user might call this at startup just to discover what's available, not intending to use all providers.

This increases the footprint of the process which is undesirable. Modify the code so the libs are unloaded after the check.

Unloading them may be difficult and error prone.

One idea may be to let the user try to load a specific provider, the ones they are interested in. But I am not seeing much point in it since Append Provider calls are doing just that.

Another idea, instead of loading the DLLs, use std::filesystem API to resolve the path and verify it is an existing file and any links point to the shipped name of the EP DLL. See ValidateExternalDataPath() for an example in tensorprotoutils.cc (you branch is way out of sync).

onnxruntime/core/session/provider_bridge_ort.cc
```diff
@@ -2069,6 +2070,84 @@ void UnloadSharedProviders() {
...
+#ifdef USE_AZURE
+      kAzureExecutionProvider,
+#endif
+  };
+
+  if (kStaticallyLinkedProviders.count(provider_name)) {
+    return true;
+  }
+
+  // For shared library providers, attempt to load the library
```
yuslepukhin: >// For shared library provider [](http://example.com/codeflow?start=2&length=30)

As the Copilot reviewer noted in this comment:

For CUDA, TensorRT, OpenVINO, DNNL, CANN, MIGraphX: TryGetProviderInfo_*() → calls Get() → calls Load() + Initialize(). This is a deep check.
For VitisAI, QNN: only s_library_*.Load() is called — this verifies the .dll/.so can be loaded and GetProvider symbol resolved, but does not call Provider::Initialize().
This means VitisAI and QNN may report as "usable" even if Initialize() would fail (e.g., missing hardware). This is an asymmetric contract.

onnxruntime/core/session/provider_bridge_ort.cc
```diff
@@ -2069,6 +2070,84 @@ void UnloadSharedProviders() {
...
+  // using the existing TryGetProviderInfo_* infrastructure
+  try {
```
yuslepukhin: > try { [](http://example.com/codeflow?start=1&length=6)

When a user calls get_usable_providers() on a machine without CUDA, every unavailable shared-library provider will emit an ERROR log line. (See TryGetProviderInfo_*()) functions down below.

This is misleading — it's expected behavior, not an error as it is currently written. Users monitoring logs for real errors will get false alarms.

onnxruntime/core/session/provider_bridge_ort.cc
```diff
@@ -2069,6 +2070,84 @@ void UnloadSharedProviders() {
...
+#ifdef USE_XNNPACK
+      kXnnpackExecutionProvider,
+#endif
+#ifdef USE_JSEP
+      kJsExecutionProvider,
+#endif
+#ifdef USE_WEBNN
+      kWebNNExecutionProvider,
+#endif
+#if defined(USE_WEBGPU) && !defined(ORT_USE_EP_API_ADAPTERS)
```
yuslepukhin: When USE_WEBGPU is defined with ORT_USE_EP_API_ADAPTERS, WebGPU is excluded from kStaticallyLinkedProviders and has no entry in the shared-library if/else chain. It falls through to "Unknown provider → return false." This is a functional gap if WebGPU is in the available list but always reads as "not usable."

Suggestion: Either add a comment explaining that plugin-EP-based providers are intentionally excluded from this API, or add handling for the adapter path.

onnxruntime/core/providers/get_execution_providers.cc
```diff
@@ -216,4 +221,37 @@ const std::vector<std::string>& GetAvailableExecutionProviderNames() {
...
+  return true;
+#endif
+}
+
+std::vector<std::string> GetUsableExecutionProviderNames() {
+  const auto& available = GetAvailableExecutionProviderNames();
+#if !defined(ORT_MINIMAL_BUILD)
+  std::vector<std::string> usable{};
+  for (const auto& provider : available) {
+    if (IsProviderLibraryLoadable(provider)) {
```
yuslepukhin: > if (IsProviderLibraryLoadable(provider)) { [](http://example.com/codeflow?start=3&length=43)

GetUsableExecutionProviderNames() calls IsProviderLibraryLoadable() per-provider, not IsExecutionProviderUsable().

Is this what's intended? 

The function above, on the other hand, IsExecutionProviderUsable() calls IsProviderLibraryLoadable().

onnxruntime/core/session/provider_bridge_ort.cc
yuslepukhin: > IsProviderLibraryLoadable(c [](http://example.com/codeflow?start=4&length=28)

This function creates yet another list of EP names among several, that will have to be kept up to date and maintained. Please, consolidated it with some other list.

onnxruntime/test/providers/get_execution_providers_test.cc
```diff
@@ -48,5 +48,48 @@ TEST(GetExecutionProvidersTest, NoDuplicates) {
...
+  const auto& available = GetAvailableExecutionProviderNames();
+  const auto usable = GetUsableExecutionProviderNames();
+
+  // Verify usable providers maintain the same relative ordering as available
+  size_t last_available_idx = 0;
+  for (const auto& provider : usable) {
+    auto it = std::find(available.begin() + last_available_idx, available.end(), provider);
+    ASSERT_NE(it, available.end())
+        << "Usable provider " << provider << " not found in available list";
+    last_available_idx = std::distance(available.begin(), it);
```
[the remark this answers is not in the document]
sagarbhure-msft: Fixed, now uses static_cast<size_t>(std::distance(available.begin(), it)) + 1 to enforce strictly increasing order.

onnxruntime/core/session/provider_bridge_ort.cc
```diff
@@ -2069,6 +2070,84 @@ void UnloadSharedProviders() {
...
+    } else if (provider_name == kMIGraphXExecutionProvider) {
+      return TryGetProviderInfo_MIGraphX() != nullptr;
+    } else if (provider_name == kVitisAIExecutionProvider) {
+      // VitisAI uses shared library but doesn't have a TryGetProviderInfo_ function
+      // exposed in the same way. Try loading the library directly.
+      auto status = s_library_vitisai.Load();
+      return status.IsOK();
+    } else if (provider_name == kQnnExecutionProvider) {
+      auto status = s_library_qnn.Load();
+      return status.IsOK();
```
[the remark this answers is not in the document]
sagarbhure-msft: Fixed , QNN's library_base_name is nullptr (static) when BUILD_QNN_EP_STATIC_LIB is set, otherwise "onnxruntime_providers_qnn".

onnxruntime/core/session/provider_bridge_ort.cc
```diff
@@ -2069,6 +2070,84 @@ void UnloadSharedProviders() {
...
+      return TryGetProviderInfo_OpenVINO() != nullptr;
+    } else if (provider_name == kDnnlExecutionProvider) {
+      return TryGetProviderInfo_Dnnl() != nullptr;
+    } else if (provider_name == kCannExecutionProvider) {
+      return TryGetProviderInfo_CANN() != nullptr;
+    } else if (provider_name == kMIGraphXExecutionProvider) {
+      return TryGetProviderInfo_MIGraphX() != nullptr;
+    } else if (provider_name == kVitisAIExecutionProvider) {
+      // VitisAI uses shared library but doesn't have a TryGetProviderInfo_ function
+      // exposed in the same way. Try loading the library directly.
```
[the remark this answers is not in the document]
sagarbhure-msft: Removed TryGetProviderInfo_*() entirely. Uses std::filesystem::exists() now - no logs.

onnxruntime/core/session/provider_bridge_ort.cc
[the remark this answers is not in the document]
sagarbhure-msft: Fixed, all providers now use identical std::filesystem::exists() check. No more asymmetry

onnxruntime/python/onnxruntime_pybind_module.cc
```diff
@@ -117,6 +117,21 @@ PYBIND11_MODULE(onnxruntime_pybind11_state, m) {
...
+      "this function verifies that each provider's runtime dependencies "
+      "(e.g. CUDA toolkit, cuDNN) are actually present and loadable. "
+      "The order of elements represents the default priority order of Execution Providers "
+      "from highest to lowest.");
+  m.def(
+      "is_provider_usable", [](const std::string& provider_name) -> bool { return IsExecutionProviderUsable(provider_name); },
+      "Check whether a specific Execution Provider is usable at runtime. "
+      "Returns True if the provider is compiled in AND its runtime dependencies "
+      "(e.g. shared libraries, hardware drivers) can be loaded. "
+      "Returns False if the provider is not available or its dependencies are missing.",
```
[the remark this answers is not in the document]
sagarbhure-msft: Fixed, both APIs now document: "Note: In minimal builds, this behaves the same as get_available_providers()."

onnxruntime/test/providers/get_execution_providers_test.cc
sagarbhure-msft: Yes, tests call IsExecutionProviderUsable() and GetUsableExecutionProviderNames() which have `#if !defined(ORT_MINIMAL_BUILD)` guards. In minimal builds these degrade gracefully to compile-time  checks; all tests remain valid

onnxruntime/core/session/provider_bridge_ort.cc
sagarbhure-msft: Completely redesigned. No libraries are loaded. Uses `std::filesystem::exists()` to check if the file is present, using `Env::Default().GetRuntimePath()` + the same library naming convention as ProviderLibrary. Zero memory footprint, zero hardware init. Branch rebased onto latest origin/main.

onnxruntime/core/session/provider_bridge_ort.cc
sagarbhure-msft: Fixed entire approach replaced. All providers use identical `std::filesystem::exists()` via library_base_name field. Uniform contract

onnxruntime/core/session/provider_bridge_ort.cc
sagarbhure-msft: Extended existing ProviderInfo struct with library_base_name. No new list.

onnxruntime/core/session/provider_bridge_ort.cc
sagarbhure-msft: Fixed  WebGPU has library_base_name = nullptr with comment explaining plugin EP adapters handle  
their own loading. Treated as static when compiled in.

onnxruntime/core/providers/get_execution_providers.cc
sagarbhure-msft: both functions now check provider.available (compile-time) and DoesProviderLibraryExist() (runtime) through the same kProvidersInPriorityOrder table. GetUsableExecutionProviderNames() iterates the table directly for efficiency rather than calling IsExecutionProviderUsable() per-provider (which would do redundant FindProviderInfo lookups

sagarbhure-msft:
Addressed all review feedback - redesigned to use filesystem checks, no library loading.
 
 Thanks @yuslepukhin and `@copilot` for the thorough review. I've completely redesigned the implementation:
 
 Key changes:
 
  - No DLL/SO loading — uses std::filesystem::exists() to check library file presence. Zero memory footprint, zero hardware init, zero error logs.
  - Single source of truth — extended existing ProviderInfo struct in kProvidersInPriorityOrder with library_base_name field. No duplicate EP lists.
  - Edge cases — QNN static build (BUILD_QNN_EP_STATIC_LIB), WebGPU + EP adapters, minimal builds all handled.
  - Branch rebased onto latest origin/main.
  - No changes to provider_bridge_ort.cc/.h — implementation is self-contained in the providers layer (5 files, down from 7)

onnxruntime/core/session/provider_bridge_ort.cc
sagarbhure-msft: Fixed  TryGetProviderInfo_*() is no longer called. The v2 implementation uses std::filesystem::exists() which produces zero log output. No more false ERROR alarms during capability queries.

yuslepukhin:
It looks like you force pushed. I am not able to see the differences with the previous version, and the local update of your branch is conflicting. I will take a look at it when I have time to deal with it.

sagarbhure-msft:
> It looks like you force pushed. I am not able to see the differences with the previous version, and the local update of your branch is conflicting. I will take a look at it when I have time to deal with it.

 Oops, sorry about the force push! Rebased + redesigned in one shot, lesson learned - incremental commits next time
 🫡
 
 Quick summary of v1 → v2: removed all `provider_bridge_ort.cc` changes, `TryGetProviderInfo_*()` calls, and the
 duplicate EP list. Everything now lives in `get_execution_providers.cc`  extended existing `ProviderInfo` struct with 
 library_base_name + `std::filesystem::exists()` check. 5 files, +206 lines, no side effects.
 
 Take your time, happy to clarify anything! 
 
 @yuslepukhin

yuslepukhin:
The potential issue with the whole approach is that we are moving to plug-in EPs. Even CUDA is moving to an ABI based plug-in approach. This means, in the future, we may not have any compiled in (except CPU) or traditional EPs.
Furthermore, we will have no idea about EPs that customers build in-house bc they will just work, and the new API will have no idea about their existence or the names of their libs.

PR #27699, opened 2026-03-17
fix potential out of boundary issue when initializer a SVMClassifier

xadupre:
### Description
If the ONNX file is malformed, it could lead to an incorrect memory access. This change enforces that does not happen.

### Motivation and Context
security issue

### DIFF
--- a/onnxruntime/core/providers/cpu/ml/svmclassifier.cc
+++ b/onnxruntime/core/providers/cpu/ml/svmclassifier.cc
@@ -72,6 +72,40 @@ SVMClassifier::SVMClassifier(const OpKernelInfo& info)
   ORT_ENFORCE(classlabels_strings_.size() > 0 || classlabels_ints_.size() > 0);
   ORT_ENFORCE(proba_.size() == probb_.size());
   ORT_ENFORCE(coefficients_.size() > 0);
+
+  // Validate attribute array sizes against the declared dimensions to prevent
+  // out-of-bounds reads from crafted models.
+  if (mode_ == SVM_TYPE::SVM_SVC) {
+    // SVC mode: coefficients layout is [class_count - 1, vector_count]
+    const size_t expected_coefficients = static_cast<size_t>(class_count_ - 1) * static_cast<size_t>(vector_count_);
+    ORT_ENFORCE(coefficients_.size() >= expected_coefficients,
+                "coefficients attribute size (", coefficients_.size(),
+                ") is smaller than expected (", expected_coefficients,
+                ") for the given class_count and vector_count.");
+
+    // rho needs one entry per classifier pair: class_count * (class_count - 1) / 2
+    const size_t num_classifiers = static_cast<size_t>(class_count_) * static_cast<size_t>(class_count_ - 1) / 2;
+    ORT_ENFORCE(rho_.size() >= num_classifiers,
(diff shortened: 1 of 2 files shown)


yuslepukhin:
The vulnerability report describes three independent OOB access sites. The current fix only addresses one of them. See comments.

Design issue: validation at inference time vs. construction time
The current fix validates at inference time (inside ComputeImpl), incurring overhead on every inference call. Since rho_, coefficients_, proba_, and probb_ are model attributes (immutable after construction), all validation should happen in the constructor instead. This:
- Catches malformed models at load time (fail-fast)
- Eliminates per-inference overhead
- Is the standard pattern used elsewhere in ORT

onnxruntime/core/providers/cpu/ml/svmclassifier.cc
```diff
@@ -287,16 +287,25 @@ Status SVMClassifier::ComputeImpl(OpKernelContext& ctx,
...
+            ORT_ENFORCE(i_coeff_row_offset + SafeInt<size_t>(start_index_j + class_j_support_count) <= coefficients_.size(),
+                        "The number of vectors per class is not consistent with the number of vectors in the model.");
+            const float* val1 = &(coefficients_[i_coeff_row_offset + SafeInt<size_t>(start_index_j)]);
+            const float* val2 = &(cur_kernels[onnxruntime::narrow<size_t>(start_index_j)]);
+            for (int64_t m = 0; m < class_j_support_count; ++m, ++val1, ++val2) {
+              sum += *val1 * *val2;
+            }
+          }
 
           sum += rho_[classifier_idx++];
```
yuslepukhin, line 335: >sum += rho_[classifier_idx++]; [](http://example.com/codeflow?start=10&length=30)

classifier_idx iterates from 0 to num_classifiers - 1 (where num_classifiers = class_count_ * (class_count_ - 1) / 2).
If rho_ has fewer entries than num_classifiers, this is an OOB heap read. The test case actually has rho = {0.1f} with 3 classes (needing 3 entries), but execution hits the coefficients_ check first and never reaches the rho_ access. #Resolved

yuslepukhin:
>           float val1 = sigmoid_probability(classifier_scores[index], proba_[index], probb_[index]);

index iterates from 0 to num_classifiers - 1.
If proba_ or probb_ are non-empty but have fewer than num_classifiers entries, this is an OOB heap read.
There is no validation at all for these arrays. #Resolved

---
Refers to: onnxruntime/core/providers/cpu/ml/svmclassifier.cc:327 in 74f0639. [](commit_id = 74f0639e721661e013e747b8bd694bf075ccbd07, deletion_comment = True)

onnxruntime/test/providers/cpu/ml/svmclassifier_test.cc
```diff
@@ -263,5 +263,35 @@ TEST(MLOpTest, SVMClassifierLinear) {
   test.Run();
 }
 
+TEST(MLOpTest, SVMClassifierOutOfBoundary) {
```
yuslepukhin: The single test case (SVMClassifierOutOfBoundary) triggers the coefficients_ check but doesn't independently test rho_, proba_, or probb_ undersizing.

Please, add more tests. #Resolved

PR #27701, opened 2026-03-17
[CUDA] Update Pad kernel to support versioning and add tests for Opset 19

### LINKED ISSUE #26393: [Performance] Missing CUDA Kernal for Pad, MaxPool, ConvTranspose opset 19-23
### Describe the issue

Starting Opset 19 I have this missing cuda kernel 
```
[I:onnxruntime:, cuda_execution_provider.cc:2922 GetCapability] CUDA kernel not found in registries for Op type: Pad node name: node_pad_3
```
And starting opset 22 I have both two more warning 
```
[I:onnxruntime:, cuda_execution_provider.cc:2922 GetCapability] CUDA kernel not found in registries for Op type: ConvTranspose node name: node_convolution
[I:onnxruntime:, cuda_execution_provider.cc:2922 GetCapability] CUDA kernel not found in registries for Op type: ConvTranspose node name: node_convolution_
1
```

I don't know If there much difference introduced by Pad op19+ and MaxPool, ConvTranspose op23+ but it would be much appreciated to have kernel for thoses Ops

### To reproduce

NA

### Urgency

_No response_

### Platform

Linux

### OS Version

Ubuntu 24.04

### ONNX Runtime Installation

Released Package

1.23.1

### ONNX Runtime API

Python

### Architecture

X64

### Execution Provider

CUDA

CUDA 12.9

### Model File

_No response_

No

lukas-folle-snkeos:
### Description
This pull request updates the CUDA kernel registration for the `Pad` operator in ONNX Runtime to use versioned kernel registration for opsets 18 through 23, and adds test coverage for CUDA-specific behavior. 

### Motivation and Context
Fixes #26393

### DIFF
--- a/onnxruntime/core/providers/cuda/cuda_execution_provider.cc
+++ b/onnxruntime/core/providers/cuda/cuda_execution_provider.cc
@@ -1441,10 +1441,10 @@ class ONNX_OPERATOR_TYPED_KERNEL_CLASS_NAME(kCudaExecutionProvider, kOnnxDomain,
 class ONNX_OPERATOR_TYPED_KERNEL_CLASS_NAME(kCudaExecutionProvider, kOnnxDomain, 18, int64_t, ReduceMax);
 class ONNX_OPERATOR_KERNEL_CLASS_NAME(kCudaExecutionProvider, kOnnxDomain, 18, ScatterElements);
 class ONNX_OPERATOR_KERNEL_CLASS_NAME(kCudaExecutionProvider, kOnnxDomain, 18, ScatterND);
-class ONNX_OPERATOR_TYPED_KERNEL_CLASS_NAME(kCudaExecutionProvider, kOnnxDomain, 18, float, Pad);
-class ONNX_OPERATOR_TYPED_KERNEL_CLASS_NAME(kCudaExecutionProvider, kOnnxDomain, 18, double, Pad);
-class ONNX_OPERATOR_TYPED_KERNEL_CLASS_NAME(kCudaExecutionProvider, kOnnxDomain, 18, MLFloat16, Pad);
-class ONNX_OPERATOR_TYPED_KERNEL_CLASS_NAME(kCudaExecutionProvider, kOnnxDomain, 18, bool, Pad);
+class ONNX_OPERATOR_VERSIONED_TYPED_KERNEL_CLASS_NAME(kCudaExecutionProvider, kOnnxDomain, 18, 23, float, Pad);
+class ONNX_OPERATOR_VERSIONED_TYPED_KERNEL_CLASS_NAME(kCudaExecutionProvider, kOnnxDomain, 18, 23, double, Pad);
+class ONNX_OPERATOR_VERSIONED_TYPED_KERNEL_CLASS_NAME(kCudaExecutionProvider, kOnnxDomain, 18, 23, MLFloat16, Pad);
+class ONNX_OPERATOR_VERSIONED_TYPED_KERNEL_CLASS_NAME(kCudaExecutionProvider, kOnnxDomain, 18, 23, bool, Pad);
 class ONNX_OPERATOR_VERSIONED_TYPED_KERNEL_CLASS_NAME(kCudaExecutionProvider, kOnnxDomain, 18, 18, float, Resize);
 class ONNX_OPERATOR_VERSIONED_TYPED_KERNEL_CLASS_NAME(kCudaExecutionProvider, kOnnxDomain, 18, 18, double, Resize);
 class ONNX_OPERATOR_VERSIONED_TYPED_KERNEL_CLASS_NAME(kCudaExecutionProvider, kOnnxDomain, 18, 18, MLFloat16, Resize);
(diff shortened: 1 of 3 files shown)


tianleiwu:
## Summary

The direction is good: moving CUDA Pad away from a single opset-18 registration and adding a CUDA-focused regression test is the right shape of change. However, the current registration range is not safe yet because it makes CUDA eligible for opset-19+ `Pad(mode="wrap")` nodes even though the CUDA kernels still implement only constant/reflect/edge, and it also stops short of the newer Pad schemas that already exist in-tree.

---

## Review

### 1. CUDA Pad Versioning (`[onnxruntime/core/providers/cuda/tensor/pad.cc](https://github.com/microsoft/onnxruntime/blob/8107200dfd6dede7b8284edcdef1cab17df68525/onnxruntime/core/providers/cuda/tensor/pad.cc)`)

**Positive**:
- Switching Pad from a single opset-18 registration to a versioned registration is the right mechanism for broadening CUDA coverage, and the matching updates in `[onnxruntime/core/providers/cuda/cuda_execution_provider.cc](https://github.com/microsoft/onnxruntime/blob/8107200dfd6dede7b8284edcdef1cab17df68525/onnxruntime/core/providers/cuda/cuda_execution_provider.cc)` keep the registry declarations consistent.

**Concern**:
- ⚠️ **`wrap` mode is now advertised without a CUDA implementation**: `PadBase` parses `mode="wrap"` into `Mode::Wrap` in `[onnxruntime/core/providers/cpu/tensor/padbase.h](https://github.com/microsoft/onnxruntime/blob/8107200dfd6dede7b8284edcdef1cab17df68525/onnxruntime/core/providers/cpu/tensor/padbase.h)`, but the CUDA kernels only define three modes and only dispatch `pad_mode` values `0`, `1`, and `2` in `[onnxruntime/core/providers/cuda/tensor/pad_impl.cu](https://github.com/microsoft/onnxruntime/blob/8107200dfd6dede7b8284edcdef1cab17df68525/onnxruntime/core/providers/cuda/tensor/pad_impl.cu)`. With this PR, opset-19 through opset-23 Pad nodes become eligible for CUDA placement, which includes `mode="wrap"` per the in-tree ONNX schema. The added test change only excludes CUDA from wrap tests; it does not stop real models from being assigned to CUDA here. That means a wrap Pad can now be placed on CUDA and hit an unsupported mode path at runtime.
  ```cpp
  // pad_impl.cu currently only dispatches 3 modes.
  switch (pad_mode) {
    case 0: /* constant */ break;
    case 1: /* reflect */ break;
    case 2: /* edge */ break;
  }
  ```

### 2. Schema Coverage And Tests (`[onnxruntime/test/providers/cpu/tensor/pad_test.cc](https://github.com/microsoft/onnxruntime/blob/8107200dfd6dede7b8284edcdef1cab17df68525/onnxruntime/test/providers/cpu/tensor/pad_test.cc)`)

**Positive**:
- The new CUDA-only MLFloat16 test is useful and directly exercises the motivating scenario instead of relying on broader typed coverage to catch the registration change indirectly.

**Concern**:
- ⚠️ **Coverage still stops before the latest Pad schemas**: even if wrap placement is fixed, the new CUDA registration ends at opset 23 while this tree already carries Pad schemas for opsets 24 and 25 in `[cmake/external/onnx/onnx/defs/operator_sets.h](https://github.com/microsoft/onnxruntime/blob/8107200dfd6dede7b8284edcdef1cab17df68525/cmake/external/onnx/onnx/defs/operator_sets.h)` and `[cmake/external/onnx/onnx/defs/tensor/defs.cc](https://github.com/microsoft/onnxruntime/blob/8107200dfd6dede7b8284edcdef1cab17df68525/cmake/external/onnx/onnx/defs/tensor/defs.cc)`. CPU already registers Pad separately for `24` and `25` in `[onnxruntime/core/providers/cpu/tensor/pad.cc](https://github.com/microsoft/onnxruntime/blob/8107200dfd6dede7b8284edcdef1cab17df68525/onnxruntime/core/providers/cpu/tensor/pad.cc)`, so a non-wrap float/double/MLFloat16/bool Pad in an opset-24 or opset-25 model would still miss CUDA after this PR. I’d expect the CUDA registration and tests to cover the newer schema bands too once the wrap-mode issue is addressed.

---

## Summary of Concerns

| 1 | High | CUDA Pad versioning | Registers opset-19+ Pad on CUDA even though `wrap` is part of those schemas and is not implemented in the CUDA kernels. |
| 2 | Suggestion | Schema coverage and tests | Registration only goes through opset 23, leaving opset 24/25 Pad models without CUDA coverage. |

---

## Verdict

REQUEST CHANGES — the current versioned registration would expose unsupported `wrap` Pad nodes to CUDA placement, so the PR is not safe to merge as-is.

tianleiwu:
@lukas-folle-snkeos, I added a new PR https://github.com/microsoft/onnxruntime/pull/27708 to extend Pad up to opset 25. Your commit and contribution is kept. Thanks.

lukas-folle-snkeos:
@tianleiwu thanks for getting to this so quickly. I am closing this PR in favor of your #27708 
Thanks for keeping my contribution!

PR #27710, opened 2026-03-17
[ Cuda] ConvTranspose-22

tianleiwu:
## Description

This PR extends the CUDA `ConvTranspose` operator registration to support ONNX opset 22. The CUDA implementation already shares the same attribute and output-shape handling used by earlier supported opsets, so this change primarily exposes the existing kernel path for opset 22 and adds regression coverage to keep that path working.

### CUDA Kernel Registration

| File | Change |
|------|--------|
| `onnxruntime/core/providers/cuda/nn/conv_transpose.cc` | Split `ConvTranspose` kernel registration into `1-10`, `11-21`, and `22` so the CUDA kernel can be selected for opset 22. |
| `onnxruntime/core/providers/cuda/cuda_execution_provider.cc` | Added matching provider-side kernel declarations and registry entries for `ConvTranspose(22)` on CUDA. |
| `onnxruntime/core/providers/cuda/cuda_nhwc_kernels.cc` | Added matching NHWC CUDA declarations and registry entries for `ConvTranspose(22)`. |

### Test Coverage

| File | Change |
|------|--------|
| `onnxruntime/test/providers/cpu/nn/conv_transpose_op_test.cc` | Added a CUDA-only opset 22 regression test that validates `ConvTranspose` with `output_shape`. |
| `onnxruntime/test/providers/cuda/nhwc/conv_transpose_test.cc` | Updated existing NHWC comparison coverage to instantiate `ConvTranspose` at opset 22. |

## Testing

- Built the touched translation units successfully with targeted `ninja` object builds:
  - `onnxruntime/core/providers/cuda/cuda_execution_provider.cc`
  - `onnxruntime/core/providers/cuda/cuda_nhwc_kernels.cc`
  - `onnxruntime/core/providers/cuda/nn/conv_transpose.cc`
  - `onnxruntime/test/providers/cpu/nn/conv_transpose_op_test.cc`
  - `onnxruntime/test/providers/cuda/nhwc/conv_transpose_test.cc`
- Ran `git diff --check` to confirm the patch is formatting-clean.
- A full `onnxruntime_test_all` / gtest runtime pass was not completed locally because the build tree regenerated and expanded into a broad rebuild; runtime verification of the new CUDA and NHWC tests should still be run in CI or a focused local test build.

## Motivation and Context
Related issue: https://github.com/microsoft/onnxruntime/issues/26393

The ONNX `ConvTranspose-22` schema keeps the same core padding and output-shape semantics used by the earlier CUDA-supported opsets, but CUDA registration stopped at opset 11. That meant models using `ConvTranspose` at opset 22 could miss CUDA kernel assignment even though the implementation path was already compatible.

This PR closes that gap by updating kernel registration and test coverage without changing the underlying CUDA compute logic.

## Checklist

- [x] Tests added/updated
- [ ] Documentation updated (if applicable)
- [x] No breaking changes (or documented in description)
- [ ] CI passes


AzHicham:
Up I really need this feature in a nightly build :) same for operator max pool

titaiwangms (approved):

titaiwangms:
## Multi-reviewer synthesis (readability + code + critical + deep)

No Critical issues. Registration split is mechanically clean; opset-22 test math hand-verified (all 14 expected values exact). Findings prioritized below.

### Major

**1. BFloat16 not registered for ConvTranspose-22.**
The only schema delta from ConvTranspose-11 → 22 is adding `bfloat16` to `T` (proven via the bundled ONNX `Changelog.md` and the down-converter restriction at `cmake/external/onnx/onnx/version_converter/convert.h:811`: `TypeRestriction("ConvTranspose", 22, 21, bfloat16_not_allowed)`). The PR registers `float / double / MLFloat16` only, while the companion Conv-22 CUDA kernel *does* register BFloat16.

Note: the CPU EP also doesn't register BFloat16 for ConvTranspose-22 (`onnxruntime/core/providers/cpu/nn/conv_transpose.cc:41-45` is `float`-only despite a "Opset 22 starts to support bfloat16" comment), so this is a consistent pre-existing gap rather than a CUDA regression — but it does mean the PR title overstates what is delivered (the only feature opset-22 actually adds is unwired). The cuDNN-frontend path in `conv_transpose.cc` already has bf16 math-type handling, though its compute-type setup may need to align with Conv's (`conv.cc:551`) before enabling.

**Recommendation:** either (a) register `BFloat16` for `kOnnxDomain` ConvTranspose-22 now (preferable — this is the only opset-22 capability gain), or (b) explicitly note in the PR description that BFloat16 ConvTranspose remains unsupported pending a follow-up that also covers the CPU EP.

### Minor

**2. NHWC test was migrated rather than extended** (`onnxruntime/test/providers/cuda/nhwc/conv_transpose_test.cc:25`, opset 14 → 22). The bump exercises the new `22+` registration but loses coverage of the newly split `[11, 21]` range. Best of both: keep one case at opset 14 (or 21) and add a separate opset-22 case so both new ranges are validated.

**3. New CUDA-only test only covers `float`** (`conv_transpose_op_test.cc:474`). `MLFloat16` and `double` are newly registered for opset 22 but uncovered. Add at least one of each (CUDA EP only, CPU fallback disabled).

**4. CUDA-specific test in CPU test file.** `ConvTranspose_2D_OutputShape_2_OpSet22_CUDA` depends on `DefaultCudaExecutionProvider()` + `kOrtSessionOptionsDisableCPUEPFallback`. Not unprecedented in this file, but moving to `test/providers/cuda/` (or a brief comment explaining the placement) would aid discovery.

### Nits

- **No comment on the BFloat16 asymmetry** between Conv-22 (registered) and ConvTranspose-22 (not registered) — a one-line comment near the macro would help future maintainers know if this is deferred / infeasible / oversight.
- **New test data duplicates `ConvTranspose_2D_OutputShape_2` verbatim** with no cross-reference. A 1-line comment ("same inputs as X; verifies opset-22 routes to CUDA kernel") clarifies intent.
- **`output_shape = {1, 1, 1, 14}` is non-canonical**: ONNX spec says output_shape should exclude batch/channel; portable form is `{1, 14}`. ORT tolerates both, and the new test inherits this from the cloned test, so not a bug — pre-existing cleanup opportunity.

### Praise

- Symmetric `[1,10] / [11,21] / 22` split via the existing macro pattern; class declarations and `BuildKernelCreateInfo` entries are balanced (verified 9↔9 for kOnnxDomain, 6↔6 for NHWC, no remaining unbounded `11+` entries).
- `OperatorKernels.md` lists the actually-registered type set rather than the broader schema-allowed set — no overpromising.
- Replacement comment in `conv_transpose.cc` ("ConvTranspose opsets 11-22 share the same CUDA implementation for the currently supported types") is more accurate than the old stub.

---
*Reviewed by a cross-family reviewer team (Claude readability + GPT code & critical + Claude-high deep). Math and registration invariants independently verified; deep-reviewer was tie-breaker on the BFloat16 framing.*

PR #27713, opened 2026-03-17
[Plugin EP] Add plugin EP APIs to retrieve ONNX operator schemas

### LINKED ISSUE #27680: [Feature Request] Expose OpSchema API for EP Plugin to enable proper TypeConstraint registration

When implementing a custom Execution Provider using the EP Plugin API (ORT 1.24+), there is currently no way to access ONNX OpSchema information. This makes it impossible to correctly set TypeConstraints when registering custom op kernels via `KernelDefBuilder::AddTypeConstraint()`.

## Background

The EP Plugin API provides `AddTypeConstraint(const char* arg_name, const OrtDataType* data_type)` for kernel registration, but the `arg_name` parameter must exactly match the TypeConstraint name defined in the ONNX OpSchema (e.g., "T", "T1", "T2").

Without access to OpSchema, plugin EPs must either:
1. Skip TypeConstraint entirely, leading to potential kernel mismatch issues
2. Hardcode TypeConstraint names for each op/domain/version combination  

## Current Behavior

In the internal VitisAI provider (`onnxruntime/core/providers/vitisai/imp/global_api.cc`), the schema is accessed via:

```cpp
auto schema = Provider_GetHost()->GetSchema(op->GetName(op), op->GetStartVersion(op), domain->domain_);
auto input_name = schema->inputs__GetName(i);
if (schema->typeConstraintMap().count(schema->inputs__GetTypeStr(i))) {
    input_name = schema->inputs__GetTypeStr(i);
}
def_builder->TypeConstraint(input_name.c_str(), DataTypeImpl::GetTensorTypeFromOnnxType(input_type));
```

However, `Provider_GetHost()` is an internal API not available to EP Plugins.

## Problem

Without proper TypeConstraint registration:
- ORT's `MatchKernelDefTypes()` function returns `true` for any type when `kernel_type_constraints` is empty
- This can cause incorrect kernel selection, leading to **data type mismatches and precision issues**
- For example, an int8 kernel might be selected for a float16 node, causing incorrect computation results

## Use Case

VitisAI EP Plugin registers QDQ (QuantizeLinear/DequantizeLinear) ops with different data types (int8, uint8, int16, uint16, float, etc.). Without proper TypeConstraint:
- Multiple kernels with the same op name but different data types are registered
- ORT cannot distinguish between them during kernel selection
- Wrong kernel may be selected, causing precision loss

## Environment

- ONNX Runtime version: 1.24+
- EP Plugin API

## Related

- PR #21688: Added OpSchema and TypeConstraint handling for internal VitisAI provider
- PR #26206: [EP ABI] Initial support for kernel-based EPs
- This request extends similar capability to EP Plugin API

## Motivation

This feature would:
1. Enable correct kernel matching for plugin EPs
2. Prevent precision issues caused by type mismatches
3. Align EP Plugin capabilities with internal provider capabilities

adrianlizarraga:
### Description

Adds C/C++ APIs to the `OrtEpApi` that allow plugin EPs to query ONNX operator schemas from ORT's global schema registry. This enables EPs to programmatically discover operator metadata (input/output names, type constraints, allowed types, since_version) needed to correctly build kernel definitions with proper type constraints.

### Motivation

Resolves #27680. Plugin EPs must provide exact type constraint names (e.g., `"T"`, `"T1"`) and allowed types when calling `KernelDefBuilder::AddTypeConstraint()`. Without schema access, EPs must either hard-code these names or skip type constraints entirely, leading to potentially incorrect kernel selection and data type mismatches at runtime.

**Why can't an EP library just link to its own ONNX library?** The ONNX `OpSchemaRegistry` is a Meyers singleton (`static` local in `Instance()`). Each shared library gets its own copy of that static variable: on Windows each DLL is isolated by default, on macOS two-level namespaces have the same effect, and on Linux behavior depends on `dlopen` flags (`RTLD_LOCAL` isolates, `RTLD_GLOBAL` creates unpredictable interposition). Even when isolation doesn't occur, the EP's registry would lack ORT's contrib and internal schemas, and version mismatches between the EP's ONNX library and ORT's vendored copy could cause silent divergence. A C API through ORT is the only reliable, portable way to query the schemas ORT actually uses.

### Changes

**New opaque types:**
- `OrtOpSchema` — owning opaque struct wrapping an `onnx::OpSchema*` with precomputed type constraint data. Allocated by `GetOpSchema`, released by `ReleaseOpSchema`.
- `OrtOpSchemaTypeConstraint` — non-owning opaque entity representing a single type constraint (e.g., "T"). Lifetime is tied to the parent `OrtOpSchema`. Each constraint carries its name, allowed types, and input/output index mappings.

**New C APIs added to `OrtEpApi` (Version 1.25, 15 functions):**

| Function | Description |
|---|---|
| `GetOpSchema` | Look up a schema by name, max opset version, and domain. Accepts `""` or `"ai.onnx"` for standard ONNX ops, `"ai.onnx.ml"` for ML ops, `"com.microsoft"` for contrib ops. |
| `ReleaseOpSchema` | Release an `OrtOpSchema` allocated by `GetOpSchema`. |
| `OpSchema_GetSinceVersion` | Get the opset version that introduced the schema. |
| `OpSchema_GetNumInputs` / `GetNumOutputs` | Input/output counts. |
| `OpSchema_GetInputName` / `GetOutputName` | Formal parameter names. |
| `OpSchema_GetInputTypeConstraint` / `GetOutputTypeConstraint` | Get the type constraint for a given input/output (O(1) lookup). Returns `nullptr` if the input/output has no type constraint. Shared constraints return the same pointer (pointer identity = shared type). |
| `OpSchema_GetTypeConstraintCount` | Number of unique type constraints. |
| `OpSchema_GetTypeConstraint` | Get the i-th type constraint by index. |
| `OpSchemaTypeConstraint_GetTypeParamName` | Get the type parameter name (e.g., `"T"`, `"T1"`). |
| `OpSchemaTypeConstraint_GetAllowedTypes` | Get the allowed type strings (e.g., `"tensor(float)"`). |
| `OpSchemaTypeConstraint_GetInputIndices` | Get input indices using this constraint. |
| `OpSchemaTypeConstraint_GetOutputIndices` | Get output indices using this constraint. |

**C++ wrappers:**
- `Ort::OpSchema` — owning wrapper around `OrtOpSchema*` (move-only, auto-releases).
- `Ort::ConstOpSchemaTypeConstraint` — non-owning wrapper around `const OrtOpSchemaTypeConstraint*`.
- `Ort::GetOpSchema()` — free function to query the registry.

**Design highlights:**
- Type constraints are eagerly precomputed during `GetOpSchema` — all subsequent accessors are O(1) with no allocation.
- `GetInputTypeConstraint`/`GetOutputTypeConstraint` return the full constraint object directly (not just a string), enabling a 2-call workflow: `GetInputTypeConstraint(0)` → `GetAllowedTypes()`.
- Pointer identity: inputs sharing a constraint (e.g., both inputs of `Add` use `"T"`) return the same `OrtOpSchemaTypeConstraint*`.
- Domain `"ai.onnx"` is normalized to `""` (the canonical ONNX domain) for transparent lookup.

**Tests:** 14 unit tests covering known/unknown ops, version boundaries, wrong domains, `"ai.onnx"` alias, schema properties (Add, Relu, LSTM), type constraint access, pointer identity for shared constraints, out-of-range errors, and the input→constraint→allowed-types workflow.

### Files

| File | Description |
|---|---|
| `onnxruntime/core/session/abi_opschema.h` | Internal struct definitions for `OrtOpSchemaTypeConstraint` and `OrtOpSchema`. |
| `include/.../onnxruntime_ep_c_api.h` | Public C API: function signatures, doc comments, opaque type declarations. |
| `onnxruntime/core/session/plugin_ep/ep_api.h` | Internal function declarations. |
| `onnxruntime/core/session/plugin_ep/ep_api.cc` | Implementation of all 15 functions + API struct initializer. |
| `include/.../onnxruntime_cxx_api.h` | C++ wrapper class declarations. |
| `include/.../onnxruntime_cxx_inline.h` | C++ wrapper inline implementations. |
| `onnxruntime/test/framework/ep_plugin_provider_test.cc` | Unit tests. |

### DIFF
--- a/include/onnxruntime/core/session/onnxruntime_cxx_inline.h
+++ b/include/onnxruntime/core/session/onnxruntime_cxx_inline.h
@@ -4036,6 +4036,7 @@ inline std::vector<size_t> OpSchemaTypeConstraintImpl<T>::GetInputIndices() cons
   const size_t* indices = nullptr;
   size_t count = 0;
   ThrowOnError(GetEpApi().OpSchemaTypeConstraint_GetInputIndices(this->p_, &indices, &count));
+  if (count == 0) return {};
   return std::vector<size_t>(indices, indices + count);
 }
 
@@ -4044,6 +4045,7 @@ inline std::vector<size_t> OpSchemaTypeConstraintImpl<T>::GetOutputIndices() con
   const size_t* indices = nullptr;
   size_t count = 0;
   ThrowOnError(GetEpApi().OpSchemaTypeConstraint_GetOutputIndices(this->p_, &indices, &count));
+  if (count == 0) return {};
   return std::vector<size_t>(indices, indices + count);
 }
 }  // namespace detail


onnxruntime/test/framework/ep_plugin_provider_test.cc
```diff
@@ -779,4 +779,66 @@ TEST(PluginExecutionProviderTest, IsConcurrentRunSupported) {
...
+  Ort::ConstOpSchema schema = Ort::GetOpSchema("Relu", 20, "com.nonexistent.domain");
+  ASSERT_EQ(static_cast<const OrtOpSchema*>(schema), nullptr);
+}
+
+// Test ConstOpSchema methods on the "Add" operator schema (2 inputs, 1 output).
+TEST(OpSchemaCxxApiTest, AddSchemaProperties) {
+  Ort::ConstOpSchema schema = Ort::GetOpSchema("Add", 20, "");
+  ASSERT_NE(static_cast<const OrtOpSchema*>(schema), nullptr);
+
+  EXPECT_EQ(schema.GetSinceVersion(), 14);
```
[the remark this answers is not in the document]
adrianlizarraga: 14 is guaranteed when retrieving an opschema for version 20. This won't change in future onnx releases since we're already at opset 26+.

onnxruntime/core/session/plugin_ep/ep_api.cc
```diff
@@ -806,6 +807,125 @@ ORT_API_STATUS_IMPL(GetEnvConfigEntries, _Outptr_ OrtKeyValuePairs** config_entr
...
+                    _In_ const char* domain, _Outptr_result_maybenull_ const OrtOpSchema** out_schema) {
+  API_IMPL_BEGIN
+  ORT_API_RETURN_IF(name == nullptr, ORT_INVALID_ARGUMENT, "name must not be null");
+  ORT_API_RETURN_IF(domain == nullptr, ORT_INVALID_ARGUMENT, "domain must not be null");
+  ORT_API_RETURN_IF(out_schema == nullptr, ORT_INVALID_ARGUMENT, "out_schema must not be null");
+
+  const auto* schema = ONNX_NAMESPACE::OpSchemaRegistry::Instance()->GetSchema(
+      name, max_inclusive_version, domain);
+  *out_schema = reinterpret_cast<const OrtOpSchema*>(schema);
+  return nullptr;
```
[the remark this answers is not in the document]
adrianlizarraga: ep_api.cc is never compiled in minimal builds. The CMake file 
onnxruntime_session.cmake:25-33 explicitly excludes all plugin_ep/*.* files when onnxruntime_MINIMAL_BUILD is set.

include/onnxruntime/core/session/onnxruntime_cxx_api.h
```diff
@@ -3515,6 +3515,52 @@ struct KernelRegistry : detail::Base<OrtKernelRegistry> {
...
+  size_t GetNumOutputs() const;
+
+  ///< Wraps OrtEpApi::OpSchema_GetOutputName
+  std::string GetOutputName(size_t index) const;
+
+  ///< Wraps OrtEpApi::OpSchema_GetOutputTypeStr
+  std::string GetOutputTypeStr(size_t index) const;
+
+  ///< Wraps OrtEpApi::OpSchema_HasTypeConstraint
+  bool HasTypeConstraint(const char* type_str) const;
```
tianleiwu: This API for type constarint is not completed. For example LSTM operator has T and T1
```
Type Constraints
T in ( tensor(bfloat16), tensor(double), tensor(float), tensor(float16) ): Constrain input and output types to float tensors.
T1 in ( tensor(int32) ): Constrain seq_lens to integer tensor.
```

It also maps each input/output to at most one type constraint.

Ideally, we define a struct like {type_str, data_types, input_indices, output_indices}, we only need one API to return a list of such struct.

adrianlizarraga: Thanks. Yes it is incomplete as I believe this was the minimum set of apis used by vitisai ep. That said, we might as well do it all at once and do it right.

I'll update.

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
```diff
@@ -1444,6 +1444,131 @@ struct OrtEpApi {
    * \since Version 1.24
    */
   ORT_API2_STATUS(GetEnvConfigEntries, _Outptr_ OrtKeyValuePairs** config_entries);
+
+  /** \brief Get an ONNX operator schema from the global registry.
```
edgchen1: if it's just for ONNX ops, maybe name it `GetOnnxOpSchema()`?

onnxruntime/core/session/plugin_ep/ep_api.cc
```diff
@@ -806,6 +807,125 @@ ORT_API_STATUS_IMPL(GetEnvConfigEntries, _Outptr_ OrtKeyValuePairs** config_entr
...
+
+ORT_API_STATUS_IMPL(OpSchema_HasTypeConstraint, _In_ const OrtOpSchema* schema, _In_ const char* type_str,
+                    _Out_ bool* out) {
+  API_IMPL_BEGIN
+  ORT_API_RETURN_IF(schema == nullptr, ORT_INVALID_ARGUMENT, "schema must not be null");
+  ORT_API_RETURN_IF(type_str == nullptr, ORT_INVALID_ARGUMENT, "type_str must not be null");
+  ORT_API_RETURN_IF(out == nullptr, ORT_INVALID_ARGUMENT, "out must not be null");
+
+  const auto* onnx_schema = reinterpret_cast<const ONNX_NAMESPACE::OpSchema*>(schema);
+  *out = onnx_schema->typeConstraintMap().count(type_str) > 0;
```
edgchen1: we can finally use `unordered_map::contains()`

onnxruntime/test/framework/ep_plugin_provider_test.cc
```diff
@@ -779,4 +779,69 @@ TEST(PluginExecutionProviderTest, IsConcurrentRunSupported) {
 #endif  // !defined(ORT_NO_EXCEPTIONS)
 }
 
+// Tests for the Ort::ConstOpSchema C++ wrapper API and Ort::GetOpSchema free function.
+// These test the C++ layer over the OrtEpApi OpSchema functions using well-known ONNX operator schemas
+// from the global ONNX schema registry.
+
+TEST(OpSchemaCxxApiTest, GetOpSchema_KnownOp_ReturnsNonNull) {
+  Ort::ConstOpSchema schema = Ort::GetOpSchema("Relu", 20, "");
```
edgchen1: should we document that the empty string is the ONNX domain? does "ai.onnx" also work?

onnxruntime/core/session/plugin_ep/ep_api.cc
```diff
@@ -806,6 +809,248 @@ ORT_API_STATUS_IMPL(GetEnvConfigEntries, _Outptr_ OrtKeyValuePairs** config_entr
...
+ORT_API_STATUS_IMPL(GetOpSchema, _In_ const char* name, _In_ int max_inclusive_version,
+                    _In_ const char* domain, _Outptr_result_maybenull_ OrtOpSchema** out_schema) {
+  API_IMPL_BEGIN
+  ORT_API_RETURN_IF(name == nullptr, ORT_INVALID_ARGUMENT, "name must not be null");
+  ORT_API_RETURN_IF(domain == nullptr, ORT_INVALID_ARGUMENT, "domain must not be null");
+  ORT_API_RETURN_IF(out_schema == nullptr, ORT_INVALID_ARGUMENT, "out_schema must not be null");
+
+  // Normalize "ai.onnx" to "" (the canonical ONNX domain used by the schema registry).
+  const char* lookup_domain = (strcmp(domain, kOnnxDomainAlias) == 0) ? kOnnxDomain : domain;
+
```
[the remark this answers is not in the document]
adrianlizarraga: add include

include/onnxruntime/core/session/onnxruntime_cxx_inline.h
```diff
@@ -3946,4 +3946,111 @@ inline Ort::KeyValuePairs GetEnvConfigEntries() {
...
+  }
+  return result;
+}
+
+template <typename T>
+inline std::vector<size_t> OpSchemaTypeConstraintImpl<T>::GetInputIndices() const {
+  const size_t* indices = nullptr;
+  size_t count = 0;
+  ThrowOnError(GetEpApi().OpSchemaTypeConstraint_GetInputIndices(this->p_, &indices, &count));
+  return std::vector<size_t>(indices, indices + count);
```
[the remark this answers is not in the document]
adrianlizarraga: Fixed.

include/onnxruntime/core/session/onnxruntime_cxx_inline.h
```diff
@@ -3946,4 +3946,111 @@ inline Ort::KeyValuePairs GetEnvConfigEntries() {
...
+}
+
+template <typename T>
+inline std::vector<size_t> OpSchemaTypeConstraintImpl<T>::GetOutputIndices() const {
+  const size_t* indices = nullptr;
+  size_t count = 0;
+  ThrowOnError(GetEpApi().OpSchemaTypeConstraint_GetOutputIndices(this->p_, &indices, &count));
+  return std::vector<size_t>(indices, indices + count);
```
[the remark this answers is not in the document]
adrianlizarraga: Fixed.

include/onnxruntime/core/session/onnxruntime_cxx_api.h
adrianlizarraga: updated

include/onnxruntime/core/session/onnxruntime_ep_c_api.h
adrianlizarraga: that registry also has ort contrib ops. I updated doc.

onnxruntime/test/framework/ep_plugin_provider_test.cc
adrianlizarraga: Yep, good catch. I also had to normalize "ai.onnx" into "" within the implementation.

include/onnxruntime/core/session/onnxruntime_cxx_api.h
```diff
@@ -3521,6 +3522,88 @@ struct KernelRegistry : detail::Base<OrtKernelRegistry> {
                    void* kernel_create_func_state);
 };
 
+namespace detail {
+/** \brief Non-owning wrapper around a `const OrtOpSchemaTypeConstraint*`.
```
edgchen1, line 3526: nit: the `detail::OpSchemaTypeConstraintImpl` class is not concerned with ownership, right? I think the comment is more applicable to `ConstOpSchemaTypeConstraint`, which already has the comment.

similar comment for `detail::OpSchemaImpl`

onnxruntime/test/framework/ep_plugin_provider_test.cc
```diff
@@ -779,4 +780,202 @@ TEST(PluginExecutionProviderTest, IsConcurrentRunSupported) {
...
+  ASSERT_NE(t1_ptr, nullptr) << "Expected to find type constraint 'T1'";
+
+  auto has_type = [](gsl::span<const std::string> types, const char* t) {
+    return std::find(types.begin(), types.end(), t) != types.end();
+  };
+
+  // T should include tensor(float) and tensor(double)
+  auto t_types = t_tc.GetAllowedTypes();
+  EXPECT_GT(t_types.size(), 0u);
+  EXPECT_TRUE(has_type(t_types, "tensor(float)")) << "Expected T to allow tensor(float)";
```
edgchen1, line 930: nit: could use `EXPECT_THAT(t_types, ::testing::Contains("tensor(float)"))`. it should output an informative error message on failure.
https://google.github.io/googletest/reference/matchers.html#container-matchers

onnxruntime/test/framework/ep_plugin_provider_test.cc
```diff
@@ -779,4 +780,202 @@ TEST(PluginExecutionProviderTest, IsConcurrentRunSupported) {
...
+
+// Test OpSchema methods on the "Add" operator (2 inputs, 1 output, shared constraint T).
+// Also tests pointer identity: inputs/output sharing a constraint return the same pointer.
+TEST(OpSchemaCxxApiTest, AddSchemaProperties) {
+  int opset_version = 20;
+  Ort::OpSchema schema = Ort::GetOpSchema("Add", opset_version, "");
+  ASSERT_NE(static_cast<OrtOpSchema*>(schema), nullptr);
+
+  // The "since version" will be <= to the opset version used to retrieve the schema.
+  EXPECT_LT(schema.GetSinceVersion(), opset_version + 1);
```
edgchen1, line 831:
```suggestion
  EXPECT_LE(schema.GetSinceVersion(), opset_version);
```

include/onnxruntime/core/session/onnxruntime_cxx_api.h
adrianlizarraga: yep, will address in a separate PR

onnxruntime/test/framework/ep_plugin_provider_test.cc
adrianlizarraga: Good point. I'll update separately

onnxruntime/test/framework/ep_plugin_provider_test.cc
adrianlizarraga: yep, will address in a separate PR

PR #27718, opened 2026-03-17
Fix OOB read in ConcatSliceElimination and UnsqueezeElimination

vraspar:
### Description
`GetSliceInfo` in `ConcatSliceElimination` returns true with empty axes/steps vectors when a Slice node omits optional inputs. 
The caller `FuseConcatSliceSubgraph` then accesses axes[0] and steps[0] without checking, reading uninitialized InlinedVector

  Root Cause

  `GetSliceInfo` did not populate ONNX spec defaults for omitted optional Slice inputs:

   - Opset v1: axes attribute is optional; if absent, vector stays empty
   - Opset ≥10: if axes input doesn't exist, the entire axes/steps parsing block is skipped

  Fix
  `concat_slice_elimination.cc:` Populate ONNX spec defaults before returning from GetSliceInfo. axes defaults to [0, 1, ...,
  len(starts)-1], steps defaults to all 1s. This matches how ORT's Slice kernel handles missing inputs in slice_helper.h.

  `unsqueeze_elimination.cc` (related variant): Added duplicate axis detection to prevent an iterator from reading past 
  tensor_proto.dims().cend() when axes contain duplicates. 

  Testing
   - Added `ConcatSliceEliminationTest_NoAxesSteps` regression test

### DIFF
--- a/onnxruntime/core/optimizer/concat_slice_elimination.cc
+++ b/onnxruntime/core/optimizer/concat_slice_elimination.cc
@@ -141,6 +141,20 @@ static bool GetSliceInfo(const Graph& graph,
   } else {
     return false;
   }
+
+  // Per ONNX Slice spec: if axes not provided, default to [0, 1, ..., len(starts)-1].
+  if (axes.empty()) {
+    axes.reserve(starts.size());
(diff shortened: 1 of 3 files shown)


vraspar:
Closing the PR in favor of  https://github.com/microsoft/onnxruntime/pull/27638

PR #27719, opened 2026-03-18
[CUDA] Refactor CUDA helpers or ops for plugin EP

tianleiwu:
## Description

This PR refactors a set of CUDA provider helpers and kernel call sites so more of the remaining CUDA ops can build cleanly when the CUDA EP is compiled as a plugin. The main goal is to reduce direct dependencies on framework-internal CUDA EP types such as `CUDAExecutionProvider` and `CudaStream`, and to move reusable CUDA type/handle helpers into header-visible utilities that are available on both sides of the plugin boundary.

These changes follow the plugin EP guidance, where the next stage of enablement is to remove kernel assumptions about provider-only infrastructure and rely on stream-handle based access instead. Along the way, the PR also fixes a few small compatibility issues in contrib CUDA kernels that surfaced while working through the remaining excluded ops.

### Shared CUDA Helper Refactoring

| File | Change |
|------|--------|
| `onnxruntime/core/providers/cuda/cuda_common_type_helpers.h` | Adds header-only CUDA type conversion and string helper utilities so `.cu` and plugin builds can consume them without relying on `cuda_common.cc`. |
| `onnxruntime/core/providers/cuda/cuda_common.h` | Includes `core/util/math.h` for local half conversion helpers and pulls in the new header-only helper definitions. |
| `onnxruntime/core/providers/cuda/cuda_common.cc` | Removes the moved helper implementations and keeps runtime GEMM option initialization focused on shared state. |
| `onnxruntime/core/providers/cuda/shared_inc/accumulation_type.h` | Adds a default accumulation type mapping so unsupported specializations no longer fail at compile time. |
| `onnxruntime/contrib_ops/cuda/math/gemm_float8.cu` | Switches float8 GEMM code to consume the new shared CUDA type helper header. |

### Stream and Handle Access Cleanup

| File | Change |
|------|--------|
| `onnxruntime/core/providers/cuda/cuda_kernel.h` | Adds `Stream*` overloads for retrieving cuBLAS and cuDNN handles so kernels do not need to reach into `CudaStream` directly. |
| `onnxruntime/core/providers/cuda/integer_gemm.cc` | Replaces direct `CudaStream` handle access with `CudaKernel` helper-based cuBLAS retrieval. |
| `onnxruntime/core/providers/cuda/reduction/reduction_ops.cc` | Reuses a single cuDNN handle derived from the ORT stream and removes repeated direct `CudaStream` assumptions. |
| `onnxruntime/core/providers/cuda/reduction/reduction_ops.h` | Stops caching a `CUDAExecutionProvider*` in reduction kernels, reducing provider coupling. |
| `onnxruntime/core/providers/cuda/rnn/cudnn_rnn_base.cc` | Switches RNN weight reorganization to retrieve the cuDNN handle through the shared stream helper. |
| `onnxruntime/core/providers/cuda/tensor/transpose.cc` | Uses the new stream-based cuBLAS handle helper in transpose fast paths. |

### Kernel-Level Plugin Compatibility Fixes

| File | Change |
|------|--------|
| `onnxruntime/core/providers/cuda/tensor/reshape.cc` | Replaces `CopyTensor` usage with explicit `cudaMemcpyAsync` on the kernel stream to avoid plugin-incompatible stream assumptions. |
| `onnxruntime/core/providers/cuda/tensor/reshape.h` | Applies the same stream-based reshape copy path to the attribute-driven reshape variant. |
| `onnxruntime/core/providers/cuda/math/clip.h` | Removes the CPU `Clip_6Base` dependency by inlining min/max attribute handling into the CUDA kernel. |
| `onnxruntime/contrib_ops/cuda/bert/group_query_attention.cc` | Stores present KV outputs once and reuses those tensors instead of re-fetching them through `context->Output()`. |
| `onnxruntime/contrib_ops/cuda/quantization/qordered_ops/qordered_qdq.cc` | Improves shape validation error reporting by emitting the tensor shape object directly. |

## Motivation and Context

The CUDA plugin EP compiles kernels into a separate shared library and can only depend on types and helpers that are visible through the EP API boundary. The background doc for this work calls out several recurring incompatibility patterns, especially direct use of `CudaStream`, direct inclusion of real CUDA EP infrastructure headers, and helper implementations that only exist in provider-owned `.cc` files.

This PR addresses that class of issues for another slice of the remaining excluded CUDA ops by:

- moving broadly useful CUDA type helpers into header-visible code,
- routing cuBLAS/cuDNN handle lookup through stream-oriented helpers instead of provider internals,
- removing a CPU-base-class dependency from `Clip_6`, and
- simplifying a few kernel call sites that were still assuming the non-plugin CUDA EP environment.

Together, these changes make the CUDA provider code more self-contained and reduce the amount of plugin-specific adaptation needed to bring the remaining CUDA ops online.

## Checklist

- [ ] Tests added/updated
- [x] Documentation updated (background captured in `docs/cuda_plugin_ep/cuda_ops_for_plugin_ep.md`)
- [x] No breaking changes
- [ ] CI passes

### DIFF
--- a/onnxruntime/core/providers/cuda/math/clip.h
+++ b/onnxruntime/core/providers/cuda/math/clip.h
@@ -4,6 +4,7 @@
 #pragma once
 #include "core/providers/cuda/cuda_kernel.h"
 #include "core/providers/cpu/math/clip.h"
+#include <limits>
 
 namespace onnxruntime {
 namespace cuda {


onnxruntime/core/providers/cuda/cuda_common.h
```diff
@@ -231,14 +232,9 @@ class HalfGemmOptions {
...
-
-const char* CublasComputeTypeToString(cublasComputeType_t ct);
 #endif
 
-cudaDataType_t ToCudaDataType(int32_t element_type);
-
 }  // namespace cuda
 }  // namespace onnxruntime
+
+#include "core/providers/cuda/cuda_common_type_helpers.h"
```
kunal-vaishnavi, line 240: Why is this `include` at the bottom of the file?

kunal-vaishnavi (approved):

onnxruntime/core/providers/cuda/cuda_common.h
tianleiwu: Those helper functions were at the end of the file so this is the safe way to change.

PR #27721, opened 2026-03-18
Fix out-of-bounds read in Gemm PrePack with malformed model dimensions

bhkumar007:
### Description
Adds input validation to GemmPackBFp32 to prevent out-of-bounds memory access in MLAS MlasSgemmTransposePackB when processing malformed ONNX models with invalid weight tensor dimensions.

### Motivation and Context
Bug:
A crafted 8-byte ONNX model can cause ORT to parse garbage tensor dimensions for a Gemm node's weight matrix (B). These unvalidated dimensions are passed through Gemm<float>::PrePack → GemmPackBFp32 → MlasGemmPackB → MlasSgemmTransposePackB, which uses the leading dimension (ldb) as a stride multiplier (e.g., b[ldb * 7]), resulting in an out-of-bounds read and crash.

Root cause:
GemmPackBFp32 did not validate that:
* Tensor dimensions are positive (negative int64_t values silently wrap to huge size_t values via static_cast)
* The tensor's backing data is large enough for the declared shape

Fix:
* Reject tensors with non-positive dimensions (<= 0) before casting int64_t shape values to size_t
* Verify the tensor has sufficient elements for K * N

### DIFF
--- a/onnxruntime/core/providers/cpu/math/gemm.cc
+++ b/onnxruntime/core/providers/cpu/math/gemm.cc
@@ -122,6 +122,12 @@ bool GemmPackBFp32(AllocatorPtr& alloc,
   const size_t K = trans_b ? static_cast<size_t>(b_shape[1]) : static_cast<size_t>(b_shape[0]);
   const size_t N = trans_b ? static_cast<size_t>(b_shape[0]) : static_cast<size_t>(b_shape[1]);
 
+  // Verify the tensor has enough data for the declared dimensions.
+  const size_t required_bytes = K * N * sizeof(float);
+  if (tensor_b.SizeInBytes() < required_bytes) {
+    return false;
+  }
+
   packed_b_size = MlasGemmPackBSize(trans_a ? CblasTrans : CblasNoTrans, trans_b ? CblasTrans : CblasNoTrans, N, K, mlas_backend_kernel_selector_config);
   if (packed_b_size == 0) {
     return false;


onnxruntime/core/providers/cpu/math/gemm.cc
```diff
@@ -113,11 +113,21 @@ bool GemmPackBFp32(AllocatorPtr& alloc,
...
+  if (b_shape[0] <= 0 || b_shape[1] <= 0) {
+    return false;
+  }
 
   const size_t K = trans_b ? static_cast<size_t>(b_shape[1]) : static_cast<size_t>(b_shape[0]);
   const size_t N = trans_b ? static_cast<size_t>(b_shape[0]) : static_cast<size_t>(b_shape[1]);
 
+  // Verify the tensor has enough data for the declared dimensions.
+  if (b_shape.Size() < 0 ||
+      static_cast<size_t>(b_shape.Size()) < K * N) {
```
edgchen1: given the previous check for `tensor_b.Shape().NumDimensions() != 2`, this seems unnecessary since `K` and `N` will be the only two dimensions.

onnxruntime/core/providers/cpu/math/gemm.cc
```diff
@@ -113,11 +113,21 @@ bool GemmPackBFp32(AllocatorPtr& alloc,
   if (tensor_b.Shape().NumDimensions() != 2) {
     return false;
   }
+
   b_shape = tensor_b.Shape();
+  if (b_shape[0] <= 0 || b_shape[1] <= 0) {
```
edgchen1: perhaps we can just put the `b_shape.Size() < 0` check here instead.

onnxruntime/core/providers/cpu/math/gemm.cc
```diff
@@ -113,11 +113,21 @@ bool GemmPackBFp32(AllocatorPtr& alloc,
...
+    return false;
+  }
 
   const size_t K = trans_b ? static_cast<size_t>(b_shape[1]) : static_cast<size_t>(b_shape[0]);
   const size_t N = trans_b ? static_cast<size_t>(b_shape[0]) : static_cast<size_t>(b_shape[1]);
 
+  // Verify the tensor has enough data for the declared dimensions.
+  if (b_shape.Size() < 0 ||
+      static_cast<size_t>(b_shape.Size()) < K * N) {
+    return false;
```
edgchen1: could you please add a test for this fix?

onnxruntime/core/providers/cpu/math/gemm.cc
bhkumar007: yes, makes sense, updated.

onnxruntime/core/providers/cpu/math/gemm.cc
bhkumar007: Added tests

PR #27724, opened 2026-03-18
Fix non-ASCII Unicode model path crash across session and provider code

sagarbhure-msft:
### Description
On Windows, `std::filesystem::path::string()` converts the internal UTF-16 representation to a narrow string using the system's active ANSI code page. When the path contains characters outside that code page (Japanese, Chinese, Korean, etc.), this throws std::system_error with 'No mapping for the Unicode character exists in the target multi-byte code page.'

This affected both the core session telemetry logging (causing `InferenceSession::Initialize()` to fail) and execution provider code (OpenVINO, TensorRT, TensorRT RTX, QNN, MIGraphX) where model paths are converted for EPContext attributes and profiling.

### Motivation and Context
Fix: Replace `.filename().string()` with `PathToUTF8String(.filename().native())` which uses `WideCharToMultiByte(CP_UTF8, ...)` and handles all Unicode characters correctly. This pattern is already used elsewhere in the codebase for path-to-string conversions.

Note: Two remaining instances in Linux-only code (`cann_utils.cc`, `device_discovery.cc`) are left as-is since `.string()` is safe on Linux where paths are already narrow strings.

Fixes microsoft/WindowsAppSDK#6173

### DIFF
--- a/onnxruntime/core/providers/nv_tensorrt_rtx/onnx_ctx_model_helper.cc
+++ b/onnxruntime/core/providers/nv_tensorrt_rtx/onnx_ctx_model_helper.cc
@@ -166,7 +166,7 @@ Status CreateCtxNode(const GraphViewer& graph_viewer,
     }
     attr_ep_cache_context->set_s(engine_data_str);
   } else {
-    std::string engine_cache_filename = ToUTF8String(std::filesystem::path(engine_cache_path).filename().native());
+    std::string engine_cache_filename = PathToUTF8String(std::filesystem::path(engine_cache_path).filename().native());
     attr_ep_cache_context->set_s(engine_cache_filename);
     std::fstream engine_cache_file(engine_cache_path, std::ios::binary | std::ios::out);
     if (engine_cache_file.is_open()) {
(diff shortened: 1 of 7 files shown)


onnxruntime/core/providers/migraphx/migraphx_execution_provider_utils.h
```diff
@@ -268,7 +268,7 @@ inline std::string GenerateGraphId(const GraphViewer& graph_viewer) {
   const fs::path path{main_graph.ModelPath()};
 
   if (path.has_filename()) {
-    const auto model_name = path.filename().string();
+    const auto model_name = PathToUTF8String(path.filename().native());
```
yuslepukhin, line 271: > PathToUTF8String [](http://example.com/codeflow?start=27&length=17)

Please, unify the usage on PathToUt8String() for readability.

sagarbhure-msft: Thanks @yuslepukhin - good catch on the inconsistency. Updated all call sites to use PathToUTF8String() uniformly. Also addressed the Copilot review comment on the same point. Please take another look when you get a chance

PR #27730, opened 2026-03-18
Fix onnxruntime-qnn python wheel build on WSL

derdeljan-msft:
### Description

Building Linux onnxruntime-qnn python wheel didn't work through WSL because QNN lib dependencies were not copied over to the wheel, so this PR resolved the issue.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -399,9 +399,7 @@ if platform.system() == "Linux" or platform.system() == "AIX":
         "libHtpPrepare.so",
     ]
     dl_libs.extend(qnn_deps)
-    is_wsl = "microsoft-standard" in platform.uname().release.lower()
-    if is_wsl:
-        libs.extend(qnn_deps)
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -399,6 +399,9 @@ def finalize_options(self):
         "libHtpPrepare.so",
     ]
     dl_libs.extend(qnn_deps)
+    is_wsl = "microsoft-standard" in platform.uname().release.lower()
+    if is_wsl:
+        libs.extend(qnn_deps)
```
edgchen1: do we need the WSL-specific condition?

this `dl_libs` and `libs` thing looks like it could be cleaned up. only `dl_libs` is used for manylinux and only `libs` is used for the non-manylinux case. short of that, maybe we can just add `qnn_deps` to both, as is done below for `nv_tensorrt_rtx_deps`.


```suggestion
    libs.extend(qnn_deps)
```

PR #27747, opened 2026-03-18
[Core] Add SpaceToDepth fusion pattern

hariharans29:
### Description
Fuses a pattern of 4 Slices + 1 Concat into 1 SpaceToDepth op (+1 Gather if the order doesn't match the expected default pattern). Saves about ~1ms for Yolox_tiny which is a sub-15ms model on an AVX512 machine. So, it is good savings.

### Motivation and Context
Improves performance especially when the SpaceToDepth operation occurs in a cheap real-time model

Original SpaceToDepth operation in the model (4 slices + 1 Concat is a lot of memory transactions that can be replaced by one optimized kernel):

image

### DIFF
--- a/onnxruntime/core/optimizer/slice_concat_to_space_to_depth_fusion.cc
+++ b/onnxruntime/core/optimizer/slice_concat_to_space_to_depth_fusion.cc
@@ -36,6 +36,13 @@ constexpr int64_t kRank = 4;
 constexpr int64_t kChannelAxis = 1;
 constexpr int64_t kHeightAxis = 2;
 constexpr int64_t kWidthAxis = 3;
+// This fusion currently only recognizes the common blocksize=2 pattern used by
+// YOLO-style focus layers: 4 Slice nodes with offsets in {0,1}x{0,1}, step=2,
+// followed by channel-axis Concat. The same idea generalizes to arbitrary
+// blocksize b by matching b^2 Slice nodes with offsets in {0..b-1}x{0..b-1},
+// step=b, and extending the phase-permutation/channel-reorder logic
+// accordingly. For now we intentionally keep the implementation limited to the
+// blocksize=2 case.
 constexpr int64_t kBlockSize = 2;
 
 int64_t NormalizeAxis(int64_t axis, int64_t rank) {
@@ -464,6 +471,10 @@ bool FuseSliceConcatToSpaceToDepth(Node& concat, Graph& graph, const logging::Lo
   }
 
   const bool is_canonical_order = phase_permutation == std::array<int64_t, 4>{0, 1, 2, 3};
+  int64_t channel_count = 0;
+  if (!is_canonical_order && !TryGetStaticChannelCount(*common_input, channel_count)) {
+    return false;
+  }
 
   InlinedVector<NodeArg*> space_to_depth_outputs;
   if (is_canonical_order) {
@@ -488,11 +499,6 @@ bool FuseSliceConcatToSpaceToDepth(Node& concat, Graph& graph, const logging::Lo
 
   Node* replacement_end = &space_to_depth;
   if (!is_canonical_order) {
-    int64_t channel_count = 0;
-    if (!TryGetStaticChannelCount(*common_input, channel_count)) {
-      return false;
-    }
-
     InlinedVector<int64_t> gather_indices;
     gather_indices.reserve(onnxruntime::narrow<size_t>(channel_count * kBlockSize * kBlockSize));
     for (const int64_t source_block_index : phase_permutation) {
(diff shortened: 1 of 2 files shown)


onnxruntime/core/optimizer/slice_concat_to_space_to_depth_fusion.cc
```diff
@@ -0,0 +1,501 @@
...
+}
+
+bool IsSupportedSpaceToDepthInputType(const NodeArg& input) {
+  const auto* type_proto = input.TypeAsProto();
+  if (type_proto == nullptr || !type_proto->has_tensor_type()) {
+    return false;
+  }
+
+  const int32_t elem_type = type_proto->tensor_type().elem_type();
+  if (elem_type != TensorProto::FLOAT && elem_type != TensorProto::DOUBLE) {
```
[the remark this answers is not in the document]
hariharans29: Left a TODO for this. Will address in the future if required.

onnxruntime/core/optimizer/graph_transformer_utils.cc
```diff
@@ -260,7 +261,7 @@ InlinedVector<std::unique_ptr<GraphTransformer>> GenerateTransformers(
       transformers.emplace_back(std::make_unique<ReshapeFusion>());
       transformers.emplace_back(std::make_unique<FreeDimensionOverrideTransformer>(
           session_options.free_dimension_overrides));
-
+      transformers.emplace_back(std::make_unique<SliceConcatToSpaceToDepthFusion>());
```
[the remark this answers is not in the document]
hariharans29: Fixed.

onnxruntime/core/optimizer/slice_concat_to_space_to_depth_fusion.cc
```diff
@@ -0,0 +1,590 @@
...
+  if (is_canonical_order) {
+    space_to_depth_outputs = {};
+  } else {
+    auto space_to_depth_output_type = MakeSpaceToDepthOutputTypeProto(*common_input);
+    space_to_depth_outputs.push_back(&graph.GetOrCreateNodeArg(graph.GenerateNodeArgName("space_to_depth_out"), &space_to_depth_output_type));
+  }
+
+  NodeArg* space_to_depth_input = graph.GetNodeArg(common_input->Name());
+
+  Node& space_to_depth = graph.AddNode(graph.GenerateNodeName("SpaceToDepth"),
```
tianleiwu, line 489: Might add an orphaned SpaceToDepth node in the graph when exit early in line 493.

**Suggested fix**: Either (a) move the `TryGetStaticChannelCount` check *before* the SpaceToDepth node is created, or (b) remove the SpaceToDepth node from the graph on the failure path:
  ```cpp
  // Option (a) — preferred: check before creating any nodes
  const bool needs_gather = !is_canonical_order;
  if (needs_gather) {
    int64_t channel_count = 0;
    if (!TryGetStaticChannelCount(*common_input, channel_count)) {
      return false;  // no graph mutation yet, safe to bail
    }
    // ... proceed with node creation ...
  }
  ```
  
Given the dangling-node concern above, a test with a dynamic channel dimension and non-canonical Slice ordering would verify the graceful failure path (once the bug is fixed).

onnxruntime/core/optimizer/slice_concat_to_space_to_depth_fusion.cc
```diff
@@ -0,0 +1,590 @@
...
+  std::array<int64_t, 4> starts;
+  std::array<int64_t, 4> ends;
+  std::array<int64_t, 4> steps;
+};
+
+constexpr int64_t kRank = 4;
+constexpr int64_t kChannelAxis = 1;
+constexpr int64_t kHeightAxis = 2;
+constexpr int64_t kWidthAxis = 3;
+constexpr int64_t kBlockSize = 2;
```
tianleiwu, line 46: The fusion is limited to blocksize=2. While this covers the most common YOLO-style focus layers, the pattern generalizes to any blocksize `b` where there are `b²` Slices with offsets in `{0..b-1}×{0..b-1}` and step=`b`. Consider adding a comment noting this limitation and the generalization path, even if only blocksize=2 is implemented now.

onnxruntime/core/optimizer/slice_concat_to_space_to_depth_fusion.cc
hariharans29: Fixed

onnxruntime/core/optimizer/slice_concat_to_space_to_depth_fusion.cc
hariharans29: Added comment

PR #27751, opened 2026-03-18
Enable KleidiAI for asymmetric 4-bit MatMulNBits on ARM64

jambayk:
### Description

Extends the KleidiAI-accelerated `MatMulNBits` path on ARM64 to support asymmetric 4-bit quantization (models with per-block zero points). Previously, KleidiAI was only used for symmetric quantization (`!HasZp` guard), and asymmetric models fell back to a significantly slower non-KleidiAI kernel.

Since KleidiAI only provides symmetric int4 micro-kernels (`kai_matmul_clamp_f32_qai8dxp_qsi4c32p`), this PR runs KleidiAI as-if symmetric (hardcoded `rhs_zero_point=8`) and applies a float-domain zero-point correction post-GEMM:

$$C_{\text{actual}} = C_{\text{symmetric}} + A_{\text{blksum}} \times B_{\text{ZpCorr}}^T$$

where:
- `BZpCorr[n, blk] = scale_b[n, blk] × (8 - zp_b[n, blk])` — precomputed at weight packing time
- `AFloatBlkSum[m, blk] = Σ A_float[m, blk_start..blk_end]` — computed per inference alongside A quantization

Key changes:
- **`UseKleidiAI()`**: Removed the `!HasZp` guard so KleidiAI is now eligible for both symmetric and asymmetric models (checks `BlkLen%32==0`, `K%BlkLen==0`, dotprod capability).
- **B packing (`SQ4BitGemmPackQuantBDataAndBlkSum`)**: Computes and stores `BZpCorr` after KleidiAI packed B data when zero points are present.
- **Workspace expansion**: Allocates space for `AFloatBlkSum` (M × BlockCountK floats) in the per-GEMM workspace for asymmetric models.
- **`ComputeAFloatBlkSum`**: NEON-vectorized (4× unrolled `vaddq_f32`) function to compute per-block float sums of A.
- **`ApplyBZpCorrection`**: NEON-vectorized correction kernel tiled 4-N-wide (`vfmaq_f32`) for L1-friendly BZpCorr reuse.
- **PrePack**: Computes `BZpCorr` during the scales PrePack (not zero_points PrePack), since ORT may erase constant inputs after marking them packed.

No changes to the symmetric path. No changes to x64. No changes to 8-bit quantization.

### Motivation and Context

Asymmetric 4-bit quantized models (e.g., GPTQ/RTN with zero points) on ARM64 were **23–72% slower** than their symmetric counterparts because KleidiAI's `sdot`/`i8mm` micro-kernels only support symmetric RHS, forcing a fallback to a slower non-KleidiAI kernel path.

This change closes most of that gap:

| Model | Seq Len | Asym/Sym (before) | Asym/Sym (after) | Asym speedup | Asym latency (after) |
|-------|---------|-------------------|------------------|--------------|----------------------|
| Qwen 1.5B | 256 | 1.35× | 1.17× | **1.16×** | 1107.8ms |
| Qwen 1.5B | 512 | 1.23× | 1.06× | **1.14×** | 2259.7ms |
| Qwen 3B | 256 | 1.43× | 1.12× | **1.28×** | 2029.7ms |
| Qwen 3B | 512 | 1.39× | 1.22× | **1.24×** | 4188.0ms |
| Qwen 7B | 256 | 1.61× | 1.11× | **1.52×** | 3661.6ms |
| Qwen 7B | 512 | 1.72× | 1.11× | **1.58×** | 7263.8ms |

The remaining 6–22% asym/sym gap comes from the extra pass over A to compute float block sums — this cannot be fused into KleidiAI's sealed A-packing function and would require an upstream KleidiAI API change.

### DIFF
--- a/onnxruntime/core/mlas/lib/qnbitgemm.cpp
+++ b/onnxruntime/core/mlas/lib/qnbitgemm.cpp
@@ -944,7 +944,9 @@ SQ4BitGemm_CompInt8(
             DataParams->C, RangeStartM, RangeCountM, RangeStartN, RangeCountN, K,
             DataParams->ldc, DataParams->Bias);
 
-        // Apply zero-point correction for asymmetric quantization:
+        // Apply zero-point correction for asymmetric quantization (KleidiAI path only).
+        // BZpCorr and AFloatBlkSum are only set when KleidiAI is active with asymmetric
+        // quantization (has zero points). On all other paths they remain nullptr.
         // C += AFloatBlkSum * BZpCorr^T  (for this tile's M/N ranges)
         if (DataParams->BZpCorr != nullptr && DataParams->AFloatBlkSum != nullptr) {
             const size_t BlockCountK = MlasDivRoundup(K, BlkLen);
--- a/onnxruntime/core/mlas/lib/qnbitgemm_kernel_neon.cpp
+++ b/onnxruntime/core/mlas/lib/qnbitgemm_kernel_neon.cpp
@@ -63,7 +63,7 @@ QNBitGemmPackQuantBDataSize(
 #endif
 
 #ifdef USE_KLEIDIAI
-        if (ComputeType == SQNBIT_CompInt8 && UseKleidiAIBase(K, BlkLen, BackendKernelSelectorConfig)) {
+        if (ComputeType == SQNBIT_CompInt8 && UseKleidiAI(K, BlkLen, BackendKernelSelectorConfig)) {
             const auto& k = GetKleidiAIGemmUKernel();
             const auto& ukernel = k.ukernel;
             const size_t nr = ukernel.get_nr();
@@ -207,7 +207,7 @@ SQ4BitGemmPackQuantBDataAndBlkSum(
     assert(BlkLen >= 16 && BlkLen % 16 == 0);
 
 #ifdef USE_KLEIDIAI
-    if (UseKleidiAIBase(K, BlkLen, BackendKernelSelectorConfig)) {
+    if (UseKleidiAI(K, BlkLen, BackendKernelSelectorConfig)) {
         const auto& k = GetKleidiAIGemmUKernel();
         const auto& ukernel = k.ukernel;
         std::byte* PackedQuantBDataBegin = PackedQuantB.PackedQuantBData;
@@ -461,7 +461,7 @@ QNBitGemmPerGemmWorkspaceSize(
         case SQNBIT_CompInt8: {
             // workspace buffer is used for block quantization of A to int8
 #ifdef USE_KLEIDIAI
-            if (BlkBitWidth == 4 && UseKleidiAIBase(K, BlkLen, BackendKernelSelectorConfig)) {
+            if (BlkBitWidth == 4 && UseKleidiAI(K, BlkLen, BackendKernelSelectorConfig)) {
                 const auto& k = (M == 1) ? GetKleidiAIGemvUKernel() : GetKleidiAIGemmUKernel();
                 const auto& ukernel = k.ukernel;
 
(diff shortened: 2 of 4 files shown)


onnxruntime/core/mlas/lib/qnbitgemm_kernel_neon.cpp
```diff
@@ -192,36 +202,68 @@ SQ4BitGemmPackQuantBDataAndBlkSum(
...
+        // Note: We intentionally use fp32 scales here (not bf16-truncated) for higher precision
+        // in the correction term. KleidiAI internally truncates scales to bf16 for the main GEMM,
+        // but the correction benefits from full fp32 precision to better approximate the true result.
+        // This may be called separately from B packing when zero points arrive later.
+        if (HasZeroPoint && QuantBZPBegin != nullptr && QuantBScaleBegin != nullptr) {
+            const size_t kleidiai_packed_size = kai_get_rhs_packed_size_rhs_pack_nxk_qsi4c32p_qsu4c32s1s0(
+                N, K, nr, kr, sr, BlkLen, kai_dt_bf16);
+            // Align offset so BZpCorr starts at a float-aligned address
+            constexpr size_t FloatAlignment = alignof(float);
+            const size_t bzpcorr_offset = (kleidiai_packed_size + FloatAlignment - 1) & ~(FloatAlignment - 1);
```
hariharans29, line 254: Are there MLAS unit tets that can be added for the ZP correction tensor and the A BlkSum components ?

onnxruntime/core/mlas/lib/qnbitgemm_kernel_neon.cpp
```diff
@@ -471,24 +522,29 @@ QNBitGemmPerGemmWorkspaceAlignment(
...
     MLAS_UNREFERENCED_PARAMETER(BlkLen);
-    MLAS_UNREFERENCED_PARAMETER(HasZp);
     return false;
 #endif
 }
 
+bool
+UseKleidiAI(size_t K, size_t BlkLen, bool HasZp, const MLAS_BACKEND_KERNEL_SELECTOR_CONFIG* BackendKernelSelectorConfig)
+{
+    return !HasZp && UseKleidiAIBase(K, BlkLen, BackendKernelSelectorConfig);
```
hariharans29: Where is UseKleidiAI() used anymore if we are now supporting assymetric quantized workloads with KAI ? I thought UseKleidiAI is superseded by UseKleidiAIBase (in which case we can remove UseKleidiAIBase and just keep UseKleidiAI ?)

onnxruntime/core/mlas/lib/qnbitgemm.cpp
```diff
@@ -943,6 +943,32 @@ SQ4BitGemm_CompInt8(
         SQ4BitGemm(BlkLen, QuantA, DataParams->PackedQuantBData,
             DataParams->C, RangeStartM, RangeCountM, RangeStartN, RangeCountN, K,
             DataParams->ldc, DataParams->Bias);
+
+        // Apply zero-point correction for asymmetric quantization:
+        // C += AFloatBlkSum * BZpCorr^T  (for this tile's M/N ranges)
+        if (DataParams->BZpCorr != nullptr && DataParams->AFloatBlkSum != nullptr) {
```
hariharans29, line 951: I take it that this condition will be true only for the KAI case ?

jambayk: yes, added a comment mentioning this

onnxruntime/core/mlas/lib/qnbitgemm_kernel_neon.cpp
jambayk: The existing MLAS tests in test_sqnbitgemm.cpp already exercise both paths. The `Symmetric=false` parameterization (line 382-408) with `SQNBIT_CompInt8` will trigger the KleidiAI asymmetric path on ARM64 — calling `MlasQNBitGemmPackQuantBData` (which computes BZpCorr), InitializeWorkspace (which computes AFloatBlkSum), and the correction in `SQ4BitGemm_CompInt8`. The test compares against a reference GEMM that does the exact dequantization math, so any BZpCorr or AFloatBlkSum bug would be caught. 

there are no isolated tests for `ComputeAFloatBlkSum` and `ApplyBZpCorrection` (i.e., testing those functions directly rather than through the full GEMM flow) though.

onnxruntime/core/mlas/lib/qnbitgemm_kernel_neon.cpp
jambayk: you are right! removed usekleidiaibase and used useklediai as the name

hariharans29 (approved):

PR #27752, opened 2026-03-18
+rotemb, +rmsnorm, reshape->opset-25, transpose->opset-24

guschmue:
for webgpu ep:
+ onnx rotary-embedding op
+ onnx rmsnorm
+ reshape-> opset-25
+ transpose -> opset-24


onnxruntime/core/providers/webgpu/llm/rotary_embedding.cc
```diff
@@ -0,0 +1,132 @@
...
+                 ? std::vector<uint32_t>({batch_stride, head_size, sequence_length * head_size, 1})
+                 : std::vector<uint32_t>({}));
+
+  // The contrib RotaryEmbeddingProgram expects inputs in order:
+  //   input(0), position_ids(1), cos_cache(2), sin_cache(3)
+  // The ONNX op has: X(0), cos_cache(1), sin_cache(2), position_ids(3, optional)
+
+  if (position_ids != nullptr) {
+    // position_ids provided: cos/sin cache is 2D (max_pos, D/2)
+    contrib::webgpu::RotaryEmbeddingProgram program{interleaved_};
```
fs-eire, line 86: technically this file should not depend on contrib ops because the build will fail if `DISABLE_CONTRIB_OPS` is defined. The more common way is to move the implementation from contrib.

fs-eire, line 86: Not directly related to this PR but introducing factories for program maybe a good idea (to avoid duplicated code to call those chain-style methods)

fs-eire (dismissed):

onnxruntime/core/providers/webgpu/llm/rotary_embedding.cc
guschmue: yeah, agree. We'll clean this up in the near future since that is going to be a large change.

guschmue: k

fs-eire (approved):

PR #27761, opened 2026-03-19
[DRAFT] A scan op to support variable-length scan outputs

gramalingam:
### Description
Add ScanVarLen contrib op: variable-length scan output concatenation

Adds a com.microsoft.Scan (v1) kernel that generalizes the ONNX Scan op to support variable-length scan outputs. Per-iteration scan outputs  are concatenated along the specified scan_output_axes, allowing the concatenation-axis dimension to differ across iterations — unlike standard Scan where each iteration must produce the same shape.

### Motivation and Context
This is a proof-of-concept, to verify such an op is implementable. The eventual plan is to add such an op to the ONNX standard. This generalization allows us to encode some computations efficiently: for example, PackedAttention which uses a compact representation for a batch of variable-length sequences by concatenating them. Any processing of such a batch of "ragged" tensors to produce another batch of "ragged" tensors can be expressed as a scan, but requires variable-length scan-outputs.

Note that this is being created as a PR against a temporary branch off main. The main goal is to get review feedback.

NOTE: The code is copilot-generated.


gramalingam:
Hi @skottmckay : this is a proof-of-concept implementation for scan with variable-length scan-outputs. (This implementation is from copilot, FYI.) Please let me know if you have any concerns. This is just a draft, because ideally I will create the new op in onnx standard, and then implement it. This is currently a contrib-op in this draft.

gramalingam:
Closing this. A candidate onnx PR based on this is in progress. https://github.com/onnx/onnx/pull/7962

PR #27764, opened 2026-03-19
Add MHA fusion for Nemotron speech conformer encoder

kunal-vaishnavi:
### Description

This PR updates the pattern matchings to perform multi-head attention fusion for the conformer encoder inside [Nemotron speech](https://huggingface.co/nvidia/nemotron-speech-streaming-en-0.6b).

image

### Motivation and Context

These changes allow the `MultiHeadAttention` op to appear in the encoder ONNX model.

### DIFF
--- a/onnxruntime/python/tools/transformers/fusion_conformer_attention.py
+++ b/onnxruntime/python/tools/transformers/fusion_conformer_attention.py
@@ -205,7 +205,7 @@ class FusionConformerAttention(FusionAttention):
 
         # Validate attention_bias: the Attention and MultiHeadAttention kernels require a 4-D
         # tensor with shape [batch_size or 1, num_heads or 1, sequence_length, total_sequence_length].
-        # Scalar or 1-D initializers (e.g. a plain QK scaling constant) must not be forwarded as 
+        # Scalar or 1-D initializers (e.g. a plain QK scaling constant) must not be forwarded as
         # attention_bias. Non-initializer values (computed positional-bias outputs) are kept as-is.
         attention_bias = add_qk.input[1]
         bias_init = self.model.get_initializer(attention_bias)


tianleiwu (approved):
Sent too soon :-)

tianleiwu (changes requested):
## Summary

The PR broadens the conformer-attention matcher to cover several new Nemotron graph shapes, and most of the new fallbacks are in the right place for that goal. The main risk is that one of those fallbacks changes an optional side-branch check into a required one, which looks like it will stop fusion on already-supported conformer graphs. There is also redundant postprocessing in the conformer optimizer that will run graph cleanup multiple extra times.

---

## Review

### 1. Attention Fusion Matching (`onnxruntime/python/tools/transformers/fusion_conformer_attention.py`)

**Positive**:
- The new fallback matchers for QKV/Q/K/V and masked-QK paths make the fusion less brittle against exporter-specific graph layouts, especially the variants without `Add` on K/V and the extra `Div` before QK masking.

**Concern**:
- ⚠️ **Existing conformer fusion likely regresses**: [`fusion_conformer_attention.py#L147`](https://github.com/microsoft/onnxruntime/blob/bf02d86b4562efec229028e925b3ad37592d6346/onnxruntime/python/tools/transformers/fusion_conformer_attention.py#L147) used to treat `extra_q_nodes` as optional. In the old code, `extra_q_nodes is None` was the normal path for standard conformer attention, and only the conformer-transducer variant populated it. The new logic now returns early when neither the old extra-Q pattern nor the new Nemotron pattern matches, so existing graphs that never had an auxiliary Q branch will stop fusing here. It also means `extra_q_nodes` is never `None` on the new alternative path, so the packed `Attention` fast path at [`fusion_conformer_attention.py#L207`](https://github.com/microsoft/onnxruntime/blob/bf02d86b4562efec229028e925b3ad37592d6346/onnxruntime/python/tools/transformers/fusion_conformer_attention.py#L207) cannot be taken anymore.
  ```python
  extra_q_nodes = self.model.match_parent_path(...)
  if extra_q_nodes is not None and q_nodes[0].op_type in ["Div", "Mul"] and q_nodes[0] != extra_q_nodes[-1]:
      logger.debug("fuse_conformer_attention: failed to match extra q path")
      return

  nemotron_extra_q_nodes = self.model.match_parent_path(...)
  if nemotron_extra_q_nodes is not None:
      extra_q_nodes = nemotron_extra_q_nodes
  ```

### 2. Conformer Optimizer Postprocess (`onnxruntime/python/tools/transformers/onnx_model_conformer.py`)

**Positive**:
- Adding conformer-specific cleanup for cast-heavy exported graphs is sensible, and keeping it in `postprocess()` is the right abstraction point.

**Concern**:
- ⚠️ **Postprocess now runs more than once**: [`onnx_model_conformer.py#L27`](https://github.com/microsoft/onnxruntime/blob/bf02d86b4562efec229028e925b3ad37592d6346/onnxruntime/python/tools/transformers/onnx_model_conformer.py#L27) calls `self.postprocess()` after `super().optimize(...)`, but `BertOnnxModel.optimize()` already calls `self.postprocess()` polymorphically. On top of that, [`onnx_model_conformer.py#L36`](https://github.com/microsoft/onnxruntime/blob/bf02d86b4562efec229028e925b3ad37592d6346/onnxruntime/python/tools/transformers/onnx_model_conformer.py#L36) invokes `super().postprocess()` and then immediately repeats `clean_graph()` and `prune_graph()` again. That means cleanup/pruning is executed multiple extra times per optimization pass, which adds avoidable work and makes the control flow harder to reason about.
  ```python
  def optimize(...):
      ...
      super().optimize(options, add_dynamic_axes)
      # Remove this second call: BertOnnxModel.optimize() already dispatched to ConformerOnnxModel.postprocess()
      # self.postprocess()

  def postprocess(self):
      super().postprocess()
      self.remove_cascaded_cast_nodes()
      self.remove_useless_cast_nodes()
      self.remove_unused_constant()
      # super().postprocess() already cleaned/pruned once
  ```

---

## Summary of Concerns

| 1 | High | Attention Fusion Matching | `extra_q_nodes` changed from optional to required, which appears to break fusion for already-supported conformer graphs. |
| 2 | Suggestion | Conformer Optimizer Postprocess | `postprocess()` is invoked redundantly and repeats cleanup/pruning work. |

---

## Verdict

REQUEST CHANGES — the new matching logic appears to introduce a real fusion regression for existing conformer models.

onnxruntime/python/tools/transformers/fusion_conformer_attention.py
```diff
@@ -203,6 +203,20 @@ def fuse(self, normalize_node, input_name_to_nodes, output_name_to_node):
             logger.debug("fuse_conformer_attention: failed to detect num_heads or hidden_size")
             return
 
+        # Validate attention_bias: the Attention and MultiHeadAttention kernels require a 4-D
+        # tensor with shape [batch_size, num_heads, seq_len, seq_len].  Scalar or 1-D initializers
```
tianleiwu: [batch_size or 1, num_heads or 1, seq_len, total_seq_len]
see https://github.com/microsoft/onnxruntime/blob/44ce4f5f13fc332285a9cd865f5e729dcca084c3/onnxruntime/contrib_ops/cpu/bert/multihead_attention_helper.h#L184

kunal-vaishnavi: Fixed

PR #27773, opened 2026-03-19
fix webnn test case for webgpu ep: 'transpose float32 1D constant tensor default op…

guschmue:
fixes a few webnn test cases for webgpu ep.
This
https://wpt.live/webnn/conformance_tests/transpose.https.any.html?gpu
should pass 100% now


onnxruntime/core/providers/webgpu/tensor/transpose.cc
```diff
@@ -102,13 +102,20 @@ Status TransposeProgram::GenerateShaderCode(ShaderHelper& shader) const {
                               << "    " << output.SetByIndices("output_indices_t(output_row, output_col)", "tile[local_id.x][local_id.y]") << "\n"
                               << "  }";
   } else {
-    shader.AdditionalImplementation() << "fn perm(i: output_indices_t)->a_indices_t {\n"
-                                         "  var a: a_indices_t;\n";
-    for (size_t i = 0; i < perm_.size(); ++i) {
-      shader.AdditionalImplementation() << "  a[" << perm_[i] << "] = i[" << i << "];\n";
+    if (perm_.size() < 2) {
```
edgchen1: do we need a unit test for this in ORT?

fs-eire: it seems that if perm is 1D, the Transpose is technically an Identity operator. We don't even need to run the shader.

A long term solution is to add a graph transformer to remove Transpose nodes with perm size == 1. 

A short term solution can be using `ComputeContext::CopyTensor()` and not running shader.

guschmue: ok, we can switch to copy

guschmue: done

PR #27778, opened 2026-03-20
fixes to pass webnn DequantizeLinear compliance tests over webgpu ep

guschmue:
This PR is on top of a previous PR and fixes the remaining issues.
https://github.com/microsoft/onnxruntime/pull/27706

All tests here should be passing now over webgpu:
https://wpt.live/webnn/conformance_tests/dequantizeLinear.https.any.html?gpu

### DIFF
--- a/onnxruntime/core/providers/webgpu/quantization/quantize_linear.cc
+++ b/onnxruntime/core/providers/webgpu/quantization/quantize_linear.cc
@@ -65,13 +65,11 @@ Status DequantizeLinearProgram::GenerateShaderCode(ShaderHelper& shader) const {
         << "let scale_value = " << scale.GetByOffset("scale_index") << ";\n";
   } else {
     // Block quantization. Scale input rank is same as input/output rank.
-    // Divide each dimension by its block size (output_shape[i] / scale_shape[i]).
+    // On the block axis, divide by block_size; on other axes, use output index directly.
     shader.MainFunctionBody() << "var scale_indices: scale_indices_t;\n";
     for (int i = 0; i < rank_; i++) {
-      std::string output_dim = GetElementAt("uniforms.output_shape", i, rank_);
-      std::string scale_dim = GetElementAt("uniforms.scale_shape", i, rank_);
-      std::string block_expr = "(" + output_dim + " / " + scale_dim + ")";
-      std::string value_expr = output.IndicesGet("output_indices", i) + " / " + block_expr;
+      std::string idx = output.IndicesGet("output_indices", i);
+      std::string value_expr = "select(" + idx + ", " + idx + " / uniforms.block_size, " + std::to_string(i) + "u == uniforms.axis)";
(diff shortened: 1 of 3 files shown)


onnxruntime/core/providers/webgpu/quantization/quantize_linear.cc
```diff
@@ -122,11 +162,11 @@ Status DequantizeLinear::ComputeInternal(ComputeContext& context) const {
   auto* output_tensor = context.Output(0, x_shape);
   int64_t x_scale_rank = x_scale->Shape().NumDimensions();
 
-  // Currently only INT8, UINT8, and INT32 are registered.
   auto x_type = x->GetElementType();
 
-  bool packed = x_type == ONNX_TENSOR_ELEMENT_DATA_TYPE_INT8 || x_type == ONNX_TENSOR_ELEMENT_DATA_TYPE_UINT8;
-  bool is_signed = x_type == ONNX_TENSOR_ELEMENT_DATA_TYPE_INT8;
+  bool packed_4bit = x_type == ONNX_TENSOR_ELEMENT_DATA_TYPE_UINT4 || x_type == ONNX_TENSOR_ELEMENT_DATA_TYPE_INT4;
+  bool packed = packed_4bit || x_type == ONNX_TENSOR_ELEMENT_DATA_TYPE_INT8 || x_type == ONNX_TENSOR_ELEMENT_DATA_TYPE_UINT8;
```
edgchen1: now we have two bools `packed` and `packed_4bit`, but it seems that there is some dependency between the two. would it be clearer to have some enum to indicate what kind of packing (if any) is done?

onnxruntime/core/providers/webgpu/quantization/quantize_linear.cc
```diff
@@ -122,11 +162,11 @@ Status DequantizeLinear::ComputeInternal(ComputeContext& context) const {
...
+  bool is_signed = x_type == ONNX_TENSOR_ELEMENT_DATA_TYPE_INT8 || x_type == ONNX_TENSOR_ELEMENT_DATA_TYPE_INT4;
```
edgchen1: nit: int32 is also signed. if the meaning is more like "is the packed type signed", maybe adjust the variable name.

edgchen1:
are we hitting the new code paths with our existing unit tests? I'm not sure if all of the relevant tests in onnxruntime\test\providers\cpu\tensor\quantize_linear_test.cc are running for WebGPU EP.

onnxruntime/core/providers/webgpu/quantization/quantize_linear.cc
guschmue: ok, switched to a enum

onnxruntime/core/providers/webgpu/quantization/quantize_linear.cc
guschmue: changed the name

guschmue:
yes, some of the ut was not enabled for webgpu. Fixed now.

Issue #27762, opened 2026-03-19
[WebGPU] Nemotron-H Mamba Einsum node throws invalid shader error


xenova:
### Describe the issue

When running an optimized mamba export, the following error is thrown:
```
available_providers=['WebGpuExecutionProvider', 'CPUExecutionProvider']
node_name=/model/layers.0/mamba/else/ssm/ssm_state_f32_tmp
equation=bcknd,bckns->bcnds
hs_chunked_shape=[1, 1, 128, 8, 16]
B_decay_shape=[1, 1, 128, 8, 16]

Running CPUExecutionProvider...
CPU succeeded: output_shape=(1, 1, 8, 16, 16), sample=-13.966281

Running WebGpuExecutionProvider...
This is expected to fail while compiling the Einsum shader.
2026-03-19 01:58:54.824 Python[64980:7279612] 2026-03-19 01:58:54.824530 [E:onnxruntime:, sequential_executor.cc:611 ExecuteKernel] Non-zero status code returned while running Einsum node. Name:'/model/layers.0/mamba/else/ssm/ssm_state_f32_tmp' Status Message: Failed to create a WebGPU compute pipeline: [Invalid ShaderModule "Einsum"] is invalid due to a previous error.
 - While validating compute stage ([Invalid ShaderModule "Einsum"], entryPoint: "main").
```

### To reproduce

codex helped cook up a standalone reproduction.

```py
import tempfile
from pathlib import Path

import numpy as np
import onnx
import onnxruntime as ort
from onnx import TensorProto, helper


MODEL_NAME = "einsum_webgpu_bug"
NODE_NAME = "/model/layers.0/mamba/else/ssm/ssm_state_f32_tmp"
EQUATION = "bcknd,bckns->bcnds"

# These shapes mirror the chunked Mamba tensors used by Nemotron-H:
# hs_chunked: [batch, chunks, chunk_size, n_heads, d_head]
# B_decay:    [batch, chunks, chunk_size, n_heads, d_state]
HS_CHUNKED_SHAPE = [1, 1, 128, 8, 16]
B_DECAY_SHAPE = [1, 1, 128, 8, 16]
OUTPUT_SHAPE = [1, 1, 8, 16, 16]


def build_model(model_path: Path) -> None:
    hs_chunked = helper.make_tensor_value_info("hs_chunked", TensorProto.FLOAT, HS_CHUNKED_SHAPE)
[...]

PR #27779, opened 2026-03-20
[WebGPU] Einsum fixes for 5D tensors

xenova:
### Description

Fixes WebGPU Einsum op by replacing the manual `uniforms.inputN_shape[idx]` string with `GetElementAt(...)` which correctly handles uniform shape access for all tensor ranks.

I also added a bunch of tests for this...

### Motivation and Context

Closes https://github.com/microsoft/onnxruntime/issues/27762


xenova:
cc @guschmue

PR #27780, opened 2026-03-20
Optimize FlashAttention for M4 Max (20x speedup)

xenova:
MultiHeadAttention
Before: 58.3s
After: 2.89
Speedup: 20x

### Motivation and Context

Tested with vision_encoder.onnx for https://huggingface.co/onnx-community/LightOnOCR-2-1B-ONNX

### DIFF
--- a/onnxruntime/contrib_ops/webgpu/bert/flash_attention.h
+++ b/onnxruntime/contrib_ops/webgpu/bert/flash_attention.h
@@ -78,7 +78,8 @@ class FlashAttentionProgram final : public Program<FlashAttentionProgram> {
                         bool is_nvidia,
                         bool q_BNSH,
                         bool use_seqlen_k = false,
-                        bool has_head_sink = false)
+                        bool has_head_sink = false,
+                        int qkv_max_k_step = 16)
       : Program{kernel_name},
         has_attention_bias_(has_attention_bias),
         is_qualcomm_(is_qualcomm),
@@ -89,7 +90,8 @@ class FlashAttentionProgram final : public Program<FlashAttentionProgram> {
         is_nvidia_(is_nvidia),
         q_BNSH_(q_BNSH),
         use_seqlen_k_(use_seqlen_k),
-        has_head_sink_(has_head_sink) {
+        has_head_sink_(has_head_sink),
+        qkv_max_k_step_(qkv_max_k_step) {
   }
 
   Status GenerateShaderCode(ShaderHelper& sh) const override;
@@ -115,6 +117,7 @@ class FlashAttentionProgram final : public Program<FlashAttentionProgram> {
   bool q_BNSH_;
   bool use_seqlen_k_;
   bool has_head_sink_;
+  int qkv_max_k_step_;
 };
 
 class FlashAttentionDecodeQKTProgram final : public Program<FlashAttentionDecodeQKTProgram> {
--- a/onnxruntime/contrib_ops/webgpu/bert/flash_attention.h
+++ b/onnxruntime/contrib_ops/webgpu/bert/flash_attention.h
@@ -75,6 +75,8 @@ class FlashAttentionProgram final : public Program<FlashAttentionProgram> {
                         int qkv_head_size,
                         int qkv_num_heads,
                         bool is_unidirectional,
+                        bool is_nvidia,
+                        bool is_apple,
                         bool q_BNSH,
                         bool use_seqlen_k = false,
                         bool has_head_sink = false)
@@ -85,24 +87,23 @@ class FlashAttentionProgram final : public Program<FlashAttentionProgram> {
         qkv_head_size_(qkv_head_size),
         qkv_num_heads_(qkv_num_heads),
         is_unidirectional_(is_unidirectional),
+        is_nvidia_(is_nvidia),
+        is_apple_(is_apple),
         q_BNSH_(q_BNSH),
         use_seqlen_k_(use_seqlen_k),
         has_head_sink_(has_head_sink) {
-    // Compute optimal max_k_step based on head_size and shared memory budget.
-    // SHM budget: 16KB = 16384 bytes.
-    // k_tile + v_tile = 2 * element_size * head_size * max_k_step
-    // element_size = 2 for f16, 4 for f32.
-    const int element_size = is_fp16 ? 2 : 4;
-    const int shm_budget = 16384;
-    int max_k_from_shm = shm_budget / (2 * element_size * qkv_head_size);
-    // On Qualcomm, keep max_k_step=16 to avoid register spilling.
-    // Other GPUs (Apple, Nvidia, Intel): use a dynamic qk_scores array, cap at 64.
-    if (is_qualcomm) {
-      max_k_step_ = 16;
-    } else if (max_k_from_shm >= 64) {
-      max_k_step_ = 64;
-    } else if (max_k_from_shm >= 32) {
-      max_k_step_ = 32;
+    if (is_apple) {
+      // On Apple GPUs, use an optimized loop-based path with dynamic max_k_step.
+      // Compute max_k_step from shared memory budget: k_tile + v_tile = 2 * element_size * head_size * max_k_step
+      const int element_size = is_fp16 ? 2 : 4;
+      int max_k_from_shm = 16384 / (2 * element_size * qkv_head_size);
+      if (max_k_from_shm >= 64) {
+        max_k_step_ = 64;
+      } else if (max_k_from_shm >= 32) {
+        max_k_step_ = 32;
+      } else {
+        max_k_step_ = 16;
+      }
     } else {
       max_k_step_ = 16;
     }
@@ -129,6 +130,8 @@ class FlashAttentionProgram final : public Program<FlashAttentionProgram> {
   int qkv_head_size_;
   int qkv_num_heads_;
   bool is_unidirectional_;
+  bool is_nvidia_;
+  bool is_apple_;
   bool q_BNSH_;
   bool use_seqlen_k_;
   bool has_head_sink_;
--- a/onnxruntime/contrib_ops/webgpu/bert/flash_attention.h
+++ b/onnxruntime/contrib_ops/webgpu/bert/flash_attention.h
@@ -97,9 +97,7 @@ class FlashAttentionProgram final : public Program<FlashAttentionProgram> {
       // Compute max_k_step from shared memory budget: k_tile + v_tile = 2 * element_size * head_size * max_k_step
       const int element_size = is_fp16 ? 2 : 4;
       int max_k_from_shm = 16384 / (2 * element_size * qkv_head_size);
-      if (max_k_from_shm >= 64) {
-        max_k_step_ = 64;
-      } else if (max_k_from_shm >= 32) {
+      if (max_k_from_shm >= 32) {
         max_k_step_ = 32;
       } else {
         max_k_step_ = 16;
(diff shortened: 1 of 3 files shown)


xenova:
@guschmue 🙏

qjia7:
Awesome! I made a similar change a few days ago to optimize Whisper locally, but your approach is more comprehensive than mine. I just tested it and observed comparable improvements to those in #27781 for Whisper. I’ll go ahead and close that one, and I’m looking forward to seeing this land soon!

onnxruntime/contrib_ops/webgpu/bert/flash_attention.wgsl.template
```diff
@@ -31,27 +29,25 @@ const min_value = q_element_t(-65504.0);
...
-// Default SHM usage limit is 16KB in Dawn.
-// vec4<f16> * head_size_vec * max_k_step = 8 * (128/4) * 16 = 4KB. 128 is head_size for phi4.
+// K and V tiles in shared memory.
+// SHM usage = 2 * sizeof(vec4<element>) * head_size_vec * max_k_step
 var<workgroup> k_tile : array<array<q_value_t, head_size_vec>, max_k_step>;
 var<workgroup> v_tile : array<array<q_value_t, head_size_vec>, max_k_step>;
 
 // Private memory per lane.
 var<private> q_tile : array<q_value_t, head_size_vec>;
+var<private> qk_scores : array<q_element_t, max_k_step>;
```
qjia7: When max_k_step = 128 (e.g., head_size=32 with f16): this allocates 128 private registers per lane for QK scores. On some GPUs, this may cause register spilling and hurt performance.  Have you tested this on less powerful devices, such as Intel Tiger Lake or Qualcomm?

xenova: > Have you tested this on less powerful devices, such as Intel Tiger Lake or Qualcomm?

Unfortunately not, I've mainly just tested on my device to be honest. Do you have recommendations or a CI that can help with this?

xenova:
> Awesome! I made a similar change a few days ago to optimize Whisper locally, but your approach is more comprehensive than mine. I just tested it and observed comparable improvements to those in #27781 for Whisper. I’ll go ahead and close that one, and I’m looking forward to seeing this land soon!

Great! 😄 I must admit that the changes I made were very optimized for my M4 Max and this specific vision encoder. But as you mentioned above, it does seem to help with Whisper too.

Also, a lot of the PR diff is removing `prefer_subgroupshuffle`... which may not be good across other devices. @guschmue lmk what you think!

onnxruntime/contrib_ops/webgpu/bert/flash_attention.wgsl.template
xenova: reducing max_k_step to 64 doesn't hurt M4 Max performance

https://github.com/microsoft/onnxruntime/pull/27780/changes/64875156921be6e0ddbd8d4083c7268271099f51

xenova:
Ran some more benchmarks on some other models.

Model | ONNX file | Op | before | after | speedup
-- | -- | -- | -- | -- | --
onnx-community/all-MiniLM-L6-v2-ONNX | model.onnx | MultiHeadAttention | 4.36ms | 2.02ms | 2.16x
onnx-community/gemma-3-270m-it-ONNX | model.onnx | GroupQueryAttention | 3.65ms | 1.63ms | 2.24x
onnx-community/LightOnOCR-2-1B-ONNX | vision_encoder.onnx | MultiHeadAttention | 192s | 9.3s | 20.71x
onnx-community/LightOnOCR-2-1B-ONNX | decoder_model_merged.onnx | GroupQueryAttention | 14.3ms | 7.4ms | 1.96x

xenova:
been testing more and more... every model sees a 2-3x performance for the MHA nodes. Hoping we can get some benchmarking done on lower-end devices so we can fast-track the PR!

onnxruntime/contrib_ops/webgpu/bert/flash_attention.wgsl.template
qjia7: I found regressions on Qualcomm for phi4. It seems that register spilling happens (2s -> 11s for FlashAttention). Maybe we should keep the original path for qualcomm (not test Tiger Lake yet).
The baseline is as below:
baseline
With this PR:
opt

xenova: Thanks so much for testing! I will keep the qualcomm path based on this feedback.

xenova: added back qualcomm path. @qjia7 an you do another round of testing? That change doesn't affect current performance on M4.

Run | Op Name | Count | Total | Avg | Min | Max | % Total | Provider(s)
-- | -- | -- | -- | -- | -- | -- | -- | --
main | MultiHeadAttention | 168 | 693.627 ms | 4.129 ms | 10.0 us | 8.340 ms | 80.26% | WebGpu
this PR (w/o qualcomm path) | MultiHeadAttention | 168 | 77.239 ms | 459.8 us | 10.0 us | 917.0 us | 31.06% | WebGpu
this PR (w/ qualcomm path) | MultiHeadAttention | 168 | 77.270 ms | 459.9 us | 10.0 us | 920.0 us | 31.20% | WebGpu

guschmue:
extra cool!

guschmue:
I can give it a run on tiger lake in the afternoon.

xenova:
> extra cool!
> I can give it a run on tiger lake in the afternoon.

Great! Hopefully we see good performance 🤞

xenova:
Based on Guenther's feedback, I updated the implementation so that we only use my optimized branch for apple hardware. Everything else falls back to original implementation. I see the performance increase on my device (m4 max), and other hardware should produce the same benchmarks as before.

@qjia7 @guschmue PTAL 🙏

onnxruntime/contrib_ops/webgpu/bert/flash_attention.wgsl.template
qjia7: Verified on Qualcomm. The perf is back. Thanks.

onnxruntime/contrib_ops/webgpu/bert/flash_attention.h
```diff
@@ -87,13 +88,31 @@ class FlashAttentionProgram final : public Program<FlashAttentionProgram> {
...
         has_head_sink_(has_head_sink) {
+    if (is_apple) {
+      // On Apple GPUs, use an optimized loop-based path with dynamic max_k_step.
+      // Compute max_k_step from shared memory budget: k_tile + v_tile = 2 * element_size * head_size * max_k_step
+      const int element_size = is_fp16 ? 2 : 4;
+      int max_k_from_shm = 16384 / (2 * element_size * qkv_head_size);
+      if (max_k_from_shm >= 64) {
+        max_k_step_ = 64;
+      } else if (max_k_from_shm >= 32) {
+        max_k_step_ = 32;
```
qjia7, line 101: Your current method is to use more registers to improvement the performance. Do you measure that how much perf gap if we use max_k_step_ = 32  instead of  max_k_step_ = 64 for M4 Max? And how about using max_k_step_ = 32 plus subgroupShuffle compared with max_k_step_ = 64 for M4 Max? If they can get the similar performance, I prefer we use max_k_step_ = 32 for apple and nvdia, which can help reduce the register pressure (such as M1). My previous machine is NV and see very good improvement for whisper with max_k_step_ = 32.

xenova: sure I can test that.

xenova: okay, max_k_step_ = 32 has no noticeable performance difference vs. 64


max_k_step_ = 32 plus subgroupShuffle causes significant issues.

qjia7: Weird that max_k_step_ = 32 plus subgroupShuffle causes significant issues. Thanks for trying. The latest change looks good to me.

xenova: yeah -- weird that it only happens for apple. maybe an upstream implementation issue in dawn?

@guschmue I think we're good to merge? 😇

sroussey:
@xenova might fix the preprocessor directive whitespace issues in case that is holding this up. definitely looking forward to this improvement!

xenova:
> @xenova might fix the preprocessor directive whitespace issues in case that is holding this up. definitely looking forward to this improvement!

We're noticing some regressions for older (~M1) apple hardware... so we're still trying to figure out what the optimal setup looks like.

kokroo:
> > @xenova might fix the preprocessor directive whitespace issues in case that is holding this up. definitely looking forward to this improvement!
> 
> We're noticing some regressions for older (~M1) apple hardware... so we're still trying to figure out what the optimal setup looks like.

+1. The fixes don't really spped anything up on my M1 Pro.

xenova:
@qjia7 it would be great to be able to get this working (in such a way that doesn't affect other hardware). Any ideas?

xenova:
noticed again with https://huggingface.co/onnx-community/depth-anything-v2-small-ONNX where my branch is around 6x faster (460ms -> 75ms)

qjia7:
> @qjia7 it would be great to be able to get this working (in such a way that doesn't affect other hardware). Any ideas?

Could you gather more details from AdapterInfo and further constrain this to M4 Max? If so, we can move forward with landing it. I can also take a deeper look into the regression causes introduced by these changes. Hopefully, we can enable this on more devices—I'm seeing significant gains on my NV device as well.

For example, I can retrieve the adapter info as shown below. I expect you should also be able to distinguish M4 Max using the adapter information.
```
vendor="nvidia"
architecture="lovelace"
device="NVIDIA RTX 2000 Ada Generation Laptop GPU"
backend_type=4, vendor_id=4318, device_id=10424

**Comment by sroussey** [comment]
I have an M2 Max I can test. What is the quickest way to do so?

**Comment by xenova** [comment]
@qjia7 the only useful information I can see is probably `architecture : "metal-3"` (vendor is apple). everything else appears blank.

**Comment by qjia7** [comment]
> @qjia7 the only useful information I can see is probably `architecture : "metal-3"` (vendor is apple). everything else appears blank.

How about providing a WebGPU EP session option, something like 
```
ep.webgpuexecutionprovider.experimentalEnableAggressiveFlashAttention = "1"
```
**Benefits**:
Safe to land — off by default, zero regression risk on any device
Device-agnostic — anyone (Apple, NVIDIA, Intel) can opt in and test
Data-driven follow-up — once we collect enough benchmarks across devices, a future PR can auto-enable it for known-good architectures

**Note**: I see this as a temporary stepping stone, not a permanent solution. For follow-up work:

Root-cause the regressions on some unexpected devices — Having the opt-in flag makes it easy to A/B test on affected machines. We can do deeper analysis on why this regresses (register spilling? workgroup size mismatch? shared memory pressure?). There may also still be room to further optimize the current shader.

File a Dawn bug for richer GPU info — Currently Dawn's AdapterInfo only exposes architecture: "metal-3" for Apple, which isn’t sufficient to distinguish M4 Max from other variants. It would be helpful to ask for more detailed GPU identification. Once Dawn exposes that information, we can follow up with a PR to automatically enable the optimization on the appropriate devices and eventually deprecate the manual opt‑in.

What do you think? @xenova @guschmue

xenova:
Hmm, I think finding a way to ship without an experimental flag would be better... but as you say, it's not really possible to know which apple device we are running on. Maybe some other way to test?

Also, do we have benchmarks to see how this PR affects other apple devices? For example, I'd much rather prefer a 20x speed up (like this PR offers) on M4 max and higher than a minor 1.1x slowdown on m3 and below.

guschmue:
let me give this another try

guschmue (approved):

guschmue:
ok, I see no real regression so ok to merge.
on m4 macmini I see not gains but I enabled it locally on nvidia and it has good gains on prefill.
Will do a PR after this one is merged to also apply this to nvidia.

xenova:
great -- thanks @guschmue! 🙏  Got many demos planned for some of the OCR models which failed previously

PR #27786, opened 2026-03-20
Model Package Support

chilo-ms:
### Description
To support the model package design, one of the goals for ORT is to automatically select the most suitable compiled EPContext binary from a collection of precompiled variants based on the EP, provider options, metadata, and available devices.

This PR is for ORT to support first phase model package. There could be other follow-up PRs in the future.

A model package is a collection of models, binaries, and metadata files organized in a hierarchically structured directory.
The directory structure is not yet finalized, so the following is just a simple example of a model package directory:

````
<model>.ortpackage/  
├── manifest.json 
├── pipeline.json 
├── configs/ 
|   ├── genai_config.json 
|   └── chat_template.jinja  
└── models/  
    └── model_name/  
        ├── metadata.json 
        |   └── Contains general information on the component model, 
        |       and specific information about each model variant 
        |       such as data types, quantization algo, EP, etc. that  
        |       is updated on add/remove of model variant 
        └── shared_weights/ (shared weights from all variants) 
            └── <checksum of weights file A>/ 
                └── model.data 
            └── <checksum of weights file B>/  
                └── model.data 
            └── ... 
        └── base model/    
            ├── model.onnx  
        └── variant A /  
            ├── optimized model.onnx (contains EPContext nodes)  
            └── [Compilation artifacts]  
        └── variant B /  
            ├── optimized model.onnx (contains EPContext nodes)  
            └── [Compilation artifacts] 

````
#### Spec and Format:
See [here](https://github.com/microsoft/onnxruntime/blob/07e55627e75da24099c582331a0f786090e6382a/onnxruntime/core/session/model_package/README.md)
#### Definitions:

- Model Package 
   - A model package defines the overall logical ‘model’ 
   - A model package contains one or more ‘component models’ 

- Component Model 
  - A component model comprises one or more ‘model variants’ 

- Model Variant 
  - A ‘model variant’ is a single ONNX or ORT format model

#### manifest.json and metadata.json

A manifest.json may look like:

````
{ 
    "model_name":  <logical_model_name>,
    "component_models": [
        <component_model_name_1>,
        <component_model_name_2>
    ]
}
````

A metadata.json for a component model may look like:
````
{ 
    "component_model_name":  <component_model_name_1>,
    "model_variants": {
         <variant_name_1>:  {
             "file": <ep_context_model_1 onnx file>,
             "constraints": {
                 "ep": <ep_name>,
                 "device": <device_type>,
                 "architecture": <hardware_architecture>
             }
         },
         <variant_name_2>:  {
             "file": <ep_context_model_2 onnx file>,
             "constraints": {
                 "ep": <ep_name>,
                 "device": <device_type>,
                 "architecture": <hardware_architecture>
             }
         }   
    }
}
````
#### Model Selection

The selection logic is implemented in `MatchesVariant()`, which evaluates the following constraints:
(Note: A constraint refers to a value under the "constraints" field in either manifest.json or metadata.json.)

- Check ep constraint
- Check device constraint
  -   For some provider-bridge EPs, they may not implement `OrtEpFactory::GetSupportedDevices`, therefore ORT
      won't have the supported device information for those EPs. In that case, ORT will skip the device constraint validation for those EPs.
  -  If provider option contains key related to device type, then the value must match the device constraint if any. 
- Check ep_compatibility_info constraint
  - ORT does not directly evaluate the architecture constraint. Instead, it relies on the ep_compatibility_info constraint, which may encode architecture information if needed.
  - The ep_compatibility_info value is expected to match the EP compatibility string stored in the EPContext model metadata. (See OrtEp::GetCompiledModelCompatibilityInfo() for how this string is generated.)
  - The EP implementation of EpFactory::ValidateCompiledModelCompatibilityInfo() is responsible for validating the compatibility string against the target device (i.e. OrtHardwareDevice) and returning the compatibility result.

#### Note

Check the unit test [here](https://github.com/microsoft/onnxruntime/pull/27786/changes#diff-bfa4122a85543ae2d80bf4cf6d9f85248e51c2276a5956af32f9bd8c8983d23a) to better understand how to use model package. 

#### Code Change

This pull request introduces significant enhancements to the execution provider (EP) selection and management infrastructure in ONNX Runtime. The main focus is on supporting more sophisticated device selection and manifest-based model packaging, as well as refactoring provider selection logic for modularity and future extensibility.

Key changes include:

- Introduction of model package context and manifest parsing to support selecting model components based on device and EP constraints.
- Refactoring of the execution provider interface and related classes to support multiple devices per provider.
- Modularization of EP/device selection, creation, and registration logic in the provider policy context.

The most important changes are:

**Model Package Context and Manifest Support**
- Added new files `model_package_context.h` and `model_package_context.cc` to implement manifest parsing, device/EP constraint matching, and component selection logic for model packages. This enables ONNX Runtime to select the most appropriate model variant based on available hardware and EP configuration. [[1]](diffhunk://#diff-006078879d52b421c973e2880c65db474aad6b21ad81ba69d387df8661bafeb2R1-R78) [[2]](diffhunk://#diff-45c29f481077e424c8969dc2198a8b40ab5908cf3b0bbf25dbeaca3ec51935d5R1-R279)

**Execution Provider Interface Enhancements**
- Updated the `IExecutionProvider` class to support construction with a list of `OrtEpDevice` pointers, and added a `GetEpDevices()` method to retrieve the supported devices. This allows plugin and bridge EPs to expose multiple devices. [[1]](diffhunk://#diff-e15769e35b807986b812aae3ff7192269e171c5846b2ff4d8ec571ec8ed57aa4R87-R104) [[2]](diffhunk://#diff-e15769e35b807986b812aae3ff7192269e171c5846b2ff4d8ec571ec8ed57aa4R203-R207)
- Updated plugin EP construction to pass the list of supported devices to the base class.

**Provider Policy Context Refactoring**
- Refactored provider policy context logic to modularize device ordering, device selection, telemetry logging, EP creation, and registration. This includes splitting the monolithic `SelectEpsForSession` into smaller methods: `OrderDevices`, `SelectEpDevices`, `LogTelemetry`, `CreateExecutionProviders`, `RegisterExecutionProviders`, and a new flow for model package-based EP selection. [[1]](diffhunk://#diff-dd9f398bec3f054aed2c930af620e3e1bfcc5b4a5d5667c4b0cd1f60ddfffda0R53-R58) [[2]](diffhunk://#diff-dd9f398bec3f054aed2c930af620e3e1bfcc5b4a5d5667c4b0cd1f60ddfffda0L118-L156) [[3]](diffhunk://#diff-dd9f398bec3f054aed2c930af620e3e1bfcc5b4a5d5667c4b0cd1f60ddfffda0L225-R199) [[4]](diffhunk://#diff-dd9f398bec3f054aed2c930af620e3e1bfcc5b4a5d5667c4b0cd1f60ddfffda0R254-R365)

These changes collectively lay the groundwork for more flexible, robust, and extensible device and EP selection in ONNX Runtime, especially in scenarios involving packaged models with multiple variants and complex hardware environments.

### DIFF
--- a/onnxruntime/core/session/model_package/model_package_context.cc
+++ b/onnxruntime/core/session/model_package/model_package_context.cc
@@ -57,29 +57,6 @@ const OrtHardwareDevice* FindMatchingHardwareDevice(std::string_view device_cons
   return nullptr;
 }
 
-bool MatchesDeviceTypeProviderOption(std::string_view provider_option_key,
-                                     std::string_view provider_option_value,
-                                     std::string_view device_constraint) {
-  if (device_constraint.empty()) {
-    return true;
-  }
-
-  const auto key_lower = ToLower(provider_option_key);
-
-  // If provider option key is related to device type, then its value must match the device constraint value.
-  //
-  // Those keys are not standardized and are defined by EPs, e.g. "device_type" for OpenVINO EP, "backend_type" for QNN EP.
-  //   - "backend_type" has valid values: "cpu", "gpu", "htp" and "saver".
-  //   - "device_type" has valid values: "CPU", "GPU", "GPU.0", "GPU.1" and "NPU".
-  //
-  // TODO: In the future, we can consider standardizing the key for device type in provider options and make it more generic for all EPs to use.
-  if (key_lower == "device_type" || key_lower == "backend_type") {
-    return ToLower(provider_option_value).find(ToLower(device_constraint)) != std::string::npos;
-  }
-
-  return true;
-}
-
 Status ValidateCompiledModelCompatibilityInfo(const VariantSelectionEpInfo& ep_info,
                                               const std::string& compatibility_info,
                                               std::vector<const OrtHardwareDevice*>& constraint_devices,
@@ -154,20 +131,6 @@ bool MatchesVariant(ModelVariantInfo& variant, const VariantSelectionEpInfo& ep_
                        << "' does not match any device supported by EP '"
                        << ep_info.ep_name << "'. Skip this variant.";
     return false;
-  }
-
-  // Additional check for provider-bridge EPs only.
-  // If provider option contains key related to device type, then the value must match the device constraint if any.
-  // Gets the target device if matched.
-  if (ep_info.hardware_devices.empty()) {
-    for (const auto& [key, value] : ep_info.ep_options) {
-      if (!MatchesDeviceTypeProviderOption(key, value, variant.device)) {
-        LOGS_DEFAULT(INFO) << "Provider option '" << key << "' with value '" << value
-                           << "' does not match device constraint '" << variant.device
-                           << "'. Skip this variant.";
-        return false;
-      }
-    }
   }
 
   // 3) Check ep_compatibility_info constraint

### CODE REVIEW

**Comment by skottmckay** [comment]
Top-level manifest.json should define the overall inputs/outputs so a user of the package knows what it does. They shouldn't have to trawl through the information to find the first and last things that will be run to infer this info.

File: onnxruntime/test/autoep/test_model_package.cc [hunk 1]
```diff
@@ -0,0 +1,177 @@
...
+    "components": [
+      {
+        "variant_name": "variant_1",
+        "file": "mul_1.onnx",
+        "constraints": {
+          "ep": "example_ep",
+          "device": "cpu",
+          "architecture": "arch1"
+        }
+      },
```
skottmckay: This seems like lower level per-variant info I would have expected to be in the component model's metadata.json not the top level manifest.

skottmckay:
Would be good to add some details about the 'how' things are done as the PR description says what has changed but doesn't cover things like how selection is being implemented.

onnxruntime/core/session/provider_policy_context.h
```diff
@@ -40,14 +40,37 @@ class ProviderPolicyContext {
...
+                      const std::vector<const OrtEpDevice*>& execution_devices,
+                      const std::vector<const OrtEpDevice*>& devices_selected);
+
+  Status CreateExecutionProviders(const Environment& env, InferenceSession& sess,
+                                  std::vector<const OrtEpDevice*>& devices_selected,
+                                  std::vector<std::unique_ptr<IExecutionProvider>>& providers);
+
+  Status RegisterExecutionProviders(InferenceSession& sess,
+                                    std::vector<std::unique_ptr<IExecutionProvider>>& providers);
+
```
skottmckay: Does anything call these? #Closed

onnxruntime/core/session/model_package_context.h
```diff
@@ -0,0 +1,78 @@
...
+  std::string ep_name;
+  OrtEpFactory* ep_factory;
+  std::vector<const OrtEpDevice*> ep_devices;
+  std::vector<const OrtHardwareDevice*> hardware_devices;
+  std::vector<const OrtKeyValuePairs*> ep_metadata;
+  ProviderOptions ep_options;
+};
+
+class ModelPackageManifestParser {
+ public:
```
skottmckay: Would be good to keep the model package handling (parsing manifest and metadata files, iterating directories etc. to create a user-friendly in-memory representation) standalone as it will be needed in other places like Foundry Local

onnxruntime/test/autoep/test_model_package.cc
chilo-ms: I think where to store this lower-level, per-variant metadata is still open for discussion. It can either reside in the top-level manifest or within each component model’s metadata.json.

Option 1: Top-level manifest:
- Pros:
  - Provides a single source of truth for the entire model package. 
  - Simplifies parsing for ORT, as it only needs to read one manifest file to obtain a complete view.

- Cons: 
  - The manifest may become overly detailed, containing extensive information about each precompiled variant.
  - Requires synchronization with the component model directories.
If component model directories are added or removed, the manifest must be updated accordingly

Option 2: Component model's metadata.json:
  - Pros:
    - Better modularity and separation of concerns. 
    - Changes to component model directories typically do not require updates to the top-level manifest.
    
- Cons:
  - ORT must scan and parse all component model directories to collect per-variant metadata.
  - May introduce additional runtime overhead during model package loading.
  

@devang-ml

onnxruntime/core/session/provider_policy_context.h
chilo-ms: good catch, that's unnecessary anymore and will remove it

onnxruntime/test/autoep/test_model_package.cc
skottmckay: Either way the same amount of info needs to be parsed. Should be trivial to scan the component model directories as the metadata should be in the top-level directory for each component model. 

I'm wary about the top-level manifest getting overwhelmingly large, especially if we have multiple component models all with multiple variants. Harder to see issues/inconsistencies and easier to get mega-config files wrong. But maybe humans will never read this stuff and it doesn't matter anymore. 

Option 2 slightly simplifies add/remove variant from package as you should only need to update the component model metadata file when doing so. For Foundry Local we will need to add/remove frequently as we will have to publish per-variant packages due to catalog being immutable and to keep the versioning specific to the variant, and merge those on-device post-download. e.g. user downloads TRT-RTX variant and CPU variant separately as they are different entries in the catalog.

I don't quite understand how Option 2 adds runtime overhead. I would have expected we have a general model package helper class. Create it by pointing it at the package directory and it parses everything into in-memory info at that point and checks the package is valid. Via that class instance I should be able to easily get things like the ordered list of component models, the available variants for each component model and things like the EP they require, and a way to get the directory of a variant to load it.

onnxruntime/core/session/utils.cc
```diff
@@ -289,6 +390,99 @@ static OrtStatus* CreateSessionAndLoadModelImpl(_In_ const OrtSessionOptions* op
...
+    }
+
+    // Build EP info from finalized providers.
+    std::vector<SelectionEpInfo> ep_infos;
+    ORT_API_RETURN_IF_STATUS_NOT_OK(GetSelectionEpInfo(options_to_use, provider_list, ep_infos));
+
+    // Parse manifest and gather components.
+    ModelPackageManifestParser parser(logging::LoggingManager::DefaultLogger());
+    std::vector<EpContextVariantInfo> components;
+    ORT_API_RETURN_IF_STATUS_NOT_OK(parser.ParseManifest(package_root, components));
```
skottmckay: nit: would suggest having a ModelPackage class that owns all this info instead of doing things piecemeal in other places. Construct the ModelPackage from the package_root. It parses and validates the info and provides getters for the code to read that. That way all the parsing and processing of the model package is in one class so if there are any issues there's one place to fix them.

This also feels like it's missing a layer. The package has one or more component models (if multiple there's a specific order they're executed in). A component model has one or more variants. But this code is reading a collection of EpContextVariantInfo so the required grouping by component model seems to be lost.  #Closed

onnxruntime/core/session/model_package_context.h
```diff
@@ -0,0 +1,80 @@
...
+
+class ModelPackageContext {
+ public:
+  ModelPackageContext() = default;
+
+  // Select components that match the provided EP/device info. If multiple
+  // components match, the one with the highest device score is chosen.
+  Status SelectComponent(gsl::span<EpContextVariantInfo> components,
+                         gsl::span<SelectionEpInfo> ep_infos,
+                         std::optional<std::filesystem::path>& selected_component_path) const;
```
skottmckay: nit: A 'context' owning the selection logic feels slightly off. Maybe that's just a naming thing as this seems more like it's implementing a selection policy for a model variant (which != component model).

chilo-ms: i was mixing the use of component model and model variant. 
You are right, here is implementing a selection policy for a model variant. I will change the naming.

onnxruntime/core/session/utils.cc
chilo-ms: A ModelPackage class that owns all the info is a good suggestion and will change the code here.
Also, i did miss the "component model" layer and will add it back.

onnxruntime/test/autoep/test_model_package.cc
chilo-ms: While we are still discussing internally, i made the `ModelPackageManifestParser::ParseManifestbe` able to parse manifest.json and metadata.json for all component models as well as their associated model variants.

If a variant appears in both, it choses metadata.json as the source of truth, but falls back to manifest.json
if metadata.json is missing required fields.

onnxruntime/core/session/model_package/model_package_context.h
```diff
@@ -0,0 +1,70 @@
...
+  // Compute a score for a variant
+  int CalculateVariantScore(const ModelVariantInfo& variant) const;
+};
+
+class ModelPackageContext {
+ public:
+  ModelPackageContext(const std::filesystem::path& package_root);
+
+  // Select the most suitable model variant
+  Status SelectModelVariant(gsl::span<SelectionEpInfo> ep_infos);
```
skottmckay: nit: I would put this in a separate class so that the ModelPackageContext is a simple data provider. There may be various usages of that data of which selecting a model variant is one (but selection logic may differ in different usages) so I think it's better to decouple. instead provide a const ModelPackageContext as input to a function that uses the data. #Closed

onnxruntime/core/session/model_package/model_package_context.cc
```diff
@@ -0,0 +1,335 @@
...
+  }
+
+  return nullptr;
+}
+
+bool MatchesDeviceTypeProviderOption(std::string_view provider_option_key,
+                                     std::string_view provider_option_value,
+                                     std::string_view device_constraint,
+                                     gsl::span<const OrtHardwareDevice* const> hardware_devices,
+                                     std::vector<const OrtHardwareDevice*>& constraint_devices) {
```
skottmckay: Why are we matching against provider option keys? That seems very fragile especially if we're hardcoding arbitrary EP specific values to look for. 

Do we not have enough info in the metadata for the variant to match against the OrtEpDevice (which includes OrtHardwareDevice) properties that we contro? e.g. that has the EP name, and the OrtDevice. If not, what would we need to add to be able to do that?

onnxruntime/core/session/model_package/model_package_context.cc
```diff
@@ -0,0 +1,335 @@
...
+                       << ep_info.ep_name << "'. Skip this variant.";
+    return false;
+  }
+
+  // 2) Check device constraint
+  bool device_ok = variant.device.empty();
+
+  // For some provider-bridge EPs, they may not implement OrtEpFactory::GetSupportedDevices, therefore ORT
+  // won't have the supported device information for those EPs and ep_info.hardware_devices will be empty.
+  // In that case, we will skip the device constraint validation for those EPs.
```
skottmckay: nit: might be clearer to say "EPs that don't implement the OrtEpFactory interface". 

If an EP implements the plugin EP factory (full plugin or provider bridge) they should always implement GetSupportedDevices - otherwise they will never produce an OrtEpDevice and there's no point them implementing any part of OrtEpFactory. 

Can we get the IExecutionProvider instance for the EP and use the OrtDevice returned by GetDevice() in the check?

onnxruntime/core/session/utils.cc
```diff
@@ -179,14 +256,55 @@ common::Status CopyStringToOutputArg(std::string_view str, const char* err_msg,
...
+      }
+    }
+  }
+#if !defined(ORT_MINIMAL_BUILD)
+  else {
+    // TEMPORARY for testing. Manually specify the EP to select.
+    auto auto_select_ep_name = sess.GetSessionOptions().config_options.GetConfigEntry("test.ep_to_select");
+    if (auto_select_ep_name) {
+      ORT_RETURN_IF_ERROR(TestAutoSelectEPsImpl(sess.GetEnvironment(), sess, *auto_select_ep_name));
+    }
```
skottmckay, line 323: This can probably be removed. Was from before the EP selection policies were implemented. Now that they exist we shouldn't need it. If there's any test still using it the test should be updated to set a selection policy delegate instead.

onnxruntime/core/session/utils.cc
```diff
@@ -289,6 +407,108 @@ static OrtStatus* CreateSessionAndLoadModelImpl(_In_ const OrtSessionOptions* op
...
+      // Build EP info from finalized providers.
+      std::vector<SelectionEpInfo> ep_infos;
+
+      ORT_API_RETURN_IF_STATUS_NOT_OK(GetSelectionEpInfo(options_to_use, provider_list, ep_infos));
+
+      if (ep_infos.empty()) {
+        return OrtApis::CreateStatus(ORT_FAIL,
+                                     "No execution providers were provided or selected for model variant selection. "
+                                     "Try specifying the model file path instead of a model package, or check the "
+                                     "EP selection policy and provider factories.");
```
skottmckay: Is there output somewhere that gives the user the list of available EPs and the list of EPs the package supports so it's easy to see what is required to fix this? As a user, if it fails, I'd love to be able to quickly check the output and know whether I need a different EP to be available (could be my registration code for example), or whether I need to download a different variant for the package.

onnxruntime/core/session/model_package/model_package_context.h
```diff
@@ -0,0 +1,70 @@
...
+  std::string ep{};
+  std::string device{};
+  std::string architecture{};
+  std::string compatibility_info{};
+  std::unordered_map<std::string, std::string> metadata{};
+  OrtCompiledModelCompatibility compiled_model_compatibility{};
+  std::filesystem::path model_path{};
+};
+
+struct SelectionEpInfo {
```
skottmckay: nit: name is a little confusing (at least to me). would VariantSelectionEpInfo be clearer? or EpSelectionInfo if it's inside a class that handles selection (so name would be something like ModelVariantSelector::EpSelectionInfo).  #Closed

onnxruntime/core/session/model_package/model_package_descriptor_parser.cc
```diff
@@ -0,0 +1,449 @@
...
+//
+// {
+//     "model_name" : <logical_model_name>,
+//     "component_models" : { // optional, if missing, ORT will discover component models by looking for folders with
+//                            // metadata.json under model_package_root/models
+//         <model_name_1> : {
+//            ...             // Could be empty.
+//         }
+//     }
+// }
```
skottmckay: Can we create some structs to formalize the schema and use the `.get<T>(...)` functionality from nlohmann json? I believe you implement a to_json function to support that.

e.g. if I have this:
```cpp
// define struct to formalize schema
struct ChatCompletionRequest {
  std::string model;                                        // "model"
   ...
};

// have static method to deserialize
void from_json(const nlohmann::json& j, ChatCompletionRequest& r) {
  // Required fields
  r.model = j.at("model").get<std::string>();
  ...
```

I can do this:
```cpp
      req_json = nlohmann::json::parse(body_str->c_str());

    // Deserialize into typed request struct
    ChatCompletionRequest req = req_json.get<ChatCompletionRequest>();
```

Easy to see what the supported format is and ensures it's current vs. comments which are likely to get out of date as someone has to remember to update them.

onnxruntime/core/session/model_package/model_package_descriptor_parser.cc
```diff
@@ -0,0 +1,449 @@
...
+// or
+//
+// {
+//     "model_name" : <logical_model_name>,
+//     "component_models" : {
+//         <model_name_1> : {
+//             "model_variants" : {
+//                 <variant_name_1> : {
+//                     "model_type": "onnx",
+//                     "file" : <ep_context_model_1 onnx file>,
```
skottmckay:
```suggestion
//                     "model_file" : <ep_context_model_1 onnx file>,
```

onnxruntime/core/session/model_package/model_package_descriptor_parser.cc
```diff
@@ -0,0 +1,449 @@
...
+// }
+//
+// or
+//
+// {
+//     "model_name" : <logical_model_name>,
+//     "component_models" : {
+//         <model_name_1> : {
+//             "model_variants" : {
+//                 <variant_name_1> : {
```
skottmckay: Is `variant_name_1` the directory name containing the variant? That makes it easy to find where the files are (subdirectory of the component model directory) and we don't need a specific value for that in the metadata. 

We may want a model_id property which I expect would match a catalog value when the model comes from there. 

e.g. if 'phi4-cpu:1' is the model id from the catalog it might be in a directory called `<component_model_dir>/phi4-cpu_1` and the model filename may just be `model.onnx`

onnxruntime/core/session/model_package/README.md
```diff
@@ -0,0 +1,180 @@
...
+            └── [Compilation artifacts] 
+        └── variant B / 
+            ├── optimized model.onnx (contains EPContext nodes) 
+            └── [Compilation artifacts] 
+````
+
+
+## Notes:
+- Only one component model is allowed in the package for now, but the format allows for multiple component models in the future.
+- If it's an ORT GenAI model variant, the variant directory contains all GenAI files. ORT GenAI used as-is.
```
skottmckay: If we're moving the GenAI config up isn't this out of date and it's strictly one ONNX model in a variant directory? i.e. no longer 'a variant directory contains a complete GenAI model' as I had in my simple 'step 1' approach.

That would also mean multiple component models are required to support things like Whisper where there are separate encoder and decoder models.

onnxruntime/core/session/model_package/README.md
```diff
@@ -0,0 +1,180 @@
...
+    "model_name":  <logical_model_name>,
+    "component_models": {
+        <model_name_1>: {
+            "model_variants": {
+                <variant_1>: {
+                    "model_type": "onnx",
+                    "file": "model_ctx.onnx",
+                    "constraints": {
+                        "ep": "TensorrtExecutionProvider",
+                        "ep_compatibility_info": "device=gpu,npu;cuda_driver_version_support=..."
```
skottmckay: Is this the compatibility info the EP factory creates during compilation or something we add during package creation? 

We have zero control over the EP factory generated compatibility string but need that to let the EP factory check if the model is compatible. It could be a base64 encoded string with binary data.

onnxruntime/core/session/model_package/README.md
```diff
@@ -0,0 +1,180 @@
...
+    }
+}
+```
+
+
+## `metadata.json` (required per component model)
+
+Location: `<package_root>/models/<component_model>/metadata.json`
+
+Purpose: Describes the variants available for a specific component model.
```
skottmckay, line 86: Variant metadata should be in one place. We either put everything in the manifest, or we put the component model specific stuff here and the manifest doesn't need a component_models value as content of that is the same as merging the data from the /models/.../metadata.json files.

onnxruntime/core/session/model_package/README.md
chilo-ms: https://github.com/microsoft/onnxruntime/pull/27786/changes/aed65c6889703bd6989c259c26e758295c56e554

My last commit updated the GenAI config and removed the "Only one component model is allowed in the package for now". I think that commit was not included when you were reviewing the code.

onnxruntime/core/session/model_package/model_package_context.cc
chilo-ms: An EP can support multiple OrtEpDevices, and application uses "device type" provider option to let ORT/EP know which specific device to run on. In this case, knowing multiple OrtEpDevices are supported is not enough to make selection. ORT might still need to check "device" constraint of the model variant against that "device type" provider option in order to select the suitable "device specific" EPContext model.

Currently OpenVINO EP and QNN EP has such provider options. TRT EP has one as well.

Agree that adding hardcoding code into selection logic is very fragile, but it seems inevitable to me for now.
In the future, we can consider standardizing the key for device type in provider options and make it more generic for all EPs to use.

skottmckay: Not quite following. With the V2 setup you have auto EP or manually registration. 

If manually registering for an EP that has multiple OrtEpDevice instances you can provide one or more. Doesn't that explicitly say the device/s to use? e.g. if I have a GPU and an NPU OrtEpDevice for the EP I could register just the GPU, just the NPU, or both. 

If doing auto selection from a model package you'd need to set a selection policy, and I would have expected that to be the mechanism to determine the OrtEpDevice/s to register.

chilo-ms: sorry for not making it clear, i'm referring to manually registration use case not auto EP.

> If manually registering for an EP that has multiple OrtEpDevice instances you can provide one or more. Doesn't that explicitly say the device/s to use? e.g. if I have a GPU and an NPU OrtEpDevice for the EP I could register just the GPU, just the NPU, or both

What about provider-bridge EP case? 
If manually registering for a provider-bridge EP that supports multiple devices via old SessionOptionsAppendExecutionProvider_XXX on a system that has multiple devices. There is no place to explicitly specify device/s to use except in provider option. In such case, i think ORT might still need to check provider option to know what device to run on and choose suitable model accordingly.

Do we want model package to support provider-bridge EPs?

onnxruntime/core/session/model_package/model_package_context.cc
chilo-ms: > Can we get the IExecutionProvider instance for the EP and use the OrtDevice returned by GetDevice() in the check?

yes, this PR already did.

It [calls](https://github.com/microsoft/onnxruntime/blob/chi/model_package_1/onnxruntime/core/session/utils.cc#L462) GetSelectionEpInfo() to get the OrtEpDevices and then get OrtHardwareDevices from there.

````c++
Status GetSelectionEpInfo(const OrtSessionOptions* session_options,
                          std::vector<std::unique_ptr<IExecutionProvider>>& provider_list,
                          std::vector<SelectionEpInfo>& ep_infos) {

  ...
 
  // Add ep devices to ep_info
  auto& ep_devices = provider->GetEpDevices();
  ep_info.ep_devices = ep_devices;
  
  ...
 
  // Add hardware devices to ep_info
  ep_info.hardware_devices.reserve(ep_devices.size());
  for (const auto& ep_device : ep_devices) {
    if (ep_device->device != nullptr) {
      ep_info.hardware_devices.push_back(ep_device->device);
    }
  }
  ...
 }
 ````
  


and check the OrtHardwareDevice against the constraint [here](https://github.com/microsoft/onnxruntime/blob/chi/model_package_1/onnxruntime/core/session/model_package/model_package_context.cc#L157).

File: onnxruntime/core/session/utils.cc (hunk 10 above)
Comment (chilo-ms) [reply onnxruntime/core/session/utils.cc]: Added a `PrintAvailableAndSelectedEpInfos()` function.

File: onnxruntime/core/session/model_package/model_package_context.h (hunk 11 above)
Comment (chilo-ms) [reply onnxruntime/core/session/model_package/model_package_context.h]: Changed to use `VariantSelectionEpInfo`

File: onnxruntime/core/session/model_package/model_package_descriptor_parser.cc (hunk 12 above)
Comment (chilo-ms) [reply onnxruntime/core/session/model_package/model_package_descriptor_parser.cc]: okay, i added following structs to formalize the schema:
VariantConstraintsSchema, VariantSchema, ComponentSchema, ManifestSchema and MetadataSchema

File: onnxruntime/core/session/model_package/README.md (hunk 16 above)
Comment (chilo-ms) [reply onnxruntime/core/session/model_package/README.md]: > Is this the compatibility info the EP factory creates during compilation or something we add during package creation?

It's the compatibility info the EP factory creates during compilation.

When using Olive to create the model package, Olive should parse the ONNX model's metadata to get that compatibility string and creates this "ep_comaptibility_info" constraint. If manually creating the package, users need to manually add this constraint, and make sure the string is the same as the compatibility string in model metadata.

As for base64 encoded string with binary data, i think it can still be converted into a valid string and can be passed around.

File: onnxruntime/core/session/model_package/model_package_descriptor_parser.cc (hunk 14 above)
Comment (chilo-ms) [reply onnxruntime/core/session/model_package/model_package_descriptor_parser.cc]: > Is variant_name_1 the directory name containing the variant?

yes, please see the code.
````C++
    ...
    ModelVariantInfo variant{};
    const std::filesystem::path model_dir = component_model_root / variant_name;
    std::filesystem::path resolved_model_path;

    if (chosen_variant->model_file.has_value()) {
      const std::filesystem::path candidate_path = model_dir / *chosen_variant->model_file;
      
    ...
````

https://github.com/microsoft/onnxruntime/blob/chi/model_package_1/onnxruntime/core/session/model_package/model_package_descriptor_parser.cc#L484

Comment (chilo-ms) [reply onnxruntime/core/session/model_package/model_package_descriptor_parser.cc]: What is the format of a model id from the catalog?
Is it "<model_variant>:\<id\>"?

File: onnxruntime/core/session/model_package/README.md (hunk 17 above)
Comment (chilo-ms) [reply onnxruntime/core/session/model_package/README.md:86]: > Variant metadata should be in one place

agree, but i think we don't have a final decision to put these in manifest or component model directory, right? 

I just thought about one use case that makes putting variant metadata in manifest not a good idea.
If application calls session creation API and passes in component model's directory, then only metadata.json of that component model can be accessible. It doesn't make sense for ORT to recursively get the current folder's parent folder to get the manifest.

BTW, i think "passing in component model's directory" use case should be supported, as i would image ORT GenAI runs the model with the given model package by creating/running multiple sessions (passing in different component model's directory) in order. Correct me if i'm wrong.

Comment (skottmckay) [reply onnxruntime/core/session/model_package/README.md:86]: Good point. Ideally ORT knows as little as possible about the rest of the model package so having to look in a parent directory and parse a file full of info that has nothing to do with it feels incorrect. I agree that it should support being given a component model directory as conceptually that represents the 'model' it's loading. 

If everything is in the manifest I think ORT would need the variant info for the model to be passed to it when creating the session, but that might be pretty cumbersome and potentially fragile.

File: onnxruntime/core/session/model_package/model_package_descriptor_parser.cc (hunk 14 above)
Comment (skottmckay) [reply onnxruntime/core/session/model_package/model_package_descriptor_parser.cc]: model id = `<model name>:<version>`. the model name typically has the variant in it like 'cuda' or 'qnn'.

e.g.
```
          "id": "Phi-4-generic-cpu:1",
          "name": "Phi-4-generic-cpu",
          "version": 1,
```

onnxruntime/core/session/model_package/model_package_context.cc
skottmckay: CUDA EP implements OrtEpFactory so it creates the OrtEpDevice instances.
WebGPU EP now has a plugin version .
TRT EP is a gap but if we need to we could implement OrtEpFactory for it.

Given that, are there any EPs we care about where the V2 registration can't be used?

onnxruntime/core/session/model_package/README.md
skottmckay: My point was more that we have no control over what's in the string but the example gives an impression that the data is somewhat structured and predictable. If there's a string created by Olive with more structured data we could use that, but AFAIK that would not be the same as the compat string the EP factory uses.

chilo-ms: i see, i replaced the example compat string with "...". 
Olive won't create more structured data; it should just copy and paste the compat string from model metadata.

onnxruntime/core/session/model_package/README.md
chilo-ms: I updated the spec and ORT to allow variant metadata only in the component model metadata.json.

onnxruntime/core/session/model_package/model_package_context.cc
chilo-ms: The EPs that we care about all have a plugin version. I removed the check for "device type" provider option.

Issue #27062, opened 2026-01-19
[Mobile] [Bug] onnxruntime-react-native fails to load model in Expo 54 standalone iOS


shaneeelliott:
### Describe the issue

InferenceSession.create() consistently fails to load ONNX models on iOS with the error:

Can't load a model: failed to load model

The same model loads and runs successfully on Android using identical code.

### To reproduce

1. Initial Setup
Model bundled via Metro: require('@/assets/models/v16/resnet18/v1/model.onnx')
Resolved using expo-asset:
const asset = Asset.fromModule(MODEL_PATH);
await asset.downloadAsync();
const modelPath = asset.localUri;
2. Applied Workaround from Issue #26738
Copied model from app bundle to writable directory:

const destPath = `${FileSystem.documentDirectory}model.onnx`;
await FileSystem.copyAsync({
    from: asset.localUri,
    to: destPath,
});
// Remove file:// prefix for iOS
const cleanPath = destPath.replace('file://', '');
await InferenceSession.create(cleanPath);
Result: Still fails with same error

3. Tried ORT Format
Converted model to 
.ort
 format
Same workaround applied
Result: Same failure
4. Session Options Tested
const sessionOptions = {
    executionProviders: ['cpu'], // Also tried with CoreML
    logSeverityLevel: 0, // Verbose logging
    graphOptimizationLevel: 'all',
    enableCpuMemArena: true,
    enableMemPattern: true,
};
Result: No additional error details, same generic failure

Observed Behavior
iOS (Failing)
LOG  [LocalInferenceService] Final model path: /var/mobile/Containers/Data/Application/.../Documents/model.onnx
LOG  [OnnxModelRunner] Loading model from: /var/mobile/Containers/Data/Application/.../Documents/model.onnx
LOG  [OnnxModelRunner] 🚀 Execution providers configured: ["cpu"]
WARN [OnnxModelRunner] ⚠️ Failed to load with hardware acceleration, trying CPU only...
WARN [OnnxModelRunner] Hardware acceleration error: Can't load a model: failed to load model
ERROR [OnnxModelRunner] ❌ Failed to load model: Can't load a model: failed to load model
The error occurs at the native layer (OnnxruntimeSessionHandler#loadModel) with no additional details.

Android (Working)
Same code, same model
Loads successfully
Inference runs correctly
Model Details
Type: ResNet18-based segmentation model
Input: [1, 768, 1024, 3] (float32)
[...]

Issue #26738, opened 2025-12-06
[Mobile] onnxruntime-react-native fails to load bundled ONNX assets in Expo standalone iOS


acrdlph:
## Environment

- React Native / Expo (standalone iOS build)
- `onnxruntime-react-native` (version: 1.23.2)
- Hermes
- Models bundled via Metro using `expo-asset`

---

## Problem

In the Expo dev client, loading a bundled model via  
`Asset.fromModule(...).localUri` works.

In the standalone iOS build,  
`InferenceSession.create(localUri)` fails when the model stays in the app bundle.

- The `localUri` is a `file://` path inside the app bundle.
- The session init fails with a generic **“failed to load model”** error.
- No model bytes are read unless we first copy the file out of the bundle.

---

## Observed Behavior

- `InferenceSession.create` rejects when `uri` points to the bundled file  
  (`file://...` in the app sandbox).
- No clear error about:
  - Path scheme
  - Bundle accessibility

---

## Workaround Applied

Copy the bundled model to a writable path before loading:

```ts
const dest =
  FileSystem.documentDirectory + basename(asset.localUri);

await FileSystem.copyAsync({
  from: asset.localUri,
  to: dest,
});
```

### To reproduce

In an expo project with `onnxruntime-react-native` installed:

1. Bundle a model (e.g., `silero_vad_v5.onnx`) via `require(...)`.
2. Resolve with:
   - `const asset = Asset.fromModule(mod);`
   - `await asset.downloadAsync();`
   - `const uri = asset.localUri;`
3. Call `InferenceSession.create(uri)` (or the equivalent load API).
4. Observe failure in a standalone iOS build; the same code works in the Expo dev client.

### Urgency

Not urgent, since workaround found.

### Platform

iOS

### OS Version

26

### ONNX Runtime Installation

Released Package

### Package Name (if 'Released Package')

onnxruntime-react-native

### ONNX Runtime Version or Commit ID

1.23.2

### ONNX Runtime API

Other / Unknown

### Architecture

Other / Unknown

### Execution Provider

Other / Unknown

shaneeelliott:
I have the same issue as this but the workaround of copying the bundle to local destination is not fixing the model loading error. I have no issue on android. Are there any other solutions to try?

kevinlam92:
@shaneeelliott

Can you try the solution from this PR: https://github.com/microsoft/onnxruntime/pull/27795 and let me know if it helped?

Issue #17773, opened 2023-10-03
[Build] React Native app does not work when i import onnx_runtime_react_native


faithyeenxin:
### Describe the issue

the aim of my react native expo app is to run a machine learning model that takes in an image and returns a value.
i followed the repository to start up [ort-rn-hello-world](https://github.com/fs-eire/ort-rn-hello-world) and got the model to run on my simulators within my computer. ie using npm start ios // npm start android.
however when i try to start it with expo start it throws me an error.

```bash
TypeError: Cannot read property 'install' of null, js engine: hermes
at node_modules/react-native/Libraries/Core/ExceptionsManager.js:105:15 in reportException
at node_modules/react-native/Libraries/Core/ExceptionsManager.js:150:4 in handleException
at node_modules/react-native/Libraries/Core/setUpErrorHandling.js:26:18 in handleError
at node_modules/expo/build/errors/ExpoErrorManager.js:27:16 in createErrorHandler
at node_modules/expo/build/errors/ExpoErrorManager.js:32:8 in <anonymous>
at node_modules/@react-native/js-polyfills/error-guard.js:51:16 in ErrorUtils.applyWithGuard
at node_modules/metro-runtime/src/polyfills/require.js:237:19 in guardedLoadModule
at http://192.168.203.40:19000/index.bundle?platform=ios&dev=true&hot=false&strict=false&minify=false:null in metroRequire
at http://192.168.203.40:19000/index.bundle?platform=ios&dev=true&hot=false&strict=false&minify=false:null in global

Invariant Violation: "main" has not been registered. This can happen if:
* Metro (the local dev server) is run from the wrong folder. Check if Metro is running, stop it and restart it in the current project.
* A module failed to load due to an error and `AppRegistry.registerComponent` wasn't called., js engine: hermes
at node_modules/react-native/Libraries/Core/ExceptionsManager.js:105:15 in reportException
at node_modules/react-native/Libraries/Core/ExceptionsManager.js:150:4 in handleException
- ... 7 more stack frames from framework internals

```

i followed the steps on the repo linked to the T - in his readme. 
however instead of expo prebuild i did expo eject 

this is my repo if you would like to inspect: [expo onnx rn app](https://github.com/faithyeenxin/expo_rn_onnx)

also, not sure if it's related but i am unable to get npm run ios on my connected mobile device. it does not seem to recognize that my device is connected. 

### Urgency

i need to get this working by 4 oct

### Target platform

iOS, MacOS

### Build script

[...]

Issue #17017, opened 2023-08-05
[Mobile] Failed to load onnx model on real iPhone device (but works on both Android and iOS simulator)


l3utterfly:
### Describe the issue

Loading an model with `Inference.create` does not work when running on real iOS devices.

It fails with the error: "Can't load a model: failed to load model from buffer"

The same code works on Android emulator, real Android device, and iOS emulator. It only fails on a real iPhone. Tested with iPhone 12,13,14

### To reproduce

Export a model from python. Load using:

```
let assetPath: string;
let base64Str: string;

if (Platform.OS === 'android') {
    assetPath = `models/${TFIDF_MODEL_FILENAME}`;
    base64Str = await RNFS.readFileAssets(assetPath, 'base64');
} else {
    assetPath = `${RNFS.MainBundlePath}/${TFIDF_MODEL_FILENAME}`;
    base64Str = await RNFS.readFile(assetPath, 'base64');
}

const bytes = base64.toByteArray(base64Str);
const array = new Uint8Array(bytes);

let buf = array.buffer as ArrayBufferLike;

// load a model
this.session = await InferenceSession.create(buf);
```

### Platform

iOS

### OS Version

16.2

### ONNX Runtime Installation

Released Package

### Package Name (if 'Released Package')

onnxruntime-react-native

### ONNX Runtime Version or Commit ID

1.15.1

### ONNX Runtime API

JavaScript

### Architecture

ARM64

### Execution Provider

Default CPU

yufenglee:
@skottmckay

skottmckay:
@YUNQIUGUO could you please take a look?

l3utterfly:
[tfidf_v2.zip](https://github.com/microsoft/onnxruntime/files/12321127/tfidf_v2.zip)

This is the model in question if it helps. It's a TF-IDF classifier.

YUNQIUGUO:
> [tfidf_v2.zip](https://github.com/microsoft/onnxruntime/files/12321127/tfidf_v2.zip)
> 
> This is the model in question if it helps. It's a TF-IDF classifier.

thanks for providing the model. that will be helpful. will try to see if I can reproduce the error locally. 

And fyi, I just tested a simple mnist.ort (https://github.com/microsoft/onnxruntime/blob/main/js/react_native/e2e/src/mnist.ort) model on iphone (12) device with the same release version 1.15.1 ort react native package in an Expo app. it successfully loads the model and is able to create an ort inference session. etc.

[...]

PR #27795, opened 2026-03-21
[React Native] Add ONNX model asset bundling to Expo config plugin

kevinlam92:
### Description

Enhances the existing Expo config plugin to automatically bundle `.onnx` and `.ort` model files into native app builds via a new `models` config option in `app.json`.

```json
["onnxruntime-react-native", { "models": ["./assets/models/my_model.onnx"] }]
```

Running `expo prebuild` copies each model into the correct platform-specific location:

- **iOS**: Xcode source root + "Copy Bundle Resources" build phase (via `IOSConfig.XcodeUtils.addResourceFileToGroup`)
- **Android**: `android/app/src/main/assets/`

When no `models` config is provided, the plugin behaves identically to the existing version.

### Motivation

I ran into this issue firsthand in my own Expo project. Models loaded fine in the dev client and simulator, but silently failed in production EAS builds with a generic "Can't load model" error. It took me a while to figure out the root cause: adding `onnx`/`ort` to Metro's `assetExts` makes the JS bundler aware of the files, but **does not guarantee they end up in the native `.app` bundle (iOS) or `assets/` directory (Android)** in EAS archive builds. The distinction between Metro asset resolution and native bundle inclusion isn't obvious, and the error messages don't point you in the right direction.

Looking through the issue tracker, it seems like a number of other Expo users have hit variations of this same problem (see linked issues below). The failure mode is particularly frustrating because everything works in development. The failure only surfaces in production builds.

Other RN libraries that need domain-specific binary assets in the native bundle (e.g., `expo-notifications` for sound files) solve this in their own config plugins. This PR follows the same pattern for ONNX model files.

### Changes

- `js/react_native/app.plugin.js`: Model bundling for both platforms, Metro `assetExts` validation, structured logger
- `js/react_native/README.md`: "Bundling ONNX Models (Expo)" section with config and runtime loading examples using `expo-asset`/`expo-file-system`
- `js/react_native/package.json`: Jest `moduleNameMapper` for test support
- `js/react_native/__tests__/app.plugin.test.js`: 7 unit tests

### Updated Developer Experience

Before this PR, an Expo user adding ONNX Runtime to their app had to:

- Add `onnxruntime-react-native` to `app.json` plugins (handles native linking only)
- Manually figure out that Metro's `assetExts` config doesn't guarantee native bundle inclusion
- Write a custom Expo config plugin to copy `.onnx` files into the Xcode project's "Copy Bundle Resources" phase
- Separately handle Android by copying models into `android/app/src/main/assets/`
- Debug a generic "Can't load model" error with no indication that the file simply isn't in the native bundle
- Discover through trial and error that dev client works but production EAS builds don't

After this PR, the same user:

1. Adds models to their existing plugin config:

```json
   ["onnxruntime-react-native", { "models": ["./assets/models/my_model.onnx"] }]
```

2. Runs expo prebuild and both iOS and Android are handled automatically
3. Gets a warning at prebuild time if their `metro.config.js` is missing `onnx`/`ort` in `assetExts`
4. Gets a warning if a model path doesn't exist (with the resolved absolute path for debugging)
5. Gets a warning if two model paths share the same basename (which would collide in the flat native assets directory)

The plugin validates early and warns loudly, so users hit errors at prebuild time rather than at runtime in production.

### Related Issues

This should help with the following issues, which appear to stem from the same underlying asset bundling gap:

- #26738  `onnxruntime-react-native` fails to load bundled ONNX assets in Expo standalone iOS
- #27062 Fails to load model in Expo 54 standalone iOS
- #17017  Failed to load model on real iPhone (works on simulator)
- #17773 React Native Expo app fails on import

Note: I haven't been able to reproduce every one of these individually, so I can't confirm this PR resolves all of them. But the common denominator is that models are working in dev but failing in production/standalone builds. This is consistent with the Metro vs native bundle gap that this PR addresses.

### DIFF
--- a/js/react_native/README.md
+++ b/js/react_native/README.md
@@ -89,7 +89,9 @@ async function resolveModelPath() {
     if (info.exists) return bundlePath;
   }
 
-  // Fallback: Metro asset system (dev server / simulator)
+  // Fallback: Metro asset system (dev server / simulator).
+  // On Android this works in both dev and production — the plugin copies
+  // models to assets/ which is where the Android native code already looks.
   const [asset] = await Asset.loadAsync(
     require("./assets/models/my_model.onnx")
   );


nehatkkhan786:
sir is this working now. i badly want to run this in my expo sdk 54 project.

PR #27804, opened 2026-03-23
[QNN-EP] Fix use-after-free of logger object

qc-tbhardwa:
### Description
Update logger object in QnnBackendManager::SetupBackend.

### Motivation and Context
While generating weight sharing context binary, Inference Session is created once for each graph. Inference session creates logger object and passes it to QnnBackendManager. QnnBackendManager stores this pointer in logger_ pointer and holds it long after Inference Session destroys Logger. On next Inference Session, another Logger object is created but QnnBackendManager do not use this as backend_setup_completed_ is already set, using this causes UAF.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
@@ -468,9 +468,7 @@ void QnnLogging(const char* format,
   }
 }
 
-Status QnnBackendManager::InitializeQnnLog(const logging::Logger& logger) {
-  logger_ = &logger;
-
+Status QnnBackendManager::InitializeQnnLog() {
   // Set Qnn log level align with Ort log level
   auto ort_log_level = logger_->GetSeverity();
   QnnLog_Level_t qnn_log_level = MapOrtSeverityToQNNLogLevel(ort_log_level);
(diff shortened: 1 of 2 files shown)


tirupath-qti:
@edgchen1 
This is a critical bug fix. Could you please help to review and trigger CI on this..

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1564,6 +1564,8 @@ Status QnnBackendManager::SetupBackend(const logging::Logger& logger,
                                        std::unordered_map<std::string, std::unique_ptr<std::vector<std::string>>>& context_bin_map,
                                        bool enable_htp_extended_udma_mode) {
   std::lock_guard<std::recursive_mutex> lock(logger_recursive_mutex_);
+  if (logger_ != &logger)
+    logger_ = &logger;
```
edgchen1, line 1566: `QnnBackendManager::InitializeQnnLog()` assigns to `logger_` and also does some additional logic - is that important here? also, it seems cleaner to have a single place that sets `logger_`.

qc-tbhardwa: I have updated code to assign `logger_` in only one place. Other logic in `QnnBackendManager::InitializeQnnLog()` does not need to be executed again.

edgchen1: ok, then I think `InitializeQnnLog()` no longer needs to have a `logger` parameter.

qc-tbhardwa:

PR #27811, opened 2026-03-23
Add dynamic dim select

TedThemistokleous:
### Description

Changes to let user select which dimension is dynamic in a model and leverage pad/slice logic so that to MIGraphX the run appears to be static but the specified dimensions are sliced out from the max size of the selected dynamic input

ORT_MIGRAPHX_DYNAMIC_DIMENSIONS_INDEX and ORT_MIGRAPHX_MAX_DYNAMIC_DIM_SIZE are used so that these values specify a non batch dimension and size that we'll also use pad/slice logic for. 

### Motivation and Context

Helps with customer models that have multiple batch and another dimension that are dynamic, such as sequence length. or other fields. Should allow us to leverage existing logic.


TedThemistokleous:
Not needed here.

PR #27815, opened 2026-03-23
Fix overflow in DmlGraphFusionHelper::ProcessInputData

adrastogi:
### Description

This change tries to address a problem in the DML EP where AlignToPow2 rounded up tensorByteSize to a 4-byte boundary before the data was read from the source buffer. This caused CreateCpuResource, CreateResource,  WriteToFile, and the inputRawData vector construction to read 1–3 bytes past the end of the original tensor data.

CreateResource and CreateCpuResource already independently align the D3D12 resource descriptor size, so they work correctly with the original (unaligned) byte count. The fix is to move the alignment to the location where it's needed. 

### Motivation and Context

This is required because it addresses a crash / incorrect behavior in the DML EP.

### DIFF
--- a/onnxruntime/test/providers/cpu/tensor/quantize_linear_test.cc
+++ b/onnxruntime/test/providers/cpu/tensor/quantize_linear_test.cc
@@ -99,8 +99,7 @@ TEST(DequantizeLinearOpTest, Int8_NonAlignedSize_Initializer) {
   test.AddInput<int8_t>("x_zero_point", {1}, {zero_point}, /*is_initializer=*/true);
   test.AddOutput<float>("y", {kNumElements}, y_expected);
 
-  test.Run(OpTester::ExpectResult::kExpectSuccess, "",
-           {kTensorrtExecutionProvider, kWebGpuExecutionProvider});
+  test.Run(OpTester::ExpectResult::kExpectSuccess, "", {kTensorrtExecutionProvider});
 }
 
 // scalar zero & scale with int4


onnxruntime/test/providers/cpu/tensor/quantize_linear_test.cc
```diff
@@ -74,6 +74,35 @@ TEST(DequantizeLinearOpTest, Int4_LargeInitializerInput) {
...
+
+  // Mark all inputs as initializers so they go through DML's ProcessInputData
+  // → UnpackInitializer → AlignToPow2 code path during graph fusion.
+  test.AddInput<int8_t>("x", {kNumElements}, x_data, /*is_initializer=*/true);
+  test.AddInput<float>("x_scale", {1}, {scale}, /*is_initializer=*/true);
+  test.AddInput<int8_t>("x_zero_point", {1}, {zero_point}, /*is_initializer=*/true);
+  test.AddOutput<float>("y", {kNumElements}, y_expected);
+
+  test.Run(OpTester::ExpectResult::kExpectSuccess, "",
+           {kTensorrtExecutionProvider, kWebGpuExecutionProvider});
```
fdwr: (resolve me) Unrelated to this PR, but just curious why the WebGPU EP fails for this simple case?

fdwr (approved):
👍

FYI @RafaelCintron

onnxruntime/test/providers/cpu/tensor/quantize_linear_test.cc
adrastogi: Oops, that was a copy-paste error from the test case right above it (picked up their exclusion list).

adrastogi: Removed the WebGPU exclusion and will have CI verify this.

adrastogi: WebGPU CI-related runs are green, marking this as resolved.

